PR #4, opened 2014-12-21
Clarification for an example

estelle:
However, the span element is not in fact broken in two for the purposes of styling.

If the span had a border or background image, and the browser acted as if the element were closed and reopened, a no-repeat background image would be displayed twice. or, if a border were present, the border would be closed and reopened.


fantasai:
Hm, so it's a little tricky: styling *is* affected because inheritance is affected. But you're right that the box properties don't behave as if there were two elements.
The whole section needs a rewrite, honestly. Not sure what to do with the pull request given that. :/

PR #13, opened 2015-03-12
[css-text-decor] Add trailing-spaces to text-decoration-skip

frivoal:


fantasai:
CSS Text Decoration L3 is in CR. This should go into L4.

PR #14, opened 2015-03-22
[css3-ui] Clean up minor leftover from nav-index removal

frivoal:


frivoal:
Equivalent commit was applied independently. No need for this pull request any longer.

PR #17, opened 2015-04-09
[css-ui] Fix all current bikeshed warnings

frivoal:


tantek:
Looks good except for the changes:

from: <a spec="css-style-attr">style attribute</a> 
to: <a href="http://dev.w3.org/csswg/css-style-attr/#style-attribute">style attribute</a> 

What is the reason NOT to use spec="css-style-attr"?
And we should not be directly linking to dev.w3.org draft URLs as references (will block us from advancing to CR).

plinss:
Rather than using absolute URLs to avoid Bikeshed warnings, you can update the linking defaults to select the proper specs:
https://github.com/tabatkins/bikeshed/blob/master/docs/definitions-autolinks.md#configuring-linking-defaults

Then Bikeshed should choose the /TR versions of the specs when generating for /TR.

frivoal:
The problem with style attribute is that the the css-style-attr spec defines the term multiple times, so if we let bikeshed auto-link, even if we point it to the right spec, it will complain that it's ambiguous. The bug is in css-style-attr, not in css-ui, but I wanted to work around it to get to 0 warnings.

Since it is a REC already (and therefore is unlikely to change any time soon), I'll just link to TR instead.

frivoal:
Link updated.

tabatkins:
If it defines the term multiple times, it's definitely a bug, but also it's probably okay for any of them to be the link target.  To just shut Bikeshed up about the error, put it in an "Ignored Terms" metadata; that way it'll silently choose the first dfn in its database without emitting a warning.

frivoal:
I thought "Ignored Terms" meant it wouldn't link at all. At least that's what the here doc claims:
https://github.com/tabatkins/bikeshed/blob/master/docs/metadata.md

tabatkins:
Whoops, ignore that documentation while I correct it.  All `Ignored Terms` does is silence the warnings/errors from autolinking that term.

tantek:
Hard-linking to the w3.org/TR drafts is a reasonable path forward for those references.

tantek:
Hard-linking to the w3.org/TR drafts is a reasonable path forward for those references.

PR #18, opened 2015-04-10
[css3-ui] Explicit file format requirements for 'cursor'

frivoal:
This commit introduces the file format requirements for the cursor property, as recently resolved.


css-ui/Overview.bs
```diff
@@ -985,6 +985,36 @@ If the user agent cannot handle any user-defined cursor,
...
+	<li>SVG, as defined in [[!SVG]], in <a href="http://www.w3.org/TR/svg-integration/#secure-animated-mode">secure animated mode</a> [[!SVG-INTEGRATION]]
+	<li>any other animated image file format that they support
+	for <<image>> in other properties,
+	such as the the 'background-image' property
+</ul>
+
+The UA may also support additional file formats.
+
+Note: At the time of writing this specification,
+the only file formats supported for cursors in all browsers are
```
tantek: This looks good except "in all browsers" is both something we can't reasonably claim, and dated text - that is, reflects a statement about the present, in a spec that will be read in the future. We should say something like "in common 2015 desktop browsers".

frivoal:
Updated the prose to according to Tantek's review.

PR #20, opened 2015-04-14
animations override !important

estelle:
My understanding is that animation property values are more important than non-animation property values, even with !important
Changed As defined in [CSS3CASCADE], animations override all normal rules, but are overridden by !important rules.
to remove the overridden statement, as all modern browsers except FF support this, and conversations in #css IRC indicate that all browsers with the exception of FF have implemented as intended.

http://codepen.io/estelle/pen/iDvBz shows green border in all browsers (IE11, Safari, Chrome and Opera) except Firefox.


tabatkins:
Cascade defines different behavior, and was decided on intentionally.  Browsers just haven't caught up yet.

PR #26, opened 2015-06-04
Florian/text decor skip

frivoal:
This implements the resolution recorded here:
https://lists.w3.org/Archives/Public/www-style/2015May/0314.html


fantasai:
Looks fine. Fix "trainling" and add an issue on whether the initial value should be easier to respecify and go ahead and merge once that's done.

PR #27, opened 2015-06-10
Florian/pre wrap

frivoal:
These are the level 3 changes for the resolutions made at the New York F2F regarding pre-wrap, and recorded here:
https://lists.w3.org/Archives/Public/www-style/2015May/0281.html


kojiishi:
lgtm

frivoal:
Thanks

PR #28, opened 2015-06-23
[css-align] Fix typo in Baseline Self-Alignment definition for Grids.

javifernandez:


atanassov:
A grid item is aligned as part of its containing grid content-alignment. The current spec text is written from the  POV of grid, thus this change isn't necessary.

tabatkins:
No, grid items align themselves within their grid area.  I think this is a correctly-identified typo.

PR #33, opened 2015-07-01
Florian/contain and clip

frivoal:
Fixing paint containment to invoke overflow:clip, as per https://lists.w3.org/Archives/Public/www-style/2015Apr/0405.html

Doing as a pull request since this touches not only the overflow spec, which I co-edit, but also the containment spec, which I don't.

### DIFF
--- a/css-containment/Overview.bs
+++ b/css-containment/Overview.bs
@@ -197,9 +197,8 @@ Paint Containment</h3>
 
 	Giving an element <dfn export>paint containment</dfn> has the following effects:
 
-	1. As specified in [[!CSS-OVERFLOW-3]],
-		the computed value the 'overflow-x' and 'overflow-y' is changed to ''overflow/clip''
-		if it would otherwise have been ''overflow/visible''.
+	1. If the computed value of overflow-x or overflow-y would otherwise have been ''overflow/visible'',
+		it must instead compute to ''overflow/clip''. [[!CSS-OVERFLOW-3]]
 		<span class=note>This means that regardless of the specified value of 'overflow',
 		the contents of the element is clipped to the element's content box,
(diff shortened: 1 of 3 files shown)


css-containment/Overview.bs
```diff
@@ -197,8 +197,12 @@ Paint Containment</h3>
 
 	Giving an element <dfn export>paint containment</dfn> has the following effects:
 
-	1. The contents of the element must be clipped to the element's content box.
-		This includes both the paint of the descendants and their geometry (similar to overflow).
+	1. As specified in [[!CSS-OVERFLOW-3]],
+		the computed value the 'overflow-x' and 'overflow-y' is changed to ''overflow/clip''
+		if it would otherwise have been ''overflow/visible''.
```
tabatkins: I prefer "If the computed value of overflow-x or overflow-y would otherwise have been 'visible', it must instead compute to 'clip'. [[!CSS-OVERFLOW-3]]"

css-containment/Overview.bs
```diff
@@ -208,7 +212,8 @@ Paint Containment</h3>
 		the UA can directly skip trying to paint its contents,
 		as they're guaranteed to be off-screen/obscured as well.
 
-	2. If the containing element is ''overflow:visible'',
+	2. If the containing element would have been ''overflow:visible'',
+		it becomes ''overflow:clip'' and
```
tabatkins: This line sounds like normative instructions; I'd prefer a parenthetical like (but is now overflow:clip due to paint containment), or something more direct, like "If the containing element's overflow has been changed to 'clip' (because it was originally visible), ..."

frivoal:
Fixed according to the review, merging.

PR #34, opened 2015-07-07
[css-containment] Fix layout containment and fragmentainer interaction

frivoal:
Fix layout containment to work better with Regions and other fragmenting things.

Discussion in this mail thread:
https://lists.w3.org/Archives/Public/www-style/2015Jul/0078.html

### DIFF
--- a/css-containment/Overview.bs
+++ b/css-containment/Overview.bs
@@ -153,12 +153,12 @@ Layout Containment</h3>
 		then they are not generated.
 		If they would exist regardless,
 		they remain part of the <a>fragmentation context</a>,
-		but do no receive any content from the <a>fragmented flow</a>.
+		but do not receive any content from the <a>fragmented flow</a>.
 
 		Specifically:
 			- <a spec=css-regions>CSS Regions</a> following the one which traps the content
 				are still considered part of the <a spec=css-regions>region chain</a>
-				as returned by the {{getRegions()}} method
+				as returned by the {{NamedFlow/getRegions()}} method
 				of the {{NamedFlow}} interface.
 			- the {{regionOverset}} attribute of the {{Region}} interface
 				of the region which traps the content
--- a/css-containment/Overview.bs
+++ b/css-containment/Overview.bs
@@ -138,6 +138,37 @@ Layout Containment</h3>
 
 	2. The element must be a <a>formatting context</a>.
 
+	3. If a <a>fragmentation context</a> participates in <a>layout containment</a>,
+		the first element with <a>layout containment</a>
+		affecting the <a>fragmentation context</a>
+		must “trap” the remainder of the <a>fragmented flow</a>.
+		<a>Fragmentation</a> must not continue past the <a>layout containment</a> boundary,
+		and the last <a>fragmentation container</a>
+		within the first <a>layout containment</a> boundary
+		is treated as if it is the last <a>fragmentation container</a>
+		in its <a>fragmentation context</a>.
+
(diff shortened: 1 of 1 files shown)


css-containment/Overview.bs
```diff
@@ -138,6 +138,37 @@ Layout Containment</h3>
...
+		within the first <a>layout containment</a> boundary
+		is treated as if it is the last <a>fragmentation container</a>
+		in its <a>fragmentation context</a>.
+
+		If subsequent <a>fragmentation containers</a> in the <a>fragmentation context</a>
+		are only generated when more content remains in the <a>fragmented flow</a>,
+		then they are not generated.
+		If they would exist regardless,
+		they remain part of the <a>fragmentation context</a>,
+		but do no receive any content from the <a>fragmented flow</a>.
```
tabatkins: "do not"

css-containment/Overview.bs
```diff
@@ -138,6 +138,37 @@ Layout Containment</h3>
...
+
+		Specifically:
+			- <a spec=css-regions>CSS Regions</a> following the one which traps the content
```
tabatkins: Do you actually need the spec='' attribute here? It looks like there's only two refs, one from the TR and one from the ED of the Regions spec.

css-containment/Overview.bs
```diff
@@ -138,6 +138,37 @@ Layout Containment</h3>
...
+				are still considered part of the <a spec=css-regions>region chain</a>
+				as returned by the {{getRegions()}} method
```
tabatkins: Just to make future problems less likely, better to qualify all attribute/method references with the interface they're from, like `{{NamedFlow/getRegions()}}`.

css-containment/Overview.bs
frivoal: Bikeshed complains if you don't. I think they're not exported.

tabatkins: Well, you know what to do. (Export those terms! You can fix the database yourself!)

frivoal:
Pushed a new commit to address the review comments.

css-containment/Overview.bs
frivoal: Maybe I can, but I don't know how to. Where is TFM for that, so that I can R it?

tabatkins: https://github.com/tabatkins/bikeshed/blob/master/docs/definitions-autolinks.md#exporting-definitions

Shorter answer: Add an `export` boolean attribute to all your "dfn" type definitions (the ones without an explicit type otherwise) that you want other specs to be able to link to.  (That is, unless they're an internal definition you just have for use within the spec itself, export it.)

frivoal: Oops, I actually knew that one, but assumed based on file extension .src.html that css-regions was not under bikeshed, and then I didn't know. But it is, so I'll fix this. How often is the database refreshed once I've done this?

plinss: The linking database is managed by Shepherd, and it parses all the CSS specs, not just those produced by Bikeshed (though the other specs may or may not have their <dfn>s marked up to be recognized properly). The spec database for drafts is updated within minutes after a push. Specs on /TR get updated at least once per day (midnight pacific time).

frivoal:
Fixed regions (https://github.com/w3c/csswg-drafts/pull/35), so I removed the explicit spec=.

css-containment/Overview.bs
frivoal: Good to know the update is a fast one. As for figuring out what to do on non bikeshed specs... Let's just bikeshed them all.

PR #35, opened 2015-07-07
Florian/regions bikeshed

frivoal:
Bikeshed clean up, also removing stuff that no longer exists, and exporting key definitions.


astearns:
Thanks, Florian.

PR #41, opened 2015-07-24
Non normative edits.

estelle:
Three separate commits. If you prefer separate pull requests instead, please let me know.

1) negative delays impact animation duration
2) animation iteration can be 0 or greater.
3) paragraph sentences reordered for clarity.

### DIFF
--- a/css-animations/Overview.html
+++ b/css-animations/Overview.html
@@ -417,7 +417,7 @@ Animations</span><a class="self-link" href="#animations"></a></h2>
 
    <p>The end of the animation is defined by the combination of the
 	<a class="property" data-link-type="propdesc" href="#propdef-animation-duration">animation-duration</a>, <a class="property" data-link-type="propdesc" href="#propdef-animation-iteration-count">animation-iteration-count</a> and <a class="property" data-link-type="propdesc" href="#propdef-animation-fill-mode">animation-fill-mode</a>
-	properties.</p>
+	properties, and <a class="property" data-link-type="propdesc" href="#propdef-animation-delay">animation-delay</a>, if negative.</p>
 
 	
    <div class="example" id="example-dee38741"><a class="self-link" href="#example-dee38741"></a>
--- a/css-animations/Overview.html
+++ b/css-animations/Overview.html
@@ -571,10 +571,10 @@ Keyframes</span><a class="self-link" href="#keyframes"></a></h2>
 	from more than one @keyframes rule with the same selector value.</p>
 
 
-   <p>If a property is not specified for a keyframe, or is specified but invalid, the animation of that
-	property proceeds as if that keyframe did not exist. Conceptually, it is as if a set of keyframes is
-	constructed for each property that is present in any of the keyframes, and an animation is run
-	independently for each property.</p>
+   <p>Conceptually, it is as if a set of keyframes is
+  constructed for each property that is present in any of the keyframes, and an animation is run
+  independently for each property. If a property is not specified for a keyframe, or is specified but invalid, the animation of that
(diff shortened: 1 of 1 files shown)


tabatkins:
Thanks for the commits, but you're editting the generated file, not the source file, so it'll get wiped out the moment the repository auto-regens the spec. ^_^

dbaron:
A few comments on the actual edits:

1) I'm not sure what this sentence even means.  But it's not clear to me why it would be affected by only negative delays and not positive ones.  I think I might prefer removing the sentence, but should probably look more closely.

2) looks great

3) I think the sentences are really mostly unrelated.  It might be better if they were 2 separate paragraphs.  But I think there's also something missing here about how properties in one keyframe rule override another.  That should have been written for the edit under "2013-02-20 minutes", but I don't think it was.

estelle:
When I resubmitted the PR to the correct file I left off 1) as the sentence seemed to read that the timing of the animationend had to do with those properties, but on re-read of your comment and the text, it doesn’t actually seem to mean anything.
In terms of 3)  what does that paragraph mean? When we animate multiple properties, in the end it's as if a separate animation was created for each property. That is why switching the animation-timing-function during a keyframe only impacts the property/value pairs within that keyframe.

-Estelle 

Estelle Weyl
estelle@weyl.org
http://www.standardista.com





On Jul 25, 2015, at 11:17 AM, David Baron <notifications@github.com> wrote:

A few comments on the actual edits:

1) I'm not sure what this sentence even means. But it's not clear to me why it would be affected by only negative delays and not positive ones. I think I might prefer removing the sentence, but should probably look more closely.

2) looks great

3) I think the sentences are really mostly unrelated. It might be better if they were 2 separate paragraphs. But I think there's also something missing here about how properties in one keyframe rule override another. That should have been written for the edit under "2013-02-20 minutes", but I don't think it was.

dbaron:
Actually, rereading (3) again, I think the order it's in does make sense, although it could certainly be clearer.

What it's trying to say is that if you have:
0% { top: 0; left: 0 }
50% { top: 50px }
100% { top: 100px; left: 100px }
and you're using an 'ease-in' timing function, top animates over two separate uses of that ease-in timing function, wherease left animates only over a single easing function.  There's no attempt to infer some value for left at 50% and act as though there are two pieces to the left animation.  (And likewise, if the timing functions were specified in the keyframes, they'd only matter for a property if that property is in a keyframe at that percentage.)

estelle:
Where it reads “property proceeds as if that keyframe did not exist”, makes it sound like the full keyframe gets obliterated, even the valid property/value pairs. The next paragraph clarifies that, so inverting the paragraphs makes the 1st sentence make more sense.
-Estelle 






On Jul 25, 2015, at 11:44 AM, David Baron <notifications@github.com> wrote:

Actually, rereading (3) again, I think the order it's in does make sense, although it could certainly be clearer.

What it's trying to say is that if you have:
0% { top: 0; left: 0 }
50% { top: 50px }
100% { top: 100px; left: 100px }
and you're using an 'ease-in' timing function, top animates over two separate uses of that ease-in timing function, wherease left animates only over a single easing function. There's no attempt to infer some value for left at 50% and act as though there are two pieces to the left animation. (And likewise, if the timing functions were specified in the keyframes, they'd only matter for a property if that property is in a keyframe at that percentage.)

birtles:
Re (1), I agree with [dbaron's comment](https://github.com/w3c/csswg-drafts/pull/41#issuecomment-124762801) and will remove that sentence altogether.

birtles:
Re (3), I agree that the sentence, as it stands, makes more sense than inverting the order. However, I hope we can make this more clear when we address issue #71.

PR #42, opened 2015-07-24
Estelle

estelle:
edit to overview.bs. Edits to overview.html should be automagically overwritten.

1) The animation may iterate less than one time. 0.25 and 0 are valid iteration counts
2) reordered the two sentences because in the original order it made it sound like keyframes, not property values, are ignored.
What does that paragraph mean? When we animate multiple properties, in the end it's as if a separate animation was created for each property. That is why switching the animation-timing-function during a keyframe only impacts the property/value pairs within that keyframe.


dbaron:
This is a continuation from #41, which has some prior discussion.

grorg:
Seems to have been incorporated.

PR #49, opened 2015-09-11
Florian/pre wrap auto trim

frivoal:
This adds text-space-collapse: preserve-auto and text-space-collapse: preserve-trim to css-text-4, and explains white-space: pre-wrap-auto in terms of the former.

Also includes a small bug fix.

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -109,7 +109,7 @@ White Space Trimming: the 'text-space-trim' property</h3>
 	<pre class="propdef">
 	Name: text-space-trim
 	Value: none | trim-inner || discard-before || discard-after
-	Initial: collapse
+	Initial: none
 	Applies to: all elements
 	Inherited: no
 	Percentages: n/a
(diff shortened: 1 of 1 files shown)


fantasai:
preserve-spaces-auto should refer explicitly to system conventions, since that's the whole point of it. It's not UA-defined. It's OS-defined.

(Also, preserve-trim should note that it's at-risk, and we're not sure if it's needed. We need use cases, and they should help clarify if both leading and trailing spaces should be trimmed.)

frivoal:
> preserve-spaces-auto should refer explicitly to system conventions, since that's the whole point of it. It's not UA-defined. It's OS-defined.

Not sure. IE's behavior and Safari's behavior are arguably following OS standards. But Chrome does the same cross platform. And we were not trying to force them to change. I think adding a "Should follow platform convensions" would be good, since indeed that's a large part of the intent, but I'd like to keep it as a should. Sounds OK?

> preserve-trim should note that it's at-risk
Will do.

frivoal:
@fantasai Is this better?

PR #50, opened 2015-09-18
[css-2015] Clarify why unstable features should be released both prefixed and unprefixed

frivoal:


frivoal:
Actually, I'm a co-editor of this document, so I'll merge this myself.

PR #52, opened 2015-09-21
Adding a description to address the case of a Grooved black border

Adenilson:
Adding a description to address the case of a Grooved black border making explicit what is the expected behavior of a UA.

For whole discussion, please see:
https://lists.w3.org/Archives/Public/www-style/2015Sep/0106.html

### DIFF
--- a/css-backgrounds/Overview.html
+++ b/css-backgrounds/Overview.html
@@ -2380,7 +2380,8 @@ background-color:      white;
    <dd> Looks as if it were carved in the canvas. (This is typically achieved
     by creating a “shadow” from two colors that are slightly lighter and
     darker than the ‘<a href="#border-color"><code
-    class=property>border-color</code></a>’.)
+    class=property>border-color</code></a>’. In case the border color
+    is black, a shade of gray must be used as the lighter color.)
 
    <dt><dfn id=ridge>‘<code class=css>ridge</code>’</dfn>
 


tabatkins:
As stated in IRC by @SimonSapin, explicitly calling out "black" isn't correct here.  Near-black colors like `#000001` have the same problems, and so do colors near white.  

What we want to do is check if the difference between the "lighter" and "darker" color is sufficiently large; when the start color is near the min/max of lightness, it's not, so we'll have to shift the colors a bit.

tabatkins:
Let's pretend for a moment that the general algorithm is "represented as HSL, the two colors are the base color, but with lightness +- 10%". (That is, 10 added/subtracted from the lightness, not multiplied by 1.1/.9.)

Then the spec text we want is something like:

> Let *base color* be the used value of 'border-color' as an HSL color.  If base color's lightness is between 10% and 90%, then *light color* and *dark color* are colors with the same hue and saturation as base color, but lightness 10% higher or lower, respectively, than base color.
> 
> Otherwise, if base color's lightness is less than 10%, dark color is black, and light color is a color with the same hue and saturation as base color, but lightness set to 20%.
> 
> Otherwise, if base color's lightness is greater than 90%, light color is white, and dark color is a color with the same hue and saturation as base color, but lightness set to 80%.

Tweak the numbers as appropriate for whatever the real light/dark adjuster is.

SimonSapin:
:+1:

Adenilson:
Sounds good!

fantasai:
Closing this request without merging, since a) it's not editing the correct file and b) the statement being added is unambiguously implied by the previous statement anyway. If we want to add an algorithm, that gets filed against Level 4...

PR #58, opened 2015-10-09
Fixes pre-wrap, pre-wrap-auto and preserve-auto

frivoal:
* (Tries to) Improve the prose describing pre-wrap-auto / preserve-auto as discussed at the last telco,
* fixes the normative behavior of pre-wrap-auto and preserve-auto. 

Overview.html is not updated (and should be updated when merging) because bikeshed is currently broken (https://github.com/tabatkins/bikeshed/issues/501)


frivoal:
In Sydney, the CSSWG has resolved in favor of an alternative solution, making this obsolete.

PR #59, opened 2015-10-14
Adding periods at the end figure's descriptions.

Adenilson:
Other specs (e.g. http://www.w3.org/TR/css3-background/#the-border-radius) include a period at the end of text describing figures. This patch includes the missing periods for all figures.


Adenilson:
Will upload a new merge request later after the other changes land.

PR #61, opened 2015-10-15
Rephrasing note about shape diversity.

Adenilson:
Changing from 'If the shapes of displays are various..' to 'For other shapes...' as it reads more naturally in English.


frivoal:
Oops, just noticed that you were fixing the .html file, not the .bs file. Can you file another PR for that?

Adenilson:
Sure thing, please check:
https://github.com/w3c/csswg-drafts/pull/63

PR #81, opened 2015-12-03
[cssom-view] Add more positioning control to scrollIntoView

zcorpan:
* Add "center" to ScrollLogicalPosition.
* Add inline dictionary member to ScrollIntoViewOptions.

Ref. https://www.w3.org/Bugs/Public/show_bug.cgi?id=17152
Ref. https://lists.w3.org/Archives/Public/www-style/2015Dec/0050.html


zcorpan:
https://log.csswg.org/irc.w3.org/css/2015-12-07/#e627335

PR #91, opened 2016-01-22
Point out case-insensitivity of i in [foo=bar i]

SimonSapin:
https://lists.w3.org/Archives/Public/www-style/2016Jan/0166.html shows that it’s easy to miss.

How does this look, @tabatkins?


SimonSapin:
Test change: https://github.com/w3c/web-platform-tests/pull/2516

zcorpan:
LGTM

PR #101, opened 2016-03-28
Removal of the rests of 'grid-template' references

SelenIT:
The `grid-template` shorthand property has been removed from the spec, but two references to it remained. Also, this change was not reflected in the 'Changes' section. Shouldn't we fix this?


tabatkins:
Thanks!

PR #102, opened 2016-04-02
Counter reset on ol too broad in scope

FearGoidte:
Reseting counters on the “ol” element means that an ordered list which
is the descendent of a sibling element of a previous ordered list will
behave as a continuation of the previous list.

### DIFF
--- a/css2/generate.html
+++ b/css2/generate.html
@@ -647,7 +647,7 @@ result is very similar to that of setting 'display:list-item' and
 'list-style: inside' on the LI element:
 
 <pre>
-OL { counter-reset: item }
+OL>LI:first-child { counter-reset: item }
 LI { display: block }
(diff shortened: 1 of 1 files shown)


tabatkins:
Just so you don't think this is hanging, I started a thread in the CSSWG mailing list about possibly fixing this *properly*. <https://lists.w3.org/Archives/Public/www-style/2016Apr/0364.html>

FearGoidte:
Thanks for the update. I find that part of the recommendation quite impenetrable so it took me some time to determine that the behaviour was correct but probably unintended.

Kind regards

--
James

tabatkins:
Fixed in Lists 3 by <https://github.com/w3c/csswg-drafts/issues/5477>

PR #104, opened 2016-04-17
Update default-TR.css

panyiyi:


dbaron:
Is there a reason to think the webmaster will allow publication of TR drafts without this line?

tabatkins:
Closing this request - our stylesheet isn't up for public editting.

(Also, the base.css that's used on TR drafts now has a readable max-width.)

FremyCompany:
It would be great though if the repo could use a stylesheet that looks more like the final one. Maybe it's just me but local builds of specs don't feature the new table of contents. Or is it because I should update bikeshed?

tabatkins:
They... should?  I need to do some fixup to our default.css, but it imports the W3C's base.css right now, which means it should look like public ones.  They certainly do on my computer, at least.

PR #110, opened 2016-05-12
[css-grid] Make placing examples follow `<integer> && <custom-ident>?`

tomalec:
I found out that Examples 33 and 37 are incompatible with `<grid-line>` definition at https://drafts.csswg.org/css-grid/#line-placement
```
<grid-line> =
  auto |
  <custom-ident> |
  [ <integer> && <custom-ident>? ] |
  [ span && [ <integer> || <custom-ident> ] ]
```
As they use:
```
grid-row: text 5 / text 7;
```
So more like `<custom-ident>? <integer>`

This PR updates examples to:
```
grid-row: 5 text / 7 text;
```

### DIFF
--- a/css-grid/Overview.bs
+++ b/css-grid/Overview.bs
@@ -2594,9 +2594,9 @@ Named Lines and Spans</h4>
 	<div class='example'>
 		<pre>
 			.six {
-				grid-row: text 5 / text 7;
+				grid-row: 5 text / 7 text;
 				/* Span between the 5th and 7th lines named "text". */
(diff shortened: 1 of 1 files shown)


tabatkins:
The `&&` in `[ <integer> && <custom-ident>? ]` indicates that the two branches must occur, but can do so in any order, so the current examples are perfectly correct (and so are yours).

I just realized that grammar block wasn't marked up properly to give you the convenient linking from syntax operators to definitions, so that's fixed now.

PR #111, opened 2016-05-13
[css-text-3] break spaces

frivoal:


css-text/Overview.bs
```diff
@@ -755,6 +755,9 @@ Languages and Typesetting</h4>
          the UA must either <a>hang</a> the <a>white space</a>
          or visually collapse their character advance widths
          such that they don't take up space in the line.
+         However, if 'overflow-wrap' is set to ''break-spaces'',
```
kojiishi: When ''break-spaces'', spaces breaks, so it no longer be "at the end of a line". I think I understand your intention but it can confuse rather than help. Maybe you could add such note that this will not happen if ''break-spaces'' if you think having text here is important?

css-text/Overview.bs
```diff
@@ -1296,7 +1299,7 @@ Line Breaking Details</h3>
 
 <pre class='propdef'>
 Name: overflow-wrap, word-wrap
-Value: normal | break-word
+Value: normal | break-word | break-spaces
```
kojiishi: thought this was for L4, but I admit that I wasn't paying much attention to which level we resolved on. Have a link to the resolution?

frivoal: I don't think the resolution (SF or Syndey) specified if it was level 3 or 4. My main reason for preferring having it in L3 is:

* This value is new, but the behavior is already present in some browsers, triggered by a different property/value: Firefox and Safari wrap spaces like this when word-break is break-all (FF all preserved spaces, Safari only nbsp) (see http://output.jsbin.com/degaqe). Since we have clarified that the break-word property is not supposed to do that and that this is supposed to be triggered by overflow-wrap:break-spaces, rather than have them remove this behavior now and put it back later under a different name, I thought it made sense to add the switch in the current Level.

Also from a spec editing point of view, L3 is easier (but I agree this is not an important reason):

* L4 does not yet have the equivalent of section 4 (#white-space-processing), and that this makes some modification to it (section 4.1.3 point 4).

css-text/Overview.bs
frivoal: When 'break-spaces' is specified, spaces should break. But this would not happen if they have already been collapsed, so I am trying to (normatively) make it explicit that you cannot do that. I thought this text was easier to understand here than in the definition of the break-spaces value. Do you think it should be moved? Do you have a suggestion of a better way to phrase it?

css-text/Overview.bs
kojiishi: Totally understandable, and I don't have preference either way, but when I want to follow the resolution, you and I having different understanding looks like a sign that it's better to double-check with WG, no? In my understanding, fantasai said L4 at the beginning of the discussion and that was applied to all the discussion, but now it's unclear to me too.

frivoal: Right. I've pinged the WG now, let's see what comes out of this.
https://lists.w3.org/Archives/Public/www-style/2016May/0134.html

frivoal: RESOLVED: Add overflow-wrap: break-spaces to level 3
http://logs.csswg.org/irc.w3.org/css/2016-05-25/#e692109

frivoal: @kojiishi ping. Do you want me to adjust something in the pull request before you can merge it, or is it fine like this?

frivoal:
@kojiishi @fantasai Ping. Can you either merge this or tell me what's wrong with it?

css-text/Overview.bs
kojiishi: By re-reading this, you're modifying an item that I don't understand, so it's not your problem, but you'd have to ask fantasai to review.

This item starts with:
> If spaces or tabs at the end of a line are non-collapsible but have 'white-space' set to ''pre-wrap''

but I can't interpret this part. Does this "but" mean "except"? Maybe clear to native speakers but this is beyond my English capability, sorry.

css-text/Overview.bs
```diff
@@ -1324,6 +1327,11 @@ Canonical order: N/A
       Shaping characters are still shaped as if the word were not
       broken, and grapheme clusters must stay together as one unit.
       No hyphenation character is inserted at the break point.
+    <dt><dfn>break-spaces</dfn></dt>
+    <dd>Line breaks are handled identically to ''overflow-wrap/normal'',
+      except that any sequence of <a>preserved</a> white space
+      that would otherwise overflow the line under these rules
+      must be broken after the last white space character that would fit the line.
```
kojiishi: So when you say
> sequence of preserved white space...must be broken

you want to break between preserved white spaces but not before/after? Can you try to clarify?

frivoal: Would this be clearer? 

> Line breaks are handled identically to ''overflow-wrap/normal'',
> except that if a sequence of <a>preserved</a> white space
> would otherwise overflow the line under these rules,
> the line must be broken after the last white space character of that sequence
> that would fit the line.

This is meant to be the same behavior, but should be more explicit. Break in places unrelated to the white space sequence are allowed as usual; it allows the break to happen after the last character in the sequence. It does not create a break opportunity between the first white space character in the sequence and whatever comes before it, to avoid wrapping a single space to the next line (I think you were the first to suggest that that would be a bad idea).

kojiishi: As a member, I still don't understand why one would want this, but as an editor, I can review PRs according to resolutions. That makes me a good reviewer though. Since I don't understand how it's used, I can review how the text reads to implementors without any background ;-)

What will happen if there are no spaces that can fit?

frivoal: Sorry for the delayed answer.

If there is no spaces that would fit, you deal with it the way you would otherwise: break the line somewhere earlier if you can, or overflow if you cannot.

I don't actually have a strong opinion whether we should allow a break before the first space in a sequence of not, but I believe you were the one who expressed a preference for disallowing it.

css-text/Overview.bs
frivoal: I think this "but" could just as easily be replaced with "and". We're talking about a subset of of the non-collapsible spaces. "If there are non-collapsible spaces and when 'white-space' is set to ''pre-wrap''" is the same as "if there are non-collapsible spaces, but [only in the case when] 'white-space' set to ''pre-wrap''"

css-text/Overview.bs
kojiishi: Please make a call, as I said, I don't understand how it's used.

So from what I understand, when all of the following condition met:

* If `white-space: pre-wrap` and
* if a non-breakable word and at least one trailing space fit within the line, and
* If subsequent spaces follow, and
* If those subsequent spaces do not fit within the line

You want to break between spaces, but not before/after?

frivoal: Very sorry for the slow answer.

> You want to break between spaces

correct, if `overflow-wrap` is `break-spaces.` you break and wrap in the middle of the space sequence instead of letting it overflow.

> but not before

Not before the first space  if `overflow-wrap` is only `break-spaces.` If it is `overflow-wrap:break-word | break-spaces`, you can also break before the first space, since `overflow-wrap: break-word` allows for breaks anywhere.

I think this is the best behavior, but I am not very strongly attached to this nuance, so if the rest of the proposal is OK and this is blocking, I am open to changing this part.

> or after?

I don't think there's anything in the wording I proposed that prohibits breaking after spaces. If the end of the line falls earlier than the last space, you should break at that earlier point, but if it falls right after that space, breaking there is the normal thing to do.

---
2 questions:
 - Are you OK with the behavior?
 - Do you think the proposed text (a) correctly and (b) understandably describes this behavior?

frivoal:
@kojiishi @fantasai this PR has been open for pretty long now, and I'd like to wrap it up. In light of [my answer](https://github.com/w3c/csswg-drafts/pull/111#discussion_r74177445) to [Koji's question](https://github.com/w3c/csswg-drafts/pull/111#discussion_r68636764), can you either merge it, or tell me if you either want more clarification or some change to the proposed text?

If there is anything left to be tweaked, I'd like to make sure that we get it done before the end of TPAC, so if you'd like me to prepare something before then, let me know what.

kojiishi:
Sorry for my slow response too but the PR doesn't look to address a case when it says:

> after the last white space character that would fit the line.

it doesn't define the behavior when there's no white space that would fit the line, correct?

Also I prefer it be effective only when white space is preserved by the `white-space` property, rather than making the preservation automatically activated by the `overflow-wrap`, unless it was discussed and resolved (sorry if it did, in that case, I'm ok.)

frivoal:
> doesn't look to address a case when it says:
>>  after the last white space character that would fit the line.
>
> it doesn't define the behavior when there's no white space that would fit the line, correct?

Hmmm. I thought the sentence was clear, but having re-read it a couple times, I see what you mean. I've pushed a rewording with a couple of clarifying notes. I think this now defines all cases. Let me know if you think it doesn't.

> Also I prefer it be effective only when white space is preserved by the white-space property

Since this value only affects what happens in runs of preserved white space, in most cases, it will not do anything if the white-space property is not preserving the white spaces. The difference with what you are suggesting is that as currently specified, it would also call for breaks in overflowing runs of &amp;nbsp;. I'd rather keep it that way: I'd find it pretty counter intuitive if a value called break-spaces did not do that.

If you agree, great. If you don't (and there's no other issue left), can we get this merged, and open a separate issue to discussed this point?

css-text/Overview.bs
```diff
@@ -1137,6 +1142,30 @@ Line Breaking Details</h3>
       Shaping characters are still shaped as if the word were not
       broken, and grapheme clusters must stay together as one unit.
       No hyphenation character is inserted at the break point.
+    <dt><dfn>break-spaces</dfn></dt>
+    <dd>Line breaks are handled as usual,
+      except that any sequence of <a>preserved</a> white space
+      that would otherwise overflow the line and <a>hang</a> as per <a href="white-space-phase-2">Trimming and Positioning</a>
```
kojiishi: isn't "#" needed for "white-space-phase-2"?

kojiishi:
Thanks for the update, that part looks good, except one possible nit I put as an inline comment. Please check.

> If you agree, great. If you don't (and there's no other issue left), can we get this merged, and open a separate issue to discuss this point?

I feel sorry for taking long on this small PR but I prefer to split it out to another PR, discuss and resolve before the merge. I suspect our code reviewers will ask for a spec change if someone tries to contribute to Blink, so I'd like not to take that long path.

* We need to check whether white spaces collapses or not on every page, so adding this condition will add slight cost for the additional _if_ statement on every page on every user. It is a very hot code path we try hard to optimize, and I don't think this property is worth the cost for everyone.
* This property is only effective when `white-space` allows wrapping. `break-space` being only effective when `white-space` does not collapse looks more consistent to me. There are other such example, such as `text-orientation`, which is only effective in vertical flow. That way, we pay the cost to check the property only in vertical flow.
* I understand how you feel it more intuitive, but from another perspective, it's unintuitive to break consecutive white spaces when there are no such white spaces. I feel more intuitive for authors to create consecutive white spaces first before s/he allows them to break.

This PR looks good to me if this part was separated.

css-text/Overview.bs
frivoal: You're right. Fixed.

kojiishi:
Discussed offline to resolve misunderstandings, lgtm.

PR #114, opened 2016-05-16
[css-flexbox] Replace 'Animatable' descriptions with 'Animation type'

birtles:
With regards to 'flex-basis', I wasn't sure if "Animation type: length,
percentage, or calc; otherwise discrete" was more clear about the case
when the value is 'content'. However, we should probably just make sure
the general wording for resolving animation types covers this case, i.e.
"if you can't stuff the value in the animation type, it's discrete".

Note that this needs https://github.com/tabatkins/bikeshed/pull/709 (the corresponding bikeshed support for Animation type) to be merged first.

### DIFF
--- a/css-flexbox/Overview.bs
+++ b/css-flexbox/Overview.bs
@@ -51,6 +51,13 @@ spec: css-writing-modes-3; type: dfn;
 	text: end
 </pre>
 
+<pre class='anchors'>
+urlPrefix: https://drafts.csswg.org/css-transitions/#animtype-; type: dfn; spec: css-transitions
+    text: integer animation type; url: integer
+    text: number animation type; url: number
+    text: length, percentage, or calc animation type; url: lpcalc
+</pre>
+
 <style>
   #axis-mapping-table [rowspan=8] { writing-mode: sideways-lr; }
   #axis-mapping-table a { white-space: nowrap; }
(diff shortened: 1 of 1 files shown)


css-flexbox/Overview.bs
```diff
@@ -1481,7 +1488,7 @@ Flexibility</h2>
 <h3 id='flex-property'>
 The 'flex' Shorthand</h3>
 
-	<pre class='propdef'>
+	<pre class='propdef shorthand'>
```
tabatkins: What's this for? We don't mark shorthand-ness on the propdef.

tabatkins:
I'm gonna go ahead and reject this PR entirely; it's working around bad markup in Transitions and insufficient DWIM support in Bikeshed, and ends up being terribly verbose and easy to mess up and not something we want to encourage.  I'll fix the two source problems and get Flexbox fixed; from there you can do the rest of the specs if you want.

birtles:
@tabatkins What's DWIM?

This PR isn't about working around bad markup in Transitions but about replacing "Animatable" with "Animation type" and specifying the actual animation type.

css-flexbox/Overview.bs
birtles: It's because without that, bikeshed ends up generating "Animation type: discrete" (the default value for that line), but with "shorthand" added bikeshed generates "Animation type: see individual properties". That's what at least one of the other shorthands in this spec does.

PR #122, opened 2016-05-18
[css-flexbox] Fix animation type of 'flex-basis'

birtles:
'width' is not an animation type and quoting width like this links to the width property.


birtles:
@tabatkins Can you please take a look? Thanks

fantasai:
This is actually not a correct fix, because it doesn't handle keywords. The link to the 'width' property was intentional, as it should animate exactly the same as 'width'. Switched it to 'discrete or keywords, otherwise as length, percentage, or calc'.

birtles:
Yeah, the intention (as mentioned in the commit for #114) is to update CSS transitions to make it more obvious that if you can't stuff something into the specified animation type, you fall back to discrete. Otherwise we'll be typing this everywhere. This is fine for now though.

PR #169, opened 2016-06-07
[cssom-view] Fix #159. Do not leak a closed node via offsetParent

### LINKED ISSUE #159: [cssom-view] HTMLElement.prototype.offsetParent leaks a node inside a shadow tree
See the context: https://github.com/w3c/webcomponents/issues/497

@rniwa, do you have any preference to this issue? I do not have a strong opinion yet.

hayatoito:
My attempt to fix https://github.com/w3c/csswg-drafts/issues/159.

@tabatkins , could you review this?

> Return the nearest ancestor <a>unclosed node</a> element of the element

"unclosed node element of the element" can be acceptable?

### DIFF
--- a/cssom-view/Overview.bs
+++ b/cssom-view/Overview.bs
@@ -1352,7 +1352,7 @@ The <dfn attribute for=HTMLElement>offsetParent</dfn> attribute must return the 
     * The element is the root element.
     * The element is <a>the HTML <code>body</code> element</a>.
     * The element's computed value of the 'position' property is ''position/fixed''.
-1. Return the nearest ancestor element of the element for which at least one of the following is true and terminate this algorithm if such an ancestor is found:
+1. Return the nearest ancestor <a lt="unclosed node">unclosed</a> element of the element for which at least one of the following is true and terminate this algorithm if such an ancestor is found: [[!DOM]]
     * The computed value of the 'position' property is not ''static''.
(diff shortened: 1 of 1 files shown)


TakayoshiKochi:
Hmm, can it be "unclosed element", just use "unclosed" as an adjective?

And in that case we need to add "unclosed element" in the glossary definition in the shadow dom spec as well in addition to "unclosed node"?

zcorpan:
You could do `<a lt="unclosed node">unclosed</a> element` or some such. Also add `[[!DOM]]` at the end of the paragraph.

zcorpan:
(The commit message should say `[cssom-view]`, but we can fix that when merging.)

tabatkins:
@hayatoito Yup, agree with @zcorpan's review. Fix those two nits, and --amend the commit message to use the right tag, and i'll merge.

hayatoito:
@zcorpan  @tabatkins Thanks! I did amend the commit.

> You could do <a lt="unclosed node">unclosed</a> element or some such. Also add [[!DOM]] at the end of the paragraph.

Done.

> (The commit message should say [cssom-view], but we can fix that when merging.)

The subject of the commit includeded `[cssom-view]`, I think. I did not change the subject.
Is that okay?

zcorpan:
Thanks! (It said `[css-omview]` rather than `[cssom-view]`)

PR #184, opened 2016-06-12
[css-flexbox] Fix comment about default `flex` values

philix:


tabatkins:
This is correcting in the wrong direction - `0 1 auto` is the correct initial value.

The actual issue here is that shorthand props don't *have* initial values, so the 'flex' block is wrong - it should just say "see individual properties", like it does for all the other rows.

SebastianZ:
> The actual issue here is that shorthand props don't _have_ initial values, so the 'flex' block is wrong - it should just say "see individual properties", like it does for all the other rows.

I think that statement is not correct, shorthands may have initial values, they are just normally built up by their related longhand values. While there is no such case I know of, e.g. `flex` could have had `none` as initial value.
Also, 'see individual properties' is not very expressive. (Therefore [MDN lists the longhands in those cases](https://developer.mozilla.org/en-US/docs/Web/CSS/flex).) So, listing the individual longhands would be better, though that's out of the scope of this change here.

Sebastian

dholbert:
@tabatkins, that's what it used to say, BTW - but the full "initial value" got explicitly added here (by @fantasai), recently:
https://github.com/w3c/csswg-drafts/commit/96708b92cbe29fc0eafeee451ccd2095d2c7e3a3#diff-9c5fc491cf3bc3eb549a06890ba15b4fL1482

Sounds like that was a mistake, though?

tabatkins:
They seem to have explicitly added the language to two shorthands for some reason, despite that not being common practice.

SebastianZ:
> The actual issue here is that shorthand props don't have initial values, so the 'flex' block is wrong - it should just say "see individual properties", like it does for all the other rows.

I've opened issue #223 to discuss this.

Sebastian

PR #185, opened 2016-06-14
[mediaqueries-4] Can't have 'not', 'and' and 'or' at the same level

frivoal:
This fixes a mistake introduced in https://github.com/w3c/csswg-drafts/commit/2f37630466463a8653832584ad3c80fd6f2e01be

Not only can you not have `or` and `and` at the same level, but `not` is handled the same as well, and also needs extra parentheses.

### DIFF
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -679,15 +679,14 @@ Combining Media Features {#media-conditions}
 
 	* <a>Media conditions</a> can be grouped by wrapping them in parentheses ''()''
 		which can then be nested within a condition the same as a single media query.
-		For example, ''not (color) or (hover)''
+		For example, ''(not (color)) or (hover)''
 		is true on devices that are monochrome
-		and/or that have hover capabilities,
-		because the ''not'' only applies to the immediately following ''(color)'' feature.
+		and/or that have hover capabilities.
 		If one instead wanted to query for a device that was monochrome and <em>didn't</em> have hover capabilities,
 		it must instead be written as ''not ((color) or (hover))''
-		(or, equivalently, as ''not (color) and not (hover)'').
+		(or, equivalently, as ''(not (color)) and (not (hover))'').
 
-	It is <em>invalid</em> to mix ''and'' and ''or'' at the same “level” of a media query.
+	It is <em>invalid</em> to mix ''and'' and ''or'' and ''not'' at the same “level” of a media query.
 	For example, ''(color) and (pointer) or (hover)'' is illegal,
 	as it's unclear what was meant.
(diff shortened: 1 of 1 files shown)


frivoal:
Making this into a pull request rather than fixing it directly on the chance that I'm wrong. I'm pretty sure of this, but then again you should know it pretty well as well, so maybe *I* missed something somewhere.

tabatkins:
Hmmmm. I'd prefer if we could do "not" at the same level as the others, but I suppose I'd need to see what browsers actually do with and/not today.  Your change is indeed correct for what @dbaron wrote in Conditional, tho.

frivoal:
And for the grammar that you have in the MQ spec as well. I think the decision was not to have any implied operator priority, and keep it fully explicit always, to make sure there's not apparent ambiguity. Not sure why you think it is less important for `not` than for `and` vs `or`.

tabatkins:
Because "not" is unary, and "and" and "or" are binary; it's very standard for unary operators to either have very high or very low binding strength, whichever is most useful in context (and in this case I think it's obvious it would have high binding, like it does in other programming languages), but it's usually difficult to predict which of two binary operators have higher precedence.

frivoal:
I think it is pretty confusing with `not` as well. Binding strongly as in other languages sounds reasonable at first, but then you have to consider that in the legacy syntax `@media not screen and (width: 12345px) {...}` it binds more weakly than `and`.

Morality: don't guess, and stick parentheses in there, which is why we resolved to require them. (Can't find the link right now, but I am pretty sure we have a resolution a few years back).

tabatkins:
Yeah, that's reasonable.

simevidas:
@frivoal `@media not screen and …` is legacy syntax? Is this an upcoming change? The spec ([ch. 2.2.1.](https://drafts.csswg.org/mediaqueries/#mq-not)) uses it normally.

frivoal:
@simevidas By legacy, I did not mean something to be phased out. I meant what we've had so far, as opposed to this new boolean logic that is (relatively) newly being introduced.

fantasai:
On 06/14/2016 10:55 PM, Florian Rivoal wrote:
> I think it is pretty confusing with |not| as well. Binding strongly as in other languages sounds reasonable at first, but then
> you have to consider that in the legacy syntax |@media not screen and (width: 12345px) {...}| it binds more weakly than |and|.
>
> Morality: don't guess, and stick parentheses in there, which is why we resolved to require them. (Can't find the link right
> now, but I am pretty sure we have a resolution a few years back).

Given we've had media queries for awhile now, the correct thing
to do is whatever Media Queries Level 3 and its implementations
do. Anything else is likely to cause confusion, by having MQs
that work in new browsers but not old ones or vice versa.

~fantasai

PR #187, opened 2016-06-14
[css-flexbox] Correct typo in `flex` shorthand initial value

### LINKED ISSUE #183: [css-flexbox-1] Wrong comment about `flex` defaults
In the table about the `flex` propery, the initial property is said to be `1 0 auto` (https://www.w3.org/TR/css-flexbox-1/#flex-property)

But a comment below says `The initial values of the flex components are equivalent to flex: 0 1 auto`.

dholbert:
This fixes #183 -- the initial value of the `flex` shorthand is `0 1 auto`, not `1 0 auto`.

### DIFF
--- a/css-flexbox/Overview.bs
+++ b/css-flexbox/Overview.bs
@@ -1491,7 +1491,7 @@ The 'flex' Shorthand</h3>
 	<pre class='propdef shorthand'>
 	Name: flex
 	Value: none | [ <<'flex-grow'>> <<'flex-shrink'>>? || <<'flex-basis'>> ]
-	Initial: 1 0 auto
+	Initial: 0 1 auto
 	Applies to: <a>flex items</a>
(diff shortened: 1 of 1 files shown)


dholbert:
CC @fantasai -- looks like this was a typo that was introduced recently, here:
https://github.com/w3c/csswg-drafts/commit/96708b92cbe29fc0eafeee451ccd2095d2c7e3a3#diff-9c5fc491cf3bc3eb549a06890ba15b4fL1482

SebastianZ:
Instead of changing the value to `0 1 auto`, it should just be removed, so Bikeshed will add 'see individual properties'. See https://github.com/w3c/csswg-drafts/pull/184#issuecomment-225682922.

Sebastian

tabatkins:
Yup, I removed the lines. Closing this.

Issue #208, opened 2016-06-21
Fix link errors reported by Bikeshed


SebastianZ:
Bikeshed notes several referencial errors in the following specifications:
- [x] [CSS Color 4](https://drafts.csswg.org/css-color-4/)
- [x] [CSS Counter Styles 3](https://drafts.csswg.org/css-counter-styles-3/)
- [x] [CSS Extensions 1](https://drafts.csswg.org/css-extensions/)
- [ ] [CSS Font Loading 3](https://drafts.csswg.org/css-font-loading-3/)
- [x] [CSS Images 4](https://drafts.csswg.org/css-images-4/)
- [x] [CSS Inline Layout 3](https://drafts.csswg.org/css-inline-3/)
- [ ] [CSS Logical Properties 1](https://drafts.csswg.org/css-logical-props-1/)
- [ ] [CSS Multicol 1](https://drafts.csswg.org/css-multicol-1/)
- [ ] [CSS Multicol 2](https://drafts.csswg.org/css-multicol-2/)
- [ ] [CSS Pseudo-Elements 4](https://drafts.csswg.org/css-pseudo-4/)
- [ ] [CSS Round Display 1](https://drafts.csswg.org/css-round-display-1/)
- [ ] [CSS Scoping](https://drafts.csswg.org/css-scoping-1/)
- [ ] [CSS Scroll Snap 1](https://drafts.csswg.org/css-scroll-snap/)
- [ ] [CSS Shapes 2](https://drafts.csswg.org/css-shapes-2/)
- [ ] [CSS 2015](https://drafts.csswg.org/css-2015/)
- [ ] [CSS Step Sizing 1](https://drafts.csswg.org/css-step-sizing-1/)
- [ ] [CSS Tables 3](https://drafts.csswg.org/css-tables-3/)
- [ ] [CSS Transforms 1](https://drafts.csswg.org/css-transforms/)
- [x] [CSS User Interface 3](https://drafts.csswg.org/css-ui-3/)
- [x] [CSS User Interface 4](https://drafts.csswg.org/css-ui-4/)
- [ ] [CSS Writing Modes 3](https://drafts.csswg.org/css-writing-modes-3/)

_Note that I didn't include WG Notes, CSS Expressive Generalizations and Gadgetry 1 and the documents listed under 'Other Documents' at https://drafts.csswg.org/. If considered valuable, the section should be added there, too._

Sebastian

frivoal:
For CSS-UI-3, I get this:

[...]

PR #212, opened 2016-06-21
Fixed link errors (part of issue #208)

SebastianZ:


frivoal:
The multicol part of this patch looks good. For other specs, see with their respective editors.

SebastianZ:
Thank you for the review, Florian! Should I do something to get a review from the other spec. authors?

Sebastian

css-logical-props/Overview.bs
```diff
@@ -22,7 +22,7 @@ Introduction</h1>
   This mapping controls the interpretation of logical keywords and properties.
 
 <h2 id="logical-controls">
-Logical Directional Values: ''block-start'', ''block-end'', ''inline-start'', ''inline-end''</h2>
+Logical Directional Values: 'block-start', 'block-end', 'inline-start', 'inline-end'</h2>
```
tabatkins: These (and the next) are wrong - `'foo'` is the shorthand for property autolinks, `''foo''` is for values, and is being used correctly.

css-logical-props/Overview.bs
```diff
@@ -10,7 +10,7 @@ ED: https://drafts.csswg.org/css-logical-props/
 Editor: Rossen Atanassov, Microsoft, ratan@microsoft.com
 Editor: Elika J. Etemad / fantasai, Invited Expert, http://fantasai.inkedblade.net/contact
 Abstract: This module introduces logical properties and values that provide the author with the ability to control layout through logical, rather than physical, direction and dimension mappings. The module defines logical properties and values for the features defined in [CSS21]. These properties are writing-mode relative equivalents of their corresponding physical properties.
-Ignored Terms: div, dl, dfn
+Ignored Terms: div, dl, dfn, block-start, block-end, inline-start, inline-end, start, end
```
tabatkins: These shouldn't be ignored; they're terms that need to link up in this spec. If the definitions are broken that's something to be fixed, not ignored.

tabatkins:
I reviewed all of them. Left some comments; the rest are good.

css-logical-props/Overview.bs
SebastianZ: Ok, I've addressed this and the next comment in https://github.com/w3c/csswg-drafts/pull/212/commits/4447eecd147b717e522e717e7aa9dfe7ddac76d6.

> If the definitions are broken that's something to be fixed, not ignored.

My thought was that they are already defined in the prose. Those values affect many different properties in several specs, so I wonder how that can be defined correctly.

Sebastian

Issue #213, opened 2016-06-22
Remove Overview.html from specs running through Bikeshed


SebastianZ:
Specifications having an Overview.bs will automatically be parsed on the server in case there is no Overview.html.
This has the advantages that Bikeshed results are shown online, links are kept up to date, Bikeshed improvements take effect without editor effort and error in Bikeshed itself are indicated.

See also the [comment from @plinss about this](https://github.com/w3c/csswg-drafts/commit/7896268ba15a2af354e74e66ccdc2331473655af#commitcomment-17962683).

Affected specifications are:

- [x] [CSS Animations 1](https://drafts.csswg.org/css-animations-1/)
- [x] [CSS Animations 2](https://drafts.csswg.org/css-animations-2/)
- [x] [CSS Backgrounds 4](https://drafts.csswg.org/css-backgrounds-4/)
- [x] [CSS Fragmentation 3](https://drafts.csswg.org/css-break/)
- [x] [CSS Conditional 3](https://drafts.csswg.org/css-conditional-3/)
- [x] [CSS Exclusions 1](https://drafts.csswg.org/css-exclusions-1/)
- [ ] [CSS Fonts 3](https://drafts.csswg.org/css-fonts-3/) (once the Overview.bs is finished)
- [x] [CSS Fonts 4](https://drafts.csswg.org/css-fonts-4/)
- [x] [CSS GCPM 4](https://drafts.csswg.org/css-gcpm-4/)
- [x] [CSS Inline Layout 3](https://drafts.csswg.org/css-inline-3/)
- [x] [CSS Multicol 1](https://drafts.csswg.org/css-multicol-1/)
- [x] [CSS Overflow 4](https://drafts.csswg.org/css-overflow-4/)
- [x] [CSS Paged Media 3](https://drafts.csswg.org/css-page-3/)
- [x] [CSS Positioned Layout 3](https://drafts.csswg.org/css-position-3/)
- [x] [CSS Round Display 1](https://drafts.csswg.org/css-round-display-1/)
- [x] [CSS Ruby 1](https://drafts.csswg.org/css-ruby-1/)
- [x] [CSS Size Adjustment 1](https://drafts.csswg.org/css-size-adjust-1/)
- [x] [CSS Step Sizing 1](https://drafts.csswg.org/css-step-sizing-1/)
- [x] [CSS Text Decoration 3](https://drafts.csswg.org/css-text-decor-3/)
[...]

PR #227, opened 2016-06-23
[css-break-3] Rerun Bikeshed

SebastianZ:


tabatkins:
I'll just kill the `.html` instead.

SebastianZ:
> I'll just kill the `.html` instead.

Perfect for me. Comes in handy for issue #213.

Sebastian

PR #228, opened 2016-06-23
Removed Overview.html to let Bikeshed be run automatically on the server (issue #213)

SebastianZ:


frivoal:
Some of the specs I edit were affected by this change. For the record, I am OK with it.

However, I know that some of the specs were intentionally kept that way by their editor, and I am not sure it is appropriate to do this change without the editor's consent. In particular, I know that @dbaron did it intentionally. I don't know if that is still his position, but it would be good to check.

SebastianZ:
In the end it was Tab's decision to merge the PR, though if some editors have a problem with this, they should raise their voice in issue #213 (and maybe add the Overview.html back in case they have really good reasons for that).

Sebastian

Issue #240, opened 2016-06-25
[css-module-bikeshed] Add 'Privacy and Security Considerations' section


SebastianZ:
To pass the Bikeshed parsing without warnings there should be a *Privacy and Security Considerations* section added.

This also pushes editors to add that information to their specs.

Sebastian

SebastianZ:
After discussiong PR #241 with @frivoal it came to my mind that maybe the editors wait until their spec. stabalized, i.e. is in CR or PR, to add that section and use the Bikeshed warning as note that they still need to add it. Is that assumption correct?

If not, should an issue be added to the spec. for this as suggested by Florian?

Sebastian

tabatkins:
No, there's no actual reasoning behind leaving it out, besides that it's additional effort and usually low-value for CSS specs.

Marking the contents as an issue is fine.

SebastianZ:
Ok, I've done that in #241.

Sebastian

tabatkins:
danke!

SebastianZ:
Bitte! :smile:

PR #241, opened 2016-06-25
Added 'Privacy and Security Considerations' section to CSS Bikeshed Module

SebastianZ:


css-module-bikeshed/Overview.bs
```diff
@@ -159,3 +159,7 @@ Internal display model: the 'foo' property {#the-foo-property}
 			DOMString baz = "qux";
 		};
 	</pre>
+
+<h2 class=no-num id=priv-sec>Privacy and Security Considerations</h2>
+
+Make some considerations about privacy and security.
```
frivoal: How about making this “Make some considerations about privacy and security.” into an issue, to make sure it doesn't get forgotten?

SebastianZ: I don't have a strong opinion about that, but there is no other issue marking a 'must edit'.

Sebastian

frivoal: Well, adding this to the template only serves to silence a Bikeshed warning telling you you need to write this part. If we replace the warning with an issue in the spec, that's fine, but there was a reason for that warning: this is a part of the spec that is at risk of being forgotten unless you are explicitly reminded about.

SebastianZ: That it can be easily forgotten is one reason I opened issue #240.
Obviously very few editors actually added this section to their specs so far. So it seemed to me the Bikeshed warning is not enough and I was wondering about how to push people more to add that section.

It's ok for me to add an issue instead, though before I do that I'm asking about this at https://github.com/w3c/csswg-drafts/issues/240#issuecomment-228660782.

Sebastian

PR #267, opened 2016-07-04
[css-grid] Replaced <length> | <percentage> by <length-percentage>

SebastianZ:


fantasai:
While I appreciate the effort, this needs to wait on the CR update of Values and Units (which is currently blocked on calc serialization... which I might just drop for a quicker CR update if it isn't sorted out soon, but anyway, we're blocked on this atm). Would you prefer that we close the PR for now, or leave it open?

SebastianZ:
Tab already merged #263, #264 and #265, which did the same for CSS Backgrounds, CSS Position and CSS Fonts.
If you feel it's too early to merge, I'd prefer to leave it open. And I'll adjust it if it cannot be merged without conflicts later.

Sebastian

tabatkins:
Nah, it's just that we're publishing a Grid CR update *right now*. We need to push a V&U update so that `<length-percentage>` is *at least* at WD level.

SebastianZ:
> Nah, it's just that we're publishing a Grid CR update _right now_.

That "right now" obviously takes a bit. :smile:

> We need to push a V&U update so that `<length-percentage>` is _at least_ at WD level.

The [minutes of the telcon from 2016-08-03](https://lists.w3.org/Archives/Public/www-style/2016Aug/0009.html) don't include this dependency. Is the requirement to publish another V&U CR before Grid CR tracked somewhere?

Sebastian

SebastianZ:
Rebased the patch. As the `<*-percentage>` values finally made it into the V&U CR, it should be ok to merge it now.

Sebastian

SebastianZ:
Thank you for merging this PR, Tab!

Sebastian

PR #323, opened 2016-07-15
[css-values] Added 'vi' and 'vb' to heading of section 5.1.2

SebastianZ:


fantasai:
vb and vi were not supposed to go into Level 3, but since that's @tabatkins's mistake, I'm merging this in and then commenting them out. >:|

PR #344, opened 2016-07-27
[css-transitions] Add ontransitionend global event handler

birtles:


birtles:
This updates CSS Transitions in the same way we updated CSS Animations to have these global event handlers.

[Mozilla bug 911987](https://bugzilla.mozilla.org/show_bug.cgi?id=911987) contains most of the history for this. Basically I proposed adding them back at TPAC 2015 but @grorg initially seemed hesitant. Then in the F2F in Feb 2016, @grorg (and a few other Apple folk who were in the room) commented that they won't be dropping support for these. As a result we should just go ahead and add them like we did for CSS Animations.

Note that as with CSS Animations I've made this include both content attributes and IDL attributes although in my testing I notice that only Safari supports the IDL attributes (Chrome does only the content attributes). It seems more consistent, however, to support both which is what this pull request specifies.

plinss:
Ack, sorry I probably merged this prematurely. I saw the branch in the hg repo and forgot to check for an outstanding PR.

In the future, please submit PRs from a personal repo instead of branches in the main repo. When in the main repo they get copied over to hg and we generally try to keep branches out of there.

birtles:
I think this is probably fine to merge (and was just going to do it anyway rather than let it rot). I'll try to use a personal repo in future.

Issue #428, opened 2016-08-25
[cssom][css-images] CSS.elementSources incompatible with CSS being an IDL namespace


zcorpan:
From https://www.w3.org/Bugs/Public/show_bug.cgi?id=29623

https://drafts.csswg.org/css-images-4/#elementsources

reads

```
partial interface CSS {
  [SameObject] readonly attribute Map elementSources;
};
```

If we switch CSS over to be IDL namespace instead of interface, it can only use operations, not attributes, as far as I can tell.

`elementSources` is not implemented anywhere, is it? Should we change it to an operation when switching to IDL namespace? Or should namespaces be able to use attributes in Web IDL?

nox:
AFAICT that specification is wrong anyway, because it's an attribute so `CSS` should be instantiated first, but then the code snippet said:

```js
CSS.elementSources.set('fancy', bg);
```

Which clearly is trying to use `elementSources` has a static attribute.

zcorpan:
Yep, indeed.

On a related note `document.cssElementMap` was removed from HTML in https://github.com/whatwg/html/pull/258 (due to lack of implementations for years).

bzbarsky:
The use of `Map` here is a concern as well.  It causes some implementation issues and couples layout to scripting in ways that are not necessarily desirable.  A maplike would make a lot more sense.

Anyway, we could add attribute support to namespaces.  We just didn't have any use cases so far, and it's hard to design in a vacuum....

tabatkins:
Yeah, that spec version is way out of date and needs to get some stuff dropped.

PR #437, opened 2016-09-02
[cssom][css-conditional][css-images] Switch `CSS` to use IDL namespace

zcorpan:
Web IDL added support for namespaces in
https://github.com/heycam/webidl/pull/121

`elementSources` in css-images-4 is commented out for now because
IDL namespaces don't support attributes yet, see #428.

Fixes part of https://www.w3.org/Bugs/Public/show_bug.cgi?id=29623

---

cc @bzbarsky @domenic @tabatkins @dbaron

### DIFF
--- a/css-conditional/Overview.bs
+++ b/css-conditional/Overview.bs
@@ -997,9 +997,9 @@ The <code>CSS</code> interface, and the <code title=''>supports()</code> functio
 The {{CSS}} interface holds useful CSS-related functions that do not belong elsewhere.
 
 <pre class='idl'>
-partial interface CSS {
-  static boolean supports(DOMString property, DOMString value);
-  static boolean supports(DOMString conditionText);
+partial namespace CSS {
+  boolean supports(DOMString property, DOMString value);
+  boolean supports(DOMString conditionText);
 };
 </pre>
 
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -401,10 +401,12 @@ Using Out-Of-Document Sources: the <code>ElementSources</code> interface</h4>
 	The <a idl>elementSources</a> Map object provides this.
 
 	<pre class='idl'>
-		partial interface CSS {
-			[SameObject] readonly attribute Map elementSources;
+		partial namespace CSS {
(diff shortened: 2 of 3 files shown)


bzbarsky:
The change seems fine.  The question is which browsers are willing to try it to see whether it's web-compatible and whether we actually want to change the spec before such trials happen.....

fantasai:
If Web-compat seems likely, then we should change the spec, and put a note that it may be reverted if Web-compat problems are found. If Web-compat seems unlikely, then maybe just put a note that we're in the process of trying this out and the spec may be updated if it ends up working.

zcorpan:
Or maybe we can just wait with merging this until someone has tried to ship.

fantasai:
Seems like an obscure approach, is the problem. This way it's documented where the feature itself is documented.

tabatkins:
I'm happy to merge this now with a note in CSSOM next to the definition that it's contingent on web-compat. Do you want me to just add that in and merge it, @zcorpan?

zcorpan:
SGTM

zcorpan:
Should add support for namespaces in web-platform-tests's idlharness.js and add an interfaces.html test for cssom (and cssom-view).

bzbarsky:
That ended up with the wrong pull request number in the commit message... ah, well.

bzbarsky:
@tabatkins Are there existing bugs reported against browsers for this?

tabatkins:
Looks like Chrome has one, at least <https://bugs.chromium.org/p/chromium/issues/detail?id=727971>

bzbarsky:
That looks like the bug on supporting namespace at all, not for "CSS should be a namespace".

I guess the question is who is responsible for making sure that UAs actually investigate whether this is shippable.

tabatkins:
The second comment on that bug is about the CSS namespace specifically.

bzbarsky:
Yes, it's about the fact that the functionality is needed for that.  I'll be pretty surprised if this change gets made in that bug.

Anyway, I landed this in Gecko in https://bugzilla.mozilla.org/show_bug.cgi?id=1455805 but in general the CSSWG really should start filing bugs when spec changes happen....

PR #444, opened 2016-09-06
[css-color] Propdef table: color computed value can be currentcolor

SimonSapin:
https://drafts.csswg.org/css-color/#currentcolor-color says that `currentcolor` substitution happens at used-value time.


SimonSapin:
I was wrong. The `color` property is the one exception to this.

PR #557, opened 2016-09-30
[css-color] Changes to CSS color spec

cabanier:
I gathered some feedback from our color people and changed the spec per their recommendations


svgeesus:
Wait, I thought I was approving changesets individually not the entire PR :(

I need responses on the source of the  chromaticities for P3 and on the transfer function for Rec.2020 before approving.

css-color/Overview.bs
```diff
@@ -1714,7 +1717,7 @@ Profiled, Device-dependent Colors</h2>
 	or an RGB colorspace (such as ProPhoto <!-- ref -->, <!-- would be good to mention AdobeRGB 1998 here too, need to check trademark and copyright issues first --> widely used by photographers), or any other color or monochrome output device which has been characterized.
 	In addition, for convenience,
 	CSS provides two predefined RGB color spaces:
-	DCI P3 [[!DCI-P3]],
+	p3d65 [[!DCI-P3]],
```
palemieux: AFAIK the closest standard definition of "P3D65" is in Annex D of SMPTE ST 2067-21:2016, where it is specified as:
* the color primaries specified in SMPTE RP 431-2, and
* the illuminant D65 specified in SMPTE RP 177.

That definition does not include a transfer function. Moreover, the COLOR.6 colorspace defined in the same specification combines P3D65 colorimetry with the PQ transfer function.

I would therefore not use "P3D65", and instead use "P3D65Gamma_X_X".

Perhaps more importantly, I do not expect content to be distributed using this particular combination of colorimetry and transfer function.

css-color/Overview.bs
```diff
@@ -1872,18 +1890,8 @@ Predefined colorspaces: ''srgb'', ''dci-p3'', and ''rec2020''.</h3>
 				<tr><th>Green chromaticity</th><td>x=0.170 y=0.797</td></tr>
 				<tr><th>Blue chromaticity</th><td>x=0.131 y=0.046</td></tr>
 				<tr><th>White chromaticity</th><td>x=0.3127 y=0.3290 (D65)</td></tr>
-				<tr><th>Transfer function</th><td>see below</td></tr>
+				<tr><th>Transfer function</th><td>1/2.4</td></tr>
```
palemieux: This is not telling the full story. Rec. 2020 specifies an OETF with an exponent of 0.45, but notes that

> in typical production practice the encoding function of image sources is adjusted so that the final
> picture has the desired look, as viewed on a reference monitor having the reference decoding
> function of Recommendation ITU-R BT.1886

BT.1886 specifies an EOTF with a gamma of 2.4.

I recommend referencing Rec.2020, instead of attempting to replicate normative text here.

palemieux:
I think the PR needs work.

css-color/Overview.bs
svgeesus: I agree that the normative reference should be Rec.2020. The "see below" transfer function with the high precision coefficients was copied from that specification. Just saying that is is aboyt 1/2.4 is just as wrong as saying that sRGB is "about 1/2.2".

css-color/Overview.bs
svgeesus: It seems that Apple, at least, wanted the P3 value for the color gamut media query specifically so they could distribute wide-gamut still imagery in their Display P3 colorspace.

I'm waiting for an ICC profile from @grorg to see what exactly their Display P3 space is like.

tabatkins:
Note that dropping a single large PR with a bunch of unrelated changes and little explanation is *not* good etiquette.  This would have been vastly better as a set of PRs, one for each set of related changes, with an actual explanation of what each PR was trying to do and why.

svgeesus:
Agreed. For some of them I can see (and agree with) the rationale; for others I asked questions which will need answers before merging.

css-color/Overview.bs
cabanier: I was told that these are the values that are currently used by Apple in the p3 profile.
I can send that profile to the group if wanted so you can double-check

cabanier:
The changes are mostly in the definition of the profiles so it's not as big of a change as it seems.

I'm tracking down where this transfer function for 2020 is coming from.

css-color/Overview.bs
palemieux: @cabanier My primary comment is that the name should not be "p3d65" since it refers to something else already. A secondary question is whether the group intends to document specific display capabilities and/or encoding colorspaces. For example, sRGB is currently the encoding colorspace for the web even though not all displays can necessarily display sRGB completely.

cabanier: @palemieux the spec intends to document popular compositing/encoding spaces.
As for the name, we can call it anything else; it's just Apple's implementation of the p3 colorspace. Ideally, we should keep it short.

css-color/Overview.bs
cabanier: I looked over our discussion wrt 2020:

> W3C needs to follow the 2020 note or else 2020 videos will look washed out in browsers, the way QTP looks washed out.
> The transfer function (1886) for reference Rec.2020 displays is gamma 2.4.
> EBU has more details on reference monitors.
> 
> For a computer in an office, Recommendation ITU-R BT.2035 does not apply. The sRGB curve was designed for this purpose. 
> So W3C should consider either gamma 2.4 for dark surrounds or sRGB for average surrounds, but not the current pick.

svgeesus:
If you pulled the chromaticity values out of an ICC profile and it is a v.4 profile, be aware that those chromaticities are already adapted to D50 and the adaptation applied is in the `chad` table. For this spec we need the unadapted primaries, as chromatic adaptation to D50 is already performed in the XYZ to Lab step.
The primaries and transfer function currently in the spec are directly from the BT.2020 spec, which is freely available.

cabanier:
I commented earlier on the transfer function. I didn't pull the values out of the profile; they were given to me (and different from the ones in the profile)

css-color/Overview.bs
svgeesus: Thanks, Rik. That sort of detailed rationale would have been very helpful along with the original PR, as tab notes. I now agree with specifying the sRGB curve for normal viewing environments, and the spec should provide the specific rationale you excerpted above.

cabanier: @svgeesus how should we add that rationale? Do you want me to update the PR?

css-color/Overview.bs
grorg: I would suggest displayp3 or display-p3. Here's the ICC file that ships with macOS Sierra.
[Display P3.icc.zip](https://github.com/w3c/csswg-drafts/files/506728/Display.P3.icc.zip)

palemieux: @grorg What happens if a display ships in the future with the same p3 primaries but different transfer characteristics and/or luminance range?

grorg: Or maybe just p3.

cabanier: Just p3 could be confusing. 'displayp3' sgtm

palemieux: @cabanier What happens if a display ships in the future with the same p3 primaries but different transfer characteristics and/or luminance range?

cabanier: @palemieux a display with such vastly different characteristics that an ICC color transform can't correct for it?

palemieux: @cabanier What do you mean by "an ICC color transform can't correct for it"? Do you mean using an ICC profile to ensure that a color encoded as P3/D65/Gamma_2_2/100nits is correctly shown on a display with broader capabilities... or something else?

css-color/Overview.bs
palemieux: FYI. See [ITU Rec.2087](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2087-0-201510-I!!PDF-E.pdf) for recommendations on "Colour conversion from Recommendation ITU-R BT.709 to Recommendation ITU-R BT.2020"

cabanier: @svgeesus I added a note for rec2020
@palemieux why did you post that link?

css-color/Overview.bs
cabanier: @grog I agree. Also, these spaces are for targeting input (colors, images, videos, etc) and not displays.

palemieux: >  Also, these spaces are for targeting input (colors, images, videos, etc) and not displays.

Makes sense to me... so "image-p3" is a color space that will be used to encode pixels in photos? I do not expect it to be used for professional content. Perhaps this was all discussed before and I appreciate your patience.

css-color/Overview.bs
palemieux: @cabanier ... because ITU Rec.2087 provides two data points that can be useful to implementers and authors:

(a) conversion from Rec 2020 to Rec 709
(b) discusses the two distinct cases, each of which use a different EOTF: "where the goal is to preserve colours seen on a Rec. 709 display" and "In the case where the source is a direct camera output and the goal is to match the colours of a direct Rec. 2020 camera output"

css-color/Overview.bs
cabanier: @palemieux: yes, for now it's to tag input colors. There's a note/todo in the spec that we need to tag the compositing space but that will likely happen in a later spec.

cabanier: @grorg I renamed p3d65 and updated the PR

css-color/Overview.bs
```diff
@@ -1717,7 +1717,7 @@ Profiled, Device-dependent Colors</h2>
 	or an RGB colorspace (such as ProPhoto <!-- ref -->, <!-- would be good to mention AdobeRGB 1998 here too, need to check trademark and copyright issues first --> widely used by photographers), or any other color or monochrome output device which has been characterized.
 	In addition, for convenience,
 	CSS provides two predefined RGB color spaces:
-	p3d65 [[!DCI-P3]],
+	display-p3 [[!DCI-P3]],
```
palemieux: RP 431-2 specifies a transfer function (gamma 2.6 and absolute XYZ coordinates) and white point there are no consistent with the use in this document. Also note that the term "p3" is not used once in RP 431-2.

At the very least, only Annex C.1 of RP 431-2 should be referenced. Ideally, RP 431-2 would not be normatively referenced, and instead the document would merely mention that the selected primaries happen to be identical to those specified in RP 431-2.

cabanier: @palemieux you are correct! Sorry. I will fix this asap

css-color/Overview.bs
Crissov: I don’t know where the `P3` part comes from, but a more useful keyword for authors would probably be `cinema` (or `movie-theater`, `film-projector` …), alongside `uhdtv`, `hdtv`, `sdtv`, `ntsc`, `pal`/`ebu`, … ([table at Wikipedia](https://en.wikipedia.org/wiki/RGB_color_space#Specifications))

cabanier: @crissov Displays are marketed as being "P3" (see Apple iMac and Ipad pro).
Also, there's already a P3 media query: https://drafts.csswg.org/mediaqueries-4/#color-gamut

Crissov: My comment equally applies to MQ4 gamuts then, I guess.

cabanier:
Should I pull something out or can it be merged as-is? (and we can address minor issues later)

cabanier:
What should I do? break it up or remove features or can the pr be merged?

svgeesus:
I still don't see a source for the different chromaticities (apart from "they were given to me"). Breaking this PR into smaller chunks would certainly help; some parts of the current PR could be merged as is while other parts are blockers.

cabanier:
Sorry that I didn't see your latest comment. Let me look into the source of the chromaticities and get back to you.
I didn't actually make that many changes. It looks like a lot because it caused changes in several areas.

cabanier:
The chromicities come from SMPTE ST 2067-21:2016 
Apple’s display profile for the sRGB curve.

cabanier:
This PR has been open for a while now. What can we do to get it merged?
I taled to the people at ICC and they also want to see these changes in the document.

css-color/Overview.bs
```diff
@@ -2653,9 +2667,6 @@ Issue: The code below is only for sRGB, and duplicates the
 			Otherwise, set the channel's value to
 			<code>((channel + 0.055) / 1.055) ^ 2.4</code>.
 
-			Note: This reverses the logarithmic power scaling of the sRGB gamut,
-			so the value of the channel is  linear with respect to the amount of light produced.
```
svgeesus: Pointing out that the value is linear wrt the light produced (i.e., it is additive) still seems valuable.

cabanier:
I don't see the changes reflected in the public draft. Is there a problem with the publishing process?

svgeesus:
I don't either, but I do see

CSS Color Module Level 4
Editor’s Draft, **14 November 2016**

I see them in the Overview.bs on GitHub.

cabanier:
something must have gone wrong in the auto-publishing process.
Who should we contact?

plinss:
If there's a problem, ping me. In this case, the server was showing the most recent push which was an accidental branch from some time back. I merged the branch and it's now showing the latest version (will look at updating the server to always show the tip).

Issue #352, opened 2016-07-28
[css-fonts][css-text] Variation Selection of Colorful (Emoji) or Monochrome Glyphs


Crissov:
In late March, on [www-style](https://lists.w3.org/Archives/Public/www-style/2016Mar/0370.html), Addison Philips posted a [request](https://lists.w3.org/Archives/Member/member-i18n-core/2016Mar/0015.html) on behalf of W3C’s I18N WG to be able to control the style of emoji characters via CSS. There were links to the relevant [I18N minutes](https://www.w3.org/2016/03/24-i18n-minutes.html) and [Unicode minutes](http://www.unicode.org/L2/L2015/15314-css-ctrl-emoji.pdf) as well as the reference document [UTR51](http://www.unicode.org/reports/tr51/proposed.html#Presentation_Style).

> `default` : style characters to use each character's default display
> `text` : style characters to use their text display
> `emoji` : style characters to use their emoji display

The issue had been brought up in [2014](https://lists.w3.org/Archives/Public/www-style/2014Aug/0322.html) already, but I cannot find a WG conclusion on it. It has been suggested to be handled by [`font-variant-color`](https://lists.w3.org/Archives/Public/www-style/2014Aug/0322.html) (John Dagget), [`font-variant-emoji`](https://lists.w3.org/Archives/Public/www-style/2016Mar/0373.html) (Xidorn Quan @upsuper), “something more generic” or [`@font-palette`](https://lists.w3.org/Archives/Public/www-style/2014Aug/0353.html) (Cameron McCormack @heycam) or “a new property” (Amelia Bellamy Royds @AmeliaBR).

I agree with some of the concerns that this is not an issue appropriate to be handled by [`font-variant-*`](https://drafts.csswg.org/css-fonts-3/#font-rend-props), not just because there’s no corresponding Open Type feature. I was about to propose the selection being part of [`text-transform`](https://drafts.csswg.org/css-text-3/#transforming), but I then realized that this would only be reasonable to _override or synthesize_ variation selectors (U+FE0E = VS-15 ‘text’, U+FE0F = VS-16 ‘emoji’). [UTR51](http://www.unicode.org/reports/tr51/proposed.html#Presentation_Style) wants VSs to take precedence over normal styling, though. I suggest we add both capabilities!

```
font-color: 'auto' | 'monochrome' | 'colorful'  | <color>+

text-transform: <text-transform-case> || <text-transform-width> || <text-transform-variation>
[...]

PR #561, opened 2016-10-02
[css-color] Add note for relationship between color property and color glyphs

upsuper:
Per [discussion in the mailing list](https://lists.w3.org/Archives/Public/www-style/2016Mar/0431.html).

### DIFF
--- a/css-color/Overview.bs
+++ b/css-color/Overview.bs
@@ -86,6 +86,15 @@ For example, to specify lime green:
 		<dd>The <<color>> type is defined in a later section.
 	</dl>
 
+    <p class="note">
+    Note: In general, this property, including its alpha component, has
+    no effect on "color glyphs", such as emoji in some fonts, which are
+    colored by a built-in palette. Some colored fonts are able to refer
+    to the foreground color, such as palette entry ''0xFFFF'' in COLR
+    table of OpenType, and ''context-fill'' value in SVG-in-OpenType. In
+    that case, the foreground color is set by this property, identical
+    to how ''<color>/currentcolor'' value works.
+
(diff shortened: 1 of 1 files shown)


Crissov:
Text color (and its opacity) should probably not affect PNGinOTF glyphs (`CBDT` and Apple), but should it also not affect fonts with `COLR` or `SVG ` tables? Should we solely rely on future explicit [`CPAL` control](https://drafts.csswg.org/css-fonts-4/#font-palette-control)?

#352 is somewhat related and I hope to see the issue on a WG agenda soon.

svgeesus:
The specification for COLR and CPAL says that the color values are in sRGB.
The specification for SVG allows whatever CSS defines.

CSS4 Fonts does, as @Crissov mentions, allow palette selection, and new palettes in CSS.

tabatkins:
I'd prefer text more like:

> Note: This property has no effect on "color glyphs", such as emoji in some fonts, which are explicitly colored by a built-in palette, unless explicitly specified otherwise by a font specification. For example, some colored fonts use a palette with the ability to refer to the "current color", such as the CPAL table in OpenType; in that case the "current color" is set by this property, identically to how the ''currentcolor'' value works.

Am I referring to things correctly here?  I'm just assuming CPAL is an OpenType table.

litherum:
I prefer @tabatkins's comment too, because it is more descriptive.

svgeesus:
> Am I referring to things correctly here? I'm just assuming CPAL is an OpenType table.

It is, yes. 
https://www.microsoft.com/typography/otspec/cpal.htm

Issue #442, opened 2016-09-05
[mediaqueries] Media Feature: "reduce motion" user setting


cookiecrook:
Tracking issue for Media Feature "prefers-reduced-motion" 

Originally accepted at W3C TPAC, October 2014:
https://lists.w3.org/Archives/Public/www-style/2014Oct/0527.html

More recently discussed in February 2016 thread:
https://lists.w3.org/Archives/Public/www-style/2016Feb/0037.html

For spec: 
https://drafts.csswg.org/mediaqueries/

cookiecrook:
Allows certain views to remove or tone down animations. Platform preference is shipping on iOS, macOS, watchOS, and tvOS. 

For example:
- launching an app on iOS changes to a subtle dissolve animation rather than a full-screen zoom.
- switching Spaces or launching Exposé on macOS Sierra

Customers regularly mention to us that common web site patterns (e.g. horizontal carousels or scroll-jacking animations) cause vestibular issues such as dizziness or vertigo. Web developers ought to be able to achieve a similar solution that app developers can achieve natively.

Some background:
http://simplyaccessible.com/article/balance-awareness/

Example calls for this support from the web development community:
https://twitter.com/zeldman/status/492805247455072256
https://twitter.com/webaxe/status/489159894860197888

cookiecrook:
Changing these user settings don't change the rendering of anything. It just conveys a user preference that allows the frameworks, native apps, or web apps to adjust for this user preference/desire/need.

I should also note these proposed names don't fit well within the "none or truthy" pattern of some existing media features. It'd be awkward to specify that `prefers-reduced-motion: none` means "user is okay with animation." The none value here may be open to misinterpretation, so considering a `default` or `no-preference` value that behaves like `none` for boolean comparisons.

prefers-reduced-motion: [ default | reduce ];
prefers-reduced-motion: [ no-preference | reduce ];

astearns:
If using 'none' is a pattern we want to maintain, then we could rename the prefers-* properties:

reduced-motion-preference: [ none | reduce ]
reduce-motion-preference: [ none | reduce ]

cookiecrook:
The core reason this needs to expose a user preference rather than change something about the view is because UI varies greatly. There's no way a UA can reliably reduce or stop motion on behalf of the user and still guarantee an understandable interface. Exposing the pref gives authors a way to do this appropriately within the context of their own interfaces.

cookiecrook:
[...]

Issue #443, opened 2016-09-05
[mediaqueries] Media Feature: "increase contrast" and/or "reduce transparency" user settings


cookiecrook:
Tracking issue for Media Feature "prefers-reduced-transparency" 

Discussed in thread:
https://lists.w3.org/Archives/Public/www-style/2016Feb/0037.html

For spec: 
https://drafts.csswg.org/mediaqueries/

cookiecrook:
Allows web developers to know they should switch certain translucent views to an opaque or significantly less translucent rendering. Platform preference is shipping on iOS, macOS, watchOS, and tvOS. This can greatly increase readability for individuals with certain vision impairments.

cookiecrook:
Changing these user settings don't change the rendering of anything. It just conveys a user preference that allows the frameworks, native apps, or web apps to adjust for this user preference/desire/need.

I should also note these proposed names don't fit well within the "none or truthy" pattern of some existing media features. It'd be awkward to specify that `prefers-reduced-motion: none` means "user is okay with animation." The none value here may be open to misinterpretation, so considering a `default` or `no-preference` value that behaves like `none` for boolean comparisons.

prefers-reduced-transparency: [ default | reduce ];
prefers-reduced-transparency: [ no-preference | reduce ];

cookiecrook:
Based on suggestion by @astearns in #442 

reduce-transparency-preference: [ none | reduce ]
increase-opacity-preference: [ none | increase ]

cookiecrook:
The core reason this needs to expose a user preference rather than change something about the view is because UI varies greatly. There's no way a UA can reliably reduce transparency on behalf of the user and still guarantee a readable, usable interface. Exposing the pref gives authors a way to do this appropriately within the context of their own interfaces.

frivoal:
(this comment is equally valid for this issue and for #442)

This seems very reasonable and useful to me.

1) I am a little worried about adding this in isolation though. I believe this is an OS X / iOS specific preference. If we list all (or even just many) such settings from all (or even just many) OSes one by one, I worry that we will end up with a very long list, from which each UA vendor will only pick a few, and authors will themselves also only respond to the few they know/care about. The intersection of both meaning that for end users, the odds of this actually working might be pretty low. Maybe we can spec anything we run into, wait a few years to see what sticks, and deprecate the rest, but I am not sure this is a great approach.

I think It would be very useful to see a broad survey of all such accessibility/preference settings across major OSes to determine:
[...]

PR #586, opened 2016-10-10
Add 'prefers-reduced-motion' media feature

grorg:
Added a new section for user preferences, with the assumption we'll get more of them.

Added the prefers-reduced-motion feature, discussed in #442 

WebKit plans to implement this asap.

### DIFF
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1783,6 +1783,55 @@ Rare Interaction Capabilities: the 'any-pointer' and 'any-hover' features</h3>
 		or to hide them altogether to reduce visual clutter if ''(any-pointer: fine)'' is false.
 	</div>
 
+
+
+<!--
+  ██     ██  ██████  ████████ ████████  ████████  ████████  ████████ ████████  ██████
+  ██     ██ ██    ██ ██       ██     ██ ██     ██ ██     ██ ██       ██       ██    ██
+  ██     ██ ██       ██       ██     ██ ██     ██ ██     ██ ██       ██       ██
+  ██     ██  ██████  ██████   ████████  ████████  ████████  ██████   ██████    ██████
+  ██     ██       ██ ██       ██   ██   ██        ██   ██   ██       ██             ██
+  ██     ██ ██    ██ ██       ██    ██  ██        ██    ██  ██       ██       ██    ██
+   ███████   ██████  ████████ ██     ██ ██        ██     ██ ████████ ██        ██████
+-->
+
+<h2 id='mf-user-preferences'>
+User Preference Media Features</h2>
+
+<h3 id="prefers-reduced-motion">
+Detecting the desire for less motion on the page: the 'prefers-reduced-motion' feature</h3>
+
+	<pre class='descdef mq'>
+	Name: prefers-reduced-motion
+	Value: none | default | reduce
+	For: @media
+	Type: discrete
+	</pre>
+
+	The 'prefers-reduced-motion' media feature is used to detect if the user
+	has requested the system minimize the amount of animation or motion it
+	uses.
+
+	<dl dfn-type=value dfn-for="@media/prefers-reduced-motion">
+		<dt><dfn>none</dfn>
+		<dd>
+			Equivalent to 'reduce'.
+
+		<dt><dfn>default</dfn>
+		<dd>
+			Indicates that the user has made no preference known
+			to the system.
+
+		<dt><dfn>reduce</dfn>
+		<dd>
+			Indicates that user has notified the system that they
+			prefer an interface that minimizes the amount of movement
+			or animation, preferably to the point where all non-essential
+			movement is removed.
+
+	</dl>
+
+
 <!--
  ██████   ██████  ████████  ████ ████████  ████████ ████ ██    ██  ██████
 ██    ██ ██    ██ ██     ██  ██  ██     ██    ██     ██  ███   ██ ██    ██
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -491,7 +491,7 @@ Evaluating Media Features in a Boolean Context</h4>
 	If the feature would be true for any value
 	<em>other than</em> the number ''0'',
 	a <<dimension>> with the value ''0'',
-	or the keywords ''none'' and ''default'',
+	or the keywords ''none'' and ''no-preference'',
 	the <a>media feature</a> evaluates to true.
 	Otherwise, it evaluates to false.
 
@@ -1803,7 +1803,7 @@ Detecting the desire for less motion on the page: the 'prefers-reduced-motion' f
 
 	<pre class='descdef mq'>
 	Name: prefers-reduced-motion
-	Value: default | reduce
+	Value: no-preference | reduce
 	For: @media
 	Type: discrete
 	</pre>
@@ -1813,7 +1813,7 @@ Detecting the desire for less motion on the page: the 'prefers-reduced-motion' f
 	uses.
 
 	<dl dfn-type=value dfn-for="@media/prefers-reduced-motion">
-		<dt><dfn>default</dfn>
+		<dt><dfn>no-preference</dfn>
 		<dd>
 			Indicates that the user has made no preference known
 			to the system. This keyword value evaluates as false
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -491,7 +491,7 @@ Evaluating Media Features in a Boolean Context</h4>
 	If the feature would be true for any value
 	<em>other than</em> the number ''0'',
 	a <<dimension>> with the value ''0'',
-	or the keyword ''none'',
+	or the keywords ''none'' and ''default'',
 	the <a>media feature</a> evaluates to true.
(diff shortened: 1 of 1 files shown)


grorg:
Oops. Hold off on this. I need to remove the None keyword.

grorg:
Ready now (fingers crossed)

grorg:
@frivoal any comments on this?

frivoal:
@grorg Sorry for the delay, I have not forgotten. This is in my todo list, gently making its way to the top. Hopefully it'll get there soon.

tabatkins:
I don't particularly like using "default" as a falsey value. It doesn't *feel* falsey.  On the other hand, I can see why "none" doesn't feel super appropriate here.

I'm okay with adding a new falsey value for the preference values, tho. What about "no" or "unknown"? Both of these feel much falsier than "default"  to me, and also seem like they might be usable in the future for other MQs in a similar situation, where "0" or "none" don't quite mesh.

cookiecrook:
`unknown` seems reasonable to me

inoas:
The current wording - e.g. **prefers-reduced-motion** does not allow any future granularity like say numbers, 3d vs 2d, or whatever comes to mind. It is very limiting in future.

Is there a particular reason these where not considered?

```
(motion-preference: all) / (motion-preference: reduced) / (motion-preference: disabled)
(prefers-motion: all) / (prefers-motion: reduced) / (prefers-motion: off)
```

frivoal:
I agree with the general shape of the proposal, the semantics and the new spec section, but I don't think the naming is great, and also dislike `default` being falsey.

My proposal back from #443 was
`reduced-motion: none | prefer`, which could possibly be expanded later with other truthy values, like for example `force` for when the UA is disabling all motions on its own, and we just want an MQ to let the author know about it so that they can react.

I don't think the `force` value is particularly likely on `reduced-motion`, but I do think it would be more likely on other similar "preference" properties, and I'd like to get the naming consistent.

`reduced-transparency: none | prefer | force`
`reduced-motion: none | prefer | force`
`color-invert: none | prefer | force`
`save-ink: none | prefer | force`

I'd also be OK with a replacing `none` with another falsey value, but `unknown` feels wrong, because it feels like an invocation of the 3-valued logic described in https://drafts.csswg.org/mediaqueries-4/#evaluating.

cookiecrook:
@inoas wrote:
> The current wording - e.g. prefers-reduced-motion - does not allow any future granularity 

Not true. Future values could be granular tokens, even a combined list, if someone wanted to implement them that way.

`prefers-reduced-motion: reduce-parallax reduce-scaling; /* etc. reduce-rotation, etc. */`

> the falsey-value issue is also being addressed in the alternative PR I hope.

Which alternative PR? Please link.

cookiecrook:
@frivoal wrote:
> My proposal back from #443 […snip…]
> ```css
reduced-transparency: none | prefer | force
reduced-motion: none | prefer | force
```

I thought we settled that this syntax was unclear. `reduced-motion: none;` will leave many to believe it means "no motion" rather than "no preference, so motion is okay."

**Comment by inoas** [comment]
@cookiecrook 
Sure it is true. The variable/setting itself declares that it is only about reduction not about disabling. 
Because of that prefers-reduced-motion: disabled is ambiguous.
If you look at my alt PR https://github.com/w3c/csswg-drafts/issues/778 you can see the issue is not appearing anymore.

motion-preference: none will simply say you have no preference, e.g. take the default.
motion-preference: disable will however disable motion, reduce will lower it...
And in future, IF DESIRED: values between 0 and 1000 would allow speeds, 0-100% would allow scaling, 0ms - 500ms would allow concrete max animation length.

**Comment by cookiecrook** [comment]
@inoas wrote:
> The variable/setting itself declares that it is only about reduction not about disabling.

If you're talking about the term "reduced" in the property name, I think we're mincing semantics. Disabling motion is a reduction of motion, albeit a complete one. 

> Because of the choice of words in this PR `prefers-reduced-motion: disabled` is ambiguous in the first place.

This argument is a straw man fallacy. The PR does not include `prefers-reduced-motion: disabled` but if the value were needed a later time, it could be an unambiguous `no-motion`.

**Comment by nattarnoff** [comment]
A complete reduction of non-interaction triggered animation has to be allowed in the settings. Given the scenario where items animate into a scene, users prone to motions sickness, vertigo, migraines, or distraction need this turned off yet usable. 

Interaction triggered animation needs to be minimized, some stopped depending on what it is. Small animations such as color fades or animated icons on buttons or links likely won't cause a trigger point, but other animations non-essential to the control function may cause problems for users. 

Reducing motion is good. But stopping/preventing/pausing animation and making the content still readable, actionable, and usable is needed. Additionally, best practices should be worded in a PE approach of no-motion, reduced-motion, full-motion structuring of CSS.

**Comment by cookiecrook** [comment]
@nattarnoff wrote:

> A complete reduction of non-interaction triggered animation has to be allowed in the settings. 

Currently, no platforms implement that in native UI or web UI. At such a time that any do, it would be appropriate to standardize in Web API. 

> Interaction triggered animation needs to be minimized, some stopped depending on what it is. Small animations such as color fades or animated icons on buttons or links likely won't cause a trigger point, but other animations non-essential to the control function may cause problems for users. Reducing motion is good. But stopping/preventing/pausing animation and making the content still readable, actionable, and usable is needed. 

Herein lies the problem. As CSS and JavaScript-based animations are implemented as primitives, there is no good way for a browser to know which animations may be triggers. Apple has chosen to expose this on its native platforms as a user preference which allows authors to adopt and determine which animations are appropriate to reduce. This takes evangelism, but seems to be working well.

**Comment by nattarnoff** [comment]
>Currently, no platforms implement that in native UI or web UI. At such a time that any do, it would be appropriate to standardize in Web API.

Why couldn't the Web API take the lead instead of waiting for native or web UI? If we establish the ability and best practice, are we not evangelizing that they should catch up and implement it? 

>there is no good way for a browser to know which animations may be triggers

Sadly after thinking through this a little more, I see this problem. Perhaps it is out of scope for this PR, but it is something that should be investigated. On the WCAG side of the fence, there is https://github.com/w3c/wcag21/issues/18 which is about defining accessible animation from interaction. SC 2.2.2 Pause, Stop, Hide covers animation that isn't by direct user interaction.

**Comment by cookiecrook** [comment]
@nattarnoff wrote:
> Why couldn't the Web API take the lead instead of waiting for native or web UI? 

I think you'll find any proposal to disable *all* animations on the Web very controversial and difficult to implement well. Also, advocacy through specification is usually frowned upon. 

Perhaps try an incubation approach in WICG instead?

**Comment by frivoal** [comment]

File: mediaqueries/Overview.bs [hunk 1]
```diff
@@ -491,7 +491,7 @@ Evaluating Media Features in a Boolean Context</h4>
 	If the feature would be true for any value
 	<em>other than</em> the number ''0'',
 	a <<dimension>> with the value ''0'',
-	or the keyword ''none'',
+	or the keywords ''none'' and ''default'',
```
cookiecrook: Please change `default` to `no-preference` per [discussion](https://github.com/w3c/csswg-drafts/issues/442#issuecomment-266637739)

mediaqueries/Overview.bs
```diff
@@ -1783,6 +1783,52 @@ Rare Interaction Capabilities: the 'any-pointer' and 'any-hover' features</h3>
...
+
+<h2 id='mf-user-preferences'>
+User Preference Media Features</h2>
+
+<h3 id="prefers-reduced-motion">
+Detecting the desire for less motion on the page: the 'prefers-reduced-motion' feature</h3>
+
+	<pre class='descdef mq'>
+	Name: prefers-reduced-motion
+	Value: default | reduce
```
cookiecrook: ditto

mediaqueries/Overview.bs
```diff
@@ -1783,6 +1783,52 @@ Rare Interaction Capabilities: the 'any-pointer' and 'any-hover' features</h3>
...
+	For: @media
+	Type: discrete
+	</pre>
+
+	The 'prefers-reduced-motion' media feature is used to detect if the user
+	has requested the system minimize the amount of animation or motion it
+	uses.
+
+	<dl dfn-type=value dfn-for="@media/prefers-reduced-motion">
+		<dt><dfn>default</dfn>
```
cookiecrook: ditto

mediaqueries/Overview.bs
```diff
@@ -1783,6 +1783,52 @@ Rare Interaction Capabilities: the 'any-pointer' and 'any-hover' features</h3>
...
+		<dt><dfn>reduce</dfn>
+		<dd>
+			Indicates that user has notified the system that they
+			prefer an interface that minimizes the amount of movement
+			or animation, preferably to the point where all non-essential
+			movement is removed.
+
+	</dl>
+
+
```
cookiecrook: Maybe add a editorial note to the discussion in #442?

frivoal: I'd rather not do that, at least not without a summary of what you want people to go look for in that giant rambling of ours.

grorg:
Updated to use "no-preference" over "default".

frivoal:
Available here. https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion

PR #591, opened 2016-10-11
[css-position] Fix some minor typos and grammatical issues

gilmoreorless:
This fixes some minor spelling mistakes, duplicated phrases and singular/plural grammatical nitpicks in the css-position spec. It also removes a dummy "example" section that appears to have been left over from the bikeshed template.

As a side comment, I found some parts of the spec difficult to read/comprehend properly. I have some ideas of improvements that can be made, but given that css-position is a new draft spec, I wasn't sure how far through the writing process it was.
What's the etiquette here? Create separate pull requests with suggested changes, or just raise an issue or two? (Or leave it be until the spec is further advanced?)

### DIFF
--- a/css-position/Overview.bs
+++ b/css-position/Overview.bs
@@ -435,7 +435,7 @@ Relative positioning</h3>
   A relatively positioned box keeps its <a>normal flow</a> size, including
   line breaks and the space originally reserved for it.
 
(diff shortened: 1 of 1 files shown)


atanassov:
Thanks for the fixes! Creating a PR for something as simple as the editorial changes here is just fine. For anything more elaborate I'd rather you start with an issue and we can go from there.

PR #596, opened 2016-10-13
[css-containment] Reuse existing terminology

### LINKED ISSUE #396: [css-containment] reuse "ink overflow" terminology
This is a follow up form https://lists.w3.org/Archives/Public/www-style/2016Feb/0320.html which was left unresolved.

point 3 of item https://drafts.csswg.org/css-containment/#containment-layout says this:

> If the contents of the element overflow the element, they must not project any "geometry" beyond the element’s normal bounds. In particular, any overflow must not be capable of causing the element’s ancestors to overflow; as far as the ancestors are concerned, the element is perfectly confined to its normal width×height bounds.

As far as I can tell, this is trying to define an effect identical to ink overflow, as defined in https://www.w3.org/TR/css-overflow-3/#ink.

Rather than redefining it, we should just reuse that term:

> If the contents of the element overflow the element, they must be treated as ink overflow

frivoal:
Closes #396

### DIFF
--- a/css-containment/Overview.bs
+++ b/css-containment/Overview.bs
@@ -221,10 +221,7 @@ Layout Containment</h3>
 				it must instead compute to ''continue/overflow''.
 
 	3. If the contents of the element overflow the element,
-		they must not project any "geometry" beyond the element's normal bounds.
-		In particular, any overflow must not be capable of causing the element's ancestors to overflow;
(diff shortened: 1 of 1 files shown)


frivoal:
@tabatkins How about merging this (or letting me know why you cannot)?

PR #601, opened 2016-10-14
[cssom] Make MediaList.item return serialization.

### LINKED ISSUE #600: [cssom] MediaList.item(index) should return a serialization
The return value of `item` is `DOMString?`, however the spec currently says:
> The item(index) method must return the media query in the collection of media queries given by index, or null, if index is greater than or equal to the number of media queries in the collection of media queries.

It should say it returns a serialization of the media query.

upsuper:
Fixes #600.


zcorpan:
❤️

PR #602, opened 2016-10-14
[css-values] Add the ''lh'' and ''rlh'' units

frivoal:
Resolution in https://lists.w3.org/Archives/Public/www-style/2016Aug/0009.html

### DIFF
--- a/css-values/Overview.bs
+++ b/css-values/Overview.bs
@@ -884,6 +884,10 @@ Relative lengths</h3>
 				as represented by the “水” (CJK water ideograph, U+6C34) glyph
 		<tr><td>''rem''
 			<td>font size of the root element
+		<tr><td>''lh''
+			<td>line height of the element
+		<tr><td>''rlh''
+			<td>line height of the root element
 		<tr><td>''vw''
 			<td>1% of viewport's width
 		<tr><td>''vh''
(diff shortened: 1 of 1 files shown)


tabatkins:
While this copies the circularity-breaker of font-size/em in handling line-height using lh, it doesn't address the new cycle we're adding when font-size uses lh and line-height uses em.

We can either break this situation by making *both* refer to the parent element's metrics, or by just making one of them refer to the parent element and letting the other one rely as usual. 

I think I prefer treating the em unit as slightly more fundamental, so that line-height/em will use the parent element's metrics, and then font-size/lh will just rely on the element's metrics as usual. However, I could easily be convinced otherwise by a decent argument.  We probably want to poll implementors on what seems best to them.

frivoal:
> While this copies the circularity-breaker of font-size/em in handling line-height using lh, it doesn't address the new cycle we're adding when font-size uses lh and line-height uses em.

I'm not seeing it.

What happens to the font-size when it is defined in `lh` is addressed: I did not add a new paragraph for that, but the existing one about the exception to `font-size` is for all font-relative length units, and `lh` is classified as one.

And as long as you have that, it doesn't matter that the `line-height` depends on the font-size (which it does even if you don't use `em`s, and use percentages or unitless numbers), since there's no loop.

Did you miss that, or am I missing something else? Can you give an example of this loop you're seeing and that I am not?

> I think I prefer treating the em unit as slightly more fundamental, so that line-height/em will use the parent element's metrics, and then font-size/lh will just rely on the element's metrics as usual. 

How would that work when line-height, instead of being in `em`, is a naked number or a % or its initial value of normal? In that case, it still depends on `font-size,` so that would be causing a loop, wouldn't it?

frivoal:
If you agree (or if we end up agreeing on something else), I'm happy to submit tests to make sure that the next person looking at this doesn't miss on the subtlety.

frivoal:
@tabatkins Could you follow up on that review?

fantasai:
I think it would be surprising if line-height: 1em; didn't make the line height exactly match the font size. On the other hand, I can see that font-size: 1lh to make the font size match the parent line-height would be useful. So I think that's a better way to break it down.

fantasai:
Accepting the pull request. We can fix up the cycle issue as an additional commit, since it's very simply additive.

PR #631, opened 2016-10-20
[cssom] Handle custom properties separately.

### LINKED ISSUE #580: [cssom] "case-sensitive match for a supported CSS property"?
In each of:
- https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty
- https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setpropertyvalue
- https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setpropertypriority

… one of the steps is:

> If property is not a case-sensitive match for a supported CSS property, terminate this algorithm. 

… with a link to:

https://drafts.csswg.org/cssom/#supported-css-property

> The term supported CSS property refers to a CSS property that the user agent implements, including any vendor-prefixed properties, but **excluding custom properties**.

(Emphasis added.)

This does not match the behavior of at least Firefox and Chromium, where the matching is (probably ASCII) case-insensitive, and custom properties are not excluded:

``` js
document.documentElement.style.setProperty("Color", "green")
document.documentElement.style.setProperty("--a", "b")
assert_equals(document.documentElement.style.getPropertyValue("color"), "green");
assert_equals(document.documentElement.style.getPropertyValue("--a"), "b");
```

### LINKED ISSUE #630: [cssom][css-variable] Methods of CSSStyleDeclaration interface doesn't handle custom properties properly
In all methods which take a `property` parameter, there is a very early step says:
> Let property be property converted to ASCII lowercase.

However, [name of custom properties is case-sensitive](https://drafts.csswg.org/css-variables/#defining-variables), so this step would mess up custom properties.

In addition to that, some of the methods also have:
> If property is not a case-sensitive match for a supported CSS property, terminate this algorithm.

This would make the method not usable for custom properties.

I think the solution is to recognize custom properties in an early step, and only apply the above steps to non-custom properties.

upsuper:
Fixes #630, fixes #580.

### DIFF
--- a/cssom/Overview.bs
+++ b/cssom/Overview.bs
@@ -108,6 +108,7 @@ spec:dom; type:interface; text:Document
 spec:html; type:dfn; text:ascii case-insensitive
 spec:html; type:dfn; text:case-sensitive
 spec:css-logical-props-1; type:property; text:inline-size
+spec:css-variables-1; type:dfn; text:custom property
 </pre>
 
 <script src=https://resources.whatwg.org/file-issue.js defer data-file-issue-url="https://github.com/w3c/csswg-drafts/issues/new?title=%5Bcssom%5D%20"></script>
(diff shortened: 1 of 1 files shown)


upsuper:
So this should fix #580.

zcorpan:
Thanks @upsuper!

@SimonSapin does this look good to you?

SimonSapin:
Looks good, thanks!

tim-loh:
I think getComputedStyle needs to be updated too:

> declarations
> All longhand properties that are supported CSS properties, in lexicographical order, with the value being the resolved value computed for obj using the style rules associated with doc.

upsuper:
You can open a new pull request for that.

PR #692, opened 2016-11-08
Remove text-orientation references from css-logical-props 

### LINKED ISSUE #691: [css-logical-props][css-writing-modes] Specify relation of text-orientation with abstract-to-physical mapping
In [writing-modes §3 Logical Box Properties](https://drafts.csswg.org/css-logical-props/#logical-box-props), it says

> Which pairs of properties share computed values depends on the computed values of writing-mode, text-orientation, and direction. For a summary of these dependencies, see [Abstract-to-Physical Mappings](https://drafts.csswg.org/css-writing-modes/#logical-to-physical)

The actual linked abstract-to-physical mappings do not mention text-orientation at all. They should, if there indeed is a dependence.

cc @heycam

Manishearth:
(fixes #691)

r? @fantasai @heycam @kojiishi

### DIFF
--- a/css-logical-props/Overview.bs
+++ b/css-logical-props/Overview.bs
@@ -145,7 +145,7 @@ Logical Box Model Properties
   but the logical and physical properties share computed values.
   Which pairs of properties share computed values
   depends on the computed values of
-  'writing-mode', 'text-orientation', and 'direction'.
+  'writing-mode' and 'direction'.
   For a summary of these dependencies, see
   <a href="https://drafts.csswg.org/css-writing-modes/#logical-to-physical">Abstract-to-Physical Mappings</a>
   in [[!CSS3-WRITING-MODES]].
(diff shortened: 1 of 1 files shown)


css-logical-props/Overview.bs
```diff
@@ -270,7 +270,7 @@ the margin- and offset- block-start/block-end/inline-start/inline-end properties
 
   These properties correspond to the 'top', 'bottom', 'left', and 'right'
   properties. The mapping depends on the <em>parent element's</em>
-  'writing-mode', 'direction', and 'text-orientation'.
+  'writing-mode', 'direction'.
```
kojiishi: nit: s/, / and / as in other places

Manishearth:
nit fixed

atanassov:
Thanks for the fix.

PR #693, opened 2016-11-08
[css-flexbox] Clarify implied minimum size definition

### LINKED ISSUE #671: [css-flexbox] Clarify automatic min size's definition if both specified size and aspect ratio are not present
https://drafts.csswg.org/css-flexbox/#min-size-auto

In the section: 

> In general, the automatic minimum size is the smaller of its content size and its specified size. However, if the box has an aspect ratio and no specified size, its automatic minimum size is the smaller of its content size and its transferred size.

The case which no specified size and no aspect ratio is present is not clear. People might get confused and use the old default `0`.  

In the [previous edition](https://hg.csswg.org/drafts/rev/25641c92d33f), it is clear that we should use `content size` in that case.

shinglyu:
Fixes #671. In the [old version](https://hg.csswg.org/drafts/rev/25641c92d33f) this case was specified, but dropped while changing from a table to a paragraph.

r? @dbaron

### DIFF
--- a/css-flexbox/Overview.bs
+++ b/css-flexbox/Overview.bs
@@ -904,6 +904,8 @@ Implied Minimum Size of Flex Items</h3>
 	However, if the box has an aspect ratio and no <a>specified size</a>,
 	its <a>automatic minimum size</a>
 	is the smaller of its <a>content size</a> and its <a>transferred size</a>.
+	If the box has neither a <a>specified size</a> nor an <a>aspect ratio</a>, 
+	its <a>automatic minimum size</a> is the <a>content size</a>.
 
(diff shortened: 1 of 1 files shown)


dbaron:
Er, the github notification I got pointed to a commit in [Shing's repository](https://github.com/shinglyu/csswg-drafts) and gave me no hint whatsoever that the comment I made on the commit would show up as a comment on a pull request in a different repository, or even that such a pull request existed!

So, anyway, /cc @tabatkins

tabatkins:
It wasn't intentionally dropped; it was intended that this was covered by the first clause - the min() of a value and something that doesn't exist is the value.

That said, I have nothing against making this more explicit. Merging.

PR #701, opened 2016-11-10
Fix wording in parsing a media query list

upsuper:
A media query list can contain no media query at all, so "the list of one or more media queries" is incorrect.


zcorpan:
Thanks!

upsuper:
Oops, did I push to a branch of csswg-drafts direct?! Sorry about that...

zcorpan:
No problem at all 🙂

Issue #708, opened 2016-11-14
[css-values] clarify that most (all?) high-dpi devices anchor on pixel, rather than physical, unit


patrickhlauke:
The section on absolute lengths https://drafts.csswg.org/css-values/#absolute-lengths currently contains the following:

> For print media and similar high-resolution devices, the anchor unit should be one of the standard physical units (inches, centimeters, etc). For lower-resolution devices, and devices with unusual viewing distances, it is recommended instead that the anchor unit be the pixel unit. For such devices it is recommended that the pixel unit refer to the whole number of device pixels that best approximates the reference pixel.

In essence, this splits devices into "print and high resolution" on one side, and "lower-resolution and unusual viewing distances". However, most (all?) high-dpi devices (including recent desktop monitors, laptops, mobile/tablet devices) currently also anchor on the pixel unit. It seems that it's really only print media that (try to) anchor on physical units (unless I'm missing some non-print device that does this?).

I think it would be important to either rewrite this section to change the distinction, or to leave the above as is but then include a note straight after clarifying that currently even high-resolution desktops/laptops/mobiles/tablets are anchored on the pixel unit.

The reason for bringing this up is that this issue seems to constantly rear its head in discussions around being able to use absolute measures like "mm" to actually map to physical size of elements *as rendered on a device* (most recently I had this discussion in the mobile accessibility task force https://www.w3.org/WAI/GL/mobile-a11y-tf/wiki/Summary_of_Research_on_Touch/Pointer_Target_Size#Comparison_of_web_content_sized_in_px_and_physical_size_of_rendered_content_on_common_device_screens which led to having to demonstrate this practically using http://codepen.io/patrickhlauke/full/zqabMR/ ... and I'm aware of similar discussions under WICG in relation to large screen / TV environments)

Happy to draft a PR if this sounds like a reasonable request.

nicksherman:
Related (perhaps duplicate?): Issue #614

patrickhlauke:
related yes. not duplicate though, I'd say (as #614 is proposing a change in what UAs do / new units of measure, while here I'm proposing making the spec match/describe reality a bit more closely)

frivoal:
I agree with this. This isn't really changing how things work, a the way units are anchored to eachother stays the same, and the recommendation to anchor the whole system to either the physical units is only a recommendation, not a requirement. But it is indeed a recommendation that does not match reality, and I don't think it is particularly good advice: what browsers do today is the right thing to do.

[...]

PR #713, opened 2016-11-15
Change anchor for high-dpi devices to pixel

patrickhlauke:
Make the spec match reality, as most (all?) high-dpi screen devices
currently also anchor on pixel unit, and not any physical unit.

Closes https://github.com/w3c/csswg-drafts/issues/708


frivoal:
Looks good to me (but I am not an editor of that spec).

patrickhlauke:
Wonderful, thank you

patrickhlauke:
See further discussion https://github.com/w3c/csswg-drafts/issues/708#issuecomment-302240095

Issue #704, opened 2016-11-13
[css-device-adapt] note/warning about not restricting zoom


patrickhlauke:
as per https://lists.w3.org/Archives/Public/www-style/2016Sep/0115.html and in light of the addition of a warning about the accessibility/usability concerns of preventing users from being able to zoom in the HTML spec with reference to the `viewport` meta (see https://github.com/w3c/html/commit/29b89343cbc12fa385143de3ddf70144d38b6095 and my subsequent tweak https://github.com/w3c/html/pull/607), i'm wondering if there's any mileage in me submitting a proposed PR with an initial draft note along the same lines for the CSS Device Adaptation spec? /cc @frivoal @stevefaulkner

frivoal:
@patrickhlauke Yes, a PR would be great.

PR #714, opened 2016-11-15
Port HTML note on zoom/accessibility

patrickhlauke:
Ports the note warning/note recently added to
https://w3c.github.io/html/document-metadata.html#the-meta-element about
not suppressing/limiting user's ability to zoom, and the note that UAs
can ignore these directives anyway.

Closes https://github.com/w3c/csswg-drafts/issues/704

### DIFF
--- a/css-device-adapt/Overview.bs
+++ b/css-device-adapt/Overview.bs
@@ -524,7 +524,9 @@ Computed value: auto, or a non-negative number or percentage as specified
 Specifies the largest allowed zoom factor. It is used as input to the
 <a href="#constraining-procedure">constraining procedure</a> to constrain
 non-''zoom/auto'' 'zoom' values, but also to limit the allowed zoom factor
-that can be set through user interaction. The UA should also use this
+that can be set through user interaction. The UA may choose to ignore
+this limit for accessbility/usability reasons – see the relevant note in
+the 'user-zoom' section. The UA should also use this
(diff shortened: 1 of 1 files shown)


css-device-adapt/Overview.bs
```diff
@@ -524,7 +524,9 @@ Computed value: auto, or a non-negative number or percentage as specified
 Specifies the largest allowed zoom factor. It is used as input to the
 <a href="#constraining-procedure">constraining procedure</a> to constrain
 non-''zoom/auto'' 'zoom' values, but also to limit the allowed zoom factor
-that can be set through user interaction. The UA should also use this
+that can be set through user interaction. Note that the UA may choose to ignore
```
frivoal: ”Note that” makes this sentence sound informative, while I believe the intent is to normatively allow UAs to ignore this.

frivoal:
Other than the comment above, looks good to me. Agenda+ing because to get confirmation from the WG, as this is a change in behavior.

css-device-adapt/Overview.bs
patrickhlauke: I originally intended it as an informative "this is what some UAs currently do anyway", but I guess at that point it makes sense to normatively allow it. I'll push a change to that effect.

patrickhlauke:
👍

Issue #690, opened 2016-11-07
[mediaqueries-4] Interaction Media Features make a rigid primary/"rare" distinction


patrickhlauke:
The section on Interaction Media Features https://drafts.csswg.org/mediaqueries-4/#mf-interaction makes, throughout its description text, a hard distinction between the "primary" input and other input types, even going as far as labelling the others as "rare".

As pointed out previously (e.g. in this article https://dev.opera.com/articles/media-features/ - which needs updating due to the removal of `on-demand`), fundamentally this hard distinction is becoming much more blurry nowadays.

For instance, on a touch-enabled laptop, or one of those new Surface Studio desktop devices, what is the *primary* input? I know the idea is that the OS/UA will make that call, but by virtue of even having that distinction in the spec, you're discounting the increasing possibility that devices have multiple equi-potent input modalities. There may well be more than one "primary" input, and labelling anything other than whatever the OS/UA decides is "primary" as being "rare" is an inappropriate qualitative statement in the spec.

At the very least, I would like to see language such as "rare" being amended/removed. Call it "secondary", at a stretch, and possibly include stronger advice that the `any-*` results should be treated not as some courtesy

> some extra conveniences or non essential controls can be offered to users who have additional ways to interact.

because if a user does rely on whatever the OS/UA deems as "rare"/"secondary", then it's not just about being "nice" to the user...if you rely on things like hovering or high pointing precision, you ARE going to stop users from being able to use your site/app if you only concentrate on the "primary".

More fundamentally, I'd actually love to see the concept of "primary" input go away altogether as it's increasingly inappropriate in multi-input environments, but I take it that'd be a step too far? 

(for what it's worth, note for instance that in the Pointer Events spec, we don't try to categorise any type of pointer by it being "primary" or not...they're all treated as equally valid/important)

patrickhlauke:
Happy to make a PR following some discussion here...but wanted to get a feel for this first from the WG

patrickhlauke:
The potentially radical solution (if we agree that multi-input devices where different input types can be of equal importance) would be to of course drop `pointer` and `hover`, and *only* rely on `any-pointer`/`any-hover` queries. A halfway solution perhaps to keep `pointer` and `hover`, but to add a very strong warning in the spec pointing out that the concept of "primary" may not be applicable to situations with multiple input devices used concurrently/interchangeably, and to advise authors to just rely on `any-*` queries...and mark `pointer` and `hover` as "at risk".

[...]

PR #715, opened 2016-11-15
[mediaqueries-4] expand/change emphasis of advice for use of pointer/hover/any-pointer/any-hover 

patrickhlauke:
Related to the discussion in https://github.com/w3c/csswg-drafts/issues/690 this is an initial expansion/change to the current advice given for `pointer`/`hover`/`any-pointer`/`any-hover`.

[edit: copying the relevant description from the individual commits here, for completeness]

* Change "rare" to "all available": "rare" is a loaded term, unnecessarily implying (without much foundation) the likelihood/frequency that a user may use inputs deemed "not primary" by the OS/UA. In addition, as `any-*` also include the pointer that is deemed primary, the "rare" moniker is even more inappropriate, as by definition the "primary" is not "rare".

* Reword the note about pointer/hover vs any-*: Change the emphasis of the note, away from making the `any-*` features sound like a nice-to-have, and more towards aknowledging that these can be useful in making a page/app adapt better/preemptively to all available input methods.

* Add new example for any-hover usage: Include the common touchscreen laptop scenario, and show how `any-hover` can be used to make two completely different judgement calls on the part of the author (one inclusive, one exclusive)

* Add additional any-* consideration note to pointer/hover: The rationale given in any-pointer/any-hover about not relying exclusively on any-* is equally valid in reverse here. This does not imply that authors must design for "lowest common denominator", but simply opens up the possibility that they may wish to keep non-primary inputs in mind when deciding on their layout/functionality.

* Soften/expand the advice in the modified note for any-*: Reworded to remove the implication that authors should "ensure" that "ideally" they take any-* into account...turning it into a "may", which explains more about the rationale for the existence of these `any-*` features without seemingly mandating that authors must follow a specific style/design decision.

While I'm still not a fan of the "primary" concept, this PR would be a compromise I could (currently) live with (in which case this PR could close the relate issue)

### DIFF
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1607,7 +1607,7 @@ Pointing Device Quality: the 'pointer' feature</h3>
 	Type: discrete
 	</pre>
 
-	The 'pointer' media feature is used to query about the presence and accuracy of a pointing device such as a mouse.
+	The 'pointer' media feature is used to query the presence and accuracy of a pointing device such as a mouse.
 	If a device has multiple input mechanisms,
 	the 'pointer' media feature must reflect the characteristics of the “primary” input mechanism,
 	as determined by the user agent.
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1747,17 +1747,17 @@ All Available Interaction Capabilities: the 'any-pointer' and 'any-hover' featur
 	or there are no pointing devices at all.
 
 	<div class="note">
-	These two media features are best suited to provide refinements over 'hover' and 'pointer',
-	and it is rarely appropriate to use them
-	without having first designed the main style and interaction mode of the page
-	to suit the primary input mechanism.
-	Once that has been taken into account, 'any-hover' and 'any-pointer' can be used
-	to see if some extra conveniences or non essential controls
-	can be offered to users who have additional ways to interact.
+	While 'pointer' and 'hover' can be used to design the main style and interaction
+	mode of the page to suit the primary input mechanism, 'any-pointer' and 'any-hover'
+	can be used to potentially take into account all possible types of pointers
+	that have been detected.
 
 	Designing a page that relies on hovering or accurate pointing
-	only because 'any-hover' or 'any-pointer' indicate that an input mechanism with these capabilities is available,
-	is likely to result in a poor experience.
+	only because 'any-hover' or 'any-pointer' indicate that at least one of the available
+	input mechanisms has these capabilities	is likely to result in a poor experience.
+	However, authors may use this information to inform their decision about the style and
+	functionality they wish to provide based on additional input mechanisms that
+	are available to the user.
 	</div>
 
 	<div class="example">
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1723,7 +1723,7 @@ Hover Capability: the 'hover' feature</h3>
 	</div>
 
 <h3 id='any-input'>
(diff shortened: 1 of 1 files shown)


mediaqueries/Overview.bs
```diff
@@ -1607,7 +1607,7 @@ Pointing Device Quality: the 'pointer' feature</h3>
 	Type: discrete
 	</pre>
 
-	The 'pointer' media feature is used to query about the presence and accuracy of a pointing device such as a mouse.
+	The 'pointer' media feature is used to query the presence and accuracy of a pointing device such as a mouse.
```
frivoal: Is this to fix incorrect grammar, or are both sentences valid but mean different things? My non native speaker brain prefers the current phrasing, but I am very possibly either missing a nuance or not noticing a mistake.

mediaqueries/Overview.bs
```diff
@@ -1709,6 +1713,10 @@ Hover Capability: the 'hover' feature</h3>
 	For accessibility reasons, even on devices that do support hovering,
 	the UA may give a value of ''hover: none'' to this media query,
 	to opt into layouts that work well without hovering.
+	In addition, even if the primary input mechanism has 'hover: hover' capability,
+	there may be additional input mechanisms available to the user where 'hover: none' is true.
+	Authors may	wish to query the 'any-hover' media feature to take these other non-hover-capable
+	potential input mechanisms into account.
```
frivoal: This does not work. While there may be other devices which cannot hover (and arguably this is a common case: every desktop computer with a mouse also has a keyboard that cannot hover), the query will not let you detect that in the way you suggest. `any-hover: none` only matches if none of the devices can hover, not merely if some of them cannot.

Maybe we should change that (although I am not convinced), but until we do, this note is incorrect.

mediaqueries/Overview.bs
```diff
@@ -1788,6 +1796,16 @@ Rare Interaction Capabilities: the 'any-pointer' and 'any-hover' features</h3>
...
+	<div class="example">
+	    On a touchscreen laptop, the built-in trackpad is usually treated as the primary
+	    input mechanism. As such, ''(hover: hover)'' is true. However, due to the presence of the
+	    touchscreen, ''(any-hover: none)'' is also true. An author may decide that in this case,
+	    the page should not rely on hover capabilities, as a user may choose to
+	    use the non-hover-capable input mechanism. On the other hand, an author may decide that
+	    use of the hover capability is essential for their page, and display a notification
+	    reminding users to only use their mouse/trackpad.
+	</div>
+
```
frivoal: Same comment as above:  `any-hover: none` only matches if none of the devices can hover, not merely if some of them cannot, so this example is incorrect.

patrickhlauke:
oh, i must have misunderstood the `any-hover:none` in that case. i'll make fixes later today so it can be merged

mediaqueries/Overview.bs
patrickhlauke: This is for grammar, but also for consistency (if you check the rest of the spec, there are 5 instances of "query the ...", e.g. " query the value of individual features", "query the ability of the output device to modify the apearance of content ..." etc

patrickhlauke:
Force-pushed a change removing the stuff relating to my incorrect understanding of `any-hover` (though I may file a separate issue about changing `any-hover` to that effect, with the example I removed in this version as a use case)

frivoal:
Thanks, merged.

> though I may file a separate issue about changing any-hover to that effect, with the example I removed in this version as a use case

Sure. And thanks for agreeing to slice this issue in smaller chunks. It seem easier to make progress this way.

Issue #737, opened 2016-11-18
[mediaqueries-4] Change when any-* evaluate to 'none'


patrickhlauke:
Currently, `any-pointer: none` and `any-hover: none` only evaluate to true when they apply to ALL pointers/input mechanisms. There are situations, however, where it would be useful if these evaluated to true if ANY pointers/input mechanisms had those characteristics. Filing a speculative PR together with this issue - see the examples I propose there for use cases. [edit: [see examples/use cases from the PR](https://github.com/w3c/csswg-drafts/pull/738/commits/08eaa2874d320210d613d3e5849707b9b7855023) ]

This would clearly be a substantive change to the spec, but one that I feel would open up some more granular possibilities for authors.

(this issue is a continuation of https://github.com/w3c/csswg-drafts/pull/715 which in turn came out of this isse https://github.com/w3c/csswg-drafts/issues/690)

tabatkins:
There are definitely some interesting use-cases to address that aren't fully addressed by (pointer) and (any-pointer) (and hover/any-hover) right now, but I don't think that generalizing (any-pointer) to allow it to match *anything* is the right approach.  The PR as written just removes *any* instructions on what values to should match.

If we can come up with good criteria for determining what any-pointer/any-hover should match (more elaborate than just "the union of capabilities", as currently defined), then there might be something here.  Otherwise, we might be looking at a different solution altogether - for example, maybe it's useful to explicitly call out the capabilities of the *secondary* input devices (so (secondary-pointer) would be just like (any-pointer), but excluding the primary devices).  Or maybe we want a JS enumeration of the input devices with some annotations of their capabilities, so apps that really care can be smart about it.  

Regardless, I think the PR as written won't work.

patrickhlauke:
>  The PR as written just removes any instructions on what values to should match.

- if ANY inputs are present that are non-hover capable, `any-hover: none` should evaluate to true
- if ANY inputs are present that are hover capable, `any-hover: hover` should evaluate to true
- if ALL inputs that are present are non-hover capable, `any-hover: none` should be true and `any-hover: hover` should be false
- if a mix of hover-capable and non-hover capable inputs is present, both `any-hover: none` and `any-hover: hover` should evaluate to true

(same for `any-pointer`, saving myself from writing out the combinations here)

Of course, an author would need to know exactly what they're asking for, and why, in order to make sensible use of this.

patrickhlauke:
[...]

PR #738, opened 2016-11-18
[mediaqueries-4] Change when any-* evaluate to 'none'

patrickhlauke:
Currently, `any-pointer: none` and `any-hover: none` only evaluate to true when they apply to ALL pointers/input mechanisms. There are situations, however, where it would be useful if these evaluated to true if ANY pointers/input mechanisms had those characteristics. See the examples in this PR for use cases where this would be the case. This would clearly be a substantive change to the spec, but one that I feel would open up some more granular possibilities for authors.

(this issue is a continuation of #715 which in turn came out of this isse #690)

Closes https://github.com/w3c/csswg-drafts/issues/737


patrickhlauke:
Closing this, as I've come around to at least part of the reasoning (for `any-pointer:none`, but not for `any-hover:none`) in https://github.com/w3c/csswg-drafts/issues/737#issuecomment-269348418

PR #740, opened 2016-11-20
Optimize images

kaelig:
I was browsing the source of this repository and noticed that images could be optimized.

So I started optimizing images using [ImageOptim](https://imageoptim.com/mac), and broke optimizations into small commits (one per spec), meaning it's easy to revert in case any of the optimizations caused regressions (unlikely).

The result: optimizing images saved more than 8 MB in total.

Note that some formats weren't optimized: SVG, Animated Gifs.

I hope this helps for users' loading performance as well as potentially reducing the hosting bill.


atanassov:
Thank you for taking the time and optimizing our images @kaelig :)

kaelig:
No problem, thanks for merging the PR!

inoas:
Thank you!

PR #754, opened 2016-11-24
[cssom] Check index for rule insertion first

### LINKED ISSUE #753: [cssom] Algorithm of "insert a CSS rule" should check the index first
All mainstream browsers agree with the same behavior that the index is checked first. Parsing the rule is obviously more expensive than checking the index, so we shouldn't try to parse it if the index is already invalid.

Test case:
```html
<!DOCTYPE html>
<style></style>
<script>
  var sheet = document.styleSheets[0];
  sheet.insertRule(">", 1);
</script>
```

This test case yields index error on Firefox, Chrome, and Edge, and if you change the index argument to `0`, they all show a syntax error, which indicates they do have a different steps than what is speced.

upsuper:
Fixes #753.


zcorpan:
Thanks!

PR #773, opened 2016-11-29
Fix typo (capitalization at start of sentence)

dholbert:
Just fixing a typo I noticed when reading this section of the spec - capitalizing a word at the start of a sentence.

(The next sentence starts with the same term & it's capitalized there, so this isn't an always-lowercase-term or anything like that.)


frivoal:
This isn't my spec, but this fix is editorial and obviously right, so merging anyway. Editors, feel free to be angry at me if you think I invaded your spec inappropriately.

PR #778, opened 2016-11-30
Add 'motion-preference' media feature

inoas:
Implements Issue: https://github.com/w3c/csswg-drafts/issues/442
Alternative to https://github.com/w3c/csswg-drafts/pull/586

Having a good wording of the feature also gets around default value evaluating to true instead of false. Ref: https://github.com/w3c/csswg-drafts/pull/586#issuecomment-262327733

### DIFF
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1846,6 +1846,7 @@ Detecting the preference for less or no motion on the page: the 'motion-preferen
 			prefer an interface that minimizes the amount of movement
 			or animation, preferably to the point where all non-essential
 			movement is removed.
+		<dt><dfn>disable</dfn>
 		<dd>
 			Indicates that user has notified the system that they
 			prefer an interface that disables movement or animation,
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1807,6 +1807,53 @@ All Available Interaction Capabilities: the 'any-pointer' and 'any-hover' featur
 	</div>
 
 <!--
+  ██     ██  ██████  ████████ ████████  ████████  ████████  ████████ ████████  ██████
+  ██     ██ ██    ██ ██       ██     ██ ██     ██ ██     ██ ██       ██       ██    ██
+  ██     ██ ██       ██       ██     ██ ██     ██ ██     ██ ██       ██       ██
+  ██     ██  ██████  ██████   ████████  ████████  ████████  ██████   ██████    ██████
+  ██     ██       ██ ██       ██   ██   ██        ██   ██   ██       ██             ██
+  ██     ██ ██    ██ ██       ██    ██  ██        ██    ██  ██       ██       ██    ██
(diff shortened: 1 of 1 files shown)


inoas:
Ref: https://github.com/w3c/csswg-drafts/issues/442

nattarnoff:
This makes sense to me as solid way to implement in CSS. By default motion is on, and we can reduce action from there. I'm struggling to see how this gets implemented though in a real world example.
For instance, as a media query, I expect something like:
```
@medis (motion-preference: none {
  // Base CSS - no motion
}
@media (motion-preference: reduced) {
  // CSS for limited motion
}
 // Balance of CSS with full animation
````
Of course this is an ideal, and most folks will probably retrofit and use these in the opposite manner initially. Is there anyway this can be read and acted upon by WAAPI/JS animation implementations?

**Comment by cookiecrook** [comment]
Why split the conversation into this alternative PR? Please just continue in either of the original threads: #442 or #586.

**Comment by inoas** [comment]
The suggestion already is in #442 - precisely here: https://github.com/w3c/csswg-drafts/issues/442#issuecomment-249442652

However it wasn't picked and as far as I can see there wasn't any argument against it (yet!).

Do you see any issues with this PR here?
Does it not solve anything?
Is it anywhere unclear?

**Comment by cookiecrook** [comment]
@inoas wrote:

> as far as I can see there wasn't any argument against it

Your [last comment in that thread](https://github.com/w3c/csswg-drafts/issues/442#issuecomment-249442652) suggested `all` as a falsey value which doesn't seem right to me. 

You also seemed to agree with the "prefers-" prefix over your suggested "-preference" suffix with the following comment:

> the idea with the prefers- prefix seems to be a good one!

**Comment by cookiecrook** [comment]
@inoas wrote:
> Where would this not work for you?

I think the forced-all-animation-off idea (`motion-preference: disable;`) is unlikely to be implemented because it would break too many things, which is why I proposed the opt-in. I'd rather see that forced-all-animation-off idea incubated. If you get author and browser support, bring it back.

**Comment by cookiecrook** [comment]
As for the `motion-preference` versus `prefers-reduced-motion` property name, let's keep discussing that in PR #586.

**Comment by cookiecrook** [comment]
Also I think your proposal would not be backwards compatible:

```css
@media (motion-preference: all) /* this would be the default, but works in no browsers today */
```

cookiecrook:
I suggest closing this one and duping it to #586

inoas:
- I think it is really bad to design keywords around what is currently possible. If you think it is not the right time to support 'disabled' then either the browser-vendor can decide so or the PR can be restricted to values `default` and `reduced` leaving the *OPTION* to later have `disabled`. However personally I'd love to be able to know if a user does not like any animations. It is still a preference setting. E.g. you *can* still offer animations, but maybe .1s long instead of .3s.
- There is no `all` but only `default`, `reduced` and `disable`. Where would this not work in browsers?
- I am perfectly fine discussing in #586 I was just setting up a PR so that it isn't just words.

frivoal:

inoas:
Closing as there has been made a decision.

PR #793, opened 2016-12-06
Add 'prefers-motion' media feature

inoas:
Implements Issue: #442
Alternative to #586

Having a good wording of the feature also gets around default value evaluating to true instead of false. Ref: #586 (comment)


frivoal:

PR #808, opened 2016-12-15
Optimize images in the css22 folder

kaelig:
Saved 226.1KB out of 565.1KB. 33.9% per file on average (up to 68.5%)


kaelig:
@tabatkins you're right. The optimization gets rid of color profiles which can have this type of effect. Looking at the color changes, though, it seems like they are desirable.

The previous PR I submitted was this one: https://github.com/w3c/csswg-drafts/pull/740

(I can't remember submitting a PR to WHATWG specs directly)

PR #828, opened 2016-12-23
[css-containment] Set spec to level 1 rather than 3

frivoal:


SebastianZ:
The URL should also be adjusted accordingly. The spec. is still reachable under https://drafts.csswg.org/css-containment-3/ but not https://drafts.csswg.org/css-containment-1/.

Sebastian

PR #833, opened 2016-12-23
Change "line feed" to "segment break" in 4.1.2

upsuper:
It doesn't seem to me there is any reason we need to distinguish between "line feed" and the other two types of "segment break", so I suppose this is probably a typo, which could be misleading.


kojiishi:
Thank you for the PR, r+.

PR #840, opened 2016-12-27
[mediaqueries-4] More explicit clarification for when any-* 'none' evaluates to true

patrickhlauke:
Adds more explicit clarification about the possibly non-intuitive way in
which `any-pointer:none`/`any-hover:none` work.

Splits out the initial mention (that "ALL pointing devices need to
match") into its own line, and provides two sets of notes/examples that
clarify scenarios where this may not be completely obvious.

Additionally, this explicitly highlights the situation that currently
cannot be detected: having more than one pointing device where at least
one is non-hover-capable (as touched on in
https://github.com/w3c/csswg-drafts/issues/737#issuecomment-269348418 )

### DIFF
--- a/css-scroll-snap/Overview.bs
+++ b/css-scroll-snap/Overview.bs
@@ -5,6 +5,7 @@ Shortname: css-scroll-snap
 TR: https://www.w3.org/TR/css-scroll-snap-1/
 Level: 1
 Status: ED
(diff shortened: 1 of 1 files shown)


patrickhlauke:
apologies, only now noticed i made pull request from master, rather than a separate branch

PR #842, opened 2016-12-27
[mediaqueries-4] More explicit clarification for when any-* 'none' evaluates to true

patrickhlauke:
Adds more explicit clarification about the possibly non-intuitive way in
which `any-pointer:none`/`any-hover:none` work.

Splits out the initial mention (that "ALL pointing devices need to
match") into its own line, and provides two sets of notes/examples that
clarify scenarios where this may not be completely obvious.

Additionally, this explicitly highlights the situation that currently
cannot be detected: having more than one pointing device where at least
one is non-hover-capable (as touched on in
https://github.com/w3c/csswg-drafts/issues/737#issuecomment-269348418 )

### DIFF
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1759,10 +1759,41 @@ All Available Interaction Capabilities: the 'any-pointer' and 'any-hover' featur
 
 	The 'any-pointer' and 'any-hover' media features are identical to the 'pointer' and 'hover' media features,
 	but they correspond to the union of capabilities of all the pointing devices available to the user.
-	More than one of their values can match,
+	In the case of 'any-pointer', more than one of the values can match,
 	if different pointing devices have different characteristics.
-	They must only match ''none'' if <em>all</em> of the pointing devices would match ''none'' for the corresponding query,
+
+	'any-pointer' and 'any-hover' must only match ''none'' if <em>all</em> of the pointing devices would match ''none'' for the corresponding query,
 	or there are no pointing devices at all.
+
+	<div class="note">
+	'any-pointer' is used to query the presence and accuracy of pointing devices.
+	It does not take into account any additional non-pointing device inputs,
+	and can not be used to test for the presence of other input mechanisms,
+	such as d-pads or keyboard-only controls,
+	that don't move an on-screen pointer.
+	'any-pointer:none' will only evaluate to true if there are no pointing devices at all present.
+	</div>
+
+	<div class="example">
+	On a traditional desktop environment with a mouse and keyboard,
+	'any-pointer:none' will be false (due to the presence of the mouse),
+	even though a non-pointer input (the keyboard) is also present.
+	</div>
+
+	<div class="note">
+	'any-hover:none' will only evaluate to true if there are no pointing devices,
+	or if all the pointing devices present lack hover capabilities.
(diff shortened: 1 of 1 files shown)


darktears:
I would like to mention for posterity that this change had a potential compat impact on Chrome.

Historically any-hover:none would match in the case of a laptop with touchscreen and keyboard (because the touch screen doesn't have hover capabilities). We read literally the "union of capabilities" and maybe the spec wasn't very detailed at the time of implementation. That said thanks for the change and the clarification.

Interaction MQ are truly dynamic in Chrome since recently : we react to the best we can to input capabilities changes from the OS. Think the case of convertible/detachable laptops where the touchpad is gone/deactivated. With that said just to clarify, I'm assuming that a tablet where a mouse/touchpad is added (through USB/Proprietary connectors like Pixel or Bluetooth) should behave like a touchscreen laptop with touchpad/mouse? Aka any-hover:none should be false when some external input is connected and true when only the touchscreen is available to the user?

I'm aligning the implementation with the spec as we speak.

patrickhlauke:
@darktears 

> Historically any-hover:none would match in the case of a laptop with touchscreen and keyboard (because the touch screen doesn't have hover capabilities). We read literally the "union of capabilities" and maybe the spec wasn't very detailed at the time of implementation. That said thanks for the change and the clarification.

Do you mean you ignored the trackpad (assuming the laptop scenario had one)? Or do you mean that both `any-hover:none` (for the touchscreen) and `any-hover:hover` (for the trackpad) evaluated to true? From discussion in https://github.com/w3c/csswg-drafts/issues/737, I believe the latter was not the intention of the spec at the time - hence my clarification.

> I'm assuming that a tablet where a mouse/touchpad is added (through USB/Proprietary connectors like Asus Transformers or Bluetooth) should behave like a touchscreen laptop with touchpad/mouse? Aka any-hover:none should be false when some external input is connected and true when only the touchscreen is available to the user?

correct. `any-hover:none` should only evaluate to true if there are NO inputs at all that are hover-capable. so, for instance, a pure touchscreen device like a phone (but if you then pair a bluetooth mouse with the phone and it results in a hover-capable mouse pointer, like on android/windows 10 on mobile, then `any-hover:none` should be false due to the mouse being detected).

darktears:
> Do you mean you ignored the trackpad (assuming the laptop scenario had one)? Or do you mean that both any-hover:none (for the touchscreen) and any-hover:hover (for the trackpad) evaluated to true? From discussion in #737, I believe the latter was not the intention of the spec at the time - hence my clarification.

I meant in the laptop with touchscreen and trackpad/mouse any-hover:none (for the touchscreen) and any-hover:hover (for the trackpad) matched to true. Therefore the clarification of this PR slightly change the behavior because in that particular case any-hover:none would not evaluate to true anymore.

> correct. any-hover:none should only evaluate to true if there are NO inputs at all that are hover-capable. so, for instance, a pure touchscreen device like a phone (but if you then pair a bluetooth mouse with the phone and it results in a hover-capable mouse pointer, like on android/windows 10 on mobile, then any-hover:none should be false due to the mouse being detected).

Ok then we need to adjust our Android implementation as well.

patrickhlauke:
@darktears thanks for clarifying. yeah i believe the confusion (even for me at the time) comes from the word "union" in the spec, and how it treats the "none" in this case.

PR #870, opened 2017-01-09
[cssom] Don't serialize shorthand if importance differs

upsuper:
For the following declarations:
```css
div {
  background: green;
  background-image: none !important;
}
```
the current spec should return something for `getPropertyValue('background')`, but all browsers (even Servo) agree on that this should return nothing because the importance of its longhands differ.

This PR fixes this issue.

It can alternatively be fixed in "serialize a CSS value" algorithm. Given that the other callsite of this algorithm with a list (which is "serialize a CSS declaration block") already handles important flag itself, I guess it is better being put in `getPropertyValue` rather than "serialize a CSS value".

### DIFF
--- a/cssom/Overview.bs
+++ b/cssom/Overview.bs
@@ -1980,7 +1980,9 @@ The <dfn method for=CSSStyleDeclaration>getPropertyValue(<var>property</var>)</d
        <li>If <var>declaration</var> is null, return the empty string and terminate these steps.
        <li>Append the <var>declaration</var> to <var>list</var>.
       </ol>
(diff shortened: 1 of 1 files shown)


FremyCompany:
That makes sense to me.

zcorpan:
Thanks, looks good. Do you have a test case you can submit to web-platform-tests or csswg-test also? :-)

upsuper:
There already exists [a web-platform test](https://github.com/w3c/web-platform-tests/blob/master/cssom/shorthand-serialization.html) covers this.

Issue #867, opened 2017-01-08
[css-color-4] When do color space conversions happen?


frivoal:
Maybe I missed something, but as far as I can tell css-color-4 is silent about whether converting colors expressed in different profiles/spaces (using `color()`, `lab()`, `lch()`, or `gray()`) to the working color space  is something that happens at computed value time, used value time, or some other time, and how that affects serialization.

Given that for color properties, the resolved value returned by getComputedStyle() is the used value, the serialization should be the same if the conversion happens at computed or used value time: the converted value. However, when the working color space is (for instance) sRGB, if we expect `color(rec2020 0 255 127)` to serialize as `color(rec2020 0 255 127)` rather than as `rgb( x, y, z)` (or `color(srgb x y z)`) then supposedly, the conversion must happen after used value time.

On the other hand, if the conversion is handled after computed value time,  presumably this means we cannot animate from `color(foo 1 2 3)` to `color(bar 4 5 6)`.

I think that being able to animate between all sorts of colors is more valuable than a more literal serialization (and we already do all sorts of changes to the computed value anyway), so I'd lead towards doing the conversion at computed value time.

I'm guessing this is the intent of the spec already, but I have not found a clear statement one way or the other.

svgeesus:
You are correct, it needs to be described and it does affect serialization. This will go in the section on working colorspace, which has yet to be written.

svgeesus:
> On the other hand, if the conversion is handled after computed value time, presumably this means we cannot animate from color(foo 1 2 3) to color(bar 4 5 6).

No, that doesn't follow.

frivoal:
> > On the other hand, if the conversion is handled after computed value time, presumably this means we cannot animate from `color(foo 1 2 3)` to `color(bar 4 5 6)`.

> No, that doesn't follow.

As I understand, animations and transitions work on computed values. If the computed value is not in the same color space, how can you animate? what is 25% of the way between `color(foo 0 0 0)` to `color(bar 0 0 100)`? 25% of the way between 0 and 100 is 25, but 25% of the way between foo and bar is what?

grorg:
I think the transitioning question should be separate #883

svgeesus:
> As I understand, animations and transitions work on computed values. If the computed value is not in the same color space, how can you animate? 

[...]

Issue #741, opened 2016-11-21
[css-ui] Computed value of currentColor


frivoal:
[css-ui-3 defines the computed value of caret-color this way](https://drafts.csswg.org/css-ui-3/#propdef-caret-color):

> The computed value for auto is auto. For <color> values, the computed value is as defined for the [CSS3COLOR] color property.

The intent was (among other things) to have `currentColor` compute to `currentColor` as a keyword, but it is not clear if this wording is good to unambiguously define it that way.

In [css-color-3](https://drafts.csswg.org/css-color-3/#currentColor-def), we're told that:
> The used value of the ‘currentColor’ keyword is the computed value of the ‘color’ property. If the ‘currentColor’ keyword is set on the ‘color’ property itself, it is treated as ‘color: inherit’. 

The first sentence says that `currentColor` is resolved at used value time, which implies that it stays as a keyword at computed value time. “treated as” in the second sentence implies that at computed value time on the `color` property, `currentColor` has been replaced with the computed value of the `color` property on the parent element.

Despite the fact that CSS-UI invokes the computed value logic of the color property, it wouldn't make much sense for `currentColor` on `caret-color` to “be treated as inherit”, but it would be possible to read it as `currentColor` being replaced by an actual color value at computed value time. Given that `caret-color` is an inherited property, this would lead to a different (undesirable) behavior than preserving it as a keyword at computed value time.

[css-color-4 also agrees](https://drafts.csswg.org/css-color/#currentcolor-color). It says the following about currentColor:
> if the value is inherited, it’s inherited as currentcolor, not as the value of the color property, so descendants will use their own color property to resolve it.

Since the intended meaning in css-ui-3 and the meaning inferred by color-4 agree: `currentColor` should computed to `currentColor` as a keyword. I suggest making this unambiguous in css-ui-3 by changing the Computed Value entry in the propdef table to:

> The computed value for auto is auto, and for currentColor is currentColor. For other <color> values, the computed value is as defined for the color property.

frivoal:
Agenda+. I expect this to be completely uncontroversial, but it is a change to a CR, so a group decision is warranted.

frivoal:
WG was in favor of this, but wanted a better wording that could more easily be reused across specs that have the same issue. That wording should be anchored on definitions / alogrithms / rules / etc found in level 3 of the color spec. If necessary (e.g. the wording we want is in level 4 or not written yet) we can make errata to color 3.

tabatkins:
[...]

Issue #480, opened 2016-09-14
[css-color] Serializing color() values


grorg:
How should values defined using color() be serialized in computed style?
1. Should color(a b c) or color(srgb a b c) values be output as rgb()?
2. Should color(someprofile a b c) values that end up inside sRGB be output as rgb()?

SebastianZ:
I tend to agree with [Tab's statement to treat `rgb()` as legacy syntax](https://github.com/w3c/csswg-drafts/issues/266#issuecomment-230898208). Therefore I think it should compute to `color(a b c)` in those cases.

Sebastian

svgeesus:
I agree with Tab and SebastianZ

grorg:
Treating it as legacy syntax is fine for input. But the output from computed style is probably parsed by a lot of content on the Web. If that all of a sudden changes to something unexpected, the world will collapse, the seas will rise, and Trump may get elected.

upsuper:
Agree with @grorg, unless someone can prove that changing this output is web-compatible.

grorg:
Wow. I'm regretting my last statement now.

cabanier:
:-( 
It's all your fault!

frivoal:
Reposting a comment I made in #742 as I am closing that one as duplicate:

---

[The css-color-4 definition of the computed value of the color property](https://drafts.csswg.org/css-color/#propdef-color) says :

> Computed value: 	an RGBA color

That doesn't sound right:

* This is not entirely clear what is meant by that. The simplest interpretation is that the value should be computed to the `rgba()` function, but maybe something else is intended.
* If computing to `rgba()` is intended, this appears to contradicts [css-color-3](https://drafts.csswg.org/css-color-3/#color0). The rule there is phrased in a somewhat ambiguous way, but seems to imply that basic colors, hex colors, and `rgb()` colors should be computed to the six digit hex value or `rgb()` functional value, while other notations should be preserved as specified.
* Browsers today (checked Chrome, Firefox, Safari, Edge, Presto) interoperably return `rgb()` (not `rgba()`) as the computed value of hex colors or named colors when exposing it through getComputedStyle. In contradiction of css-color-3, All also do so for `hsl()` colors. And but all but Edge also for `rgba()` and `hsla()` colors with an alpha of 1 (Edge compute both to `rgba()`).
* How to convert `color()` colors to `rgba()` is non trivial and lossy. It needs to happen eventually to be able to paint, but computed value time seems wrong.

[...]

PR #871, opened 2017-01-09
Resolving colors

frivoal:
This is a proposal to resolve the lack of a generally usable definition for how colors, and in particular `currentcolor` are turned into computed values, used values, inherited...

This should help address #741, #480, and #867 (and maybe other issues I am not aware of).

If we agree on the concept, we should probably try to backport the relevant subset of this to css-colors-3. I can do this as an update to this PR or as a separate one, as the WG / editors prefer.

I have for now left undefined (and marked as an issue) the computed values of the functions that are not yet implemented. We should have that discussion as well, but it need not delay the introduction of the general concept and anchoring terminology.

For the functions that are implemented, the definition I provided is based on my observations of current inter-operable behavior.

### DIFF
--- a/css-color/Overview.bs
+++ b/css-color/Overview.bs
@@ -184,7 +184,7 @@ Resolving Color values</h2>
 		so descendants will use their own 'color' property to resolve it.
 
 		<li>
-		The computed value and used value of ''<color>/transparent'' is ''rgba(0,0,0,0)''.
+		The computed value and used value of ''<color>/transparent'' is ''rgba(0, 0, 0, 0)''.
 
 		Issue: Gecko disagrees, and serializes any <<color>> with an alpha channel of 0 as transparent.
 		No other browser does that though.
--- a/css-color/Overview.bs
+++ b/css-color/Overview.bs
@@ -368,7 +368,7 @@ Named Colors</h2>
 	Most of the rest
 	come from one version of the X11 color system,
 	used in Unix-derived systems to specify colors for the console.
-	(Two special color values, ''transparent'' and ''currentcolor'',
+	(Two special color values, ''transparent'' and ''<color>/currentcolor'',
 		are specially defined in their own sections.)
 
 	Note: The history of the X11 color system is interesting,
@@ -705,7 +705,7 @@ The ''<color>/currentcolor'' keyword</h3>
 	Its <a>computed value</a> and <a>used values</a> are determined by <a>resolving color values</a>.
 
 	<div class='example'>
-		Here's a simple example showing how to use the ''currentcolor'' keyword:
+		Here's a simple example showing how to use the ''<color>/currentcolor'' keyword:
 
 		<pre class="lang-css">
 		.foo {
@@ -726,7 +726,7 @@ The ''<color>/currentcolor'' keyword</h3>
 
 	<div class='example'>
 		For example, the 'text-emphasis-color' property [[CSS3-TEXT-DECOR]],
-		whose initial value is ''currentcolor'',
+		whose initial value is ''<color>/currentcolor'',
 		by default matches the text color
 		even as the 'color' property changes across elements.
 
@@ -744,7 +744,7 @@ The ''<color>/currentcolor'' keyword</h3>
 	</div>
 
 	Note: Multi-word keywords in CSS usually separate their component words with hyphens.
-	''currentcolor'' doesn't, because it was originally introduced in SVG
+	''<color>/currentcolor'' doesn't, because it was originally introduced in SVG
 	as a special attribute value spelled "currentColor",
 	rather than a CSS value.
 	Only later did CSS pick it up,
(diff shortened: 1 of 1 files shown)


frivoal:
Agenda+ to discuss if this is the way to go, and if tweaks are needed.

css-color/Overview.bs
```diff
@@ -161,6 +160,75 @@ Notes On Using Colors</h3>
...
+		comma separated ''rgb()'' colors without an alpha channel,
+		and comma separated ''hsl()'' colors without an alpha channel,
+		is the equivalent numeric value in comma separated ''rgb()'' notation omitting the alpha value.
+
+		<li>
+		The computed value and used value of
+		4 and 8 digits <a>hex colors</a>,
+		comma separated ''rgba()'' colors with an alpha channel,
+		and comma separated ''hsla()'' colors with an alpha channel,
+		is the equivalent numeric value in comma separated ''rgba()'' notation with the alpha value.
```
upsuper: This doesn't match what browsers currently do. Browsers use `rgb()` for opaque colors even if the color is specified from `rgba()`. See http://software.hixie.ch/utilities/js/live-dom-viewer/saved/4780

css-color/Overview.bs
```diff
@@ -161,6 +160,75 @@ Notes On Using Colors</h3>
...
+		the used value of ''<color>/currentcolor'' is the same as
+		the used value of the 'color' property on the same element.
+
+		Note: This means that if the ''<color>/currentcolor'' value is inherited,
+		it’s inherited as a keyword,
+		not as the value of the 'color' property,
+		so descendants will use their own 'color' property to resolve it.
+
+		<li>
+		The computed value and used value of ''<color>/transparent'' is ''rgba(0,0,0,0)''.
```
upsuper: Should be `rgba(0, 0, 0, 0)`.

FWIW, it seems for all colors whose alpha channel is zero, Gecko returns `transparent` as the resolve value, which may make some sense given they are effectively the same for rendering. But other browsers' behavior that returns the `rgba` value seems to be better regarding consistency.

frivoal: I am not a huge fan of that behavior, but that's indeed what gecko does. Added an issue to call it out, so that we don't loose track of this difference and remember to resolve it later.

css-color/Overview.bs
frivoal: Thanks. Added a new commit to reflect that.

css-color/Overview.bs
```diff
@@ -60,12 +60,11 @@ Foreground Color: the 'color' property</h2>
 	Inherited: yes
 	Percentages: N/A
 	Media: visual
-	Computed value: an RGBA color
+	Computed value: See <a>resolving color values</a>
```
dbaron: I think this is actually wrong for the <code>color</code> property, since for <code>color</code>, <code>currentColor</code> should be resolved at computed value time.  (I can't think of how this is observable today, but I'd much prefer it be defined this way.)

css-color/Overview.bs
upsuper: Oh, probably I should use "BTW" instead of "FWIW" here. The main issue I want to raise here is a small editoral issue that there should be a whitespace after each comma. `rgba(0,0,0,0)` without whitespaces doesn't match any browser, and could cause confusing.

css-color/Overview.bs
frivoal: I've fixed the definition now, so this reference should now be fine.

css-color/Overview.bs
```diff
@@ -691,7 +758,7 @@ The ''currentcolor'' keyword</h3>
 	</div>
 
 	Note: Multi-word keywords in CSS usually separate their component words with hyphens.
-	''currentcolor'' doesn't, because it was originally introduced in SVG
+	''<color>/currentcolor'' doesn't, because it was originally introduced in SVG
 	as a special attribute value spelled "currentColor",
```
svgeesus: No, it was introduced as a property value (and properties in SVG can be expressed in presentational attributes, as well as in style attributes, style elements, and external stylesheets).

css-color/Overview.bs
```diff
@@ -184,7 +184,7 @@ Resolving Color values</h2>
 		so descendants will use their own 'color' property to resolve it.
 
 		<li>
-		The computed value and used value of ''<color>/transparent'' is ''rgba(0,0,0,0)''.
+		The computed value and used value of ''<color>/transparent'' is ''rgba(0, 0, 0, 0)''.
```
svgeesus: Maybe worth calling out explicitly that the serialization syntax (with commas) differs - for historical reasons - from the comma-less syntax which CSS4 color is now using (commas now being reserved for lists of alternatives).

svgeesus:
Merging because reviews seem generally positive and review comments have been addressed. My comments about inline issues can easily be fixed later and should not hold up a merge.

PR #882, opened 2017-01-11
[cssom][css-transforms] Allow other specs to specify 'resolved value'

### LINKED ISSUE #392: [cssom][css-transforms] Conflicting on resolved value of various properties
CSSOM has a list of properties which should have their resolved value be their used value rather than computed value. That list doesn't include `perspective-origin`. However, CSS Transforms spec says the resolved value of `perspective-origin` should be used value.

I guess one of them should be changed to match the other.

zcorpan:
Fixes #392.


zcorpan:
cc @upsuper @smfr

upsuper:
It looks fine to me. (I wonder whether there is an easy way to find out what spec defines other special cases. But it is probably not a big issue I guess.)

PR #848, opened 2017-01-01
[css-values] Adding new units: en, cap, bs/hs; m/dm/mu, dd/cc, thm/mdl, dot, yd/ft/hd, sx/tx/twip, ax/ay/bx/by/cx/cy; tip/tap/mark/pad/span/sole; moa/soa, pi, perm; min/h, now/mom, beat/blink, jif; bpm

Crissov:
My git-fu isn’t strong enough yet to split this nicely into separate pull requests.

# Length units

## Viewport-relative length units

* `vi` and `vb` logical viewport units. – were already included (#113, #323)

## Relative length units

* Add the [`ic` and] `en` advance units. (#357; #315)
* Add the `cap` unit. (#660, #315, #661)
* Add the `bs` and `hs` stroke units. (#239, #459, #838)

## Absolute length units

* Add the `m`, `dm` and `mu` metric units. (#315)
* Add the `dd` and `cc` metric typographic units. (#315)
* Add the `thm`, `mdl` metric English units.
* Add the `dot` greatest common divisor unit. (#315) 
* Add the `yd`, `ft` and `hd` English units. (#315)
* Add the `sx`, `tx` and `twip` English fractional typographic units. (#315, #378)
* Add the `ax`, `ay`, `bx`, `by`, `cx` and `cy` paper units. (#328, #378)

## Anthropometric/physiological length units 

* Add the `tip`, `tap`, `mark`, `pad`, `span` and `sole` anthropometric units. (#614, #837)

# Angle units

* Add the `moa` and `soa` angular units.
* Add the `pi` circular unit. (#309)
* Add the `perm` slope unit.

# Duration units

* Add the classic `min` and `h` units.
* Add the psychological `now` and `mom` units. (#837)
* Add the physiological (heart) `beat` and (eye) `blink` units. (#837)
* Add the technical `jif` unit.

# Frequency units 

* Add the classic `bpm` unit.

Issue #660, opened 2016-10-28
[css-values] cap-height unit?


tbredin:
I've just found out about the `ch` and `ex` units described here:
https://www.w3.org/TR/css3-values/#relative-lengths

Cap height: This is something that could become extremely useful in vertical rhythm projects such as [Sassline](https://github.com/jakegiltsoff/sassline) and [MegaType](https://github.com/studiothick/megatype), which go to great length to attempt to find this height through other means (plus a lot of manual config adjustment). When combined with `ex` (and a speculative [baseline position](https://github.com/w3c/csswg-drafts/issues/661)), we also start to build up a very useful picture of a font's vertical metrics.  

The best measure of this is usually the height of a capital `H` in any font, but could likely also fall back to `I` or `X`. 

Is this something that might be considered in future?

[Further reading on some existing issues with typographic metrics on the web](http://www.studiothick.com/essays/web-typography-is-broken)

simevidas:
Google has proposed (and is implementing) the [CSS Step Sizing module](https://www.chromestatus.com/feature/5734273533345792), which enables “vertical rhythm using CSS”. What do you think? 😊

SelenIT:
There is also the [CSS Line Grid module](https://drafts.csswg.org/css-line-grid/) that also provides a CSS mechanism for vertical rhythm (including baseline snapping to regular grid lines).

But I believe that having better access to the font metrics from CSS would have more use cases other than that. For example, it can be aligning the icon with text (both top and bottom) or removing extra vertical spacing before text (e.g. in the approach described [in this article](https://medium.com/eightshapes-llc/space-in-design-systems-188bcbae0d62#9406)) without having to use magic numbers for each particular font.

fantasai:
RESOLVED: add cap unit to Values and Units Level 4

tabatkins:
Added.

karlhorky:
Thanks for this proposal for a `cap` unit for allowing font-sizing via the capital heights, really cool!

Does anyone know of any browsers that have considered or started implementation of this? I've been looking for tracking bugs, but haven't been successful...

Crissov:
Apparently none yet. https://caniuse.com/mdn-css_types_length_cap

SebastianZ:
I've now [created a feature request](https://bugzilla.mozilla.org/show_bug.cgi?id=1702924) to implement it in Gecko (Firefox).

Sebastian

karlhorky:
[...]

PR #902, opened 2017-01-13
[css-values] cap unit and copyedit

Crissov:
This is the trimmed down version of #848. It just adds `cap` #660, rearranges relative length units, fixes several minor mistakes and adds some line breaks to better manage possible future unit additions.


tabatkins:
You've mixed together at least four different commits here, I think.  It makes it much more difficult to plumb history in the future when a commit does so many different things at once.  Please submit only the cap unit PR; the other PRs can be done separately and are likely to be merged faster.

Crissov:
Yeah, I’m still struggling a bit with git, but shouldn’t you be able to fix that with a [squash merge](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits)?

fantasai:
No, because we'd still have to separate out the changes: you've mixed up whitespace changes, a new feature, rearranging paragraphs, adding annotations, and a whole bunch of other changes into one PR. These each need to be separate commits. I can split these out manually, but in the future don't do this. I'd rather you emailed me separate patches if you can't figure out git than send a pile of tangled up commits as a PR.

tabatkins:
I'll go ahead and close this; it can't be submitted as written, and while it could be cut down to only one change and then have a squash merge done, that can just as easily be done in a new PR.

PR #956, opened 2017-01-18
replaced 2015 with 2017

SelenIT:


frivoal:
Nice catch, thank you.

PR #959, opened 2017-01-19
[css-line-grid] Make the dependency on line-height-step explicit.

### LINKED ISSUE #491: [css-logical-props] `border‐radius`
`border‐radius` was left out of [css-logical-props](https://drafts.csswg.org/css-logical-props/) for no aparent particularly good reason.

I think the following properties should be present in the spec:

```
border-start-start-radius
```

```
border-start-end-radius
```

```
border-end-start-radius
```

```
border-end-end-radius
```

The first part would be related to the block axis and the second part to the inline axis.

And alternative would be `border-block-start-inline-start-radius`, but that feels unnecessarily verbose.

### LINKED ISSUE #939: [css-rhythm][css-line-grid] Work out the interaction between line-height-step and line-grid
If we're not careful, [the line-height-step property](https://drafts.csswg.org/css-rhythm/#line-height-step) and the [line grid](https://drafts.csswg.org/css-line-grid/) will interfere instead of complementing each other.

The grid created by`line-grid:create` needs to take `line-height-step` into account. As far as I can tell, [the current wording](https://drafts.csswg.org/css-line-grid/#valdef-line-grid-create) already does that implicitly, but we should probably make it explicit to avoid any misunderstanding.

frivoal:
~~Closes #491~~ (that was a typo)

Closes #939

### DIFF
--- a/css-line-grid/Overview.bs
+++ b/css-line-grid/Overview.bs
@@ -234,7 +234,8 @@ Defining a <dfn>Line Grid</dfn>: the 'line-grid' property</h2>
 
 		<dt><dfn>create</dfn>
 		<dd>Box creates a new <a>line grid</a>
-			using its own font and line layout settings.
+			using its own font and line layout settings,
+			including any adjustment to the line height caused by the 'line-height-step' property.
 			The <a>line grid</a> consists of a series
 			of horizontal lines corresponding
 			to all the baselines


SebastianZ:
@frivoal So, you should reopen #491.

Sebastian

PR #969, opened 2017-01-23
[css-transforms-2] Individual transform properties animate with none

### LINKED ISSUE #968: [css-transforms-2] Individual transform animation to and from none
Animate of the 'translate', 'rotate' and 'scale' should be smooth, even when animating to or from 'none'.

ewilligers:
When 'translate', 'rotate' or 'scale' are animating or transitioning, and
the from value or to value (but not both) is none, the value none is
replaced by the equivalent identity value (0px for translate, 0deg for
rotate, 1 for scale).

resolves #968

### DIFF
--- a/css-transforms-2/Overview.bs
+++ b/css-transforms-2/Overview.bs
@@ -590,6 +590,9 @@ while all other values
 create a stacking context and containing block,
 per usual for transforms.
 
+When 'translate', 'rotate' or 'scale' are animating or transitioning, and the from value or to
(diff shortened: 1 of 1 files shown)


smfr:
Is "an independent transform property" defined? I would prefer to list the properties explicitly.

ewilligers:
> I would prefer to list the properties explicitly.

Done.

smfr:
Thanks.

Issue #967, opened 2017-01-21
[mediaqueries-4] do `hover` and `any-hover` only make sense for *pointer-based* input mechanisms?


patrickhlauke:
Following the discussions in https://github.com/w3c/csswg-drafts/issues/737 and taking a step back, I think I may have had a slight shift in perception on how I could reconcile my most fundamental/philosophical issues with what the spec seems to be trying to do...

One of the scenarios that I felt very uncomfortable with for a long time was the fact that keyboard-type inputs aren't sufficiently covered by the spec. My fear in this context is that authors may do something like querying `pointer:none`, `hover:none`, `any-pointer:none`, `any-hover:none` and if those did *not* evaluate to true, assume that there was no keyboard (or similar non-pointer, non-hover-capable input).

Now in the case of `pointer` / `any-pointer`, I've been able to reconcile this in https://github.com/w3c/csswg-drafts/pull/842 with the note I added

> 'any-pointer' is used to query the presence and accuracy of pointing devices.	It does not take into account any additional non-pointing device inputs, and can not be used to test for the presence of other input mechanisms, such as d-pads or keyboard-only controls, that don't move an on-screen pointer. 'any-pointer:none' will only evaluate to true if there are no pointing devices at all present.

I'm now wondering if something similar (clarifying text/wording) could also help in the case of `hover` / `any-hover`, and this would depend on the following: does the concept of "hovering" only really make sense in the context of pointer-based inputs? Inputs that have an on-screen or physical pointer/cursor?

I can't think of a situation where an input *would* or *could* be hover-capable if it didn't have the concept of a pointer (and this also seems to be reflected in the table for https://drafts.csswg.org/mediaqueries-4/#mf-interaction). Is the very idea of hovering perhaps intrinsically tied to having a "pointer", and therefore not applicable at all in the context of non-pointer-based inputs?

And if we can agree on that, perhaps the cleanest way to address my concern about keyboard/keyboard-like input mechanisms not being covered would be to, instead of talking about "input mechanisms" in the spec (which is quite a broad, generic term, which *would* cover keyboards), the spec text should quite explicitly talk about something like "pointer-based input mechanisms". By using this more specific language, it would be immediately clear that these media features are not even *meant* to help authors detect if there are other input mechanisms, like keyboards, present. pointer-accuracy and hover-capability would only apply to any pointer-based inputs, and authors would still need to use their judgement to know that there may be other non-pointer-based input mechanisms present (and this could be reinforced in a note).

[...]

PR #973, opened 2017-01-23
[mediaqueries-4] make explicit that `pointer`, `hover`, `any-pointer`, `any-hover` relate only to pointing devices

patrickhlauke:
Closes https://github.com/w3c/csswg-drafts/issues/967

### DIFF
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1609,6 +1609,12 @@ Interaction Media Features</h2>
 	in response to changes in the user environment 
 	or in the way the user is interacting with the UA.
 
+	Note: The 'pointer', 'hover', 'any-pointer' and 'any-hover' features only relate to the characteristics,
+	or the complete absence, of pointing devices,
+	and can not be used to detect the presence of non-pointing device input mechanisms such as keyboards.
+	Authors should take into account the potential presence of non-pointing device inputs,
+	regardless of which values are matched when querying these features. 
+
 <h3 id="pointer">
 Pointing Device Quality: the 'pointer' feature</h3>
 
@@ -1788,6 +1794,9 @@ All Available Interaction Capabilities: the 'any-pointer' and 'any-hover' featur
 	As such, it should be understood as a query to test if any hover-capable pointing devices are present,
 	rather than whether or not any of the pointing devices is hover-incapable.
 	The latter scenario can currently not be determined using 'any-hover' or any other interaction media feature.
+	Additionally, it does not take into account any non-pointing device inputs,
+	such as d-pads or keyboard-only controls,
+	which by their very nature are also not hover-capable.
 	</div>
 
 	<div class="example">
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1604,7 +1604,10 @@ Interaction Media Features</h2>
 	the expectation is that User Agents should make this determination 
 	by combining knowledge about the device/environment they are running on, 
 	the number and type of pointing devices available, 
-	and a notion of which of these is generally and/or currently being used. 
+	and a notion of which of these is generally and/or currently being used.
+	In situations where the primary input mechanism for a device is not a pointing device,
+	but there is a secondary – and less frequently used – input that is a pointing devices,
+	the user agent may decide to treat the non-pointing device as the primary (resulting in 'pointer: none').
 	User Agents may also decide to dynamically change what type of pointing device is deemed to be primary, 
 	in response to changes in the user environment 
 	or in the way the user is interacting with the UA.
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -1597,15 +1597,15 @@ Interaction Media Features</h2>
 		</style>
 	</div>
 
-	The 'pointer' and 'hover' features relate to the characteristics of the “primary” input mechanism, 
-	while 'any-pointer' and 'any-hover' can be used to query the properties of all potentially available input mechanisms.
+	The 'pointer' and 'hover' features relate to the characteristics of the “primary” pointing device, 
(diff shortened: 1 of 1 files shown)


patrickhlauke:
While in #967 i proposed "pointer-based input mechanisms", i saw that the spec already used "pointing devices" as a phrase, so for consistency went with this one...

inoas:
OT: So is iOS-Safari not behaving strict to the spec when it render elements you tap onto as having `:hover` triggered? Ref: http://codepen.io/anon/pen/jywdxM

patrickhlauke:
yours is a question more generally about the spec, and not about something in this PR, but:

iOS uses heuristics to make `:hover` apply under certain conditions after a user has tapped the screen, but it's undisputable that the touchscreen itself does not support actual hovering, and not every tap (e.g. on non-actionable elements) results in a hover behavior either.

see https://drafts.csswg.org/mediaqueries-4/#hover in reference to `hover: none`:

> Pointing systems that can hover, but for which doing so is inconvenient and not part of the normal way they are used, also match this value

iOS' faked hovering would fall under this category.

In any case, this spec does not say either way when user agents should or shouldn't apply `:hover`...it specifies the `hover` media feature itself, which reflects the capabilities of the pointing device.

inoas:
OT/Noise: Maybe it could be helpful for the future to specify that touch-only (without 3d-touch / pressing sensitivity at least) devices are no pointing devices (as there is no real way to distinquish between pointing and clicking/tapping)...

The spec you are correcting also speaks about multiple devices. It should trigger `:hover` for the pointing device of highest priority, not just any primary device (imagine touch+pointing on the same screen, touch being the primary device?). It could also explicitly exclude non-pointing devices such as (2d) touch.

However it would be great to have `:tap` or `:touch` to be able differentiate.

patrickhlauke:
i think you're misunderstanding how the term "pointing device" is used here in the spec (and elsewhere, such as in https://w3c.github.io/pointerevents/). nothing to do with "pointing" as separate from potential activation.

If you do want to propose a new distinction/value for the media features here, i'd suggest opening a new issue.

Lastly, this spec does not deal with defining pseudo-classes like `:hover` or the hypothetical `:tap` / `:touch` you propose, so suggest actually going to https://drafts.csswg.org/selectors-4/ and proposing things there...

inoas:
Thanks for clarification! and the link to pointerevents

patrickhlauke:
as much as i hate pestering... @frivoal does this look ok to merge (based on your positive comment https://github.com/w3c/csswg-drafts/issues/967#issuecomment-274419675) ?

patrickhlauke:
bumping this for @tabatkins' consideration as well...

tabatkins:
Looks pretty reasonable to me. @frivoal, any objections to merging?

frivoal:
This is probably good, but I wonder: this is fine for `any-*` but what if the input mechanism deemed primary is not a pointing device? Say, a device with a keyboard deemed primary, and a gamepad. Should `pointer` match `none` because that's the case of the keyboard, or should it match `coarse`, because within the pointing devices, the gamepad, by virtue of being the only one, is necessarily primary, even if overall it wouldn't be?

This is possibly enough of a corner case not to matter, but I am not sure.

mediaqueries/Overview.bs
```diff
@@ -1597,18 +1597,24 @@ Interaction Media Features</h2>
...
-	Note: While this specification does not define how User Agents should decide what the “primary” input is, 
+	Note: While this specification does not define how User Agents should decide what the “primary” pointing device is, 
 	the expectation is that User Agents should make this determination 
 	by combining knowledge about the device/environment they are running on, 
-	the number and type of input mechanisms available, 
-	and a notion of which of these inputs is generally and/or currently being used. 
-	User Agents may also decide to dynamically change what type of input is deemed to be primary, 
+	the number and type of pointing devices available, 
+	and a notion of which of these is generally and/or currently being used. 
+	User Agents may also decide to dynamically change what type of pointing device is deemed to be primary,
```
patrickhlauke: could add an extra example right here: "In situations where the primary input mechanism for a device is not a pointing device, but there is a secondary – and less frequently used –  input that is a pointing devices, the user agent may decide to keep treating the non-pointing device as the primary (resulting in `pointer: none`), or they may opt to "promote" the secondary input to be identified as the primary pointing device (resulting in `pointer: fine` or `pointer: coarse`).

patrickhlauke:
> what if the input mechanism deemed primary is not a pointing device? Say, a device with a keyboard deemed primary, and a gamepad. Should pointer match none because that's the case of the keyboard, or should it match coarse, because within the pointing devices, the gamepad, by virtue of being the only one, is necessarily primary, even if overall it wouldn't be?

i'd say that's the remit of the UA to decide in that case. i'd tend to lean towards "if the UA deems the keyboard to be the primary input mechanism, and the gamepad is only used very infrequently, then there's no primary pointer, but only an `any-` pointer input. what about adding an extra example to that first note? (just added it as a comment)

basically...up to the UA, and this spec won't say what's right and it's up to the UA. or is that too handwavy?

mediaqueries/Overview.bs
frivoal: Sounds good. I'd shorten it to this:

In situations where the primary input mechanism for a device is not a pointing device, but there is a secondary – and less frequently used – input that is a pointing devices, the user agent may decide to treat the non-pointing device as the primary (resulting in `pointer: none`).

If you can update the pull request with that, I'll merge.

patrickhlauke:
urgh, i seem to have completely messed up this pr by doing a rebase. let me see if i can fix that...

mediaqueries/Overview.bs
patrickhlauke: Done. Agree the terse versions works just as well.

PR #978, opened 2017-01-24
[css-color] Editorial: correct case for JS 'Math'

zcorpan:


SebastianZ:
The [`Math` object](http://www.ecma-international.org/ecma-262/7.0/index.html#sec-math-object) does currently not provide functions like `matrix()` or `multiply()`.

Sebastian

zcorpan:
Oh... right... But the sample code here doesn't provide implementations of those, either?

Closing this since the suggested change is bogus.

SebastianZ:
> Oh... right... But the sample code here doesn't provide implementations of those, either?

Correct. Those would need to be added to the sample code.

Sebastian

svgeesus:
> But the sample code here doesn't provide implementations of those, either?

Right, that is why it uses the math library to do so :) since that library does provide them. You can look at the code easily, it is in the repo as math.js

https://github.com/w3c/csswg-drafts/tree/master/css-color

PR #994, opened 2017-01-31
[cssom-view] Return real values for colorDepth and pixelDepth (closes #993)

### LINKED ISSUE #993: [cssom-view] Make colorDepth/pixelDepth return useful information again
**Context:** this is related to the [Screen](https://drafts.csswg.org/cssom-view/#screen) interface, in particular the [colorDepth](https://drafts.csswg.org/cssom-view/#dom-screen-colordepth) and [pixelDepth](https://drafts.csswg.org/cssom-view/#dom-screen-pixeldepth)

**Some bug history:**
* https://www.w3.org/Bugs/Public/show_bug.cgi?id=17522
* https://www.w3.org/Bugs/Public/show_bug.cgi?id=14071

I am suggesting that we do the following changes:
- `colorDepth` and `pixelDepth` continue to return the same value;
- `colorDepth` and `pixelDepth` should return the number of bits allocated to colors on the output device or an approximation to the best of the UA's knowledge;
- 24 is to be used as a default value for privacy considerations or when the UA can't find the value;
- the `color` MQ should return a consistent values with `colorDepth` and `pixelDepth` (eg. 8 when the properties return 24 or 10 when 30 is returned by the properties)

We could also recommend 24 to be used if the value would be below 24. However, Blink already exposes monochrome in the `color` MQ so there might be benefit of keeping consistency between APIs in some cases where the value would be below 24.

The purpose for this change is for websites to be able to detect whether it is worth providing content that would require high color depth output.

The main concern would be fingerprinting. The treat seems fairly weak compared to the current fingerprinting surface on the Web. It will most likely add one or two bits of entropy (24/30/48) which can probably be linked to other properties of the display (CSS color gamut, screen size, etc), thus likely reducing the practical benefit for fingerprinting.

mounirlamouri:
Fixes #993

### DIFF
--- a/cssom-view/Overview.bs
+++ b/cssom-view/Overview.bs
@@ -81,6 +81,7 @@ urlPrefix: http://heycam.github.io/webidl/
         text: long
     type: dfn; text: converted to an IDL value; url: dfn-convert-ecmascript-to-idl-value
 url: https://www.w3.org/TR/DOM-Level-3-Events/#interface-MouseEvent; type: interface; text: MouseEvent
+url: https://drafts.csswg.org/mediaqueries-4/#mf-colors; type: dfn; text: color media query
 </pre>
 
 <pre class='link-defaults'>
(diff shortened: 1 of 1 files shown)


mounirlamouri:
@zcorpan PTAL

mounirlamouri:
Thank you for the comments @dbaron. I have applied them apart from the example one: do you have recommendations for a better, less misleading, example? In practices, I would be surprised if the non-comforming implementations also implement the css color gamut which hopefully makes the example not insane in practice but I would be happy to put something else in the spec :)

dbaron:
While that example itself is fine because such implementations wouldn't match the p3 media query, people might start from that code and modify it to do other things.

Would the high-gamut devices report 48?  If so, maybe it's better to test >= 48?

mounirlamouri:
I've updated the example to use `>= 48`. It actually makes a lot of sense :)

I've also removed the 24 clamping following discussion with @dbaron and @zcorpan on IRC.

cssom-view/Overview.bs
```diff
@@ -896,13 +897,35 @@ screen area</a>. <!--fingerprint-->
...
 
-The <dfn attribute for=Screen>colorDepth</dfn> attribute must return 24.
+The <dfn attribute for=Screen>colorDepth</dfn> and <dfn attribute
+for=Screen>pixelDepth</dfn> attributes should return the number of bits
+allocated to colors for a pixel in the output device, excluding the alpha
+channel. If the user agent is not able to return the number of bits used by the
+output device, it should return the closest estimation such as, for example, the
+number of bits used by the frame buffer sent to the display or any internal
+representation that would be the closest to the value the output device would
+use. The user agent must return the same value for these attributes and the
```
zcorpan: Same value seems wrong. The color media query is per component, whereas here is all components together. There's also this complication:

> If different color components are represented by different number of bits, the smallest number is used.

I suppose an integer-divide by 3 should give the same value as the color media query?

mounirlamouri:
@zcorpan, I've rephrased. WDYT?

zcorpan:
Looks good to me.

mounirlamouri:
@zcorpan I assume we need @dbaron's lgtm now? :) Is there someone else who should have a look?

zcorpan:
Final review from @dbaron would be nice.

PR #1019, opened 2017-02-10
[cssom][css-grid][css-transforms] Make 'transform-origin' and 'perspe…

### LINKED ISSUE #392: [cssom][css-transforms] Conflicting on resolved value of various properties

zcorpan:
…ctive-origin' resolved value like 'height'

Fixes #392.

### DIFF
--- a/css-grid/Overview.bs
+++ b/css-grid/Overview.bs
@@ -1719,6 +1719,9 @@ Flexible Lengths: the ''fr'' unit</h4>
 <h4 id='resolved-track-list'>
 Resolved Values</h4>
 
+	The 'grid-template-rows' and 'grid-template-columns' properties are
+	<a>resolved value special case properties</a>. [[!CSSOM]]
+
 	When an element's 'display' is ''grid'' or ''inline-grid'' and it generates a box,
 	the <a href="https://www.w3.org/TR/cssom/#resolved-values">resolved value</a>
 	of the 'grid-template-rows' and 'grid-template-columns' properties is the used value,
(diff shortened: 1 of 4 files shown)


mrego:
Just one question, shouldn't `grid-row-gap` and `grid-column-gap` be marked too?
https://drafts.csswg.org/css-grid/#propdef-grid-column-gap

And one nit, the commit says `css-grid` but doesn't talk about grid properties. Probably it'd be nice to update the commit title and description.

zcorpan:
> Just one question, shouldn't `grid-row-gap` and `grid-column-gap` be marked too?
https://drafts.csswg.org/css-grid/#propdef-grid-column-gap

No, the spec for those doesn't say anything about "resolved value", so they work like normal properties. (If implementations disagree, please file a new issue.)

> And one nit, the commit says `css-grid` but doesn't talk about grid properties. Probably it'd be nice to update the commit title and description.

OK, yeah.

Thanks!

zcorpan:
Proposed commit message when merging

```
[cssom][css-grid][css-transforms] Make 'transform-origin' and 'perspective-origin' resolved value like 'height'

Also mention that 'grid-template-rows' and 'grid-template-columns'
are resolved value special case properties (no normative change).

Fixes #392.

mrego:
>> Just one question, shouldn't grid-row-gap and grid-column-gap be marked too?
>> https://drafts.csswg.org/css-grid/#propdef-grid-column-gap

> No, the spec for those doesn't say anything about "resolved value", so they work like normal properties. (If implementations disagree, please file a new issue.)

Yes sorry for the noise, I was mixing in my mind again *computed* vs *resolved* value.

fantasai:
Looks OK to me.

PR #1035, opened 2017-02-14
Replace example screenshot image with SVG version for Example 22

huijing:
This is the first time I'm submitting a pull request to the csswg-drafts so I'm not sure if this is the correct process.


tabatkins:
Your editor appears to have replaced every single tab in the .bs file with two spaces. Fix that, and I'm happy to merge.

huijing:
I'm so sorry about that. I should have checked the git diff before committing. I have fixed the file, and pushed it to my own fork. Do I need to submit a new pull request? I'll merged the latest version from the w3c master branch already.

tabatkins:
Nope, you're good. Thanks!

PR #1066, opened 2017-02-28
SVG images for alignment spec - positional alignment

rachelandrew:
Some SVG images for the alignment specification positional alignment section.

I have the Sketch document here if you can think of clearer examples, or want additional labels etc.

### DIFF
--- a/css-align/Overview.bs
+++ b/css-align/Overview.bs
@@ -234,19 +234,39 @@ Positional Alignment: the ''center'', ''start'', ''end'', ''self-start'', ''self
 		<dt><dfn for="<self-position>, <content-position>">center</dfn> <small>(self, content)</small>
 		<dd>Centers the <a>alignment subject</a> within its <a>alignment container</a>.
 
+		<figure>
+			<img src="images/positional-center.svg" alt="For example, given one item with the inline axis running left to right the item will center inside the alignment container, with an equal amount of space left and right. With two items the items will group together in the center maintaining an equal amount of space left and right.">
+		</figure>
+
 		<dt><dfn for="<self-position>, <content-position>">start</dfn> <small>(self, content)</small>
 		<dd>Aligns the <a>alignment subject</a> to be flush with the <a>alignment container</a>’s start edge in the appropriate axis.
 
+		<figure>
+			<img src="images/positional-start.svg" alt="For example, given an inline axis running left to right, one or more items set to start will align to the left-hand edge of the alignment container.">
+		</figure>
+
 		<dt><dfn for="<self-position>, <content-position>">end</dfn> <small>(self, content)</small>
 		<dd>Aligns the <a>alignment subject</a> to be flush with the <a>alignment container</a>’s end edge in the appropriate axis.
+
+		<figure>
+			<img src="images/positional-end.svg" alt="For example, given an inline axis running left to right, one or more items set to end will align to the right-hand edge of the alignment container.">
+		</figure>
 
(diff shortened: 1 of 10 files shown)


fantasai:
Hm, so in the interest of consistency, I'm thinking these should either follow the style in the overview, or follow the style with the distribution keywords. Also wondering if there's a way to convey what start actually means more clearly? The images make flex-start, self-start, start, and left all seem identical...

fantasai:
(Happy to merge the branch in anyway, though, if you're comfortable making improvements live. But also okay if you want to sketch things out elsewhere before finalizing.)

rachelandrew:
Yes I could think of a useful way to show the difference - if you have any ideas I can give it a go

fantasai:
I got nothing.  :(

rachelandrew:
I was thinking about this. I could switch start and end lines (so start being on the right) which might help people to understand that start isn't always equal to left.

self-start is the hardest one though to come up with a clear example that isn't a mess of labels.

fantasai:
Yeah, but then why is it on the right? In English it's always to the left...

I guess another important point would be that these operate in the vertical axis as well. :) Like, start=top for align-self in Grid or in a horizontal flexbox, but start=left for justify-self in the same situations.

For self-start vs start, maybe show the letters A, 文,  א, and α in each box, and then using the align-self model in flex, show that they all align to the left for start, but the Hebrew letter aligns right for self-start? If we make it a wrap-reverse flexbox (with 5, 6, 7 as the next few boxes to show the ordering) then we can also demonstrate flex-start vs start by wrapping the last few items to a new line, but maybe that's a bit much... it'll need a caption for sure, but then that might be appropriate for the specificness of flex-*.

rachelandrew:
I think if you want to show vertical axis as well we end up with a number of diagrams for each possibility. Which is fine, but it could end up being an entire grid of tiny drawings.

rachelandrew:
Back on this now I've finished launching all the things. How about something like this (this is just for start, so we will end up with a little set for each). I've also used a similar style as the earlier images in the spec.

I think this format gives enough to be able to demonstrate some of the differences between the types.

screenshot 2017-04-15 11 36 36

tabatkins:
We don't need the arrows in the diagrams (they're there in the generic diagrams for each property, to indicate where things can move; here nothing's moving, as it's already positioned), but otherwise :+1: for the diagram revisions.  If you can produce those I'll happily merge.

(Sorry for the delay, but I just merged and reviewed the older ones, and I agree with @fantasai’s earlier critiques.)

rachelandrew:
So I have:

**start**

align-start

**end**

align-end

**center*

align-center

**self-start**

(which I'm still not sure about)

align-self-start

**flex-start**

align-flex-start

**flex-end**

align-flex-end

**left**

align-left

**right**

align-right

Thoughts? Changes/additional images? Once we like them I'll turn them into SVG.

svgeesus:
@rachelandrew if you could link your GitHub account with your W3C one the IPR bot would be happy.

@fantasai are there being merged later after changes, or should this be closed?

alshamiri5 (approved):

svgeesus:
@rachelandrew  there are requests for further changes from @fantasai are you able to make those changes?

Or (given the long time since this PR was opened, and that other illustrations have been added meanwhile) should this PR be closed? The changes in https://github.com/w3c/csswg-drafts/pull/1066#issuecomment-319973888 look good to me personally, but I am not one of the editors.

Also closing/re-opening just to kick the bot into action

Issue #984, opened 2017-01-25
[mediaqueries-4] min-width / max-width / min-height / max-height and fractional viewport dimensions


patrickhlauke:
Currently, `min-` and `max-` prefixes are equivalent to `>=` and `<=` (see https://drafts.csswg.org/mediaqueries-4/#mq-min-max)

This is fine under the assumption that the width/height of the viewport will always be integers, but will fall apart in the case of fractional dimensions (admittedly an edge case?) - see for instance https://github.com/twbs/bootstrap/issues/21552

Wondering if it would be possible to change these so that one of them is "including", while the other is "excluding". For instance, changing the definition to make `min-` equivalent to `>=` but `max-` equivalent to `<` only?

patrickhlauke:
alternatively, for backwards-compatibility with existing content, perhaps having a new modifier for the value part to indicate it's "excluding", thus allowing authors to choose which one they want to make "excluding" (assuming browsers would then infer the rest from context, i.e. if it's used for a `min-` then the value with modifier results in a `>` only, and in case of `max-` with modifier value a `<` only).

Spitballing, based on mathematical notation for intervals, maybe the `]` and `[` characters (where `[` could be assumed as the default for backwards compat)? Something like `min-width: ]320px` / `max-width: [320px` perhaps?

Marat-Tanalin:
The `@media (width < 600px)` syntax stability is [confirmed](https://lists.w3.org/Archives/Public/www-style/2017Feb/0036.html) by Tab Atkins in a [discussion](https://lists.w3.org/Archives/Public/www-style/2017Feb/0034.html) started today at www-style@w3.org.

I’ve filed a [feature request](https://bugzilla.mozilla.org/show_bug.cgi?id=1337174) for Firefox.

patrickhlauke:
As https://drafts.csswg.org/mediaqueries/#mq-range-context would cover the problem of fractional viewport dimensions, is it worth adding a note (or series of notes) in the spec to point out that:

- `min-`/`max-` queries for `width`/`height` can cause problems (or gaps) with fractional viewport dimensions, and that this problem is better solved with the explicit use of `>`/`<=` etc?
- in the range context section, have a matching note reiterating this

If yes, happy to have a stab at this, just let me know @frivoal @tabatkins

PR #1083, opened 2017-03-07
[mediaqueries-4] Add note about min-/max- shortcoming vs proper range context

patrickhlauke:
Closes https://github.com/w3c/csswg-drafts/issues/984

### DIFF
--- a/mediaqueries/Overview.bs
+++ b/mediaqueries/Overview.bs
@@ -634,6 +634,34 @@ Using “min-” and “max-” Prefixes On Range Features</h4>
 			For example, ''(max-width: 40em)'' is equivalent to ''(width <= 40em)''.
 	</ul>
 
+	Note: because “min-” and “max-” both equate to range comparisons that <strong>include</strong> the value,
+	they may be limiting in certain situations.
+
+	<div class='example'>
+		For instance,
+		authors trying to define different styles based on a breakpoint in the viewport width using “min-” and “max-”
+		would generally offset the values they're comparing,
+		to ensure that both queries don't evaluate to true simultaneously.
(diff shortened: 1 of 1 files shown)


tabatkins:
:+1: to the note (tho given its length, I might reformat it into a `<details>`), but please remove the extra `<div class=note>` you've added; it's extraneous, as a paragraph starting with "Note: " is automatically picked up by Bikeshed and formatted accordingly already.

patrickhlauke:
> it's extraneous, as a paragraph starting with "Note: " is automatically picked up by Bikeshed

oops, wasn't aware of that. just force-pushed an updated version of this PR. see what you think (wrapped the whole "For instance..." part into a `<div class="example">` which perhaps makes more sense now...

tabatkins:
Ooh, much better. Thanks!

phistuck:
Is this supposed to work on any browser at the moment?
`data:text/html,<!doctype html><style>@media (width < 320px) { html {background: black} }</style>`
Does not show black on my desktop Chrome or Firefox.

patrickhlauke:
@phistuck this isn't really the best place to ask about current browser support, but: currently, I believe range context is not supported by any (?) browsers. (additionally, once they support it, it would only be browsers/windows with a viewport smaller than 320px that show black, so on desktop this would only happen if you made the window small and/or zoomed in)

phistuck:
@patrickhlauke -
1. You are correct, I apologize - I got here from a Twitter suggestion for solving that problem, so I let myself go a bit because I thought it was a relevant suggestion and perhaps not a futuristic feature. ;)
2. It was a mistake, I meant `>` (I tried that as well).

Thank you!

Issue #1158, opened 2017-04-01
[css-pseudo] Clearer ::first-line definition


Loirooriol:
[CSS Pseudo-Elements says](https://drafts.csswg.org/css-pseudo-4/#first-line-pseudo)

> The ::first-line pseudo-element describes the contents of the first formatted line of its originating element.

> The first formatted line of an element must occur inside a block-level descendant in the same flow

But which block-level descendant? The spec is so vague. It should be more clear. I tried to fill the gaps with what current implementations do, I think it would be

 - A block container box which has block-level boxes inside it only has a `::first-line` pseudo-element if the first in-flow block-level child is also a block container with a `::first-line`. In that case, the `::first-line` is nested just outside the `::first-line` of that child.
 - A block container box which has inline-level boxes inside it always has a `::first-line` pseudo-element. The `::first-line` is nested inside the block container as outermost as possible, and contains the inline-level boxes in the first line box.
 - No other box has a `::first-line` pseudo-element.

This would make this example more understandable: https://jsfiddle.net/oyw62a3t/

```html
<div><p></p>Foo</div>
```
```css
div { color: red; }
div::first-line { color: green; }
```

Since the `p` element contains no line, the first line of the `div` is Foo. However, Foo is red on Firefox, Chrome and Edge. This seems strange with the current definition, but in my proposal it becomes clear.

fantasai:
Alright, finally got around to rewriting the ::first-line section. :) It's needed serious work for ages. Thanks for the super clear outline of what the definition should look like, it's much much better now!
https://drafts.csswg.org/css-pseudo-4/#first-line-pseudo
Lmk if there's anything left to fix; otherwise please close out the issue~

Loirooriol:
I don't remember the details about this issue, but my proposal looks reasonable and the edits seem good :)

PR #1174, opened 2017-04-05
[css-pseudo] only propagate ::first-line to ancestors in BFC

Loirooriol:
The current text was copied from CSS2.1, but nowadays there are block containers which do not participate in a BFC other than table-cell and inline-block, e.g. flex items and grid items.

This is a minor fix, I think the whole algorithm should be replaced with something like #1158.

### DIFF
--- a/css-pseudo/Overview.bs
+++ b/css-pseudo/Overview.bs
@@ -152,7 +152,7 @@ Finding the First Formatted Line</h4>
     (assuming that both <code>P</code> and <code>DIV</code> are blocks).
   </div>
 
-  The first line of a table-cell or inline-block
+  The first line of a <a>block container</a> which does not participate in a <a>block formatting context</a>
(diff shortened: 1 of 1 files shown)


Loirooriol:
Maybe table captions should also be excluded (#1192).

Loirooriol:
An alternative wording would be

> The first line of a block container that is not block box
> cannot be the first formatted line of an ancestor element.

fantasai:
I think this is mostly correct, but it's not clear what happens with multicol. Partly this is because in the past multicol was considered a special type of block container, and right now it's not entirely clear--so we should call it out explicitly.

fantasai:
@Loirooriol Wasn't able to easily resolve the conflicts, so I just generated a new fix. Basically the same wording, except I also added a parenthetical to handle table captions. Gonna close this PR.

PR #1177, opened 2017-04-06
[selectors-4] Make :active :hover and :focus-within work on the flat tree

### LINKED ISSUE #1135: [selectors-4] About shadow-including descendants
For three pseudo-classes ([`:hover`](https://drafts.csswg.org/selectors-4/#the-hover-pseudo), [`:active`](https://drafts.csswg.org/selectors-4/#the-active-pseudo) and [`:focus-within`](https://drafts.csswg.org/selectors-4/#the-focus-within-pseudo)) the spec has a similar text:
> An element also matches :hover if one of its shadow-including descendants matches :hover.

The reference to [*shadow-including descendants*](https://dom.spec.whatwg.org/#concept-shadow-including-descendant) was [introduced back in April 2016 by @tabatkins](https://github.com/w3c/csswg-drafts/commit/eb5e7c1b1546305bda16c6b531bc369de4bc0c81).

However it seems it doesn't match the behavior in either Chrome and Safari.
Using an example from @rune-opera:
```html
  <div id="host">
    <span>Hover me</span>
  </div>
  <script>
    host.attachShadow({mode:"open"}).innerHTML = "<style>#shadowDiv:hover { border: 5px solid green }</style><div id='shadowDiv'><slot></slot></div>";
  </script>
```

In both Blink and WebKit a green border appears on the `shadowDiv` when you hover the `<span>`. This seems the right behavior from the user point of view.
However, the `<span>` is not a shadow-including descendant of the `shadowDiv`. Shouldn't the spec mention the *[flat tree](https://drafts.csswg.org/css-scoping/#flat-tree)* instead?

This discussion started on the [`blink-dev` thread about `:focus-within`](https://groups.google.com/a/chromium.org/d/msg/blink-dev/V3RNBhQelSg/nGRIWjy6CgAJ).

frivoal:
They were incorrectly defined as applying to shadow-including descendants,
while the intent was for them to work on flat tree descendants

Closes #1135

### DIFF
--- a/selectors/Overview.bs
+++ b/selectors/Overview.bs
@@ -1989,7 +1989,10 @@ The Pointer Hover Pseudo-class: '':hover''</h3>
 	(e.g., a pen device that does not detect hovering)
 	are still conforming.
 
-	An element also matches '':hover'' if one of its <a>shadow-including descendants</a> matches '':hover''.
+	An element also matches '':hover''
+	if one of its descendants in the <a>flat tree</a>
+	(including non-element nodes, such as text nodes)
+	matches the above conditions.
 
 	Document languages may define additional ways in which an element can match '':hover''.
 	For example, [[HTML5]] defines a labeled control element as <a href="http://www.whatwg.org/html/selectors.html#selector-hover">matching <code>:hover</code></a>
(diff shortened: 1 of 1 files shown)


tabatkins:
Whoops, I just accidentally committed identical changes without seeing this PR. Sorry for wasting your time, Florian. :(

frivoal:
> Whoops, I just accidentally committed identical changes without seeing this PR. Sorry for wasting your time, Florian. :(

No pb, form my point of view, the effect is the same: I made a PR and the changes I wanted end up in the spec. That you end-up retying the same thing instead of just clicking merge isn't my problem :)

PR #1245, opened 2017-04-19
Create a Media Queries 5 specification

grorg:
Reinstate the content that was taken out of Level 4, and
add 'prefers-reduced-motion'.

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1,0 +1,331 @@
+<pre class='metadata'>
+Title: Media Queries Level 5
+Group: csswg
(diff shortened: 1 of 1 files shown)


grorg:
I don't really care about being an "editor", but I figured that since I created the document I should put my name there. Happy for it to be removed, especially if it means I do less work 😄

frivoal:
Looks sane to me, merging. Further tweaks to be done after merging. Thank you for submitting this.

@astearns @atanassov Adding a co-editor to a draft is the chairs' prerogative. Let me/us know if you'd like a set of co-editors other than what this draft has: Dean, Tab & me.

Issue #825, opened 2016-12-21
[css-fonts] CSSFontFaceRule  does not seem Web compatible


svgeesus:
CSSFontFaceRule has a style property, introduced in DOM Level 2 Style.
 
DOM 2 Style says style property is of type CSSStyleDeclaration
  https://www.w3.org/TR/2000/REC-DOM-Level-2-Style-20001113/css.html#CSS-CSSFontFaceRule
  https://www.w3.org/TR/2000/REC-DOM-Level-2-Style-20001113/css.html#CSS-CSSStyleDeclaration

while CSS3 Fonts says style property is of type DOMString.
   https://drafts.csswg.org/css-fonts/Fonts.html#om-fontface

Informal testing indicates that in browsers (tested Firefox 51, Chrome 57, Edge 14) the style property is an object, per Dom 2 Style. It isn't a DOMString and it is not clear how Web compatible it would be to change the type of the property.

Quick test http://codepen.io/svgeesus/pen/ZBZbbv

svgeesus:
(That test has a separate issue where family == 'sample' works on Chrome and Edge, but family == '"sample"' is needed on Firefox, with double quoting).

svgeesus:
So, is this just a copy paste error in the Fonts 3 spec? If style was CSSStyleDeclaration while family, src, etc were DOMString that would make sense to me, the old one for backwards compat and the new ones for convenience.

tabatkins:
This was no error, it was an intentional "let's fix this better" effort by jdaggett years ago that never panned out.  Old 2.1 spec gave CSSFontFaceRule a .style property, which worked identically to style rules, but jdaggett tried to fix it to just put the valid descriptors directly on the CSSFontFaceRule, like we do with other at-rules.

However, no browser ever picked up the change, including Firefox, which John was nominally in charge of (for the relevant section). :/  This needs to be reverted to match reality, unless browsers actually think they can change their implementation. (I'm pretty sure Chrome's position is that we probably can't.)

upsuper:
I still hope to push forward this change. Could we keep a `style` attribute inside `CSSFontFaceRule` which points to itself, so that we don't use `CSSStyleDeclaration` which includes tons of unrelated properties?

The currently state doesn't make much sense to me.

litherum:
We definitely can't break existing code, which treats it as a CSSStyleDeclaration (using cssText and getPropertyValue()). Given this constraint, I'm not sure that we can do this.

[...]

PR #1247, opened 2017-04-19
[css-fonts] Move CSSFontFaceRule attributes to Level 4

SimonSapin:
Per https://github.com/w3c/csswg-drafts/issues/825#issuecomment-295143110


SimonSapin:
CC @litherum 

(Leaving #825 open because the desired new attributes are not implemented.)

css-fonts/Fonts.src.html
```diff
@@ -4469,17 +4469,15 @@ <h3 id="om-fontface">The <code>CSSFontFaceRule</code> interface</h3>
...
-  attribute DOMString weight;
-  attribute DOMString stretch;
-  attribute DOMString unicodeRange;
-  attribute DOMString variant;
-  attribute DOMString featureSettings;
+  // Intentionally left blank in this specification.
 }</pre>
 
-<p>The DOM Level 2 Style specification [[DOM-LEVEL-2-STYLE]] defined a different variant of this rule.  This definition supercedes that one.</p>
+<p class="note">The CSS Working Group would like <code>CSSFontFaceRule</code> to have
```
litherum: Do we really want to describe the aspirations of the WG? Shouldn't we just state the reason there are no attributes in the object?

PR #1264, opened 2017-04-20
[css-display] Run-in can be inserted in IFC root. #1228

### LINKED ISSUE #1228: [css-display] Can run-in be inserted as child of block which establishes an inline formatting context?
CSS Display defines [Run-In Layout](https://drafts.csswg.org/css-display/#run-in-layout) as

> A run-in box behaves exactly as an inline-level box, except:
> 
>  - If a run-in sequence is immediately followed by a block box that does not establish a new formatting context, it is inserted as direct children of the block box [...]
>  - Otherwise, an anonymous block box is generated around the run-in sequence and all immediately following inline-level content (up to, but not including, the next run-in sequence, if any).
>  - A run-in box with a flow inner display type inlinifies its contents. 

The following example is provided:

> 
> ```html
> <dl class='dict'>
>   <dt>dictionary
>   <dd>a book that lists the words of a language in alphabetical
>       order and gives their meaning, or that gives the equivalent
>       words in a different language.
>   <dt>glossary
>   <dd>an alphabetical list of terms or words found in or relating
>       to a specific subject, text, or dialect, with explanations; a
>       brief dictionary.
> </dl>
> ```
> ```css
> .dict > dt {
>   display: run-in;
> }
> .dict > dt::after {
>   content: ": "
> }
> ```
> Which is formatted as:
> <pre><b>dictionary:</b> a book that lists the words of a language
> in alphabetical order and explains their meaning.
> 
><b>glossary:</b> an alphabetical list of terms or words found
> in or relating to a specific subject, text, or dialect,
> with explanations; a brief dictionary.
> </pre>

However, the `dd` establishes an inline formatting context according to [CSS 2.1](https://www.w3.org/TR/CSS21/visuren.html#block-boxes):

> A block container box either contains only block-level boxes or establishes an inline formatting context and thus contains only inline-level boxes.

So I think that either the example is wrong and the `dt` should not be inserted inside the `dd`, or that the definition should allow a run-in sequence to be inserted as direct children of a block box which establishes an inline formatting context.

Loirooriol:
If a block box establishes a new formatting context, it can either be a block or an inline one. Run-in sequences should be allowed to be inserted into the block box in the latter case. Fixes #1228

### DIFF
--- a/css-display/Overview.bs
+++ b/css-display/Overview.bs
@@ -703,7 +703,7 @@ Run-In Layout</h2>
 	<ul>
 		<li>
 			If a <a>run-in sequence</a> is immediately followed by a block box
-			that does not establish a new <a>formatting context</a>,
+			that does not establish a new <a>block formatting context</a>,
 			it is inserted as direct children of the block box
 			after its ''::marker'' pseudo-element's boxes (if any),
 			but preceding any other boxes generates by the contents of the block


tabatkins:
Whoops, we handled this (with an identical edit) just now while reviewing the issues list. I'll go ahead and close the PR. ^_^

Loirooriol:
@tabatkins No problem. Maybe you can merge #1368 instead?

tabatkins:
hehe, np

PR #1266, opened 2017-04-21
[cssom][all] Add CSSOMString and replace DOMString usage with it

### LINKED ISSUE #1217: [cssom][all specs defining IDL] Consider USVString instead of DOMString, replacing surrogates with U+FFFD
CSSOM uses WebIDL’s `DOMString` type for all string parameters and return values. It corresponds to JavaScript strings: arbitrary sequences of 16-bit code units. These are usually interpreted as UTF-16, but they’re not necessarily well-formed in UTF-16: they can contain [unpaired surrogate code units](https://simonsapin.github.io/wtf-8/#surrogates-code-units). I sometimes call this encoding WTF-16.

(Character encoding decoders never emit surrogates when decoding bytes from the network, even when decoding UTF-16BE or UTF-16LE. So surrogates can’t end up in a string that way, only through JS.)

WebIDL also defines `USVString` which is a Unicode string. (A sequence of Unicode scalar values, excluding surrogate code points.) When [converting to it from a JavaScript string](https://heycam.github.io/webidl/#es-USVString), unpaired surrogate are [replaced with the replacement character U+FFFD](https://heycam.github.io/webidl/#dfn-obtain-unicode).

As far as I know all major browser engines currently use WTF-16 internally, so they preserve unpaired surrogates "by default" when strings go through various browser components where no code is actively looking for those.

In Firefox, we’re working on a new style system ([Stylo, a.k.a. Quantum CSS](https://wiki.mozilla.org/Quantum/Stylo)) where strings are represented with Rust’s native `&str` type. `&str` uses UTF-8 bytes for its in-memory representation of Unicode and guarantees (as part of the type’s contract) that these bytes are well-formed in UTF-8. Unicode designed UTF-8 to specifically exclude surrogate code points, in order to be compatible with (well-formed) UTF-16. As a consequence, well-formed UTF-8 (and `&str`) can not represent all JavaScript strings without some sort of escape sequence mechanism.

Stylo currently replaces unpaired surrogates with U+FFFD when converting JS strings to UTF-8. This is equivalent to defining WebIDL interfaces with `USVString` instead of `DOMString`. This is a deviation from specified and currently-interoperable behavior.

It would be possible to make Stylo preserve surrogates (for example by moving everything to [WTF-8](https://simonsapin.github.io/wtf-8/)). However we’re inclined not to. Preserving surrogates is an historical accident, not a feature. I argue that any occurrence of surrogates in a JS string is likely an error, and coming up with an example where not preserving them in CSSOM makes an observable difference is extremely convoluted. For example:

http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=5012
```html
<!DOCTYPE html>
<style></style>
<script>
document.documentElement.classList.add('\uD800');
document.styleSheets[0].insertRule('.\uD800:before { content: "Surrogates can be used in class names." }', 0);
document.styleSheets[0].insertRule('.\uD801:before { content: "Surrogates seem to be mapped to U+FFFD." }', 1);
</script>
```

So I would like to propose changing CSSOM and other CSS specifications that declare WebIDL interfaces to use `USVString` instead of `DOMString`. This makes CSS syntax “Unicode-clean”, and enable implementations to use UTF-8 internally.

CSSWG discussed and rejected [in 2014](https://lists.w3.org/Archives/Public/www-style/2014Jun/0060.html) a proposal that was effectively the same. However neither `USVString` nor Stylo existed at the time. What has changed is that WebIDL now gives us the tool to easily specify this change, and one major implementation is on a path to likely to make this change.

SimonSapin:
`CSSOMString` is an IDL `typedef` of either `USVString` or `DOMString`, chosen by implementations.

CSSWG resolution: https://github.com/w3c/csswg-drafts/issues/1217#issuecomment-295053842. Fixes #1217.

Each replaced occurrence is one of:

* CSS syntax
* A name (for example a property name) that also occurs in CSS syntax
* `Stylesheet::type`, which is always `text/css`.
* <s>`Stylesheet::title`, which is set from the eponymous HTML content attribute of [`<style>`](https://html.spec.whatwg.org/multipage/semantics.html#attr-style-title) and [`<link>`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-title) elements.</s>

### DIFF
--- a/cssom/Overview.bs
+++ b/cssom/Overview.bs
@@ -281,13 +281,14 @@ typedef DOMString CSSOMString;
 	This choice effectively allows implementations to do this replacement,
 	but does not require it.
 
-	This enables an implementation to use UTF-8 internally to represent strings in memory.
+	Using {{USVString}} enables an implementation
+	to use UTF-8 internally to represent strings in memory.
 	Since well-formed UTF-8 specifically disallows <a>surrogate</a> code points,
 	it effectively requires this replacement.
 
 	On the other hand,
 	implementations that internally represent strings as 16-bit <a>code units</a>
-	may prefer to avoid the cost of doing this replacement.
+	might prefer to avoid the cost of doing this replacement.
 </div>
 
 
@@ -489,11 +490,11 @@ An object that implements the <code>MediaList</code> interface has an associated
 <pre class=idl>
 [LegacyArrayClass]
 interface MediaList {
-  [TreatNullAs=EmptyString] stringifier attribute DOMString mediaText;
+  [TreatNullAs=EmptyString] stringifier attribute CSSOMString mediaText;
   readonly attribute unsigned long length;
-  getter DOMString? item(unsigned long index);
-  void appendMedium(DOMString medium);
-  void deleteMedium(DOMString medium);
+  getter CSSOMString? item(unsigned long index);
+  void appendMedium(CSSOMString medium);
+  void deleteMedium(CSSOMString medium);
 };
 </pre>
 
@@ -843,7 +844,7 @@ The {{StyleSheet}} interface represents an abstract, base style sheet.
 
 <pre class=idl>
 interface StyleSheet {
-  readonly attribute DOMString type;
+  readonly attribute CSSOMString type;
   readonly attribute USVString? href;
   readonly attribute (Element or ProcessingInstruction)? ownerNode;
   readonly attribute StyleSheet? parentStyleSheet;
@@ -882,7 +883,7 @@ The {{CSSStyleSheet}} interface represents a <a>CSS style sheet</a>.
 interface CSSStyleSheet : StyleSheet {
   readonly attribute CSSRule? ownerRule;
   [SameObject] readonly attribute CSSRuleList cssRules;
-  unsigned long insertRule(DOMString rule, optional unsigned long index = 0);
+  unsigned long insertRule(CSSOMString rule, optional unsigned long index = 0);
   void deleteRule(unsigned long index);
 };
 </pre>
@@ -1591,7 +1592,7 @@ interface CSSRule {
   const unsigned short MARGIN_RULE = 9;
   const unsigned short NAMESPACE_RULE = 10;
   readonly attribute unsigned short type;
-  attribute DOMString cssText;
+  attribute CSSOMString cssText;
   readonly attribute CSSRule? parentRule;
   readonly attribute CSSStyleSheet? parentStyleSheet;
 };
@@ -1647,7 +1648,7 @@ The <code>CSSStyleRule</code> interface represents a style rule.
 
 <pre class=idl>
 interface CSSStyleRule : CSSRule {
-  attribute DOMString selectorText;
+  attribute CSSOMString selectorText;
   [SameObject, PutForwards=cssText] readonly attribute CSSStyleDeclaration style;
 };
 </pre>
@@ -1713,7 +1714,7 @@ The <code>CSSGroupingRule</code> interface represents an at-rule that contains o
 <pre class=idl>
 interface CSSGroupingRule : CSSRule {
   [SameObject] readonly attribute CSSRuleList cssRules;
-  unsigned long insertRule(DOMString rule, optional unsigned long index = 0);
+  unsigned long insertRule(CSSOMString rule, optional unsigned long index = 0);
   void deleteRule(unsigned long index);
 };
 </pre>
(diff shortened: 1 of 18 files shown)


cssom/Overview.bs
```diff
@@ -256,6 +256,41 @@ the last item. Unless otherwise specified, an empty list is serialized as the
...
+Most strings in CSSOM interfaces use the <dfn>CSSOMString</dfn> type.
+Each implementation chooses to define it as either {{USVString}} or {{DOMString}}:
+
+<pre class=idl>
+typedef USVString CSSOMString;
+</pre>
+
+Or, alternatively:
+
+<pre class="def lang-webidl">
```
zcorpan: Can use `class="idl extract"` here instead.

cssom/Overview.bs
```diff
@@ -256,6 +256,41 @@ the last item. Unless otherwise specified, an empty list is serialized as the
...
+<div class=note>
+	The difference is only observable from web content
+	when <a>surrogate</a> code units are involved.
+	{{DOMString}} would preserve them,
+	whereas {{USVString}} would replace them with U+FFFD REPLACEMENT CHARACTER.
+
+	This choice effectively allows implementations to do this replacement,
+	but does not require it.
+
+	This enables an implementation to use UTF-8 internally to represent strings in memory.
```
zcorpan: s/This/Using {{USVString}}/

cssom/Overview.bs
```diff
@@ -256,6 +256,41 @@ the last item. Unless otherwise specified, an empty list is serialized as the
...
+	Since well-formed UTF-8 specifically disallows <a>surrogate</a> code points,
+	it effectively requires this replacement.
+
+	On the other hand,
+	implementations that internally represent strings as 16-bit <a>code units</a>
+	may prefer to avoid the cost of doing this replacement.
```
zcorpan: Don't use may in a note

zcorpan:
> * `Stylesheet::title`, which is set from the eponymous HTML content attribute of [`<style>`](https://html.spec.whatwg.org/multipage/semantics.html#attr-style-title) and [`<link>`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-title) elements.

Can you explain why this one shouldn't be `DOMString`?

zcorpan:
(Thanks for fixing link errors in cssom!)

cssom/Overview.bs
SimonSapin: With that Bikeshed gives:

```
FATAL ERROR: Multiple local 'typedef' <dfn>s have the same linking text 'CSSOMString'.

cssom/Overview.bs
SimonSapin: Done.

cssom/Overview.bs
SimonSapin: Replaced with “might”.

cssom/Overview.bs
zcorpan: Hmm... @tabatkins should this work?

SimonSapin:
>> * `Stylesheet::title`, which is set from the eponymous HTML content attribute of [`<style>`](https://html.spec.whatwg.org/multipage/semantics.html#attr-style-title) and [`<link>`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-title) elements.
>
> Can you explain why this one shouldn't be `DOMString`?

I did hesitate about this one, it just seemed more “within style system territory”. I don’t feel strongly about it, though. In Stylo we can (and possibly already do) keep this on the C++ side rather than the Rust side of the bindings. (And Servo needs to decide what to do about every `DOMString`, not just stylesheet titles.)

zcorpan:
So `title` on the HTML side is on `HTMLElement` and is `DOMString`. Then "create a CSS stylesheet" the `title` content attribute is passed for the [title](https://drafts.csswg.org/cssom/#concept-css-style-sheet-title), which is then updated if the content attribute is changed. [`styleSheet.title`](https://drafts.csswg.org/cssom/#dom-stylesheet-title) then returns that value... which is a `DOMString`.

I'm open to changing how this works, but since the string will exist as a `DOMString` anyway, it seems kinda pointless to use `CSSOMString` on `StyleSheet`.

SimonSapin:
Changed `Stylesheet::title` back to `DOMString?`.

zcorpan:
Filed https://github.com/w3c/web-platform-tests/issues/5641 to follow up on tests. The `href` change affects everyone, so might also need bugs on browsers.

cssom/Overview.bs
tabatkins: I haven't added the "extract" support yet.  (The bug is still open!)

(And in any case, I think it's probably better to do exactly this - instead of claiming something is WebIDL and then opting out, just ask it to be highlighted as WebIDL but otherwise be meaningless text.)

PR #1304, opened 2017-04-28
[selectors-4] Clarifying :nth-child and friends

### LINKED ISSUE #1292: [selectors-4] :nth-child and friends are either underdefined or defined wrong
All of these

* https://drafts.csswg.org/selectors/#the-nth-of-type-pseudo
* https://drafts.csswg.org/selectors/#the-nth-last-child-pseudo
* https://drafts.csswg.org/selectors/#the-nth-of-type-pseudo
* https://drafts.csswg.org/selectors/#the-nth-last-of-type-pseudo

have a definition that includes “the An+Bth element” with no details on what that means. This is either wrong (since this is using “the element” singular, where as multiple elements could match) or at least under-defined. A and B are from https://drafts.csswg.org/css-syntax/#anb, but what is n?

This is a regression from Selectors Level 3, which has a precise definition:

https://drafts.csswg.org/selectors-3/#nth-child-pseudo
> The :nth-child(an+b) pseudo-class notation represents an element that has an+b-1 siblings before it in the document tree, for any positive integer or zero value of n, and has a parent element.  

Notable edge cases that should be well-defined include n<0 and A=0.

Loirooriol:
It was not completely clear what “the An+Bth element” meant, because according to CSS Syntax, this notation represents a list of elements. Fixes #1292

### DIFF
--- a/selectors/Overview.bs
+++ b/selectors/Overview.bs
@@ -2543,9 +2543,8 @@ Child-indexed Pseudo-classes</h3>
 '':nth-child()'' pseudo-class</h4>
 
 	The <dfn id='nth-child-pseudo' lt=":nth-child()">:nth-child(<var>An+B</var> [of <var>S</var>]? )</dfn>
-	pseudo-class notation represents the <var>An+B</var>th element
-	that matches the <a>selector list</a> <var>S</var>
-	among its <a>inclusive siblings</a>.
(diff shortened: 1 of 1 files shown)


tabatkins:
I've fixed it in a different way in <https://github.com/w3c/csswg-drafts/issues/1292#issuecomment-298095519> be7492ce3

PR #1326, opened 2017-05-04
[css-counter-styles] Exclude 'none' from <counter-style-name>

### LINKED ISSUE #1295: [css-counter-styles] Exclude 'none' from <counter-style-name>
https://drafts.csswg.org/css-counter-styles/#typedef-counter-style-name
> The general form of an `@counter-style` rule is:
>
> ```css
> @counter-style <counter-style-name> { <declaration-list> }
> ```
>
> where `<counter-style-name>` is a `<custom-ident>`.
> If a counter style’s name is an ASCII case-insensitive match for "decimal" or "none", the `@counter-style` rule is invalid.

We’ve clarified in #1285 that the last sentence only applies to at-rules, and not other uses of `<counter-style-name>`. This makes sense for `decimal` which is a style that always exists (it just can’t be overridden). Allowing `none` however is useless since it can’t ever refer to an existing counter style.

I suggest changing the spec to make `none` (ASCII-case-insensitive) an invalid value of `<counter-style-name>` in all contexts, like CSS-wide keywords already are.

CC @upsuper, @tabatkins

SimonSapin:
… in all contexts, not just `@counter-style` rules.

Fix #1295.

### DIFF
--- a/css-counter-styles/Overview.bs
+++ b/css-counter-styles/Overview.bs
@@ -152,10 +152,12 @@ Defining Custom Counter Styles: the ''@counter-style'' rule</h2>
 
 	<pre>@counter-style <<counter-style-name>> { <<declaration-list>> }</pre>
 
-	where <dfn>&lt;counter-style-name></dfn> is a <<custom-ident>>.
+	<dfn>&lt;counter-style-name></dfn> is a <<custom-ident>>
+	that is not an <a>ASCII case-insensitive</a> match for "none".
 
-	If a counter style's name is an <a>ASCII case-insensitive</a> match for "decimal" or "none",
-	the ''@counter-style'' rule is invalid.
+	Additionally, in the context of the prelude of an ''@counter-style'' rule,
+	an <a>ASCII case-insensitive</a> match for "decimal" is also an invalid <<counter-style-name>>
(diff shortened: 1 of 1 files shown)


SimonSapin:
CC @tabatkins

css-counter-styles/Overview.bs
```diff
@@ -152,10 +152,12 @@ Defining Custom Counter Styles: the ''@counter-style'' rule</h2>
 
 	<pre>@counter-style <<counter-style-name>> { <<declaration-list>> }</pre>
 
-	where <dfn>&lt;counter-style-name></dfn> is a <<custom-ident>>.
+	<dfn>&lt;counter-style-name></dfn> is a <<custom-ident>>
+	that is not an <a>ASCII case-insensitive</a> match for "none".
 
-	If a counter style's name is an <a>ASCII case-insensitive</a> match for "decimal" or "none",
-	the ''@counter-style'' rule is invalid.
+	Addtionally, in the context of the prelude of an ''@counter-style'' rule,
```
tabatkins: "Additionally"

SimonSapin: Fixed.

PR #1387, opened 2017-05-17
[css-namespaces] Restore Terminology section

SimonSapin:
It was removed in https://github.com/w3c/csswg-drafts/commit/23f06fe37e7185bd44e8408cd533b2f5110ad586, presumably accidentally.

It includes a normative statements that was missing for the ED:

> In CSS Namespaces a namespace name consisting of the empty string is taken to represent the null namespace or lack of a namespace.

### DIFF
--- a/css-namespaces/Overview.bs
+++ b/css-namespaces/Overview.bs
@@ -61,6 +61,34 @@ Introduction</h2>
 	rather than possibly match them incorrectly.
 
 
+<h3 id="terminology">
+Terminology</h3>
+
+	Besides terms introduced by this specification,
+	CSS Namespaces uses the terminology defined in Namespaces in XML 1.0. [[!XML-NAMES]]
+	However, the syntax defined here is not restricted to representing XML element and attribute names
+	and may represent other kinds of namespaces as defined by the host language.
+
+	In CSS Namespaces a namespace name consisting of the empty string
+	is taken to represent the null namespace
+	or lack of a namespace.
+
+	<div class="example">
+		For example, given the namespace declarations:
+
+		<pre>
+			@namespace empty "";
+			@namespace "";
+		</pre>
+
+		The <a>type selectors</a>
+		<code>elem</code>,
+		<code>|elem</code>,
+		and <code>empty|elem</code>
+		are equivalent.
+	</div>
+
+
 <h2 id="declaration">Declaring namespaces: the ''@namespace'' rule</h2>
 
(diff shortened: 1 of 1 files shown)


SimonSapin:
CC @tabatkins, @bzbarsky 

And because I don’t want to throw it away, here is the evidence I gathered this morning to argue for a normative spec change, before I realized the statement I wanted was already in the spec, just not in the ED.

<details>
The css-namespaces spec distinguishes these two cases:

https://drafts.csswg.org/css-namespaces/#syntax
> All strings—including the empty string and strings representing invalid URIs—are valid namespace names in `@namespace` declarations.

https://drafts.csswg.org/css-namespaces/#css-qualified-name
> The prefix of a qualified name may be omitted to indicate that the name belongs to no namespace, i.e. that the namespace name part of the expanded name has no value.

----

However, the rest of the platform ensure that it is not possible for an element’s or attribute’s name to have an empty string namespace name, as opposed to no namespace. Any attempt to construct the former results in the latter.

Or, equivalently, the empty namespace name and the lack and namespace can be considered to be the same thing.

https://dom.spec.whatwg.org/#namespaces
> To validate and extract a namespace and qualifiedName, run these steps:
>
> 1. If namespace is the empty string, set it to null. 
> 2. […]

https://dom.spec.whatwg.org/#locate-a-namespace
> […] an attribute whose namespace is the XMLNS namespace, namespace prefix is null, and local name is "xmlns", then return its value if it is not the empty string, and null otherwise. 

https://www.w3.org/TR/REC-xml-names/#iri-use
>  The empty string, though it is a legal URI reference, cannot be used as a namespace name. 

https://www.w3.org/TR/REC-xml-names/#nsc-NoPrefixUndecl
> Namespace constraint: No Prefix Undeclaring
> In a namespace declaration for a prefix (i.e., where the NSAttName is a PrefixedAttName), the attribute value MUST NOT be empty.

https://www.w3.org/TR/REC-xml-names/#defaulting
> The attribute value in a default namespace declaration MAY be empty. This has the same effect, within the scope of the declaration, of there being no default namespace. 

----

These tests pass in all of Firefox, Chromium, Safari, and Edge:

* http://w3c-test.org/css/css-namespaces-3/prefix-002.xml
* http://w3c-test.org/css/css-namespaces-3/prefix-003.xml
* http://w3c-test.org/css/css-namespaces-3/prefix-004.xml

(They are served with a `Content-Type: application/xml` HTTP response header, so they’re parsed with the XML parser rather than the HTML parser.)
</details>

PR #1389, opened 2017-05-17
[cssom-1] Add missing extract class on dashed_attribute example

SaschaNaz:
https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-_dashed_attribute

```WebIDL
partial interface CSSStyleDeclaration {
  [CEReactions, TreatNullAs=EmptyString] attribute CSSOMString _dashed_attribute;
};
```

This IDL snippet does not have `extract` class attached and thus collected on IDL index, whereas other _camel_cased_attribute and _webkit_cased_attribute examples are not collected.


zcorpan:
Thanks!

PR #1467, opened 2017-05-26
[css-align] Fixes Example 1 behavior in Firefox

SelenIT:
Currently, Firefox wraps the word "Authoritarianism" in the Example 1, making it hard to understand the difference between unsafe and safe centering. Setting `white-space:nowrap` to it makes it more clear.


tabatkins:
Omg, and the `visibility:collapse` example has the same problem! Thanks for pointing this out.

PR #1478, opened 2017-05-30
[css-display] Link "replaced element" to CSS Display instead of CSS2

Loirooriol:
CSS Display defines "replaced element" in the same spec, so there is no need to link CSS2.

### DIFF
--- a/css-display/Overview.bs
+++ b/css-display/Overview.bs
@@ -131,9 +131,9 @@ Box Layout Modes: the 'display' property</h2>
 	which consists of the two basic qualities of how an element generates boxes:
 
 	* the <dfn export local-lt="inner">inner display type</dfn>,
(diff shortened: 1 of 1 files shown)


SelenIT:
By the way, shouldn't the definition of the "replaced element" in CSS be synchronized with the corresponding definition in HTML? For example, the MDN page, based on CSS definition, lists `input` and `textarea` as replaced elements, while the 'Rendering' section of the HTML spec explicitly lists them [as non-replaced](https://html.spec.whatwg.org/multipage/#toc-rendering) (and introduces the term 'widget' instead).

tabatkins:
@Loirooriol Ah, thanks, those links pre-date us adding the definition to this spec.

@SelenIT Yes, there's a category of "semi-replaced" that we've been edging around in CSSWG discussions for a little bit that we need to formalize at some point - their layout is described by CSS, but they have some intrinsic sizing qualities like a replaced element does. `input` and `textarea` both fall into this.

PR #1498, opened 2017-06-04
Allow -n in an-b, allow comments within +n and -n

Manishearth:
Currently the grammar doesn't allow for `-n` within `:nth-child`, it only deals with `+n`. However, `-n` (and things like `-n+5`) are accepted by browsers.

This fixes the discrepancy.

Additionally, the parsing here is a bit tricky -- while the `+`/`-` and `n` are separate tokens, the spec does not allow for spaces between them. Chrome follows this, and as of  https://bugzilla.mozilla.org/show_bug.cgi?id=1364009, firefox does too. However, both allow comments. While this is probably implicit, I'm clarifying this within the spec.

See also: https://github.com/servo/rust-cssparser/issues/153,

r? @tabatkins @SimonSapin

### DIFF
--- a/css-syntax/Overview.bs
+++ b/css-syntax/Overview.bs
@@ -2655,23 +2655,23 @@ The <code>&lt;an+b></code> type</h3>
 		  <var>&lt;integer></var> |
 
 		  <var>&lt;n-dimension></var> |
-		  '+'?<sup><a href="#anb-plus">†</a></sup> n |
+		  ['+' | '-']?<sup><a href="#anb-plus">†</a></sup> n |
 		  -n |
 
 		  <var>&lt;ndashdigit-dimension></var> |
-		  '+'?<sup><a href="#anb-plus">†</a></sup> <var>&lt;ndashdigit-ident></var> |
+		  ['+' | '-']?<sup><a href="#anb-plus">†</a></sup> <var>&lt;ndashdigit-ident></var> |
 		  <var>&lt;dashndashdigit-ident></var> |
 
 		  <var>&lt;n-dimension></var> <var>&lt;signed-integer></var> |
-		  '+'?<sup><a href="#anb-plus">†</a></sup> n <var>&lt;signed-integer></var> |
+		  ['+' | '-']?<sup><a href="#anb-plus">†</a></sup> n <var>&lt;signed-integer></var> |
 		  -n <var>&lt;signed-integer></var> |
 
 		  <var>&lt;ndash-dimension></var> <var>&lt;signless-integer></var> |
-		  '+'?<sup><a href="#anb-plus">†</a></sup> n- <var>&lt;signless-integer></var> |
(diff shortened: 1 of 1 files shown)


tabatkins:
The grammar *does* allow for `-n` and `-n+5`. Every single "group" of clauses has the first line handle a no-sign initial, the second handle an initial `+`, and the third handle an initial `-`.

What it doesn't allow is `-` and `n` to be separate tokens; since we're already requiring no spaces between the sign and the `n`, the only reason to parse them as separate tokens *on purpose* is to specifically allow comments between them. We can't avoid that for `+n`, because that parses as two tokens no matter what, but there's no particular reason to specifically allow `-/**/n` to work - it's just more effort for literally zero payoff.

Chrome does not allow this, by the way: 

```html
<!DOCTYPE html>
<p>foo<p>foo<p>foo
<style>p:nth-child( -/**/n + 2 ) { color: red; }</style>
```

doesn't turn anything red, while `+/**/n + 2` does (correctly).  It looks like Firefox *does* allow it, giving the same effect as `-n + 2`, which I assume is because they don't follow the spec and instead just reparse the source text with a specialized an+b parser.

tabatkins:
Oh, and note that comments are allowed *anywhere* between two tokens. This is implicit in ["consume a token"](https://drafts.csswg.org/css-syntax/#consume-token) - whenever you ask for the next token, it automatically consumes (and silently discards) any comments as well.  In [Serialization](https://drafts.csswg.org/css-syntax/#serialization), it's specified that impls are *allowed* to remember where comments were, and re-insert them during serialization if they wish, but this information must have no effect on parsing - as far as CSS parsing and grammars are concerned, comments don't exist.

SimonSapin:
The initial message of this PR is misleading. This is specifically about the `-/**/n` case.

tabatkins:
Sure, but I addressed that specifically in my response, too.  Chrome does *not* allow `-/**/n + 2` as far as I can tell, so the premise is flawed.

Manishearth:
Yeah, I hadn't misread the spec -- but I felt it should be symmetric.

For some reason it seemed like Chrome accepted this? Maybe it was a separate case, I was doing a lot of testing to make Firefox match Chrome and the spec.

Will remove this behavior from Firefox then.

tabatkins:
Theoretical purity (symmetry) doesn't apply when we're dealing with "what random-ass places in the middle of expressions can people put comments in?". ^_^  There's a lot of things like this that are just artifacts of CSS tokenization, not something meaningful.

Manishearth:
Fair. Firefox has a bajillion tests for "what random-ass places can folks put comments in", but I'm just going to remove the ones that test for this behavior and fix up the parsing to match this.

tabatkins:
*Testing* for those things is good; people will put them anywhere they're allowed, because users tessellate the possibility space of an API. But trying to make decisions about where to allow them based on aesthetics is not. ^_^

Manishearth:
Agreed

PR #1532, opened 2017-06-15
[css2] Linkify a mention of the 'height' property

dholbert:
In section CSS2 10.3.2, there are a series of paragraphs that start out "If 'height' and 'width' ..." with both properties linkified.  One of them neglects to linkify 'height'. This pull request fixes that.


nghi234:
No1

tabatkins:
I have no idea how CSS2's build system works, but I'll pretend that this is correct and approve it. ^_^

PR #1540, opened 2017-06-19
[cssom-view] Clarify definition of scrolling box

### LINKED ISSUE #1526: [cssom-view] "Elements and viewports have an associated scrol..."
https://drafts.csswg.org/cssom-view/#terminology

> Elements and viewports have an associated scrolling box if has a scrolling mechanism or it overflows its content area and the used value of the overflow\-x or overflow\-y property is hidden\. \[CSS3\-BOX\]
> 
> An element is potentially scrollable if all of the following conditions are true:​
> 
> The element has an associated CSS layout box\.
> The element is not the HTML body element, or it is and the root element’s used value of the overflow\-x or overflow\-y properties is not visible\.
> The element’s used value of the overflow\-x or overflow\-y properties is not visible\.
> Note:​ An element that is potentially scrollable might not have a scrolling box\. For instance, it could have overflow set to auto but not have its content overflowing its content area\.

The definition of scrolling box looks wrong. Certainly `scroll` and `auto` should also cause a scrolling box.

The definition of potentially scrollable also looks wrong w.r.t. `overflow: clip`.

zcorpan:
Fixes #1526.

### DIFF
--- a/cssom-view/Overview.bs
+++ b/cssom-view/Overview.bs
@@ -125,8 +125,9 @@ HTML.
 <dfn>Content edge</dfn>, <dfn>padding edge</dfn>, <dfn>border edge</dfn>, <dfn>margin edge</dfn>, and <dfn>viewport</dfn> are defined by CSS.
 <!-- These *should* just link to CSS2.1, but it doesn't have actual anchors to link to, because it's THE WORST. -->
 
-Elements and <a>viewports</a> have an associated <dfn>scrolling box</dfn> if has a scrolling mechanism<!--XXX xref--> or it overflows
-its content area<!--XXX xref--> and the used value of the 'overflow-x' or 'overflow-y' property is ''overflow/hidden''. [[!CSS3-BOX]]
+An element or <a>viewport</a> has an associated <dfn>scrolling box</dfn> if it has a <a>scroll
+container</a> and <a>scrollable overflow</a>. This box is the same as the <a>scrollport</a>.
+[[!CSS-OVERFLOW-3]]
 
 An element <var>body</var> (which will be <a>the HTML <code>body</code> element</a>) is
 <dfn>potentially scrollable</dfn> if all of the following conditions are true:


dbaron:
Is there a way to review a generated version that shows what these link to?  It's not clear to me what spec that would be in...

zcorpan:
Generated HTML is
```html
   <p>An element or <a data-link-type="dfn" href="#viewport" id="ref-for-viewport">viewport</a> has an associated <dfn class="dfn-paneled" data-dfn-type="dfn" data-noexport="" id="scrolling-box">scrolling box</dfn> if it has a <a data-link-type="dfn" href="https://drafts.csswg.org/css-overflow-3/#scroll-container" id="ref-for-scroll-container">scroll
container</a> and <a data-link-type="dfn" href="https://drafts.csswg.org/css-overflow-3/#scrollable-overflow" id="ref-for-scrollable-overflow">scrollable overflow</a>. This box is the same as the <a data-link-type="dfn" href="https://drafts.csswg.org/css-overflow-3/#scrollport" id="ref-for-scrollport">scrollport</a>. <a data-link-type="biblio" href="#biblio-css-overflow-3">[CSS-OVERFLOW-3]</a></p>
```

tigercosmos:
Shall we land this PR?

dbaron:
@zcorpan - Sorry for taking so long to get back to this:

It's not clear to me why you're conditioning this on having scrollable overflow.  (And, actually, I don't think it's really defined what *having* scrollable overflow means.  I'd generally interpret that to mean that something that is included in the scrollable overflow (as the term is defined) actually overflows the padding-box.  Strictly speaking, it's defined as a set, so it's always there.)  If your intent is to condition it on whether scrollable overflow is overflowing the padding box... I'm not sure why you'd want to do that.  My instinct would be to define the concept for everything that is a scroll container.  It seems cleaner, and not doing that seems likely to be problematic for something like `overflow: scroll` without any overflow.

I'm also a little worried about element/box terminology and has/is terminology.  A [scroll container](https://drafts.csswg.org/css-overflow-3/#overflow-properties) is defined as a thing that *boxes* *are*.  This seems to say it's a thing that *elements* *have*.  I think it's probably good to use boxes.  I don't actually care about has vs. is except for being consistent.

dbaron:
Also, if it's the same as the scrollport, why make a new term rather than using that term?

dbaron:
Also, I guess I've looked through the uses of the term now (which I probably should have done earlier), and I suppose I see that given the uses are all related to the APIs that cause movement on scrolling, why you'd want to condition it on having scrollable overflow.  However, it seems like that usage precludes things like being able to control whether removal of content (*if* it causes a box to no longer have overflow) triggers a smooth or instant scroll.  So even given that usage, it seems better to have the term be consistently defined whether or not there's currently overflow, and handle all cases of the scrollable range changing consistently, whether or not the change in the scrollable range means there's no longer any room for movement.  And it seems like the restrictions on the range you're able to scroll to should be able to handle all the cases that not being able to scroll at all can.

(I'd also note that sometimes there's scrollable overflow that doesn't cause any ability to scroll, e.g., if it's above the scrollport.)

zcorpan:
Replacing with scrollport seems good. I don't recall why I defined it like this.

PR #1542, opened 2017-06-20
[mediaqueries-5] remove level from shortname metadata

tobie:
This causes the GitHub Issues link to point to https://github.com/w3c/csswg-drafts/labels/mediaqueries-5-5 (instead of https://github.com/w3c/csswg-drafts/labels/mediaqueries-5).

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1,7 +1,7 @@
 <pre class='metadata'>
 Title: Media Queries Level 5
 Group: csswg
-Shortname: mediaqueries-5
+Shortname: mediaqueries
 Level: 5
 Status: ED
 Status Text: Once complete, this specification will include and extend <cite>Media Queries Level&nbsp;4.</cite> [[MEDIAQUERIES-4]]


frivoal:
Hmmm. Yes, it does solve that problem, but then it introduces another one instead:
if it has the same shortname as media queries level 4, bikeshed considers that it replaces level 4 entirely, which is not true, since it is a diff spec which only contains the new bits, and leaves level 4 to define the rest.

Applying your patch, bikeshed starts to complain about not finding definitions that level 5 refers to, but are defined in level 4. I get these errors for instance:

```
LINK ERROR: No 'dfn' refs found for 'media queries' that are marked for export.
<a data-link-type="dfn" data-lt="Media Queries">Media Queries</a>
LINK ERROR: No 'dfn' refs found for 'media feature' that are marked for export.
<a data-link-type="dfn" data-lt="media feature">media feature</a>
LINK ERROR: No 'type' refs found for '<media-query-list>' that are marked for export.
<a class="production" data-link-type="type" data-lt="&lt;media-query-list&gt;">&lt;media-query-list&gt;</a>
LINK ERROR: No 'dfn' refs found for 'range context' that are marked for export.
<a data-link-type="dfn" data-lt="range context">range context</a>```

@tabatkins What's the writing way to set up the spec so that bikeshed both generates the correct links (as with @tobie 's patch) AND keeps understanding links to things defined in the previous level (as with the spec as it is now)?

tobie:
> Hmmm. Yes, it does solve that problem, but then it introduces another one instead:

Oops. :-/ Sorry!

Tagging @tabatkins, as the [previous tagging](https://github.com/w3c/csswg-drafts/pull/1542#issuecomment-309739170) got lost in markdown issues.

tabatkins:
Bikeshed doesn't handle diff specs very well yet; the "lots of links are defined in the previous version" behavior they want directly conflicts with the "links to a previous version are almost always accidental references to a removed feature" behavior that all other specs want.

That said, I just got an idea for handling this better. I'll see what I can do.

tabatkins:
This is now tracked in <https://github.com/tabatkins/bikeshed/issues/1050>.

fantasai:
@frivoal MQ4 is CR, I think you should just merge this change. The fact that all the links break means you need to merge in the L4 content, but you gotta do that anyway.

frivoal:
MQ5 is no longer a delta spec, and this shortname trickery has been fixed.

PR #1560, opened 2017-06-25
[css-flexbox][css-grid] Removed redundant exclusion of column-* properties from applying to flex and grid containers

SebastianZ:
Removes the [now redundant](https://github.com/w3c/csswg-drafts/commit/52ebb10453232307562a4e2c69e82565da9fd39b) exclusion of `column-*` properties from applying to flex and grid containers.

Sebastian

### DIFF
--- a/css-flexbox/Overview.bs
+++ b/css-flexbox/Overview.bs
@@ -28,7 +28,7 @@ Former Editor: Neil Deakin, Mozilla Corporation, enndeakin@gmail.com
 Former Editor: Ian Hickson, formerly of Opera Software, ian@hixie.ch
 Former Editor: David Hyatt, formerly of Netscape Corporation, hyatt@apple.com
 !Issues List: <a href="https://drafts.csswg.org/css-flexbox-1/issues">https://drafts.csswg.org/css-flexbox-1/issues</a>
-Ignored Terms: column-*, auto, first formatted line, first letter, static position
+Ignored Terms: auto, first formatted line, first letter, static position
 Ignored Vars: item’s own max-content, maximum min-content among all items
 </pre>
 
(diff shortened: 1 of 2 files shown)


fantasai:
*sigh* This isn't redundant until css-multicol gets republished. :/ Which, given it's not actively maintained, isn't likely to be very soon.

SebastianZ:
> *sigh* This isn't redundant until css-multicol gets republished. :/ Which, given it's not actively maintained, isn't likely to be very soon.

Hmm, I totally missed that. So, I guess the next steps are to find someone to take over the spec. and then republish it. (It's time anyway, the last revision is more than six years old.)

Sebastian

SebastianZ:
But wait, couldn't Florian do the republishing?

Sebastian

frivoal:
I could republish multicol, and I suppose the current state of the ED is better than the current state of the CR. However, I have not reviewed the last 6 years worth of comments and resolutions, much less addressed all said comments and applied all said resolutions. So if we republish now, it will need to be as a WD, not as a CR.

If we want to republish as a CR, there's a fair chunk of work to do first.

rachelandrew:
I'd be willing to help with that if I had some mentoring initially. I know the spec and am interested and have experience both on the web and epublishing side of things.

frivoal:
@rachelandrew, I think it would be terrific if you could become an editor of the multicol spec, and I am more than happy help you get started with editing and publishing. As for me, I cannot commit to doing all the heavy lifting myself, but I'm happy to stay involved and do a part.

Maintenance of multicol is on the agenda for the Paris F2F next week. Let's think about our respective time commitments and see if two co-editors are enough, or if we need a third cook in that kitchen.

rachelandrew:
Great - let's have a chat about it in Paris. If you'd be willing to help me get started I am more than happy to work on it.

PR #1570, opened 2017-06-28
Mark css-overflow-4 transition as done

frivoal:


svgeesus:
I generally mark them as done only when they have actually been published, ie the webmaster agrees they are okay.

svgeesus:
Oh good catch, I hadn't noticed someone had already done it (but not updated this document)

Issue #1457, opened 2017-05-23
[css-contain] Becoming a formatting context


Loirooriol:
I see various problems with the [Becoming a formatting context](https://drafts.csswg.org/css-display-3/#becoming-formatting-context) section.

> In some circumstances (See [\[CSS-CONTAIN-1\]](https://drafts.csswg.org/css-display-3/#biblio-css-contain-1) or [\[CSS3-MULTICOL\]](https://drafts.csswg.org/css-display-3/#biblio-css3-multicol) for examples), an element may need to become a formatting context.

"See" should be lowercase.

Situations that CSS2.1 said produced a block formatting context when in fact they only needed to become a formatting context should be listed here. This includes out-of-flow boxes, and boxes to which `overflow` applies and is different than `visible`. Maybe also mention block containers that are not block boxes.

> If the element already establishes a formatting context of any kind, this condition is satisfied.

Except for inline formatting contexts! Inline formatting contexts are not contained at all, they are e.g. affected by floats, so `contain: layout` is not satisfied by a block container that establishes an inline formatting context, it must establish a block formatting context instead.

And what about ruby formatting contexts? I know very little about them, but I don't think establishing a ruby formatting context is enough neither.

I propose:

 - If the inner display type is `flow-root`, `table`, `flex` or `grid`, the condition is satisfied.
 - If the inner display type is `flow`, it is changed to `flow-root`.
 - An element with a `ruby` inner display type always (not just if the outer display type is different than `inline`) generates two boxes: a principal box with the inner display type set to `flow` and the outer display type set to the outer display type of the element, and an inline-level ruby container. "Becoming a formatting context" only affects the principal box, changing its inner display type to `flow-root`.
 - Not sure about layout internal types.

Loirooriol:
This note also needs to be fixed:

> Exclusions are able to affect content across formatting context boundaries. (At time of writing, they are the only layout feature that can.)

Maybe it would be better to say that inline (and ruby?) are pseudo-formatting contexts or some thing different than "real" formatting contexts.

fantasai:
[...]

Issue #1071, opened 2017-03-03
[css-multicol] 'column-span' and tables


dbaron:
The [definition of `column-span`](https://drafts.csswg.org/css-multicol/#column-span) says:
> Applies to: 	in-flow block-level elements

The [definition of block-level element](https://drafts.csswg.org/css2/visuren.html#block-level) says that it applies to things with `display: block`, `display: list-item` and `display:table`.

Then the definition of column-span goes on to say:
> The element establishes a new block formatting context.

This means that (1) `column-span: all` applies to tables and (2) when it does, they become block formatting contexts.

Having a single edge case in which tables can establish block formatting contexts doesn't make sense.  This shouldn't happen.

That said, given the lack of consideration to this issue, I wonder whether `column-span` is really intended to apply to tables.

What do existing implementations do?  (IIRC, Gecko is the last engine to implement `column-span`, which would have made me expect the spec to be in better shape than it appears to be.)

/cc @tabatkins @gregwhitworth @neerjapancholi

gregwhitworth:
Looking at this testcase: https://jsfiddle.net/kememj9w/2/

It works on tables in Chrome/Edge (didn't test Safari). Granted following normal table layout rules it shrinks to fit its content so you have to set the width to `100%` in order to consume the available width of the columns that `span: all` provides it.

![image](https://cloud.githubusercontent.com/assets/865244/23574800/c6735424-0037-11e7-9dca-25e269e1b35d.png)

frivoal:
I think this is merely imprecise wording, not an actual expectation that tables should establish a BFC. The goal was, IIRC, the margins of descendants of two adjoining spanners would not collapse with each other, and that floats from one would not intrude in the other.  For that, block spanners need to establish a BFC, but tables are fine e as is.

FremyCompany:
Just wanted to mention that this property would have to be hoisted if we want to apply it on tables (like every single css property except background and border, but hey as long as we want the spec to be a whitelist...)

mstensho:
The way I see it, the goal was to keep column row content from interacting with spanners, and to keep things inside spanners from interacting with things inside other spanners, be it margins of floats. Just saying that they establish a new block formatting context may be too simplistic.

[...]

PR #1588, opened 2017-07-07
[css-multicol] Revert part of a change about formatting contexts

frivoal:
Stating that the element establishes a new formatting context, without
saying what kind of formatting context it needs to establish is not
specific enough. Also, saying that it must establish a "new" formatting
context is ambiguous or misleading for elements that would establish one
anyway.

The current definition of "becomes a formatting context" is
problematic (see https://github.com/w3c/csswg-drafts/issues/1457), so we
need to fix it, but it is precisely aimed at solving this: defining what
kind of formatting context to establish for the boxes that wouldn't
otherwise, and being clear about which boxes are fine as is and don't
need to be changed.

### DIFF
--- a/css-multicol/Overview.bs
+++ b/css-multicol/Overview.bs
@@ -952,7 +952,7 @@ Spanning columns</h2>
 			Content in the normal flow that appears before the
 			element is automatically balanced across all columns before the
 			element appears.
-			The element establishes a new <a>formatting context</a>.
+			The element <a>becomes a formatting context</a>.
 
 			Note: Whether the element establishes a new <a>formatting context</a>
 			does not depend on whether the element is a descendent of a multicol or not.


frivoal:
@fantasai You have [changed](https://github.com/w3c/csswg-drafts/commit/c54cda7e2a79dec7240c292cf96a84390ceaa992) the [commit I made](https://github.com/w3c/csswg-drafts/commit/f3dc11e73b920ea652bb439b17f6ef63e2ab376b) to solve #1071. I disagree with (part of) your commit, but as I don't want to start an editing war where we revert each other, I've made this pull request with my fix to your fix, to be merged (or closed) once we agree.

The reason for my disagreement is explained in the comment above (which is the commit message).

@tabatkins, since you approved my original commit, feel free to chime in to either back me up or repent for ever thinking this was a good idea.

tabatkins:
I agree with @fantasai's revert in general - despite my original +1, the text for "becomes a formatting context" is indeed incorrect and needs fixing.  I do agree that we need to link to an algo for this process, rather than just relying on "establishes a new FC", for the reasons you state.

frivoal:
@tabatkins I'm a little confused. Are you saying that we should merge this (and link back to "becomes a formatting context") once the text for "becomes a formatting context" is fixed, or that we shouldn't link back and fix this some other way? To my feeble mind, your comment seems to implies both.

tabatkins:
Yes, it's both. Basically the text right now is bad and we should fix it, which we'll do after f2f discussion, and so there's no much reason to worry about the language right at this moment.

PR #1647, opened 2017-07-26
[css-box] fix ex.12 'margin-bottom' -> 'margin'

retrazil:
There's en error in example 12.
`margin-bottom: 2em 0 1em 0`


FremyCompany:
Thanks!

PR #1650, opened 2017-07-26
[css-align] Fix reference to unexisting justify-items: auto.

### LINKED ISSUE #1649: [css-align] Spec still references justify-items: auto
In https://drafts.csswg.org/css-align/#justify-items-property:

> Computed value: specified value, except for auto (see prose)

But the `auto` value is no more, that should probably refer to `legacy`.

emilio:
Fixes #1649.


tabatkins:
Thanks!

PR #1698, opened 2017-08-05
[css-box] Fixed example 26

retrazil:
Example 26 has lot of errors. I've replaced ambiguous 'body' with 'content'.  Also attached all the container (or boxes/ pseudo code divs) names with actual css ids and replaced pseudo code like <scroller> with actual div tags.

### DIFF
--- a/css-box/Overview.html
+++ b/css-box/Overview.html
@@ -3092,13 +3092,13 @@ with an overflow parent. Style sheet: </p>
 <pre>container { position: relative; border: solid; }
 scroller { overflow: scroll; height: 5em; margin: 5em; }
 satellite { position: absolute; top: 0; }
-body { height: 10em; }
+content { height: 10em; }
 </pre>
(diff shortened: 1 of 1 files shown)


tabatkins:
None of these are actually wrong - it's not valid HTML, sure, but the tag structure and selectors are otherwise fine.  And the 'border' declaration is correct as is.

retrazil:
Why use pseudo code in one example while all the rest of the examples contain valid html code ? <body/> there can be easily confused with the actual html <body> tag. 

'border' declaration is correct like you say.

frivoal:
@tabatkins This PR should be closed or accepted. Keeping it open forever does not help anyone.

@retrazil The original example isn't pseudo code, it would work in if loaded in a browser. Yes, it does use non existing elements with made up semantics, but the valid markup you suggested is more verbose, which may distracts from the point the example wants to make.

PR #1709, opened 2017-08-10
MQ5 User Prefs: prefers-reduced-transparency

cookiecrook:
Discussed at recent F2F meeting, and in CSS Issue #443

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -478,6 +478,36 @@ Detecting the desire for less motion on the page: the 'prefers-reduced-motion' f
 
 	</dl>
 
+<h3 id="prefers-reduced-transparency">
+Detecting the desire to reduce the use of transparent or translucent layering effects on the page: the 'prefers-reduced-transparency' feature</h3>
+
+	<pre class='descdef mq'>
+	Name: prefers-reduced-transparency
+	Value: no-preference | reduce
+	For: @media
+	Type: discrete
+	</pre>
+
+	The 'prefers-reduced-transparency' media feature is used to detect if the user
+	has requested the system minimize the amount of transparent or translecent 
+	layering effects it uses.
+
+	<dl dfn-type=value dfn-for="@media/prefers-reduced-transparency">
+		<dt><dfn>no-preference</dfn>
+		<dd>
+			Indicates that the user has made no preference known
+			to the system. This keyword value evaluates as false
+			in the <a>boolean context</a>.
+
+		<dt><dfn>reduce</dfn>
+		<dd>
+			Indicates that user has notified the system that they
+			prefer an interface that minimizes the amount of 
+			transparent or translucent layer effects.
+
+	</dl>
+
+
 <h2 id="changes" class="no-num">
 Changes</h2>
 <h3 id="changes-2017">
@@ -493,6 +523,10 @@ The following changes were made to this specification since the
 
 	<li>
 		Add 'prefers-reduced-motion' media feature.
+
+	<li>
+		Add 'prefers-reduced-transparency' media feature.		
+
 </ul>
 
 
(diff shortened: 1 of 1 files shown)


inoas:
Aside that - the MQ feature name never should have included the direction of the scale and an implication that you cannot request prefers-disabled-transparency etc - this is great and will mean authors can help those visually impaired.

cookiecrook:
@inoas wrote: 
> the MQ feature name never should have included the direction of the scale and an implication that you cannot request prefers-disabled-transparency

We talked about this in the other thread. It could always be expanded at a later with values that still make sense: For example…

```css
prefers-reduced-transparency: [ none |  reduce | force | opaque | bikeshed]
```

> this is great and will mean authors can help those visually impaired.
👍

mediaqueries-5/Overview.bs
```diff
@@ -478,6 +478,36 @@ Detecting the desire for less motion on the page: the 'prefers-reduced-motion' f
 
 	</dl>
 
+<h3 id="prefers-reduced-transparency">
+Detecting the desire to reduce the use of transparent or translucent layering effects on the page: the 'prefers-reduced-transparency' feature</h3>
```
frivoal: Could you shorten this title a bit? This a bit unwieldy, and does not need to reproduce the whole content of the section. How about:
> Detecting the desire for reduced transparency on the page: the 'prefers-reduced-transparency' feature

inoas: Transparency is a boolean concept; translucency and opacity are not. If the feature name contains `transparency` for the sake of easy vs more correct language, then the docs can easily contain `translucent` and/or `opacity` if authors are search the docs for those (more correct) terms.

But - as long as @cookiecrook hadn't had more specifics in mind - skipping the `layering effects` vs just `page` 👍 

> Detecting the desire for reduced transparency or translucency on the page: the 'prefers-reduced-transparency' feature

frivoal: > Transparency is a boolean concept; translucency and opacity are not.

I'm affraid I'll discagree on this one. I check the dictionary, and it gave not hit of this nuance. https://encrypted.google.com/search?q=partially+transparent&tbm=isch gives plenty of relevant results. Apple used this phrasing in their System Preferences to express the same concept. Using the word transparency is fine.

inoas: Okay, but clearly many UI effects nowadays combine a certain levels of opacity/transparency with certain levels of translucency (see example graphics on https://en.wikipedia.org/wiki/Transparency_and_translucency) and as long as this MQ should detect a combined preference then transparency for the name and translucency as part of the docs seems reasonable. My guess is still that apple picked "transparency" as that is an easy concept to get to every user.

`opacity` is btw the key word in CSS used to describe (non boolean shades of) *transparency of layers*, not transparency.

cookiecrook:
Change addressed.

frivoal:
@grorg You may be interested in this thing I just merged here. If something's wrong, please open a new issue, thx.

cookiecrook:
Thanks @frivoal

mediaqueries-5/Overview.bs
tabatkins: Agree that the standard English usage of all of these related terms are all roughly equivalent, and all encompass degrees.

For example, I was taught that opaque/translucent/transparent refer to the degree of opacity: something is translucent when light passes thru, but it's not clear enough to actually see details on the far side; something is transparent when you can actually see thru it to at least some degree.  (The boundary is obviously very fuzzy, no pun intended, but the categories are clear enough, no pun intended.  "Privacy glass" is more on the translucent side, cloudy plastic is translucent, etc.  Normal windows are transparent.  Rice paper is translucent.)

In other words, the definitions are definitely intertwined and uncertain enough that we can really use any of them here; there's no way to defend any particular definition as Definitely Correct.

PR #1721, opened 2017-08-15
[css-transforms-2] Refer to offset, not motion

ewilligers:
The shorthand defined by CSS Motion Path is named
'offset', not 'motion'.
https://drafts.fxtf.org/motion-1/#propdef-offset

The 'Current Transformation Matrix' no longer refers
to the obsolete name 'motion'.


nainar:
LGTM! :)

PR #1767, opened 2017-08-30
Ask for web-platform-tests in CONTRIBUTING.md

### LINKED ISSUE #1755: [all] Consider policy to ask for web-platform-tests
See https://github.com/w3c/fxtf-drafts/pull/215 for this for fxtf.

I would like to adopt this policy for cssom and cssom-view. Having tests and browser bugs for each normative spec change has turned out to be a really good way to reduce the time it takes for that change to be interoperably implemented, which is one of the main goals of standards.

We have lots of specs here. I don't know if this policy makes sense for all CSS specs. But I think it does for most of them. I would be happy to do this for only cssom and cssom-view for now, but I'd like the CSS WG to consider this for every spec, so we can decide what's best and apply that.

Case study: Chromium recently sent an [Intent to Implement and Ship for the 'q' unit](https://groups.google.com/a/chromium.org/d/msg/blink-dev/RRUYxYvsRCk/7VillBFrCAAJ), and found that there was only a test for it for ``, there was no CSS-specific test. With this policy, there would be some tests written and ideally also browser bugs filed asking to implement at the same time as it was added to the specification.

zcorpan:
(This is two commits for easier reviewing; should be squashed when merging.)

Closes #1755.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,3 +20,12 @@ If you added a contributor by mistake, you can remove them in a comment with:
 
 If you are making a pull request on behalf of someone else but you had no part in designing the 
 feature, you can remove yourself with the above syntax.
+
+# Tests
+
+For normative changes, a corresponding
+[web-platform-tests](https://github.com/w3c/web-platform-tests) PR is highly appreciated. Typically,
+both PRs will be merged at the same time. Note that a test change that contradicts the spec should
+not be merged before the corresponding spec change. If testing is not practical, please explain why
+and if appropriate [file an issue](https://github.com/w3c/web-platform-tests/issues/new) to follow
+up later. Add the `type:untestable` or `type:missing-coverage` label as appropriate.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ corresponding spec change. If testing is not practical, please explain why and i
 [file an issue](https://github.com/w3c/web-platform-tests/issues/new) to follow up later. Add the
 `type:untestable` or `type:missing-coverage` label as appropriate.
 
-The pre-CR specifications are currently:
+The pre-CR specifications with this testing requirement are currently:
 
 * cssom
 * cssom-view
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,11 +26,11 @@ feature, you can remove yourself with the above syntax.
 For normative changes for any specification in
 [CR or later](https://www.w3.org/Style/CSS/current-work) as well as the pre-CR specifications listed
 below, a corresponding [web-platform-tests](https://github.com/w3c/web-platform-tests) PR must be
-provided; for other specifications it is highly appreciated. Typically, both PRs will be merged at
-the same time. Note that a test change that contradicts the spec should not be merged before the
-corresponding spec change. If testing is not practical, please explain why and if appropriate
-[file an issue](https://github.com/w3c/web-platform-tests/issues/new) to follow up later. Add the
-`type:untestable` or `type:missing-coverage` label as appropriate.
+provided, except if testing is not practical; for other specifications it is usually appreciated.
+Typically, both PRs will be merged at the same time. Note that a test change that contradicts the
+spec should not be merged before the corresponding spec change. If testing is not practical, please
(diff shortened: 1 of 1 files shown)


frivoal:
Someone else than me ( @astearns or @atanassov  ?) should review this, but looks good to me.

CONTRIBUTING.md
```diff
@@ -20,3 +20,19 @@ If you added a contributor by mistake, you can remove them in a comment with:
...
+For normative changes for any specification in
+[CR or later](https://www.w3.org/Style/CSS/current-work) as well as the pre-CR specifications listed
+below, a corresponding [web-platform-tests](https://github.com/w3c/web-platform-tests) PR must be
+provided; for other specifications it is highly appreciated. Typically, both PRs will be merged at
+the same time. Note that a test change that contradicts the spec should not be merged before the
+corresponding spec change. If testing is not practical, please explain why and if appropriate
+[file an issue](https://github.com/w3c/web-platform-tests/issues/new) to follow up later. Add the
+`type:untestable` or `type:missing-coverage` label as appropriate.
+
+The pre-CR specifications are currently:
```
astearns: How about "The pre-CR specifications with this testing requirement are currently:"

zcorpan: So changed.

CONTRIBUTING.md
```diff
@@ -20,3 +20,19 @@ If you added a contributor by mistake, you can remove them in a comment with:
 
 If you are making a pull request on behalf of someone else but you had no part in designing the 
 feature, you can remove yourself with the above syntax.
+
+# Tests
+
+For normative changes for any specification in
+[CR or later](https://www.w3.org/Style/CSS/current-work) as well as the pre-CR specifications listed
+below, a corresponding [web-platform-tests](https://github.com/w3c/web-platform-tests) PR must be
```
foolip: I've been asking for a policy like this, so obvious this makes me happy, but I think it's important that there's some leeway here. The "testing is not practical" case will probably come up at least occasionally at CR or later, and if everyone agrees, then this "must" ought not get in the way of progress. I think that if this is downgraded to a "should", then editor's can use their best judgement about when to file issues instead of requiring tests.

astearns: The way I'm reading it, the 'must' is to have a wpt PR *or* an explanation/issue for followup. It's still at the editor's discretion, but they must do one or the other.

foolip: Perhaps I'm reading this too much like a format spec, but I'm not able to parse it that way, "a corresponding PR must be provided" sounds quite final. Of course, it's only with RFC 2119 goggles that changing it to "should" helps. Interested to hear how others read this, and don't mind it if other have a more lenient interpretation.

fantasai: I think if the intent is the way @astearns is reading it, then it needs to be reworded to be more clear about the either-or.

CONTRIBUTING.md
```diff
@@ -20,3 +20,19 @@ If you added a contributor by mistake, you can remove them in a comment with:
...
+provided; for other specifications it is highly appreciated. Typically, both PRs will be merged at
```
fantasai: “for other specifications it is highly appreciated” I would probably downgrade this a bit to maybe s/highly/often/. I would not have appreciated people submitting tests to CSS Grid in the earlier exploratory stages of the spec: that just means more work for for us to go through the for-sure outdated and possibly entirely useless tests later on--and discourages submissions without tests even in cases where they wouldn't actually be helpful.

CONTRIBUTING.md
```diff
@@ -20,3 +20,19 @@ If you added a contributor by mistake, you can remove them in a comment with:
...
+the same time. Note that a test change that contradicts the spec should not be merged before the
+corresponding spec change. If testing is not practical, please explain why and if appropriate
+[file an issue](https://github.com/w3c/web-platform-tests/issues/new) to follow up later. Add the
+`type:untestable` or `type:missing-coverage` label as appropriate.
+
+The pre-CR specifications with this testing requirement are currently:
+
+* cssom
+* cssom-view
```
fantasai: css-overflow-3 and css-position-3 will probably fall under this list as well, although I haven't looked at them lately, so fine to leave out for now.

zcorpan:
I've tweaked the wording, PTAL.

Issue #1783, opened 2017-09-07
[cssom-view] Does scroll-behavior affect the sequential focus navigation?


jihyerish:
CSSOM Spec says about `scroll-behavior`, : 

> The scroll-behavior property specifies the scrolling behavior for a scrolling box, when scrolling happens due to navigation or CSSOM scrolling APIs. Any other scrolls, e.g. those that are performed by the user, are not affected by this property. 

Does "**navigation**" here also include the [sequential focus navigation](https://html.spec.whatwg.org/#sequential-focus-navigation) triggered by pressing <kbd>Tab</kbd>?

When I see the HTML Spec and the implementation related to  `scroll-behavior`, it only works for [navigation to a fragment](https://html.spec.whatwg.org/#scroll-to-fragid), not for  the [sequential focus navigation](https://html.spec.whatwg.org/#sequential-focus-navigation).
But I think the [sequential focus navigation](https://html.spec.whatwg.org/#sequential-focus-navigation) also need to be affected by this property.

jihyerish:
cc @zcorpan @RByers 
There is a discussion about customizing scroll behavior of Element.focus(), https://github.com/whatwg/html/issues/834.
Could we also customize the scroll behavior of sequential focus navigation with `scroll-behavior`?

zcorpan:
Including sequential focus navigation and focus() makes sense to me and it has been requested before. Right now it falls under "any other scrolls".

https://bugzilla.mozilla.org/show_bug.cgi?id=1257600
https://lists.w3.org/Archives/Public/www-style/2016Mar/0055.html
https://github.com/whatwg/html/issues/94#issuecomment-198020847

@mstange

jihyerish:
@zcorpan Thanks for the references! 
Can we add some description to the spec such as:
- CSSOM View: 
  - Types of navigation which can specify the scrolling behavior
  - Types of scroll which are not affected by `scroll-behavior`
- HTML: Add a step about adjusting `scroll-behavior` in the [sequential focus navigation algorithm](https://html.spec.whatwg.org/#sequential-focus-navigation)

Issue #1720, opened 2017-08-14
[cssom-view] scrollIntoView({}) dictionary default should match scrollIntoView(true)


annevk:
Why isn't the IDL as follows:

    void scrollIntoView(optional (boolean or ScrollToOptions) arg = false);

Overloading is better to avoid when possible.

And the dictionary having different defaults than false also seems wrong. Passing {} shouldn't have different results from passing undefined or passing nothing at all.

(Via https://bugzilla.mozilla.org/show_bug.cgi?id=1389274.)

zcorpan:
(Commenting here instead of creating a duplicate issue.)

https://drafts.csswg.org/cssom-view/#extension-to-the-element-interface

annevk suggests the dictionary defaults for scrollIntoView should be the same as omitting the argument and passing in `true` (which is start-nearest). Currently the dictionary defaults are instead center-center, because in https://www.w3.org/Bugs/Public/show_bug.cgi?id=17152 it was argued to be a common case so there should be a convenient way to get center-center.

Due to legacy we can't change the behavior of omitted argument or passing `true`, but I think we can still change the dictionary defaults since they are not yet interoperably implemented. I also think having the defaults match between dictionary and boolean makes for a less surprising and easier to learn API.


| Test case                               | EdgeHTML | Blink | Gecko | WebKit |
|-----------------------------------------|----------|-------|-------|--------|
| http://software.hixie.ch/utilities/js/live-dom-viewer/saved/5300     | top         | center      | top      |  top      |

Current spec:
```
dictionary ScrollIntoViewOptions : ScrollOptions {
  ScrollLogicalPosition block = "center";
  ScrollLogicalPosition inline = "center";
};
```
Suggested change:
```
dictionary ScrollIntoViewOptions : ScrollOptions {
  ScrollLogicalPosition block = "start";
  ScrollLogicalPosition inline = "nearest";
};
```

Related: https://github.com/w3c/csswg-drafts/issues/1367

zcorpan:
cc @SaschaNaz @bzbarsky @travisleithead @RByers

bzbarsky:
Having {} match undefined sounds fine to me from first principles; it's certainly least-surprising.  I can't speak to whether that trumps the reasoning from https://www.w3.org/Bugs/Public/show_bug.cgi?id=17152

zcorpan:
This issue is blocked on
https://github.com/w3c/csswg-drafts/pull/1505
[...]

PR #1805, opened 2017-09-12
[cssom-view] Add notIfViewed to ScrollIntoViewOptions; add FocusScrollOptions

zcorpan:
notIfViewed was originally suggested in
http://www.w3.org/mid/CAE3TfZPmNWvz1z7XPqFjtg5S+m_9BOmNNHsOZuSrR2z2AgyNHA@mail.gmail.com

Now it is needed for HTMLElement focus(options) to match the behavior
of focus() in browsers. Since the defaults for focus() is different
from the defaults for scrollIntoView(), a new dictionary
FocusScrollOptions is introduced. For now the defaults for block
and inline are UA-defined.

https://github.com/whatwg/html/pull/2787#issuecomment-328397430

### DIFF
--- a/cssom-view/Overview.bs
+++ b/cssom-view/Overview.bs
@@ -1073,14 +1073,15 @@ Note: This {{DOMRect}} object is not <a spec=html>live</a>.
 <h2 id=extension-to-the-element-interface>Extensions to the {{Element}} Interface</h2>
 
 <pre class=idl>
+enum ScrollIntoViewMode { "always", "if-needed" };
 enum ScrollLogicalPosition { "start", "center", "end", "nearest" };
 dictionary ScrollIntoViewOptions : ScrollOptions {
-  boolean notIfViewed = false;
+  ScrollIntoViewMode scroll = "always";
   ScrollLogicalPosition block = "start";
   ScrollLogicalPosition inline = "nearest";
 };
 dictionary FocusScrollOptions : ScrollOptions {
-  boolean notIfViewed = true;
+  ScrollIntoViewMode scroll = "if-needed";
   ScrollLogicalPosition block;
   ScrollLogicalPosition inline;
 };
@@ -1153,16 +1154,16 @@ The <dfn method for=Element caniuse=scrollintoview>scrollIntoView(<var>arg</var>
 1. Let <var>behavior</var> be "<code>auto</code>".
 1. Let <var>block</var> be "<code>start</code>".
 1. Let <var>inline</var> be "<code>nearest</code>".
-1. Let <var>notIfViewed</var> be false.
+1. Let <var>scrollMode</var> be "<code>always</code>".
 1. If <var>arg</var> is a {{ScrollIntoViewOptions}} dictionary, then:
     1. Set <var>behavior</var> to the {{ScrollOptions/behavior}} dictionary member of <var>options</var>.
     1. Set <var>block</var> to the {{ScrollIntoViewOptions/block}} dictionary member of <var>options</var>.
     1. Set <var>inline</var> to the {{ScrollIntoViewOptions/inline}} dictionary member of <var>options</var>.
-    1. Set <var>notIfViewed</var> to the {{ScrollIntoViewOptions/notIfViewed}} dictionary member of <var>options</var>.
+    1. Set <var>scrollMode</var> to the {{ScrollIntoViewOptions/scroll}} dictionary member of <var>options</var>.
 1. Otherwise, if <var>arg</var> is false, then set <var>block</var> to "<code>end</code>".
 1. If the element does not have any associated <a>layout box</a>, then return.
 1. <a lt='scroll an element into view'>Scroll the element into view</a>
-    with <var>behavior</var>, <var>block</var>, <var>inline</var>, and <var>notIfViewed</var>.
+    with <var>behavior</var>, <var>block</var>, <var>inline</var>, and <var>scrollMode</var>.
 1. Optionally perform some other action that brings the element to the user's attention.
 
 The <dfn method for=Element lt="scroll(options)|scroll(x, y)">scroll()</dfn> method must run these steps:
@@ -1315,7 +1316,7 @@ The <dfn attribute for=Element>clientHeight</dfn> attribute must run these steps
 
 To <dfn>scroll an element into view</dfn> <var>element</var>,
 with a scroll behavior <var>behavior</var>,
-a boolean indicating to not scroll if the element is already in view <var>notIfViewed</var>,
+a mode indicating whether to scroll if the element is already in view <var>scrollMode</var>,
 optionally a block flow direction position <var>block</var> (undefined if not given),
 and optionally an inline base direction position <var>inline</var> (undefined if not given),
 means to run these steps for each ancestor element or <a>viewport</a> that establishes
@@ -1329,7 +1330,7 @@ a <a>scrolling box</a> <var>scrolling box</var>, in order of innermost to outerm
 1. Let <var>scrolling box edge D</var> be the <a>ending edge</a> in the <a>inline base direction</a> of <var>scrolling box</var>, and let <var>element edge D</var> be <var>element bounding border box</var>'s edge on the same physical side as that of <var>scrolling box edge D</var>.
 1. Let <var>element width</var> be the distance between <var>element edge C</var> and <var>element edge D</var>.
 1. Let <var>scrolling box width</var> be the distance between <var>scrolling box edge C</var> and <var>scrolling box edge D</var>.
-1. If <var>notIfViewed</var> is true, and <var>element</var> is entirely in view already, then return.
+1. If <var>scrollMode</var> is "<code>if-needed</code>", and <var>element</var> is entirely in view already, then return.
 
         ISSUE: Define "entirely in view".
 
--- a/cssom-view/Overview.bs
+++ b/cssom-view/Overview.bs
@@ -1075,8 +1075,14 @@ Note: This {{DOMRect}} object is not <a spec=html>live</a>.
 <pre class=idl>
 enum ScrollLogicalPosition { "start", "center", "end", "nearest" };
 dictionary ScrollIntoViewOptions : ScrollOptions {
+  boolean notIfViewed = false;
   ScrollLogicalPosition block = "start";
   ScrollLogicalPosition inline = "nearest";
+};
+dictionary FocusScrollOptions : ScrollOptions {
+  boolean notIfViewed = true;
+  ScrollLogicalPosition block;
+  ScrollLogicalPosition inline;
 };
 
 partial interface Element {
@@ -1099,6 +1105,13 @@ partial interface Element {
   readonly attribute long clientHeight;
 };
 </pre>
+
+The {{ScrollIntoViewOptions}} dictionary is used by {{Element/scrollIntoView()}} and similar APIs
+that primarily scroll something into view.
+
+The {{FocusScrollOptions}} dictionary is similar to {{ScrollIntoViewOptions}}, but has different
+defaults, and is used by {{HTMLElement}}'s {{HTMLElement/focus()}} and similar APIs that primarily
+move focus to something, or equivalent, and secondarily scroll it into view.
 
 The <dfn method for=Element>getClientRects()</dfn> method, when invoked, must return the result of the following algorithm:
 
@@ -1140,14 +1153,16 @@ The <dfn method for=Element caniuse=scrollintoview>scrollIntoView(<var>arg</var>
 1. Let <var>behavior</var> be "<code>auto</code>".
 1. Let <var>block</var> be "<code>start</code>".
 1. Let <var>inline</var> be "<code>nearest</code>".
+1. Let <var>notIfViewed</var> be false.
 1. If <var>arg</var> is a {{ScrollIntoViewOptions}} dictionary, then:
     1. Set <var>behavior</var> to the {{ScrollOptions/behavior}} dictionary member of <var>options</var>.
     1. Set <var>block</var> to the {{ScrollIntoViewOptions/block}} dictionary member of <var>options</var>.
     1. Set <var>inline</var> to the {{ScrollIntoViewOptions/inline}} dictionary member of <var>options</var>.
+    1. Set <var>notIfViewed</var> to the {{ScrollIntoViewOptions/notIfViewed}} dictionary member of <var>options</var>.
 1. Otherwise, if <var>arg</var> is false, then set <var>block</var> to "<code>end</code>".
 1. If the element does not have any associated <a>layout box</a>, then return.
 1. <a lt='scroll an element into view'>Scroll the element into view</a>
-    with <var>behavior</var>, <var>block</var>, and <var>inline</var>.
+    with <var>behavior</var>, <var>block</var>, <var>inline</var>, and <var>notIfViewed</var>.
 1. Optionally perform some other action that brings the element to the user's attention.
 
 The <dfn method for=Element lt="scroll(options)|scroll(x, y)">scroll()</dfn> method must run these steps:
@@ -1300,8 +1315,9 @@ The <dfn attribute for=Element>clientHeight</dfn> attribute must run these steps
 
 To <dfn>scroll an element into view</dfn> <var>element</var>,
 with a scroll behavior <var>behavior</var>,
-a block flow direction position <var>block</var>,
-and an inline base direction position <var>inline</var>,
+a boolean indicating to not scroll if the element is already in view <var>notIfViewed</var>,
+optionally a block flow direction position <var>block</var> (undefined if not given),
+and optionally an inline base direction position <var>inline</var> (undefined if not given),
 means to run these steps for each ancestor element or <a>viewport</a> that establishes
 a <a>scrolling box</a> <var>scrolling box</var>, in order of innermost to outermost <a>scrolling box</a>:
 
@@ -1313,6 +1329,15 @@ a <a>scrolling box</a> <var>scrolling box</var>, in order of innermost to outerm
 1. Let <var>scrolling box edge D</var> be the <a>ending edge</a> in the <a>inline base direction</a> of <var>scrolling box</var>, and let <var>element edge D</var> be <var>element bounding border box</var>'s edge on the same physical side as that of <var>scrolling box edge D</var>.
 1. Let <var>element width</var> be the distance between <var>element edge C</var> and <var>element edge D</var>.
 1. Let <var>scrolling box width</var> be the distance between <var>scrolling box edge C</var> and <var>scrolling box edge D</var>.
+1. If <var>notIfViewed</var> is true, and <var>element</var> is entirely in view already, then return.
+
+        ISSUE: Define "entirely in view".
+
+1. If <var>block</var> is undefined, set <var>block</var> to a UA-defined value.
+1. If <var>inline</var> is undefined, set <var>inline</var> to a UA-defined value.
+
+        ISSUE: Define defaults for <var>block</var> and <var>inline</var> for {{FocusScrollOptions}}.
+
 1. Let <var>position</var> be the scroll position <var>scrolling box</var> would have by following these steps:
 
     1. If <var>block</var> is "<code>start</code>", then align <var>element edge A</var> with <var>scrolling box edge A</var>.


zcorpan:
Note that this introduces a boolean dictionary member with default value `true`, which is generally not recommended.

> It is strongly suggested not to use a default value of true for boolean-typed arguments, as this can be confusing for authors who might otherwise expect the default conversion of undefined to be used (i.e., false).

https://heycam.github.io/webidl/#dfn-optional-argument-default-value

However, the alternative would be to use different members that both default to `false` but have different semantics (i.e., `notIfViewed = false` and `evenIfViewed = false`), which also seems like it would cause confusion. cc @tobie

tobie:
That's a tough one. There's no enum that would work, is there?

zcorpan:
@tobie suggested an alternative name: `unlessViewed`.

I'm open for suggestions with enums or other API shapes.

tobie:
This? :-/

```js
scrollIntoView({ scroll: "always" });
scrollIntoView({ scroll: "only-when-hidden" });
```

espadrine:
I like `only-when-hidden`. Maybe `scrollIntoView({ onlyWhenHidden: true })`?

tobie:
> Maybe scrollIntoView({ onlyWhenHidden: true })?

Well, then you're back with the boolean defaulting to true issue.

espadrine:
onlyWhenHidden would default to false, wouldn't it? It does the opposite of
ifNotViewed, which defaulted to true.

tobie:
Well, the issue is both conditions are the default in different APIs.

zcorpan:
```
focus({ preventScroll: true });
focus({ scrollOptions: { scroll: "always" } });
focus({ scrollOptions: { scroll: "if-needed" } });
scrollIntoView({ scroll: "always" });
scrollIntoView({ scroll: "if-needed" });
```
"if-needed" would be familiar for developers using `scrollIntoViewIfNeeded()`.

zcorpan:
OK updated the PR per my previous comment.

zcorpan:
Another approach could be to add "-if-needed" suffixes to the keywords in LogicalScrollPosition. The current `nearest` already has "if needed" semantics, as specified. This would give more control as it lets developers specify "if needed" separately for the two dimensions.

```
focus({ preventScroll: true });
focus({ scrollOptions: { block: "center", inline: "center" } });
focus({ scrollOptions: { block: "center-if-needed", inline: "center-if-needed" } });
scrollIntoView({ block: "start" });
scrollIntoView({ block: "start-if-needed" });
```
IDL
```
enum ScrollLogicalPosition { "start", "start-if-needed", "center", "center-if-needed", "end", "end-if-needed", "nearest" };
```
(No change to `nearest` for compat.)

tobie:
That looks interesting.

jihyerish:
I don't have a strong preference between these two: 
- Add `scroll` dictionary member scroll which can have `always` or `if-needed` as a value.
- Add "-if-needed" suffixes to the keywords in ScrollLogicalPosition

But it looks like the second choice has more benefits for developers.

Apart from that, when `focus({ scrollOptions: { scroll: "always" } });`, does `always` mean that the scrolling happens to the element wherever it is positioned?
Or does it work only excluding the element which is entirely out of the view?
I think `always` should work for the element which is entirely or partially in the view.

If it's not, there is no way to do that case.

Position | -- | --  | -- | -- 
-- | -- | -- | -- | --
Entirely in the view | O | X | X | X
Partially in the view | O | O | X | X
Entirely out of the view | O | O | O | X
IDL | focus({   scrollOptions: { scroll: "always" } }); | ?!  | focus({ scrollOptions: { scroll: "if-needed" } }); | focus({ preventScroll: true });

zcorpan:
The proposal here says to abort if the element is entirely in view and "if-needed" is specified. If I understand correctly that should cover your use case. But always doesn't mean that scrolling will always take place--e.g., if nearest/nearest is specified and the element is already in view, no scrolling happens. (Or if there's no overflow, there's nothing to scroll.)

jihyerish:
Sorry, I mixed it up. I thought if "if-needed" is specified and the element is **entirely out of view**, then scrolling happens.
Then, the use case for scrolling only when the element is entirely out of view cannot be supported with this, right? That case is what I wanted to solve at the very beginning. (https://github.com/whatwg/html/issues/834#issuecomment-295643901)
Also in Firefox, Element.focus() works like that.

So I think it also need to cover this case.

jihyerish:
Currently, the result of Element.focus() on browsers is like this:

 Position of an element | Chrome | Firefox
-- | -- | --
Entirely in the view | No scrolling | No scrolling
Partially in the view | Scrolling (align: nearest) | No scrolling
Entirely out of the view | Scrolling (align: center) |  Scrolling (align: nearest)

You can see how it works [here](https://jihyerish.github.io/all-about-focus/).

zcorpan:
OK, then I think I have misunderstood the requirements. I thought you were happy with a way to turn off scrolling entirely (with `preventScroll` in `focus()`).

So you want to be able to have no scrolling for partially in view, and scroll into view for entirely out of view, correct?

Maybe instead of having increasingly complicated dictionaries to cover all possible cases, we could have *only* `preventScroll` in `focus()` (and `select()` or whatever else), then a separate API to check if an element is entirely in view, partially in view, or entirely out of view, and finally you can call (or not) `scrollIntoView()` with the positioning you want for the given case?

jihyerish:
> So you want to be able to have no scrolling for partially in view, and scroll into view for entirely out of view, correct?

Yes, no scrolling for partially and entirely in view.
The separate API can solve the problem here. Then will it be the new API?

anawhj:
At this stage, I would like to summarize the current status and suggest some opinions on this PR.
I attached some comments(//) into the WebIDL below to clarify the meaning of the terms.
```javascript
enum ScrollBehavior { "auto", "instant", "smooth" };
enum ScrollIntoViewMode { "always", "if-needed" }; // 1. prefer "if-needed" than suffixes"
enum ScrollLogicalPosition { "start", "center", "end", "nearest" };

dictionary ScrollOptions {
  ScrollBehavior behavior = "auto";
};

dictionary ScrollIntoViewOptions : ScrollOptions {
  ScrollIntoViewMode scroll = "always";
  ScrollLogicalPosition block = "start";
  ScrollLogicalPosition inline = "nearest";
};

dictionary FocusScrollOptions : ScrollOptions {
  ScrollIntoViewMode scroll = "if-needed"; // 2. scroll to scrollMode
  ScrollLogicalPosition block = "start"; // 3. default value: same as ScrollIntoViewOptions
  ScrollLogicalPosition inline = "nearest";
};

dictionary FocusOptions {
  boolean preventScroll = false;
  FocusScrollOptions scrollOptions;
};

void focus(optional FocusOptions options);

// For instances (applied the comments above)
elm.focus();
elm.focus({preventScroll: false});
elm.focus({scrollOptions: {behavior: "smooth"}});
elm.focus({scrollOptions: {behavior: "smooth", scrollMode: "always", block: "center"}});
elm.scrollIntoView({behavior: "smooth", scrollMode: "if-hidden", block: "nearest"});
```

First of all, the current IDL seems so complicated enough to support the scrolling options for the focus function, so I totally agree @zcorpan's opinion to prevent having increasingly complicated dictionaries.

Honestly, before the UA support for smooth scrolling, the feature has been supported by several libraries or polyfills using requestAnimationFrame(). However, in case of scrolling options for the focus function, there seems no way to control the scrolling behavior when the focus is moved to another element due to several reasons such as focus-related events with no cancelable. In addition, the important thing is that developers want to use the native focus function in order to support the focus-related a11y features and one of the biggest challenges is to handle the scrolling.

With such a background, I would strongly suggest that focus function can embrace the FocusScrollOptions dictionary as well as the preventScroll property. The WebIDL above couldn't be happier to me, while the case of partially view would be handled by other API (e.g. IntersectionObserver).

Regarding the WebIDL above, I would like to put some comments, but it's not the strong opinions. The basic design of WebIDL seems well-organized enough to me.

1. **prefer "if-needed" than suffixes of "-if-needed",  prefer "if-hidden" than "if-needed"**
We need an option to specify whether to trigger scrolling or not when the element is fully in the view. Basically, scrollIntoView() enables the scrolling in that case, but focus() doesn't, so we defined the FocusScrollOption dictionary with different default value from the ScrollIntoViewOptions dictionary. As I see, there seems no correlation between "if-needed" that can control to happen scrolling and the ScrollLogicalPosition dictionary that specifies the position after scrolling. In many cases, developers want only to use the "if-needed" option, not to use the "block" or "inline" options, so I'm in favor of the ScrollIntoViewMode approach than the suffixes of "-if-needed" approach.
In addition, it would be difficult to understand the meaning of "if-**needed**" even now. I agree on the design of "if-needed" in consideration of `scrollIntoViewIfNeeded", but it's not the standard API (only support in WebKit and Blink). If this work proceed well, scrollIntoView with "if-needed" could be same as the scrollIntoViewIfNeeded so that it could be deprecated naturally. It seems so difficult to define a new term instead of "if-needed", but I suggest "if-hidden" that means **scrolling only if the element is hidden (partially or entirely in the view)**. I guess "if-hidden" could be used in scrollIntoView method to support several use cases while it has even been supported in JS libraries. Surely no problem to me with "if-needed" as is for BC.

2. **scroll to scrollMode**
`scroll` property name as the ScrollIntoViewMode type seems kind of general term, so I prefer to change the term like scrollMode.
```javascript
elm.scrollIntoView({behavior: "smooth", scroll: "if-needed"});
elm.scrollIntoView({behavior: "smooth", scrollMode: "if-needed"});
```

3. **The default values of FocusScrollOptions would be same as ScrollIntoViewOptions**
{block: start, inline: nearest} would be good for FocusScrollOptions (same as ScrollIntoViewOptions)

zcorpan:
Thank you for your analysis and comments @anawhj! 

Let me try to state my understanding, so we can reach common ground.

* `preventScroll: true` + IntersectionObserver + `scrollIntoView()` handles all use cases, but might not be so convenient.
* Making https://drafts.csswg.org/cssom-view/#smooth-scrolling apply to scrolls caused by moved focus would be a convenient way to enable smooth scrolling for `focus()`. https://github.com/w3c/csswg-drafts/issues/1783
* Customizing scrolling for `focus()` in a convenient manner, but less clear and less powerful compared to the first item in this list, could be done with a nested dictionary. The semantics and naming here is still under discussion. We agree that it shouldn't cover differentiating for the partially visible case, since that would make it too complicated.

To your points:

1. <blockquote>I suggest "if-hidden" that means scrolling only if the element is hidden (partially or entirely in the view).</blockquote>
   
   Sounds good to me.
2. <blockquote> scroll to scrollMode </blockquote>
   
   Sounds good to me. 
3. <blockquote> The default values of FocusScrollOptions would be same as ScrollIntoViewOptions</blockquote> 
   
   Why? The primary reason for introducing `FocusScrollOptions`, as explained in https://github.com/whatwg/html/pull/2787#issuecomment-328801350 , is that omitting the dictionary should be equivalent to specifying an empty dictionary. Compare with https://github.com/w3c/csswg-drafts/issues/1720 . As noted by @jihyerish above https://github.com/w3c/csswg-drafts/pull/1805#issuecomment-331383688 the current browser behavior for `focus()` does not match that of `scrollIntoView()`. Therefore, the defaults for focus need to be different. Since we don't yet have interoperability on how `focus()` scrolls, what those defaults should be is up to the UA. We should work towards interoperability on that, but that can be separated imo.

zcorpan:
I rebased this branch (for `cssom-view` → `cssom-view-1` rename) and added 2 new commits to rename `if-needed` to `if-hidden`, made it not scroll when partially visible, and renamed `scroll` to `scrollMode`, per above comments.

frivoal:
> preventScroll: true + IntersectionObserver + scrollIntoView() handles all use cases, but might not be so convenient.

I agree both that it would work, and that it's not so convenient.

Rather than adding another boolean or another value to an enum in the dictionary, which would  force us to think hard about naming to make the whole thing non confusing, couldn't we add a numeric `threshold`, with the same meaning as the `threshold` in `IntersectionObserver`, applied only in the `if-hidden` case?

That seems to cover everything, without excessive complexity either for authors who want to use this, nor for those who don't care, nor on the implementation side.

zcorpan:
@frivoal I think that wouldn't help particularly. At least it wouldn't help to achieve different alignment between partially in view and hidden, like in Chrome for `focus()`.

zcorpan:
I took a stab at defining how `if-hidden` should work, in terms of "element edge A" et. al. that the algorithm uses already for calculating scroll position.

zcorpan:
It would be good to have the WG give input on how we should reach interoperability on how `focus()` scrolls; see @jihyerish [comment above](https://github.com/w3c/csswg-drafts/pull/1805#issuecomment-331383688) with browser results.

As the state is right now, the spec (with this PR) can't explain the Chrome, Opera, Safari behavior. The Firefox, Edge behavior is like:
```
// elm.focus() is equivalent to
elm.focus({ scrollOptions: { scrollMode: "if-hidden", block: "nearest", inline: "nearest" }});
```

domenic:
I think it should be mode, instead of scrollMode? It's redundant in all of the method calls it's used in so far.

I see this thread is full of lots of bikeshedding already, so apologies for my contribution to it.

jihyerish:
Thanks for the updates @zcorpan! I think these are the remaining issues in here:

1.  Define how to handle the partially viewed element
This is difficult to handle. As @zcorpan pointed out, covering all the cases (mentioned in https://github.com/w3c/csswg-drafts/pull/1805#issuecomment-331109979) is complicated.
So I'm happy with the definition of `if-hidden` which can scroll the element if it's entirely out of view.
IntersectionObserver can help with the partially viewed element.

2. Default values of FocusScrollOptions
I prefer `{block: start, inline: nearest}` or `{block: nearest, inline: nearest}` as default values.
I hope we can get a resolution about this in the conf call. 
  - Defaults of ScrollIntoViewOptions: `{block: start, inline: nearest}`
  - Defaults of Firefox, Edge: block: `{block: nearest, inline: nearest}`
  - Defaults of Chrome, Opera, Safari: `{block: center, inline: center}`

frivoal:
@zcorpan 
>I think that wouldn't help particularly.

Would it not?

* `elm.focus({scrollMode: "if-hidden", threshold: 0 });` would trigger scrolling if even 1 pixel of elm is hidden
* `elm.focus({scrollMode: "if-hidden", threshold: 1 });` would trigger scrolling only if the whole elm is hidden
* `elm.focus({scrollMode: "if-hidden", threshold: 0.5 });` would trigger scrolling if the half or more of elm is hidden

That seems to work, and to cover @jihyerish 's use case just fine (as well as intermediate situations through values other than 0 or 1).

>  At least it wouldn't help to achieve different alignment between partially in view and hidden, like in Chrome for focus().

Fair enough, but neither do the other proposals discussed so far, and that didn't seem blocking.

dbaron:
For what it's worth, the [current state of Gecko's internal API for scrolling an element into view](https://searchfox.org/mozilla-central/rev/dca019c94bf3a840ed7ff50261483410cfece24f/layout/base/nsIPresShell.h#654-756) may be interesting.  In particular, it provides 3-way control (scroll-always, scroll-if-not-visible, scroll-if-not-fully-visible) that's independent for each axis.  And it provides many other things; it requires a bit of digging to figure out what they're all used for.

cssom-view-1/Overview.bs
```diff
@@ -1335,16 +1335,21 @@ a <a>scrolling box</a> <var>scrolling box</var>, in order of innermost to outerm
...
-        ISSUE: Define "partially or entirely in view".
-
 1. If <var>block</var> is undefined, set <var>block</var> to a UA-defined value.
 1. If <var>inline</var> is undefined, set <var>inline</var> to a UA-defined value.
 
         ISSUE: Define defaults for <var>block</var> and <var>inline</var> for {{FocusScrollOptions}}.
 
-1. Let <var>position</var> be the scroll position <var>scrolling box</var> would have by following these steps:
+1. Let <var>hidden</var> be false.
+1. If <var>element edge B</var> is outside (before) <var>scrolling box edge A</var>, or
```
anawhj: It seems that 'if-hidden' works only when an element is entirely outside the scrolling box, but I think the coverage of 'if-hidden' would be two cases like 'entirely out of view' and 'partially in view', while 'always' should support the two cases plus an 'entirely in view' as the three cases totally.

In other words, "if-hidden" means scrolling only if an element is hidden (partially or entirely out the view).

The cases like 'scrollIntoViewIfNeeded' method and 'notIfViewed' property works as above now.
https://lists.w3.org/Archives/Public/www-style/2012May/0808.html

In terms of real use cases, 'if-hidden' seems to work as above.

jihyerish: I also agree with 'if-hidden' works when an element is entirely or partially out of the view.
Sorry for the confusion!

zcorpan: OK, I must have misunderstood @anawhj earlier comment.

jihyerish:
FYI, [the fragment link scrolling](https://html.spec.whatwg.org/#scroll-to-fragid) works the same in all browsers as: 

Position of an element | display: block | display: inline
-- | -- | -- 
Entirely in the view | Scrolling to start | No scrolling
Partially in the view | Scrolling to start | Scrolling to nearest
Entirely out of the view | Scrolling to start | Scrolling to nearest

This could be a reference to decide the default values of `FocusScrollOptions`.

zcorpan:
We discussed this in the CSS WG telecon yesterday. Minutes at https://www.w3.org/2017/10/18-css-irc#T16-08-44

zcorpan:
Here is my recommendation on how to best proceed.

What we should focus on right now:

* Add only `preventScroll` to `focus()` (without `scrollOptions`), and try to get that adopted cross-browser. This should be an easier sell than the whole package sans consensus. See https://extensiblewebmanifesto.org/
* Develop a JS library that uses the primitives `preventScroll: true` + IntersectionObserver + `scrollIntoView()` to expose a convenient API for web developers.
* Ask browser vendors to consider if they want to get [full interoperability](https://docs.google.com/document/d/1LSuLWJDP02rlC9bOlidL6DzBV5kSkV5bW5Pled8HGC8/edit#heading=h.y7jw138np12j) on default scrolling behavior for `focus()`, and if so what the behavior should be. I can file issues.

What we can do later:

* When `preventScroll` is shipped and JS libraries exist to use it, study how these features are actually being used on the web, to inform what the for-convenience API should handle and how.
* If/when `focus()` default scrolling behavior is more interoperable, we have a stable foundation to extend the API with `scrollOptions` (or something) that can explain the default scrolling behavior and allow customizing alignment etc.

anawhj:
@zcorpan Thank you for the summary. I agree to add only `preventScroll` to focus(), while other features such as smooth scrolling, alignment in the scrolling box could be handled using other APIs.

In addition, `scrollOptions` such as scrolling behavior and customizing alignment would be investigated aligned with CSS Scroll Snap. Regarding interoperability of alignment(block/inline) for focus() summarized in [the comment](https://github.com/w3c/csswg-drafts/pull/1805#issuecomment-331383688) above, interests from browser vendors seems to be important. Hopefully it will be unified as a single solution, but we should carefully consider it as it has a significant influence in terms of web UX.

jihyerish:
@zcorpan Thanks for wrapping up the current issue in here. 
I'll be happy only if preventScroll is added to focus() (without scrollOptions) for right now.

Also, for the future work about FocusScrollOptions, it would be nice to discuss about: 
- Figuring out the default values of FocusScrollOptions
- Finding the way to handle the partially viewed element (for now, it's intersectionObserver but is there any better solution?)
- Considering the alignment with other focus cases. (scrolling by tab focusing, the fragment link scrolling, etc)
- Investigating the alignment with CSS Scroll Snap

zcorpan:
> Ask browser vendors to consider if they want to get [full interoperability](https://docs.google.com/document/d/1LSuLWJDP02rlC9bOlidL6DzBV5kSkV5bW5Pled8HGC8/edit#heading=h.y7jw138np12j) on default scrolling behavior for `focus()`, and if so what the behavior should be. I can file issues.

https://bugzilla.mozilla.org/show_bug.cgi?id=1410112
https://bugs.webkit.org/show_bug.cgi?id=178515
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14298983/
https://bugs.chromium.org/p/chromium/issues/detail?id=776368

Loirooriol:
It seems all comments consider "Partially in the view" as a single case, but it can be subdivided depending on whether there is available space to show a greater fraction of the element or not. The `"nearest"` mode scrolls in the former but not in the latter. Some people might want this kind of behavior for other modes.

stipsan:
Hey ya'll! I would love to see this move forward 😄 

Are we waiting for vendors to give their feedback or can we move forward and get their input later?

I maintain a related [polyfill for this proposal](https://github.com/stipsan/scroll-into-view-if-needed) and based on feedback I'm getting, and the growing popularity of the package itself, I think it's safe to say that developers are very interested.

If there's anything I can do to help move things forward then just say the word! 👋

jihyerish:
@stipsan 
Thanks for sharing your feedback, and the awesome polyfill! :+1: 
Yes, for now, we are gathering more feedback from vendors and developers.

As I mentioned in [here](https://github.com/w3c/csswg-drafts/pull/1805#issuecomment-337814855), I'm planning to do items on the list in the near future!

alshamiri5 (approved):

zcorpan:
Current status:

> https://bugzilla.mozilla.org/show_bug.cgi?id=1410112

No change, but @annevk said it seems ok to align with Chrome and Safari.

> https://bugs.webkit.org/show_bug.cgi?id=178515

No activity.

> https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14298983/

Edge has switched to Chromium.

> https://bugs.chromium.org/p/chromium/issues/detail?id=776368

Closed as WontFix ("there is no update from the spec discussion in more than 1 year").

frivoal:
This pull request is made with a branch in the main repo, rather than from a fork. This continuously puts strain on the build server, which is rebuilding it all the time to keep the bikeshed references current. This is OK for branches that need to have a preview, but this seems stale. It should be (updated and) merged or closed.

zcorpan:
Has there been any movement on the default scrolling behavior for `focus()`?

I think it's ok to close this, it can be reopened if there is new interest. We could file an issue so it doesn't get forgotten.

anawhj:
> Has there been any movement on the default scrolling behavior for focus()?

Yep! It has been merged via [a commit](https://github.com/whatwg/html/commit/7a579c1cf3fddc68a634e74c12c457af258e6a80.) in Oct 2017.
You can quickly understand how `element.focus({preventScroll: true});` work on [this](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/focus).

I believe that the valuable changes in this thread would be needed for many developers.
Its [polyfill](https://github.com/stipsan/scroll-into-view-if-needed) has been widely used for several use cases such as custom scroller, arrow keys based navigation.

In my side(LG webOS), there are requirements about the following functionalities.

A: `element.focus({block: start});`
(It can be used for both aligning it in a customized way and mediating the different default behaviors between Chrome and Firefox.)

B: `element.scrollIntoView({scroll: if-needed});` or `element.scrollIntoView({preventScroll: true});`.
(The latter could be used with IntersectionObserver for selecting the 'already inside view' case.)

All the cases are mostly covered with the already proposed changes in this thread.
How could we drive this item? What should we discuss with more?

zcorpan:
Moved to https://github.com/w3c/csswg-drafts/pull/5677

PR #1816, opened 2017-09-14
[cssom][cssom-view] Update editors

zcorpan:
See https://lists.w3.org/Archives/Public/www-style/2017Sep/0023.html


zcorpan:
"NOBODY (OOPS!)" was the intended result. 🙂

PR #1826, opened 2017-09-19
Fix typo (<var> wanted to be </var)

dholbert:
This fixes a typo in @fantasai's commit https://github.com/w3c/csswg-drafts/commit/1ddd3039929a92d20a63dd3ae5cbc6996b7fa702

(Without this typo-fix, a bunch of this paragraph ends up being incorrectly italicized.)


dholbert:
@fantasai, I didn't want to commit directly (though it looks like GitHub gives me the power) -- would you mind merging this if it looks good? It's just fixing a typo (we had an `<var>` open tag when we wanted a `</var>` close tag).

PR #1847, opened 2017-10-04
[css-break] Remove <i> autolinks

xfq:
According to tabatkins/bikeshed#1042, `<i>` autolinks will be removed in the future.

### DIFF
--- a/css-break/Overview.bs
+++ b/css-break/Overview.bs
@@ -58,7 +58,7 @@ Module Interactions</h3>
 <h3 id="values">
 Values</h3>
 
-  This specification follows the <a href="https://www.w3.org/TR/CSS21/about.html#property-defs">CSS property definition conventions</a> from [[!CSS2]].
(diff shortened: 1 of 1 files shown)


tabatkins:
👍

xfq:
Maybe I should change all the `[CSS21]` refs to `[CSS2]`, not the other way around...

Currently, most CSS 2 refs are `[CSS21]` and there's one `[CSS2]` (which was introduced in https://github.com/w3c/csswg-drafts/commit/7d80249c7df1d62b64e284902a61ebfa788302a0), and it causes a fatal error in Bikeshed:

```
FATAL ERROR: The biblio refs [[CSS2]] and [[CSS21]] are both aliases of the same base reference [[CSS2]]. Please choose one name and use it consistently.
```

tabatkins:
Either way works.

xfq:
OK. Thanks. I'll merge this PR if there's no objection until Tuesday.

PR #1857, opened 2017-10-10
[css-values] Use upper case for Q unit

ewilligers:
"The unit 'Q' is very common in Japanese typesetting"
https://lists.w3.org/Archives/Public/www-style/2013Nov/0302.html
http://tosche.net/2013/10/font-size-in-the-metric-system_e.html

### DIFF
--- a/css-values-3/Overview.bs
+++ b/css-values-3/Overview.bs
@@ -1048,13 +1048,13 @@ Viewport-percentage lengths: the ''vw'', ''vh'', ''vmin'', ''vmax'' units</h4>
 -->
 
 <h3 id="absolute-lengths">
-Absolute lengths: the ''cm'', ''mm'', ''q'', ''in'', ''pt'', ''pc'', ''px'' units</h3>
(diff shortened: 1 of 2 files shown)


frivoal:
By convention, units and keywords in css are specified in lower case, but as the spec says: "Like keywords, unit identifiers are ASCII case-insensitive". Therefore "q" and "Q" are the same thing, and it is already possible to use the unit with an upper case "Q" or a lower case "q", as the author prefers, even without this proposed change.

This proposed change has no normative effect, and is an editorial change to follow the casing convention of the "Q" unit instead of the casing convention of CSS.

I would be in favor of continuing to follow internal CSS conventions over external ones, but I will note that this convention is already not fully followed, as the "hz" unit is described as "Hz", so there is precedent.

ewilligers:
> This proposed change has no normative effect

Case is normative in Typed OM [factory function](https://drafts.css-houdini.org/css-typed-om-1/#numeric-factory) names - we should [choose](https://github.com/w3c/css-houdini-drafts/issues/504) between Q() Hz() and q() hz().

PR #1863, opened 2017-10-12
Delete a couple of stale files

frivoal:
These files seem to be temporary work file that have made it to the repo
by accident. Getting rid of them.


frivoal:
@fantasai It looks like these were committed by you. Safe to delete, or were these meant to be there?

frivoal:
Bah, merging. 99% sure it's just an accidental commit. If the files were meant to save some content for later, they're still in the repo history.

PR #1864, opened 2017-10-12
[css-break-3] Clarify the definition of widows and orphans

### LINKED ISSUE #1823: [css-break] Clarify definition of widow and orphans.
[The section on orphans, widows in css-break](https://drafts.csswg.org/css-break/#widows-orphans) says this:

> The `orphans` [resp. `widow`] property specifies the minimum number of line boxes in a block container that must be left in a fragment before [resp. after] a fragmentation break.
> [...]
> If a block contains fewer lines than the value of widows or orphans, the rule simply becomes that all lines in the block must be kept together. 

While it is clear to me what this is supposed to do if the block container only contains inline-level content, the meaning is less obvious when it has block level descendants.

Take this example:
```html
<style>div {  widows: 3; }</style>
<div>
  <p>
    line 1<br>
    line 2
  </p>
  <p>
    line 3<br>
    line 4
  </p>
</div>
```

https://www.w3.org/TR/css-display-3/#block-container is pretty clear that block containers can either contain online inline-level boxes or block level boxes, but the terminology isn't 100% clear (as far as I can tell) about whether a block container that contains block level elements can still be said to have "line boxes in it” (indirectly, via its block level children that themselves are block containers containing inlines), or if it should be considered not to have "line boxes in it”.

1. If the above spec prose is to be interpreted as applying only to line boxes that are direct children of  block containers that establish an inline formatting context, then `widows:3` forbids breaks inside both `<p>`, but does not prevent breaks between line 2 and 3: 

> If a block contains fewer lines than the value of widows or orphans, the rule simply becomes that all lines in the block must be kept together. 

The block has no (direct) lines, there's nothing to keep together, and the break is allowed.

2. If the above prose is meant to apply to line boxes that are descendants of the block container, at any depth (but presumably within the same BFC), the the break between line 3 and 4 would be forbidden.

---

I think the spec intends the first one. If that is the case, I suggest clarifying by changing the "Applies to" line to "Block container boxes that establish new inline formatting contexts" and/or by adding the following prose: “The `widows` and `orphan` properties have no effect on block container boxes that contains only block-level boxes”.

frivoal:
Closes #1823

### DIFF
--- a/css-break-3/Overview.bs
+++ b/css-break-3/Overview.bs
@@ -418,7 +418,7 @@ Breaks Between Lines: 'orphans', 'widows'</h3>
 
   <p>
     The 'widows' and 'orphans' properties have no effect
-    on <a>block container boxes</a> that contains only <a>block-level</a> boxes.
+    on <a>block container boxes</a> that contain only <a>block-level</a> boxes.
   </p>
   <p>
     The 'orphans' property specifies the minimum number
--- a/css-break-3/Overview.bs
+++ b/css-break-3/Overview.bs
@@ -412,10 +412,14 @@ Breaks Between Lines: 'orphans', 'widows'</h3>
   Name: orphans, widows
   Value: <<integer>>
   Initial: 2
-  Applies to: block containers
+  Applies to: <a>block containers</a> that establish new <a>inline formatting contexts</a>
   Inherited: yes
   </pre>
 
+  <p>
+    The 'widows' and 'orphans' properties have no effect
+    on <a>block container boxes</a> that contains only <a>block-level</a> boxes.
+  </p>
   <p>
     The 'orphans' property specifies the minimum number
     of line boxes in a block container


fantasai:
This is not confusing enough that it needs to be stated twice. Just fixing the Applies to line seems enough. If you're really insistent to duplicate the point, then make the paragraph after the propdef box a note *at the end of the section* so it's not before the actual description of what the property is for (which is way more important imho than pointing out where it doesn't apply).

Also, stop putting things that don't need WG discussion on the Agenda just because you want my attention. You have my phone number, if you're that desperate you can just call me!!! >:[

frivoal:
> This is not confusing enough that it needs to be stated twice. Just fixing the Applies to line seems enough. 

Done.

> Also, stop putting things that don't need WG discussion on the Agenda just because you want my attention. You have my phone number, if you're that desperate you can just call me!!!

Sorry, that wasn't the intent. I don't want your attention specifically, I just want feedback in general, and you're not the only editor of that spec. I do appreciate when I get your feedback, but I am not trying to single you out. Since you edit all the specs, I could call you anytime I want feedback on something, but that doesn't scale very well. That said, regardless of intent, if this is irking you, I'm doing it wrong. So point taken, I'll try to be more conservative when adding things to the agenda.

frivoal:
Oops. I intended to apply the fix the merge, but clicked merge before pushing the fix. I've [applied it on the master branch](https://github.com/w3c/csswg-drafts/commit/98927f19fe4e48b2c93d893741523e8d898529af), the the final result is OK, but sorry for messing that up.

Issue #1828, opened 2017-09-22
[css-pseudo-4] styling the spelling error marker


frivoal:
With the `::spelling-error` pseudo introduced [in css-pseudo-4](https://drafts.csswg.org/css-pseudo/#highlight-selectors), if you want for example to highlight spelling mistakes with a green wavy underline, you can do this:
```css
::spelling-error {
  text-decoration: wavy green underline;
}
```
However, this only works if the browser's original spelling mistake marker was created using the same mechanism in the UA stylesheet. If it is created some other way, what you'd get instead would be this:
![screenshot_from_2017-09-22_10-54-09](https://user-images.githubusercontent.com/113268/30741284-2b20a43a-9fcf-11e7-9458-e97e698eaa89.png)

Since that's not what the author would want, I think we should normatively require that browsers that support `::spelling-error` and `::grammar-error` must create the error indicators using these pseudos in the UA stylesheet.

tabatkins:
That's literally the entire intent of the feature, yes, so if it's not yet clear in the text, it should be made so.

(You don't have access to the spelling/grammar checker yourself, all you have is the ability to style its output via these pseudo-elements. If your styling didn't override the built-in one, it would be 100% useless.)

FremyCompany:
I agree

frivoal:
Added a pull request (#1881) to try and address this based on what we discussed during the call. Feedback appreciated

PR #1881, opened 2017-10-16
Spelling & grammar markers

frivoal:
This pull requests is an attempt to solve https://github.com/w3c/csswg-drafts/issues/1828, based on what I think we were converging towards in during the [conf call](https://github.com/w3c/csswg-drafts/issues/1828#issuecomment-334319887).

A possible variant of this would be to add the two new values to `text-decoration-style` instead of `text-decoration-line`. Arguably, if UAs wanted to allow it, `text-decoration: overline spelling-error` is slightly more likely to be a meaningful combination than `text-decoration: solid spelling-error`.

However, neither are motivated by real use cases, and the syntax I chose in this PR means the feature can be activated with just `text-decoration: spelling-error` or `text-decoration-line: spelling-error`, rather than having to specify `text-decoration: spelling-error underline`, expecting the underline part to be ignored, in order to avoid the longhand expansion to `text-decoration-line: none; text-decoration-style: spelling-error;` that would happen if it were left out.

### DIFF
--- a/css-text-decor-3/Overview.bs
+++ b/css-text-decor-3/Overview.bs
@@ -18,6 +18,7 @@ Abstract: This module contains the features of CSS relating to text decoration, 
 At Risk: The 'text-decoration-skip' property.
 At Risk: The line positioning rules.
 At Risk: The ability to place both emphasis marks and ruby on the same base text.
+At Risk: The ''spelling-error'' and ''grammar-error'' values of the 'text-decoration-line' property.
 Link Defaults: css-color-3 (property) color
 Use <i> Autolinks: yes
 </pre>
@@ -179,7 +180,7 @@ Text Decoration Lines: the 'text-decoration-line' property</h3>
 
   <pre class="propdef">
   Name: text-decoration-line
-  Value: none | [ underline || overline || line-through || blink ]
+  Value: none | [ underline || overline || line-through || blink ] | spelling-error | grammar-error
   Initial: none
   Inherited: no (but see prose, above)
   </pre>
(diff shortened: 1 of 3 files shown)


css-text-decor-3/Overview.bs
```diff
@@ -203,8 +204,27 @@ Text Decoration Lines: the 'text-decoration-line' property</h3>
         Conforming user agents may simply not blink the text.
         Note that not blinking the text is one technique to satisfy <a href="https://www.w3.org/TR/UAAG/guidelines.html#tech-on-off-blinking-text">checkpoint 3.3 of WAI-UAAG</a>.
         This value is <strong>deprecated</strong> in favor of Animations [[CSS3-ANIMATIONS]].
+    <dt><dfn value for="text-tecoration-line">spelling-error</dfn>
+      <dd>This values indicates the type of text decoration used by the User Agent
+          to highlight spelling mistakes.
+	  It's appearance is UA defined.
```
astearns: Its

css-text-decor-3/Overview.bs
```diff
@@ -203,8 +204,27 @@ Text Decoration Lines: the 'text-decoration-line' property</h3>
         Conforming user agents may simply not blink the text.
         Note that not blinking the text is one technique to satisfy <a href="https://www.w3.org/TR/UAAG/guidelines.html#tech-on-off-blinking-text">checkpoint 3.3 of WAI-UAAG</a>.
         This value is <strong>deprecated</strong> in favor of Animations [[CSS3-ANIMATIONS]].
+    <dt><dfn value for="text-tecoration-line">spelling-error</dfn>
+      <dd>This values indicates the type of text decoration used by the User Agent
```
astearns: value

css-text-decor-3/Overview.bs
```diff
@@ -203,8 +204,27 @@ Text Decoration Lines: the 'text-decoration-line' property</h3>
...
+          to highlight spelling mistakes.
+	  It's appearance is UA defined.
+	  <span class=note>It is often rendered as a red wavy underline.</span>
+    <dt><dfn value for="text-decoration-line">grammar-error</dfn>
+      <dd>This values indicates the type of text decoration used by the User Agent
```
astearns: value

css-text-decor-3/Overview.bs
```diff
@@ -203,8 +204,27 @@ Text Decoration Lines: the 'text-decoration-line' property</h3>
...
+          to highlight grammar mistakes.
+	  It's appearance is UA defined.
```
astearns: Its

PR #1884, opened 2017-10-17
[mediaqueries-4] Expand example for min-/max- shortcoming vs proper range context

patrickhlauke:
A follow-up to https://github.com/w3c/csswg-drafts/pull/1083 that further expands the example to also include use of high-precision values (while we wait for browsers to support range context queries)


patrickhlauke:
Apologies for rehashing this further, but as the use of decimal values seems the most robust workaround currently, it would be good to get it documented officially somewhere authoritative /cc @frivoal @tabatkins

frivoal:
Seems reasonable to me. Approved.

PR #1979, opened 2017-11-12
[selectors-4] edited major changes list

estelle:
In the css selectors 4 specification it reads under major changes since working draft: "Removed restriction on combinators within :matches(), :not(), :nth-match(), and :nth-last-match()."
:nth-match(), and :nth-last-match() are not in the current draft other than in this line, so it should likely read "Removed restriction on combinators within :matches() and not:not()"
and on another line:
"Removed :nth-match(), and :nth-last-match()"
location: https://drafts.csswg.org/selectors-4/#changes
Issue #1997

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -3597,7 +3597,8 @@ Changes</h2>
 					</ul>
 				</li>
 			</ul>
-		<li>Removed restriction on combinators within '':matches()'', '':not()'', '':nth-match()'', and '':nth-last-match()''.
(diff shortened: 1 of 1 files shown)


SelenIT:
Maybe the second line should read something like the following?
> Removed :nth-match(), and :nth-last-match() and introduced the An + B of S syntax for :nth-child() and :nth-last-child() instead, with no restriction on combinators for the S part

Also, maybe the Issue 22 after these lines can be now dropped, since these selector are already implemented (and even shipped) by WebKit?

PR #1986, opened 2017-11-14
[css-syntax] Trim whitespace from the beginning/ending of a declarati…

### LINKED ISSUE #774: [css-syntax] Trim whitespace around declarations?
Right now, the Syntax spec is clear about what the value of a declaration is - [it's all the tokens between the : and the ;](https://drafts.csswg.org/css-syntax/#consume-declaration).  In particular, this includes any initial or final whitespace, which is pretty common in declarations like `color: red;` or in omitted-semicolon situations like:

```css
.foo {
  color:red
}
```

When serializing non-custom properties, this whitespace doesn't matter; the property value has been parsed into a specialized data structure, and serialization just works off of that - it doesn't even attempt to preserve source whitespace.  This means, for example, that you can be sure the first character of 'color' will be a non-whitespace char.  

But the value of a custom property is the token list it parses as, so technically the initial/final whitespace is part of its value and should be serialized with it.  Add to this the resolution at TPAC that custom properties should store their source representations so they can reserialize *exactly* as written (rather than, in some situations, serializing with extra comments inserted).  On the other hand, this can result in unexpected whitespace in the serialization, as the initial/final whitespace isn't really considered to be "part of the value". (This is why I'm reporting it - one of the engineers on our team was writing some code with custom properties and got tripped up by this.)

So my question is: do we want to alter the ["consume a declaration" algorithm](https://drafts.csswg.org/css-syntax/#consume-declaration) to trim initial/final whitespace tokens? That dodges the resolution, as it won't be part of the value anymore. This will have zero effect on ordinary properties, just custom properties, and will likely bring custom property serialization slightly more in line with author expectations.

tabatkins:
…on's value. Fixes #774.

Needs a testcase before merging.

### DIFF
--- a/css-syntax-3/Overview.bs
+++ b/css-syntax-3/Overview.bs
@@ -2386,7 +2386,7 @@ Consume a declaration</h4>
 	<a>Consume the next input token</a>.
 	Create a new declaration
 	with its name set to the value of the <a>current input token</a>
-	and its value initially set to the empty list.
+	and its value initially set to the empty [=list=].
 
 	<ol>
 		<li>
@@ -2401,6 +2401,10 @@ Consume a declaration</h4>
 			Otherwise, <a>consume the next input token</a>.
 
 		<li>
+			While the <a>next input token</a> is a <<whitespace-token>>,
+			<a>consume the next input token</a>.
+
+		<li>
 			As long as the <a>next input token</a> is anything other than an <<EOF-token>>,
 			<a>consume a component value</a>
 			and append it to the declaration's value.
@@ -2411,6 +2415,10 @@ Consume a declaration</h4>
 			followed by an <<ident-token>> with a value that is an <a>ASCII case-insensitive</a> match for "important",
 			remove them from the declaration's value
 			and set the declaration's <var>important</var> flag to true.
+
+		<li>
+			While the last token in the declaration's value is a <<whitespace-token>>,
+			[=list/remove=] that token.
 
 		<li>
 			Return the declaration.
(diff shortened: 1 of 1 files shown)


css-syntax-3/Overview.bs
```diff
@@ -2386,7 +2386,7 @@ Consume a declaration</h4>
 	<a>Consume the next input token</a>.
 	Create a new declaration
 	with its name set to the value of the <a>current input token</a>
-	and its value initially set to the empty list.
+	and its value initially set to the empty [=list=].
```
kalenobr: [=list=]

csnardi:
Is there any reason why this hasn't been merged yet? It seems there is a testcase ready to be merged in WPT (with rebasing) at https://github.com/tabatkins/web-platform-tests/commit/1e27529ed3fa4c2cf0959ae236a74cec4562ee0a, and other WPT testcases can be adjusted at the same time as well.

tabatkins:
I never merged it because I never successfully got wpt running so that I could run and lint the test. ^_^  But if it's fine, I'm happy to merge.

csnardi:
Travis helpfully runs & lints tests for PRs to WPT, so you should be able to check by rebasing and opening a pull request.

I was just mentioning this as I was going through some of the Variables tests, which are partially incorrect per the current spec (or partially correct), but would be affected by this spec change.

tabatkins:
Ooh, didn't know. Will do so now.

mgol:
Either I'm reading it incorrectly or CSS Custom Property tests haven't been updated to account for this change; I submitted https://github.com/web-platform-tests/wpt/issues/34880.

Issue #1796, opened 2017-09-11
[css21][css-inline] Meta bug for line height


frivoal:
This is the meta bug to track a series of issues based on the findings of https://github.com/w3c/csswg-drafts/issues/938#issuecomment-314690847.

The general premise is: css21 is partly ambiguous, partly self-contradictory about how height measurements of non replaced inline level boxes are supposed to happen. Implementations are not fully interoperable. The goal here is to find what is interoperable, to agree on what to do for things that are not, and to fix the specs accordingly.

**Note that I am only testing what happens with a single inline element, although I am using fallback fonts. Multiple inline elements on the line is a separate discussion.**

frivoal:
Things all implementations agree about:
* https://github.com/w3c/csswg-drafts/issues/1797
* https://github.com/w3c/csswg-drafts/issues/1801
* https://github.com/w3c/csswg-drafts/issues/1802
* https://github.com/w3c/csswg-drafts/issues/1800
* https://github.com/w3c/csswg-drafts/issues/1804

Disagreements:
* https://github.com/w3c/csswg-drafts/issues/1798

TL;DR: We have interoperability in the general case both for the height of inline boxes, the position of the baseline therein, and for the height of the content area of inline boxes. Specs should be updated accordingly. The remaining cases where interop is lacking are odd corner cases, which we should harmonize and specify.

frivoal:
@astearns @atanassov How do you want to manage/chair this? I am reasonably confident in my conclusions, but this should get meaningful review from others before we amend 2.1. I can write Pull Requests for spec changes and for tests to go with it, and these should get close scrutiny, but presumably we want (at least) some high level discussions prior to that.

I can try to present this during a teleconf, and see if we can get a resolution. If we don't, I'll whiteboard this at TPAC. If we do, I write the PRs, ask for review, and we revisit during TPAC, hopefully to resolve to merge them.

Pinging a few people who should probably review this sooner or later: @atanassov @FremyCompany @fantasai @dbaron @smfr @tabatkins @litherum

kojiishi:
This is the one we talked about in Paris, correct? Can you point which one Blink behaves differently from other engines?

frivoal:
@kojiishi Yes, it's that one. The difference is actually limited to a more narrow case than I had thought at first, but it is there. Details in #1798.

kojiishi:
Ok, thanks, I'll take a look at #1798. Please let me know if there were other issues I should look at to see what Blink does.

[...]

PR #1993, opened 2017-11-16
Line height fixes

frivoal:
This is a series of patches to apply the lessons learned in https://github.com/w3c/csswg-drafts/issues/1796 and sub-issues.

The overarching theme is to remove contradictions and ambiguities from CSS2 about line-height and to align with implementations.

I will be submitting a corresponding test pull request shortly, and cross link it with this one when I'm done.


frivoal:
ping @bert-github @dbaron @fantasai @FremyCompany @litherum @svgeesus for review.

frivoal:
I edited both the source and the output because I made that edit while I my application for invited expert had not yet been processed, and I therefore did not have access rights to Bert's preprocessor. I believe the manual edit matches what the processor would have done. I'm happy to regenerate if that helps with anything.As for css2.1's directories, we currently have 2: one that matches the 2.1 REC (css21), and one that has all the changes folded in (css2). I edited the later. As some other point it time, we'll have to cherry pick things from there to make a new REC, but that's a separate discussion.

PR #1998, opened 2017-11-19
[css-overflow-3] Fix ED link

xfq:
Shepherd uses L4 for the unleveled URL.


frivoal:
@plinss Shepherd should probably not use L4 for the unleveled URL, as L3 is the "current" thing, and L4 is the speculative thing.

In the meanwhile, this PR looks good.

plinss:
Set level 3 to the "current version"

Issue #1983, opened 2017-11-14
[css-color-4] Spec doesn't specify whether implementation must honor <decimal> in rgb/rgba or if it can truncate to an <integer>


hichris1234:
[The RGB functions: rgb() and rgba()](https://drafts.csswg.org/css-color/#rgb-functions) states that:

> These values come from the fact that many graphics engines store the color channels internally as a single byte, which can hold integers between 0 and 255. However, the CSS syntax allows full \<number\>s, not just \<integer\>s, for authoring convenience.

However, this section of the spec is unclear as to whether an implementation must try to honor the color channels with decimals preserved or if the implementation can truncate/round the \<number\> to an integer. Edge has implemented this part of the spec, and appears to simply truncate the number. [Chromium is thinking about implementing it](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/zH51Atd-Vag), but wants to clarify this section first.

svgeesus:
Thanks for opening this, agree it needs to be clarified. The "for authoring convenience" is misleading and should be removed. This isn't just syntactic sugar.

Especially given the move to 10-bit and 12-bit per channel, the intent of the  the reason is to allow greater precision. Whether implementations `should` or `must` store the greater precision needs to be discussed. This also affects serialization.

tabatkins:
We make no guarantees whatsoever on numeric precision - UAs can store data at whatever precision they want to store it at.  They could do 4-bit color if they felt like it, it would just be ugly on anything more modern than a console.

annevk:
Rounding seems reasonable, but serializing an integer sometimes and sometimes something containing a dot seems like asking for trouble.

annevk:
The other problem here is that if two UAs both use say 8-bit it's unclear whether 254.5 means 254 or 255. I think if UAs use an equal amount of bits they need to have the same behavior.

Nadya678:
Is anyone able to see this difference?

svgeesus:
> Is anyone able to see this difference? 

As we move to wider gamuts and higher dynamic range then yes, banding due to rounding to 8 bits is visibly noticeable. For sRGB, not really.

tabatkins:
Yeah, specifying the required rounding behavior would probably be good.

annevk:
We should also have tests ensuring the serialization of CSS colors used on `<canvas>` are not affected. At least not until `<canvas>` supports more than 8 bits.

[...]

PR #2086, opened 2017-12-06
[css-color-4] Clarify required precision and rounding behavior for color channels

csnardi:
Helps to address #1983 as to the importance of honoring authored numbers, and the required rounding behavior if the precision as authored is not honored.


csnardi:
@svgeesus Since it was resolved that rounding was the proper behavior, could this be merged in?

Issue #2036, opened 2017-12-04
[selectors-4] Rename :focus-ring to :focus-indicator


alice:
https://drafts.csswg.org/selectors-4/#the-focusring-pseudo

- The current name causes confusion as authors tend to assume it refers to modifying the style of the default focus ring
- It is also misleading as it seems to imply that the author-provided focus indicator should be a ring, whereas in many cases author-provided focus styles are not

`:focus-indicator` does not carry this semantic baggage.

Furthermore, the phrase "focus indicator" is used in [WCAG guidelines](https://www.w3.org/TR/WCAG20-TECHS/G195.html), creating a semantic link between these two notions.

alice:
A couple of comments on the status quo regarding focus indicators in general:

- Chrome and Firefox and Edge all set focus on click. Safari does not. 
- The CSS spec has nothing at all to say about when a focus ring *should* be drawn.
  - This is WAI - browsers should be able to determine when they wish to show a focus indicator.
  - This selector will allow styling *that* case, as opposed to `:focus` which will (and should continue to) match whenever the element is focused.
- The current behaviour for a `<button>` element in Chrome and Edge is what we wish to normalise:
  - Clicking sets focus
  - `:focus` matches, but no focus ring is drawn
- In Chrome, this is effectively a hack: there is a user agent style for `:focus` which matches and should apparently be applied, but it is hard coded not to show on click.
  - For elements without this special exception, the UA style matches and a focus ring is shown.

alice:
http://output.jsbin.com/kijafe/1 demonstrates the difference between `<button>` default focus styling and `<div>` (or custom element) default focus styling.

bkardell:
Just wanted to add that Alice and I have been talking about adding this to the agenda since the week before TPAC, I 100% support this. Just to add briefly: The fact that ':focus-ring' isn't necessarily a ring, nor does it necessarily have a 100% standard 'thing' that it does (some people use border, some people use backgrounds, certain things indicate just fine without) has been confusing, in actual practice, for a number of authors and testers I've encountered.  In all cases this was resolved by pointing to existing documents explaining that it is really a focus _indicator_ we're interested in.

FremyCompany:
I have no strong opinion, but since there is a call to rename, I'll add my two cents to the discussion and propose `:focus-visible` (but do not intend to argue about it, it's just an idea for you to consider)

AmeliaBR:
[...]

PR #2104, opened 2017-12-12
[selectors-4] Rename :focus-ring to :focus-visible

alice:
Per discussion in #2036


frivoal:
Looks good to me.

Just one little non essential comment:
This has been up under the focus-ring name for a while, so there could be quite a few inbound links. Just in case, it could be nice to keep a (hidden) element with the old ids (`the-focusring-pseudo`, `focus-ring-pseudo`) near the new ones, to keep the links working.

alice:
Good idea, will fix.

alice:
Done.

alice:
I can't pull this but seems like it's ready to go?

Issue #2037, opened 2017-12-04
[selectors] :link and :visited are not mutually exclusive in implementations


victoriasu:
According to the [spec](https://drafts.csswg.org/selectors-4/#link), :link and :visited should be mutually exclusive, however in:
Chrome, Firefox, and Safari
- Style rules for :link and :visited are being applied to the same element (a visited link)
Edge
- All links are using the :link style and not the :visited style

Tested using:
https://jsfiddle.net/victoriaytsu/ujz85sor/

lilles:
This is for privacy reasons. See: https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector

Your spec link says:

"Since it is possible for style sheet authors to abuse the :link and :visited pseudo-classes to determine which sites a user has visited without the user’s consent, UAs may treat all links as unvisited links or implement other measures to preserve the user’s privacy while rendering visited and unvisited links differently."

So Edge is doing the former thing and Chrome/Firefox/Safari are doing the "other measures"?

emilio:
Yeah, querySelector and such never match `:visited` on Gecko. Similarly `getComputedStyle` doesn't return `:visited` styles and such.

dbaron:
For a single selector matching operation, I think they still are mutually exclusive, though.  At least based on [what I originally described](https://dbaron.org/mozilla/visited-privacy).

FremyCompany:
I think the difference in Edge is just an invalidation issue for ":link" because we are supposed to behave like other browsers here. https://jsfiddle.net/gck56f1h/ shows we still apply an outline but with wrong color once we visit the link, it doesn't make much sense to me.

FremyCompany:
Though actually loading the demo in other browsers, I am not sure I understand their behavior either. I think the problem is that you cannot remove all ":link" styles when the link is visited because otherwise this might leak somehow, so only a couple of properties will match using ":visited" and other properties will match using ":link" and this creates a weird state.

Maybe saying ":link" always matches would help solve the craziness but there might be a few websites broken by specificity of ":link" overriding their ":visited" styles.

FremyCompany:
Looking at the test case a bit more, here is what I understand the difference between Edge and other browsers to be:

Edge:
- Apply all :link rules and, for allowed properties, also apply all :visited rules (at the same time)

Other browsers:
- Apply all rules that do not have :visited (including :link rules)
- Apply all rules that have :visited or do have neither :link nor :visited

[...]

PR #2105, opened 2017-12-12
Rewrite :visited rules to match reality, based on dbaron's spec.

tabatkins:
Per #2037, adapts [dbaron's spec](https://dbaron.org/mozilla/visited-privacy) into more generic language, and adds several examples and some surrounding verbiage.


selectors-4/Overview.bs
```diff
@@ -1906,35 +1912,187 @@ The Hyperlink Pseudo-class: '':any-link''</h3>
 <h3 id="link">
 The Link History Pseudo-classes: '':link'' and '':visited''</h3>
 
-	User agents commonly display unvisited <a href="#the-any-link-pseudo">hyperlinks</a> differently from
-	previously visited ones. Selectors
-	provides the pseudo-classes <dfn id='link-pseudo'>:link</dfn> and
-	<dfn id='visited-pseudo'>:visited</dfn> to distinguish them:
+	User agents commonly display unvisited <a href="#the-any-link-pseudo">hyperlinks</a>
```
therealglazou: I would say "User agents commonly render unvisited hyperlinks" or "Visual user agents commonly display unvisited hyperlinks".

ionas:
Now instead of messing with the spec and creating a dirty implementation that works in weird ways (like explained in dbarons blog) why could the standards body not simply remove/deprecate `:visited` or at least move it into a different security profile (such as trusted or local single page applications only).

Then to enable the use case (notify users of a link that has already been visited) browser vendors could implement a tool-tip that is rendered ontop of the canvas not reachable by getComputedStyle, PointerEvents or anything else as it is a OS-tooltip, which would somehow notify the user about the state of visit.

Now this is not the most convenient for users, but it is the most secure.

Another way around this would be security profiles where :visited would always work on urls of the same domain, maybe even across subdomains of the same domain, but never onto foreign domains.

Foreign domains could then implement something like access control origin where they declare which domains would show the users visiting history, users could then accept those shared history suggestions/lists.

While all these are just bits of ideas, I am pretty sure at some point in future dbarons implementation will bite back because some browser API want's to read what's the pure render color below the mouse pointer or whatnot. `:visited` security constraints will be forgotten quickly or ignored by "greater needs".

tabatkins:
:visited already exists and is commonly used; dropping it entirely isn't realistic.

There've been discussions (off-list, I think) about figuring out ways to expose :visited styling "safely" in some contexts: allow it to work normally for same-origin links; or remember what links you *actually clicked* from a given domain, or visited *from* with a Referer header (and thus, what history could have been snooped on via JS or server code anyway), and let it work normally for those links, etc.

dbaron:
Regarding "In full, the behavior is actually that the '':link'' pseudo-class applies to all links, ":  I don't think "applies to all links" sounds right here.  It might be better to restructure this somehow to say that `::link` applies to all links that `::visited` doesn't apply to.

(But it seems like this PR is pretty messed up... the "Files changed" view for the PR as a whole doesn't have the first commit in it... and trying to comment on the commit with the actual changes fails with "Start commit oid is not part of the pull request".)

tabatkins:
Yeah, I have no clue what the PR is doing; I'll have to manually reconstruct it.

dbaron:
(I was going to continue to review further... but I haven't actually read further yet.)

alshamiri5 (approved):

astearns:
It looks like the substantive parts of this PR were merged in some other fashion

PR #2107, opened 2017-12-13
Added note for :link and :visited pseudo classes

### LINKED ISSUE #2037: [selectors] :link and :visited are not mutually exclusive in implementations

victoriasu:
UAs may display chosen colors for :visited and :link elements, but only
return the :link colors from getComputedStyle(). For non-color
properties, UAs may treat all links as unvisited.

fixes #2037

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -1936,6 +1936,10 @@ The Link History Pseudo-classes: '':link'' and '':visited''</h3>
 	or implement other measures to preserve the user's privacy
 	while rendering visited and unvisited links differently.
 
+	Note: UAs may display chosen colors for :visited and :link elements,
+	but only return the :link colors from getComputedStyle().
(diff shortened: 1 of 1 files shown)


victoriasu:
This might be redundant now that Tab has also proposed a change to the spec at https://github.com/w3c/csswg-drafts/pull/2105

svgeesus:
So, is this going to be merged as-is, merged with changes, or closed?

tabatkins:
Closing because defining this fully is a lot more complicated than a single-paragraph callout.

PR #2169, opened 2018-01-06
[css-align-3] Add left|right to *-position grammar

sideshowbarker:
This change adds the `left` and `right` keywords to the grammar terms
`<self-position>` and `<content-position>`.

That addition to the grammar brings it into sync with the existing spec prose,
which already includes `left` and `right` among the *positional alignment*
keywords collection, and states, “Two grammar terms are used to denote this
collection of values: `<self-position>` [and] `<content-position>`”.


tabatkins:
Yeah, I talked with @fantasai about this and she reminded me that we very explicitly did *not* want left/right in justify-content, nor the align-* properties; they're only intended for inline layout.

sideshowbarker:
> Yeah, I talked with @fantasai about this and she reminded me that we very explicitly did *not* want left/right in justify-content, nor the align-* properties; they're only intended for inline layout.

OK, but then problem remains that the spec as currently written doesn’t define `align-items: left` and `align-items: right` as valid.

Issue #913, opened 2017-01-13
[css-transforms-2] Explicitly specify that a transform is a containing block for all descendants


smfr:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=16328

Comment:  https://www.w3.org/Bugs/Public/show_bug.cgi?id=16328
Comment:  http://www.w3.org/mid/514E058C.5060105@exyr.org

Also
Comment: https://www.w3.org/mid/24BE3FE7-60C9-4690-A01C-F2291460DE7F@adobe.com
Comment:  https://www.w3.org/mid/20151125002414.GA22745@pescadero.dbaron.org

mikemcguire:
WOAH! moving issues to github, +1 for that and +1 for resolving this issue

tomasdev:
Are there any plans on fixing this?

http://jsbin.com/jazuwus/edit?html,css,output

asiljan:
+1

chrishtr:
Is this issue removing transform as a containing block for fixed-position elements? I don't think
that is a good idea, as it:
* makes clipping and filter effects hard to define and implement, and in general ill-defined because
it's impossible to "filter" across multiple 2d planes.
* introduces difficult and complex code to implement compositing
* makes it hard to provide clean fast paths for transforms

mikemcguire:
Im not going to pretend to understand the difficulties in implementing the client code cause I'm completely unqualified. But want to put my two cents in. When I ran into this issue I sat stumped forever. While this issue is now well documented on the web, it took me forever to actually figure out that the transform is what was causing the issue and thus being able to find the chrome bug filed and eventually the original Bugzilla spec bug to request an update. 

It seems counter-intuitive to not update the spec with what seems to be a universal expectation to accommodate the Chrome team. Especially since other browsers are accommodating what seems to be universally expected.

Additionally, while the spec for positioned elements outlines this issue, I find it difficult to understand the language used in the spec. Front-end devs could potentially waste a lot of time looking for a fix, and possibly explains a lack of interest for this issue. I personally find that the transforms spec more clearly outlines the behavior than the positioning spec.

RByers:
I believe #1945 tracks doing exactly that - updating the spec to match Chrome's behavior.  Is that right @chrishtr @smfr?  Should we merge these issues somehow to reflect the current thinking?

RByers:
[...]

PR #2185, opened 2018-01-13
[css-transform] Define "containing block for all descendants"

trchen1033:
Add a new definition "containing block for all descendants" and specify
that values other than 'none' on 'transform', 'transform-style',
'perspective' will cause an element to establish a containg block for
all descendants.

### DIFF
--- a/css-transforms-1/Overview.bs
+++ b/css-transforms-1/Overview.bs
@@ -316,9 +316,46 @@ For elements whose layout is governed by the CSS box model, the transform proper
 
 For elements whose layout is governed by the CSS box model, any value other than ''transform/none'' for the transform results in the creation of a stacking context. Implementations must paint the layer it creates, within its parent stacking context, at the same stacking order that would be used if it were a positioned element with ''z-index: 0''. If an element with a transform is positioned, the 'z-index' property applies as described in [[!CSS2]], except that ''z-index/auto'' is treated as ''0'' since a new stacking context is always created.
 
-For elements whose layout is governed by the CSS box model, any value other than ''transform/none'' for the transform also causes the element to become a containing block, and the object acts as a containing block for fixed positioned descendants.
+For elements whose layout is governed by the CSS box model, any value other than ''transform/none'' for the transform also causes the element to establish a <dfn export>containing block for all descendants</dfn>. Its padding box will be used to layout for all of its absolute-position descendants, fixed-position descendants, and descendant fixed background attachments.
 
-Issue(w3c/csswg-drafts#913): Is this effect on ''position: fixed'' necessary? If so, need to go into more detail here about why fixed positioned objects should do this, i.e., that it's much harder to implement otherwise.
+<div class="example">
+To demostrate the effect of [=containing block for all descendants=] on fixed-position descendants, the following code snippets should behave identically:
+<pre>
+&lt;style>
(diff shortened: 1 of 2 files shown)


dirkschulze:
Proposal to address https://github.com/w3c/csswg-drafts/issues/913

chrishtr:
Tab can you review/merge this?

css-transforms-1/Overview.bs
```diff
@@ -316,9 +316,46 @@ For elements whose layout is governed by the CSS box model, the transform proper
...
 
-For elements whose layout is governed by the CSS box model, any value other than ''transform/none'' for the transform also causes the element to become a containing block, and the object acts as a containing block for fixed positioned descendants.
+For elements whose layout is governed by the CSS box model, any value other than ''transform/none'' for the transform also causes the element's padding box to form a containing block for all of its out-of-flow descendants, including both absolute-position and fixed-position descendants, and fixed background attachments on descendants and itself.
 
-Issue(w3c/csswg-drafts#913): Is this effect on ''position: fixed'' necessary? If so, need to go into more detail here about why fixed positioned objects should do this, i.e., that it's much harder to implement otherwise.
+<div class="example">
+An element with 'transform' other than ''transform/none'' establishes <dfn export>containing block for all descendants</dfn>. To demostrate the effect on fixed-position descendants, the following code snippets should behave identically:
```
tabatkins: Can't define things in an example block, this needs to be moved to a normative section (probably just the preceding paragraph, where you're currently *referencing* the definition.

css-transforms-2/Overview.bs
```diff
@@ -62,7 +62,7 @@ Module Interactions {#module-interactions}
 
 The <a>3D transform functions</a> here extend the set of functions for the 'transform' property.
 
-Some values of 'perspective', 'transform-style' and 'backface-visibility' result in the creation of a <a href="https://www.w3.org/TR/CSS2/visuren.html#containing-block">containing block</a>, and/or the creation of a <a spec="css2">stacking context</a>.
+Some values of 'perspective', 'transform-style' and 'backface-visibility' result in the creation of a [=containing block for all descendants=], and/or the creation of a <a spec="css21">stacking context</a>.
```
tabatkins: This document doesnt' appear to define the term anywhere. (Unless it was already defined in some section that doesn't show up in this diff? Bikeshed doesn't seem to know the term, tho.) You can't rely on preceding versions of the document for a definition, it needs to be duplicated into here as well.

tabatkins:
r+ after requested changes

css-transforms-1/Overview.bs
trchen1033: Done.

css-transforms-2/Overview.bs
tabatkins: The problem is that the definition lives in a previous version of the same shortname. Bikeshed has automatic logic to ignore previous versions of a spec, because that's nearly always what you want to do.  In particular, if a later version omits a definition that the previous version had, this was probably for a good reason! You thus don't want autolinking to continue to point to the old version; you want it to fail so you can notice the breakage and update things.

You want to make sure that the latest version of a given spec has all the definitions that you expect to need; in this case that means adding a `<dfn>` somewhere in *this* document.

PR #2196, opened 2018-01-17
[css-color-4] consistently name color-mod()

### LINKED ISSUE #868: [css-color-4] Is color-mod() = color(…) a typo?
```
color-mod() = color( [ <color> | <hue> ] <color-adjuster>* )
```

Source: https://drafts.csswg.org/css-color/#modifying-colors

Why is `color-mod` on the left and `color` on the right? Shouldn’t the same identifier be on both sides?

---

Note to self: Notify https://cloudfour.com/thinks/building-themes-with-css4-color-features/

jonathantneal:
Uses the `color-mod()` function name consistently in order to avoid confusion with the `color()` function.

Resolves #868

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -2310,7 +2310,7 @@ Modifying Colors: the ''color-mod()'' function</h2>
 	Issue: Actually add the CNS thing.
 
 	<pre class='prod'>
-		<dfn>color-mod()</dfn> = color( [ <<color>> | <<hue>> ] <<color-adjuster>>* )
+		<dfn>color-mod()</dfn> = color-mod( [ <<color>> | <<hue>> ] <<color-adjuster>>* )
 		<dfn>&lt;color-adjuster></dfn> =
 		    [red( | green( | blue( | alpha( | a(] ['+' | '-']? [<<number>> | <<percentage>>] ) |
 		    [red( | green( | blue( | alpha( | a(] '*' <<percentage>> ) |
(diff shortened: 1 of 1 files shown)


praveenpuglia:
I see this being merged but..
here - https://www.w3.org/TR/css-color-4/#funcdef-color-mod

I still see inconsistencies. 
Example - 
**Example 11** in the spec uses the word `color` and not `color-mod` as function name. 

Will this be updated later or something?

AmeliaBR:
@praveenpuglia Merging a commit into the master branch automatically updates the editor's draft at https://drafts.csswg.org/css-color-4/

The "TR" link on w3.org is updated less frequently, after a formal process with the W3C.

praveenpuglia:
But the draft only has a link to `color-mod` function which redirects to the TR page. This seems confusing. Couldn't the latest version live in the editor's draft instead of pointing to the TR?

AmeliaBR:
Ah, it looks like the entire `color-mod` proposal was removed (while waiting for a major syntax re-writing) in https://github.com/w3c/csswg-drafts/commit/034b063697c3dadf144504f52e0858a79cd84414#diff-4dcdf0ecb00e997ce9e5b7d522a0cb62

But one reference remained by accident, and because it no longer had any definition in the same file to link to, it was automatically linked to the last TR definition.

Could you file a separate issue, and tag @TabAtkins in it?

praveenpuglia:
Sure! :)

PR #2197, opened 2018-01-17
[css-color-4] correct blend() adjuster calculation

### LINKED ISSUE #2195: [css-color] blend() percentages are inverted?
Are the instructions for blending colors inverted? 

According to the [`color-mod()` spec](https://drafts.csswg.org/css-color/#funcdef-color-mod):

```
color-mod() = color( [ <color> | <hue> ] <color-adjuster>* )
```

> The first argument specifies the **base color**... After the **base color**, zero or more `<color-adjuster>`s can be specified. Each `<color-adjuster>` modifies the color in some way, passing a new **base color** to the next `<color-adjuster>` in the list.

Then later, according to the [`blend()` spec](https://drafts.csswg.org/css-color/#blend-adjuster):

```
blend( <color> <percentage> [rgb | hsl | hwb]? )
```

> To determine the resulting color... linearly interpolate each of the channels of the colors according to the given `<percentage>`, where `0%` produces the specified `<color>` and `100%` produces the **base color**.

This seems backwards. For example:

```
color-mod(blue blend(red 100%))
```

In this example the **base color** is `blue` and is blended with `red` at `100%`, which produces ... `blue`?

jonathantneal:
Correct the `blend()` adjuster calculation so that a base color does not change when a specified color is 0% blended in.

This will consequently correct `tint()` and `shade()` adjusters, as they rely on the behavior of `blend()`.

Resolves #2195

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -2607,8 +2607,8 @@ Color Blending: the ''blend'' and ''blenda'' adjusters</h3>
 			interpret the |base color| and the given color in the appropriate color space
 			(RGB, HSL, or HWB).
 			Linearly interpolate each of the channels of the colors according to the given <<percentage>>,
-			where ''0%'' produces the specified <<color>>
-			and ''100%'' produces the |base color|.
+			where ''100%'' produces the specified <<color>>
+			and ''0%'' produces the |base color|.
 
 			If the color space is ''hsl'' or ''hwb'',
 			interpolate the hue channel either clockwise or counterclockwise,


svgeesus:
@tabatkins is it worth accepting corrections to the adjusters, given we want to rethink them from scratch?

jonathantneal:
I believe it will be helpful to have the correct information in your existing reference.

I do hope it is worth it. As an idea, tinting a color 0%, or mixing 0% black with a color should not turn that color 100% black.

tabatkins:
Yeah, might as well fix it while it's here.

svgeesus:
Agreed

PR #2264, opened 2018-02-02
[selectors-4] :nth-*-child(n of S) selectors should be Level 3/4

SelenIT:
It's true that `:nth-child(n)`/`:nth-last-child(n)` have been added in Level 3, but `of S` syntax is Level 4 addition. I believe they should be marked as Level 3/4, like the `:not()` selector (also introduced in Level 3, but extended in Level 4 with the ability to contain selector lists).


fantasai:
Thanks!

Issue #1559, opened 2017-06-23
[cssom]Inline style manipulation is somewhere between underspecified and incorrectly specified


bzbarsky:
A simple example.  Say we start at https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty  and walk through the steps.  The property value is parsed, the new value is set, etc.  This is all fine.

But there's a comefrom bit in https://drafts.csswg.org/cssom/#the-elementcssinlinestyle-interface that says:

> Mutating the declarations must set the style content attribute on the context object to the serialization of the declarations. If the style content attribute is set, changed or removed, the declarations must be updated as appropriate.

Where does that fit into the above-linked algorithm?  Does it run inside https://drafts.csswg.org/cssom/#set-a-css-declaration and if so does it run after step 3 _and_ after step 4, or just after step 4, or something else?  This is observable via mutation observers.

Or does it run at all?  I'm pretty sure that in browsers mutating the inline style via CSSOM does NOT do a serialize-and-reparse thing.  And this is black-box distinguishable (in browsers; not sure about per spec, because the processing model for this stuff is somewhat underdefined in various other ways) from doing the serialize-and-reparse, as far as I can tell, if base URIs are changed.

upsuper:
So, I think the bits to fix this are:
1. Define "attribute change steps" in either CSS Style Attributes or CSSOM (probably the former?) to accept change from setting attribute via DOM.
2. Change the set functions and delete function to run [change an attribute](https://dom.spec.whatwg.org/#concept-element-attributes-change) and [append an attribute](https://dom.spec.whatwg.org/#concept-element-attributes-append).

The hard part is that this create a slightly unfortunate reparsing in "attribute change steps", it's probably not too bad. Browsers should be able to optimize out that easily without observable difference. Actually this is already the case for `DOMTokenList`.

upsuper:
Actually... just "set an attribute value" algorithm.

bzbarsky:
> Browsers should be able to optimize out that easily without observable difference.

No, they can't, as I already pointed out in this issue in https://github.com/w3c/csswg-drafts/issues/1559#issue-238215652 -- serialize-and-reparse is black-box different from not serializing and reparsing wrt base URI handling.  And probably also wrt CSP.

PR #2268, opened 2018-02-03
[cssom-1] Record mutation for changes to CSSStyleDeclaration.

upsuper:
To ensure that we record the mutation to style attribute for MutationObserver.

Most of this should match what browsers currently do, at least it seems to match what Blink does. Gecko's behavior on `removeProperty` is a bit inconsistent in this case, and we want to fix it.


upsuper:
cc @bzbarsky

bzbarsky:
See #1559

zcorpan:
Why was this closed?

upsuper:
Because #2269 should be a better way to address the whole issue.

Issue #1898, opened 2017-10-23
[css-logical-1] [css-cascade-3] The all longhand probably shouldn't set logical properties.


emilio:
Or if it should, order of the expansion of `all` should be specified.

From https://drafts.csswg.org/css-cascade-3/#all-shorthand:

> The all property is a shorthand that resets all CSS properties except direction and unicode-bidi. It only accepts the CSS-wide keywords.

This includes logical properties, and this is a problem because that means that it produces unexpected results when used from CSSOM (where the position a property declaration appears on matters).

In particular, we came across https://bugzilla.mozilla.org/show_bug.cgi?id=1410028, which happens because of the internal order the properties are reset.

In particular, it's not the same if `all` expands to:

```
padding-inline-start: initial;
padding-left: initial;
```

that if it expands the other way around.

i.e., this two pieces of code should be equivalent:

```html
<!doctype html>
<script>
let setStyle = (el, props) => {
  for (prop in props)
    el.style.setProperty(prop, props[prop]);
};

let logical = document.createElement('div');
let physical = document.createElement('div');

setStyle(physical, {
  all: 'initial',
  'background-color': 'blue',
  display: 'block',
  width: '100px',
  height: '100px',
  padding: '10px'
});

setStyle(logical, {
  all: 'initial',
  'background-color': 'blue',
  display: 'block',
  width: '100px',
  height: '100px',
  'padding-inline-start': '10px',
  'padding-inline-end': '10px',
  'padding-block-start': '10px',
  'padding-block-end': '10px'
});

document.documentElement.appendChild(physical);
document.documentElement.appendChild(logical);
</script>
```

But they aren't in current browsers.

upsuper:
The problem of making `all` not expand to logical properties is that it in some case it would fail to reset all properties in some cases. For example, if someone has
```javascript
elem.style.paddingLeft = '10px';
elem.style.paddingInlineStart = '10px';
elem.style.all = 'initial';
```
people should be expecting that there is no padding anymore. However, if `all` doesn't reset logical properties, we would still have `padding-inline-start` left there.

I guess the solution should be having the order defined in the way that, logical properties go before physical ones. This doesn't fix the issue that the two pieces of code having different behavior, but it at least provides a behavior compatible with pre-logical world, and can consistently reset all properties with `all`.

emilio:
[...]

PR #2269, opened 2018-02-03
[cssom-1] Ensure inline style and corresponding declaration block ates each other properly

### LINKED ISSUE #1559: [cssom]Inline style manipulation is somewhere between underspecified and incorrectly specified

upsuper:
This mostly follows how [`DOMTokenList`](https://dom.spec.whatwg.org/#interface-domtokenlist) works in the DOM spec, which has a similar situation that having a live object reflecting an attribute.

This should fix #1559.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1952,6 +1952,10 @@ the DOM a <a>CSS declaration block</a> is a
  <dt><dfn>owner node</dfn>
  <dd>The {{Element}} that the <a>CSS declaration block</a> is associated with, if any, or
  null otherwise.
+
+ <dt><dfn>updating flag</dfn>
+ <dd>Unset by default. Set when the <a>CSS declaration block</a> is updating
+ the <a for="CSSStyleDeclaration">owner node</a>'s <code>style</code> attribute.
 </dl>
 
 To <dfn>parse a CSS declaration block</dfn> from a string <var>string</var>, follow these steps:
@@ -2038,6 +2042,7 @@ A <a>CSS declaration block</a> has these <a>attribute change steps</a> for its <
 
 <ol>
  <li>If the <a for="CSSStyleDeclaration">readonly flag</a> is set, terminate these steps.
+ <li>If the <a for="CSSStyleDeclaration">updating flag</a> is set, terminate these steps.
  <li>If <var>localName</var> is not "<code>style</code>", or <var>namespace</var> is not null, terminate these steps.
  <li>If <var>value</var> is null, empty the <a for="CSSStyleDeclaration">declarations</a>.
  <li>Otherwise, let the <a for="CSSStyleDeclaration">declarations</a> be the result of <a>parse a CSS declaration block</a>
@@ -2053,6 +2058,18 @@ When a <a>CSS declaration block</a> object is created, then:
  given null, "<code>style</code>", and <var>owner node</var>.
  <li>If <var>value</var> is not null, let the <a for="CSSStyleDeclaration">declarations</a> be the result of
  <a>parse a CSS declaration block</a> from a string <var>value</var>.
+</ol>
+
+To <dfn>update style attribute for</dfn> <var>declaration block</var> means to run the steps below:
+
+<ol>
+ <li>Assert: <var>declaration block</var>'s <a for="CSSStyleDeclaration">readonly flag</a> is unset.
+ <li>Let <var>owner node</var> be <var>declaration block</var>'s <a for="CSSStyleDeclaration">owner node</a>.
+ <li>If <var>owner node</var> is null, terminate these steps.
+ <li>Set <var>declaration block</var>'s <a for="CSSStyleDeclaration">updating flag</a>.
+ <li><a>Set an attribute value</a> for <var>owner node</var> using "<code>style</code>" and the result of
+ <a lt="serialize a CSS declaration block">serializing</a> <var>declaration block</var>.
+ <li>Unset <var>declaration block</var>'s <a for="CSSStyleDeclaration">updating flag</a>.
 </ol>
 
 The <dfn id=concept-shorthands-preferred-order>preferred order</dfn> of a list of shorthand properties <var>shorthands</var> is as follows:
@@ -2100,8 +2117,7 @@ Setting the {{CSSStyleDeclaration/cssText}} attribute must run these steps:
  <li>Empty the <a for="CSSStyleDeclaration">declarations</a>.
  <li><a lt="Parse a CSS declaration block">Parse</a> the given value and, if the return value is not the empty list, insert the items in the list
  into the <a for="CSSStyleDeclaration">declarations</a>, in <a>specified order</a>.
- <li>If the <a for="CSSStyleDeclaration">owner node</a> is not null, <a>set an attribute value</a> for <a for="CSSStyleDeclaration">owner node</a>
- using "<code>style</code>" and the result of <a lt="serialize a CSS declaration block">serializing</a> the <a for="CSSStyleDeclaration">declarations</a>.
+ <li><a>Update style attribute for</a> the <a>CSS declaration block</a>.
 </ol>
 
 The <dfn attribute for=CSSStyleDeclaration>length</dfn> attribute must return the number of <a lt="CSS declaration">CSS
@@ -2187,8 +2203,7 @@ The <dfn method for=CSSStyleDeclaration>setProperty(<var>property</var>, <var>va
  <li>Otherwise, <a lt="set a CSS declaration">set the CSS declaration</a> <var>property</var> with value <var>component value list</var>, with
  the <i>important</i> flag set if <var>priority</var> is not the empty string, and unset otherwise, and with the list of declarations being the
  <a for="CSSStyleDeclaration">declarations</a>.
- <li>If the <a for="CSSStyleDeclaration">owner node</a> is not null, <a>set an attribute value</a> for <a for="CSSStyleDeclaration">owner node</a>
- using "<code>style</code>" and the result of <a lt="serialize a CSS declaration block">serializing</a> the <a for="CSSStyleDeclaration">declarations</a>.
+ <li><a>Update style attribute for</a> the <a>CSS declaration block</a>.
 </ol>
 
 To <dfn>set a CSS declaration</dfn> <var>property</var> with a value <var>component value list</var> and optionally with an <i>important</i> flag set, in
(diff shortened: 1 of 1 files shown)


upsuper:
cc @bzbarsky

upsuper:
Note that this cannot bikeshed correctly at this moment, because "attribute change steps" wasn't actually exported from the DOM spec until whatwg/dom#567 which was merged several hours ago.

upsuper:
WebKit and Blink don't agree with the part of `setProperty`. They don't update the attribute when the setting doesn't change value. It seems tricky to me in the sense that I don't think we have a formal definition anywhere about how to compare two values of a property, do we?

Also, if we follow the resolution in #1898 to make `setProperty` append new value rather than mutate the value in-place, the behavior of WebKit and Blink may become more trickier to be right. Given these, I tend to keep the that part as-is.

upsuper:
Also wpt test for this change: w3c/web-platform-tests#9376.

cssom-1/Overview.bs
```diff
@@ -2034,6 +2034,25 @@ Note: The serialization of an empty CSS declaration block is the empty string.
...
+ <li>Assert: the <a for="CSSStyleDeclaration">readonly flag</a> is unset.
+ <li>If <var>localName</var> is not "<code>style</code>", or <var>namespace</var> is not null, terminate these steps.
+ <li>If <var>value</var> is null, empty the <a for="CSSStyleDeclaration">declarations</a>.
+ <li>Otherwise, let the <a for="CSSStyleDeclaration">declarations</a> be the result of <a>parse a CSS declaration block</a>
+ from a string <var>value</var>.
+</ol>
+
+When a <a>CSS declaration block</a> object is created, then:
+<ol>
+ <li>Let <var>owner node</var> be the <a for="CSSStyleDeclaration">owner node</a>.
```
zcorpan: Need to also check if the readonly flag is set. Now this will try to get an attribute when doing `getComputedStyle(elm)` I think.

upsuper: Oh, you're right. The algorithm for `attribute change steps` also need updates.

upsuper: Updated to make this algorithm and `attribute change steps` check `readonly flag`.

zcorpan:
> They don't update the attribute when the setting doesn't change value. It seems tricky to me in the sense that I don't think we have a formal definition anywhere about how to compare two values of a property, do we?

We don't I think. But maybe figuring this out can be part of #1898 ?

upsuper:
> We don't I think. But maybe figuring this out can be part of #1898 ?

Yeah, sure, I guess `setProperty` part definitely needs further tweaking for that issue.

cssom-1/Overview.bs
```diff
@@ -2034,6 +2034,25 @@ Note: The serialization of an empty CSS declaration block is the empty string.
 Note: The serialization of a non-empty CSS declaration block does not include any surrounding whitespace, i.e., no whitespace appears
 before the first property name and no whitespace appears after the final semicolon delimiter that follows the last property value.
 
+A <a>CSS declaration block</a> has these <a>attribute change steps</a> for its <a for="CSSStyleDeclaration">owner node</a>:
+<ol>
+ <li>If the <a for="CSSStyleDeclaration">readonly flag</a> is set, terminate these steps.
```
bzbarsky: How can this even happen?  The cases when there is an owner node are currently not readonly and there are no plans I am aware of to make them readonly...

There should probably be a note here explaining what this test is about, because it's pretty confusing.

cssom-1/Overview.bs
```diff
@@ -2034,6 +2034,25 @@ Note: The serialization of an empty CSS declaration block is the empty string.
...
+ from a string <var>value</var>.
+</ol>
+
+When a <a>CSS declaration block</a> object is created, then:
+<ol>
+ <li>Let <var>owner node</var> be the <a for="CSSStyleDeclaration">owner node</a>.
+ <li>If <var>owner node</var> is null, or the <a for="CSSStyleDeclaration">readonly flag</a> is set, terminate these steps.
+ <li>Let <var>value</var> be the result of <a lt="get an attribute by namespace and local name">getting an attribute</a>
+ given null, "<code>style</code>", and <var>owner node</var>.
+ <li>If <var>value</var> is not null, let the <a for="CSSStyleDeclaration">declarations</a> be the result of
```
bzbarsky: s/let the/let/ ?

cssom-1/Overview.bs
```diff
@@ -2079,6 +2098,8 @@ Setting the {{CSSStyleDeclaration/cssText}} attribute must run these steps:
  <li>Empty the <a for="CSSStyleDeclaration">declarations</a>.
  <li><a lt="Parse a CSS declaration block">Parse</a> the given value and, if the return value is not the empty list, insert the items in the list
  into the <a for="CSSStyleDeclaration">declarations</a>, in <a>specified order</a>.
+ <li>If the <a for="CSSStyleDeclaration">owner node</a> is not null, <a>set an attribute value</a> for <a for="CSSStyleDeclaration">owner node</a>
```
bzbarsky: I really think we should prevent this attribute set from leading to a re-parse.  It's a much simpler model than having it re-parse, because in the presence of CSP parsing from the string does NOT give the same results as the thing that was serialized to the string, in general.  And since UAs are going to optimize out the reparsing for basic sanity reasons, this is likely to lead to divergence between what the spec theoretically says (but no one would ever do) and what UAs do.

There are also issues around base URIs: set vs serialize-and-reparse treat them differently.  I did explicitly mention this in https://github.com/w3c/csswg-drafts/issues/1559 as a problem with the serialize-and-reparse approach.

Not reviewing the rest of the changes for now given that, because they will need to be modified anyway.

cssom-1/Overview.bs
zcorpan: getComputedStyle sets owner node per the spec. But maybe that is a spec bug?

https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle

bzbarsky: Oh, I missed that somehow.  OK, then this is probably fine.  That said, this is the first actual _use_ of owner node that I can see.  So if it simplifies things, we could probably avoid setting owner node in `getComputedStyle` instead of dealing with it here....

zcorpan: Yeah I think we can set it to null there.

upsuper: We can discuss this in a separate PR. I don't think that needs to block this change.

Actually, I have a feeling that having the declaration block from `getComputedStyle` have the owner node set makes sense. Not sure how useful it is, though.

cssom-1/Overview.bs
upsuper: It seems to me that there is a convention that we use "the" before associated properties, and no "the" before variables, so I add "the" here. Not sure if my understanding is correct.

cssom-1/Overview.bs
upsuper: Do you have any suggestion for how can we avoid serialize-and-reparse? It seems to me if we want to avoid that, we would need to duplicate many definitions from the DOM spec just for getting rid of the "run attribute change steps" part.

Is there any existing algorithm in the DOM spec helps doing that? If not, I suggest we merge this change to get the spec closer to what browsers do currently, and file a new issue to address the serialize-and-reparse issue, which may need some help from the DOM spec side.

bzbarsky: You could set a flag while _you_ are setting the attribute, and not do any attribute change steps while that flag is set, right?

upsuper: Hmmm, okay, fair. Thanks for the suggestion.

upsuper:
@zcorpan @bzbarsky ping. Are you two fine with the updated version? (The second commit is the change on top of what you have reviewed.)

cssom-1/Overview.bs
zcorpan: Something like that, yeah, though probably not entirely consistent. But we should probably adopt Infra Standard conventions when we update things or write new material for CSSOM (like saying "then return" instead of "terminate these steps").

upsuper:
Thanks for the review and the new test :)

zcorpan:
Browser bugs
Safari https://bugs.webkit.org/show_bug.cgi?id=182687
Edge https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/15908562/
Firefox - @upsuper ?
Chrome - passes the tests

upsuper:
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1197705

I removed the test of setting identical value triggering mutation record. That test would fail Chrome as well, but anyway... We can add that later after we address #1898.

cssom-1/Overview.bs
```diff
@@ -2034,6 +2038,40 @@ Note: The serialization of an empty CSS declaration block is the empty string.
...
+<ol>
+ <li>If the <a for="CSSStyleDeclaration">readonly flag</a> is set, terminate these steps.
+ <li>If the <a for="CSSStyleDeclaration">updating flag</a> is set, terminate these steps.
+ <li>If <var>localName</var> is not "<code>style</code>", or <var>namespace</var> is not null, terminate these steps.
+ <li>If <var>value</var> is null, empty the <a for="CSSStyleDeclaration">declarations</a>.
+ <li>Otherwise, let the <a for="CSSStyleDeclaration">declarations</a> be the result of <a>parse a CSS declaration block</a>
+ from a string <var>value</var>.
+</ol>
+
+When a <a>CSS declaration block</a> object is created, then:
```
bzbarsky: @upsuper Why is this timing correct? Especially since nothing specifies when these objects are actually created, last I checked...

cssom-1/Overview.bs
```diff
@@ -2034,6 +2038,40 @@ Note: The serialization of an empty CSS declaration block is the empty string.
...
+</ol>
+
+To <dfn>update style attribute for</dfn> <var>declaration block</var> means to run the steps below:
+
+<ol>
+ <li>Assert: <var>declaration block</var>'s <a for="CSSStyleDeclaration">readonly flag</a> is unset.
+ <li>Let <var>owner node</var> be <var>declaration block</var>'s <a for="CSSStyleDeclaration">owner node</a>.
+ <li>If <var>owner node</var> is null, terminate these steps.
+ <li>Set <var>declaration block</var>'s <a for="CSSStyleDeclaration">updating flag</a>.
+ <li><a>Set an attribute value</a> for <var>owner node</var> using "<code>style</code>" and the result of
```
bzbarsky: This doesn't handle mutation events right, but I guess neither does anything else....

cssom-1/Overview.bs
upsuper: As far as it is returned from `Element.style`, does it matter when the objects are actually created? Browser needs to create it at some point, but before script gets it somehow, whether it exists doesn't really matter, does it?

bzbarsky: It matters because the base URI is different at different points in time, depending on what the node document of the element is.  So depending on exactly when the creation/parsing happen you will get different behavior for relative URIs in property values....

Maybe that doesn't affect what CSSOM sees, just the actual styling behavior.

upsuper: Hmm... okay. Sounds like something worth raising for css-style-attr I guess... The interaction between the two specs would be fun...

PR #2277, opened 2018-02-06
Append / to ED link

foolip:
https://drafts.csswg.org/selectors redirects.


foolip:
Noticed because a [crappy script](https://github.com/foolip/day-to-day/blob/master/build/specs.js) I wrote assumes trailing slashes in drafts.csswg.org URLs to avoid capturing it in a regex. I'll fix the script, but might as well use the canonical URL here too.

PR #2299, opened 2018-02-11
[cssom-1] Remove or replace "terminate ..."

upsuper:
This change removes all the "terminate these steps" and "terminate this algorithm" from this spec. They are either removed due to redundancy (already following a "return" or "throw") or replaced with "return".


upsuper:
Following @zcorpan's comment at [#2269 (comment)](https://github.com/w3c/csswg-drafts/pull/2269#discussion_r167449109).

Issue #1872, opened 2017-10-15
[css-contain] Clarify style containment effect 1


Loirooriol:
I don't understand what the first effect of [style containment](https://drafts.csswg.org/css-contain/#containment-style) is supposed to do.

> The following properties must have no effect on descendants of the element:
> - `break-*` (and the related aliases)
> - `bookmark-*`
> - `string-set`

For starters, the definitions of `bookmark-*` and `string-set` don't make sense (see #1830 and #1829). In fact I think the whole CSS Content spec should be marked as "Not ready for implementation" or something until it is rewritten. So I don't think these properties should be mentioned in CSS Contain, which is a CR.

So let's focus on `break-*`. I don't understand the effect. Does it mean that `break-*` properties in selectors that match descendants do not have any effect? Or that `break-*` property values on the contained element don't affect descendants? For example, from [unforced breaks](https://drafts.csswg.org/css-break-3/#unforced-breaks),

> if all of them are `auto` and a common ancestor of all the elements has a `break-inside` value of `avoid`, then breaking here is not allowed.

If this ancestor is the contained element, does it mean that `break-inside: avoid` has no effect, is treated as `auto`, and thus a break between its descendants is allowed?

If so please clarify the sentence to something like

> The following properties, when set on the contained element or one of its ancestors, must have no effect on descendants of the element

But then I'm not much sure about how this is different from [scoped properties](https://drafts.csswg.org/css-contain/#scoped-property) described in effect 2.

frivoal:
I've made the editorial changes based on the previous discussion (move some effects of style containement from css-contain to css-content, clarify that `break-*` means `break-after` and `break-before`, not `break-inside`), but we still need to resolve whether or not style containment affects `break-before` and `break-after`.

Based on the discussion minitued above, I think a good case has been made that it should not, and that we should remove that from the spec, but @tabatkins seemed to think that maybe it should.

Agenda+ to sort this out. @tabatkins if you have not changed your mind, please have some arguments ready.

Loirooriol:
Fixed in 763c0f1fc2d4e23a9f71f748e0371bd5ef244f48

PR #2310, opened 2018-02-15
Move effects of containment on regions from [css-contain-1] to [css-regions-1]

frivoal:
As discussed in https://github.com/w3c/csswg-drafts/issues/1872#issuecomment-365686723, this PR moves the effect of containment on regions from the containment spec to the regions spec.

@astearns I vouch for the css-contain part of this patch. Can you review (and hopefully approve, then merge) the css-regions part?

### DIFF
--- a/css-regions-1/Overview.bs
+++ b/css-regions-1/Overview.bs
@@ -280,7 +280,8 @@ Last region</h3>
 	A <a>CSS region</a> is deemed to be the <dfn lt="last usable region | last usable CSS region">last usable region</dfn>
 	in a <a>region chain</a>
 	if it is the first region in that chain to have <a>layout containment</a>,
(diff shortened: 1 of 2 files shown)


css-regions-1/Overview.bs
```diff
@@ -274,6 +274,14 @@ Region chain</h3>
 	into a <a>region chain</a>
 	according to the document order.
 
+<h3 id=last-region>
+Last region</h3>
+
+	A <a>CSS region</a> is deemed to be the <dfn lt="last usable region | last usable CSS region">last usable region</dfn>
+	in a <a>region chain</a>
+	if it is the first region in that chain to have <a>layout containment</a>,
```
astearns: Could you add a "(See [[!CSS-CONTAIN-1]])" here?

PR #2335, opened 2018-02-20
Add missing comma between "multi-column" & "flex"

dholbert:
Before this fix, the spec text says "...provide this functionality for multi-column flex, and grid layout" (with links to these three distinct layout specs).  Clearly there's a comma missing between "multi-column" and "flex".

Link to this spec section, for reference: https://drafts.csswg.org/css-align/#gaps


tabatkins:
The *one* downside of semantic linebreaks is that sometimes you just *assume* you've written a comma at the end of a line. ^_^

frivoal:
@tabatkins 
We should adopt semantic breaks
Let us in verse compose our specs
Just style it all with white-space: pre
From punctuation we break free

PR #2338, opened 2018-02-20
Explicitly specify what gets set to what with MCW

Manishearth:
"A is B" can mean both "A = B" and "B = A". It usually means the former; but in this case it's the latter, and that's confusing.

### DIFF
--- a/css2/tables.html
+++ b/css2/tables.html
@@ -996,7 +996,7 @@ final layout and may demand more than one pass.
   formatted content may span any number of lines but may not overflow
   the cell box. If the specified <a href="visudet.html#propdef-width" class="noxref"><span
   class="propinst-width">'width'</span></a> (W) of the cell is greater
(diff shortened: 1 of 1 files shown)


css2/tables.html
```diff
@@ -996,7 +996,7 @@ <h4>17.5.2.2 <a name="auto-table-layout">Automatic table layout</a></h4>
   formatted content may span any number of lines but may not overflow
   the cell box. If the specified <a href="visudet.html#propdef-width" class="noxref"><span
   class="propinst-width">'width'</span></a> (W) of the cell is greater
-  than MCW, W is the minimum cell width. A value of 'auto' means that
+  than MCW, the minimum cell widthis set to W. A value of 'auto' means that
```
tabatkins: "widthis" typo

Manishearth: gah, used the github editor, and it javascripted on me whilst typing. Will fix.

Manishearth:
updated

PR #2377, opened 2018-03-02
Use `<custom-ident>` consistently for counter names

nox:
The functions `counter()` and `counters()` were badly specified, using
`<ident>` for the counter names, even though `counter-increment` and friends
will only accept `<custom-ident>` for those names.


nox:
There is a test for this in https://github.com/servo/servo/pull/20171.

PR #2391, opened 2018-03-05
Scroll effect

Jul10l1r4:
Based: [MDN - Scroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) :heart:

### DIFF
--- a/default.css
+++ b/default.css
@@ -14,6 +14,7 @@
 		position: relative;
 		clear: both;
 		color: #600;
+		scroll-behavior: smooth;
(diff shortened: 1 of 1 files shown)


tabatkins:
Looks reasonable to me. @fantasai, you have any objection?

fantasai:
Ummm, where do we ever use programmatic scrolling inside an example code block? :) Because afaict from the spec that's the only place this would apply...

PR #2407, opened 2018-03-06
Add :focus-visible recommendation for tabindex/appearance.

robdodson:
Although the `:focus-visible` spec says it's up to the UA's heuristic to determine when to match, I wanted to add a note to recommend that `:focus-visible` not match on mouse-focused elements with `tabindex=0` or modified `appearance`. What do y'all think?

cc @alice @bkardell @tabatkins

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -2295,6 +2295,30 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
 			apply '':focus-visible''.
 	</div>
 
+	<div class="note">
+		Note: It is a common practice by UAs to always draw a focus indicator on any
+		element with a <pre>tabindex</pre> attribute, or any element whose
+		''appearance'' has been altered. However, it is recommended that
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -2295,6 +2295,30 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
 			apply '':focus-visible''.
 	</div>
 
+	<div class="note">
+		Note: It is a common practice by UAs to always draw a focus indicator on any
```
alice: I would replace this with a note about UA stylesheets and `:focus` as the usual mechanism for showing a focus ring, except in a few cases concerning elements which have their default appearance and exhibit something like the behaviour `:focus-visible` should emulate.

Then, follow up with a comment about the eventual goal being to recommend UAs use `:focus-visible` in place of `:focus` so that the default behaviour will be more nuanced.

robdodson:
I'm going to close this PR while we wordsmith this a bit more.

PR #2415, opened 2018-03-07
[css-text-4] Update several links

xfq:


css-text-4/Overview.bs
```diff
@@ -335,7 +335,7 @@ Inline breaks between boxes: the 'wrap-before'/'wrap-after' properties</h3>
 		<dd>
 			Force a <a>flex line</a> break immediately before/after the box
 			if the box is a <a>flex item</a>
-			in a <a href="https://www.w3.org/TR/css-flexbox-1/#multi_line">multi-line flex container</a>.
+			in a <a>multi-line flex container</a>.
```
ewilligers: Should be href="https://www.w3.org/TR/css-flexbox-1/#multi-line-flex-container"

tabatkins: Yeah, I've run into this bug before. I'll merge this, and figure out the bug separately.

xfq:
Thanks @tabatkins!

Issue #2068, opened 2017-12-05
[web-animations-1] Drop the *ReadOnly interfaces


birtles:
_From @birtles on April 7, 2017 6:27_

Originally we introduced the *ReadOnly interfaces for CSS. The logic went something like this:

* It's pretty much impossible to map changes made via API back to style (we tried, and there were lots of cases where it's just not clear what the expected behavior should be, particularly when the list length changes.)
* So we decided to make it a one-way mapping, i.e. style updates the API objects but changing the objects doesn't update style.
* As a result you need to know when an object gets a particular timing property from style and when it gets it from the API.
* We tried doing this on a property-by-property basis but it leads to a very confusing API. How do you tell which properties you've overridden? How do you reset them so that they start reflecting style again? Do we add three methods per property?
* Instead we opted for an API where you very deliberately have to break from style as follows:

```js
const animation = div.getAnimations()[0];
animation.effect = new KeyframeEffect(animation.effect); // Make a mutable copy
// Now you can tweak it and its obvious that changes to style do not apply
effect.timing.duration = 3000; // This works
div.style.animationDuration = '4s'; // This doesn't
```

However, in #168 I'm proposing we simplify the timing interfaces to have only:

```js
animation.effect.getTiming();
animation.effect.setTiming(); // Resets all properties to their defaults
animation.effect.setTimingProperty({ duration: 3000 }; // Overrides a single property
```

The last one is basically equivalent to:

```js
effect.setTiming(Object.assign(effect.getTiming(), { duration: 3000 }));
```

I'm now wondering if there's some way we can exploit this setup (and similarly `setKeyframes()`) to signal which properties you want to override. Perhaps we could change the timing dictionary to have default values which we interpret as the initial value for script-generated animations and the style-specified value for CSS animations/transitions such that `setTiming()` resets everything?

That would certainly be nicer for authors: being able to override the delay of a single instance of a CSS animation programmatically is certainly something I've wanted to do, and it would be nice not to have to completely divorce it from style changes in order to do so. It would also avoid some of the confusion and complexity around the *ReadOnly interfaces so it's probably worth considering at least.

_Copied from original issue: w3c/web-animations#185_

birtles:
_From @alancutter on April 7, 2017 6:58_

What happens when you do
```javascript
const animation = div.getAnimations()[0];
animation.effect.setTimingProperty({ duration: 3000 });
div.style.animationDuration = '4s';
```
under this new scheme?

[...]

PR #2432, opened 2018-03-12
[web-animations-1] Replace *ReadOnly interfaces and timing objects

### LINKED ISSUE #2055: [web-animations-1] Add a constructor for AnimationEffectTiming and AnimationEffectTimingReadOnly
_From @birtles on September 30, 2015 4:57_

https://w3c.github.io/web-animations/#issue-f6fa9df9
https://w3c.github.io/web-animations/#issue-f6fa9df90

We should then use that constructor from the KeyframeEffectReadOnly constructor:
https://w3c.github.io/web-animations/#issue-c84185b9

_Copied from original issue: w3c/web-animations#121_

### LINKED ISSUE #2065: [web-animations-1] Replace AnimationEffectTiming(ReadOnly) with setTiming() ?
_From @birtles on September 14, 2016 7:47_

Just a thought about how we could simplify the API and make it more consistent.

Currently, we manage timing properties using the following interfaces:
- [`AnimationEffectTimingProperties`](https://w3c.github.io/web-animations/#dictdef-animationeffecttimingproperties) &mdash; property bag for passing to constructors etc.
- [`ComputedTimingProperties`](https://w3c.github.io/web-animations/#the-computedtimingproperties-dictionary) &mdash; extends `AnimationEffectTimingProperties` with some extra properties we compute
- [`AnimationEffectTimingReadOnly`](https://w3c.github.io/web-animations/#the-animationeffecttimingreadonly-interface) &mdash; Live object with getters for reading the specified timing properties. Fortunately, this happens to have the same members as `AnimationEffectTimingProperties` so you can pass it to constructors/methods that expect a `AnimationEffectTimingProperties` and it should work.
- [`AnimationEffectTiming`](https://w3c.github.io/web-animations/#the-animationeffecttiming-interface) &mdash; Subclass of `AnimationEffectTimingReadOnly` that adds getters for the different parameters (used by mutable effects, i.e. those not tied to CSS)

In practice it looks like this:

``` js
// Set up a new animation
const anim = elem.animate({...}, { duration: 2000, fill: 'auto' });

// Read back the specified values
anim.effect.timing.duration; // 2000
anim.effect.timing.fill; // 'auto'

// Read back computed values
anim.effect.getComputedTiming().duration; // 2000
anim.effect.getComputedTiming().fill; // 'none'
// NOTE: |fill| and |duration| are a little bit special in that they allow 'auto'
// values that get computed into something else. It's not particularly useful, though,
// until we have groups (where, e.g., a group's auto duration computes to the length
// of its children).

// Read other computed values
anim.effect.getComputedTiming().activeDuration // 2000
anim.effect.getComputedTiming().progress // 0.0

// Update specified values
anim.effect.timing.duration *= 2;
anim.effect.timing.iterations++;
```

The parallel here is with style:
- `anim.effect.timing` =~ `elem.style` (sort of, kind of)
- `anim.effect.getComputedTiming` =~ `window.getComputedStyle(elem)`

_However_, for keyframes we do something completely different. We don't have live keyframe objects since that would be hard and instead we just use [`getKeyframes()`](https://w3c.github.io/web-animations/#dom-keyframeeffectreadonly-getkeyframes) and [`setKeyframes()`](https://w3c.github.io/web-animations/#dom-keyframeeffect-setkeyframes).

In addition, keyframe offsets can be computed. The author can omit them and we'll fill them in. We _don't_ just fill in the `offset` member with the computed value when we return keyframes using `getKeyframes()`, however, since if we did that you couldn't take the result of `getKeyframes()` and pass it to `setKeyframes()` on another effect without freezing the offsets at that point (and losing the ability to auto-distribute keyframes from then on). Instead, we return the computed offset as a separate member: `computedOffset`.

So my thought here was:
- Drop `AnimationEffectTimingReadOnly`.
- Drop `AnimationEffectTiming`.
- Drop `effect.timing`.
- Rename `effect.getComputedTiming()` to `effect.getTiming()`.
- Extend `ComputedTimingProperties` to include `computedDuration` and `computedFill`.
- Add `effect.setTiming()` that takes an `AnimationEffectTimingProperties` (which `ComputedTimingProperties` extends so you could pass the result of `getTiming()` to `setTiming()`).

Advantages:
- 2 fewer interfaces
- Fewer live objects
- Consistency within the API &mdash; keyframes and timing properties follow the same idiom
- Bulk setting of timing properties is easier

Disadvantages:
- More awkward to update an individual timing property, e.g.

``` js
// Current API
anim.effect.timing.duration *= 2;
// Proposed API
const updatedTiming = anim.effect.getTiming();
updatedTiming.duration *= 2;
anim.effect.setTiming(updatedTiming);
```
- More inefficient to update individual timing properties since you need to read and write these property bags. From what I understand, at least for Gecko, that can be quite slow.

I do wonder how common those last two items are, though. (And for the first one, if you're just blindingly overwriting the value, you can use Object.assign() or, one day, the object spread operator to write it more compactly.)

The other consideration is this proposal _may_ be less forwards compatible. For example, suppose we extend `iterations` to take a keyword value. Presumably it will still compute to a float. Code that currently calls `anim.effect.getComputedTiming().iterations` would continue to work when this new property is used. However, under this proposal, code that calls `anim.effect.getTiming().iterations` would need to be updated to call `anim.effect.getTiming.computedIterations` instead.

That said, I think we already have this problem in that anyone currently calling `anim.effect.timing.iterations` and assuming it will be a float (which I expect is not uncommon since it's less to type than `anim.effect.getComputedTiming().iterations`) will have the same problem.

@shans, @dstockwell Any thoughts about this?

@heycam, @bzbarsky Is returning snapshot objects generally the preferred approach over maintaining live objects?

_Copied from original issue: w3c/web-animations#168_

birtles:
This fixes #2065 and #2068. It also fixes #2055.

Summary of changes:
* Replaced `AnimationEffectReadOnly` with `AnimationEffect`.
* Replaced the timing member of `AnimationEffectReadOnly` with the `getTiming()` and `updateTiming()` methods on `AnimationEffect`.
* Removed the `AnimationEffectTimingReadOnly` and `AnimationEffectTiming` interfaces.
* Renamed the `AnimationEffectTimingProperties` dictionary to `EffectTiming`.
* Renamed the `ComputedTimingProperties` dictionary to `ComputedEffectTiming`.
* Introduced the `OptionalEffectTiming` dictionary for use with the `updateTiming()` method.
* Removed `KeyframeEffectReadOnly`, leaving only `KeyframeEffect`.

The relevant parts of the new IDL look something like:

```webidl
[Exposed=Window]
interface AnimationEffect {
    EffectTiming         getTiming();
    ComputedEffectTiming getComputedTiming();
    void                 updateTiming(optional OptionalEffectTiming timing);
};

dictionary EffectTiming {
    double                             delay = 0;
    double                             endDelay = 0;
    FillMode                           fill = "auto";
    double                             iterationStart = 0.0;
    unrestricted double                iterations = 1.0;
    (unrestricted double or DOMString) duration = "auto";
    PlaybackDirection                  direction = "normal";
    DOMString                          easing = "linear";
};

dictionary OptionalEffectTiming {
    double                             delay;
    double                             endDelay;
    FillMode                           fill;
    double                             iterationStart;
    unrestricted double                iterations;
    (unrestricted double or DOMString) duration;
    PlaybackDirection                  direction;
    DOMString                          easing;
};

dictionary ComputedEffectTiming : EffectTiming {
    unrestricted double  endTime;
    unrestricted double  activeDuration;
    double?              localTime;
    double?              progress;
    unrestricted double? currentIteration;
};

[Exposed=Window,
 Constructor ((Element or CSSPseudoElement)? target,
              object? keyframes,
              optional (unrestricted double or KeyframeEffectOptions) options),
 Constructor (KeyframeEffect source)]
interface KeyframeEffect : AnimationEffect {
    attribute (Element or CSSPseudoElement)? target;
    attribute IterationCompositeOperation    iterationComposite;
    attribute CompositeOperation             composite;
    sequence<object> getKeyframes ();
    void             setKeyframes (object? keyframes);
};
```

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -96,12 +96,13 @@ Abstract: This specification defines a model for synchronization and
     for interacting with this model and it is expected that further
     specifications will define declarative means for exposing these
     features.
-Ignored Terms: double, boolean, DOMString, unsigned long, unrestricted double, (animationeffectreadonly or effectcallback), (unrestricted double or DOMString)
+Ignored Terms: double, boolean, DOMString, unsigned long, unrestricted double, (unrestricted double or DOMString)
 Ignored Vars: RGB|A, G, t
 Link Defaults: css-transforms (property) transform
 </pre>
 <pre class="anchors">
 urlPrefix: https://heycam.github.io/webidl/#dfn-; type: dfn; spec: webidl
+    text: present
     text: platform object
     text: nullable; url: nullable-type
     text: throw
@@ -283,7 +284,7 @@ elem.animate({ transform: 'scale(0)', opacity: 0 }, 300);
 // Print the id of any opacity animations on elem
 elem.getAnimations().filter(
   animation =&gt;
-    animation.effect instanceof KeyframeEffectReadOnly &&
+    animation.effect instanceof KeyframeEffect &&
     animation.effect.getKeyframes().some(
       frame =&gt; frame.hasOwnProperty('opacity')
     )
@@ -299,7 +300,7 @@ elem.getAnimations().filter(
 // Slow down and replay any transform animations
 elem.getAnimations().filter(
   animation =&gt;
-    animation.effect instanceof KeyframeEffectReadOnly &&
+    animation.effect instanceof KeyframeEffect &&
     animation.effect.getKeyframes().some(
       frame =&gt; frame.hasOwnProperty('transform')
     )
@@ -1567,7 +1568,7 @@ an animation is typically performed asynchronously.
 var animation = elem.animate({ left: '100px' }, 2000);
 animation.playbackRate = 2;
 animation.currentTime = 1000; // animation is now finished
-animation.effect.timing.iterationCount = 2; // animation is no longer finished
+animation.effect.updateTiming({ iterationCount: 2 }); // animation is no longer finished
 </pre></div>
 
 The one exception to this asynchronous behavior is when the <a>finish an
@@ -1918,11 +1919,11 @@ animation into a <a lt="finished play state">finished</a> animation
 without restarting by setting the [=start time=] such as below:
 
 <div class='example'><pre class='lang-javascript'>
-animation.effect.timing.duration = 5000;
+animation.effect.updateTiming({ duration: 5000 });
 animation.currentTime = 4000;
 animation.pause();
 animation.ready.then(function() {
-  animation.effect.timing.duration = 3000;
+  animation.effect.updateTiming({ duration: 3000 });
   alert(animation.playState); // Displays 'paused'
   animation.startTime =
     document.timeline.currentTime - animation.currentTime * animation.playbackRate;
@@ -3916,11 +3917,11 @@ API by the {{Animation}} interface.
 
 <pre class='idl'>
 [Exposed=Window,
- Constructor (optional AnimationEffectReadOnly? effect = null,
+ Constructor (optional AnimationEffect? effect = null,
               optional AnimationTimeline? timeline)]
 interface Animation : EventTarget {
              attribute DOMString                id;
-             attribute AnimationEffectReadOnly? effect;
+             attribute AnimationEffect?         effect;
              attribute AnimationTimeline?       timeline;
              attribute double?                  startTime;
              attribute double?                  currentTime;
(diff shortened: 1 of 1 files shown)


birtles:
The one item I should call attention to is that I haven't added a `setTiming()` method. I figure it is less useful and easy to add later if necessary.

web-animations-1/Overview.bs
```diff
@@ -4261,156 +4327,99 @@ interface AnimationEffectTimingReadOnly {
...
+::  Fill backwards and forwards.
 
-The {{AnimationEffectTimingProperties}} dictionary encapsulates the timing
-properties of an {{AnimationEffectReadOnly}} so that they can be set in bulk (as
-with the {{Animation()}} constructor) or returned as a readonly snapshot (as
-with the {{AnimationEffectReadOnly/getComputedTiming()}} method of the
-{{AnimationEffectReadOnly}} interface).
+:   <code>auto</code>
+::  No fill.
+    In a subsequent level of this specification, this will produce different
```
stephenmcgruer: s/will/may ?

web-animations-1/Overview.bs
```diff
@@ -4261,156 +4327,99 @@ interface AnimationEffectTimingReadOnly {
...
 
-<div class='members'>
+:   <code>normal</code>
+::  All iterations are played as specified.
 
-:   <dfn dict-member for=AnimationEffectTimingProperties>delay</dfn>
-::  See the {{AnimationEffectTiming/delay}} attribute of
-    the {{AnimationEffectTiming}} interface.
+:   <code>reverse</code>
+::  All iterations are played in the reverse direction from the way
```
stephenmcgruer: s/way/order ?

rolivo:
Bastard HDP , you kill web-animations specification

birtles:
@rolivo Please raise your specific concerns as a separate issue. From your comment I can't tell what the concern is.

rolivo:
You take away all the beauty of web technologies, with this change they reduce the use cases that the api allows, they only dedicate themselves to destroy the specifics. GroupEffects, SequenceEffects, terrible losses and now this, The only thing that is worthwhile of the web technologies is webassembly, the rest is crap.

astearns:
@rolivo this isn't the right place to bring up issues, and the way you're engaging isn't welcome. If you have a specific concern to raise, and you can do this in a respectful manner, please open an issue.

PR #2436, opened 2018-03-13
[css-fonts-4] Larger font size can also have a11y implications

xfq:
See https://github.com/w3c/i18n-drafts/issues/125 for an example.


xfq:
Also, I might be missing something, but it looks like the "user-agent style sheet" in this sentence should be "user style sheet", because the user agent style sheet is the default style sheet of the UA, instead of the user-specified one.

Relevant spec: https://drafts.csswg.org/css-cascade/#cascading-origins

AmeliaBR:
> Making text smaller is an unusual acessibility requirement

Not that unusual. I personally find too-large body text annoying & difficult to read (and I really hate the trend towards using vw units in font-size to increase it on wider screens, often opening up dev tools to fix it).

It becomes a serious accessibility issue for people with limited field of vision, who need a constrained-width column of reasonably-sized text, not big text that stretches across the entire screen.

svgeesus:
Goof points, @AmeliaBR

PR #2457, opened 2018-03-20
[css-text-3] Fix Thai sample

### LINKED ISSUE #2455: [css-text] Questionable Thai words
https://drafts.csswg.org/css-text-3/#word-break-property

แและ·ตัวอย่าง·การเขียน·ภาษาไทย has two many แ characters at the start.

Also, are we sure about the word segmentation for these examples? I tested it by using Firefox to wrap the text, and ended up with the following words:

และ•ตัวอย่าง•การ•เขียน•ภาษา•ไทย

Literally, that means "and sample how write language thai". The current spec word divisions seem to map rather to that of the translation(?)

frivoal:
Closes #2455


r12a:
> (I'll raise a separate issue anyway, because there are no break points before the punctuation in wb:break-all.)

Ah, i may be confused about the difference between word-break:all and line-break:anywhere again.

frivoal:
> even so it's a bit strange, especially given that the example sentence starts with Chinese

I don not believe that there are established punctuation conventions for text that spans 4 unrelated writing systems, so a full stop is just as good as anything else. The punctuation is indeed there to show `word-break:break-all` has no expected effect on punctuation (as `line-break:anywhere` is what does that).

> Ah, i may be confused about the difference between word-break:all and line-break:anywhere again.

Yes, but I cannot blame you. These properties and values are all terribly named (although `line-break:anywhere` isn't the worst), but by and large we're stuck with them.

frivoal:
I am so confused at how github works. Why is it that I made this commit, you merged it, and the commit history [shows you  as the author of this commit](https://github.com/w3c/csswg-drafts/commit/3a640ed2403d3a4002f629bcb7189021bd524265)? Not of the merge commit, of the actual commit.

I noticed because [twitter told me](https://twitter.com/CSScommits/status/976619788431749120) you  fixed this, and I was wondering why you didn't merge the commit I had made. Maybe I had got it wrong somehow? I cannot read Thai, so it is very possible. But no, it turns out you did merge my PR, but github claims you wrote it. Probably a side effect of rebase-and-merge?

I don't care about being given credit for this particular commit, but I it seems bad that github seems to be messing with attribution in general. Beyond affecting who gets credit, it will also hide that the spec has external contributors when their Pull Requests get merged by WG members, which seems problematic from an IP point of view.

@fantasai did you do something unusual when merging this, making my worries about github unfounded?

fantasai:
I committed the same changes as you, and then realized you had made a PR for them, and so I just merged it--which was a valid action, but merged an empty diff, so it closed the PR without adding a changeset into the logs.

frivoal:
Aha. Makes more sense. So much for my conspiracy theory about github's evil rewriting :)

PR #2461, opened 2018-03-21
Fix a build error in Bikeshed

### LINKED ISSUE #2460: [cssom] Bikeshed build broken by "Use Web IDL's new-ish interface mixins concept"
https://drafts.csswg.org/cssom/ hasn't been updated after https://github.com/w3c/csswg-drafts/pull/2123 and https://drafts.csswg.org/bikeshed/cssom/ says:
```
FATAL ERROR: Line doesn't match the grammar `k:v; k:v; k:v`:
ProcessingInstruction includes LinkStyle;
Traceback (most recent call last):
  File "/sites/api.csswg.org/bikeshed/bikeshed.py", line 5, in 
    bikeshed.main()
  File "/sites/api.csswg.org/bikeshed/bikeshed/cli.py", line 202, in main
    doc.preprocess()
  File "/sites/api.csswg.org/bikeshed/bikeshed/Spec.py", line 125, in preprocess
    self.lines = datablocks.transformDataBlocks(self, self.lines)
  File "/sites/api.csswg.org/bikeshed/bikeshed/datablocks.py", line 81, in transformDataBlocks
    doc=doc)
  File "/sites/api.csswg.org/bikeshed/bikeshed/datablocks.py", line 644, in transformInclude
    el = "
Please correct the error and update the source file.

Last Bikeshed run at 2018-03-20 15:26:10 PDT
```

I've also filed https://github.com/tabatkins/bikeshed/issues/1207

romandev:
This fixes #2460.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1245,7 +1245,9 @@ must also define when a <a>CSS style sheet</a> is
 
 <!-- XXX load/error events -->
 
-<pre class="idl">ProcessingInstruction includes LinkStyle;</pre>
+<pre class=idl>
+ProcessingInstruction includes LinkStyle;
+</pre>
 
 The <dfn>prolog</dfn> refers to <a>nodes</a> that are children of the
 <a>document</a> and are not <a>following</a> the


romandev:
@tabatkins, @foolip PTAL

tabatkins:
Hm, this shouldn't be necessary - I thought I allowed single-line `pre` just fine. (When it is multiline, I do purposely only allow the start/end tag by themselves on their line, with the content on intervening lines.)  I'll look into it.

tabatkins:
Oh, I see the problem! Since I look for data-blocks before HTML parsing (so they can include things that would make HTML unhappy), I do a *really* minimal parse to figure out what things are.

In particular, this build error is caused by me finding that the line starts with `<pre`, and then seeing that it includes the substring `include`, which classifies it as an include-block rather than an ordinary `pre`. :(

I should bite the bullet and actually do a little bit of real parsing of the tag; that bit of the HTML parser isn't complicated.

tabatkins:
The build error has been fixed on Bikeshed' side, so closing this.

PR #2467, opened 2018-03-23
rephrase to make it clear which shadow root

fergald:
When dealing with mappings on host elements there are 2 possible meanings to "shadow root" - the one attached to the host and the one containing the host. In the spec, only one makes sense, so it's unambiguous but this change makes it explicit.

### DIFF
--- a/css-shadow-parts-1/Overview.bs
+++ b/css-shadow-parts-1/Overview.bs
@@ -133,17 +133,17 @@ Each part mapping is one of:
 
 <dl class=switch>
     : <code>ident</code>
-    :: Adds «[ ident → el ]» to the <a>shadow root's</a> <a>shadow part map</a>.
+    :: Adds «[ ident → el ]» to the <a>shadow part map</a> of the <a>shadow root</a> containing el.
 
     : <code>ident1 => ident2</code>
     :: If el is a <a>shadow host</a>,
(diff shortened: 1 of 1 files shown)


fergald:
@tabatkins

fergald:
@tabatkins

fergald:
pr-preview of this is [here](https://pr-preview.s3.amazonaws.com/fergald/csswg-drafts/4/f815c69...5f9d66e.html) it's against my own repo since pr-preview doesn't currently work for w3c/csswg-drafts

css-shadow-parts-1/Overview.bs
```diff
@@ -7,7 +7,7 @@ Status: ED
 Work Status: exploring
 URL: http://drafts.csswg.org/css-shadow-parts/
 Editor: Tab Atkins-Bittner, Google, http://xanthir.com/contact/, w3cid 42199
-Abstract: This specification defines the ''::part()'' and ''::theme()'' pseudo-elements on <a>shadow hosts</a>, allowing <a>shadow hosts</a> to selectively expose chosen elements from their <a>shadow tree</a> to the outside page for styling purposes.
```
kojiishi: Can you explain why we want to take these links out?

fergald: That shouldn't be happening. I think I took the wrong base for this PR. I will fix that.

PR #2470, opened 2018-03-23
[mediaqueries-4] Move 'pointer'/'hover' note, reword "can potentially" to "should strongly consider"

patrickhlauke:
Based on https://www.smashingmagazine.com/2018/02/media-queries-responsive-design-2018/, and in particular based on the article's note

> "I’d suggest that you should take great care in using these, as making a user switch from their primary pointing device is likely to result in a poor user experience."

I think the fact that relying purely on `pointer` and `hover` can also be just as bad has been missed out by having the note where it is. Moving it right to the start (and strenghtening the suggestion) would probably help in making sure authors don't get too complacent about relying just on whatever the browser/device determines to be the primary pointer input, as ignoring any alternative inputs would conversely force users to only use the primary input even if they don't want to, for whatever reason.


frivoal:
Sounds reasonable to me. merging.

patrickhlauke:
super, thanks @frivoal

Issue #2414, opened 2018-03-07
[css-shadow-parts] IDL for attributes


fergald:
As discussed at Web Components F2F (2018-03, Tokyo) and in [TAG issue](https://github.com/w3ctag/design-reviews/issues/230#issuecomment-370644319) update [spec](https://drafts.csswg.org/css-shadow-parts/) to include IDL for part, partmap, theme to allow programmatic API.

domenic:
Since these are supposed to be "superglobal", the best precedent is things like https://dom.spec.whatwg.org/#dom-element-id

fergald:
Posting discussion about programmatic from #2411 to here.

@fergald
> > In my model the partmap is a dictionary where the values are Set()s of part names. So deleting would be partmap['a'].delete('b') and to remove "a" entirely, delete partmap['a'] (although partmap['a'].clear() would have the same effect).

> > What's the tradeoff between a dictionary of sets and something with an explicit API. Are there operations that dictionaries/sets allow that we wouldn't want or conversely are there operations dictionaries/sets don't allow that we would want and couldn't add later without breaking compatibility? I don't really want to write up an API for something that behaves essentially as a dictionary of sets if I can avoid it.

@rniwa
> Well, browser engines can't observe changes to a dictionary so your proposal as currently stands will make partmap readonly, and any changes to it will be discarded. It's even weirder because it has [SameObject] on it. It means the browser engine has to update the dictionary continuously? I don't think it works because then author scripts can override property descriptors of a dictionary, and all sorts of insanity ensues.

domenic:
I would strongly suggest just exposing these as strings in IDL for now. There is no need for a more complicated data structure, at least until authors ask for one.

We have precedent for microsyntaxes being reflected as simple strings in at least `<img srcset>`, `<iframe csp>`, `<iframe allow>`, and `<link sizes>`.

We can always add something more complex later, depending on how we find developers are using these things, e.g. feature policy is looking at a *read-only* API that gives a view onto the `allow` attribute. But for now, no need.

domenic:
Concretely, and with regard to the feedback in https://github.com/w3c/csswg-drafts/issues/2411#issuecomment-377768803, I think this means:

- Say "content attribute" in most (all?) places you current say "attribute". You can omit it if it's quite obvious, e.g. I think "the [part] attribute" is fine as long as [part] links to the `<dfn>` which will say "The **part** content attribute".
[...]

PR #2475, opened 2018-03-26
[css-shadow-parts] Add IDL for partList and partMap

fergald:
Add basic IDL. I'm not very confident about this change. I expect lots of comments.

@hayatoito @domenic @TakayoshiKochi @tabatkins 

Progresses #2414

### DIFF
--- a/css-shadow-parts-1/Overview.bs
+++ b/css-shadow-parts-1/Overview.bs
@@ -278,3 +278,23 @@ but never match additional <a>shadow-part pseudo-elements</a>.
     would select just the one button's label,
     ignoring any other labels.
 </div>
+
+Extensions to the {{Element}} Interface {#idl}
+==============================================
+
+<pre class=idl>
+dictionary PartMap {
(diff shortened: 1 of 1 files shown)


domenic:
This design for partMap doesn't really work, I'm sorry to say. Dictionaries are not good for representing map data structures; among other things, they cannot be returned from attributes.

I would suggest one of two designs:

- partMap is just a DOMString. It's not a token list, and there's no pre-made structure for it, so just do the natural thing. This is what we do for e.g. ``'s `srcset=""` attribute.
- partMap is a new multimap. For this you'll need something similar to https://xhr.spec.whatwg.org/#interface-formdata (although less complicated since you just need DOMString, not (File or USVString)). This would be quite a bit of work.

I'd suggest just doing the DOMString version for now. We can always add the more complicated version later, with a different name (similar to `className` vs. `classList` or `rel` vs. `relList`).

fergald:
Going to withdraw this and come back with something different in a bit.

PR #2481, opened 2018-03-28
Explain focus-visible intent and guidance.

robdodson:
@tabatkins — we wanted to more thoroughly explain the intent of the focus-visible pseudo, and improve the conformance guidelines. This is our first stab at it, any feedback you (or others) have is much appreciated.

cc @alice @bkardell

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -2256,47 +2256,55 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
 		like clicking on the button with a mouse pointer.
 	</div>
 
-	This pseudo-class is intended to give developers the ability to style focus
-	selectively, in order to encourage visible, usable focus styles, and
-	discourage the practice of disabling the default user agent focus style.
+	The goal of this selector is to explain how the browser's default focus
+	indicator behavior works in order to encourage more visible, usable focus
+	styles, and discourage the practice of disabling the default user agent focus
+	style.
 
-	When UAs choose to specially indicate focus on an element,
-	or whether they specially indicate focus at all,
-	is UA-dependent.
-	Different UAs,
-	the same UA on different operating systems,
-	or the same UA on the same OS,
-	but with different user settings,
-	can make different choices as to when an element matches '':focus-visible''.
+	For example, a `:focus` style will always be applied
+	to a focused element, regardless of _how_ it was focused. Whereas a
+	`:focus-visible` style would only be applied to an element if it matches the
+	same heuristic that the UA uses to draw its default focus indicator.
 
-	Since part of the problem is user agents' default focus indicator behaviour,
-	it is recommended that browsers implementing the `:focus-visible`
-	selector also use it in their user agent stylesheets to show the default focus
-	indicator.
+	The user agent should take into account the available signals as to whether
+	the user might find focus indication useful. The result of these heuristics
+	must determine whether `:focus-visible` should match on the currently focused
+	element.
 
-	<div class=note>
-		The user agent should take into account the available signals as to whether
-		the user might find focus indication useful. The result of these heuristics
-		must determine whether `:focus-visible` should match on the currently
-		focused element.
+	By default, if there are no signals indicating that focus should be
+	indicated, then focus should not be indicated on the focused element, unless
+	the user has expressed a preference for always seeing a focus ring.
 
-		By default, if there are no signals indicating that focus should be
-		indicated, then focus SHOULD NOT be indicated on the focused element, unless
-		the user has expressed a preference for always seeing a focus ring.
+	The following may be considered signals:
 
-		The following may be considered signals:
+	* An element which requires the use of a keyboard, such as a text input,
+	should always have a visible focus indicator.
+	* If a user is using keyboard-based navigation by using the tab and arrow
+	keys, the focused element should be indicated. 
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -2256,6 +2256,10 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
 		like clicking on the button with a mouse pointer.
 	</div>
 
+	This pseudo-class is intended to give developers the ability to style focus
+	selectively, in order to encourage visible, usable focus styles, and
+	discourage the practice of disabling the default user agent focus style.
+
 	<div class=advisement>
```
alice: Should this section be removed, as it's essentially duplicated below?

robdodson: fixed

selectors-4/Overview.bs
```diff
@@ -2280,19 +2269,34 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
...
-
-		* If the element received focus via a keyboard interaction,
-			including indirectly,
-			such as triggering a dialog
-			by pressing a button using the keyboard,
-			apply '':focus-visible''.
-		* If a keyboard event occurs while an element is focused,
-			even if the element wasn't focused by a keyboard interaction,
-			apply '':focus-visible''.
+	Since part of the problem is user agents' default focus indicator behaviour,
```
dbaron: Please prefer US spelling of behavior.

dbaron: I wonder if "the problem" should be spelled out in slightly more words, as well, since it doesn't seem tied to the previous paragraph.

selectors-4/Overview.bs
```diff
@@ -2256,20 +2256,9 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
 		like clicking on the button with a mouse pointer.
 	</div>
 
-	<div class=advisement>
-		Page authors should follow these guidelines
```
dbaron: It seems a little disappointing to remove all of the advice for authors.  It's not clear to me that the old advice was the best that could be offered, but this does seem like an area where it's worth offering some expert advice on what the right thing really is.

selectors-4/Overview.bs
```diff
@@ -2256,20 +2256,9 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
...
-			(such as text fields or buttons),
-			use '':focus-visible''.
-		* Otherwise, if the element is emulating a text input,
-			or something else that is <em>intended</em> to receive keyboard interaction,
-			use '':focus''.
-		* Otherwise,
-			use '':focus-visible''.
-	</div>
+	This pseudo-class is intended to give developers the ability to style focus
+	selectively, in order to encourage visible, usable focus styles, and
```
triblondon: As a developer, personally I find the previous form of words more straightforward and easier to understand.  I'm struggling particularly with the word 'selectively'.  The purpose of `:focus-visible` is, as I understand it, to style the focus ring while not overriding the UA's discretion about whether to display it.  Therefore this mechanism intentionally doesn't give the developer control over that discretion, whereas I'd interpret 'selectively' in this text as implying that it does.

selectors-4/Overview.bs
```diff
@@ -2280,19 +2269,34 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
...
-			including indirectly,
-			such as triggering a dialog
-			by pressing a button using the keyboard,
-			apply '':focus-visible''.
-		* If a keyboard event occurs while an element is focused,
-			even if the element wasn't focused by a keyboard interaction,
-			apply '':focus-visible''.
+	Since part of the problem is user agents' default focus indicator behaviour,
+	it is recommended that browsers implementing the `:focus-visible`
+	selector also use it in their user agent stylesheets to show the default focus
```
dbaron: I don't think "show" is quite the right verb here.  It's possible it could even be "explain", or perhaps "specify".  It might also help to add an additional clarifying sentence saying that the point is that when `:focus-visible` matches should correspond to the UA's default behavior for when to show focus indicators.  In fact, it might be worth stating *that* as a conformance requirement (and then saying that an easy way to implement that requirement may be by using `:focus-visible` in the UA style sheet.)

robdodson:
Thank you for the reviews @dbaron and @triblondon! I'll revise and ping the thread again when it's ready for another look.

selectors-4/Overview.bs
```diff
@@ -2256,20 +2256,9 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
...
-			or something else that is <em>intended</em> to receive keyboard interaction,
-			use '':focus''.
-		* Otherwise,
-			use '':focus-visible''.
-	</div>
+	This pseudo-class is intended to give developers the ability to style focus
+	selectively, in order to encourage visible, usable focus styles, and
+	discourage the practice of disabling the default user agent focus style.
 
 	When UAs choose to specially indicate focus on an element,
```
alice: I think this paragraph confuses the issue about the problem we're trying to solve. The problem is the practice of disabling the default UA focus style, mentioned in the previous paragraph.

alice:
Thanks for the feedback - we're going to close this PR for now while we figure out exactly what advice we want to give browser implementors and authors respectively.

PR #2516, opened 2018-04-07
[cssom] Have property setters append new declaration rather than set in place

### LINKED ISSUE #1898: [css-logical-1] [css-cascade-3] The all longhand probably shouldn't set logical properties.

upsuper:
This resolves #1898.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2209,8 +2209,8 @@ a list of declarations <var>declarations</var>, follow these steps:
 
 <ol>
  <li>If <var>property</var> is a <a>case-sensitive</a> match for a <a for="CSS declaration">property
- name</a> of a <a>CSS declaration</a> in <var>declarations</var>, let <var>declaration</var> be that <a>CSS declaration</a>.
- <li>Otherwise, append a new <a>CSS declaration</a> with the <a for="CSS declaration">property name</a> <var>property</var>
+ name</a> of a <a>CSS declaration</a> in <var>declarations</var>, remove that <a>CSS declaration</a>.
+ <li>Append a new <a>CSS declaration</a> with the <a for="CSS declaration">property name</a> <var>property</var>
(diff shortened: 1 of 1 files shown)


upsuper:
I'll have a look at the wpt and see if there's any existing to update.

upsuper:
Test is in w3c/web-platform-tests#10354.

upsuper:
@emilio Well, I don't think you should merge it without approval from the current editor really...

astearns:
@upsuper we added Emilio as editor today

Issue #2149, opened 2017-12-30
[cssom] getComputedStyle only supports ::before/::after


gsnedders:
https://drafts.csswg.org/cssom/#extensions-to-the-window-interface

Roughly, from what I've tested:

* Chrome, Firefox, and Safari all support gCS `::first-letter` and `::first-line`.
* Firefox throws `NS_ERROR_NOT_AVAILABLE` for `::marker` (and doesn't support `::marker` AFAIK), Chrome doesn't support it so returns the element, and Safari does support `::marker` and does support it for gCS
* Firefox supports gCS `::placeholder`, Chrome and Safari does not (despite supporting the feature)

Is there any reason not to allow all pseudo-elements for gCS?

upsuper:
I think it makes sense to support all pseudo-elements which the user-agent supports, and it probably should also be specified that what should happen when the pseudo-element is not supported.

Personally I think throwing an exception (maybe `TypeError`) is better than returning the style of the element, but I can accept argument on that.

Note that, Firefox currently throws `NS_ERROR_NOT_AVAILABLE` when getting property value from the object returned from gCS in that case, and gCS itself doesn't throw, which, I would think, is a bug. I think gCS should throw, or everything should work smoothly.

FWIW, there could be tricky cases for supporting returning all properties on all pseudo-elements, for example, since many pseudo-elements have only limited properties applied, user-agents may want to avoid computing the rest if possible, and may optimize their storage, which can lead to observable difference on returned value.

Also some properties which use used value rather than computed value as resolve value might be tricky in some cases, although it isn't clear to me what the cases would be.

FremyCompany:
Edge supports a broad range of them, and like Chrome we fallback to returning the style of the element if we are not able to return the pseudo element style. We are supposed to support -webkit-input-placeholder but I am not sure it's actually working in the end, this could be just a bug. 

I believe the advantage of returning a functioning style object when asked about a pseudo-element you don't support is that you don't crash, with a bit of chance the script will do just fine with slighly wrong values.

PR #2534, opened 2018-04-11
[cssom] Fix #2403 and #2149

### LINKED ISSUE #2403: [cssom] Consider changing the definition of getComputedStyle when called from a different document to match reality.
Somewhat related to https://github.com/w3c/csswg-drafts/issues/1548 and https://github.com/w3c/csswg-drafts/issues/1964.

Right now the spec doesn't really match any implementations. Implementations usually return the style of the element styled as the document it's in (Firefox doesn't in some cases, I can try to explain if people are curious).

In any case, for a visible iframe and an element in the document, the behavior of all engines is definitely not the one required by the spec in https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle:

> Note: This means that even if obj is in a different document (e.g. one fetched via XMLHttpRequest) it will still use the style rules associated with the document that is associated with the global object on which getComputedStyle() was invoked to compute the CSS declaration block.

And they just return the style of the element for the document they're in.

cc @lilles @FremyCompany @bzbarsky @dbaron

emilio:
See individual commits for reference.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2789,6 +2789,11 @@ steps:
    <dt><a for="CSSStyleDeclaration">declarations</a>
    <dd>All longhand properties that are <a lt="supported CSS property">supported CSS properties</a>, in lexicographical order, with the value being
    the <a>resolved value</a> computed for <var>obj</var> using the style rules associated with <var>doc</var>.
+
+   Issue: There are UAs that handle shorthands, and all UAs handle shorthands
+   that used to be longhands like 'overflow', see
+   <a href="https://github.com/w3c/csswg-drafts/issues/2529">#2529</a>.
+
    <dt><a for="CSSStyleDeclaration">parent CSS rule</a>
    <dd>Null.
    <dt><a for="CSSStyleDeclaration">owner node</a>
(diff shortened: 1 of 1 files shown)


cssom-1/Overview.bs
```diff
@@ -2791,9 +2790,9 @@ steps:
    <dd>All longhand properties that are <a lt="supported CSS property">supported CSS properties</a>, in lexicographical order, with the value being
    the <a>resolved value</a> computed for <var>obj</var> using the style rules associated with <var>doc</var>.
 
-   Note: This means that even if <var>obj</var> is in a different document (e.g. one fetched via <code>XMLHttpRequest</code>) it will still use
```
emilio: This should go in the first commit.

cssom-1/Overview.bs
```diff
@@ -2791,9 +2790,9 @@ steps:
...
-   the style rules associated with the document that is associated with the global object on which
-   {{Window/getComputedStyle()}} was invoked to compute the <a>CSS declaration block</a>.
+   Issue: There are UAs that handle shorthands, and all UAs handle shorthands
+   that used to be shorthands like 'overflow', see
```
emilio: and s/shorthands/longhands/ here.

PR #2544, opened 2018-04-11
Replace all CSS2 copies with a single css2/ equal to the 2011/6 REC

gsnedders:
This is from the css21/ directory in b739b88dc9d2252de3c6bfcf27647f15a56d63ea with changes to variables in the Makefile to make the build exactly match https://www.w3.org/TR/2011/REC-CSS2-20110607 ("W3C Recommendation 07 June 2011, edited in place 12 April 2016 to point to new work").

This is stage one of the plan for ongoing maintenance of CSS 2.1 (#2553) along with tagging the previous revision on master to this landing.


tantek:
[@gsnedders](https://twitter.com/gsnedders) this looks good. I spot checked the 550(!!!) files changed and the diffs I saw were consistent with what I would expect with reverting to 2011 CSS2.1 REC with 2016 updates.

(Originally published at: http://tantek.com/2018/102/t1/)

gsnedders:
FWIW, all I've done here is rebase it onto master (with no changes to the first commit) and added a second commit to ignore all files from the build system except for those already checked in. I'll merge this soon unless anyone objects.

PR #2563, opened 2018-04-13
Fix #1548.

### LINKED ISSUE #1548: [cssom] [css-display] [css-values] Clarify behavior of window.getComputedStyle on detached subtrees, and the definition of "the root element".
Consider the following test-case:

```html
<!doctype html>
<style>
:root { font-size: 100px; }
</style>
<script>
let detached = document.createElement('div');
detached.style.display = "inline";
detached.style.fontSize = "1rem";
alert(getComputedStyle(detached).display);
alert(getComputedStyle(detached).fontSize);
</script>
```

The behavior across browsers is as follows:

 * Gecko: `block`, `16px`.
 * Edge: `inline`, `16px`
 * Gecko (with Servo's style system): `inline`, `100px`.
 * Blink and WebKit: Empty string in both cases.

Per https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle, there isn't anything off-hand that justifies WebKit/Blink's behavior here.

Now, for the other interop problems the question is: what is the definition of "the root element"?

There are two spec sections that potentially apply here:

https://drafts.csswg.org/css-display/#transformations (1):

> The root element’s display type is always blockified. Additionally, a display of contents computes to block on the root element.

and https://drafts.csswg.org/css-values/#rem (2):

> Equal to the computed value of font-size on the root element. When specified on the font-size property of the root element, the rem units refer to the property’s initial value.

So the different behaviours are as follows:

 * Gecko takes the detached element as "the root element". For Gecko "the root element" is effectively the root of the inheritance subtree. That means that in Gecko, for example, `rem` units don't work as expected on the `::backdrop` pseudo (since it inherits from nothing, and thus `1rem` will always be the initial font-size).

 * Edge seems to apply that same definition for (2), but not for (1).

 * Servo uses the document element of the document for both sections of the spec (which I think it's consistent, and doesn't have the problem with `::backdrop`).

 * Blink and WebKit seem to just not compute style if the element is outside of the document, which seems wrong per https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle.

So, two questions:

 * Is the behavior of Blink/WebKit justified by any reasonable interpretation of the spec, or is it just a browser bug? It sounds like the second to me, but worth confirming.

 * What is the definition of "the root element"?

Edge's behavior seems inconsistent here, but both of the Gecko or Servo's interpretation look consistent to me.

Gecko's behavior have the inconveniency of `rem` units on `::backdrop` though. I haven't checked other browser's behaviour in depth with regards to `rem` units and `::backdrop`. From code inspection, I think Blink at least uses the document element's font-size[1].

[1]: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/css/resolver/ElementResolveContext.cpp?l=63&rcl=709a3cbb9e85dd0c4572ee30a32bda68140fbb52

emilio:
See individual commits for reference.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2787,7 +2787,7 @@ steps:
   and its <a>shadow-including root</a>'s <a>browsing context</a> either doesn't have a
   <a>browsing context container</a>, or that <a>browsing context container</a>
   is <a>being rendered</a>,
-  let <var>decls</var> be a list of all longhand properties that are
+  set <var>decls</var> to a list of all longhand properties that are
   <a lt="supported CSS property">supported CSS properties</a>, in lexicographical order,
   with the value being the <a>resolved value</a> computed for <var>obj</var>
   using the style rules associated with <var>doc</var>.
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2784,9 +2784,9 @@ steps:
 
  <li>Let <var>decls</var> be an empty list of <a>CSS declarations</a>.
  <li>If <var>elt</var> is <a>connected</a>, part of the <a>flat tree</a>,
-  and its <a>shadow-including root</a>'s <a>browsing context</a> either doesn't have a
-  <a>browsing context container</a>, or that <a>browsing context container</a>
-  is <a>being rendered</a>,
+  and its <a>shadow-including root</a> has a <a>browsing context</a> which either
+  doesn't have a <a>browsing context container</a>, or whose
+  <a>browsing context container</a> is <a>being rendered</a>,
   set <var>decls</var> to a list of all longhand properties that are
   <a lt="supported CSS property">supported CSS properties</a>, in lexicographical order,
   with the value being the <a>resolved value</a> computed for <var>obj</var>
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2779,19 +2779,24 @@ steps:
   versions. That is, both <code>:before</code> and <code>::before</code> should
   match above.
 
+ <li>Let <var>decls</var> be an empty list of <a>CSS declarations</a>.
+ <li>If <var>elt</var> is <a>connected</a> and part of the <a>flat tree</a>,
+  let <var>decls</var> be a list of all longhand properties that are
+  <a lt="supported CSS property">supported CSS properties</a>, in lexicographical order,
+  with the value being the <a>resolved value</a> computed for <var>obj</var>
+  using the style rules associated with <var>doc</var>.
+
+  Issue: There are UAs that handle shorthands, and all UAs handle shorthands
+  that used to be longhands like 'overflow', see
+  <a href="https://github.com/w3c/csswg-drafts/issues/2529">#2529</a>.
+
  <li>
   Return a live <a>CSS declaration block</a> with the following properties:
   <dl>
    <dt><a for="CSSStyleDeclaration">readonly flag</a>
    <dd>Set.
    <dt><a for="CSSStyleDeclaration">declarations</a>
-   <dd>All longhand properties that are <a lt="supported CSS property">supported CSS properties</a>, in lexicographical order, with the value being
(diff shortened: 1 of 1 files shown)


cssom-1/Overview.bs
```diff
@@ -2779,19 +2782,27 @@ steps:
   versions. That is, both <code>:before</code> and <code>::before</code> should
   match above.
 
+ <li>Let <var>decls</var> be an empty list of <a>CSS declarations</a>.
+ <li>If <var>elt</var> is <a>connected</a>, part of the <a>flat tree</a>,
+  and its <a>shadow-including root</a>'s <a>browsing context</a> either doesn't have a
+  <a>browsing context container</a>, or that <a>browsing context container</a>
+  is <a>being rendered</a>,
+  let <var>decls</var> be a list of all longhand properties that are
```
annevk: You want to use set here, since you're overwriting something. See https://infra.spec.whatwg.org/. (Older specifications are inconsistent with this unfortunately.)

cssom-1/Overview.bs
```diff
@@ -2779,19 +2782,27 @@ steps:
   versions. That is, both <code>:before</code> and <code>::before</code> should
   match above.
 
+ <li>Let <var>decls</var> be an empty list of <a>CSS declarations</a>.
+ <li>If <var>elt</var> is <a>connected</a>, part of the <a>flat tree</a>,
+  and its <a>shadow-including root</a>'s <a>browsing context</a> either doesn't have a
+  <a>browsing context container</a>, or that <a>browsing context container</a>
+  is <a>being rendered</a>,
```
annevk: When would it (not) be rendered?

cssom-1/Overview.bs
```diff
@@ -2779,19 +2782,27 @@ steps:
   versions. That is, both <code>:before</code> and <code>::before</code> should
   match above.
 
+ <li>Let <var>decls</var> be an empty list of <a>CSS declarations</a>.
+ <li>If <var>elt</var> is <a>connected</a>, part of the <a>flat tree</a>,
+  and its <a>shadow-including root</a>'s <a>browsing context</a> either doesn't have a
```
annevk: What if the root doesn't have a browsing context?

emilio: Could you put an example of a document which has no browsing context? It's not clear to me when can that happen.

emilio: This is why I asked you for review btw, thanks a lot for taking a look :-)

annevk: `new Document()`, `xhr.responseXML`, `new DOMParser().parseFromString(...)`, ...

cssom-1/Overview.bs
annevk: Ah yeah, that makes sense.

cssom-1/Overview.bs
emilio: Alright, I think we don't want to return styles there either, since those documents have no rendering behavior. Pushed two fixups for the issues you raised, mind taking a look? Thanks!

annevk:
This looks fine to me, editorially and behaviorally, but I do wonder whether it matches implementations. In my own testing many years ago I've found that `getComputedStyle(someElementFromAnotherDocumentWithoutBrowsingContext)` does end up returning styles, if I remember correctly.

It's probably worth accompanying this change with web-platform-tests. (I don't know if the CSS WG has such a policy yet, but at the WHATWG each normative change to a standard requires corresponding tests and bugs against failing implementations.)

emilio:
Yeah, we decided this in #1548. This is kind of a mess in terms of interop right now. I added tests for this in https://github.com/w3c/web-platform-tests/pull/10422.

emilio:
Basically, this is where we want to move. It may not be possible to not return styles in display: none iframes because of web compat, but we decided to give it a shot.

annevk:
Cool. I'd generally wait for at least one browser to ship before committing on the model, but seems like most things (apart from browser bugs) are covered here.

Issue #2554, opened 2018-04-12
Review old CSS2 edits


gsnedders:
- [ ] a54258755ea2a845dfaa677ecbe63802dc607313 explicitly specify what gets set to what with MCW
 - [ ] 0b331afcb77d23f589281e0d181fa0abdd03606d [css2] Fix typo
 - [ ] 71aaa7d27702af3b4d389e8db8f9b28368a4cecc [css2] Editorial tweaks about line-height
 - [ ] 4fd269813a66ec3f84eba7d95150d7715eedf86a [css2] Tighten the definition of the content area height for inlines
 - [ ] d8f466a53b3e09c28099377a85f8cee3816773bb [css2] Fix line-height:normal with only fallback fonts
 - [ ] b55a7344841b52c42e9962354169a2c34742ff90 [css2] Make line-height:normal match implementations
 - [ ] a3c7617bafdc3ea7201b77dde93719ecc2fea8ca [css2] Make line-height match implementations for values other than normal
 - [ ] c2a93bf97d118caa67a834394b38ee40a554ca6e [css2] Correct misleading note
 - [ ] e75b25a7259155b146680460b613785ed451b6a7 [CSS2] Typo. Regenerated.
 - [ ] 3652e8b825e6c50a96a7023cb51dc2bb5fcac6db [CSS2] Added omitted text back in and made the links to corresponding level-3 modules into notes. Regenerated the HTML.
 - [ ] c048c4132510a4e0c99fb8eff1bf4dc98b0c0596 [CSS2] Regenerated.
 - [ ] c262ac90a69786c759d28eacc4a7715906e8470a [CSS2] Added links to relevant tests.
 - [ ] 2614704bce8c2e82c2bcc85493f5b3a6d8b75bb2 [CSS2] Regenerated.
 - [ ] 99b7998e2e06218e3f0f693463194878023a2bb5 [CSS2] Added date to the change about baselines of overflowing inline-block.
 - [ ] 288a8807619dd592a1d7690689770803b4cea4e3 [CSS2] Scripts accept both name= and id= on input, only generate id= on output. Replaced the definition of the color property by a reference to css3-color. Replaced some name= by id=. Updated links to editors' draft and to ERCIM. Refer to css-syntax-3 for white space, because there is no text in the syntax chapter anymore. Removed redundant class=example. Regenerated HTML files. Trying a different link checker. Add support for links to properties defined in CSS modules and no longer in CSS2 itself.
[...]

PR #2570, opened 2018-04-14
[CSS2] Reapply most editorial changes not by Bert

gsnedders:
This starts on #2554.

### DIFF
--- a/css2/cover.html
+++ b/css2/cover.html
@@ -23,7 +23,7 @@
 
 <div class="head">
   <p><a href="http://www.w3.org/"><img alt="W3C" height="48"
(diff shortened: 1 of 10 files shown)


frivoal:
Github tells me my review is requested, but at the same time the "DO NOT LAND" comments suggests this isn't ready.

Either way, I had a look. I cannot vouch for the completeness of this (and suspect many things need to be added, although I did not check), but at least the bits that are there seem fine.

gsnedders:
@frivoal basically the "css2/" in csswg-drafts claims to be the REC, and we probably don't want any changes in it till the status gets changed to a 2.2 ED (which is #2581).

PR #2573, opened 2018-04-15
[css-pseudo-4] EventTarget is not a mixin

saschanaz:
`implements` is only for mixins.


foolip:
@tabatkins can you review/merge? I've run into the same problem and also had to fix it, found this PR when looking to send one myself.

foolip:
Thanks @ericwilligers and @tabatkins.

This change hasn't propagated to reffy-reports yet, I've filed https://github.com/tidoust/reffy/issues/141 about that.

Issue #2008, opened 2017-11-24
[css2] Naming of revision of CSS 2.1


gsnedders:
As far as I can tell, looking through WG minutes, we never decided what to call the revision of CSS 2.1 incorporating errata. I think we have two options:

 * CSS 2.1 2nd Edition
 * CSS 2.2

I'm in favour of CSS 2.1 2nd Edition because with the exception of scientific notation for numbers, we haven't added/removed any features, it's a simple update to incorporate errata, and as such in my view it doesn't constitute a new subversion of CSS 2.

SelenIT:
Doesn't three formatting context instead of two (Table formatting context added) count as a feature?

frivoal:
Which one are you talking about: CSS2.1 + with all the work in progress errata, or the one with the limited subset or errata that are ready for REC republication?

I think the later does not need a new name at all. It's just CSS2.1, and if you want to be very specific, you can still say CSS2.1, 07 June 2011 Recommendation, or CSS2.1 29 February 2019.

For the former, it probably should have a name, but not a new version number, since it isn't meant to be a new document that ever gets published as a replacement for the 2.1 REC, just a staging area for changes. So I'd call it something along these lines: "CSS2.1, errata staging draft"

SelenIT:
I'd prefer following the pattern consistent with the HTML spec development:

- CSS2.1 2nd Edition (like [HTML5.1 2nd Edition](https://www.w3.org/TR/html51/)) as a republished REC, with the limited subset of errata incorporated;
- its own Editor's draft (like https://rawgit.com/w3c/html/html5.1-2/single-page.html) as its errata staging draft, without the separate published verion;
- CSS2.2 (like [HTML5.2](https://www.w3.org/TR/html52/)), with all CSS2.1 errata incorporated and all sections that have been updated with newer CSS modules replaced with links to those modules, as a future CSS2.x release that will eventually replace the CSS2.1 as the "[Latest CSS level 2](http://www.w3.org/TR/CSS2/)";
- [Editors draft of "the latest CSS level 2"](https://drafts.csswg.org/css2/) pointing to the latest state of CSS2.x development (like https://w3c.github.io/html/).

[...]

PR #2581, opened 2018-04-16
[CSS2] change status to ED, bump version to 2.2

gsnedders:
See https://github.com/w3c/csswg-drafts/issues/2008#issuecomment-381088973 for the last two commits, we should decide what we want to do.

e.g., do we want the spec to say:

>  If the containing block's width depends on this element, then the resulting layout is undefined in CSS 2.2.

or: 

> 
>  If the containing block's width depends on this element, then the resulting layout is undefined in CSS 2.

If we want the former, we need to drop the last commit on this PR; if we want the latter, we should squash the last two commits together and reword the commit message.

Aside from that, I wouldn't worry too much about the changes in cover.src; we can always reword them later, it's more important to have something that claims it's a 2.2 ED to start modifying generally.


frivoal:
Given that specref / bikeshed make `[[CSS2]]` link to 2.1 (and presumably will update that to link to 2.2 when 2.2 gets to REC), I think it would be better to have 2.2 nth edition generally refer to CSS2, except when specifically discussing the differences between 2.0, 2.1, and 2.2

However, to do that, this last commit seems insufficient, since the about.html section keeps using CSS2 to refer to CSS2.0, in order to explain how it supersedes it.

I suggest adding a sentence like:

> The original &lt;a href="... dated URL...">1998 Recommendation of CSS2&lt;/a> is hereafter referred to as CSS2.0 in order to distinguish it from CSS2.1 and CSS2.2, even though its original name did not carry this ".0" qualifier.

then disambiguate the relevant CSS2 references accordingly.

gsnedders:
> However, to do that, this last commit seems insufficient, since the about.html section keeps using CSS2 to refer to CSS2.0, in order to explain how it supersedes it.

I don't think we have any precedent for calling it "2.0" (which given the full title of later revisions would imply it's "CSS level 2 revision 0").

frivoal:
Sure, but we need to call the old REC something other than just CSS2 if we want CSS2 to refer the latest of CSS2 revision x, n<sup>th</sup> edition, for any x and any n.

I think CSS2.0 works for this purpose. It is not what the thing was initially called, but is unambiguous, and anybody who knows of the 1998 REC should easily guess that it is what we're talking about.

Although that term has no precedent in writing (that I know of), I think the phrase is used occasionally in speech.

gsnedders:
> re: CSS 2.0 - don't make up a new old version number never before used. If need be, use "CSS2 1998" in prose to disambiguate because that's when that REC was.

Still need to go through the spec and change all instances of "CSS2" (no whitespace) to "CSS2 1998" or whatever.

> This is the right edit to also update Editors (just the two of us IIRC) and add Previous Editors (move all prior editors here, plus add Elika per f2f resolution).

Your suggestion before was to leave me out for now, FWIW, and I've done that for now.

gsnedders:
@tantek this now uses "CSS2 (1998)" throughout wherever it means to refer to it

css2/cover.src
```diff
@@ -78,39 +94,32 @@
...
-CSS&nbsp;2.1, and then they should only reference CSS2 for those
-features, or preferably reference such feature(s) in the respective
-CSS3 Module that includes those feature(s)).
+<p>CSS&nbsp;2.2 is derived from and is intended to replace the
+CSS&nbsp;2.1 and CSS2&nbsp;(1998)
+specifications. See <a href="changes.html">appendix C of this
+document</a> for changes from CSS&nbsp;2.1,
+and <a href="/TR/2011/REC-CSS21-2011-06-07/changes.html">appendix C of
+CSS&nbsp;2.1 for changes from CSS2&nbsp;(1998)</a> (noting that certain CSS2&nbsp;(1998) features
+were removed from CSS&nbsp;2 in CSS&nbsp2.1). All future
```
Ms2ger: Missing semicolon in the latter nbsp.

gsnedders: There's lots of entities that are missing semicolons. :( Because, y'know, it's valid SGML and software should cope. Or something.

PR #2597, opened 2018-04-20
Establishing a new formatting context

frivoal:
As discussed in the resolution of #1457 we should add to css-display a linkable definition for establishing a (new) formatting context without explicitely saying which kind, and still have the right thing happen unambiguously when possible, while being explicit that it is a spec bug to invoke this definition in the cases where it's not meaningful/possible.

This PR attempts to add this definition, and to update all the specs that should use it to do so.

I am not particularly attached to the phrasing I am proposing here. I believe it to be correct, but certainly welcome more elegant ways of saying the same thing if anyone has suggestions.

Review from @fantasai or @tabatkins (but particularly @fantasai) appreciated.

Changes in css-contain css-multicol css-align and css-rhythm should be particularly safe, since this is just adding a link without rephrasing anything.

Changes to the other specs are meant to be no-ops that merely unify the way we talk about this, without any normative effect.

### DIFF
--- a/css-display-3/Overview.bs
+++ b/css-display-3/Overview.bs
@@ -1055,6 +1055,28 @@ Appendix A: Glossary</h2>
 			<p class="example">
 				As another example, margins do not collapse across [=formatting context=] boundaries.
 
+			Specifications may call for an element or box to
+			<dfn export lt="establishes a formatting context | establish a formatting context | establish a new formatting context | establishes a new formatting context">establish a formatting context</dfn>
+			without further precision as to the type of formatting context or how to do so.
+			This operation is not applicable to
+			non-atomic inline-level boxes
+			nor to boxes with a layout-internal display-type.
+			Specifications that invoke this
+			must either exclude such boxes,
+			or to blockify them first.
+			If the box is a block container that does not establish a BFC,
+			it is made to establish a BFC.
+			Otherwise, the box already establishes a formatting context, and this operation is a no-op.
+			This has no effect on the computed value of the display property.
+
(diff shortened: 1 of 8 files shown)


Loirooriol:
I think https://drafts.csswg.org/css-display-3/#transformations should say that blockifying causes the element to establish a new formatting context. Then there would be no need to explicitly say that e.g. flex items are both blockified and establish a new formatting context, the former would imply the later.

frivoal:
@Loirooriol Could you file that as a separate issue? This is kind of related, but it seems to me to be an independent decision to make, so a separate issue seems better (also, this issue is already way too long).

css-display-3/Overview.bs
```diff
@@ -1047,6 +1047,28 @@ Appendix A: Glossary</h2>
 			<p class="example">
 				As another example, margins do not collapse across [=formatting context=] boundaries.
 
+			Specifications may call for an element or box to
+			<dfn export lt="establishes a formatting context | establish a formatting context | establish a new formatting context | establishes a new formatting context">establish a formatting context</dfn>
+			without further precision as to the type of formatting context or how to do so.
+			This operation is not applicable to
+			non-atomic inline-level boxes
+			nor to boxes with a layout-internal display-type.
```
Loirooriol: The correct terminology seems "display type" without hyphen. And I reiterate that display types should probably belong only to elements (and pseudo-elements), not boxes, but this problem is all over the spec.

fantasai:
We had to do some substantial rewriting on this, so closing this PR. (We'll link to it when we commit our results.)

PR #2603, opened 2018-04-22
Mention "CSS namespace" rather than "CSS interface"

heycam:


dbaron:
Could you add an `<a id="the-css-interface">` around the contents of the h3 so that the old id works too?

PR #2620, opened 2018-04-27
[css-grid-2] Updating initial example to match per-axis model

rachelandrew:
The initial example was based on the two-axis model, updated it to be per-axis.


fantasai:
I think this particular case doesn't matter, actually: each li is only one row tall, so subgridding both or just subgridding the columns will give the same result.

PR #2621, opened 2018-04-27
[css-contain] Typo: use plural

Loirooriol:
There are 2 effects listed below


frivoal:
I need to get less lousy as spelling and grammar.

PR #2636, opened 2018-05-02
Fix typo: "header element" to "heading element"

kbav:
Change "header element" to "heading element". "Heading" is intended here, since it is referring to HTML heading elements above, and "heading" is also used in a prior description (line 1271).


fantasai:
Thanks! Merged in. :)

PR #2666, opened 2018-05-09
[css-animations] Change an AnimationEvent member from float to double

stephenmcgruer:
As per the web-idl spec, float should only be used if there is a specific
reason, as double much more closely matches the ECMAScript Number
type. In the case of AnimationEvent's 'elapsedTime' member there does
not appear to be such a reason.

Furthermore, switching to double allows the elapsedTime to reach higher
values without overflowing. A rough calculation (with some assumptions
and possible errors) suggests that this change will raise the celing
from ~4.5 hours to years.


stephenmcgruer:
@birtles for review

birtles:
Looks good to me. I checked that this change doesn't break any changes in Firefox's test suite and it seems fine.

birtles:
@stephenmcgruer Do you mind submitting a similar PR for transitions?

PR #2670, opened 2018-05-11
[css-contain] Fix typo

mrego:


css-contain-1/Overview.bs
```diff
@@ -168,7 +168,7 @@ Types of Containment</h2>
 	restricting the effects that its descendants can have on the rest of the page in various ways.
 	<a>Containment</a> enables much more powerful optimizations by user agents,
 	and helps authors compose their page out of functional units,
-	as it limits how widely an effect a given change can have on a document.
+	as it limits how widely effect a given change can have on a document.
```
frivoal: I agree with you that the original sentence has a problem, but I think your fix is wrong too :) It should be:
> "[...] how wide an effect a given change can have on a document."
or
> "[...] how widely a given change can affect a document."

Pick your favorite.

mrego: I guessed it was not right yet.

I picked the second one, it seems clearer to me.

PR #2706, opened 2018-05-25
[css-logical] Typo: use plural

Loirooriol:
`float` and `clear` are 2 properties, so use plural.


dontcallmedom:

PR #2707, opened 2018-05-27
Convert CSS2 vector image source files to SVGs

### LINKED ISSUE #2583: Convert image source files to SVGs
CSS2 has images in loads of different formats: *.eps, *.fig, *.gif, *.idraw, *.png, *.ps, and *.svg. (There's also *.jpe?g, but that's unused and removed in #2569.) The build system converts all of these into PNGs.

The *.eps, *.fig, *.idraw, and *.ps files should be converted into SVGs, rather than using proprietary or rare formats. The *.gifs should be converted into *.pngs and we should just drop the *.gifs entirely (having the gifs there and having the build system convert them is just… pointless).

Once we have all the images in SVGs, we can then ask a further question as to whether we should use the SVGs in the CSS2 spec.

And, uh, hey, I did this (also) on my flight to CA: https://github.com/w3c/csswg-drafts/compare/master...gsnedders:css2-svg

Note this isn't yet landable: a lot of the *.pngs have a fair change from the current spec.

gsnedders:
Fix #2583: convert CSS2 vector image source files to SVGs

This gets rid of the various other formats we have, standardising entirely on SVG rather than using the current Xfig and Autodesk Graphic (formerly iDraw) files. Note these are programmatic conversions and the SVG files aren't particularly readable; these should be cleaned up as when we need to modify them.

Notable changes, presumably either from the source file not matching the committed PNG or the conversion process not working:

 * The current PNGs all seem fairly whitewashed on every device I have here, seemingly because of the gamma value set in them.
 * clip.png's fill colour of the shaded blocks has changed

Things definitely needing fixed before merging:

 - [x] boxdimeg.png is broken by the font size getting smaller
 - [x] first-letter.png looks pretty broken (font changes)
 - [x] first-letter2.png likewise
 - [x] float2p.png is broken by the font size getting smaller
 - [x] floatclear.png is broken by the font size getting smaller
 - [x] initial-cap.png is broken by the font size getting smaller
 - [x] tbl-border-conflict.png has dramatically changed viewport size
 - [x] tbl-empty.png is dramatically different (this is almost certainly just the committed PNG and idraw being very different)
 - [x] this branch also regenerates images generated from .gifs; we should drop those from the SVG conversion commit (and probably get rid of the GIF source, given GIF to PNG conversion helps nobody)
 - [x] remove use of Myriad Pro
 - [x] remove use of Verdana
 - [x] make zopfli optional

That said, please take a look and see if you can find anything else broken.


css2/Makefile
```diff
@@ -268,12 +264,11 @@ LN= ln
...
-%.png: %.ps; convert -density 96x96 $< $@
-%.png: %.idraw; convert -density 96x96 $< $@
-%.png: %.svg; convert -density 96x96 $< $@
-%.png: %.gif; convert -density 96x96 $< $@
-%.eps: %.fig; fig2dev -L eps -F -c $< $@
+%.big.png: %.svg; convert -density 96x96 $< $@
+%.big.png: %.gif; convert -density 96x96 $< $@
+
+%.png: %.big.png
+	zopflipng -y -m $< $@
```
gsnedders: This should probably be made optional, though it's probably a good idea by default (to minimise the object size of the blobs committed to the repo, given they'll become part of the history forever).

gsnedders: This isn't outdated.

css2/images/anon-block.svg
```diff
@@ -0,0 +1,79 @@
...
+       transform="scale(0.1)"
+       id="g1212"><path
+         id="path1214"
+         style="fill:#bfbfbf;fill-opacity:1;fill-rule:evenodd;stroke:none"
+         d="m 24,379.301 h 2832 v 1097.4 H 24 Z" /><path
+         id="path1216"
+         style="fill:#404040;fill-opacity:1;fill-rule:evenodd;stroke:none"
+         d="m 62.0078,432 h 2745.98 V 896 H 62.0078 Z" /><path
+         id="path1218"
+         style="fill:#808080;fill-opacity:1;fill-rule:evenodd;stroke:none"
```
dirkschulze: Is this Inkscape? A looooot of nonsense :P

gsnedders: Yarp. Sadly the only way to convert these from their source files without the image coming out entirely corrupt, from my trials of things that did less craziness.

gsnedders:
cc/ @fantasai if you want to review this

gsnedders:
`for (let el of document.querySelectorAll("[data-file-type='.png'] button.rendered")) el.click()` in case it helps anyone review by just looking at the changes to the PNGs

fantasai:
@gsnedders I trust you. If you've looked at the output and there aren't pieces of images missing, I'm sure it's fine. :)

PR #2713, opened 2018-05-28
[css-transforms-2] Explicitly reset the temp matrix in matrix recomposition.

emilio:
The idea of the blocks (IIUC) are to reset `temp` to be an identity matrix in
case it's been used before.

However if only skew[0] and skew[2] are non-zero, this code didn't manage to
accomplish it.

Make it so that the reset is explicit after setting a matrix member instead of
try to undo all other possible mutations that have happened before.

### DIFF
--- a/css-transforms-2/Overview.bs
+++ b/css-transforms-2/Overview.bs
@@ -1073,14 +1073,14 @@ matrix = multiply(matrix, rotationMatrix)
 if (skew[2])
 		temp[2][1] = skew[2]
 		matrix = multiply(matrix, temp)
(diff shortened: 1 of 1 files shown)


dirkschulze:
If resetting is not the intention there still would be something wrong with the current code in the spec. Your change does seem to make sense but I’d like to look over it again first.

svgeesus:
close-and-reopen to kick the ipr bot

svgeesus:
@dirkschulze @emilio is this ancient PR still wanted?

emilio:
I think so?

PR #2726, opened 2018-05-31
[mediaqueries-3] Mention ASCII case-insensitivity, not only case-insensitivity.

emilio:
Not that it's observable I think, since there are no defined media features with non-ascii characters, but...

### DIFF
--- a/mediaqueries-3/Overview.html
+++ b/mediaqueries-3/Overview.html
@@ -593,8 +593,9 @@ Y  y|\\0{0,4}(59|79)(\r\n|[ \t\r\n\f])?|\\y</pre>
   <p><code>RESOLUTION</code> is to be added to the CSS2 <code>term</code>
    production.
 
-  <p>CSS style sheets are generally case-insensitive, and this is also the
-   case for media queries.
+ <p>CSS style sheets are generally <a
+   href="https://infra.spec.whatwg.org/#ascii-case-insensitive">ASCII
+   case-insensitive</a>, and this is also the case for media queries.
 
   <p>In addition to conforming to the syntax, each media query needs to use
    media types and media features according to their respective specification
--- a/mediaqueries-3/Overview.src.html
+++ b/mediaqueries-3/Overview.src.html
@@ -432,8 +432,9 @@ Y  y|\\0{0,4}(59|79)(\r\n|[ \t\r\n\f])?|\\y</pre>
 <p><code>RESOLUTION</code> is to be added to the CSS2
 <code>term</code> production.</p>
 
-<p>CSS style sheets are generally case-insensitive, and this is
-also the case for media queries.</p>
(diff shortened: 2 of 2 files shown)


frivoal:
As that there's no effective difference between case insensitive and ASCII case insensitive in this case since there are no defined media features with non-ascii characters, I think this PR ends up just being editorial. Since updating a REC is relatively process heavy, I'm tempted to rejected it. It's not wrong, but it seems too little value for too much work.

@atanassov @astearns @svgeesus, what do you think?

astearns:
I think it's not worth making the change to level 3

emilio:
Sure, sounds fine, closing then. Thanks!

dbaron:
There is a detectable difference between ASCII-case-insensitive and Unicode-case-insensitive.  For example, should "HEİGHT" (with a dot on the capital I) be a valid media feature?  The capital dotted I (İ) should lowercase to a standard ASCII lowercase i.

emilio:
Reopening based on that.

svgeesus:
> Since updating a REC is relatively process heavy, I'm tempted to rejected it. It's not wrong, but it seems too little value for too much work.

True, although lobbing it into an erratum is pretty easy, if we decide to accept this.

svgeesus:
Please link `ASCII case-insensitive` as requested by @tabatkins

emilio:
Done

svgeesus:
Now I need to list that as an erratum ...

frivoal:
@svgeesus Should we make a test for that, and the update the MQ3 REC? Erratum are hard to read and easy to miss. I think the answer's yes, and if the process turns out to have too much friction to make this convenient, the process should be fixed and that gives me something to do on the AB.

frivoal:
Also, if we're going to be errataing / republishing a REC, I think we should actually have a WG resolution on the change. @astearns @atanassov ?

emilio:
https://wpt.fyi/results/css/mediaqueries/mq-case-insensitive-001.html is already there and tests the case-insensitive / ascii-case-insensitive difference.

We landed it as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1464091, and it passes in all browsers.

So this is a bit of a no-brainer IMO, but sure, should be fine to have a resolution for this if you want :)

frivoal:
> tests [...] passes in all browsers.
Great.

> So this is a bit of a no-brainer IMO, but sure, should be fine to have a resolution for this if you want :)

Sure, I have no disagreement about the technical issue. But there are checks along the way to publishing things on TR, and I think leaving a clear trail of how we got there helps reduce friction.

svgeesus:
> @svgeesus Should we make a test for that, and the update the MQ3 REC?

Yes, all errata _should_ have tests (and all errata rolled into a new edition of a spec _must_ have tests).

Issue #2458, opened 2018-03-20
[css-ui-4][css-sizing-3] Who defines box-sizing?


annevk:
It seems both CSS Sizing and CSS UI put a claim on it. This makes linking it from other standards hard.

(In particular, I noticed this when looking at warnings for https://quirks.spec.whatwg.org/.)

Loirooriol:
From #1906 it seems it should be CSS Sizing.

annevk:
Thanks, it seems that was semi-fixed as nobody updated UI...

fantasai:
@frivoal Mind removing from ui-4?

frivoal:
@fantasai 
I'd like to (since I was the one who reported #1906), but the the definition was rewritten (by you and/or Tab) when moving adding it to css-sizing, and I had not yet taken the time to check whether anything normative had been lost or changed during the rewrite.

Now I have, so here's my review:

----

The rewrite is generally an improvement, both in purely editorial terms, and because it makes us of more anchoring terminology than what was available for the css-ui version.

However:

1. The statement that `box-sizing` affects flex-basis is good, but a new addition to the definition of `box-sizing`. I cannot find any test that supports it. There's a [2012 resolution](https://lists.w3.org/Archives/Public/www-style/2012May/0520.html) about flexbox that supports it and it was in flexbox before, so that addition is correct, but adding tests when modifying post-CR normative prose, or at least flagging the need for it, would be good.
2. Same comment about "Values affected by box-sizing include [...] those used in functional notations such as fit-content()". That addition is a good one, but it should be supported by tests, and currently isn't.
3. The new examples are good, but you dropped the one that showed how `box-sizing` let you put two `width: 50%` things with a `1em` border and have them fit side by side as their outer size adds up to `100%`. I think that was worth having. Even if we've invented things like flexbox since that example was written, it remains true.
4. The CSS-UI version defined terms line `inner min width` and monkey patched css2.1 to use these terms where appropriate rather than the ambiguous `minimum width` which might refer to the inner one or the outer one, as well as `content width` as opposed to just `width`. The resulting monkey patch was not pleasant to read (or to write), but CSS21 was judged sufficiently non obvious that @bzbarsky had requested (multiple times) full disambiguation, and that writing it (and the corresponding tests) uncovered bugs. 
    You dropped both the term definitions and the money-patch, and replaced it with a much tighter statement that terms like `width`, `minimum width` and friends always refer to the inner size, defining them to be non-ambiguous.
[...]

PR #2739, opened 2018-06-06
Box sizing migration

frivoal:
This PR solves https://github.com/w3c/csswg-drafts/issues/2458


css-sizing-3/Overview.bs
```diff
@@ -53,6 +53,8 @@ Module interactions</h3>
 	<p>This module extends the 'width', 'height', 'min-width', 'min-height', 'max-width', 'max-height', and 'column-width'
 	features defined in [[!CSS2]] chapter 10 and in [[!CSS3COL]]
 
+	<p>The definition of the 'box-sizing' property in module supersedes the one in [[CSS-UI-3]].
```
fantasai: I think I'd probably make this a note?

css-sizing-3/Overview.bs
```diff
@@ -620,22 +622,67 @@ Box Edges for Sizing: the 'box-sizing' property</h3>
...
 	<a lt="max width">maximum (max) width</a>, and
 	<a lt="max height">maximum (max) height</a>
-	refer to the <a lt="inner size">inner</a> size
+	generally refer to the <a lt="inner size">inner</a> size
 	(<a lt="content box">content-box</a> size)
 	of a <a>box</a>
 	unless otherwise indicated.
 
+	Note: [[CSS-UI-3#box-sizing]] provides an explicit disambiguation of these terms
+	for the <a href="https://www.w3.org/TR/CSS21/visudet.html">Visual formatting model details</a> section of [[CSS2]].
```
fantasai: s/Note:/Refer to/; s/provides/for/

css-sizing-3/Overview.bs
```diff
@@ -620,22 +622,67 @@ Box Edges for Sizing: the 'box-sizing' property</h3>
...
+
+	<div class=advisement>To avoid ambiguities,
+	specification authors should avoid ambiguous terms such as width or height without further qualification,
```
fantasai: “should avoid ambiguous uses of terms such as width or height without further qualification”

because sometimes the disambiguation is formed by context and not needed in a specific instance of the term--it's ambiguous usage that's a problem

css-ui-4/Overview.bs
```diff
@@ -140,196 +140,14 @@ which itself replaced and superseded the following:
 		<a href="https://www.w3.org/TR/2000/WD-css3-userint-20000216">User Interface for CSS3 (16 February 2000)</a> [[CSS-UI-3]]
 </ul>
 
+One feature previously defined in [[CSS-UI-3]] has been moved to [[CSS-SIZING-3]] rather than this specification.
```
fantasai: Delete this sentence and the next two headings, mark the following added text as a note, and add empty anchors for all the removed anchors, e.g. via <a id=".."></a> or somesuch. This way links won't be broken, but we also don't have to carry around an empty section.

css-sizing-3/Overview.bs
frivoal: I made all the requested changes except this one.

If it is a note, then css-sizing does not normatively supersede css-ui, and we have two specs both claiming to the the canonical version. That sounds less than ideal. Merging as is for expediency, but feel free to fix after the fact if you still disagree.

PR #2743, opened 2018-06-06
[css-text-3] %age text-indent refers to the width of the content box

### LINKED ISSUE #2394: [css-text-3] percentage on text-indent
We've discussed this previously, but did not conclude. Old thread starts here:
https://lists.w3.org/Archives/Public/www-style/2015Feb/0250.html

CSS2.1 and CSS-TEXT-3 seem to agree that text-indent applies to block level elements, and that percentages are resolved against the width of containing block, i.e. the parent.

Firefox violates that, and they resolve the percentage against the content area of the element itself.

Even though this seems like a spec violation, it seems to be nicer behavior, as discussed for example in this email: https://lists.w3.org/Archives/Public/www-style/2015Feb/0273.html

>  given
> 
> ```
>    #outer { width: 500px; }
>    #inner { width: 200px; text-indent: 10%; }
> 
>    <div id="outer">
>      <div id="inner">
>        Lorem ipsum dolor sit amet...
> ```
> 
> an author will expect the text to have a first-line indent of 20px, not 
> 50px.
> 
> We also get the surprising result that, assuming <p> doesn't have its 
> own text-indent specified, the two lines in
> 
> ```
>    <div id="outer">
>      <div id="inner">
>        Line one.
>        <p>Line two.
> ```
> 
> have different indents. Again, I don't think authors would expect this.

Also, here's another mail where @fantasai argues that refering to the containing block has some odd consequences in tables.
https://lists.w3.org/Archives/Public/www-style/2010Sep/0431.html

Since both behaviors (+ bug based variations) are attested in browsers, we're probably not overly constrained by web compat. Going by the width of the content area of the element makes more sense to me (and to everyone who participated in the previous email threads AFAICT) so I suggest we switch to that.  

Most typically, text-indents are specified in lengths, not in percentages, so maybe we shouldn't overthink this, but we should get consistency anyway.

----

Note: In discussions of this topic, the question of interaction between percentage text-indents and intrinsic sizing is also occasionally raised, but it is present regardless of the solution we pick, so I think these can remain independent, and there's another issue for that: https://github.com/w3c/csswg-drafts/issues/1597

frivoal:
Closes #2394

### DIFF
--- a/css-text-3/Overview.bs
+++ b/css-text-3/Overview.bs
@@ -1988,7 +1988,7 @@ Cursive Scripts</h4>
     Initial: 0
     Applies to: block containers
     Inherited: yes
-    Percentages: refers to width of containing block
+    Percentages: refers to <a lt="inner size">inner</a> <a>inline size</a> of the element
     Computed value: the percentage as specified or the absolute length, plus any keywords as specified
     Animation type: <a href="https://drafts.csswg.org/css-transitions-1/#animtype-lpcalc">length, percentage, or calc</a>, but only if keywords match
     Canonical order: <abbr title="follows order of property value definition">per grammar</abbr>


fantasai:
Closed via https://github.com/w3c/csswg-drafts/commit/3f8e4e8924935639f2f6935c496dd0c2300fc1ac

PR #2763, opened 2018-06-14
[css-contain] Disable size containment on table elements

### LINKED ISSUE #2746: [css-contain] SIze contain shouldn't apply to tables
[The section on size containment](https://drafts.csswg.org/css-contain/#containment-size) says size containment has no effect on internal table elements, however it seems like it also shouldn't apply to tables (which aren't internal table elements). 

cc: @dbaron

mrego:
Closes #2746.


frivoal:
The patch itself was good, but going forward, an addition to the change section and to the DoC would be welcome. It is much easier to keep track of if we do it as we go.

Done in 2bfc0914fb5f387583754aa0f809a34739dd1f03

PR #2797, opened 2018-06-19
[css-animations] Change definition of GlobalEventHandlers to mixin

csnardi:
`GlobalEventHandlers` is now defined as a `mixin` per https://html.spec.whatwg.org/multipage/webappapis.html#globaleventhandlers. Update the partial interface's definition to indicate that it is a `mixin`.


birtles:

PR #2801, opened 2018-06-21
[css-contain] property names linkified

### LINKED ISSUE #2800: [css-contain] 'counter-increment' and 'content' property names should be linkified (but are not), in style containment section
In [css-contain-1 section 3.3 "Style Containment"](https://drafts.csswg.org/css-contain-1/#containment-style), the "counter-set" property-name is nicely linkified, but "counter-increment" and "content" are not, for some reason.

Let's make them linkified, if we can!

(I imagine this is some bikeshed / spec-HTML-building infrastructure issue, under the hood, and perhaps there needs to be some sort of disambiguation or something in the bikeshed markup? *shrug*)

ewilligers:
'counter-increment' and 'content' property names are now working links.

resolves #2800

### DIFF
--- a/css-contain-1/Overview.bs
+++ b/css-contain-1/Overview.bs
@@ -15,7 +15,6 @@ Editor: Tab Atkins, Google, http://xanthir.com/contact/, w3cid 42199
 Editor: Florian Rivoal, On behalf of Bloomberg, https://florian.rivoal.net/, w3cid 43241
 Abstract: This CSS module describes the 'contain' property, which indicates that the element's subtree is independent of the rest of the page. This enables heavy optimizations by user agents when used well.
 Ignored Terms: scrollWidth, scrollHeight, clientWidth, clientHeight
-Link Defaults: css-lists-3 (property) counter-increment
 </pre>
 <pre class="anchors">
 spec: css2; urlPrefix: https://www.w3.org/TR/CSS2/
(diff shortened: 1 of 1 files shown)


frivoal:
Sorry, this needs to be reverted:

1) linking to `css-content` instead of `css2` is not appropriate. `css-content` is a known-to-be buggy draft. It is not as bad as it used to be, as some maintenance work has been done to it, but I don't think it has reached the point where it is more accurate than `css2`. Case in point: it claims that the `content` property applies to all elements, which is absolutely not true in existing implementations. `css2` correctly says that it applies to ":before and :after pseudo-elements "

2) independently of the point above, neither `css-content` nor `css-list-3` are mature enough to be accepted without strong justifications as normative references for a document which is trying to exit CR, so we will have to change these links anyway before next publication of `css-contain`

frivoal:
sub-point on `css-list-3`: for `counter-set`, `css-list-3` is vaguely acceptable, as:
1) `counter-set` is not defined anywhere else
2) it can be argued that `counter-set` is almost identical to `counter-reset`, which is in a REC.
That makes it possible to argue in favor of using css-list-3 as a reference for `counter-set`, but harder to argue in favor of it for things that exist in CSS2.

frivoal:
Commit https://github.com/w3c/csswg-drafts/commit/26d5bf68b31f7d53888739f339ec44f2f1f72726 fixes this. And since it uses css2 rather than css21, we do get cross-spec links.

Issue #2629, opened 2018-04-30
[css-env-1] defining variables


ghost:
This is a follow up to: https://drafts.csswg.org/css-env-1/#issue-ed0904e4

The current Safari implementation does not have a way to define variables. A noticeable difference from var is the missing double dash at the beginning of variable names e.g. env(safe-area-inset-let). Maybe we could prefix developer defined variables with the double dash and UA defined variables should not have the double dash.

This would make it easier to differentiate between the two types of variables, would stop the developer overriding UA variables and ensure consistency with what has already been shipped.

tabatkins:
Yeah, that's the intent - author-defined ones would be `--foo`, so they live in a separate namespace from UA-defined ones, and we don't have to worry about clashes.

bramus:
Closing this one in favor of https://github.com/w3c/csswg-drafts/issues/2627 which discusses potential syntaxes to define custom envvars.

PR #2807, opened 2018-06-22
[css-env-1] Define safe area inset variables #2629

rebeccahughes:
Define CSS safe area inset variables as part of #2629

### DIFF
--- a/css-env-1/Overview.bs
+++ b/css-env-1/Overview.bs
@@ -77,6 +77,46 @@ tho user agents may define additional undocumented [=environment variables=]:
 
 * Issue: define list of predefined vars
 
+Safe area inset variables {#safe-area-insets}
+------------------------------------------------------------------
+
+<table>
+	<tr>
+		<th>Name</th>
+		<th>Value</th>
+		<th>Media</th>
+	</tr>
+	<tr>
+		<td>safe-area-inset-top</td>
+		<td><<length>></td>
+		<td>visual</td>
+	</tr>
+	<tr>
+		<td>safe-area-inset-right</td>
+		<td><<length>></td>
+		<td>visual</td>
+	</tr>
+	<tr>
+		<td>safe-area-inset-bottom</td>
+		<td><<length>></td>
+		<td>visual</td>
+	</tr>
+	<tr>
+		<td>safe-area-inset-left</td>
+		<td><<length>></td>
+		<td>visual</td>
+	</tr>
+</table>
+
+The safe area insets are four [=environment variables=] that define a rectangle by
+its top, right, bottom, and left insets from the edge of the viewport. For rectangular
+displays, these must all be zero, but for nonrectangular displays they must form a
+rectangle, chosen by the user agent, such that all content inside the rectangle is
+visible, and such that reducing any of the insets would cause some content inside of
+the rectangle to be invisible due to the nonrectangular nature of the display. This
+allows authors to limit the layout of essential content to the space inside of the
+safe area rectangle.
+
 Using Environment Variables: the ''env()'' notation {#env-function}
 ===================================================================
 


rebeccahughes:
cc: @tabatkins @grorg

css-env-1/Overview.bs
```diff
@@ -77,6 +77,53 @@ tho user agents may define additional undocumented [=environment variables=]:
...
+	Applies to: all elements
+	Value: <<length>>
+	Initial: 0
+	Inherited: yes
+	Media: visual
+	Animatable: no
+</pre>
+
+The safe area insets are four [=environment variables=] that define the top, right, bottom and
+left insets from the edge of the document. Page content that is in this safe area may be
```
astearns: 'document' seems wrong here - shouldn't it be something like 'viewport'?

astearns: And I've been assuming the safe area is the bit inside the insets, so it I think it would be 'content that is outside this safe area' or 'content that is inside these insets'

rebeccahughes: Sounds good - fixed.

css-env-1/Overview.bs
```diff
@@ -77,6 +77,53 @@ tho user agents may define additional undocumented [=environment variables=]:
 
 * Issue: define list of predefined vars
 
+Safe area inset variables {#safe-area-insets}
+------------------------------------------------------------------
+
+<pre class=propdef>
```
dbaron: I think defining these using `propdef` entries doesn't seem right.  The only aspects of the propdef tables that actually seem to apply here are "Name", "Value", and maybe "Media".  Furthermore, using propdef entries will put them in property indexes, which will be very confusing.

I think in the absence of a better mechanism this should probably be something more like a hand-written table, and if a better mechanism is needed it can be developed later.

css-env-1/Overview.bs
```diff
@@ -77,6 +77,53 @@ tho user agents may define additional undocumented [=environment variables=]:
...
+	Value: <<length>>
+	Initial: 0
+	Inherited: yes
+	Media: visual
+	Animatable: no
+</pre>
+
+The safe area insets are four [=environment variables=] that define the top, right, bottom and
+left insets from the edge of the viewport. Page content that is inside these insets may be
+obscured by the shape of the display.
```
dbaron: I think this description could be clearer about what these insets are for to readers not already familiar with the concept, and should probably also have stronger implementation conformance criteria.  I'd suggest rewording to something like the following:

> The safe area insets are four [=environment variables=] that define a rectangle by its top, right, bottom, and left insets from the edge of the viewport.  For rectangular displays, these must all be zero, but for nonrectangular displays they must form a rectangle, chosen by the user agent, such that all content inside the rectangle is visible, and such that reducing any of the insets would cause some content inside of the rectangle to be invisible due to the nonrectangular nature of the display.  This allows authors to limit the layout of essential content to the space inside of the safe area rectangle.

rebeccahughes: Fixed

css-env-1/Overview.bs
rebeccahughes: Fixed

dbaron:
Thanks.  This looks good to me.

Since you kept the "Media" entry in the table, I think the one thing that's still remaining is defining what happens when the medium in question doesn't match.  That's beyond the definition of the safe area inset variables.  I suspect a decent answer would be that the variable isn't present.  But it should probably be defined; I filed #2819 to follow up on that.

rebeccahughes:
Thanks!

astearns:
@rebeccahughes do you work at Google?

rebeccahughes:
@astearns Yes I do

PR #2836, opened 2018-06-27
[css-round-display] Refer to Backgrounds & Borders

ewilligers:
The CSS Borders spec (CSS3-BORDER) is obsolete.
We now have CSS Backgrounds and Borders (CSS3BG).


jihyerish:
Thanks for this update, Eric!

PR #2865, opened 2018-07-03
Mark some IDL bits as examples.

### LINKED ISSUE #2486: [cssom] "partial interface CSSStyleDeclaration {   [CERe..."
The following snippets are not specification IDL, but example output, so should be flagged as `note` (or `example`) to be ignored by IDL scrapers.

https://drafts.csswg.org/cssom/#example-5b8953f5

> partial interface CSSStyleDeclaration \{
>   \[CEReactions\] attribute \[TreatNullAs=​EmptyString\] CSSOMString \_camel\_cased\_attribute;
> \};

> partial interface CSSStyleDeclaration {
>   [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _webkit_cased_attribute;
> };

> partial interface CSSStyleDeclaration {
>   [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _dashed_attribute;
> };

emilio:
Since they don't define attributes themselves, but a class of them.

Fixes #2486

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2251,7 +2251,7 @@ the following partial interface applies where <var>camel-cased attribute</var>
 is obtained by running the <a>CSS property to IDL attribute</a> algorithm for
 <var>property</var>.
 
-<pre class="idl extract">
+<pre class="idl example">
 partial interface CSSStyleDeclaration {
   [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _<var>camel_cased_attribute</var>;
 };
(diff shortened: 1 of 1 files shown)


tabatkins:
The "extract" class is used by WHATWG specs (and now recognized by Bikeshed) as a non-normative class already. (I don't know why...)

emilio:
It's not clear to me what the implications here are, @zcorpan should we just close #2486 with no change then?

svgeesus:
So, should this be merged or should it be closed without change?

svgeesus:
@zcorpan @emilio

zcorpan:
Yeah. The purpose is to exclude it from the IDL index, which it does.

PR #2897, opened 2018-07-09
Explain :focus-visible intent, add examples and UA guidance

alice:
@tabatkins Following up on #2481, trying to more clearly explain the intent of this pseudo-class and provide better author and UA guidance.

Thanks to @patrickhlauke and @AmeliaBR for inspiration for the second example (not sure how to credit you in the spec).

cc @robdodson @bkardell

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -2240,12 +2240,11 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
 	<em>and</em> the user agent determines via heuristics
 	that the focus should be made evident on the element.
 
-	<div class=note>
-	The intent of '':focus-visible'' is 
+	
+	Note: The intent of '':focus-visible'' is 
 	to allow authors to provide <em>clearly identifiable</em> focus styles
 	which are visible when a user is likely to need to understand where focus is,
 	and not visible in other cases.
-	</div>
 
 	<div class="example">
 		In this example, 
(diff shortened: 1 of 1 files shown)


AmeliaBR:
I'm concerned about the phrase "element which does <em>not</em> support user input".

Technically, every focusable element supports user input. That input may only be activation, though. Maybe make it "text input" explicitly?

selectors-4/Overview.bs
```diff
@@ -2237,62 +2237,84 @@ The Focus-Indicated Pseudo-class: '':focus-visible''</h3>
...
+		should <strong>always</strong> match '':focus-visible'' when focused.</li>
+		<li>If the user interacts with the page via the keyboard,
+		the currently focused element should match '':focus-visible''
+		(i.e. keyboard usage may change whether this pseudo-class matches
+		even if it doesn't affect '':focus''). 
+		<li>If the user interacts with the page via a pointing device,
+		such that the focus is moved to a new element 
+		which does <em>not</em> support user input,
+		the newly focused element 
+		should <strong>not</strong> not match '':focus-visible''.</li>
```
robdodson: "not not" match?

alice:
@AmeliaBR Yeah, I hesitated over that text, but I wasn't sure about "text input" because it's also numeric, date/time etc so I hedged. 

How about this?

> (...) keyboard input (for example, any element which may trigger any type of virtual keyboard to be shown if a physical keyboard is not available) (...)

alice:
(Added a bullet point regarding user preferences at the top of the heuristics list.)

bkardell:
Yeah, we discussed this many times... it is wonky to define in terms we have - even 'virtual keyboard' is potentially a little confusing if you think of it as the qwerty thing, really it's pretty much any means of inputting text of some kind into some kind of control. I think the text here is pretty good

dbaron:
@tabatkins @fantasai curious about the state of this PR.  (See link to TAG issue above; we keep pushing it forward on the TAG agenda.)

robdodson:
friendly ping, @tabatkins or @fantasai do y'all have any thoughts?

slightlyoff:
Per https://github.com/w3ctag/design-reviews/issues/233, consider this a friendly ping @tabatkins

tabatkins:
Just did some very minor cleanup of markup, and tweaked some 2119 term usage. Otherwise this looks great, thanks so much @alice!

FremyCompany:
Travis Leitheid from Microsoft realized there were a couple of typos in the code example, so I fixed them in https://github.com/w3c/csswg-drafts/commit/afbffa43d4dd84537a130075d916b712e5f5f36f.

PR #2922, opened 2018-07-13
[css-logical-1] Add some concepts for logical properties for other specs to use

upsuper:
Specifically I'd like to reference them in cssom spec.

### DIFF
--- a/css-logical-1/Overview.bs
+++ b/css-logical-1/Overview.bs
@@ -220,6 +220,22 @@ Flow-Relative Box Model Properties</h2>
   depends on the element’s computed values of
   'writing-mode', 'direction', and 'text-orientation'.
 
+  Each set of flow-relative longhand properties and their parallel
+  physical properties form a <dfn export>logical property group</dfn>.
+
+  Each property in a <a>logical property group</a> has
+  a <dfn export>logical kind</dfn>. Flow-relative properties'
+  <a>logical kind</a> is "flow-relative", and physical properties'
+  <a>logical kind</a> is "physical".
+
+  Note: For example, 'block-size', 'inline-size', 'width', and 'height'
+  form a <a>logical property group</a>, while 'margin-block-start',
+  'margin-block-end', 'margin-inline-start', 'margin-inline-end',
+  'margin-top', 'margin-right', 'margin-bottom', and 'margin-left'
+  form another.
+
+  Note: Each property is in at most one <a>logical property group</a>.
+
   Note: Depending on the element’s own writing mode
   simplifies the cascading calculations
   and gives a straightforward model for authors to reason about.


css-logical-1/Overview.bs
```diff
@@ -220,6 +220,12 @@ Flow-Relative Box Model Properties</h2>
   depends on the element’s computed values of
   'writing-mode', 'direction', and 'text-orientation'.
 
+  Each set of flow-relative properties and their parallel physical
+  properties are in the same <dfn export>logical property group</dfn>.
```
zcorpan: It's not clear to me which these groups are. Can it be made more explicit?

css-logical-1/Overview.bs
```diff
@@ -220,6 +220,22 @@ Flow-Relative Box Model Properties</h2>
...
+  Each property in a <a>logical property group</a> has
+  a <dfn export>logical kind</dfn>. Flow-relative properties'
+  <a>logical kind</a> is "flow-relative", and physical properties'
+  <a>logical kind</a> is "physical".
+
+  Note: For example, 'block-size', 'inline-size', 'width', and 'height'
+  form a <a>logical property group</a>, while 'margin-block-start',
+  'margin-block-end', 'margin-inline-start', 'margin-inline-end',
+  'margin-top', 'margin-right', 'margin-bottom', and 'margin-left'
+  form another.
```
Loirooriol: What about shorthand properties like `margin-block`?

upsuper: Regarding the proposed use in #2924, it doesn't matter. Shorthand properties are expanded to longhands during parsing.

Loirooriol: Yes, but this should be properly defined independently of #2924. "Each set of flow-relative properties" seems to include shorthands, but they are not in the example. You can say "Each set of flow-relative longhand properties"

upsuper: Changed that according to your suggestion. Thanks.

FremyCompany:
FWIW I just reviewed this, and this looks fine to me

fantasai:
OK, so, I think this needs some revision. For example, an example shouldn't be marked as a note, it should be marked as an example. :) But it's easier for me to just merge it and edit it than to ask for fiddly wording changes that are annoying to pass back and forth so gonna just do that...

upsuper:
Thanks!

fantasai:
Notes on changes made...
* Format source per http://rhodesmill.org/brandon/2012/one-sentence-per-line/
* Moved definition of “logical property group” to the first paragraph, it fits better there. Makes more sense to define what something is first, then how it behaves second.
* This paragraph
  > Each property in a <a>logical property group</a> has  a <dfn export>logical kind</dfn>. Flow-relative properties' <a>logical kind</a> is "flow-relative", and physical properties' <a>logical kind</a> is "physical".”

  is just obnoxiously redundant for humans (A is A, and B is B), so I took a different approach to defining the term. Also renamed the term, since “kind” is a mostly-meaningless word, so I prefer not to use it if I can find something better.

PR #2924, opened 2018-07-13
[cssom-1] Replace steps of set a CSS declaration with some constraints

upsuper:
(This depends on concepts being added to css-logical in #2922.)

This commit replaces the steps of "set a CSS declaration" with some constraints, so that user agents are allowed to sort the declarations in whatever way they want as far as those constraints hold.

It's not clear to me what is the right wording for this.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -118,6 +118,8 @@ spec:html; type:dfn; text:ascii case-insensitive
 spec:html; type:dfn; text:case-sensitive
 spec:infra; type:dfn; text:list
 spec:css-logical; type:property; text:inline-size
+spec:css-logical; type:dfn; text:logical property group
+spec:css-logical; type:dfn; text:logical kind
 spec:css-variables-1; type:dfn; text:custom property
 spec:selectors-3; type:selector; text:::before
 spec:selectors-3; type:selector; text:::after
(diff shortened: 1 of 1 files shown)


upsuper:
This would definitely need a working group resolution.

cssom-1/Overview.bs
```diff
@@ -2190,28 +2192,51 @@ The <dfn method for=CSSStyleDeclaration>setProperty(<var>property</var>, <var>va
...
+  or removed from <var>declarations</var>.
+
+ <li>
+  If there are <a>CSS declaration</a>s in <var>declarations</var> whose
+  <a for="CSS declaration">property name</a> is in the same <a>logical property group</a> as
+  <var>property</var>, but has a different <a>logical kind</a>, <var>target declaration</var> must be
+  at an index after all of those <a>CSS declaration</a>s.
+
+ <li>
+  The steps must return true if the serialization of <var>declarations</var> is changed as a result of
```
zcorpan: Word this as "If the serialization of declarations was changed as result of these steps, then return true. Otherwise, return false."

cssom-1/Overview.bs
```diff
@@ -2190,28 +2192,51 @@ The <dfn method for=CSSStyleDeclaration>setProperty(<var>property</var>, <var>va
...
+  being <var>component value list</var>. The <a for="CSS declaration">important flag</a> of this
+  <a>CSS declaration</a> must be <a>set</a> if <i>important</i> flag is set, and <a>unset</a> otherwise.
+  Such declaration is referenced as the <var>target declaration</var> below.
+
+ <li>
+  Any <a>CSS declaration</a> which is not the <var>target declaration</var> must not be changed, inserted,
+  or removed from <var>declarations</var>.
+
+ <li>
+  If there are <a>CSS declaration</a>s in <var>declarations</var> whose
```
zcorpan: Move the `s` inside the `<a>`

cssom-1/Overview.bs
upsuper: This wording "If ... then ... otherwise" sounds like part of steps, but this should be a constraint I think.

upsuper:
To adopt this change, we need to remove two existing wpt and introduce a looser wpt. I'm considering doing this in [Gecko bug 1473180](https://bugzilla.mozilla.org/show_bug.cgi?id=1473180). It's not clear to me whether I can change the corresponding wpt without a resolution from the working group.

emilio:
@upsuper can you open an issue sketching the problem and the proposal and put the Agenda+ label?

I'll attend this week's call and I can try getting a resolution there.

upsuper:
I don't think we need a separate issue for that. I can put the problem and add Agenda+ on this PR directly.

So in #1898 we resolved (again) that seting property should always append the new declaration to the end of the declaration block, which was added to spec in #2516. However, this change has a performance impact, because browsers would be required to trigger mutation observer, and flush style. (To be honest, flushing style may be avoidable, but it could be complicated to fix given how things are working nowadays.)

Because of this, [the change to Firefox has been put behind a pref and off by default](https://bugzilla.mozilla.org/show_bug.cgi?id=1461285), and [the change to Chrome has been backed out](https://bugs.chromium.org/p/chromium/issues/detail?id=845944#c35).

In the end, in #1898 what we really want is that physical properties and flow-relative properties can override each other correctly. There are possible algorithms to realize that without sacrificing performance in usual cases, but such algorithms may be non-trivial and it could become tricky to spec and have everyone agree on one.

Thus, instead I propose that we just put down some basic requirements for this algorithm and allow implementations to explore. There would be observable behavior difference, but I don't think that would be too bad. If authors have been relying on the order, the original change itself would have been a breaking change.

FremyCompany:
👍

gsnedders:
> Because of this, the change to Firefox has been put behind a pref and off by default, and the change to Chrome has been backed out.

So what's the current behaviour? How interoperable are they? What would it take to converge behaviour in some other way to what we previously resolved in #1898?

upsuper:
> Feel free to merge this, probably after #2922 lands as well.

I think we should have that landed first.

> So what's the current behaviour? How interoperable are they?

Current behavior is to always update existing declaration with the same property name. I believe at least Blink and Gecko both do this, likely WebKit as well. Edge may have some other idea.

Blink (and probably WebKit as well) doesn't trigger mutation observer if the value and flag aren't changed, while Gecko did. That caused performance issue for us and thus we fixed that in 62 or 63.

> What would it take to converge behaviour in some other way to what we previously resolved in #1898?

I don't think we are going to use the algorithm we previously resolved in #1898 because that has webcompat issue.

We may eventually converge to some algorithm, but it's probably not going to happen very soon, as a conformant algorithm without webcompat issue is going to be somewhat complicated, and getting everyone agree on every details could take quite a bit of time.

Also there are other stuff currently in my mind that we should probably add to this constraints which may make things more complicated...

upsuper:
BTW, the corresponding web-platform test is going to be landed in web-platform-tests/wpt#12065.

cssom-1/Overview.bs
```diff
@@ -2190,28 +2192,101 @@ The <dfn method for=CSSStyleDeclaration>setProperty(<var>property</var>, <var>va
...
+
+ <li>
+  The steps must return true if the serialization of <var>declarations</var> was changed as result
+  of the steps. It may return false otherwise.
+</ul>
+
+Issue: Should we add something like "Any observable side effect must not be made outside
+<var>declarations</var>"? The current constraints sound like a hole for undefined behavior.
+
+Note: The steps of <a>set a CSS declaration</a> are not defined in this level of CSSOM. The user agent may
```
FremyCompany: nit: plural: `user agents may use different algorithms as long as the contraints above hold`

Issue #2735, opened 2018-06-04
[mediaqueries-5] Dark Mode in CSS


jonathantneal:
Could “Dark Mode” be something we leverage in CSS? Android has it. [Windows](http://bgr.com/2018/05/09/windows-10-dark-mode-file-explorer-how-to/) and [MacOS](https://www.engadget.com/2018/06/04/macos-mojave-includes-dark-mode/) are getting it. We already have system-ui fonts and (deprecated) system colors.

Related to #1286 but not necessarily the same as this isn’t about _high-contrast_ as much as it is about ”_dark on light_” (Light Mode) vs ”_light on dark_” (Dark Mode).

If apps can leverage this, it would be swell if webapps could leverage this.

eeeps:
Maybe a good candidate for an [environment variable](https://drafts.csswg.org/css-env-1/#environment)?

tabatkins:
No, this is definitely an MQ. env() is for piping *values* into CSS that you can then use in various properties. This has nothing to do with that; it's about letting the page know that a particular preference exists, so it can respond in arbitrary ways.

FremyCompany:
(Generally supportive of the idea)

WebReflection:
if anyhow useful, the Linux GNOME Adwaita Dark theme has been around for 4+ years and it looks stunning: https://github.com/GNOME/gtk/tree/master/gtk/theme/Adwaita

GNOME Dark mode is the default for pretty much every GNOME user I know (including me) so in case you need a reference for implementation, being Adwaita theme OSS maybe it could help speeding up things.

Spoiler: it looks just like macOS dark theme ...

WebReflection:
P.S. out of experience, browsers blindly applying all CSS had issues with pretty much every website assuming inputs have white background and text on those is black ... without special form elements, links, and others pre-styled overrides, there might be undesired consequences.

TL;DR I'd love to have a dark theme default on the Web, but it'll take years from now to be widely available. UI is brand for most browsers (companies behind) too.

jonjohnjohnson:
@WebReflection Guessing this issue isn't about browsers blindly applying all CSS, as much as it's about offering a media feature that allows a website access to a users "dark mode" system settings to then style accordingly.

As an aside, has anyone ever found themselves wanting to style around "night shift" features? Not just ambient light, but a users explicit setting, even while it affects a whole screens output?

jonathantneal:
> ... we are considering a CSS media query.
[...]

PR #2930, opened 2018-07-16
[mediaqueries-5] Add initial versions of prefers-contrast and color-scheme

grorg:


grorg:
I've now linked my accounts.

grorg:
@frivoal since i think you were the one most vocally suggesting `forced-foo` variants, I wanted you to take a look before I committed this.

If I don't hear from you in a few days, I'll land it anyway and we can take it up as an issue.

AmeliaBR:
Linking to the issue with all the discussion: https://github.com/w3c/csswg-drafts/issues/2735

(@grorg it would have been helpful if you'd done that at the time.)

PR #2935, opened 2018-07-18
Define overflow directions in terms of block-end and inline-end.

### LINKED ISSUE #1569: [cssom-view] Doubts regarding the overflow directions description
This issue would be a very simple problem about [the definition of overflow directions](https://drafts.csswg.org/cssom-view/#overflow-directions).

According to the combination of a block flow direction and an inline base direction, the overflow direction can be determined. For example, if the block flow direction is top-to-bottom and the inline base direction is left-to-right, the overflow direction is rightward and downward. On the basis of [`CSS Writing Mode`](https://drafts.csswg.org/css-writing-modes-4/#text-flow), [`CSSOM View`](https://drafts.csswg.org/cssom-view) describes the combination and the overflow direction accordingly, but I think some cases would be wrong.

In the [`2. terminology`](https://drafts.csswg.org/cssom-view/#terminology) chapter in `CSSOM View`, it specifies that if the block flow direction is left-to-right and the inline base direction is left-to-right, the overflow directions is `rightward and downward`. But as far as my understanding, the overflow directions should be modified to `rightward and upward` in that case. Besides, there would be some errors like this case near the section.

cc - @zcorpan, @frivoal

dbaron:
Fixes #1569.

(See #1569 for discussion.)

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -139,24 +139,7 @@ An element <var>body</var> (which will be <a>the HTML <code>body</code> element<
 Note: A <{body}> element that is <a>potentially scrollable</a> might not have a <a>scrolling box</a>.
 For instance, it could have a used value of 'overflow' being ''overflow/auto'' but not have its content overflowing its content area.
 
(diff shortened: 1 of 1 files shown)


dbaron:
While I'd still be interested in comments from @fantasai, I've gone ahead and merged this rather than leave the spec in its current state.

PR #2961, opened 2018-07-28
[css-ui-4] Deal with locked editable elements and user-select #285

### LINKED ISSUE #285: [css-ui] Readonly form control and user-select
Currently, for user-select, CSS UI 4 says
> on editable elements where the computed value is always `contain` regardless of the specified value

and it defines "editable elements" as
> an editable element is either an editing host or a mutable form control with textual content, such as textarea.

It seems this would exclude readonly form controls from the computation rule above, because according to the HTML Standard, readonly form controls are not mutable. But I think that rule should be applied to those controls as well.

frivoal:
Closes #285

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -1542,6 +1542,11 @@ except:
 		regardless of the specified value
 
 	<li>
+		on <a>locked editable element</a>s
+		where the computed value is ''user-select/contain''
+		if the specified value was ''user-select/text'' or ''user-select/auto''
+
+	<li>
 		when the specified value is ''user-select/auto'',
 		which computes one of the other values as defined below
 </ol>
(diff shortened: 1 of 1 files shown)


css-ui-4/Overview.bs
```diff
@@ -1541,6 +1541,11 @@ except:
 		where the computed value is always ''user-select/contain''
 		regardless of the specified value
 
+	<li>
+		on <a>locked editable element</a>s
+		where the computed value is ''user-select/contain''
+		if the specified value was ''user-select/text'' or ''user-select/auto''
```
upsuper: As far as the handling of `auto` for this case is mentioned below, it can probably be removed here.

css-ui-4/Overview.bs
```diff
@@ -1551,6 +1556,10 @@ an <dfn>editable element</dfn> is either
 an <a href="https://w3c.github.io/editing/contentEditable.html#dfn-editing-host">editing host</a>
 or a <a href="https://www.w3.org/TR/html/sec-forms.html#mutable">mutable</a> form control with textual content,
 such as <{textarea}>.
+A <dfn>locked editable element</dfn> is
+a normally <a href="https://www.w3.org/TR/html/sec-forms.html#mutable">mutable</a> form control with textual content,
+when it is placed in a non mutable state,
+such as a <{textarea}> with the <a href="https://html.spec.whatwg.org/multipage/input.html#the-readonly-attribute"><code>readonly</code></a> attribute.
```
upsuper: It seems to me that this definition is probably not clear enough, for example, it is unclear whether a disabled form control is a "locked editable element". I guess the ambiguity is from "normally mutable form control". Is a disabled form control considered normally mutable? Or is being disabled just a non-mutable state?

Maybe explicitly referencing `readonly` attribute rather than listing it as an example may help? Something like "a otherwise mutable form control ... when it is placed in a non-mutable state by readonly attribute".

alshamiri5 (approved):

Issue #1963, opened 2017-11-09
[css-scrollbars][css-scrollbars-1] 'overflow' does not apply to all FC roots


Loirooriol:
From https://drafts.csswg.org/css-scrollbars/#scrollbar-color-properties,

> Applies to: block containers and boxes that establish a formatting context [same as elements that [overflow](https://drafts.csswg.org/css-overflow-3/#propdef-overflow) applies to]

However, they are not the same. `overflow` does not apply to

 - Inline formatting context roots. `overflow` makes the block container establish a BFC instead, and an anonymous IFC root without `overflow` is generated.
 - Ruby formatting context roots. They are inline-like boxes, so `overflow` does not apply.
 - Table formatting context roots. `overflow` is used on the table wrapper box instead of on the table box.

So I would just say

> Applies to: [same as elements that [overflow](https://drafts.csswg.org/css-overflow-3/#propdef-overflow) applies to]

PR #2969, opened 2018-07-30
[css-scrollbars-1] Applies to boxes to which 'overflow' applies... which isn't quite FC roots. #1963

fantasai:


upsuper:
Although I'm not the editor... it seems to be the right thing to do.

PR #2979, opened 2018-08-01
removing accidental markup

dret:


tabatkins:
weird, thanks for the catch!

PR #2992, opened 2018-08-06
[css-scrollbars-1] Loosen requirement for scrollbar-width:thin

upsuper:
I think the main point of having thin scrollbars is to give content more room when the space is already tight, so it is probably better using the concept of "scrollbar gutter" for this value, which is also clearer when overlay scrollbar / auto-expand scrollbar involves.

Also when the scrollbar is already fully overlay (e.g. on macOS when trackpad is used, or on Android), I believe there is no point to create an even thinner scrollbar, so the value probably shouldn't use a "must".


upsuper:
Updated, and remove the gutter part. It now just mentions thinner scrollbar.

PR #3036, opened 2018-08-19
[css-logical] border radii not shorthands

### LINKED ISSUE #3035: [css-logical] Border radii are not shorthands
https://drafts.csswg.org/css-logical/#border-radius-shorthands defines "Flow-relative Border-radius Shorthands".

Not sure if I'm missing something, but I think these are longhands instead of shorthands.

ewilligers:
The Flow-relative Border Radius properties are longhands, not
shorthands.

resolves #3035


css-logical-1/Overview.bs
```diff
@@ -641,7 +641,7 @@ the 'border-block-start', 'border-block-end', 'border-inline-start', 'border-inl
   respectively.
 
 <h3 id="border-radius-shorthands">
```
Loirooriol: The ID should probably be fixed too

ewilligers: Done

PR #3053, opened 2018-08-25
[css-syntax-3] Missing character after Unicode name

Netmosfera:
Adds missing `(\)` after `U+005C REVERSE SOLIDUS`


dontcallmedom:

PR #3055, opened 2018-08-27
[css-align] flex-end without semicolon

ewilligers:
The production rules for self-position and content-position
previously ended with `| flex-end;`

The semicolon is not needed.


ewilligers:
@fantasai @tabatkins please review

PR #3090, opened 2018-09-10
Update refs to the box model

xfq:
Follow-up of #3078.

### DIFF
--- a/css-contain-1/Overview.bs
+++ b/css-contain-1/Overview.bs
@@ -23,6 +23,7 @@ spec: css2; urlPrefix: https://www.w3.org/TR/CSS2/
 	type: value; for: content; url: generate.html#value-def-close-quote; text: close-quote
 	type: value; for: content; url: generate.html#value-def-no-open-quote; text: no-open-quote
 	type: value; for: content; url: generate.html#value-def-no-close-quote; text: no-close-quote
+	type: dfn; url: box.html#padding-edge; text: padding edge
 
 spec: css-backgrounds-3;
 	type: dfn; url: https://drafts.csswg.org/css-backgrounds-3/#corner-clipping; text: corner clipping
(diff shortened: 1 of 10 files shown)


css-contain-1/Overview.bs
```diff
@@ -23,7 +23,6 @@ spec: css2; urlPrefix: https://www.w3.org/TR/CSS2/
 	type: value; for: content; url: generate.html#value-def-close-quote; text: close-quote
 	type: value; for: content; url: generate.html#value-def-no-open-quote; text: no-open-quote
 	type: value; for: content; url: generate.html#value-def-no-close-quote; text: no-close-quote
-	type: dfn; url: box.html#padding-edge; text: padding edge
```
frivoal: Actually, I think I'd rather keep this one linking to css2, because css-contain is in CR (to be updated soon). To avoid complications around publication request, I think it is best to avoid taking dependencies on a newish spec with non-CR status. Css-box only contains things that were in the CSS2.1 REC, so we could argue it away, but on the flip side, since it doesn't bring any thing new , then there's no rush in updating.

css-ui-3/Overview.bs
```diff
@@ -81,7 +81,6 @@ spec:css2; type:property; text:background-image
 <!-- Some of the following definitions have existing exported defintions from other specs,
 but not from specs that are OK to link to from a CR -->
 <pre class="anchors">
-url: https://www.w3.org/TR/CSS2/box.html#border-edge; type: dfn; spec: CSS2; text: border edge
```
frivoal: Same-ish comment as in css-contain: css-ui-3 is a REC, and if we need to publish an update, I'd rather not be tripped up by dependencies to a non REC, especially for no non-editorial gains.

xfq:
Fair enough, @frivoal. I pushed a commit to revert the changes to css-contain-1 & css-ui-3.

frivoal:
The changes to other specs look good to me, but I'm not the editor. I suggest you wait a bit more to see if someone has an issue, and otherwise merge.

One side comment: your commit also removes a number of trailing spaces at the end of lines. Removing them is good, but it would make reviews easier to do that in a separate commit or pull request in the future.

xfq:
> The changes to other specs look good to me, but I'm not the editor. I suggest you wait a bit more to see if someone has an issue, and otherwise merge.

Will do. Thank you!

> One side comment: your commit also removes a number of trailing spaces at the end of lines. Removing them is good, but it would make reviews easier to do that in a separate commit or pull request in the future.

Agreed.

(For the reviewers: you can add `?w=1` after the URL to hide whitespace changes. Sorry for the inconvenience.)

PR #3113, opened 2018-09-15
[css-scrollbar] Add text for contrast concern of scrollbar colors

### LINKED ISSUE #2968: [css-scrollbars-1] Clarify contrast expectations
What happens if an author specifies `scrollbar-face-color: black; scrollbar-track-color: black`? Is that expected to do something reasonable or unreasonable? Does the UA perform fixup? Is the scrollbar expected to have borders or other aspects which would provide sufficient contrast? Is it expected to be a WCAG violation?

upsuper:
This intends to fix #2968.


upsuper:
Updated the link per review comment.

PR #3129, opened 2018-09-20
[css-fonts-4] Stop claiming that 'ch' uses the first available font, since it doesn't

dbaron:
css-fonts-4 claims that the 'ch' unit uses the "first available font", but [it doesn't](https://drafts.csswg.org/css-values-4/#ch).  This revises the spec to stop making this claim.

Note that the same claim is made in css-fonts-3.  Given its status, I'm not sure whether to suggest the same edit there...

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -1012,7 +1012,7 @@ Relative sizing: the 'font-size-adjust' property</h3>
 	The value of 'font-size-adjust' affects the used value of 'font-size'
 	but does not affect the computed value.
 	It affects the size of relative units
-	that are based on font metrics of the <a>first available font</a>
+	that are based on font metrics
 	such as <code>ex</code> and <code>ch</code>
 	but does not affect the size of <code>em</code> units.
 	Since numeric values of 'line-height'
@@ -3175,7 +3175,7 @@ to ensure that the results are as consistent as possible across user agents,
 given an identical set of available fonts and rendering technology.
 
 The <dfn export>first available font</dfn>,
-used for example in the definition of <a>font-relative lengths</a> such as ''ex'' and ''ch''
+used for example in the definition of <a>font-relative lengths</a> such as ''ex''
 or in the definition of the 'line-height' property
 is defined to be the first available font
 that would match the U+0020 (space) character


AmeliaBR:
Your "it doesn't" link leads to the following text:

> Equal to the used advance measure of the “0” (ZERO, U+0030) glyph in the font used to render it.

This text could use with some clarification, too.  It took me a moment to realize that "it" referred back to the "0", and not the element as a whole.  And since there might not be a "0" in the element's text content, maybe a better wording would be (borrowing from the definition of "first available font"):

> Equal to the used advance measure of the “0” (ZERO, U+0030) glyph in the first available font that would match that character, given font families in the ‘font-family’ list (or a user agent's default font if none are available).

(and similar wording for the `ic` unit).

litherum:
It’s pretty unfortunate that ch units can cause fonts to download.

WebKit doesn’t do this; it just uses the primary font, and if the primary font doesn’t support “0” then it uses that font’s .notdef glyph. No downloads necessary.

https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/platform/graphics/Font.cpp#L123

https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/css/CSSPrimitiveValue.cpp#L661

emilio:
I'm not the most familiar with Gecko's font metrics code, but it seems that we use the first valid font as well:

  https://searchfox.org/mozilla-central/rev/0b8ed772d24605d7cb44c1af6d59e4ca023bd5f5/layout/style/ServoBindings.cpp#2557

And fall back to `aveCharWidth` if there's no `0` glyph, which is computed in a somewhat complicated way which I'm not confident enough to describe.

@jfkthame can probably correct if anything above is wrong.

jfkthame:
> It’s pretty unfortunate that ch units can cause fonts to download.

Can they?

It seems to me the definition of "first available font" is slightly unclear:

> The **_first available font_**, used for example in the definition of font-relative lengths such as 'ex' and 'ch' or in the definition of the 'line-height' property is defined to be the first available font that would match the U+0020 (space) character given font families in the 'font-family' list (or a user agent’s default font if none are available).

But this doesn't say what it means for a font to be "available". If the font-family list includes `@font-face` families for which resources have not (yet) been downloaded, are those fonts "available" for the purpose of computing font-relative lengths or not?

AFAICT, the spec usually uses "available" to refer to a font resource that has been downloaded and is ready for use, so I've assumed that was the intent here as well; but in §4.1 it seems to regard any font defined by a `@font-face` rule as being "available":

> A given set of `@font-face` rules define a set of fonts available for use within the documents that contain these rules.

without regard for whether the relevant resources have actually been fetched.

[1] https://drafts.csswg.org/css-fonts-4/#first-available-font

svgeesus:
> Note that the same claim is made in css-fonts-3. Given its status, I'm not sure whether to suggest the same edit there...

Once we have agreement, the same change can be listed as an erratum on Fonts 3.

svgeesus:
Hmm that would also call into question [this test](https://test.csswg.org/harness/test/css-fonts-4_dev/first-available-font-002/) currently passed by Gecko, Edge, Blink and even Trident.

dbaron:
So this is a PR to make a referencing spec agree with the spec that normatively defines the relevant behavior.  If folks disagree with the behavior, I think it would be good to open an issue against the latter spec (css-values).

dbaron:
Also, I don't think https://test.csswg.org/harness/test/css-fonts-4_dev/first-available-font-002/ really distinguishes the case, since it's a font that has *neither* a space nor a zero (it has only a lowercase a).  It would be more relevant to the question of how to define `ch` units to try variants of that test that have one but not the other of space and zero (probably both ways).

svgeesus:
I agree that V&U defines `ch` to use the _font that renders it_, not _the first available font_. Thus, I agree with this PR.

litherum:
> > It’s pretty unfortunate that ch units can cause fonts to download.

> Can they?

Right now, [Values and Units](https://drafts.csswg.org/css-values-4/#ch) says:

> Equal to the used advance measure of the “0” (ZERO, U+0030) glyph in the font used to render it.

Seems fairly unambiguous to me.

PR #3141, opened 2018-09-25
[css-transforms-2] Add definitions for transform list addition and accumulation

birtles:


css-transforms-2/Overview.bs
```diff
@@ -1106,6 +1106,65 @@ The transform functions <<matrix()>>, ''matrix3d()'' and ''perspective()'' get c
 
 For interpolations with the primitive ''rotate3d()'', the direction vectors of the transform functions get normalized first. If the normalized vectors are equal, the rotation angle gets interpolated numerically. Otherwise the transform functions get converted into 4x4 matrices first and interpolated as defined in section <a href="#matrix-interpolation">Interpolation of Matrices</a> afterwards.
 
+Addition and accumulation of transform lists {#combining-transform-lists}
+============================================
+
+<a lt="value addition">Addition</a> of transform lists is
+defined as <a spec=infra>list</a> concatenation
+such that <var ignore>V<sub>result</sub></var> is equal to
+<var ignore>V<sub>b</sub></var> <a spec=infra for=list>appended</a> to
```
tabatkins: Where do these `b` and `a`s come from? (The fact that you had to `ignore` them is a warning sign - they should be defined as inputs, *then* used. Wrap it in a `<div algorithm="transform list addition">` too, so their use is scoped to this algorithm and won't spuriously be counted as uses of those variables elsewhere.

css-transforms-2/Overview.bs
```diff
@@ -1106,6 +1106,65 @@ The transform functions <<matrix()>>, ''matrix3d()'' and ''perspective()'' get c
 
 For interpolations with the primitive ''rotate3d()'', the direction vectors of the transform functions get normalized first. If the normalized vectors are equal, the rotation angle gets interpolated numerically. Otherwise the transform functions get converted into 4x4 matrices first and interpolated as defined in section <a href="#matrix-interpolation">Interpolation of Matrices</a> afterwards.
 
+Addition and accumulation of transform lists {#combining-transform-lists}
+============================================
+
+<a lt="value addition">Addition</a> of transform lists is
+defined as <a spec=infra>list</a> concatenation
```
tabatkins: You'll be using these a lot, I recommend just putting these terms into `link-defaults` for the spec.

css-transforms-2/Overview.bs
tabatkins: `[=list/appended=]` is easier to read and write. ^_^

css-transforms-2/Overview.bs
```diff
@@ -1106,6 +1106,65 @@ The transform functions <<matrix()>>, ''matrix3d()'' and ''perspective()'' get c
...
+<a>identity transform function</a>
+(e.g. scale parameters and matrix elements <var ignore>m11</var>, <var
+ignore>m22</var>, <var ignore>m33</var>, and <var ignore>m44</var>),
+which combine using <dfn>accumulation for one-based values</dfn> as follows:
+
+  <var ignore>V<sub>result</sub></var> =
+  <var ignore>V<sub>a</sub></var> + <var ignore>V<sub>b</sub></var> - 1
+
+<div class="note">
+The above definition preserves the intent of <a lt="value
```
tabatkins: Please indent the contents of block elements. (I think the rest of the spec is inconsistent in this regard, but it's better to be correct here than consistent.)

css-transforms-2/Overview.bs
```diff
@@ -1106,6 +1106,65 @@ The transform functions <<matrix()>>, ''matrix3d()'' and ''perspective()'' get c
...
+However, instead of interpolating the individual parameters,
+they are combined using arithmetic addition--
+except in the case of parameters whose value is one in the
+<a>identity transform function</a>
+(e.g. scale parameters and matrix elements <var ignore>m11</var>, <var
+ignore>m22</var>, <var ignore>m33</var>, and <var ignore>m44</var>),
+which combine using <dfn>accumulation for one-based values</dfn> as follows:
+
+  <var ignore>V<sub>result</sub></var> =
+  <var ignore>V<sub>a</sub></var> + <var ignore>V<sub>b</sub></var> - 1
```
tabatkins: These two lines are separated and indented in the source, but all this will do in the output is make them inline in a new paragraph. Was that intended?

Also, again with using `<div algorithm>`, and defining the input vars to your algorithm and then using them; that way you won't need to `ignore` them all.

(and note that you can use `|V|` instead of `<var>V</var>`, to make the var usage less intrusive.)

css-transforms-2/Overview.bs
```diff
@@ -1106,6 +1106,65 @@ The transform functions <<matrix()>>, ''matrix3d()'' and ''perspective()'' get c
...
+and allows an animation such as:
+
+<pre class="lang-javascript">
+div.animate(
+  { transform: ['scale(1)', 'scale(2)'] },
+  {
+    duration: 1000,
+    easing: 'ease',
+  }
+);</pre>
```
tabatkins: Put your `</pre>` on its own line.

css-transforms-2/Overview.bs
```diff
@@ -1106,6 +1106,65 @@ The transform functions <<matrix()>>, ''matrix3d()'' and ''perspective()'' get c
...
+<pre class="lang-javascript">
+div.animate(
+  { transform: ['scale(1)', 'scale(2)'] },
+  {
+    duration: 1000,
+    easing: 'ease',
+    <strong>iterations: 5,
+    iterationComposite: 'accumulate',</strong>
+  }
+);</pre>
```
tabatkins: Here too.

css-transforms-2/Overview.bs
birtles: They come from here: https://drafts.csswg.org/css-values-4/#combining-values
How do I define them as inputs?

css-transforms-2/Overview.bs
birtles: I looked up the bikeshed docs, but I couldn't work out how to setup an alias with `link-defaults` (from "value addition" → "addition"). I've done it with `anchors` before, but not `link-defaults`.

birtles: Oh, this comment was referring to the `list` link -- I see! Sorry I assumed it was the first link.

css-transforms-2/Overview.bs
birtles: This too is referencing the notation from css-values-4. I guess we should work out how to link it somehow?

Unfortunately `|V<sub>result</sub>|` doesn't parse as a variable so we have to use `<var>V<sub>result</sub></var>`.

birtles:
I've addressed the review feedback I could, but I need guidance about how we should approach the variables. Does it makes sense to make them definitions in css-values-4? Or is there some other way to link to them there?

css-transforms-2/Overview.bs
tabatkins: Just in prose, like:

> Addition of two transform lists |a| and |b| is defined as...

I don't *yet* have definition ability for algorithm arguments, but for now making the inputs clear in prose will do the job.

css-transforms-2/Overview.bs
tabatkins: Ah, yes, markup will indeed disturb that. All the *other* variables can be done simpler. ^_^

birtles: Oh, I missed the other variables. I tried switching `m11` etc. to `|m11|` but then I got warnings about unreferenced variables. I found another reference to `<var ignore>m11</var>` in the doc and dropped the `ignore` attribute but it didn't fix it because I had scoped my `m11` to the "transform list accumulation" algorithm.

I wonder if there's some way to make the "Mathematical Description of Transform Functions" section define these variables? They're currently defined in an image there.

birtles:
Ok, everything should be addressed except the nicer markup for `|m11|` variables since I'm not sure if there's a good solution there (in effect those variables are defined by an image).

I also added a separate commit to fix some link errors / warnings in the existing file. I wasn't able to fix the link errors to SVG `<set>` and `<animate>` elements however since as far as I can tell neither SVG 1.1 not the separate [SVG Animations module](https://svgwg.org/specs/animations/) are currently indexed.

birtles:
@tabatkins review ping?

PR #3145, opened 2018-09-25
[css-text-3] clarify that U+3000 is not affected by word-break

### LINKED ISSUE #3142: [css-text-3] word-break: keep-all should not suppress break opportunities after Ideographic Spaces
The spec [defines `keep-all`](https://drafts.csswg.org/css-text-3/#valdef-word-break-keep-all) as:

> Breaking is forbidden within “words”: implicit soft wrap opportunities between typographic letter units (or other typographic character units belonging to the NU, AL, AI, or ID Unicode line breaking classes [UAX14]) are suppressed

U+3000 IDEOGRAPHIC SPACE is not included in any of the listed classes, so the break opportunity that is normally found after it should be preserved. However, it is not in Firefox.

While that is merely an implementation bug, it's a relatively easy mistake to make to assume (incorrectly) that IDEOGRAPHIC SPACE belongs to the class ID=IDEOGRAPHIC.

I don't think the normative prose needs any change, but a note (and a test) seem appropriate.

frivoal:
Closes #3142

### DIFF
--- a/css-text-3/Overview.bs
+++ b/css-text-3/Overview.bs
@@ -994,6 +994,10 @@ Line Breaking Details</h3>
         <p class=note>This is the other common behavior for Korean (which uses spaces between words),
         and is also useful for mixed-script text where CJK snippets are mixed
         into another language that uses spaces for separation.</dd>
+
+	Note: Despite its name, U+3000 IDEOGRAPHIC SPACE does not belong to the ID Unicode line breaking class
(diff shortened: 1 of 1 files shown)


frivoal:
Updated this pull request to try and have the same effect in much less text. A whole note seemed overkill, but given that there are implementation mistakes, a extra little bit of insisting to make sure the text is not misunderstood seems worth it.

PR #3180, opened 2018-09-28
[css-contain-1] Add bikeshed wpt info

frivoal:
Adding the bikeshed wpt info for tests in the wpt test suite.

@tabatkins I think I'm doing this right, but I don't understand if this is safe to merge. If I understand correctly, any test added to wpt that isn't referenced here will generate a fatal error, and block the spec from being auto-regenerated on http://drafts.csswg.org/, which seems bad.

Generating a warning is totally useful, but I fatal error seems a problem. Am I misunderstanding something?


frivoal:
hmmm. `<wpt-rest>` generates a fatal error if there aren't more tests, so that doesn't seem to help either.

tabatkins:
Fatal errors don't stop auto-publishing on csswg.org. You can see several specs with red status on <https://drafts.csswg.org/> ^_^

frivoal:
> According to current WPT, you're missing: [...]

Damn, these were added since yesterday!

> Fatal errors don't stop auto-publishing on csswg.org.

aha. I was running `bikeshed spec -f` and that didn't work, but that's because the correct invocation is `bikeshed -f spec`

PR #3198, opened 2018-10-08
[css-fonts][css-multicol][css-scroll-anchoring][css-scrollbars][css-shapes][css-size-adjust][css-transforms][css-ui] Update Computed Value and Animation lines to new format.

tabatkins:
Update a bunch of propdefs' Computed Value and Animation lines to the new format.

@litherum @frivoal @dirkschulze @rachelandrew @upsuper @astearns @dbaron Please review your respective specs; we'll merge when everyone gets a review in.

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -325,7 +325,7 @@ The 'shape-inside' Property</h4>
 		Applies To: block-level elements
 		Inherited: no
 		Computed Value: computed lengths for <<basic-shape>>, the absolute URI for <<uri>>, otherwise as specified</td>
-		Animatable: as specified for <<basic-shape>>, otherwise no
+		Animation type: as defined for <<basic-shape>>, otherwise no
 	</pre>
 
 	The values of this property have the following meanings:
@@ -459,7 +459,7 @@ The 'shape-padding' Property</h4>
 		Applies To: block-level elements
 		Inherited: no
 		Computed Value: the absolute length
-		Animatable: as <a href="https://www.w3.org/TR/css3-transitions/#animtype-lpcalc">length, percentage, or calc</a>
+		Animation type: by computed value
 	</pre>
 
 	<div class="example">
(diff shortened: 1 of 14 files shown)


css-ui-4/Overview.bs
```diff
@@ -1324,6 +1322,7 @@ Applies to: all enabled elements
 Inherited: no
 Percentages: N/A
 Computed value: as specified
+Animation type: not animatable
```
frivoal: I think this one should be discrete. If you're animating the relative position of things, it seems reasonable to change the `nav-*` mapping half way through the animation to reflect the new locations. Animating discretely from one keyword to another doesn't seem to have any particular implementation difficulty either (no box to tear up or down halfway through the animation or any such shenanigans), so I don't see any reason to make this not animatable.

frivoal:
Multicol 1 & 2, as well as css-ui 3 and 4 OK, except for 1 comment: https://github.com/w3c/csswg-drafts/pull/3198#discussion_r223525648

svgeesus:
I'm not requesting changes to the patch. I'm just saying that updating the files here doesn't update the Rec, and that the .src.html format is a pain in the ass.

(sent in a hurry while on vacation)

css-ui-4/Overview.bs
tabatkins: Agree, I don't see a reason to make this non-animatable.

css-shapes-2/Overview.bs
```diff
@@ -325,7 +325,7 @@ The 'shape-inside' Property</h4>
 		Applies To: block-level elements
 		Inherited: no
 		Computed Value: computed lengths for <<basic-shape>>, the absolute URI for <<uri>>, otherwise as specified</td>
-		Animatable: as specified for <<basic-shape>>, otherwise no
+		Animation type: as defined for <<basic-shape>>, otherwise no
```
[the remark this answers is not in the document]
fantasai: Yes.

frivoal:
@svgeesus , It's your turn to deal with fonts-3, as you mentioned in https://github.com/w3c/csswg-drafts/pull/3198#issuecomment-428503433

PR #3215, opened 2018-10-16
[css-transforms-1] Specify the "pad, interpolate prefix, then smoosh" interpolation behavior

### LINKED ISSUE #927: [css-transforms] Smarter interpolation of truncated transform lists
https://www.w3.org/mid/CALXxKfCKZd+P0mH2oFyDkBgps56WqwQnLSMpKk1nrqiFXScqVA@mail.gmail.com
https://www.w3.org/mid/CAAWBYDBPafEbiXNy=Mw4a9w1jmvK6dUJzq28PPjN7APgtjqAEw@mail.gmail.com

Need to decide what goes into Level 1, and what to call the new identify functions. Related to https://github.com/w3c/csswg-drafts/issues/898

birtles:
This fixes #927.

### DIFF
--- a/css-transforms-1/Overview.bs
+++ b/css-transforms-1/Overview.bs
@@ -1092,44 +1092,75 @@ The scaling causes a non-invertible CTM for the coordinate space of the div box.
 Interpolation of Transforms {#interpolation-of-transforms}
 ==========================================================
 
-When animating or transitioning transforms, the transform function lists must be interpolated. For interpolation between one transform <em>from-transform</em> and a second transforms <em>to-transform</em>, the rules described below are applied.
+[=Interpolation=] of transform function lists is performed as follows:
 
 <ul>
-  <li id="none-none-animation">
-    If both the <em>from-</em> and <em>to-transform</em> are ''transform/none'':
-      * There is no interpolation necessary. The computed value stays ''transform/none''.
+  <li id=none-none-interpolation>
+    If both <var>V<sub>a</sub></var> and <var>V<sub>b</sub></var>
+    are ''transform/none'':
+      * <var>V<sub>result</sub></var> is ''transform/none''.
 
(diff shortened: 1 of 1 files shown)


birtles:
Anyone care to review this?

We've already landed it in Firefox and upstreamed the tests to wpt.

birtles:
Review ping? @tabatkins?

dirkschulze:
@birtles I will review it on Monday. I’d like to look at it before it lands.

css-transforms-1/Overview.bs
```diff
@@ -1092,45 +1092,69 @@ The scaling causes a non-invertible CTM for the coordinate space of the div box.
...
+    <div class=example>
+        For example,
+        if <var>V<sub>a</sub></var> is ''scale(1) rotate(90deg) translate(20px)''
+        and <var>V<sub>b</sub></var> is ''scale(2) translate(10px) rotate(270deg)'',
+        the ''scale(1)'' and ''scale(2)'' functions will be interpolated
+        according to [[#interpolation-of-transform-functions]]
+        while the remainder of each list--
+        ''rotate(90deg) translate(20px)'' and ''translate(10px) rotate(270deg)''--
+        will first be converted to equivalent 4x4 matrices
+        and then interpolated as described in [[#matrix-interpolation]].
```
tabatkins: This example, if I'm thinking correctly, gives the same result whether you use whole-property matrix interpolation or this smarter interpolation.  I recommend providing an example that showcases the differences between the two; an obvious example would be having the common prefix be a `rotate(0deg)` and `rotate(360deg)`, which is hugely different.

Then you can also have a small note in this that the spec previously used whole-property matrix interpolation, which would have a different effect.

tabatkins:
Approved, but I have one suggestion for improving an example.

PR #3223, opened 2018-10-18
[css-syntax-3] (non substantive) add another missing <dd> tag

Netmosfera:
Found another occurrence. I've made sure it's the last one.


dontcallmedom:

PR #3252, opened 2018-10-26
[css-print] correct Bikeshed Shortname variable

Crissov:
It erroneously used `css-mobile` instead of `css-print`.


Crissov:

Issue #2905, opened 2018-07-11
[css-overflow-3] Allowing (or not) alternate ellipsis behavior for block-overflow


frivoal:
The spec says:
> If there is no next fragmentation container and thus the remainder of the content after the break would be discarded, then the UA may visually replace the contents of the line, as it does for text-overflow.

in https://github.com/w3c/csswg-drafts/issues/390#issuecomment-372143424, Mats Palmgren said:

> I think this "may" should be changed to a "must", or alternatively, the entire sentence be removed. I don't see any upsides to allowing UAs to have incompatible behavior for this common case.

The idea was that the behavior allowed by this "may" sentence was worse, and that we wanted browsers to do the full thing, but that it was more involved, and that doing it as a visual replacement was a tolerable approximation (only in the case of no subsequent fragmentainer).

If implementors are OK with going with the full thing, I'd be happy about deleting that.

frivoal:
Follow up from @fantasai 

> Your third option for placing the ellipsis seems pretty reasonable to me. When drafting up the spec we wanted to allow implementations to re-use as much of the existing text-overflow machinery as possible, hence the MAY -- as long as you're otherwise discarding content after the break, placing it effectively “underneath” the ellipsis for event handling is fine. It's only when the content needs to be pushed to another container that we need to make the ellipsis replace it layout-wise rather than just paint-wise.
> 
> In typical cases, the author will not be able to tell the difference between the two behaviors. But layout replacement can affect the height of the line, which makes it a more complicated mechanism to implement correctly (and as noted, can introduce cycles, which then need to be broken). I would rather not require that extra complexity for this case, where it is not needed and the behavior change in real layouts is slight. (At some point, as we expand CSS's ability to break flows between boxes, we may want to require layout replacement here, to make discarded flows consistent with continued ones; but since such a change seems unlikely to break pages, it seems safe to allow the easier thing now.)

frivoal:
in https://github.com/w3c/csswg-drafts/issues/390#issuecomment-372143424 @MatsPalmgren also said:

> >    The means of breaking any resulting cycles is up to the UA.
>
> Again, vague spec text like this will likely lead to incompatible behavior in different UAs. I'd prefer the spec to say something concrete about how to solve it and make that behavior mandatory.

kojiishi:
It's not clear what "visually replace the contents" means, but from "as it does for text-overflow", if it means UA to layout all the remainder of the content but clipped, I hope us not to require it.

[...]

PR #3263, opened 2018-10-29
[css-overflow-3] Inserting the block ellipsis should not cause relayouts

frivoal:
Treating it as ''line-height: 0'' should handle that.

This is slightly different from the resolution taken in #2905, but this should be equally safe with regards to relayout, while still allowing text alignment and especially justification to work fine.


frivoal:
reviewed by @fantasai

PR #3291, opened 2018-11-07
[css-transforms] [web-animations-1] csswg github

ewilligers:
Issues should be created in
https://github.com/w3c/csswg-drafts/issues
not
https://github.com/w3c/fxtf-drafts/issues

### DIFF
--- a/css-transforms-1/Overview.bs
+++ b/css-transforms-1/Overview.bs
@@ -10,7 +10,7 @@ Previous Version: https://www.w3.org/TR/2012/WD-css3-transforms-20120911/
 Previous Version: https://www.w3.org/TR/2012/WD-css3-transforms-20120403/
 Shortname: css-transforms
 Level: 1
-Group: fxtf
+Group: csswg
 Editor: Simon Fraser, Apple Inc http://www.apple.com/, simon.fraser@apple.com, w3cid 44066
 Editor: Dean Jackson, Apple Inc http://www.apple.com/, dino@apple.com, w3cid 42080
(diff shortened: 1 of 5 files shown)


birtles:
I think we need WG resolution on this. I got resolution to move the spec to csswg-drafts but not explicitly to move the spec to the CSSWG proper.

There was some discussion about this in the past:

  https://lists.w3.org/Archives/Public/www-style/2016Dec/0042.html

But I believe @fantasai might still have had some reservations about it.

tabatkins:
Moving the spec to the csswg-drafts repo definitely means issues should be in the csswg-drafts repo. You're drawing a distinction that doesn't exist.

I'll go ahead and merge.

dirkschulze:
@tabatkins the issues is that now the header of the spec changes. With the merge, the spec is no longer an FX spec (governed by SVG WG and CSS WG) But a pure CSS WG spec. The change should get reverted and maybe Bikeshed could allow a different repo on demand from a spec.

dirkschulze:
Oh, this now affects 3 specifications! Thought it was limited to 1 :(

birtles:
> @tabatkins the issues is that now the header of the spec changes. With the merge, the spec is no longer an FX spec (governed by SVG WG and CSS WG) But a pure CSS WG spec

Right, that was my concern. Last time we discussed doing that, some were not entirely on board with making that change.

tabatkins:
Is the SVGWG still even chartered? I thought that was the point.

dirkschulze:
@tabatkins The condemned live longer.

And yes, this change would require a WG resolution from both sides. I do not expect an approval from the SVG WG just yet. At least not on specifications that used to be part of SVG. (Web animations might be different.)

PR #3295, opened 2018-11-07
[css-sizing-4] Typo: s/as tall as it is high/as tall as it is wide/

Loirooriol:
"tall" and "high" refer to the same dimension so it doesn't really make sense.


Loirooriol:
@ewilligers Can I merge the PR now that I have your approval? Or do I need the approval of an editor of the spec or something else?

ewilligers:

Issue #2435, opened 2018-03-13
[css-animations] Serialize all keyframe names as strings


csnardi:
After #118 and #801, support for naming a keyframe the same as a reserved keyword by using a `<string>` was added to parsing and CSSOM setting. [The WPT test](https://github.com/w3c/web-platform-tests/blob/master/css/cssom/CSSKeyframesRule.html#L59-L69) for this part of the spec seems to assume that if a keyword could be parsed as an `<ident>`, then it would be serialized as a string (with quotation marks); otherwise, no quotation marks would be used. However, this seems needlessly complex; wouldn't it be easiest just to serialize all keyframe names as strings? That way, it would be ensured that re-parsing the output would generate an identical rule, without having to check if the name is an `<ident>`.

fantasai:
This is likely Web-incompatible, since Firefox and Blink currently return bare identifiers in getComputedStyle.
It's also inconsistent with the principle that CSS-internal user-chosen names use identifiers and strings are only used when interfacing with an external system.

I think what would make more sense is to exclude the CSS-wide keywords and 'none' from being animation names, as the `<custom-ident>` type recommends. We could still parse in strings as required by https://github.com/w3c/csswg-drafts/issues/118 but they would compute to `<custom-ident>` and any that match an excluded keyword would be invalid. This is effectively what we do for line names in CSS Grid, for example.

ewilligers:
Do we need a more precise resolution before updating the spec and tests? I gather from a [PR discussion](https://github.com/w3c/csswg-drafts/pull/3296) that we need to serialize as strings whenever we do not have a valid animation-name custom identifier (e.g. `'"Initial"'`, `'"None"'`, `'" x "'`). Also, do we want to ban the empty string `'""'` ?

yisibl:
Is there any progress here?

javifernandez:
Is the CSSWG resolution mentioned https://github.com/w3c/csswg-drafts/issues/2435#issuecomment-428646888 still valid ? 

I've seen that WebKit has [updated the implementation](https://github.com/WebKit/WebKit/pull/38484) to match the current spec, which is what Firefox already implements; Chrome diverges, though.

javifernandez:
@fantasai is the PR #12985 enough to cover the requested edits ? Id so, perhaps we can close this issue.

cdoublev:
I am confused. :worried:

Quoting the new text:

[...]

PR #3296, opened 2018-11-07
[css-animations] animation-name can contain string

ewilligers:
The computed value for animation-name can contain strings.

For example, the following are valid:
@keyframes "initial" { ... }
@keyframes "None" { ... }

The computed value for animation-name
  `"initial", "None", None`
should be
  `"initial", "None", none`

Without quotes, the computed value would not round trip.

### DIFF
--- a/css-animations-1/Overview.bs
+++ b/css-animations-1/Overview.bs
@@ -53,6 +53,7 @@ spec:cssom-1; type:interface; text:CSSRule
 spec:html; type:dfn; text:case-sensitive
 spec:css-backgrounds-3; type:property; text:background-image
 spec:css2; type:property; text:display
+spec:css-values-4; type:dfn; text:string
 </pre>
 
 <h2 id="intro">
@@ -472,7 +473,7 @@ The 'animation-name' property</h3>
 	Applies to: all elements
 	Inherited: no
 	Percentages: N/A
-	Computed value: list, each item either a case-sensitive <a>css identifier</a> or the keyword ''animation-name/none''
+	Computed value: list, each item either a case-sensitive <a>css identifier</a> or <a>string</a>, or the keyword ''animation-name/none''
 	Animation type: not animatable
 	Canonical order: per grammar
 	</pre>


ewilligers:
Proposed [WPT](https://github.com/web-platform-tests/wpt/pull/13970)

tabatkins:
Hm. I guess the question is; do we actually preserve this information in the underlying value? An alternative is to always make it an identifier, but serialize as a string when it's the handful of disallowed idents.

I don't have a strong opinion; I'm fine with matching implementations.

emilio:
See #2435

tabatkins:
Cool, so that suggests that we should keep the computed value as idents, and just specify that serialization produces strings in a few limited cases.

birtles:
@ericwilligers I think this PR is no longer correct. If I understand the resolution from #2435 correctly, we no longer want to allow `"initial"` as an animation-name, for example.

From what I understand, it sounds like the behavior we want is:

1. Accept `<string>` or `<custom-ident>` as we currently do.
2. If `<string>` is one of the CSS-wide keywords or `"default"` then it is invalid (at parse / variable-resolution time) as per the rules for custom-idents: https://www.w3.org/TR/css-values-3/#custom-idents
3. When serializing, if the value doesn't have [whitespace](https://www.w3.org/TR/css-syntax-3/#whitespace)(??) serialize without quotes (i.e. as `<custom-ident>`), otherwise serialize as `<string>`. I'm not sure what other cases require quotes or if there is machinery defining this somewhere else? (I looked in CSS Fonts 4 but didn't find anything there.)

ewilligers:
I missed #2435 
> RESOLVED: serialize all keyframes as identifiers

This implies we never serialize as strings, so `"1"` and `"  "` and `"initial"` and `"None"` can not be valid keyframe names. I suspect  `"  hello  "` should either not be a valid keyframe name, or it should serialize as `hello`.

This PR can be discarded.

birtles:
> This implies we never serialize as strings

I believe that resolution doesn't capture what was actually discussed. Or at least, when I re-read through the minutes my understanding was that we would still serialize as strings when needed. Tab's comment above also seems to suggest that.

birtles:
> 3\. When serializing, if the value doesn't have [whitespace](https://www.w3.org/TR/css-syntax-3/#whitespace)(??) serialize without quotes (i.e. as `<custom-ident>`), otherwise serialize as `<string>`. I'm not sure what other cases require quotes or if there is machinery defining this somewhere else? (I looked in CSS Fonts 4 but didn't find anything there.)

Maybe it would just be better to say that anything that can be serialized as a `<custom-ident>` is serialized as such, and anything else is serialized as a string.

birtles:
Cancelling review request as per my comments above, I believe this PR doesn't capture the changes discussed in the minutes.

ewilligers:
> Maybe it would just be better to say that anything that can be serialized as a `<custom-ident>` is serialized as such, and anything else is serialized as a string.

https://jsfiddle.net/ericwilligers/5na1Ldpb/ shows that Firefox computes `"Default"` `"Initial"` `"None"` as `Default` `Initial` `None`; two of these do not round trip and `Default` should not round trip either.

ewilligers:
Note also that Firefox preserves `'""'` as a specified value, but gives none as the computed value.

upsuper:
> Note also that Firefox preserves `'""'` as a specified value, but gives none as the computed value.

IIRC this is a known bug of Firefox.

yisibl:
> > Note also that Firefox preserves `'""'` as a specified value, but gives none as the computed value.
> 
> IIRC this is a known bug of Firefox.

https://bugzilla.mozilla.org/show_bug.cgi?id=1554737

PR #3304, opened 2018-11-08
[css-transitions-1] [css-animations-1] GlobalEventHandlers is a mixin

saschanaz:
...and thus should only be extended by `partial interface mixin`s.


saschanaz:

birtles:
Thank you!

PR #3313, opened 2018-11-13
[cssom] Return CSSStyleSheet from LinkStyle

### LINKED ISSUE #3311: [cssom] LinkStyle.sheet to return CSSStyleSheet instead of StyleSheet?
https://drafts.csswg.org/cssom/#the-linkstyle-interface

>The associated CSS style sheet of a node is the CSS style sheet in the list of document or shadow root CSS style sheets of which the owner node implements the LinkStyle interface.
>```webidl
>interface mixin LinkStyle {
>  readonly attribute StyleSheet? sheet;
>};
>```
>The sheet attribute must return the associated CSS style sheet for the node or null if there is no associated CSS style sheet.

The prose says `sheet` returns a CSS style sheet but the IDL says it returns a superclass instance (`StyleSheet`). Can we fix the IDL to return `CSSStyleSheet`?

saschanaz:
Fixes #3311

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1205,7 +1205,7 @@ interface.
 
 <pre class=idl>
 interface mixin LinkStyle {
-  readonly attribute StyleSheet? sheet;
+  readonly attribute CSSStyleSheet? sheet;
 };
 </pre>
 


saschanaz:

emilio:
Thanks!

PR #3316, opened 2018-11-14
[cssom] Return CSSStyleSheet instead of StyleSheet

saschanaz:
Resolves https://github.com/w3c/csswg-drafts/issues/3311#issuecomment-438402828.


saschanaz:

PR #3317, opened 2018-11-14
[selectors4] :is() has specificity

### LINKED ISSUE #3258: [selectors-4] Rename :matches() to :is()
Breaking this one out from https://github.com/w3c/csswg-drafts/issues/2143#issuecomment-433627117 ; @gibson042 wrote:

> @fantasai We may not live in an ideal world, but I think we do live in one that's close enough to introduce `:is()` and redefine `:matches()` as a deprecated alias of it. What do you think?

The benefits would be: It's much shorter to type, and it makes a clear pairing with :not(), which is its opposite.
The downsides would be: It's already shipping in WebKit as `:matches()`.

Here's the snippet of [prior CSSWG discussion](https://lists.w3.org/Archives/Public/www-style/2018Jul/0027.html), excerpted from [the naming of `:where()`](https://github.com/w3c/csswg-drafts/issues/2143#issuecomment-433627117):

```
  frremy: Why use the :matches() name at all, it's a bad name
  leaverou: We're stuck with it anyway
...
  leaverou: Would you object to :is() if it did what :matches()
            currently does, and :match() is the 0 specificity one?
  fantasai: No, I would not
...
  frremy: :matches() was called that way because the DOM API is called that
  fantasai: Actually, it's very old, original :matches() proposal was from hixie
            a long time ago, before :any()
  <dbaron> I implemented :-moz-any() in Gecko in 2010 in
           https://bugzilla.mozilla.org/show_bug.cgi?id=544834
  ericwilligers: It has been in webkit for a long time
  astearns: How much web content uses :matches?
  iank: Not a lot, looking at usecounter
  ericwilligers: usecounter may be wrong

  Rossen: Is either 2 or 7 an option? they involve changing :matches. webkit people,
          can we do that?
  myles: We could
  myles: We would be moderately interested in updating, but it's low priority
  dino: Why ?
  fantasai: Because a well-named proposal otherwise is :is(), but it is bad if
            it's specificity doesn't match :not()
  fantasai: So we could change :matches() to be the 0 specificity one,
            freeing :is() to do what :matches() does now
...
  [subsequently eliminated :matches() from the list of proposals for #2143,
   so this was not further discussed]
```

I believe other implementations besides WebKit are getting close to shipping `:matches()` so if we're renaming this, we need to do it asap.

ewilligers:
The functional pseudo-class like :where() but with specificity
is named :is() instead of :matches().
https://github.com/w3c/csswg-drafts/issues/3258#issuecomment-438742190

resolves #3258

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -133,7 +133,7 @@ Selectors Overview</h2>
 				<td>[[#negation]]
 				<td>3/4
 			<tr>
-				<td><code>E:matches(<var>s1</var>, <var>s2</var>, &hellip;)</code>
+				<td><code>E:is(<var>s1</var>, <var>s2</var>, &hellip;)</code>
 				<td>an E element that matches <a>compound selector</a> <var>s1</var>
 						and/or <a>compound selector</a> <var>s2</var>
 				<td>[[#matches]]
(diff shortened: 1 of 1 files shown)


fantasai:
Drat, didn't see the PR before pushing the changes. Thanks for the diffs, but I'm gonna have to close the PR!

PR #3324, opened 2018-11-15
Migrate spatial navigation

frivoal:
This imports the spatial navigation spec from WICG, including the whole git history, but removing irrelevant files.


frivoal:
When merging this, https://github.com/WICG/spatial-navigation/pull/136 should be merged as well.

frivoal:
Btw, the resolution to do this is over there: https://w3.logbot.info/css/20181023#c451857

Issue #3341, opened 2018-11-25
[cssom-1] Relationship between owner node and associated css style sheet


domfarolino:
The definition of [associated CSS style sheet](https://drafts.csswg.org/cssom/#associated-css-style-sheet) is a bit fuzzy to me, and I don't think captures the intended definition. Is it trying to say one of the following:

 - Only nodes that implement the LinkStyle interface can have "associated CSS style sheets", or...
 - More specifically, given a node that implements the LinkStyle interface, its associated CSS style sheet is the style sheet (from a larger list of sheets) whose owner node is the given node.

From this I'm trying to understand the relationship between a CSS style sheet's owner node and a some node's associated style sheet. Given how the spec is written, it seems that CSS style sheets can have owner nodes that either implement the LinkStyle interface or not, but only nodes that implement the LinkStyle interface have "associated CSS style sheets". Does this mean that a CSS style sheet can have an owner node that does not implement LinkStyle, and as such, does not have an associated CSS style sheet?

upsuper:
Given
> The associated CSS style sheet of a node is the CSS style sheet in the list of document or shadow root CSS style sheets of which the owner node implements the LinkStyle interface.

and
> **owner node**: Specified when created. The DOM node associated with the CSS style sheet or null if there is no associated DOM node. 

I don't think it is possible that a style sheet can have an owner node which doesn't implement `LinkStyle` as far as the two "associated"s refer to the same thing.

emilio:
I think #3343 clarifies this, but feel free to reopen otherwise. Thanks again!

PR #3343, opened 2018-11-25
Clarify associated sheet

domfarolino:
[cssom-1] Clarify the relationship between a CSS style sheet's owner node, and the owner node's associated CSS style sheet. See https://github.com/w3c/csswg-drafts/issues/3341.


domfarolino:

PR #3396, opened 2018-12-05
[css-transforms][web-animations] Change deliverers from FXTF to CSSWG

dirkschulze:
Change deliverers from FXTF to CSSWG. @birtles could you approve the web animations part please?


dirkschulze:
For reference from this PR, WG decision: https://lists.w3.org/Archives/Public/www-style/2016Dec/0042.html

Issue #3329, opened 2018-11-16
[resize-observer] Which box information should we pass to the callback


gregwhitworth:
We now allow authors to determine which box(es) they wish to observe on layout changes. Now we need to discuss which boxes we provide the information on. We have a few options:

Take the following example: If the author sets `content-box` and `border-box` to be observed, which box dimensions would you expect to receive passed to your callback?

1. The box(es) that were requested to be observed: so in the example above, `content-box` and `border-box` information
2. They receive all available CSS boxes that are observable: They will receive `border-box`, `content-box`, `scroll-box` dimensions.

**Option 1 Pros** is good because it makes sense that I requested a specific set of boxes to be observed and those are the ones that you returned information on.

**Option 1 Cons** If the sets the observer to the `border-box` but desires to know something about an inner box, they may request it in some other manner causing us to do additional work which defeats the purpose of this API altogether.

**Option 2 Pros** You can set one observation and get information about all of the boxes, so if you have a scenario in which you want to observe the border box but check the content or scroll box - this is possible without having to observe them.

**Option 2 Cons** It clutters up the response, if I only care about the `border-box` but I get results for every other box this may be a nuisance.

My personal opinion is to go with option 2 as I think people will either over observe to get the potential rects they care about or will ask for them in some other manner. That said, I don't have a strong opinion about this as I think it will be user preference. I have a [twitter poll](https://twitter.com/gregwhitworth/status/1062954297757904898) going about this in hopes of crowd sourcing some thoughts from web devs.

atotic:
My vote is also for #2, report all boxes. 

Observing content-box and border-box at the same time increases likelihood of "undelivered notifications error". We do not want developers observing multiple sizes just to be able to read their values out of the ResizeObserverEntry.

gregwhitworth:
> increases likelihood of "undelivered notifications error"

Then what is the reasoning behind allowing a sequence of options to be passed in? Maybe we should limit it to only one option being passed in as to avoid this issue?

atotic:
> Then what is the reasoning behind allowing a sequence of options to be passed in? Maybe we should limit it to only one option being passed in as to avoid this issue?

Good point. I think single option will be vast majority of usage. For the rare case of multiple sizes being observed, developers can use multiple ResizeObservers.

gregwhitworth:
> Good point. I think single option will be vast majority of usage. For the rare case of multiple sizes being observed, developers can use multiple ResizeObservers.

[...]

PR #3409, opened 2018-12-08
[resize-observer-1] Update algorithms with new boxes

atotic:
[resize-observer-1] Update algorithms with new boxes

A large pull request, algorithm redo touched everything.

The main work was modifying algorithms to work with multiple boxes. This work should be complete. I assumed RO is only observing 1 box at a time, as discussed in #3329.

Minor work included:
- fixing some of the spec bugs that were opened by domenic in the old repo.
- switching from contentRect to borderBoxSize in examples, no need to showcase old API.

One last large [remaining problem](https://github.com/WICG/ResizeObserver/issues/46) is ResizeObservation. It should not be an idl interface, but a bag of slots and algorithms. I'll do this next week. 

I am not very proficient with bikeshed, especially references and slots. There is probably a lot of room for cleanup by an expert.


atotic:
I've just connected my w3c account to github.

Issue #2644, opened 2018-05-06
[css-fonts] Clarify what ligatures are optional


upsuper:
Recently we got [a bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1458053) that when `text-justify: inter-character` is specified, some characters which should be composed are instead separated.

I did some investigation and found that the reason is when that is specified, or a non-zero `letter-spacing` is specified, Gecko disables features `liga` and `clig` based on the CSS Text spec which states:
> When the effective spacing between two characters is not zero (due to either justification or a non-zero value of letter-spacing), user agents should not apply optional ligatures.

while in this case, the composition is applied via `clig` feature. It sounds like `clig` is probably not an optional ligature.

But according to @jfkthame in [the bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1458053#c5), the example in [OpenType feature registry for `clig`](https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae#a-nameclig-idcligatag-clig) clearly indicates it should be optional ligatures as it falls in the same category as ligatures like "fi". On the other hand, he also noticed that in shaping spec of some scripts, `clig` is also mentioned, which makes it less clear whether it is required, so maybe this should be conditionally treated optional.

It would probably be helpful if the spec can provide recommendation on what ligatures should be considered optional. Also any feedback from i18n wg would be great.

r12a:
Ok, this is a bit like a ball of spaghetti. Let me try to unravel it a little.

First of all, the word បាទ used for the example is in the Khmer script, not Myanmar (see https://r12a.github.io/uniview/?charlist=%E1%9E%94%E1%9E%B6%E1%9E%91). 

Second, the ligation occurs between a base character and a combining character, and my understanding is that they should not normally be separated by `text-justify:inter-character`, since they constitute a basic grapheme cluster. The units around which spacing occurs, according to https://drafts.csswg.org/css-text-3/#valdef-text-justify-inter-character, are what CSS calls 'typographic character units'. They are defined at https://drafts.csswg.org/css-text-3/#typographic-character-unit, which implies that a grapheme cluster is usually not broken (ie. you wouldn't want to separate បា).

However, third, in Thai (which is a different script again), the particular vowel-sign represented by the second character apparently is separated in certain similar circumstances, though not identical. This is actually described in an example in the spec (note that the example involves an additional combining character, the anusvara):

[...]

PR #3430, opened 2018-12-11
[css-fonts-3][css-fonts-4] letter-spacing disables optional ligatures. #2644

bungeman:
A non-default value for letter-spacing should disable all
optional ligatures, not just common ligatures. Defer to css-text for
optional ligature definition.

Test that letter-spacing disables discretionary ligatures in particular at
https://chromium-review.googlesource.com/c/chromium/src/+/1294874

### DIFF
--- a/css-fonts-3/Fonts.html
+++ b/css-fonts-3/Fonts.html
@@ -4975,7 +4975,7 @@ body { font-feature-settings: "hwid"; /* Half-width OpenType feature */ }
     title="font-feature-settings!!property">‘<code
     class=property>font-feature-settings</code>’</a>. For example, setting
     a non-default value for the ‘<code
-    class=property>letter-spacing</code>’ property disables common
+    class=property>letter-spacing</code>’ property disables optional
     ligatures.
 
    <li>Font features implied by the value of <a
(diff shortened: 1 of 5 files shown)


svgeesus:
Thanks for the PR and test.

> Defer to css-text for optional ligature definition.

Do you mean [this wording](https://drafts.csswg.org/css-text-3/#letter-spacing-property):

> When the effective spacing between two characters is not zero (due to either justification or a non-zero value of letter-spacing), user agents **should not**  apply optional ligatures, i.e. those that are not defined as required for fundamentally correct glyph shaping

or [this non-normative note](https://drafts.csswg.org/css-text-3/#letter-spacing-property)?

> Note: In OpenType, required ligatures are expected to be associated to the rlig feature. All other ligatures are therefore considered optional. In some cases, however, UA or platform heuristics apply additional ligatures in order to handle broken fonts; this specification does not define or override such exceptional handling.

I agree that optional, rather than just common, ligatures is correct here.

Oh - the ipr bot is complaining, could you [link your GitHub account with your W3C account](https://labs.w3.org/repo-manager/pr/id/w3c/csswg-drafts/3430) please?

bungeman:
Since the checks are now passing, I assume that I've finally hit all the buttons and done all of the verifying required.

By "Defer to css-text for optional ligature definition." I mean both of the sections you mention. css-text here is defining optional ligatures as "those that are not required for fundamentally correct glyph shaping" and that is a great general definition, especially in the css-text specification. The note about OpenType is also good in that it clarifies what this means for the 95% use case. If there were also a note about TrueType shaping in there it would hit 99.9% or so. However, this note is already straying a bit into the css-fonts specification area, so I understand why it is a note.

PR #3472, opened 2018-12-28
[css-text-3] Clarify confusing sentence

frivoal:
@fantasai this PR corrects a sentence which I find somewhat confusing / misleading.

Correcting it to what I believe is the "obvious" correct meaning, but asking you for review in case I am more confused than I thought.


fantasai:
Rewrote your spec changes.

PR #3490, opened 2019-01-08
[web-animations-1] Make the definition of "is current" reflect the associated animation's playback rate

### LINKED ISSUE #3193: [web-animations-1] The definition of a current effect should reflect the playback rate
This has come up in a couple of contexts for me recently, but I'm starting to think that the definition of a **[current](https://drafts.csswg.org/web-animations-1/#current)** effect should reflect the sign of the playback rate.

If you have an animation with a negative playback rate that is in the end delay phase, it should probably be returned by `getAnimations()` even if it is not filling.

birtles:
Fixes #3193.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2226,9 +2226,8 @@ follows:
 
 :   <a>current</a>
 ::  Corresponds to an <a>animation effect</a> that is either <a>in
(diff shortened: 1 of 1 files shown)


birtles:
Not sure why, but I can't add @graouts as a reviewer.

PR #3500, opened 2019-01-10
typo fix

estelle:
`<string>` was considered a tag, so switched to `<dfn><<string>></dfn>`  so it would be visible


dontcallmedom:

tabatkins:
The correct markup for things like this is just to mark it as `for` the appropriate thing. That's used regularly in property definitions. I'll edit the PR and merge it.

PR #3558, opened 2019-01-26
[css-fonts-3] Fix type on onload eventHandler

### LINKED ISSUE #3553: [css-fonts-3] Typo in onload eventHandler
The following pages include an encoded '<' character in the onload eventHandler.

```
window.onload = function () {
  if (!("devicePixelRatio" in window && window.devicePixelRatio > 1)) return;
  var i, hiresElements = document.getElementsByClassName("hires");
  for (i = 0; i &lt; hiresElements.length; i++) {
    var h = hiresElements[i];
    if (h.tagName != "IMG") continue;
    var src = h.getAttribute("src");
    var src2x = src.replace(/\.\w+$/, function(m) { return "@2x" + m; });
    h.src = src2x;
  }
}
```
https://github.com/w3c/csswg-drafts/blob/40eeef57df9ac1c6908588332d3bd6a10eb17bb8/css-fonts-3/Overview.html#L6417
https://github.com/w3c/csswg-drafts/blob/40eeef57df9ac1c6908588332d3bd6a10eb17bb8/css-fonts-3/Fonts.html#L6423

ewilligers:
resolves #3553

### DIFF
--- a/css-fonts-3/Fonts.html
+++ b/css-fonts-3/Fonts.html
@@ -6420,7 +6420,7 @@ interface CSSFontFaceRule : CSSRule {
 window.onload = function () {
   if (!("devicePixelRatio" in window && window.devicePixelRatio > 1)) return;
   var i, hiresElements = document.getElementsByClassName("hires");
-  for (i = 0; i &lt; hiresElements.length; i++) {
+  for (i = 0; i < hiresElements.length; i++) {
     var h = hiresElements[i];
     if (h.tagName != "IMG") continue;
     var src = h.getAttribute("src");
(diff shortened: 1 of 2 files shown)


svgeesus:
Hi @ericwilligers thanks for this. 

Fonts 3 (and Color 3) use an old, unmaintained preprocessor and build Overview.html from Overview.html.src. But that tool is now broken and the published (W3C Recommendation) version of Fonts 3 has a bunch of edits on the post-processed output.

I need to first land the edited (Rec) version back as a new ED. I also plan to delete or archive the .src file (and the redundant Fonts.html) at that time. Then I can apply your patch on top of it. I will also need to generate an erratum for the Rec, also based on your patch.

Which is by way of explaining the `do not merge yet` review result.

svgeesus:
Merged, also now listed [in the Fonts 3 errata](https://www.w3.org/Style/2018/REC-css-fonts-3-20180920-errata.html)

PR #3572, opened 2019-01-30
[css-images-4] Uncomment IDL attribtue

### LINKED ISSUE #428: [cssom][css-images] CSS.elementSources incompatible with CSS being an IDL namespace

saschanaz:
Closes #428

Because current Web IDL spec [allows namespaces to include readonly attributes](https://heycam.github.io/webidl/#idl-namespaces).

Also `Map` is replaced to `any` because AFAIK there is no valid way to represent ES map object in IDL.

### DIFF
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -877,11 +877,9 @@ Using Out-Of-Document Sources: the <code>ElementSources</code> interface</h4>
 
 	<pre class='idl'>
 		partial namespace CSS {
-			// [SameObject] readonly attribute Map elementSources;
+			[SameObject] readonly attribute any elementSources;
 		};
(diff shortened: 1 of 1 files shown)


saschanaz:

PR #3574, opened 2019-01-30
[css-ui] appearance: menulist-button is different from auto

### LINKED ISSUE #3551: [css-ui] Make appearance: menulist-button match WebKit/Chromium
See https://bugzilla.mozilla.org/show_bug.cgi?id=1481615

I believe this value would cause a drop-down `select` to render with its CSS styles (including a down-arrow button), like when you specify a border or a background. Firefox currently makes it equivalent to 'menulist'.

Possibly both approaches are web-compatible enough, but I think we might as well match WebKit/Chromium on this since this "appearance" is needed anyway when border/background is applied.

cc @MatsPalmgren @tkent-google @frivoal

zcorpan:
Fixes #3551

Tests: https://github.com/web-platform-tests/wpt/pull/14996

cc @MatsPalmgren @tkent-google

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -86,6 +86,7 @@ spec:css-backgrounds-3; type:property; text:border-width
 spec:css-overflow-4; type:property; text:text-overflow
 spec:selectors-4; type:selector; text::enabled
 spec:selectors-4; type:selector; text::disabled
+spec:css-color-4; type:property; text:color
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


zcorpan:
Gecko bug
https://bugzilla.mozilla.org/show_bug.cgi?id=1481615

The test that fails in Firefox in https://github.com/web-platform-tests/wpt/pull/14996 is `/css/css-ui/appearance-menulist-button-002.html`

Issue #3538, opened 2019-01-21
[cssom-view] Add pageZoomFactor to window interface


EiraGe:
`window.devicePixelRatio` includes both the device scale and the page zoom. However, there is no easy way to get the page zoom factor alone. 

Developers use different ways to get the zoom factor, but they are either inaccurate or not specced
(e.g window.outerWidth / window.innerWidth, and window.getComputedStyle()['zoom'])

In addition, as mentioned in this [comment](https://github.com/w3c/uievents/issues/150#issuecomment-455258687), `event.screenX/Y` are in DIP(device independent pixels), which includes zoom, and `clientX/Y` are in CSS pixel. Lacking page zoom factor causes developer having trouble converting between the coordinates spaces.

I therefore propose adding a new **window.pageZoomFactor** to expose the zoom factor.

mustaqahmed:
Thanks for raising this, Ella.

To clarify, `pageZoomFactor` here means the ctrl+/- zoom in browser, as defined [here](https://drafts.csswg.org/cssom-view-1/#page-zoom), which is different from pinch zoom and device scale factor.  We have three related coordinate spaces here: physical pixels, device independent pixels (DIPs) and CSS pixels:
![image of proposed factor vs deviceScaleFactor](https://eirage.github.io/Conversion%20factors.png)

We already have [`window.devicePixelRatio`](https://drafts.csswg.org/cssom-view/#dom-window-devicepixelratio) to convert between physical and CSS pixels, but there is currently no way to convert any of these to DIPs or back.  The proposed `pageZoomFactor` fills the gap.

upsuper:
The first question would be, why should content be aware of zoom level? I think content are supposed to be responsive to just the viewport, and the viewport changes based on the zoom level.

Ideally I hope content would never be able to be aware of zoom level. In Gecko, both `window.outerWidth` and `window.innerWidth` are based on the zoomed coordinate space. I would argue that we should change `event.screen{X,Y}` the same way, but that could be tricky.

EiraGe:
> I would argue that we should change event.screen{X,Y} the same way, but that could be tricky. 

It's definitely nicer to have `event.screenX/Y` based on the zoomed coordinate space too, but it's very risky to change it since all major browser (except edge) have them unzoomed for many years.

the `window.outerWidth` and window.innerWidth` in different coordinate space is definitely a bug in chrome. It's only a crazy way developers using to getting the zoom factor.

> why should content be aware of zoom level?

[...]

PR #3580, opened 2019-01-31
[cssom-view]Add pageZoomFactor to window interface

EiraGe:
As discussed in #3538, adding pageZoomFactor to give developers more flexibility when


EiraGe:
@mustaqahmed could you please take a quick look on this?

PR #3618, opened 2019-02-05
Address typo in ellipse() syntax

### LINKED ISSUE #3609: [css-shapes] Minor typo in ellipse() syntax
The ellipse syntax is defined as `ellipse( <shape-radius>{2}? [ at <position> ]? )`

Based on the css-values definition, it looks like a brace multiplier followed by an optional element is illegal.

Shouldn't it be `ellipse( [<shape-radius>{2}]? [ at <position> ]? )`

https://github.com/w3c/csswg-drafts/blob/6f9324adc8d6207923dbf8e78d018ace795f4276/css-shapes-1/Overview.bs#L379

ewilligers:
An ellipse may have an x-radius and y-radius, or neither.

resolves #3609

### DIFF
--- a/css-shapes-1/Overview.bs
+++ b/css-shapes-1/Overview.bs
@@ -376,7 +376,7 @@ Supported Shapes</h3>
 			</ul>
 		</dd>
 		<dt><dfn>ellipse()</dfn> =
-			ellipse( <<shape-radius>>{2}? [ at <<position>> ]? )
+			ellipse( [ <<shape-radius>>{2} ]? [ at <<position>> ]? )
 		</dt>
 		<dd>
 			<ul>


mfwitten:
If it takes more than 23 days for someone to resolve this pull request, then there is something objectively and indisputably wrong with the review process.

It has taken more than 23 days.

ewilligers:
@astearns @atanassov please review

astearns:
@ewilligers I'm fine taking the change, but I'm not certain why the group is required here.

ewilligers:
https://drafts.csswg.org/css-values/#component-combinators

> Every type, keyword, or bracketed group may be followed by **one** of the following modifiers:
> - A  question mark (?) indicates that the preceding type, word, or group is optional (occurs zero or one times).
> - A single number in curly braces ({A}) indicates that the preceding type, word, or group occurs A times.

Two modifiers are not permitted.

mfwitten:
Why was my comment deleted, @w3c? Please reinstate it for the record.

astearns:
It didn't add anything to the discussion here. For the record, everything that happens in the W3C GitHub org is separately archived.

mfwitten:
Well, that's your opinion.

I disagree, and in a discussion about a project, other readers should be allowed to form their own opinions of my legitimate input.

There's a strong correlation between my comment and the resolution of this pull request; indeed, there's a good argument for a causal relationship, so it seems suspicious to suggest that my comment didn't add anything to the discussion.

Please reinstate my comment, because what people read is *this* discussion (not some separate archive), and I want my name (*mfwitten*) associated with my comment, not with @w3c's comment that my input has been deleted, which makes it appear as though my input is somehow uncouth.

I mean, before deletion, there was just my one comment, which led to the swift resolution of this pull request. Now, there is a discussion about @w3c's deletion of my comment; it seems to me that it is the deletion of my comment that has detracted from the discussion and that is leading rapidly to dampened progress or cooperation.

If a person feels compelled to provide input, then that's not something that you should ignore, let alone explicitly silence; such input represents a perspective that others might share. At best, such silencing tends to dissuade future contributions, some of which you might actually consider to be valuable.

Have a little leeway, you know, for the sake of [noise margins](https://en.wikipedia.org/wiki/Static_discipline).

plinss:
Michael, frankly the tone of your comment was rude and disrespectful if not outright abusive. IMO it was also in violation of the [W3C Code of Ethics and Professional Conduct](https://www.w3.org/Consortium/cepc/).

If I were you I'd be grateful that there _isn't_ a readily accessible record of your behavior. Alan was doing you a favor by deleting it. But instead you choose to double-down and suggest that demonstrating being abusive is the way to get things done.

This is not acceptable and is not the way to foster involvement in a community where many of the participants are volunteering their time uncompensated. You have no right to demand action by anyone for any reason.

Not all input is welcome, and silencing disruptive and disrespectful input actually does **more** to encourage future contributions from people who don't care to be at the receiving end of such behavior.

Perhaps you'd benefit from reading some of the educational materials of the [Positive Work Environment Task Force](https://www.w3.org/Consortium/pwe/#Education).

PR #3639, opened 2019-02-09
Use Sam's correct name

frivoal:
See https://gsnedders.com/who-even for details.


frivoal:
It's everything a case insensitive grep could find.

PR #3657, opened 2019-02-19
[css-images-4] Fix gradient color stop syntax

### LINKED ISSUE #1334: [css-images-4] Gradient color stop syntax is incorrect
The [gradient color stop syntax as currently defined in CSS Images 4](https://drafts.csswg.org/css-images-4/#color-stop-syntax) makes a stop position mandatory for every stop. It should be optional.

Since some of that content is due to be integrated into CSS Images 3 (per #1284), it's would be helpful to get a correction sooner rather than later.

AmeliaBR:
Closes #1334

I'm pretty sure this was the only change required. Not sure why I didn't make a PR at the time I filed the issue, but I guess I was busy…

### DIFF
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -1336,7 +1336,7 @@ Gradient Color-Stops</h3>
 	<pre class=prod>
 		<dfn>&lt;color-stop-list></dfn> =
 			[ <<linear-color-stop>> [, <<linear-color-hint>>]? ]# , <<linear-color-stop>>
-		<dfn>&lt;linear-color-stop></dfn> = <<color>> && <<color-stop-length>>
+		<dfn>&lt;linear-color-stop></dfn> = <<color>> && <<color-stop-length>>?
 		<dfn>&lt;linear-color-hint></dfn> = <<length-percentage>>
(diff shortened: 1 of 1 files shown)


fantasai:
While we're here... why is the color and length reorderable in Level 4? There was a fixed order in Level 3, was that a change we decided to make or an error in the spec?

AmeliaBR:
@fantasai That's a good question. [I just tested it](https://codepen.io/AmeliaBR/full/f0ce61aa57da3a3793738c8039e1631a) in Chrome, Edge, and Firefox, and none support the length value before the color.

The Git blame isn't helping me find a reason, either. The `&&` syntax goes back to the [first commit for Images 4](https://github.com/w3c/csswg-drafts/commit/c0bb64d19fd683b4352e929545eaa27facc5aa15#diff-d391915319a8fcf35f6ad0aa2b4aa8c1R673).

As an author, I wouldn't mind having the extra flexibility.  I've made the mistake of writing it in the wrong order & then having to debug why my gradient wasn't showing up.

But we'd need commitments to actually implement it — maybe at the same time as implementing the two-value syntax? But then we'd need to discuss whether `blue, 20px green 30px, blue` is a valid two-value form.

So maybe that's a separate issue?

PR #3690, opened 2019-02-28
[css-tables-3] Remove file `debug.log'

mfwitten:
It was presumably a mistake to add it to the repository.


dontcallmedom:

astearns:
We don't need anything from this person. Someone else can clean up this file, or replace its contents with a not on how not to behave: https://github.com/Netatalk/Netatalk/issues/107

(he ran through something similar with us yesterday, sealioning us through email when he got annoyed that the github conversation was locked)

PR #3723, opened 2019-03-12
[css-scoping-1] Update Changes section

xfq:


alshamiri5 (approved):

frivoal:
This pull request is made with a branch in the main repo, rather than from a fork. This continuously puts strain on the build server, which is rebuilding it all the time to keep the bikeshed references current. This is OK for branches that need to have a preview, but this seems stale. It should be merged or closed.

tabatkins (approved):

PR #3725, opened 2019-03-12
[css-ruby-1] Update Changes section

xfq:
Also integrated changes from previous publications.


fantasai:
> Also integrated changes from previous publications.

You didn't have to, but it's nice. :)

Issue #3401, opened 2018-12-06
[css-nav-1] changing the spatnav scroll controls from JS to declarative


frivoal:
During TPAC, several engineers from Google (including I believe @eaenet and @chrishtr ) expressed concern over the `navbeforescroll` event, on the grounds that having to run any kind of synchronous JS right before scrolling (and even more so in the middle of scrolling, as could be the case if the user tries to repeatedly navigate in the same direction) was a performance concern.

Changing from an even to an observer would slightly attenuate that concern as it is cheaper, but not completely as that still involves running JS synchronously at that point.

In theory, I believe the 3 events (`navbeforescroll`, `navbeforefocus`, and `navnotarget`) are necessary and sufficient to be able to intercept spatnav before anything happens. However, in practice, I can only think of one use case that calls for `navbeforescroll`: the one illustrated in [example 5](https://drafts.csswg.org/css-nav-1/#focus-only), which allows the authors to switch the behavior from searching for candidates from within just the visible elements in the container to within all elements in the container regardless of visibility.

In [the spatial navigation steps](https://drafts.csswg.org/css-nav-1/#spatial-navigation-steps ), steps 5.1 and 7 invoke [`finding focusable areas`](https://drafts.csswg.org/css-nav-1/#find-focusable-areas), relying on the optional `visibleOnly` argument defaulting to `true`. Invoking it with `visibleOnly` set to `false` would switch to the other behavior we're trying to cover here.

Providing a declarative solution to this problem is very easy, and has none of the performance downsides: an inherited css property whose computed value on `searchOrigin` in step 5.1 or on `container` in step 7 would act as the switch between the two modes. Strawman name:
```css
spatial-navigation-candidates: visible | all
```

If we add this, I think we could remove `navbeforescroll` and not lose any useful capability, solving the performance concern.

jihyerish:
I agree with removing `navbeforescroll` event, in terms of the scroll latency.

After dropping `navbeforescroll` event, when the `navnotarget` event occurs?
(1) When there isn’t any candidate in the spatial navigation container and it cannot be scrolled at the same time. (current spec)
(2) When there isn’t any visible candidate in the spatial navigation container no matter what it can be scrolled more or not.

frivoal:
If `spatial-navigation-candidates: visible` (the initial value), then it would be (1). If `spatial-navigation-candidates: all`, then it would be:
[...]

PR #3732, opened 2019-03-15
[css-nav-1] Add spatial-navigation-action CSS Property

jihyerish:
Add the `spatial-navigation-action` CSS property which can specify the spatial navigation behavior of the container.

IDL:
```css
spatial-navigation-action: auto | focus | scroll
```
* `auto`: The directional input moves the focus to the visible candidates inside the spatial navigation container. If there isn't any visible candidate, it scrolls the container.
* `focus`: The directional input moves the focus to candidates inside the spatial navigation container. The user cannot manually scroll the container.
* `scroll`: The directional input scrolls the container manually.
(`scroll` value is noted as 'at-risk')

Resolves: https://github.com/w3c/csswg-drafts/issues/3401


css-nav-1/Overview.bs
```diff
@@ -1392,6 +1250,79 @@ Note: the 'spatial-navigation-contain' property is <a>at-risk</a>.
...
 </div>
 
+<h3 id=css-property-spatialnavigationaction>
+Defining the spatial navigation behavior : the 'spatial-navigation-action' property</h3>
+
+<pre class='propdef'>
+Name: spatial-navigation-action
+Value: auto | focus | scroll
+Initial: auto
+Applies to: <a>spatial navigation focus container</a>
```
jihyerish: @frivoal Which is the most correct description among:
(1) all element
(2) spatial navigation focus container
(3) scrollable spatial navigation focus container

(1) is possible, but specifying this property is meaningless for the element which isn't a spatial navigation container.

anawhj: You can refer to the following links in terms of the consideration above.
https://www.w3.org/TR/cssom-view-1/#smooth-scrolling
https://www.w3.org/TR/css-scroll-snap-1/#scroll-padding

jihyerish: I think this property is meaningful for the scrollable element.
Other properties such as https://drafts.csswg.org/cssom-view/#smooth-scrolling uses 
`scrolling box` as its applied element, so I'll change to it.

anawhj: Great! seems no problem now. +1

Issue #2258, opened 2018-02-02
[selectors-4] :defined


annevk:
HTML defines `:defined`: https://html.spec.whatwg.org/multipage/semantics-other.html#selector-defined.

1. Is that enough for a definition of a new pseudo-class?
2. Is it fine for HTML to define this?
3. Should Selectors contain a registry of externally defined selectors?

(I'm asking this in part since we're considering a `:heading(level)` pseudo-class.)

See also: https://github.com/w3c/webcomponents/issues/665.

cc @domenic @snuggs

fantasai:
Can I get a layperson's definition of :defined and what it's used for? :)

Wrt :heading(), we've gotten requests for this an declined them so far. If it's something we want to add, we can add it to Selectors--it's definitely not an HTML-specific concept, so it should go into Selectors and not HTML.

annevk:
`:defined` matches all elements that had no problem during construction, roughly. It's useful since you can now implement your own element class and sometimes you want to match elements for which the constructor has not yet run (or vice versa).

As for headings, that sounds reasonable, though we'll run into the data model issue again (though arguably we do already for other things, so it doesn't necessarily add to the problem).

snuggs:
To embellish on @annevk's statement with an example.

```html
<foo-bar></foo-bar>

<style>
  foo-bar { visibility: hidden }
  foo-bar:defined { visibility: show }
</style>
```

If a user wanted to preserve real estate but not display until element is `defined`.

@annevk does that help or hurt?

snuggs:
@annevk @fantasai is this good to close now that is on track with CSS Level4?

annevk:
I think it would make sense to close this when the PR that adds this to Selectors Level 4 lands.

hober:
> I think it would make sense to close this when the PR that adds this to Selectors Level 4 lands.

As w3c/csswg-drafts#3735 landed a while ago, can we close this issue now?

PR #3735, opened 2019-03-15
[selectors-4] Add :defined pseudo-class

spectranaut:
[selectors-4] Add the `:defined` pseudo-class as discussed in #2258. 

The original definition has been moved verbatim from the html spec and a note and examples added. PR to remove the [definition from the html spec from WHATWG](https://github.com/whatwg/html/pull/4433).

A test for this definition has previously been submitted: https://github.com/ewilligers/web-platform-tests/commit/92a561a3121a1831214a7b9f47931904ae3df5e7

Worked on with @zcorpan and @isaacdurazo.

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -1415,6 +1415,33 @@ Namespaces in Elemental Selectors</h3>
 	that has not been previously <a href="#nsdecl">declared</a>
 	is an <a>invalid selector</a>.
 
+<h3 id="the-defined-pseudo">
+The Defined Pseudo-class: '':defined()''</h3>
+
+	The <dfn id='defined-pseudo'>:defined</dfn> <a>pseudo-class</a> matches any element that is
+        <a for=Element>defined</a>.
+
+	Note: The '':defined'' pseudo-class is used to differentiate between a <a>custom element</a> that
+        has been successfully constructed from custom elements that have not yet been constructed or have
+        failed to be constructed.
+
+	<div class="example">
+		For example, the '':defined'' pseudo-class can be used to hide a custom element before a
+                script defines the custom element but after it is initially rendered:
+
+		<pre>custom-element { visibility: hidden }</pre>
+		<pre>custom-element:defined { visibility: visible }</pre>
+	</div>
+
+	An element that is not a custom element will also match the '':defined'' pseudo-class.
+
+	<div class="example">
+		For example, this style sheet will turn all paragraph text green:
+
+		<pre>p:defined { color: green }</pre>
+	</div>
+
+
 <h2 id="attribute-selectors">
 Attribute selectors</h2>
 
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -1419,15 +1419,15 @@ Namespaces in Elemental Selectors</h3>
 The Defined Pseudo-class: '':defined()''</h3>
 
 	The <dfn id='defined-pseudo'>:defined</dfn> <a>pseudo-class</a> matches any element that is
-        <a for=Element>defined</a>.
+	<a for=Element>defined</a>.
 
 	Note: The '':defined'' pseudo-class is used to differentiate between a <a>custom element</a> that
-        has been successfully constructed from custom elements that have not yet been constructed or have
-        failed to be constructed.
+	has been successfully constructed from custom elements that have not yet been constructed or have
+	failed to be constructed.
 
 	<div class="example">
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -1415,6 +1415,32 @@ Namespaces in Elemental Selectors</h3>
 	that has not been previously <a href="#nsdecl">declared</a>
 	is an <a>invalid selector</a>.
 
+<h3 id="the-defined-pseudo">
+The Defined Pseudo-class: '':defined()''</h3>
```
zcorpan: Remove `()`

annevk:
See my comments at https://github.com/whatwg/html/pull/4433#issuecomment-473514957. I don't think a change to how we do layering between CSS and HTML has been discussed so this specification should probably define this header more abstractly while leaving the actual matching requirements for DOM/HTML to HTML.

zcorpan:
@annevk ok. Per the meeting discussion in https://github.com/w3c/csswg-drafts/issues/2258#issuecomment-440756225

> &lt;dael> Rossen: If this has impl in various UA and since it's a selector makes sense as part of selectors unless we're against it in the way it's defined. We should accept it and refine
> &lt;dael> florian: Sounds good
> &lt;dael> Rossen: Other comments or Objections to adding :defined to selectors L4?
> &lt;dael> RESOLVED: Add :defined to selectors L4

As I understand this, the WG resolved to take the spec as-is (for now). cc @atanassov

Do we want this to follow the model of e.g. `:link` instead?

annevk:
I think either Selectors embraces DOM/HTML and is defined on top of that data model or it keeps things abstract (which is shaky btw, there's some open issues around that) as it tries to do today. Doing both is too confusing though.

fantasai:
I agree with @annevk and I believe the resolution was to add it at the same level of definition as e.g. :link etc, as he expects.

zcorpan:
I've changed this so that it has a high-level definition in Selectors, and has the examples be specific for HTML.

zcorpan:
There are some tests in wpt for `:defined` already:

```
$ git grep ":defined"
css/selectors/invalidation/defined.html:    <title>CSS Selectors Invalidation: :defined</title>
css/selectors/invalidation/defined.html:    <meta name="assert" content="This tests that the :defined selector is effective">
css/selectors/invalidation/defined.html:      #a1:defined {
css/selectors/invalidation/defined.html:      :defined + #b1 {
css/selectors/invalidation/defined.html:      :defined > #c1 {
css/selectors/invalidation/defined.html:      div + :defined + * #d1 {
css/selectors/invalidation/defined.html:      }, ":defined selector is effective");
custom-elements/Document-createElement.html:    assert_false(instance2.matches(':defined'));
custom-elements/pseudo-class-defined.html:  :defined { color:${defined}; }
custom-elements/pseudo-class-defined.html:  :not(:defined) { color:${not_defined}; }
custom-elements/pseudo-class-defined.html:  // Test ":defined" changes when the custom element was defined.
custom-elements/pseudo-class-defined.html:    assert_equals(element.matches(':defined'), expected, 'matches(":defined")');
custom-elements/pseudo-class-defined.html:    assert_equals(element.matches(':not(:defined)'), !expected, 'matches(":not(:defined")');
custom-elements/pseudo-class-defined.html:  }, `${description} should ${expected ? 'be' : 'not be'} :defined`);
```

selectors-4/Overview.bs
annevk: Seems this comment still applies 😊

This part looks reasonable to me otherwise.

zcorpan:
```suggestion
The Defined Pseudo-class: '':defined''</h3>
```

zcorpan:
Ping @fantasai @tabatkins :)

tabatkins:
Ah sorry for the delay.

Hmm. Overall this is fine, but now it sounds *really* abstract, to the point that I don't think I could figure out what it meant if I didn't already know about custom elements and their definitions.

I'll do a little tweaking and then accept it.

zcorpan:
Thanks @tabatkins!

bkardell:
If we get to more than just a single global registry, does this still hold up?

zcorpan:
@bkardell this says it's up to the host language to say what matches `:defined`. For HTML this is https://html.spec.whatwg.org/multipage/semantics-other.html#selector-defined - that could be changed in HTML, and SVG could say something else, without needing updates to the Selectors spec.

PR #3643, opened 2019-02-11
[css-regions-1] Convert Region to a mixin

foolip:
Following https://github.com/heycam/webidl/pull/433.

PR #3737, opened 2019-03-17
[css-regions-1] Convert Region to a mixin (reprise)

foolip:
The actual `mixin` keyword was overlooked in https://github.com/w3c/csswg-drafts/pull/3643.


foolip:
@astearns @tabatkins PTAL.

Issue #1950, opened 2017-11-07
[css-transforms-2] transform-style: flat should not create stacking context, and 3D context grouping


smfr:
Discussion about how transform-style: flat should work.

mattwoodrow:
We should also resolve what happens when the author specifies transform-style:preserve-3d, but a grouping property forces it to flat.

Consider a testcase that has a preserve-3d element, and then you later add a grouping property that doesn't force a stacking context and containing block (overflow:hidden, clip, clip-path etc).

I find it a bit weird from an implementation standpoint that the change handling for overflow:hidden (which normally has nothing to do with containing blocks) needs to check if we also have transform-style:preserve-3d, and disable us from being a containing block. It's a weird dependency.

I think authors might also find it weird that adding a grouping property not only changes the visual rendering of preserve-3d, but also has the layout change of removing the containing block.

I'd prefer if the flattening effect of grouping properties applied to the rendering model, but not the layout. We could spec that as defining the stacking context/containing block parts of transform-style:preserve-3d are determined by the computed value of transform-style.

mattwoodrow:
I wrote up an explanation of how Gecko interprets the spec and how the implementation works, along with a bunch of testcases to show some of the differences: https://docs.google.com/document/d/1yb4a_uhTG3KmcbGPta4B9p67v1HO_qY8ZMk-otGOwTc

It looks like the primary difference is that WebKit considers a flattening element (element with a flattening property, or stacking context with transform-style:flat) to be what establishes a 3d rendering context for its children (but not itself), whereas both Gecko and blink consider an element with transform-style:preserve-3d to be what establishes a 3d rendering context for itself and its children.

I think the blink/Gecko interpretation will be simpler to specify, since the root of the 3d rendering context needs to be a stacking context to create a flattened representation. Flattening properties (like overflow:scroll) aren't necessarily a stacking context, so we'd need to make them one, if (and only if) they happen to have 3d descendants. In Gecko/blink the root always has transform-style:preserve, which is explicitly a stacking context.

Commenting here since the css-meeting bot did, but this is effectively the same underlying issue as #3138, #1944, #1951 and #1952.

mattwoodrow:
We also discussed how WebKit/blink look to non-parent ancestors to determine if we're participating in a 3d rendering context (spec says look at containing block, current behaviour appears to be stacking context), whereas Gecko looks at the direct parent element.

[...]

PR #3750, opened 2019-03-20
[css-transforms-2] Revert to the TR rendering model, where transform-style:preserve-3d establishes the 3d rendering context. #1950

mattwoodrow:
This changes the 3d rendering model to be closer to the old TR (and blink/gecko's implementation), where transform-style:preserve-3d is what establishes/extends a rendering context, rather than flat.

It also makes transform-style:preserve use parent elements instead of containing blocks (or stacking contexts) to resolve handling of overflow:hidden (which is neither).

### DIFF
--- a/css-transforms-2/Overview.bs
+++ b/css-transforms-2/Overview.bs
@@ -51,7 +51,7 @@ New transform functions for the 'transform' property allow three-dimensional tra
 and additional properties make working with three-dimensional transforms easier,
 and allow the author to control how nested three-dimensional transformed elements interact.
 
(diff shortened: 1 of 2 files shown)


smfr:
I don't have time to do a full review of this at the moment, but I wonder if we should keep working on a branch for a while to avoid having the ED bounce between two significantly different versions, until we have consensus about reverting to the older version of the spec.

I think getting that consensus should also involves some more data-driven estimations of web compatibility. Maybe implementors can gather information about sites using 3D transforms (e.g. from bug reports) so we can see what's important for compat.

Issue #3752, opened 2019-03-20
[css-scroll-snap-1] Example 11 should also include scrollbar arrows


egirard:
https://drafts.csswg.org/css-scroll-snap-1/#intended-direction should specify that clicking on the scrollbar arrows defines an intended direction. Other examples have specified what happens if you move the scrollbar thumb or page by clicking above or below the thumb. Seems odd to have left this one out.

PR #3753, opened 2019-03-20
[css-scroll-snap-1] Include scroll arrows in "intended direction" examples. #3752

egirard:
As per https://github.com/w3c/csswg-drafts/issues/3752

All other scrollbar operations are listed in the examples. Seems odd to not include the arrows.


dontcallmedom:

Issue #3384, opened 2018-12-03
[css-nav-1] Improve the distance function


frivoal:
---

Migrated from https://github.com/WICG/spatial-navigation/issues/114
Originally created by @jihyerish on *Thu, 11 Oct 2018 08:34:46 GMT*

---
The distance function in the spec which measures the distance between two elements causes unexpected focus movement.
It needs to consider:
- Multiple candidate elements are adjacent to the starting point element (currently focused element)
- Transformed elements
- The element with rounded corner

So the function needs some experiment for getting the more reasonable result and simplifying.

Also, the function in the polyfill doesn't match exactly with the spec; it needs to be fixed.

jihyerish:
I think this issue is duplicated one with https://github.com/w3c/csswg-drafts/issues/3388.
So, I'll close this issue.

jihyerish:
## Problem
When the spatial navigation is used for the real pages, there are unexpected behavior in UX point of view because of the distance function above.

### Case 1
The focus moves to the unexpected element with the smallest Euclidean distance but not aligned with the search origin in the navigation direction.
e.g.) [Related test case](https://wicg.github.io/spatial-navigation/tests/ux/spatnav-distance-function-grid-001.html)

### Case 2
The focus moves to the unexpected candidate when there are multiple candidates which have the same Euclidean distance.
e.g.) [Related test case](https://wicg.github.io/spatial-navigation/tests/ux/spatnav-distance-function-grid-003.html)

### Case 3
The focus moves to the unexpected candidate when the Euclidean distances of candidates are slightly different (about 1px), but the degree of alignment with the search origin are significantly different.
e.g.) [Related test case](https://wicg.github.io/spatial-navigation/tests/ux/spatnav-distance-function-grid-align-002.html)

## Approach
The distance formula needs to change as below:

#### 1. Remove the factor about calculating the absolute distance in the navigation direction between a candidate and the search origin.
In _**Case 1**_, the aligned elements are more desirable element to gain the focus. But calculating the absolute distance in the navigation direction (B Factor of the distance formula in the spec) penalizes elements which are aligned, so it’s the reason for the unexpected behavior.

[See the test result](https://wicg.github.io/spatial-navigation/tests/ux/result.html#2-remove-redundant-distance-value)

#### 2. Add the factor about calculating the degree of alignment between a candidate and the search origin.

This aims to deal with the issues of _**Case 2**_ and _**Case 3**_.

[...]

Issue #3353, opened 2018-11-29
[css-nav-1] Define points precisely for measuring the distance 


jihyerish:
In the [selecting the best candidate](https://drafts.csswg.org/css-nav-1/#select-the-best-candidate) step, the distance is measured for each candidate from the starting point. 
But it's unclear that which points of the boundary box area are chosen. (for P1 and P2)
It would be 
(1) the center point of the  boundary box 
(2) the center point of edges of each element
(3) the closest point along the edges of each element (Implementation in Blink)

* Related issue : https://github.com/w3ctag/design-reviews/issues/287

jihyerish:
I'd like to suggest selecting the point as the **_closest point along the edges_** of each element (Implementation in Blink).

Selecting the point from the center of an element is also considerable, but it cannot solve the problem when candidate elements are aligned with the same Euclidean distance from the currently focused element.

I've implemented test cases in UX point of view, and tested for several ways to select points.
Please see the result in [here](https://wicg.github.io/spatial-navigation/tests/ux/result.html#1-selection-of-points).

PR #3755, opened 2019-03-22
[css-nav-1] Modify the definition of the distance function

jihyerish:
This changes aim to update the spec with the improved distance function (Fixes https://github.com/w3c/csswg-drafts/issues/3384)
  - Returns more expected result when the elements are aligned (Fixes https://github.com/w3c/csswg-drafts/issues/3388)
- Added precise definition of points which are used for calculating the distance (Fixes https://github.com/w3c/csswg-drafts/issues/3353)
- Added the description about calculating the distance when the search origin is a point


css-nav-1/Overview.bs
```diff
@@ -1236,37 +1236,69 @@ follow the following steps:
...
+</div>
+
+<div algorithm="to calculate the distance">
+To <dfn lt="calculating the distance | to calculate the distance">calculate the <var>distance</var></dfn>, the function is defined as below:
+
+<pre class="prod">
+    <var>distance</var> = <var>euclidian</var> - <var>alignment</var> + <var>displacement</var> - <var>sqrt(Overlap)</var>
+</pre>
+
+The meaning of each factor in the function is as follows:
```
anawhj: each factor -> each factors

css-nav-1/Overview.bs
```diff
@@ -1236,37 +1236,69 @@ follow the following steps:
...
-                <dd>The absolute distance in the <var>dir</var> direction between <var>P1</var> and <var>P2</var>
-
-              <dt><var>C</var>:
-                <dd>The absolute distance in the direction which is orthogonal to <var>dir</var> between <var>P1</var> and <var>P2</var>
-
-              <dt><var>D</var>:
-                <dd>The square root of the area of intersection between the <a>boundary boxes</a> of <var>candidate</var> and <var>searchOrigin</var>
-            </dl>
-        3. Return the item of the <var>candidates</var> set that has the smallest <var>distance</var>.
+            1. If the <var>searchOrigin</var> is an element, then find the points <var>P1</var> inside the <a>boundary box</a>
```
anawhj: find the points P1 -> find the point P1

css-nav-1/Overview.bs
```diff
@@ -1236,37 +1236,69 @@ follow the following steps:
...
-
-              <dt><var>D</var>:
-                <dd>The square root of the area of intersection between the <a>boundary boxes</a> of <var>candidate</var> and <var>searchOrigin</var>
-            </dl>
-        3. Return the item of the <var>candidates</var> set that has the smallest <var>distance</var>.
+            1. If the <var>searchOrigin</var> is an element, then find the points <var>P1</var> inside the <a>boundary box</a> 
+                of <var>searchOrigin</var> and P2 inside the boundary box of <var>candidate</var>.
+            2. Else if the <var>searchOrigin</var> is a point, then let the point <var>P1</var> be the <var>searchOrigin</var> 
+                and find the point <var>P2</var> inside the <a>boundary box</a>.
+        3. Let <var>intersectRect</var> is intersected area between a candidate and the search origin.
```
anawhj: is intersected area -> be an intersected area

css-nav-1/Overview.bs
jihyerish: 'each factor' is the correct grammar.

css-nav-1/Overview.bs
jihyerish: The word 'points' represents both P1 and P2 (in the next line). Therefore I think the current expression is correct.

css-nav-1/Overview.bs
jihyerish: Thanks, I'll fix this.

css-nav-1/Overview.bs
anawhj: I was a bit confused. Totally agree on it.

css-nav-1/Overview.bs
anawhj: Agree on it.

css-nav-1/Overview.bs
```diff
@@ -1236,38 +1236,100 @@ follow the following steps:
...
+            * Else if the <var>dir</var> is {{SpatialNavigationDirection/up}} or {{SpatialNavigationDirection/down}}, 2
+
+    <dt><var>alignment</var>
+    <dd>The degree of alignment in <var>dir</var> between <var>searchOrigin</var> and <var>candidate</var>
+    <dd>Defined as <pre class="prod"><var>alignBias</var> * <var>alignWeight</var></pre>
 
+        * <var>alignBias</var>
 
+            * If the <var>dir</var> is {{SpatialNavigationDirection/left}} or {{SpatialNavigationDirection/right}}, 
+            height of <var>intersectRect</var> / height of <var>candidate</var>
```
hugoholgersson: In #3388 you wrote "length of the edge of the search origin". Here you wrote _candidate_'s edge" instead... Mistake?

anawhj: The edge of search origin on the comment (#3388) has been replaced with the current expression on the spec which describes an edge of candidate. If you have another opinion, please put it here so that we could consider. With minor updates (figures of some weights), we are going to apply the updated expression into Blink.

BTW, the polyfill would operate it with the edge of search origin that seems not to match with the current spec. @jihyerish please check and update on it.
https://github.com/WICG/spatial-navigation/blob/master/polyfill/spatial-navigation-polyfill.js#L1059

hugoholgersson: I created [crbug.com/958845](https://crbug.com/958845) for the Blink work. 

See this example:

```
   ____________________ top edge of container or other focusable. 20 units wide.
   YYYYY XXXXXXXXXXXX
```
Here we want X to win so we do 12/20 .

`12 / 20 > 5 / 20` => we give X a **larger** `alignment` because we want to decrease the distance to X more than we decrease the distance to Y. (`alignment` is a negative term in the formula).

jihyerish: Yes, you're right. My mistake...
Thanks for pointing this out! I'll fix this.

jihyerish: This has been fixed by https://github.com/w3c/csswg-drafts/commit/bfe77ef53581fafe9a2a97696ef0413f8910e3d4.

Issue #2331, opened 2018-02-17
[css-values] Trigonometric functions


LeaVerou:
I keep stumbling on use cases that need trigonometric functions to be solved, and I always have to resort to JS or a preprocessor for them, which is a pain. Basically almost any time you need to match two things and angles are involved.

Sadly I couldn't find most of my use cases, but here are two: 

- https://codepen.io/leaverou/pen/JpyZMO - cos, atan (via JS)
- [play.csssecrets.io/folded-corner-realistic](http://play.csssecrets.io/folded-corner-realistic) — sin, cos (precomputed)

Also the following use sqrt (precomputed), but would need trigonometry in the general case :

- [play.csssecrets.io/inner-rounding](http://play.csssecrets.io/inner-rounding)
- [play.csssecrets.io/diamond-images](http://play.csssecrets.io/diamond-images)
- [play.csssecrets.io/bevel-corners](http://play.csssecrets.io/bevel-corners)
- [play.csssecrets.io/folded-corner](http://play.csssecrets.io/folded-corner)

(credits: @MeFoDy for collecting these from my book)

What do you think? Can we make it happen?

thebabydino:
My primary use case is for syncing rotation angles and scale factors or translation distances when I animate multiple elements such that corners/ edges match throughout the animation.

In [this demo](https://codepen.io/thebabydino/pen/JNZGwE/), I need to sync the rotation angle of the bright yellow squares and the post-skew scale factor of the orange squares. This is done by doing the trigonometric computations in the JS and then setting the values to the custom properties used in the transform.

A way to do it all from the CSS would be immensely useful.

I can try to fake it with different timing functions for rotation angles and scale factors or translation distances as I did in [this demo](https://codepen.io/thebabydino/pen/ybExWm/), but that's just hardcoding the cubic-bezier() for every case.

Crissov:
Adding these would require a solution to enter π other than `0.5turn` etc. #309

LeaVerou:
> Adding these would require a solution to enter π other than 0.5turn etc. #309

Um, why? (Yes, I did read the issue you linked to)

MeFoDy:
Some awesome use cases in [presentation](http://askd.rocks/pres/css/) by @askd:
<details><summary>Triangle</summary>

```css
/* http://askd.rocks/pres/css/#triangle */
.triangle {
  position: relative;
  width: var(--w);
  height: var(--h);
  overflow: hidden;
}
.triangle::before {
[...]

PR #3761, opened 2019-03-26
[css-values-4] Clean up trig function definitions

AmeliaBR:
Related to #2331 and commit b8935e7df14ed544c51a991cf10f15bf077eb603

The only normative change (from Tab's commit) is that `atan2()`
is changed to support values with dimensions (so long as both dimensions are the same).
I don't see any implementation reason not to allow it;
they type checking rules are the same as `hypot`
(except that the types cancel out, so don't affect the final type of the function).

Most other corrections should be obvious in the diff.

I changed the note about atan2 because it used terminology
based on a y-axis-up coordinate system
and CSS/SVG uses a y-axis-down system.

I changed the notes about simplifying at computed value time
to refer to _calculations_ instead of _math functions_,
because some complex nested functions cannot be simplified early.
(E.g., `hypot(100%, 5em)` can't be simplified until percentages can be turned into lengths.)

**Edits still required:**

The [Syntax](https://drafts.csswg.org/css-values/#calc-syntax) and [Serialization](https://drafts.csswg.org/css-values/#calc-serialize) sections should probably include entries for the new functions. But I wasn't sure how that should look.

*PS,* Thanks for the main commit @tabatkins! I'd said I'd give it a go, but who knows when I would have got to it.

(And I probably would have just linked to ECMAScript for all the stuff to do with infinities and NaN.)

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -2472,13 +2472,13 @@ Trigonometric Functions: ''sin()'', ''cos()'', ''tan()'', ''asin()'', ''acos()''
 	''sin()'', ''cos()'', ''tan()'', ''asin()'', ''acos()'', ''atan()'', and ''atan2()''--
 	compute the various basic trigonometric relationships.
 
-	The <dfn lt="sin()">sin(A)</dfn>, <dfn lt="cos(A)">cos()</dfn>, and <dfn lt="tan()">tan(A)</dfn> functions
+	The <dfn lt="sin()">sin(A)</dfn>, <dfn lt="cos()">cos(A)</dfn>, and <dfn lt="tan()">tan(A)</dfn> functions
 	all contain a single [=calculation=]
 	which must resolve to either a <<number>>
 	or an <<angle>>,
 	and compute their corresponding function
 	by interpreting the result of their argument as radians.
-	(That is, ''sin(45deg)'', ''sin(.125)'', and ''sin(3.14159 / 4 * 1rad)''
+	(That is, ''sin(45deg)'', ''sin(.125turn)'', and ''sin(3.14159 / 4 * 1rad)''
 	all represent the same value,
 	approximately ''.707''.)
 	They all represent a <<number>>;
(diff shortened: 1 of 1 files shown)


css-values-4/Overview.bs
```diff
@@ -2472,13 +2472,13 @@ Trigonometric Functions: ''sin()'', ''cos()'', ''tan()'', ''asin()'', ''acos()''
...
 
-	The <dfn lt="sin()">sin(A)</dfn>, <dfn lt="cos(A)">cos()</dfn>, and <dfn lt="tan()">tan(A)</dfn> functions
+	The <dfn lt="sin()">sin(A)</dfn>, <dfn lt="cos()">cos(A)</dfn>, and <dfn lt="tan()">tan(A)</dfn> functions
 	all contain a single [=calculation=]
 	which must resolve to either a <<number>>
 	or an <<angle>>,
 	and compute their corresponding function
 	by interpreting the result of their argument as radians.
-	(That is, ''sin(45deg)'', ''sin(.125)'', and ''sin(3.14159 / 4 * 1rad)''
+	(That is, ''sin(45deg)'', ''sin(.125turn)'', and ''sin(3.14159 / 4 * 1rad)''
```
tabatkins: Whoops, indeed. Mind switching the third calculation to omit the rad, so e have an example of passing in raw numbers?

css-values-4/Overview.bs
```diff
@@ -2499,21 +2499,23 @@ Trigonometric Functions: ''sin()'', ''cos()'', ''tan()'', ''asin()'', ''acos()''
...
-	''atan2(1, -1)'', corresponding to the point (-1, 1) in the NW of the plane,
+	but it gives a better answer when the point in question includes negative components on one dimension.
+	''atan2(1, -1)'', corresponding to the point (-1, 1),
 	returns ''135deg'',
-	distinct from ''atan2(-1, 1)'', corresponding to the point (1, -1) in the SE of the plane,
+	distinct from ''atan2(-1, 1)'', corresponding to the point (1, -1),
 	which returns ''-45deg''.
-	However, as ''1 / -1'' and ''-1 / 1'' both resolve to ''-1'',
-	''atan()'' returns the same ''-45deg'' for both points.
+	In contrast, as ''atan(1 / -1)'' and ''atan(-1 / 1)'' both return''-45deg'',
```
tabatkins: Remove "as"

css-values-4/Overview.bs
```diff
@@ -2499,21 +2499,23 @@ Trigonometric Functions: ''sin()'', ''cos()'', ''tan()'', ''asin()'', ''acos()''
...
+	or else the function is invalid.
+	The function returns the <<angle>>
 	between the positive X-axis and the point (B,A).
 	The returned angle must be normalized to the interval (''-180deg'', ''180deg'']
 	(that is, greater than ''-180deg'', and less than or equal to ''180deg'').
 
 	Note: ''atan2(Y, X)'' is <em>generally</em> equivalent to ''atan(Y / X)'',
-	but it gives a better answer when the point in question is in the second (NW) or third (SW) quadrants of the plane.
-	''atan2(1, -1)'', corresponding to the point (-1, 1) in the NW of the plane,
+	but it gives a better answer when the point in question includes negative components on one dimension.
```
tabatkins: Not quite right; fourth quadrant (SE) gives the same results in both functions, despite one negative value.

AmeliaBR: Ok, new wording is "may include negative components". Even if sometimes the value is the same, it is still a "better answer" because you are confident of what it represents.

tabatkins:
Thanks for the fixes!

PR #3772, opened 2019-03-28
[css-conditional-3] @supports is not at risk.

emilio:
There are multiple interoperable implementations of it.


svgeesus:
[yes, clearly](https://test.csswg.org/harness/results/css-conditional-3_dev/grouped/section/6/)

PR #3773, opened 2019-03-28
[selectors] Export the definition of unknown -webkit- pseudo-elements.

emilio:
I want to reference it from css-conditional, in order to define the @supports
selector function.


tabatkins:
Please feel free to just go ahead and commit this kind of thing in the future. ^_^

PR #3774, opened 2019-03-28
[css-conditional] Define @supports selector function.

### LINKED ISSUE #3207: [css-conditional-4] selector() feature query function for selector support
I think one feature we should add to Conditional Rules Level 4 is a feature query function to test for selector support, presumably called `selector()`.

I've started a [draft of an explainer for this](https://github.com/dbaron/css-supports-functions/blob/master/explainer.md), from which I quote the relevant section:

-----

Having feature queries based on selectors is probably the highest priority addition, both because it seems the most likely to be important in the near future, and because it is the simplest.  It has come up recently because CSS authors using the features in the [CSS Scrollbars Module Level 1](https://drafts.csswg.org/css-scrollbars-1/) may want to query for support of those features, and may also want to query for support of the `::-webkit-scrollbar` pseudo-element.

I believe a possible syntax for such a feature was discussed at one point early in the development of `@supports`, but I can't find meeting minutes of the discussion.  I think the best way to extend the syntax is using a `selector()` function, as in:

```css
@supports (scrollbar-width: thin) or selector(::-webkit-scrollbar) {
  #chatwindow { overflow: scroll; scrollbar-width: thin }
  #chatwindow::-webkit-scrollbar { width: 5px }
}
```

It is worth noting that this usage requires that the `selector()` feature query function *not* copy the quirks that are present in selector parsing, where all `::-webkit-*` pseudo-elements behave as though they're supported (and allow the selector to parse).  Instead, inside of the `selector()` feature query function, this quirk should not be present, and a `selector()` function containing a `::-webkit-*` pseudo-element should only be true if the pseudo-element is *actually* supported.

Other than this, the function should work just as the tests for support/validity work in the [rules on handling invalid selectors](https://drafts.csswg.org/selectors-4/#invalid).

emilio:
Depends on #3773.

Closes #3207.

### DIFF
--- a/css-conditional-4/Overview.bs
+++ b/css-conditional-4/Overview.bs
@@ -27,6 +27,28 @@ and have not yet been copied here.
 
 Issue: In the future, copy the contents of [[css3-conditional]] into this document.
 
+<h2 id="at-supports-ext">Extensions to the ''@supports'' rule</h2>
+
+This level of the specification extends the <<supports-feature>> syntax as follows:
+
+<pre class="prod def" nohighlight>
+	<dfn>&lt;supports-feature></dfn> = <<supports-selector-fn>> | <<supports-decl>>
+	<dfn>&lt;supports-selector-fn></dfn> = selector( <<complex-selector>> )
+</pre>
+
+: <<supports-selector-fn>>
+::
+	The result is true if the UA
+	<a href="#dfn-support-selector">supports the selector</a>
+	provided as an argument to the function.
+
+<h3 id="support-definition-ext">Extensions to the definition of support</h3>
+
+A CSS processor is considered to
+<dfn export for=CSS id="dfn-support-selector">support a CSS selector</dfn>
+if it accepts that selector (rather than discarding it as a
+parse error), and that selector doesn't contain
+<a>unknown -webkit- pseudo-elements</a>.
 
 <h2 class=no-num id="acknowledgments">Acknowledgments</h2>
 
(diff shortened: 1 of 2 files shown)


emilio:
@dbaron I've put it in conditional-3 since there's no conditional-4 spec yet, but let me know if you want me to just copy conditional-3 into conditional-4 before or something.

dbaron:
I don't think it should go in css-conditional-3; the other features there are very stable.

It probably belongs in css-conditional-4 -- the question is whether to start it as a delta specification or as a copy.  I think a delta spec is probably better for now (so we can maintain the features in conditional-3 in one place for now).

I'll look at starting that, and then you can revise this PR to put it there?

dbaron:
I created #3777; I think you should merge this on top of that once it lands.

emilio:
SGTM :)

emilio:
Ok, rebased on top of that (kept it as two different commits for convenience).

css-conditional-4/Overview.bs
```diff
@@ -27,6 +27,28 @@ and have not yet been copied here.
 
 Issue: In the future, copy the contents of [[css3-conditional]] into this document.
 
+<h2 id="at-supports-ext">Extensions to the ''@supports'' rule</h2>
+
+This level of the specification extends the <<supports-feature>> syntax as follows:
+
+<pre class="prod def" nohighlight>
+	<dfn>&lt;supports-feature></dfn> = <<supports-selector-fn>> | <<supports-decl>>
```
ExE-Boss: Shouldn’t `<<supports-decl>>` be first?

emilio: Doesn't really matter since they cannot conflict, right?

emilio:
I'd still like @dbaron to take a look in any case, since I want to make sure this is the right way to do the delta spec thing :)

css-conditional-4/Overview.bs
ExE-Boss: They shouldn’t, but I prefer it to be alphabetically sorted and/or appended because it looks nicer.

ExE-Boss:
I’ve updated https://github.com/mdn/data/pull/289 and https://github.com/mdn/kumascript/pull/1110 with information for this.

dbaron:
Oh, when I merged this I thought you wanted the two commits *not* squashed... but now I realize you wanted them squashed.  Oh well...

PR #3788, opened 2019-04-01
[selectors-4] Fixes according latest changes

lahmatiy:
- Removed `:drop()` from the at-risk feature list since removed from the spec
- Added `s` to `Ignored Vars`. Not sure its needed
- Added `s` to `<attr-modifier>` syntax


lahmatiy:

PR #3797, opened 2019-04-04
[css-easing-1] Reword to refer to easing functions and not assume they are applied to animations

### LINKED ISSUE #3205: [css-easing-1] Make text more generic (less specific to animations)
Splitting this off from issue #1577 which I closed because the spec renaming is done.

We still need to edit the spec so that it doesn't assume the transfer functions are being applied to animations, but recognizes that can also be used for, e.g. gradient easing etc.

birtles:
Closes #3205.


nongaum2112:
Re.... Call

PR #3808, opened 2019-04-05
Tweaks and fixes to css-nav-1

frivoal:
This is a follow up to https://github.com/w3c/csswg-drafts/pull/3732, https://github.com/w3c/csswg-drafts/pull/3755, and https://github.com/w3c/csswg-drafts/pull/3795.

It fixes a range of grammatical and editorial issues, and clarifies some ambiguities.


css-nav-1/Overview.bs
```diff
@@ -937,8 +943,12 @@ Additional <a>spatial navigation containers</a> can be created using the 'spatia
 Navigation</h3>
 
 <figure class="verbose non-normative">
-<figcaption>This figure is not normative. It gives an overview of the processing model further defined in this section.</figcaption>
-<object type="image/svg+xml" data="images/spatnav_processing_model_diagram.svg"></object>
+    <figcaption>
+        This figure is not normative.
+        It gives an overview of the processing model further defined in this section,
+        assuming that the 'spatial-navigation-action' property has its initial value of ''spatial-navigation-action/auto''.
```
jihyerish: :+1:

css-nav-1/Overview.bs
```diff
@@ -1131,105 +1155,114 @@ run the following steps:
...
-            unless it too is overlapped with another higher item at the same distance, recursively.    
+            unless it too is overlapped with another higher item at the same distance, recursively.
+
 </div>
 
-<div algorithm="to calculate the distance">
-To <dfn lt="calculating the distance | to calculate the distance">calculate the <var>distance</var></dfn>, the function is defined as below:
+<div algorithm="to find the shortest the distance">
+To <dfn lt="find the shortest distance">to find the shortest distance</dfn>
+between a <var>reference</var> and a <var>candidate</var>
```
jihyerish: Doesn't "reference" need to be "searchOrigin"?
I can guess "reference" is the same as "searchOrigin", but using a new variable may cause confusion.

css-nav-1/Overview.bs
```diff
@@ -1131,105 +1155,114 @@ run the following steps:
...
-The meaning of each factor in the function is as follows:
+The meaning of each term is as follows:
 
 <dl dfn-type="value" dfn-for="spatial-navigation-distance-function">
     <dt><var>euclidean</var>
-    <dd>The euclidean distance between <var>P1</var> and <var>P2</var>
+    <dd>
+        The euclidean distance between <var>P1</var> and <var>P2</var>
 
     <dt><var>displacement</var>
```
jihyerish: I wasn't sure that this "displacement" term is appropriate.
Does it make sense?

css-nav-1/Overview.bs
```diff
@@ -1131,105 +1155,114 @@ run the following steps:
...
+        <pre><code>
+            <var>displacement</var> = (absolute distance on the axis orthogonal to <var>dir</var> between <var>P1</var> and <var>P2</var> +
+                            <var>orthogonalBias</var>) *
+                           <var>orthogonalWeight</var>
+        </code></pre>
+
+    <dt><var>orthogonalBias</var>:
+    <dd>
+        * If the <var>dir</var> is {{SpatialNavigationDirection/left}} or {{SpatialNavigationDirection/right}},
+            the height of the axis-aligned bounding box of <var>reference</var> / 2
```
jihyerish: I'm just curious - I don't understand clearly about "axis-aligned bounding box of reference".
Is it different from "bounding box of reference"?

css-nav-1/Overview.bs
```diff
@@ -1131,105 +1155,114 @@ run the following steps:
...
-NOTE: The values of <var>alignWeight</var> and <var>orthogonalWeight</var> in the function are decided based on 
-        the collected <a href="https://wicg.github.io/spatial-navigation/tests/ux/list.html">UX test cases</a>. 
-        It aims to show the generally expected result in the real pages.
-        Everyone is welcome to contribute to the test cases!
+Note: This general formula was picked from several plausible alternatives,
+based on which one most often match intuition
+when used to select the best candidate
+in a series of <a href="https://wicg.github.io/spatial-navigation/tests/ux/list.html">UX test cases</a>.
+Similarly, the values of <var>alignWeight</var> and <var>orthogonalWeight</var>
+were also determined experimentatlly based on the same test cases.
```
jihyerish: experimentatlly -> experimentally

css-nav-1/Overview.bs
```diff
@@ -1131,105 +1155,114 @@ run the following steps:
...
+The resulting forumla is somewhat complicated,
```
jihyerish: forumla -> formula

css-nav-1/Overview.bs
```diff
@@ -1348,56 +1381,95 @@ Note: the 'spatial-navigation-contain' property is <a>at-risk</a>.
...
+
+        If the currently focused element is a [=scroll container=],
+        it is scrolled in the direction requested without changing which element is in focus,
+        regardless of the presence of focusable descendants.
+
+        Note: This means that spatial navigation can be used
+        to move the focus to a [=scroll container=] and to scroll it,
+        but not to move the focus to its descendants.
+        However, if the focus is moved to a descendant by some other mean
+        (such as pressing the <span class=key>Tab</span> key or using the <{Element/focus()}> method)
```
jihyerish: Broken link: <{Element/focus()}> -> <{{Element/focus()}}>

css-nav-1/Overview.bs
```diff
@@ -1348,56 +1381,95 @@ Note: the 'spatial-navigation-contain' property is <a>at-risk</a>.
...
-Applies to: <a>scroll container</a>
-Inherited: true
+Applies to: <a>scroll containers</a>
+Inherited: no
 </pre>
 
-When the user triggers spatial navigation in a given direction, the 'spatial-navigation-action' decides
-the proper spatial navigation behavior.
+When the focus is inside of a scroll container and the user triggers spatial navigation,
+it is somewhat ambiguous whether they are requesting that the focus be moved in that direction,
```
jihyerish: "focus be moved" -> "focus is moved"

css-nav-1/Overview.bs
```diff
@@ -1348,56 +1381,95 @@ Note: the 'spatial-navigation-contain' property is <a>at-risk</a>.
...
+or whether the document should be scrolled in that direction.
+By default, this is automatically determined,
+but this property allows the author decide between focusing or scrolling.
```
jihyerish: "allows the author decide" -> "allows the author to decide"

css-nav-1/Overview.bs
```diff
@@ -1348,56 +1381,95 @@ Note: the 'spatial-navigation-contain' property is <a>at-risk</a>.
...
+        Otherwise, the [=scroll container=] is scrolled in the direction requested.
 
     <dt><dfn>focus</dfn>
-    <dd>The directional input moves the focus to all candidates inside a scroll container. The container <a>cannot be scrolled manually</a>.
+    <dd>
+        The focus is moved to the nearest focusable element within the [=scroll container=],
+        regardless of whether it is visible.
+        If there are none,
+        the scroll container is <em>not</em> scrolled,
+        and the search continues up the tree instead.
```
jihyerish: "up the tree" seems a little bit ambiguous.
How about "up the ancestry chain" that is already used in the spec?
(In the overview section)

css-nav-1/Overview.bs
```diff
@@ -1348,56 +1381,95 @@ Note: the 'spatial-navigation-contain' property is <a>at-risk</a>.
...
-        then directional input only works for manually scrolling the container.
+        Note: If the ''spatial-navigation-action/focus'' value is given to 'spatial-navigation-action',
+        <a event>navnotarget</a> event occurs when there isn’t any visible candidate in the given direction within the viewport of
+        the <a>spatial navigation container</a> even if the container can be scrolled more.
 
-    Note: The 'scroll' value is <a>at-risk</a>. 
+    <dt><dfn>scroll</dfn>
+    <dd>
+        If the currently focused element is not itself a [=scroll container=],
+        this value on an ancestor [=scroll container=] has the same effect as ''spatial-navigation-action/auto''.
```
jihyerish: This can be misinterpreted that the ancestor in this case always behaves with 'auto' value.
Adding more detail about this is only for the default case would be helpful.

css-nav-1/Overview.bs
```diff
@@ -1348,56 +1381,95 @@ Note: the 'spatial-navigation-contain' property is <a>at-risk</a>.
...
+        However, if the focus is moved to a descendant by some other mean
+        (such as pressing the <span class=key>Tab</span> key or using the <{Element/focus()}> method)
+        spatial navigation can be used to move the focus to other focusable descendants.
+
+        Note: The ''spatial-navigation-ation/scroll'' value is <a>at-risk</a>.
 </dl>
 
+Note: Earlier version of this specification did not offer a declarative way
+to opt into the behavior defined by ''spatial-navigation-action/focus'',
+and instead offered an cancellable event that would be fired before scrolling,
```
jihyerish: an cancellable -> a cancellable

css-nav-1/Overview.bs
```diff
@@ -1131,105 +1155,114 @@ run the following steps:
...
             In that case, return that item instead,
-            unless it too is overlapped with another higher item at the same distance, recursively.    
+            unless it too is overlapped with another higher item at the same distance, recursively.
+
 </div>
 
-<div algorithm="to calculate the distance">
-To <dfn lt="calculating the distance | to calculate the distance">calculate the <var>distance</var></dfn>, the function is defined as below:
+<div algorithm="to find the shortest the distance">
+To <dfn lt="find the shortest distance">to find the shortest distance</dfn>
```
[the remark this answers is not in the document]
frivoal: You're right, fixing it.

css-nav-1/Overview.bs
frivoal: It doesn't need to be. Algorithms work like functions. They have their local arguments, so this is something like
```
let searchOrigin = ...;
let dir = ...;
let candidates = ...;
candidates.map( candidate => find_the_shortest_distance(searchOrigin, candidate, dir) );
...
function find_the_shortest_distance(reference, candidate, dir) {.....}
```

css-nav-1/Overview.bs
frivoal: Not sure either, but it is clearly defined and used, and is not exposed through any API. If a native speaker finds a better word later, we can always change it.

css-nav-1/Overview.bs
frivoal: I mean if the box is rotated, we're not talking about the rotated bounding box, but the one that is aligned with the horizontal and vertical axises.

css-nav-1/Overview.bs
frivoal: Right. Thanks.

css-nav-1/Overview.bs
frivoal: thanks

css-nav-1/Overview.bs
frivoal: No, actually this one is correct: https://en.wikipedia.org/wiki/English_subjunctive

css-nav-1/Overview.bs
jihyerish: Oh, yes, on second thought, using a separate variable name for each algorithm is clearer than before.
I agree that this is good as it is.

css-nav-1/Overview.bs
frivoal: thanks

css-nav-1/Overview.bs
frivoal: good idea.

css-nav-1/Overview.bs
jihyerish: :+1:

css-nav-1/Overview.bs
jihyerish: Oh, I was confused. Thanks!

css-nav-1/Overview.bs
jihyerish: Oh, now I understand that. Thanks!

css-nav-1/Overview.bs
frivoal: Do you have a better suggestion? Otherwise, there's all the details in section 8.2, so I think it is OK.

css-nav-1/Overview.bs
frivoal: thanks

css-nav-1/Overview.bs
jihyerish: You're right. I didn't connect to that section. I agree with you.

PR #3823, opened 2019-04-12
[css-values-3] link to a test

TalbotG:


dontcallmedom:

ewilligers:
Why not use the same approach as css-contain-1/css-flexbox-1/css-multicol-1?
They use `WPT Path Prefix:` and don't have `WPT Display: inline`.

frivoal:
I was with @TalbotG when making this commit. `WPT Display: inline` was committed accidentally, we only meant to use it locally for testing. Sorry about that.

As for `WPT Path Prefix:` it triggers warnings / errors when there are test files in the directory you point to that you have not yet added to the spec. That's useful when you're done adding all / most tests and want to be informed about what's missing and what's just been added, but until then, it's mostly noise.

PR #3833, opened 2019-04-16
[web-animations-1] Define animation replacing behavior

### LINKED ISSUE #3689: [web-animations-1] Alternative to FillAnimation: replace events
In #3210 and #2054 I have been working on a proposal to allow filling animations to not leak memory. I have [prepared spec text](https://github.com/w3c/csswg-drafts/compare/master...birtles:fill-animation) for that and implemented (but not yet landed) it in Firefox but it is quite complex. This is an alternative approach that might be simpler.

Summary
---------

Proposal: automatically cancel any completely replaced filling script animations and fire a 'replace' event that includes their animated style so the author can manually set that on the element if they wish to preserve the animation's effect:

e.g. instead of

```js
left.addEventListener('click', () => {
  target.animate(
    { transform: 'translateX(-80px)', composite: 'add' },
    { duration: 1000, easing: 'ease', fill: 'forwards' }
  );
});
```
[(Live example)](https://codepen.io/birtles/pen/jJbNmo)

The author would be required to write:

```js
left.addEventListener('click', () => {
  const animation = target.animate(
    { transform: 'translateX(-80px)', composite: 'add' },
    { duration: 1000, easing: 'ease', fill: 'forwards' }
  );
  animation.onreplace = evt => {
    target.style.transform = evt.computedStyle.transform;
  };
});
```

When additive animations are not in use, the authors would typically be free to ignore the event unless they planned to cancel the overriding animation.

Proposal
--------

Define a forever-filling animation as:

*   The existence of the animation is _not_ prescribed by markup.
    That is, it is _not_ a CSS animation with an owning element, nor a CSS transition with an owning element.
*   The animation's play state is finished.
*   The animation is associated with a monotonically increasing timeline.
*   The animation has an associated target effect.
*   The target effect associated with the animation is in effect (i.e. it has the appropriate fill mode).

Any time a forever-filling animation finishes, run the following steps:

1.  Look up the effect stack for the target element of the animation that finished.
    (Once we introduce `GroupEffect`s, this will involve visiting all target elements.)

2.  Let _replaced animations_ be an empty set.

3.  Proceeding from the bottom of the effect stack[1], for each forever-filling animation, _animation_, where _all_ the target properties (once resolved to physical longhand properties) are also specified by one or more forever-filling animations that have a _greater_ composite order:

    1.  Calculate the computed value of each of _animation_'s target effect's target properties at its current position in the effect stack.

    2.  Add (_animation_, _computed values_) to _replaced animations_.

4.  For each pair (_animation_, _computed values_) in _replaced animations_:

    1.  Cancel _animation_ (but don't dispatch a 'cancel' event? See below).

    2.  Create an `AnimationReplaceEvent`, _replaceEvent_.

        (I'm not sure if we should re-use `AnimationPlaybackEvent` or not here.)

    3.  Set _replaceEvent_’s `type` attribute to 'replace'.

    4.  Set _replaceEvent_’s `computedStyle` attribute to _computed values_.

        (For implementations that implement CSS Typed OM, there is also
        a `computedStyleMap` attribute.)

    5.  Set _replaceEvent_’s `timelineTime` attribute to the current time of the timeline with which _animation_ is associated. If _animation_ is not associated with a timeline, or the timeline is inactive, let timelineTime be null.

    6.  If _animation_ has a document for timing, then append _replaceEvent_ to its document for timing's pending animation event queue along with its target, animation. For the scheduled event time, use the result of converting the most recently finished overriding animation’s target effect end to an origin-relative time.

        Otherwise, queue a task to dispatch _replaceEvent_ at animation. The task source for this task is the DOM manipulation task source.

[1] Obviously no one would actually implement this as proceeding from the bottom of the stack. You'd almost certainly want to proceed from the top and go down.  The point here is just to ensure that the replaced animations and hence the replace events are dispatched in order from lower in stack to higher.

Since replace events are dispatched in composite order, script can blindly just set the computed style on the target element and wait for subsequent events to override it.

Given that replace events are typically queued as part of updating timing and run before updating the rendering, there should be no visual glitch as a result of updating style in this manner.

In order to generate the replace events an additional style flush will typically be required to compute the correct computed style.  Implementations might want to avoid doing this when there are no replace event listeners registered. Such an optimization could possibly produce Web-observable changes with regards to dispatching transitions so the behavior with regard to style change events might need to be specified.

IDL
---

```webidl
partial interface Animation {
    attribute EventHandler onreplace;
}

[Exposed=Window,
 Constructor (DOMString type, optional AnimationReplaceEvent eventInitDict)]
interface AnimationReplaceEvent : Event {
    readonly attribute double? timelineTime;
    readonly attribute CSSStyleDeclaration computedStyle;
    readonly attribute StylePropertyMapReadOnly computedStyleMap;
};
dictionary AnimationReplaceEventInit : EventInit {
    double? timelineTime = null;
};
```

Relationship to finish events
-----------------------------

Finish events sort before replace events. It is the finishing of a higher-priority animation that triggers the replacing.

For an animation that is replaced as soon as it finishes, the finish event still fires first since logically it finishes before it is replaced.

Relationship to the `finished` Promise
--------------------------------------

Since finishing happens before replacing, the `finished` promise will be resolved and hence will not be rejected when an animation is subsequently canceled by being replaced.

Relationship to cancel events
-----------------------------

I suspect cancel events should _not_ be fired even though we effectively cancel these animations. Dispatching cancel events will likely confuse authors and might break existing content. Rather, we should consider replacing a special kind of canceling; neutering perhaps.

Side effects
------------

Replace events give the author the computed animated style for the effect that is being canceled so that they can apply it directly. However, since the returned value will be comprised of computed values for longhand (and probably physical-only) properties, any context-sensitive values (`em` units, CSS variables(?) etc.) will be lost. Furthermore, since CSS cannot yet represent additive values in static contexts, any additive behavior will also be lost.

These side effects should at least be somewhat apparent from the naming (`computedStyle`) and steps taken to achieve the effect.

Compatibility
-------------

Currently only effects that fully replace the underlying value are shipped so the only situation where this proposal would produce an observable change is if an overriding forever-filling animation is canceled while there is an underlying forever-filling animation.  Under this proposal, the underlying animation would have been canceled so the result would differ.

I don't expect this is common but usage of the cancel() method is [surprisingly high (3% of page loads!)](https://www.chromestatus.com/metrics/feature/timeline/popularity/699)

Issues
------

Ideally it would be good to offer a way for authors to opt-out of the replacing behavior.

For example, if, in the replace event callback, the author could re-instate the animation to its filling state, they could avoid the side effects mentioned above. This might produce a situation that effectively leaks memory, but it would be at the author's explicit direction much as if they created new Elements ad infinitum (and unlike the case where they fire off filling animations without necessarily being aware the animations will never disappear).

I don't have any idea for what a suitable API for reinstating such an Animation would be, however.

birtles:
This fixes #3689.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3762,8 +3762,8 @@ if <em>all</em> of the following conditions are true:
 
 *   The existence of the [=animation=] is <em>not</em> prescribed by markup.
     That is, it is <em>not</em>
-    a CSS animation with an [=owning element (animation)|owning element=], or
-    a CSS transition with an [=owning element (transition)|owning element=].
+    a CSS animation with an [=owning element (animation)|owning element=],
+    nor a CSS transition with an [=owning element (transition)|owning element=].
 *   The [=animation=]'s [=play state=] is [=finished play state|finished=].
 *   The [=animation=]'s [=replace state=] is <em>not</em>
     [=removed replace state|removed=].
@@ -3789,7 +3789,7 @@ then for every [=animation=], |animation|, that:
     [=associated with an animation|associated=] with |animation|,
     an [=animation effect=] associated with a [=replaceable=] [=animation=]
     with a higher [=composite order=] than |animation|
-    that includes the same [=target property=].
+    that includes the same [=target property=]
 
 perform the following steps:
 
@@ -3811,9 +3811,11 @@ perform the following steps:
 1.   If |animation| has a [=document for timing=], then append
      |removeEvent| to its [=document for timing=]'s
      [=pending animation event queue=] along with its target, |animation|.
-     For the [=scheduled event time=], use the result of <a lt="animation
-     time to origin-relative time">converting</a>
-     |animation|'s [=target effect end=] to an origin-relative time.
+     For the [=scheduled event time=], use the result of
+     applying the procedure to convert
+     [=timeline time to origin-relative time=] to
+     the [=timeline current time|current time=] of the [=timeline=]
+     with which |animation| is associated.
 
      Otherwise, [=queue a task=] to [=dispatch=] |removeEvent| at
      |animation|.
@@ -4112,7 +4114,7 @@ interface Animation : EventTarget {
 :   <dfn method for=Animation lt='persist()'>void persist()</dfn>
 ::  Sets this animation's [=replace state=] to
     [=persisted replace state|persisted=].
-:   <dfn method for=Animation lt='commitStyles()'>void commitStyles()</dfn>
+:   <dfn method for=Animation lt='commitStyles()'>void commitStyles(properties)</dfn>
 ::  Writes the current [=effect values=]
     produced by this animation's [=animation effects=]
     to their corresponding [=target elements=]' inline style
@@ -4123,8 +4125,8 @@ interface Animation : EventTarget {
     <div class=note>
 
       This is useful for retaining the effect of an animation after it has been
-      removed (see [[#removing-replaced-animations]]) or canceled without
-      retaining the actual animation.
+      removed (see [[#removing-replaced-animations]]) without retaining the
+      actual animation.
 
       Note that the values committed are the <em>computed</em> values
       produced by the [=animation effects=] at the time when this method is
@@ -4139,6 +4141,10 @@ interface Animation : EventTarget {
       the {{Animation/persist()}} method may be used,
       but note that doing so will mean the animation continues to consume
       resources.
+
+      Also note that as with other methods defined on this interface,
+      calling this method does <em>not</em> trigger a [=style change event=]
+      (see [[#model-liveness]]).
 
     </div>
 
@@ -4155,7 +4161,7 @@ interface Animation : EventTarget {
 </div>
 
 To <dfn>commit computed styles</dfn> for an [=animation=],
-|animation|, given an optional list of property names, |properties|:
+|animation|, given an optional list of CSS property names, |properties|:
 
 1.   Let |longhandProperties| be an empty [=ordered set|set=] of
      CSS property names.
@@ -4168,13 +4174,17 @@ To <dfn>commit computed styles</dfn> for an [=animation=],
 
      1.   For each CSS property name, |propertyName|, in |properties|:
 
-          1.   If |propertyName| corresponds to a CSS shorthand property name,
-               <a for=set>append</a> each longhand property
+          <dl class=switch>
+
+          :    If |propertyName| corresponds to a CSS shorthand property name,
+          ::   <a for=set>append</a> each longhand property
                that |propertyName| maps to,
                to |longhandProperties|.
 
-          1.   Otherwise,
-               <a for=set>append</a> |propertyName| to |longhandProperties|.
+          :    Otherwise,
+          ::   <a for=set>append</a> |propertyName| to |longhandProperties|.
+
+          </dl>
 
 1.   Let |targets| be the [=ordered set|set=] of all [=target elements=]
      for [=animation effects=] [=associated with an animation|associated=]
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2077,7 +2077,7 @@ is absent or has no observable result.
 ::  Queued whenever an animation enters the <a>idle play state</a> from
     another state. Creating a new <a>animation</a> that is initially
     idle does <em>not</em> generate a new <a>cancel event</a>.
-:   <dfn lt="replace event">replace</dfn>
+:   <dfn lt="remove event">remove</dfn>
 ::  Queued whenever an animation is automatically removed.
     See [[#replacing-animations]].
 
@@ -3753,7 +3753,7 @@ The [=animation effects=] of an [=animation=]
 whose [=replace state=] is [=removed replace state|removed=]
 are not included in the [=effect stacks=] of their [=target properties=].
 
-<h4 id="performing-replacement">Performing replacement</h4>
+<h4 id="removing-replaced-animations">Removing replaced animations</h4>
 
 An [=animation=] is
 <dfn lt="replaceable animation" local-lt="replaceable">replaceable</dfn>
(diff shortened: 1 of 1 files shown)


birtles:
This still needs work:

* There are three issues in the PR that should probably be addressed before merging
* The wording for the commit styles procedure is very wordy and could almost certainly be improved

I've tried to make the descriptions very general so that they "just work" when we introduce group effects. That's at least partly why some of them are a bit convoluted. (In my experience, the implementation for this is simpler than the spec text.)

For now, though, hopefully this is ready enough for feedback.

birtles:
I've made all the outstanding changes I had in mind for this PR so it should be ready to review now.

birtles:
@graouts @stephenmcgruer @flackr review ping?

web-animations-1/Overview.bs
```diff
@@ -3960,9 +4111,164 @@ interface Animation : EventTarget {
...
+               whose [=target element=] is |target| and which include
+               |property| as a [=target property=] to |partialEffectStack|.
+
+          1.   Remove from |partialEffectStack| any [=animation effects=]
+               whose [=associated with an animation|associated=] [=animation=]
+               has a higher [=composite order=] than |animation|.
+
+          1.   Let |effect value| be the result of calculating the result of
+               |partialEffectStack| for |property| using |target|'s computed
+               style (see [[#calculating-the-result-of-an-effect-stack]]).
```
flackr: Can this still result in growing lists? I.e. if we have composite: 'add' transforms might this result in an ever growing transform list or is the effect value the transform matrix? What about filter lists?

web-animations-1/Overview.bs
```diff
@@ -3682,10 +3706,129 @@ as follows:
...
+     applying the procedure to convert
+     [=timeline time to origin-relative time=] to
+     the [=timeline current time|current time=] of the [=timeline=]
+     with which |animation| is associated.
+
+     Otherwise, [=queue a task=] to [=dispatch=] |removeEvent| at
+     |animation|.
+     The task source for this task is the [=DOM manipulation task
+     source=].
+
```
flackr: Should the animation then be removed? Is the replace state change sufficient to imply the animations is removed?

web-animations-1/Overview.bs
```diff
@@ -3960,9 +4111,164 @@ interface Animation : EventTarget {
     method unpauses the animation and, if the animation has already finished
     playing in the reversed direction, seeks to the start of the <a>target
     effect</a>.
+:   <dfn method for=Animation lt='persist()'>void persist()</dfn>
+::  Sets this animation's [=replace state=] to
+    [=persisted replace state|persisted=].
+:   <dfn method for=Animation lt='commitStyles()'>void commitStyles(properties)</dfn>
```
flackr: Do you expect that only applying some of the animation styles but not all of them will be a common use case? Might it make sense to start with an API where all properties are applied and allow for upgrading in the future?

web-animations-1/Overview.bs
```diff
@@ -3960,9 +4111,164 @@ interface Animation : EventTarget {
...
+::  Writes the current [=effect values=]
+    produced by this animation's [=animation effects=]
+    to their corresponding [=target elements=]' inline style
+    using the [=commit computed styles=] procedure
```
flackr: What's the motivation for adding a commit mechanism over a way to retrieve the animation's effective output? I haven't fully thought through the implications but it seems a simpler API to simply pass a dictionary of values with the replace event.

flackr:
I'm a little unclear what the state of a removed animation is, i.e. can it be replayed? retargeted? Does it's replace state change if it is?

web-animations-1/Overview.bs
birtles: Yes, setting the replace state to removed at the beginning of the procedure is sufficient for it to stop being returned from `getAnimations()` and stop contributing to the effect stack.

web-animations-1/Overview.bs
birtles: Yes, I'm not sure. I added this at @graouts' suggestion but I'm fine with leaving it off and adding it later if that's preferable. (And when I implemented it I made this feature a separate patch to prove it could be incrementally added.)

web-animations-1/Overview.bs
birtles: I believe the motivation was simply to avoid authors having to add boilerplate code. Unfortunately, passing a dictionary of values has two problems:

1. It doesn't extend easily to `GroupEffect`s. We'd need to add a `Map` or something to the event instead, to map from target elements to computed style.

2. It requires us to either _always_ flush style in order to produce the correct computed style for the current frame, or never flush style and return potentially out-of-date style (see [examples here](https://github.com/w3c/csswg-drafts/issues/3689#issuecomment-486475363)).

    At least always flushing style means we only do it once, regardless of how many animations are being removed in the current frame. Perhaps we could even spec that it only happens if there is at least one `remove` event listener, I'm not sure. Flushing style at that point in the frame might be a bit odd though -- triggering transitions before we run `requestAnimationFrame` callbacks.

    The advantage of the `commitStyles()` approach is that, even assuming we make it flush style, the author is in charge of calling it so it won't flush styles unless they use it and the side effects are more understandable. The disadvantage is that it means that if several animations are removed in the same frame, then calling `commitStyles()` several times will flush and dirty style multiple times.

Having a means to get the computed style of an `Animation` / `AnimationEffect` is nice because it might be more generally useful but it doesn't really help avoid the [boilerplate @graouts was concerned about](https://github.com/w3c/csswg-drafts/issues/3689#issuecomment-478049033).

web-animations-1/Overview.bs
birtles: Yes, it can. The computed value for transforms is a list, it just happens to be [serialized as a single matrix value](https://drafts.csswg.org/css-transforms/#serialization-of-the-computed-value). For filters too, there is a similar issue.

birtles:
> I'm a little unclear what the state of a removed animation is, i.e. can it be replayed? retargeted? Does it's replace state change if it is?

The replace state is part of the animation model so it's entirely orthogonal to any timing. All it means is that:
1. Its effect(s) won't contribute to the effect stack(s) of their target elements.
1. It won't be returned from `getAnimations()`

So:
* A removed animation can be replayed and will dispatch the appropriate events etc. It just won't have any affect on style.
* It can be retargeted.
* It won't change replace state when it is retargeted. The only thing that can change its replace state once it is `removed` is calling `persist()`.

I have a few tests ([1](https://hg.mozilla.org/try/rev/315ce410dd99a435c29b3788ca92e74bd33767d1#l5.2), [2](https://hg.mozilla.org/try/rev/f2477993a04a9c0056d702a44e84f64eed528f37#l6.3), [3](https://hg.mozilla.org/try/rev/3869f7e3aa65fd7adc11610c17e24bc3864cae39#l7.3)) for these things that may or may not be useful in understanding the behavior.

web-animations-1/Overview.bs
graouts: I think we should start without any arguments and consider adding a list of properties in a future version if we feel it is warranted.

birtles:
I've dropped the `properties` parameter from `commitStyles` for now.

PR #3851, opened 2019-04-19
[css-transitions-2][css-animations-2] Specify composite order for ::marker pseudo elements

birtles:

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -165,6 +165,7 @@ last) as follows:
     With regard to pseudo-elements, the sort order is as follows:
 
     *  element
+    *  ::marker
     *  ::before
     *  ::after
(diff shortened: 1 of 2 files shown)


birtles:
@MatsPalmgren I've specced this to match what you implemented in Gecko but I was wondering if there was any particular reason you chose that order? Is it because typically ::marker pseudo elements appear before ::before pseudo elements?

birtles:
(I'm currently working on web-platform-tests for this which I will merge via mozilla-central.)

MatsPalmgren:
@birtles No particular reason, since I wasn't really aware there was any particular order to pay attention to when I wrote that code.  But yes, `::marker` is considered to be before `::before` in the document order.

birtles:
Great, thank you!

PR #3868, opened 2019-04-23
[css-text] force-hang end-of-line pre-wrap spaces before soft wraps

### LINKED ISSUE #3440: [css-text][css-sizing] When to/not to include preserved trailing spaces
We found browsers are inconsistent on the topic on `pre-wrap`. `pre` always include trailing spaces.

Feature|Blink|Edge|Gecko|WebKit
---|---|---|---|---
`text-align: start`|Don't include|Don't include|Don't include|Don't include
`text-align: end / center`|Include|Include|Include|Don't include
`text-align: justify`|If !wrap|Do not justify|Do not justify|Do not justify
`min-content`|Don't include|Don't include|Include|Don't include
`max-content`|Include|Include|Include|Include

Can we define these? Are there more cases we should cover?

Note: table updated on June 2, 2019. `text-align: left` added, and some values updated. Changes maybe from inaccurate old tests, or browsers updated, not sure.

frivoal:
Closes #3440


frivoal:
@fantasai Can you review this pull request? It is supposed to implement the changes resolved in #3440, but I want to make sure I'm not misunderstanding anything.

@kojiishi Can you also review this? In particular, please check the example, to make sure that what this does is in line with your understanding. Since reading the example in diff form is pretty annoying, here's a screenshot of what it looks like:
Screenshot 2019-07-02 22 42 18

frivoal:
Updated examples to show the same situations as in #4095 (only example 8 gives a different result):
Screen Shot 2019-07-27 at 14 18 30
Screen Shot 2019-07-27 at 14 18 35
Screen Shot 2019-07-27 at 14 18 40

frivoal:
Closing. This was rejected, as #4095 was accepted instead.

PR #3888, opened 2019-05-02
[resize-observer-1] Fix some observedBox terminology issues

dholbert:
Just fixing some typos.

This commit replaces 'specificBox' with 'observedBox' in several places (which is the actual defined term for this concept in the spec).  This also fixes one mention of "bounding-box" which really meant to say "border-box".


dholbert:
CC @atotic

gregwhitworth:
Thanks @dholbert

PR #3891, opened 2019-05-04
[css-size-adjust] text-size-adjust is non-negative

ewilligers:
Like other sizing properties, text-size-adjust does not
accept negative values.

### DIFF
--- a/css-size-adjust-1/Overview.bs
+++ b/css-size-adjust-1/Overview.bs
@@ -66,7 +66,7 @@ Module interactions {#placement}
 Values {#values}
 ----------------
 
-  This specification follows the <a href="https://www.w3.org/TR/CSS21/about.html#property-defs">CSS property definition conventions</a> from [[!CSS2]].
+  This specification follows the <a href="https://www.w3.org/TR/CSS21/about.html#property-defs">CSS property definition conventions</a> from [[!CSS21]].
   Value types not defined in this specification are defined in CSS Values & Units [[!CSS-VALUES-3]].
   Other CSS modules may expand the definitions of these value types.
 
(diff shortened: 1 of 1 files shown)


ewilligers:
Note also that in Firefox and Blink, 'none' computes to '100%'.

Perhaps the spec should change
Computed value: specfied keyword or percentage 
to
Computed value: specfied percentage, or 'auto'

emilio:
> Note also that in Firefox and Blink, 'none' computes to '100%'.

How are you determining this? I don't think Firefox supports `text-size-adjust: <percentage>` to begin with.

ewilligers:
My mistake. I guess I made the mistake because a [test](https://chromium-review.googlesource.com/c/chromium/src/+/1595202/1/third_party/blink/web_tests/external/wpt/css/css-size-adjust/parsing/text-size-adjust-computed.html) passes in Firefox, but that would be because Firefox doesn't support the property, so test_computed_value succeeds immediately.

tabatkins:
Lack of support causing a test to pass means the test is written badly - test_computed_value() should fail or throw in that case.

ewilligers:
> Lack of support causing a test to pass means the test is written badly - test_computed_value() should fail or throw in that case.

Test now changed: [chromium](https://chromium-review.googlesource.com/c/chromium/src/+/1609850) / [WPT](https://github.com/web-platform-tests/wpt/pull/16845)

ewilligers:
@dbaron please review

ewilligers:
@tantek please review.

This avoids the warning
> LINE ~69: The biblio refs [[CSS2]] and [[CSS21]] are both aliases of the same base reference [[CSS2]]. Please choose one name and use it consistently.
and forbids negative values. Note the browsers already reject negative values:
https://wpt.fyi/results/css/css-size-adjust/parsing/text-size-adjust-invalid.html?label=master&label=experimental

ewilligers:
@litherum please review.
WPTs exist for [valid non-negative](https://wpt.fyi/results/css/css-size-adjust/parsing/text-size-adjust-valid.html?label=experimental&label=master&aligned) and [invalid negative](https://wpt.fyi/results/css/css-size-adjust/parsing/text-size-adjust-invalid.html?label=experimental&label=master&aligned) percentages.

Issue #2921, opened 2018-07-13
[css-values] Automatic parsing of value definitions


tidoust:
Context for this issue is that @dontcallmedom and I spent some time integrating CSS specs in the list of specs crawled by [Reffy](https//github.com/tidoust/reffy). Our goal was to extract and parse value definitions for CSS properties and descriptors from all CSS specs, first step that could then perhaps be used to detect potential anomalies, automate the creation of parsing tests, or create tools that list CSS properties (apart from the detection of a few anomalies which led to the issues I reported yesterday and today on individual specs, we haven't had time yet to look into using the result of this parsing).

This exercise was also meant as an occasion for us to take a deeper look at how CSS specs are written. It is quite possible that we misunderstood a few things, we're much more familiar with API specs in practice. Also, as opposed to API specs where the automatic extraction of IDL content allows to create tests, and actual stubs for implementation, the automatic extraction and parsing of value definitions of CSS properties may perhaps not be seen as an interesting goal or a priority, because that does not trigger major interop issues in practice.

That being said, taking for granted that the goal of the [Value Definition Syntax](https://drafts.csswg.org/css-values-4/#value-defs) is to ease the automatic parsing of values, we noted a few potential issues:

1. [Keyword values](https://drafts.csswg.org/css-values-4/#keywords) do not allow for some of the values actually used in specs. The syntax defines keyword values as identifiers which conform to the `<ident-token>` grammar. Unless we read that definition incorrectly, this does not allow keywords to start with a digit or with an `@`. Some values need that such as [`glyph-orientation-vertical`](https://drafts.csswg.org/css-writing-modes-4/#propdef-glyph-orientation-vertical), [`font-weight`](https://drafts.csswg.org/css-fonts-3/#font-weight-prop) (although value was replaced by `<number>` in Level 4), or [`<feature-type>`](https://drafts.csswg.org/css-fonts-4/#font-feature-values-syntax) (e.g. `@stylistic`).

[...]

Issue #355, opened 2016-07-28
[css-values] Define value syntax that limits <integer>, <number>, <length>, etc. to ranges


AmeliaBR:
Many properties only accept positive values.  Negative values are illegal, and are parse errors.  This is currently only defined in prose for the relevant property.  Examples: width, height, line-height.

In contrast, other properties have logical bounds, but values outside these bounds are parsed as normal and only clamped at used value time. Examples: opacity, numeric parameters in color functions.

I'm sure it would be useful for both authors and implementers if the property value syntax explicitly indicated whether negative values would be a parse error or not.  That requires new value types in [CSS Values & Units](https://drafts.csswg.org/css-values/).

I'm not sure if there are any other common "illegal value" restrictions currently defined in prose, but it's worth looking around to see if there are, and if they can also be standardized into the value definition syntax.

tabatkins:
Well, "positive" is only allowed for integers, not numbers or dimensions, because it constitutes an open range otherwise, which is informally disallowed.  We do use non-negative (sometimes paired with prose that floors it at some non-zero value), but that's harder to express easily.

AmeliaBR:
Mathematically, "positive" usually means zero or greater, which is what I was thinking of, although depending on how the CSS parser handles -0, you might need some extra text in there.

"non-zero positive integer" would be another possible type, if that's actually required anywhere.

frivoal:
> "non-zero positive integer" would be another possible type, if that's actually required anywhere.

It is: https://github.com/w3c/csswg-drafts/issues/341#issuecomment-234938653

tabatkins:
> Mathematically, "positive" usually means zero or greater, which is what I was thinking of, although depending on how the CSS parser handles -0, you might need some extra text in there.

That's not really true in my experience? Zero is neither positive nor negative, and there's a lot of stuff backing that up: the relatively common usage of "non-negative" in precise technical contexts, the implementations of sign() functions in various languages, the fact that +0 and -0 are mathematically equivalent, etc.  CSS has been pretty precise about this, because we've gotten bug reports about "why isn't 0 valid?" when we were imprecise and just used "positive".  At best, whether 0 is considered "positive" or not is ambiguous and context-sensitive.

AmeliaBR:
> At best, whether 0 is considered "positive" or not is ambiguous and context-sensitive.

[...]

PR #3894, opened 2019-05-06
Use bracket range notation in syntax grammar [css-values-3][css-values-4][css-backgrounds-3][css-backgrounds-4][css-multicol-1][css-multicol-2][css-flexbox-1][css-counter-styles-3]

AmeliaBR:
Partially addresses #355

Includes some minor fixes to the [definition of the notation](https://drafts.csswg.org/css-values-3/#numeric-ranges) @fantasai added to css-values-3.
The definition is then copied over to css-values-4 (which is what you get for just /css-values/ now).
The term "bracketed range notation" is exported for use in other specs, primarily to explain why I'm changing things.

This initial PR includes syntax edits to the specs listed as [Stable CR in the Current Work web page](https://www.w3.org/Style/CSS/current-work), and later levels of those same modules:

- CSS Backgrounds and Borders Level 3 & 4
- CSS Conditional Rules Level — no edits required
- CSS Multi-column Layout Level 1 & 2
- CSS Values and Units Level 3 & 4— edits described above
- CSS Flexible Box Layout Level 1
- CSS Cascading and Inheritance Level 3 — no edits required
- CSS Writing Modes Level 3 — no edits required, although see https://github.com/w3c/csswg-drafts/issues/2921#issuecomment-489494006
- CSS Counter Styles Level 3

See the individual commit messages for details of the changes. I've tried to match code style to each document.

Note that this currently creates some non-fatal bikeshed errors, see https://github.com/tabatkins/bikeshed/issues/1467
If that's going to take a while to get fixed, @tabatkins, it might be worth a separate PR just to merge the changes to CSS Values.

I'll be continuing to work through the remaining CR and WD specs, but that may be spread out over a few weeks, so no need to delay merging these edits waiting for an omnibus pull request. (That'll just increase the risk of merge conflicts!)


frivoal:
Reviewed multicol 1&2. The edits are correct, but I wonder: you both added the bracket range notation in the grammar **and** left the equivalent restrictions in prose. Shouldn't we drop the prose now that the grammar expresses the same thing unambiguously?

AmeliaBR:
@frivoal Because the bracket syntax is new, I thought it best to leave in the prose for clarification.

But maybe I should tidy up the prose as I go along, too?  There are a lot of different phrases used in different specs to describe these restrictions. I could switch them all to the simple "Negative values are not allowed." And maybe minimize duplication (e.g., only include that phrase once per property) when its unambiguous that it applies to multiple data types.

frivoal:
My take would be to drop it from the prose as soon as it's added to the grammar, and write new tests to make sure it isn't ignored. But that's just me. Your approach isn't wrong either.

Can we have a third opinion?

astearns:
I'm happy with redundancy - explaining the range in prose is another way to learn the new grammar.

AmeliaBR:
Ok. I'll rewrite the prose to be more consistent / less verbose, but will leave some prose in place.

AmeliaBR:
>  write new tests to make sure it isn't ignored.

For me, I'm less worried about the new grammar being ignored by implementations & more worried about it being inscrutable to authors.  I'm assuming that all these restrictions are already tested for any spec that's at stable CR, though I haven't been checking.

AmeliaBR:
Pushed a tidy-up of the prose restrictions for these specs. Also expanded the note in CSS Values:

<blockquote>
<p class="note" role="note"><span>Note:</span> At the time of writing,
	the <a data-link-type="dfn" href="#bracketed-range-notation" id="ref-for-bracketed-range-notation①">bracketed range notation</a> is new;
	thus in most CSS specifications
	any range limitations are described only in prose.
	This does not make them any less binding.
	Common prose restrictions include
	“Negative values are not allowed” or
	“Negative values are invalid”
	to indicate a <code>[0,&infin;]</code> range.
	For an <a class="production css" data-link-type="type" href="#integer-value" id="ref-for-integer-value④">&lt;integer&gt;</a>, “Values must be greater than 0”
	is equivalent to <span class="production" id="ref-for-integer-value⑤">&lt;integer [1, ∞]&gt;</span>.</p></blockquote>

fantasai:
Cherry-picked out (and tweaked) the css-values changes, so you'll want to drop those from the PR. Don't merge until we republish on /TR. :) Shouldn't take more than a week or two.

AmeliaBR:
Thanks @fantasai. I've merged your changes into the PR, so they won't get messed up when this PR gets merged into master.

fantasai:
@AmeliaBR Where you're tweaking the prose... I think using the term "invalid" is better because it more cleanly hooks into the parsing implications.

fantasai:
@AmeliaBR Would you mind updating to use "invalid" per https://github.com/w3c/csswg-drafts/pull/3894#issuecomment-516998671 ? css-values-3 has been republished with the range notation, so after that's fixed we should be OK to merge

svgeesus:
@AmeliaBR could you take a look at the three conflicts? It would be good to merge this PR.

AmeliaBR:
Thanks @tabatkins (it was on my to-do, but I'm still sick & with a backlog across many projects)

PR #3900, opened 2019-05-08
[cssom] Define legacy CSSStyleSheet methods as implemented in Blink.

### LINKED ISSUE #3814: [cssom] Figure out what to do with non-standard CSSStyleSheet methods in WebKit / Blink
WebKit / Blink implement some legacy Microsoft CSSOM functions on CSSStyleSheet, which if I'm not wrong come from Microsoft:

https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/css/css_style_sheet.idl?l=37-40&rcl=6dc816e543555790648ec5af1ca827431a7d5807

 * `rules` just calls `cssRules`.
 * `removeRule` just calls `deleteRule`.
 * `addRule` just string-concats a selector and a style block and calls `insertRule`.

It's not too hard to add them to Gecko, but `addRule` in particular seems rather hacky / inconsistent with the rest of the APIs. We haven't seen any compat issue due to not having them (but I just got bit by it when trying to test something).

Would there be any chance on collecting data and potentially removing them? Or is all hope lost / is not worth the churn and should I just put them in the spec?

emilio:
Fixes #3814.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -949,6 +949,39 @@ The <dfn method for=CSSStyleSheet>deleteRule(<var>index</var>)</dfn> method must
  <li><a>Remove a CSS rule</a> in the <a for=CSSStyleSheet>CSS rules</a> at <var>index</var>.
 </ol>
 
+#### Legacy CSSStyleSheet members #### {#legacy-css-style-sheet-members}
+
+<p class=advisement>
+  Note: These members are required for compatibility with existing sites.
+  Authors are encouraged to use the standard members defined above instead,
+  which are consistent with {{CSSGroupingRule}}.
+</p>
+
+<pre class=idl>
+partial interface CSSStyleSheet {
+  [SameObject] readonly attribute CSSRuleList rules;
+  long addRule(optional DOMString selector = "undefined", optional DOMString style = "undefined", optional unsigned long index);
+  void removeRule(optional unsigned long index = 0);
+};
+</pre>
+
+The <dfn attribute for=CSSStyleSheet>rules</dfn> attribute must follow the same steps as {{CSSStyleSheet/cssRules}}, and return the same object {{CSSStyleSheet/cssRules}} would return.
+
+The <dfn method for=CSSStyleSheet>removeRule(<var>index</var>)</dfn> method must run the same steps as {{CSSStyleSheet/deleteRule()}}.
+
+The <dfn method for=CSSStyleSheet>addRule(<var>selector</var>, <var>block</var>, <var>optionalIndex</var>)</dfn> method must run the following steps:
+
+<ol>
+ <li>Let <var>rule</var> be an empty string.
+ <li>Append <var>selector</var> to <var>rule</var>.
+ <li>Append <code>" { "</code> to <var>rule</var>.
+ <li>If <var>block</var> is not empty, append <var>block</var>, followed by a space, to <var>rule</var>.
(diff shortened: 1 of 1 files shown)


emilio:
WPT tests in https://bugzilla.mozilla.org/show_bug.cgi?id=1545823

cssom-1/Overview.bs
```diff
@@ -949,6 +953,25 @@ The <dfn method for=CSSStyleSheet>deleteRule(<var>index</var>)</dfn> method must
...
+
+The <dfn attribute for=CSSStyleSheet>removeRule(<var>index</var>)</dfn> method must run the same steps as {{CSSStyleSheet/deleteRule()}}.
+
+The <dfn method for=CSSStyleSheet>addRule(<var>selector</var>, <var>block</var>, <var>optionalIndex</var>)</dfn> method must run the following steps:
+
+<ol>
+ <li>Let <var>rule</var> be an empty string.
+ <li>Append <var>selector</var> to <var>rule</var>.
+ <li>Append <code>" { "</code> to <var>rule</var>.
+ <li>If <var>block</var> is not empty, append <var>block</var>, followed by an space, to <var>rule</var>.
```
bzbarsky: "a space"

cssom-1/Overview.bs
```diff
@@ -949,6 +953,25 @@ The <dfn method for=CSSStyleSheet>deleteRule(<var>index</var>)</dfn> method must
...
+ <li>Append <code>"}"</code> to <var>rule</var>
+ <li>Let <var>index</var> be <var>optionalIndex</var> if provided, or the number of <a for=CSSStyleSheet>CSS rules</a> in the stylesheet otherwise.
+ <li>Call {{CSSStyleSheet/insertRule()}}, with <var>rule</var> and <var>index</var> as arguments.
```
bzbarsky: Should return its return value, right?

emilio: No, it returns -1:

http://webkit.crisal.io/webkit/rev/6cc00baaf9835dc4abdf7b16e72cff60ecd7bb88/Source/WebCore/css/CSSStyleSheet.cpp#324

Fixed.

cssom-1/Overview.bs
emilio: Fixed.

cssom-1/Overview.bs
```diff
@@ -911,6 +911,10 @@ interface CSSStyleSheet : StyleSheet {
   [SameObject] readonly attribute CSSRuleList cssRules;
   unsigned long insertRule(CSSOMString rule, optional unsigned long index = 0);
   void deleteRule(unsigned long index);
+
+  [SameObject] readonly attribute CSSRuleList rules;
+  long addRule(optional DOMString selector = "undefined", optional DOMString style = "undefined", optional unsigned long index);
```
domenic: Wow. I guess in Blink this is done via https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/css/css_style_sheet.idl?q=css_style_sheet.idl&sq=package:chromium&g=0&l=1 which bottoms out at https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/bindings/IDLExtendedAttributes.md#DefaultValue_p

/cc @yukishiino as Blink bindings team member. I hope we can get rid of [DefaultValue] and prevent any future APIs from ending up this weird.

emilio:
Tweaked the text so that the header is properly nested under the previous section, and add some advisement text to discourage usage. Normative text is the same.

PR #3901, opened 2019-05-09
[web-animations-1] Add advice about not using indefinitely filling animations

birtles:

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2345,6 +2345,67 @@ The possible <a>fill modes</a> are:
 The normative definition of these modes is incorporated in the
 calculation of the <a>active time</a> in [[#calculating-the-active-time]].
 
+<div class=advisement>
+<p>
+Authors are discouraged from using [=fill modes=] to produce animations whose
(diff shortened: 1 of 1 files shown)


birtles:
> I wonder if we could get something added to https://developer.mozilla.org/en-US/docs/Web/API/EffectTiming/fill as well?

Sure! Will do. (I'll merge this first though so I can link to it.)

Also, I'm sure you're aware but just in case, MDN is publicly editable (you can login with GitHub) so feel free to edit away if you notice anything amiss.

birtles:
> I wonder if we could get something added to https://developer.mozilla.org/en-US/docs/Web/API/EffectTiming/fill as well?

Edited: https://developer.mozilla.org/en-US/docs/Web/API/EffectTiming/fill$compare?locale=en-US&to=1549696&from=1494907

stephenmcgruer:
> Also, I'm sure you're aware but just in case, MDN is publicly editable (you can login with GitHub) so feel free to edit away if you notice anything amiss.

I was not aware! Thanks for letting me know.

PR #3902, opened 2019-05-09
[web-animations-1] Add parameter for fetching all animations on a subtree

### LINKED ISSUE #2058: [web-animations-1] Add parameter to getAnimations for getting animations in descendants too
_From @birtles on March 17, 2016 0:48_

As discussed and minuted here: The ideas we came up with are here: https://docs.google.com/document/d/1lQbgJ0fZYGn1_qboj59G-MOILK33z4kmBdjgtSHF2gA/edit#heading=h.cioj28lsgoa2

A more concrete elaboration on that proposal is described here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1254761#c0

_Copied from original issue: w3c/web-animations#146_

birtles:
This fixes #2058.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -5648,10 +5648,15 @@ the {{Animatable}} interface mixin.
 interface mixin Animatable {
     Animation           animate (object? keyframes,
                                  optional (unrestricted double or KeyframeAnimationOptions) options);
-    sequence&lt;Animation&gt; getAnimations ();
+    sequence&lt;Animation&gt; getAnimations (optional GetAnimationsOptions options);
 };
+
 dictionary KeyframeAnimationOptions : KeyframeEffectOptions {
     DOMString id = "";
+};
+
+dictionary GetAnimationsOptions {
+  boolean subtree = false;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


web-animations-1/Overview.bs
```diff
@@ -5721,11 +5726,14 @@ animation.play();</pre>
...
-    sequence&lt;Animation&gt; getAnimations()</dfn>
+:   <dfn method for=Animatable lt="getAnimations(options)">
+    sequence&lt;Animation&gt; getAnimations(options)</dfn>
 ::  Returns the set of [=relevant=] {{Animation}} objects
     that contain at least one <a>animation effect</a> whose
-    <a>target element</a> is this object.
+    <a>target element</a> is this object, or,
+    if an {{Animatable/getAnimations(options)/options}} parameter
+    is passed with {{GetAnimationsOptions/subtree}} set to true,
+    an [=inclusive descendant=] of this object.
```
stephenmcgruer: Do we need to specify anything regarding shadow DOM, or does inclusive descendant automatically exclude them?

birtles: As I understand it, "inclusive descendant" automatically excludes them. To include them you need to use the [shadow including inclusive descendant](https://dom.spec.whatwg.org/#concept-shadow-including-inclusive-descendant) concept.

stephenmcgruer: Ack, thanks.

annevk:
So the idea is to not tackle shadow roots at all? Will you add tests to ensure nothing from shadow roots is returned?

birtles:
> So the idea is to not tackle shadow roots at all? Will you add tests to ensure nothing from shadow roots is returned?

Right. Yes, I need to update the implementation in Firefox to match the spec terms and I plan to add WPT at that point. I'll make sure to add one for shadow roots at the same time.

PR #3911, opened 2019-05-10
[css-shapes] link to closest-side definition

ewilligers:
Use ''closest-side'' syntax so the keyword is formatted as CSS syntax.
Use link-defaults to prevent definitions in css-images-3 or motion-1 from
being chosen by bikeshed.
Use `<a href="#closest-side">` to link to the definition in CSS Shapes.

Note:
If basic shapes including the closest-side keyword were defined inside a
property such as shape-outside, we could use the
''shape-outide/closest-side'' bikeshed syntax.


ewilligers:
@tabatkins please review

PR #3913, opened 2019-05-10
[css-display] link to css-position for 'fixed'

ewilligers:
This avoid the blikeshed warning

LINK ERROR: Multiple possible 'fixed' maybe refs.
Arbitrarily chose https://drafts.csswg.org/css-position-3/#valdef-position-fixed
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:css-backgrounds-3; type:value; text:fixed
spec:css-counter-styles-3; type:value; text:fixed
spec:css-position-3; type:value; text:fixed
spec:css-device-adapt-1; type:value; text:fixed

### DIFF
--- a/css-display-3/Overview.bs
+++ b/css-display-3/Overview.bs
@@ -29,6 +29,7 @@ spec:css2;
 spec: html;
 	type: dfn; text: rendered legend
 	type: element; text: a
+spec:css-position-3; type:value; text:fixed
 spec:css-pseudo-4; type:selector;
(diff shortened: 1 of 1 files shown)


fantasai:
This seems to have been fixed by specifying ''position/fixed'' in the source already.

ewilligers:
Fixed by https://github.com/w3c/csswg-drafts/commit/65ba881a95d5da281477ee22372101881865615d

PR #3914, opened 2019-05-10
[css-regions] Consistent references to CSS3-BREAK

ewilligers:
Avoid the bikeshed warning:
The biblio refs [[css-break-3]] and [[CSS3-BREAK]] are both aliases of the same base reference [[css-break-3]]. Please choose one name and use it consistently.


ewilligers:
@astearns please review

PR #3917, opened 2019-05-11
Link to CSS Values definition of ex units

ewilligers:
Link to css-values definition of ex units
instead of css-inline-3 leading-trim keyword ex.
(Specs already link to css-values for em.)

css-cascade-3 links to mediaqueries-4 instead
of css-cascade-4 for media-query.


ewilligers:
@tabatkins please review. 

An alternative would be for bikeshed to by default
link to the css-values definition of ex units
instead of css-inline-3 leading-trim keyword ex.

tabatkins:
I don't really like special-casing things like that. I'd rather have a linking error than silently link to the wrong thing. Thanks for the fixes!

PR #3918, opened 2019-05-11
[css-values] fix indentation

ewilligers:
Avoid bikeshed error report:
FATAL ERROR: Line 3251 isn't indented enough (needs 1 indent) to be valid Markdown:

Fix reference to `<<calc-value>>`


ewilligers:
@tabatkins please review

PR #3919, opened 2019-05-11
[css-inline] [css-scroll-anchoring] content area

ewilligers:
Select the css-box-3 definition of 'content area',
not the description in css-exclusions-1.

Avoid bikeshed message:
Multiple possible 'content area' dfn refs.


ewilligers:
@tabatkins please review

PR #3921, opened 2019-05-11
[css-exclusions] Link to definitions of 'none'

ewilligers:
We now link to the `float` and `wrap-through` definitions
of 'none', as appropriate.

Avoid bikeshed warning:
Ambiguous for-less link for 'none'


ewilligers:
@atanassov please review

PR #3929, opened 2019-05-14
[cssom] Move TreatNullAs to the correct position

saschanaz:
https://drafts.csswg.org/cssom-1/#the-cssstyledeclaration-interface

`[TreatNullAs=EmptyString]` should stick to IDL types instead of arguments. cc @gsnedders


saschanaz:

gsnedders:
see also https://github.com/heycam/webidl/issues/691

Issue #3848, opened 2019-04-18
[css-color-adjust-1] Disallow repetition of color-scheme keywords?


tabatkins:
Currently 'color-scheme' is defined to allow repeated keywords, like `color-scheme: light light`, and the repetitions are just dropped from the computed value.

I copied this from Rune's initial spec, but I'm not sure if we want it; we don't usually allow meaningful things to be repeated. Would it be okay to just call this invalid?

frivoal:
I don't see a useful meaning to it, so I'd support dropping this.

emilio:
I suspect @lilles did that because if we allow parsing arbitrary identifiers for future compat, then you need to keep some kind of hash set of strings or such while parsing the property, to know if there are duplicates (since the engine doesn't have any internal representation for those)...

But I'm not sure why parsing arbitrary identifiers is ok, it seems undesirable, at least from the `@supports` and such point-of-view.

tabatkins:
> But I'm not sure why parsing arbitrary identifiers is ok

Because CSS isn't super-great at handling properties that are additive sets in the first place, let alone ones that'll expand over time. It would really suck if you designed your page to handle all the color schemes, including the new `sepia` one just added in color-adjust-2, but an older UA drops the property and you end up supporting none of them, because `sepia` isn't recognized.

emilio:
Well, I mean, you'd use something like `color-scheme: light; color-scheme: light sepia;` right?

emilio:
Otherwise there's no way to know if the UA _actually_ supports the sepia color-scheme. `@supports (color-scheme: whatever)` would return true.

tabatkins:
Even if the UA *knows about* `sepia`, there's no guarantee that it *supports* that value, and @supports wouldn't help you there. Luckily, `color-scheme` doesn't really care; it's all about giving the UA *permission* to use a particular color scheme. Unless you use `only`, you don't have any control over the scheme in the first place. (And even then, there's still no guarantee that your desired scheme is supported by the UA, regardless of whether it's recognized.)

emilio:
Ah, ok, fine then, sorry for the noise :)

lilles:
We do allow repetition of property names in transition-property, and names in animation-name, for instance.

emilio:
They don't go away in the computed style as mentioned on the original comment on this issue though, right? I think Rune's argument about consistency with these other properties makes sense.

lilles:
[...]

PR #3930, opened 2019-05-14
[css-color-adjust-1] Make computed value of color-scheme as specified. Issue #3848.

lilles:


lilles:
> We spent a lot of effort cleaning up the Computed Value lines to be more precise, so that TypedOM and Web Animations have something they can rely on. This change seems like a regression.

Changed to just remove the de-duplication/removal text.

PR #3942, opened 2019-05-17
[css-ui] appearance: Remove 'button-bevel'

zcorpan:
The 'button-bevel' value is removed in Chromium 75:

https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/ldH_uIfk_QE/Ua0bGyLIAwAJ

Tests: https://github.com/web-platform-tests/wpt/pull/14980

---

cc @tkent-google @MatsPalmgren

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -1796,7 +1796,7 @@ so that CSS can be used to restyle them.
 
 <pre class="propdef">
 Name: appearance
-Value: ''appearance/none'' | ''auto'' | ''button'' | ''textfield'' | ''menulist-button'' | <<compat>>
+Value: ''appearance/none'' | ''auto'' | ''button'' | ''textfield'' | ''menulist-button'' | <<compat-auto>>
 Initial: none
 Applies To: all elements
 Inherited: no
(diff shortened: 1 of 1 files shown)


MatsPalmgren:
`'listbox'` is [implemented in Gecko](https://searchfox.org/mozilla-central/search?q=StyleAppearance%3A%3AListbox) and is different from `'none'` (different border size/colors/style and something about the focus ring, maybe also something on Windows which I didn't test but the widget code handles it (see widget/windows/nsNativeThemeWin.cpp in the link), so I think it needs to behave as `'auto'`. `'listbox'` also appears to be implemented in Safari.

tkent-google:
I confirmed WebKit had ```listbox``` implementation.  Chrome recognizes ```listbox``` keyword, but does nothing for it.
We should not remove ``listbox``` keyword.

zcorpan:
Thank you @MatsPalmgren and @tkent-google 

In my testing, I thought 'listbox' was close enough to 'none' such that it could be an alias. But I've reverted this so that it's an alias to 'auto' again.

zcorpan:
Browser bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1554150
https://bugs.webkit.org/show_bug.cgi?id=198219

Here's a Live DOM Viewer test to check which values (of those removed in Chromium 75) are supported:
 http://software.hixie.ch/utilities/js/live-dom-viewer/saved/6943

frivoal:
Looks good to me. We should be aligning the compat values with the subset of things implemented by browsers that are needed for compat. Values that are being dropped by browsers shouldn't stay in the spec.

Merging.

zcorpan:
Thank you, @frivoal!

Issue #1084, opened 2017-03-07
[css-scroll-snap] Should negative scroll-padding be allowed?


fantasai:
Was flipping through [old comments](https://github.com/w3c/csswg-drafts/issues/395#issuecomment-241158192) and noted that we forgot to consider whether `scroll-snap-padding` should be restricted to non-negative values. In light of that discussion, I think we should go ahead and disallow negative values.

Thoughts?

tabatkins:
Padding is already restricted to non-negative, and scroll-snap-padding's purpose is to make sure the snapped points don't get overlapped by other stuff positioned at the edges of the screen.  A negative scroll-snap-padding would void that, as the scroller *itself* would overlap it.

So :+1:

frivoal:
Agreed. I suspect that if somebody actually came up with use cases, these would more likely be solved by introducing a `scroll-(snap-)margin` and allowing that to be negative.

tabatkins:
Agenda+ to approve this normative change (restricting scroll-padding to non-negative values only).

astearns:
I'm assuming we need one or more testcases for this change

astearns:
Reopening for testcase need.

PR #3943, opened 2019-05-20
[css-scroll-snap] scroll-padding is not negative.

ewilligers:
Discussed in #1084 and
https://drafts.csswg.org/css-scroll-snap-1/#changes-20161020


fantasai:
Happy for you to merge this in, but please first insert a line break before "Negative". We're following http://rhodesmill.org/brandon/2012/one-sentence-per-line/ :)

PR #3959, opened 2019-05-23
[web-animations-1] Move definition of getAnimations from Document to DocumentOrShadowRoot

### LINKED ISSUE #3951: [web-animations-1] Add `getAnimations()` to `ShadowRoot`
Currently the following will return an empty array:

```js
const shadow = host.attachShadow({ mode: 'open' });
const elem = document.createElement('div');
shadow.appendChild(elem);
elem.animate(...);

// Get the animations from the host
// --> Returns empty array because getAnimations does not cross shadow-tree boundaries
host.getAnimations({ subtree: true });
```

To make working with shadow trees easier we could add:

```webidl
partial interface ShadowRoot {
  sequence<Animation> getAnimations();
};
```

(Or should we add this to `DocumentFragment` instead?)

birtles:
This fixes #3951.


birtles:
Thanks Eric!

PR #3961, opened 2019-05-23
[css-nav-1] Fix a few typos and reword one bit that's unclear.

dbaron:
This fixes a few typos and rewords one bit that I noticed while reviewing w3ctag/design-reviews#287.


jihyerish:
Thank you for fixing errors and reviewing the document!

Issue #3024, opened 2018-08-17
[css-ui] Unprefix 'appearance' and/or make the spec web-compatible


zcorpan:
In https://github.com/whatwg/html/issues/3912 there is a proposal to specify two new values of the 'appearance' for the fieldset and legend elements.

I note that none of WebKit, Chromium, Gecko or Edge support 'appearance' unprefixed, but they do support 'webkit-appearance'.

http://software.hixie.ch/utilities/js/live-dom-viewer/saved/6114

Would it be reasonable to unprefix it?

cc @mstensho @MatsPalmgren @dholbert @emilio @thejohnjansen @smfr

zcorpan:
annevk pointed me to https://groups.google.com/forum/#!msg/mozilla.dev.platform/Fq6OH5SWeRo/lxHDXFnwCQAJ

annevk:
@jwatt

emilio:
The current spec (https://drafts.csswg.org/css-ui/#propdef-appearance) is not web-compatible at all. I'd be against unprefixing it without a proper spec for the property.

cc @FremyCompany too

zcorpan:
I think it would be good to have a spec that is closer to what browsers are willing to implement. I may be able to work on that in the near term.

zcorpan:
Also see https://github.com/whatwg/compat/pull/104

frivoal:
The specified version isn't meant to be exactly identical to the prefixed version, as the prefixed version is sort of insane. The biggest problem with the prefixed version is that there's not value that lets native form control just do their normal thing. You need a gigantic list of values, some of which are meant to apply to non-standard pseudo elements that represent parts of form controls.

As far as I know, no browser is planning on harmonizing the full list of pseudos+values.

The proposed "standard" version has an "auto" value to cope with that problem, so you don't need the full laundry list of values (and associated pseudos) for it to work at all. `auto` does nothing on divs, and does the right thin on the right type of control.

That said, we could add a few select values that people do want to opt into, either because they're genuinely useful, or for compat reasons. Browser representatives in the css wg have so far objected to doing that, but there's nothing in the design of the property that prevents us from adding more than just `auto` and `none`.

If we do add more, there are two ways to go about it. `appearance: foo` (with `foo` being something else than `auto` and `none`, such as `buttton`, or `fieldset`…) could mean either:
*  just turn anything into a foo, so you can turn regular divs into foos, you can turn bars into foos, etc.
* do the same as `auto`

[...]

PR #3966, opened 2019-05-24
[css-ui] Restrict appearance: button more

zcorpan:
Per web compat analysis [1], the `button` value can be restricted to

* `a`
* `button`
* `input type=button/reset/submit`
* drop-down box `select`

Part of #3024

Tests: TODO

[1] https://github.com/w3c/csswg-drafts/issues/3024#issuecomment-493818911

---

cc @tkent-google @foolip @MatsPalmgren @emilio @jwatt

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -87,6 +87,7 @@ spec:css-overflow-4; type:property; text:text-overflow
 spec:selectors-4; type:selector; text::enabled
 spec:selectors-4; type:selector; text::disabled
 spec:css-color-4; type:property; text:color
+spec:html; type:element; text:a
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


emilio:
How does this restriction work, is it at used value time? (looks like it from the spec text)

zcorpan:
Yes. Keywords don't resolve/compute to other keywords, just "act as if".

frivoal:
Looks good to me, but for one thing: I am a little surprised there's not web-compat need for `button` to have an effect on arbitrary divs. Did the compat study look include plain (non replaced, non-control) elements in "others" that and conclude it was fine, or was this left out?

zcorpan:
I think "others" included any element.

@tkent-google which elements does Chromium currently allow to have `button` appearance?

tkent-google:
> I think "others" included any element.
> 
> @tkent-google which elements does Chromium currently allow to have `button` appearance?

Chrome 79 stable:  `button` works with arbitrary elements except for `<select>` and `<a>`.  We haven't received significant site-compatibility issues, and I think we can continue to exclude `<select>` and `<a>`.

Chrome 80 beta: `button` will work only with `<button>` and `<input type=button/color/reset/submit>`.  It's still in Beta.  We don't know whether wen can keep this behavior or not yet.

frivoal:
The discussion has moved on since this PR, and https://github.com/w3c/csswg-drafts/issues/5174 has the latest on this topic.

PR #3968, opened 2019-05-24
[css-ui] Disallow internal pseudo-elements and 'appearance' values

zcorpan:
Pseudo-elements for internal parts of controls (e.g., `::-webkit-search-cancel-button`)
and related 'appearance' values (e.g., `searchfield-cancel-button`)
must not be exposed to author origin CSS or in `getComputedStyle()`.

If user agents find that they can't remove support for something,
the pseudo-element and the appearance value should both be specified.

Tests:
- 'appearance' values: already covered by css/css-ui/appearance-cssom-001.html
- pseudo-elements: TODO
- getComputedStyle: TODO

Part of #3024.

---

cc @tkent-google @foolip @MatsPalmgren @emilio

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -1804,25 +1804,6 @@ Computed value: specified keyword
 Animation type: discrete
 </pre>
 
-<div class="note">
-	Note: This specification intentionally refrains from making the appearance
-	of all possible form controls and sub-controls available as values,
-	as had previously been attempted by earlier proposals for this property
-	and by several UA vendors in experimental implementations.
-	Experience has shown that such a list would be very long and not practical to maintain,
-	and UAs would need to add non-standard values
-	to account for the behavior of non-standard pseudo-elements
-	sometimes used to implement form controls.
-	Moreover, many values of such an enumeration only make sense
-	on a single element or pseudo-element,
-	and are never used outside of the UA stylesheet.
-	The ability provided by this approach to turn any arbitrary element
-	into any arbitrary form control or subcontrol
-	was also a significant source of complexity,
-	despite being not particularly useful.
-	The model provided by this specification is deliberately simpler.
-</div>
-
 <dl dfn-type=value dfn-for=appearance>
 	<dt><dfn>none</dfn>
 	<dd>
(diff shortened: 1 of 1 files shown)


FremyCompany:
> ```css
> ::-webkit-progress-bar { z-index: 1 }
> ```
> ```js
> assert_not_equals(getComputedStyle(elm, pe).zIndex, '1', elm.outerHTML);
> ```
> https://github.com/web-platform-tests/wpt/pull/17008/commits/c12c73e6235e994b08216721f43e80c318a2efd0

Wait, what? You are requiring getComputedStyle to not work on absolutely no property on pseudo-elements? This is wild, and absolutely not something that seems up for discussion or suggested by anyone as far as I'm concerned.

The only changed required is that the appearance property on those pseudos must return a value that is supported by the property, and that is absolutely logical and shouldn't require any specific text except maybe a note; browser who support special values internally can convert those values to 'auto' if those pseudos currently have a proprietary value today (either for real, or just for gCS purposes). 

But in the test you verify that zIndex lies to author about having a value it actually has, this is significantly way beyond the scope of speccing the appearance property.

FremyCompany:
cc @frivoal @fantasai

zcorpan:
Well there's also no spec for those pseudo-elements, and `getComputedStyle` previously only supported `::before` and `::after`. But you're right that this is not strictly about the `appearance` property.

What I tried to address here is the concern raised at the 2018 TPAC f2f, about what `getComputedStyle()` should return for pseudo-elements where the UA stylesheet has a value that's not specified. Then I thought only `::before` and `::after` was supported, and therefore there was no problem. But CSSOM now allows any pseudo-element it seems.

I can file a new issue for CSSOM to discuss this.

zcorpan:
https://github.com/w3c/csswg-drafts/issues/3980

css-ui-4/Overview.bs
```diff
@@ -1888,6 +1869,23 @@ Animation type: discrete
 
 </dl>
 
+Some controls have internal parts, or sub controls.
+For example, an <code class="lang-markup">&lt;input type=search></code> element
+could have a clear button to clear the text.
```
jugglinmike: In this context, "a clear button" might be taken to mean "a button with no background color." Scare quotes would help avoid this ("a "clear" button") as would removing the adjective altogether (which seems acceptable given the phrase is immediately followed by a description of the button's purpose).

zcorpan:
I'm marking this "Needs Design / Proposal" to indicate that this PR is blocked on a resolution of #3980

zcorpan:
@FremyCompany 

> The only changed required is that the appearance property on those pseudos must return a value that is supported by the property, and that is absolutely logical and shouldn't require any specific text except maybe a note; browsers who support special values internally can convert those values to 'auto' if those pseudos currently have a proprietary value today (either for real, or just for gCS purposes).

Yes, I think you are correct, or more simply, the css-ui spec could require such values to be the `auto` value.

alshamiri5 (approved):

PR #3969, opened 2019-05-24
[css-ui] Remove warning about 'appearance'

### LINKED ISSUE #1250: [css-ui] 'auto' as the initial value for the 'appearance' property isn't web-compatible
https://drafts.csswg.org/css-ui-4/#appearance-switching
> Value: 	auto | none 

For this [testcase](https://bug1357655.bmoattachments.org/attachment.cgi?id=8859626), the output in  the Console in Chrome, Safari and Edge is:
> div: none
> select with -webkit-appearance:inherit: none
> button: button
> button w. -webkit-appearance:initial: none

I think that makes it clear that the initial value for `-webkit-appearance` is `none` in those UAs.  I believe the initial value for `appearance` must be the same.

After implementing `appearance` according to spec  in Gecko (with `auto` as the initial value and  `-webkit-appearance` as an alias) we [found](https://bugzilla.mozilla.org/show_bug.cgi?id=1357655) that this isn't web-compatible.

We will now try with `none` as the initial value instead, and override that with `appearance:auto` in the UA sheet for elements that have a native theme in Gecko, `<button> <select>` etc.

zcorpan:
Closes #1250.

---

cc @tkent-google @foolip @MatsPalmgren @emilio @jwatt

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -1749,41 +1749,6 @@ so this specification allows flexibility.
 
 <h3 id="appearance-switching" caniuse="css-appearance">Switching appearance</h3>
 
-<div class=issue>
-	<strong>This is a work-in-progress, not cleared for shipping</strong> (see <a href="https://github.com/w3c/csswg-drafts/issues/1250">Github issue 1250</a>).
-
-	This previously existed as a prefixed form in most browsers.
-	Before standardization, in addition to none,
-	the possible values were a very long list of all the ways an element could look;
-	each value being responsible for giving form control their specific look and feel.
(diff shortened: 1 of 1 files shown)


zcorpan:
@frivoal ping

lukebjerring:
@frivoal - are you able to take a look at this?

frivoal:
Thanks for the suggestion. Decided to keep part of this issue and turn it into a note instead of deleting it entirely. Fixed using https://github.com/w3c/csswg-drafts/commit/0c820686b6c97e521dfa6cfd31c73038aad98f71

PR #3977, opened 2019-05-27
[css-values-3] Bikeshed-added 34 links to tests in Overview.bs

TalbotG:
List of tests bikeshed-added with <wpt> into Overview.bs are 

absolute-length-units-001.html
angle-units-001.html
angle-units-002.html
angle-units-003.html
angle-units-004.html
angle-units-005.html
attr-color-invalid-cast.html
attr-color-invalid-fallback.html
attr-color-valid.html
attr-invalid-type-001.html
attr-invalid-type-002.html
attr-invalid-type-008.html
attr-length-invalid-cast.html
attr-length-invalid-fallback.html
attr-length-valid-zero-nofallback.html
attr-length-valid-zero.html
attr-length-valid.html
attr-px-invalid-cast.html
attr-px-invalid-fallback.html
attr-px-valid.html
calc-angle-values.html
calc-background-position-002.html
calc-in-calc.html	
calc-in-counter-001.xhtml
calc-in-media-queries-001.html
calc-in-media-queries-002.html
calc-invalid-range-clamping.html	
calc-letter-spacing.html
calc-min-height.html
calc-nesting.html
calc-numbers.html
calc-parenthesis-stack.html
calc-time-values.html
calc-unit-analysis.html

I want to soon add the remaining ones

### DIFF
--- a/css-values-3/Overview.bs
+++ b/css-values-3/Overview.bs
@@ -791,10 +791,6 @@ Numbers with Units: <a>dimension</a> values</h3>
 	It corresponds to the <<dimension-token>> production
 	in the <a href="https://www.w3.org/TR/css-syntax/">CSS Syntax Module</a> [[!CSS3SYN]].
 	Like keywords, unit identifiers are <a>ASCII case-insensitive</a>.
-	
-	<wpt>
-	css/css-values/angle-units-003.html
-	</wpt>
 
 	CSS uses <<dimension>>s to specify
 	distances (<<length>>),
@@ -935,6 +931,10 @@ Distance Units: the <<length>> type</h2>
 	user agents must approximate it in the <a lt="actual value">actual</a> value.
 
 	There are two types of length units: <a lt="relative length">relative</a> and <a lt="absolute length">absolute</a>.
+
+	<wpt>
+	css/css-values/calc-unit-analysis.html
+	</wpt>
 
 <h3 id="relative-lengths">
 Relative Lengths</h3>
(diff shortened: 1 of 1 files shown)


frivoal:
Thanks. These all seem to be in the right sections, which is great. However, to help evaluated spec coverage more precisely, it would be helpful to split the `<wpt>` tags into smaller ones place them at the end of the relevant paragraph, rather than at the end of the section.

Also, sometimes a test may exercise statements in two or more paragraphs. When that is the case, I think it is good practice to list the test in the `<wpt>` tag after both paragraphs:
* this gives a better picture of the coverage
* if at a later time, the editor modifies some paragraph of the specification, they should check the related tests to see if they need to be updated to. This works best if the tests can be found next to all the spec statements they verify.

However, this is just my take on this. @tabatkins or @fantasai, you're co-editing this spec. Do you agree, or would you rather merge this PR as is, or do you have some other preference?

TalbotG:
> it would be helpful to split the `<wpt>` tags into smaller ones place them at the end of the relevant paragraph, rather than at the end of the section.

I actually wanted to do that. But I thought that I could not or that I was not supposed to do that or that it was not adequate or not requested. Breaking `<wpt>` tags into smaller ones and to place them at the end of the relevant paragraph will most likely involve more time for careful reading (of tests - especially those that I did not create - and of the spec). There is no exact science on the best/ideal spot to insert a `<wpt>`. But I will try to place smaller `<wpt>` at the most relevant spots from now on.

frivoal:
> I actually wanted to do that. But I thought that I could not or that I was not supposed to do that or that it was not adequate or not requested. 

You absolutely can. We cannot make `<wpt>` that apply to each sentence, but you can place one after each block level element (`<p>`, `<div>`, `<li>`...)

tabatkins:
> Do you agree, or would you rather merge this PR as is, or do you have some other preference?

I agree with this philosophy overall. The closer the WPT is to the statement being tested, the better, and you can have the same test occur multiple times without issue.

(I'm for merging this now, tho, so I'm glad @fantasai did.)

Issue #3980, opened 2019-05-27
[cssom] Should getComputedStyle() support non-standard pseudo-elements?


zcorpan:
https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle

See https://github.com/w3c/csswg-drafts/pull/3968#issuecomment-496033399

Why does `getComputedStyle()` support any pseudo-element? Previously it only supported `::before` and `::after`.

This is an issue for `appearance` because UAs can support non-standard values for non-standard pseudo-elements. (Though css-ui could require `auto` to be used instead.)

Do we want such pseudo-elements to be available in author CSS? Do we want to support them in `getComputedStyle()`?

cc @frivoal @fantasai @FremyCompany

FremyCompany:
We resolved on this last year, based on the fact all browsers de-facto already allowed it to some extent at least. The discussion is saved there if you are looking for some detail in particular:

https://github.com/w3c/csswg-drafts/issues/2149#issuecomment-380386188

zcorpan:
Thank you. So the rationale was "browsers seem to allow it".

The [wpt PR](https://github.com/web-platform-tests/wpt/pull/10422) for that change added a test for `getComputedStyle(div, "totallynotapseudo"),` but no other pseudo-elements (apart from before/after which were already tested).

https://github.com/web-platform-tests/wpt/pull/17008 tests non-standard pseudos for controls.

https://wpt.fyi/results/css/css-ui/control-internal-pseudo-elements-getComputedStyle-001.html?sha=c12c73e623 shows results for Firefox and Safari. Chrome results are missing for some reason, but it passes all the tests for me. Which is to say:

* Firefox supports `::-moz-*` pseudos in gCS.
* Chrome and Safari don't support any of the tested pseudos in gCS.

FremyCompany:
If you could add some tests covering the previous decision, that would be awesome :-)

birtles:
Firefox also supports `::marker` as far as I can tell.

gmksupernova:
{"error": {"message": "", "code": 404}}

zcorpan:
@gmksupernova https://github.com/web-platform-tests/wpt.fyi/issues/1322

AmeliaBR:
I was testing unrecognized pseudo-element selectors in the context of #4456 and discovered how much of an inconsistency this issue is.

Test case: https://codepen.io/AmeliaBR/pen/ZEEJeWm

Summary:

[...]

PR #3985, opened 2019-05-30
[css-ui] Hide proprietary appearance values #3968

jugglinmike:
This is an attempt to implement the following suggestion from gh-3968:

> The only changed required is that the appearance property on those pseudos must return a value that is supported by the property, and that is absolutely logical and shouldn't require any specific text except maybe a note; browsers who support special values internally can convert those values to 'auto' if those pseudos currently have a proprietary value today (either for real, or just for gCS purposes).

However, in gh-3980, the motivating use case is under discussion. Depending on how that is resolved (e.g. by disabling access to non-standard pseudo-elements), this change may not be necessary. Even in that case, standardizing this behavior could improve consistency in a more immediate timeframe.

The text I've written feels primitive to me, but I can't think of a better way to normatively handle a disallowed situation. For instance, it would be cleaner to allow UAs to extend the set of valid values, but that would also allow authors to use apply those proprietary values in their own code. That doesn't seem like what's desired here.

An alternative to the property's computed value may be [CSSOM's definition of "resolved value"](https://drafts.csswg.org/cssom/#resolved-value). That's a little more indirect, but indirection may be warranted for an edge case like this.

/cc @FremyCompany @zcorpan

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -1800,7 +1800,7 @@ Value: ''appearance/none'' | ''auto'' | ''button'' | ''textfield'' | ''menulist-
 Initial: none
 Applies To: all elements
 Inherited: no
-Computed value: specified keyword
+Computed value: See below
 Animation type: discrete
 </pre>
 
@@ -1887,6 +1887,10 @@ Animation type: discrete
 
 
 </dl>
+
+The computed value is the specified value unless it is not included in the set
+of values defined by this specification. In that case, it is the value
+''appearance/auto''.
 
 <div class=example>
 	An author wanting to customize the look and feel of check boxes in HTML could use the following:


FremyCompany:
Sounds good by me, but I would phrase this a bit more explicitly in the tabl:

```
Computed value: as specified (except for proprietary values, see notes below)
```

emilio:
> You can serialize CSS values without going through getComputedStyle(), in particular element.style and document.styleSheets[n].cssRules[m]. Changing the computed style doesn't affect these I believe.

But you cannot set those as an author, and you cannot access UA stylesheets via CSSOM, right? At least the way I plan to go about non-standard values in Gecko is "removing as much as possible, leave the rest only exposed to UA stylesheets".

zcorpan:
Right, but that setup requires that the non-standard keywords aren't parsed in author-origin CSS (what about user origin?), which would be a separate requirement.

jugglinmike:
@zcorpan I assumed that was a requirement (above, "That doesn't seem like what's desired here.").

Is it acceptable to allow authors to set `appearance` to a proprietary value?

zcorpan:
I suppose UAs aren't allowed to support values that aren't listed in the Value grammar, but this isn't particularly clear (to me, at least). From the spec change here, a possible interpretation is that non-standard values should continue to work in author CSS but getting their value from `getComputedStyle()` should give `auto`.

I don't think authors should be able to set `appearance` to non-standard values.

alshamiri5 (approved):

Issue #3990, opened 2019-05-30
[css-ui] "control" is undefined


jugglinmike:
gh-3971 generalized the specification's references to "form controls" to simply "controls". However, neither term has a precise definition in either CSS or HTML. This is particularly hazardous because the term "control" may be interpreted to involve interactivity, but (as noted in gh-3971) it is actually intended to include non-interactive elements like `<meter>`.

https://drafts.csswg.org/css-ui-4/#control-styling

bkardell:
Probably much to the dismay of everyone, I would like to ask the potentially naive, and maybe already asked question: is a control really just something which has shadow dom?  Like, if you don't hinge it on something like 'interactive element' and it gets wonky and hard to define -- doesn't it wind up being more or less anything with a shadow dom and the conversion more or less the same as those being had by folks producing custom elements with the same sorts of concerns/disagreements about whether you should expose parts/thems them and how?

jugglinmike:
So long as we're being naive... If "control" were defined as "anything with a shadow DOM," then could `appearance: none` simply be interpreted as "do not render the shadow DOM"?

SelenIT:
Does the term _control_ used here differ from the term _widget_ used in the "Rendering" section [of the HTML standard](https://html.spec.whatwg.org/multipage/rendering.html#widgets)? If the difference is not substantial, maybe it would make sense to reuse the same term in both places?

jugglinmike:
HTML doesn't define the term "widget" in a way that Bikeshed is aware of. We could link informally using anchor tags, but that's prone to rot. To properly reference the term "widgets" here, we'll need to update HTML to formally define it.

@frivoal could you verify that replacing the undefined term "control" with [the term "widget" as defined in HTML](https://html.spec.whatwg.org/#widgets) is appropriate?

jugglinmike:
@frivoal I'd be happy to submit a patch changing "control" to "widget," but I'd like to confirm that such a change would be acceptable. Can you let me know?

zcorpan:
I think "widget" for these things is OK.

"Widget" is also a term in ARIA, which is a class of roles, closely related to these but also including e.g. `link` (which wouldn't be an HTML/CSS "widget").

jugglinmike:
Thanks, @zcorpan! I've proposed the first step (defining the term) over in whatwg/html:

https://github.com/whatwg/html/pull/4725

zcorpan:
@gregwhitworth it covers these elements for HTML:

[...]

PR #3991, opened 2019-05-30
[css-ui] Define  "native appearance" and rename control to widget

jugglinmike:
[css-ui] Define `control` #3990


css-ui-4/Overview.bs
```diff
@@ -1785,13 +1785,16 @@ so this specification allows flexibility.
 </div>
 
 While the way most elements in a document look can be fully controlled by CSS,
-controls are typically rendered by UAs using native UI controls of the host operating system,
+<a>controls</a> are typically rendered by UAs using native UI controls of the host operating system,
 which can neither be replicated nor styled using CSS.
 
+The term <dfn>control</dfn> denotes the following [[HTML]] elements:
```
bkardell: Is it good to be quite this specific? What if things change? Are there disagreements even on some in terms of support? Could you say instead "Examples of controls in HTML are: ....."

jugglinmike: This definition, like the spec itself, is meant to be canonical. Ideally, if things change, they'd start by changing here. Even if we find ourselves updating the definition reactively, it seems to me that completeness is a goal of the standard. From that perspective, suggesting a partial list is less helpful than defining a complete set.

bkardell: is <summary>/<details>  not a control then?

jugglinmike: Good question! I'll defer to the editor on that.

frivoal: I'm thinking that maintaining this in a css spec is the wrong place. Tracking how the HTML spec changes will never be perfectly synchronized. Ideally, we'd just refer to a term defined in the HTML spec, and they'd update it when they add new elements. However, there doesn't seem to be one already.

SelenIT: > Ideally, we'd just refer to a term defined in the HTML spec

Doesn't the existing term [_widget_](https://html.spec.whatwg.org/multipage/rendering.html#widgets) satisfy this criteria? BTW, it already includes `<details>`.

zcorpan:
I have changed this PR and https://github.com/whatwg/html/pull/4725 as per discussion in that PR. Please review.

frivoal:
Not sure why the IPR checker is stuck, all contributors to this PR are good to go, so I'll merge anyway.

PR #4054, opened 2019-06-24
[css-transforms-1] Simplify transform function syntaxes

lahmatiy:
Simplify syntaxes for `matrix()`, `translate()`, `scale()` and `skew()`

### DIFF
--- a/css-transforms-1/Overview.bs
+++ b/css-transforms-1/Overview.bs
@@ -1529,6 +1529,7 @@ At this point there are no information about potential privacy or security conce
 <h3 class="no-num" id="CR20190214">Since the <a href="https://www.w3.org/TR/2019/CR-css-transforms-1-20190214/">14 February 2019 Candidate Recommendation</a></h3>
 
 * Relax syntax of <a element-attr for>transform</a> attribute: Do not require commas between <<transform-list>> items.
+* Simplified grammar of transform functions. (No normative impact.)
 * Editorial changes.
 
 <h3 class="no-num" id="WD20181130">Since the <a href="https://www.w3.org/TR/2018/WD-css-transforms-1-20181130/">30 November 2018 Working Draft</a></h3>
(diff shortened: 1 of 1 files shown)


lahmatiy:

fantasai:
Needs an entry in the Changes section, e.g. “Simplified grammar of transform functions. (No normative impact.)”, but otherwise it looks correct afaict.

lahmatiy:
@fantasai Done.

Btw I'm not sure such entries is needed in that case. At least no entries were added in my other PRs with similar fixes.

tabatkins:
Yeah, non-normative changes aren't always explicitly tracked. But it's still good practice to do so. ^_^

lahmatiy:
@tabatkins Thank you! Could you take a look at the relevant PR https://github.com/w3c/csswg-drafts/pull/4053? (it's just about `matrix3d()`)

dirkschulze:
Are the changes to translate (and others correct? Isn’t it possible to write `translate(5,)` after the change now?

dirkschulze:
@tabatkins @lahmatiy https://github.com/w3c/csswg-drafts/pull/4054#issuecomment-506630749

dirkschulze:
(I’d also appreciate to wait for the feedback of one of the spec editors before merging. The editors agreed not to merge PRs w/o feedback of other editors given the current state of the spec.)

ewilligers:
> Isn’t it possible to write `translate(5,)` after the change now?

https://www.w3.org/TR/css-values-4/#component-types

> Commas specified in the grammar are implicitly omissible in some circumstances, when used to separate optional terms in the grammar. Within a top-level list in a property or other CSS value, or a function’s argument list, a comma specified in the grammar must be omitted if:

> all items following the comma have been omitted 

> `example( first? , second? , third? )`

> Given this grammar, ... and `example(first,)` are invalid. 

> If commas were not implicitly omittable, the grammar would have to be much more complicated to properly express the ways that the arguments can be omitted, greatly obscuring the simplicity of the feature.

tabatkins:
As this was a non-substantial grammar fix, and I'm one of the editors of the spec *defining* said grammar, I figured it was fine to do a merge. I wouldn't merge anything subject-matter-related without one of y'all editors approving. ^_^

PR #4062, opened 2019-06-26
[css-round-display] Refer to css-shapes-2

bangseongbeom:
'shape-inside' property was postponed in level 1 and now resides in level 2.


jihyerish:

PR #4064, opened 2019-06-28
[css-grid-1] Fix Previous Versions

xfq:
Add missing versions to fix https://lists.w3.org/Archives/Public/site-comments/2019Jun/0003.html


Loirooriol:
This seems good, but I don't get why some old versions are missing in https://www.w3.org/TR/2017/CR-css-grid-1-20171214/, but https://drafts.csswg.org/css-grid/ seems complete?

tabatkins:
Because the Hg bridge is weird and bad and sometimes the EDs end up getting built from the latest PR. There's some badness where the bridge loses track of what branch different commits are coming from.

PR #4069, opened 2019-06-29
[css-scroll-snap-1] Make <'scroll-padding'> syntax uniform with others

lahmatiy:
Move `auto` keyword in the beginning like in any other syntaxes of the spec.


lahmatiy:

PR #4070, opened 2019-06-29
[css-counter-styles-3] Fix <counter-style> syntax to refer to <symbols()> instead of defining it

lahmatiy:
In current form of `counter-style` syntax, `symbols()` is not a reference for the type:
```
<counter-style> = <counter-style-name> | symbols();
```
This change fix that.


lahmatiy:

PR #4081, opened 2019-07-03
[css-values-3][css-values-4] Fix example in Property Value Examples

lahmatiy:
- Replaced `text-shadow` for `box-shadow` since `text-shadow` has no `inset` component
- Removed unnecessary grouping ([`<shadow>` definition](https://drafts.csswg.org/css-backgrounds/#typedef-shadow))

According to spec, the order of terms should be `none | [ <color>? && <length>{2,4} && inset? ]#`. However, left it as is since it's not a big deal.


lahmatiy:

PR #4087, opened 2019-07-05
[css-text-3] Normatively disallow break-word

sideshowbarker:
This change adds a normative *“Authors must not”* statement prohibiting authors from using the (deprecated) `break-word` value with the `word-break` property.

This change aligns with similar normative author-conformance prohibitions in other specs which define certain features as deprecated; see, for example, the language used in the Media Queries spec with regard to the deprecated `device-width`, etc., media features:

https://drafts.csswg.org/mediaqueries-4/#mf-deprecated
> The following media features are deprecated. They are kept for
> backward compatibility, but are not appropriate for newly written style
> sheets. Authors must not use them.

### DIFF
--- a/css-text-3/Overview.bs
+++ b/css-text-3/Overview.bs
@@ -2449,6 +2449,8 @@ Line Breaking Details</h3>
       ''word-break: normal'' and ''overflow-wrap: anywhere'',
       regardless of the actual value of the 'overflow-wrap' property.
       <!-- This is officially sad -->
+      However, the ''word-break/break-word'' keyword is not appropriate for
+      newly-written style sheets. Authors must not use it.
 
       <wpt>
         white-space/pre-wrap-008.html


sideshowbarker:

frivoal:
LGTM.

In the case of media queries, the deprecated features were effectively misfeatures. Here, the behavior itself is fine, just terribly named and triggered off the wrong property, so I am not sure the *must* is quite as strongly justified. But there's a cleaner way to do this, so I'm ok with being a bit pushy in favor of the better way.

fantasai:
Hi @sideshowbarker,
I don't think we can make this a MUST requirement yet... the newer syntax via `overflow-wrap` is not as widely supported, so having a MUST-level requirement to not use this older syntax is not practical to require. If we want to advise a particular authoring practice here, probably the best we can do at the moment is to encourage using the new syntax alongside the old one.

fantasai:
Currently blocked on https://bugs.webkit.org/show_bug.cgi?id=195345

frivoal:
Discussed this offline with @sideshowbarker (a while back, only remembering it now), and this is not so much an issue about css-text-3 specifically, but rather a more generic meta question: css should define and use consistent terminology for words like "deprecated", "obsolete", and other similar terms.

Here are some of the possible meanings for such words (non exhaustive list):
1. Using this is a bad idea (but if you do, here's how it works)
2. this is an old bad name/API for a useful thing which has a new good name/API. Use the new one if you can, be mindful of compat.
3. This is going away, don't count on it being there, or on continuing to do anything

I think, at present, CSS specs have some cases of 1 (https://drafts.csswg.org/mediaqueries-4/#mf-deprecated), some cases of 2 (https://drafts.csswg.org/css-text-3/#valdef-word-break-break-word. I can't quite tell if https://www.w3.org/TR/css-color-3/#css2-system is 1 or 3. We call all these "deprecated".

Should we clean up our terminology to account for this nuance? Is there some systematic (and formally defined) existing use of these words by other groups we could align to?

Note that this is about individual features, and is different from obsoleting / superseding / rescinding entire specifications. That's clearly defined [in the Process](https://www.w3.org/2020/Process-20200915/#tr-end).

alshamiri5 (approved):

frivoal:
This is no longer relevant as a pull request, but I've filed the generalized problem as a standalone issue:  https://github.com/w3c/csswg-drafts/issues/5644

PR #4089, opened 2019-07-05
[css-animations] animation-delay is not animatable

ewilligers:
Replace the old phrasing
Animatable: no
with the new phrasing
Animation type: not animatable


birtles:
Thanks!

Issue #3804, opened 2019-04-04
[css-color-4] Support high-contrast/dark mode colors


fantasai:
With Dark Mode and MSFT's “high contrast” mode (which is more accurately a “specific contrast” mode since it also represents low-contrast settings), we anticipate that authors will want the ability to re-use relevant system colors. Revisiting the CSS System Colors, @tabatkins and I have come up with the following list of colors (based on the available [MSFT high-contrast settings](https://raw.githubusercontent.com/MicrosoftEdge/MSEdgeExplainers/master/HighContrast/TurnOnHighContrast.jpg)) which we propose to add*/undeprecate for these purposes:

- Canvas* (we'd have liked to call this “Background” but that's already used for something else)
- Text*
- LinkText*
- VisitedText*
- Highlight
- HighlightText
- GrayText (could be aliased to InactiveText or DisabledText for clarity)
- ButtonFace (could be aliased to Button for consistency)
- ButtonText

Note: MSFT doesn't distinguish visited vs. unvisited links, but as some systems may we included keywords for both variants: they can be mapped to the same color as needed.

smfr:
WebKit's user agent CSS file contains the following system color values:

background-color: -apple-system-control-background;
background-color: ButtonFace;
color: -webkit-activelink;
color: -webkit-link;
color: ActiveButtonText;
color: ButtonText;
color: GrayText;
color: text;

So of those, the ones not covered by your list above are:

background-color: -apple-system-control-background;
color: -webkit-activelink;
color: ActiveButtonText;

fantasai:
We mainly wanted to hit the colors that are needed to handle forced-colors mode. (MSFT's contrast feature or similar), so our initial list was intentionally limited to the colors MSFT uses in that mode: those are the colors that authors need to be able to access. Extending system colors more generally is possible, but we felt it would be a separate issue so as not to complicate solving this particular problem asap. Not sure how the Apple colors fit into that idea? (What even is -apple-system-control-background)

smfr:
-apple-system-control-background is what we use to render the background of some <input> elements.

AmeliaBR:
How are the proposed `Canvas` and `Text` colors different from the CSS2 `Window` and `WindowText` colors?  Are these expected to be the same colors, but just with better names?

In general, I'd prefer not to introduce new keywords/aliases if there isn't a strong reason for them, since user agents are going to need to continue to support the old keywords anyway.

[...]

Issue #3873, opened 2019-04-24
[css-color-4] Prevent fingerprinting with deprecated system colors


AmeliaBR:
In #3804, we agreed to un-deprecate system colors that have important use cases for accessibility (high contrast mode) and dark mode theming.

I'd like to suggest that the spec get more strict about what "deprecated" means for the remaining colors.  Specifically, **user agents should not expose any user-specific data through the deprecated color names**; they should standardize the values for these colors so that they can't be used as fingerprinting data. The colors could still be adjusted for browser & OS, with light & dark mode variants, since that doesn't expose any new information relative to what's already exposed by user agent strings and media queries.

Of the colors that we are keeping deprecated, the problematic one that I know of is `Background`.  In [the spec](https://drafts.csswg.org/css-color-3/#css-system), this is the "desktop background".  On my windows system, both Firefox and EdgeHTML do expose my custom OS desktop background color. Chrome instead exposes my custom OS theme accent color. Which are sufficiently different that I can't see any user benefit from theming to match my colors, but I can see a significant fingerprinting vector.

PS, [Codepen with all the system colors & their definitions, on various colored backgrounds](https://codepen.io/AmeliaBR/full/JbaZLX), if you want to poke around to see what is being exposed for you

svgeesus:
related to #3847

tabatkins:
Agenda+ to see if we can upgrade the current text "encouraging" either using static defaults or mapping to the undeprecated system colors to being a MUST.

fantasai:
Dropping a note here to say: don't forget to say how these values compute, if they compute to other keyword values or to a fixed color or to themselves. (I'm guessing fewer keywords that compute to themselves here is better for implementations.)

tabatkins:
Depends on whether we want them to respond to color-scheme or not.

AmeliaBR:
> Depends on whether we want them to respond to color-scheme or not.

Given that these are legacy values for legacy compat & we don't want authors using them in new content, we can probably safely define them all as light mode values.  If you want to use system colors in a dark mode theme, only use the non-deprecated values.

But, there should probably be a clear warning to authors about that!

tabatkins:
Yeah that sounds reasonable.

svgeesus:
Allright, no-one had any suggestions so:

- anything to do with borders is mapped to `ButtonBorder` because that is the only non-deprecated border color
- backgrounds are mapped to `Canvas`
- text names are mapped to `CanvasText`
[...]

PR #4091, opened 2019-07-05
[css-color-4] Add new set of CSS system colors (addresses #3804)

melanierichards:
Preserves language around system colors from css-color-3; moves undeprecated values from the appendix, leaving behind those which are still deprecated.

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -129,7 +129,7 @@ Representing sRGB Colors: the <<color>> type</h2>
 	<dfn>&lt;color></dfn> = <<rgb()>> | <<rgba()>> | <<hsl()>> | <<hsla()>> |
 	          <<hwb()>> | <<gray()>> | <<device-cmyk()>> |
 	          <<hex-color>> | <<named-color>> | currentcolor |
-	          <<deprecated-system-color>>
+	          <<system-color>>
 	</pre>
 
 	Some operations work differently on <dfn export>achromatic</dfn> colors.
@@ -209,7 +209,7 @@ Resolving Color values</h2>
 
 		<li>
 		The computed value and used value of
-		<a>named colors</a> (including <<deprecated-system-color>> colors),
+		<a>named colors</a> (including <<system-color>> colors),
 		3 and 6 digits <a>hex colors</a>,
 		4 and 8 digits <a>hex colors</a> with an explicitly opaque alpha channel,
 		comma separated ''rgb()'' colors without an alpha channel,
@@ -366,11 +366,18 @@ The RGB hexadecimal notations: ''#RRGGBB''</h3>
 
 
 
-<h2 id='named-colors'>
-Named Colors</h2>
+<h2 id='color-keywords'>
+Color Keywords</h2>
 
 	In addition to the various numeric syntaxes for <<color>>s,
-	CSS defines a large set of <dfn id="named-color" lt="named color" export>named colors</dfn> that can be used instead,
+	CSS defines several sets of color keywords that can be used instead—
+	each with their own advantages or use cases.
+
+
+<h3 id="named-colors">
+Named Colors</h3>
+	
+	CSS defines a large set of <dfn id="named-color" lt="named color" export>named colors</dfn>,
 	so that common colors can be written and read more easily.
 	A <dfn>&lt;named-color></dfn> is written as an <<ident>>,
 	accepted anywhere a <<color>> is.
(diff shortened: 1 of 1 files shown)


css-color-4/Overview.bs
```diff
@@ -129,7 +129,7 @@ Representing sRGB Colors: the <<color>> type</h2>
 	<dfn>&lt;color></dfn> = <<rgb()>> | <<rgba()>> | <<hsl()>> | <<hsla()>> |
 	          <<hwb()>> | <<gray()>> | <<device-cmyk()>> |
 	          <<hex-color>> | <<named-color>> | currentcolor |
-	          <<deprecated-system-color>>
+	          <<system-color>>
```
dbaron: It looks like you define both `<system-color>` and `<deprecated-system-color>`, so don't both of them still need to be referenced here?  Though I think it would likely be better to have a single term that includes both of them (probably `<system-color>`), but that doesn't look like how it's defined in this PR.

css-color-4/Overview.bs
```diff
@@ -2759,11 +2809,11 @@ Default Style Rules</h2>
 
 </pre>
 
-<h2 id="system-colors" class="no-num">
+<h2 id="deprecated-system-colors" class="no-num">
 Appendix A: Deprecated CSS System Colors</h2>
 
 	Earlier versions of CSS defined several additional named color keywords,
-	the <dfn>&lt;deprecated-system-color></dfn>s,
+	the <dfn id="deprecated-system-color" lt="deprecated system colors" export>system colors</dfn>&lt;deprecated-system-color></dfn>s,
```
dbaron: This change seems a bit strange; it's not clear why the `<deprecated-system-color>` text remains after the `dfn` element.  It might be OK as it was...

css-color-4/Overview.bs
```diff
@@ -2332,6 +2332,56 @@ Converting Between Uncalibrated CMYK and RGB-Based Colors</h3>
 		<li>fallback color must be set to the input color
 	</ul>
 
+
+<h2 id="css-system-colors">
+CSS System Colors</h2>
```
dbaron: As I mentioned above, I think it would likely be better if something in this section included the deprecated system colors in the `<system-color>` term so that other pieces of spec wouldn't miss half of them.

css-color-4/Overview.bs
melanierichards: This looks like a weird paste error I didn't catch, thanks @dbaron!

css-color-4/Overview.bs
```diff
@@ -2332,6 +2332,56 @@ Converting Between Uncalibrated CMYK and RGB-Based Colors</h3>
...
+
+	<dl id="system-color-values" data-dfn-type="value" dfn-for="<system-color>" export>
+		<dt><dfn>ActiveText</dfn>
+		<dd>Text in active links.
+		<dt><dfn>ButtonFace</dfn>
+		<dd>The face background color for push buttons.
+		<dt><dfn>ButtonText</dfn>
+		<dd>Text on push buttons.
+		<dt><dfn>Canvas</dfn>
+		<dd>Window background.
```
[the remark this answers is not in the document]
melanierichards: I carried forward a lot of language from css-color-3, I like your suggestion better :)

css-color-4/Overview.bs
```diff
@@ -2332,6 +2332,56 @@ Converting Between Uncalibrated CMYK and RGB-Based Colors</h3>
...
+	should be mapped to the nearest system color value, or to a default color. 
+	Note that some profiles of CSS may not support System Colors at all.
+
+	The following lists the additional values for color-related CSS values and their 
+	general meaning. This list includes a subset of system colors defined in CSS2; 
+	authors should not use additional system colors documented in <a>deprecated system colors</a>.
+	
+	Any color property can take one of the following names. Although 
+	these are case-insensitive, it is recommended that the mixed capitalization shown 
+	below be used, to make the names more legible.
```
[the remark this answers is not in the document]
melanierichards: Carried forward old language here, let me soften based on feedback.

melanierichards:
@dbaron @AmeliaBR one thing I wanted to ask after I take the PR out of draft mode was whether it makes sense to have this new section on system colors as its own section or the spec, or if it would make more sense as a subsection under named colors. Do you all have thoughts/opinions on that?

css-color-4/Overview.bs
melanierichards: Besides removing the definition of `<deprecated-system-color>`, would you mind clarifying what you'd like changed here? Move the deprecated list out of the appendix and into this section (though of course still noting they are deprecated)?

AmeliaBR:
> whether it makes sense to have this new section on system colors as its own section of the spec, or if it would make more sense as a subsection under named colors. 

A subsection makes sense. Or maybe [the top-level section](https://drafts.csswg.org/css-color-4/#named-colors) should be renamed "Color keywords", and then have four subsections below it:

- named colors
- system colors
- transparent
- currentcolor

melanierichards:
Commit has been updated to include the following changes:

* Moved system colors section under "named colors" section
* Addressed inline feedback, with the exception of clarification on how folks want deprecated system colors integrated into the definition (if other changes made don't suffice)
* Updated language in security & privacy, and referenced from system colors section
* Moved from length-based line breaks to semantic line breaks

AmeliaBR:
Thanks for the fast updates, Melanie!

On security & privacy, the issue is not just about spoofing, it's also about fingerprinting. See also #3873.

melanierichards:
@AmeliaBR The security & privacy section of the spec already has a mention for the fingerprinting risk, do we feel it needs more detail? Seems like #3873 might be out of scope for this particular PR, or at least we might need to discuss first as a group?

@Crissov @svgeesus was your upvote on Amelia's comment in favor of renaming the "Named colors" section to "Color keywords"? For now I had just moved system colors within the "Named colors" section.

Thanks all for the feedback!

AmeliaBR:
I think your current text re security & privacy is accurate. Any discussion of stricter requirements on UAs can happen in the other issue.

Crissov:
Iʼm I in the favor of renaming the section and adding sub-headings as suggested.

css-color-4/Overview.bs
svgeesus: @dbaron it looks like this PR is awaiting your clarification of what changes you requested.

svgeesus:
> @svgeesus was your upvote on Amelia's comment in favor of renaming the "Named colors" section to "Color keywords"?

Yes, it was.

melanierichards:
Just renamed "Named colors" section => "Color keywords" as requested (and sheepishly added myself to acknowledgements per @AmeliaBR's comment).

Apologies for the intermittent delay on these last changes!

PR #4106, opened 2019-07-12
[css-values] sin and cos can reach -1

ewilligers:
resolves #4104

PR #4112, opened 2019-07-13
[css-values] sin and cos can reach -1

ewilligers:
Use minus sign instead of hyphen.

Use increasing order for tan, improving consistency
as we use increasing order for sin and cos.


ewilligers:
This addresses the review comments in https://github.com/w3c/csswg-drafts/pull/4106

Loirooriol:
Actually I don't get if the spec is supposed to use hyphens or minus signs, e.g. there are 9 occurrences of `−∞` and 8 of `-∞`, it's very inconsistent. @tabatkins Can you clarify?

tabatkins:
Either works; math minus is prettier, I just don't always remember to copypaste it. Happy to take an alignment PR shifting them all one direction.

Issue #607, opened 2016-10-14
[css-tables] Should min-width support percentages?


FremyCompany:
Edge and Chrome ignores "min-width: %" while Firefox applies it properly.
Firefox continues to apply a "min-width: %" when a "width: px" is specified (the width is then ignored).

We would like to spec Edge/Chrome behavior here. Any opinion?
cc: @dbaron

FremyCompany:
**Summary of the issue**
> Edge and Chrome ignores "min-width: %" while Firefox applies it. 

**Face to face recommendation**
> Resolve that min-width is ignored when its value is a percentage on table-internal elements

**Examples**
> https://wptest.center/#/tdspq9

gregwhitworth:
It's worth noting webkit matches Chrome/Edge in the testcase as well: https://jsfiddle.net/gvco0t54/
 
![image](https://user-images.githubusercontent.com/865244/28835507-d3653dc6-76e5-11e7-987a-0c58a48f8980.png)

PR #4127, opened 2019-07-18
[css-tables-3] Remove vestiges of min-width percents

davidsgrogan:
[css-tables-3] Remove vestiages of min-width percents from #607

WG resolved to ignore percent min-width for column widths in #607 but
not everything was removed.


davidsgrogan:
I just linked my github and w3c accounts but looks like there's a delay

davidsgrogan:

FremyCompany:
Thanks!

PR #4128, opened 2019-07-18
[css-tables-3] Editorial: Fix constrainedness references

davidsgrogan:
bikeshed didn't give the definition the ② ③ links when the
references were done with a href.


FremyCompany:

PR #4130, opened 2019-07-19
[css-writing-modes] Clarify how propagation from body works

### LINKED ISSUE #3066: [css-writing-modes] Does vertical writing mode of an HTML body element cause an orthogonal flow?
https://drafts.csswg.org/css-writing-modes-3/#principal-flow

> As a special case for handling HTML documents, if the root element has a body child element [HTML], the principal writing mode is instead taken from the values of writing-mode and direction on the first such child element instead of taken from the root element. Note that this does not affect the values of writing-mode or direction on the root element itself. 

When applying `writing-mode: vertical-rl` to `body`, `html` has writing-mode value `horizontal-tb` (initial value) according to the note and there happens an orthogonal flow between `html` and `body`?

Each browsers behave differently about this:

* Edge: An orthogonal flow doesn't happen. `body { writing-mode: vertical-rl }` behaves as same as `html { writing-mode: vertical-rl }`.
* Chrome: It seems an orthogonal flow happens since the margin of body child elements doesn't collapse with the margin of a body element.
  * I filed https://bugs.chromium.org/p/chromium/issues/detail?id=845008 and it marked as WontFix.
* Firefox: Propagation of writing-mode is not supported. https://bugzilla.mozilla.org/show_bug.cgi?id=1102175

According to [the old spec text](https://github.com/w3c/csswg-drafts/commit/a0cca817f7abb6e57becd276c8ce08ce0a5a0271), an orthogonal flow can't happen.

> The propagated values are treated as if they were specified on the root element.

frivoal:
'writing-mode' and 'direction' are propagated up from the body element.
This affects the root element in addition to the ICB and viewport,
but only for used values, not for computed values.

The same propagation is optionally allowed for 'text-orientation'.

Closes #3066

### DIFF
--- a/css-writing-modes-3/Overview.bs
+++ b/css-writing-modes-3/Overview.bs
@@ -8,6 +8,8 @@ Work Status: Testing
 Group: csswg
 ED: https://drafts.csswg.org/css-writing-modes-3/
 TR: https://www.w3.org/TR/css-writing-modes-3/
+Previous Version: https://www.w3.org/TR/2018/CR-css-writing-modes-3-20180524/
+Previous Version: https://www.w3.org/TR/2017/CR-css-writing-modes-3-20171207/
 Previous Version: https://www.w3.org/TR/2015/CR-css-writing-modes-3-20151215/
 Issue Tracking: Tracker http://www.w3.org/Style/CSS/Tracker/products/30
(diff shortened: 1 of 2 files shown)


frivoal:
There was a typo-induced bug in this PR. Fixed it: https://github.com/w3c/csswg-drafts/commit/42ecbc8ec0a32a5b76dd5cb12147827d523bbc4c

PR #4145, opened 2019-07-23
[css-page-3] Revert to ED

xfq:
Post publication edit. Also remove the generated HTML.


xfq:
> The edit to Overview.bs is fine, but instead of updating Overview.html, just delete it, and it will be regenerated server-side.

Yep. That's what I did in this PR.

frivoal:
> That's what I did in this PR.

Oops, sorry, I looked a the diff wrong, and somehow thought that you had regenerated it. Merged.

Issue #3779, opened 2019-03-29
[css-overflow] Overflow propagation when the element propagated from is display: none


emilio:
https://drafts.csswg.org/css-overflow/#overflow-propagation says that some overflow properties should be propagated to the viewport from the html or body element:

However this happens regardless of the display value of the document element or the body. For example, the following examples show scrollbars:

* `data:text/html,<html style="display: none"><body style="overflow: scroll">`
* `data:text/html,<html style="display: none; overflow: scroll">`

This is the only case where Gecko needs to resolve a style in a `display: none` subtree for rendering (apart of getComputedStyle), and I find it somewhat silly.

Seems enough of an edge-case that could probably be changed. I wonder if there would be interest from other vendors in doing this change?

cc @lilles

emilio:
Basically, my point is that this should be closer to the background style propagation, which only happens if the document element is actually displayed.

fantasai:
I think from an author standpoint this case doesn't matter, so personally I don't care. Happy to do whatever is expedient for implementers.

lilles:
Fine for us from an implementation pov.

lilles:
Chrome even propagates the background, which is not according to spec.

lilles:
For backgrounds: https://bugs.chromium.org/p/chromium/issues/detail?id=947873

lilles:
There is the same issue for https://drafts.csswg.org/css-writing-modes-3/#principal-flow

Chrome starts with the horizontal scrollbar to the right for direction:rtl regardless of display:none or not on body below.

```
<!DOCTYPE html>
<style>
  html { overflow: scroll; width: 5000px; height: 100px }
  body { direction: rtl; display: none }
</style>
<body></body>
```

PR #4148, opened 2019-07-24
[css-overflow-3] #3779 Do not propagate overflow from boxless elements.

lilles:
Resolution: https://github.com/w3c/csswg-drafts/issues/3779#issuecomment-481762912

### DIFF
--- a/css-overflow-3/Overview.bs
+++ b/css-overflow-3/Overview.bs
@@ -489,7 +489,7 @@ Overflow Viewport Propagation</h3>
 	The element from which the value is propagated must then have
 	a used 'overflow' value of ''overflow/visible''.
 
-	The overflow values are not applied to the viewport if no boxes are
+	The overflow values are not propagated to the viewport if no boxes are
 	generated for the element whose overflow values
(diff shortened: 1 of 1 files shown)


css-overflow-3/Overview.bs
```diff
@@ -489,6 +489,11 @@ Overflow Viewport Propagation</h3>
 	The element from which the value is propagated must then have
 	a used 'overflow' value of ''overflow/visible''.
 
+	The overflow values are not applied to the viewport if no boxes are
```
frivoal:
```suggestion
	The overflow values are not propagated to the viewport if no boxes are
```

css-overflow-3/Overview.bs
```diff
@@ -489,6 +489,11 @@ Overflow Viewport Propagation</h3>
...
+	generated for the element whose overflow values
+	would be used for the canvas (for example, if the root element has
```
frivoal:
```suggestion
	would be used for the viewport (for example, if the root element has
```

Issue #4032, opened 2019-06-13
[resize-observer] svg interaction


bkardell:
ResizeObserver mentions SVG specially several times (see, for example https://drafts.csswg.org/resize-observer/#calculate-box-size).  I don't recall the intent of this, but I am wondering if perhaps it is intended for observing the resizing of things _inside_ the SVG and not the SVG itself?  It seems that if I had some construct with a svg with `width="100%"` the SVG itself is resizing, but if it is observing the boundingBox then those things will not be reported.  This [little pen](https://codepen.io/bkardell/pen/RzwNmv) illustrates what I mean - here, I believe the path should fill green, but it doesn't because it seems that what is being observed is the bounding box.  The pen also updates the reporting of the size of that bounding box every 100ms and as you can see, it never updates.

This seems strange and it does seem (at least from what I can tell) implemented this way in all of the browsers.  As an author, it's true that you can work around this by wrapping a div around it and monitoring that instead, but that seems like a strange workaround and like maybe this creates a need for special understanding.  It seems what we want to observe here is not this?

AmeliaBR:
For SVG elements that have associated CSS layout boxes, you should definitely be able to access the CSS content and border boxes, the same as on an `<img>` or container element. This applies to the root `<svg>` in an SVG document and any inline `<svg>` element that is a direct child of HTML elements (or other non-SVG elements) or a direct child of an SVG `<foreignObject>` element, which creates an internal CSS layout context.  Other specifications have special rules for SVG elements with or without CSS boxes, including [`transform-origin`(https://drafts.csswg.org/css-transforms/#transform-origin-property) and some [CSSOM View APIs](https://drafts.csswg.org/cssom-view/#svg-layout-box).

For SVG elements _without_ CSS layout boxes (i.e, shape or group elements, or `<svg>` nested inside other `<svg>`), then `content-box` and `border-box` aren't defined. So perhaps it makes sense to default to the SVG [object bounding box](https://svgwg.org/svg2-draft/coords.html#BoundingBoxes) size.  This will only be useful in limited cases — where the container `<svg>` does not have a `viewBox` attribute and this element or its children are defined using percentage dimensions, which will therefore change as the container is resized — but most other cases could be handled by listening for resize events on the SVG that _does_ have CSS layout boxes, _if_ those are handled as normal CSS boxes.

gregwhitworth:
[...]

PR #4150, opened 2019-07-24
[resize-observer] remove SVG specific text

gregwhitworth:
Per resolution in #4032 I have removed all SVG specific text since since the SVG root element results in CSS boxes and should follow the same model as the rest. I also added this to the changes section. @AmeliaBR & @bkardell can you please review. cc: @atotic

### DIFF
--- a/resize-observer-1/Overview.bs
+++ b/resize-observer-1/Overview.bs
@@ -234,8 +234,6 @@ interface ResizeObserver {
 
         2. Clear the {{ResizeObserver/activeTargets}} list.
 
-        3. Remove |this| from {{Document}}.<var>resizeObservers</var> slot.
-
 </div>
 
 <h3 id="resize-observer-callback">ResizeObserverCallback</h3>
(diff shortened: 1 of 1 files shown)


atotic:
I do not know enough about interaction between css/svg specs. Does this change means that currently reported contentRect in ROv0 should change? If so, can we also modify RO SVG tests to reflect correct values per standard:

https://github.com/web-platform-tests/wpt/blob/master/resize-observer/svg.html

AmeliaBR:
A couple notes:

- Can you rebase these changes on top of the latest from master? (or just merge in master for now, then squash when the PR is accepted) It makes it hard to review if there are irrelevant changes. E.g., [you're deleting non-SVG instructions re `contentRect`](https://github.com/w3c/csswg-drafts/pull/4150/files#diff-fc4cc4fe5e6535f140c79bbdd1dbabc8L309), is that intentional?

- There should be a note/issue somewhere that the behavior is currently not defined for SVG elements without CSS layout boxes, with a link to w3c/svgwg#713, maybe mentioning as a warning to authors that current implementations return the bounding box values.

- I think @atanassov [wanted the spec to be explicit](https://github.com/w3c/csswg-drafts/issues/4032#issuecomment-510137495) about which SVG elements _do_ have CSS layout box (and therefore should behave as for other CSS boxes).  I don't think that needs to be normative, but it would be useful to have in a note.

bkardell:
Yeah, my only feedback really is what @AmeliaBR said above.. I think the decision was to describe/mention which of these do have/are observing their CSS layout boxes and a note/link to svg issue on what to do with others

frivoal:
This pull request is made with a branch in the main repo, rather than from a fork. This continuously puts strain on the build server, which is rebuilding it all the time to keep the bikeshed references current. This is OK for branches that need to have a preview, but this seems stale. It should be (updated and) merged or closed.

atotic:
This request has been open for more then a year. 
@gregwhitworth do you want to just close it?

tabatkins:
Ping @gregwhitworth?

gregwhitworth:
@tabatkins haven't had a chance to do this just yet. I'll try to get around to it this weekend.

gregwhitworth:
I have not had a chance to make the necessary edits to complete this PR. I have copied the changes locally and will follow the fork method but have deleted the branch that this was associated with. Until I can get to the content I'm going to close this PR.

PR #4153, opened 2019-07-27
[css-text-4] Add word-boundary-detection and word-boundary-expansion

frivoal:


css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+	This property allows the author to decide
+	whether and how
+	the User Agent must analyse the content
+	to determine where word boundaries are,
+	and to insert [=virtual word boundaries=] accordingly.
+
+	A <dfn data-dfn-for='' data-dfn-type=dfn>virtual word boundary</dfn> is similar to the presence
+	of the ZERO WIDTH SPACE (U+200B) character:
+	it introduces a [=soft wrap opportunity=]
+	and is affected by the 'word-boundary-expansion' property;
```
fantasai: End sentence.

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+	its presence has no effect on text shaping,
+	nor on 'word-spacing'.
+	However, its insertion must have no effect on the underlying content,
```
fantasai: Inserting [=virtual word boundaries=] must have ...

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+	and must not affect the content of a plain text copy &amp; paste operation.
+
+	<dl dfn-type=value dfn-for=word-boundary-detection>
+		<dt><dfn>manual</dfn>
+		<dd>
+			The User Agent must not insert [=virtual word boundaries=].
```
fantasai: Add some more explanatory text for authors, maybe:

Linguistic analysis is not used to determine [=line wrapping opportunities=] in any language or writing system.

or something better...

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+
+			If the element’s [=content language=],
+			as represented in BCP 47 syntax [[BCP47]],
+			does <em>not</em> match the language range described by the computed value's <<lang>>
+			in an extended filtering operation
+			per [[RFC4647]] <cite>Matching of Language Tags</cite> (section 3.3.2)
+			with both the [=content language=] and <<lang>>
+			<i>canonicalized</i> and converted to <i>extlang form</i>
+			as per section 4.5 of [[!RFC5646]],
+			then the [=used value=] is ''word-boundary-detection/manual'',
```
fantasai: Should this be normal?

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+		<dd>
+			This value directs the User Agent to perform language-specific content analysis
+			to determine where to insert [=virtual word boundaries=].
+
+			<dfn dfn-type=type><<lang>></dfn> must be a valid CSS <<ident>> or <<string>>.
+			It represents an IETF BCP 47 language range
+			(see [[BCP47]]).
+			If the UA does not support word-boundary detection
+			for <em>all</em> languages represented by the specified range,
+			it must reject that value at parse-time.
```
fantasai: "rejected at parse-time" is not the right terminology. You want to say it's invalid. E.g.

If the UA does not support word-boundary detection for all languages represented by the specified language range, that specified value is invalid (and will cause the declaration to be ignored).

or

Only `<<lang>>` values for which the UA has support of all languages represented by the specified range are valid.

or somesuch.

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+
+		However, if the User Agent supports a generic word-boundary detection system
+		that is suitable for Chinese in general,
+		it is expected to accept the broad ''auto(zh)'' characterization,
+		as well as any more specific ones,
+		such as ''auto(zh-yue)'', ''auto(zh-Hant-HK)'', ''auto(zh-Hans-SG)'', or ''auto(zh-hak).
+	</div>
+
+	<div class=example>
+		Specifying the language for which the word boundary detection is to be performed
```
fantasai: and making unsupported language ranges invalid

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+
+			Note: This implies that for languages such as English
+			where words are separated by spaces or other separating characters,
+			''word-boundary-detection/auto(<lang>)'' has no effect.
+
+		<li>
+			between a [=typographic letter unit=]
+			and a subsequent [=typographic character unit=] from the [[!UNICODE]] Pe or Pf classes,
+			or between a [=typographic letter unit=]
+			and a preceeding [=typographic character unit=] from the [[!UNICODE]] Ps or Pi classes,
```
fantasai: IIRC Pf/Pi classes are characters that are sometimes used as opening and sometimes used as closing, so this might not be quite correct.

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+
+		<dt><dfn>ideographic-space</dfn>
+		<dd>
+			Instances of U+200B ZERO WIDTH SPACE
+			within the [=text run=] children of this element
+			are replaced by U+3000 IDEOGRAPHIC SPACE.
+	</dl>
+
+	The User Agent must not replace
+	instances of U+200B imediately preceding or following
```
fantasai: immediately

css-text-4/Overview.bs
```diff
@@ -31,8 +35,519 @@ Introduction</h2>
...
+	Like 'text-transform', this property transforms text for styling purposes.
+	It has no effect on the underlying content,
+	and must not affect the content of a plain text copy &amp; paste operation.
+
+	<div class=note>Note: The effects of this property are similar
+	to those of the 'text-transform' property.
+	However, it is defined as a separate property
+	rather than additional values to 'text-transform' because:
+	* This property needs to take effect before [[CSS-TEXT-3#white-space-rules]],
+		but 'text-transform' happens after that.
```
fantasai: explain why

css-text-4/Overview.bs
frivoal: Good point. I've though some more about it, and derived updated rules from UAX14 line breaking classes, rather than directly from unicode general categories.

frivoal:
Ok, I've addressed your comments and merged. The largest change I made was in response to your comment about Pi an Pf. Thinking about it, not only did I agree, but I though this section would work out better using UAX14 line breaking classes than unicode general categories. See https://github.com/w3c/csswg-drafts/pull/4153/commits/87a41925005e2f67a3738f5bc840a884d369d754

If you think that's actually a step back, let me know (here or in a new issue), and I'll revert to what it was prior to that, with Pi and Pf removed.

PR #4179, opened 2019-08-07
[cssom] Fix ElementCSSInlineStyle section.

### LINKED ISSUE #4174: [cssom-1] ElementCSSInlineStyle is not an "interface"
The spec headings and prose talk about "the ElementCSSInlineStyle Interface".  It's not an interface, though, it's a mixin.

emilio:
It's a mixing now.

Fixes #4174

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2782,10 +2782,12 @@ DOM Access to CSS Declaration Blocks {#dom-access-to-css-declaration-blocks}
 ============================================================================
 
 
-The {{ElementCSSInlineStyle}} Interface {#the-elementcssinlinestyle-interface}
+The {{ElementCSSInlineStyle}} Mixin {#the-elementcssinlinestyle-mixing}
 ------------------------------------------------------------------------------
 
-The <code>ElementCSSInlineStyle</code> interface provides access to inline style properties of an element.
+<a name="the-elementcssinlinestyle-interface"><!-- Legacy link --></a>
+
+The <code>ElementCSSInlineStyle</code> mixin provides access to inline style properties of an element.
 
 <pre class=idl>
 interface mixin ElementCSSInlineStyle {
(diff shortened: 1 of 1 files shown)


emilio:
@tabatkins so this fails bikeshed with:

> FATAL ERROR: Autolink <a name="the-elementcssinlinestyle-interface" data-link-type="dfn"></a> has no linktext.

Which makes sense, but it feels like there has to be a way to make this or something like this work.

I couldn't find anything in the docs... Do I just need to add a `Note: This section used to be ...` or such so that I can have some text to link or something?

cssom-1/Overview.bs
```diff
@@ -2782,10 +2782,12 @@ DOM Access to CSS Declaration Blocks {#dom-access-to-css-declaration-blocks}
 ============================================================================
 
 
-The {{ElementCSSInlineStyle}} Interface {#the-elementcssinlinestyle-interface}
+The {{ElementCSSInlineStyle}} Mixin {#the-elementcssinlinestyle-mixing}
```
tabatkins:
```suggestion
The {{ElementCSSInlineStyle}} Mixin {#the-elementcssinlinestyle-mixin}
```

tabatkins:
Two ways to fix that bikeshed complaint: the immediate solution, and the correct solution (which is basically just as easy). ^_^

Immediate solution is to use `<span id>` instead of `<a name>`; any `<a>` without an `href` is assumed to be an autolink by Bikeshed. Theoretically I could rule out `<a name>` too, but that's a legacy feature anyway; you should use IDs instead.

The correct solution is to rewrite the markdown heading into an `<h3>`, and use the `oldids` attribute. [docs](https://tabatkins.github.io/bikeshed/#id-gen)

tabatkins:
And I went ahead and did it. R+ from me.

emilio:
Thanks!

emilio:
Blerg, I thought Github was going to be smart enough to add you as a co-author if I chose to squash the commits... Seems it wasn't, sorry :(

tabatkins:
I think I have enough commits in the repo already, don't worry about it. ^_^

PR #4183, opened 2019-08-10
[css-inline] line-height cannot have two values

ewilligers:
line-height must be normal or a number or a length-percentage.

It cannot be a number followed by a length-percentage.

This brings the definition of line-height into consistency with CSS 2.
https://www.w3.org/TR/CSS2/visudet.html#propdef-line-height


ewilligers:
@fantasai please review

PR #4194, opened 2019-08-13
[css-transforms-2] Use reference-box for translate property. 

### LINKED ISSUE #4193: [css-transforms-2] The reference box for resolving percentage on translate
Per the spec definition of [`translate`](https://drafts.csswg.org/css-transforms-2/#propdef-translate), we resolve its percentage by _containing block_:
> relative to the width of the containing block (for the first value) or the height (for the second value) 

Here, we use _containing block_. However, for [`transform`](https://drafts.csswg.org/css-transforms-1/#transform-property) property, we use its [reference box](https://drafts.csswg.org/css-transforms-1/#reference-box) to resolve the percentage:
> refer to the size of reference box

The _containing block_ has a special meaning, I think, based on the spec of [visual formatting model](https://drafts.csswg.org/css2/visudet.html#containing-block-details). It may refer to the content edge or padding edge, based on its `position` value. I checked the implmentation of Blink and Gecko, and it seems both use `transform-box` (default: view-box) as the reference box for both `transform` and `translate`. So my suggestion is to update how we resolve the percentage for `translate`. i.e. Use _reference box_, instead of _containing block_.

BorisChiou:
Fixes #4193.

### DIFF
--- a/css-transforms-2/Overview.bs
+++ b/css-transforms-2/Overview.bs
@@ -552,7 +552,7 @@ Value: none | <<length-percentage>> [ <<length-percentage>> <<length>>? ]?
 Initial: none
 Applies to: <a>transformable elements</a>
 Inherited: no
-Percentages: relative to the width of the containing block (for the first value) or the height (for the second value)
+Percentages: relative to the width of the <a>reference box</a> (for the first value) or the height (for the second value)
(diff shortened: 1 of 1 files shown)


ewilligers:

BorisChiou:
Thanks!

BorisChiou:
BTW, I don't have the write access to land this, so probably need authors or others to help me to merge this. Thanks. :)

PR #4212, opened 2019-08-17
[selectors-4] Canonicalize into extlang form prior to :lang() matching

### LINKED ISSUE #4154: Consider Canonicalization of language tags in :lang() selector maching
Cantonese may be described with any of the following language tags:
* `zh-HK` (most common, legacy)
* `zh-yue` (also legacy)
* `yue` (current most appropriate value)

Even though these refer to the same language, `:lang(yue)` would only match the last one.

[RFC 5646 section 4.5](https://tools.ietf.org/html/rfc5646#section-4.5) together with [the IANA Language Subtag Registry](https://www.iana.org/assignments/language-subtag-registry) define canonicalization and mappings that would allow to match the last two (by extlang form). [RFC4647 section 3.2](https://tools.ietf.org/html/rfc4647#section-3.2) says we _should_ use this.

The last paragraph of that same section 3.2 also says we may also want to consider mappings like zh-HK to zh-yue, and that too could seem appropriate, but as far as I know there’s no equivalent to IANA Language Subtag registry for such mappings.

While it is possible for authors to deal with this themself by using `:lang(yue, zh-yue, zh-HK)`, to the extend we can automate this, they shouldn't have to deal with this.

(I'm taking Cantonese as an example, but the same can be said about other languages)

frivoal:
Fixes #4154


alshamiri5 (approved):

tantek (approved):
This looks like a good improvement to the Working Draft, so I'd be ok seeing it optimistically merged.
Are there any tests that exercise this more detailed treatment so we can check existing implementation :lang support to see how well it is already supported (or needs bug fixing)? Also to evaluate any potential webcompat challenges?

svgeesus:
@tantek thanks for the review!

@frivoal  This PR from 2019 has of course accumulated merge conflicts in the interim. Could you rebase or otherwise handle these, so it can be merged?

svgeesus:
Actually that was a very easy line-rewrapping conflict, so merging.

frivoal:
Happy that the two of you liked the PR, but shouldn't we get a WG resolution first?

cc: @astearns @atanassov

astearns:
It looks like the corresponding issue stalled with a promise of “further guidance” from the internationalization folks. Perhaps you could mention this issue and the PR on the joint call on the 19th, and then we could either re-open the issue with a call for an async resolution, or re-open the issue with further discussion?

svgeesus:
My apologies, I thought this had a resolution.

frivoal:
We do now have a resolution to accept these edits. See https://github.com/w3c/csswg-drafts/issues/4154

PR #4214, opened 2019-08-17
[selectors-4] Fix two links in selectors overview table

shvaikalesh:


shvaikalesh:

PR #4225, opened 2019-08-20
[css-fonts-4] Add explicit [Exposed=Window]

saschanaz:
Required after heycam/webidl#423. (Found from web-platform-tests/wpt#18382)


saschanaz:

PR #4233, opened 2019-08-22
[cssom-view-1] Add dictionary default values

saschanaz:

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -445,11 +445,11 @@ partial interface Window {
     [Replaceable] readonly attribute double pageXOffset;
     [Replaceable] readonly attribute double scrollY;
     [Replaceable] readonly attribute double pageYOffset;
-    void scroll(optional ScrollToOptions options);
+    void scroll(optional ScrollToOptions options = {});
     void scroll(unrestricted double x, unrestricted double y);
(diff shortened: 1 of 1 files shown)


saschanaz:

cssom-view-1/Overview.bs
```diff
@@ -1072,12 +1072,12 @@ dictionary ScrollIntoViewOptions : ScrollOptions {
 partial interface Element {
   DOMRectList getClientRects();
   [NewObject] DOMRect getBoundingClientRect();
-  void scrollIntoView(optional (boolean or ScrollIntoViewOptions) arg);
-  void scroll(optional ScrollToOptions options);
+  void scrollIntoView(optional (boolean or ScrollIntoViewOptions) arg = {});
```
dbaron: Are you sure this is the right thing given the `(boolean or ScrollIntoViewOptions)`?

saschanaz: Yes. Previously the default value was implicitly an empty dictionary when the union includes non-required dictionary type, and now it must be explicit.

saschanaz: Check out the relevant spec change: https://github.com/heycam/webidl/pull/750/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR2245

PR #4235, opened 2019-08-22
[web-animations-1] Add dictionary default values

saschanaz:

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3960,7 +3960,7 @@ dictionary DocumentTimelineOptions {
 };
 
 [Exposed=Window,
- Constructor(optional DocumentTimelineOptions options)]
+ Constructor(optional DocumentTimelineOptions options = {})]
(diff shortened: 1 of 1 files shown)


saschanaz:

birtles:
Oh, actually this needs a little more work. e.g. the [`KeyframeEffect` constructor](https://drafts.csswg.org/web-animations-1/#dom-keyframeeffect-keyframeeffect) has three lines:

* If options is a KeyframeEffectOptions object ...
* If options is a double ...
* Otherwise (options is undefined) ...

But the last option is no longer necessary.

I think that might be all.

birtles:
(Unrelated, but something else I noticed while looking into this was that step 4 of the description of [`animate`](https://drafts.csswg.org/web-animations-1/#dom-animatable-animate) fails to account for when `options` is a `double`.)

birtles:
I addressed the above issues in 3fafc22d659eccb596c502f6c77070e39970af6b.

PR #4238, opened 2019-08-23
Defined @namespace using CSS syntax (fixes #649)

### LINKED ISSUE #649: [css-namespaces-3] Replace CSS 2.1 syntax definition by CSS syntax
The [minutes from 2016-10-12](https://lists.w3.org/Archives/Public/www-style/2016Oct/0091.html) resolved on removing the CSS grammar section from CSS 2.2 and instead refer to the CSS Syntax module. Though the CSS Namespaces Module still [uses that grammar for the definition of the @namespace rule](https://drafts.csswg.org/css-namespaces-3/#syntax).

This syntax should be replaced by something like this:

```
@namespace <namespace-prefix>? [ <string> | <url> ];

where
<namespace-prefix> = <ident>
```

Sebastian

SebastianZ:
[css-namespaces-3] This replaces the CSS 2.1 syntax used to define the `@namespace` rule by the newer [CSS Syntax 3](https://drafts.csswg.org/css-syntax-3/) rules used in other specs.

### DIFF
--- a/css-namespaces-3/Overview.bs
+++ b/css-namespaces-3/Overview.bs
@@ -145,21 +145,13 @@ Terminology</h3>
 <h3 id="syntax">
 Syntax</h3>
 
-	The syntax for the ''@namespace'' rule is as follows
-	(using the notation from the <a href="https://www.w3.org/TR/CSS21/grammar.html">Grammar appendix of CSS 2.1</a> [[!CSS21]]):
+	The syntax for the ''@namespace'' rule is:
 
 	<pre>
-		namespace
-		  : NAMESPACE_SYM S* [namespace_prefix S*]? [STRING|URI] S* ';' S*
-		  ;
-		namespace_prefix
-		  : IDENT
-		  ;
+		@namespace <<namespace-prefix>>? [ <<string>> | <<url>> ] ;
+
+		<dfn>&lt;namespace-prefix&gt;</dfn> = <<ident>>
 	</pre>
-
-	with the new token:
-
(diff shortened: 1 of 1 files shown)


SebastianZ:
As I immediately got an error after creating the PR due to not having my GitHub account connected with my W3C account, I've done so now.

svgeesus:
@fantasai do you think we should merge this? It is certainly better to use the more modern syntax. Namespaces 3 is a Rec so that would also mean issuing an erratum (which I can help with, if needed). Or indeed we could go through the steps to republish as an updated Recommendation.

I feel this is an editorial change and has no effect on actual implementations.

(I will deal with the IPR bot false positive, if we merge)

tabatkins:
I think we should do it; the current definition isn't well-founded since CSS isn't defined over that grammar definition any longer.

SebastianZ:
Sounds like this change could just be merged, but please let me know if I still have to change something.

Sebastian

fantasai:
@svgeesus Given css-syntax-3 is a CR now, and it's effectively editorial, yeah, I think it's fine to merge. I'd want @tabatkins to review the changes though, since I'm not quite as familiar with the new syntax definition requirements. :)

Also, hopefully Process 2020 goes through and we should be able to publish an editorial update to the REC without too much hassle. ^_^

tabatkins:
r+

SebastianZ:
Thank you for merging this change, @tabatkins!

Sebastian

PR #4256, opened 2019-08-30
[css-grid-2] Add missed `partial` class in markup

lahmatiy:
Added `partial` class as in other specs


lahmatiy:

PR #4257, opened 2019-08-30
[web-animations-2] Define a title as in other specs

lahmatiy:
That's a single spec that define a title via `<h1>`. Now it defined as in any other spec


lahmatiy:

tabatkins:
(Note that the old form is perfectly okay per Bikeshed; it looks for the first `h1` and takes it as the title of the spec if no `Title` metadata is present. It's also required if you want to put markup in your title. But I do prefer keeping things similar in structure when possible.)

PR #4258, opened 2019-08-30
[css-overflow-3] Change class order in markup as in other specs

lahmatiy:
That's a single spec where order of CSS class goes in that ways. It would be great to keep consistent order of classes in such cases.


lahmatiy:

frivoal:
I don't mind this pull request, but the order of classes in the html class attribute is not significant, so any body relying on a particular order is calling for trouble.

lahmatiy:
@frivoal Yep, the order of classes in not significant for styling, but useful for search and uniform.

PR #4265, opened 2019-09-02
Uniform `Inherited` values across specs

lahmatiy:
With this patch there will be the following list of values for `Inherited`:
- yes
- no
- no (but see prose)
- no (but see prose, above)
- n/a

Not sure we need to left `n/a`, but not sure about replacement


lahmatiy:

PR #4268, opened 2019-09-02
[selectors-4] Make links more precise in selectors overview table

shvaikalesh:
Follow-up of #4214.


shvaikalesh:

PR #4269, opened 2019-09-03
[css-text-3] Disambiguate "space"

### LINKED ISSUE #4267: [css-text] full-width spaces not covered in the collapsible/preserved space definition
According to the last draft of the CSS Text specification, this is the definition of the 'white space' term:

https://drafts.csswg.org/css-text-3/#white-space
> Except where specified otherwise, White space processing in CSS affects only the document white space characters: spaces (U+0020), tabs (U+0009), and segment breaks. 

First of all, I wonder what "Except where specified otherwise" actually means; in my opinion this statement makes harder to understand other dependent terms, like the ones I'm trying to clarify in this issue later.

Then, it's stated in the spec that full-width spaces, like x3000 (ideographic space), are defined as 'other space separators': 

 https://drafts.csswg.org/css-text-3/#other-space-separators
> Besides Space (U+0020) and No-Break Space (U+00A0), Unicode [UNICODE] defines a number of additional space separator characters. In this specification all characters in the Unicode Zs category (See [UAX44]) except Space (U+0020) and No-Break Space (U+00A0) are collectively referred to as other space separators. 

So at this point, I understand that every time I read "white space" in the spec it doesn't apply to full-width spaces like U+3000 (ideographic space). Is this correct ? 

So then we have the definition of "collapsible space", which is specified as follow:

https://drafts.csswg.org/css-text-3/#collapse
> If white-space is set to normal, nowrap, or pre-line, white space characters are considered collapsible

Since the definition above uses the term "white space", I understand that full-width spaces are excluded from that definition; hence ideographic spaces are never collapsible spaces. Is this correct ? 

Then there is the statement about soft wrapping opportunities activated by the white-space property:

> If white-space is set to pre, pre-wrap, or break-spaces, any sequence of spaces is treated as a sequence of non-breaking spaces. However, for pre-wrap, a soft wrap opportunity exists at the end of a sequence of spaces and/or tabs, while for break-spaces, a soft wrap opportunity exists after every space and every tab. 

In the statement above, the term uses is just "space", so I assume that such term includes also the "other space separator" category; hence, there are soft wrap opportunities **at the end** of the sequence of ideographic spaces. Again, I wonder whether this assumption about full-width spaces is correct.

It's also interesting to consider that there may be soft wrap opportunities in the middle of a sequence of full-width spaces (considered as other space separators) if the value of the white-space is either "break-spaces" or "normal"; see your comment on issue #4180, as follows:

> wrapping opportunities are allowed after every ideographic space (cf: https://www.unicode.org/reports/tr14/tr14-39.html#BA)

I guess this depends on the value of the line-break property, correct ? Is this the case of regular white spaces (eg. U+0020) so that 'white-space: normal' may allow wrap opportunities after every space ?

Finally we have the definition of 'preserved white space', which is described by the spec as follows:

https://drafts.csswg.org/css-text-3/#preserved-white-space
> White space that was not removed or collapsed due to white space processing is called preserved white space.

Again, since the statement above uses the term "white space" it implies that any character in the "other space separators" group is excluded; hence, I understand that full-width spaces, like U+3000, can't be considered preserved spaces. Is this correct ? 

These definitions are very important, specially for the specification of the white-space property and its different values, which most of them use the term "white space", without mentioning the 'other space separator' term. They have some implications about how to compute the intrinsic size when there are 'other space separator' characters involved. But lets first clarify the definitions above before entering on these other issues.

frivoal:
Linkify and use non ambiguous terminology for "white space", "spaces",
and related terminology throughout normative prose.

Closes #4267

### DIFF
--- a/css-text-3/Overview.bs
+++ b/css-text-3/Overview.bs
@@ -619,7 +619,7 @@ Text Processing</h3>
     <p>Text transformation happens after [[#white-space-phase-1]]
       but before [[#white-space-phase-2]].
       This means that ''full-width'' only transforms
-      U+0020 spaces to U+3000 within <a>preserved</a> white space.
+      U+0020 spaces to U+3000 within <a>preserved</a> [=white space=].
 
       <wpt>
         text-transform/text-transform-fullwidth-006.html
@@ -788,10 +788,10 @@ Text Processing</h3>
     <dt><dfn>break-spaces</dfn></dt>
     <dd>The behavior is identical to that of ''white-space/pre-wrap'',
       except that:
-      * Any sequence of <a>preserved</a> white space always takes up space,
+      * Any sequence of <a>preserved</a> [=white space=] always takes up space,
         including at the end of the line.
       * A line breaking opportunity exists after every <a>preserved</a> <a>white space</a> character,
-        including between white space characters.
+        including between [=white space characters=].
 
         <wpt>
           overflow-wrap/overflow-wrap-break-word-002.html
@@ -832,7 +832,7 @@ Text Processing</h3>
           white-space/break-spaces-tab-006.html
         </wpt>
 
-      Such preserved spaces take up space and do not [=hang=],
+      Such [=preserved=] [=white space characters=] take up space and do not [=hang=],
       and thus affect the box's intrinsic sizes
       ([=min-content size=] and [=max-content size=]).
 
@@ -841,12 +841,12 @@ Text Processing</h3>
         white-space/white-space-intrinsic-size-002.html
       </wpt>
 
-      <p class="note">Note: This value does not guarantee that there will never be any overflow due to spaces:
-      for example, if the line length is so short that even a single space does not fit,
+      <p class="note">Note: This value does not guarantee that there will never be any overflow due to white space:
+      for example, if the line length is so short that even a single white space character does not fit,
       overflow is unavoidable.</p>
 
     <dt><dfn>pre-line</dfn></dt>
-    <dd>Like ''white-space/normal'', this value collapses consecutive spaces and allows wrapping,
+    <dd>Like ''white-space/normal'', this value collapses consecutive [=white space characters=] and allows wrapping,
       but preserves <a>segment breaks</a> in the source as <a>forced line breaks</a>.
 
       <wpt pathprefix="/css/CSS2/text/">
@@ -941,15 +941,15 @@ Text Processing</h3>
   </table>
 
   <p>See <a href="#white-space-processing">White Space Processing Rules</a>
-    for details on how white space collapses. An informative summary of
+    for details on how [=white space=] collapses. An informative summary of
     collapsing (''white-space/normal'' and ''nowrap'') is presented below:
     <ul>
       <li>A sequence of segment breaks and other <a>white space</a> between two
         Chinese, Japanese, or Yi characters collapses into nothing.
-      <li>A zero width space before or after a white space sequence
+      <li>A zero width space before or after a [=white space=] sequence
         containing a segment break causes the entire sequence of <a>white space</a>
         to collapse into a zero width space.
-      <li>Otherwise, consecutive <a>white space</a> collapses into a single space.
+      <li>Otherwise, consecutive <a>white space</a> collapses into a single [=space=].
     </ul>
 
   <p>See <a href="#line-breaking">Line Breaking</a>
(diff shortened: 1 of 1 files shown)


css-text-3/Overview.bs
```diff
@@ -1237,11 +1237,11 @@ Text Processing</h3>
       </ol>
     </li>
     <li><p>If 'white-space' is set to ''pre'', ''pre-wrap'', or ''break-spaces'',
-      any sequence of spaces is treated as a sequence of non-breaking spaces.
+      any sequence of [=preserved white space=] is treated as a sequence of non-breaking spaces.
```
fantasai: This is incorrect. We only do this for spaces, not for tabs.

css-text-3/Overview.bs
```diff
@@ -1594,7 +1594,7 @@ Text Processing</h3>
 
       <li>A sequence at the end of a line
         (ignoring any intervening <a>inline box</a> boundaries)
-        of <a>collapsible</a> spaces (U+0020)
+        of <a>collapsible</a> [=white space=]
```
fantasai: This needs to say spaces, because line breaks are not necessarily removed.

css-text-3/Overview.bs
```diff
@@ -1549,7 +1549,7 @@ Text Processing</h3>
 
     <p>As each line is laid out,</p>
     <ol>
-      <li>A sequence of collapsible spaces at the beginning of a line
+      <li>A sequence of [=collapsible=] [=white space=] at the beginning of a line
```
fantasai: This should stay spaces as well, since we've converted collapsible white space to spaces.

css-text-3/Overview.bs
```diff
@@ -1670,7 +1670,9 @@ Text Processing</h3>
              Note: [=Hanging=] the white space rather than collapsing it
              allows users to see the space when selecting or editing text.
           <li>If 'white-space' is set to ''break-spaces'',
-             [=hanging=] or collapsing the advance width of the spaces or tabs
+             [=hanging=] or collapsing the advance width of the [=white space=],
+	     [=preserved=] [=tabs=],
+	     or [=other space separators=]
```
fantasai: I think spaces -> white space here wasn't necessary given we're calling out tabs explicitly?

css-text-3/Overview.bs
frivoal: well, we don't do that for tabs that have been preserved as tabs, but tabs (and new lines) that have been converted to spaces and didn't collapse away do count. I was trying to capture that this is not just for U+0020 that have started their life as such, but I guess you're right that tabs that have been preserved as tabs are part of `preserved white space`, and that those are not included.

Possibilities:
* just revert to "spaces", counting on the fact that people will understand that this includes the things that weren't spaces initially but have been converted to spaces at some earlier step. (I'm a little worried about that, because while correct, that stretches the definition of "space" a little bit beyond U+0020, which might make people mistakenly think it includes even more things, such as other space separators).
* rephrase this to make the above point explicit (how?)
* change the definition of `preserved white space` to exclude tabs that have been preserved as tabs, and new lines that have been preserved as new lines, and replace `preserved white space` with "preserved white space, preserved tabs, and preserved newlines" where appropriate.

css-text-3/Overview.bs
frivoal: Aren't **collapsible** line breaks removed (because they're turned into spaces), and it's only preserved non collapsible line breaks that say?

css-text-3/Overview.bs
frivoal: Same question as https://github.com/w3c/csswg-drafts/pull/4269/files#r320544442

css-text-3/Overview.bs
frivoal: This should probably be rephrased, but what I'm trying to do ties into what I was saying in https://github.com/w3c/csswg-drafts/pull/4269/files#r320544043.
This includes:
* preserved spaces (regardless of whether they started started their life as U+0020, U+0009, or a line feed)
* preserved tabs
* other space separators

but does not include preserved new lines.

So if we do the redefining of `preserved white space` proposed in https://github.com/w3c/csswg-drafts/pull/4269/files#r320544043, this would become
> [...] the advance width of the preserved white space, preserved tabs, or other space separators [...]

css-text-3/Overview.bs
fantasai: At this point any white space that was previously something else and is converted to a space *is* a space. What it was is not relevant.

css-text-3/Overview.bs
fantasai: Sure, but now that they're spaces they should just be called spaces.

css-text-3/Overview.bs
fantasai: Same answer

PR #4274, opened 2019-09-04
[css-animations-1][css-backgrounds-4][css-fonts-4] Fix multiplier usage in propdef/descdef

lahmatiy:
- a multiplier should stick to a term
- `#` multiplier should go before `{N,M}`


lahmatiy:

birtles:
I'm not too familiar with CSS values syntax, but it checks out against [the spec](https://drafts.csswg.org/css-values-4/#component-multipliers), and since no one else seems to be around to review this I'll take it.

PR #4280, opened 2019-09-06
[css-backgrounds] computed background-position-*

ewilligers:
Computed length-percentage values for background-position longhands
is now specified using the same terminology as for other properties.

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -17,6 +17,7 @@ Warning: Not Ready
 
 <pre class="link-defaults">
 spec:css-text-4; type:value; text:collapse
+spec:css-shapes-2; type:function; text:path()
 </pre>
 
 <h2 id="intro">
@@ -89,7 +90,7 @@ Background Positioning Longhands: the 'background-position-x', 'background-posit
 		Initial: 0%
 		Inherited: no
 		Percentages: refer to width of background positioning area <em>minus</em> width of background image
-		Computed value: A list, each item consisting of: an offset given as a combination of an absolute length and a percentage, plus an origin keyword
+		Computed value: A list, each item consisting of: an offset given as a computed <<length-percentage>> value, plus an origin keyword
 	</pre>
 
 	This property specifies the background position's horizontal component.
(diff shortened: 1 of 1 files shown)


ewilligers:
computed `<length-percentage>` value terminology is used consistently:

https://drafts.csswg.org/css-position/#propdef-bottom
https://drafts.csswg.org/css-position/#propdef-inset-after
the keyword auto or a computed `<length-percentage>` value

https://drafts.csswg.org/css-scroll-snap-1/#padding-longhands-logical
https://drafts.csswg.org/css-scroll-snap-1/#propdef-scroll-padding
the keyword auto or a computed `<length-percentage>` value

https://drafts.csswg.org/css-align-3/#propdef-column-gap
https://drafts.csswg.org/css-flexbox/#propdef-flex-basis
specified keyword, else a computed `<length-percentage>` value

https://drafts.csswg.org/css-sizing-3/#propdef-max-height
https://drafts.csswg.org/css-sizing-3/#propdef-min-width
as specified, with `<length-percentage>` values computed

https://drafts.csswg.org/css-backgrounds/#background-position
list, each item a pair of offsets (horizontal and vertical) from the top left origin each given as a computed `<length-percentage>` value

https://drafts.csswg.org/css-backgrounds/#border-image-width
four values, each either a number, the keyword auto, or a computed `<length-percentage>` value

https://www.w3.org/TR/css-text-3/#propdef-text-indent
computed `<length-percentage>` value, plus any specified keywords



The exceptions are offset-distance (hence https://github.com/w3c/fxtf-drafts/pull/370) and CSS2 vertical-align:

https://www.w3.org/TR/CSS2/visudet.html#propdef-vertical-align
for `<percentage>` and `<length>` the absolute length, otherwise as specified

The following properties compute lengths:

https://drafts.csswg.org/css-fonts-4/#font-size-prop
an absolute length, as clamped by font-min-size and font-max-size

https://drafts.csswg.org/css-inline-3/#propdef-line-height
the specified keyword, a number, or a computed `<length>` value

LeaVerou:
LGTM, thanks!

PR #4289, opened 2019-09-09
Correct spelling of `id="scollbar-gutter-property"`

Malvoz:
"scollbar" => "scrollbar".


Malvoz:

PR #4290, opened 2019-09-10
[css-text-3] Hang trailing collapsible other space separators instead of removing them

### LINKED ISSUE #4180: [css-text] How to handle leading ideographic space sequences ?
I'm not sure whether the spec covers some extreme cases related to full-width spaces, specifically ideographich spaces (x3000). 

As far as I know, leading ideographich space sequences should not be collapsed, even when the white-space property allows to do so (normal, nowrap, or pre-line). However, I'm not sure what to do in cases like this one:

`<div width: 2em; white-space:normal">&#x3000;&#x3000;XX</div>`

First of all, 'white-space: normal' allows wrapping, as it's [stated](https://drafts.csswg.org/css-text-3/#valdef-white-space-normal) in the spec: 

> Lines may wrap at allowed soft wrap opportunities, as determined by the line-breaking rules in effect, in order to minimize inline-axis overflow. 

The [Phase II: Trimming and Positioning](https://drafts.csswg.org/css-text-3/#white-space-phase-2) section describes several scenarios and steps to evaluate, but I don't know which one this cases should be applied to:

> 1- A sequence of collapsible spaces at the beginning of a line (ignoring any intervening inline box boundaries) is removed. 

In this case, these are full-width spaces, so they can not be considered as collapsible, despite the value of the white-space property (normal).

> 3- A sequence at the end of a line (ignoring any intervening inline box boundaries) of collapsible spaces (U+0020) and/or other space separators whose white-space value collapses spaces is removed. 

If we allow breaking after the last U+3000 space, the space sequence will fill the whole line. Should we apply this step and remove the sequence ? In that case, it'd be the same than applying the step 1, which I think is not what we want for ideographic spaces.

So, I wonder which of the following results would be the expected one, according to the current specs, for this case:

<table>
<tr><td><b>Result 1</b></td><td><b>Result 2</b></td><td><b>Result 3</b></td></tr>
<tr><td></td><td></td><td></td></tr>
</table>

frivoal:
Closes #4180

### DIFF
--- a/css-text-3/Overview.bs
+++ b/css-text-3/Overview.bs
@@ -1595,18 +1595,12 @@ Text Processing</h3>
       <li>A sequence at the end of a line
         (ignoring any intervening <a>inline box</a> boundaries)
         of <a>collapsible</a> [=spaces=]
-        and/or [=other space separators=] whose 'white-space' value collapses spaces
         is removed.
 
         <wpt>
           white-space/line-edge-white-space-collapse-001.html
           white-space/pre-float-001.html
           white-space/pre-wrap-float-001.html
-          white-space/trailing-ideographic-space-001.html
-          white-space/trailing-ideographic-space-002.html
-          white-space/trailing-other-space-separators-001.html
-          white-space/trailing-other-space-separators-003.html
-          white-space/trailing-other-space-separators-004.html
           white-space/trailing-space-before-br-001.html
         </wpt>
         <wpt pathprefix="/css/CSS2/text/">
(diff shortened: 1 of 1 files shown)


css-text-3/Overview.bs
```diff
@@ -1621,10 +1615,20 @@ Text Processing</h3>
           white-space-processing-047.xht
         </wpt>
       <li>If there remains any sequence of <a>white space</a>,
-        [=other space separators=],
-	and/or [=preserved=] [=tabs=]
+        and/or [=other space separators=],
         at the end of a line:
         <ul>
+          <li>If 'white-space' is set to ''white-space/normal'', ''white-space/nowrap'', or ''white-space/pre-line'',
```
kojiishi: I think the condition is 1) when trailing collapsible spaces are removed, or 2) when trailing collapsible spaces hang. I understand we need some more discussion when to hang for `pre-wrap` but is this intention agreeable?

Issue #4293, opened 2019-09-10
[css-writing-modes-3] Fix typo


frogcat:
https://github.com/w3c/csswg-drafts/blob/dc5f3893f1b861b08655a93865a2115935797bb4/css-writing-modes-3/Overview.bs#L2509

2091 -> 2019

PR #4294, opened 2019-09-10
[css-writing-modes-3] Fix typo #4293

frogcat:


ewilligers:

Issue #4008, opened 2019-06-05
[web-animations] Level 2 Spec maintenance and planning


majido:
## Current Status
The current [web animations level 2](https://drafts.csswg.org/web-animations-2/) spec has not been updated and it looks fairly out of date. 

After discussing with other folks working on web-animations spec it was clear that we all agree on this and it was suggested that the two main ideas potentially worth saving from level-2 are:
1. Group Effects
2. Custom Effects

There are at least a few other dependent specifications that I think may require changes in level 2 to ensure better integrations:
- Animation Worklet
- Scroll Timeline

Here is a [full list](https://github.com/w3c/csswg-drafts/labels/web-animations-2) of open issues.

## Group Effect Prototype

We have started looking more closely at Group Effect and we are implementing a prototype in Blink.  As such I want to ensure its specification moves forward as well. 

On the spec side, we have started by writing an [explainer](https://github.com/yi-gu/group_effects/blob/master/README.md) that tries to motivate the feature and also explains fairly minor changes we like to make to the current design. That is a good first step but I want to make sure we have a plan on how to make the necessary web-animations spec changes which brings me to this issue.

## Proposal for Level-2

I propose the following:
1. Archive the current level-2.
2. Make a fresh copy of level-1 as level-2
3. Move over individual features of interest to the new spec, starting with GroupEffect. Update the wording/idl to match the changes.
4. Make any design adjustment we want to make to each feature.
ongoing: Keep level-2 updated by merging current level-1 regularly. 

We can help with this process as we work on Group Effect and other features. :)

I think that gives us a good base to restart the work on level-2 and continue adding features that make sense.

Note that this all assumes we don't expect a whole lot of changes to level-1 anymore which I think is a fair assumption. But please correct me if that is not the case.

birtles:
Thanks for bringing this up. I'm a little concerned about the burden of syncing the two levels, but let's give it a try and see how it goes.

majido:
I had a discussion with @fantasai in F2F about alternative ways to do this. It appears that perhaps we can do a **diff spec**.

While currently bikeshed does not properly support publishing of diff specs but they should be fine as long as they are in draft state. That should remove (or reduce) the burden of syncing between levels.

I will investigate this first and if it turns out that is not idea we can also do as suggested in the original post.

birtles:
It's already a delta spec: https://drafts.csswg.org/web-animations-2/#delta

PR #4302, opened 2019-09-13
[web-animations-2] Update the GroupEffect sections in level 2 spec

yi-gu:
We have started looking more closely at GroupEffect ([explainer](https://github.com/yi-gu/group_effect/blob/master/README.md), [prototype](https://chromium-review.googlesource.com/c/chromium/src/+/1710156)). As such we want to ensure its specification moves forward as well as part of [1].

In this PR we made some fairly minor changes to the current design. PTAL. Thanks!

[1] https://github.com/w3c/csswg-drafts/issues/4008

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -9,7 +9,7 @@ Group: CSSWG
 URL: https://drafts.csswg.org/scroll-animations-1/
 ED: https://drafts.csswg.org/scroll-animations-1/
 Shortname: scroll-animations-1
-Abstract: Defines an API and markup for creating animations that are tied to 
+Abstract: Defines an API and markup for creating animations that are tied to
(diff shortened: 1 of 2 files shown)


birtles:
I tried to build this PR locally but got a bunch of link errors. Is it supposed to work?

Also, any chance we could replace those PNG images with SVG images in a similar style to the existing ones? (Preferably not the UML class diagram style.)

I'm not totally convinced that `StaggerEffect` is necessary. It seems a lot of stagger-type effects use a decreasing delta (e.g. 0s, 1s, 1.8s, 2.5s, 3s, 3.3s, 3.4s) or use some degree of random jitter in the amount of stagger. It would seem like script can easily generate a `ParallelEffect` to achieve these cases and the original `StaggerEffect` case.

The link to the explainer at the start of this thread also appears to be broken.

Issue #4300, opened 2019-09-12
[web-animations-1] Propose moving iterationComposite to web-animations-2


stephenmcgruer:
This is a proposal to move the [iteration composite operation](https://drafts.csswg.org/web-animations-1/#iteration-composite-operation) concept to the level 2 version of the spec. If adopted, [section 5.5](https://drafts.csswg.org/web-animations-1/#iteration-composite-operation) would be removed, as well as all references to the iteration composite operation.

### Motivation

This proposal is primarily motivated by the difficulty of implementing iterationComposite in Chromium (as our interpolation stack would have to change significantly to support it), backed by the claim that we would not lose too much utility in moving it to level 2. The goal is to clear the path for all web browsers to ship web-animations-1 in an interoperable way.

It is hard to judge what iterationComposite brings to the spec. It is not covered in the listed [use-cases](https://drafts.csswg.org/web-animations-1/#use-cases) for web-animations-1. I was unable to find any JavaScript animation libraries offering a similar property in a quick search, but they may exist. If there are use-cases, I would be interested to hear them.

Based on the [WPT tests](https://wpt.fyi/results/web-animations/animation-model/keyframe-effects/effect-value-iteration-composite-operation.html?label=master&label=experimental&aligned), only Firefox implements iteration composite operations today.

### Implications

Although less ergonomic, I believe web developers should be able to mimic the behavior of iterationComposite by setting the correct keyframes on the animation. (Please correct me if this is not the case.) So the main implication is a loss of ergnomics, if the web developer wants to do something iterationComposite-like.

graouts:
I have no objection.

stephenmcgruer:
Spec has been updated; remaining work is to figure out how to clearly differentiate the WPT tests as level-2. In https://github.com/w3c/csswg-drafts/pull/4303 @foolip suggested just using the existing meta tag, and then wpt.fyi might add an ability to filter by spec, however I was just thinking this doesn't really work for things like interface tests...

Still, I will take a look today to see what can be done :)

foolip:
See https://github.com/web-platform-tests/wpt.fyi/issues/65 for where that discussion goes, of if you want to make it go a particular way :)

birtles:
Spec work is done, WPT meta changes are probably less important now that Chrome implements this.

I can see there's at least:

- web-animations/interfaces/KeyframeEffect/iterationComposite.html
- web-animations/animation-model/keyframe-effects/effect-value-iteration-composite-operation.html

[...]

PR #4303, opened 2019-09-13
[web-animations-1][web-animations-2] Move iterationComposite to web-animations-2

stephenmcgruer:
This PR moves the `iterationComposite` concept from the level 1 spec to the level 2 spec, as motivated in https://github.com/w3c/csswg-drafts/issues/4300

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -1247,7 +1247,8 @@ Update the description of an effect that targets a non-animated property as:
 The procedure for computing the <a>effect value</a> of a single property
 referenced by a <a>keyframe effect</a> as one of its <a>target properties</a>,
 for a given <var ignore=''>iteration progress</var>, <var>current iteration</var> and
-<var ignore=''>underlying value</var> is amended by inserting the following step:
+<var ignore=''>underlying value</var> is amended by inserting the following step
+after the step to apply the keyframe effect composite mode.
 
 12. For each <var>keyframe</var> in <var ignore=''>interval endpoints</var>:
 
@@ -2117,7 +2118,7 @@ SequenceEffect implements GroupEffectMutable;
 The <a>KeyframeEffect</a> interface is modified to add the following:
 
 <pre class='idl'>
-interface KeyframeEffect : AnimationEffect {
+partial interface KeyframeEffect {
     attribute IterationCompositeOperation    iterationComposite;
 };
 
@@ -2188,10 +2189,10 @@ new SequenceEffect(
 <h4 id="the-keyframeeffectoptions-dictionary">The KeyframeEffectOptions dictionary</h4>
 
 The <a>KeyframeEffectOptions</a> dictionary interface is modified to add the
-following attribute:
+following member:
 
 <pre class='idl'>
-dictionary KeyframeEffectOptions : EffectTiming {
+partial dictionary KeyframeEffectOptions {
     IterationCompositeOperation iterationComposite = "replace";
 };
 </pre>
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -1229,6 +1229,42 @@ Update the description of an effect that targets a non-animated property as:
 >     delaying the fulfilment of an <a>animation</a>'s <a>current finished
 >     promise</a>.
 
+<h3 id="keyframe-effects">Keyframe Effects</h3>
+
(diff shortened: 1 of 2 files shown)


stephenmcgruer:
Antoine, Brian - not sure what we should do with WPT tests for `iterationComposite` - should they be marked tentative, are we somehow else differentiating level-2 tests, or are we just leaving them all and its up to each of us to maintain what we think are the level-1-relevant tests?

flackr:
I'd suggest marking them tentative for now, although perhaps longterm it would make sense to differentiate spec levels in wpt. I seem to recall ideas around including or referencing the test results from the spec at which point being able to differentiate spec level seems desirable. Thoughts, @foolip ?

foolip:
If level 2 is published somewhere then they need not be tentative. You can add a link to the spec and if you need to filter based on those links perhaps we can come up with something in wpt.fyi?

stephenmcgruer:
> perhaps we can come up with something in wpt.fyi?

I think this is a great idea! I'm not sure how widespread the use of the `<link rel="help" href="spec-link">` tags are, but the ability to search the whole of wpt.fyi by spec would be pretty neat, especially as we look at different levels of the same spec. Though to be fair, I'm not sure if anyone consistently keeps those separate so maybe that would limit impact...

foolip:
Feature request is at https://github.com/web-platform-tests/wpt.fyi/issues/65.

web-animations-2/Overview.bs
```diff
@@ -1229,6 +1240,42 @@ Update the description of an effect that targets a non-animated property as:
...
 >     promise</a>.
 
+<h3 id="keyframe-effects">Keyframe Effects</h3>
+
+<h4 id="the-effect-value-of-a-keyframe-animation-effect">The effect value of a keyframe effect</h4>
+
+The procedure for computing the <a>effect value</a> of a single property
+referenced by a <a>keyframe effect</a> as one of its <a>target properties</a>,
+for a given <var ignore=''>iteration progress</var>, <var>current iteration</var> and
+<var ignore=''>underlying value</var> is amended by inserting the following step:
```
birtles: ...is amended by inserting the following step after the step to apply the keyframe effect composite mode.

web-animations-2/Overview.bs
```diff
@@ -2043,12 +2114,48 @@ SequenceEffect implements GroupEffectMutable;
 <h3 id="the-keyframeeffect-interfaces">The <code>KeyframeEffectReadOnly</code>
   and <code>KeyframeEffect</code> interfaces</h3>
 
-Add:
+The <a>KeyframeEffect</a> interface is modified to add the following:
 
 <pre class='idl'>
+interface KeyframeEffect : AnimationEffect {
```
birtles: I believe this should be `partial interface`?

web-animations-2/Overview.bs
```diff
@@ -2078,6 +2185,46 @@ new SequenceEffect(
 
 </div>
 
+<h4 id="the-keyframeeffectoptions-dictionary">The KeyframeEffectOptions dictionary</h4>
+
+The <a>KeyframeEffectOptions</a> dictionary interface is modified to add the
+following attribute:
+
+<pre class='idl'>
+dictionary KeyframeEffectOptions : EffectTiming {
```
birtles: `partial dictionary`

web-animations-2/Overview.bs
```diff
@@ -2078,6 +2185,46 @@ new SequenceEffect(
 
 </div>
 
+<h4 id="the-keyframeeffectoptions-dictionary">The KeyframeEffectOptions dictionary</h4>
+
+The <a>KeyframeEffectOptions</a> dictionary interface is modified to add the
+following attribute:
```
birtles: following member

web-animations-2/Overview.bs
stephenmcgruer: Ah, TIL about `partial interface`s, thanks :)

web-animations-2/Overview.bs
stephenmcgruer: Done.

web-animations-2/Overview.bs
stephenmcgruer: Done.

web-animations-2/Overview.bs
stephenmcgruer: Done.

web-animations-2/Overview.bs
stephenmcgruer: Done.

Issue #4299, opened 2019-09-12
[web-animations-1] Propose moving mutable timelines to web-animations-2


stephenmcgruer:
This is a proposal to move the [mutability of timelines](https://drafts.csswg.org/web-animations-1/#dom-animation-timeline) on the Animation interface to the level 2 version of the spec. If adopted, the timeline attribute of the [Animation interface](https://drafts.csswg.org/web-animations-1/#animation) would be marked **readonly** in the level 1 spec.

### Motivation

This proposal is motivated on two fronts: firstly, the utility of mutable timelines in web-animations-1, and secondly the difficulty of implementing them in Chromium (our implementation presumes a lot about the timeline not changing). The goal is to clear the path for all web browsers to ship web-animations-1 in an interoperable way.

In https://github.com/w3c/csswg-drafts/issues/3759 I requested input on the use case for `timeline` being writable. The conversation was focused on ScrollTimeline, but no other usecases for mutable timelines were presented. Briefly scanning the spec, I did not find mutable timelines required for any of the use-cases.

To be clear; I have no disagreement with mutable timelines, they just don't have clear utility to me in web-animations-1 today.

Based on the [WPT tests](https://wpt.fyi/results/web-animations/interfaces/Animation/idlharness.window.html?label=master&label=experimental&aligned), I believe timelines are currently mutable in Firefox and Safari, and not mutable in Chromium.

### Implications

There are three transitions possible with the web-animations-1 spec today:
  * moving from one DocumentTimeline to another,
  * moving from a DocumentTimeline to null,
  * moving from null to a DocumentTimeline.

The first of these does not seem to have much use. The most one can do with a different DocumentTimeline is apply an offset, which could also be done by modifying the Animation's timing.

Going to/from null does have an effect - it makes the current time null/non-null, which then essentially makes the Animation 'happen' or not. However it is unclear to me if there is a situation where a web developer would prefer this over pausing, finishing, or cancelling their animation (and if there is such a situation, it feels like that highlights a hole in our API surface?). I would welcome input if someone believes there is an important use-case here.

graouts:
I agree too, I don't see any pressing practical use for this feature.

PR #4304, opened 2019-09-13
[web-animations-1][web-animations-2] Move mutable timelines to web-animations-2

stephenmcgruer:
Closes https://github.com/w3c/csswg-drafts/issues/4299

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3996,7 +3996,7 @@ API by the {{Animation}} interface.
 interface Animation : EventTarget {
              attribute DOMString                id;
              attribute AnimationEffect?         effect;
-             attribute AnimationTimeline?       timeline;
+    readonly attribute AnimationTimeline?       timeline;
              attribute double?                  startTime;
              attribute double?                  currentTime;
              attribute double                   playbackRate;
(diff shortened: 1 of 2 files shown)


stephenmcgruer:
I dont *think* anything else needs updated, but I was surprised at how small this change was.

stephenmcgruer:
> I think section 4.4.1. Setting the timeline of an animation should be moved to level 2, right

@birtles 

I'm undecided about this. It is still used in the `Animation (effect, timeline)` constructor. That said, the timeline can't change so the steps there are probably irrelevant (have not confirmed yet). I'll need to convince myself that `setting the timeline of an animation` can't change anything during the `Animation` constructor, will look at that today :)

(Can't get github to actually show me the comment in the file, I think because it must have been made outside of the diff range? Not sure; clicking 'View changes' on your comment shows me nothing at all...)

birtles:
> (Can't get github to actually show me the comment in the file, I think because it must have been made outside of the diff range? Not sure; clicking 'View changes' on your comment shows me nothing at all...)

I think I probably made it as a review comment as opposed to an inline comment.

> I'm undecided about this. It is still used in the `Animation (effect, timeline)` constructor. That said, the timeline can't change so the steps there are probably irrelevant (have not confirmed yet). I'll need to convince myself that `setting the timeline of an animation` can't change anything during the `Animation` constructor, will look at that today :)

I was just going off where bikeshed said that procedure was linked-to from. I didn't actually check the constructor itself so if it is used there, then we should probably leave it in level 1.

stephenmcgruer:
Ack. I switched to a partial interface for `Animation`, I think otherwise this is good to go.

birtles:
Thanks. I suspect we can simplify the set the timeline procedure if it is only used by the `Animation` constructor but perhaps we can do that later.

stephenmcgruer:
Hrm, seems like neither https://drafts.csswg.org/web-animations-2/ nor the level 1 version have been updated automatically after this merge. Do we know what process usually generates them and how to check on it? @birtles @graouts

birtles:
Sometimes the spec generation process gets blocked. I normally ping @tabatkins when that happens.

birtles:
It appears to have run now.

Issue #4288, opened 2019-09-09
[css-transitions-2] transition-timing-function and getComputedTiming()


george-steel:
Should the easing function specified in `transition-timing-function` be applied to the global timing function of the CSSTransition's KeyframeEffect (and be returned by `getComputedTiming()`) or should it be applied to the first keyframe? First keyframe is consistent with CSSAnimation while global timing seems to be implied by the reversing behavior, as the global progress it talks about is affected by the easing function and it would make sense if getComputedTiming returned this value.

birtles:
That's a great question. In Firefox we apply it to the first keyframe for consistency with CSS animations, but I can see how it might be more intuitive for an author if it applied to the effect instead (so you can substitute out the keyframe values without affecting the easing).

Furthermore, regarding the reversing behavior you identified, it looks like we ran into the same issue in Firefox and have an outstanding issue to moving the easing to the effect instead ([code comment](https://searchfox.org/mozilla-central/rev/588814f2edddf0e132d77d326ddae50911e8bad1/layout/style/nsTransitionManager.cpp#695-699), [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1292001)).

graouts:
We discussed this during the [2019-09-10 telcon](https://docs.google.com/document/d/1j3oeLpgX6kneG9lEgZQ80kitdyN-vbAYGPiwtYyN3tw) and I agree that this is desirable. I think at this stage we have general agreement to pursue this, @george-steel could you make a spec edit and create a pull request?

PR #4306, opened 2019-09-13
[css-transitions-2] Add section of transition-timing-function and getComputedTiming (issue #4288)

george-steel:
[css-transitions-2] Add section of transition-timing-function and getComputedTiming (issue #4288)

### DIFF
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -80,15 +80,16 @@ sub-property that is animatable. The <dfn>expanded transition property
 name</dfn> of a transition is the name of the longhand sub-property for which
 the transition was generated (e.g. 'border-left-width').
 
-## The transition timing function ## {#teransition-timing-function}
+## The transition timing function ## {#transition-timing-function}
 
 When creating a {{KeyframeEffect}} for a transition, the effect's {{EffectTiming/easing}}
 option is set to the value of the matching 'transition-timing-function' property,
 which must be used by {{AnimationEffect/getComputedTiming()}}.
 
-Note, this behavior was chosen over the {{CSSAnimation}} behavior of setting
-the timing functions of individual keyframes in order to make use of the
-<a>transformed progress</a> with reversing a transition.
+Note: Unlike CSS animations, the timing function for CSS transitions applies to
+the animation effect as opposed to the individual keyframes since this allows
+it to be reflected in the <a>transformed progress</a> as used when calculating
+the reversing shortening factor.
 
 ## Owning element ## {#owning-element-section}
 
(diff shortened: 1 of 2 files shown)


css-transitions-2/Overview.bs
```diff
@@ -80,6 +80,16 @@ sub-property that is animatable. The <dfn>expanded transition property
 name</dfn> of a transition is the name of the longhand sub-property for which
 the transition was generated (e.g. 'border-left-width').
 
+## The transition timing function ## {#teransition-timing-function}
```
ewilligers: transition-...

css-transitions-2/Overview.bs
```diff
@@ -80,6 +80,16 @@ sub-property that is animatable. The <dfn>expanded transition property
...
+
+When creating a {{KeyframeEffect}} for a transition, the effect's {{EffectTiming/easing}}
+option is set to the value of the matching 'transition-timing-function' property,
+which must be used by {{AnimationEffect/getComputedTiming()}}.
+
+Note, this behavior was chosen over the {{CSSAnimation}} behavior of setting
+the timing functions of individual keyframes in order to make use of the
+<a>transformed progress</a> with reversing a transition.
```
ewilligers: s/with/when/

css-transitions-2/Overview.bs
```diff
@@ -80,6 +80,16 @@ sub-property that is animatable. The <dfn>expanded transition property
 name</dfn> of a transition is the name of the longhand sub-property for which
 the transition was generated (e.g. 'border-left-width').
 
+## The transition timing function ## {#teransition-timing-function}
+
+When creating a {{KeyframeEffect}} for a transition, the effect's {{EffectTiming/easing}}
+option is set to the value of the matching 'transition-timing-function' property,
+which must be used by {{AnimationEffect/getComputedTiming()}}.
+
+Note, this behavior was chosen over the {{CSSAnimation}} behavior of setting
```
birtles: If we use `Note:` instead, does this get formatted as a note?

birtles: Also, rather than referring to `CSSAnimation`, it might make more sense to just refer to CSS Animations. e.g.

```
Note: Unlike CSS animations, the timing function for CSS transitions applies to the animation effect as opposed to the individual keyframes since this allows it to be reflected in the [=transformed progress=] as used when calculating the reversing shortening factor.
```

css-transitions-2/Overview.bs
george-steel: done

css-transitions-2/Overview.bs
george-steel: n/a due to Brian's change

css-transitions-2/Overview.bs
george-steel: done. Thanks for the reword.

PR #4370, opened 2019-09-25
[css-fonts-4] Clarify the behavior of the 'very small' block period, …

### LINKED ISSUE #4108: [css-fonts-4] font-display: optional without relayout
https://drafts.csswg.org/css-fonts-4/#font-display-desc

I originally thought `font-display: optional` was designed to provide web fonts in a way that avoided swapping or flash-of-invisible-text, at the expense of getting a fallback font for the first visit. This is kinda backed up by a note in the spec:

> the font is used if it’s already downloaded and available, but otherwise a fallback is used for the rest of the page’s lifetime instead. The font might download in the background and be available to future page loads

However, the prose doesn't really define the behaviour.

https://static-misc.glitch.me/optional-font-load/

In Chrome & Firefox at least, the page may briefly render with invisible text (using layout from a fallback font), causing the green bar to move when the page is laid out again using the web font. This is the kind of layout instability I thought `font-display: optional` was designed to avoid.

**Proposal:**

When the browser discovers a part of the page that needs a `font-display: optional` web font, it must:

1. Block rendering.
1. Makes an [`"only-if-cached"`](https://fetch.spec.whatwg.org/#concept-request-cache-mode) fetch for the resource.
1. Let *font* be null.
1. If there's a valid font response that isn't stale, set *font* to that resource.
1. Otherwise, set *font* to the result of getting the next item in the `font-family` stack.
1. Unblock rendering (if it hasn't already been unblocked during the previous line).
1. If there was no response, or the response was stale, make a low-priority request for the font resource (just to populate the cache for next time).

We already block rendering when getting regular font resources from disk, so blocking rendering shouldn't really be a problem here, unless we expect the above to be significantly slower than loading installed fonts.

Do you think we can safely change `font-display: optional` to behave similar to above? If not, do we need a new value that behaves like this?

tabatkins:
…to resolve #4108.

@litherum Would you mind reviewing? I wasn't sure if you wanted to maintain ownership of the font-display section of the spec, or if it was okay for me to still make edits.

Also @jakearchibald, does this new wording work for you?

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -2668,8 +2668,7 @@ for example, forcing all fonts to have a ''0s'' <a>block period</a>.
 
 	<dt><dfn>fallback</dfn>
 	<dd>
-		Gives the font face an extremely small <a>block period</a>
-		(''100ms'' or less is recommended in most cases)
+		Gives the font face a [=no-network=] <a>block period</a>
 		and a short <a>swap period</a>
 		(''3s'' is recommended in most cases).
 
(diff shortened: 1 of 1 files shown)


litherum:
This is confusing because it sounds like “no network” describes behavior, but it’s being presented in a context where the term is being presented as a simple duration.

Are you sure this concept isn’t better described as behaviors (like “check the cache but not the network”) instead of simple durations?

tabatkins:
Hm, I see your point about the naming.

I tried describing the behavior more explicitly first, but since it needs to apply to two values mostly identically, it felt really weird.  *That said*, reviewing this again with fresh hours-later eyes, I'm still not actually addressing Jake's point, which is that we don't want it to render with invisible-fallback at all if that's avoidable (but phrasing in terms of block periods still allows exactly that behavior). I'll take another crack at it tomorrow.

jakearchibald:
Do we have the concept of "block the rendering" anywhere else? I think we only have parser blocking.

But yeah, it seems worth defining, since it's how font rendering works for all fonts that aren't web fonts.

tabatkins:
Not explicitly, but I'm comfortable hand-waving about it.

svgeesus:
@jakearchibald  I'm reading your https://github.com/w3c/csswg-drafts/pull/4370#issuecomment-535387079 as a "changes requested" review, to clarify the issue mentioned by @litherum is that right?

svgeesus:
pinging @jakearchibald to clarify whether you request changes or not. @litherum I agree with your point. @tabatkins could you reword and if necessary rebase so we can merge this?

jakearchibald:
Sorry for missing this. 

I filed #4108 because browsers have layout instability when using `font-display: optional`.

"block" in this spec seems to mean "block the rendering of this font". It doesn't seem to mention the rendering of other content that loads afterwards. Right now, when a font is blocked, it's still laid out using the fallback font. Then, the intended font loads, and the page is re-laid out with the metrics of the loaded font. This creates layout instability.

It feels like the solution needs to define a new kind of "block", one that goes beyond the text. It needs to block all further layout until it's had a chance to see if that font is immediately available.

The changes here don't seem to require that kind of layout stability, so I guess I'm requesting changes, yeah.

jakearchibald:
(but also I'm not really part of this group or familiar with how strictly things should be defined, so if you all think I'm wrong, ignore me)

tabatkins:
Actually, the changes already made it into the spec, so I should close this PR. ^_^  @jakearchibald, feel free to review <https://drafts.csswg.org/css-fonts-4/#valdef-font-face-font-display-optional> to make sure it matches what you want; I'm pretty sure it does now, tho in a declarative way (requiring no layout shifts and allowing a delay to achieve that) rather than defining what it means to block the rest of layout.

PR #4374, opened 2019-09-26
[web-animations-1] Fix box-shadow addition/accumulation

stephenmcgruer:
To align with other list-list properties, box-shadow
addition/accumulation is changed to be defined as concatenation for
addition (previously component-wise addition) and component-wise
addition for accumulation (previously unspec'd).

Fixes https://github.com/w3c/csswg-drafts/issues/4330

Issue #4330, opened 2019-09-18
[css-backgrounds-3] Clarification on animation additive composite behavior for box-shadow


stephenmcgruer:
Currently the spec for box-shadow (https://drafts.csswg.org/css-backgrounds/#box-shadow) does not appear (to my quick reading) to give any instructions for additive composition (https://drafts.csswg.org/web-animations-1/#effect-composition) when animating.

In the absence of a behavior for addition, the fallback is to use replace (https://drafts.csswg.org/css-values-4/#combining-values).

Firefox instead appears to implement box-shadow interpolation as list concatenation, as shown by https://output.jsbin.com/pikiguv. Chrome appears to do component-wise addition. Which, if either, is correct?

EDIT: Note, you will need composited modes enabled in the browsers if you test them. For Chrome, that's chrome://flags, set "experimental-web-platform-features" to enabled. For Firefox, you can test using Firefox Nightly (has it enabled by default).

stephenmcgruer:
cc @birtles

stephenmcgruer:
For the record, generally with list-y things, add is spec'd as list concatenation and accumulate is spec'd as component-wise addition. 

Also, it's very possible this is spec'd for `box-shadow` and I overlooked it, so please lmk if so.

flackr:
From a developer ergonomics point of view, I think concatenation makes more sense. It's as if you are adding another shadow. As for what is correct - it is animatable as a shadow list. The definition for how to animate a shadow list moved to web-animations-1 from css-transitions-1 https://www.w3.org/TR/web-animations-1/#animating-shadow-lists which suggests it should be component-wise.

stephenmcgruer:
Dammit, I though we killed all the overrides in the animations specs -_-. I didn't realize they had jumped to web-animations-1 from css-transitions-1 :(.

I would quite like to kill https://www.w3.org/TR/web-animations-1/#animation-types; what is stopping us from moving the information into the relevant specs?

`text-shadow` has the same interop discrepancy btw, and I don't think it made it to web-animations-1

PR #4375, opened 2019-09-26
[css-text-decor-3] Specify animation addition/accumulation behavior for text-shadow

stephenmcgruer:
Previously there was no definition of text-shadow composition. To match the behavior of other list-like properties such as transform lists (https://drafts.csswg.org/css-transforms-2/#combining-transform-lists) and box-shadow (https://github.com/w3c/csswg-drafts/pull/4374), this PR specs addition as list concatenation and accumulation as component-wise addition.

See https://github.com/w3c/csswg-drafts/issues/4330

### DIFF
--- a/css-text-decor-3/Overview.bs
+++ b/css-text-decor-3/Overview.bs
@@ -899,9 +899,10 @@ concatenation such that <var ignore>V<sub>result</sub></var> is equal to
 <var>V<sub>a</sub></var> [=list/extended=] with <var>V<sub>b</sub></var>.
 
 <a lt="value accumulation">Accumulation</a> of two text-shadow lists
-<var>V<sub>a</sub></var> and <var>V<sub>b</sub></var> follows the same
-approach as interpolation of two text-shadow lists. The algorithm is defined
-in the <b>Animation type</b> section of 'text-shadow'.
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
Ping and cc @birtles

fantasai:
If there's already a definition for addition and accumulation of shadows, why aren't we reusing it? I don't think text-shadow should get its own one-off definition here.

stephenmcgruer:
This may be overly pedantic. I'm not very familiar with box-shadow and text-shadow, so it's unclear to me whether they differ enough to have different definitions for addition/accumulation. I would have thought if they were alike enough then we would have extracted a common value for them out and could define things like computed-style and animation on that?

But if a `css-text-decor-3` would rather we just point at https://drafts.csswg.org/web-animations-1/#animating-shadow-lists, I'm happy to do that. I feel it is important to do *something* here, since Chrome and Firefox ended up with different behaviors due to this not being defined in the specs.

fantasai:
Committed cross-references between text-shadow and Web Animations' shadow lists. Closing out this PR. @stephenmcgruer Let me know if there's any problem (preferably by filing a new issue; I'm unlikely to notice replies here, tbh).

Issue #94, opened 2016-02-02
[css-tables] Calc for table layout


tomalec:
Are there any plans to specify `calc` behavior for widths and heights in table layout?
Implementations varies across browsers (http://jsbin.com/juqafi/1/edit?html,css,output)

> Given the complexities of width and height calculations on table cells and table elements, math expressions involving percentages for widths and heights on table columns, table column groups, table rows, table row groups, and table cells in both auto and fixed layout tables MAY be treated as if auto had been specified.

I suspect the complexity of the math grows when it comes to mix unit (personally, I feel it's just a set of mathematical equations, to be solved in not more sophisticated way that sizing algorithm for Grid Layout).

However, I would say that expressions like:
- `calc(10% + 20%)` http://jsbin.com/juqafi/1/edit?html,css,output
- `calc(100%/7)` https://bugzilla.mozilla.org/show_bug.cgi?id=957915

Should be really easy to compute, resolve, and apply.

miyconst:
+1

tabatkins:
The Tables module is now receiving some love, so we'll see!

I agree that there are some "trivial" calc()s that *should* be possible without making anything complicated, but because of that, they're also avoidable.  It's *easier* to write `calc(100% / 7)` than `14.28%`, but the diff isn't huge either.

FremyCompany:
I don't think this is reflected in the latest CSS Tables draft, but some investigations on this exact topic were conducted by yours truly.

The sad conclusion is that there is probably no way to allow what you want in a way that would work continuously, as width:0% and width:0px have very different behaviors. Heck, even width:10% can have a different behavior than width:10% if specified on consecutive but otherwise perfectly identical column, tables are a complete mess. CSS Grids resolve width distribution in a way that makes it possible to seamlessly mix units, but CSS Tables don't. 

I feel no difficulty in supporting calc expression resolving to only one particular unit, but that might mislead authors in believing that calc is entirely supported. 

I do not entirely discard the idea of making calc work, but I suspect this cannot make the cut for our current level of the spec, because we would need to add a few new features which is against the spirit of the spec right now, which would add fluidity to the currently one-or-zero behavior of the width distribution. We have so many challenges on that topic already...

FremyCompany:
[...]

Issue #3482, opened 2019-01-04
[css-values] Differences between calc() handling for <length-percentage> values.


emilio:
I was doing some work in Gecko and realized various inconsistencies with how we treat `calc()`. I saw that other browsers have other (different) inconsistencies. I'm fixing some of those where it seems reasonable and we all agree. But I'm having trouble to find what the model is supposed to be.

I'll try to break this down to simpler questions...

### Should `calc(%)` always be treated the same as `%`

It looks to me like it should, but in Gecko right now this is not always the case. For example, in the containing-block-size quirk, Gecko will only apply it to non-calc percentages, intentionally:

https://searchfox.org/mozilla-central/rev/ecf61f8f3904549f5d65a8a511dbd7ea4fd1a51d/layout/generic/ReflowInput.cpp#2162

```html
<style>
  body { margin: 0 }
  div {
    height: 100%;
    float: left;
    width: 50%;
    background: green;
  }
  div + div {
    height: calc(100%);
    background: orange;
  }
</style>
<div></div>
<div></div>
```

### Should `calc(% + 0px)` be equivalent to `calc(%)`?

I intuitively thought it should, but browsers differ here. For example, these two tables look different in blink / webkit:

```html
<!doctype html>
<table border>                                                                                                                                                                                                     
  <tr>                                                                                                                                                                                                             
    <td style="width: calc(50%)">x</td>                                                                                                                                                                            
    <td style="width: 100px">y</td>                                                                                                                                                                                
</table>
<table border>                                                                                                                                                                                                     
  <tr>                                                                                                                                                                                                             
    <td style="width: calc(50% + 0px)">x</td>                                                                                                                                                                            
    <td style="width: 100px">y</td>                                                                                                                                                                                
</table>
```

I think they should render the same though (though I could be convinced otherwise).

### Should `calc(<length> + 0%)` be the same as `calc(<length>)`?

I tend to think it should **not**, since we special-case percentages in quite a few places, and `0%` is not the same as `0px` everywhere (I'm pretty sure it's not the same in grid track sizing for example). Also intrinsic sizing, and this is a test passing in all browsers:

  https://searchfox.org/mozilla-central/rev/ecf61f8f3904549f5d65a8a511dbd7ea4fd1a51d/testing/web-platform/tests/css/css-sizing/intrinsic-percent-non-replaced-001.html#38

[...]

PR #4389, opened 2019-09-30
[css-values] Change MAY into MUST regarding calc() on table

xiaochengh:
#94 has made a resolution that certain `calc()` mixing both lengths and percentages MUST be treated as `auto` instead of MAY. Updating spec accordingly.

Relevant WPT are updated in https://chromium-review.googlesource.com/c/chromium/src/+/1832619.

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -3333,7 +3333,7 @@ Computed Value</h3>
 	math expressions mixing both percentages and lengths for widths and heights on
 	table columns, table column groups, table rows, table row groups, and table cells
 	in both auto and fixed layout tables
-	MAY be treated as if ''width/auto'' had been specified.
+	MUST be treated as if ''width/auto'' had been specified.
 
 
 <h3 id='calc-range'>


davidsgrogan:
At first glance the simple s/MAY/MUST/ seems to prohibit treating calc(10% + 0) as 10%.

> <heycam> emilio: what about calc(10% + 0)?
> <heycam> ... that's simplifies to 10% in all browsers
> <heycam> Rossen: yes we've resolved that previously

emilio:
That is a good point... https://github.com/w3c/csswg-drafts/issues/3482#issuecomment-451590359 seems to imply it should be treated the same:

> > calc(% + 0px) vs calc(%)
>
> Sure. There's no case in which these two should be different. Lengths never have magical behavior, so adding it to a % shouldn't add any complications, and in the rare cases where length+% is different than just resolving the % and adding the two together (bg-position), a 0px length is still a no-op.

css-values-4/Overview.bs
```diff
@@ -3333,7 +3333,7 @@ Computed Value</h3>
 	math expressions mixing both percentages and lengths for widths and heights on
```
xiaochengh:
```suggestion
	math expressions mixing both percentages and non-zero lengths for widths and heights on
```
According to #3482, it is safe to drop 0 length from <length-percentage>. Then how about making it explicit here?

FremyCompany:
LGTM (but I'm not the editor of that spec)

css-values-4/Overview.bs
davidsgrogan: > According to #3482, it is safe to drop 0 length from <length-percentage>. Then how about making it explicit here?

SGTM. As far as I can tell, the resolution from https://github.com/w3c/csswg-drafts/issues/3482 isn't in any spec. Are you proposing to add it here explicitly?

xiaochengh: @tabatkins Could you help?

I'm not familiar with the editing process. What should we do here?

PR #4400, opened 2019-10-08
[css-values-4] Fix typos in comma-separated-list multiplier notes

lahmatiy:
- Added missed chars in example definition, since it should be as original defined
- Remove doubling `<` and `>` around "pairing", since it renders odd

image


lahmatiy:

PR #4403, opened 2019-10-09
[css-inline-3] Use ASCII filenames

SamB:
The server seems not to like non-ASCII filenames, so ...

----

This is neither normative nor substantive, just an attempt to get https://drafts.csswg.org/css-inline-3/#drop-initial to actually display all of the figures.

I am not aware of any patents covering this technique.


SamB:
Another option, of course, would be to fix whatever is going wrong on the server.

tabatkins:
Eh, I'd rather go ahead and push a testcase for desirable behavior thru the system, rather than forever limit ourselves.

plinss:
Pushing a test case through everything would be fine, so long as all the stakeholders in the various systems are on board and available to deal with repercussions. Personally I have higher priority things to work on than chasing this on the draft server at the moment. Especially when the work-around is so trivial.

Bear in mind, we're not just necessarily talking about one or two programs, but potential OS and file system issues across multiple versions of multiple OSes, as well as a long-tail of "who's going to get bit by this down the road just because they tried to download a spec and whatever tool didn't encode the filenames properly". Just seems like a lot of work for very little reward here.

Also, this isn't a "we'll have to live with it forever" thing, it's more of a question of "is this worth dealing with right now?" cost vs benefit thing.

PR #4408, opened 2019-10-10
[cssom-view] Update {add,remove}Listener() to hook into DOM correctly

### LINKED ISSUE #4407: [cssom-view] The addListener(listener) spec doesn't match with current DOM spec
https://drafts.csswg.org/cssom-view/#example-5ecb609e

> The addListener\(listener\) method must run these steps:​
> 
> If listener is null, terminate these steps\.
> 
> Append an event listener to the associated list of event listeners with type set to change, callback set to listener, and capture set to false, unless there already is an event listener in that list with the same type, callback, and capture\.
> 
> The removeListener\(listener\) method must run these steps:​
> 
> Remove an event listener from the associated list of event listeners, whose type is change, callback is listener, and capture is false\.

This doesn't hook into DOM correctly anymore, I believe. It probably did when it was written but the DOM spec has since changed.

See https://dom.spec.whatwg.org/#dom-eventtarget-addeventlistener

cssom-view should say something like

> The `addListener(callback)` method, when invoked, must run these steps:
> 1. [Add an event listener] with the [context object] and an [event listener] whose **type** is `change`, **callback** is *callback*, **capture** is false, **passive** is false, and **once** is false.
>
> The `removeListener(callback)` method, when invoked, must run these steps:
> 1. If the [context object]’s [event listener list] [contains] an [event listener] whose **type** is `change`, **callback** is *callback*, and **capture** is false, then [remove an event listener] with the [context object] and that [event listener].

Related tests: https://github.com/web-platform-tests/wpt/pull/18390 cc @shvaikalesh @annevk

zcorpan:
Fixes #4407.

Tests: https://github.com/web-platform-tests/wpt/pull/18390

---

@annevk are some of the terms here not exported in DOM but should be?

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -771,8 +771,8 @@ When asked to <dfn>evaluate media queries and report changes</dfn> for a {{Docum
 interface MediaQueryList : EventTarget {
   readonly attribute CSSOMString media;
   readonly attribute boolean matches;
-  void addListener(EventListener? listener);
-  void removeListener(EventListener? listener);
+  void addListener(EventListener? callback);
+  void removeListener(EventListener? callback);
            attribute EventHandler onchange;
 };
 </pre>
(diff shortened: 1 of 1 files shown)


cssom-view-1/Overview.bs
```diff
@@ -783,22 +783,24 @@ the associated <a>media</a>.
...
-1. Append an <a>event listener</a> to the associated list of <a>event listeners</a>
-    with <b>type</b> set to <code>change</code>,
-    <b>callback</b> set to <var>listener</var>,
-    and <b>capture</b> set to false,
-    unless there already is an <a>event listener</a> in that list
-    with the same <b>type</b>, <b>callback</b>, and <b>capture</b>.
+1. <a spec=dom>Add an event listener</a> with the <a spec=dom>context object</a>
+    and an <a spec=dom>event listener</a> whose
+    <a spec=dom for="event listener">type</a> is <code>change</code>,
+    and <a spec=dom for="event listener">callback</a> is |callback|.
```
emilio: This should probably specify what happens when `callback` is null explicitly, which is nothing?

Though I guess per https://dom.spec.whatwg.org/#add-an-event-listener step 2 we're alright, so maybe not worth it.

zcorpan: Yeah, no need to check it multiple times.

cssom-view-1/Overview.bs
```diff
@@ -783,22 +783,24 @@ the associated <a>media</a>.
...
-    whose <b>type</b> is <code>change</code>,
-    <b>callback</b> is <var>listener</var>,
-    and <b>capture</b> is false.
+1. If the <a spec=dom>context object</a>’s <a spec=dom for=EventTarget>event listener list</a>
+    <a for=list>contains</a> an <a>event listener</a> whose
+    <a spec=dom for="event listener">type</a> is <code>change</code>,
+    <a spec=dom for="event listener">callback</a> is |callback|,
+    and <a spec=dom for="event listener">capture</a> is false,
+    then <a spec=dom>remove an event listener</a> with
+    the <a spec=dom>context object</a> and that <a spec=dom>event listener</a>.
```
annevk: I wonder if we should fold this conditional into "remove an event listener".

zcorpan: That would make sense to me.

annevk: I think currently it does not because HTML doesn't need it, but it would effectively be a no-op for HTML so probably okay. Alternative we have this and an inner variant that HTML invokes directly.

zcorpan: No strong opinion. But I don't think we need to block this PR on this tweak. We can always change again in a followup.

annevk: I filed https://github.com/whatwg/dom/issues/788.

Issue #3879, opened 2019-04-27
[css-text] Hanging/collapsing of fixed-width spaces not specified


briansmith:
https://www.w3.org/TR/css-text-3/#white-space-rules says "White space processing in CSS affects only the document white space characters: spaces (U+0020), tabs (U+0009), and segment breaks."

However, it is often/usually desirable for the collapsing/hanging behavior to apply to fixed-width spaces when the fixed-width space would cause a line break. Consider:

```html
<p>1&#x205f;+&#x205f;1&#x2005;=&#x2005;2</p>
```
Which renders as:
> <p>1&#x205f;+&#x205f;1&#x2005;=&#x2005;2</p>

The spaces on either side of the `+` are U+205F Medium Mathematical Space and the spaces on either side of the `=` are U+2005. (This is the spacing recommended in http://www.unicode.org/notes/tn28 for these operators.)

Lines wrap at fixed-width spaces already (at least in the browsers I tested). However, in the browsers I tested, these fixed-width spaces do not currently collapse or hang at the end of a line like U+0020. Either collapsing/hanging of these spaces like U+0020 should be the default behavior, or there should be some way to opt into them being treated like U+0020 for the purposes of collapsing/hanging.

frivoal:
Doing something like this could mean changing either or both of these sentences in [4.1.2](https://drafts.csswg.org/css-text-3/#white-space-phase-2) to cover more types of spaces, (maybe U+2000 to U+200A and U+205F)
* > A sequence at the end of a line (ignoring any intervening inline box boundaries) of collapsible spaces (U+0020) and/or ideographic spaces (U+3000) whose white-space value collapses spaces is removed. 
* > If there remains any sequence of white space and/or ideographic spaces (U+3000) at the end of a line: 
  > [... rules about pre-wrap and break-spaces..]

Conceptually, that makes sense to me.

litherum:
There's like a million billion different flavors of spaces in Unicode. What was the reasoning for not including them since their inception?

fantasai:
@litherum Legacy, I imagine. We can gather them up and call them by category Zs, fwiw. http://www.fileformat.info/info/unicode/category/Zs/list.htm Hanging Ogham Space Mark is a bit odd, though, since it's usually visible.

briansmith:
> <fantasai> fantasai: I think we knew that nbsp couldn't hang

I'm not disagreeing, but I don't understand that is special about nbsp that makes it necessary for it to be excepted.

[...]

Issue #2500, opened 2018-04-04
[css-text-3] line breaks and ideographic space


frivoal:
(reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1450228)

Browsers differ in how they handle line breaks for  IDEOGRAPHIC SPACE (U+3000):
* Firefox allows break before and after
* Chrome and Safari forbid a break before, and allow a break after
* Edge forbids a break before and allow a break after, but also allows hanging overflow

[try it here](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cstyle%3E%0Ap%20%7B%0A%20%20%20%20width%3A%2010em%3B%0A%7D%0A%3C%2Fstyle%3E%0A%3Cp%3E%E3%81%82%E3%81%84%E3%81%86%E3%81%88%E3%81%8A%E3%81%A0%E3%82%8D%E3%81%86%E3%81%8B%EF%BC%9F%E3%80%80%E3%81%9D%E3%82%8C%E3%81%AF%E4%BD%95%E3%81%AE%E3%81%9F%E3%82%81%E3%81%AB%EF%BC%81%E3%80%80%E7%A7%81%E3%81%AF%E3%80%81%E3%81%8B%E3%81%8D%E3%81%8F%E3%81%91%E3%81%93%E3%81%A0%E3%80%82%3C%2Fp%3E)

Quoting @MurakamiShinyu

> This is important because IDEOGRAPHIC SPACE is normally used after "！" or "？" in the middle of a paragraph to keep 1em space after such punctuation marks as explained in JLREQ [3.1.6 Positioning of Dividing Punctuation Marks (Question Mark and Exclamation Mark) and Hyphens](https://www.w3.org/TR/jlreq/#positioning_of_dividing_punctuation_marks)

That logic would favor Edge's behavior.

UAX-14 says that IDEOGRAPHIC SPACE (U+3000) has class BA (for break after), which would support the behavior of Chrome/Safari/Edge over Firefox's, but:
* I cannot I cannot find any reference in css-text-3 to UAX-14's BA class
* The line-break property does not mention IDEOGRAPHIC SPACE (U+3000)

Should we reference the BA class somehow?
Should we directly list DEOGRAPHIC SPACE (U+3000) in some level of line-break?
Should we say something about allowing/requiring it to hang?

fantasai:
[...]

PR #4420, opened 2019-10-15
[css-text] Clarify handling of other space separators in white-space dfn

frivoal:
The definition of the various values of the white-space property in
section 3 does not go into much detail about exactly how they work,
and just provides a high level intro to what they do. The details are
provided in section 4, and in particular 4.1 and 4.3 (so called
“phase 1” and “phase 2” of white space processing).

While this phase 2 has been updated (see https://github.com/w3c/csswg-drafts/issues/2500, https://github.com/w3c/csswg-drafts/issues/3879, or https://github.com/w3c/csswg-drafts/issues/4180)
to define that not only space characters, but also other space
separators hang / don't hang / wrap / don't wrap, etc, based on the
value of the white-space property, the high level definition of these
values was not updated to reflect that.

While this is not necessarily a bug, as it is already called out that
details are in section 4, apparent contradictions or omissions can be
confusing.

Therefore, in the part of these definitions that explicitly talked
about white space wrapping/hanging (or not), include other space
separators as well, as defined in 4.3.


frivoal:
ping @fantasai for review

PR #4426, opened 2019-10-16
[cssom] Fix a typo in #serialize-a-local

joshgelbard:
> To serialize a URL means to create a string represented by "url(", followed by the serialization of the URL as a string, followed by ")".
> 
> To serialize a LOCAL means to create a string represented by "local(", followed by the serialization of the URL as a string, followed by ")".

`s/URL/LOCAL/` in that last line. Looks like it was copied from the rule above and incompletely replaced.


emilio:

PR #4435, opened 2019-10-18
[css-size-adjust] Specify text-size-adjust more fully

litherum:
As per the discussion at https://lists.w3.org/Archives/Public/www-style/2019Oct/0007.html.

### DIFF
--- a/css-size-adjust-1/Overview.bs
+++ b/css-size-adjust-1/Overview.bs
@@ -9,6 +9,7 @@ TR: https://www.w3.org/TR/css-size-adjust/
 ED: https://drafts.csswg.org/css-size-adjust/
 Editor: L. David Baron, Mozilla https://www.mozilla.org/, https://dbaron.org/, w3cid 15393
 Editor: Tantek &Ccedil;elik, Mozilla https://www.mozilla.org/, http://tantek.com/, w3cid 1464
+Editor: Myles C. Maxfield, Apple Inc., mmaxfield@apple.com, w3cid 77180
 Abstract: This module contains features of CSS relating to one possible mechanism for adapting pages designed for desktop computer displays for display on smaller screens such as those of mobile phones.  This mechanism involves displaying a scaled down display of the Web page and allowing the user to pan and zoom within that display, but within that scaled down display making certain text and similar elements larger than specified by the page author in order to ensure that when a block of wrapped text is zoomed to the width of the device (so it can be read without side-to-side scrolling for each line), the text is large enough to be readable.
 Status Text: <p>The following features are at risk:</p> <ul> <li>the reference to [[!CSS-TEXT-4]]&apos;s 'text-wrap' property</li> </ul>
 </pre>
@@ -88,7 +89,7 @@ This section defines the default size adjustment rules.  These rules
 are then referenced by the definition of the 'text-size-adjust' property
 in the following section.
 
-The result of the adjustment must be a computed value of the 'text-size'
+The result of the adjustment must be a computed value of the 'font-size'
 property and/or the 'line-height' property.
 
 Note: Therefore, this adjustment is polyfill-able.
--- a/css-size-adjust-1/Overview.bs
+++ b/css-size-adjust-1/Overview.bs
@@ -1,5 +1,5 @@
 <pre class="metadata">
-Title: CSS Mobile Text Size Adjustment Module Level 1
+Title: CSS Text Size Adjustment Module Level 1
 Status: ED
 Work Status: Revising
 Shortname: css-size-adjust
@@ -44,6 +44,12 @@ Introduction {#intro}
   time, this needs to be done with minimal disruption to the overall
   design of the page.
 
+  Even on a display which is large enough to prevent horizontal scrolling,
+  when a user does not zoom into the page, text size adjustment may still
+  be desirable. For example, if the viewport rules [[!CSS-DEVICE-ADAPT]]
+  describe a default zoom, text may need to be adjusted in order to be
+  readable at that default zoom level.
+
   While implementations of CSS are not required to use this
   technique, this module describes how implementations of CSS that do
   use this technique must do so.
@@ -75,20 +81,24 @@ Values {#values}
   also accept the <a>CSS-wide keywords</a> keywords as their property value.
   For readability they have not been repeated explicitly.
 
-Default size adjustment {#default-adjustment}
-=============================================
+Size adjustment {#size-adjustment}
+==================================
 
 This section defines the default size adjustment rules.  These rules
 are then referenced by the definition of the 'text-size-adjust' property
 in the following section.
 
-Issue: All of the subsections of this section need significant
-refinement:  additional detail, verification that the detail already
-present is correct, etc.
+The result of the adjustment must be a computed value of the 'text-size'
+property and/or the 'line-height' property.
 
-Issue: It's not clear how much this section should define
-precise behavior versus how much it should allow future room for
-innovation and improvement.
+Note: Therefore, this adjustment is polyfill-able.
+
+Different User Agents are quite free to make different decisions about viewing conditions,
+which therefore means that the same page may be adjusted differently in different User Agents.
+In general, the User Agent is better able to make these decisions than the page author.
+Page authors are recommended to test a variety of User Agents, devices, and viewing conditions
+when authoring their pages. Text size adjustment can be observed, controlled, and disabled by
+page authors.
 
 Types of boxes adjusted {#default-adjustment-types}
 ---------------------------------------------------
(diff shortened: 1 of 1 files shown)


css-size-adjust-1/Overview.bs
```diff
@@ -75,20 +82,24 @@ Values {#values}
   also accept the <a>CSS-wide keywords</a> keywords as their property value.
   For readability they have not been repeated explicitly.
 
-Default size adjustment {#default-adjustment}
-=============================================
+Size adjustment {#size-adjustment}
```
hober: I assume you shouldn't change the ID here, so old fragment links continue to work.

css-size-adjust-1/Overview.bs
```diff
@@ -97,54 +108,90 @@ The default size adjustment affects text and form controls, whether
 those form controls contain text (e.g., text inputs, selects) or do not
 (e.g., radio buttons, checkboxes).
 
-Conditions that suppress adjustment {#default-adjustment-conditions}
---------------------------------------------------------------------
+Adjustment Conditions {#adjustment-conditions}
```
hober: Same ID stability concern as before.

css-size-adjust-1/Overview.bs
```diff
@@ -97,54 +108,90 @@ The default size adjustment affects text and form controls, whether
...
+  even on a desktop machine</li>
+
+  <li>Elements which are extremely narrow. These are usually structural</li>
+
+  <li>Elements which only contain whitespace</li>
 </ul>
 
-Calculation of default adjustment {#default-adjustment-calculation}
--------------------------------------------------------------------
+Calculation of adjustment {#adjustment-calculation}
```
hober: ID stability.

css-size-adjust-1/Overview.bs
```diff
@@ -163,21 +210,52 @@ Size adjustment control: the 'text-size-adjust' property {#adjustment-control}
...
-      <dd>When displaying on a small device, renderers must not do size adjustment but instead the computed value of 'font-size' must be multiplied by this percentage.
-      <p class="note">Note: This means that 'text-size-adjust: 100%;' is equivalent to 'text-size-adjust: none;'.</p></dd>
+      <dd>Provided as a hint to the User Agent about a suggested adjustment amount. User agents may consider this percentage value when computing an adjusted text size.</dd>
+      <p class="note">Note: Therefore, if a User Agent naively applies the percentage to the specified text size, 'text-size-adjust: 100%;' is equivalent to 'text-size-adjust: none;'.</p></dd>
   </dl>
 
-  Issue: "Small device" is not well defined. We should clearly specify
-  when text size adjustment should apply. For example, in practice a
-  large tablet is considered by vendors to be a small device.
+Determining and Reacting to Adjustment Effects
```
hober: Missing ID.

css-size-adjust-1/Overview.bs
```diff
@@ -44,6 +45,12 @@ Introduction {#intro}
   time, this needs to be done with minimal disruption to the overall
   design of the page.
 
+  Even on a display which is large enough to prevent horizontal scrolling,
+  when a user does not zoom into the page, text size adjustment may still
+  be desirable. For example, if the viewport rules [[!CSS-DEVICE-ADAPT]]
+  describe a default zoom, text may need to be adjusted in order to be
+  readable at that default zoom level.
```
dbaron: Is this specific to zoomed-out effects (zoom < `1.0`), or is it more general than that?  (I'm not sure I entirely understand the scenario this sentence is describing.)

css-size-adjust-1/Overview.bs
```diff
@@ -44,6 +45,12 @@ Introduction {#intro}
   time, this needs to be done with minimal disruption to the overall
   design of the page.
 
+  Even on a display which is large enough to prevent horizontal scrolling,
```
dbaron: I think you're better off without this comma, i.e., "a display which is large enough to prevent horizontal scrolling when the user does not zoom into the page".

I also don't think "prevent" is quite the right word here.  Maybe "to prevent" -> "not to require", if that's what's intended?  If it's not what's intended, then maybe something else?

css-size-adjust-1/Overview.bs
```diff
@@ -75,20 +82,24 @@ Values {#values}
...
-refinement:  additional detail, verification that the detail already
-present is correct, etc.
+The result of the adjustment must be a computed value of the 'font-size'
+property and/or the 'line-height' property.
+
+Note: Therefore, this adjustment is polyfill-able.
 
-Issue: It's not clear how much this section should define
-precise behavior versus how much it should allow future room for
-innovation and improvement.
```
dbaron: I think it might still be worth leaving this issue for now.  Your edits certainly change the balance on this question, but it might still be worth thinking about further.

css-size-adjust-1/Overview.bs
```diff
@@ -97,54 +108,90 @@ The default size adjustment affects text and form controls, whether
...
-conditions are associated with layouts for which the user experience
-would be worsened by size adjustment rather than improved by it.  These
-conditions are:
+Note: In general, User Agents are trying to determine which elements are
+integral to the structure of the page, and thus shouldn't be adjusted,
+and which elements are allowed to flow freely and thus can be adjusted
+without undesirable rendering artifacts.
 
+Here are some criteria which User Agents may consider when making this
+determination:
```
dbaron: I think this list is still all worded in the form of things that would or might lead to a decision *not* to do text size adjustment.  So I think it's still worth saying that, so it's clear.  (The previous header for the list says that -- and you also said it explicitly inside the first item, but I think it's true for all the items.  And if it's not, i.e., if some of the items are the other way around, then I think it's worth restructuring them to match the rest.)

css-size-adjust-1/Overview.bs
```diff
@@ -97,54 +108,90 @@ The default size adjustment affects text and form controls, whether
...
 <ul>
-  <li>when the total amount of text in the block formatting context (see
+  <li>When the total amount of text in the block formatting context (see
```
dbaron: technically I think this should probably still be lowercase since they're not complete sentences.  (They can still end with a period, since they're *followed* by complete sentences in many cases.)

css-size-adjust-1/Overview.bs
```diff
@@ -97,54 +108,90 @@ The default size adjustment affects text and form controls, whether
...
-  <li>when the objects to be adjusted have 'white-space' of ''white-space/pre'' or
+  <li>When the objects to be adjusted have 'white-space' of ''white-space/pre'' or
   ''white-space/nowrap'' (see [[!CSS21]]) or a 'text-wrap' of ''text-wrap/nowrap'' (see
   [[!CSS-TEXT-4]]).</li>
 
+  <li>Editability of an element, either because it's a form element or because it is ''contenteditable''</li>
+
+  <li>Element contents; empty elements might not benefit from being autosized</li>
+
+  <li>Elements which have compute styles that are fairly unique among the page. These are often structural
```
dbaron: "compute" -> "computed"

css-size-adjust-1/Overview.bs
```diff
@@ -97,54 +108,90 @@ The default size adjustment affects text and form controls, whether
...
+  and might provide more broken layout than they benefit from size adjustment</li>
+
+  <li>Elements lay out to be extremely wide compared to the viewport. These would cause horizontal scrolling
```
dbaron: probably something like "lay out to be" "whose width is"

dbaron: That said, in Gecko we don't consider this a condition for disabling adjustment -- instead, we cap the width used to min(viewport inline-size, element inline-size) in [this code](https://searchfox.org/mozilla-central/rev/9b4b41b95cbcda63f565bdc24411e15248f91d83/layout/generic/nsFontInflationData.cpp#209-225).  So it's perhaps worth mentioning that option as well?

css-size-adjust-1/Overview.bs
```diff
@@ -97,54 +108,90 @@ The default size adjustment affects text and form controls, whether
...
+Note: In general, User Agents are trying to determine how big to make an
+element so that it is readable, and how small it has to be to not overflow
+or get clipped by other nearby elements.
 
-The size adjustment involves multiplication of sizes by a ratio
-determined by the minimum block text size and the computed value of
-'font-size'.  This ratio must be at least the first divided by the
-second; however, in order to maintain differentiations between font
-sizes, it should often be slightly larger.  <span class="issue">Define
-this with more detail/precision.</span>
```
dbaron: I think it's probably worth retaining this text as an example, and perhaps adding another example if you have one.  It may also be worth adding the "differentiations between font sizes" point to the list of criteria.

css-size-adjust-1/Overview.bs
```diff
@@ -163,21 +210,52 @@ Size adjustment control: the 'text-size-adjust' property {#adjustment-control}
...
-      <dd>Renderers must use the <a href="#default-adjustment">default size adjustment</a> when displaying on a small device.</dd>
+      <dd>Renderers must use the <a href="#default-adjustment">default size adjustment</a> that is best tailored to the individual user's and device's viewing conditions.</dd>
+      <p class="note">Note: On most desktop machines, the best adjustment is no adjustment.</p></dd>
     <dt><dfn>none</dfn></dt>
-      <dd>Renderers must not do size adjustment when displaying on a small device.</dd>
+      <dd>Renderers must not adjust sizes.</dd>
     <dt><dfn>&lt;percentage&gt;</dfn></dt>
-      <dd>When displaying on a small device, renderers must not do size adjustment but instead the computed value of 'font-size' must be multiplied by this percentage.
-      <p class="note">Note: This means that 'text-size-adjust: 100%;' is equivalent to 'text-size-adjust: none;'.</p></dd>
+      <dd>Provided as a hint to the User Agent about a suggested adjustment amount. User agents may consider this percentage value when computing an adjusted text size.</dd>
```
dbaron: I don't think this is a sufficient definition.  It needs to say what the percentage means (which the old text did -- although it's possible it was wrong), even if it then says that UAs don't have to honor it.  It would be better to say what the inputs into the decision of whether to honor it are.

css-size-adjust-1/Overview.bs
```diff
@@ -163,21 +210,52 @@ Size adjustment control: the 'text-size-adjust' property {#adjustment-control}
...
+      <p class="note">Note: Therefore, if a User Agent naively applies the percentage to the specified text size, 'text-size-adjust: 100%;' is equivalent to 'text-size-adjust: none;'.</p></dd>
   </dl>
 
-  Issue: "Small device" is not well defined. We should clearly specify
-  when text size adjustment should apply. For example, in practice a
-  large tablet is considered by vendors to be a small device.
+Determining and Reacting to Adjustment Effects {#adjustment-reaction}
+=====================================================================
+
+This section is non-normative
```
dbaron: Use `<em>This section is non-normative.</em>` (with `em` element and `.`).

alshamiri5 (approved):

Issue #4301, opened 2019-09-12
[web-animations-1] Dependency on CSSPseudoElement


stephenmcgruer:
If we ship getAnimations, by default we will expose animations on pseudo-elements (accessible via either `document.getAnimations()`, or `element.getAnimations({subtree:true})`). These animations may have been created by either CSS Animations or CSS Transitions (not Web Animations at time of writing, as `Element.pseudo()` has not shipped). The web developer can then call `animation.effect.target`, and get a handle to… what? [CSSPseudoElement](https://www.w3.org/TR/css-pseudo-4/#CSSPseudoElement-interface) has not shipped, so what do we return here?

This issue explores options for how to handle this, so we can actually ship getAnimations.

**Option 1 - ship CSSPseudoElement**

The simplest option - 'just' ship CSSPseudoElement. This would obviously solve the problem, but it's unclear if the browser vendors generally agree with shipping it. The exact nature of the interface might still need work (and the spec says that more methods are likely to be added to it). This is very easy to implement for node-like pseudo-elements but would be nontrivial (due to inheriting EventTarget) to implement for first-line and selection.

**Option 2 - Separate interfaces for Node-like pseudo-elements, ::selection, and ::first-line**

The term pseudo-element currently refers to a number of different behaviors. Many of them (before, after, backdrop, and arguably first-letter) behave like auto-generated elements and support full styling. As such, inheriting from EventTarget makes sense for them (along with some mutators like animation). The CSSPseudoElement interface makes sense here.

[Selection](https://www.w3.org/TR/css-pseudo-4/#highlight-pseudos) operates on an overlay system. In the spec, the highlighted areas are painted as if ::selection (which is restricted to non-layout properties) applies as a pseudo-class. [First-line](https://www.w3.org/TR/css-pseudo-4/#first-line-pseudo) is even more unusual as the pseudo-element's styling (and that of its surroundings) affect the extent of its contents, thus there is no static node structure. It does not make sense for these to have a CSSPseudoElement interface that implements EventTarget.

If we separated out psuedo-elements into different types in the IDL, we could ship CSSPseudoElement for the types that make sense, and disallow animations on those that don't.

**Option 3 - add pseudoSelector property to AnimationEffect**

This option sidesteps the problem, but at the cost of developer ergonomics. The idea would be that a pseudo-element animation would still have a `target` of the underlying Element, and then a `pseudoSelector` property that would return the appropriate selector.

[...]

PR #4437, opened 2019-10-21
[web-animations-1] Eliminate usage of CSSPseudoElement

george-steel:
Fix issue #4301

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -547,7 +547,7 @@ of the core timing model but are layered on top.
 
 Similarly, the <a href="#reaching-the-end" section>finishing behavior</a> of
 animations means that dynamic changes to the end time of
-the media [=associated effect=] of an animation may produce a
+the media ([=associated effect=]) of an animation may produce a
 different result depending on when the change occurs.
 This behavior is somewhat unfortunate but has been deemed intuitive
 and consistent with HTML.
@@ -775,7 +775,7 @@ An <dfn id="concept-animation">animation</dfn> connects a single <a>animation
 effect</a>, called its <dfn>associated effect</dfn>, to a <a>timeline</a> and
 provides playback control.
 Both of these associations are optional and configurable such that
-an <a>animation</a> may have no associated <a>associated effect</a> or
+an <a>animation</a> may have no <a>associated effect</a> or
 <a>timeline</a> at a given moment.
 
 An [=animation's=] <dfn>document for timing</dfn> is the {{Document}} with which
@@ -847,8 +847,7 @@ becomes <a lt="idle play state">idle</a>, is when the procedure to
 
 </div>
 
-<h4 id="setting-the-associated-effect">Setting the associated effect of an
-  animation</h4>
+### Setting the associated effect of an animation ### {#setting-the-target-effect}
 
 The procedure to <dfn>set the associated effect of an animation</dfn>,
 <var>animation</var>, to <var>new effect</var> which may be null, is as
@@ -1413,7 +1412,7 @@ Graphically, the effect of limiting the current time is shown below.
        alt="The effect of limiting the current time of an animation.">
   <figcaption>
     The effect of limiting the <a>current time</a> of an <a>animation</a>
-    with a start time of 1s, a associated effect of length 3s, and
+    with a start time of 1s, an [=associated effect=] of length 3s, and
     a positive [=playback rate=].
     After the <a>current time</a> of the animation reaches the end of the
     associated effect, it is capped at 3s.
@@ -3809,7 +3808,7 @@ if <em>all</em> of the following conditions are true:
     [=timeline=].
 *   The [=animation=] has an [=associated effect=].
 *   The [=animation=]'s [=associated effect=] is [=in effect=].
-*   The [=associated effect=] has an [=effect target=].
+*   The [=animation=]'s [=associated effect=] has an [=effect target=].
 
 When asked to <dfn>remove replaced animations</dfn>
 for a {{Document}}, |doc|,
@@ -4809,8 +4808,8 @@ interface KeyframeEffect : AnimationEffect {
 :   <dfn attribute for=KeyframeEffect>pseudoElement</dfn>
 ::  The [=target pseudo-selector=].
     This is `null` if the [=effect target=] is an {{Element}} or is itself `null`.
-    When not `null`, this must be a [=pseudo-element=] selector with valid 
-    [[selectors-4#pseudo-element-syntax]].
+    When not `null`, this must be a [=pseudo-element=] selector with
+    [[selectors-4#pseudo-element-syntax|valid syntax]].
     This will not accept invalid values and will always return the selector in
     its canonical form (i.e. `::before` instead of `:before`).
 
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -547,7 +547,7 @@ of the core timing model but are layered on top.
 
 Similarly, the <a href="#reaching-the-end" section>finishing behavior</a> of
 animations means that dynamic changes to the end time of
-the media (target effect) of an animation may produce a
+the media [=associated effect=] of an animation may produce a
 different result depending on when the change occurs.
 This behavior is somewhat unfortunate but has been deemed intuitive
 and consistent with HTML.
@@ -772,10 +772,10 @@ analogous to that of a DVD player and a DVD.
 </div>
 
 An <dfn id="concept-animation">animation</dfn> connects a single <a>animation
-effect</a>, called its <dfn>target effect</dfn>, to a <a>timeline</a> and
+effect</a>, called its <dfn>associated effect</dfn>, to a <a>timeline</a> and
 provides playback control.
 Both of these associations are optional and configurable such that
-an <a>animation</a> may have no associated <a>target effect</a> or
+an <a>animation</a> may have no associated <a>associated effect</a> or
 <a>timeline</a> at a given moment.
 
 An [=animation's=] <dfn>document for timing</dfn> is the {{Document}} with which
@@ -784,7 +784,7 @@ If an animation is not associated with a timeline, or its timeline is not
 associated with a document, then it has no [=document for timing=].
 
 An <a>animation</a>'s <dfn>start time</dfn> is the
-<a>time value</a> of its <a>timeline</a> when its <a>target effect</a>
+<a>time value</a> of its <a>timeline</a> when its <a>associated effect</a>
 is scheduled to begin playback.
 An animation's start time is initially <a>unresolved</a>.
 
@@ -847,14 +847,14 @@ becomes <a lt="idle play state">idle</a>, is when the procedure to
 
 </div>
 
-<h4 id="setting-the-target-effect">Setting the target effect of an
+<h4 id="setting-the-associated-effect">Setting the associated effect of an
   animation</h4>
 
-The procedure to <dfn>set the target effect of an animation</dfn>,
+The procedure to <dfn>set the associated effect of an animation</dfn>,
 <var>animation</var>, to <var>new effect</var> which may be null, is as
 follows:
 
-1.  Let <var>old effect</var> be the current <a>target effect</a> of
+1.  Let <var>old effect</var> be the current <a>associated effect</a> of
     <var>animation</var>, if any.
 1.  If <var>new effect</var> is the same object as <var>old effect</var>,
     abort this procedure.
@@ -864,11 +864,11 @@ follows:
     to run as soon as <var>animation</var> is <a>ready</a> to play <var>new
     effect</var>.
 1.  If <var>new effect</var> is not <code>null</code> and
-    if <var>new effect</var> is the <a>target effect</a> of another
+    if <var>new effect</var> is the <a>associated effect</a> of another
     <a>animation</a>, <var>previous animation</var>, run the procedure to <a>set
-    the target effect of an animation</a> (this procedure) on <var>previous
+    the associated effect of an animation</a> (this procedure) on <var>previous
     animation</var> passing null as <var>new effect</var>.
-1.  Let the <a>target effect</a> of <var>animation</var> be <var>new
+1.  Let the <a>associated effect</a> of <var>animation</var> be <var>new
     effect</var>.
 1.  Run the procedure to <a>update an animation's finished state</a> for
     <var>animation</var> with the <var>did seek</var> flag set to false, and
@@ -891,7 +891,7 @@ The procedure to <dfn>reset an animation's pending tasks</dfn> for
 
 ### The current time of an animation ### {#the-current-time-of-an-animation}
 
-<a>Animations</a> provide a <a>time value</a> to their <a>target
+<a>Animations</a> provide a <a>time value</a> to their <a>associated
 effect</a> called the animation's <dfn>current time</dfn>.
 
 The <a>current time</a> is calculated from the first
@@ -1055,7 +1055,7 @@ is as follows:
     <var>animation</var> with the <var>did seek</var> flag set to true, and
     the <var>synchronously notify</var> flag set to false.
 
-### Waiting for the target effect ### {#waiting-for-the-target-effect}
+### Waiting for the associated effect ### {#waiting-for-the-target-effect}
 
 <div class='informative-bg'>
 
@@ -1086,7 +1086,7 @@ An animation is <dfn>ready</dfn> at the first moment where <em>both</em> of the
 following conditions are true:
 
 *   the user agent has completed any setup required to begin the playback of
-    the animation's <a>target effect</a>
+    the animation's <a>associated effect</a>
     including rendering the first frame of any <a>keyframe
     effect</a>.
 
@@ -1152,7 +1152,7 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
 
         *   <a>current time</a> is <a>unresolved</a>, or
         *   <a>current time</a> &lt; zero, or
-        *   <a>current time</a> &ge; <a>target effect end</a>,
+        *   <a>current time</a> &ge; <a>associated effect end</a>,
 
     ::  Set <var>animation</var>'s <a>hold time</a> to zero.
     :   If |animation|'s [=effective playback rate=] &lt; 0,
@@ -1161,12 +1161,12 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
 
         *   <a>current time</a> is <a>unresolved</a>, or
         *   <a>current time</a> &le; zero, or
-        *   <a>current time</a> &gt; <a>target effect end</a>,
+        *   <a>current time</a> &gt; <a>associated effect end</a>,
 
-    ::  If <a>target effect end</a> is positive infinity,
+    ::  If <a>associated effect end</a> is positive infinity,
         <a>throw</a> an "{{InvalidStateError}}" {{DOMException}} and
         abort these steps.
-        Otherwise, set <var>animation</var>'s <a>hold time</a> to <a>target
+        Otherwise, set <var>animation</var>'s <a>hold time</a> to <a>associated
         effect end</a>.
     :   If |animation|'s [=effective playback rate=] = 0 and |animation|'s
         [=current time=] is [=unresolved=],
@@ -1256,7 +1256,7 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
 
         <div class="note">
             Note that the order of the above two steps is important
-            since it means that an animation with zero-length <a>target
+            since it means that an animation with zero-length <a>associated
             effect</a> will resolve its <a>current ready promise</a>
             before its <a>current finished promise</a>.
         </div>
@@ -1324,12 +1324,12 @@ follows:
     ::   Let <var>animation</var>'s <a>hold time</a> be zero.
 
     :    Otherwise,
-    ::   If <a>target effect end</a> for <var>animation</var> is positive
+    ::   If <a>associated effect end</a> for <var>animation</var> is positive
          infinity,
          <a>throw</a> an "{{InvalidStateError}}" {{DOMException}}
          and abort these steps.
          Otherwise, let <var>animation</var>'s <a>hold time</a> be
-         <a>target effect end</a>.
+         <a>associated effect end</a>.
 
     </div>
 
@@ -1343,13 +1343,13 @@ follows:
 1.  Schedule a task to be executed at the first possible moment after
     the user agent has performed any processing necessary to suspend
     the playback of
-    <var>animation</var>'s <a>target effect</a>, if any.
+    <var>animation</var>'s <a>associated effect</a>, if any.
     The task shall perform the following steps:
 
     1.  Let <var>ready time</var> be the time value of the timeline associated
         with <var>animation</var> at the moment when the user agent completed
         processing necessary to suspend playback of <var>animation</var>'s
-        <a>target effect</a>.
+        <a>associated effect</a>.
 
     1.  If <var>animation</var>'s [=start time=]
         is <a lt=unresolved>resolved</a> and its <a>hold time</a> is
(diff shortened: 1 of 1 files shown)


george-steel:
@stephenmcgruer @birtles here's my first draft. I'm still not sure what to do about [commitStyles](https://drafts.csswg.org/web-animations-1/#dom-animation-commitstyles) as the DOM has no notion of inline style for pseudo-elements.

birtles:
> I'm still not sure what to do about [commitStyles](https://drafts.csswg.org/web-animations-1/#dom-animation-commitstyles) as the DOM has no notion of inline style for pseudo-elements.

I'm not sure you need to do anything? `commitStyles` [throws](https://drafts.csswg.org/web-animations-1/#commit-computed-styles) if any of the targets is a pseudo-element.

web-animations-1/Overview.bs
```diff
@@ -3188,9 +3189,14 @@ Keyframe effects {#keyframe-effects}
...
-an element or pseudo-element such as <code>::before</code> or
-<code>::after</code> [[!SELECT]] referred to as the <dfn>target element</dfn>.
+effect</a> that use the output of the timing model to update CSS properties of
+an element known as the <dfn>target element</dfn> or one of its <a>pseudo-elements</a>
+(in which case the target element is the <a>originating element</a> of the pseudo-element).
+
+The combination of <a>target element</a> and <a>pseudo-element</a> selector
+referred to as the <dfn>effect target</dfn>. In the case that the effect target does not exist
+(for example when it targets a `::after` pseudo-element where one does not exist)
+the effect applies no styles until the target comes into existence.
```
birtles: I'd rather we don't change the meaning of "target element". This should be an API-only change, the model itself should stay the same, right?

birtles: Actually, may it's ok. I'd probably just prefer to reword this so that the definition of "effect target" comes before "target element".

So the first paragraph should be just:

> Keyframe effects are a kind of animation effect that use the output of the timing model to update CSS properties for
an element or pseudo-element such as ::before or ::after [[!SELECT]] referred to as the effect target.

web-animations-1/Overview.bs
```diff
@@ -3290,7 +3296,7 @@ To produce <a>computed keyframe offsets</a>, we define a procedure to
 
 <a>Computed keyframes</a> are produced using the following procedure.
 Note that this procedure is only performed on a <a>keyframe effect</a> having
-a <a>target element</a> for which computed property values can be calculated.
+a existing <a>effect target</a> for which computed property values can be calculated.
```
birtles: I think if you're going to introduce "existing" here, I think you need to link to a definition of what it means for a target to exist.

web-animations-1/Overview.bs
```diff
@@ -3236,7 +3242,7 @@ Before calculating the <a>effect value</a> of a <a>keyframe effect</a>,
 the property values specified on its <a>keyframes</a> are resolved to
 <a>computed values</a>, and the offset to use for any keyframes with a null
 <a>keyframe offset</a> is computed. The result of resolving these values is
-a set of <dfn>computed keyframes</dfn>.
+a set of <dfn>computed keyframes</dfn> for a particular moment in time.
```
birtles: Not sure this addition is needed?

web-animations-1/Overview.bs
```diff
@@ -3357,7 +3363,7 @@ given <var>iteration progress</var>, <var ignore=''>current iteration</var> and
 1.  If <a>animation type</a> of the <var>target property</var> is
     <a>not animatable</a> abort this procedure
     since the effect cannot be applied.
-1.  Define the <dfn>neutral value for composition</dfn> as a value
+1.  Define the <dfn>identity value for composition</dfn> as a value
```
birtles: Is this change related?

web-animations-1/Overview.bs
```diff
@@ -3301,24 +3307,24 @@ a <a>target element</a> for which computed property values can be calculated.
...
         For logical properties [[!CSS-LOGICAL-1]], add the [=equivalent physical
         properties=] [[!CSS-WRITING-MODES-4]] based on the computed value of
-        'writing-mode' and/or 'direction' for <a>target element</a>.
+        'writing-mode' and/or 'direction' for the <a>effect target</a>.
 
-        For example, if <var>keyframe</var> has a value of &lsquo;12pt&rsquo;
+        For example, if <var>keyframe</var> has a value of ‘12pt’
         for the 'border-width' property, the user agent may produce a computed
-        value of &lsquo;16px&rsquo; for each of the longhand properties:
+        value of ‘16px’ for each of the longhand properties:
```
birtles: Should these changes be a separate patch?

web-animations-1/Overview.bs
```diff
@@ -4788,11 +4796,17 @@ interface KeyframeEffect : AnimationEffect {
...
+::  The <a>target element</a> being animated by this object
+    (either the <a>effect target</a> if it is an element or its
+    <a>originating element</a> if it is a pseudo-element).
+    This may be `null` for animations that do not target
     a specific element such as an animation that produces a sound
     using an audio API.
 
+:   <dfn attribute for=KeyframeEffect>pseudoElement</dfn>
+::  The <a>effect target</a>'s <a>pseudo-element</a> selector if the effect target is a pseudo-element.
+    This is `null` if the <a>effect target</a> is an element or has no specific target.
```
birtles: Does this need to define that this is `::before` or `::after` or `::marker`? i.e. both the specific syntax used (not the old-style `:before` etc.) and the set of pseudo elements supported?

web-animations-1/Overview.bs
```diff
@@ -5743,9 +5762,9 @@ animation.play();</pre>
         <em>and</em>
     * Its [=replace state=] is <em>not</em> [=removed replace state|removed=].
 
-    For the purposes of this method
-    a [=pseudo-element=] is considered to be
-    a [=child=] of its [=originating element=].
+    Note: Since an {{Animation}} targeting a <a>pseudo-element</a> has the
+    <a>originating element</a> as its <a>target element</a>, calling {{Animatable/getAnimations}}
+    on an element will return <a>animations</a> targeting both the element itself and its pseudo-elements.
```
birtles: It should only return them if `subtree: true` is set. I think the want to keep roughly the original spec text here.

birtles:
I've done a first pass review for now. It's looking pretty good but I'd like to have another look after those changes have been made. It would also be easier to review if we can move the cosmetic and unrelated changes to a separate PR / patch.

web-animations-1/Overview.bs
george-steel: Perhaps you're right and this and the previous change should go in a separate cleanup patch.

web-animations-1/Overview.bs
george-steel: I was trying to clarify something that got misread and lead to bugs in our current implementation.

web-animations-1/Overview.bs
george-steel: Existing means present in the layout tree, such as a `::before` with a content property, as opposed to one without content that doesn't appear. The idea is to hang Animations off of their target Elements which have a well-defined lifecycle and make Animation lifecycle independent of the ill-defined pseudo-element lifecycle. The Animation always runs and has a visible effect iff the pseudo-element does.

web-animations-1/Overview.bs
george-steel: My idea was that this since there's now the notion of a target Element to hang the Animation off, we could apply animations to all valid pseudo-elements: both the tree-abiding ones, ones that act as alternate style slots (such as ::first-line, ::selection, only supporting the properties those selectors support), and shadow part styles. I see the restriction to before after and marker as an accidental effect of needing a layout object as the effect target.

web-animations-1/Overview.bs
george-steel: See my above comments about hanging Animations off of elements.

george-steel: One alternative proposed by @stephenmcgruer if to add a `pseudos` option to select whether just to get Animations targeting the element's base style or whether to get Animations targeting target pseudo-elements as well (i.e. all Animations hanging off the element).

web-animations-1/Overview.bs
george-steel: Good idea to reorder.

web-animations-1/Overview.bs
birtles: That sounds like a separate issue to this PR which requires further discussion. (I know Gecko has no intention of implementing animations on a wider range of pseudo-elements at this stage.)

For now we still need to define / link to the appropriate syntax here.

web-animations-1/Overview.bs
birtles: Let's keep the existing behavior for now. If you'd like to change it, let's use a separate issue / PR for that.

web-animations-1/Overview.bs
birtles: Can you elaborate on that? I suspect there's probably a better way to clarify in that case.

web-animations-1/Overview.bs
birtles: That still sounds like existing wants some sort of definition.

web-animations-1/Overview.bs
george-steel: Are you saying we should add a restriction of which pseudo-elements to support? Currently there is no such restriction in the spec other than the fact that CSSPseudoElement:EventTarget is too much of a mess to even come close to implementing in all but those cases.

Specifying that is returns pseudo-elements in the modern syntax is a good idea.

web-animations-1/Overview.bs
george-steel: I think there is, I plan on proposing an (isomorphic) refactor to this section to clarify it as a separate patch. Perhaps this should wait until then. The idea was that this depends on what the inherited styles are at when they're calculated, which must be the same time they're used (i.e. a keyframes do not capture parent styles at the moment of their creation).

web-animations-1/Overview.bs
george-steel: Ok, it's a bit odd, but having a separate issue for that makes sense. I was trying to keep the invariant of it returns everything hanging off that element instead of the invariant of iit only returning the same set of animations as before.

web-animations-1/Overview.bs
george-steel: Maybe just remove "existing" since "for which computed property values can be calculated" does the job. As this procedure is affine, it can always be skipped when there is nothing to style.

web-animations-1/Overview.bs
birtles: No, no need to an explicit restriction.

web-animations-1/Overview.bs
birtles: Ok that makes sense. And it probably would be best as a separate refactor.

web-animations-1/Overview.bs
birtles: Ok, makes sense.

george-steel:
@birtles I made the changes and split out the cosmetic stuff (it's no longer in this PR).

web-animations-1/Overview.bs
```diff
@@ -3799,14 +3805,14 @@ if <em>all</em> of the following conditions are true:
     [=timeline=].
 *   The [=animation=] has an associated [=target effect=].
 *   The [=target effect=] associated with the [=animation=] is [=in effect=].
-*   The [=target effect=] has an associated [=target element=].
+*   The [=target effect=] has an associated [=effect target=].
```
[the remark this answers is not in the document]
stephenmcgruer: @birtles for your thoughts here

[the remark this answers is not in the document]
george-steel: I agree this is akward, your solution looks good to me.

george-steel:
Sorry about the force push, A bunch of brian's changes required splitting commits between this PR and a future one.

web-animations-1/Overview.bs
```diff
@@ -3187,10 +3188,15 @@ the <a>animation type</a> is <a>discrete</a>.
...
-<code>::after</code> [[!SELECT]] referred to as the <dfn>target element</dfn>.
+<dfn lt="keyframe effect">Keyframe effects</dfn> are a kind of [=animation
+effect=] that use the output of the timing model to update CSS properties of
+an element or [=pseudo-element=] (such as `::before` or `::after` [[!select]])
+referred to as the <dfn>effect target</dfn>.
+
+In order for the [=effect target=] to have a well-defined lifecycle,
+it is reperesented as an {{Element}} known as the <dfn>target element</dfn>
+(if the target is a [=pseudo-element=] this is the target's [=originating element=])
+and a [=pseudo-element=] selector (which is `null` if the [=effect target=] is an element).
```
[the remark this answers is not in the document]
george-steel: done

web-animations-1/Overview.bs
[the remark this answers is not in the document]
birtles: I think `effect` is too vague a term for the model (it's fine for the API though). Any old key frame effect is an effect so this can easily get confusing. We mean to say the particular effect that the animation is playing.

I agree we should keep the term `effect target` but rename `target effect` to something else.

At one point I think we called the `target effect` the `source effect` but that's slightly inaccurate. An animation _drives_ an effect, but is quite happy to keep ticking along without one. So it's not really the "source" for anything.

`client effect`? `applied effect`? `associated effect`?

I think I like "associated effect" best?

web-animations-1/Overview.bs
```diff
@@ -4788,11 +4795,21 @@ interface KeyframeEffect : AnimationEffect {
...
     a specific element such as an animation that produces a sound
     using an audio API.
 
+:   <dfn attribute for=KeyframeEffect>pseudoElement</dfn>
+::  The [=effect target=]'s [=pseudo-element=] selector.
+    This is `null` if the [=effect target=] is an element the effect
+    has or has no specific target.
+    When not `null`, this must be a valid [=pseudo-element=] selector
+    and will always return the modern syntax
+    (i.e. it will always reutrn `::before` and never `:before`).
```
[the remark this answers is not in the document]
birtles: Yes, we need something better than "modern syntax". Ideally a link to wherever the syntax is defined and a example in braces (e.g. `::before`) so implementers and authors alike know what to expect. I seem to remember dbaron encouraging me to spec that at some point which might be why CSSPseudoElement has the syntax explicitly listed:

  https://drafts.csswg.org/css-pseudo-4/#CSSPseudoElement-interface

birtles:
> We seem to have lost Brian's earlier comments (possibly due to a force-push?), so please make sure to check the history and see that we don't lose track of anything that wasn't yet resolved.

Sorry, I should have done a proper review. I kept thinking, "just one more comment!"

web-animations-1/Overview.bs
```diff
@@ -3187,10 +3188,19 @@ the <a>animation type</a> is <a>discrete</a>.
...
-effect</a> that use the output of the timing model to update CSS properties for
-an element or pseudo-element such as <code>::before</code> or
-<code>::after</code> [[!SELECT]] referred to as the <dfn>target element</dfn>.
+<dfn lt="keyframe effect">Keyframe effects</dfn> are a kind of [=animation
+effect=] that use the output of the timing model to update CSS properties of
+an element or [=pseudo-element=] (such as `::before` or `::after` [[!select]])
+referred to as the <dfn>effect target</dfn>.
+
+In order for the [=effect target=] to have a well-defined lifecycle,
+it is reperesented as an {{Element}} known as the <dfn>target element</dfn>
```
stephenmcgruer: s/reperesented/represented

web-animations-1/Overview.bs
```diff
@@ -3187,10 +3188,19 @@ the <a>animation type</a> is <a>discrete</a>.
...
+and a [=pseudo-element=] selector knows as the <dfn>target pseudo-selector</dfn>.
+If the [=effect target=] is an {{Element}}, the [=target element=] is that
+element and the [=target pseudo-selector=] is `null`.
+If the [=effect target=] is a [=pseudo-element=], the [=target element=] is
+its [=originating element=], and the [=target pseudo-selector=] is as required
+to specify thatparticular [=pseudo-element=].
```
stephenmcgruer: s/thatparticular/that particular

web-animations-1/Overview.bs
```diff
@@ -3187,10 +3188,19 @@ the <a>animation type</a> is <a>discrete</a>.
...
+an element or [=pseudo-element=] (such as `::before` or `::after` [[!select]])
+referred to as the <dfn>effect target</dfn>.
+
+In order for the [=effect target=] to have a well-defined lifecycle,
+it is reperesented as an {{Element}} known as the <dfn>target element</dfn>
+and a [=pseudo-element=] selector knows as the <dfn>target pseudo-selector</dfn>.
+If the [=effect target=] is an {{Element}}, the [=target element=] is that
+element and the [=target pseudo-selector=] is `null`.
+If the [=effect target=] is a [=pseudo-element=], the [=target element=] is
+its [=originating element=], and the [=target pseudo-selector=] is as required
```
stephenmcgruer: Super nit; the comma on this line is not needed.

web-animations-1/Overview.bs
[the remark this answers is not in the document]
stephenmcgruer: `associated effect` SGTM. It's somewhat vague, but good enough I think.

web-animations-1/Overview.bs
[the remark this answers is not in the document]
stephenmcgruer: https://www.w3.org/TR/selectors-4/#pseudo-element-syntax is the syntax, right?

web-animations-1/Overview.bs
```diff
@@ -3807,9 +3807,9 @@ if <em>all</em> of the following conditions are true:
     [=removed replace state|removed=].
 *   The [=animation=] is associated with a [=monotonically increasing=]
     [=timeline=].
-*   The [=animation=] has an associated [=target effect=].
-*   The [=target effect=] associated with the [=animation=] is [=in effect=].
-*   The [=target effect=] has an associated [=effect target=].
+*   The [=animation=] has an [=associated effect=].
+*   The [=animation=]'s [=associated effect=] is [=in effect=].
+*   The [=associated effect=] has an [=effect target=].
```
birtles: (Nit: I guess for consistency we should make the last bullet point also say `The [=animation=]'s [=associated effect=]...`.)

web-animations-1/Overview.bs
```diff
@@ -4809,10 +4809,10 @@ interface KeyframeEffect : AnimationEffect {
 :   <dfn attribute for=KeyframeEffect>pseudoElement</dfn>
 ::  The [=target pseudo-selector=].
     This is `null` if the [=effect target=] is an {{Element}} or is itself `null`.
-    When not `null`, this must be a valid [=pseudo-element=] selector,
-    attempting to set it to an invalid value will have no effect.
-    and will always return using the the double-colon syntax
-    (i.e. always `::before` and never `:before`).
+    When not `null`, this must be a [=pseudo-element=] selector with valid 
+    [[selectors-4#pseudo-element-syntax]].
```
birtles: Just curious, how does this actually render? Does it say, "...a pseudo element selector with valid syntax"?

web-animations-1/Overview.bs
```diff
@@ -4809,10 +4809,10 @@ interface KeyframeEffect : AnimationEffect {
...
+    This will not accept invalid values and will always return the selector in
+    its canonical form (i.e. `::before` instead of `:before`).
```
birtles: I think this isn't strict enough. The linked-to syntax section still allows the single colon format whereas I think we want to explicitly reject that for this API since we don't want to propagate that form any further.

web-animations-1/Overview.bs
```diff
@@ -772,10 +772,10 @@ analogous to that of a DVD player and a DVD.
 </div>
 
 An <dfn id="concept-animation">animation</dfn> connects a single <a>animation
-effect</a>, called its <dfn>target effect</dfn>, to a <a>timeline</a> and
+effect</a>, called its <dfn>associated effect</dfn>, to a <a>timeline</a> and
 provides playback control.
 Both of these associations are optional and configurable such that
-an <a>animation</a> may have no associated <a>target effect</a> or
+an <a>animation</a> may have no associated <a>associated effect</a> or
```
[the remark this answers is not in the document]
george-steel: I thought I got all of those. Must have missed that one.

web-animations-1/Overview.bs
george-steel: I was aiming for consistency with other specs which accept pseudo-element selectors. Most I've encountered so far will accept but never return the legacy syntax. Should the question of whether to change that convention be handled here or should that be a separate issue/PR?

web-animations-1/Overview.bs
george-steel: oops, I accidentally had it include the section number, fixing that now.

web-animations-1/Overview.bs
```diff
@@ -847,14 +847,14 @@ becomes <a lt="idle play state">idle</a>, is when the procedure to
 
 </div>
 
-<h4 id="setting-the-target-effect">Setting the target effect of an
+<h4 id="setting-the-associated-effect">Setting the associated effect of an
```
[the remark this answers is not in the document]
george-steel: Unfortunately there isn't, I'm changing the id back for now.

web-animations-1/Overview.bs
```diff
@@ -1413,26 +1413,26 @@ Graphically, the effect of limiting the current time is shown below.
        alt="The effect of limiting the current time of an animation.">
   <figcaption>
     The effect of limiting the <a>current time</a> of an <a>animation</a>
-    with a start time of 1s, a target effect of length 3s, and
+    with a start time of 1s, a associated effect of length 3s, and
```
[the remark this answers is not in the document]
george-steel: The hazards of running a regex. Thanks for catching this.

web-animations-1/Overview.bs
george-steel: Now it renders like `a pseudo element selector with [=valid syntax=]`

web-animations-1/Overview.bs
george-steel: @stephenmcgruer

web-animations-1/Overview.bs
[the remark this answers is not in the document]
george-steel: @stephenmcgruer What do you think?

web-animations-1/Overview.bs
birtles: Yes, please file a separate issue for this. I seem to remember getting advice at one point that we shouldn't accept the old syntax either.

web-animations-1/Overview.bs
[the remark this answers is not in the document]
birtles: The [bikeshed docs](https://tabatkins.github.io/bikeshed/#id-gen) say:
> If a heading changed significantly, so that you want to change the ID, but you want links to the old heading ID to still work, put the old ID in an oldids="" attribute on the heading element. If there are multiple, comma-separate them.

george-steel:
fixed the oldids

stephenmcgruer:
Awesome! Can you file the issue for disallowing the old syntax @george-steel , if you haven't already?

Looking forward to the WPT test changes to accompany this too :)

george-steel:
@birtles I'm not quite sure about your reasoning for disallowing the old syntax in this one specific case. Since it was your idea, would you mind filing the issue?

birtles:
> @birtles I'm not quite sure about your reasoning for disallowing the old syntax in this one specific case. Since it was your idea, would you mind filing the issue?

Filed #4472.

web-animations-1/Overview.bs
```diff
@@ -3187,10 +3187,19 @@ the <a>animation type</a> is <a>discrete</a>.
...
-an element or pseudo-element such as <code>::before</code> or
-<code>::after</code> [[!SELECT]] referred to as the <dfn>target element</dfn>.
+<dfn lt="keyframe effect">Keyframe effects</dfn> are a kind of [=animation
+effect=] that use the output of the timing model to update CSS properties of
+an element or [=pseudo-element=] (such as `::before` or `::after` [[!select]])
+referred to as the <dfn>effect target</dfn>.
+
+In order for the [=effect target=] to have a well-defined lifecycle,
+it is represented as an {{Element}} known as the <dfn>target element</dfn>
+and a [=pseudo-element=] selector knows as the <dfn>target pseudo-selector</dfn>.
```
Loirooriol: Typo: knows -> known

PR #4446, opened 2019-10-24
Correct translation calculation in recomposition of decomposed transform

jlfwong:
As far as I can tell, this calculation for recomposing the transform is incorrect.

If you apply the given pseudocode description of decomposing and recomposing to the following matrix:

```
[ 1 1 10
  0 1 20 ]
```

(where `(t_x, t_y) = (10, 20)`)

Then you'd expect to get back the same matrix (decompose & recompose should be inverse functions).

However, if you follow the specification as given, then I think you get back the following instead:

```
[ 1 1 14.14
  0 1 24.14 ]
```

### DIFF
--- a/css-transforms-1/Overview.bs
+++ b/css-transforms-1/Overview.bs
@@ -1346,8 +1346,8 @@ matrix[1][0] = m21
 matrix[1][1] = m22
 
 // Translate matrix.
-matrix[3][0] = translate[0] * m11 + translate[1] * m21
-matrix[3][1] = translate[0] * m12 + translate[1] * m22
+matrix[3][0] = translate[0]
+matrix[3][1] = translate[1]
 
 // Rotate matrix.
 angle = deg2rad(angle);


tabatkins:
@smfr, @chrishtr, @emilio, what's right here?

emilio:
I think this is right, and this is a bug in the spec.

Firefox code is [here for decomposition](https://searchfox.org/mozilla-central/rev/e8e4a8641f3e6728ea42ae30905ce2145bfd5568/servo/components/style/values/animated/transform.rs#158), and [here for recomposition](https://searchfox.org/mozilla-central/rev/e8e4a8641f3e6728ea42ae30905ce2145bfd5568/servo/components/style/values/animated/transform.rs#228).

I found some other bugs in the 3d algorithm when dealing with compat issues (https://github.com/w3c/csswg-drafts/pull/2713 for example).

But yeah, I think the spec as written doesn't preserve the matrix across decomposition and recomposition, which is clearly a bug.

cc @dirkschulze

PR #4490, opened 2019-11-06
[css-color-4] Align color example comments

DannyS712:
No change to the actual content, just align the end of the HTML comments. Currently, only the first 2 are aligned, despite having different lengths; the alternative would be to not align any of the comment ends, to be consistent with a single space before `*/`


DannyS712:
Please mark this as `non-substantive` - "if the contribution does not concern a normative part of a specification, or is editorial in nature (e.g. fixing typos or examples), the contribution can be marked as non-substantive"

PR #4523, opened 2019-11-21
[css-nav-1] Improve processing model

jihyerish:
* Modify the Processing model section with improved readability
   * Add the glossary section
* Additional: Modify the IDL format for Navigation Event


triple-underscore:
The last “Dispatches navnotarget event ...” in the spatial navigation steps, which was “fire an event named *navbeforefocus*” before this update, seems unintentional change of event type, isn't it?

jihyerish:
@triple-underscore , thank you so much!
Yes, you're right. It was my mistake, so I fixed it via https://github.com/w3c/csswg-drafts/commit/b7055b2d97560e68f8a32a7fea6a3a730a04271d.

PR #4527, opened 2019-11-22
Fixed a lot of spelling and grammar mistakes, added some formatting a…

SebastianZ:
I've corrected a bunch of spelling and grammar mistakes within the CSS UI 4 spec. and added some missing references.

Sebastian


css-ui-4/Overview.bs
```diff
@@ -331,21 +331,21 @@ regardless of color background.
...
-The 'outline' property is a shorthand property,
+The 'outline' property is a shorthand property
 and sets all three of 'outline-style',
 'outline-width',
 and 'outline-color'.
 
 Note: The outline is the same on all sides.
 In contrast to borders,
-there are no ''outline-top'' or ''outline-left'' etc. properties.
+there are no ''outline-top'', ''outline-left'' etc. properties.
```
tabatkins: Need another comma between outline-left and etc here.

css-ui-4/Overview.bs
```diff
@@ -411,10 +411,10 @@ then the outline is outset from the <a>border edge</a> by that amount.
 to shrink into the border box.
 Both the height and the width of outside of the shape
```
tabatkins:
```suggestion
Both the height and the width of the outside of the shape
```

css-ui-4/Overview.bs
```diff
@@ -529,7 +529,7 @@ The user agent must allow the user to resize the element
 with no other constraints than what is imposed by
 'min-width', 'max-width', 'min-height', and 'max-height'.
 
-Note: There may be situations where user attempts to resize an element
+Note: There may be situations, in which user attempts to resize an element
```
tabatkins: no comma

css-ui-4/Overview.bs
```diff
@@ -1740,10 +1740,9 @@ Note: The details of the aliasing mechanism is intentionally left up to the UA.
 Making ''-webkit-user-select'' a shorthand property of 'user-select'
 is known to be an effective approach,
 and new implementers should consider it.
-However UAs supporting ''-webkit-user-select'' as an alias of 'user-select' through other means exist,
-without adverse consequences to compatibility,
-so this specification allows flexibility.
-
+However, UAs supporting ''-webkit-user-select'' as an alias of 'user-select' through other means exist,
+without adverse consequences to compatibility.
+So this specification allows flexibility.
```
tabatkins: this line should remain part of the previous sentence, as originally written. (the addition of a comma after 'however' is good, tho)

css-ui-4/Overview.bs
```diff
@@ -1794,7 +1793,7 @@ so that CSS can be used to restyle them.
 	and a "none" value that suppresses "native" look.
 
 	However, there is evidence that this alone is not web compatible,
-	due in part to uses such as:
+	in part due to uses such as:
```
tabatkins: This was better as-is.

css-ui-4/Overview.bs
```diff
@@ -1859,19 +1858,19 @@ so that CSS can be used to restyle them.
...
+		for earlier non-standard versions of this property.
 		They all have the same effect as ''appearance/auto''.
 
 		<pre class=prod style="white-space: normal"><<compat-auto>> =  <dfn>searchfield</dfn> | <dfn>textarea</dfn> | <dfn>push-button</dfn> | <dfn>slider-horizontal</dfn> | <dfn>checkbox</dfn> | <dfn>radio</dfn> | <dfn>square-button</dfn> | <dfn>menulist</dfn> | <dfn>listbox</dfn> | <dfn>meter</dfn> | <dfn>progress-bar</dfn></pre>
 
 		Issue: When ''appearance/auto'' is widely supported,
 		recommend using it instead.
 
-		Issue: If any of these value is not needed for web compat,
+		Issue: If any of these values is not needed for web compat,
```
tabatkins:
```suggestion
		Issue: If any of these values are not needed for web compat,
```

css-ui-4/Overview.bs
```diff
@@ -1977,10 +1976,10 @@ Effects of 'appearance' on Semantic Aspects of Elements</h4>
 
 UAs must preserve aspects of the control
 which are necessary to operate the control with its original semantics.
-The UA may however give them a different look and feel
+The UA may however give it a different look and feel
 as long as it remains possible to operate the control.
-Aspects of a control which are merely needed to observe the state the control is in
-and are not needed to operate it
+Aspects of a control, which are merely needed to observe the state the control is in
```
tabatkins: these two lines were better as-is

css-ui-4/Overview.bs
```diff
@@ -2036,8 +2035,8 @@ that are typical of the element whose appearance it acquires.
 <h3 id=control-specific-rules>
 Control Specific Rules</h3>
 
-Issue: Maybe some or all of this section
-should be moved to the [[HTML]] spec
+Issue: Maybe some part or all of this section
```
tabatkins: this line was better as-is

css-ui-4/Overview.bs
SebastianZ: Right. Added.

css-ui-4/Overview.bs
SebastianZ: Done.

css-ui-4/Overview.bs
SebastianZ: That was new to me. Thanks!

css-ui-4/Overview.bs
SebastianZ: Ok, added the comma back.

css-ui-4/Overview.bs
SebastianZ: Hmm, that sounded strange to me, but you're the native speaker, so I trust you.

css-ui-4/Overview.bs
SebastianZ: Right. Changed it.

css-ui-4/Overview.bs
SebastianZ: Ok, changed it back.

css-ui-4/Overview.bs
SebastianZ: Removed "part" again.

css-ui-4/Overview.bs
```diff
@@ -707,7 +707,7 @@ Values have the following meanings:
 			If the values are unspecified,
 			then the intrinsic hotspot defined inside the image resource itself is used.
 			If both the values are unspecific and the referenced cursor has no defined hotspot,
-			the effect is as if a value of "0 0" were specified.
+			the effect is as if a value of "0 0" was specified.
```
Loirooriol: Not a native speaker, but this seems an unreal conditional, so "were" was correct?
https://www.grammarly.com/blog/conditional-sentences-was-instead-of-were/

ewilligers: I hadn't heard of 'unreal conditional' before, but I wouldn't read "0 0" as being implausible here.

As a native speaker, I would always use 'was' here due to the non-plural "a value".

The following might also have worked: "the effect is as if the values `0 0` were specified."

SebastianZ: I am also not a native speaker, though I looked up the grammar about unreal conditionals before I created the PR and came to the same conclusion as Eric.

The page you referred to clearly states "When you’re talking about a possibility that did happen or might be true, use “was” and “were” as you normally would."

So, according to that, "was" is correct.

Sebastian

Loirooriol: Well, we are under the assumption that "both the values are unspecific", so it's impossible for "0 0" to be specified. Then it seems an unreal conditional to me, but it's not like this matters much.

SebastianZ:
Thank you for merging these changes, @frivoal! I totally missed that.

Sebastian

Issue #2370, opened 2018-02-28
[css-mediaqueries] Proposal: "prefers-reduced-data"


Malvoz:
Would CSSWG consider this proposal of <s>detecting</s> exposing if the [`Save-Data`](http://httpwg.org/http-extensions/client-hints.html#save-data) request header from the [Client Hint](https://github.com/httpwg/http-extensions#client-hints) HTTP extension is on/off to allow for usage with CSS?

### The `Save-Data` HTTP request header:
> ...a signal indicating explicit user opt-in into a reduced data usage mode on the client, and when communicated to origins allows them to deliver alternate content honoring such preference - e.g. smaller image and video resources, alternate markup...<br>
_[- http://httpwg.org/http-extensions/client-hints.html#save-data](http://httpwg.org/http-extensions/client-hints.html#save-data)_

### The 'prefers-reduced-data' feature:
<table>
  <tr>
    <th>Name:</th>
    <th>For:</th>
     <th>Value:</th>
<th>Type:</th>
  </tr>
  <tr>
    <td><code>prefers-reduced-data</code></td>
    <td><code>@media</code></td>
    <td><code>no-preference | reduce</code></td>
<td><code>discrete</code></td>
  </tr>
</table>

### Example usage:
```
.image {
  background-image: url("image/heavy.jpg");
}

@media (prefers-reduced-data: reduce) { 
  /* Save-Data: On */
  .image {
    background-image: url("image/light.webp");
  }
}
```

Relevant articles, including use cases:
- [https://developers.google.com/web/updates/2016/02/save-data](https://developers.google.com/web/updates/2016/02/save-data)
- [https://www.ctrl.blog/entry/http-save-data](https://www.ctrl.blog/entry/http-save-data)
- [https://css-tricks.com/help-users-save-data/](https://css-tricks.com/help-users-save-data/)

frivoal:
Well, this would not **detect** the HTTP header, since that is sent from the client to the server, and this runs on the client, but we could have this and trigger it on the same conditions that cause the client to send that header.

frivoal:
The CSS-WG has considered this, and thinks this sounds generally reasonable, but we would like to hear more about what this would be used for. The cases that were brought up during our meeting all already had solutions.

Valid use cases for this that aren't already addressed by any existing thing would be a good justification to add this, so we're looking for examples of that.

Malvoz:
Adding example to use cases:
[...]

PR #4575, opened 2019-12-09
[css-mediaqueries] Add 'prefers-reduced-data' (addresses #2370)

huijing:
Content pulled from discussion within #2370

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -719,6 +719,10 @@ Detecting the desire for reduced data usage when loading a page: the 'prefers-re
 	It might be a system-wide setting exposed by the Operating System,
 	or a setting controlled by the User Agent.
 	
+	Note: User Agents may consider setting this based on the same user 
+	or system preference as they use to set the Save-Data HTTP request 
+	header.
+	
 	<div class="example">
 		For example, a site could honour the preference of a user who 
 		has turned on data-saving mode by serving a smaller image.
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -721,17 +721,17 @@ Detecting the desire for reduced data usage when loading a page: the 'prefers-re
 	
 	<div class="example">
 		For example, a site could honour the preference of a user who 
-		has turned on data-saving mode by not loading webfonts.
+		has turned on data-saving mode by serving a smaller image.
 
 		<pre>
-		body {
-			font-family: 'Maiola, serif;
+		.image {
+			background-image: url("images/light.jpg");
 		}
 
 		@media (prefers-reduced-data: reduce) { 
 			/* Save-Data: On */
-			body {
-				font-family: serif;
+			.image {
+				background-image: url("images/heavy.jpg");
 			}
 		}
 		</pre>
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -720,7 +720,8 @@ Detecting the desire for reduced data usage when loading a page: the 'prefers-re
 	or a setting controlled by the User Agent.
 	
 	Note: User Agents may consider setting this based on the same user 
-	or system preference as they use to set the Save-Data HTTP request 
+	or system preference as they use to set the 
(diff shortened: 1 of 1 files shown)


frivoal:
Thanks @huijing, this looks good. One suggestion: could you add a note saying that User Agents may consider setting this based on the same user or system preference as they use to set the Save-Data HTTP request header?

mediaqueries-5/Overview.bs
```diff
@@ -688,6 +688,56 @@ Detecting a forced color palette: the 'forced-colors' feature</h3>
...
+Detecting the desire for reduced data usage when loading a page: the 'prefers-reduced-data' feature</h3>
+
+	<pre class='descdef mq'>
+	Name: prefers-reduced-data
+	Value: no-preference | reduce
+	For: @media
+	Type: discrete
+	</pre>
+
+	The 'prefers-reduced-data' media feature is used to detect is an user
```
yoavweiss: s/is an user/if a user/

mediaqueries-5/Overview.bs
```diff
@@ -688,6 +688,56 @@ Detecting a forced color palette: the 'forced-colors' feature</h3>
...
+
+		<pre>
+		.image {
+			background-image: url("image/heavy.jpg");
+		}
+
+		@media (prefers-reduced-data: reduce) { 
+			/* Save-Data: On */
+			.image {
+				background-image: url("image/light.webp");
```
yoavweiss: There can be a future where we'd have user agents that support this MQ but not webp, so I would caution against coupling "lightweight" with browser support for formats. Having a way to do client side format negotiation for BG images would be a separate (and potentially worthwhile) use case IMO.

mediaqueries-5/Overview.bs
huijing: Ah, typo, my bad.

mediaqueries-5/Overview.bs
huijing: Should I switch out this use-case with another one from the thread? Or change the term "lightweight"?

yoavweiss: I'd just switch out the "webp" extension in favour of a "jpg"

huijing: Makes sense. Let me update that.

mediaqueries-5/Overview.bs
```diff
@@ -688,6 +688,60 @@ Detecting a forced color palette: the 'forced-colors' feature</h3>
...
+
+		<pre>
+		.image {
+			background-image: url("images/light.jpg");
+		}
+
+		@media (prefers-reduced-data: reduce) { 
+			/* Save-Data: On */
+			.image {
+				background-image: url("images/heavy.jpg");
```
yoavweiss: I think "light" and "heavy" switched places :)

mediaqueries-5/Overview.bs
```diff
@@ -688,6 +688,60 @@ Detecting a forced color palette: the 'forced-colors' feature</h3>
...
+			Indicates that user has expressed the preference for
+			lightweight alternate content.
+	</dl>
+	
+	The method by which the user expresses their preference can vary.
+	It might be a system-wide setting exposed by the Operating System,
+	or a setting controlled by the User Agent.
+	
+	Note: User Agents may consider setting this based on the same user 
+	or system preference as they use to set the Save-Data HTTP request
```
yoavweiss: You can point to https://wicg.github.io/savedata/ as a reference link

mediaqueries-5/Overview.bs
huijing: I am sorry.

mediaqueries-5/Overview.bs
huijing: Added link.

mediaqueries-5/Overview.bs
yoavweiss: no worries :)

frivoal:
Merged! thanks @huijing

yoavweiss:
Yay!! 🎉

ryantownsend:
Super excited about this! 😁

huijing:
omg, this is the highlight of my day :)

Issue #4475, opened 2019-10-31
[css-grid] Resolved values of grid-template-rows/columns don't round-trip


Loirooriol:
According to @fantasai in https://github.com/w3c/csswg-drafts/issues/4444#issuecomment-548006702,

> Roundtripping getComputedStyle() values is probably the most fundamental rule about defining them.

However, `grid-template-rows` and `grid-template-columns` don't always round-trip. That's because they set explicit tracks, which start just after the 1st grid line. However, when serialized, they also include implicit tracks, which may exist before the 1st grid line.

```js
gridContainer.style.gridAutoRows = "1px";
gridContainer.style.gridTemplateRows = "2px";
var cs = getComputedStyle(gridContainer);
cs.gridTemplateRows; // "2px"
gridItem.style.gridRow = "auto / 1";
gridContainer.style.gridTemplateRows = cs.gridTemplateRows; // "1px 2px"
gridContainer.style.gridTemplateRows = cs.gridTemplateRows; // "1px 1px 2px"
gridContainer.style.gridTemplateRows = cs.gridTemplateRows; // "1px 1px 1px 2px"
gridContainer.style.gridTemplateRows = cs.gridTemplateRows; // "1px 1px 1px 1px 2px"
```

On the one hand, knowing the final size of these implicit tracks can be useful. On the other hand, failing to round-trip sucks.

Loirooriol:
I have been thinking that we may be able to keep leading implicit tracks in the resolved value and also ensure round-tripping. We only need to allow `grid-template-rows` and `grid-template-columns` to define explicit tracks before the grid line number 1.

The current basic syntax is
```
<explicit-track-list> = [ <line-names>? <track-size> ]+ <line-names>?
<line-names>          = '[' <custom-ident>* ']'
```

Maybe we could change `<line-names>` to `'[' [ <custom-ident>* || 1? ] ']'`, with the constraint that this `1` can only appear once in the `<explicit-track-list>`.

This would define that line name to be the number 1. Tracks before that line would belong to the explicit grid, but items wouldn't be able to use integers in `grid-row/column` to be placed in such tracks (since 0 is invalid and negative integers refer to the end). Items could still reference line names in order to be placed there.

Then resolved values could preserve this `1` line name if it was specified, and otherwise it could be inserted when there are leading implicit tracks. The previous example would become
```js
cs.gridTemplateRows; // "2px"
gridItem.style.gridRow = "auto / 1";
[...]

PR #4584, opened 2019-12-11
[css-grid-1] Exclude implicit tracks from track listing resolved value. #4475

Loirooriol:
In order to have round-tripping. As resolved in
https://github.com/w3c/csswg-drafts/issues/4475#issuecomment-553525609


Loirooriol:
Closing this since some pages rely on this information. Won't be possible to exclude implicit tracks until we add a CSSOM API to obtain this information.

Issue #4588, opened 2019-12-12
[css-highlight-api] Inline styles for Highlight should apply on top of styles provided by ::highlight


sanketj:
This section (https://drafts.csswg.org/css-highlight-api-1/#c-and-h) of the highlight API spec mentions that "the User Agent must use the styles provided in the style attribute of the corresponding highlight when no applicable property has been specified via ::highlight()". This seems to imply that if a ::highlight has been specified, then the ::highlight styles should "win" over those coming from the .style attribute. What was initially proposed in the highlight API explainer (https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/highlight/explainer.md#application-of-css-properties) was that "the Highlight object's 'inline' style (i.e. properties set directly on the .style member) will be applied on top of the cascaded values for the given highlight identifier", which is the reverse of what we currently have spec'ed.

Not sure if this was a deliberate change but I think it is important that styles set via the .style attribute apply on top of ::highlight styles. This is a useful convenience for authors to be able to override the highlight styles from script. I can see it being especially useful when combined with highlight events (https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/highlight/events-explainer.md). For example, if the author wants to change the background color of a highlight when it is clicked, it is convenient to attach a click event listener to the highlight and then set the .style attribute in the callback.

cc: @frivoal @BoCupp-Microsoft @gked

frivoal:
This difference was deliberate when I drafted the document, but it's not something I'm completely sure about either, and did want to discuss. Here's the logic I had when I wrote this:

1) There was already a "default styling" behavior for the built-in pseudos, used for ::selection and ::active selection, defining how styles using the pseudo-element and styles coming from elsewhere would interact, so I thought that it made sense to try and use the same mechanics there rather than define something similar but different.

[...]

Issue #4602, opened 2019-12-14
[css-highlight-api] How should inline highlight styles with 'inherit' values be treated?


sanketj:
Moving this issue over from: https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/110

This is predicated on the assumption that highlight styles coming from the style attribute should apply on top of styles coming from ::highlight, which is being discussed here: https://github.com/w3c/csswg-drafts/issues/4588. We should probably resolve on that issue before tackling this one.

As described in the Highlight API explainer (https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/highlight/explainer.md#application-of-css-properties), inline styles for Highlight are applied on top of the cascaded styles. This means that if a style sheet is styling a highlight group with color: green, but an inline style on the same highlight group specifies color: red, the highlighted text will appear red (assuming no other styles are in play). However, what should the behavior be if the inline style specifies color: inherit? My thinking below...

Specifying an inherit value has no effect - it is equivalent to not specifying any value for the property. The spec (https://drafts.csswg.org/css-pseudo-4/#highlight-cascade) says that a highlight pseudo element should inherit values "from the corresponding highlight pseudo-element of its originating element’s parent element". For ranges, there is no single parent element, but rather there can be multiple containing elements that cover parts of the range. So we can say that each part of the highlight inherits values from the highlight pseudo element of that part's containing element. These are actually the same values that would be applied if the highlight's inline 'inherit' style is left off.

Consider the following scenario as an example.

```
<head>
<style>
:root::highlight(example-highlight) {
    color: blue;
}
div::highlight(example-highlight) {
    color: red;
}
</style>
</head>
<body><div>a</div>b<script>
    let div = document.querySelector('div');
    let highlightRange = new Range();
    highlightRange.setStart(div, 0);
    highlightRange.setEnd(document.body, 2);
    let highlight = new Highlight('example-highlight', highlightRange);
    CSS.highlights.add(highlight);
</script>
</body>
</html>
```

The above markup should render with a red 'a' and blue 'b'.

[...]

Issue #4398, opened 2019-10-04
[css-pseudo-4] Semantics of CSSPseudoElement : EventTarget


george-steel:
The CSSPseudoElement IDL specifies an inheritance from EventTarget, and this leaves a not a questions as to how enent handling on pseudo-elements should behave (which has no section on the spec). Is this interface only for animation events (transition events already target the parent but include a pseudoElement field to specify the pseudo-type)?

Currently system events (including click) which hit a pseudo-element are retargeted to the parent element before propagation. Should pseudo-elements be able to listen for these events, and if so, should `event.target` be the the pseudo-element (a breaking change) or the parent element.

Also, while the current system of event bubbling is well-defined for for the tree hierarchy for elements, some pseudo-elements (first-line, selection) feature multiple inheritance. How would pseudo-element events handle this. In normal event handling, a bubbled event always bubbles to it's parent element. However, when dealing with pseudo-elements, we have (following CSS cascade rules) we have the chain
    parent -> parent::first-line -> child -> child::first-line
How would we handle bubbling in this case? Do we break normal order and follow the CSS inheritance hierarchy? Do we propagate through all the pseudo-elements first?

Loirooriol:
I think this can only make sense for tree-abiding pseudo-elements: `::marker`, `::before` and `::after`. Not for things like `::first-line`.

bfgeek:
@flackr

PR #4607, opened 2019-12-16
[css-highlight-api][css-pseudo] add style attribute to pseudo-elements and use it in css-highlight-api as a replacement for the style attribute on Highlight

frivoal:
Exploratory Pull Request drafting what I suspect would be a good solution to #4588 (and to #4602 as well).

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -22,6 +22,7 @@ Indent: 2
 spec:css-color-3; type:value; text:currentcolor
 spec:css-color-3; type:property; text:color
 spec:fill-stroke-3; type:property; text:stroke-width
+spec:css-style-attr; type:dfn; text:style attribute
 </pre>
 
 <h2 id="intro">Introduction</h2>
@@ -864,6 +865,7 @@ Additions to the CSS Object Model</h2>
     interface CSSPseudoElement : EventTarget {
         readonly attribute CSSOMString type;
         readonly attribute Element element;
+        attribute CSSStyleDeclaration style;
     };
   </pre>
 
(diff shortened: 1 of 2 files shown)


css-pseudo-4/Overview.bs
```diff
@@ -883,12 +885,17 @@ Additions to the CSS Object Model</h2>
...
+  * the initially empty [=declarations=] list represents a [=style attribute=] for this pseudo-element
+          in the sence of [[!css-style-attr]] and as used in [[!css-cascade-4]].
+
   Note: This interface may be extended in the future
-  to other pseudo-element types
-  and/or to allow setting style information
-  through a {{CSSStyleDeclaration}} <code>style</code> attribute.
-  The current functionality is limited
-  to that which is needed to support [[web-animations-1]].
+  to other pseudo-element types.
```
sanketj: The CSSPseudoElement interface currently only supports `::before`, `::after` and `::marker` (lines 876-879). These are all [tree-abiding pseudo elements](https://www.w3.org/TR/css-pseudo-4/#treelike), meaning that they have a well-defined tree relationship with their owning element, and as a result, we can craft a sensible event flow when one of these receive an event.

CSSPseudoElement does not support (at least yet!) any pseudo elements that inherit styles from other pseudo elements, like ::selection or ::highlight, and not even pseudos that apply styles to multiple elements, like ::first-line. These seem harder to support because it's not clear what the event flow should be and how we would implement it (related issue: #4398 ). If we keep the same event path as we do currently, we probably won't deliver events to non-owning DOM elements that are covered by the pseudo. We would need to define a new event path and make sure it also works for other types of events (like pointer events - these are important for [highlight events](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/highlight/events-explainer.md)).

I'm happy to brainstorm about this further (assuming people want to expand CSSPseudoElement to non-tree abiding elements). If either you and @fantasai have any ideas or know of any discussions that have taken place about this before, I would love to hear them. I'll also ponder on this and try to bring something to discuss during our next highlight API discussion call.

frivoal:
Closed #4588 and #4602 differently. See https://github.com/w3c/csswg-drafts/commit/78f022fa4ace03ddb4ad4544f949fb95e3c1ee42

PR #4616, opened 2019-12-19
[web-animations-1] Resolve remaining issues missed in pseudo-element change.

### LINKED ISSUE #4502: [css-animations-2] List of pseudo-elements in composite order is non-exhaustive
In the definition of Animation [composite order](https://drafts.csswg.org/css-animations-2/#animation-composite-order), the list of pseudo-elements only includes `::marker`, `::before`, and `::after`. Since many other pseudo-elements exist, and there is no section of the spec restricting animations to these particular pseudo-elements, How should we resolve this? I can think of a few possibilities.

1. Replace the specific list with "pseudo-elements in alphabetical order by selector", as is done to order transitions by property.
2. Add an entry for "other pseudo-elements in alphabetical order by selector".
3. Add a canonical ordering to the pseudo-element spec and reference it. This would have to keep up with changes.
4. Add a restriction on which pseudo-elements may be animated to css-animations-1 and web-animations

### LINKED ISSUE #4586: [web-animations] Error handling in KeyframeEffect.pseudoElement
What should KeyframeEffect.pseudoElement do if it is set in an invalid or unsupported value? I can see a few possibilities here.

1. Throw an exception. This can easily break backwards compatibility as the set of pseudo-elements is not currently fixed. Running an animation with a new pseudoelement on an old browser can crash the whole script. I think it makes sense for inputs that don't start with "::" though.

2. Set pseudoElement to null. This fails without crashing but has the disadvantage of causing animations to happen on the wrong element in the case of an unsupported pseudo-element.

3. Leave pseudoElement unchanged. Again, can cause false positives.

4. Set pseudoElement to the text given and do not animate.

5. Set pseudoElement to s sentinel value (such as `::unknown`) which has no effect. Allows for error detection without the possibility of crashes.

My personal preference is for option 5.

### LINKED ISSUE #4701: [web-animations] Procedure not specified when property values cannot be calculated for effect target
In the [section on computed keyframes](https://drafts.csswg.org/web-animations-1/#calculating-computed-keyframes) it says

> Computed keyframes are produced using the following procedure. Note that this procedure is only performed on a keyframe effect having an effect target for which computed property values can be calculated.

All of the following procedures for applying an effect rely on the computed keyframes. However, there is nothing in the spec saying that those only apply when property values can be calculated. We should add a note to the keyframe effect intro clarifying this.

george-steel:
Fixup of #4437 addressing #4301.

* Add pseudoElement option to KeyframeEffect constructor body
* Fix #4502 adding catch-all pseudo-element case to composite order
* Fix #4586 adding error handling to KeyframeEffect.pseudoElement
* Fix #4701 making note of case when property values cannot be calculated

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -167,7 +167,8 @@ last) as follows:
     *  element
     *  ::marker
     *  ::before
-    *  other pseudo-elements in alphabetical order by selector
+    *  any other pseudo-elements not mentioned specifically in this list,
+       sorted in ascending order by the Unicode codepoints that make up each selector
     *  ::after
     *  element children
 
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -167,6 +167,7 @@ last) as follows:
     *  element
     *  ::marker
     *  ::before
+    *  other pseudo-elements in alphabetical order by selector
     *  ::after
     *  element children
 
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3355,8 +3355,7 @@ an <a>effect target</a> for which computed property values can be calculated.
 1.  Return <var>computed keyframes</var>.
 
 
-<h4 id="the-effect-value-of-a-keyframe-animation-effect">The effect value of
-  a keyframe effect</h4>
+### The effect value of a keyframe effect ### {#the-effect-value-of-a-keyframe-animation-effect}
 
 The <a>effect value</a> of a single property referenced by a <a>keyframe
 effect</a> as one of its <a lt="target property">target properties</a>, for a
@@ -3370,6 +3369,8 @@ given <var>iteration progress</var>, <var ignore=''>current iteration</var> and
 1.  If <a>animation type</a> of the <var>target property</var> is
     <a>not animatable</a> abort this procedure
     since the effect cannot be applied.
+1.  If the [=effect target=] is `null` or cannot have computed property values
+    calsulated, abort this procedure.
 1.  Define the <dfn>neutral value for composition</dfn> as a value
     which, when combined with an <a>underlying value</a> using the <a
     lt="composite operation add">add</a> <a>composite
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3201,6 +3201,11 @@ If the [=effect target=] is a [=pseudo-element=], the [=target element=] is
 its [=originating element=] and the [=target pseudo-selector=] is as required
 to specify that particular [=pseudo-element=].
 
+Note that not all [=effect targets=] specified in this manner (such as ''::part()''
+pseudo-elements, unsupported/invalid pseudo-elements, and `null`) have
+computed property values defined. Such targets are treated as if all
+properties are [=not animatable=].
+
 ### Keyframes ### {#keyframes-section}
 
 The <a>effect values</a> for a <a>keyframe effect</a>
@@ -4699,6 +4704,18 @@ interface KeyframeEffect : AnimationEffect {
 
     1.  Set the <a>target element</a> of <var>effect</var> to <var>target</var>.
 
+    1.  Set the [=target pseudo-selector=] to the result corresponding to the first
+        matching condition from below.
+
+        :   If <var>options</var> is a {{KeyframeEffectOptions}} object with a
+            {{KeyframeEffectOptions/pseudoElement}} property,
+        ::  Set the [=target pseudo-selector=] to the value of that property,
+            following the validation procedure from the
+            {{KeyframeEffect/pseudoElement}} setter.
+
+        :   Otherwise,
+        ::  Set the [=target pseudo-selector=] to `null`.
+
     1.  Let <var>timing input</var> be the result corresponding to the first
         matching condition from below.
 
@@ -4810,8 +4827,15 @@ interface KeyframeEffect : AnimationEffect {
     This is `null` if the [=effect target=] is an {{Element}} or is itself `null`.
     When not `null`, this must be a [=pseudo-element=] selector with
     [[selectors-4#pseudo-element-syntax|valid syntax]].
-    This will not accept invalid values and will always return the selector in
-    its canonical form (i.e. `::before` instead of `:before`).
+
+    This will only accept `null` or syntactically-valid selectors
+    (i.e. `:before`, `:after`, `:first-letter`, `:first-line`, or a string starting with `::`).
+    If such a string is not given, this will throw a {{DOMException}} with
+    error name {{SyntaxError}}.
+
+    If one of the legacy single-colon selectors is given, the selector is
+    changed to its double-colon version. If the empth string is given,
+    the selector is changed to `null`.
 
 :   <dfn attribute for=KeyframeEffect>composite</dfn>
 ::  The <a>composite operation</a> used to composite this
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3304,8 +3304,7 @@ To produce <a>computed keyframe offsets</a>, we define a procedure to
 
 <a>Computed keyframes</a> are produced using the following procedure.
 Note that this procedure is only performed on a <a>keyframe effect</a> having
-an <a>effect target</a> for which computed property values can be calculated,
-as otherwise all propereties are <a>not animatable</a>.
+an <a>effect target</a> for which computed property values can be calculated.
 
 1.  Let <var>computed keyframes</var> be an empty list of <a>keyframes</a>.
 
@@ -3364,9 +3363,6 @@ The <a>effect value</a> of a single property referenced by a <a>keyframe
 effect</a> as one of its <a lt="target property">target properties</a>, for a
 given <var>iteration progress</var>, <var ignore=''>current iteration</var> and
 <var>underlying value</var> is calculated as follows.
-Note that this procedure is only performed on a <a>keyframe effect</a> having
-an <a>effect target</a> for which computed property values can be calculated,
-as otherwise case all propereties are <a>not animatable</a>.
 
 1.  If <var>iteration progress</var> is <a>unresolved</a> abort this
     procedure.
@@ -3734,9 +3730,6 @@ will be added to &lsquo;Transition declarations&rsquo; level of the cascade.
 
 The <a>composited value</a> calculated for a CSS <a>target
 property</a> is applied using the following process.
-Note that this procedure is only performed on a <a>keyframe effect</a> having
-an <a>effect target</a> for which computed property values can be calculated,
-as otherwise case all propereties are <a>not animatable</a>.
 
 1.  Calculate the <var>base value</var> of the property as
     the value generated for that property by computing the computed value
@@ -4711,8 +4704,8 @@ interface KeyframeEffect : AnimationEffect {
 
     1.  Set the <a>target element</a> of <var>effect</var> to <var>target</var>.
 
-    1.  Set the <a>target pseudo-selector</a> of <var>effect</var> to the result
-        corresponding to the first matching condition from below.
+    1.  Set the [=target pseudo-selector=] to the result corresponding to the first
+        matching condition from below.
 
         :   If <var>options</var> is a {{KeyframeEffectOptions}} object with a
             {{KeyframeEffectOptions/pseudoElement}} property,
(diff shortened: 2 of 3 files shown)


george-steel:
@stephenmcgruer @birtles

web-animations-1/Overview.bs
```diff
@@ -4699,6 +4699,10 @@ interface KeyframeEffect : AnimationEffect {
 
     1.  Set the <a>target element</a> of <var>effect</var> to <var>target</var>.
 
+    1.  If <var>options</var> is a {{KeyframeEffectOptions}} object with a
```
stephenmcgruer: Is there a reason not to do this in the same step (or at least beside) the setting of `KeyframeEffect/composite` below?

web-animations-1/Overview.bs
```diff
@@ -4699,6 +4699,10 @@ interface KeyframeEffect : AnimationEffect {
...
+        {{KeyframeEffectOptions/pseudoElement}} property, let the <a>target
+        pseudo-selector</a> be the value of that property. Let it be `null` otherwise.
```
birtles: We need to define the error-handling behavior here (as I just commented on in https://github.com/w3c/csswg-drafts/issues/4586#issuecomment-568360267).

stephenmcgruer: If we move it to the same step as `composite` then it gets the same behavior as writing to the setter, which we can then spec (if not already done) as throwing for error cases.

birtles: Yes, I agree. Although, I think it should still be a separate step, probably before assigning `composite` (so that if _both_ are in error, we throw the error for the `pseudoElement` property instead). But we should use the same sort of prose, i.e. invoking the same error handling as for setting the property.

birtles:
@george-steel Anything I can do to help get this landed?

george-steel:
Sorry, I was delayed by the holidays and fixing a memory leak. I'll have another go at this in the next few days.

web-animations-1/Overview.bs
george-steel: done

web-animations-1/Overview.bs
george-steel: done

george-steel:
@birtles @stephenmcgruer Next draft is ready.

I agree with your advice and decided that the error handling and a couple of other issues needed to be fixed at the same time as the constructor. What I wrote here reflects the tentative decisions we made for blink in the absence of a spec.

css-animations-2/Overview.bs
```diff
@@ -167,6 +167,7 @@ last) as follows:
     *  element
     *  ::marker
     *  ::before
+    *  other pseudo-elements in alphabetical order by selector
```
stephenmcgruer: This seems reasonable to me, but I defer to if @birtles has any thoughts.

Editorial comment: I'm not expert, but I know web-animations uses the verbage: 'sorted in ascending order by the Unicode codepoints', so that may be preferable here (e.g. 'any other pseudo-elements not mentioned specifically in this list, sorted in ascending order by the Unicode codepoints that make up each selector').

web-animations-1/Overview.bs
```diff
@@ -3201,6 +3201,11 @@ If the [=effect target=] is a [=pseudo-element=], the [=target element=] is
 its [=originating element=] and the [=target pseudo-selector=] is as required
 to specify that particular [=pseudo-element=].
 
+Note that not all [=effect targets=] specified in this manner (such as ''::part()''
```
stephenmcgruer: I think this is attempting to address https://github.com/w3c/csswg-drafts/issues/4701, right?

I'm not opposed to this approach, though I wonder if we should just have an early exit in `5.3.3. The effect value of a keyframe effect` instead.

web-animations-1/Overview.bs
```diff
@@ -4699,6 +4704,18 @@ interface KeyframeEffect : AnimationEffect {
 
     1.  Set the <a>target element</a> of <var>effect</var> to <var>target</var>.
 
+    1.  Set the [=target pseudo-selector=] to the result corresponding to the first
```
stephenmcgruer: Editorial nits:

- Above line uses `<a>target element</a>`, this uses `[=target pseudo selector=]`.
- Above line mentions `effect`, this omits it.

web-animations-1/Overview.bs
```diff
@@ -4810,8 +4827,15 @@ interface KeyframeEffect : AnimationEffect {
...
     [[selectors-4#pseudo-element-syntax|valid syntax]].
-    This will not accept invalid values and will always return the selector in
-    its canonical form (i.e. `::before` instead of `:before`).
+
+    This will only accept `null` or syntactically-valid selectors
+    (i.e. `:before`, `:after`, `:first-letter`, `:first-line`, or a string starting with `::`).
+    If such a string is not given, this will throw a {{DOMException}} with
+    error name {{SyntaxError}}.
+
+    If one of the legacy single-colon selectors is given, the selector is
```
stephenmcgruer: This needs to be above the 'If such a string' line, otherwise we throw the exception before ever getting here.

web-animations-1/Overview.bs
```diff
@@ -4810,8 +4827,15 @@ interface KeyframeEffect : AnimationEffect {
...
+    changed to its double-colon version. If the empth string is given,
```
stephenmcgruer: Ditto re being above the 'If such a string' line (also, 'empty' I presume :D)

css-animations-2/Overview.bs
george-steel: done

web-animations-1/Overview.bs
george-steel: done

web-animations-1/Overview.bs
george-steel: fixed. Also, are pseudo-element selectors case-sensitive. If not, we should canonicalize the case as well.

web-animations-1/Overview.bs
george-steel: I kept this and added reminders to the other procedures just like `5.3.2. Calculating computed keyframes` has.

web-animations-1/Overview.bs
```diff
@@ -3304,7 +3304,8 @@ To produce <a>computed keyframe offsets</a>, we define a procedure to
 
 <a>Computed keyframes</a> are produced using the following procedure.
 Note that this procedure is only performed on a <a>keyframe effect</a> having
-an <a>effect target</a> for which computed property values can be calculated.
+an <a>effect target</a> for which computed property values can be calculated,
+as otherwise all propereties are <a>not animatable</a>.
```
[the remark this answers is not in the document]
birtles: Although, I'm not exactly sure what the purpose of this change is. How is it related to pseudo elements? What is the difference between saying the procedure is not applied and the properties are "not animatable"?

It's also a bit weird to say they are "not animatable" since that's something that is normally defined for the property itself. We should instead say something like they follow the same logic as not animatable properties or something. Once I understand the purpose of the change I can help suggest wording.

web-animations-1/Overview.bs
```diff
@@ -3363,6 +3364,9 @@ The <a>effect value</a> of a single property referenced by a <a>keyframe
 effect</a> as one of its <a lt="target property">target properties</a>, for a
 given <var>iteration progress</var>, <var ignore=''>current iteration</var> and
 <var>underlying value</var> is calculated as follows.
+Note that this procedure is only performed on a <a>keyframe effect</a> having
+an <a>effect target</a> for which computed property values can be calculated,
+as otherwise case all propereties are <a>not animatable</a>.
```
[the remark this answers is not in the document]
birtles: As above, I'd probably avoid invoking the "not animatable" language since that's really an animation type and I think it is confusing to use that concept here. It's probably better just to say this procedure is not run on that case, or, better still, add a step that early returns if the the effect target is in a context where we can't calculate property values.

web-animations-1/Overview.bs
```diff
@@ -4699,6 +4711,18 @@ interface KeyframeEffect : AnimationEffect {
...
     1.  Set the <a>target element</a> of <var>effect</var> to <var>target</var>.
 
+    1.  Set the <a>target pseudo-selector</a> of <var>effect</var> to the result
+        corresponding to the first matching condition from below.
+
+        :   If <var>options</var> is a {{KeyframeEffectOptions}} object with a
+            {{KeyframeEffectOptions/pseudoElement}} property,
+        ::  Set the [=target pseudo-selector=] to the value of that property,
+            following the validation procedure from the
+            {{KeyframeEffect/pseudoElement}} setter.
```
birtles: Nit: For consistency, could we follow the same format we use for `composite`?

e.g.

```bs
Set the [=target pseudo-selector=] to the value of the {{KeyframeEffectOptions/pseudoElement}} property.

When assigning this property, the error-handling defined for the corresponding setter on the
{{KeyframeEffect}} interface is applied. If the setter requires an exception to be thrown, this procedure
must throw the same exception and abort all further steps.
```

web-animations-1/Overview.bs
birtles: It looks like it still says `the empth string`.

Could we rewrite this as something a bit more imperative like:

```
The [=target pseudo-selector=].

`null` if this effect has no [=effect target=] or
if the [=effect target=] is an element (i.e. not a pseudo-element).
When the [=effect target=] is a pseudo-element,
this specifies the pseudo-element selector (e.g. "::before").

On setting,
sets the [=target pseudo-selector=] of the [=animation effect=]
to the provided value after applying the following exceptions:

*   If the provided value is not a syntactically valid
    <pseudo-element-selector> the user agent must [=throw=]
    a {{DOMException}} with error name {{TypeError}} and leave the
    [=target pseudo-selector=] of this [=animation effect=] unchanged.

*   If one of the legacy Selectors Level 2 single-colon selectors
    (':before', ':after', ':first-letter', or ':first-line') is
    specified, the [=target pseudo-selector=] must be set to the
    equivalent two-colon selector (e.g. '::before').

*   If the empty string is specified, the [=target pseudo-selector=] is 
    cleared (hence getting this member subsequently would return
    `null`).

```

Although, now that I look at it, do we need that last condition?

Isn't it better to only allow _either_ `null` or an empty string, but not both?

web-animations-1/Overview.bs
```diff
@@ -4704,8 +4711,8 @@ interface KeyframeEffect : AnimationEffect {
 
     1.  Set the <a>target element</a> of <var>effect</var> to <var>target</var>.
 
-    1.  Set the [=target pseudo-selector=] to the result corresponding to the first
-        matching condition from below.
+    1.  Set the <a>target pseudo-selector</a> of <var>effect</var> to the result
```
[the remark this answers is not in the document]
george-steel: @stephenmcgruer requested this change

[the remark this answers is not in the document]
stephenmcgruer: Sorry, I had seen the use of `<a>` in the line above and asked George to match it. Didn't realize `[= =]` was a more modern syntax and we were trying to switch to it.

Apologies for wasting your time George, but can you switch it back to `[= =]` then?

web-animations-1/Overview.bs
george-steel: moot

web-animations-1/Overview.bs
george-steel: done

web-animations-1/Overview.bs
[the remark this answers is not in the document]
george-steel: good point

web-animations-1/Overview.bs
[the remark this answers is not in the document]
george-steel: removed

web-animations-1/Overview.bs
```diff
@@ -3730,6 +3734,9 @@ will be added to &lsquo;Transition declarations&rsquo; level of the cascade.
 
 The <a>composited value</a> calculated for a CSS <a>target
 property</a> is applied using the following process.
+Note that this procedure is only performed on a <a>keyframe effect</a> having
+an <a>effect target</a> for which computed property values can be calculated,
+as otherwise case all propereties are <a>not animatable</a>.
```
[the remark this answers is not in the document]
george-steel: removed

web-animations-1/Overview.bs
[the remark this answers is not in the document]
george-steel: reverted

web-animations-1/Overview.bs
george-steel: done

web-animations-1/Overview.bs
```diff
@@ -3201,6 +3201,10 @@ If the [=effect target=] is a [=pseudo-element=], the [=target element=] is
 its [=originating element=] and the [=target pseudo-selector=] is as required
 to specify that particular [=pseudo-element=].
 
+Note that not all [=effect targets=] specified in this manner (such as ''::part()''
+pseudo-elements, unsupported/invalid pseudo-elements, and `null`) have
```
birtles: I'm not sure invalid pseudo-elements can exist, right? They're rejected at the API boundary?

What does ``null`` refer to here? The lack of a target element? (Represented by `null` in the programming interface) Or the lack of a target pseudo-selector?

web-animations-1/Overview.bs
```diff
@@ -3366,6 +3369,8 @@ given <var>iteration progress</var>, <var ignore=''>current iteration</var> and
 1.  If <a>animation type</a> of the <var>target property</var> is
     <a>not animatable</a> abort this procedure
     since the effect cannot be applied.
+1.  If the [=effect target=] is `null` or cannot have computed property values
```
birtles: Nit: Technically I guess this should say, "The [=keyframe effect=] does not have an [=effect target=]...` since we're talking about the model here. They fact that the lack of an associated effect target is represented by `null` in the programming interface is specific to the programming interface.

PR #4641, opened 2019-12-31
[css-ruby] Describe position and size of ruby level container when empty

### LINKED ISSUE #4631: [css-ruby] Size and position of empty ruby level containers
The spec currently says:
> Each *ruby base container* is sized and positioned to contain exactly all of its *ruby bases*’ margin boxes.

and
> Each *ruby annotation container* is sized and positioned to contain exactly all of its *ruby annotations*’ margin boxes.

However, it is unclear what should happen when a ruby level container doesn't have any child.

I suggest that for empty ruby base container, it should probably behave as if there is an empty ruby base inside, so that it would laid out like a normal empty inline element.

I'm not sure what would be a useful behavior for ruby annotation container. I kinda tend to think that it should keep being zero size in that case, but we may need to define its position somehow as well?

This issue is raised because of [a bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1603354) to Gecko after I [changed the behavior](https://bugzilla.mozilla.org/show_bug.cgi?id=1565681) to match the spec.

upsuper:
This intends to fix #4631.

I think this can be considered to be a bug fix and thus wouldn't really need to go through the working group.

### DIFF
--- a/css-ruby-1/Overview.bs
+++ b/css-ruby-1/Overview.bs
@@ -629,6 +629,8 @@ Ruby Layout</h2>
 	exactly as if its <i>ruby bases</i> were a regular sequence of inline boxes.
 	Each <i>ruby base container</i> is sized and positioned
 	to contain exactly all of its <i>ruby bases</i>’ margin boxes.
+	If it has no <i>ruby base</i> inside,
+	it behaves as if there is an empty one.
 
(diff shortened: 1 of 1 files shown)


frivoal:
Fixed in https://github.com/w3c/csswg-drafts/pull/4941, closing

PR #4644, opened 2020-01-02
Fix typos in &lt; and &gt; html entities

renatahodovan:


frivoal:
This is a [class 1 change](https://www.w3.org/2019/Process-20190301/#correction-classes), which therefore does not require a [Contributor License Grants](https://www.w3.org/2019/Process-20190301/#contributor-license). Merging.

Issue #3229, opened 2018-10-21
[css-text-4] Text-spacing is too strict


litherum:
https://drafts.csswg.org/css-text-4/#text-spacing-property

The spec [states](https://drafts.csswg.org/css-text-4/#valdef-text-spacing-ideograph-alpha):
> `ideograph-alpha`
> Creates 1/4em extra spacing between runs of ideographs and non-ideographic letters.

... and includes definitions of "ideographs" and "non-ideographic letters."

[Later](https://drafts.csswg.org/css-text-4/#fullwidth-collapsing), the spec also provides required collapsing rules:

> Set fullwidth opening punctuation half-width if the previous character is a fullwidth opening punctuation, fullwidth middle dot punctuation, or ideographic space (U+3000), or if the previous character is a fullwidth closing punctuation of an equivalent or larger font-size. Else set it full-width.
> Set fullwidth closing punctuation half-width if the next character is a fullwidth closing punctuation, fullwidth middle dot punctuation, or ideographic space (U+3000), or if the next character is a fullwidth opening punctuation of a larger font-size. Else set it full-width.

When discussing this with our team of designers at Apple, it turns out that we have our own similar behavior for native text. Our design team at Apple does increase spacing between script changes and decrease spacing between adjacent punctuation on a line, and the details of how we do it are similar but don't exactly match [JLREQ](https://w3c.github.io/jlreq/) or CSS. We used JLREQ as an input to the design, but there were other inputs too. From speaking to our designers, it turns out the specific rules for exactly how much space to expand or shrink can vary from one design house to the next.

Therefore, I'd like to propose the following changes to the spec:
1. In the above description of `ideographic-alpha` and `ideographic-numeric`, change "Creates 1/4em extra spacing between" to "Creates extra spacing between."
2. Provide some wiggle room around what the UA determines as "ideographs" and "non-ideographic characters" for the purpose of this property. The spec should continue to state that the desire is to add spacing between ideographs and non-ideographic letters, but the spec should let the classification be UA-dependent.
3. Remove "(or the font has proportional fullwidth punctuation glyphs)." Fullwidth characters should be fullwidth; if they're not, the font is broken.
4. In the [Fullwidth Punctuation Collapsing](https://drafts.csswg.org/css-text-4/#fullwidth-collapsing) section, change "Set fullwidth ... punctuation half-width if the" to "Set fullwidth ... punctuation narrower". Similarly to above, allow some wiggle room around exactly which characters this rule applies to.
[...]

PR #4657, opened 2020-01-08
[css-text-4] Add auto value to text-spacing

litherum:
Closes https://github.com/w3c/csswg-drafts/issues/3229

https://lists.w3.org/Archives/Public/www-style/2018Nov/0006.html

> RESOLVED: Add a new text-spacing:auto value, which is not the initial value. (Issue #3229)

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -1873,7 +1873,7 @@ Character Class Spacing: the 'text-spacing' property</h3>
 
 	<pre class="propdef">
 	Name: text-spacing
-	Value: normal | none |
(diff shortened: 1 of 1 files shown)


astearns:
Should we define this as following platform conventions or just “let the UA decide”? Is it actually possible for Safari/Chrome/Firefox to match behavior on MacOS, for example? Or should I expect spacing to differ on Firefox depending on whether it's running on MacOS, Win10 or Linux?

litherum:
I suppose it can mean "let the UA decide" until other browsers want to implement different behaviors on different platforms. I can edit the PR.

Issue #4471, opened 2019-10-30
[cssom-1][mediaqueries-5][css-color] Dealing with bi-plane (video/graphics) when reporting values


gregwhitworth:
Hey folks,

Over in the Media WG we're trying to help web developers serving media content provide the best end user experience and we've encountered an interesting scenario that we'd like to get the CSSWG's feedback on. @chcunningham did a great job summarizing the *very long* discussion down to the various options we feel are available [here](https://github.com/w3c/media-capabilities/issues/135#issuecomment-547673815).

> New summary:
>
> For most UA implementations, video is like any other element in terms of its participation in the CSS and Document object model. But TVs (and cast dongles) have distinct code/hw paths ("planes") for rendering video vs other elements ("graphics"). Generally, video can be rendered at the devices native resolution (e.g. 4K) and can take full advantage of the devices HDR/WCG capabilities, while graphics are likely rasterized at a much lower resolution (720p or 1080p) with ~sRGB 8 bit color. Many TVs have this problem. TVs are an very important media consumer. This problem is not predicted to work itself out with time.
>
>With this in mind, we aim to design an API that works for TV and non-TV UAs to signal when a Screen supports HDR and/or 4K resolution. We've discussed 3 approaches.
>
>**1. clean but half-broken:** Add Screen.hdrSupported alongside existing Screen.width and height properties
>
>Like width and height, hdrSupported is assumed to describe hdr support for video and graphics alike. This is a natural extension of the current Screen object, but it offers TVs no way to signal their distinct video vs graphics capabilities.
>
>TVs can pick one plane to describe and lie about the other. Lying is obviously unpleasant, and it has various trade-offs / breaks depending on which plane is reported and specific use cases (covered earlier in this thread).
>
>**2. verbose but functional:** Add video* prefixed media queries media features for the video plane. The running list of features to prefix includes: color, color-gamut, color-index, device-aspect-ratio, resolution, height, width and (new) hdrSupported
>
>TVs will use the video* prefixed versions to surface the hdr and 4k support for video, while signalling sdr and 720p-1080p for graphics via the existing (unprefixed) properties. For non-TVs, the video* prefixed values will all be the same as their unprefixed counterparts.
>
>This is equivalent (and could accompany) an earlier proposal to add a new "video" property bag to Screen for video plane properties.
>
>This completely solves the problem. It also standardizes this TV quirk into the web platform in a highly visible manner.

[...]

PR #4678, opened 2020-01-17
Adding of video-* properties and dynamic-range to handle bi-plane scenarios

gregwhitworth:
As resolved in issue #4471 I have added in:

```
video-color-gamut
video-dynamic-range
video-width
video-height
video-resolution
dynamic-range
```
And a brief description of the issue along with normative text of when to answer differently between a single or bi-plane scenario.

Specific to dynamic-range we placed a general definition there and are interested in thoughts since standard & high may fluctuate over time but I'd expect the spec text too as well.

### DIFF
--- a/mediaqueries-4/Overview.bs
+++ b/mediaqueries-4/Overview.bs
@@ -1706,45 +1706,47 @@ Dynamic Range: the 'dynamic-range' feature</h3>
 	</pre>
 </h3>
 
-High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
+'@media/dynamic-range' represents the combination of brightness and color
+depth.
 
-<dl dfn-type=value dfn-for="dynamic-range">
+<dl dfn-type=value dfn-for="@media/dynamic-range">
 	<dt><dfn>high</dfn>
 	<dd>
 		If the following are true for the UA and the output device:
 		* max brightness is considered high
 		* color depth is greater than 24 bit
-		* has a <a href="#contrast-of-display">high contrast ratio</a>
+		* has a <a>high contrast ratio</a>
 	<dt><dfn>standard</dfn>
 	<dd>
 		If the following are true for the UA and the output device:
-		* max brightness is considered standards
+		* max brightness is considered standard
 		* color depth is 24 bit or 8 bit per color component of RGB
-		* has a <a href="#contrast-of-display">low contrast ratio</a>
+		* has a <a>low contrast ratio</a>
 </dl>
 
 <h3 id="contrast-brightness-of-display">
 	Determining contrast and brightness of display</h3>
 
 This text is non-normative. There currently is no agreed upon standardization of
-high contrast ratio, low contrast ratio, max brightness or low brightness as it
-relates to displays. As such, the determination for <a>dynamic-range</a> and
-<a>video-dymanic-range</a> will be defined by the user agent.
+<dfn export>high contrast ratio</dfn>, <dfn export>low contrast ratio</dfn>,
+max brightness or low brightness as it relates to displays. As such, the
+determination for '@media/dynamic-range' and '@media/video-dynamic-range' will
+be defined by the user agent.
 
 <h3 id="video-prefixed-properties">
-Handling bi-plane scenarios</h3>
+Video Prefixed Properties scenarios</h3>
 
-	In some scenarios, such as TVs, the user agents may have a graphics
-	plane dedicated to video that has its own discrete hardware. In order
-	to allow UAs in environments that have more than one plane for graphics
-	rendering the following properties should be used to correctly answer
-	video specific queries.
+    Some user agents, including many TVs, render video and graphics in two
+	separate "planes" (bi-plane) with distinct screen characteristics. A set of
+	video-prefixed features is provided to describe the video plane.
 
-	For any UA on a bi-plane device you **MUST** return the values for:
-	<a>video-color-gamut</a>; <a>video-width</a>; <a>video-height</a>; <a>video-resolution</a>;
-	<a>video-dynamic-range</a>; based on the plane that renders video. If the UA is not in a bi-plane
-	scenario then it **MUST** return the values of the single plane
-	(they're effectively aliases to their non-video counterparts).
+	Any bi-plane implementation **MUST** return values based on the video plane
+	for the following features: '@media/video-color-gamut'; '@media/video-width';
+	'@media/video-height'; '@media/video-resolution'; '@media/video-dynamic-range'.
+	All other features **MUST** return values based on the graphics plane.
+
+    Non bi-plane implementations **MUST** return the same values for
+	video-prefixed features and their non-prefixed counterparts.
 
 <h3 id="video-color-gamut">
 Video Color Display Quality: the 'video-color-gamut' feature</h3>
@@ -1757,7 +1759,7 @@ Video Color Display Quality: the 'video-color-gamut' feature</h3>
 	</pre>
 </h3>
 
-The 'video-color-gamut' media feature describes the approximate range of colors
+The '@media/video-color-gamut' media feature describes the approximate range of colors
 	that are supported by the UA and output device's video plane.
 	That is, if the UA receives content with colors in the specified space
 	it can cause the output device to render the appropriate color,
@@ -1776,21 +1778,22 @@ Video Dynamic Range: the 'video-dynamic-range' feature</h3>
 	</pre>
 </h3>
 
-High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
+'@media/video-dynamic-range' has the same values and definitions as '@media/dynamic-range',
+but describes the UA's and device's video plane.
 
 <dl dfn-type=value dfn-for="video-dynamic-range">
 	<dt><dfn>high</dfn>
 	<dd>
 		If the following are true for the UA and the output device's video plane:
-		* color gamut is <a>p3</a> or <a>rec2020</a>
+		* color gamut is ''p3'' or ''rec2020''
 		* color depth is greater than 24 bit
-		* has a <a href="#contrast-of-display">high contrast ratio</a>
+		* has a high contrast ratio
 	<dt><dfn>standard</dfn>
 	<dd>
 		If the following are true for the UA and the output device's video plane:
-		* color gamut is <a>sRGB</a>
+		* color gamut is sRGB
 		* color depth is 24 bit or 8 bit per color component of RGB
-		* has a <a href="#contrast-of-display">low contrast ratio</a>
+		* has a low contrast ratio
 </dl>
 
 <h3 id="video-width">
@@ -1840,13 +1843,11 @@ Video Display Resolution: the 'video-resolution' feature</h3>
 	Type: range
 	</pre>
 
-	The 'video-resolution' media feature describes the resolution of the output
+	The '@media/video-resolution' media feature describes the resolution of the output
 	device's video plane, i.e. the density of the pixels, taking into account the <a spec=cssom-view>page zoom</a>
 	but assuming a <a spec=cssom-view>pinch zoom</a> of 1.0.
 
 	The 'video-resolution' media feature is <a>false in the negative range</a>
-
-<p class="issue">There needs to be a solution for dynamic range</p>
 
 <!--
 ████ ██    ██ ████████ ████████ ████████     ███     ██████  ████████ ████  ███████  ██    ██
--- a/mediaqueries-4/Overview.bs
+++ b/mediaqueries-4/Overview.bs
@@ -1695,8 +1695,36 @@ Color Display Quality: the 'color-gamut' feature</h3>
 	you can use this feature in a negated boolean-context fashion:
 	''not (color-gamut)''.
 
+<h3 id="dynamic-range">
+Dynamic Range: the 'dynamic-range' feature</h3>
+
+	<pre class='descdef mq'>
+	Name: dynamic-range
+	Value: standard | high
+	For: @media
+	Type: discrete
+	</pre>
+</h3>
+
+High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
+
+<dl dfn-type=value dfn-for="dynamic-range">
+	<dt><dfn>high</dfn>
+	<dd>
+		If the following are true for the UA and the output device:
+		* color gamut is <a>p3</a>, <a>rec2020</a>
+		* color depth is greater than 24 bit
+		* has a high contrast ratio
+	<dt><dfn>standard</dfn>
+	<dd>
+		If the following are true for the UA and the output device:
+		* color gamut is <a>sRGB</a>
+		* color depth is 24 bit or 8 bit per color component of RGB
+		* has a low contrast ratio
+</dl>
+
 <h3 id="video-prefixed-properties">
-Handling bi-plane displays</h3>
+Handling bi-plane scenarios</h3>
 
 	In some scenarios, such as TVs, they have a graphics plane dedicated 
 	to video that has its own discrete hardware. In order to allow UAs in 
@@ -1704,13 +1732,14 @@ Handling bi-plane displays</h3>
 	following properties should be used to correctly answer video specific 
 	queries.
 
-	For any useragent on a bi-plane device you **MUST** return the values for: 
-	<a>video-color-gamut</a>; <a>video-width</a>; <a>video-height</a>; <a>video-resolution</a>; 
-	based on the plane that renders video. If the useragent is not in a bi-plane scenario then 
-	you **MUST** return the values of the single plane (they're effectively aliases to their non-video counterparts).
+	For any UA on a bi-plane device you **MUST** return the values for: 
+	<a>video-color-gamut</a>; <a>video-width</a>; <a>video-height</a>; <a>video-resolution</a>;
+	<a>video-dynamic-range</a>; based on the plane that renders video. If the UA is not in a bi-plane 
+	scenario then it **MUST** return the values of the single plane 
+	(they're effectively aliases to their non-video counterparts).
 
 <h3 id="video-color-gamut">
-Color Display Quality: the 'video-color-gamut' feature</h3>
+Video Color Display Quality: the 'video-color-gamut' feature</h3>
 
 	<pre class='descdef mq'>
 	Name: video-color-gamut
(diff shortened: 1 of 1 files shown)


gregwhitworth:
ccing: @vi-dot-cpp @AmeliaBR @chcunningham @mwatson2

Crissov:
So, although we all agree that the `color-gamut` keywords are badly designed, but need to stay as is because they are already shipping, we still copy them verbatim to `video-color-gamut`?

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,151 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+	</pre>
+</h3>
+
+High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
+
+<dl dfn-type=value dfn-for="dynamic-range">
+	<dt><dfn>high</dfn>
+	<dd>
+		If the following are true for the UA and the output device:
+		* color gamut is <a>p3</a>, <a>rec2020</a>
```
svgeesus: Is that an "or"?

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,151 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+		* color depth is greater than 24 bit
+		* has a high contrast ratio
```
svgeesus: is contrast ratio defined?

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,151 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+	</pre>
+</h3>
+
+High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
+
+<dl dfn-type=value dfn-for="video-dynamic-range">
+	<dt><dfn>high</dfn>
+	<dd>
+		If the following are true for the UA and the output device's video plane:
+		* color gamut is <a>p3</a>, <a>rec2020</a>
```
svgeesus: same comment as above.

I wonder if a link is better (for maintainability) than the text, like 

"High and standard have the same meanings as [link to the other one]

mediaqueries-4/Overview.bs
gregwhitworth: Hmmmm.... as it relates to displays I'm not sure. @vi-dot-cpp @chcunningham @mwatson2 do you all happen to know if this is define anywhere? If not, we'll need to define what "has a high contrast ratio" means

mwatson2:
It would of course be great to reference some widely agreed definition of High Dynamic Range or Wide Color Gamut as they pertain to display capabilities. However, I'm not aware that any such thing exists. Given any candidate for such a thing, someone will make a display that is 1 nit less bright or has 0.1% P3 smaller gamut and claim (accurately) that the experience on that display is just as deserving of the moniker.

In any case, the question that the site wishes to answer with these properties is whether the user experience will be better if the HDR / WCG version of the content is provided (and mapped down to the display capability) than if the SDR / NCG version is provided (and mapped up to the display capability). This of course is a matter of opinion. The opinion that matters in this case is the users (or, some may argue, the filmmakers).

So, I think the best we can do is ask the user's agent to answer this question on behalf of the user.

svgeesus:
@gregwhitworth I'm nt worried enough to want to hold up this PR

svgeesus:
So I think we can merge this?

gregwhitworth:
@svgeesus I made your modifications you asked for. I also added some non-normative text regarding the current state of handling contrast of the display that I'd like you to review quickly. If you're fine with all of that please merge it in.

svgeesus:
@gregwhitworth thanks for that, all looks good to me. I pinged @frivoal just as the CSSWG f2f was ending, he wants to review more closely and perhaps tighten some wording before merging.

chcunningham:
Sorry folks, I'm late to the party. Let me make a pass on the language real quick.

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+This text is non-normative. There currently is no agreed upon standardization of 
+high contrast ratio or low contrast ratio as it relates to displays. 
+As such, the determination for <a>dynamic-range</a> and <a>video-dymanic-range</a> 
+will be defined by the user agent.
+
+<h3 id="video-prefixed-properties">
+Handling bi-plane scenarios</h3>
+
+	In some scenarios, such as TVs, they have a graphics plane dedicated
```
chcunningham: Grammar nit: "they" is funny and a little ambiguous here. Maybe "user agents"

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+	Type: discrete
+	</pre>
+</h3>
+
+High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
+
+<dl dfn-type=value dfn-for="dynamic-range">
+	<dt><dfn>high</dfn>
+	<dd>
+		If the following are true for the UA and the output device:
```
chcunningham: I'd remove color gamut and add in max brightness. 

The core considerations for high dynamic range are more focused on brightness: high maximum brighness, high contrast ratio, and high bit depth. Similar to contrast ratio, the meaning of "high" for these other attributes is also not standardized. Something like >= 400 nits and >= 10 bits per channel, but I wouldn't write those values into the spec. Combined (for all channels) bit depth >= 24 is probably fine given the history/current spec for that attribute. 

Color gamut is distinct and well covered by the existing attribute. Its true that we will usually find HDR displays to have one of these color gamuts, but its not inherent in the definition of HDR and we'll be more future proof being less specific about the color gamuts.

mediaqueries-4/Overview.bs
gregwhitworth: is it user agents or the output display?

gregwhitworth: I went ahead and made it user agent for now

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+This text is non-normative. There currently is no agreed upon standardization of
+high contrast ratio, low contrast ratio, max brightness or low brightness as it
+relates to displays. As such, the determination for <a>dynamic-range</a> and
+<a>video-dymanic-range</a> will be defined by the user agent.
+
+<h3 id="video-prefixed-properties">
+Handling bi-plane scenarios</h3>
+
+	In some scenarios, such as TVs, the user agents may have a graphics
```
chcunningham: I think the language here could be a little more direct. I couldn't find a great way to do this on a per-word basis, so I'm proposing the following as wholesale replacement. Written in my voice, so some difference in style - please feel free to ignore and/or revise further. Apologies for the large edit.

---

Some user agents, including many TVs, render video and graphics in two separate "planes" (bi-plane) with distinct screen characteristics. A set of video- prefixed features is provided to describe the video plane.

Any bi-plane implementation **MUST** return values based on the video plane for the following features: <a>video-color-gamut</a>; <a>video-width</a>; <a>video-height</a>; <a>video-resolution</a>; <a>video-dynamic-range</a>. All other features **MUST** return values based on the graphics plane. 

Non bi-plane implementations **MUST** return the same values for video-prefixed features and their non-prefixed counterparts.

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+	<pre class='descdef mq'>
+	Name: video-dynamic-range
+	Value: standard | high
+	For: @media
+	Type: discrete
+	</pre>
+</h3>
+
+High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
```
chcunningham: ditto here for transfer function and color gamut

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+<dl dfn-type=value dfn-for="video-dynamic-range">
```
chcunningham: I'd make the same edits here as for dynamic-range. But you might instead give a pointer to the non-prefixed feature like for video-color-gamut. ex: "Dynamic range values and definitions are the same as <a href="#dynamic-range">dynamic-range</a>"

chcunningham: If you take that second option, be sure to add "by the UA and output device's video plane." to the end of the intro sentence on l.1779.

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+	Type: range
+	</pre>
+
+	The 'video-resolution' media feature describes the resolution of the output
+	device's video plane, i.e. the density of the pixels, taking into account the <a spec=cssom-view>page zoom</a>
+	but assuming a <a spec=cssom-view>pinch zoom</a> of 1.0.
+
+	The 'video-resolution' media feature is <a>false in the negative range</a>
+
+<p class="issue">There needs to be a solution for dynamic range</p>
```
chcunningham: Do you mean to leave this here?

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+<h3 id="contrast-brightness-of-display">
+	Determining contrast and brightness of display</h3>
+
+This text is non-normative. There currently is no agreed upon standardization of
+high contrast ratio, low contrast ratio, max brightness or low brightness as it
+relates to displays. As such, the determination for <a>dynamic-range</a> and
+<a>video-dymanic-range</a> will be defined by the user agent.
+
+<h3 id="video-prefixed-properties">
+Handling bi-plane scenarios</h3>
```
chcunningham: I like "Video Prefixed Properties". Purely subjective. 

I was curious to see what this would look like relative to the other features, so I dumped the raw text into https://api.csswg.org/bikeshed/, but this produces some build errors that stopped the output from generating. LMK if I'm holding it wrong.

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,159 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+	<pre class='descdef mq'>
+	Name: dynamic-range
+	Value: standard | high
+	For: @media
+	Type: discrete
+	</pre>
+</h3>
+
+High dynamic range (HDR) represents the combination of brightness, color depth, transfer function, and color gamut.
```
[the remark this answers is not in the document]
gregwhitworth: > color gamut to be consistent

I'm torn, technically this is describing what HDR is but I also agree that we don't bring into account transfer function in the heuristic (maybe we should?). Besides that, color gamut is included in the heuristic unless I'm mis-understanding your point.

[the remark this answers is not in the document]
chcunningham: HDR TVs screen's will generally have particular eotfs and color gamuts, but "dynamic range" is a distinct concept describing luminance. See [these comments](https://github.com/w3c/csswg-drafts/issues/4471#issuecomment-548165527) from me and @mwatson2 in the issue thread.

Removing mention of color gamut and eotf will make the intro sentence more consistent with the changes you recently made to the definitions of "high" and "standard" below.

mediaqueries-4/Overview.bs
vi-dot-cpp: The bikeshed compile issues have been fixed. You can try again.

mediaqueries-4/Overview.bs
vi-dot-cpp: I took your recommendation -- thanks!

vi-dot-cpp:
@chcunningham Thanks for your comments. They should all be addressed -- let me know if issues remain. cc: @gregwhitworth @gregwfreedman

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,160 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+	Determining contrast and brightness of display</h3>
+
+This text is non-normative. There currently is no agreed upon standardization of
+<dfn export>high contrast ratio</dfn>, <dfn export>low contrast ratio</dfn>,
+max brightness or low brightness as it relates to displays. As such, the
+determination for '@media/dynamic-range' and '@media/video-dynamic-range' will
+be defined by the user agent.
+
+<h3 id="video-prefixed-properties">
+Video Prefixed Properties scenarios</h3>
```
chcunningham: Sorry, let me revise my advise. I'd name this "Video Prefixed Feature", not properties, and no mention of scenarios. "Features" is more correct from the CSS pov.

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,160 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+    Some user agents, including many TVs, render video and graphics in two
+	separate "planes" (bi-plane) with distinct screen characteristics. A set of
+	video-prefixed features is provided to describe the video plane.
+
+	Any bi-plane implementation **MUST** return values based on the video plane
```
chcunningham: Change **MUST** to simply MUST or must. I used that formatting earlier just to give emphasis (bold) the github comment, but it seems most references to "must" in this spec are simply lower case non-bold (with one a few being caps non-bold). None have the surrounding asterisks.

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,160 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+	<pre class='descdef mq'>
+	Name: dynamic-range
+	Value: standard | high
+	For: @media
+	Type: discrete
+	</pre>
+</h3>
+
+'@media/dynamic-range' represents the combination of brightness and color
```
chcunningham: I would change to:

'@media/dynamic-range' represents the combination of max brightness, color depth, and contrast ratio that are supported by the UA and output device.

This makes it more consistent with the detailed definition and also more consistent with the header for color-gamut.

mediaqueries-4/Overview.bs
```diff
@@ -1695,6 +1695,160 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+
+	<pre class='descdef mq'>
+	Name: video-dynamic-range
+	Value: standard | high
+	For: @media
+	Type: discrete
+	</pre>
+</h3>
+
+'@media/video-dynamic-range' has the same values and definitions as '@media/dynamic-range',
```
chcunningham: I would change to:
'''
'@media/video-dynamic-range' represents the combination of max brightness, color depth, and contrast ratio that are supported by the UA and output device's video plane.

Supported values are the same as <a href="#dynamic-range">dynamic-range</a>.
'''

mediaqueries-4/Overview.bs
chcunningham: Now that you're referencing the values for non-prefixed dynamic-range, you can remove this whole dfn block of value definitions and just rely on the reference. This is consistent with video-color-gamut.

chcunningham: (repasting so this doesn't git hidden in the resolved thread)

Now that you're referencing the values for non-prefixed dynamic-range, you can remove this whole dfn block of value definitions and just rely on the reference. This is consistent with video-color-gamut.

mediaqueries-4/Overview.bs
vi-dot-cpp: I changed all the instances of this -- thanks.

mediaqueries-4/Overview.bs
vi-dot-cpp: >git hidden

:)

vi-dot-cpp:
Ah, thanks -- I had missed some of the more obvious edits. The latest commit should address your CR.

mediaqueries-4/Overview.bs
```diff
@@ -1733,19 +1733,18 @@ max brightness or low brightness as it relates to displays. As such, the
 determination for '@media/dynamic-range' and '@media/video-dynamic-range' will
 be defined by the user agent.
 
-<h3 id="video-prefixed-properties">
-Video Prefixed Properties scenarios</h3>
+<h3 id="video-prefixed-properties">Video Prefixed Feature</h3>
```
chcunningham: Add 's' to end of Feature (sorry, my typo). Change id to match: id="video-prefixed-features".

mediaqueries-4/Overview.bs
```diff
@@ -1778,23 +1777,11 @@ Video Dynamic Range: the 'video-dynamic-range' feature</h3>
...
-		* color depth is greater than 24 bit
-		* has a high contrast ratio
-	<dt><dfn>standard</dfn>
-	<dd>
-		If the following are true for the UA and the output device's video plane:
-		* color gamut is sRGB
-		* color depth is 24 bit or 8 bit per color component of RGB
-		* has a low contrast ratio
-</dl>
+Supported values are the same as dynamic-range.
```
chcunningham: Replace last word with 
```
<a href="#dynamic-range">dynamic-range</a> 
```
to match video-color-gamut.

vi-dot-cpp:
Thanks Chris!

gregwhitworth:
Ok, since we've gotten @chcunningham sign off I'd like to get a review from @frivoal since @svgeesus [mentioned ](https://github.com/w3c/csswg-drafts/pull/4678#issuecomment-578246852)he'd like to provide some feedback.

svgeesus:
That would be nice, but the recent changes have improved this nicely so I would also be fine with merging as-is.

gregwhitworth:
Ok, based on @svgeesus feedback I'm going to merge it in - @frivoal I'm more than happy to address any feedback you may have if you find any issues. If editorial in nature; as always feel free to do them.

frivoal:
Hi @gregwhitworth, I'm sorry for being so far behind. I do have some feedback, which I'll put it here first, but depending on responses it'll make its way into separate issues / PRs.

## procedural / editorial 
* This PR was made against mq4, but the original issue was tagged against mq5, and I do think it would be a better fit there. Also, MQ4 is in CR already, so effectively under a (soft) feature freeze.
* I'd make all the `video-*` media features be subsections of `6.7. Video Prefixed Features` (turn them from `<h3>`s into `<h4>`s): they're not full fledged sections, but just saying "video-foo is the same as foo, in except in this different context" with some clarifications, so I don't think they should have the same importance in the table of contents.

## Substantive
* For section `6.6 Determining contrast and brightness of display`, I'd do it a little differently. I think non-normative isn't exactly what we're shooting for, and "UA defined" would be a better phrasing. Also, while I agree it should be up to the UA rather than tightly defined by the spec, it still needs some degree of guidance about what the intent is. We are talking about the screens that have the ability to do "brighter than white" highlights and the like, as typically found in video, rather than screens which have a "normal" top brightness and particularly deep blacks, which could also reasonably be described as being high contrast, but aren't what we're interested in here. While leaving the specifics UA defined, there should be some text to express the intent of that definition.

* The definition of `video-width` (same thing for `video-height`) includes this:
    > For continuous media, this is the width of the viewport (as described by CSS2, section 9.1.1 [[!CSS21]]) including the size of a rendered scroll bar (if any). For paged media, this is the width of the page box (as described by CSS2, section 13.2 [[!CSS21]]).

  That text comes from the definition of `width`, but I don't think it is relevant here: I don't think there currently exists such a thing as a paginated media with a dedicated video plane (or a paginated video plane being itself paginated).
  On the other hand, we should add a mention that this is about the size of the entire video plane, not the layout-dependent portion of it that is being clipped and composited with the rest of the page.

* The definition of `video-width` (same thing for video-height) include this:
    > lengths are interpreted according to [[#units]].
 
   I suspect that means we're doing things like measuring in CSS px, not in device pixels. If so, we should be very explicit about this, as this may be counter-intuitive for video. Also, maybe we should consider actually reporting physical pixels here when there is that separate video plane (and if so, also returning 1 as the video-resolution).


----

Depending on the responses to the points above, I might file issues / PRs differently, but for now, I expect I'll probably directly commit on the master branch for the first 2 points, propose PRs for the next 2, and to open a new issue for the last one.

vi-dot-cpp:
Hi @frivoal. Thank you for your review -- your feedback loos good to me and I trust your purview here. Am I understanding correctly that you intend to make Pull Requests for these proposals?

P.S. I hope you are well.

gregwfreedman:
hi @frivoal. thanks for reviewing. i agree with your first four points.

with respect to
> lengths are interpreted according to [[#units]].

i think that was just to be consistent with other CSS features. the only intent of `video-height` and `video-width` is to determine physical pixels of the video plane.  given that, what would be your recommendation? if we change the units to physical pixels, then we probably don't need `video-resolution`.

frivoal:
follow up issues / PR filled:
* https://github.com/w3c/csswg-drafts/issues/5041
* https://github.com/w3c/csswg-drafts/issues/5042
* https://github.com/w3c/csswg-drafts/issues/5043
* https://github.com/w3c/csswg-drafts/issues/5044

Issue #4247, opened 2019-08-27
[css-scroll-anchoring-1] Can anchor node be an inline-block?


emilio:
If you go to [this test-case](https://bug1540203.bmoattachments.org/attachment.cgi?id=9088572), then scroll so that the red box is only partially visible, then click it, what should happen?

Firefox and Chrome disagree on this one, and I think Firefox is right per spec. However the user expected Chrome's behavior.

In particular:

 * Firefox chooses the text as the scroll anchor, and thus when the box grows the text keeps its position relative to the viewport.

 * Chrome seems to choose the red box as an anchor, and thus when it grows it moves the text down.

I think per spec Firefox is right. https://drafts.csswg.org/css-scroll-anchoring/#anchor-node-selection says:

> The anchor node is either a non-anonymous block box or a text node.

Where block box is https://drafts.csswg.org/css-display-3/#block-box. An inline-block element is clearly not a "block box" (its display outside is `inline`), so the next eligible anchor node is the text.

However it seems Chromium does select inline-blocks as anchor nodes?

Is this an oversight in the specification, or a bug in Chromium?

cc @skobes @eqrion @dholbert

skobes:
I think this is a spec bug; we did not intend to exclude inline-block elements from being anchor nodes.

Probably any box-shaped element should be suitable for anchoring to.

emilio:
What should the spec say about stuff like inline-blocks or inlines split across multiple lines?

emilio:
In particular, the spec says:

> The scroll anchoring bounding rect of a DOM node N is N’s scrollable overflow rectangle if N is a block box, or the bounding rect of its line boxes if N is a text node. 

Also, the spec seems messy here. N cannot be both a box and a node.

emilio:
Also, blink seems to [skip inlines](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/layout/scroll_anchor.cc?l=276&rcl=79f03ab2a29a9977b9ad444d9c29530a4273da30), how does that make sense while not skipping inline-blocks?

So, basically, the blink condition for whether something it's an acceptable anchor if I'm reading the code right is:

```cpp
!IsAnonymous() && !IsInline() && (IsText() || IsBox())
```

What does `IsBox()` really mean? what is the rect that blink uses for boxes that are not blocks? Why does the above condition make sense?

skobes:
[...]

PR #4682, opened 2020-01-18
[scroll-anchoring-1] Allow anchor nodes to be any non-atomic inline box

chrishtr:
This resolves issue #4247 .

### DIFF
--- a/css-scroll-anchoring-1/Overview.bs
+++ b/css-scroll-anchoring-1/Overview.bs
@@ -81,7 +81,7 @@ and close to the block start edge of its <a>optimal viewing region</a>.
 Note: If the user agent does not support the 'scroll-padding' property,
 the optimal viewing region of the scrolling box is equivalent to its <a>content area</a>.
 
-An anchor node can be any non-<a>atomic inline</a> <a>box</a>.
+An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
 The anchor node is always a <a>descendant</a> of the <a>scrolling box</a>.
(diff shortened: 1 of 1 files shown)


chrishtr:
WPT tests for this include:

css/css-scroll-anchoring/image-001.html
css/css-scroll-anchoring/inline-block-002.html

css-scroll-anchoring-1/Overview.bs
```diff
@@ -81,7 +81,7 @@ and close to the block start edge of its <a>optimal viewing region</a>.
 Note: If the user agent does not support the 'scroll-padding' property,
 the optimal viewing region of the scrolling box is equivalent to its <a>content area</a>.
 
-The anchor node is either a non-anonymous <a>block box</a> or a <a>text node</a>.
+An anchor node can be any non-<a>atomic inline</a> <a>box</a>.
```
emilio: I think this reads a bit weird, but it may be because I'm not a native English speaker. When I see this, it kinda wants to read "Any inline box which is non-atomic", which is exactly the opposite of what we want. I guess in the spec it would be a bit clearer because `atomic inline` is a concept on its own, but maybe this should read "`any box except <a>inline boxes</a> that are non-<a for="atomic inline">atomic</a>`" or such (whatever makes bikeshed generate the right markup)?

emilio:
(And thanks so much for taking time to update the spec!)

css-scroll-anchoring-1/Overview.bs
chrishtr: Adjusted to make it clearer.

PR #4684, opened 2020-01-18
[resize-observer] exclude ResizeObservation IDL from Reffy

foolip:
This IDL is non-normative and is excluded by class="exclude":
https://github.com/tidoust/reffy/blob/d886aca8457f85bd7c89c9daf855d27def0d635d/src/cli/extract-webidl.js#L148


foolip:
@atotic @gregwhitworth

atotic:
Great, thanks for fixing my misunderstanding of how idl's and specs interact.

foolip:
Hmm, it's still in https://drafts.csswg.org/resize-observer/#idl-index which means it's also still in reffy-reports and web-platform-tests.

@tabatkins for the index, does one have to use `class="extract"` instead of `class="exclude"`?

PR #4692, opened 2020-01-23
Revert "[scroll-anchor-1] Add scroll-anchor suppression trigger for overflow-anchor"

chrishtr:
Reverts w3c/csswg-drafts#4681

I don't think we want this after all. On change of overflow-anchor, a layout has not actually occurred, and we just need to re-compute the anchor, not suppress the whole action for the next
layout.

css/css-scroll-anchoring/subtree-exclusion.html is such a case, that would break otherwise.


chrishtr:
I don't think the spec says anything about when the algorithm is run. I think it assumes that the anchor node selection algorithm is always up to date with the latest state of the DOM.

Issue #4347, opened 2018-11-20
[scroll-animations] ScrollTimeline with unresolved time value shouldn't be considered as "inactive" in some cases


yi-gu:
By [spec](https://www.w3.org/TR/web-animations-1/#responding-to-a-newly-inactive-timeline), if the animation timeline becomes newly inactive, i.e. its current time is newly unresolved, if the previous current time is resolved, we should use the previous current time to run the animation.

However, for **scroll timeline**, we are expecting different behaviors. By [spec](https://wicg.github.io/scroll-animations/#current-time-algorithm), the current time of scroll timeline is unresolved
1. If scrollSource does not currently have a CSS layout box, or if its layout box is not a scroll container;
2. If current scroll offset is less than startScrollOffset, return an unresolved time value if fill is none or forwards
3. If current scroll offset is greater than or equal to endScrollOffset, return an unresolved time value if fill is none or backwards

For 1, applying the last current time to hold the time value is reasonable. The timeline becomes **inactive** in such case.
For 2 and 3, applying the last current time may actually force a fill-mode. Rather, the animation should be disabled in such case. i.e. timeline's time value remains **unresolved**.

In another word, we should differentiate the concepts **inactive** and **unresolved** for scroll timeline. A scroll timeline is **inactive** in case 1 and the last current time (if resolved) should be applied. In case 2 and 3, the scroll timeline's time value remains **unresolved** and the animation should be disabled.

majido:
Duplicate of #2066

Issue #4325, opened 2019-09-17
[scroll-animations] ScrollTimeline fill-mode has asymmetric interaction with animation fill modes.


flackr:
ScrollTimeline's fill mode is a bit awkward to use. See this demo here:
https://jsbin.com/haworeb/edit?html,output

The ScrollTimeline is active from scroll offset 25 to 75 to observe the asymmetric behavior:
* Before the ScrollTimeline range, if the ScrollTimeline has `fill: both` or `fill: before` the currentTime is 0 so regardless of the effect's fill, the effect has it's initial value.
* After the ScrollTimeline range, if the ScrollTimeline has `fill: both` or `fill: after` the currentTime is the effect duration which means that the effect's fill mode decides whether the last value fills.
* If ScrollTimeline has `fill: none` the effect's fill mode doesn't matter.

I think this makes the fill on ScrollTimeline difficult to use since you need to pair it with the effect. I think that as a developer it would be easier if ScrollTimeline didn't have fill, and just exceeded the time range (i.e. < 0 or > timeRange) when it was outside of the scroll range. This way I would only have to worry about the effect's fill. Do you think there are use cases for separately wanting to rely on the scroll timeline's fill?

majido:
The discussion for this issue happened mostly on #2066 (In particular see this [comment](
https://github.com/w3c/csswg-drafts/issues/2066#issuecomment-577424074)).

I will summarize my understanding of the consensus here:

- Remove ScrollTimeline.fill
- Introduce readonly attribute [timeline](https://drafts.csswg.org/web-animations-1/#animationtimeline) state in Web-Animation that can be:  {`Inactive`, `BeforeStart`, `InRange`, `AfterEnd`}
- In the case of ScrollTimeline
  - `Inactive`: No source, or non-scroller source   => timeline.currentTime is undefined.
  - `BeforeStart`:  when scroll offset < start           => timeline.currentTime is 0
  - `InRange`:       when start <= scroll offset < end => timeline.currentTime is calculated based on existing formular.
  - `AfterEnd`:       when end <= scroll offset   => timeline.currenTime is timeRange.
[...]

PR #4750, opened 2020-02-05
[scroll-animations-1] Removed fill from ScrollTimeline

JTensai:
Removed fill from ScrollTimeline in favor of returning Infinity values for before start and after end states of the timeline.

Also fixed edit link by correcting shortname. 

Related Issues:
https://github.com/w3c/csswg-drafts/issues/2066 (Discussion for change)
https://github.com/w3c/csswg-drafts/issues/4347 (Solved by this change)

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -184,6 +184,7 @@ Using the CSS markup:
   }
   #union-circle {
     animation-name: union-circle;
+    animation-fill-mode: forwards;
     animation-timeline: scroll(element(#container), vertical, "250px", "300px");
   }
   @keyframes left-circle {
@@ -214,7 +215,7 @@ if (window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
   var right = leftCircle.animate({ transform: 'translate(350px)' }, 1000);
   right.timeline = circleTimeline;
 
-  var union = unionCircle.animate({ opacity: 1 }, 1000);
+  var union = unionCircle.animate({ opacity: 1 }, { duration: 1000, fill: "forwards" });
   union.timeline = new ScrollTimeline({
     scrollSource: scrollableElement,
     startScrollOffset: '250px',
@@ -257,6 +258,7 @@ as follows:
     height: 30px;
     background: red;
     animation: progress 1s linear;
+    animation-fill-mode: forwards;
     animation-timeline: scroll(element(#body));
   }
 }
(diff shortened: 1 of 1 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -7,13 +7,14 @@ Work Status: exploring
...
 ED: https://drafts.csswg.org/scroll-animations-1/
-Shortname: scroll-animations
+Shortname: scroll-animations-1
 Abstract: Defines an API and markup for creating animations that are either
           triggered by or tied to the scroll offset of a scroll container.
 Editor: Brian Birtles <bbirtles@mozilla.com>
 Editor: Botond Ballo <botond@mozilla.com>
 Editor: Stephen McGruer <smcgruer@google.com>
 Editor: Antoine Quint <graouts@apple.com>
+Editor: Jordan Taylor <jortaylo@microsoft.com>
```
majido: FYI you don't need to be an editor to make this change.

https://github.com/w3c/csswg-drafts/blob/master/CONTRIBUTING.md#after-discussion

scroll-animations-1/Overview.bs
```diff
@@ -542,25 +508,9 @@ as follows:
 1.  Otherwise, let <var>current scroll offset</var> be the current scroll offset of {{scrollSource}}
     in the direction specified by {{orientation}}.
 
-1.  If <var>current scroll offset</var> is less than {{startScrollOffset}}, return an unresolved
-    time value if {{fill}} is <code>none</code> or <code>forwards</code>, or 0 otherwise.
+1.  If <var>current scroll offset</var> is less than {{startScrollOffset}}, return -Infinity.
```
majido: So if I understand correctly the output is not fully decided yet. I expect this to change once the discussion is resolved here:
https://github.com/w3c/csswg-drafts/issues/2066#issuecomment-580384210

birtles:
I've commented a number of times in #2066 that I'd prefer we avoid infinity values and instead use explicit states for this.

JTensai:
I have changed the returned values to 0 and time range since I believe we have settled on that at this point. I also updated the examples as requested.

birtles:
> I have changed the returned values to 0 and time range since I believe we have settled on that at this point.

Thank you!

scroll-animations-1/Overview.bs
```diff
@@ -542,25 +509,9 @@ as follows:
 1.  Otherwise, let <var>current scroll offset</var> be the current scroll offset of {{scrollSource}}
     in the direction specified by {{orientation}}.
 
-1.  If <var>current scroll offset</var> is less than {{startScrollOffset}}, return an unresolved
-    time value if {{fill}} is <code>none</code> or <code>forwards</code>, or 0 otherwise.
+1.  If <var>current scroll offset</var> is less than {{startScrollOffset}}, return 0.
```
majido: This is correct. However one thing to note is that in this situation the scroll timeline would be in the "InBefore" state which should cause the animation to be in the same state. So while timeline time is 0 but only animations with "fill:backwards" would be active.

At the moment we cannot specify this since `timeline.state` is not specified. But I think we should add a TODO here.

TODO(4750): Define what is the correct timeline state based on scroll offset.

majido: I summarized my understanding of the current plan here.

https://github.com/w3c/csswg-drafts/issues/4325#issuecomment-585295758

Issue #4338, opened 2019-06-05
CSS Syntax design considerations and alternatives 


majido:
This issue is meant to help consider alternatives and different design for scroll timeline css syntax assuming the current general working models. Note that if we drastically alter the working models of scroll-linked animations these design discussions may not apply. 

## Assumptions
[scroll-linked animation interaction model](https://wicg.github.io/scroll-animations/#scroll-linked-animations-usecase): basically, ScrollTimeline outputs time values that directly drive a regular web Animation by setting its currentTime.

ScrollTimeline properties:
 * start, end offset
 * scroll source (single element)
 * orientation, timeRange, fill mode

### A) Timeline Property 
This is the current standing syntax proposal which basically adds a new animation property that can be used to adjust each individual animations. The property takes a css function which allows various timeline props to be configures using that function parameters.

```css
  @keyframes progress {
    from {
      width: 0vw;
    }
    to {
      width: 100vw;
    }
  }
  .progress {
    animation-name: progress;
    animation-duration: 1s;
    animation-timing-function: linear;
    /* Assume the HTML element has id 'root' */
    animation-timeline: scroll(element(#root), vertical);
  }
```

Pros:
 - Simple to add and understand
- Fairly simple to implement i.e., no need to introduce a new @rule syntax.

Cons:
 - Not possible to use a single timeline instance with multiple animations (This may actually be a pro too :) )
- Jamming all various properties of the scroll timeline into a css function seems a bit awkward.

Note that w3c/scroll-animations#25 points our that the current `scroll()` function syntax need revision but that is a minor change.

### B) @Timeline rule

This is mostly based on the proposal in w3c/scroll-animations#48 but only limited to parts relevant to current interaction model. 
 
```css 
@timeline root-scroller-timeline {
  timeline-type: scroll; 
  timeline-node: selector(:root) ; /* the element that controls this timeline;
         the selector would match a single element using document.querySelector() rules */
  scroll-direction: block; /* descriptors that are specific to scroll timelines */
  scroll-offset-start: 5rem;
  scroll-offset-end: calc(100% - 5rem); /* scroll-offset-end instead of end-scroll-offset
          so that we could have a scroll-offset shorthand */ 
}
@keyframes progress {
  from {
    width: 0vw;
  }
  to {
    width: 100vw;
  }
}
.progress {
   animation-name: progress;
   animation-duration: duration(root-scroller-timeline); /* maybe also use it with calc. */
[...]

Issue #4336, opened 2019-09-17
Use more succinct names for start/end


majido:
This is a bit of bikeshedding but as I have been building demos based on current API I find that some of the property names used for scroll timeline options are unnecessary verbose. So I suggest the following changes:

- `startScrollOffset` -> `start`
- `endScrollOffset` -> `end`
- `scrollSource` -> `source` (optional)

This first two are particularly important once we also allow element-based start and end concepts (as proposed in https://github.com/WICG/scroll-animations/issues/51) in addition to vanilla scroll offsets: 

That will give you the following IDL
```webidl 
dictionary ScrollTimelineOptions {
  Element? source = null;
  ScrollDirection orientation = "block";
  DOMString start = "auto";
  DOMString end = "auto";
  (double or ScrollTimelineAutoKeyword) timeRange = "auto";
  FillMode fill = "none";
};
```

majido:
Specification has been updated. Marking this as fixed.

PR #4751, opened 2020-02-05
[scroll-animations] Update scroll timeline css syntax and properties (#4338, #4336)

majido:
Make improvements to CSS syntax based on previous CSSWG discussions. 

## New at-rule
This addresses issue #4338 .

Introduce @timeline at-rule. The rule is used to specify parameters for scroll timeline (but can later
be used for regular timelines as well).

The animation-timeline property would select the appropriate at-rule based on its name.

I have also added language to allow existing `animation-name` property to be used for this as well. This will improve ergonomics IMO. But I can drop this part and have it as a separate PR if there is concern.

## Shorter names for Scroll Timeline properties 
This addresses  #4336 

Use shorter names for properties:
 - startScrollOffset -> start
 - endScrollOffset -> end
 - scrollSource -> source (optional)

## Other changes

Minor changes to avoid bikeshed WARNING:
 - disambiguate reference to rAF
 - Reference layout cycle

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -268,8 +268,8 @@ Typically, the scroll bar provides this visual indication but
 applications may wish to hide the scroll bar for aesthetic or useability
 reasons.
 
-Using the 'animation-timeline' property, this example could be written
-as follows:
+Using the updated 'animation' shorthand that includes 'animation-timeline', 
+this example could be written as follows:
 
 <pre class='lang-css'>
 @media (prefers-reduced-motion: no-preference) {
@@ -285,14 +285,11 @@ as follows:
     width: 0px;
     height: 30px;
     background: red;
-    animation: progress 1s linear;
-    animation-fill-mode: forwards;
-    animation-timeline: progress-timeline;
+    animation: 1s linear forwards progress progress-timeline;
   }
-
-
 }
 </pre>
+
 
 If we use this API for this case, the example code will be as follow:
 
@@ -625,11 +622,19 @@ In this case, each possible value of type <<keyframes-name>> from
     {{DocumentTimeline}}, more specifically the <a>default document
     timeline</a>.
 
-Note: Notice that the behavior for the case when no timeline with given name was
-found is different for either property. This ensures backward compatibility
-since most existing animations which have 'animation-name' specified expect to
-use <a>default document timeline</a>.
+Note: Notice that the behavior for the case where no timeline with the given
+name is found is different for these two properties. This ensures backward
+compatibility because all existing time-based animations with 'animation-name'
+specified expect to use <a>default document timeline</a>.
 
+### Changes to the 'animation' shorthand property ### {#animation-shorthand}
+
+The 'animation' shorthand property syntax is updated to accept an additional
+optional <<timeline-name>>.
+
+	<dfn>&lt;single-animation></dfn> = <<time>> || <<easing-function>> || <<time>> || <<single-animation-iteration-count>> || <<single-animation-direction>> || <<single-animation-fill-mode>> || <<single-animation-play-state>> || [ none | <<keyframes-name>> ] || [ none | <<timeline-name>> ]
+
+Issue: Update css-animations spec instead of monkey-patching it here.
 
 ## The '@timeline' at-rule ## {#timeline-at-rule}
 
@@ -691,8 +696,12 @@ have the following effects:
 'scroll-source' descriptor determines the scroll timeline's {{source}}. If
 specified as an '<selector()>' the scroll timeline's {{source}} is the
 [=scroll container=] identified by the <<id-selector>>, otherwise if not
-specified or none then it is the animation target's nearest scrollable ancestor.
+specified or none then it is the the {{scrollingElement}} of the {{Document}} 
+<a lt="document associated with a window">associated</a> with the {{Window}}
+that is the <a>current global object</a>.
 
+Issue(4338): consider choosing animation target's nearest scrollable ancestor
+instead of document's scrolling Element
 
 <pre class='descdef'>
   Name: scroll-orientation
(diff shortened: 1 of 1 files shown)


birtles:
I had a quick glance at this today and it looks good so far. I wonder if I'm the best reviewer though?

majido:
@birtles thanks for the high level check! I agree that we should use your limited time more effectively so I rather you stay focused on #2066  and potential PR for that.

As I suggested offline we should share the review load for this spec across all editors and I am working to ensure we have enough active editors on this specification to help make forward progress.

@graouts or @theres-waldo would either of you be able to review this for scroll-linked animations?

scroll-animations-1/Overview.bs
```diff
@@ -589,42 +615,178 @@ Computed value: As specified
...
+  Name: scroll-source
+  For: @timeline
+  Value: selector(<<id-selector>>) | none
+  Initial: none
+</pre>
+
+'scroll-source' descriptor determines the scroll timeline's {{source}}. If
+specified as an '<selector()>' the scroll timeline's {{source}} is the
+[=scroll container=] identified by the <<id-selector>>, otherwise if not
+specified or none then it is the animation target's nearest scrollable ancestor.
```
ogerchikov: Based on [this](https://drafts.csswg.org/scroll-animations/#scrolltimeline-interface ) ScrollTimeline.source attribute is set to document.scrollingElement if no scroll source is passed in the constructor. I would expect the same behavior for javascript and CSS specification of scroll timeline.

scroll-animations-1/Overview.bs
```diff
@@ -559,22 +587,20 @@ as follows:
...
-      <code>(<var>current scroll offset</var> - {{startScrollOffset}}) / ({{endScrollOffset}} - {{startScrollOffset}}) &times; [=effective time range=]</code>
+      <code>(<var>current scroll offset</var> - {{start}}) / ({{end}} - {{start}}) &times; [=effective time range=]</code>
     </blockquote>
 
-</div>  <!-- link-for-hint="ScrollTimeline" -->
-
-
 ## The 'animation-timeline' property ## {#animation-timeline}
 
 A {{ScrollTimeline}} may be applied to a CSS Animation [[CSS3-ANIMATIONS]] using
```
ogerchikov: I think this specification should mention how/if animation-timeline participates in animation shorthand property definition.

scroll-animations-1/Overview.bs
```diff
@@ -559,22 +587,20 @@ as follows:
...
 
     <blockquote>
-      <code>(<var>current scroll offset</var> - {{startScrollOffset}}) / ({{endScrollOffset}} - {{startScrollOffset}}) &times; [=effective time range=]</code>
+      <code>(<var>current scroll offset</var> - {{start}}) / ({{end}} - {{start}}) &times; [=effective time range=]</code>
     </blockquote>
 
-</div>  <!-- link-for-hint="ScrollTimeline" -->
-
-
 ## The 'animation-timeline' property ## {#animation-timeline}
```
ogerchikov: Perhaps not for this PR, but I would expect animation-timeline specification to live in CSS-ANIMATIONS spec ([css-animations-2](https://drafts.csswg.org/css-animations-2/)?)

scroll-animations-1/Overview.bs
```diff
@@ -589,42 +615,178 @@ Computed value: As specified
...
+  Initial: time
+</pre>
 
-    The <<single-animation-fill-mode>> value, if specified, determines the timeline's {{fill}}.
+The 'timeline-type' determines the type of timeline and whose possible values
+have the following effects:
+
+:   time
+::  The animation's [=timeline=] is a {{DocumentTimeline}}, more specifically
+    the <a>default document timeline</a>.
```
ogerchikov: Any reason to limit to the default document timeline not any document timeline?

scroll-animations-1/Overview.bs
majido: I am not sure if that is necessary. Unless we decide to move Scroll Timeline into web-animations as well. Otherwise keeping the css syntax and its JS equivalent next to each other makes more sense to me.

scroll-animations-1/Overview.bs
majido: This is the [behavior of the animation constructor](https://drafts.csswg.org/web-animations/#ref-for-dom-animation-animation) and I think we should match that.

Beside we have to choose  *a* timeline for auto. It is not clear to me why anything other than "default document timeline" makes sense.

scroll-animations-1/Overview.bs
majido: Good point. Added a section to have this property participate in shorthand. At the moment this is kept in this spec but I expect it to become part of css-animations-2

scroll-animations-1/Overview.bs
majido: Good point. There was a note in CSSWG group discussion to use nearest ancestor. But you are right
about consistency. For now I changed it to use scrollingElement same as Javascript.

Added an issue note to consider using nearest ancestor in this case. We can revisit this in future.

scroll-animations-1/Overview.bs
ogerchikov: From what I read the only time based timeline in this spec allowed is the default document timeline. Web Animations API  allows instantiating document timelines and attaching animations to them.  Is it intentionally that we are limiting CSS animations to be attached to arbitrary document timeline?

majido: You are correct that timeline can be change in Web Animations API but not in CSS Animations.
I am not intending to change the time-based CSS Animation capabilities that in this patch thus the current wording.
But I believe the timeline syntax here is flexible enough to enable this and bring CSS Animation in parity with WebAnimation in this regard.

As for why the current limitation in CSS Animations, I can only guess:
  - There is very limited value to having a non-default timeline at the moment.
 - There was not CSS Syntax specified for timelines.

Given these two I can imagine why no one bothered allowing custom timelines in CSS Animations. This patch will rectify #2 but does not just add custom time-based timeline in CSS Animations.

majido: I will merge this PR as is. Happy to make modifications if my answer is not accurate or there is a better way to word this.

Issue #4576, opened 2019-12-09
[css-text] Atomic inlines being equivalent to ID for line breaking is not web-compatible


kojiishi:
The current CSS Text [Line Breaking Details] says:

> The line breaking behavior of a replaced element or other atomic inline is equivalent to an ideographic character (Unicode linebreaking class ID [UAX14]), and additionally, for Web-compatibility, introduces a soft wrap opportunity between itself and any adjacent U+00A0 NO-BREAK SPACE character.

We've got a report saying this is not web-compatible. Tested that all browsers always break before and after atomic inlines.

Is this an intentional change that all other browsers are willing to match?

@fantasai @litherum @emilio @jfkthame

[Line Breaking Details]: https://drafts.csswg.org/css-text-3/#line-break-details

frivoal:
My understanding was that "equivalent to ID" is considered the more useful behavior in theory, but that it alone is not be web compatible. However, Equivalent to ID **and** breaks between adjacent U+00A0 was expected to be web compatible. Have you found other cases where the specified behavior would cause breakage?

kojiishi:
Two cases were reported; atomic inline + ':' U+003A COLON and atomic inline + ';' U+003B SEMICOLON. In both cases, authors expect there's a break opportunity between them.

kojiishi:
This issue was reported at
https://crbug.com/1025715: No breaking opportunity between SVG and ':' character

litherum:
I guess the question is: Is there a use case for wanting strict line-breaking in a paragraph, but still wanting breaking opportunities on both sides of images? I think yes, these two things are fairly distinct, and it's entirely reasonable for an author to want this combination of behaviors.

litherum:
I suppose another option is to add a new value to the [`wrap-before` and `wrap-after` properties](https://drafts.csswg.org/css-text-4/#wrap-before). Maybe something like `wrap-before: ideograph` or something.

frivoal:
> I guess the question is: Is there a use case for wanting strict line-breaking in a paragraph, but still wanting breaking opportunities on both sides of images?

In that case, you could use

p { word-break: strict; }
img { word-break: auto; }

frivoal:
I've made the edit corresponding to the resolution, and landed the corresponding test updates. Closing this issue, as that solves the original problem reported.

[...]

PR #4755, opened 2020-02-07
[css-text-3] replaced elements and atomic inlines with non default line-break

frivoal:
This is a follow up to #4576

This makes replaced elements and atomic inlines with non default values for `line-break` behave the same as characters with line-breaking class ID for line breaking purposes. This gives the expected behavior for images used as “gaiji” and as emoji.

This cannot be the default behavior for compat reasons, but having it as an opt-in based on the rarely used line-break property should be safe, and opens up for further tweaking of the line breaking behavior via the usual set of properties.

### DIFF
--- a/css-text-3/Overview.bs
+++ b/css-text-3/Overview.bs
@@ -2395,6 +2395,12 @@ Line Breaking Details</h3>
       line-breaking/line-breaking-atomic-nowrap-001.html
       </wpt>
 
+      Note: Once an [=atomic inline=] is treated as an ideographic charcter
+      due to a non ''line-break/auto'' value of 'line-break',
+      properties such as 'word-break'
+      that change how line breaking works with respect ideographic characters
+      also affect the [=atomic inline=].
+
     <li>For <a>soft wrap opportunities</a> created by characters that disappear at the line break (e.g. U+0020 SPACE),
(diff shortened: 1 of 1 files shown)


kojiishi:
I would like us to use any other properties than `line-break` for this purpose.

The effect of using incorrect values of `line-break` is visible only to some scripts. It is possible for authors to misunderstand and apply, for instance:
```css
html { line-break: strict; }
```
but authors will not notice it has undesired side effects until they test it with CJK content. This can lead to a situation where global sites to use undesired typography only for CJK content.

/cc @r12a @himorin

fantasai:
@kojiishi The initial value of `line-break` right now is `auto`, so non-CJK sites can use `line-break: normal` to make the change in behavior. Also I think, it's easy to say "put it on the image", `img { line-break: normal; }`, then it won't affect any other characters. Maybe we can make this an example so it's clearer.

frivoal:
@fantasai, added an example: https://github.com/w3c/csswg-drafts/pull/4755/commits/34c697ba3304b2c6fe016e01c09313db914af967. Feedback welcome.

frivoal:
Closed as the resolution in https://github.com/w3c/csswg-drafts/issues/4949#issuecomment-624705207 decided on a different approach.

PR #4769, opened 2020-02-12
[css-fonts-4] Clean up `font-min-size` and `font-max-size`

Malvoz:
The properties were removed in https://github.com/w3c/csswg-drafts/commit/ebca02f2de74aadd45f675ed794cd686a2852d87.


Malvoz:

PR #4786, opened 2020-02-18
[css-text-3] Adding missing word "element"

captainbrosset:
Just a minor update, it seems to me like the word "element" is missing from the sentence:

> [...] the rules for determining the content language of an HTML are defined in [...]


tabatkins:
thanks for the catch!

PR #4806, opened 2020-02-25
[scroll-animations-1] Add majidvp as editor

majido:
Adding myself (majidvp@google.com) as editor to the spec.
Also moving Stephen (smcgruer@google.com) to former editor section.


majido:
It is the id that identifies your account in w3c: https://tabatkins.github.io/bikeshed/#metadata

I think CSSWG asked these to be added at some point by I cannot find where it was. For most spec it was auto-added but I think this is before I migrated this spec to csswg. Other specs have it see for example [this](https://github.com/w3c/csswg-drafts/blob/master/css-animations-1/Overview.bs).

I will  add other editors w3cid in a separate PR.

PR #4835, opened 2020-03-06
[css-inline-3] text-orientation: sideways-left is not a thing

litherum:
https://drafts.csswg.org/css-writing-modes-3/#text-orientation

### DIFF
--- a/css-inline-3/Overview.bs
+++ b/css-inline-3/Overview.bs
@@ -192,7 +192,7 @@ Dominant Baselines: the 'dominant-baseline' property</h3>
 		<dd>
 			Equivalent to ''dominant-baseline/alphabetic'' in <a>horizontal writing modes</a>
 			and in <a>vertical writing modes</a>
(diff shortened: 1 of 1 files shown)


css-inline-3/Overview.bs
```diff
@@ -192,7 +192,7 @@ Dominant Baselines: the 'dominant-baseline' property</h3>
 		<dd>
 			Equivalent to ''dominant-baseline/alphabetic'' in <a>horizontal writing modes</a>
 			and in <a>vertical writing modes</a>
-			when 'text-orientation' is ''sideways'', ''sideways-right'', or ''sideways-left''.
+			when 'text-orientation' is ''sideways'', or ''sideways-right''.
```
kojiishi: `sideways-right` is not a thing either, so just `sideways`?

kojiishi: nvm, I found it's still there "if needed for backward compatibility reasons".

Issue #2080, opened 2017-12-05
[web-animations-1] Remove "responding to a newly inactive timeline" section?


birtles:
See: https://drafts.csswg.org/web-animations-1/#responding-to-a-newly-inactive-timeline

Is it needed in this level of the spec?

birtles:
Fixed in #4842

Issue #2075, opened 2017-12-05
[web-animations-1] Don't preserve current time for scroll linked animations when changing playbackRate


birtles:
_From @stephenmcgruer on November 3, 2017 14:28_

According to the spec, current time is calculated as:

`current time = (timeline time - start time) × playback rate`

This makes sense for DocumentTimeline, where we want to provide an offset from the start of the animation, but it doesn't seem to make sense for a ScrollTimeline (https://wicg.github.io/scroll-animations/#scrolltimeline-interface). Consider the following case:

```
let effect = new KeyframeEffect(
    foo,
    [ { width: "0vw" }, { width: "100vw" } ],
    { duration: 1000, easing: "linear" });
let timeline = new ScrollTimeline({
    scrollSource: document.documentElement,
    orientation: "block",
    range: "auto"
});
let anim1 = new Animation(effect, timeline);
anim1.play();

document.documentElement.scrollTop = "50vh";
let anim2 = new Animation(effect, timeline);
anim2.play();

// According to the spec:
anim1.currentTime == 500
anim2.currentTime == 0
```

The start time is set from the timeline time. This means that if the user manages to scroll the `scrollSource` before the animation setup script runs (e.g. perhaps because the script has not yet loaded over the network), the animation will be offset by whatever value the user has already scrolled - which I don't think is what the developer would want.

A first thought (courtesy of majidvp@) is to differentiate between montonic (use start offset) and non-monotonic (don't use start offset) timelines. DocumentTimeline is obviously monotonic, whereas ScrollTimeline isn't.

_Copied from original issue: w3c/web-animations#204_

birtles:
#174 is possibly related.

majido:
Any other ideas on how this can be fixed beside treating these timeline differently? 
+ @theres-waldo for visibility since this involves scroll timeline.

FYI, the link above is no longer correct after migration it should be #2066.

majido:
So far, for our implementation of ScrollTimeline in Chrome we have treated scroll linked animations as it they always have a set Zero start time (i.e., as if they always start at 0). 

However this breaks some other behaviors in web-animations specifically where *startTime* is adjusted during seeking. The main two cases that I know about are: 
1. when currentTime of a playing animation is set via animation.currentTime.
2. when playbackRate is changed, in order to maintain the previous currentTime and prevents a jump in the animation output.

[...]

PR #4842, opened 2020-03-06
[web-animations-1] Initialize start time of scroll animations to zero (#2075)

ogerchikov:
Initializing start time of scroll animations to zero is required to have animation effect progress in sync with the timeline progress.

Changes made:
* Removed responding to a newly inactive timeline section (#2080).
* Updated playing/pausing animation procedures to initialize start time to zero for finite timelines. This makes animation current time equal to scroll timeline current time or unresolved if the timeline is inactive.
* Updated pausing animation procedure to execute animation pending pause task when inactive timeline becomes active. This makes ready time resolved at the time of the execution and, as a result, produces resolved hold_time.
* Updated definition of idle play state to include check for unresolved start time. This is required so
playing animations attached to inactive timeline return "running" state. Without this change play pending animations attached to inactive timeline return "running" state, while playing animations attached to inactive timeline return "idle" state.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -1917,11 +1917,18 @@ non-normative description is also provided:
 
 :   <a lt="idle play state">idle</a>
 ::  The <a>current time</a> of the animation is <a>unresolved</a> and
+    the <a>start time</a> of the animation is <a>unresolved</a> and
     there are no pending tasks.
     In this state the animation has no effect.
 :   <a lt="running play state">running</a>
 ::  The animation has a resolved <a>current time</a> that changes on each
-    <a>animation frame</a> (provided the [=playback rate=] is not zero).
+    <a>animation frame</a> (provided the [=playback rate=] is not zero
+    and the <a>timeline</a> is [=monotonically increasing=]). If the
+    <a>timeline</a> is not [=monotonically increasing=], the
+    <a>current time</a> may not change on each <a>animation frame</a> or
+    is unresolved if the <a>timeline</a> is
+    <a lt="inactive timeline">inactive</a>. In the latter case
+    the animation has no effect.
 :   <a lt="paused play state">paused</a>
 ::  The animation has been suspended and the <a>current time</a>
     is no longer changing.
@@ -1939,6 +1946,8 @@ condition from the following:
 
 :   <em>All</em> of the following conditions are true:
     *   The <a>current time</a> of <var>animation</var> is <a>unresolved</a>,
+        <em>and</em>
+    *   the <a>start time</a> of <var>animation</var> is <a>unresolved</a>,
         <em>and</em>
     *   <var>animation</var> does <em>not</em> have <em>either</em>
         a <a>pending play task</a> <em>or</em> a <a>pending pause task</a>,
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -1127,7 +1127,8 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
     <div class="switch">
     :   If |animation|'s <a>current time</a> is <a>unresolved</a>,
         <a>start time</a> is <a>unresolved</a>, the associated <a>timeline</a>
-        is finite and the <var>auto-rewind</var> flag is true
+        is not [=monotonically increasing=] and the <var>auto-rewind</var> flag
+        is true
     ::  Set <var>animation</var>'s <a>start time</a> to zero.
     :   If |animation|'s [=effective playback rate=] &gt; 0,
         the <var>auto-rewind</var> flag is true and <em>either</em>
@@ -1304,7 +1305,7 @@ follows:
 
     <div class="switch">
     :   If |animation|'s <a>start time</a> is <a>unresolved</a> and the
-        associated <a>timeline</a> is finite
+        associated <a>timeline</a> is not [=monotonically increasing=]
     ::  Let <var>animation</var>'s <a>start time</a> be zero.
 
     :    If <var>animation</var>'s [=playback rate=] is &ge; 0,
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -1126,12 +1126,8 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
 
     <div class="switch">
     :   If |animation|'s <a>current time</a> is <a>unresolved</a>,
-        <a>start time</a> is <a>unresolved</a>, the <var>auto-rewind</var>
-        flag is true and <em>either</em>:
-        *   the |animation| has no associated <a>timeline</a>, or
-        *   the associated <a>timeline</a> is
-            <a lt="inactive timeline">inactive</a>
-
+        <a>start time</a> is <a>unresolved</a>, the associated <a>timeline</a>
+        is finite and the <var>auto-rewind</var> flag is true
     ::  Set <var>animation</var>'s <a>start time</a> to zero.
     :   If |animation|'s [=effective playback rate=] &gt; 0,
         the <var>auto-rewind</var> flag is true and <em>either</em>
@@ -1307,11 +1303,8 @@ follows:
     perform the steps according to the first matching condition from below:
 
     <div class="switch">
-    :   If |animation|'s <a>start time</a> is <a>unresolved</a> and
-        <em>either</em>:
-        *   the |animation| has no associated <a>timeline</a>, or
-        *   the associated <a>timeline</a> is
-            <a lt="inactive timeline">inactive</a>
+    :   If |animation|'s <a>start time</a> is <a>unresolved</a> and the
+        associated <a>timeline</a> is finite
     ::  Let <var>animation</var>'s <a>start time</a> be zero.
 
     :    If <var>animation</var>'s [=playback rate=] is &ge; 0,
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -1121,8 +1121,9 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
     <var>animation</var> has a <a>pending pause task</a>, and false otherwise.
 1.  Let <var>has pending ready promise</var> be a boolean flag that is
     initially false.
-1.  Let <var>start or hold time updated</var> be a boolean flag that is
-    initially false.
+1.  Let <var>performed seek</var> be a boolean flag that is initially false.
+1.  Let <var>has finite timeline</var> be true if |animation| has an associated
+    <a>timeline</a> that is not [=monotonically increasing=].
 1.  Perform the steps corresponding to the <em>first</em> matching
     condition from the following, if any:
 
@@ -1136,16 +1137,18 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
         *   <a>current time</a> &lt; zero, or
         *   <a>current time</a> &ge; <a>associated effect end</a>,
 
-    ::  1.  Set <var>start or hold time updated</var> to true.
-        1.  Update either <var>animation</var>’s <a>hold time</a> or
-            <a>start time</a> as follows:
+    ::  1.  Set <var>performed seek</var> to true.
+        1.  Update either <var>animation</var>’s <a>start time</a> or
+            <a>hold time</a> as follows:
+
             <div class="switch">
 
-            :   If |animation| has no associated <a>timeline</a> or the
-                associated <a>timeline</a> is [=monotonically increasing=],
-            ::  Set <var>animation</var>'s <a>hold time</a> to zero.
-            :   Otherwise,
+            :   If |has finite timeline| is true,
             ::  Set <var>animation</var>'s <a>start time</a> to zero.
+            :   Otherwise,
+            ::  Set <var>animation</var>'s <a>hold time</a> to zero.
+
+            </div>
 
     :   If |animation|'s [=effective playback rate=] &lt; 0,
         the <var>auto-rewind</var> flag is true and <em>either</em>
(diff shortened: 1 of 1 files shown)


web-animations-1/Overview.bs
```diff
@@ -1150,7 +1125,14 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
     condition from the following, if any:
 
     <div class="switch">
-
+    :   If |animation|'s <a>current time</a> is <a>unresolved</a>,
+        <a>start time</a> is <a>unresolved</a>, the <var>auto-rewind</var>
+        flag is true and <em>either</em>:
+        *   the |animation| has no associated <a>timeline</a>, or
+        *   the associated <a>timeline</a> is
+            <a lt="inactive timeline">inactive</a>
```
birtles: I'm afraid I don't quite understand the reasoning behind these conditions. I thought we discussed setting the start time to zero for scroll-based timelines? But they won't necessarily have no timeline, right?

web-animations-1/Overview.bs
```diff
@@ -1185,9 +1167,10 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
     1.  Cancel that task.
     1.  Set <var>has pending ready promise</var> to true.
 
-1.  If the following three conditions are <em>all</em> satisfied:
+1.  If the following four conditions are <em>all</em> satisfied:
 
     * |animation|'s [=hold time=] is [=unresolved=], and
+    * |animation|'s [=start time=] is [=unresolved=], and
```
birtles: I'm not quite sure I follow this but I think it probably relates back to my [comment here](https://github.com/w3c/csswg-drafts/issues/2066#issuecomment-598018685).

web-animations-1/Overview.bs
```diff
@@ -1324,6 +1307,12 @@ follows:
     perform the steps according to the first matching condition from below:
 
     <div class="switch">
+    :   If |animation|'s <a>start time</a> is <a>unresolved</a> and
+        <em>either</em>:
+        *   the |animation| has no associated <a>timeline</a>, or
+        *   the associated <a>timeline</a> is
+            <a lt="inactive timeline">inactive</a>
```
birtles: As with my first comment, I'm afraid I don't follow why we have these particular conditions.

web-animations-1/Overview.bs
ogerchikov: Setting start time to zero and hold time to null produces unresolved current time. This is to ensure no animation effect is applied when the timeline is inactive.

web-animations-1/Overview.bs
ogerchikov: The explanation is the same as in the first comment.

web-animations-1/Overview.bs
ogerchikov: Just to be clear regarding your comment about zero start time initialization for scroll timeline, I am planning to make this change later, but as part of a separate PR. The goal of this PR is just handling inactive/null timelines.

birtles: I see. So the tricky part is making an inactive timeline produce an unresolved current time in the animation whilst also still allowing animations without a timeline to be seeked to a particular time.

This is complicated because we might have Web compat issues here given that Chrome and Firefox both ship the `Animation()` constructor so the behavior of calling `play()` on animations without a timeline is already observable.

That said, we probably don't commonly encounter inactive timelines, so I wonder if we can make this change simply apply to inactive timelines?

Or is it a bit odd that calling `play()` on an animation without a timeline behaves differently from calling `play()` on an animation attached to an inactive timeline? From my understanding, both would make the animation "running" and pending, but in the first case the `currentTime` would be zero, but in the second case the `currentTime` would be `null`.

ogerchikov: With #4899 proposal, if accepted, play/pause procedures will be updated to initialize start time to zero for both active and inactive finite timelines. Event without #4899, we probably should be initializing start time to 0 for finite timeline linked animations regardless of the timeline activeness.

Having said this, we need to decide which behavior we expect from animations without a timeline. In #2066 the "half" agreement was to treat a null timeline as inactive timeline, meaning animations don't expose animation effect. If this indeed the desired behavior, perhaps we can find a way to phaseout already shipped behavior in favor of the desired one.

ogerchikov:
I repurposed this change to initialize start time of scroll animations to 0 instead of handling inactive timelines.

web-animations-1/Overview.bs
ogerchikov: Changed to initialize start time of finite timeline animations to 0. No changes are made for the animations that don't have a timeline. Note, in this change we initialize start time to 0 regardless of the timeline activeness. It achieves the desired effect when the timeline is active (anim.currentTime==timeline.currentTime) and when the timeline is inactive (animation.currentTime = null).

web-animations-1/Overview.bs
ogerchikov: This is required to avoid early exit if start time is set.

web-animations-1/Overview.bs
ogerchikov: Please see the first comment.

web-animations-1/Overview.bs
```diff
@@ -1150,7 +1125,10 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
     condition from the following, if any:
 
     <div class="switch">
-
+    :   If |animation|'s <a>current time</a> is <a>unresolved</a>,
+        <a>start time</a> is <a>unresolved</a>, the associated <a>timeline</a>
+        is finite and the <var>auto-rewind</var> flag is true
```
ogerchikov: Definition of finite timeline depends on #4862 and will be updated/referenced when the corresponding PR is merged.

ogerchikov:
@majido @birtles - can you please take a look at these changes? I believe we are in agreement that start time of scroll animations should be initialized to 0.

web-animations-1/Overview.bs
```diff
@@ -1150,7 +1125,11 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
     condition from the following, if any:
 
     <div class="switch">
-
+    :   If |animation|'s <a>current time</a> is <a>unresolved</a>,
+        <a>start time</a> is <a>unresolved</a>, the associated <a>timeline</a>
+        is not [=monotonically increasing=] and the <var>auto-rewind</var> flag
```
birtles, line 1130: This needs to clarify what happens when there is no associated timeline.

Looking at this more, I feel a little unsure about this abstraction. If the animation has a negative playback rate, it seems like we should seek to the end of the animation. Maybe we should keep the conditions as-is, but alternative between setting the hold time or start time based on the type of timeline? Would that work?

web-animations-1/Overview.bs
birtles: I'm not sure about this. Won't this mean that calling `play()` on an already running animation makes it become pending? (Since the hold time is unresolved in that case but the start time is not -- previous we would abort the procedure in that case but now we won't.)

web-animations-1/Overview.bs
```diff
@@ -1324,6 +1304,9 @@ follows:
     perform the steps according to the first matching condition from below:
 
     <div class="switch">
+    :   If |animation|'s <a>start time</a> is <a>unresolved</a> and the
+        associated <a>timeline</a> is not [=monotonically increasing=]
+    ::  Let <var>animation</var>'s <a>start time</a> be zero.
```
birtles, line 1309: (Presumably whatever we decide for the play procedure will also apply here.)

web-animations-1/Overview.bs
birtles: We could probably fix the above by simply adding some extra flag to this procedure to indicate if we made a change that means that we need to wait for a pending task to run. This might be a more elegant way, however.

ogerchikov: Added the flag.

web-animations-1/Overview.bs
ogerchikov: Changed per your suggestion.

web-animations-1/Overview.bs
ogerchikov: Changed similarly to Play.

ogerchikov:
@birtles - thanks for your feedback! All your comments are addressed, please take a look. Note, I intentionally don't change  null timelines behaviors. I am happy to do so but not in this PR.

web-animations-1/Overview.bs
```diff
@@ -1930,11 +1966,18 @@ non-normative description is also provided:
...
 :   <a lt="running play state">running</a>
 ::  The animation has a resolved <a>current time</a> that changes on each
-    <a>animation frame</a> (provided the [=playback rate=] is not zero).
+    <a>animation frame</a> (provided the [=playback rate=] is not zero
+    and the <a>timeline</a> is [=monotonically increasing=]). If the
+    <a>timeline</a> is not [=monotonically increasing=], the
+    <a>current time</a> may not change on each <a>animation frame</a> or
+    is unresolved if the <a>timeline</a> is
+    <a lt="inactive timeline">inactive</a>. In the latter case
+    the animation has no effect.
```
birtles, line 1980: Perhaps this could be simplified to something like...

> The animation has a resolved current time that changes on each animation frame (provided the playback rate is not zero and the animation has an associated timeline that is active and monotonically increasing).

web-animations-1/Overview.bs
```diff
@@ -1325,16 +1336,37 @@ follows:
...
 
-    :    If <var>animation</var>'s [=playback rate=] is &ge; 0,
-    ::   Let <var>animation</var>'s <a>hold time</a> be zero.
+    :   If <var>animation</var>'s [=playback rate=] is &ge; 0,
+    ::  Update either <var>animation</var>’s <a>hold time</a> or
+        <a>start time</a> as follows:
+        <div class="switch">
+
+        :   If |animation| has no associated <a>timeline</a> or the
+            associated <a>timeline</a> is [=monotonically increasing=],
```
birtles, line 1345: As with the play procedure, it might be worth introducing a local variable to avoid repeating this condition twice?

web-animations-1/Overview.bs
```diff
@@ -1953,6 +1996,8 @@ condition from the following:
 :   <em>All</em> of the following conditions are true:
     *   The <a>current time</a> of <var>animation</var> is <a>unresolved</a>,
         <em>and</em>
+    *   the <a>start time</a> of <var>animation</var> is <a>unresolved</a>,
+        <em>and</em>
```
birtles, line 2000: I'm a little bit nervous about changing this part since I'm afraid it might have unintended consequences that I haven't been able to anticipate. However, if we can make this change and none of the existing WPT break, then I guess it's ok.

ogerchikov: Without this change, having current time unresolved automatically means start time being unresolved. That's how I convinced myself it should not be any backward compatibility problems.

web-animations-1/Overview.bs
```diff
@@ -1159,23 +1136,56 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
         *   <a>current time</a> &lt; zero, or
         *   <a>current time</a> &ge; <a>associated effect end</a>,
 
-    ::  Set <var>animation</var>'s <a>hold time</a> to zero.
+    ::  1.  Set <var>start or hold time updated</var> to true.
+        1.  Update either <var>animation</var>’s <a>hold time</a> or
+            <a>start time</a> as follows:
+            <div class="switch">
```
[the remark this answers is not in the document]
ogerchikov: Yesterday bikeshed complained about indents on closing divs so I just removed them. Today I finally figured out how to close - it must be an empty line before closing div.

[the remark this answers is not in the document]
ogerchikov: Fixed.

web-animations-1/Overview.bs
```diff
@@ -1159,23 +1136,56 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
...
+
+            :   If |animation| has no associated <a>timeline</a> or the
+                associated <a>timeline</a> is [=monotonically increasing=],
```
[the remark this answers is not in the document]
ogerchikov: Fixed.

web-animations-1/Overview.bs
```diff
@@ -1146,6 +1121,8 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
     <var>animation</var> has a <a>pending pause task</a>, and false otherwise.
 1.  Let <var>has pending ready promise</var> be a boolean flag that is
     initially false.
+1.  Let <var>start or hold time updated</var> be a boolean flag that is
+    initially false.
```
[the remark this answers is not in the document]
ogerchikov: Fixed.

web-animations-1/Overview.bs
ogerchikov: Changed per your suggestion. However I still feel that we are not explaining behavior of animations attached to active/inactive finite timelines in terms of current time update and having effect.

web-animations-1/Overview.bs
ogerchikov: FIxed.

ogerchikov:
@birtles - thanks again for reviewing and providing great feedback. All your comments are addressed.

web-animations-1/Overview.bs
birtles: > 
> 
> Without this change, having current time unresolved automatically means start time being unresolved. That's how I convinced myself it should not be any backward compatibility problems.

Is that true? If you have no timeline or an inactive timeline, I think you can still have a resolved start time with an unresolved current time?

web-animations-1/Overview.bs
birtles: Yes, that's true. But I think this is non-normative text so perhaps it doesn't need to describe everything?

birtles:
Looks great to me. I'll be interested to hear your answer to the one question I added but feel free to merge this (or do you not have access to do it?).

web-animations-1/Overview.bs
ogerchikov: After you mentioned null timeline, the case I can think of is when running animation.timeline is set to null. However in Chrome animation.timeline is read only, so hopefully user scripts don't rely on this behavior.

In regards to inactive timelines, I don't think DocumentTimeline is expected to become newly inactive.  Even if it does become inactive, the spec says we should use previous resolved time. "Responding to newly inactive timelines" is not implemented in Chrome, so, again, hopefully user scripts don't rely on this behavior.

I think the major impact of this change is that we are changing the mental model of animation play state. Before, 'idle'/'running' implied activeness of a timeline the animation is attached to, while now 'idle' play state can only be achieved by user actions - cancelling a running animation or not calling play/pause.

birtles: > After you mentioned null timeline, the case I can think of is when running animation.timeline is set to null. However in Chrome animation.timeline is read only, so hopefully user scripts don't rely on this behavior.

I think Chrome is shipping the `Animation()` constructor, however, which allows creating animations without a timeline so I think it's possible it's already being relied on. It does seem unlikely, however.

PR #4854, opened 2020-03-10
[web-animations-1] Added phase to Animation Timeline #4325 

JTensai:
Adding phase to Animation Timeline requires changes to multiple spec's let me know if they can be done at once in the same PR or if they need to be split into individual PR's.

I'm sure there are some places I missed that need to reference the new timeline phase. If you can think of any please comment and let me know so I can adjust the spec as needed.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -621,13 +621,6 @@ purpose of synchronization.
 At any given moment, a [=timeline=] has a single current [=time value=] known
 simply as the timeline's <dfn lt="timeline current time">current time</dfn>.
 
-A <a>timeline</a> may not always be able to return a meaningful <a>time
-value</a>, but only an <a>unresolved</a> time value. For example, it may
-be defined relative to a moment that has yet to occur, such as the firing of
-a document's load event.
-A <a>timeline</a> is considered to be <dfn lt="inactive timeline">inactive</dfn>
-when its <a>time value</a> is <a>unresolved</a>.
-
 A <a>timeline</a> is <dfn export lt="monotonically increasing timeline"
 local-lt="monotonically increasing">monotonically increasing</dfn> if its
 reported [=timeline current time|current time=] is always greater than or equal
@@ -708,6 +701,22 @@ animation model to change, but these operations in themselves do not create
 a new [=animation frame=], rather they merely update the current <a>animation
 frame</a>.
 
+### Timeline Phase ### {#timeline-phases}
+
+A [=timeline=] may be in one of four possible 
+<dfn export lt="timeline phase">phases</dfn>:
+
+1.  <dfn export lt="timeline inactive phase">inactive</dfn>
+1.  <dfn export lt="timeline before phase">before</dfn>
+1.  <dfn export lt="timeline active phase">active</dfn>
+1.  <dfn export lt="timeline after phase">after</dfn>
+
+<div class="informative-bg"><em>This section is non-normative</em>
+
+[=timeline phase|Timeline phase=] is used to give additional context to 
+[=timeline current time=] when it would otherwise be ambiguous.
+
+</div>
 ### Document timelines ### {#document-timelines}
 
 A <dfn>document timeline</dfn> is a type of <a>timeline</a> that is <a
@@ -721,15 +730,29 @@ time</dfn>.
 Issue(2079): There must be a better term than &ldquo;origin time&rdquo;&mdash;
              it's too similar to &ldquo;time origin&rdquo;.
 
-Prior to establishing the [=time origin=] for its associated document,
-a [=document timeline=] is <a lt="inactive timeline">inactive</a>.
+
+The [=timeline phase|phase=] of a <a>document timeline</a> at a given moment is
+the [=timeline phase|phase=] corresponding to the <em>first</em> matching
+condition from the following:
+
+<div class="switch">
+
+:   If the associated document is not an [=active document=] or its
+    [=time origin=] has not yet been established
+::  &rarr; [=timeline inactive phase|inactive=]
+:   Otherwise,
+::  &rarr; [=timeline active phase|active=]
+
+</div>
+
+A [=document timeline=] in the [=timeline inactive phase|inactive phase=] can
+also be called an <dfn>inactive timeline</dfn>
+
+An [=inactive timeline=] has an [=unresolved=]
+[=timeline current time|current time=].
 
 After a [=document timeline=] becomes [=inactive timeline|active=], it is
 [=monotonically increasing=].
-
-A <a>document timeline</a> that is associated with a {{Document}} which is not
-an <a>active document</a> is also considered to be
-<a lt="inactive timeline">inactive</a>.
 
 To <a lt="timeline time to origin-relative time">convert a timeline
 time, |timeline time|, to an origin-relative time</a> for a document timeline,
@@ -796,7 +819,7 @@ An animation's start time is initially <a>unresolved</a>.
 
 An <a>animation</a> also maintains a <dfn>hold time</dfn> <a>time value</a>
 which is used to fix the animation's output <a>time value</a>, called its
-<a>current time</a>, in circumstances such as pausing</a>.
+<a>current time</a>, in circumstances such as pausing.
 The <a>hold time</a> is initially <a>unresolved</a>.
 
 In order to establish the relative ordering of conflicting <a>animations</a>,
@@ -3926,6 +3949,7 @@ The <code>AnimationTimeline</code> interface {#the-animationtimeline-interface}
 [Exposed=Window]
 interface AnimationTimeline {
     readonly attribute double? currentTime;
+    readonly attribute TimelinePhase phase;
 };
 </pre>
 
@@ -3935,6 +3959,8 @@ interface AnimationTimeline {
 ::  Returns the <a lt="timeline current time">current time</a> for this timeline
     or <code>null</code> if this timeline is <a lt="inactive
     timeline">inactive</a>.
+:   <dfn attribute for=AnimationTimeline>phase</dfn>
+::  Returns the <a lt="timeline phase">phase</a> for this timeline.
 
 </div>
 
@@ -4301,6 +4327,21 @@ enum AnimationReplaceState { "active", "removed", "persisted" };
 ::  Corresponds to the [=removed replace state=].
 :   <code>persisted</code>
 ::  Corresponds to the [=persisted replace state=].
+
+### The <code>TimelinePhase</code> enumeration ### {#the-timelinephase-enumeration}
+
+<pre class='idl'>
+enum TimelinePhase { "inactive", "before", "active", "after" };
+</pre>
+
+:   <code>inactive</code>
+::  Corresponds to the [=timeline inactive phase=].
+:   <code>before</code>
+::  Corresponds to the [=timeline before phase=].
+:   <code>active</code>
+::  Corresponds to the [=timeline active phase=].
+:   <code>after</code>
+::  Corresponds to the [=timeline after phase=].
 
 The <code>AnimationEffect</code> interface {#the-animationeffect-interface}
 ----------------------------------------
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -711,12 +711,12 @@ A [=timeline=] may be in one of four possible
 1.  <dfn export lt="timeline active phase">active</dfn>
 1.  <dfn export lt="timeline after phase">after</dfn>
 
-<div class="informative-bg"><em>This section is non-normative</em>
+A [=timeline=] in the [=timeline inactive phase|inactive phase=] can
+also be called an <dfn>inactive timeline</dfn>
 
-[=timeline phase|Timeline phase=] is used to give additional context to 
-[=timeline current time=] when it would otherwise be ambiguous.
+An [=inactive timeline=] has an [=unresolved=]
+[=timeline current time|current time=].
 
-</div>
 ### Document timelines ### {#document-timelines}
 
 A <dfn>document timeline</dfn> is a type of <a>timeline</a> that is <a
@@ -744,12 +744,6 @@ condition from the following:
 ::  &rarr; [=timeline active phase|active=]
 
 </div>
-
-A [=document timeline=] in the [=timeline inactive phase|inactive phase=] can
-also be called an <dfn>inactive timeline</dfn>
-
-An [=inactive timeline=] has an [=unresolved=]
-[=timeline current time|current time=].
 
 After a [=document timeline=] becomes [=inactive timeline|active=], it is
 [=monotonically increasing=].


web-animations-1/Overview.bs
```diff
@@ -628,6 +628,21 @@ a document's load event.
 A <a>timeline</a> is considered to be <dfn lt="inactive timeline">inactive</dfn>
 when its <a>time value</a> is <a>unresolved</a>.
 
+A [=timeline=] may be in one of four possible <dfn>phase</dfn>s:
+
+* <dfn lt="inactive phase">inactive</dfn>
+* <dfn lt="before phase">before</dfn>
+* <dfn lt="active phase">active</dfn>
+* <dfn lt="after phase">after</dfn>
```
birtles: This should probably be integrated with the previous paragraph (e.g. introduce the phases and then say that when it's inactive there's no meaningful time value).

web-animations-1/Overview.bs
```diff
@@ -628,6 +628,21 @@ a document's load event.
...
+
+Calculating which phase a timeline is in is handled by each type of [=timeline=]
```
birtles: I don't think we need this sentence. But we probably do need to say that document timelines are always active.

web-animations-1/Overview.bs
```diff
@@ -628,6 +628,21 @@ a document's load event.
 A <a>timeline</a> is considered to be <dfn lt="inactive timeline">inactive</dfn>
 when its <a>time value</a> is <a>unresolved</a>.
 
+A [=timeline=] may be in one of four possible <dfn>phase</dfn>s:
```
birtles: Would it be better to scope phase to something like "timeline phase"?

e.g. `<dfn lt="timeline phase">phase</dfn>`.

Then you can link to it as `[=timeline phase=]` or `[=timeline phase|phase=]` for when repeating "timeline" is unnecessary. (And if that becomes too onerous we can add `local-lt="phase"`.)

web-animations-1/Overview.bs
```diff
@@ -2296,12 +2311,14 @@ definitions:
...
 animation effect's <a>local time</a> is not <a>unresolved</a> and
-<em>either</em> of the following conditions are met:
+<em>any</em> of the following conditions are met:
 
 1.  the <a>local time</a> is less than the <a>before-active boundary time</a>,
     <em>or</em>
 1.  the <a>animation direction</a> is &lsquo;backwards&rsquo; and the <a>local
     time</a> is equal to the <a>before-active boundary time</a>.
+1.  the <a>animation timeline</a> <a>phase</a> corresponding to this effect
+    is "before"
```
birtles: This should use the ["associated with a timeline"](https://drafts.csswg.org/web-animations-1/#associated-with-a-timeline) concept. It needs a period too. And the previous bullet point needs to drop the period and add `, <em>or</em>`.

web-animations-1/Overview.bs
```diff
@@ -4302,7 +4320,23 @@ enum AnimationReplaceState { "active", "removed", "persisted" };
...
+::  Corresponds to the [=inactive timeline phase=].
+:   <code>before</code>
+::  Corresponds to the [=before timeline phase=].
+:   <code>active</code>
+::  Corresponds to the [=active timeline phase=].
+:   <code>after</code>
+::  Corresponds to the [=after timeline phase=].
+
 The <code>AnimationEffect</code> interface {#the-animationeffect-interface}
+
```
birtles: Just curious, is this blank line needed?

ogerchikov:
Should hold_time concept be update to include hold_phase? What if an animation is paused at timeline.currentTime of zero and 'before' phase?

web-animations-1/Overview.bs
JTensai: According to the spec, there are cases where a document timeline can be inactive:
"A document timeline that is associated with a Document which is not an active document is also considered to be inactive."
I have removed the unnecessary sentence though. I will update the spec in regards to Document Timeline to explain how it determines what phase it is in.

JTensai:
> Should hold_time concept be update to include hold_phase? What if an animation is paused at timeline.currentTime of zero and 'before' phase?

Good catch, I will add the concept of hold_phase in the next commit.

web-animations-1/Overview.bs
```diff
@@ -708,6 +701,32 @@ animation model to change, but these operations in themselves do not create
...
+:   <dfn export lt="timeline active phase">active</dfn>
+::  The timeline current time is after its start point and before its end point.
+:   <dfn export lt="timeline after phase">after</dfn>
+::  The timeline current time is at or beyond its end point.
+
+</div>
+
+[=timeline phase|Timeline phase=] is used to give additional context to 
+[=timeline current time|current time=]. For example, if a [=timeline=] has a
+delayed start, [=timeline current time|current time=] of 0 would be ambiguous
```
birtles: What is a delayed start?

web-animations-1/Overview.bs
```diff
@@ -731,6 +750,19 @@ A <a>document timeline</a> that is associated with a {{Document}} which is not
...
+The [=timeline phase|phase=] of a <a>document timeline</a> at a given moment is
+the [=timeline phase|phase=] corresponding to the <em>first</em> matching
+condition from the following:
+
+<div class="switch">
+
+:   If the <a>document timeline</a> is <a lt="inactive timeline">inactive</a>:
+::  &rarr; [=timeline inactive phase|inactive=]
+:   Otherwise,
+::  &rarr; [=timeline active phase|active=]
```
birtles: I'm probably missing something but this seems like a circular reference? Can't we just say that it's active except for in such-and-such a situation where it's inactive?

web-animations-1/Overview.bs
```diff
@@ -2296,12 +2328,21 @@ definitions:
...
+    time</a> is equal to the <a>before-active boundary time</a>,
+    <em>or</em>
+1.  the effect is <a>associated with a timeline</a> <em>and</em> the associated
+    <a>timeline</a>'s [=timeline phase|phase=] is
+    [=timeline before phase|before=]
+
+
+
+
+
```
birtles: This whitespace seems unnecessary.

web-animations-1/Overview.bs
```diff
@@ -3926,6 +3967,7 @@ The <code>AnimationTimeline</code> interface {#the-animationtimeline-interface}
 [Exposed=Window]
 interface AnimationTimeline {
     readonly attribute double? currentTime;
+    readonly attribute TimelinePhase phase;
```
birtles: There should probably be a description of the member somewhere that refers back to the definition in the model.

web-animations-1/Overview.bs
JTensai: Added example: "a ScrollTimeline startScrollOffset greater than 0"

web-animations-1/Overview.bs
JTensai: It's not quite circular but I understand the confusion. "Inactive" is defined differently for each type of timeline. "Inactive" is defined for DocumentTimeline above:

"Prior to establishing the time origin for its associated document, a document timeline is inactive."
"A document timeline that is associated with a document which is not an active document is also considered to be inactive"

So depending on the document, timeline can be inactive, and based on whether or not the timeline is active we set the phase.

A lot of that was existing, should I change it around to make it less confusing?

JTensai:
Added concept of hold phase. Ready for review again.

web-animations-1/Overview.bs
```diff
@@ -708,6 +701,35 @@ animation model to change, but these operations in themselves do not create
...
+    </dfn> when its <a>time value</a> is <a>unresolved</a>. This behavior is
+    further defined by each type of timeline.
+:   <dfn export lt="timeline before phase">before</dfn>
+::  The timeline current time is before its start point.
+:   <dfn export lt="timeline active phase">active</dfn>
+::  The timeline current time is after its start point and before its end point.
+    Note: There is a special case for ScrollTimeline that can make the end point
+    inclusive.
+:   <dfn export lt="timeline after phase">after</dfn>
+::  The timeline current time is at or beyond its end point.
```
birtles, line 723: I don't think this makes sense to have these as exported definitions that are non-normative. I think if we are linking to these we should at least just list them in normative text. We can then follow up with a non-normative section describing them in more detail.

web-animations-1/Overview.bs
```diff
@@ -708,6 +701,35 @@ animation model to change, but these operations in themselves do not create
...
+
+</div>
+
+[=timeline phase|Timeline phase=] is used to give additional context to 
+[=timeline current time|current time=]. For example, if a [=timeline=] has a
+delayed start (e.g. a ScrollTimeline startScrollOffset greater than 0),
+[=timeline current time|current time=] of 0 would be ambiguous since it could be
+any amount of time before the start point or it could be at the start time.
```
birtles, line 731: I think this probably belongs in the non-normative section. And it's probably best to avoid referring to specific interface members of the scroll-timeline specification (since that will change). Instead I'd just describe it in slightly more generic terms.

web-animations-1/Overview.bs
```diff
@@ -731,6 +753,19 @@ A <a>document timeline</a> that is associated with a {{Document}} which is not
...
 <a lt="inactive timeline">inactive</a>.
 
+The [=timeline phase|phase=] of a <a>document timeline</a> at a given moment is
+the [=timeline phase|phase=] corresponding to the <em>first</em> matching
+condition from the following:
+
+<div class="switch">
+
+:   If the <a>document timeline</a> is <a lt="inactive timeline">inactive</a>:
+::  &rarr; [=timeline inactive phase|inactive=]
```
birtles, line 763: This also seems odd to me. I don't think there are any circumstances defined in this spec for a document timeline to be inactive so we should just say it is always active. When we find circumstances where that's not the case we can list them as exceptions (e.g. XHR docs etc.).

web-animations-1/Overview.bs
```diff
@@ -796,9 +831,22 @@ An animation's start time is initially <a>unresolved</a>.
 
 An <a>animation</a> also maintains a <dfn>hold time</dfn> <a>time value</a>
 which is used to fix the animation's output <a>time value</a>, called its
-<a>current time</a>, in circumstances such as pausing</a>.
+<a>current time</a>, in circumstances such as pausing.
 The <a>hold time</a> is initially <a>unresolved</a>.
 
+An <a>animation</a> also maintains a <dfn>hold phase</dfn>
+[=timeline phase|phase=] which is used by <a>animation effect</a> to calculate
```
birtles, line 838: This currently reads:

"An animation also maintains a hold phase phase" (← notice the double "phase").

web-animations-1/Overview.bs
```diff
@@ -796,9 +831,22 @@ An animation's start time is initially <a>unresolved</a>.
...
+its correct phase in circumstances such as pausing. The <a>hold phase</a> is
```
birtles, line 839: This should probably link to the animation effect phase section.

web-animations-1/Overview.bs
```diff
@@ -796,9 +831,22 @@ An animation's start time is initially <a>unresolved</a>.
...
+initially <a>unresolved</a>. Whenever <a>hold time</a> is changed,
+<a>hold phase</a> should be updated as follows:
```
birtles, line 841: Nit: In general we should start a new line for each sentences. It makes subsequent diffs smaller.

Also, "should" has a special meaning in specs as something that can be ignored. Instead it would be better to simply make it descriptive:

"Whenever an animation's hold time is changed, its hold phase is updated as follows:"

web-animations-1/Overview.bs
```diff
@@ -2296,12 +2344,16 @@ definitions:
...
 
 1.  the <a>local time</a> is less than the <a>before-active boundary time</a>,
     <em>or</em>
 1.  the <a>animation direction</a> is &lsquo;backwards&rsquo; and the <a>local
-    time</a> is equal to the <a>before-active boundary time</a>.
+    time</a> is equal to the <a>before-active boundary time</a>,
+    <em>or</em>
+1.  the effect is <a>associated with a timeline</a> <em>and</em> the associated
+    <a>timeline</a>'s [=timeline phase|phase=] is
+    [=timeline before phase|before=]
```
birtles, line 2356: Two questions:

* Is this the behavior we want? It effectively means you can't seek an animation associated with a timeline in the before/after phase which seems wrong?
* Do we need to do something similar for the after phase?

web-animations-1/Overview.bs
```diff
@@ -708,6 +701,35 @@ animation model to change, but these operations in themselves do not create
...
+A [=timeline=] may be in one of four possible 
+<dfn export lt="timeline phase">phases</dfn> for each of which, a non-normative
+description is also provided:
+
+<div class=informative-bg>
+
+:   <dfn export lt="timeline inactive phase">inactive</dfn>
+::  A <a>timeline</a> is considered to be <dfn lt="inactive timeline">inactive
+    </dfn> when its <a>time value</a> is <a>unresolved</a>. This behavior is
+    further defined by each type of timeline.
```
birtles, line 715: This still feels really odd to me. It feels like it should be the other way around. When the timeline is inactive it will produce an unresolved time (and then we define the circumstances that make it inactive).

web-animations-1/Overview.bs
JTensai: Done, I moved the requirements for being inactive to the phase calculation for document timeline.

web-animations-1/Overview.bs
JTensai: Going to add this concept in a later PR that covers the effects in more detail.

web-animations-1/Overview.bs
JTensai: Removing holde phase concept for now. Going to add this concept in a later PR that covers the effects in more detail.

web-animations-1/Overview.bs
JTensai: Removing for now.

web-animations-1/Overview.bs
JTensai: Deferring hold phase until a later PR.

web-animations-1/Overview.bs
JTensai: Done, I moved the requirements for being inactive to the phase calculation for document timeline.

web-animations-1/Overview.bs
JTensai: Deferring adding the details of how phase affects effects until a later PR.

JTensai:
I messed some things up trying to do a rebase onto a more recent master. I ended up losing the commit history of this PR, but it shouldn't matter much. Feedback has been addressed and is ready for review again :)

web-animations-1/Overview.bs
```diff
@@ -721,16 +730,30 @@ time</dfn>.
...
+:   If the associated document is not an [=active document=] or its
+    [=time origin=] has not yet been established
+::  &rarr; [=timeline inactive phase|inactive=]
+:   Otherwise,
+::  &rarr; [=timeline active phase|active=]
+
+</div>
+
+A [=document timeline=] in the [=timeline inactive phase|inactive phase=] can
+also be called an <dfn>inactive timeline</dfn>
```
birtles, line 749: Is this definition supposed to be limited to document timelines? I think it applies to all types of timelines, right?

web-animations-1/Overview.bs
```diff
@@ -708,6 +701,22 @@ animation model to change, but these operations in themselves do not create
...
+
+1.  <dfn export lt="timeline inactive phase">inactive</dfn>
+1.  <dfn export lt="timeline before phase">before</dfn>
+1.  <dfn export lt="timeline active phase">active</dfn>
+1.  <dfn export lt="timeline after phase">after</dfn>
+
+<div class="informative-bg"><em>This section is non-normative</em>
+
+[=timeline phase|Timeline phase=] is used to give additional context to 
+[=timeline current time=] when it would otherwise be ambiguous.
```
birtles, line 717: This doesn't really explain anything. We should probably give an example.

web-animations-1/Overview.bs
JTensai: Great catch, moving this up to the timeline phase section.

web-animations-1/Overview.bs
JTensai: Removing this section until a later time when we can reference scroll timeline phases as an example.

PR #4859, opened 2020-03-11
[css-transforms-1] Fix typo and simplify transforms syntax

mclegrand:
`wsp*` is already a prefix of `comma-wsp` and thus is useless here

Non-substantive change


dontcallmedom:

PR #4873, opened 2020-03-13
[web-animations-1] Mark ready and finished promise rejections as handled

### LINKED ISSUE #4556: [web-animations-1] Should ready / finished rejections be marked as handled?
In https://github.com/web-platform-tests/wpt/issues/18932 there was some discussion about unhandled rejections for the `.ready` promise and I wasn't sure what the correct behavior should be.

This morning I discovered https://github.com/whatwg/streams/issues/547 where for streams we decided to explicitly mark rejections for "state promises" like `.ready` as handled. I wonder if we should be doing the same for `.ready` and `.finished` for the `Animation` interface?

birtles:
Fixes #4556.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -889,6 +889,8 @@ The procedure to <dfn>reset an animation's pending tasks</dfn> for
 1.  [=Apply any pending playback rate=] on |animation|.
 1.  <a lt="reject a Promise">Reject</a> <var>animation</var>'s <a>current ready
     promise</a> with a DOMException named "AbortError".
+1.  Set the \[[PromiseIsHandled]] internal slot of |animation|'s
+    <a>current ready promise</a> to true.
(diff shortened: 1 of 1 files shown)


graouts:
Can we write tests for this? I don't think this state is inspectable using DOM APIs, or is it?

flackr:
I think just ensuring the ready promise is resolved without error on cancellation or setting another effect is sufficient to test this.

birtles:
> Can we write tests for this? I don't think this state is inspectable using DOM APIs, or is it?

See https://github.com/w3c/csswg-drafts/issues/4556#issuecomment-598588755

i.e. https://hg.mozilla.org/try/rev/54bc69142ded40d5a5e5b8fef909bfd3fe1dab3e#l6.12

We can test it by registering for unhandledrejection events and checking they are _not_ thrown.

PR #4876, opened 2020-03-15
[meta] Update ISSUE_TEMPLATE to explain spec shortnames

AmeliaBR:
Add some more guidance for new contributors, who might not understand what the spec shortnames are.


Psoffie009:
> Add some more guidance for new contributors, who might not understand what the spec shortnames are.

pswan004:
> Add some more guidance for new contributors, who might not understand what the spec shortnames are.

PR #4884, opened 2020-03-17
Define terms in CSS2 in a Bikeshed compatible way

### LINKED ISSUE #4880: Migrate CSS2 prop/value/type dfn to Bikeshed/Shepherd syntax
Shepherd currently has [magic support](http://hg.csswg.org/dev/specification/file/4a110a36c1cc/python/specification/specificationparser.py#l998) for how CSS2 is marked up; this will break when we republish as we'll no longer be using `a[name]` for them.

While we could just modify Shepherd here, for propdefs it's essentially no work (changing the build system), and for the others it's probably worthwhile in case we want to move CSS2 to BS in future.

gsnedders:
Fixes #4880.

(Tab, not expecting you to do a detailed review, but please do look quickly at the "add data-dfn-type" commits and check the *.html files look sane!)

### DIFF
--- a/css2/about.html
+++ b/css2/about.html
@@ -192,7 +192,7 @@ information that resembles the following:</p>
 
 <div class="propdef">
 <dl><dt>
-<span class="index-def" title="'property-name'"><dfn id="propdef-property-name" class="propdef-title" data-dfn-type="property" data-lt="property-name"><strong>'property-name'</strong></dfn></span>
+<span class="index-def" data-term="'property-name'"><dfn id="propdef-property-name" class="propdef-title" data-dfn-type="property" data-lt="property-name"><strong>'property-name'</strong></dfn></span>
 <dd>
 <table class="propinfo def">
 <tr><td><em>Value:</em>&nbsp;&nbsp;<td>legal values &amp; syntax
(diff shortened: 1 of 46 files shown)


tabatkins:
Sorry, but a bunch of the properties/values/etc have single quotes in their link text; we'll need to either remove those or put in a `data-lt` with the plain name. (Shepherd's heuristics remove those, but Bikeshed takes it literally.)

Minor nit, but the existing `<strong>`s were really just there to make things bold, not for any semantic purpose. W3C's CSS already bolds `<dfn>`s, so those aren't necessary even for styling any longer.

gsnedders:
> Sorry, but a bunch of the properties/values/etc have single quotes in their link text; we'll need to either remove those or put in a `data-lt` with the plain name. (Shepherd's heuristics remove those, but Bikeshed takes it literally.)

Ah, that's what I get for only looking at what Shepherd does!

> Minor nit, but the existing `<strong>`s were really just there to make things bold, not for any semantic purpose. W3C's CSS already bolds `<dfn>`s, so those aren't necessary even for styling any longer.

There's plenty of redundant stuff like that, on the whole feel like that easier to cleanup manually and leaving that for now at least.

tabatkins:
> There's plenty of redundant stuff like that, on the whole feel like that easier to cleanup manually and leaving that for now at least.

Yeah, that's def fine.

PR #4889, opened 2020-03-19
[resize-observer-1] Fix IDL dictionary default syntax

domenic:
Part of https://github.com/heycam/webidl/issues/758.


tabatkins:
This was fixed as part of the void->undefined switcharoo.

Issue #4862, opened 2020-03-11
[scroll-animations] Support progress-based animations on finite timelines


flackr:
# Problem

Currently, when you write an animation based on a ScrollTimeline, you have to provide an arbitrary animation duration, e.g.:
```js
document.getElementById('parallax').animate(
    { transform: ['translateY(0)', 'translateY(100px)']},
    { duration: 10000, // Totally arbitrary!
      fill: 'both',
      timeline: new ScrollTimeline({
          endScrollOffset: '200px'})
    });
```

By default, ScrollTimeline's `auto` value for [effective scroll range](https://drafts.csswg.org/scroll-animations/#efffective-time-range-algorithm) will assume the range of the animation. This means that whatever duration is supplied, scroling from (in this case) 0px to 200px will advance the animation from 0 to 100%. Even worse, having too small of a value entered for duration can result in precision problems (see #4353) due to the [microsecond precision on animation timers](https://www.w3.org/TR/web-animations-1/#precision-of-time-values).

# Proposal

Allow omitting the duration for an animation driven by finite timeline.

Concretely:
* Introduce a finite timeline type - whose range must not be infinite.
* If the animation timeline is finite (i.e. ScrollTimeline is finite, DocumentTimeline is not as it has no end value), duration may be omitted for new animations.
* When duration is omitted, the duration will be set to the end value for the timeline.
  * If the timeline has timeRange `auto`, the animation and timeline will have a duration / timeRange of 100 (i.e. 100%).

For the above example, this would make the following valid:
```js
document.getElementById('parallax').animate(
    { transform: ['translateY(0)', 'translateY(100px)']},
    { fill: 'both',
      timeline: new ScrollTimeline({
          endScrollOffset: '200px'})
    });
```

And would result in a duration and timeRange of 100 (percent).

# Open questions

* Is there a better way to express the non-time based progress fallback? i.e. For ScrollTimeline we could use pixels.
* Are the precision requirements on non time based animations different than time based animations? If not, the default time range for progress animations needs to be large enough to be precise.

Note: This is technically a change of the web-animations, but the reasons for it are due to finite timelines such as ScrollTimeline.

@birtles @graouts @majido WDYT? I can put together a prototype of this in my [polyfill](https://github.com/flackr/scroll-timeline) and include other examples if it would be helpful.

flackr:
[...]

PR #4890, opened 2020-03-20
[web-animations-1] [scroll-animations-1] Support automatic duration scroll animations #4862

flackr:
Change the implicit handling of auto duration animations to use the duration of the supplied timeline, giving ScrollTimeline an implicit duration of 100s corresponding to 100% of their progress.

I think this is about the minimum that could be done to support the use case in #4862 and also doesn't significantly complicate web animations. While I like the idea of supporting arbitrary percentages specified in the future, it's easy to see how the specified duration 'auto' implicitly maps to 100% of the duration.

The only change to existing behavior is that no supplied animation duration would result in an infinite duration (from the DocumentTimeline) rather than 0 duration as defined today.


scroll-animations-1/Overview.bs
```diff
@@ -385,12 +383,14 @@ interface ScrollTimeline : AnimationTimeline {
   readonly attribute ScrollDirection orientation;
   readonly attribute DOMString startScrollOffset;
   readonly attribute DOMString endScrollOffset;
-  readonly attribute (double or ScrollTimelineAutoKeyword) timeRange;
 };
 </pre>
 
 A <dfn>scroll timeline</dfn> is an {{AnimationTimeline}} whose time values are determined
 not by wall-clock time, but by the progress of scrolling in a [=scroll container=].
+The {{AnimationTimeline/duration}} of a <a>scroll timeline</a> is 100s to provide
```
flackr: It might be nice if this were still a percentage but the definition of [time values](https://www.w3.org/TR/web-animations-1/#time-value-section) already suggests the correspondence to milliseconds is weak.

scroll-animations-1/Overview.bs
```diff
@@ -465,40 +465,6 @@ not by wall-clock time, but by the progress of scrolling in a [=scroll container
     :   <<percentage>>
     ::  A percentage distance along {{scrollSource}}'s scroll range in {{orientation}}.
 
-:   <dfn attribute for=ScrollTimeline>timeRange</dfn>
```
flackr: Another option would be keeping timeRange and just having auto map to the default 100s value. In practice I expect most people to use the automatic duration / timeRange though and so would suggest that we keep this as an option for later if needed.

majido: I rather remove the value for now.

web-animations-1/Overview.bs
```diff
@@ -4518,8 +4527,9 @@ dictionary OptionalEffectTiming {
     time taken to complete a single iteration of the <a>animation
     effect</a>.
 
-    In this level of this specification, the string value <code>auto</code>
-    is treated as the value zero for the purpose of timing model calculations
+    The string value <code>auto</code> is treated as the
```
majido: Aww, GH does not allow me to comment on other lines that is not in your change 👎 

But a few lines below there is 
```
This is a forwards-compatiblity measure since a future level of this
    specification is expected to introduce group effects where the
    <code>auto</code> value expands to include the duration of the child
    effects.

````

With your change where auto is taken to mean timeline's duration, is it possible to have the flexibility to interpret it differently?
e.g., prefer's child effect effect durations if timeline duration is infinite?

Comment (flackr) [reply web-animations-1/Overview.bs:4530]: My thinking was actually if this was a group effect auto would look at the children's timing, add up all of the non-auto children timings and distribute the remaining time on the timeline between the auto children. i.e.
Having children of durations [auto, 12s, 8, auto] on a 100s ScrollTimeline would turn into [40s, 12s, 8s, 40s]

In this manner it can always be seen to be looking at the children, but if there are no children it is distributing remaining time from the timeline.

File: scroll-animations-1/Overview.bs (hunk 2 above)
Comment (flackr) [reply scroll-animations-1/Overview.bs:468]: Agreed, I think it's easier to add the property back if it proves to still be necessary than to have a property which we may have difficulty removing if it becomes unergonomic - i.e. if we decide that percentages have different treatment and ScrollTimeline should only output percentages.

**Comment by flackr** [comment]
> This looks very reasonable to me. I am surprised how little change it required.
> 
> > The only change to existing behavior is that no supplied animation duration would result in an infinite duration (from the DocumentTimeline) rather than 0 duration as defined today.
> 
> What is the web-compat impact of this?

It's actually very minimal, and could be removed if we felt necessary. Previously the value `auto` for EffectTiming duration would be treated as 0, and with this proposal as is, it would now be infinite. This means if you didn't specify a duration for an animation it would hold its 0 value forever instead of having no effect. I would expect that 'auto' duration animations are extremely rare as it's not currently a meaningful value.

As I was suggesting, we could make an exemption that `auto` for infinite duration timelines still resolves to 0, which would make it completely backwards compatible with the existing web surface.

File: web-animations-1/Overview.bs (hunk 3 above)
Comment (birtles) [reply web-animations-1/Overview.bs:4530]: I added a few comments similar to this in #4907. It would be nice to get a consistent approach that covers all these three cases: scroll timelines, group effects, and keyframe effects with time-based keyframe offsets.

File: web-animations-1/Overview.bs [hunk 4]
```diff
@@ -726,6 +729,8 @@ is calculated as a fixed offset from the |now| timestamp provided each time the
 [=update animations and send events=] procedure is run.
 This fixed offset is referred to as the document timeline's <dfn>origin
 time</dfn>.
+The <a lt="timeline duration">duration</a> of a [=document timeline=> is
```
birtles, line 732: Markup error here: `[=document timeline=>`

scroll-animations-1/Overview.bs
```diff
@@ -631,60 +610,13 @@ if (window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
...
-    [=target effect end=] of infinity, the [=effective time range=]
-    is zero.
-
-:   Otherwise,
-::  The [=effective time range=] is the {{ScrollTimeline}}'s
-    {{timeRange}}.
-
-</div>
-
-### The effective scroll range of a {{ScrollTimeline}} ### {#effective-scroll-range-algorithm}
```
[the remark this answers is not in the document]
birtles: Yes, I think it is still used in the current time calculation.

web-animations-1/Overview.bs
birtles: Also, maybe a new paragraph would be better?

web-animations-1/Overview.bs
```diff
@@ -620,6 +620,9 @@ purpose of synchronization.
 
 At any given moment, a [=timeline=] has a single current [=time value=] known
 simply as the timeline's <dfn lt="timeline current time">current time</dfn>.
+The <dfn lt="timeline duration">duration</a> of a timeline gives the
+maximum value a timeline may generate for
+[=timeline current time|current time=].
```
birtles, line 625: Perhaps we could make this a new paragraph and explain a little what it's purpose is?

web-animations-1/Overview.bs
```diff
@@ -3974,6 +3979,7 @@ The <code>AnimationTimeline</code> interface {#the-animationtimeline-interface}
 [Exposed=Window]
 interface AnimationTimeline {
     readonly attribute double? currentTime;
+    readonly attribute double? duration;
```
birtles, line 3982: I guess this needs to be `unrestricted double` in order to represent infinity?

web-animations-1/Overview.bs
```diff
@@ -4412,10 +4421,8 @@ apart from the timing model.
         {{AnimationEffect/getComputedTiming()}} must return a number
         corresponding to the calculated value of the <a>iteration duration</a>
         as defined in the description of the {{EffectTiming/duration}} member of
-        the {{EffectTiming}} interface.
-
-        In this level of the specification, that simply means that an
-        <code>auto</code> value is replaced by zero.
+        the {{EffectTiming}} interface. A percentage value is represented as its
+        double value (i.e. 100% returns 1.0).
```
birtles, line 4425: Doesn't this lose information? i.e. it doesn't roundtrip and it's not possible to tell whether 1.0 represents a percent or a millisecond time?

web-animations-1/Overview.bs
```diff
@@ -4505,12 +4512,20 @@ dictionary OptionalEffectTiming {
     the number of milliseconds from the [=start time=] of the associated
     <a>animation</a> to the start of the <a>active interval</a>.
 
+    If the <a>iteration duration</a> resolves to a percentage, the start
```
birtles, line 4515: (Nit: What does "resolves to" mean here?)

web-animations-1/Overview.bs
```diff
@@ -4580,8 +4595,16 @@ dictionary OptionalEffectTiming {
     time taken to complete a single iteration of the <a>animation
     effect</a>.
 
-    In this level of this specification, the string value <code>auto</code>
-    is treated as the value zero for the purpose of timing model calculations
+    If the <a lt="timeline duration">duration</dfn> of the Animation's
+    <a>timeline</a> is a percentage, a non <code>auto</code> duration will
+    be treated as 0. Future levels of this specification are expected to
+    introduce support for scaling duration, iterations and start/end delays
+    to fill a percentage timeline duration.
```
birtles, line 4602: We probably should update the paragraph before this, and then describe the syntax for percentages (and how they are encoded as strings).

Also, it seems like the duration of a timeline can't be a percentage?

web-animations-1/Overview.bs
```diff
@@ -4580,8 +4595,16 @@ dictionary OptionalEffectTiming {
...
+
+    The string value <code>auto</code> is treated as the
+    <a lt="timeline duration">duration</dfn> of the Animation's <a>timeline</a>
+    (zero if the timeline is null) divided by the <a>iteration count</a>
+    for the purpose of timing model calculations
```
birtles, line 4607: Isn't this a breaking change? Won't `auto' now resolve to infinity for cases where it used to resolve to 0?

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
flackr: Good catch, this was a badly resolved merge conflict.

web-animations-1/Overview.bs
flackr: I switched this to having the document timeline (i.e. non-finite timelines) have a null duration leading to a 0-time but let me know if we should switch this back.

web-animations-1/Overview.bs
flackr: It's expected that getComputedTiming can lose information though, right? I thought we had settled on the progress being a double in the [0-1] range in our previous discussions.

web-animations-1/Overview.bs
flackr: I switched the wording on this to is treated as which matches the wording for the iteration duration when "auto" is specified.

web-animations-1/Overview.bs
flackr: Hmm, my thinking was that internally we would recognize that the timeline duration is a percentage even though the value by inspection would be 1.0 (consistent with getComputedStyle). However, I'm guessing the issue is that if it's not represented in the WebIDL return value we can't recognize this.

Some options:
* Define and use a subclass of AnimationTimeline (i.e. ProgressTimeline) and check the instance type to determine which behavior to apply.
* Use a string value for duration

Any other ideas? Any recommendations?

web-animations-1/Overview.bs
flackr: This is calculated to be 0 now.

web-animations-1/Overview.bs
birtles: Oh good point. I think it's probably helpful to ask what we want `getTiming()` and `getComputedTiming()` to return when we later introduce effects that are a percentage of their parent group's time (or ancestor timeline if we introduced fixed duration document timelines for example).

I think in that case we might want something like:

* `getTiming().duration` to return [`CSSNumberish`](https://drafts.css-houdini.org/css-typed-om-1/#typedefdef-cssnumberish), i.e. a `double` in the case of a fixed time (for compatibility) or a `CSSNumberValue` to represent a percentage.
* `getComputedTiming().duration` to return a number which is either the resolved duration in ms (in part for compatibility, and in part so we can determine the actual duration for complex group effects) or a value between [0, 1] or [0, 100] to represent a percentage of an auto duration timeline.

What do you think?

flackr: I like this, if we do the same for the timeline duration then we can know when a defined duration on the timeline is a time (where time-based values in the animations don't need to be dropped) or a percentage (where we convert time based animations to their respective proportions). I'll work on making this change.

flackr: Actually we already specify that the return value from getTiming() must be auto if the duration is specified as auto so we should be returning the string 'auto' from getTiming which will round trip. I still like the idea of using CSSNumberish for the timeline duration though so that we clearly know whether the duration is time-based or progress (percentage) based.

scroll-animations-1/Overview.bs
```diff
@@ -357,13 +355,13 @@ interface ScrollTimeline : AnimationTimeline {
...
   readonly attribute DOMString end;
-  readonly attribute (double or ScrollTimelineAutoKeyword) timeRange;
 };
 </pre>
 
 A <dfn>scroll timeline</dfn> is an {{AnimationTimeline}} whose time values are
 determined not by wall-clock time, but by the progress of scrolling in a
-[=scroll container=].
+[=scroll container=]. The {{AnimationTimeline/duration}} of a <a>scroll timeline</a> is
+100%.
```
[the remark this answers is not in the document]
flackr: Done.

web-animations-1/Overview.bs
birtles: Yes, sorry I was getting a bit confused because I forgot we are only trying to cover "auto" in this patch.

(My primary concern is making sure we have a clear path for supporting percentages in general in future so I am trying to think about how we will cover that.)

flackr: We could consider returning CSSNumberish from computed timing - that would sidestep the issue of converting the percentages to a double value. It will mean changing a lot of values to be CSSNumberish, but this might be cleaner in the long run. WDYT?

birtles: I'm not really sure. Once we are able to use percentages that resolve to times (e.g. `duration` is 50% of my parent group) I wonder if it would be useful to have `getComputedTiming()` return the actual number of seconds it resolves to? In which case, should percentages that resolve against scroll timelines resolve to pixels?

Unfortunately `getComputedStyle` doesn't give an entirely clear precedent here since it sometimes resolves percentages (e.g. for `width` and `height`) but otherwise doesn't. I guess the intention is that it should always return percentages. So maybe we should return CSSNumberish instead?

It would be interesting to know if anyone else has an opinion on this.

flackr: > I'm not really sure. Once we are able to use percentages that resolve to times (e.g. `duration` is 50% of my parent group) I wonder if it would be useful to have `getComputedTiming()` return the actual number of seconds it resolves to?

I could see this being useful, and might be worth doing.

> In which case, should percentages that resolve against scroll timelines resolve to pixels?

This depends on what you consider to be the underlying value of the ScrollTimeline right? It could be pixels but it could also only operate in percentage of scroll or percentage of duration?

> Unfortunately `getComputedStyle` doesn't give an entirely clear precedent here since it sometimes resolves percentages (e.g. for `width` and `height`) but otherwise doesn't. I guess the intention is that it should always return percentages. So maybe we should return CSSNumberish instead?

I think returning CSSNumberish and keeping percentages in the computed timing makes sense. One thing that I believe supports this is this note from [css-cascade](https://drafts.csswg.org/css-cascade/#computed) (emphasis mine):

> Note: In general, the computed value resolves the specified value as far as possible **without laying out the document** or performing other expensive or hard-to-parallelize operations, such as resolving network requests or retrieving values other than from the element and its parent.

cc/ @andruud

andruud: Leaving percentages intact sounds entirely reasonable.

Although depending on layout in this case is probably less painful than normal, given the snapshotting described in [Avoiding cycles with layout](https://drafts.csswg.org/scroll-animations-1/#avoiding-cycles), which means we can resolve the percentages against whatever is known from the previous snapshot?

> returning CSSNumberish

Note that (from brief inspection of css-typed-om & Blink): there seems to be an "understanding" that a _return value_ of `CSSNumberish` means you always get a `CSSNumericValue`. So _readonly_ `CSSNumberish` attributes would probably ideally be `CSSNumericValue` attributes. So since this would be the first deviation from this (I think), we should probably specify under which circumstances we can expect a raw double.

web-animations-1/Overview.bs
```diff
@@ -4580,8 +4624,16 @@ dictionary OptionalEffectTiming {
     time taken to complete a single iteration of the <a>animation
     effect</a>.
 
-    In this level of this specification, the string value <code>auto</code>
-    is treated as the value zero for the purpose of timing model calculations
+    If the <a lt="timeline duration">duration</dfn> of the Animation's
+    <a>timeline</a> is a percentage, a non <code>auto</code> duration will
```
andruud, line 4628: Since you _just_ required the duration to be a percentage, do you need to mention non-auto?

web-animations-1/Overview.bs
birtles: > Note that (from brief inspection of css-typed-om & Blink): there seems to be an "understanding" that a _return value_ of `CSSNumberish` means you always get a `CSSNumericValue`. So _readonly_ `CSSNumberish` attributes would probably ideally be `CSSNumericValue` attributes. So since this would be the first deviation from this (I think), we should probably specify under which circumstances we can expect a raw double.

That sounds reasonable. If we were designing this from scratch I think we would always return a `CSSNumericValue` and it is only compatibility constraints that mean we sometimes need to return an actual number.

web-animations-1/Overview.bs
```diff
@@ -822,6 +831,29 @@ in which they are created. Certain <a lt="animation class">classes of
...
+time-based animation to a proportional animation</dfn>, effectively the
+[=timeline duration=] is divided up by the relative proportions of the
+specified times of the <a>animation</a>, <var>animation</var>. Concretely:
+
+1.  If the [=iteration duration=] is auto, then perform the following steps.
+
+    1.   Compute [=start delay=] and [=end delay=] to 0, as it is not possible
+         to mix time and proportions. Future versions may allow these properties
+         to be assigned percentages.
+
```
ogerchikov, line 844: Need "otherwise" close? The rest of the bullets are applied when iteration duration is not "auto".

web-animations-1/Overview.bs
```diff
@@ -822,6 +831,29 @@ in which they are created. Certain <a lt="animation class">classes of
...
+1.  Let <var>total time</var> be equal to the result of evaluating
+    <code>|start delay| + |iteration count| * |iteration duration| + |end delay|</code>.
+
+1.  Compute [=start delay=] to be the result of evaluating
+    <code>|start delay| / |total time|</code>.
+
+1.  Compute [=iteration duration=] to be the result of evaluating
+    <code>|iteration duration| / |total time|</code>.
```
ogerchikov, line 852: Need to divide by  iteration count.

web-animations-1/Overview.bs
```diff
@@ -822,6 +831,29 @@ in which they are created. Certain <a lt="animation class">classes of
...
+
+1.  Compute [=end delay=] to be the result of evaluating
+    <code>|end delay| / |total time|</code>.
+
```
ogerchikov, line 856: In this calculation the resulted values sum to 1. However when iteration duration is "auto", the timing values sum to timeline.duration. I think for consistency and correctness the values in both cases should sum to timeline.duration.

web-animations-1/Overview.bs
flackr: Thanks, done.

web-animations-1/Overview.bs
flackr: I think it's correct as is, e.g. with start delay 2, iteration count 5, iteration duration 1, end delay 3:
total time is 2 + 5 * 1 + 3 = 10
iteration duration is calculated as 1 / 10 = 10%

web-animations-1/Overview.bs
flackr: Good catch, it's possible timeline duration will not always be 100% and we should match that here. Done.

web-animations-1/Overview.bs
flackr: I just required the timeline duration to be a percentage, however the non-auto duration refers to the animation duration.

alshamiri5 (approved):

flackr:
@birtles I've moved all of the web-animations-1 changes to deltas in web-animations-2. @JTensai is currently prototyping this in chromium and I think has not run into any issues. Happy to wait until we have a working implementation to land this but just wanted to confirm that this addresses the main concerns.

web-animations-2/Overview.bs
```diff
@@ -143,10 +143,51 @@ Along with the following updated description:
...
 
+<h3 id="timelines">Timelines</h3>
+
+Add:
+
+> The <dfn lt="timeline duration">duration</a> of a timeline gives the
+> maximum value a timeline may generate for
+> [=timeline current time|current time=]. This is used to calculate the
+> [=iteration duration=] when the string <code>auto</code> is specified such that
+> the animation fills the available time.
```
birtles, line 154: A few thoughts here:

* Firstly, this is part of the model description, as opposed to the API, so we generally try to phrase things in terms of the concepts rather than the specific interface types. (I'm not sure if that's a good idea after all, but that's the state of affairs at the moment.)
* Secondly, I think we need to say where the `auto` value is specified -- in this case I think we are talking about the target effect of an animation association with the timeline. However...
* Thirdly, in level 2 of the spec, `auto` is also used to allow groups to have a duration calculated based on the lengths of their children. At the model level, `auto` means "use the intrinsic iteration duration".

So, I think this sentence should probably be something like:

> This is used in calculating the intrinsic iteration duration of the target effect of animations associated with the timeline.

Which brings us to something I think you might not want to hear: I think we probably need to specify how these different cases interact.

I wonder if can do something like the following:

1. Redefine the _intrinsic iteration duration_ as being either a number or a percentage/proportion or possibly a calc expression of both.
2. Define the _intrinsic iteration duration_ of keyframe effects (or all animation effects that don't supply their own definition for intrinsic iteration duration) as being 100% (instead of 0s).
3. Rework the definition of the _intrinsic iteration duration_ of group effects to similarly produce either a number or proportion or potentially a `calc()` if we allow that.

    In a sense, these steps produce a bottom-up initial calculation of the target effect's intrinsic iteration duration.
4. (Optionally define the timeline duration of a document timeline as being zero instead of null since I think that might make the spec simpler if there's no particular need for `null`.)
5. Define a "resolve computed intervals" procedure where we start with the target effect of an animation, and resolve its duration, start delay, and end delay based on the particular combination of its:

    * Timeline duration
    * Specified iteration duration or intrinsic iteration duration
    * Specified start delay and end delay

    This would be similar to the conversion procedure defined in this spec. In the simple case, nothing would be need to be resolved as everything would already be in milliseconds. Similarly, when we have everything in percentages, there would be nothing to do.

    In other cases, where we have a millisecond duration timeline (e.g. `DocumentTimeline` with its 0ms duration) and proportion times, we would resolve the proportions to milliseconds. (This would allow a `KeyframeEffect` with an `auto` duration to continue producing a resolved duration of 0s.)

    In yet other cases where we have a percentage timeline and millisecond times, we would convert the milliseconds to percentages as we have in the conversion algorithm in this spec.

    Then of course, there are the odd cases where we have a mix we don't support and where we need to use some sensible behavior. I think it's worth at least thinking about how we could support `calc(50% + 2s)`.
6. Recursively apply this behavior to the target effect's child effects.

I'm really sorry. I know you wanted to keep this patch as simple as possible just to get this working for scroll timelines, but I think we need to work out how this integrates with group effects.

On the positive side, if the above approach works, we'll have solved percentage based timing in general and made something really useful, I think.

I think it would also extend naturally to other use cases like `KeyframeEffect`s where the keyframe offsets are specified in milliseconds. We would simply plug those values into the calculation of the intrinsic iteration duration of the keyframe effect and the rest would just work.

birtles (changes requested):
Hi Rob, I'm really sorry. I wrote a long comment here because I think as we add this to level 2 we need to work out how it integrates with groups.

web-animations-2/Overview.bs
birtles: I should clarify step 6:

> 6. Recursively apply this behavior to the target effect's child effects.

I don't think we'd need to actually write an algorithm to recursively visit the children. Instead we could just write a definition for _computed iteration duration_, _computed start delay_ etc. in terms of their parent time source's intrinsic iteration duration and then we'd naturally get the recursive behavior (i.e. the bottom-up followed by top-down traversal).

kevers-google: Hi Brian,
I've jumped in to assist Rob on the spec work and have a new pull request (https://github.com/w3c/csswg-drafts/pull/6337), which hopefully addresses the outstanding issues.

Some of the key changes:
* intrinsic iteration duration updated to be either a percent or time.  In the case of a progress based timeline, the percent is simply 100% / iteration_count.  Note that in the future, if start and end delay are updated to allow percentages, the intrinsic iteration duration becomes (100% - start - end) / iteration_count.  Once start and end percentages are supported, we will need to be able to handle start + end > 100%, which would otherwise result in a negative intrinsic iteration duration.  This should be fairly easy to handle. Just raising the issue now so that we don't lose track of it.
* Updated to procedure for calculating the intrinsic iteration duration to support group/sequence effects as well as progress based timelines.

Note that mixing times and percentages opens up some awkward edge cases that are not easily addressed (examples below).  Ultimately, we can easily convert form entirely time based to percentage based.  The inverse is only possible if knowing the effect end time.  For now, I was planning to stick with Rob's suggestion of start and end being set to 0 if the timeline is progress based.

Currently the iteration duration can be a time or 'auto'.  In the future, we could extend to time 'auto' or percent. I would prefer to defer introducing percentages outside of 'auto' to avoid complications that would be introduced until all of the details can be worked out.  The value for effect end time depends on iteration duration, which in turn could be the intrinsic iteration duration if 'auto'.  The intrinsic iteration duration in turn can depend on effect end time. 

Let's start with a few cases that would work, and them move to the ugly cases:

Case 1:

.anim {
  animation: parallax 1s 0.25s scroll-timeline;
}
.anim2 {
  Animation: parallax 2s scroll-timeline; 
}

Here everything is time based, so the conversion to percentages for use with a progress based scroll timeline is trivial.

anim: iteration duration = 100% * (1s / 1.5s) = 66.67%, start delay = end delay = 100% * 0.25s / 1.5s = 16.67%
anim2: iteration duration = 100%, start delay = end delay = 0%

Case 2:

.anim {
  animation-name: parallax;
  animation-delay: 20%; /* once supported as a percent */
  Animation-timeline: scroll-timeline;
}

intrinsic iteration duration = 100% - 20% = 80% = iteration duration

Case 3:

.anim {
  animation-name: parallax;
  animation-duration: 1s;
  animation-delay: 20%;
}

Now we would need to be prepared to a bit more work to resolve the timing by collecting times and percentages.

end time = 1s + 20% = 100% ==> 1s  = 80% ==> 1.2.s = 100%.

Note: at this stage we can get inconsistent timings (e.g. 0s animation + delay != 100%). 

Case 4: 

.anim {
  animation-name: parallax;
  animation-duration: 75%;
  animation-delay: 1s;
}

Similar approach to case 3.

end time = 75% + 1s = 100% ==> 1s = 25% ==> 4s = 100%.

Case 5:

.anim {
  animation-name: parallax;
  animation-delay: 2s;
  Animation-iteration-count: 2;
}

Now for an ugly case.

end time = 2*(intrinsic iteration duration) + 2s  = 100% 
==> 2*((100% - 2s))/2 + 2s = 100%
==> 100% = 100%

Here the intrinsic iteration duration is arbitrary. It may make sense to use the minimum solution (0s/0%) in this case. I welcome further discussion in this area, but would prefer to defer allowing mixing of percentages and times to a followup pull request.  I believe have a pretty solid base, which can be extended in a way that won't require backpedaling.

birtles: Hi Kevin,

Thank you so much for following up on this. Unfortunately I'm going to be quite busy for the next few days. Do you mind if I get back to you on this after that? Please feel to ping me if I'm taking too long or if you have any particular time constraints.

kevers-google: No rush.

birtles: Thanks for your patience on this. I've finally reviewed #6337 and it looks good.

Responding to your comment,

> Case 1:
> 
> .anim {
> animation: parallax 1s 0.25s scroll-timeline;
> }
> .anim2 {
> Animation: parallax 2s scroll-timeline;
> }
> 
> Here everything is time based, so the conversion to percentages for use with a progress based scroll timeline is trivial.
> 
> anim: iteration duration = 100% * (1s / 1.5s) = 66.67%, start delay = end delay = 100% * 0.25s / 1.5s = 16.67%
> anim2: iteration duration = 100%, start delay = end delay = 0%

I'm curious why start and end delay end up being made the same. Shouldn't the `0.25s` from the `animation` declaration be mapped to start delay only?

Overall though it's looking great. I'm a bit uncomfortable about having combinations of content that don't work now but which we expect to work in future.

It would be good to add another PR at some point filling out those cases. I'd even be to say, for example, when an animation is directly or indirectly associated with a progress-based timeline, if it uses a combination of percentages and times, we just ignore the times (i.e. use their default values) if need be. At least that's consistent, I guess.

Maybe that's roughly equivalent to what you have, though. From what I tell though, with #6337 if you specified duration 5s, start delay 5%, end delay 2s, we'd end up ignoring _both_ start delay and end delay as opposed to simply ignoring the time-based end delay.

kevers-google: Oops...yes case 1 should be  a 1.25s animation which then maps to 100% so that the iteration duration and start delay can be converted to percentages.

flackr:
Closing as this has been continued in #6337

Issue #4683, opened 2020-01-18
[resize-observer] attributes cannot be sequence<T>


foolip:
There are 3 such attributes in https://drafts.csswg.org/resize-observer/#resize-observer-entry-interface

https://heycam.github.io/webidl/#idl-sequence says "Sequences must not be used as the type of an attribute or constant." This is because it would return a new JS array every time.

These should probably be FrozenArray<T> instead.

foolip:
@atotic @gregwhitworth

stephenmcgruer:
Looks like there's still one left; ResizeObservation has a lastReportedSizes attribute that is a sequence. (https://drafts.csswg.org/resize-observer/#dom-resizeobservation-lastreportedsizes)

dlibby-:
Good catch - that interface is just an internal, non-normative web idl, as explained by this [note](https://drafts.csswg.org/resize-observer/#resize-observation-interface):

`This section is non-normative. ResizeObservation is an example struct that can be used in implementation of Resize Observer. It is being included here in order to help provide clarity during the processing model. It effectively holds observation information for a single Element. This interface is not visible to Javascript.`

I'll go ahead and update that as well for clarity

saschanaz:
>Good catch - that interface is just an internal, non-normative web idl

How about using `.example` CSS class there?

saschanaz:
Fixed by #4909

Issue #3839, opened 2019-04-17
[resize-observer] the example "ResizeObservation"


dholbert:
* please tag the issue title with the spec's shortname, like `[css-foo]`
* please link to the spec section you're talking about, or at least the spec

https://wpt.fyi/results/resize-observer/idlharness.window.html shows a bunch of cross-browser test failures for the `ResizeObservation interface`, which come from chunk of IDL:

https://github.com/web-platform-tests/wpt/blob/a8d97c0c7c58cdc2def51573e30c2cad92923d75/interfaces/resize-observer.idl#L39-L43

```
interface ResizeObservation {
    readonly attribute Element target;
    readonly attribute ResizeObserverBoxOptions observedBox;
    readonly attribute ResizeObserverSize lastReportedSize;
};
```

This interface is mentioned in the spec, but it's clearly labeled as an "example" which is *not* supposed to be exposed to JS:
> **§ 3.1. ResizeObservation example struct**
> This section is not normative. ResizeObservation is an example struct that can be used in implementation of Resize Observer. It is being included here in order to help provide clarity during the processing model. It effectively holds observation information for a single Element. This interface is not visible to Javascript.
> [...]
> interface ResizeObservation {

https://drafts.csswg.org/resize-observer/#resize-observation-interface

Unfortunately, the WPT `interfaces/resize-observer.idl` file is autogenerated, so we can't just edit it to remove this example interface. It'll just come back whenever the file gets regenerated, presumably.  So I think we need to edit the spec to make it clearer (to the Reffy tool) that the

dholbert:
Also, worth stating for the record -- both Chrome 76 and Firefox Nightly correctly "fail" all of the ResizeObservation subtests right now, as shown on [wpt.fyi](https://wpt.fyi/results/resize-observer/idlharness.window.html).  (And other engines don't implement ResizeObserver at all.)

So, all implementations are correct here, and we just need a spec-fix (which will then implicitly fix the test, since the test is generated by the spec).

gregwhitworth:
@dholbert I asked on the testing IRC channel, but I'll ask here too. I have no knowledge of the Reffy tool, so how do I change the spec to make it so that it isn't using this example IDL as actual IDL?

dholbert:
I'm afraid I have no knowledge of the Reffy tool either.  (I only figured out that it was involved via following the trail of clues to figure out where the test data was coming from.)

bkardell:
[...]

PR #4898, opened 2020-03-24
[resize-observer-1] attributes cannot be sequence<T> #4683

sahirv:
Fix invalid WebIDL by changing sequence to FrozenArray for lastReportedSizes in the ResizeObservation interface. Currently there are still normative references to this non-normative interface, discussed in #3839. These will be addressed in a follow-up.

### DIFF
--- a/resize-observer-1/Overview.bs
+++ b/resize-observer-1/Overview.bs
@@ -318,7 +318,7 @@ interface ResizeObservation {
     : <dfn>observedBox</dfn>
     :: Which box is being observed.
     : <dfn>lastReportedSizes</dfn>
-    :: Ordered frozen array of last reported sizes.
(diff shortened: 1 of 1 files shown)


resize-observer-1/Overview.bs
```diff
@@ -318,7 +318,7 @@ interface ResizeObservation {
     : <dfn>observedBox</dfn>
     :: Which box is being observed.
     : <dfn>lastReportedSizes</dfn>
-    :: Ordered sequence of last reported sizes.
+    :: Ordered frozen array of last reported sizes.
```
dlibby-: Let's just say 'Ordered array' instead since the Frozen-ness of the array is not particularly relevant.

PR #4938, opened 2020-04-10
[css-ruby] Clarify the box model of base/annotation containers

### LINKED ISSUE #4937: [css-ruby] Be more specific about the box model of base/annotation container boxes
(this issue is similar to, but different from #4935 and #4936)

[Section 3](https://drafts.csswg.org/css-ruby-1/#ruby-layout) tells us that:
* > Each ruby base container is sized and positioned to contain exactly all of its ruby bases’ margin boxes.
* > Each ruby annotation container is sized and positioned to contain exactly all of its ruby annotations’ margin boxes

Presumably that means "**The content area of** each ruby base/annotation container…" (and then any borders, margins, and padding go outside of that). This should be clarified.

The same section 3 also states:
> Ruby annotation containers are stacked outward over or under their corresponding ruby base container, without any intervening space.

This could be interpreted to mean either:
* **the content box of** ruby annotation containers are stacked outward over or under **the content box of** their corresponding ruby base container, **ignoring any border, margin, or padding,** without any intervening space.
* **the margin box of** ruby annotation containers are stacked outward over or under **the margin box of** their corresponding ruby base container, without any intervening space.

[Section 3.2](https://drafts.csswg.org/css-ruby-1/#box-style) gives a hint:

> […] the UA is not required to support any of the box properties (borders, margins, padding), any of the background properties or outline properties, or any other property that illustrates the bounds of the box on ruby base container boxes, ruby annotation container boxes […]

Given that, I suspect the former (content box) interpretation is indented, otherwise whether borders, margins, and padding on base container boxes and annotation container boxes affect layout out nor would be implementation dependent, which would be bad. But this should be made explicit.

frivoal:
Closes #4937

### DIFF
--- a/css-ruby-1/Overview.bs
+++ b/css-ruby-1/Overview.bs
@@ -647,8 +647,8 @@ Ruby Layout</h2>
 	its [=base level=] is laid out on the line,
 	aligned according to its 'vertical-align' property
 	exactly as if its [=ruby bases=] were a regular sequence of inline boxes.
-	Each [=ruby base container=] is sized and positioned
-	to contain exactly all of its [=ruby bases=]’ margin boxes.
+	The [=content box=] of each [=ruby base container=] is sized and positioned
+	to contain exactly all of its [=ruby bases=]’ [=margin boxes=].
 
 	[=Ruby annotations=] associated with the [=base level=]
 	are then positioned with respect to their [=ruby base boxes=]
@@ -656,14 +656,16 @@ Ruby Layout</h2>
 	[=Ruby annotations=] within a level (within a single [=ruby container=])
 	are aligned to each other as if they were inline boxes
 	participating in the same inline formatting context.
-	Each [=ruby annotation container=] is sized and positioned
-	to contain exactly all of its [=ruby annotations=]’ margin boxes.
+	The [=content box=] of each [=ruby annotation container=] is sized and positioned
+	to contain exactly all of its [=ruby annotations=]’ [=margin boxes=].
 
-	[=Ruby annotation containers=] are stacked outward
-	over or under their corresponding [=ruby base container=],
+	The [=content boxes=] of [=ruby annotation containers=] are stacked outward
+	over or under the [=content box=] of their corresponding [=ruby base container=],
+	(ignoring any [=margin=], [=border=], or [=padding=]
+	on either the [=ruby annotation containers=] or the [=ruby base containers=])
 	without any intervening space.
 
-	Issue: Should block-axis margins collapse?
+	Issue: Should the block-axis margins of ruby bases and ruby containers collapse?
 	This makes layout more robust,
 	but is inconsistent with how inlines behave along the inline-axis.
 


css-ruby-1/Overview.bs
```diff
@@ -647,23 +647,25 @@ Ruby Layout</h2>
 	its [=base level=] is laid out on the line,
 	aligned according to its 'vertical-align' property
 	exactly as if its [=ruby bases=] were a regular sequence of inline boxes.
-	Each [=ruby base container=] is sized and positioned
-	to contain exactly all of its [=ruby bases=]’ margin boxes.
+	The [=content box=] of each [=ruby base container=] is sized and positioned
+	to contain exactly all of its [=ruby bases=]’ [=margin boxes=].
```
nigelmegitt, line 651: Is "contain exactly" intended to mean that it is the _smallest_ size possible that completely contains the margin boxes? It might be worth being explicit about this, since any extension relative to that size would still contain the boxes. 

It is probably implicit that the content box is a simple rectangle, but given that the union of margin boxes might not be a simple rectangle, consider making it explicit.

Same comments for ruby annotation container's content box size, line 659.

css-ruby-1/Overview.bs
```diff
@@ -647,23 +647,25 @@ Ruby Layout</h2>
...
 	[=Ruby annotations=] associated with the [=base level=]
 	are then positioned with respect to their [=ruby base boxes=]
 	according to the applicable 'ruby-position' values.
 	[=Ruby annotations=] within a level (within a single [=ruby container=])
 	are aligned to each other as if they were inline boxes
 	participating in the same inline formatting context.
-	Each [=ruby annotation container=] is sized and positioned
-	to contain exactly all of its [=ruby annotations=]’ margin boxes.
+	The [=content box=] of each [=ruby annotation container=] is sized and positioned
+	to contain exactly all of its [=ruby annotations=]’ [=margin boxes=].
```
nigelmegitt, line 660: Same comments as on line 651.

css-ruby-1/Overview.bs
```diff
@@ -647,23 +647,25 @@ Ruby Layout</h2>
...
 
-	[=Ruby annotation containers=] are stacked outward
-	over or under their corresponding [=ruby base container=],
+	The [=content boxes=] of [=ruby annotation containers=] are stacked outward
+	over or under the [=content box=] of their corresponding [=ruby base container=],
```
nigelmegitt, line 663: For reference, see also TTML2's [`tts:rubyPosition`](https://www.w3.org/TR/ttml2/#style-attribute-rubyPosition) which specifies whether under or over should be used for each line in the ruby text. If there is a similar CSS control for this it would be useful to reference it here, since there are use cases where it is important to specify the position.

css-ruby-1/Overview.bs
nigelmegitt: I raised the request for a defined term "contains exactly" at #4950.

frivoal:
This PR is made redundant by https://github.com/w3c/csswg-drafts/commit/41cdb828cfb8ba42b50f0ba86a057f6f49388dce

PR #4939, opened 2020-04-10
[css-color-4] Fix rec2020 white chromaticity.

jdashg:
A previous reformat introduced an error.


svgeesus:
Oh wow, nice catch. Thanks!

Issue #4947, opened 2020-04-13
[cssom] Define order of custom properties in computed style.


emilio:
In #1316 we resolved they should be in the computed style but we didn't resolve the ordering.

PR #4948, opened 2020-04-13
[cssom] Include custom properties in getComputedStyle().

### LINKED ISSUE #1316: [cssom] should custom properties be exposed on computed style declarations?
https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle

The spec currently says that the list of declarations that a `CSSStyleDeclaration` returned by `getComputedStyle` includes only "supported CSS properties", which means that it doesn't include custom properties.  I think it makes sense to expose these through computed style objects.

Gecko does this, both via `item()` and `getPropertyValue()`.  For `item()` (and the indexed getter) custom properties appear after all of the built-in properties, sorted by the order that they happen to cascade in (though that should probably become lexicographically sorted or something).

emilio:
Closes #1316. #4947 for the ordering issue.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2853,11 +2853,16 @@ steps:
   set <var>decls</var> to a list of all longhand properties that are
   <a lt="supported CSS property">supported CSS properties</a>, in lexicographical order,
   with the value being the <a>resolved value</a> computed for <var>obj</var>
-  using the style rules associated with <var>doc</var>.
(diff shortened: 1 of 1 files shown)


emilio:
(Not sure about my english there, is `whose` correct?)

tabatkins:
Yeah, "whose" is correct there, it's okay to lightly anthropomorphize.

emilio:
Thanks Tab :)

PR #4967, opened 2020-04-17
[cssom-1] Improve accuracy of <color> serialization

mathiasbynens:
Either the word "integer" was missing from most of these steps, or it was accidentally added in one step. ~~This patch assumes the former.~~

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2603,19 +2603,19 @@ depends on the component, as follows:
  <li>The string "<code>rgb(</code>".
  <li>The shortest base-ten integer serialization of the color's red component.
  <li>The string "<code>, </code>".
- <li>The shortest base-ten serialization of the color's green component.
+ <li>The shortest base-ten integer serialization of the color's green component.
  <li>The string "<code>, </code>".
- <li>The shortest base-ten serialization of the color's blue component.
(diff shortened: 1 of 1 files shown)


mathiasbynens:
> Also, while implementations were historically limited to 8 bits (and thus, integers in the range 0 to 255) this is being changed, to allow higher precision while keeping the historical 0 to 255 range; thus, these would no longer be integers.

Thanks. I’ve updated the patch accordingly.

emilio (approved):
Anyhow looks good.

mathiasbynens:
The IPR bot report takes me to https://www.w3.org/2004/01/pp-impl/32061/join, which explains there's a form I should fill out, but all links point to the same document (which doesn't contain any forms as far as I can tell). 🤔  Is there a way for me to waive my IPR on this silly one-word patch? I am a W3C invited expert and work for a W3C member organization.

mathiasbynens:

svgeesus:
> The IPR bot report takes me to https://www.w3.org/2004/01/pp-impl/32061/join, which explains there's a form I should fill out,

That is the "join the CSS WG" form, which is filled out by the AC Rep not the participant.

I suspect your W3C account and your GitHub account are not linked, so the bot does not know who you are.

PR #5033, opened 2020-04-30
[web-animations-1] Add timeline to animate() options (#5013)

### LINKED ISSUE #5013: [web-animations-1] Allow setting timeline in Element.animate()
## Proposal

Web Animations allows creation of `AnimationTimeline` but it does not support using a custom timeline in [`element.animate()` method](https://drafts.csswg.org/web-animations/#dom-animatable-animate). The only option is to use Animation constructor directly.

I believe as we get closer to having `ScrollTimeline` the ability to continue to use the familiar and simple to use `Element.animate()` for scroll linked animations is important. This issue is here to consider introducing a way to to pass in a timeline as part of `Element.animate(keyframes, options)`.

## Syntax Options

If it is decided that we want this, here are some ideas on syntax:

### Extend options to have a timeline member
This is an example of how it may work:
```js
const timeline = new ScrollTimeline({source: $some_scroller, timeRange: 1000});
$div.animate({opacity: [0, 1]}, {delay: 100, duration: 500, timeline: timeline});
```
Pros:
  * Simple to undestand
 * keeps timing related options in a single dictionary (I consider timeline a timing related option)
Cons:
 * At the moment spec passes options verbatim to KeyframeEffect but timeline is an animation concept so we have to separate these in spec before passing the options to keyframes. This is not a big deal IMHO.

FWIW, this is the method that @flackr  has chosen in [their scroll-timeline polyfill](https://github.com/flackr/scroll-timeline/).

### Add a third parameter just for timeline

This is an example of how it may work:
```js
const timeline = new ScrollTimeline({source: $some_scroller, timeRange: 1000});
$div.animate({opacity: [0, 1]}, {delay: 100, duration: 500}, {timeline: timeline} /* or just timeline */);
```

I find this a bit less ergonomic that the previous one. It may be more appealing if there were other animation specific options.

majido:
Fixes #5013 

The options dictionary in `Element.animate` now has an optional nullable timeline
member. This allows timeline to be set via that method in addition to Animation
constructor.

Here are some examples of how this may be used:
```
// Create animation with a null timeline.
div.animate({color: 'red' }, { duration: 1000, timeline: null});
```

```
// Create a scroll-linked animation.
div.animate({color: 'red' }, { duration: 1000, timeline: new ScrollTimeline()});
```

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -5814,30 +5814,21 @@ dictionary GetAnimationsOptions {
         If the above procedure causes an exception to be thrown, propagate the
         exception and abort this procedure.
 
+    1.  If <var>options</var> is a {{KeyframeAnimationOptions}} object, let
+        <var>timeline</var> be the <code>timeline</code> member of
+        <var>options</var> or, if <code>timeline</code> member of
+        <var>options</var> is missing, be the <a>default document timeline</a>
+        of the <a>node document</a> of the element on which this method was
+        called.
+
     1.  Construct a new {{Animation}} object, <var>animation</var>, in
         the <a>relevant Realm</a> of <var>target</var> by using the
         same procedure as the {{Animation()}} constructor, passing
-        <var>effect</var> as the argument of the same name, and
-        the <a>default document timeline</a> of the <a>node document</a>
-        of the element on which this method was called as the
-        <var>timeline</var> argument.
+        <var>effect</var> and <var>timeline</var> as arguments of the same name.
 
-    1.  If <var>options</var> is a {{KeyframeAnimationOptions}} object, run the
-        following steps:
-
-        1.  Assign the value of the <code>id</code> member of <var>options</var>
-            to <var>animation</var>'s {{Animation/id}} attribute.
-
-        2.  Let <var>timeline</var> be the <code>timeline</code> member of
-            <var>options</var> or, if <code>timeline</code> member of
-            <var>options</var> is missing, be the
-            <a>default document timeline</a> of the {{Document}}
-            <a lt="document associated with a window">associated</a> with the
-            {{Window}} that is the <a>current global object</a>.
-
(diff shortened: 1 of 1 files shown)


web-animations-1/Overview.bs
```diff
@@ -5856,10 +5870,14 @@ animation.play();</pre>
...
-        options)}} constructor as the <var>options</var> parameter and has the
-        same interpretation as defined for that constructor.
+    ::  The timing and animation options for the created {{KeyframeEffect}} and
+        {{Animation}}.
+        For {{KeyframeEffect}} this value is passed to the
+        {{KeyframeEffect(target, keyframes, options)}} constructor as the
+        <var>options</var> parameter and has the same interpretation as defined
+        for that constructor. For {{Animation}}, <code>id</code> and
+        <code>timeline</code> member gets assigned to the created
+        {{Animation}}'s {{Animation/id}} and {{Animation/timeline}} attributes.
```
birtles, line 5880: This section is a bit odd, I wonder if it is needed?

web-animations-1/Overview.bs
```diff
@@ -5821,9 +5822,22 @@ dictionary GetAnimationsOptions {
...
+        2.  Let <var>timeline</var> be the <code>timeline</code> member of
+            <var>options</var> or, if <code>timeline</code> member of
+            <var>options</var> is missing, be the
+            <a>default document timeline</a> of the {{Document}}
+            <a lt="document associated with a window">associated</a> with the
+            {{Window}} that is the <a>current global object</a>.
+
+        3.  Run the procedure to <a>set the timeline of an animation</a> on
+            <var>animation</var> passing <var>timeline</var> as the <var>new
+            timeline</var>.
```
birtles, line 5840: I wonder if it is observable if an implementation passes the timeline directly to the `Animation` constructor as opposed to setting it after the fact? Maybe it's not currently? If not that's fine, but otherwise we might want to consider what the preferred behavior is.

PR #5041, opened 2020-05-05
Editorial rewrite of HDR text

frivoal:
This is an attempt to rewrite the section of media-queries that defines and uses brightness and contrast ratio.

I think this makes the intent clearer, removes ambiguities, while keeping the details of the implementation up to user agents as intended.

This corresponds to the "For section 6.6 …" bullet point of https://github.com/w3c/csswg-drafts/pull/4678#issuecomment-610725135.

Review from @svgeesus, @gregwhitworth, @vi-dot-cpp, or @gregwfreedman very much appreciated.

### DIFF
--- a/mediaqueries-4/Overview.bs
+++ b/mediaqueries-4/Overview.bs
@@ -1706,26 +1706,47 @@ color depth, and contrast ratio that are supported by the UA and output device.
 <dl dfn-type=value dfn-for="@media/dynamic-range">
 	<dt><dfn>high</dfn>
 	<dd>
-		If the following are true for the UA and the output device:
-		* max brightness is considered high
-		* color depth is greater than 24 bit
-		* has a <a>high contrast ratio</a>
+		The combination of the User Agent and the output device
+		fulfill all of the following criteria:
+		* it has a [=high peak brightness=]
+		* it has a [=high contrast ratio=]
+		* its color depth is greater than 24 bit or 8 bit per color component of RGB
 	<dt><dfn>standard</dfn>
 	<dd>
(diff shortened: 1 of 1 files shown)


mediaqueries-4/Overview.bs
```diff
@@ -1706,26 +1706,47 @@ color depth, and contrast ratio that are supported by the UA and output device.
...
-This text is non-normative. There currently is no agreed upon standardization of
-<dfn export>high contrast ratio</dfn>, <dfn export>low contrast ratio</dfn>,
-max brightness or low brightness as it relates to displays. As such, the
-determination for '@media/dynamic-range' and '@media/video-dynamic-range' will
-be defined by the user agent.
+	<dfn export>Peak brightness</dfn> refers to how bright the brightest point
+	a light-emitting device such as an LCD screen can produce,
+	or in the case of a light reflective device such as paper or e-ink,
+	the point at which it least absorbs light.
+	Some devices can only produce their [=peak brightness=]
```
gregwhitworth, line 1727: I'm curious if denoting that they can only have peak brightness for brief amounts of time is relevant to defining peak brightness. Maybe this should be a note if you're wanting authors to be aware of it?

frivoal: I though it was useful because this is a thing that happens with HDR screens which is unfamiliar to people only used to SDR screen, and I didn't want people to take the opposite assumption (only sustainable levels of brightness count) out of habit or by mistake.

But yeah, I don't mind turning it into a note if you think that's better.

Issue #5018, opened 2020-04-28
[css-scroll-anchoring-1] Anchor selection should prefer 'active' nodes 


vmpstr:
[Anchor selection algorithm](https://drafts.csswg.org/css-scroll-anchoring/#anchor-node-selection) only considers the DOM relationship of the nodes and their visibility in the scroller.

There are some situations where scroll initiated by a user-agent algorithm, such as find-in-page, causes us to select a suboptimal scroll anchor which pushes the find-in-page match outside of the viewport. 

This situation is possible if there is, for example, an intersection observer, which sizes on-screen elements differently from off-screen elements. When find-in-page match is scrolled to the center, we can pick a scroll anchor near the top of the viewport and after resizing and adjusting the scroll position end up in a situations where the find-in-page match is not visible.

One way to fix this is to prefer certain nodes in the anchor selection algorithm. Some potential interesting nodes (in no particular order) are the following:
- Find-in-page active match
- An element which was the target of scrollIntoView() "recently" (since the last anchoring adjustment?)
- An element with focus
- An element containing user selection

(These should only be selected if they are at least partially visible, and are a descendant of the scroller, of course)

There may be more interesting nodes, or maybe some of the ones I listed are not as interesting, but I think the user-agents can use information other than DOM structure to select a nicer anchor

Any thoughts?

chrishtr:
Another interesting "node" is an editing or selection caret.

emilio:
In Gecko we use the concept of "active" node, to some extent, as in, where does focus navigation, start from or such, and it is the selection position of either the focused element (if it owns a selection) or the per-document selection otherwise. We move that when you click on the page for example.

So for example if you click on the text of @chrishtr's comment and then tab from there you go to the starting point.

It'd be nice not to invent more complex concepts for something like this, but I don't know what other browsers do for tab navigation for example.

vmpstr:
> One thing to be careful of though is that we don't interfere with draggable interfaces. I.e. imagine a scrollable list where you can drag items to reorder - we wouldn't want dragging an item to try to keep that node anchored in the scroller as it would no longer be connected to your cursor / finger.

If I understand correctly, then this could already be a problem since the current anchor selection could select a node that is being dragged, right? It's just not as predictable

> It'd be nice not to invent more complex concepts for something like this, but I don't know what other browsers do for tab navigation for example.

[...]

PR #5050, opened 2020-05-07
[css-scroll-anchoring-1] Add priority candidates for scroll anchor selection #5018

vmpstr:
The PR adds scroll anchoring priority candidates which would be used before the DOM order children are considered as scroll anchors.

#5018

### DIFF
--- a/css-scroll-anchoring-1/Overview.bs
+++ b/css-scroll-anchoring-1/Overview.bs
@@ -76,7 +76,8 @@ Anchor Node Selection</h3>
 
 Each <a>scrolling box</a> aims to select an <a>anchor node</a>
 that is deep in the DOM
-and close to the block start edge of its <a>optimal viewing region</a>.
+and either should be prioritized as an important DOM node or is close to the
+block start edge of its <a>optimal viewing region</a>.
 
 Note: If the user agent does not support the 'scroll-padding' property,
 the optimal viewing region of the scrolling box is equivalent to its <a>content area</a>.
@@ -85,16 +86,41 @@ An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
 The anchor node is always a <a>descendant</a> of the <a>scrolling box</a>.
 In some cases, a scrolling box may not select any anchor node.
 
+An element |C| is a <dfn id="anchor-viable-candidate">viable candidate</dfn>
+for becoming a scroll anchor for a scrolling box |S| if it meets all of the
+following criteria:
+  * |C| is a non-<a>atomic inline</a>.
+  * |C| is <a>partially visible</a> or <a>fully visible</a> in |S|
+  * |C| is a descendant of |S|
+  * |C| is not in an <a>excluded subtree</a>
+  * None of the ancestors of |C| up to |S| are in an <a>excluded subtree</a>
+
+Some elements are considered to be <dfn
+id="anchor-priority-candidates">priority candidates</dfn> for anchor selection:
+  1. An element that <a href="https://html.spec.whatwg.org/multipage/semantics-other.html#selector-focus">has focus</a>.
+  2. An element containing the <a href="https://drafts.csswg.org/css-ui-3/#insertion-caret">insertion caret</a>.
+  3. An element containing the current active selected match of the
+      find-in-page user-agent algorithm. If the match spans multiple elements, then
+      consider only the first such element.
+
+Note that if the <a>priority candidate</a> is an <a>atomic inline</a> element,
+then instead consider its nearest non-atomic inline ancestor element as the
+priority candidate.
+
 <div algorithm>
 	The <dfn id="anchoring-algorithm">anchor node selection algorithm</dfn>
 	for a scrolling box |S| is as follows:
 
-	1. If |S| is associated with an element
-		whose computed value of the 'overflow-anchor' property is ''overflow-anchor/none'',
-		then do not select an anchor node for |S|.
-	2. Otherwise, for each DOM child |N| of the element or document associated with |S|,
-		perform the <a>candidate examination algorithm</a> for |N| in |S|,
-		and terminate if it selects an anchor node.
+  1. If |S| is associated with an element whose computed value of the
+    'overflow-anchor' property is ''overflow-anchor/none'', then do not select an
+    anchor node for |S|.
+  2. Otherwise, for each <a>priority candidate</a> |PC| in order specified,
+    check if |PC| is a <a>viable candidate</a> in |S|. If so, select it as an
+    anchor node and terminate.
+  3. Otherwise, for each DOM child |N| of the element or document associated
+    with |S|, perform the <a>candidate examination algorithm</a> for |N| in |S|,
+    and terminate if it selects an anchor node.
+
 </div>
 
 <div algorithm>
--- a/css-scroll-anchoring-1/Overview.bs
+++ b/css-scroll-anchoring-1/Overview.bs
@@ -97,7 +97,7 @@ following criteria:
 
 Some elements are considered to be <dfn
 id="anchor-priority-candidates">priority candidates</dfn> for anchor selection:
-  1. An element that <a href="https://html.spec.whatwg.org/multipage/semantics-other.html#selector-focus">has focus</a>.
+  1. An element that <a href="https://html.spec.whatwg.org/multipage/interaction.html#dom-document-activeelement">has focus</a>.
   2. An element containing the <a href="https://drafts.csswg.org/css-ui-3/#insertion-caret">insertion caret</a>.
   3. An element containing the current active selected match of the
       find-in-page user-agent algorithm. If the match spans multiple elements, then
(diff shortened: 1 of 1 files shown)


css-scroll-anchoring-1/Overview.bs
```diff
@@ -85,16 +86,41 @@ An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
 The anchor node is always a <a>descendant</a> of the <a>scrolling box</a>.
 In some cases, a scrolling box may not select any anchor node.
 
+An element |C| is a <dfn id="anchor-viable-candidate">viable candidate</dfn>
+for becoming a scroll anchor for a scrolling box |S| if it meets all of the
+following criteria:
+  * |C| is a non-<a>atomic inline</a>.
```
emilio, line 92: This doesn't make sense. Atomic inlines are fine, it's non-atomic-inlines the ones that can't be anchors.

css-scroll-anchoring-1/Overview.bs
```diff
@@ -85,16 +86,41 @@ An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
...
+  * |C| is <a>partially visible</a> or <a>fully visible</a> in |S|
+  * |C| is a descendant of |S|
+  * |C| is not in an <a>excluded subtree</a>
+  * None of the ancestors of |C| up to |S| are in an <a>excluded subtree</a>
+
+Some elements are considered to be <dfn
+id="anchor-priority-candidates">priority candidates</dfn> for anchor selection:
+  1. An element that <a href="https://html.spec.whatwg.org/multipage/semantics-other.html#selector-focus">has focus</a>.
+  2. An element containing the <a href="https://drafts.csswg.org/css-ui-3/#insertion-caret">insertion caret</a>.
+  3. An element containing the current active selected match of the
```
emilio, line 102: This is a bit ill-defined, IMO. What does a current active selected match mean? Firefox allows you to highlight all occurrences, is only the first one active? We should probably have a proper spec for find-in-page if we're making web-observable behavior depend on it.

css-scroll-anchoring-1/Overview.bs
```diff
@@ -85,16 +86,41 @@ An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
...
+      find-in-page user-agent algorithm. If the match spans multiple elements, then
+      consider only the first such element.
+
+Note that if the <a>priority candidate</a> is an <a>atomic inline</a> element,
```
emilio, line 106: This doesn't really make sense, atomic inlines are just fine.

css-scroll-anchoring-1/Overview.bs
```diff
@@ -85,16 +86,41 @@ An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
...
+following criteria:
+  * |C| is a non-<a>atomic inline</a>.
+  * |C| is <a>partially visible</a> or <a>fully visible</a> in |S|
+  * |C| is a descendant of |S|
+  * |C| is not in an <a>excluded subtree</a>
+  * None of the ancestors of |C| up to |S| are in an <a>excluded subtree</a>
+
+Some elements are considered to be <dfn
+id="anchor-priority-candidates">priority candidates</dfn> for anchor selection:
+  1. An element that <a href="https://html.spec.whatwg.org/multipage/semantics-other.html#selector-focus">has focus</a>.
```
emilio, line 100: Are these priorities sorted? As in, should we always prefer the focused element over the caret? Or find-in-page?

css-scroll-anchoring-1/Overview.bs
vmpstr: Oh you're absolutely right, the text I added has the negated meaning of atomic/non-atomic inlines. I fixed it up.

css-scroll-anchoring-1/Overview.bs
vmpstr: You're right that it would be great to have the find-in-page algorithm specified somewhere.. I don't know if that's something that can be done easily though since we need to agree on the details to match all of the browsers.

For your specific questions here, although Firefox, Safari, and Chrome all highlight (or are capable of highlighting) all matches, the active one is usually highlighted differently. By active I mean the one that is brought into view if it's not in view yet. This is the match that advances when the user clicks 'next' or 'previous' in the search dialog.

In lieu of a proper spec for find-in-page, I would appreciate any thoughts on how this could be phrased better...

css-scroll-anchoring-1/Overview.bs
vmpstr: I've updated the text with the corrections

css-scroll-anchoring-1/Overview.bs
vmpstr: Yes, in the algorithm below I say
"2. Otherwise, for each priority candidate _PC_ **in order specified**, ..."

In other words this block here is listing out the priority candidates, and at use time in the algorithm, we're specifying that the order matters.

Does that make sense or do you think we should reference the importance of the order here as well?

emilio: No, makes sense :)

emilio: I think this `#selector-focus` is not the right definition to link to. In particular, per spec it should be true for shadow hosts that contain focused elements. I think it should only choose the deepest focused node.

css-scroll-anchoring-1/Overview.bs
emilio: Wording looks ok, it's just very fishy to reference something that is not specified in a spec like this.

css-scroll-anchoring-1/Overview.bs
```diff
@@ -85,16 +86,41 @@ An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
...
 	The <dfn id="anchoring-algorithm">anchor node selection algorithm</dfn>
 	for a scrolling box |S| is as follows:
 
-	1. If |S| is associated with an element
-		whose computed value of the 'overflow-anchor' property is ''overflow-anchor/none'',
-		then do not select an anchor node for |S|.
-	2. Otherwise, for each DOM child |N| of the element or document associated with |S|,
-		perform the <a>candidate examination algorithm</a> for |N| in |S|,
-		and terminate if it selects an anchor node.
+  1. If |S| is associated with an element whose computed value of the
```
emilio, line 114: Is the indentation change here intentional?

css-scroll-anchoring-1/Overview.bs
vmpstr: Yeah, I agree. I think it's worthwhile for us to at least file an issue to specify the find-in-page algorithm in some way, since it is a common enough operation

css-scroll-anchoring-1/Overview.bs
vmpstr: Yeah makes sense, I've updated it to reference https://html.spec.whatwg.org/multipage/interaction.html#dom-document-activeelement instead which I think is exactly what we're looking for here.

css-scroll-anchoring-1/Overview.bs
vmpstr: No, this was me struggling with bikeshed :) I undid the unintentional changes.

css-scroll-anchoring-1/Overview.bs
```diff
@@ -97,7 +97,7 @@ following criteria:
 
 Some elements are considered to be <dfn
 id="anchor-priority-candidates">priority candidates</dfn> for anchor selection:
-  1. An element that <a href="https://html.spec.whatwg.org/multipage/semantics-other.html#selector-focus">has focus</a>.
+  1. An element that <a href="https://html.spec.whatwg.org/multipage/interaction.html#dom-document-activeelement">has focus</a>.
```
emilio, line 100: Probably also not what you want to link to, for two reasons:

 * It retargets shadow DOM nodes, so you avoid anchoring focused nodes inside shadow trees and anchoring to the host instead.

 * It returns the `<body>` element, which means that if you have no focus, you only anchor to the body, which is definitely not what you want.

vmpstr: I've spent some time looking for the focused element I mean, and I'm starting to realize that focus is difficult.

I think this might be the another definition here:
https://www.w3.org/TR/html51/editing.html#focused

It does seem to talk about retargetting to the document, but it seems that it's meant to ensure that the focused element is the deepest element that matches the given criteria. It does seem to default to the document if there is no selection.

I think the path forward here is to point to one of these focus definitions but also say that we should exclude the body/document from being a potential priority candidate.

Or maybe just explicitly say that it is a deepest element that has focus (ie responds to the :focus selector)

What do you think?

emilio: We shouldn't link to the w3c html spec. The whatwg version is the right one :)

We should probably just say that this is "The DOM anchor of the focused area of the document.", where "focused area" is https://html.spec.whatwg.org/multipage/interaction.html#focused-area-of-the-document and "DOM anchor" is https://html.spec.whatwg.org/multipage/interaction.html#dom-anchor.

That's effectively the unretargeted active element, without falling back to `<body>`. Now that's quite a mouthful. Maybe @annevk / @domenic can export that in a way that it has a nicer name from html?

vmpstr: Thanks for helping me out with the text!

I got hung up on "focused element" as opposed to "focused area". The text you have makes sense, and I've updated to that.

css-scroll-anchoring-1/Overview.bs
```diff
@@ -85,16 +86,43 @@ An anchor node can be any <a>box</a> except one for a non-<a>atomic inline</a>.
...
+following criteria:
+  * |C| is an element that is not a non-<a>atomic inline</a>.
+  * |C| is <a>partially visible</a> or <a>fully visible</a> in |S|
+  * |C| is a descendant of |S|
+  * |C| is not in an <a>excluded subtree</a>
+  * None of the ancestors of |C| up to |S| are in an <a>excluded subtree</a>
+
+Some elements are considered to be <dfn
+id="anchor-priority-candidates">priority candidates</dfn> for anchor selection:
+  1. The <a href="https://html.spec.whatwg.org/multipage/interaction.html#dom-anchor">DOM anchor</a>
```
emilio, line 100: I think using bikeshed there may be a way to just use `<a>DOM anchor</a>` and so, importing the definitions from HTML.

vmpstr: I've updated this (at the cost of adding an anchors block at the top, but I think the change is worthwhile since it keeps the .bs file readable).

vmpstr:
I removed the insertion caret from the priority lists, since I think having the focused element be prioritized over it largely handles it. In other words, I'm failing to come up with any reasonable example that has an insertion caret but not a focused element.

emilio:
Caret browsing in Gecko?

vmpstr:
It was specified as insertion caret here in the text here: https://drafts.csswg.org/css-ui-3/#propdef-caret-color

I think the Gecko caret browsing is what the non-normative text refers to as a navigation caret. We can definitely add that back in, but perhaps with different wording

Issue #5096, opened 2020-05-21
[web-animations-1] Non-normative mention of "iterationCount" instead of "iterations"


graouts:
There is a non-normative paragraph in the Web Animations spec that mentions `iterationCount` as a key used for the object passed to the `updateTiming()` method. This should be `iterations`!

PR #5097, opened 2020-05-21
[web-animations-1] Non-normative mention of "iterationCount" instead of "iterations"

graouts:
[web-animations-1] There is no `iterationCount` property but there is an `iterations` property. Fix some non-normative text to use the correct property name. This addresses issue #5096.


birtles:
Thanks!

PR #5108, opened 2020-05-24
[css-values-4] initial work on vhc unit proposal text. Resolves #4329 

### LINKED ISSUE #4329: [css-values-4] Add vhc value
## Latest info

Added `lvh`, `svh`, `dvh`, `lvw`, `svw`, and `dvw` units.

See https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-863677668 and https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-880040513 for more details. 🎉 

Original content below:
## Background

`vh` is [defined](https://drafts.csswg.org/css-values-4/#viewport-relative-lengths) as `Equal to 1% of the height of the initial containing block.` Perhaps the current **behaviour**<sup> [1](https://bugs.webkit.org/show_bug.cgi?id=141832) [2](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/BK0oHURgmJ4)</sup> could also be defined as
>Equal to 1% of the height of the initial containing block _with user agent chrome minimized_.

In other words, on devices where the browser chrome changes size (e.g. mobile devices), `100vh` is actually larger than the viewport when the browser chrome is maximized, and thus overflows.

A brief history of the `vh` unit is outlined here https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-542420036

## Proposal

`vhc` (c = "with chrome", but it could be something else) could be defined as 
>Equal to 1% of the height of the initial containing block, with user agent chrome maximized

On devices without a changing chrome size (e.g. desktop devices), `1vh === 1vhc`.

## Advantages and Drawbacks

[Pros and Cons list for each proposal](https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-547551485)

Original content of this section:

A drawback of `vh` units is that content will be cut off when (1) you load a page and (2) when you scroll upwards, while the content fits when you scroll down and the chrome is minimized. 

The `vhc` unit would be the inverse: content would fit the page when (1) you load the page and (2) when you scroll upwards, but you would see additional content (or whitespace, depending on the implementation) when the chrome is minimized.

It also would provide a better experience than `vh` for games and other full-screen content that doesn't or shouldn't scroll at all.

In the end, this proposal does not completely solve the issue of `1vh` needing to be different values at different times. However, it _does_ at least give the developer a choice in which value that they want to use.

(For what it's worth, my personal preference is that I would end up using `vhc` units for responsive designs, because I would rather have additional content visible or some whitespace added, rather than have content cut off and not visible.)

## Alternative Proposals

* [css env variables](https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-542971506)
* [vh-sizing](https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-543344105)

And [a breakdown](https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-547551485) of the pros and cons of each proposal so far

## Current Workarounds

As it stands, web developers that want to have a full-height website are either reliant on javascript <sup> [1](https://css-tricks.com/the-trick-to-viewport-units-on-mobile/) [2](https://stackoverflow.com/a/43575432) [3](https://github.com/mvasin/react-div-100vh)</sup> to get `vh` units to not cut off content, or just tend to avoid <sup>[1](https://chanind.github.io/javascript/2019/09/28/avoid-100vh-on-mobile-web.html)</sup> using `vh` units altogether.

## Unrelated

This is my first time proposing, so if I did something wrong or need to improve something please let me know! Thank you for your patience. :)

frehner:
Resolves #4329 

I don't think `vhc` is (or should be) the final unit's name; however, I did want to start working on the text for this new unit, and then we can swap out once that's decided.

This is my first time proposing changes to the spec, so I probably did something wrong. Let me know and I'll fix it :)

### DIFF
--- a/css-values-3/Overview.bs
+++ b/css-values-3/Overview.bs
@@ -1005,7 +1005,9 @@ Relative Lengths</h3>
 		<tr><td>''vw''
 			<td>1% of viewport's width
 		<tr><td>''vh''
-			<td>1% of viewport's height
+			<td>1% of viewport's height when user agent chrome is minimized
+		<tr><td>''vhc''
+			<td>1% of viewport's height when user agent chrome is maximized
 		<tr><td>''vmin''
 			<td>1% of viewport's smaller dimension
 		<tr><td>''vmax''
@@ -1178,7 +1180,13 @@ Viewport-percentage Lengths: the ''vw'', ''vh'', ''vmin'', ''vmax'' units</h4>
 
 		<dt><dfn id="vh" lt="vh">vh unit</dfn>
 		<dd>
-			Equal to 1% of the height of the initial containing block.
+			When user agent chrome does not change size, it is equal to 1% of the height of the initial containing block.
+			When user agent chrome does change size, it is equal to 1% of the height of the initial containing block with the user agent chrome at its smallest size.
+			
+		<dt><dfn id="vhc" lt="vhc">vhc unit</dfn>
+		<dd>
+			When user agent chrome does not change size, it is equal to 1% of the height of the initial containing block.
+			When user agent chrome does change size, it is equal to 1% of the height of the initial containing block with the user agent chrome at its largest size.
 
 		<dt><dfn id="vmin" lt="vmin">vmin unit</dfn>
 		<dd>
@@ -1188,6 +1196,10 @@ Viewport-percentage Lengths: the ''vw'', ''vh'', ''vmin'', ''vmax'' units</h4>
 		<dd>
 			Equal to the larger of ''vw'' or ''vh''.
 	</dl>
+	
+	Note: The vh and vhc units are the same size when the user agent chrome does not change size
+	(for example, desktop user agents generally do not change chrome size, so vh and vhc will be the same in these situations).
+	For user agent whose chrome does change size, vhc is the smaller of the two units.
 
 	<wpt>
 	css/css-regions/interactivity/resizing/regions-resizing-003.html


astearns:
@tabatkins @fantasai 

@frehner you should have gotten an email from someone at the W3C about intellectual property commitments

frehner:
>@frehner you should have gotten an email from someone at the W3C about intellectual property commitments

I haven't received anything yet as far as I can tell. I also just checked my spam folder just in case

frehner:
I created and linked my w3c account to my github account. However, I'm not entirely sure if I need to do more (e.g. join the csswg? and can I do that if I'm not part of a sponsor company?)

brunoais:
For the name, I like the `vvh` [@bmakuh](https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-661351346) suggests.

johannesodland:
> For the name, I like the `vvh` [@bmakuh](https://github.com/w3c/csswg-drafts/issues/4329#issuecomment-661351346) suggests.

Won't vvh be confusing, if it refers to the visual viewport? 
The visual viewport is what's visible on the screen with or without chrome and even after pinch-zoom. 
The new viewport unit refers 1% of the screen height "with chrome".


When chrome is hidden, the visual viewport will refer to the screen with chrome hidden, but vhc will still refer to 1% of the height with chrome. 
When the user pinch-zooms the visual viewport will refer to the area that is visible on the screen. Vhc will still refer to 1% of the height with chrome.

brunoais:
I don't understand what you are mentioning about `vvh` as Visual Viewport Height being confusing.

johannesodland:
> I don't understand what you are mentioning about `vvh` as Visual Viewport Height being confusing.

visual viewport vs vu

The draft is clarifying that 1 `vh` equals 1% of the height of the initial containing block with the user agent chrome at its smallest size. 1 `vhc` is defined as 1% of the height of the initial containing block with the user agent chrome at its largest size. 

Both values are defined in terms of the layout / fixed viewport.

Above there are three screenshots from a simulated iphone.

**Left: Chrome at smallest size**
On the left the viewport has user agent chrome at its smallest size.
Here the visual viewport height is equal to `100vh` (749px)

**Middle: Chrome at largest size**
In the middle the viewport has user agent chrome at its largest size.
Here the visual viewport height is equal to `100vhc` (635px)

**Right: With scale**
On the right the viewport has been scaled. (The viewport can be scaled either through pinch zooming, or by focusing on an input element.) The visual viewport height is now 437px, and not equal to `100vh` nor `100vhc`.


The height of the visual viewport will some times be equal `100vh` and other times be equal to `100vhc`. When scaled it will be a different value all together. Naming one of the units after the visual viewport is confusing, as the visual viewport changes and will equal either of the units at different states.

Maybe @bokand can weigh in here?

brunoais:
@johannesodland That doesn't seem to be my interpretation of how `vhc` should work.
In my interpretation, `vhc` will adjust including and between the 749px and 635px depending on the browser's actual visual viewport size. It's those and everything in between so it also reflects when the chrome is, for example, only half size extended.

Is our conflicting interpretations the confusion part? Because if what happens is actually your interpretation, this new property being suggested is, for the most part, barely useful at all.

johannesodland:
@brunoais I think our interpretations are definitely different. 

This PR specs a vhc unit as folows:

> **vhc unit** 
When user agent chrome does not change size, it is equal to 1% of the height of the initial containing block.
When user agent chrome does change size, it is equal to 1% of the height of the initial containing block with the user agent chrome at its largest size.
			
			
As I interpret it it, this viewport unit is defined relative to the ICB and will only change when a window is resized and the ICB changes size. It is stable and does not change when UA chrome changes size. (It will be 635px on the iPhone in my example, as long as the phone is not rotated.)

If the unit changed with the UA chrome we would get all the reflow issues that we had with vh earlier.

Having this unit in addition to vh is very useful, at least for us. It allows us to use css to style content that fits inside the browser chrome without using JS.

Viewport units matching the visual viewport could be useful too as a way to reference the [visual viewport API](https://wicg.github.io/visual-viewport/index.html) values in CSS, but as I read the original issue and the PR this unit is defined relative to the ICB.

frehner:
> @brunoais I think our interpretations are definitely different.
> 
> This PR specs a vhc unit as folows:
> 
> > **vhc unit**
> > When user agent chrome does not change size, it is equal to 1% of the height of the initial containing block.
> > When user agent chrome does change size, it is equal to 1% of the height of the initial containing block with the user agent chrome at its largest size.
> 
> As I interpret it it, this viewport unit is defined relative to the ICB and will only change when a window is resized and the ICB changes size. It is stable and does not change when UA chrome changes size. (It will be 635px on the iPhone in my example, as long as the phone is not rotated.)
> 
> If the unit changed with the UA chrome we would get all the reflow issues that we had with vh earlier.
> 
> Having this unit in addition to vh is very useful, at least for us. It allows us to use css to style content that fits inside the browser chrome without using JS.
> 
> Viewport units matching the visual viewport could be useful too as a way to reference the [visual viewport API](https://wicg.github.io/visual-viewport/index.html) values in CSS, but as I read the original issue and the PR this unit is defined relative to the ICB.

This is the correct interpretation of the vhc unit. Having a value that changes causes issues with rendering; see the discussion in the issue for details.

brunoais:
Ok. Definitely my bad. Then I guess I will try to come up with another proposal that does just that.

frehner:
> Ok. Definitely my bad. Then I guess I will try to come up with another proposal that does as I remembered.

Sounds great! When you make it, link to it from the original issue if you don’t mind! I would love to follow it.

johannesodland:
@frehner [CSS Values 4](https://www.w3.org/TR/css-values-4/#viewport-relative-lengths) has logical viewport-percentage lengths `vb` and `vi` as well. Should these be considered as well?

frehner:
> @frehner [CSS Values 4](https://www.w3.org/TR/css-values-4/#viewport-relative-lengths) has logical viewport-percentage lengths `vb` and `vi` as well. Should these be considered as well?

Seems like they should, yes. I'm not sure if here or the related issue would be the place to bring that up with the WG though.

johannesodland:
> Seems like they should, yes. I'm not sure if here or the related issue would be the place to bring that up with the WG though.

You’re right. I’ll bring it up in th issue 👍🏼

fantasai:
Hi @frehner ! @tabatkins and I committed https://github.com/w3c/csswg-drafts/commit/0f78ba4ee462bf3ea879a122789dc153780b88f6 as a replacement of this PR. Not that there was anything specifically wrong with yours, it was a good set of edits! But we needed a lot more detail in the spec, and also a bit of restructuring seemed necessary, so I'm going to close out this PR without accepting. We did add your name to the Acknowledgements section, though, to say thanks for your efforts on this issue. ;) You did a great job of channeling the discussion and helping to make this happen. <3

frehner:
> Hi @frehner ! @tabatkins and I committed [0f78ba4](https://github.com/w3c/csswg-drafts/commit/0f78ba4ee462bf3ea879a122789dc153780b88f6) as a replacement of this PR. Not that there was anything specifically wrong with yours, it was a good set of edits! But we needed a lot more detail in the spec, and also a bit of restructuring seemed necessary, so I'm going to close out this PR without accepting. We did add your name to the Acknowledgements section, though, to say thanks for your efforts on this issue. ;) You did a great job of channeling the discussion and helping to make this happen. <3

Thank you both so much; I’m excited to see progress, and honestly it it makes me a little giddy to see my name in there as well. 😃

Issue #5100, opened 2020-05-22
[css-backgrounds] Order of values for border shorthands is unclear


zjopy:
The definitions of the order of values for the shorthands `border-color`, `border-style` and `border-width` seem a bit unclear. For all three the [spec](https://www.w3.org/TR/css-backgrounds-3/#the-border-color) states

> If it has four values, they set top, right, bottom and left in that order. If left is missing, it is the same as right; if bottom is missing, it is the same as top; if right is missing, it is the same as top.

So what happens with three or two values, e.g. `border-width: a b c` or `border-width: a b`? How does it know which side I left out?

By trial and error I know the order is the same as for the `margin` and `padding` shorthands. The definitions of `margin` and `padding` are much clearer though. The [spec](https://www.w3.org/TR/css-box-4/#margin-shorthand) states

> If there is only one component value, it applies to all sides. If there are two values, the top and bottom (padding) are set to the first value and the right and left (padding) are set to the second. If there are three values, the top is set to the first value, the left and right are set to the second, and the bottom is set to the third.

I would suggest taking over this part of the definition from `margin` / `padding` for the border properties as well.

Loirooriol:
The full syntax is like `border-width: top right bottom left`.
If you only provide 3 values (`border-width: top right bottom`), then left is missing.
If you only provide 2 values (`border-width: top right`), then bottom and left are missing.
If you only provide 1 value (`border-width: top`), then right, bottom and left are missing.

So you know which sides are missing because you match greedily, sides can only be omitted at the end.

zjopy:
Ok thanks, I understand now. The sentence assumed that for each missing value the ones before were also missing. So a somewhat clearer formulation would be

> If left is missing, it is the same as right; if ALSO bottom is missing, it is the same as top; if ALSO right is missing, it is the same as top.

 Still the `margin` / `padding` definitions seem the clearest.

PR #5116, opened 2020-05-26
[css-background-3] Clarify order of values for border shorthand #5100

zjopy:
Addresses #5100.


astearns:
@fantasai @bradkemper this looks completely editorial to me, but I don't have an opinion on whether this way of describing omitted values is more or less clear.

Issue #1751, opened 2017-08-23
[mediaqueries-4 ] Deprecate 'speech' media type as well?


patrickhlauke:
Possibly too late now for MQ4, but just wondering if there is any known support in the wild for the "speech" media type https://www.w3.org/TR/mediaqueries-4/#media-types

To my knowledge, the answer is no...which makes me wonder if that should also be deprecated along `braille`, `aural` and co.

patrickhlauke:
FWIW https://twitter.com/patrick_h_lauke/status/900325928076816384 coming up dry so far...

frivoal:
Thanks for bringing this to my attention.

The current definition of `speech` is lousy, as screen readers, which it claims should match it, are a better fit for `screen` (plus some extra capabilities), since they do work on the result of a visual layout. If that's really what it's for, it should be deprecated, because it is wrong.

If we want to use it for a set of user agents completely disjoint from visual UAs (e.g. “Siri/Alexa,  read me the wikipedia page about strawman arguments”), then it would make some more sense as a media type. Which still leaves us with a question of whether anyone implements that.

Either way, it should not stay the way it is.

@tabatkins Any thoughts? Who do we get in touch with to know if they are interested in this for pure-speach UAs (not screen readers)?

patrickhlauke:
> The current definition of speech is lousy, as screen readers, which it claims should match it, are a better fit for screen (plus some extra capabilities), since they do work on the result of a visual layout. If that's really what it's for, it should be deprecated, because it is wrong.

Yup, and also noting that screen readers are also used not just by 100% blind users, but those with partial sight as well, and even users with good eyesight but cognitive issues. Plus of course the fact that SRs sit on top of browsers, and the browsers clearly behave as `screen` user agents (although they may be able to detect there's AT running, they don't switch to now pretending to be `speech` devices, which would be wrong anyway per the previous bit about SR use).

tabatkins:
I only kept it because it was supposedly what things that used CSS Speech should identify as.  If there's no actual use-cases (or the use-cases are all merely theoretical, and all UAs in reality just report "screen" and assistive agents are ok with this), then I'm totally fine with dropping it and just going with the screen/print dichotomy.

frivoal:
[...]

PR #5140, opened 2020-05-30
[css-speech-1] Rephrase background information about media types

patrickhlauke:
Since the `speech` media type has now also been deprecated (see https://github.com/w3c/csswg-drafts/issues/1751#event-3381393863), this clarifies that the CSS properties defined for CSS Speech Module apply to all media types, and that they cannot be scoped in the way that was described


frivoal:
Seems reasonable to me

PR #5149, opened 2020-06-02
[css-inline-3] Extend “applies to” for SVG text

### LINKED ISSUE #3411: [css-inline] CSS properties appply to SVG elements
Some of the properties in CSS inline apply to certain SVG elements as well. According to SVG2:
* `alignment-baseline` to `tspan` and `textPath` elements
* `baseline-shift` to `tspan` and `textPath` elements
* `dominent-baseline` to [text content elements](https://svgwg.org/svg2-draft/text.html#TermTextContentElement)
* `inline-sizing` to `text` elements (at-risk for SVG 2)
* `line-height`  to [text content elements](https://svgwg.org/svg2-draft/text.html#TermTextContentElement) (SVG2 actually misses to specify it) CC @AmeliaBR 

Those are the properties currently covered by SVG's text algorithms. Some of the other properties in CSS inline layout should apply to certain SVG elements as well in the future.

AmeliaBR:
Closes #3411
See issue for discussion/reasoning & differences from SVG 1.

Bikeshed error that needs fixing:
the “current text position” definition from SVG text
is not being exported in a way that Bikeshed is picking up,
even with hints about which spec to look for it in.
(@TabAtkins, can I fix this here, or do I need to add some attributes to the SVG spec?)

Changes should be copied over to css-inline-4 after review.

### DIFF
--- a/css-inline-3/Overview.bs
+++ b/css-inline-3/Overview.bs
@@ -32,6 +32,8 @@ At Risk: the 'initial-letter-wrap' property
 spec:css-align-3; type:dfn; text:alignment baseline
 spec:css-box; type:dfn; text:content area
 spec:css-break-3; type:dfn; text:fragment
+spec:css-shapes-1; type:property; text:shape-margin
+spec:svg2; type:dfn; text: current text position
 </pre>
 
 <h2 id="intro">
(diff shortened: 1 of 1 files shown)


tabatkins:
"current text position" is definitely present in the db and linkable, but it's not exported, so you'll need to specify the spec to override that and link it anyway.

You can either do that inline `<l spec=svg2>[=current text position=]</l>` or by putting it in `<pre class=link-defaults>`.

AmeliaBR:
> or by putting it in `<pre class=link-defaults>`

I _had_ tried that, but I hadn't used the correct shortname for the SVG2 spec. Fixed now.

@fantasai, ready to merge if you're OK with how I handled the properties with inconsistent implementations. Note that this is only the edits to CSS Inline 3, you'll want to port to Inline 4

AmeliaBR:
> @fantasai, ready to merge if you're OK

Actually, NOW it's ready to merge, after I pulled in your latest edits & fixed the merge conflicts…

You probably want to review carefully.

fantasai:
Sorry, thanks! ^_^

PR #5186, opened 2020-06-08
[scroll-animations-1][web-animations-1] Added phase calculation for ScrollTimeline #4325

JTensai:
I added section describing how scroll timeline phase is calculated. Inclusive end exception that was removed in a previous PR has been added back in through the phase calculation (an explanatory note for the exception has been included). 

I also reformatted the section for calculating current time to match the new phase calculation formatting.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -709,22 +709,23 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
 
 ### The phase of a {{ScrollTimeline}} ### {#phase-algorithm}
 
-The [=phase=] of a {{ScrollTimeline}} is calculated from the first matching
-condition from below:
+The [=phase=] of a {{ScrollTimeline}} is calculated as follows:
+
+1.  If <em>any</em> of the following are true:
+
+    * {{source}} is null, or
+    * {{source}} does not currently have a [=CSS layout box=], or
+    * {{source}}'s layout box is not a [=scroll container=], or
+    * [=effective scroll range=] is null.
+
+    :  return [=timeline inactive phase|inactive=] and abort remaining steps.
+
+1.  Let <var>current scroll offset</var> be the current scroll offset of
+    {{source}} in the direction specified by {{orientation}}.
+
+1.  Return the result corresponding to the first matching condition from below:
 
 <div class="switch">
-
-:   If <em>any</em> of the following are true:
-
-    1.  {{source}} is null, or
-    1.  {{source}} does not currently have a [=CSS layout box=], or
-    1.  {{source}}'s layout box is not a [=scroll container=], or
-    1.  [=effective scroll range=] is null
-
-::  return [=timeline inactive phase|inactive=]
-
-:   Otherwise, let <var>current scroll offset</var> be the current scroll offset
-    of {{source}} in the direction specified by {{orientation}}.
 
 :   If <var>current scroll offset</var> is less than [=effective start offset=]:
 ::  return [=before phase=]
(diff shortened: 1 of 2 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -648,6 +653,27 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
...
+
+1.  Otherwise, let <var>current scroll offset</var> be the current scroll offset
+    of {{source}} in the direction specified by {{orientation}}.
+
+1.  If [=effective scroll range=] is null, return [=inactive phase=].
+
+1.  If <var>current scroll offset</var> is less than [=effective start offset=],
+    return [=before phase=].
+
+1.  If <var>current scroll offset</var> is greater than or equal to [=effective
```
ogerchikov, line 672: Please add end inclusivity condition (https://github.com/w3c/csswg-drafts/issues/4325#issuecomment-606723725).

scroll-animations-1/Overview.bs
```diff
@@ -648,6 +653,27 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
...
+
+The [=phase=] of a {{ScrollTimeline}} is calculated as follows:
+
+1.  If {{source}} is null, does not currently have a [=CSS layout box=], or if
+    its layout box is not a [=scroll container=], return [=inactive phase=].
+
+1.  Otherwise, let <var>current scroll offset</var> be the current scroll offset
+    of {{source}} in the direction specified by {{orientation}}.
+
+1.  If [=effective scroll range=] is null, return [=inactive phase=].
```
ogerchikov, line 667: I suggest including this into first condition.

scroll-animations-1/Overview.bs
```diff
@@ -648,35 +653,72 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
...
+
+    1.  {{source}} is null, or
+    1.  {{source}} does not currently have a [=CSS layout box=], or
+    1.  {{source}}'s layout box is not a [=scroll container=], or
+    1.  [=effective scroll range=] is null
 
-1.  Otherwise, let <var>current scroll offset</var> be the current scroll offset
+::  return [=timeline inactive phase|inactive=]
+
+:   Otherwise, let <var>current scroll offset</var> be the current scroll offset
```
birtles, line 673: Something is amiss with the markup here. I'm not sure what it is supposed to look like but it currently renders like this:

image

Might I suggest the first step should be an ordered list with a nested unordered list? e.g.

1. If any of the following are true:
    * source is null, or
    * source does not currently have a CSS layout box, or
    * source's layout box is not a scroll container, or
    * effective scroll range is null

    return inactive and abort these steps.
2. Let current scroll offset be the current scroll offset of source in the direction specified by orientation.
3. Return the result corresponding to the first matching condition from below:
...

scroll-animations-1/Overview.bs
```diff
@@ -648,35 +653,72 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
...
 
-1.  If <var>current scroll offset</var> is greater than or equal to [=effective
-    end offset=], return [=effective time range=].
+### The current time of a {{ScrollTimeline}} ### {#current-time-algorithm}
 
-1.  Return the result of evaluating the following expression:
+The [=current time=] of a {{ScrollTimeline}} is calculated from the first
+matching condition from below:
+
+<div class="switch">
```
birtles, line 694: Similarly, I think this could be reworked to be easier to follow using a similar approach to the phase section.

scroll-animations-1/Overview.bs
JTensai: I based the formatting on https://drafts.csswg.org/web-animations/#the-current-time-of-an-animation but I like your suggestion better. I'll update it in the next commit.

scroll-animations-1/Overview.bs
```diff
@@ -648,35 +653,75 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
...
+The [=phase=] of a {{ScrollTimeline}} is calculated as follows:
+
+1.  If <em>any</em> of the following are true:
+
+    * {{source}} is null, or
+    * {{source}} does not currently have a [=CSS layout box=], or
+    * {{source}}'s layout box is not a [=scroll container=], or
+    * [=effective scroll range=] is null.
+
+    :  return [=timeline inactive phase|inactive=] and abort remaining steps.
```
birtles, line 668: Nit: I don't think the `:` here is necessary?

scroll-animations-1/Overview.bs
```diff
@@ -648,35 +653,75 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
...
 
-1.  Otherwise, let <var>current scroll offset</var> be the current scroll offset
-    of {{source}} in the direction specified by {{orientation}}.
+    * {{source}} is null, or
+    * {{source}} does not currently have a [=CSS layout box=], or
+    * {{source}}'s layout box is not a [=scroll container=], or
+    * [=effective scroll range=] is null.
 
-1.  If [=effective scroll range=] is null, return an unresolved time value.
+    : return an unresolved time value and abort remaining steps.
```
birtles, line 701: Likewise here.

scroll-animations-1/Overview.bs
```diff
@@ -648,35 +653,86 @@ The procedure to calculate <dfn>effective scroll range</dfn> of a
...
 
-1.  If <var>current scroll offset</var> is greater than or equal to [=effective
-    end offset=], return [=effective time range=].
+:   If <var>current scroll offset</var> is greater than or equal to [=effective
+    end offset=]:
+::  The [=current time=] is the [=effective time range=].
 
-1.  Return the result of evaluating the following expression:
+:   Otherwise, the [=current time=] is the result of evaluating the following
+    expression:
```
birtles, line 729: Nit:

```
:   Otherwise
::  The [=current time=] is the result of evaluating...
```

(i.e. split up the line so it is actually a `<dt>`/`<dd>` pair)

PR #5193, opened 2020-06-11
[scroll-animations] Add CSSScrollTimelineRule

andruud:


andruud:
cc @majido

scroll-animations-1/Overview.bs
```diff
@@ -804,6 +804,52 @@ instead of document's scrolling Element
...
+  </dd>
+
+  <dt><dfn>end</dfn></dt>
+  <dd>
+    The 'end' descriptor associated with the ''@scroll-timeline'', or "auto" if not specified.
+  </dd>
+
+  <dt><dfn>scrollTimeRange</dfn></dt>
+  <dd>
+    The 'scroll-time-range' descriptor associated with the ''@scroll-timeline'', or "auto" if not specified.
```
majido, line 849: Ops 'scroll-time-range' is a mistake in the spec that needs to be fixed. It was missed in the patch to [update the syntax](https://github.com/w3c/csswg-drafts/commit/7fd843a197f9515985b09adb5ad3018f390c44b7#")

I would be 'time-range'. But I am happy to merge this and fix all instances together.

PR #5194, opened 2020-06-11
[biblio.ref] Update URLs of css-backgrounds-4 and css-page-template-1

tidoust:
This update adjusts the URLs of the entries for CSS Backgrounds and Borders Module Level 4 and CSS Pagination Templates Module Level 3, so that they target the final document directly; and refreshes the list of authors for CSS Backgrounds and Borders Module Level 4.

### DIFF
--- a/biblio.ref
+++ b/biblio.ref
@@ -879,8 +879,8 @@
 %T CSS Pagination Templates Module Level 3
 %A Alan Stearns
 %S Proposal for a CSS module
-%O (Retrieved 4 April 2012)
-%U https://drafts.csswg.org/css3-page-template/
+%O (Retrieved 11 June 2020)
+%U https://drafts.csswg.org/css-page-template-1/
 
 %L CSS3-POSITIONING
 %T CSS Positioned Layout Module Level 3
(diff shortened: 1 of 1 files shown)


svgeesus:
Thanks, but does anything use the ancient and unmaintained biblio ref file?

Bikeshed uses [tobie/specref](https://github.com/tobie/specref) and CSS2 is in process of being updated to use Bikeshed so I don't think this file is used anymore.

tidoust:
Well, Specref uses this "ancient and unmaintained" biblio ref file to build entries in [csswg.json](https://github.com/tobie/specref/blob/master/refs/csswg.json) (and notably `CSS3-PAGE-TEMPLATE` and `CSS4BACKGROUND`) through the [csswg.js script](https://github.com/tobie/specref/blob/master/scripts/csswg.js) that runs every few hours.

Specref ignores most of the entries in `biblio.ref` because they are already defined elsewhere, typically because the specs have already been published to /TR/. It still relies on the info in this file for some entries.

svgeesus:
Looking at the [biblio.ref history](https://github.com/w3c/csswg-drafts/commits/master/biblio.ref) with 4 edits in the last 4 years, syncing every few hours seems excessive. 

I'm okay to merge this, but it seems better long-term to just add directly to specref.

I guess I should chat with @tobie about this.

PR #5196, opened 2020-06-11
Convert CSS2 to Bikeshed

### LINKED ISSUE #2220: Convert CSS2 to use Bikeshed
I would like to convert CSS2 to Bikeshed, for the following reasons:

 * Stylistic consistency with other CSS specs (for the sake of readability, etc., especially including the xref box to other refs)
 * Making it easier to xref things to/from other Bikeshed specs
 * Getting rid of generated code from the repo (i.e., the built copy of the spec)

Risks:

 * Accidental changes to normative text
 * Not being able to generate a multipage version and breaking xref links to it (I think we can do this with post-processing with not too much difficulty, though)

https://github.com/gsnedders/csswg-drafts/commits/css2-bs is some vague start towards converting it, albeit with 488 linking errors, and obviously bogus references to other later CSS modules.

I'm happy to do further work on converting it if the group thinks this is useful.

gsnedders:
Fixes #2220.

While in total this is a big change, each of these commits individually is comparatively small and reviewable. The only exception is 82079cb, but the main change there is "just" concatenating all the *.src files, so should be easily reproducible.


plinss:
I'm a bit concerned about concatenating all the files.

I accept that Bikeshed doesn't support multipage specs yet, but I expect it to eventually. And CSS2 is large enough that there's benefit to having both single page and multipage versions. It depends on how Bikeshed will handle multipage specs, whether it will be able to split a single source file to produce a multipage version or it will concatenate multiple sources itself to produce the single page version. And it's likely too soon to know which of those approaches it may require.

There's also author and reviewer pain in having a single large source file rather than a number of more manageable files.

I'm wondering if, in the meanwhile, it would be better to preserve the individual source files and have a singe additional file that's nothing but a set of includes.

cc @tabatkins for any insights into how Bikeshed may wind up supporting multipage specs.

gsnedders:
@plinss by way of comparison, the ten largest `Overview.bs`:

```
156K    css-color-4/Overview.bs
156K    selectors-4/Overview.bs
172K    css-values-4/Overview.bs
176K    css-tables-3/Overview.bs
228K    css-flexbox-1/Overview.bs
240K    css-grid-1/Overview.bs
256K    css-text-3/Overview.bs
260K    web-animations-1/Overview.bs
264K    css-fonts-4/Overview.bs
664K    css2/Overview.bs
```

plinss:
@gsnedders good data, but if anything that tells me there are more specs that we should consider having multipage versions of when Bikeshed can handle it (and maybe multipage support should get bumped in priority). 

To be clear, I'm perfectly fine with re-releasing CSS2 as only a single page spec until Bikeshed gets multipage support. I highly value the work you're doing here, and think the value in the Bikeshed conversion outweighs the pain of a large single page spec for those that read it (though it may violate the priority of constituencies in some regards, the improved cross-linking and consistent presentation offsets that).

I just want to be sure that when Bikeshed does support multipage, we're not in a situation where converting CSS2 back into a multipage spec is too much of a burden. If Bikeshed will be able to support auto-splitting a single input file, then this is mostly a non-issue (except for potential editor/reviewer pain of dealing with a large source file). Keeping the source files split just feels like keeping our options open. I accept that it could also cause editor pain if updates often need to span files, but I don't expect it to get updated much after the conversion. 

All in all, please take this for what it is, a concern, not an objection.

tabatkins:
My intent is indeed that multipage output will be supported from a single source file; having each "page" be a separate file that you include in is possible (and probable!), but won't be necessary. So don't take that as a concern here.

----

That said, I do agree that keeping the chapters as separate files which are included together into the final document probably makes for an easier editting experience most of the time (unless you're doing find-and-replace across the whole thing, but even then, a good editor can do that across multiple files in a folder).  And yeah, those very large specs *should* be split; I've thought about doing precisely that for Flexbox, Grid, Color, and Selectors, because they're very difficult to navigate currently.

On the other hand, I'm not the one doing this work, so if it would be a pain to reverse the combining, whatever. ^_^  However, I would at minimum ask for some big-text section separators (<https://tabatkins.github.io/bikeshed/#big-text>) so it can be reasonably navigated by editors with minimaps.

PR #5217, opened 2020-06-16
[scroll-animations] Ignore @scroll-timeline rules in shadow trees

### LINKED ISSUE #5167: [scroll-animations] @scroll-timeline in shadow trees
Should we temporarily spec that `@scroll-timeline` is ignored in shadow trees, and add a note that this will change in the future? (See similar text in css-properties-values-api).

There is currently ongoing work in the WG regarding how @-rules and "name-defining things + references" in general should work in shadow trees, and it's a quite complicated problem to solve.

It doesn't really make sense to attempt to solve this "locally" via the scroll-animations spec, and the feature will be held back for a long time (probably) if we choose to block on this.

cc @majido

andruud:
For now, that is.

Fixes #5167.


andruud:
cc @majido

PR #5255, opened 2020-06-23
[web-animations-1] Specify how to compute property values

### LINKED ISSUE #5125: [css-animations] Resolving dependencies in keyframes
This came up (again) during @kevers-google's in-progress implementation of getKeyframes in Chrome. Since getKeyframes directly exposes computed keyframes, it forces us to deal with how/when dependencies in the keyframe are resolved in more detail.

We should in my opinion take a "broad" approach of thinking about all dependencies, and not specify detailed behavior per case. For example, `1em` has a dependency on `font-size`, and `var(--x)` has a dependency on `--x`. In terms of dependencies, `1em` and `var(--x)` are very similar, so for consistency, these cases should be treated the same way.

Browsers generally don't agree on how interpolation actually happens when complicated dependencies are involved. I won't enumerate all differences, but instead highlight this (well-known) example:

```
@keyframes test {
  from { font-size: 2px; width: 10em; }
  to { font-size: 4px; width: 20em; }
}

div {
  font-size: 10px;
  animation: test 10s -5s linear paused; /* t=0.5 */
}
```

The computed value of `width` on `div` for the different browsers:

 * Firefox: `150px`. (1)
 * Safari: `50px`. (2)
 * Chrome: `45px`. (3)

@kevers-google, @flackr and myself have had several discussion about this recently, and three different ways of dealing with it surfaced in those discussions:

## 1. Resolving against the pre-animated style

In this model, the `em` units would resolve against the would-be computed value of `font-size` without animation effects, i.e. `10px` in the first example. The interpolation for `width` would take place in absolute values, between `100px` and `200px`. For `var()` references, they would behave similarly: the value would be substituted using the pre-animated computed value of the referenced custom property.

I think this solution is a nice one in terms of technical complexity, but I'm not sure if it matches author expectation. 

This appears to be what Firefox is doing (for `em` units).

## 2. Resolving against local values in the keyframe

In this approach, each keyframe is basically a list of declarations that's added to the cascade, and we derive computed values from that. In approximated spec language:

 * Let _base cascade_ be the cascade as it would be without animation/transition affects added.
 * For each specified keyframe:
   * Let _keyframe cascade_ be a copy of the base cascade
   * Add the declarations of the keyframe to the keyframe cascade, at the animations level.
   * Produce a _computed keyframe_ by substituting each value in the specified keyframe with the corresponding computed value from the keyframe cascade.
 * Interpolation then happens between computed keyframes.

In the first example, `em` units would then resolve against the `font-size` specified locally in the keyframe (if present). The interpolation of `width` would be between `20px` and `80px`.

This approach would match my initial expectation as an author, I think. But not sure how practical it would be to implement this. Although Safari has the apparent behavior of this model already, at least for the example I gave.

Note that an interesting side-effect of this suggestion is that !important declarations from the base style can end up in the computed keyframes. 

## 3. Resolving against the (animated) computed value

This is what Chrome is currently doing for `em` units (but not for `var()`). In the example, `width` interpolates in `em`-space from `10em` to `20em`, at the same time as `font-size` interpolates from `2px` to `4px`. The `em` unit is resolved computed-value time (as apposed to before the effect value is added to the cascade), against the animated `font-size` at that time. So Chrome produces `45px` in the first example via `15em * 3px = 45px`.

For `var()` it would mean that, for this example:

```
@keyframes test {
  from { --a: 10px; --b: 30px: width: var(--a); }
    to { --a: 20px; --b: 40px: width: var(--b); }
}

div {
  animation: test 1s;
}
```

The animation would add a value to the cascade that can be illustrated by: `calc(var(--a) + t * (var(--b) - var(--a)))`. In other words, like for `em` resolution, the `var()` resolution would be delayed until computed-value time. (Again, Chrome does not currently do this, but it would be the consistent thing to do in this model).

Not sure what I think about this behavior. On the one hand it produces unnecessarily complex animation behavior, but on the other hand it means the concept of "computed values in keyframes" can be mostly avoided, which has a certain simplicity.

Note that getKeyframes in this model should in my opinion not return the keyframes as if "all property values are replaced with their computed values" as the spec currently suggest, as this would misrepresent the endpoints of the interpolation.

---

So at this point I was hoping for some feedback regarding which approach is better, whether or not the author really cares about this at all, how compatible each solution is w.r.t. web-animations, and whether or not there are other approaches that should be considered. Thanks.

cc @birtles @dbaron

PS: I realize this topic has been discussed many times before, and I apologize for not digging up all prior discussion in advance. (Please add links). At the same time, I hope we can avoid treating prior resolutions as automatically holy and immutable, since they apparently failed (so far) to effectuate any real consensus in browser behavior.

andruud:
When it comes to the term "computed value", the spec is currently
very confused. It appears to be making several references to
"the" computed value (as in [1]), when in reality the intention is
(I believe) to apply the *process* that calculates the computed
value ([1]) to the value specified (locally) in the keyframe. This
process is not described in a reusable way anywhere (I think), so I've
added the algorithm "compute a property value" for this. I'm using
the term "property value", since it's already extensively used in this
spec. Possibly it would be more suited for css-cascade, but I'm
proposing it for this spec until we need it somewhere else.

In general I want to maintain a clear distinction between when we
mean the actual computed value [1], and when we mean some value that
has undergone the "compute a property value" process. Hence, I think
we should avoid referring to a value that has undergone the
"compute a property value" process as "a" computed value, as it's
too easily confused with *the* computed value.

I've included some text and an examples that hopefully clarify
the expected interpolation behavior a little bit. It seemed somewhat
natural to place this in the "Computing property values" section,
since resolving dependencies is part of the process when computing
values.

Fixes #5125.

[1] https://drafts.csswg.org/css-cascade/#computed

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3311,7 +3311,8 @@ result.
   this algorithm.
 
 This algorithm implies that property values specified in keyframes can
-establish order dependencies. When <a>computing a property value</a>, the
+establish order dependencies.
+When [=compute a property value|computing a property value=], the
 <a>computed values</a> of dependencies held by <var>value</var> must be
 calculated <em>first</em>.
 
@@ -3327,8 +3328,8 @@ In this example, in order to <a>compute a property value</a> for
 <code>10em</code>, we must know the <a>computed value</a> of
 'font-size' on the <a>target element</a>, which in turn is determined
 by the <a>effect value</a> for 'font-size', which in turn requires that we
-compute property values for 'font-size'. Hence, computing property
-values are subject to ordering constraints.
+compute property values for 'font-size'.
+Hence, computing property values are subject to ordering constraints.
 </div>
 
 </div>
@@ -3606,8 +3607,9 @@ Note: this procedure permits overlapping <a>keyframes</a>.
     <a>keyframe</a> in <var>keyframes</var> respectively.
 
 <div class="note">
-    Note that <a>computed keyframes</a> are "live": user-agents must behave
-    as if they are recreated every time the <a>effect value</a> is calculated.
+    Note that <a>computed keyframes</a> are "live":
+    user-agents must behave as if they are recreated every time the
+    <a>effect value</a> is calculated.
 
     For example, if there is an ongoing transition on the 'font-size' property
     from <code>10px</code> to <code>20px</code>, a property value specified as
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -3297,13 +3297,49 @@ If the <a>keyframe-specific composite operation</a> for a <a>keyframe</a>
 is not set, the <a>composite operation</a> specified for the
 <a>keyframe effect</a> as a whole is used for values specified in that keyframe.
 
+### Computing property values ### {#computing-property-values}
+
+<div algorithm>
+To <dfn>compute a property value</dfn> given a property <var>property</var>,
+a value <var>value</var>, and an {{Element}} <var>element</var>:
+resolve <var>value</var> according to the "Computed Value" line of the
+<var>property</var>'s definition table, using the <a>computed values</a> of
+<var>element</var> as the context for resolving dependencies, and return the
+result.
+
+  Note: The <a>computed values</a> on <var>element</var> are not affected by
+  this algorithm.
+
+This algorithm implies that property values specified in keyframes can
+establish order dependencies. When <a>computing a property value</a>, the
+<a>computed values</a> of dependencies held by <var>value</var> must be
+calculated <em>first</em>.
+
+<div class="example">
+<pre class="lang-javascript">
+var animation = elem.animate([{ fontSize: '10px', width: '10em' },
+                              { fontSize: '20px', width: '20em' }], 1000);
+animation.currentTime = 500;
+console.log(getComputedStyle(elem).fontSize); // Should be 15px
+console.log(getComputedStyle(elem).width); // Should be 225px
+</pre>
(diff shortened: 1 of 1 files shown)


andruud:
@birtles : Please review. I will try to tackle css-animations/getKeyframes as well, but since web-animations is responsible for the core animation behavior, we have to start here.

(Also cc @kevers-google, @flackr)

web-animations-1/Overview.bs
```diff
@@ -3297,13 +3297,49 @@ If the <a>keyframe-specific composite operation</a> for a <a>keyframe</a>
...
+
+<div class="example">
+<pre class="lang-javascript">
+var animation = elem.animate([{ fontSize: '10px', width: '10em' },
+                              { fontSize: '20px', width: '20em' }], 1000);
+animation.currentTime = 500;
+console.log(getComputedStyle(elem).fontSize); // Should be 15px
+console.log(getComputedStyle(elem).width); // Should be 225px
+</pre>
+In this example, in order to <a>compute a property value</a> for
```
birtles, line 3326: Nit: this link doesn't work for me. I think it needs to be `[=compute a property value|computing a property value=]`.

andruud: Ah, thanks. Fixed.

(Although the broken link didn't appear on the source line you highlighted).

andruud:
> semantic line-breaking

OK, I made that change in the most obvious places, at least.

andruud:
@birtles FYI, I don't have write access, so I can't merge this. (In case you're waiting for me).

birtles:
Sorry, I didn't realize you don't have write access!

PR #5256, opened 2020-06-24
[css-color-5] Fix intermediate color in color-mix(red yellow 65%) example.

dbaron:
The main bug was reported both in a comment in the spec and in
https://github.com/w3ctag/design-reviews/issues/526#issuecomment-649100113

This also fixes dropping of a "9" digit in the red's chroma midway
through the calculations.

Note that the correct intermediate is well out of sRGB.  I just chose to
use it anyway and add a note about such cases.


dbaron:
Though it occurs to me that my "some of the intermediate colors" note might be wrong because it seems likely to be *all* of the intermediate colors.

dbaron:
Though I guess this is inconsistent with the handling of out-of-gamut colors in example 8.

LeaVerou:
We should either do gamut mapping for `--color` or not show anything. We definitely don't want to do clipping.
Also, it appears that there is another problem with this example, but I'll open a new issue for that.

dbaron:
I changed it to an entirely different example that doesn't go out of gamut (by using colors at the endpoints that are both darker and mix sRGB primaries, although in hindsight I may have only needed the latter difference).

Issue #5211, opened 2020-06-14
[scroll-animations-1] Handling of unspecified value for source seems incorrect


majido:
Here is the wording for source descriptor:

> otherwise if not specified or none then it is the the scrollingElement of the Document associated with the Window that is the current global object.

Note that 'source' is defaulted to 'none' which is interpreted as document root scroller. This seems like a bad choice to have none mean something. We should probably use 'auto' for this purpose and reserve 'none' for situation where source is null.

andruud:
I agree, `auto` is better.

majido:
Spec is fixed. I have filed a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1100907) to fix Chromium implementation.

Issue #5202, opened 2020-06-12
[scroll-animations-1] Something is odd with the handling of `source` in the ScrollTimeline constructor


birtles:
The [`ScrollTimeline` constructor](https://drafts.csswg.org/scroll-animations-1/#dom-scrolltimeline-scrolltimeline) text has the following:

> Creates a new ScrollTimeline object using the following procedure:
> 1. Let timeline be a new ScrollTimeline object.
> 1. Let source be the result corresponding to the first matching condition from below.
>     **If the source value of options is non-null,**
>     * Let source be source
> 
>     **Otherwise (source is null):**
>     * Let source be the scrollingElement of the Document associated with the Window that is the current global object.
> 
>     Note: source may still be null after this step, e.g. if the Document has no scrollingElement.
> 1. Set the source of timeline to source.

Which is odd because `source` is not nullable in the `ScrollTimeline` spec. So how can we assign `source` if it is null?

Also, "Let source be source" is odd. Clearly it means the `source` member of `options` but it needs to say that.

birtles:
If we decide that a `source` can be `null` then procedures like the one to [resolve a container-based offset](https://drafts.csswg.org/scroll-animations-1/#resolve-a-container-based-offset) will need to be updated to say what to do in that case.

birtles:
I notice that the procedure to [resolve an element-based offset](https://drafts.csswg.org/scroll-animations-1/#resolve-an-element-based-offset) handles a null `source` so I suspect the IDL and "resolve a container-based offset" need to be updated (we should also check the other algorithms too, however).

majido:
I guess the idea is that if the `ScrollTimelineOptions` may have a null source (which is the default value). And if that is given as input we resolve it to root scroller. So constructed timeline will never have a null source.

However thinking about this more, I wonder we should allow a three state in our handling
- Unspecified: Use default value as document scroller e.g., `new ScrollTimeline({source: null})`  
- Explicitly set to null:  set source to to null  e.g., `new ScrollTimeline({source: null}) `
- Explicitly set to element:  set source to element  e.g., `new ScrollTimeline({source: $div}) `


This is similar to how timeline is handler on `Animation` constructor.
I can make this change and clear up the wording mistake as well.

birtles:
> I guess the idea is that if the `ScrollTimelineOptions` may have a null source (which is the default value). And if that is given as input we resolve it to root scroller. So constructed timeline will never have a null source.

[...]

PR #5263, opened 2020-06-25
[scroll-animations] let source be explicitly nullable (#5202, #5211)

majido:
Fixing #5202 and #5211 

Changes:
 - Fix IDL to make source nullable.
 - In CSS source will accept 'auto' and 'none' with none
   mapping to null.
 - In JS, we now map missing 'source' to document scrolling element
   (i.e., auto behavior) and explicitly null 'source' to null.
 - Correctly check null source when calculating offsets.

Other minor cleanups.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -359,7 +359,7 @@ dictionary ScrollTimelineOptions {
 [Exposed=Window]
 interface ScrollTimeline : AnimationTimeline {
   constructor(optional ScrollTimelineOptions options = {});
-  readonly attribute Element source;
+  readonly attribute Element? source;
   readonly attribute ScrollDirection orientation;
   readonly attribute (DOMString or ElementBasedOffset) start;
   readonly attribute (DOMString or ElementBasedOffset) end;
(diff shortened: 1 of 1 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -471,18 +475,26 @@ one the following values:
...
+1.  If <em>any</em> of the following are true:
+
+    * {{source}} is null, or
+    * {{source}} does not currently have a [=CSS layout box=], or
+    * {{source}}'s layout box is not a [=scroll container=].
+
+    The [=effective scroll offset=] is null.
+
+1.  The [=effective scroll offset=] is the scroll offset corresponding to the
+    first matching condition from the following:
```
birtles, line 487: This seems odd. Should the second step say, "Otherwise, ..." ? Or should the first step return early?

majido: Good catch. Fixed.

majido:
Spec is fixed. Have filed a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1100907) to fix Chromium implementation.

Issue #4735, opened 2020-02-04
[css-color-5] When mixing hue, there are two ways round the hue range


smfr:
https://drafts.csswg.org/css-color-5/#colormix

Mixing hue in lch() leads to ambiguity (or undesirable results) because hue is an angle around the circle, so there are two paths between any two values. In some cases, authors may want a hue mix that traverses the 0deg/360deg point, at other times not.

svgeesus:
I have been wondering about that myself. Maybe an extra parameter with two values, to go 'the long way round' or 'the short way round' with shorter path being the default if unspecified? It would also need tie logic, so that both paths are available when the path is exactly 180deg.

jrus:
Unless explicitly specified, this should always be the shortest arc.

tabatkins:
Once the spec is fixed to not mod the angle eagerly, the consistent answer is to just do a linear interpolation from the start to the end angle. If you want a particular direction, you can specify the angles to cause that direction, just like you do with 'rotate'.

jrus:
> the consistent answer is to just do a linear interpolation from the start to the end angle

The big problem is that many authors are going to want to set gradient stops or color mix endpoints either (a) from coordinates provided by a graphical tool, (b) in a different coordinate system [mixing in polar-coordinate CIELAB (L*, C*, h*) space is apparently the default irrespective of the space in which a color is specified], or (c) from a specific named color not specified adjacent to the usage in a gradient/color mix.

When this happens for an red and a purple, the gradient should go the short way through red–purple, but from what I can tell according to this proposal will instead go all the way around via orange→yellow→green→blue.

This is going to be very confusing for authors, and difficult or even impossible for them to fix in a straightforward way.

faceless2:
You've mentioned authors using "a graphical tool" consistently in your case for using the shortest path between the angles. Presumably this tool could also convert a gradient from 340° to 20° into -20° to 20° in the generated CSS.

jrus:
There are 2 cases: (1) an author uses a graphical tool to obtain coordinates, and then copy/pastes those coordinates into a document. That puts the burden of figuring out what multiple of 360° to get the desired behavior onto the end user; (2) the graphical tool generates a whole gradient or color mix or whatever, and outputs CSS. In that case you are complicating the tool programmer’s job and making it more likely that their tools will have undesired behavior.

[...]

PR #5278, opened 2020-07-01
[css-color-5] Adjustments to edge cases and set notation for hue interpolation types

dbaron:
This makes a few adjustments to the definitions of hue interpolation
types to follow up on the discussions in #4735.

First, it adjusts the pseudo-code so that for 'shorter' and 'longer',
angles that differ by 180 degrees will always go in the decreasing
direction, as described in
https://github.com/w3c/csswg-drafts/issues/4735#issuecomment-650254063

Second, it adjusts the set notation to match the pseudo-code.  (They
were not matching even prior to this change.)  The set notation
previously seemed to assume that there were absolute-value functions
present that were not actually there.  However, given the asymmetry of
always going in the decreasing direction, it's more correct to write it
without absolute values.

### DIFF
--- a/css-color-5/Overview.bs
+++ b/css-color-5/Overview.bs
@@ -752,20 +752,20 @@ Unless the type of hue interpolation is ''specified'', both angles need to be co
 One way to do this is <code><i>θ</i> = ((<i>θ</i> % 360) + 360) % 360</code>.
 
 : ''shorter''
-:: Angles are adjusted so that θ₂ - θ₁ ∈ [0, 180). In pseudo-Javascript:
+:: Angles are adjusted so that θ₂ - θ₁ ∈ [-180, 180). In pseudo-Javascript:
 	<pre>
 	if (θ₂ - θ₁ >= 180) {
 		θ₁ += 360;
 	}
-	else if (θ₂ - θ₁ <= -180) {
+	else if (θ₂ - θ₁ < -180) {
 		θ₂ += 360;
 	}
 	</pre>
 
 : ''longer''
-:: Angles are adjusted so that θ₂ - θ₁ ∈ [180, 360). In pseudo-Javascript:
+:: Angles are adjusted so that θ₂ - θ₁ ∈ {(-360, -180], 0, (180, 360)}. In pseudo-Javascript:
 	<pre>
-	if (0 < θ₂ - θ₁ < 180) {
+	if (0 < θ₂ - θ₁ <= 180) {
 	  θ₁ += 360;
 	}
 	else if (-180 < θ₂ - θ₁ < 0) {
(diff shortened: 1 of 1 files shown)


dbaron:
(This is following up to the changes made in 4e88f2c4ac6983857fce23599f2f42d03d3d81c1.)

css-color-5/Overview.bs
```diff
@@ -745,20 +745,20 @@ Unless the type of hue interpolation is ''specified'', both angles need to be co
...
 	}
-	else if (θ₂ - θ₁ <= -180) {
+	else if (θ₂ - θ₁ < -180) {
 		θ₂ += 360;
 	}
 	</pre>
 
 : ''longer''
-:: Angles are adjusted so that θ₂ - θ₁ ∈ [180, 360). In pseudo-Javascript:
+:: Angles are adjusted so that θ₂ - θ₁ ∈ {(-360, -180], 0, (180, 360)}. In pseudo-Javascript:
```
LeaVerou, line 759: I find this rather hard to read, can we prioritize readability over conciseness?

css-color-5/Overview.bs
```diff
@@ -745,20 +745,20 @@ Unless the type of hue interpolation is ''specified'', both angles need to be co
...
 	<pre>
-	if (0 < θ₂ - θ₁ < 180) {
+	if (0 < θ₂ - θ₁ <= 180) {
```
LeaVerou, line 761: Why is this needed? Currently on the equal case, you just don't add 360. It's still well-defined. Is there a reason why we want to enter this clause?

dbaron: The reason for wanting to do this is to follow the principle from @svgeesus that angles that differ by 180 should go in the `decreasing` direction.  Without this change, the direction they go would depend on what angles they were, since it would be determined by the constraining to `[0, 360)`.

css-color-5/Overview.bs
dbaron: By just changing it to an English sentence rather than set notation?  Or some other way?

css-color-5/Overview.bs
LeaVerou: After thinking about it some more, I think if θ₂ - θ₁ == 180 (so θ₂ == θ₁ + 180) then interpolation should be the same as `increasing`, whereas when θ₂ - θ₁ == -180 (so θ₂ == θ₁ - 180), interpolation should be the same as decreasing. This seems to be less surprise-inducing than going in the same direction regardless of whether it's θ₂ < θ₁ or θ₂ > θ₁.
I'll think of how to put that in these algorithms tomorrow (or if you want to give it a shot, fine by me!)

css-color-5/Overview.bs
LeaVerou: 0 is not needed if we make the sets closed, and we can combine [-360, -180] and [180, 360] by using the absolute value of the difference.

css-color-5/Overview.bs
svgeesus: Yes, @LeaVerou convinced me that basing the direction at exactly 180 on the sign of the angle difference made more sense.

css-color-5/Overview.bs
dbaron: I'm still pretty hesitant to write a rule that says that interpolation of *identical* hues goes around the circle -- both because we'd have to choose which way, and because I think it might be problematic if we applied such rules to transitions and animations, which I think may depend in some ways on interpolation of identical values not going through other values.

I'll make the other changes shortly.

LeaVerou: Makes sense, yeah.

LeaVerou:
LGTM, and even if we tweak further eventually, the changes bring it to a much better state than the current one, so merging.

PR #5285, opened 2020-07-04
Fix #2586: drop the informative 'aural' "definition"

### LINKED ISSUE #2586: App. A: Aural style sheets is informative yet sounds normative
This goes all the way back to 2002 when it was made informative in ac046f711d.

> This chapter is informative. 

 - [ ] It's an appendix, not a chapter.

> The properties in this appendix apply to a media type 'aural', that was introduced in CSS2. The type 'aural' is now deprecated.

 - [ ] Are we informatively deprecating aural?

> User agents should allow the values corresponding to '0' and '100' to be set by the listener. 

 - [ ] Are we using RFC 2119 "should" informatively? What does it mean that a UA should implement this as an informative statement?

> This property specifies whether text will be rendered aurally and if so, in what manner.

 - [ ] Does it? I thought the property didn't really exist; the appendix is, after all, informative.

> For information, here is the list of properties implemented by Emacspeak, a speech subsystem for the Emacs editor.

- [ ] Why are we informatively giving an implementation report for an informative appendix in the spec?

More generally, why is this appendix even here? We obviously don't expect/want anyone to implement it (otherwise it would've been made normative again), so what do we gain by having it here? Should we define `display: run-in` (which was also in CSS2) informatively as well and rename it "Appendix A. Abandoned Features"?

gsnedders:
Fixes #2586.

This is technical an editorial change (hah!) given it's informative in CSS 2.1, but this is a sufficiently major change that I'm posting this for review first.


frivoal:
I support this. Informative parts of RECs is not the right place for (abandoned) incubation. So, we should get rid of this.

css2/Overview.bs
```diff
@@ -16753,12 +15676,9 @@ Colors</a></h3>
 
 <h3 id="changes">Changes</h3>
 
-<!--
-<h3 id="c1.1"><a href="about.html#css2.1-v-css2">Section 1.1 CSS 2.1 vs CSS 2</a></h3>
```
astearns, line 16757: GitHub blame fails at letting me know when this id was added on line 16757 - any concerns about breaking links to it? Also, I'm wondering if any/all of the ids in the removed section should be added to this change entry, so that links to the removed section go here?

gsnedders: We dropped the changes section pretty early on, and decided not to care about links to it. We can re-add at the least the 2016 anchors for all the 2.1 changes, but the previous decision had been we don't need to bother keeping them.

gsnedders: (Feel free to open a new issue if you believe we should maintain all the IDs that appear in the Changes section in 2.1 or if we should maintain the entire Changes list from 2.0)

PR #5294, opened 2020-07-07
[css-images-3] `image-orientation: none` applies to non-decorative images too. #5245

### LINKED ISSUE #5245: [css-images-3] change which images image-orientation applies to
In #4165 it was resolved that EXIF orientation applies to all images by default. There was some discussion about which images the `image-orientation` property and talk of a resolution, but no actual resolution.

This matters because we've got two sets of WPT tests added in the last six months which disagree - see https://github.com/web-platform-tests/wpt/issues/18549#issuecomment-645532399 - and different results in Firefox and Chrome (see [a](http://wpt.live/css/css-images/image-orientation/image-orientation-background-image.html), [b](http://wpt.live/css/css-images/image-orientation/image-orientation-none-content-images.html)). So we need a once-and-for-all resolution on this.

@heycam made the same observation in https://github.com/w3c/csswg-drafts/issues/4165#issuecomment-587949977, and we'd like to implement this but can't until it's clarified.

So the question is: does the `image-orientation` property apply to "decorative" images, such as `background-image` and `border-image`? And if it does not (as currently specified), the second question is can we tightly define "decorative" images.

@zcorpan made a full list at https://github.com/web-platform-tests/wpt/issues/18549#issuecomment-522631537

> CSS
> * ::before/::after + content
> * background-image
> * list-style-image
> * border-image
> * cursor
>
> SVG
> * image
>
> HTML
> * img
> * canvas drawImage() + getImageData()
> * video poster
> * input type=image
> * embed
> * object
>
> load an image in an iframe
> load an image in a top-level doc (this already respects EXIF in all browsers I believe)
> Favicon link rel=icon / favicon.ico (manual tests?)
> web app manifest icon (manual test?)

* I would expect "content" images to be those displayed from `content`, `list-style-image` (which generates content), SVG `image`, HTML`img`, `embed`, `object`, and `video` poster. The `image-orientation` property should apply to all of these.
* I would expect "decorative" images to be `background-image`, `border-image` and `mask-image`.
* `image-orientation` does not apply to cursor, top-level docs, favicon or web-app manifests. 
* I'm slightly unsure about exactly where iframe fits.

cc @schenney-chromium

heycam:
Fixes #5245.

### DIFF
--- a/css-images-3/Overview.bs
+++ b/css-images-3/Overview.bs
@@ -25,6 +25,10 @@ Ignored Vars: start image, end image
 Link Defaults: css2 (property) display
 Include Can I Use Panels: yes
 </pre>
+<pre class=link-defaults>
+spec:html; type:element; text:link
+spec:filter-effects-1; type:element; text:feimage
+</pre>
 
 Introduction {#intro}
 =====================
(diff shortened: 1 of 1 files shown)


heycam:
I'm adding tests in https://phabricator.services.mozilla.com/D82471.

fantasai:
Sorry for taking so long on this; there were a number of adjustments I knew I wanted to make so I procrastinated a lot on the review. But thanks for making sure the spec covers a wide range of the implications of the WG resolution!

Pushed some edits to the spec, which are better for having your PR to work from. :) I think we still have a bit of adjustment to do to clearly describe e.g. the SVG cases, so I'm going to leave this PR open until we're 100% sure we covered everything you have here.

fantasai:
OK.  I think I've got everything covered now @heycam. Feel free to review and complain if not. :) In the meantime, I'm going to close this PR.

Issue #628, opened 2016-10-20
[css2] CSS 2.2 does not reflect changes to the color type from CSS 3 Color, which is a Rec


svgeesus:
CSS 2.2 section 4.3.6 Colors
https://www.w3.org/TR/CSS22/syndata.html#value-def-color

This section doesn't seem to reflect the changes made in CSS3 Color, which has been a W3C Recommendation for a while. I thought we were adding pointers from 2.2 to newer work, or indeed deleting sections from 2.2 and replacing them with a pointer to newer work?

SelenIT:
Maybe it would make sense to delete all sections marked as replaced by newer spec [in the "Official Definition" of CSS](https://drafts.csswg.org/css-2015/#css) per CSS-2015 snapshot and replace them with pointers to the newer modules? It would be much easier to read the spec without need to check if the particular section has been superseded by something newer or not.

tabatkins:
That is our general policy, yes - delete if it's entirely superceded, with a pointer to the new spec. (For exactly the reasons you give.)

fantasai:
@tabatkins That is incorrect. Our policy was to *add* a pointer to the new spec, not to remove the section.

Issue #5114, opened 2020-05-26
Versioning policy and backwards incompatible changes


gsnedders:
I started writing this as a comment on #5111, but it seems like I've never actually posted an issue for all the times we've spoken about this. This also arguably has implications on #4770.

> [It] seems useful, to distinguish [specs such as cascade-3] from other old RECs which are unmaintained and wrong. People should be reading cascade-4, but they're not going to run into problems if they read cascade-3. On the other hand, the same isn't true about reading CSS2.0 instead of CSS2.1.

Honestly this just touches on our inconsistent habits with versioning again. I would much rather we actually tackled the whole versioning issue generally (where level 1 is nowadays defined informatively in the CSS Snapshot with the superseded spec advised against; level 2 is defined by 2.1; level 3 is defined by the modules; etc.).

At the moment if we make a backwards incompatible change to Level 4, our practice is to make behaviour undefined/optional in L2 and L3 (and ignore L1), which means we have to republish a number of RECs with substantive changes each time we make a backwards incompatible change, and that number is only going to increase over time. We also have a long term ambition to fully obsolete CSS2, which will then lead to the question as to where the normative definition of Level 2 will live.

We have a number of options here, as I see it:

1. Continue down the status-quo, accepting republishing multiple RECs when we do make normative changes.
2. Rewrite the L[n] spec to define itself as a subset of those in L[n+1].
3. Make the L[n] spec define behaviour for all L[<n] (this should never be divergent; this should just be including the different value spaces for properties and noting certain things are only defined for level k and above.

tabatkins:
> We also have a long term ambition to fully obsolete CSS2, which will then lead to the question as to where the normative definition of Level 2 will live.

I don't think we *will* have a normative definition of Level 2 at that point. It'll be obsoleted just as much as CSS 1 is.  We're not far from that point already.

> At the moment if we make a backwards incompatible change to Level 4, our practice is to make behaviour undefined/optional in L2 and L3 (and ignore L1),

I don't think that's our practice, either?

CSS2 had a lot of stuff declared undefined near the end of its process, not because of backwards-incompatible changes in the level 3 module, but because the behavior wasn't yet consistent between browsers and so we couldn't get 2+ passes for tests for any reasonable defined behavior.

[...]

PR #5298, opened 2020-07-07
[CSS2] define <color> by reference to CSS Color 3

gsnedders:
Fixes https://github.com/w3c/csswg-drafts/issues/628.

See also: https://github.com/w3c/csswg-drafts/issues/5114.

This is in many ways me experimenting with what the working group actually thinks of some of the proposals in both those issues. :smile:

### DIFF
--- a/css2/Overview.bs
+++ b/css2/Overview.bs
@@ -167,6 +167,14 @@ Required IDs: abs-non-replaced-height, abs-non-replaced-width, absolutely-positi
     }
 }
 </pre>
+
+<pre class=link-defaults>
+spec:css-color-3; type:value; text:blue
+spec:css-color-3; type:value; text:red
+spec:css-color-3; type:value; text:silver
+</pre>
+
+
 <style type="text/css">
  span.colorsquare { float:left; width:5em; height:3em; text-align:center; padding:1.2em 0 .8em }
  span.colorname { font-weight:bold }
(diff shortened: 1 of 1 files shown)


frivoal:
Overall, this seems fine to me. Css2 should be updated to say the same thing as css-color 3, as #628 argues, and that's certainly the easiest way to do that.

A minor note: SVG11 is an informative reference of CSS2, but a normative reference of CSS-COLOR3. As it is itself a REC, I don't believe that's an issue.

So, from a normative point of view, I think there's no issue with this, and I support it.

From a patent policy point of view, the effect of doing this is not necessarily identical to adjusting the css2 text to mean the same as the level 3 text and republishing that.
Currently, we have patent commitment:
- on the css2.1 content, by the members who were there then
- on the css-color-3 content, by the members who were there then
Eventually we will also have patent commitment:
- on the css-color-4, which is a superset of 3, by the members who've involved in its development

If we do what you suggest, that's all we get.

If we publish a new CSS2.x REC with the level-2 content corrected to match level 3, we'd also get patent commitment on the corrected level 2 material from members involved during that time frame. I don't think there would be any member who'd end up committed to this adjusted level 2 who's not already committed to level4, and I'm not convinced we'd even get there sooner.

So, even considering the Patent Policy implications, I still think this is a good move.

css2/Overview.bs
```diff
@@ -2684,142 +2692,16 @@ p:before {content: counter(par-num, upper-roman) ". "}
...
-em { color: rgb(110%, 0%, 0%) }  /* clipped to rgb(100%,0%,0%) */
-</pre>
-</div>
-
-<p>Other devices, such as printers, have different gamuts than sRGB;
-some colors outside the 0..255 sRGB range will be representable
-(inside the device gamut), while other colors inside the 0..255 sRGB
-range will be outside the device gamut and will thus be mapped.
-</p>
-<div class="note"><p><em><strong>Note.</strong> Mapping or clipping of
```
svgeesus, line 2819: cringing at the wooliness of that paragraph, good to have dropped it

gsnedders:
I guess we probably want to get a WG resolution on this given this both includes substantive changes and patent commitments.

gsnedders:
(FWIW, I'm not gonna be on the call this week, hopefully will be by next week; @svgeesus can you handle the above?)

fantasai:
I don't think we should take this change. I think we should definitely add a note pointing to css-color-3, but I don't think we should be removing the entire section. 

I also would like to see CSS2 republished cleanly, and then changes made on top of that, rather than piling up a bunch of changes in the ED without much clarity wrt what's happening on /TR?

svgeesus:
@fantasai could you explain why obsolete material should not be removed?

svgeesus:
related https://github.com/w3c/csswg-drafts/issues/5114

gsnedders:
**RESOLVED**: sections entirely replaced by RECs are marked with obsoletion notices recommending the REC, but not removed

PR #5300, opened 2020-07-08
[scroll-animations-1] Use CSSTypedOM instead of strings (#5213)

### LINKED ISSUE #5213: [scroll-animations-1] Use of DOMString in API surface
Small point that came up in the [TAG Review](https://github.com/w3ctag/design-reviews/issues/521), the [ScrollTimeline](https://drafts.csswg.org/scroll-animations-1/#scrolltimeline-interface) interface is currently using DOMString to represent keyword and length/percentage values. This should use the appropriate [CSS Typed OM](https://drafts.css-houdini.org/css-typed-om/) classes instead.

Using strings for CSS values is a historical artifact we're trying to eliminate, new APIs shouldn't use them.

majido:
Use appropriate CSSTypedOM instead of DOMStrings in the API. Fixes #5213.

Changes:
 - For offsets use `CSSNumericValue` or `CSSKeywordish` instead of DOMString
 - When setting the offset, the grammars for these values is check to ensure only the valid subset can be set.
 - When computing effective scroll offset check the offset actual type and interpret it accordingly
 - Update examples to showcase the API and add noted to clarify the usage.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -19,6 +19,7 @@ Editor: Olga Gerchikov, Microsoft, gerchiko@microsoft.com
 
 Former editor: Mantaroh Yoshinaga
 Former editor: Stephen McGruer, Google, smcgruer@google.com
+Markup Shorthands: markdown yes
 </pre>
 <pre class=anchors>
 urlPrefix: https://w3c.github.io/web-animations/; type: dfn; spec: web-animations
@@ -183,8 +184,8 @@ if (window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
 
   const collisionTimeline = new ScrollTimeline({
     source: scrollableElement,
-    start: '200px',
-    end: '300px'
+    start: 200,
+    end: 300
   });
 
   const left = leftCircle.animate({ transform: 'translate(300px)' }, 1000);
@@ -196,8 +197,8 @@ if (window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
   const union = unionCircle.animate({ opacity: 1 }, { duration: 1000, fill: "forwards" });
   union.timeline = new ScrollTimeline({
     source: scrollableElement,
-    start: '250px',
-    end: '300px'
+    start: 250,
+    end: 300
   });
 }
 </pre>
@@ -230,6 +231,8 @@ this example could be written as follows:
 @media (prefers-reduced-motion: no-preference) {
   @scroll-timeline progress-timeline {
     source: selector(#body);
+    start: 0;
+    end: 100%;
   }
 
   @keyframes progress {
@@ -251,7 +254,10 @@ If we use this API for this case, the example code will be as follow:
 if (window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
   var animation = div.animate({ width: '100%' }, { duration: 1000, fill: "forwards" });
   animation.timeline = new ScrollTimeline(
-    { start: '0px' }
+    { 
+      start: 0,
+      end: CSS.percent(100)
+    }
   );
 }
 </pre>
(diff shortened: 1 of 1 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -487,21 +492,37 @@ The procedure to <dfn>resolve a container-based offset</dfn> given
...
     <div class="switch">
 
-    :   <var>offset</var> is <code>auto</code>
+    :   If <var>offset</var> is <code>auto</code>:
     ::  Either the beginning or the ending of {{source}}'s scroll range in
         {{orientation}} depending on whether the offset is {{start}} or {{end}}.
 
-    :   <var>offset</var> is a <<length-percentage>>
+    :   If <var>offset</var> is a {{CSSNumberish}} and matches <<number>>:
+    ::  The distance indicated by the value in <a value>pixel unit</a> along
```
andruud, line 500: I was under the impression that treating plain numbers as `px` was a pattern we didn't want? Maybe get someone else's opinion on this. (@tabatkins ?)

scroll-animations-1/Overview.bs
```diff
@@ -487,21 +492,37 @@ The procedure to <dfn>resolve a container-based offset</dfn> given
...
+        {{source}}'s scroll range in {{orientation}}.
+
+
+    :   If <var>offset</var> is a {{CSSNumberish}} and matches
```
andruud, line 504: Can we link "matches" to https://drafts.css-houdini.org/css-typed-om-1/#cssstylevalue-match-a-grammar ?

scroll-animations-1/Overview.bs
```diff
@@ -348,11 +354,14 @@ behavior under different directionalities and writing modes.
 <pre class="idl">
 enum ScrollTimelineAutoKeyword { "auto" };
 
+typedef (CSSNumberish or ScrollTimelineAutoKeyword) ContainerBasedOffset;
```
andruud, line 357: We should IMO also accept `CSSKeywordValue` matching `auto`.

Would you want to use `CSSKeywordish` here (if it existed)?

andruud: My bad, CSSKeywordish already exists: https://drafts.css-houdini.org/css-typed-om-1/#typedefdef-csskeywordish

Do we want to use that somehow, then?

scroll-animations-1/Overview.bs
majido: I personally felt allowing numbers to be interpreted as `px` would make the API easier to use in Javascript. Specially as it matches all legacy scroll related APIs where numbers are treated as pixels


But I can also see at least two possible concerns:
 - Automatic type conversion can be a footgun but perhaps this is limited enough to be safe
 - Having every downstream spec convert `<numbers>` differently would be bad. So perhaps have a common way to do this i.e., `<number>` -> `<length>` with pixel for all JS API's that accept length but want to accept number too.

BTW, does TypedOM have a special handling for `0` length for which in CSS the unit is optional?

In any case, this is a trade-off and @tabatkins expert's opinion is appreciated. :)

plinss: We really don't want to interpret bare numbers as px. The definition of CSSNumberish interprets numeric values as a CSS "number" not a length. `offset` should simply be a CSSNumericValue or a CSSStyleValue.

scroll-animations-1/Overview.bs
majido: Done. As part of this I now have a procedure that rectifies strings into CSSKeyword and rejects invalid values at the time of setting the value.

scroll-animations-1/Overview.bs
majido: Done.

scroll-animations-1/Overview.bs
majido: Done. I removed CSSNumberish and just use CSSNumericValue which I think is the right type.

Also added a procedure to validate the incoming values when they are being set.

PR #5311, opened 2020-07-09
[CSS2] inherit should use computed, not specified, value

### LINKED ISSUE #2768: 'inherit' keyword incorrectly resolves to the specified value of parent
From https://lists.w3.org/Archives/Public/www-style/2011Oct/0482.html:

   - RESOLVED: Fix wording in CSS2.1 sections 6.2.1 and 6.1.1 to say that 'inherit' causes the property's specified value to be its parent's computed value.
               http://www.w3.org/Bugs/Public/show_bug.cgi?id=14420

Previously fixed in [2.1 errata](https://www.w3.org/Style/css2-updates/REC-CSS2-20110607-errata.html#s.6.2.1).

gsnedders:
Fixes #2768.

From https://lists.w3.org/Archives/Public/www-style/2011Oct/0482.html:

   - RESOLVED: Fix wording in CSS2.1 sections 6.2.1 and 6.1.1 to say
               that 'inherit' causes the property's specified value
               to be its parent's computed value.
               http://www.w3.org/Bugs/Public/show_bug.cgi?id=14420

Originally fixed in 2aa48e8ef and ab60bee0d.

Co-authored-by: Sam Sneddon <me@gsnedders.com>

### DIFF
--- a/css2/Overview.bs
+++ b/css2/Overview.bs
@@ -15,6 +15,7 @@ Former Editor: Editor: Bert Bos, W3C, mailto:bert@w3.org, w3cid 3343
 Former Editor: Elika J. Etemad / fantasai, Invited Expert, http://fantasai.inkedblade.net/contact, w3cid 35400
 Former Editor: Ian Hickson, ian@hixie.ch
 Former Editor: H&aring;kon Wium Lie, Opera Software, howcome@opera.com
+WPT Display: inline
 Abstract:
   This specification defines Cascading Style Sheets level&nbsp;2 (CSS&nbsp;2)
   revision&nbsp;2 (CSS&nbsp;2.2). CSS is a style sheet language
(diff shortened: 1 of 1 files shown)


css2/Overview.bs
```diff
@@ -15,6 +15,7 @@ Former Editor: Editor: Bert Bos, W3C, mailto:bert@w3.org, w3cid 3343
 Former Editor: Elika J. Etemad / fantasai, Invited Expert, http://fantasai.inkedblade.net/contact, w3cid 35400
 Former Editor: Ian Hickson, ian@hixie.ch
 Former Editor: H&aring;kon Wium Lie, Opera Software, howcome@opera.com
+WPT Display: inline
```
frivoal, line 18: Did you really intend to commit this line? Feels like something you may have used locally and forgot to remove before committing.

PR #5331, opened 2020-07-16
[css + web-animations] Handle getKeyframes for CSSAnimations

andruud:
In css-animations-2, instead of generating keyframes resembling
computed keyframes, generate specified keyframes. Computing the
keyframes at keyframe-creation-time creates too many questions
regarding when/how property values inside the keyframes are computed.
We can instead rely on the existing process for calculating computed
keyframes, and apply that process at the time getKeyframes is called.

This means that var() references (and dependencies in general, such
as 'em' units) remain until computed-keyframe-time.

This isn't a problem for shorthand expansion at keyframe-creation-time:
if there's a var() in a shorthand, it expands using the "pending
substitution value" [1] for its longhands.

We can continue to resolve logical properties to physical at keyframe-
creation-time as well, assuming 'direction' and 'writing-mode' are not
allowed in keyframes. (FWIW, it's not allowed in Blink, but I'm not
sure if this is spec'd).

In web-animations-1, in the algorithm for getKeyframes, we either
use the specified keyframes (+ compute missing offset) as the source,
or the computed keyframes, depending on circumstance. This
"circumstance" could definitely be tightened up a bit (instead of
referring to whether or not calls to setKeyframes have taken place),
but it probably makes sense to do this as a whole for all the
"tainting" behavior described by css-animations-2 (Animations chapter),
so I haven't tried to address that here.

I'm not sure if we need to explicitly add somewhere that the computed
keyframes contains the necessary offset/easing/composite values, or
if it's already "understood" that these exist also for computed
keyframes. (I didn't make a change here).

Finally, the fact that getKeyframes returns objects listed as
"dictionary ComputedKeyframe" is a bit awkward, since it's not
consistent with the spec's "main" definition of computed keyframes.
We should probably rename it.

[1] https://drafts.csswg.org/css-variables/#pending-substitution-value

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -299,8 +299,6 @@ and the position of the animation in |element|'s 'animation-name' list,
      1.   Iterate over all declarations in the keyframe block and
           add them to |keyframe| such that:
 
-          *     All variable references are resolved to their current values.
-
           *     Each shorthand property is expanded to its longhand
                 subproperties.
 
@@ -318,9 +316,7 @@ and the position of the animation in |element|'s 'animation-name' list,
                 have already added at this same |keyframe offset|,
                 they should be skipped.
 
-           *    All property values are replaced with their computed values.
-
-     1.   Add each physical longhand property name that was added to |keyframe|
+     1.   Add each property name that was added to |keyframe|
           to |animated properties|.
 
 1.   If there is no keyframe in |keyframes| with offset 0,
@@ -362,20 +358,6 @@ It could be rewritten so this is not required but that will likely change
 the behavior for some edge cases.
 We should verify what current implementations do and possible remove the
 requirement to iterate in reverse.
-
-Issue: In practice, implementations will likely maintain specified values and
-variable references internally and only resolve them to computed values when
-{{KeyframeEffect/getKeyframes()}} is called on a {{KeyframeEffect}}
-associated with a {{CSSAnimation}}
-(and, more specifically, a {{CSSAnimation}} whose effect has not been replaced
-and whose keyframes have not been overridden
-using {{KeyframeEffect/setKeyframes()}}).
-It is not clear that soing so would produce an observable difference in
-behavior, however.
-If it does, the above procedure might need to be adjusted so that the
-requirement to expand variables and convert to property values to computed
-values is limited to the result of {{KeyframeEffect/getKeyframes()}} and
-not the internal representation of keyframes.
 
 ## The 'animation-duration' property ## {#animation-duration}
 
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -5049,9 +5049,21 @@ interface KeyframeEffect : AnimationEffect {
     used to prepare the result of this method is defined in prose below:
 
     1.  Let <var>result</var> be an empty sequence of objects.
-    1.  Let <var>keyframes</var> be the result of applying the procedure to
-        <a>compute missing keyframe offsets</a> to the <a>keyframes</a>
-        for this <a>keyframe effect</a>.
+    1.  Let <var>keyframes</var> be one of the following:
+
+        1.  If this <a>keyframe effect</a> is associated with a
+            {{CSSAnimation}}, and its <a>keyframes</a> have not been replaced
+            by a successful call to {{KeyframeEffect/setKeyframes()}};
+            the <a>computed keyframes</a> for this <a>keyframe effect</a>.
+
+        1.  Otherwise, the result of applying the procedure
+            <a>compute missing keyframe offsets</a> to the <a>keyframes</a>
+            for this <a>keyframe effect</a>.
+
+        Note: We return <a>computed keyframes</a> for CSS Animations
+              because not all keyframes specified in CSS can be
+              represented by a dictionary.
+
     1.  For each <var>keyframe</var> in <var>keyframes</var>
         perform the following steps:
 
@@ -5077,9 +5089,8 @@ interface KeyframeEffect : AnimationEffect {
             keyframe-specific <a>timing function</a>, and
             <a>keyframe-specific composite operation</a>
             values of <var>keyframe</var>.
-        1.  For each animation property-value pair specified on
-            <var>keyframe</var>, <var>declaration</var>, perform the following
-            steps:
+        1.  For each animation property-value pair in <var>keyframe</var>,
+            <var>declaration</var>, perform the following steps:
 
             1.  Let <var>property name</var> be the result of applying the
                 <a>animation property name to IDL attribute name</a> algorithm


andruud:
cc @birtles

birtles:
> Computing the
> keyframes at keyframe-creation-time creates too many questions
> regarding when/how property values inside the keyframes are computed.

What does keyframe-creation-time refer to here?

I _think_ I understand the proposal here and I _think_ I like it. That is I think you might have solved the problem I have wanted to solve for a long time through the "pending substitution value" concept (which I wasn't familiar with) and by making round-tripping even less likely to succeed. At the time I was reluctant to go that direction but I think given that there are already many other places where round-tripping doesn't work due to variables this is ok.

andruud:
> What does keyframe-creation-time refer to here?

In css-animations-2, the spec says "[...] keyframe objects are generated as follows". By keyframe-creation-time, I meant whenever _that_ keyframe object generation is taking place, which should be during the style resolve which "discovers" that a valid animation was specified in CSS. This handled somewhat by css-animations-1 ([Animations chapter](https://drafts.csswg.org/css-animations-1/#animations)), although it deals mostly with Animation objects, _implying_ the creation of keyframe objects when animations are started/updated.

Previously we'd generate these keyframe objects right off the bat to something that resembles _computed_ keyframes. My proposal here is that we instead generate something closer to _specfied_ keyframes, and then lean on [Calculating computed keyframes](https://drafts.csswg.org/web-animations-1/#calculating-computed-keyframes) for the rest.

> pending substitution value [...] making round-tripping even less likely to succeed

The keyframe objects generated in my proposal here are not directly observable as-is, since we're also specifying that `getKeyframes` returns _computed keyframes_ (for keyframes generated by CSS). In other words, the declarations in the generated keyframes will undergo [Computing properties values](https://drafts.csswg.org/web-animations-1/#computing-property-values) before exiting through `getKeyframes`. This means that any pending substitution values will have been eliminated before they are observed.

For example:

```
  :root {
    --x: 1px 2px 3px 4px;
    --y: 5px 6px 7px 8px;
  }

  @keyframes test1 {
    from { margin: var(--x); }
    to { margin: var(--y); }
  }
```

The keyframes object we'd generate from the above can be illustrated by:

```
  @keyframes test1 {
    from {
      margin-top: <pending-substitution: var(--x)>;
      margin-right: <pending-substitution: var(--x)>;
      margin-bottom: <pending-substitution: var(--x)>;
      margin-left: <pending-substitution: var(--x)>;
    }
    to {
      margin-top: <pending-substitution: var(--y)>;
      margin-right: <pending-substitution: var(--y)>;
      margin-bottom: <pending-substitution: var(--y)>;
      margin-left: <pending-substitution: var(--y)>;
    }
  }
```

(Note that pending-substitution values implicitly capture the shorthand it originates from, such that we know how to parse the value after substitution).

Then, when calling `getKeyframes`, we calculate and return the computed keyframes, which can be illustrated by:

```
  @keyframes test1 {
    from {
      margin-top: 1px;
      margin-right: 2px;
      margin-bottom: 3px;
      margin-left: 4px;
    }
    to {
      margin-top: 5px;
      margin-right: 6px;
      margin-bottom: 7px;
      margin-left: 8px;
    }
  }
```

This round-trips at least somewhat successfully (though it's no longer responsive to changes in `--x`/`--y`, of course).


If you had instead included `--x` in the keyframes itself:

```
  @keyframes test2 {
    from { margin: var(--x); --x: 1px 2px 3px 4px;}
    to { margin: var(--x); --x: 5px 6px 7px 8px; }
  }
```

We can illustrate the generated keyframes as:

```
  @keyframes test2 {
    from {
      margin-top: <pending-substitution: var(--x)>;
      margin-right: <pending-substitution: var(--x)>;
      margin-bottom: <pending-substitution: var(--x)>;
      margin-left: <pending-substitution: var(--x)>;
      --x: 1px 2px 3px 4px;
    }
    to {
      margin-top: <pending-substitution: var(--x)>;
      margin-right: <pending-substitution: var(--x)>;
      margin-bottom: <pending-substitution: var(--x)>;
      margin-left: <pending-substitution: var(--x)>;
      --x: 5px 6px 7px 8px;
    }
  }
```

And then the computed keyframes (and hence the result from `getKeyframes`) would depend on the current time of the animation. At 50%, the value of `--x` would just have flipped to the value specified by the `to` frame, so we get computed keyframes (again using CSS syntax for illustration purposes):

```
  @keyframes test2 {
    from {
      margin-top: 5px;
      margin-right: 6px;
      margin-bottom: 7px;
      margin-left: 8px;
      --x: 1px 2px 3px 4px;
    }
    to {
      margin-top: 5px;
      margin-right: 6px;
      margin-bottom: 7px;
      margin-left: 8px;
      --x: 5px 6px 7px 8px;
    }
  }
```

This (and similar situations with complicated dependencies) will only round-trip correctly for the current point in time, but we already discussed this in #5125. But it won't fail "catastrophically" like it sounds like you were thinking.

>> When computing the keyframe, a var() reference will substitute whatever the computed value of the referenced custom property currently is, not to the value locally defined by the same keyframe.

> I'm not sure if that's the behavior we want for CSS keyframes is it?

The above question was imported from [elsewhere](https://github.com/web-platform-tests/wpt/pull/24655#issuecomment-661091458), but I'll reply here since it pertains to this spec edit.

This is exactly what #5125 was about: how (and _when_) to resolve dependencies. The bullet point I'm removing here ("All variable references are resolved to their current values") is presicely the thing that prompted that issue in the first place, since it was not clear (to many people) what this meant or how it was supposed to work.

I'm really only effectuating what we already concluded with in #5125, namely that dependencies behave according to the G-β behavior. In practice this means that keyframes that are generated by CSS keep `var()` references around, and the substitution of these are delayed until computed-keyframe-time. (And how to deal with these dependencies at computed-keyframe-time is now reasonably well defined as of recently). This gives us a nice consistent model for how dependencies are resolved, I think.

Sure, we _could_ define some special rules, and resolve the `var()`s against _something_ at the time keyframes are generated. Then there would be no `var()`s left to resolve by the time [Computing properties values](https://drafts.csswg.org/web-animations-1/#computing-property-values) is invoked. But this would effectively re-open #5125, as it sounds like a violation of G-β. And at this point I don't see why we should go looking for more trouble, when the model we already concluded with in #5125 seems to work quite well. :-)

birtles:
> > What does keyframe-creation-time refer to here?
> 
> In css-animations-2, the spec says "[...] keyframe objects are generated as follows". By keyframe-creation-time, I meant whenever _that_ keyframe object generation is taking place, which should be during the style resolve which "discovers" that a valid animation was specified in CSS. This handled somewhat by css-animations-1 ([Animations chapter](https://drafts.csswg.org/css-animations-1/#animations)), although it deals mostly with Animation objects, _implying_ the creation of keyframe objects when animations are started/updated.
> 
> Previously we'd generate these keyframe objects right off the bat to something that resembles _computed_ keyframes. My proposal here is that we instead generate something closer to _specfied_ keyframes, and then lean on [Calculating computed keyframes](https://drafts.csswg.org/web-animations-1/#calculating-computed-keyframes) for the rest.

Thanks, that's makes sense.

> This (and similar situations with complicated dependencies) will only round-trip correctly for the current point in time, but we already discussed this in #5125. But it won't fail "catastrophically" like it sounds like you were thinking.

That's great. Thanks for the thorough explanation.

> > > When computing the keyframe, a var() reference will substitute whatever the computed value of the referenced custom property currently is, not to the value locally defined by the same keyframe.
> 
> > I'm not sure if that's the behavior we want for CSS keyframes is it?
> 
> The above question was imported from [elsewhere](https://github.com/web-platform-tests/wpt/pull/24655#issuecomment-661091458), but I'll reply here since it pertains to this spec edit.
> 
> This is exactly what #5125 was about: how (and _when_) to resolve dependencies. The bullet point I'm removing here ("All variable references are resolved to their current values") is presicely the thing that prompted that issue in the first place, since it was not clear (to many people) what this meant or how it was supposed to work.

Sorry, I misunderstood that comment. I thought it meant we would ignore the custom property values defined in keyframes altogether.

> I'm really only effectuating what we already concluded with in #5125, namely that dependencies behave according to the G-β behavior. In practice this means that keyframes that are generated by CSS keep `var()` references around, and the substitution of these are delayed until computed-keyframe-time. (And how to deal with these dependencies at computed-keyframe-time is now reasonably well defined as of recently). This gives us a nice consistent model for how dependencies are resolved, I think.
> 
> Sure, we _could_ define some special rules, and resolve the `var()`s against _something_ at the time keyframes are generated. Then there would be no `var()`s left to resolve by the time [Computing properties values](https://drafts.csswg.org/web-animations-1/#computing-property-values) is invoked. But this would effectively re-open #5125, as it sounds like a violation of G-β. And at this point I don't see why we should go looking for more trouble, when the model we already concluded with in #5125 seems to work quite well. :-)

That sounds great. Thank you so much for all your work on this.

PR #5377, opened 2020-07-31
[css-syntax-3] Reference identifier-start instead of name-start

jonathantneal:
This updates the label of a reference to the **identifier-start code point** definition in the CSS Syntax specification.

The **identifier code point** definition contains a potentially mislabeled [reference to a “_**name-start** code point_”](https://drafts.csswg.org/date/2020-07-22T16:50:27/css-syntax-3/#identifier-code-point):

> **identifier code point**
> A name-start code point, a digit, or U+002D HYPHEN-MINUS (-).

In that definition, “_name-start code point_” links to the **identifier code point** definition.

I believe it should be relabeled to “_identifier-start code point_”:

> **identifier code point**
> An identifier-start code point, a digit, or U+002D HYPHEN-MINUS (-).

### DIFF
--- a/css-syntax-3/Overview.bs
+++ b/css-syntax-3/Overview.bs
@@ -789,7 +789,7 @@ Definitions</h3>
 		<dd>
 			A <a>code point</a> with a value equal to or greater than U+0080 &lt;control>.
 
-		<dt><dfn export lt="identifier-start code point" oldids="name-start-code-point">identifier-start code point</dfn>
+		<dt><dfn export lt="identifier-start code point | name-start code point" oldids="name-start-code-point">identifier-start code point</dfn>
 		<dd>
 			A <a>letter</a>,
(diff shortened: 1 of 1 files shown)


css-syntax-3/Overview.bs
```diff
@@ -789,15 +789,15 @@ Definitions</h3>
 		<dd>
 			A <a>code point</a> with a value equal to or greater than U+0080 &lt;control>.
 
-		<dt><dfn export lt="identifier-start code point | name-start code point" oldids="name-start-code-point">identifier-start code point</dfn>
```
tabatkins, line 792: This needs to stay, so old references continue to link.

jonathantneal: Sure thing! This is done in 7103a9f.

Should I also (or alternatively) add anything to `oldids`?

PR #5391, opened 2020-08-04
[css-inline-3] Fix the code point for maqaf

xfq:
https://drafts.csswg.org/css-inline-3/#baseline-synthesis-fonts

`U+05B3` is a Hebrew niqqud (vowel) sign, instead of a maqaf (a punctuation). The correct code point should be `U+05BE` instead.


fantasai:
Replaced maqaf with He, so closing this PR...

Issue #5400, opened 2020-08-05
[web-animations-1] Play state of an animation with an unresolved timeline


kevers-google:
Recently the rules for determining the play state of an animation were updated to add the condition of having an unresolved start time for the animation to be idle. This constraint was added to accommodate non-monotonically-increasing (i.e. scroll) timelines which should not be flagged as idle if they have a start time but an inactive timeline; however, it does break the play state for cases where the timeline is unresolved.  Explicitly setting the start time of an animation with an unresolved timeline should leave the animation in an idle state. This issue can be resolved by relaxing the constraint to require that either the start time or the timeline be unresolved.

https://drafts.csswg.org/web-animations-1/#play-states

kevers-google:
cc @birtles @ogerchikov @majido @flackr 

Related to issue: https://github.com/w3c/csswg-drafts/issues/2066

Changes to how play state is calculated for supporting scroll timelines introduced changes in the behavior of set timeline. These side effects would not have been visible in WPT tests in Blink since setting the timeline was not supported. 

Suggesting the following change to play state:

 All of the following conditions are true:
    The current time of animation is unresolved, and
    the start time of animation is unresolved **or the timeline is unresolved**, and
   animation does not have either a pending play task or a pending pause task,

→ idle

If the timeline is explicitly set to null, should the play state be allowed to be in a running state?  

With this change, the changes in support of scroll timelines do not introduce inconsistencies between browsers for null timelines.

flackr:
Just for clarification, by unresolved timeline I assume you mean there is no associated timeline for the animation?

kevers-google:
Correct

flackr:
This makes sense to me. Can you make a PR?

We have some testing of setting timelines in web-animations/timing-model/animations/setting-the-timeline-of-an-animation.html and Firefox and Safari are passing these tests - do we have test coverage which was broken by this change and/or will we after we make it?

kevers-google:
Firefox is passing due to not having the test for startTime being unresolved in the idle check. 
https://searchfox.org/mozilla-central/source/dom/animation/Animation.cpp#480

The following test was updated with the algorithm change and currently fails on Firefox and Safari:
wpt/web-animations/timing-model/animations/setting-the-timeline-of-an-animation.html

Once setTimeline is supported in Blink the following tests will have new failure reasons based on the plate state:
wpt/css/css-animations/event-dispatch.tentative.html
[...]

PR #5409, opened 2020-08-07
[web-animations-1] Update the procedure for determining if the animation play state is idle.

kevers-google:
https://github.com/w3c/csswg-drafts/issues/5400

Explicitly setting the start time on an animation with a null timeline causes it to be flagged as being in the running state. The constraint that the start time be unresolved was added recently to accommodate scroll-timelines, which may have an unresolved current time even though the start time has been set. Relaxing the constraint to required that the start time be unresolved or the timeline be null restores the previous behavior for null timelines without breaking scroll timelines.


flackr (commented):
I think this makes sense, but needs to be updated to resolve the merge conflicts.

birtles:
I thought we concluded in #5400 that we should not do this?

kevers-google:
Abandoning the PR based on #5400.

PR #5414, opened 2020-08-10
[css-syntax] Simplify ident-like URL consumption

jonathantneal:
The specification reads:

> While the next two input code points are whitespace, consume the next input code point.

If I understand correctly, this will repeatedly consume whitespace until 0 or 1 whitespace input code points remain.

Then it reads:

> If the next one or two input code points U+0022 QUOTATION MARK (`"`), U+0027 APOSTROPHE (`'`), or whitespace followed by U+0022 QUOTATION MARK (`"`) or U+0027 APOSTROPHE (`'`), then create a `<function-token>`...

Could this be simplified?

> While the next input code point is whitespace, consume it. If the next input code point is a U+0022 QUOTATION MARK (`"`) or U+0027 APOSTROPHE (`'`), then create a `<function-token>`...

### DIFF
--- a/css-syntax-3/Overview.bs
+++ b/css-syntax-3/Overview.bs
@@ -1136,11 +1136,9 @@ Consume an ident-like token</h4>
 	If |string|’s value is an <a>ASCII case-insensitive</a> match for "url",
 	and the <a>next input code point</a> is U+0028 LEFT PARENTHESIS ((),
 	consume it.
-	While the <a lt="next input code point">next two input code points</a> are <a>whitespace</a>,
-	consume the <a>next input code point</a>.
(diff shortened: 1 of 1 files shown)


tabatkins:
...huh, I have no idea why I wrote it that way.

tabatkins:
Ah, looking at the history, I changed *from* your suggested text to the current text in <https://github.com/csstools/csswg-drafts/commit/5f673863c0b1195fc4b8933b2e4d3893b1f8eae0>, and now that I see it, I understand why - if there is any whitespace between the open-paren and the string, I need to preserve it, so the next token produced is a whitespace token.

(I can't just emit the function-token immediately and let normal whitespace processing handle that; I need to scan forward and see if it's going to be a normal function (containing a string) or if it needs to be specially parsed as a url-token (unquoted).)

Issue #5433, opened 2020-08-14
[mediaqueries-5] duplication of `forced-colors: active` and `prefers-contrast: forced`


cookiecrook:
[mediaqueries-5] duplication of [`forced-colors: active`](https://drafts.csswg.org/mediaqueries-5/#descdef-media-forced-colors) and [`prefers-contrast: forced`](https://drafts.csswg.org/mediaqueries-5/#valdef-media-prefers-contrast-forced)

Breaking this discussion out of the large mega thread in #2943. 

I believe these statements are true. 
1. AFAICT, `forced-colors: active` and `prefers-contrast: forced` are exact duplications of each other. 
2. I am only aware of one implementation of forced colors: on Windows.
3. `forced-colors` is already implemented in Edge (and possibly in the Chrome HC extension.)

Opinions:
- I'm not sure why two properties are needed. I think I'd heard both that the `forced` value is intended as a replacement for the `forced-colors` media feature, and I've also heard that these are intended to be complementary and co-exist. Which is it, and why?
- Since there is only one implementation (Windows), I worry that shoehorning a single `forced` value into `prefers-constrast` is unnecessarily limiting. Another platform (iOS, Mac, ChromeOS, Android, Linux) may support a different forced mode in the future, and `forced-colors` would be easier to extend.

Editorial: [`prefers-contrast: forced`](https://drafts.csswg.org/mediaqueries-5/#valdef-media-prefers-contrast-forced) cross-link to [`forced-colors: active`](https://drafts.csswg.org/mediaqueries-5/#descdef-media-forced-colors) an confirms the duplication. If both are kept, the reciprocal link and explanatory prose should be added, too.

cookiecrook:
Even though I'm a WG member, I'm not able to add the requested labels. Is that intentional?

astearns:
Nope, it's just a manual step for us to add people to the right team. You should be good now.

frivoal:
Whether or not its the right conclusion is a separate question, but the current idea is:

They are indeed the exact same thing. We wish we had though of defining it as `prefers-contrast: forced` first, but we didn't, so MS shipped `force-color: active` already. We thought we still liked `prefers-contrast: forced` better, so we defined that, but since `forced-colors: active` shipped, we're keeping it as well for compat reasons. The fact that one is the desired syntax, and the other is the legacy alias isn't currently clear in the spec, but an editorial cleanup should make that more obvious.

So, with that in mind, I think we can take 3 stances:
[...]

Issue #3856, opened 2019-04-19
[css-color-adjust-1][mediaqueries-5] Fold `forced-colors` and `prefers-contrast`?


fantasai:
Was trying to work out authoring advice for responding to forced-colors mode and drafted up

> Authors are encouraged to simplify the contrast in their pages when '@media/forced-colors' is ''active'', reducing effects such as shadows, fades, blurs, filters, gradients, and image or pattern fills that add complexity to discerning shape outlines.

But what I noticed is that this advice is probably also applicable any time `(prefers-contrast)` is true (i.e. when it is `low` or `high` but not `no-preference`). Which made me wonder if `forced-colors: active` should instead be `prefers-contrast: forced`?

fantasai:
More info on forced colors mode, for anyone who's dropping into the issue without context: discussion in https://github.com/w3c/csswg-drafts/issues/3807 on MSFT proposal described in https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/HighContrast/explainer.md yielding current spec text at https://drafts.csswg.org/css-color-adjust-1/#forced

fantasai:
Wondering also how `(prefers-reduced-transparency)` plays into this... Transparency effects, pattern fills, and bad contrast all interfere with shape recognition.

Basically, I think it'd be nice if authors could have very clear triggers for when to make what adjustments, and if we could make those triggers as simple and straightforward as possible as media queries, and incorporate that authoring advice into the spec.

AmeliaBR:
I don't think we should be confusing `forced-colors` and `prefers-contrast` at all.  If there is authoring advice that is specific to creating good contrast, it should go under the `prefers-contrast` MQ.  A `forced-color` setting (as currently defined) is going to turn off text shadows and block out busy backgrounds behind text, anyway, so I'm not sure why this extra advice is necessary.

Now, it may make sense to combine `prefers-reduced-transparency` and `prefers-contrast`.  But then, I'm not quite sure who's asking for the reduced transparency case — the MacOS no-transparency setting is as much a performance option as it is an accessibility one.

I _do_ think web browsers should automatically update the `prefers-contrast` state automatically if they can correctly deduce that the forced color scheme is high/low contrast.  

("correctly" being the key point.  WHCM can't even figure out if my custom color scheme is dark-on-light or light-on-dark & doesn't give me any way to specify it — it's a hidden setting based on whichever built-in color scheme you started with before you customized it. grumble grumble grouch…)

[...]

PR #5434, opened 2020-08-16
Refactoring of the forced-colors and prefers-contrast media feature

frivoal:
This editorial reorganization of the text is intended to make the
overlap between forced-colors:active and prefers-contrast:forced
more readily understandable: one being the syntax that the CSS Working
Group finds more desirable, the other being retained for compatibility
with content issued against an earlier draft (see #3856).

Prior to this refactoring, the definition of this mode was also spread
between the two media features, causing confusion. This commits
consolidates the whole thing in one place.

Related to #5433

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -2680,10 +2680,13 @@ Detecting the desire for increased or decreased color contrast from elements on 
 			has a high or low contrast,
 			one of ''prefers-contrast: more'' or ''prefers-contrast: less'' should match
 			in addition to ''prefers-contrast: forced''.
+
+			Note:
 			Similarly,
 			if the forced color palette chosen by the user
 			fits within one of the color schemes described by 'prefers-color-scheme',
-			the corresponding value should also match.
+			the corresponding value will also match.
+			See [[css-color-adjust-1#forced]] for details.
 	</dl>
 
 	<div class=note>
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -2624,7 +2624,7 @@ Detecting the desire for reduced transparency on the page: the 'prefers-reduced-
 	ISSUE: How does this interact with preferences around e.g. pattern fills and backgrounds?
 	They're not about transparency, but they also interfere with shape recognition.
 
-<h3 id="prefers-contrast">
+<h3 id="prefers-contrast" oldids="forced-colors">
 Detecting the desire for increased or decreased color contrast from elements on the page: the 'prefers-contrast' feature</h3>
 
 	<pre class='descdef mq'>
@@ -2634,11 +2634,11 @@ Detecting the desire for increased or decreased color contrast from elements on 
 	Type: discrete
 	</pre>
 
-	The 'prefers-contrast' media feature is used to detect if the user
-	has requested the system increase or decrease the amount of contrast
-	between adjacent colors. For example, many users have difficulty
-	reading text that has a small difference in contrast to the text
-	background and would prefer a larger contrast.
+	The 'prefers-contrast' media feature is used to detect
+	if the user has requested an increase or decrease the amount of contrast used in the page,
+	expecting an adjustment to the contrast ratio between adjacent colors,
+	and to how much elements stand out visually,
+	for example by adjusting their borders.
 
 	<dl dfn-type=value dfn-for="@media/prefers-contrast">
 		<dt><dfn>no-preference</dfn>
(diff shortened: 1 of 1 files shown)


frivoal:
Note: This isn't the place to discuss whether the merge of these two media features is a good idea (see #3856 and #5433) for that. I'm just trying to make the current design/resolution easier to read. Whether we want to overturn this decision or not is a separate discussion.

frivoal:
@tabatkins this is (IMO) purely editorial, so I could have committed this directly, but I wouldn't mind a co-editor's review on this one, as this topic has proven reasonably subtle.

mediaqueries-5/Overview.bs
```diff
@@ -2634,11 +2634,11 @@ Detecting the desire for increased or decreased color contrast from elements on
...
-	The 'prefers-contrast' media feature is used to detect if the user
-	has requested the system increase or decrease the amount of contrast
-	between adjacent colors. For example, many users have difficulty
-	reading text that has a small difference in contrast to the text
-	background and would prefer a larger contrast.
+	The 'prefers-contrast' media feature is used to detect
+	if the user has requested an increase or decrease the amount of contrast used in the page,
+	expecting an adjustment to the contrast ratio between adjacent colors,
+	and to how much elements stand out visually,
+	for example by adjusting their borders.
```
tabatkins, line 2641: Not quite sure how the end of this sentence hooks up to the start. Seems like perhaps a change in subject? It could probably stand being broken into a separate sentence.

mediaqueries-5/Overview.bs
```diff
@@ -2673,10 +2675,15 @@ Detecting the desire for increased or decreased color contrast from elements on
 				or some other arrangement that is neither particularly low or high contrast.
 			</div>
 
-			Note: When the User Agent can determine whether the particular forced color scheme
-			chosen by the user is high or low contrast,
-			one of ''prefers-contrast: more'' or ''prefers-contrast: less'' should match,
+			Additionally, when the user agent can determine
+			whether the forced color palette chosen by the user
+			has a high or low contrast,
+			one of ''prefers-contrast: more'' or ''prefers-contrast: less'' should match
```
tabatkins, line 2681:
```suggestion
			one of ''prefers-contrast: more'' or ''prefers-contrast: less'' must match
```

mediaqueries-5/Overview.bs
```diff
@@ -2673,10 +2675,15 @@ Detecting the desire for increased or decreased color contrast from elements on
...
 			in addition to ''prefers-contrast: forced''.
+			Similarly,
+			if the forced color palette chosen by the user
+			fits within one of the color schemes described by 'prefers-color-scheme',
+			the corresponding value should also match.
```
tabatkins, line 2686: This requirement is already specified in color-adjust, with more detail. We can probably drop it (or pull it to a Note?)

mediaqueries-5/Overview.bs
frivoal: Given that we're not setting an explicit criteria by which to judge whether the forced color palette chosen by the user has a high or low contrast, I don't know if there's really a usable difference between must and should. I went with should be cause if the browser doesn't do it, it's not observable whether that is because it did not judge the contrast to be high/low, or because it did, but chose not to make the prefers-contrast query match.

I figured that if we cannot tell the two apart, insisting on a must didn't buy us anything.

I don't necessarily oppose switching to must, but can you help me see how that helps?

mediaqueries-5/Overview.bs
frivoal: Nice catch. I hadn't noticed color-adjust had this with more detail already. Turning into note+cross-ref.

mediaqueries-5/Overview.bs
frivoal: Adding parenthesis to show how I meant to group these statements, and implied bits between [brackets]:

> The 'prefers-contrast' media feature is used to detect
> if the user has requested an increase or decrease in the amount of contrast used in the page,
> [due to that user] expecting an adjustment to (the contrast ratio between adjacent colors,)
> and to (how much elements stand out visually,
> for example by adjusting their borders).

Here's a different attempt:

> The 'prefers-contrast' media feature is used to detect
> if the user has requested an increase or decrease in the amount of contrast used in the page.
> This could be responded to by adjusting the contrast ratio between adjacent colors,
> or by changing how much elements stand out visually,
> for example by adjusting their borders.

alshamiri5 (approved):

mediaqueries-5/Overview.bs
frivoal: @fantasai @tabatkins I understand both, cause I wrote both :) Could the two of you agree as to which one's better, or suggest a third alternative that makes everyone happy?

mediaqueries-5/Overview.bs
```diff
@@ -2702,6 +2712,47 @@ Detecting the desire for increased or decreased color contrast from elements on
 		result in a high or low contrast (or something else).
 	</div>
 
+	<div class=example>
+		<ul>
```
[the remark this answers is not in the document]
frivoal: Added.

PR #5439, opened 2020-08-17
[css-contain-2] Specify that the first content-visibility: auto visibility check happens immediately

vmpstr:
When a new content-visibility: auto element is added, the check to determine
its visibility needs to occur in the same frame to avoid a flash of blank content.

This PR adds a clarification to this effect.

### DIFF
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -1292,7 +1292,20 @@ Restrictions and Clarifications {#cv-notes}
 		and not cause any forced layouts.
 	</div>
 
-4. For the purposes of scrolling operations,
+4. The initial determination of visibility for ''content-visibility: auto'' must
+    happen in the same frame that determined an existence of a new
+    ''content-visibility: auto'' element.
+
+  <div class=note>
+    When an element first gains ''content-visibility: auto'', it may or may not
+    be positioned on screen. The determination of this state and thus
+    determination of whether this element is [=skipped=] must happen in the
+    same frame. If it does not, then there is a possibility of producing blank
+    content in the element's place since visibility check and [=skipped=] state
+    update would be deferred to the next frame.
+  </div>
+
+5. For the purposes of scrolling operations,
 	such as {{Element/scrollIntoView()}},
 	an element with ''content-visibility: auto''
 	that is [=skipping its contents=]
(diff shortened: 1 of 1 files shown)


css-contain-2/Overview.bs
```diff
@@ -1292,7 +1292,20 @@ Restrictions and Clarifications {#cv-notes}
 		and not cause any forced layouts.
 	</div>
 
-4. For the purposes of scrolling operations,
+4. The initial determination of visibility for ''content-visibility: auto'' must
+    happen in the same frame that determined an existence of a new
```
chrishtr, line 1296: What about nested content-visibility: auto ?

vmpstr: Hmm. I think this is a good question. I would say that the wording stands: if the nested auto is the first time we're evaluating visibility then it should happen in the same frame.

The problem here is that we want to avoid a flash of blank. It's different if this was off-screen and then enters the screen in which case there are other mechanisms to prevent the flash (margin)

WDYT?

chrishtr: Right now your implementation uses a loop in the update-the-rendering steps that is equivalent to a ResizeObserver, but stops after one loop. To implement the semantics you're suggesting we'd have to have a loop like this:

```
while (true) {
  updateStyleAndLayout();
  if (!thereAreNewlyDiscoveredContentVisibilityAutoElements())
    break;
  deliverIntersectionObservationsToNewlyDiscoveredContentVisibilityAutoElements();
}
```

As implemented above this will have at most as many while loops as there are nestings in`content-visibility: auto`.

I think this is fine actually; the looping semantics and code structure are exactly the same as `ResizeObserver`, except the observer is a special sync-on-first-frame IntersectionObserver callback. (Note that ResizeObserver also has a don't-infinite-loop requirement that it goes deeper into the DOM on each while loop iteration, which is similar to the invariant above about nested `content-visibility:auto`.)

vmpstr: Yeah, that's pretty much the intent (and how it works in Chrome)

@tabatkins do you mind taking a quick look as well?

tabatkins: This seems fine to me from a spec perspective; the detail of implementation there is above me.

frivoal:
resolution to accept this change recorded here: https://logs.csswg.org/irc.w3.org/css/2020-12-16/#e1381652

PR #5456, opened 2020-08-20
[css-values-4] Fix types in css-values-4 draft

sottar:
I found this when I was reading [css-values-4](https://www.w3.org/TR/css-values-4/#combine-colors),  I think it's color here, not number.

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -1965,9 +1965,9 @@ Combination of <<color>></h4>
 	(that is, colors for which the red, green, and blue components specified
 	have been multiplied by the alpha).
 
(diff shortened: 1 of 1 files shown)


css-values-4/Overview.bs
```diff
@@ -1965,9 +1965,9 @@ Combination of <<color>></h4>
 	(that is, colors for which the red, green, and blue components specified
 	have been multiplied by the alpha).
 
-	<a>Addition</a> of <<number>> is likewise defined as
+	<a>Addition</a> of <<color>> is likewise defined as
 	the independent <a>addition</a> of each component
-	as a <<number>>
```
tabatkins, line 1970: This one should stay as number, since it's talking about the value of a color component.

PR #5460, opened 2020-08-21
[scroll-animations-1] Remove default vaule for ScrollTimelineOptions.source

andruud:
Follow-up to #5263, which specifies that a "missing" source shall
result in using the Document's scrollingElement, but if we give
source a default value of null, it can never be missing.


andruud:
No response from @flackr, can you have a look at this @ogerchikov?

flackr (approved):
LGTM, sorry for the delay on this

ogerchikov:
Looks good.

PR #5470, opened 2020-08-25
[css-scroll-anchoring-1] Reduce the scope of focused element

vmpstr:
This reduces the scope of the focused area priority candidate to only the elements that permit text entry. Notably, this excludes typical buttons and anchor tags

Also small fix-up to active match to reference the spec.

### DIFF
--- a/css-scroll-anchoring-1/Overview.bs
+++ b/css-scroll-anchoring-1/Overview.bs
@@ -35,6 +35,20 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/;
         type: dfn;
             text:DOM anchor
             text:focused area of the document
+            text:active match; url: fip-active-match
+    urlPrefix: form-elements.html
+        type: dfn;
+            text:textarea; url: the-textarea-element
+    urlPrefix: input.html
+        type: dfn;
+            text:input; url: the-input-element
+    urlPrefix: form-control-infrastructure.html
+        type: dfn;
+            text:mutable; url: concept-fe-mutable
+spec: html; urlPrefix: https://w3c.github.io/editing/docs/execCommand/;
+    urlPrefix: index.html
+        type: dfn;
+            text:editable
 </pre>
 
 <h2 id=intro>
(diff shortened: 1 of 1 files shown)


vmpstr:
@emilio Can you take a look please? We found that as previously specced, things like buttons would become priority candidates and thus be scroll anchors. However, this breaks the "show more" button pattern since instead of scrolling the button off-screen, it's anchored and the content is instead added above it. Although there's a solution (adding overflow-anchor: none to the button), the sheer size of the impact seems to indicate the preference for previous behavior for buttons.

(let me know if the language here makes sense as well :) )

/cc @chrishtr @domenic

css-scroll-anchoring-1/Overview.bs
```diff
@@ -35,6 +35,20 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/;
...
+            text:active match; url: fip-active-match
+    urlPrefix: form-elements.html
+        type: dfn;
+            text:textarea; url: the-textarea-element
+    urlPrefix: input.html
+        type: dfn;
+            text:input; url: the-input-element
+    urlPrefix: form-control-infrastructure.html
+        type: dfn;
+            text:mutable; url: concept-fe-mutable
```
domenic, line 47: You should be able to get rid of these if you use `<{input}>` and `<{textarea}>` instead of `[=input=]` and `[=textarea=]`.

vmpstr: Done. Thanks!

emilio:
I'm not quite sure this makes sense for tab navigation though. When you tab-navigate, you actually want the focused button/link/etc on screen, you don't want any of the focus moving offscreen, I think.

If you're ok with that though, I guess this is fine.

vmpstr:
> I'm not quite sure this makes sense for tab navigation though. When you tab-navigate, you actually want the focused button/link/etc on screen, you don't want any of the focus moving offscreen, I think.

Yeah that's true. I think this case is probably less of an issue than an intentional attempt by the site to move a focused element off-screen (like a click to add content), although I don't really have evidence to support my claim other than the fact that we haven't had bugs filed for that case. We did get a bunch of bugs filed when we implemented the focus priority candidate

alshamiri5 (approved):

domenic (approved):

Issue #5428, opened 2020-08-13
[css-nesting] Update mentions of :matches() to :is() or :where()


tomhodgins:
Just noticed the text of the CSS Nesting spec mentions `:matches()` in a number of places that should probably be updated to [`:is()`](https://drafts.csswg.org/selectors-4/#matches) or [`:where()`](https://drafts.csswg.org/selectors-4/#zero-matches).

Also, a quick search of this repository shows another file where `:matches()` could maybe be updated for clarity as well:

- https://github.com/w3c/csswg-drafts/blob/4223887dd3d671c9b9c55033538b94ef08a9eec3/selectors-4/deferred-for-level-5

PR #5474, opened 2020-08-26
Update mentions of `:matches(…)` to `:is(…)` or `:where(…)`

ExE-Boss:
Fixes <https://github.com/w3c/csswg-drafts/issues/5428>


tabatkins:
thanks!

Issue #4792, opened 2020-02-19
Interoperable font metrics via explicit font metrics overrides


litherum:
Migrated from https://github.com/w3c/csswg-drafts/issues/2228#issuecomment-380762748

In font files, there exist at least three different ascent/descent values:
- `ascender` and `descender` from the [`hhea` table](https://docs.microsoft.com/en-us/typography/opentype/spec/hhea)
- `sTypoAscender` and `sTypoDescender` from the [`OS/2` table](https://docs.microsoft.com/en-us/typography/opentype/spec/os2)
- `usWinAscent` and `usWinDescent` from the [`OS/2` table](https://docs.microsoft.com/en-us/typography/opentype/spec/os2)
- maybe others???

Different browsers / OSes use different values. In fact, some even use values that are not present in the font file at all. This leads to content being laid out differently on different browsers / OSes. https://github.com/w3c/csswg-drafts/issues/2228 is an example of one. There have been others.

There have been efforts to try to introduce "interoperable" metrics into CSS (https://github.com/w3c/csswg-drafts/issues/3978 is one, https://github.com/w3c/csswg-drafts/issues/938 is related, there are a few others too), to force browsers to agree on which metrics they use. Many of the existing proposals so far have some downsides:
- They require browsers to parse font files, rather than relying on special-purpose font parsing libraries (such as CoreText). A tenet of the web platform is that browsers should be able to delegate responsibility to helper libraries.
- They cause all text to be laid out differently, across every website in the entire web. Such a massive change is too scary from a product development standpoint.
- Any approach which is opt-in for specific cases causes text metrics to not match throughout the page. This can lead to inconsistent spacing and poor typography / webpage design.

However, interoperable font metrics _are_ desirable for many page authors. Here's a compromise proposal that (I believe) solves all the above problems:

What if we add a new descriptor into `@font-face` to let the CSS author override the metrics of the font? Something like this:

```CSS
@font-face {
    font-family: "WebFont";
    src: ...;
    ascent-override: 80%;
    descent-override: 20%;
}
```

Those percentage values are resolved by multiplying by the font size (em box). Negative values are illegal. The browser doesn't have to parse any font files, all browsers will agree on the resolved value, and all text in all situations drawn with that font will use these values.

There's already precedent for this: the `font-feature-values` and `font-variation-values` descriptors in `@font-face` overrides the defaults present in the font file.

kojiishi:
Thank you for writing this up! /cc @kojiishi @drott

[...]

Issue #2228, opened 2018-01-26
[css-inline] Consistent vertical positioning of large and medium-sized text across OSs


tobireif:
The large text on the page
https://tobireif.com/non_site_stuff/test_case_for_font_position_report/
gets positioned differently eg in Chrome on Mac vs in Chrome on Windows:

The space above the large text is 102px in Chrome on Mac
vs 125px in Chrome on Windows.

The space below the large text is 75px in Chrome on Mac
vs 52px in Chrome on Windows.

The large-text positioning / the space above and below the large text should be the same in Chrome across OSs, for this font and for all fonts where the effective position currently is different across OSs.

Also see eg

https://app.crossbrowsertesting.com/public/ib8393c20a124bac/livetests/10424300/snapshots/zfa955e684685aa97890

vs

https://app.crossbrowsertesting.com/public/ib8393c20a124bac/livetests/10424303/snapshots/z7586faac855fc19fb22

I know that there are (probably many) other pages/sites/layouts where medium/large text and its position is important.

When a font does not have platform-specific metrics the browser should ensure the same vertical position across OSs.

When a font does have platform-specific metrics I need a way to override (eg using CSS) the relevant aspects in order to ensure the same vertical position across OSs.

Please prepend eg "[css-inline-4]" to the ticket title if appropriate.

Also see these three browser issue reports:
https://bugs.chromium.org/p/chromium/issues/detail?id=800693
https://bugzilla.mozilla.org/show_bug.cgi?id=1429347
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/15204186/

tobireif:
Another test-page:
https://tobireif.com/non_site_stuff/test_case_for_font_position_report_yet_another_font/

tobireif:
From the above test page:

Chrome on MacOS:

![chrome_mac_os](https://user-images.githubusercontent.com/80411/35439338-8cb9f178-029a-11e8-86aa-cf19bf836e6c.png)

Chrome on Windows:

![chrome_windows](https://user-images.githubusercontent.com/80411/35439340-90798dc8-029a-11e8-88a4-5518f09775f3.png)

tobireif:
Should the browser position the large text in the above example screenshots consistently across OSs?

Or could CSS offer a property?

fantasai:
@litherum might have some insight on this...

tobireif:
Just to clarify: I don't have a preference.

[...]

PR #5521, opened 2020-09-16
Add @font-face descriptors for overriding font metrics

xiaochengh:
Following the [resolution](https://github.com/w3c/csswg-drafts/issues/4792#issuecomment-693528301), this patch adds the following @font-face descriptors to CSS Fonts Level 4:
- `ascent-override: normal | <percentage>`
- `descent-override: normal | <percentage>`
- `line-gap-override: normal | <percentage>`

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -2933,17 +2933,20 @@ For: @font-face
 Initial: normal
 </pre>
 
-These 'ascent-override', 'descent-override' and 'line-gap-override' descriptors define the values
-of the ascent, descent and line gap metrics of the font.
+The 'ascent-override', 'descent-override' and 'line-gap-override' descriptors define the
+<a spec="CSS-INLINE-3">ascent metric</a>, <a spec="CSS-INLINE-3">descent metric</a> and
+<a spec="CSS-INLINE-3">line gap metric</a> of the font.
 
-When the descriptor value is 'normal', the metric value is obtained from the font file directly.
+When the descriptor value is 'normal', the corresponding metric value is obtained from the
+font file directly.
 
 Note: User agents may choose different pieces of data from the font file as the metric value,
 leading to different layouts. For example, Chrome on Windows uses sTypoAscender, sTypoDescender
-and sTypoLineGap from the font's OS/2 table, while on Mac, it uses those in the HHEA table if available.
+and sTypoLineGap from the font's OS/2 table, while on Mac, it uses those in the HHEA table if
+available.
 
-When the descriptor value is a percentage, the metric value is set to the given percentage of the
-em size of the font.
+When the descriptor value is a percentage, the corresponding metric value is resolved as the
+given percentage multiplied by the used font size. Negative values are illegal.
 
 <div class="example">
 	We may override the metrics of Arial as follows:
@@ -2951,8 +2954,8 @@ em size of the font.
 <pre>@font-face {
   font-family: arial;
   src: local(Arial), url("http://example.com/arial.woff");
-  ascent-override: 60%;
-  descent-override: 40%;
+  ascent-override: 80%;
+  descent-override: 20%;
   line-gap-override: 0%;
 }
 </pre>
@@ -2965,7 +2968,7 @@ font-size: 20px;
 line-height: normal;
 </pre>
 
-	Each line box will be exactly 20px high. Text baseline will be positioned at 12px
+	Each line box will be exactly 20px high. Text baseline will be positioned at 16px
 	below line box top.
 
 </div>
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -2908,6 +2908,68 @@ Initial: normal
 
 This descriptor defines initial settings that apply when the font defined by an @font-face rule is rendered. It does not affect font selection. Values are identical to those defined for the 'font-language-override!!property' property defined below except that the value inherit is omitted. When multiple font feature descriptors, properties, or variations are used, the cumulative effect on text rendering is detailed in the section [[#font-feature-variation-resolution]] below.
 
+<h3 id="font-metrics-override-desc">
+Default font metrics overriding:
+the 'ascent-override', 'descent-override' and 'line-gap-override' descriptors</h3>
+
+<pre class='descdef'>
+Name: ascent-override
+Value: normal | <<percentage>>
+For: @font-face
+Initial: normal
+</pre>
+
+<pre class='descdef'>
+Name: descent-override
+Value: normal | <<percentage>>
+For: @font-face
+Initial: normal
+</pre>
+
+<pre class='descdef'>
+Name: line-gap-override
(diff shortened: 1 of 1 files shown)


css-fonts-4/Overview.bs
```diff
@@ -2908,6 +2908,71 @@ Initial: normal
...
+When the descriptor value is 'normal', the corresponding metric value is obtained from the
+font file directly.
+
+Note: User agents may choose different pieces of data from the font file as the metric value,
+leading to different layouts. For example, Chrome on Windows uses sTypoAscender, sTypoDescender
+and sTypoLineGap from the font's OS/2 table, while on Mac, it uses those in the HHEA table if
+available.
+
+When the descriptor value is a percentage, the corresponding metric value is resolved as the
+given percentage multiplied by the used font size. Negative values are illegal.
```
chrishtr, line 2949: s/illegal/ignored as invalid/ maybe?

css-fonts-4/Overview.bs
```diff
@@ -2908,6 +2908,71 @@ Initial: normal
...
+<pre class='descdef'>
+Name: line-gap-override
+Value: normal | <<percentage>>
+For: @font-face
+Initial: normal
+</pre>
+
+The 'ascent-override', 'descent-override' and 'line-gap-override' descriptors define the
+<a spec="CSS-INLINE-3">ascent metric</a>, <a spec="CSS-INLINE-3">descent metric</a> and
+<a spec="CSS-INLINE-3">line gap metric</a> of the font.
```
chrishtr, line 2938:
```suggestion
<a spec="CSS-INLINE-3">line gap metric</a> of the font, respectively.
```

css-fonts-4/Overview.bs
```diff
@@ -2908,6 +2908,71 @@ Initial: normal
...
+
+When the descriptor value is 'normal', the corresponding metric value is obtained from the
+font file directly.
+
+Note: User agents may choose different pieces of data from the font file as the metric value,
+leading to different layouts. For example, Chrome on Windows uses sTypoAscender, sTypoDescender
```
litherum, line 2944: I don't think there's precedence for mentioning one specific browser inside a CSS spec.

css-fonts-4/Overview.bs
litherum: It's important to mention what "illegal" or "invalid" mean. We probably want them to be parse-time errors. So if you have
```
ascent-override: 60%;
ascent-override: -70%;
```
the value that gets used is the 60%.

css-fonts-4/Overview.bs
```diff
@@ -2908,6 +2908,71 @@ Initial: normal
...
+and sTypoLineGap from the font's OS/2 table, while on Mac, it uses those in the HHEA table if
+available.
+
+When the descriptor value is a percentage, the corresponding metric value is resolved as the
```
litherum, line 2948: I think we should add a sentence talking about how this percentage gets resolved against different values for different elements. So if you have

```
@font-face {
    ...
    ascent-override: 50%;
}
<span style="font-size: 20px;">...
    <span style="font-size: 150%;">...</span>
</span>
```
the ascent used in the outer element would be `10px` whereas the ascent used in the inner element would be `15px`.

css-fonts-4/Overview.bs
```diff
@@ -2908,6 +2908,71 @@ Initial: normal
...
+given percentage multiplied by the used font size. Negative values are illegal.
+
```
litherum, line 2950: It may also be good to mention something about how this works with `src: local(...)`.

Also, one of the natural use cases for this is to have a fallback font use metrics that are closer to the primary font. Perhaps an example showing this in use with fallback fonts would be valuable?

litherum (approved):

css-fonts-4/Overview.bs
tobireif: > For example, Chrome on Windows uses sTypoAscender, sTypoDescender
> [... and every browser does something different on every platform]

I really wish the spec would require all browsers on all platforms to do the same thing (I know that's a red flag for some people, but it would be opt-in), in order to solve the issue #2228 properly (without requiring web developers to figure out and supply percentages).

svgeesus: Yes, I think it is enough to mention that different implementations on different platforms use metrics drawn from different places in the font.

css-fonts-4/Overview.bs
xiaochengh:
```suggestion
given percentage multiplied by the used font size. Negative values are invalid at parse time.
```

css-fonts-4/Overview.bs
```diff
@@ -2908,6 +2908,71 @@ Initial: normal
...
+<a spec="CSS-INLINE-3">ascent metric</a>, <a spec="CSS-INLINE-3">descent metric</a> and
+<a spec="CSS-INLINE-3">line gap metric</a> of the font, respectively.
+
+When the descriptor value is 'normal', the corresponding metric value is obtained from the
+font file directly.
+
+Note: User agents may choose different pieces of data from the font file as the metric value,
+leading to different layouts. For example, Chrome on Windows uses sTypoAscender, sTypoDescender
+and sTypoLineGap from the font's OS/2 table, while on Mac, it uses those in the HHEA table if
+available.
```
xiaochengh, line 2946:
```suggestion
Note: User agents may draw data from different places from the font file as the metric values,
which results in different text layouts.
```

css-fonts-4/Overview.bs
xiaochengh: Revised accordingly.

css-fonts-4/Overview.bs
xiaochengh: Done.

xiaochengh:
Updated with all comments addressed. PTAL.

fantasai:
I believe this would go into Fonts Level 5, though?

alshamiri5 (approved):

xiaochengh:
I don't have a strong opinion which level it should go into.

@tabatkins Could you advise?

astearns:
@alshamiri5 please stop spamming our PRs with reviews.

svgeesus:
> I believe this would go into Fonts Level 5, though?

That spec doesn't yet exist and if created, would need to be a delta rather than a copy of fonts 4. Where it goes depends on how rapidly we get to consensus and get implementations. I could be persuaded either way.

svgeesus (approved):
Thanks for incorporating feedback to date. This LGTM.

svgeesus:
As both editors have approved, I'm going to go ahead and merge this. If it later turns out that this should be moved off to a Fonts 5 we can do that, along with any other featurs that are lagging on consensus or implementer uptake.

@xiaochengh thank you so much for the clear explanations, willingness to incorporate changes, and experimental implementation which all made this feature request super simple to get consensus around!

xiaochengh:
Thanks to everyone for reviewing!

Issue #5045, opened 2020-05-05
[mediaqueries] In-gamut media query


LeaVerou:
We currently have the [`color-gamut` media query](https://drafts.csswg.org/mediaqueries-4/#color-gamut). While generally useful, this is a rather crude measure of the output device's gamut, and currently only supports RGB color spaces. Since there is no ordering of gamuts, we also cannot really have `min/max-color-gamut` either, either it's "approximately" P3 (whatever that means), or it's not.

I would like a to propose an `in-gamut: <color>` media feature, that supports any CSS color, and matches if that particular color is within the output device's gamut. This will enable authors to more accurately determine the gamut of the output device, based on what matters in their own particular use cases.

faceless2:
You had me rushing to the spec for this one! The fallback in color() is if the original color is _invalid_ -  which I understood to be (for example) if the ICC profile it refers to is missing. I don't think it's used if it's out of gamut. Is that the intention?

Presumably to test if (for example) P3 was in-gamut, you'd have to take representative samples across the entire colorspace, convert to XYZ and then convert back to the device colorspace to see if it's clipped?

svgeesus:
@leaverou wrote:

> already allow providing a fallback in color() if the specified color is _out of gamut_

@faceless2 i
> if the original color is _invalid_ 

Aha, both reasonable interpretations of wooly prose:

> The color function takes one or more comma-separated arguments, with each argument specifying a color, and later colors acting as "fallback" if an earlier color **can’t be displayed** (for example, if the colorspace it specifies hasn’t been loaded yet).

`can't be displayed` is undefined except for an illustrative, parenthetical example.

To test a single color to be in-gamut, for additive colorspaces, you just need to check if any component is negative, or greater than 1.0 /  100%. (While using a code path that does not prematurely clip, obviously).

For non-additive colorspaces, like print, you also need to check that TAC is not exceeded.

svgeesus:
This would e a very valuable media query. @frivoal let me know if you would like me to suggest prose text.

Removing the color-4 tag as it does not affect that specification (although it would make it easier for stylesheet authors to use that specification).

frivoal:
@svgeesus I know just enough about colors to be dangerous. I think I know how to define this, but it might indeed be better if you gave it a shot.

LeaVerou:
[...]

PR #5523, opened 2020-09-17
[mediaqueries-5] Draft text for in-gamut MQ, see #5045

svgeesus:
Has some worked examples.

Specifies a default sRGB if the output device is not known, which is fine for screens but not for printers, hence should.

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1730,6 +1730,115 @@ Color Display Quality: the 'color-gamut' feature</h3>
 	you can use this feature in a negated boolean-context fashion:
 	''not (color-gamut)''.
 
+<h3 id="in-gamut">
+	Displayable Color: the 'in-gamut' feature</h3>
+
+	<pre class='descdef mq'>
+		Name: in-gamut
+		Value: <<color>>
+		For: @media
+		Type:  discrete
+	</pre>
+
+	The '@media/in-gamut' media feature
+	is used to query whether a given <<color>>
+	lies within the <a>gamut</a> of the output device,
+	and thus,
+	it can be accurately displayed.
+
+	<div class="example">
+		The following media query tests whether primary green,
+		in the <a>display-p3</a> colorspace,
+		is in gamut for the output device.
+
+		<pre class="lang-css">@media (in-gamut:color(display-p3 0 1 0)) { … }</pre>
+	</div>
+
+	To determine whether the color is in gamut,
+	it is first converted to CIE XYZ or CIE Lab,
+	and then converted to the colorspace of the output device.
+
+	If the value for each color component,
+	in the output colorspace,
+	lies in the range [0.0-1.0],
+	the color is in-gamut.
+	Otherwise, (if there are negative components, or components greater than one),
+	it is out of gamut.
+
+	<!--
+		This example was calculated with
+		https://colorjs.io/apps/convert/?color=color(display-p3%200%201%200)&precision=4
+	-->
+
+	<div class="example">
+		For example, if the display is an sRGB screen,
+		the display-p3 color from the previous example
+		has the following components: red -2.9058, green 1.0183, blue -1.0162.
+		It is therefore out of gamut for this screen.
+	</div>
+
+	If the colorspace of the output device is unknown,
+	then for Web compatibility,
+	it should be assumed to be <a>sRGB</a>.
+
+	<!--
+		This example was calculated with
+		https://colorjs.io/apps/convert/?color=color(a98-rgb%200.48%200.86%200.95)&precision=3
+	-->
+	<div class="example">
+		The following media query tests whether a sky blue,
+		in the <a>a98-rgb</a> colorspace,
+		is out of gamut for the output device,
+		and if so, displays an out of gamut warning.
+
+		<pre class="lang-css">@media not (in-gamut:color(a98-rgb 0.48 0.86 0.95)) {
+				p.swatch {border: thin dashed red; }
+			}</pre>
+	</div>
+
+	The media query is not restricted to RGB output devices.
+
+	<!--
+		This example was calculated using the ICCMax reference implementation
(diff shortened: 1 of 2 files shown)


mediaqueries-5/Overview.bs
```diff
@@ -1730,6 +1730,115 @@ Color Display Quality: the 'color-gamut' feature</h3>
...
+		MacBeth-Lab.txt
+	-->
+
+	<div class="example">
+		This example tests whether a given color,
+		which represents the cyan on a ColorChecker chart
+		and is specified in FOGRA39 CMYK colorspace,
+		is in gamut for the output device.
+
+		<pre class="lang-css">@color-profile --fogra39 {
```
tabatkins, line 1821: There's no need to do this sort of source contortion - you can just line up all your code properly starting on the next line, and Bikeshed'll ensure the output is good.

tabatkins (approved):

alshamiri5 (approved):

svgeesus:
@florian thoughts?

frivoal:
I think the write up is clear and precise, so it gives a clear understanding of how this works.

However, I think it's missing an explanation of why you'd want to use this, and what kind of situations it's for. The reason I think this may be important is that I suspect this is an easy thing to misunderstand, and to use when you didn't need to, or to fail to notice that it could be applicable to problems you have.

If I'm understanding things correctly, most of the time, the right way to deal with out of gamut colors is not to use this, and instead to reach for [rendering-intent](https://drafts.csswg.org/css-color-4/#descdef-color-profile-rendering-intent). This would be a more specialized tool, that you'd typically use when developing something like a photo editor or something of the sort, where you'd want to display a bright color as a warning if some of the colors end up out of gamut or something. But if it's that, I'm not completely sure why you'd want to check against the gamut of the output device. Checking against so selected target color space that you intent to save your edited image into would make more sense to me. More over, if that's the case, I'd guess this would be most useful not just with raw colors whose value you specify directly, but with colors process through one or more layers of the `color-adjust()` function for colors-5. The syntax description suggests you can do that, but the examples don't show case it.

The fogra39 input being checked against a high-end 7 color printer output almost makes sense as a use case, as that would let you soft-proof something before printing it, showing clear warnings if something would go off. But I don't think that would quite work either, given that as far as I know, CSS User Agents (at least mainstream ones known as browsers) never render to a printer with a specific color space, but always render to PDF, which by itself doesn't have a clearly bounded color space, and so the answer of the query would always be true. Maybe I am misunderstanding this too and it would actually work for some reason that escapes me, or maybe it would indeed only work in a not-a-browser CSS-UA, but then it should be called out in order not to lure people into thinking it would work in browsers while in reality it wouldn't.

Also, whether it's for soft proofing or for image editing, it also feels it would be very useful if in addition to be able to test single colors, you could pass an `<<image>>`, and get to know if any part of the image would be out of gamut.

Depending on what the actual use cases are, I wonder if the fact that you cannot test whether a color / image is in gamut after having been through a [css filter](https://drafts.fxtf.org/filter-effects/#FilterProperty) may be limiting. Whether that's a sign that I don't understand the use cases, or that this is an incomplete tool, I don't know.


Either way, even though I feel I understand **how** it works, I am reasonably confused as to the applicability of this feature. I think we need to have a bit of a description of what this is or isn't for, and to change or expand the examples to relate them not just to the calculation itself, but to actual situations you'd want to use this in.

faceless2:
I can't speak for most of this, but you mentioned PDF and my ears started burning...

Generally I'd agree with Florian's statement that PDF can represent any color in any color-space, so would always return true here. There are specific situations where that might not be the case, but they're always going to be user-agent specific - for example, a PDF can be created with an "Output Intent" color-space, which describes the intended printing conditions for the PDF. This concept used with both PDF/A or PDF/X. Targeting a specific Output Intent is possible with several of the CSS to PDF engines.

It's possible you _might_ want to ensure all colors are within that color-space when you're generating the PDF, but this is a decision based on the intended use of the PDF by whoever is making it. It's not a hard requirement based on the file format.  Specifying an Output Intent is fairly common. Limiting colors to only be within that Output Intent is less common.

What's more common in PDF is disallowing a particular color-space entirely - for example, disallowing all RGB colors, or everything but CMYK. The "in-gamut" test wouldn't really help here; a media-query for this would need to look more like
`@media (colorspace:display-p3)`, so it could be used to select between RGB and CMYK bitmap images (for example). This query would be "can I accept content in this colorspace". However that's obviously a different issue to the one under discussion.

frivoal (changes requested):
Marking as requesting changes, based on https://github.com/w3c/csswg-drafts/pull/5523#issuecomment-701235855

frivoal:
@svgeesus did you close this because you agreed with my skeptical feedback, or did this move some place else?

svgeesus:
Because we were all told off for creating branches in this repo, which messes with the GitHub <-> Mercurial synchronization; and since this PR didn't look like it would be accepted anyway.

LeaVerou:
> Because we were all told off for creating branches in this repo, which messes with the GitHub <-> Mercurial synchronization; and since this PR didn't look like it would be accepted anyway.

So is there a different PR somewhere for this, so that we could iterate?

tabatkins:
And if might be useful to read <https://www.xanthir.com/b4hf0> if you want to still do the PR from your personal fork

frivoal:
@svgeesus Regardless of git branching shenanigans, I'd be curious to hear your response to https://github.com/w3c/csswg-drafts/pull/5523#issuecomment-701235855. I usually feel I have a decent grasp on colors, as evidenced by generally agreeing with you, who definitely know what you're talking about. But here we seem to be in disagreement, so I'm puzzled :)

LeaVerou:
@frivoal The way I see it, the use cases are:

1. More granular detection of the output device's color gamut than the [`color-gamut`](https://drafts.csswg.org/mediaqueries/#color-gamut) media feature.
2. More control over fallback in cases where you don't want to depend on the UA's default gamut mapping algorithm

PR #5532, opened 2020-09-18
Editorial: Removed duplicate "keywords" from Value Definitions section

### LINKED ISSUE #5516: A bunch of specs say "CSS-wide keywords keywords" (repeating the word "keywords") in their "Value Definitions" section
This applies to various specs, as well as the sample "Foo" spec:
https://drafts.csswg.org/css-module-bikeshed/#values
https://www.w3.org/TR/css-display-3/#values
https://www.w3.org/TR/css-sizing-3/#values
...and more.

These specs all say (emphasis added):
>  all properties defined in this specification also accept the CSS-wide **keywords keywords** as their property value

I think the second "keywords" is a typo, right?

Google Search to find the affected specs:
https://www.google.com/search?client=firefox-b-1-d&q=site%3Aw3.org+%22CSS-wide+keywords+keywords%22

CC @tabatkins @fantasai

SebastianZ:
This fixes #5516 by removing the repeated "keywords" words from the Value Definitions section of all specifications.

The reason why so many specs have this error is that the css-module-bikeshed spec. previously had that mistake, which got fixed already by @svgeesus in https://github.com/w3c/csswg-drafts/commit/794fb4708c6a7bf210f9ae399adab260bca0a464.

Sebastian

### DIFF
--- a/css-align-3/Overview.bs
+++ b/css-align-3/Overview.bs
@@ -126,7 +126,7 @@ Value Definitions</h3>
 
 	In addition to the property-specific values listed in their definitions,
 	all properties defined in this specification
-	also accept the <a>CSS-wide keywords</a> keywords as their property value.
+	also accept the <a>CSS-wide keywords</a> as their property value.
 	For readability they have not been repeated explicitly.
 
 <h3 id="partial">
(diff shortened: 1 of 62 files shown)


MatsPalmgren:
Can someone please update the copy that https://api.csswg.org/bikeshed/ uses?
I still get this error when using that as a template.

tabatkins:
What do you mean by that?

MatsPalmgren:
@tabatkins duh, the bug was in my local copy-pasted text. (In my head I thought the whole "Value Definitions" chapter came from the template somehow.)

Issue #5389, opened 2020-08-03
[css-fonts-4] Managing the scaling of font-size within MathML rendering


bkardell:
(Part of #5384 - MathML Core related CSS)

Note: We're very interested in CSSWG feedback on this - it has already changed once based on feedback and tests updated, etc - and there has been some additional feedback that perhsaps this should be a single value instead which indicates that perhaps it should consult another (new property).  We're ivery nterested to nail this one down better.

---

Script levels in MathML are about managing the font-size in subexpressions (like scripts and fractions).  [This example](https://mathml-refresh.github.io/mathml-core/#font-size-scriptlevel-example) illustrates and describes some of the concepts/implications.  The method for doing this is not quite as simple as inheriting, so the MathML-Core spec introduces a new `scriptlevel()` function (See example in the [MathML Core spec](https://mathml-refresh.github.io/mathml-core/#the-math-script-level-property) )
 
Depending on context (exponents, fractions, etc) and particulars of rendering style (compact/inline), the font-size of of parts of expressions scale down at different and fairly well established rates - for example, the scaling factor adjustment of a superscript is not the same as that of a root's prescript.  The example below shows several contextual scaling variations.

![rendering of An exponent with an exponent, a root with a prescript and a fraction containing another fraction](https://mathml-refresh.github.io/mathml-core/examples/example-font-size-scriptlevel.png)

The common rules derive from TexBook, but they are subtle and deserve to be exposed for tweaking and extension by authors.  To this end, MathML adds a new `scriptlevel` function for CSS Fonts Module Level 4 which allows authors to express that the value should be automatically derived (default), or explicitly set to a specific level or that it should be the current level + some integer with rules defined in [section 4.5 of the spec](https://mathml-refresh.github.io/mathml-core/#the-math-script-level-property)

Tests in https://github.com/web-platform-tests/wpt/tree/master/css/css-fonts/math-script-level-and-math-style

faceless2:
We'd implemented the original `math-script-level` property while the core spec was evolving. For what it's worth, this proposal looks a lot better to me. An awful lot better.

Only one suggestion really - would it be an idea to reset the internal script-level to 0 if the font-size is specified as anything other than a `scriptlevel()` function?

I ask because, as specified, "the internal scriptlevel is 0 on the root element" - that's the root element of the document, not of the math equation?

If so, then my issue is that I can't reset the internal script level _and_ change the font-size at the same time. And I don't see anything that limits the use of this to MathML element only. 

[...]

Issue #5387, opened 2020-08-03
[math] math style (minimizing logical height of an equation)


bkardell:
(Part of #5384 - MathML Core related CSS)

In some cases in mathematical rendering it is desirable to minimize the logical height of an equation.  This is generally done by common norms about font-sizing adjustments to parts of the equation, how some operators are drawn, how/where scripts are measured, and how vertical gaps and shifts are applied.  Below is an example of the same equation rendered in two common ways, the one on the left applying the common rules to minimize the logical height, the one on the left, not.

![math with normal and compact math-styles](https://mathml-refresh.github.io/mathml-core/examples/example-math-style.png)

For this purpose, we introduce the `math-style` property with values `normal` (default) and `compact`.

MathML supported this via an attribute called `displaystyle` which was either the string 'true' or the string 'false' ()the rules and the term were derived from The TeXbook).  [MathML-Core introduces](https://mathml-refresh.github.io/mathml-core/#the-math-style-property) the `math-style` CSS property and defines how it works for rendering and layout. It maps this attribute's values as a presentational hint to the CSS property for reasons of compatibility with widely deployed content and the existing ecosystem of tools, but strongly recommends authors use CSS.

Tests in https://github.com/web-platform-tests/wpt/tree/master/css/css-fonts/math-script-level-and-math-style

fred-wang:
@bkardell @fantasai  Given https://github.com/w3c/csswg-drafts/pull/5536#issuecomment-697114884 and that the definition is already present in https://w3c.github.io/mathml-core/#the-math-style-property I guess we can close this, as that was done for https://github.com/w3c/csswg-drafts/issues/5388 and https://github.com/w3c/csswg-drafts/issues/5389  ?

PR #5536, opened 2020-09-22
[css-fonts] Add font-size: math, math-depth and math-style

fred-wang:
This imports the proposal from MathML Core to manage font scaling in math
formulas using a new 'math-depth' property. This mechanism involves a new
'math-style' property that describes compactness of formulas as well as
a new 'math' keyword to specify relative font-size change using the a
scale factor calculated from the 'math-depth' changes.

https://github.com/w3c/csswg-drafts/issues/5389
https://github.com/w3c/csswg-drafts/issues/5387
https://github.com/w3c/csswg-drafts/issues/5389#issuecomment-694361834
https://github.com/w3c/csswg-drafts/issues/5387#issuecomment-690677637


fred-wang:
This probably needs improvements, feedback welcome ;-)

Current test/implementation status is:

Tests: css/css-fonts/math-script-level-and-math-style/ (need to be updated and extended) as well as indirect effect on MathML in mathml/

Chrome:
math-style implemented, https://chromium-review.googlesource.com/c/chromium/src/+/2421110
math-level to be re-implemented, https://bugs.chromium.org/p/chromium/issues/detail?id=1023292
font-size: math not implemented yet

Firefox:
exposing internal -moz-math-display, https://bugzilla.mozilla.org/show_bug.cgi?id=1665975
internal -moz-math-level not exposed yet
font-size: math not implemented yet

Safari:
math-style being implemented, https://bugs.webkit.org/show_bug.cgi?id=216702
math-level not implemented yet
font-size: math not implemented yet

fantasai:
Hi @fred-wang ! I think we could use a bit of discussion about where exactly all this goes, but I do suspect it fits better into the MathML Core spec rather than the css-fonts-4 spec. The core processing of this all is fitted to the MathML layout model, and not particularly relevant generally to CSS, is why I'm thinking this.

(Note: for the `display` and `font-size` keywords, if you're using Bikeshed it's possible to use class="propdef partial" to indicate adding to, rather than redefining, the property; I'm not sure what the equivalent syntax would be in ReSpec.)

css-fonts-4/Overview.bs
```diff
@@ -838,7 +838,7 @@ Font size: the 'font-size' property</h3>
 
     <pre class="propdef">
     Name: font-size
-    Value: <<absolute-size>> | <<relative-size>> | <<length-percentage>>
+    Value: <<absolute-size>> | <<relative-size>> | <<length-percentage>> | math
```
fred-wang, line 841: This change is a mistake, it must be reverted

fred-wang:
Closing this per https://github.com/w3c/csswg-drafts/pull/5536#issuecomment-697114884

- math-depth and math-style are in MathML Core.
- font-size: math was added in https://github.com/w3c/csswg-drafts/commit/09efeb0f1948f5aec4afc3672c1ab0c264abab36

PR #5538, opened 2020-09-22
[css-font-loading-3] Add override attributes to FontFace interface following resolution in #4792

xiaochengh:
[css-font-loading-3] Add override attributes to FontFace interface following resolution in https://github.com/w3c/csswg-drafts/issues/4792#issuecomment-693528301


xiaochengh:
@tabatkins Could you review this?

Issue #3846, opened 2019-04-18
[css-color-adjust-1] What happens with multiple <meta>s?


tabatkins:
What should be the behavior if there are multiple `<meta name=color-scheme>` elements with different values?

My preference is that we honor the first and ignore the rest. This allows us to paint the background correctly as soon as possible. 

However, Device Adaptation specifies that the relevant `meta`s should be interpreted as `style` elements specifying a `@viewport`, so that they can cascade together; this means that the last `meta` affecting a particular @viewport descriptor "wins".

What rule should we use?

lilles:
When you say "honor the first", what does that mean for meta elements inserted by a script?

- If there are no color-scheme meta elements and you insert one?
- If there is a color-scheme meta and you insert another one before it with a script in head?

Just apply the first that is inserted into the document either via parser or script?

tabatkins:
The details are in the spec now for that - whenever the set of metas is changed, we re-evaluate which one is first.

smfr:
We should do whatever browsers do now for other meta tags. For the viewport meta tag (and other meta tags as far as I can see), iOS uses the last one. This sounds like something HTML should spec.

lilles:
Fwiw, theme-color is spec'ed to be the first (which is also what Blink implements):

https://html.spec.whatwg.org/multipage/semantics.html#meta-theme-color

For viewport, the Blink implementation applies the last that was inserted into the document, or had its attributes modified, not necessarily the last in the tree-order. This is a behavior we inherited from WebKit.

smfr:
Has the ship sailed on HTML speccing the same behavior for all <meta> tags?

domenic:
FWIW of the meta tags that HTML defines processing models for, the following are the behaviors I see:

- `application-name`:
  - First that matches one of the user's preferred languages based on the `lang=""` attribute.
  - Does not state how to handle dynamic changes; just gives an algorithm for one-shot determining the application name.
  - Unclear whether ones inside Shadow DOM are ignored or counted (uses ambiguous "in a Document")
  - Inside and outside the `head` are treated the same.
- `keywords`:
  - concatenate, split on commas, remove duplicates.
  - Does not state how to handle dynamic changes.
  - Very unclear whether ones inside Shadow DOM are ignored or counted.
  - Inside and outside the `head` are treated the same.
- `referrer`:
[...]

PR #5540, opened 2020-09-23
[css-color-adjust-1] Move color-scheme meta to HTML #3846

lilles:
Per resolution: https://github.com/w3c/csswg-drafts/issues/3846#issuecomment-648923121.

Add a reference to the HTML specification from the specification of the color-scheme property.

HTML pull request: https://github.com/whatwg/html/pull/5938


fantasai (changes requested):
Should be a Note, since the normative definition is in HTML, right?
Also, please move it below the note about what light and dark mean.
Otherwise seems fine.

PR #5543, opened 2020-09-23
[css-display-3] Link to definition of inline block

twilco:
I think this link to the definition of inline block would be helpful because inline blocks are not defined [in the glossary](https://drafts.csswg.org/css-display/#glossary) with the other box types.  Inline blocks are instead only defined as an alias to inline boxes [here in the generated box table](https://drafts.csswg.org/css-display/#inline-block), which one might miss when reading the paragraph in this diff.


fantasai:
Fair. Merged. :)

Issue #4931, opened 2020-04-09
[css-grid][css-contain] Clarify that `contain:size` affects track sizing


MatsPalmgren:
Consider the following case:
```html
<div style="display:inline-grid; contain:size;">
  <item style="background:red">Hello Kitty</item>
</div>
```

My understanding of the CSS Containment spec is that the intent of `contain:size` is that all size calculations should be done as if the element has no children:
https://drafts.csswg.org/css-contain/#containment-size
Chromium and Gecko agrees that the content-box size in the above case is 0x0, which is correct. I tend to think that the track sizes also should be zero and thus that it should render the same as:
```html
<div style="display:inline-grid; grid:0/0;">
  <item style="background:red">Hello Kitty</item>
</div>
```
Neither Chromium nor Gecko does that currently, but I intend to fix this in Gecko ([bug 1488878](https://bugzilla.mozilla.org/show_bug.cgi?id=1488878)).

I can't find any spec text dealing with this case explicitly though, so I would like CSS Grid to clarify that  when the Grid container has `contain:size` the [Track Sizing Algorithm](https://drafts.csswg.org/css-grid/#algo-track-sizing) should be run as if the container has no items. This will have the effect that any [intrinsic track sizes](https://drafts.csswg.org/css-grid/#intrinsic-sizing-function) will be zero.

MatsPalmgren:
(I've filed a Chromium [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1069329).)

bfgeek:
I don't believe that is the intent of the containment specification. From the spec:

> When calculating the size of the containing box, including when computing its intrinsic size, it must be treated as having no contents.

This is effectively just an early out when computing the intrinsic sizes for an element.

> Then, its contents must then be laid out into the containing box's resolved size.

Any layout algorithm should not then change its behaviour after the above step. E.g. these two cases should render the same:

```html
<div style="display:inline-grid; contain:size;">
  <item style="background:red">Hello Kitty</item>
</div>
```

```html
<div style="display:inline-grid; width: 0px;">
  <item style="background:red">Hello Kitty</item>
</div>
```

Loirooriol:
Chromium and Firefox are not interoperable in these cases:

```html
<div style="display: inline-grid; grid-auto-columns: 50px; height: 50px; contain: size; border: 5px solid">
  <div style="background: yellow"></div>
</div>
```

```html
[...]

PR #5596, opened 2020-10-08
[css-contain-2] Clarify how sizing containment works

frivoal:
This is not meant to change the expected behavior, however the original text was not very clearly phrased, hence this rewrite.

This is an attempt to solve #4931, currently written against level 2, but to be back ported to the L1 REC if we like it.

### DIFF
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -46,6 +46,10 @@ spec:css-sizing-4; type:property; text:contain-intrinsic-size
 spec:css-sizing-4; type:property; text:aspect-ratio
 spec:intersection-observer; type:dfn; text:intersection root
 spec:css-sizing-3; type:value; for:height; text:auto
+spec:css-sizing-3; type:property; text:width
+spec:css-sizing-3; type:property; text:height
+spec:css-sizing-3; type:value; for:width; text:min-content
+spec:css-sizing-3; type:value; for:width; text:max-content
 </pre>
 
 <h2 id='intro'>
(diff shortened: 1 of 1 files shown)


frivoal:
The diff is reasonably hard to read, so here's a rendered version, for ease of review (this replaces the first item in ordered list in [§ 3.1](https://drafts.csswg.org/css-contain-2/#containment-size)):


> 
> 1. Laying out a size-contained box and its content is conceptually done in two phases:
> 
>     **Sizing as if empty**
>     The used width and height of the containing box are determined as if performing a normal layout of the box, except that it is treated as having no content—not even through pseudo elements such as ::before, ::after, or ::marker.
> 
>     Replaced elements must be treated as having an intrinsic width and height of 0.
> 
>     All CSS properties of the containing box are taken into account as they would be when performing layout norally.
> 
>     Note: Even when the element’s sizing properties are auto this does not necessarily make the element zero-sized: properties set on the element itself, such as the columns property or the grid property, continue to be taken into account.
> 
>     **Laying out in-place**
>     The containing box's content (including any pseudo-element) must then be laid out into the now fixed-size containing box normally. 
> 
>     Note: size containment does not suppress baseline alignment. See layout containment for that.
> 
> 2. The intrinsic sizes of the containing box are determined as if the element had no content, following the same logic as when sizing as if empty.
> 
>     Note: This affects explit invocations of the min-content or max-content keywords, as well as any calculation that depends on these measurement, such as sizing grid tracks into which a size contained item is placed, or if fit-content sizing the contained box’s parent.

frivoal:
Also note this commented out line in the source: https://github.com/w3c/csswg-drafts/pull/5596/files#diff-5832b51ac3ccd578073f26e6b031fa9eR532, which would allow us to let other spec opt into excluding some properties from the *sizing as if empty* phase.

frivoal:
Closing in favor of #5643 the same thing rebased on the latest master branch.

PR #5600, opened 2020-10-09
Update Overview.bs

lvmm:
Changed initial from auto to none for text-decoration-skip-ink-property. Reason: the auto setting leaves the default behaviour at the discretion of implementer which is never a good thing - a standard should be well defined and unambiguous, and may change the look of sites without knowledge of their creators and owners which is not a good idea either.


astearns:
@lvmm changes like these are made by the editors after some discussion and consensus is reached in the working group over the change. If you want to argue for making this change, please open an issue.

PR #5603, opened 2020-10-09
Add a dfn for -webkit-user-select

gsnedders:


frivoal (approved):

frivoal:
Thanks, @gsnedders .

Note: It's usually preferred to send pull requests from cloned repositories rather than make branches within the csswg-drafts repository, because https://drafts.csswg.org/ builds the whole spec suite for every branch in that repository (served at https://drafts.csswg.org/branch/gsnedders-patch-1 for this patch, if that's still live by the time you look at it), which slows down the server quite a bit when there are many of those.

If you need to a long lived branch with server side builds, then fine, but this type of PR wasn't that, and so it's best done as a PR from a fork.

PR #5616, opened 2020-10-13
[css-nesting] additional examples

argyleink:
Added examples with nested `:is` and `:not` and an invalid example. Publishing draft PR, will be adding to it over the next week or so.

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -232,7 +232,7 @@ Direct Nesting {#direct}
 		}
 		/* equivalent to
 		   .foo { color: blue; }
-		   .foo .foo .bar .foo .baz .foo .qux { color: blue; }
+		   .foo .bar .foo .baz .foo .qux { color: blue; }
 		 */
 
 		.foo {
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -203,8 +203,8 @@ Direct Nesting {#direct}
 			& > .bar { color: red; }
 		}
 		/* equivalent to
-		   .foo { color: blue; }
-		   .foo > .bar { color: red; }
+			 .foo { color: blue; }
+			 .foo > .bar { color: red; }
 		 */
 
 		.foo {
@@ -212,8 +212,8 @@ Direct Nesting {#direct}
 			&.bar { color: red; }
 		}
 		/* equivalent to
-		   .foo { color: blue; }
-		   .foo.bar { color: red; }
+			 .foo { color: blue; }
+			 .foo.bar { color: red; }
 		 */
 
 		.foo, .bar {
@@ -221,9 +221,9 @@ Direct Nesting {#direct}
 			& + .baz, &.qux { color: red; }
 		}
 		/* equivalent to
-		   .foo, .bar { color: blue; }
-		   :is(.foo, .bar) + .baz,
-		   :is(.foo, .bar).qux { color: red; }
+			 .foo, .bar { color: blue; }
+			 :is(.foo, .bar) + .baz,
+			 :is(.foo, .bar).qux { color: red; }
 		 */
 
 		.foo {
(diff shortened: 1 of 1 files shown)


css-nesting-1/Overview.bs
```diff
@@ -225,6 +225,20 @@ Direct Nesting {#direct}
...
+
+		ul, ol {
+			&:not(.foo,.bar) > .baz { color: red; }
+		}
+		/* equivalent to
+			 :is(ul,ol):not(.foo,.bar) > .baz { color: red; }
+		 */
+
+		.foo {
+			&:is(.bar, &.baz) { color: red; }
```
andruud, line 237: Hmm, I'm not sure this (`&.baz`) is valid without `@nest`.

css-nesting-1/Overview.bs
```diff
@@ -248,6 +262,12 @@ Direct Nesting {#direct}
 		}
 		/* Invalid because the second selector in the list doesn't
 		   contain a nesting selector. */
+
+		.foo {
+			color: red;
+			&&.bar { color: blue; }
+		}
+		/* Invalid because the 1st & isn't at the start of the selector */
```
andruud, line 270: Do you mean the 2nd &?

css-nesting-1/Overview.bs
argyleink: I agree, I think an `@nest` should be required here

css-nesting-1/Overview.bs
argyleink: I didnt. Let me update the verbiage to be more clear 👍

css-nesting-1/Overview.bs
tabatkins: No, that's fine. The only requirement is that the overall selector needs to start with an `&` (or else it'll instead be parsed as a (likely invalid) declaration. @nest lets you violate that rule and do whatever.

So in this case it's fine, the selector starts with an `&`.

css-nesting-1/Overview.bs
```diff
@@ -225,6 +225,72 @@ Direct Nesting {#direct}
...
 		   :is(.foo, .bar).qux { color: red; }
 		 */
+
+		.foo {
+			color: blue;
+			& .bar & .baz & .qux { color: red; }
+		}
+		/* equivalent to
+		   .foo { color: blue; }
+		   .foo .foo .bar .foo .baz .foo .qux { color: blue; }
```
tabatkins, line 235: Where'd this first `.foo` come from?

css-nesting-1/Overview.bs
```diff
@@ -225,6 +225,72 @@ Direct Nesting {#direct}
...
+
+		.foo {
+			color: blue;
+			& { padding: 2ch; }
+		}
+		/* equivalent to
+		   .foo { 
+		   	 color: blue; 
+		   	 padding: 2ch;
+		   }
```
tabatkins, line 246: Would probably expand this more literally *first* (as two rules that both have `.foo` as the selector), then say it's also equivalent to this "both in one rule" example.

css-nesting-1/Overview.bs
```diff
@@ -242,6 +308,13 @@ Direct Nesting {#direct}
 		}
 		/* Invalid because & isn't in the first compound selector */
 
+		.foo {
+			color: red;
+			&&.bar { color: blue; }
+		}
+		/* Invalid because the 1st & is no longer 
+		   at the start of a nesting selector */
```
tabatkins, line 316: No, invalid because the *second* `&` isn't at the start of a *compound* selector.

tabatkins (changes requested):

css-nesting-1/Overview.bs
argyleink: a mistake 👍  removed it thx!

css-nesting-1/Overview.bs
```diff
@@ -242,12 +313,19 @@ Direct Nesting {#direct}
 		}
 		/* Invalid because & isn't in the first compound selector */
 
+		.foo {
+			color: red;
+			&&.bar { color: blue; }
+		}
+		/* Invalid because the 1st & is no longer
```
tabatkins, line 320: Still wrong; the first `&` definitely is at the start of the compound selector. It's the second that's causing the problem. ^_^

argyleink: updated 👍

tabatkins:
k, feel free to un-Draft this PR when you're ready

tabatkins (approved):

Issue #5421, opened 2020-08-12
[css-fonts-4] The spec should not allow UAs to ignore user-installed fonts


r12a:
10.3. Preinstalled Fonts and User-Installed Fonts
https://drafts.csswg.org/css-fonts-4/#preinstalled-and-user-installed-fonts

> User Agents may choose to ignore User-Installed Fonts for the purpose of the Font Matching Algorithm.

!!!! This will create consternation for a great many international users around the world. A huge number of people use languages which are not supported, or are not supported well, by preinstalled fonts. The almost universal approach, currently, for people who use less common or endangered languages on the Web is to find a page where someone has created a font (and often keyboard) that you are expected to download in order to see or create content.  (Those fonts are often used by applications other than browsers, too.)

Furthermore, quite often there are also particular font styles that content authors want to see used, often contrastively, and there's a likelihood that a user will have installed additional fonts (such as using a Kano font for Nigerian ajami text, or a Mool font style to distinguish Khmer headings, etc.) in order to make the text readable/locally relevant. For many languages, the system may provide a Noto font, but usually only the one (and generally the Noto fonts don't reflect the design users would want to see for their text, since it focuses on producing Noto-harmonised, large, sans-serif, monoline glyphs).

Of course, this plays into the issues related to privacy, but if the motivation for the sentence is that, it should say so, at least. But that discussion is not yet concluded, so in the meantime the i18n WG would like to see that sentence removed.

xfq:
If I remember correctly, the font based fingerprinting discussions started last year, but this text has been in the spec for at least [2 years](https://github.com/w3c/csswg-drafts/commit/a6ef2d9420509ec3a9ac9c9f899d4d0588381381#). Perhaps it's for another reason?

astearns:
@litherum knows better, but my understanding is that this sentence was added mostly to describe reality - in that Safari had already implemented this behavior and continues to behave this way. The subsequent fingerprinting discussion has happened in part to figure out how to address problems with this behavior.

Two of the suggestions we have considered are:

1) Allow some User-Installed Fonts to match for particular languages where many web pages depend on them
2) User agents that have this font-matching behavior MUST provide UI to let the user opt-in to User-Installed Font matching

Would either or both of these requirements make the behavior more acceptable to the i18n WG?

r12a:
> Safari had already implemented this behavior

[...]

PR #5625, opened 2020-10-15
[css-fonts-4] Add explanatory text about tradeoffs on blocking local fonts, particularly wrt i18n considerations

litherum:
Closes https://github.com/w3c/csswg-drafts/issues/5421.


frivoal:
Would you have a problem tweaking this by adding something like:

> User agents that choose to make some fonts unavailable by default for privacy reasons should offer a way for users to make specific fonts available.

astearns (approved):

litherum:
> Would you have a problem tweaking this by adding something like:
> 
> > User agents that choose to make some fonts unavailable by default for privacy reasons should offer a way for users to make specific fonts available.

I would have a problem with this. It's a user agent's choice whether or not they want to allow users to opt-in to bad privacy. Also, CSS should not dictate what UI is present in the browser chrome. Adding browser UI has cost, both on the developer and the user (in the form of additional complexity) and it's up to the UA to determine if the cost is worth it.

All we can do here is make the point that some fonts might be required for i18n, and let the UA determine how best to satisfy that requirement.

jfkthame:
> > Would you have a problem tweaking this by adding something like:
> > > User agents that choose to make some fonts unavailable by default for privacy reasons should offer a way for users to make specific fonts available.
> 
> I would have a problem with this. It's a user agent's choice whether or not they want to allow users to opt-in to bad privacy.

There's a trade-off between privacy and functionality here; we all realize that. The functionality involved is sometimes purely a matter of aesthetics, but other times it may make the difference between a user being able to read a page or not. While a user agent may make various choices with regard to *default* behavior, it seems severe for a *user* agent to entirely block functionality that may be critical to the usability of web pages the user wants to see.

Florian's proposed text uses *should* rather than *must*, so it does not impose a mandatory requirement on UAs, but I think it sends a worthwhile signal about the importance of this issue.

fantasai:
> It's a user agent's choice whether or not they want to allow users to opt-in to bad privacy.

It's our job to design tech that supports all languages worldwide, and it shouldn't be conformant to make it practically impossible for a normal user to read minority-language pages on their computer.

css-fonts-4/Overview.bs
```diff
@@ -6146,7 +6146,9 @@ Web Fonts shadow Installed Fonts, so if an Installed Font has the same family na
 <h3 id="preinstalled-and-user-installed-fonts">
 Preinstalled Fonts and User-Installed Fonts</h3>
 
-Users may choose to install fonts on their devices. User-Installed Fonts are installed by an explicit action by the user, such as clicking an "Install" button or copying a file into a particular directory on their device. Such fonts are User-Installed Fonts and also are Installed Fonts. Web content authors should not expect the presence of user-installed fonts, because there is no guarantee any user will have performed the action to install a specific font. User Agents may choose to ignore User-Installed Fonts for the purpose of the <a href="#font-matching-algorithm">Font Matching Algorithm</a>.
+Users may choose to install fonts on their devices. User-Installed Fonts are installed by an explicit action by the user, such as clicking an "Install" button or copying a file into a particular directory on their device. Such fonts are User-Installed Fonts and also are Installed Fonts. Web content authors should not expect the presence of user-installed fonts, because there is no guarantee any user will have performed the action to install a specific font.
```
fantasai, line 6149: s/Users may/Users can/ since this is a statement of fact not an allowance in this spec.

(I also, btw, recommend adopting https://rhodesmill.org/brandon/2012/one-sentence-per-line/ since it makes diffing a lot easier...)

css-fonts-4/Overview.bs
```diff
@@ -6146,7 +6146,9 @@ Web Fonts shadow Installed Fonts, so if an Installed Font has the same family na
...
+
+Note: The set of <a href="#installed-fonts">installed fonts</a> available in the <a href="#font-matching-algorithm">Font Matching Algorithm</a> is explicitly undefined. The available set of fonts <a href="https://panopticlick.eff.org">is</a> used by trackers to fingerprint users and reduce their privacy. However, some <a href="#installed-fonts">installed fonts</a>, even some <a href="#preinstalled-and-user-installed-fonts">user-installed fonts</a>, are required to make languages readable. User Agents may choose to make all installed fonts available for language support and design integrity reasons, or may choose to make some fonts unavailable for privacy reasons. In addition, User Agents may have additional facilities for fine-tuning this balance, such as interfaces to allow users to explicitly make certain fonts available or unavailable. Different User Agents, even running on the same Operating System, are expected to strike different balances here.
```
fantasai, line 6151: If you want to allow the UA to exclude certain installed fonts from the list of fonts available in the Font Matching Algorithm, add an explicit normative statement that says that. Adding a statement here that the set of fonts available to the Font Matching Algorithm is undefined is disingenuous. Also, putting it in a note implies that it's a restatement of a normative definition elsewhere.

If you want to assert that such a normative definition exists because you haven't explicitly forbidden such subsetting, then it's not clear to me that the allowance is limited to installed fonts only: if you're allowing subsetting of installed by omission of forbidding it, doesn't that mean you're also allowing subsetting of web fonts, because that also is not forbidden?

fantasai (changes requested):
Lastly, I don't think this closes the issue, so I would omit that comment from the commit message so that you can commit any improvements in this PR without closing the issue.

litherum:
@fantasai

> It's our job to design tech that supports all languages worldwide

It's also our job to protect our users' privacy.

You're right that there are good arguments for allowing some user-installed fonts. There are also good arguments for disallowing all of them, too. That's the point I'm trying to make - it's up to each UA to balance these competing desires.

frivoal:
> it's up to each UA to balance these competing desires.

I agree.

> It's also our job to protect our users' privacy.

By turning off local fonts **by default** you are protecting your users' privacy. This is indeed an acceptable choice given that tradeoff.

However, by making it impossible to turn certain fonts back on even when the user cannot read the page without them, you are not protecting that user's privacy, you're turning them into a non user: they cannot use the browser to read the page they wanted.

(Semi-sarcastic / semi-true addition: in that case, the browser is no more useful to that user than a brick, since neither a brick nor a browser than cannot display their language will let them browse that page. However, the browser is worse than a brick in terms of privacy, since an actual brick will have zero fingerprinting surface, but which is not the case of the no-local-font browser)

css-fonts-4/Overview.bs
svgeesus: This focuses totally on UAs, with little mention of user action to allow or disallow particular fonts (per-site or globally).

svgeesus: > If you want to allow the UA to exclude certain installed fonts from the list of fonts available in the Font Matching Algorithm, add an explicit normative statement that says that.

Strongly agree.

css-fonts-4/Overview.bs
svgeesus: perhaps /should not expect/should not count on/

svgeesus (changes requested):

svgeesus:
As this PR has bitrotted in the interim, and requested changes have not been made, I pulled out the proposed text, added the requested changes, and added it in https://github.com/w3c/csswg-drafts/commit/b0ac8dd0ee00450ead86efdb79e30d0781ea6f28

Issue #5627, opened 2020-10-16
[cssom] [mediaqueries] Media query serialization doesn't work for newer spec features


emilio:
I recently became confused because WebKit / Blink sorted the media features in media query lists, but @lilles [did point out](https://bugs.chromium.org/p/chromium/issues/detail?id=1138859) to me that https://drafts.csswg.org/cssom/#serialize-a-media-query does say that:

> 5. Sort the media features in lexicographical order.

But this has gone very out of date with newer mediaqueries features (like nested expressions, etc), which for now only Gecko supports. This also gets more interesting when we support the multi-value expression syntax in https://github.com/w3c/csswg-drafts/issues/2791...

Gecko has never implemented the sorting step, so I'd propose to drop it: we don't know of any compat issue due to that, it was just that I was looking at interop around `MediaQueryList` as a result of a Gecko bug. Also, Gecko's behavior is more intuitive to me but maybe I'm biased :)

If we don't drop it, we need to define how the sorting works for stuff like nested expressions and such. Basically, we'd need to define sorting among two arbitrary `<media-condition>`s.

Thoughts?

lilles:
Drop sorting. I can't imagine anyone relying on non-interoperable re-ordering of media query expressions (famous last words).

Blink also currently removes duplicates, which isn't part of the algorithm, but is indicated by an example in the spec. We should remove that too.

tabatkins:
Yup, we should just preserve the expression as-is.

I'll go fix the example that shows off deduplication.

PR #5638, opened 2020-10-19
[cssom-1] Drop sorting and deduplication from media query serialization.

emilio:
Per https://github.com/w3c/csswg-drafts/issues/5627#issuecomment-712475204

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -375,8 +375,6 @@ To
  single SPACE (U+0020), followed by "<code>and</code>", followed by a single SPACE
  (U+0020), to <var>s</var>.
 
- <li>Sort the [=media features=] in lexicographical order.
-
  <li>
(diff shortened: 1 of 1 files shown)


cssom-1/Overview.bs
```diff
@@ -375,7 +375,7 @@ To
  single SPACE (U+0020), followed by "<code>and</code>", followed by a single SPACE
  (U+0020), to <var>s</var>.
 
- <li>Sort the [=media features=] in lexicographical order.
+ Issue: The `and` bit needs updating for newer spec features.
```
tabatkins, line 378: No, that's still correct; a type is *always* joined to the rest of the MQ with an `and`. You can't say `print or (grid)`.

tabatkins (changes requested):
r+ except for the unnecessary issue; the rest looks good.

PR #5643, opened 2020-10-21
[css-contain-2] Clarify how sizing containment works

frivoal:
This is not meant to change the expected behavior, however the original text was not very clearly phrased, hence this rewrite.

This is an attempt to solve #4931, along the lines of the so-called option α of https://github.com/w3c/csswg-drafts/issues/4931#issuecomment-704676915. It is currently written against level 2, but to be back ported to the L1 REC if we like it.

The diff is reasonably hard to read, so for ease of review, here's the resulting new section 3.1: http://files.florian.rivoal.net/css-contain-2/Overview.html#containment-size

Also note the commented out line in the source, which would allow us to let other specs opt into excluding some properties from the *sizing as if empty* phase, in order to do the so-called option β of https://github.com/w3c/csswg-drafts/issues/4931#issuecomment-704676915

### DIFF
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -480,7 +480,31 @@ Size Containment</h3>
 	contain-size-056.html
 	</wpt>
 
-	1. Laying out a [=size containment box=] and its content is conceptually done in two phases:
+	1.
+		The [=intrinsic sizes=] of the [=size containment box=]
+		are determined as if the element had no content,
+		following the same logic as when [=sizing as if empty=].
+
+		Note: This affects explicit invocations of the ''min-content'' or ''max-content'' keywords,
+		as well as any calculation that depends on these measurement,
+		such as sizing <span class=informative>[=grid tracks=]</span> into which a size contained item is placed,
+		or if [=fit-content sizing=] the containment box's parent.
+
+		<wpt>
+		contain-size-013.html
+		contain-size-041.html
+		contain-size-042.html
+		contain-size-replaced-001.html
+		contain-size-replaced-002.html
+		contain-size-replaced-003a.html
+		contain-size-replaced-003b.html
+		contain-size-replaced-003c.html
+		contain-size-replaced-004.html
+		contain-size-replaced-005.html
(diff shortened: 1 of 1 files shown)


frivoal:
Here's a nicer preview: https://drafts.csswg.org/branch/contain-4931-option-alpha/css-contain-2/#containment-size

css-contain-2/Overview.bs
```diff
@@ -528,7 +573,15 @@ Size Containment</h3>
 		contain-size-baseline-001.html
 		</wpt>
 
-		<a>Replaced elements</a> must be treated as having an intrinsic width and height of 0.
+	2.
+		The [=intrinsic sizes=] of the [=size containment box=]
+		are determined as if the element had no content,
+		following the same logic as when [=sizing as if empty=].
+
+		Note: This affects explit invocations of the ''min-content'' or ''max-content'' keywords,
```
tabatkins, line 581:
```suggestion
		Note: This affects explicit invocations of the ''min-content'' or ''max-content'' keywords,
```

css-contain-2/Overview.bs
```diff
@@ -544,40 +597,32 @@ Size Containment</h3>
...
 
 		<wpt>
 		contain-size-breaks-001.html
 		contain-size-monolithic-001.html
 		</wpt>
 
+	<div class=example>
+		Given the following markup and style, the image would be sized to 100px by 100px,
+		as the aspect ratio set by the 'aspect-ratio' property takes effect.
+		<pre><code class=lang-css>
```
tabatkins, line 610: Just checking - does this highlight properly?

tabatkins (changes requested):

frivoal:
the csswg draft server isn't building this PR properly. In the meanwhile you can read it from here: http://files.florian.rivoal.net/css-contain-2/Overview.html#containment-size

css-contain-2/Overview.bs
frivoal: Yes. I was hoping it would work, and it just did. Is the markup considered bad?

PR #5645, opened 2020-10-21
[css-nesting-1] initial idea for nested conditionals

argyleink:
[css-nesting-1] 
This PR adds a section on nesting conditionals and provides a small set of examples to get the conversation started.

**Example**
```css
.foo {
  display: grid;

  @media (orientation: landscape) {
    & { 
      grid-auto-flow: column; 
    }
  }
}

/* equivalent to */
.foo { display: grid; }

@media (orientation: landscape) {
  .foo { 
    grid-auto-flow: column; 
  }
}
```

fixes https://github.com/tabatkins/specs/issues/56

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -432,6 +432,64 @@ The Nesting At-Rule: ''@nest'' {#at-nest}
 		</pre>
 	</div>
 
+Nesting Conditional Rules {#conditionals}
+-----------------------------------------------
+
+  A style rule can have any number of <a>conditional rules</a> inside of it, of any type,
+  intermixed with any number of declarations, in any order.
+
+  The presence of a nested conditional engages the same logic as if ''@nest'' was present. 
+  The nested conditional rules must contain a <a>nesting selector</a> and follow the same 
+  rules as outlined in <a href="#direct">direct nesting</a>. 
+
+  <div class="example">
+    For example, the following conditional nestings are valid:
+
+    <pre class="lang-css">
+    .foo {
+      display: grid;
+
+      @media (orientation: landscape) {
+        & { 
+          grid-auto-flow: column; 
+        }
+      }
+    }
+    /* equivalent to
+       .foo { display: grid; }
+
+       @media (orientation: portrait) {
+         .foo { 
+           grid-auto-flow: column; 
+         }
+       }
+     */
+    </pre>
+
+    But the following are invalid:
+
+    <pre class=lang-css>
+    .foo {
+      display: grid;
+
+      @media (orientation: landscape) {
+        grid-auto-flow: column;
+      }
+    }
+    /* Invalid because there's no nesting selector */
+
+    .foo {
+      color: red;
+      
+      @media (min-width: 480px) {
+        & h1, h2 { color: blue; }
+      }
+    }
+    /* Invalid because not all selectors in the list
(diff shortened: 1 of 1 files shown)


argyleink:
I'm considering adding an invalid example that tries to use `@nest @media (...) {}`, I think we should disallow it. Thoughts?

tabatkins:
Yes, @nest takes a selector as its prelude and properties as its body; it's not just a generic token you throw in front of anything getting nested. Might be worthwhile illustrating that. ^_^

fromaline:
@argyleink very useful idea at first glance! <br>
Currently this is not in spec, as i see from [tabatkins/specs#56](https://github.com/tabatkins/specs/issues/56)
```css
.foo {
  color: red;
  @media (min-width: 20rem) {
    color: blue;
    @media (max-width: 30rem) {
      color: green;
    }
  }
}
/* equivalent to */
.foo {
  color: red;
}
@media (min-width: 20rem) {
  .foo {
    color: blue;
  }
}
@media (min-width: 20rem) and (max-width: 30rem) {
  .foo {
    color: green;
  }
}
```

You've add this example.
```css
.foo {
  display: grid;

  @media (orientation: landscape) {
    & { 
      grid-auto-flow: column; 
    }
  }
}
/* equivalent to */
.foo { display: grid; }

@media (orientation: landscape) {
  .foo { 
    grid-auto-flow: column; 
  }
}
```

So could we nest `@media` inside `&`? 
For me it seems fully valid, because `&` refers to parent selector and from your example we could nest media inside a selector
```css
.foo {
  display: grid;

  & {
    @media (orientation: landscape) {
      grid-auto-flow: column;
    }
  }
}
/* equvialent to */
.foo {
  display: grid;
}

@media (orientation: landscape) {
  .foo {
    grid-auto-flow: column;
  }
}
```

So if we combine upper syntax we could achieve `@media` nesting. Should it work?
```css
.foo {
  display: grid;
  
  @media (max-width: 500px) {
    & {
      @media (orientation: landscape) {
        grid-auto-flow: column;
      }
    }
  }
}
/* is equvialent to? */
.foo {
  display: grid;
}

@media (max-width: 500px) and (orientation: landscape) {
  .foo {
    grid-auto-flow: column;
  }
}
```

argyleink:
Thanks for requesting clarification on nesting conditionals inside conditionals and posting some pseudo code 🙂 

I'd like to make a spec addition that demonstrates the syntax, here's my current version:

```css
.foo {
  display: grid;

  @media (orientation: landscape) {
    & { 
      grid-auto-flow: column; 
    }

    @media (min-inline-size > 1024px) {
      & {
        max-inline-size: 1024px;
      }
    }
  }
}
/* equivalent to
   .foo { display: grid; }

   @media (orientation: landscape) {
     .foo { 
       grid-auto-flow: column; 
     }
   }

   @media (orientation: landscape) and (min-inline-size > 1024px) {
     .foo { 
       max-inline-size: 1024px;
     }
   }
 */
```

nested conditionals were already triggering nesting, so by putting one a conditional inside a conditional, the `&` sigil resolves to the nearest resolvable selector, in this case `.foo`

less to write, and still consistent with the syntax?

fromaline:
Yeah, I see why your syntax is more consistent. It requires a selector (`&`) inside `@media` querie as opposed to my syntax, which does not follow this rule.

PR #5666, opened 2020-10-27
[css-animations-2] Describe behavior for changes to animation-timeline

andruud:


andruud:
@kevers-google @ogerchikov We need to somehow connect the recent work re. setTimeline with CSS Animations. What do you think about this?

PR #5673, opened 2020-10-28
[css-scoping-1] Let :host/:host-context/::slotted take a single compound

### LINKED ISSUE #2158: [css-scoping] Consider making ::slotted and :host take a single argument.
As far as I know, no engine implements ::slotted with multiple arguments, and keeping it with a single argument allows to optimize it more easily (see https://bugzilla.mozilla.org/show_bug.cgi?id=1425757).

andruud:
This takes the liberty of also applying the related resolution to `:host-context`, although that's not mentioned in the issue.

Resolves #2158

### DIFF
--- a/css-scoping-1/Overview.bs
+++ b/css-scoping-1/Overview.bs
@@ -259,7 +259,7 @@ Selecting Into the Light: the '':host'', '':host()'', and '':host-context()'' ps
 	The <dfn selector id="selectordef-host-function">:host()</dfn> function pseudo-class
 	has the syntax:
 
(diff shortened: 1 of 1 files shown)


andruud:
cc @tabatkins @emilio 

I've been trying to understand exactly how specificity is supposed to work for :host() etc. Currently the spec says "that of a pseudo-class, plus the specificity of its argument", without sufficiently handling the fact that "the argument" is a _list_. Anyway, this becomes a non-issue if we only allow one selector.

Issue #4778, opened 2020-02-14
[cssom-view] Inconsistencies between partially offscreen scroll-into-view


bokand:
I've noticed some inconsistencies between engines in how scrollIntoView() and focus() scroll partially offscreen elements.

In https://crbug.com/916631 I removed a magic constant in Blink for elements partially offscreen horizontally - this was [inherited from](https://crrev.com/ee5bc3acd0986e8fa87df7f804c49c120b8c9bc2) WebKit and meant that an element horizontally onscreen by at least 32 pixels was considered "fully visible" for ScrollIntoView.

However, this led to bugs like https://crbug.com/1036817. I realized that for vertical-only list controls, there isn't a good way to make them clip horizontally (overflow-x: hidden) but navigable vertically; moving focus to a vertically offscreen element should scroll it to view but we shouldn't scroll horizontally.

This appears to have been the intent of the magic constant but I think for element.scrollIntoView makes it hard for authors to reason about hence the above change in Blink (to match Gecko; @emilio added a [WPT test](https://wpt.fyi/results/css/cssom-view/scrollIntoView-horizontal-partially-visible.html?label=experimental&label=master&aligned) for this too).

It seems like it'd make sense to apply this special non-scrolling behavior just for focus()? The [focus() spec](https://html.spec.whatwg.org/#dom-focus) gives UA some latitude in how the scrollIntoView is done but I wonder if we should try to align behavior between engines? Also, not scrolling technically isn't a valid option for ScrollIntoViewOptions AFAICT.

Interestingly, it appears Firefox does avoid scrolling a partial element onscreen from focus. I'm going to make the same change in Blink but I'm having a hard time exactly determining Gecko's behavior. It looks like there's some kind of threshold too (10px onscreen is scrolled but 50px is not) but testing locally on Linux in FF73 I see focus() scrolls both axes into view whereas (using browserstack) on Windows/Mac it only scrolls the vertical axis. @emilio - can you help me understand what's going on here?

Here's a test page that demonstrates the differences between scrollIntoView and focus with differing amounts of scrollport overlap: https://jsbin.com/sewewam

@smfr for WebKit and whether we can align behavior here.

emilio:
[...]

PR #5677, opened 2020-10-28
[cssom-view] Add notIfViewed to ScrollIntoViewOptions; add FocusScrollOptions

zcorpan:
This is the same as https://github.com/w3c/csswg-drafts/pull/1805 but from a fork to address  https://lists.w3.org/Archives/Member/w3c-css-wg/2020OctDec/0050.html

`Agenda+` for https://github.com/w3c/csswg-drafts/pull/1805#issuecomment-694781722 and following comments. @frivoal @anawhj @jihyerish @stipsan

---

> notIfViewed was originally suggested in
> http://www.w3.org/mid/CAE3TfZPmNWvz1z7XPqFjtg5S+m_9BOmNNHsOZuSrR2z2AgyNHA@mail.gmail.com

> Now it is needed for HTMLElement focus(options) to match the behavior
> of focus() in browsers. Since the defaults for focus() is different
> from the defaults for scrollIntoView(), a new dictionary
> FocusScrollOptions is introduced. For now the defaults for block
> and inline are UA-defined.
>
> https://github.com/whatwg/html/pull/2787#issuecomment-328397430

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1065,8 +1065,14 @@ Note: This {{DOMRect}} object is not <a spec=html>live</a>.
 <pre class=idl>
 enum ScrollLogicalPosition { "start", "center", "end", "nearest" };
 dictionary ScrollIntoViewOptions : ScrollOptions {
+  boolean notIfViewed = false;
   ScrollLogicalPosition block = "start";
   ScrollLogicalPosition inline = "nearest";
+};
+dictionary FocusScrollOptions : ScrollOptions {
+  boolean notIfViewed = true;
+  ScrollLogicalPosition block;
+  ScrollLogicalPosition inline;
 };
 
 partial interface Element {
@@ -1089,6 +1095,13 @@ partial interface Element {
   readonly attribute long clientHeight;
 };
 </pre>
+
+The {{ScrollIntoViewOptions}} dictionary is used by {{Element/scrollIntoView()}} and similar APIs
+that primarily scroll something into view.
+
+The {{FocusScrollOptions}} dictionary is similar to {{ScrollIntoViewOptions}}, but has different
+defaults, and is used by {{HTMLElement}}'s {{HTMLElement/focus()}} and similar APIs that primarily
+move focus to something, or equivalent, and secondarily scroll it into view.
 
 The <dfn method for=Element>getClientRects()</dfn> method, when invoked, must return the result of the following algorithm:
 
@@ -1130,14 +1143,16 @@ The <dfn method for=Element caniuse=scrollintoview>scrollIntoView(<var>arg</var>
 1. Let <var>behavior</var> be "<code>auto</code>".
 1. Let <var>block</var> be "<code>start</code>".
 1. Let <var>inline</var> be "<code>nearest</code>".
+1. Let <var>notIfViewed</var> be false.
 1. If <var>arg</var> is a {{ScrollIntoViewOptions}} dictionary, then:
     1. Set <var>behavior</var> to the {{ScrollOptions/behavior}} dictionary member of <var>options</var>.
     1. Set <var>block</var> to the {{ScrollIntoViewOptions/block}} dictionary member of <var>options</var>.
     1. Set <var>inline</var> to the {{ScrollIntoViewOptions/inline}} dictionary member of <var>options</var>.
+    1. Set <var>notIfViewed</var> to the {{ScrollIntoViewOptions/notIfViewed}} dictionary member of <var>options</var>.
 1. Otherwise, if <var>arg</var> is false, then set <var>block</var> to "<code>end</code>".
 1. If the element does not have any associated <a>layout box</a>, then return.
 1. <a lt='scroll an element into view'>Scroll the element into view</a>
-    with <var>behavior</var>, <var>block</var>, and <var>inline</var>.
+    with <var>behavior</var>, <var>block</var>, <var>inline</var>, and <var>notIfViewed</var>.
 1. Optionally perform some other action that brings the element to the user's attention.
 
 The <dfn method for=Element lt="scroll(options)|scroll(x, y)">scroll()</dfn> method must run these steps:
@@ -1290,8 +1305,9 @@ The <dfn attribute for=Element>clientHeight</dfn> attribute must run these steps
 
 To <dfn>scroll an element into view</dfn> <var>element</var>,
 with a scroll behavior <var>behavior</var>,
-a block flow direction position <var>block</var>,
-and an inline base direction position <var>inline</var>,
+a boolean indicating to not scroll if the element is already in view <var>notIfViewed</var>,
+optionally a block flow direction position <var>block</var> (undefined if not given),
+and optionally an inline base direction position <var>inline</var> (undefined if not given),
 means to run these steps for each ancestor element or <a>viewport</a> that establishes
 a <a>scrolling box</a> <var>scrolling box</var>, in order of innermost to outermost <a>scrolling box</a>:
 
@@ -1305,6 +1321,15 @@ a <a>scrolling box</a> <var>scrolling box</var>, in order of innermost to outerm
 1. Let <var>scrolling box height</var> be the distance between <var>scrolling box edge A</var> and <var>scrolling box edge B</var>.
 1. Let <var>element width</var> be the distance between <var>element edge C</var> and <var>element edge D</var>.
 1. Let <var>scrolling box width</var> be the distance between <var>scrolling box edge C</var> and <var>scrolling box edge D</var>.
+1. If <var>notIfViewed</var> is true, and <var>element</var> is entirely in view already, then return.
+
+        ISSUE: Define "entirely in view".
+
+1. If <var>block</var> is undefined, set <var>block</var> to a UA-defined value.
+1. If <var>inline</var> is undefined, set <var>inline</var> to a UA-defined value.
+
+        ISSUE: Define defaults for <var>block</var> and <var>inline</var> for {{FocusScrollOptions}}.
+
 1. Let <var>position</var> be the scroll position <var>scrolling box</var> would have by following these steps:
 
     1. If <var>block</var> is "<code>start</code>", then align <var>element edge A</var> with <var>scrolling box edge A</var>.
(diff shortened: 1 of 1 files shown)


zcorpan:
@anawhj asked (https://github.com/w3c/csswg-drafts/pull/1805#issuecomment-716278718)

> How could we drive this item? What should we discuss with more?

For `focus()` extensions, see https://github.com/whatwg/html/issues/834#issuecomment-339728104

The interop problem of the default scrolling behavior for `focus()` still exists, but could be solved I believe by changing Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1410112

For `scrollIntoView()` extensions, I think it would be good to review what parts of 
https://github.com/stipsan/scroll-into-view-if-needed developers are using and should be in the web platform directly, and whether the API shape in this PR is the right one.

stipsan:
>For `scrollIntoView()` extensions, I think it would be good to review what parts of
https://github.com/stipsan/scroll-into-view-if-needed developers are using and should be in the web platform directly, and whether the API shape in this PR is the right one.

Let me know if there's anything I can do 🙂   In my experience `scrollMode: "if-needed"` is often enough for most.
Then we have the common use case where devs are confused/surprised when they experience `overflow: hidden` frames getting scrolled, common enough to warrant [adding a workaround](https://github.com/stipsan/compute-scroll-into-view#skipoverflowhiddenelements).

The [`custom behavior`](https://github.com/stipsan/scroll-into-view-if-needed#function) API was added as an escape hatch for the other corner cases instead of bloating the API surface further. Since then a growing number of libraries prefer this pattern, as they want control over how the scrolling operations are performed: https://www.npmjs.com/browse/depended/compute-scroll-into-view
It could be helpful to have a similar utility in the platform, allowing us to calculate what should scroll where, but let us handle the actual scrolling 🤔

zcorpan:
@stipsan thanks!

For `overflow: hidden`, is there a reason `overflow: clip` when you want to disallow scrolling isn't a better solution?

The custom behavior API seems interesting!

stipsan:
> For `overflow: hidden`, is there a reason `overflow: clip` when you want to disallow scrolling isn't a better solution?

I should've been more clear about that one. I mentioned the `overflow: hidden` confusion to only highlight the fact, not to propose that the `skipOverflowHiddenElements` workaround should be in the `scrollIntoView` spec. The `overflow: clip` behavior is already supported by the polyfill since it's [in the spec](https://drafts.csswg.org/css-overflow-3/#valdef-overflow-clip). Once Safari and Chrome [support it](https://caniuse.com/mdn-css_properties_overflow_clip) we'll drop the workaround 😄

othermaciej:
The title mentioned "notIfViewed" but I don't see that term or anything like it anywhere in the PR.

zcorpan:
@othermaciej thanks, fixed the title. (The API shape changed since the original proposal in #1805.)

Apologies for missing the telecon twice. I'll try again next week.

anawhj:
> For focus() extensions, see whatwg/html#834 (comment)

In the thread(whatwg/html#834), there were requirements as follows:
* Disabling scrolling on focus()
-> Now, we can disable to scroll the element with `element.focus({preventScroll: true});`. (https://github.com/whatwg/html/pull/2787)
* Smooth scrolling on focus()
-> Now, we can smoothly scroll to the element by `element.focus()` with `scroll-behavior: smooth`.
* Alignment of scrolling on focus()
-> We can't do it for now. The default alignment is even different between Chrome(center) and Firefox(nearest).

If we have `FocusScrollOptions` for the last remaining requirement, I believe it could be used for several use cases.
(e.g. handle browser compatibility issue, avoid to use custom scroller, support several UX req.)

For the purpose, we could simply refer to [`scrollIntoviewOptions`](https://www.w3.org/TR/cssom-view-1/#dictdef-scrollintoviewoptions) with block and inline properties.

My proposal is:
Can we rename the dictionary from [`scrollIntoviewOptions`](https://www.w3.org/TR/cssom-view-1/#dictdef-scrollintoviewoptions) to `scrollAllignmentOptions' as more general term?
I referred Scroll Alignment term to https://scroll-into-view-if-needed.netlify.app/#scroll-alignment.
The renamed dictionary can be referred for the additional focus() extension in whatwg side.
It doesn't affect to web developers without any WebAPI change.

anawhj:
> the interop difference in alignment on focus() should be fixed by browsers directly

I agree with it, but it hasn't been able to be unified for a long time, so authors need to have an option for handling it as one behavior now. Hopefully, it can be completed by https://bugzilla.mozilla.org/show_bug.cgi?id=1410112.

> I wonder if it makes sense to have a CSS property for focus alignment, which could take effect both for focus() and when the user moves focus with the Tab key.

If we define a new CSS property for focus alignment, it needs to work as `scroll-behavior` does. The scroll-behavior takes effect both cases now. (focus() method and seq. navigation by tab key) Do we have further consideration for defining the new CSS property? If it needs much works to do, it could be simply handled by FocusScrollOptions as parameters inside focus() without a new CSS property.
e.g. `scroll-alignment-[x|y]: [auto | start | end | center]`

> It's a separate dictionary from ScrollIntoViewOptions because scrollIntoView() and focus() have different default behavior (always vs. if-needed, and likely also default alignment).

If we can define the different default behaviors as a unified one in CSSOM-VIEW spec, ScrollIntoViewOptions could be used for several usages incl. focus(). If it would be difficult, I'm okay to put FocusScrollOptions separately from ScrollIntoViewOptions.

zcorpan:
Next steps:

- [ ] Review Scroll Snap spec and see how well the scroll alignment use cases are handled already.
- [ ] If the scroll snap scroll alignment properties are sufficient, remove the scroll alignment parts from this PR. (Otherwise, explain and discuss again with the WG.)
- [ ] Specify "if-needed" in terms of Scoll Snap so that 'scroll-margin' and 'scroll-padding' are taken into account when determining whether an element is "in view".
- [ ] Specify default alignment for `focus()` to match WebKit/Chromium, and fix Gecko to match. (AFAIK, so far no objections, and it's the shortest path to interop, and the behavior doesn't seem unreasonable.)

stipsan:
> Specify "if-needed" in terms of Scoll Snap so that 'scroll-margin' and 'scroll-padding' are taken into account when determining whether an element is "in view".
Great news! I'll update the polyfill to follow [this](https://github.com/stipsan/scroll-into-view-if-needed/issues/386) spec when I have some time off soon 😄 .

jihyerish:
How does this option work with nested overflow or iframe scrolling? Is this option applied to all descendant elements by default?

ShivamJoker:
Any updates on this?

anawhj:
@ShivamJoker 
If you have any use case on this, could you briefly share it here? :)

ShivamJoker:
Yes sure.

I have made an accordion library which renders 100s of list but it get's to bottom and user expands it, it can't automatically scroll into view.

Demo screen recording - 

https://user-images.githubusercontent.com/23727670/155487582-a8518bd2-7f35-400e-b663-00e85b3a8dd8.mov

If I add scrollIntoView all the elements will move even if it is not needed, and `scrollIntoViewIfNeeded` doesn't have smooth scrolling option and is not supported in majority of the browsers.

It would be helpful in such cases.

zcorpan:
`scrollIntoViewIfNeeded` usage seems to be rising over time: https://chromestatus.com/metrics/feature/timeline/popularity/389

If `scrollIntoViewIfNeeded` needs to be part of the web platform for compat (currently it's not implemented in Gecko), another option would be to standardize it and add smooth scrolling and positioning capabilities to it, instead of adding "if needed" to `scrollIntoView`. Thoughts?

stipsan:
@zcorpan do you also suggest adding `scroll-behavior: smooth;` to it as well?

And what about `.focus()`? There's been earlier proposals for aligning `.focus()` with `.scrollIntoView`, giving it `.focus({scrollMode: "if-needed", block: "nearest", inline: "nearest", behavior: "smooth"})`. But if `scrollIntoViewIfNeeded` is standardised instead of `.scrollIntoView({scrollMode: "if-needed"})` does that mean we should create a `.focusIntoViewIfNeeded()`?

Also worth noting that the demand for the proposed spec is on a steady increase as well: https://npm-stat.com/charts.html?package=scroll-into-view-if-needed&from=2018-03-16&to=2023-03-16 🤔

frank-topel-dbi:
Why has this issue been closed? Afaik there still is no cross-browser `scrollIntoViewIfNeeded`. What is the suggested way to achieve this? Is it still using third-party code for a feature that has been requested more than 7 years ago?

zcorpan:
@stipsan thinking about this again, I think extending `scrollIntoView()` and `focus()` makes more sense. We could also standardize `scrollIntoViewIfNeeded()` for web compat but not add an options bag.

@frank-topel-dbi yes, use polyfill for now.

stipsan:
> @stipsan thinking about this again, I think extending `scrollIntoView()` and `focus()` makes more sense. We could also standardize `scrollIntoViewIfNeeded()` for web compat but not add an options bag.

Sounds good 👍

theres-waldo:
>    * [ ]  Specify default alignment for `focus()` to match WebKit/Chromium, and fix Gecko to match. (AFAIK, so far no objections, and it's the shortest path to interop, and the behavior doesn't seem unreasonable.)

I believe this has since been done:

 * Spec issue: https://github.com/whatwg/html/pull/9527
 * Web platform test: https://github.com/web-platform-tests/wpt/pull/41029
 * Gecko fix: https://bugzilla.mozilla.org/show_bug.cgi?id=1842679

zcorpan:
> It's unclear to me what `scroll alignment` means here. `scroll-snap-align` is applicable only for `scroll snap container`s. So if the target scroll container's scroll-snap-type is `none`, it doesn't affect at all.

Scroll alignment refers to the `block` and `inline` members in this PR.

I think `scrollIntoView` should keep these, so that alignment can be controlled in cases where there is no scroll snap container, and so that it can be set to something else even if there is a scroll snap container (similar to how you can set `smooth` to override CSS). Also "nearest" is not covered by scroll snap AFAICT.

When `block` or `inline` are undefined, it would make sense to use `scroll-snap-align` (if applicable) before using default values. Also see https://github.com/w3c/csswg-drafts/issues/4778 

> 
> Whereas `scroll-margin` and `scroll-padding` are applicable as the names stand for.

OK, they should probably be honored by `scrollIntoView()`.

PR #5688, opened 2020-10-30
[css-transforms-1] Fix small mistake (dots count)

afdw:


afdw:

Issue #5520, opened 2020-09-16
[css-transforms-1] Example 5 is wrong : order is reversed


mclegrand:
The order of things in example 5 of https://github.com/w3c/csswg-drafts/blob/master/css-transforms-1/Overview.bs#L284 is wrong and contradicts both section 10 and example 2 (which are right), as explained by the following commented file :

(in red : what *actually* happens "if the div element gets translated by 80px to the bottom left direction, then scaled up by 150% and finally rotated by 45°."; in blue what happens if you just right-add transformations (in particular, you can't get from the translated object to the scaled+translated object with a origin-centered scaling, and you can't get from the translated+scaled one to the final one with a origin-centered rotation )); in green the "natural" mental transformations : take the original object, rotate the result, translate the result
```xml
<svg xmlns="http://www.w3.org/2000/svg">

<rect x="40" y="40" width="70" height="70" opacity="0.5"/> <!-- original element -->
<rect x="40" y="40" width="70" height="70" opacity="0.6" fill="green" transform="translate(80,80) scale(1.5) rotate(45)"/> <!-- resulting element -->


<!-- not the correct visual path to the result -->
<rect x="120" y="120" width="70" height="70" opacity="0.5" fill="red"/><!-- translated-->
<rect x="180" y="180" width="105" height="105" opacity="0.5" fill="red"/><!-- translated then scaled -->
<rect x="180" y="180" width="105" height="105" opacity="0.5" transform="rotate(45)" fill="red"/> <!-- translated then scaled then rotated (WRONG RESULT) -->



<!-- not the correct visual path to the result (because of multiplication order) -->
<g transform="translate(80,80)">
  <rect x="40" y="40" width="70" height="70" opacity="0.5" fill="blue"/>
  <g transform="scale(1.5)">
          <rect x="40" y="40" width="70" height="70" opacity="0.5" fill="blue"/> <!-- different from the "translated then scaled" -->
    <g transform="rotate(45)">
            <rect x="40" y="40" width="70" height="70" opacity="0.5" fill="blue"/> <!-- resulting element -->
    </g>
  </g>
</g>

<!-- correct visual/mental path to the result : rotated first (goes to the left of the image because it started on the diagonal), then scale it (stays to the left, lower), then move it diagonally -->
<g transform="rotate(45)">
  <rect x="40" y="40" width="70" height="70" opacity="0.5" fill="green"/>
</g>

<g transform="scale(1.5) rotate(45)">
[...]

PR #5690, opened 2020-10-31
[css-transforms-1] Fix mistakes in Example 5 (closes #4767)

### LINKED ISSUE #4767: [css-transforms] Error in description of transform in Section 3 example 5
[https://www.w3.org/TR/css-transforms-1/](https://www.w3.org/TR/css-transforms-1/)

In example 5 of section 3, the first paragraph under the CSS snippet reads "The visual appareance is as if the div element gets translated by 80px to the bottom left direction, then scaled up by 150% and finally rotated by 45°."  In addition to the misspelling of "appearance", the <div> element gets translated by 80px to the bottom __right__ direction.  (The diagram underneath the text is correct.)

afdw:


afdw:

astearns:
@smfr this seems fine to me, but I’d like to get your eyes on it.

mclegrand:
Ah, I missed this MR when I finally decided to act on my report #5520 this morning, sorry. I think the illustration also needs an update

afdw:
@mclegrand my PR was designed so that it corresponds to illustration. See http://web.cse.ohio-state.edu/~wang.3602/courses/cse5542-2013-spring/6-Transformation_II.pdf, there are 2 different ways to think about transformations. So I think my version is actually fully coherent if we think about it as post-multiplication / transform with respect to the local origin and basis. But [the canvas specification](https://html.spec.whatwg.org/multipage/canvas.html#current-transformation-matrix) implies the other variant (pre-multiplication / transform with respect to the global origin and basis). So what do you think? Which approach will be better?

mclegrand:
Well, the important thing is that it's the same example as in Example 1 and Example 2 :) 

IMO, the whole spec (mostly section 3) revolves around the CTM (current transformation matrix) concept which (still IMO) works best when thought in terms of pre-multiplication ( f(x) = F×x multiplied by the left ). But even with post-mul, I don't think the illustration makes sense as it shows an object being modified and not a coordinate system ?

afdw:
> But even with post-mul, I don't think the illustration makes sense as it shows an object being modified and not a coordinate system ?

Yes, the axis are kind of implicit currently, and should probable be added.

> IMO, the whole spec (mostly section 3) revolves around the CTM (current transformation matrix) concept which (still IMO) works best when thought in terms of pre-multiplication ( f(x) = F×x multiplied by the left ).

My opinion is that ideally both approaches to thinking about it should be described. I think it would eliminate the confusion one might have here. As this is a specification about transformations, this should should not be too unrelated I think?

svgeesus:
kicking the ipr bot by closing and opening.

@smfr I see @astearns was asking for your review

PR #5702, opened 2020-11-06
 [css-transforms-1] Fix 5520

### LINKED ISSUE #5520: [css-transforms-1] Example 5 is wrong : order is reversed

mclegrand:
Fixes #5520


mclegrand:
(also fixes #4767 : bottom left → bottom right)

svgeesus:
There is a [four year old pull request](https://github.com/w3c/csswg-drafts/pull/5702) to fix this issue. Could one of the editors take a look??

svgeesus:
Closing and re-opening to kick the IPR bot into activity

dbaron:
So there's a lot going on here.

Part of what's difficult about this stuff is that it makes sense to think about most of these things in either order.  You can think of the math as matrices that operate on row-vectors or matrices that operate on column-vectors.  The spec is pretty consistent about using column vectors.

Likewise, you can visually think about the transforms as transforming coordinates starting from the root coordinate space and transforming your way to the leaf, or starting from the leaf element and transforming it to the root.

Every time I read anything about transforms, I have to start from first principles to figure out which way I'm looking at it.  (I'm an editor of the transforms spec and I don't remember what the order of the transforms functions means; I just figure it out again every time I need to know.)  Maybe that's a bad sign, though.

I haven't looked at the text diffs much yet.  However, I took a look at the image diffs, and I think they're substantively incorrect, although I haven't yet thought about which way makes more sense.  In particular, the image currently in the spec geometrically matches [this testcase](https://dbaron.org/css/test/2025/pr5702-1), whereas the revised image is a bit like but doesn't actually match [this revised testcase](https://dbaron.org/css/test/2025/pr5702-2).  So I think even if we want to flip the logic of the image around, the new image isn't quite right.

I'll try to dig in more later and figure out whether I agree with the textual changes proposed here.

PR #5711, opened 2020-11-10
[css-shapes-2] Add a shape() syntax, which accepts shape-segments equivalent to SVG path

### LINKED ISSUE #5674: [css-shapes] Allow CSS grammar for path shapes
Currently, when using shapes in e.g. `clip-path`, some features are only allowed in polygon, and some only in path.

- Polygon: CSS custom properties as coordinates, relative or other non-pixel units
- Path: Rounded corners with quadratic/bezier curves

So it's currently impossible to create a non-rect polygon that has both rounded corners and relative (or em) coordinates, or custom properties, for example a speech bubble with an arrow.

I suggest one of the following:
- Allow some form of radius in `polygon()` (easier)
- Allow resolving of non-pixel units in `path()` (more powerful)

noamr:
Add a draw() shape syntax, which accepts shape-segments equivalent to SVG path segments.

Fixes #5674

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -170,7 +170,7 @@ Supported Shapes</h3>
 			</ul>
 		<dt>
 			<pre class=prod>
-				<dfn>shape()</dfn> = shape( [<<fill-rule>>,]? from <<coordinate-pair>>, <<draw-command>>#)
+				<dfn>shape()</dfn> = shape( [<<fill-rule>>]? from <<coordinate-pair>>, <<draw-command>>#)
 			</pre>
 		<dd dfn-type=value dfn-for="shape()">
 			<ul>
@@ -221,7 +221,7 @@ Supported Shapes</h3>
 		<dd>
 			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataCubicBezierCommands">smooth cubic curve</a> command if a second <<coordinate-pair>> is provided,
 			otherwise corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataQuadraticBezierCommands">smooth quadratic curve</a> command.
-		<dt><dfn><<arc-command>></dfn> = arc <<by-to>> <<coordinate-pair>> <<arc-radius>> [<<arc-sweep>> || <<arc-large>> || <<angle>>]
+		<dt><dfn><<arc-command>></dfn> = arc [<<by-to>> <<coordinate-pair>> <<arc-radius>> || <<arc-sweep>> || <<arc-large>> || <<angle>>]
 		<dd>
 			Corresponds to an <a href="https://www.w3.org/TR/SVG/paths.html#PathDataEllipticalArcCommands">arc</a> command.
 		<dt><dfn><<arc-radius>></dfn> = of <<length-percentage>>{1, 2}
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -221,7 +221,7 @@ Supported Shapes</h3>
 		<dd>
 			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataCubicBezierCommands">smooth cubic curve</a> command if a second <<coordinate-pair>> is provided,
 			otherwise corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataQuadraticBezierCommands">smooth quadratic curve</a> command.
-		<dt><dfn><<arc-command>></dfn> = arc <<by-to>> <<coordinate-pair>> <<arc-radius>> <<arc-sweep>>? <<arc-large>>? <<angle>>?
+		<dt><dfn><<arc-command>></dfn> = arc <<by-to>> <<coordinate-pair>> <<arc-radius>> [<<arc-sweep>> || <<arc-large>> || <<angle>>]
 		<dd>
 			Corresponds to an <a href="https://www.w3.org/TR/SVG/paths.html#PathDataEllipticalArcCommands">arc</a> command.
 		<dt><dfn><<arc-radius>></dfn> = of <<length-percentage>>{1, 2}
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -185,7 +185,7 @@ Supported Shapes</h3>
 	<dl>
 		<dt><dfn><<coordinate-pair>></dfn> = <<length-percentage>>{2}
 		<dd>Defines a pair of coordinates x & y.
-		<dt><dfn><<draw-command>></dfn> = <<move-command>> | <<line-command>> |
+		<dt><dfn><<draw-command>></dfn> = <<move-command>> | <<line-command>> | <<hv-line-command>> |
 			<<curve-command>> | <<smooth-command>> | <<arc-command>> | close
 		<dd>
 			Defines a single draw command, equivalent to an <a href="https://www.w3.org/TR/SVG/paths.html#PathDataGeneralInformation">SVG draw command</a>.
@@ -202,10 +202,13 @@ Supported Shapes</h3>
 		<dt><dfn><<move-command>></dfn> = move <<by-to>> <<coordinate-pair>>
 		<dd>
 			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataMovetoCommands">moveto</a> command.
-		<dt><dfn><<line-command>></dfn> = line <<by-to>> [<<length-percentage>> | x | y] <<length-percentage>>
+		<dt><dfn><<line-command>></dfn> = line <<by-to>> <<coordinate-pair>>
 		<dd>
 			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands">lineto</a> command.
 			If '''x''' or '''y''' are present instead of a <<length-percentage>>, the line will be horizontal or vertical, respectively.
+		<dt><dfn><<line-command>></dfn> = [hline | vline] <<by-to>> <<length-percentage>>
+		<dd>
+			Corresponds to a horizontal or vertical <a href="https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands">lineto</a> command.
 		<dt><dfn><<curve-command>></dfn> = curve <<by-to>> <<coordinate-pair>> via <<coordinate-pair>>{1,2}
 		<dd>
 			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataQuadraticBezierCommands">quadratic curve</a> command if one <<coordinate-pair>> is provided,
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -170,14 +170,13 @@ Supported Shapes</h3>
 			</ul>
 		<dt>
 			<pre class=prod>
-				<dfn>shape()</dfn> = shape( [<<fill-rule>>,]? [<<corner>>, ]? <<draw-command>>#)
+				<dfn>shape()</dfn> = shape( [<<fill-rule>>,]? from <<coordinate-pair>>, <<draw-command>>#)
 			</pre>
 		<dd dfn-type=value dfn-for="shape()">
 			<ul>
 				<li>
-					The <<corner>> represents the starting point for the first draw-command, as well as the coordinate system
-					to use for all the draw commands.
-					Default value when omittted is '''left top'''.
+					The <<coordinate-pair>> represents the starting point for the first draw-command,
+					in physical coordinates.
 				<li>
 					The sequence of <dfn><<draw-command>></dfn>s represent commands of an
 					<a href="https://www.w3.org/TR/SVG11/paths.html#PathData">SVG Path</a>.
@@ -189,10 +188,6 @@ Supported Shapes</h3>
 	<dl>
 		<dt><dfn><<coordinate-pair>></dfn> = <<length-percentage>>{2}
 		<dd>Defines a pair of coordinates x & y.
-		<dt><dfn><<shape-origin>></dfn> = [ [left | right] && [top | bottom] ] | [ [inline-start | inline-end] && [block-start | block-end ] ]
-		<dd>
-			Defines the origin point (0, 0) of the shape's coordinate system. All coordinates in the shape will be computed relatively
-			to the specified point, based on the <a>writing mode</a> of the element's <a>containing block</a>.
 
 		<dt><dfn><<draw-command>></dfn> = <<move-command>> | <<line-command>> | <<hv-line-command>> |
 			<<curve-command>> | <<smooth-command>> | <<arc-command>> | close
@@ -218,7 +213,7 @@ Supported Shapes</h3>
 		<dt><dfn><<hv-line-command>></dfn> = [hline | vline] <<by-to>> <<length-percentage>>
 		<dd>
 			Corresponds to a horizontal or vertical <a href="https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands">lineto</a> command.
-	n		<dt><dfn><<curve-command>></dfn> = curve <<by-to>> <<coordinate-pair>> via <<coordinate-pair>>{1,2}
+			<dt><dfn><<curve-command>></dfn> = curve <<by-to>> <<coordinate-pair>> via <<coordinate-pair>>{1,2}
 		<dd>
 			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataQuadraticBezierCommands">quadratic curve</a> command if one <<coordinate-pair>> is provided,
 			otherwise corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataCubicBezierCommands">cubic curve</a> command.
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -142,19 +142,6 @@ Supported Shapes</h3>
 	<dl>
 		<dt>
 			<pre class=prod>
-				<dfn>draw()</dfn> = draw( [<<fill-rule>>,]? from <<coordinate-pair>>, <<draw-command>>#])
-			</pre>
-		</dt>
-		<dd dfn-type=value dfn-for="draw()">
-			<ul>
-				<li>
-					The <<coordinate-pair>> represents the initial point of the path.
-				<li>
-					The sequence of <dfn><<draw-command>></dfn>s represent commands of an
-					<a href="https://www.w3.org/TR/SVG11/paths.html#PathData">SVG Path</a>.
-			</ul>
-		<dt>
-			<pre class=prod>
 				<dfn>path()</dfn> = path( [<<fill-rule>>,]? <<string>> )
 			</pre>
 		<dd dfn-type=value dfn-for="path()">
@@ -179,6 +166,18 @@ Supported Shapes</h3>
 					in the path string.
 					For the initial direction follow SVG 1.1.
 			</ul>
+		<dt>
+			<pre class=prod>
+				<dfn>draw()</dfn> = draw( [<<fill-rule>>,]? from <<coordinate-pair>>, <<draw-command>>#)
+			</pre>
+		<dd dfn-type=value dfn-for="draw()">
+			<ul>
+				<li>
+					The <<coordinate-pair>> represents the initial point of the path.
+				<li>
+					The sequence of <dfn><<draw-command>></dfn>s represent commands of an
+					<a href="https://www.w3.org/TR/SVG11/paths.html#PathData">SVG Path</a>.
+			</ul>
 	</dl>
 	
 	The arguments not defined above are defined as follows:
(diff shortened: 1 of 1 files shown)


astearns:
Thanks for writing up the proposal! It might be better to put this into css-shapes-2, as everything in css-shapes-1 is a bit further along (with an implementation and tests), and we should likely get that finished in the process and push new work to the next level.

@tabatkins can you take a look, since you have been responding in the issue?

noamr:
> Thanks for writing up the proposal! It might be better to put this into css-shapes-2, as everything in css-shapes-1 is a bit further along (with an implementation and tests), and we should likely get that finished in the process and push new work to the next level.

Of course, will move to css-shapes-2.

noamr:
@tabatkins this is an initial go at this, I'm sure it will require further examples and so, but wanted to align on the syntax.

tabatkins:
Looks nice! I like the by/to distinction, seems like a reasonable way to handle the rel/abs distinction.

The only thing I don't like is the `horizontal`/`vertical` keywords, because they're so long; it's shorter to just type `line 0 X` instead of `vertical X`.  I think we can get away with just shortening those to `h` and `v`.

noamr:
> Looks nice! I like the by/to distinction, seems like a reasonable way to handle the rel/abs distinction.
> 
> The only thing I don't like is the `horizontal`/`vertical` keywords, because they're so long; it's shorter to just type `line 0 X` instead of `vertical X`. I think we can get away with just shortening those to `h` and `v`.

The thing is that with absolute coordinates, `line to X 0` is not the same as `horizontal to X`. 
I was thinking as an alternative to use the `auto` keyword (`line to X auto` and `line to auto Y`). WDYT?

Something about the one-character EBNF feels less readable to me, but maybe just for h/v it's OK.

tabatkins:
Yeah, the one-character bit bothered me too, since the other commands aren't doing it (and dont' need to).

What about `hline` and `vline`, then?

noamr:
> Yeah, the one-character bit bothered me too, since the other commands aren't doing it (and dont' need to).
> 
> What about `hline` and `vline`, then?

LGTM!

Crissov:
I recently amended my last comment to mention alternative keywords: horizontal `level` and vertical `plummet`. 

| `l` | `L` | `h` | `H` | `v` | `V` |
|---|---|---|---|---|---|
| `line by <length> <length>` | `line to <length> <length>` | `horizontal by <length>` | `horizontal to <length>` | `vertical by <length>` | `vertical to <length>` |
| `line by <length> <length>` | `line to <length> <length>` | `level by <length>` | `level to <length>` | `plummet by <length>` | `plummet to <length>` |
| `line by <length> <length>` | `line to <length> <length>` | `line by <length> 0` | `line to <length> auto` | `line by 0 <length>` | `line to auto <length>` |
| `line by <length> <length>` | `line to <length> <length>` | `[left\|right] by <length>` | `[left\|right] to <length>` | `[top\|bottom] by <length>` | `[top\|bottom] to <length>` |
| `line by <length> <length>` | `line to <length> <length>` | `[left\|right] by <length>` | `[left\|right] to <length>` | `[up\|down] by <length>` | `[up\|down] to <length>` |

noamr:
> I recently amended my last comment in #5674 to mention alternative keywords: horizontal `level` and vertical `plummet`.
Oh, I missed that! Not sure I like `level` and `plummet`. But `up/down/left/right` sounds appealing.

noamr:
A couple of things I could use feedback about:

- WDYT of my suggestion to use `draw()` as the function keyword? It corresponds with SVG's `d` and is more specific than `shape()` and differentiates from `path()`.
- Not sure about including an `arc` syntax at all. It's been a long outstanding issue in SVG to try to find a better syntax for it, I feel that maybe we're making that clunky syntax last longer than it should. Perhaps bezier curves are sufficient for the first version (bezier curves and arcs are interchangeable);
- Thinking that we can do a better job with animations than how `path()` is animated: enable animating between lines/curves/arcs by converting all of them to bezier curves.

noamr:
> > I recently amended my last comment in #5674 to mention alternative keywords: horizontal `level` and vertical `plummet`.
> 
> Oh, I missed that! Not sure I like `level` and `plummet`. But `up/down/left/right` sounds appealing.

How about the algebra terms `rise` and `run`? I think they would be the most accurate but `hline` and `vline` still seem like the most understandable terms.

noamr:
New revision uses `hline` / `vline`. I think they're the most straightforward.

As for `arc`, since arcs are complex-ish and there's lots of tutorials out there around the existing SVG syntax, I kept it consistent for now.

ByteEater-pl:
Additional advantages of `down` and `left` are that they obviate the need to multiply by -1.

noamr:
> Additional advantages of `down` and `left` are that they obviate the need to multiply by -1.

But they would be confusing in absolute coordinates... e.g. `left to 100%` would actually be going right.

noamr:
New PR based on @smfr's comments and others:
- Using `shape` rather than `draw`
- Allowing logical properties by defining a potentially-logical origin corner rather than an arbitrary starting point.

noamr:
@tabatkins I think it's ready for a review round.

tabatkins (approved):
Looks good. Lots of small editorial nits, but it's not worth going over in this review; I need to do an editorial pass over this entire spec.

tabatkins:
Wait, spoke too soon, didn't notice the "corner" design. We should discuss that separately and make sure we're happy with the design there; I suggest dropping it from the initial proposal and just defaulting to the "standard" physical coordinate system that paths already use.

tabatkins:
Anyway, Agenda+ to discuss this addition to the spec.

noamr:
> Wait, spoke too soon, didn't notice the "corner" design. We should discuss that separately and make sure we're happy with the design there; I suggest dropping it from the initial proposal and just defaulting to the "standard" physical coordinate system that paths already use.

OK, I'll switch to physical - and will propose a separate discussion about logical coordinates in shapes.
Thanks!

faceless2:
Won't make the call to discuss, but this LGTM too. I think a straight drop-in replacement for the existing SVG syntax is a good idea. For that reason I also agree with keeping logical coordinates well away is a good idea, at least or now. I'm also not sure about "corner" - given you can replace `corner top right` with `move to 100% 0`, I don't think it adds anything. Maybe it was more useful when logical coordinates were an option; I'd consider dropping it until they're readded.

Otherwise, great! It's definitely better than string concatenation :-)

(oh, a quick addendum re. one of your earlier comments - arcs are not directly replaceable with beziers, as they are treated differently when used with SVG markers: a 270° arc with end markers has just the two, but drawing the same curve with beziers would result in more. So the concept needs to remain in the grammar. Yes the syntax is awkward, but it's well established, and using `small|large` and `cw|ccw` will certainly help)

noamr:
> Won't make the call to discuss, but this LGTM too. I think a straight drop-in replacement for the existing SVG syntax is a good idea. For that reason I also agree with keeping logical coordinates well away is a good idea, at least or now. I'm also not sure about "corner" - given you can replace `corner top right` with `move to 100% 0`, I don't think it adds anything. Maybe it was more useful when logical coordinates were an option; I'd consider dropping it until they're readded.
> 

Yes, the corner thing makes sense only in the context of logical coordinates. Will wait with that.
As a Hebrew speaker I'd love to help out figure how shapes could become logical-property friendly (but no in this PR).

> (oh, a quick addendum re. one of your earlier comments - arcs are not directly replaceable with beziers, as they are treated differently when used with SVG markers: a 270° arc with end markers has just the two, but drawing the same curve with beziers would result in more. So the concept needs to remain in the grammar. Yes the syntax is awkward, but it's well established, and using `small|large` and `cw|ccw` will certainly help)

Good point, I didn't knot at! Though SVG markers don't play a role in CSS shapes. Do they? In which case, arcs and curves are interchangeable.

faceless2:
> Good point, I didn't knot at! Though SVG markers don't play a role in CSS shapes. Do they? In which case, arcs and curves are interchangeable.

True, currently. But (from https://github.com/w3c/svgwg/issues/119 and https://github.com/w3c/svgwg/issues/320) the "d" property uses the path() function (see also PR https://github.com/w3c/svgwg/pull/374). Although it doesn't yet reference css-shapes, it seems to me that where `path()` applies, `shape()` would always apply too. And this looks pretty good to me:

```svg
<svg>
 <style>
   path {
     d: shape(move to 0 0 arc to 100% 100% cw small 90deg);
     marker-start: url(#marker);
     marker-end: url(#marker);
   }
 </style>
 <marker id="marker" markerWidth="10" markerHeight="10" refX="5" refY="5">
  <circle cx="5" cy="5" r="5"/>
 </marker>
 <path></path>
</svg>
```


Actually, that's another point - ideally the grammar for "arc" would allow the `radius`, `arc-sweep` `arc-large` and `angle` to be reordered, as this can be done without ambiguity. I _think_ that means it should look like `arc <by-to> <coordinate-pair> [<radius> || <arc-sweep> || <arc-large> || <angle>]`, but don't quote me on that. No doubt these are some of the editorial nits Tab will be picking.

noamr:
> > Good point, I didn't knot at! Though SVG markers don't play a role in CSS shapes. Do they? In which case, arcs and curves are interchangeable.
> 
> True, currently. But (from [w3c/svgwg#119](https://github.com/w3c/svgwg/issues/119) and [w3c/svgwg#320](https://github.com/w3c/svgwg/issues/320)) the "d" property uses the path() function (see also PR [w3c/svgwg#374](https://github.com/w3c/svgwg/pull/374)). Although it doesn't yet reference css-shapes, it seems to me that where `path()` applies, `shape()` would always apply too. And this looks pretty good to me:
> 
> ```
> <svg>
>  <style>
>    path {
>      d: shape(move to 0 0 arc to 100% 100% cw small 90deg);
>      marker-start: url(#marker);
>      marker-end: url(#marker);
>    }
>  </style>
>  <marker id="marker" markerWidth="10" markerHeight="10" refX="5" refY="5">
>   <circle cx="5" cy="5" r="5"/>
>  </marker>
>  <path></path>
> </svg>
> ```
> 
> Actually, that's another point - ideally the grammar for "arc" would allow the `radius`, `arc-sweep` `arc-large` and `angle` to be reordered, as this can be done without ambiguity. I _think_ that means it should look like `arc <by-to> <coordinate-pair> [<radius> || <arc-sweep> || <arc-large> || <angle>]`, but don't quote me on that. No doubt these are some of the editorial nits Tab will be picking.



> > Good point, I didn't knot at! Though SVG markers don't play a role in CSS shapes. Do they? In which case, arcs and curves are interchangeable.
> 
> True, currently. But (from [w3c/svgwg#119](https://github.com/w3c/svgwg/issues/119) and [w3c/svgwg#320](https://github.com/w3c/svgwg/issues/320)) the "d" property uses the path() function (see also PR [w3c/svgwg#374](https://github.com/w3c/svgwg/pull/374)). Although it doesn't yet reference css-shapes, it seems to me that where `path()` applies, `shape()` would always apply too. And this looks pretty good to me:
> 
> ```
> <svg>
>  <style>
>    path {
>      d: shape(move to 0 0 arc to 100% 100% cw small 90deg);
>      marker-start: url(#marker);
>      marker-end: url(#marker);
>    }
>  </style>
>  <marker id="marker" markerWidth="10" markerHeight="10" refX="5" refY="5">
>   <circle cx="5" cy="5" r="5"/>
>  </marker>
>  <path></path>
> </svg>
> ```
> 
> Actually, that's another point - ideally the grammar for "arc" would allow the `radius`, `arc-sweep` `arc-large` and `angle` to be reordered, as this can be done without ambiguity. I _think_ that means it should look like `arc <by-to> <coordinate-pair> [<radius> || <arc-sweep> || <arc-large> || <angle>]`, but don't quote me on that. No doubt these are some of the editorial nits Tab will be picking.

Yes I think making arc parameters order-agnostic would be better. Let me revise.

noamr:
This is my first contribution to CSSWG (woohoo!), so I applied as an invited expert.

> Anyway, Agenda+ to discuss this addition to the spec.

Should I wait till it's discussed in the agenda or merge it in?

astearns:
> > Anyway, Agenda+ to discuss this addition to the spec.
> 
> Should I wait till it's discussed in the agenda or merge it in?

Let’s wait for the discussion. There’s also the IPR commitment that needs to be made in some form or other. Are you talking to @svgeesus about that?

svgeesus:
> There’s also the IPR commitment that needs to be made in some form or other. Are you talking to @svgeesus about that?

I asked for the IPR commitment, and it has been made. I don't understand why the flag hasn't cleared. Ah! I had to press the "revalidate" button. All good now.

css-shapes-2/Overview.bs
```diff
@@ -165,6 +168,79 @@ Supported Shapes</h3>
 					in the path string.
 					For the initial direction follow SVG 1.1.
 			</ul>
+		<dt>
+			<pre class=prod>
+				<dfn>shape()</dfn> = shape( [<<fill-rule>>,]? from <<coordinate-pair>>, <<draw-command>>#)
```
fantasai, line 173: I'd drop the comma after fill-rule, it's cleaner without.

css-shapes-2/Overview.bs
```diff
@@ -165,6 +168,79 @@ Supported Shapes</h3>
...
+					The sequence of <dfn><<draw-command>></dfn>s represent commands of an
+					<a href="https://www.w3.org/TR/SVG11/paths.html#PathData">SVG Path</a>.
+			</ul>
+	</dl>
+	
+	The arguments not defined above are defined as follows:
+
+	<dl>
+		<dt><dfn><<coordinate-pair>></dfn> = <<length-percentage>>{2}
+		<dd>Defines a pair of coordinates x & y.
```
fantasai, line 190: Should this be a `<<position>>`?

css-shapes-2/Overview.bs
noamr: Fine with me!

css-shapes-2/Overview.bs
```diff
@@ -165,6 +168,79 @@ Supported Shapes</h3>
...
+			Corresponds to a horizontal or vertical <a href="https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands">lineto</a> command.
+			<dt><dfn><<curve-command>></dfn> = curve <<by-to>> <<coordinate-pair>> via <<coordinate-pair>>{1,2}
+		<dd>
+			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataQuadraticBezierCommands">quadratic curve</a> command if one <<coordinate-pair>> is provided,
+			otherwise corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataCubicBezierCommands">cubic curve</a> command.
+		<dt><dfn><<smooth-command>></dfn> = smooth <<by-to>> <<coordinate-pair>> [via <<coordinate-pair>>]?
+		<dd>
+			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataCubicBezierCommands">smooth cubic curve</a> command if a second <<coordinate-pair>> is provided,
+			otherwise corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataQuadraticBezierCommands">smooth quadratic curve</a> command.
+		<dt><dfn><<arc-command>></dfn> = arc <<by-to>> <<coordinate-pair>> <<arc-radius>> [<<arc-sweep>> || <<arc-large>> || <<angle>>]
```
fantasai, line 224: Suggest allowing a bit more reordering, specifically:
`arc [<<by-to>> <<coordinate-pair>> <<arc-radius>> || <<arc-sweep>> || <<arc-large>> || <<angle>>]`

css-shapes-2/Overview.bs
noamr: Sure

css-shapes-2/Overview.bs
noamr: OK

fantasai:
Overall looks great!!!

fantasai:
You need to add yourself to the acknowledgements, though. =)

noamr:
> Overall looks great!!!

Thanks!
> You need to also add yourself to the acknowledgements, though. =)

How do I do that?

fantasai:
Good point. There isn't one yet. Need to add one, and add you to it. :) I suppose we can make that a separate issue.

css-shapes-2/Overview.bs
noamr: Actually, the problem is that <<position>> contains `left`/`top` etc, which don't make sense for relative coordinates. WDYT?

tabatkins: Hmm, to handle this properly we'd need to split out the abs/rel at a higher level, rather than just via the `by`/`to` keyword, so we could use the right non-terminal in each branch. Should be a fairly light bit of editting, just would make the spec a bit more verbose.

noamr: Got it, I'll revise.

fantasai (approved):

css-shapes-2/Overview.bs
fantasai: Ah, agreed. Definitely shouldn't have `<<position>>` everywhere.

@tabatkins Not sure what you mean, do you mean like the `<<corner>>` proposal earlier? I think that made a lot of sense, actually, switching the entire path into a different coordinate system so both absolute and relative coordinates switch over. Though doing that in a separate issue/PR also makes sense.

noamr: New revision tackles this.

noamr: @tabatkins @fantasai I'm not sure which one I prefer, using `<<position>>` or limiting to coordinate pairs.

On one hand, `<<position>>` allows us to use ``left``, ``center`` etc. OTOH, when we add logical-property support to shapes, e.g. as the previous `<<corner>>` proposal, they would make less sense. Also `polygon()` uses `<<length-percentage>>` and not `<<position>>`, so maybe for the sake of consistency it's best to leave shapes with `<<coordinate-pair>>` for the time being. WDYT?

fantasai: Yes, agree to leave as <<coordinate-pair>> here. We can always expand it out later if we decide it's a good idea, but I think it's good to evaluate that separately (and in consideration of what we want to do for logical coords).

noamr: Ok. Reverted `<<position>>` change. I think the patch is ready for being on the agenda!

grorg (commented):
Typo Corresponds.

css-shapes-2/Overview.bs
```diff
@@ -165,6 +168,79 @@ Supported Shapes</h3>
...
+			If ''cw'' is set, the arc's <a href="https://www.w3.org/TR/SVG/paths.html#PathDataEllipticalArcCommands">sweep</a> flag is set to 1.
+			If it is not set, the automatic value is used.
+		<dt><dfn><<arc-large>></dfn> = [large | small]?
+		<dd>
+			If ''large'' is set, the arc's <a href="https://www.w3.org/TR/SVG/paths.html#PathDataEllipticalArcCommands">large</a> flag is set to 1.
+			If ''small'' is set, the arc's <a href="https://www.w3.org/TR/SVG/paths.html#PathDataEllipticalArcCommands">large</a> flag is set to 0.
+			If it is not set, the automatic value is used.
+		<dt>close
+		<dd>
+			Corrsepponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataClosePathCommand">closepath</a> command.
```
grorg, line 243: Typo. Corresponds.

css-shapes-2/Overview.bs
```diff
@@ -165,6 +168,79 @@ Supported Shapes</h3>
...
+		<dd>
+			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataMovetoCommands">moveto</a> command.
+		<dt><dfn><<line-command>></dfn> = line <<by-to>> <<coordinate-pair>>
+		<dd>
+			Corresponds to a <a href="https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands">lineto</a> command.
+			If '''x''' or '''y''' are present instead of a <<length-percentage>>, the line will be horizontal or vertical, respectively.
+		<dt><dfn><<hv-line-command>></dfn> = [hline | vline] <<by-to>> <<length-percentage>>
+		<dd>
+			Corresponds to a horizontal or vertical <a href="https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands">lineto</a> command.
+			<dt><dfn><<curve-command>></dfn> = curve <<by-to>> <<coordinate-pair>> via <<coordinate-pair>>{1,2}
```
grorg, line 216: would prefer `using` over `via`. The curve does not go via the control points.

grorg: Also, I wonder if the end point should come after the control points.

grorg: Lastly (hopefully), additional spec text should point out that an animation of the `shape` would work the same as SVG, but that `curve` is a bit different since the type of curve is defined by the number of parameters (rather than an identifier like SVG).

grorg (changes requested):

grorg:
Ha. @tabatkins merged before I pressed the "Submit Review" button.

tabatkins:
No worries, I'm gonna do a comprehensive edit pass over the spec anyway, so I'll make sure your comments are addressed.

PR #5719, opened 2020-11-12
[css-color-4] Add missing / to close span for ButtonBorder

dlibby-:
[css-color-4] Add missing / to close span for ButtonBorder so that the swatch doesn't include the description text.


svgeesus (approved):
Thanks!

PR #5736, opened 2020-11-19
[css-text-3] Fix spec status

xfq:
Not sure why it's a CG-DRAFT...

/cc @fantasai


svgeesus (approved):
Obviously correct fix

PR #5739, opened 2020-11-21
[mediaqueries-5] Fix grammar typo in advisment for prefers-contrast:forced 

fuzzbomb:
Non-substantive edit.

In an advisement section on `prefers-contrast`, concerning "forced"...

> This does not necessarily indicates a preference for more contrast.

"indicates" should be "indicate"


frivoal:
Thanks!

PR #5751, opened 2020-11-30
[scroll-animations-1] Specify liveness of element references

### LINKED ISSUE #5289: [scroll-animations] Elements returned from selector() should be live
It's possible for the `source` of a `@scroll-timeline` to change its meaning between frames. E.g. for `source:selector(#foo)`, which element is identified by `#foo` can change if attributes are moved around or elements are added/removed. When this happens, the subsequent frame / updating-of-the-scroll-timeline-state should then reflect whatever the new DOM state is.

This applies to element-based offsets as well.

cc @majido

andruud:
Resolves #5289.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -955,6 +955,13 @@ condition from the following:
 
 </div>
 
+Elements referenced by the {{source}} descriptor are "live":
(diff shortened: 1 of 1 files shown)


flackr:
Closing as obsolete.

PR #5755, opened 2020-11-30
[scroll-animations-1] Make @scroll-timeline forwards-compatible

### LINKED ISSUE #5109: [scroll-animations] Ignore unknown descriptors
We should allow (but ignore if unknown) _any_ descriptor, otherwise it's a problem for forwards compat.

> The &lt;declaration-list> inside of @scroll-timeline rule can only contain the descriptors defined in this section.

Could be changed to (borrowing from css-properties-values-api): "Unknown descriptors are invalid and ignored, but do not invalidate the @scroll-timeline rule".

andruud:
Copied from css-counter-styles.

Resolves #5109.


fantasai:
Closing as no longer relevant. since we redesigned the feature.

PR #5757, opened 2020-12-02
[css-font-loading] Remove step 3 from FontFaceSet.check()

### LINKED ISSUE #5744: [css-font-loading] FontFaceSet.check() method: reality vs spec vs privacy
The description in the spec of the [FontFaceSet.check() method](https://drafts.csswg.org/css-font-loading/#font-face-set-check) does not match reality as implemented in any of the major browsers I've tried; moreover, the behavior called for by the spec is bad for privacy as it trivially exposes the set of installed fonts.

Consider the behavior of `document.fonts.check("12px foobar")`, assuming no font family named "foobar" is present.

According to step 3 of the algorithm in the spec,

>  If _found faces_ is false, throw an XXX error and abort this algorithm.

this should throw an (unspecified?) error.

No browser I have tested does this. It seems that:

* Gecko (Firefox) returns `true`, which makes sense in terms of the general description of what `check()` does:

> If the given text/font combo will render without attempting to use any unloaded or currently-loading fonts, this method will return true

but does not take account of step 3 in the algorithm, which is explicitly called out as the second "special case" in the note.

* Webkit (Safari) also returns `true`.

* Blink (Chrome, Brave) returns `false`, which I believe must be considered a bug: the given font/text will not "use any unloaded or currently-loading fonts", so this is both misleading to the author (it implies the given font is known but currently unloaded or loading) _and_ non-spec-compliant.

The fact that browsers are supposed (per spec) to _throw_ if `check()` is called with a non-existent font name was noted 5 years ago in bug reports against both (Gecko)[https://bugzilla.mozilla.org/show_bug.cgi?id=1252821#c7] and (Blink)[https://bugs.chromium.org/p/chromium/issues/detail?id=591602#c7], but the situation remains unchanged: no browser does so.

Given that:

* Nobody has apparently implemented step 3 of the specified algorithm (so the web cannot be depending on it); and

* If implemented, it would offer authors a trivial way to probe the set of installed system fonts, which represents an attractive fingerprinting vector;

I propose we should simply remove this step (and the corresponding note about the special-case behavior) from the spec.

(I am aware that font fingerprinting can also be achieved through `@font-face`, as mentioned [in the spec](https://drafts.csswg.org/css-font-loading/#priv-sec), but `FontFaceSet.check()` with the throw-on-unknown-font behavior would make it particularly easy and efficient for authors.)

jfkthame:
This fixes #5744 by removing the exception-throwing step in the algorithm (that nobody has implemented), and adjusting the related note to reflect this.

### DIFF
--- a/css-font-loading-3/Overview.bs
+++ b/css-font-loading-3/Overview.bs
@@ -843,14 +843,10 @@ The <code>check()</code> method</h3>
 			the method returns <code>true</code>,
 			as the text will be rendered in the UA's fallback font instead,
 			and won't trigger any font loads.
-		* If none of the specified fonts exist,
-			even though this is technically similar to the previous case
-			(in that text rendered with that font list would just use the UA fallback font),
-			the method instead throws an error.
-			This is because such a situation is almost certainly either a typo,
-			or the result of changing the name of a downloadable font
-			and forgetting to update all places the old name was used,
-			and an error is more useful than a vacuous <code>true</code>.
+		* Likewise, if none of the specified fonts exist (for example, names are mis-spelled),
+			the method also returns <code>true</code>,
+			because using this font list will not trigger any loads;
+			instead, fallback will occur.
 	</div>
 
 	When the <dfn method for="FontFaceSet" lt="check(font, text)|check(font)">check</dfn>(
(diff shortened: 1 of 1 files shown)


svgeesus:
@jfkthame have you linked your W3C account with your GitHub account. Our IPR bot claims not to know who you work for. We do, of course, but the bot won't listen to us.

svgeesus (approved):
I agree that this is closer to what is actually interoperably implemented.

jfkthame:
> @jfkthame have you linked your W3C account with your GitHub account. Our IPR bot claims not to know who you work for. We do, of course, but the bot won't listen to us.

I have now. :) (I wasn't aware of that procedure/requirement. But it turned out to be trivial, of course.)

Issue #5303, opened 2020-07-09
[css-text-3][css-fonts-3][css-text-decor-3] Distinguish applying to text vs applying to boxes


fantasai:
We've historically only used the 'Applies to' lines to apply to boxes, but it seems we do need to distinguish whether something applies to an inline box vs. line box vs. text:

- For ::marker, a lot of properties don't apply to the marker box, because we don't know what kind of box an outside marker is yet. But they should inherit through to the text and apply there.
- For `display: contents`, properties can be applied to an element that generates no box, but still be pass through to its text.

So I'm thinking we should update the propdef tables of properties that apply to text to say "and text", so we can better define how these cases work. I think the rough idea is to do that for:
- `color` (and maybe all of css-color-adjust?)
- all font properties
- all inheritable css-text-decor properties
- all properties in css-text that apply to inline boxes?

and, going forward, not do that for any properties that rely on knowing box geometry.

(See https://www.w3.org/TR/css-display-3/#intro for discussion of boxes/elements/text runs etc.)

Does this make sense?

Loirooriol:
My assumption was that all inherited properties that apply to inline boxes, also apply to text.

Seems my assumption was wrong in the case of `line-height` on Firefox, but on Chromium it does apply to text.

frivoal:
Proposed PR in https://github.com/w3c/csswg-drafts/pull/5761

TalbotG:
CSS3Text property (white-space) that applies to inline boxes also apply to text

Tests: 
http://www.gtalbot.org/BrowserBugsSection/CSS3Text/white-space-applies-to-text-007.html

http://www.gtalbot.org/BrowserBugsSection/CSS3Text/white-space-applies-to-text-008.html

References:
http://www.gtalbot.org/BrowserBugsSection/CSS3Text/reference/white-space-applies-to-text-007-ref.html

http://www.gtalbot.org/BrowserBugsSection/CSS3Text/reference/white-space-applies-to-text-008-ref.html

I will update [PR26877](https://github.com/web-platform-tests/wpt/pull/26877) in a few min.

PR #5761, opened 2020-12-02
[css-text][css-text-decor][css-writing-modes][css-color][css-fonts][css-color-adjust] Be explicit about which properties apply to text. #5303

fantasai:


fantasai:
Not sure I caught them all, but I tried. :)

svgeesus (approved):
Yes, this seems correct to me.

Loirooriol:
Are you trying to cover all the specs here? At quick glance, things like `cursor` from css-ui should probably be added too.

fantasai:
@Loirooriol I think `cursor` is a bit tricky. We don't define the geometric bounds of text, except insofar as the glyph outline is defined, so if we were to say it applies to text, we wouldn't be able to say where it applies.

frivoal (approved):

css-writing-modes-4/Overview.bs
```diff
@@ -2537,7 +2537,7 @@ Horizontal-in-Vertical Composition: the 'text-combine-upright' property</h3>
   Name: text-combine-upright
   Value: none | all | [ digits &lt;integer&gt;? ]
   Initial: none
-  Applies to: non-replaced inline elements
+  Applies to: [=inline boxes=] and text
```
Loirooriol, line 2540: `text-combine-upright: all` is defined with

> Attempt to typeset horizontally all consecutive typographic character units within the box

So it's not clear what happens for text. Maybe "within the box or text run"?

css-text-3/Overview.bs
```diff
@@ -824,7 +824,7 @@ Order of Operations</h4>
   Name: white-space
   Value: normal | pre | nowrap | pre-wrap | break-spaces | pre-line
   Initial: normal
-  Applies to: <a>inline boxes</a>
+  Applies to: text
```
Loirooriol, line 827: The text below says

> whether and how <a>white space</a> inside the element is collapsed

Saying "element" seems strange if the property applies to text.

Loirooriol (approved):
Looks good, but I think some surrounding text may need to be tweaked.

css-text-3/Overview.bs
fantasai: Deleted “inside the element”.

css-text-4/Overview.bs
```diff
@@ -815,7 +815,7 @@ White Space Collapsing: the 'text-space-collapse' property</h3>
 	Name: text-space-collapse
 	Value: collapse | discard | preserve | preserve-breaks | preserve-spaces
 	Initial: collapse
-	Applies to: all elements
+	Applies to: text
```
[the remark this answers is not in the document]
fantasai: Deleted “inside the element”

css-writing-modes-4/Overview.bs
fantasai: Hmm... this is an interesting point. I was originally thinking that applying this to text would only make sense for the `digits` value (which is why I didn't edit css-writing-modes-3) and not the `all` value. Can make that clearer, or can say that it only applies to boxes, or can apply it to text runs as you say. I'm not sure there's a real use case for the last one, so I think we should go with whatever makes the most sense to implementers. CC @litherum @kojiishi @jfkthame  What do you think? The situation is `span { display: contents; text-combine-upright: <something>; }`: should this have an effect or have no effect? If so should it work for both `all` and `digits` or only for one of them? It doesn't seem particularly useful, so the question is what's easiest to implement. :)

Loirooriol:
Should `writing-mode` apply to text?

```html
<span style="display: contents; writing-mode: vertical-lr">foo</span>
```

In Firefox it works just like without `display: contents`. In Chromium, the letters are upright, but each one in a different line, and the lines are shifted to the right, weird.

css-writing-modes-4/Overview.bs
Loirooriol:
```html
<div style="writing-mode: vertical-lr">
  <span style="display: contents; text-combine-upright: all">foo</span>
```
has effect in Firefox and Chromium. They don't support `digits`.

fantasai: OK, made the fixes in https://github.com/w3c/csswg-drafts/commit/0b31fa93ebe3630000c86f63c5f0dfe0da05280a

fantasai:
@Loirooriol I think not, it's supposed to create an inline block and I think that needs a box!!

Loirooriol:
Yes, but we could say that text with an orthogonal writing mode is placed inside an anonymous inline-block with the writing mode of the text.

css-writing-modes-4/Overview.bs
Loirooriol: Seems WebKit has `-webkit-text-combine: horizontal` instead of `text-combine-upright: all`, and it has no effect with `display: contents`. But it's a non-inherited property, so it's expected that the text doesn't inherit it, I guess.

PR #5788, opened 2020-12-13
[css-pseudo] Add text for the ::file-selector-button pseudo-element.

### LINKED ISSUE #5049: [css-selectors] Can you standardise a pseudo-element selector for file inputs?
As per this [Mozilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1635675), there are already vendor-specific selectors to style buttons rendered within `<input type=file>` controls for Webkit and Trident engines. Webkit alone represents the majority market share, EdgeHTML is now "legacy", leaving Gecko/Quantum as the only current major engine without it. Mozilla is reluctant to add their own vendor-prefixed selector when the sensible choice would be to make `::file-upload-button` official.

emilio:
Closes #5049

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -891,6 +891,22 @@ Placeholder Input: the ''::placeholder'' pseudo-element</h3>
   It's not clear how this should work, but it may be worth doing.
   See <a href="https://github.com/w3c/csswg-drafts/issues/2517">Issue 2417</a>.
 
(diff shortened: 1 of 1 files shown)


annevk:
Should this also be defined in HTML as that defines the relevant widget? (Or integrated in a more formal manner.)

cc @zcorpan

zcorpan:
Yeah. The relevant part is https://html.spec.whatwg.org/multipage/rendering.html#the-input-element-as-a-file-upload-control

Issue #4912, opened 2020-03-31
[scroll-animations] Support series of scroll offsets


mattgperry:
Inspired by [the proposal](https://github.com/w3c/csswg-drafts/issues/4337) to allow Element-based offsets, I propose we can simplify the `startScrollOffset` and `endScrollOffset` props into a single `scrollOffsets` array.

This would allow us to leverage `KeyframeEffect`'s ability to accept more than two keyframes in an animation.

# Example

The "Example 1. Reveal / Unreveal" example in the above ticket shows the definition of two `ScrollTimeline`s, like this:

```javascript
const revealTimeline = new ScrollTimeline({
  startScrollOffset: { target: image, edge: 'start', threshold: 0 },
  endScrollOffset: { target: image, edge: 'start', threshold: 100 },
});

const unrevealTimeline = new ScrollTimeline({
  startScrollOffset:{ target: image, edge: 'end', threshold: 100}
  endScrollOffset:{ target: image, edge: 'end', threshold:0}
});
```

To create one conceptually unique effect (fade in while on screen), we create two timelines, two effects, and two animations.

Instead, we can half this to just one, with a single `scrollOffset` array:

```javascript
const timeline = new ScrollTimeline({
  source: scroller,
  scrollOffsets: [100, 200, 800, 900]
});
```

Taken from @majido's [example](https://gist.github.com/majido/45af2f8466f906722d58b29945a80ad3#file-scroll-offset-array-reaveal_unreveal-js), this would change the above to:

```javascript
const timeline = new ScrollTimeline({
  scrollOffsets: [
     { target: image, edge: 'start', threshold: 0 },
     { target: image, edge: 'start', threshold: 100 },
     { target: image, edge: 'end', threshold: 100 }, 
     { target: image, edge: 'end', threshold: 0 }
  ]
});

const effect = new KeyframeEffect(
  image,
  { opacity: [0, 1, 1, 0]},
  { duration: 1000, fill: both }
);

const revealUnrevleaAnimation = new Animation(effect, timeline);
revealUnrevleaAnimation.play();
```

majido:
This is an interesting idea which I am in favor of. I can see how multiple scroll offsets can simplify the API and is more ergonomics due to the fact that it can be paired with multi-frame keyframe effects.

Follow the processing logic for keyframes here are some ideas on how to spec and implement this:

   - Consider single-entry offsets as being the "end"
[...]

PR #5803, opened 2020-12-17
[scroll-animations-1] Support for multiple scroll offsets #4912

ogerchikov:
Described specification for ScrollTimeline.scrollOffsets API.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -41,6 +41,9 @@ urlPrefix: https://drafts.csswg.org/cssom-view/; type: dfn; spec: cssom-view-1
     text: overflow direction; url: overflow-directions
 urlPrefix: https://html.spec.whatwg.org/multipage/browsers.html; type: dfn; spec: html
     text: document associated with a window; url: concept-document-window
+urlPrefix: https://drafts.csswg.org/css-animations-1/; type: dfn; spec: css-animations-1
+    text: findRule; url: interface-csskeyframesrule-findrule
+    text: CSSKeyframeRule; url: csskeyframerule
 </pre>
 <pre class=link-defaults>
 spec:html; type:dfn; for:/; text:browsing context
@@ -156,13 +159,13 @@ Using the CSS markup:
   @scroll-timeline collision-timeline {
     source: selector(#container);
     orientation: block;
-    scrollOffsets:  200px, 300px;
+    scroll-offsets: 200px, 300px;
   }
 
   @scroll-timeline union-timeline {
     source: selector(#container);
     orientation: block;
-    scrollOffsets:  250px, 300px;
+    scroll-offsets: 250px, 300px;
   }
 
   @keyframes left-circle {
@@ -230,7 +233,7 @@ this example could be written as follows:
 @media (prefers-reduced-motion: no-preference) {
   @scroll-timeline progress-timeline {
     source: selector(#body);
-    scrollOffsets: 0, 100%;
+    scroll-offsets: 0, 100%;
   }
 
   @keyframes progress {
@@ -288,7 +291,7 @@ if (window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
   var scrollTimeline = new ScrollTimeline({
     source: scrollableElement,
     orientation: "vertical",
-    scrollOffsets: '0px', '200px'
+    scrollOffsets: [CSS.px(0), CSS.px(250)]
   });
   animation.timeline = scrollTimeline;
 
@@ -733,7 +736,7 @@ The same logic can be done in CSS markup:
 
   @scroll-timeline image-in-scrollport {
     source: selector(#container);
-    scrollOffsets: selector(#image) end, selector(#image) start;
+    scroll-offsets: selector(#image) end, selector(#image) start;
   }
 
   #target {
@@ -787,7 +790,7 @@ as follows:
     1.  Run the procedure to [=resolve a scroll timeline offset=] for 'auto'
         with the <var>is first</var> flag set to <var>first offset</var> and
         add the resulted value into <var>effective scroll offsets</var>.
-1.  If {{ScrollTimeline/scrollOffsets}} has one element
+1.  If {{ScrollTimeline/scrollOffsets}} has exactly one element
     1.  Run the procedure to [=resolve a scroll timeline offset=] for 'auto'
         with the <var>is first</var> flag set to <var>first offset</var> and
         add the resulted value into <var>effective scroll offsets</var>.
@@ -835,7 +838,8 @@ range is handled before calling this procedure, as part of [=current time=]
 calculation.
 
 Note: The rationale behind choosing last matching offset in the array is to
-be consistent with CSS property overrides.
+be consistent with [=findRule=] method for finding which [=CSSKeyframeRule=]
+applies in a comma separated list of overlapping keyframes.
 
 ### The phase of a {{ScrollTimeline}} ### {#phase-algorithm}
 
@@ -994,13 +998,13 @@ instead of document's scrolling Element for `auto`.
 'orientation' descriptor determines the scroll timeline's {{orientation}}.
 
 <pre class='descdef'>
-  Name: scrollOffsets
+  Name: scroll-offsets
   For: @scroll-timeline
   Value: [ none | <<scroll-timeline-offset>> ]#
   Initial: none
 </pre>
 
-'scrollOffsets' descriptor determines the scroll timeline's {{scrollOffsets}}.
+'scroll-offsets' descriptor determines the scroll timeline's {{scrollOffsets}}.
 
 [=Scroll timeline offsets=] in CSS are represented by the
 <<scroll-timeline-offset>> type:
@@ -1073,7 +1077,7 @@ interface CSSScrollTimelineRule : CSSRule {
 
   <dt><dfn>scrollOffsets</dfn></dt>
   <dd>
-    The 'scrollOffsets' descriptor associated with the ''@scroll-timeline'', or "none" if not specified.
+    The 'scroll-offsets' descriptor associated with the ''@scroll-timeline'', or "none" if not specified.
   </dd>
 
   <dt><dfn>timeRange</dfn></dt>
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -794,7 +794,7 @@ as follows:
     1.  Set <var>first offset</var> to false.
 1.  For each <var>scroll offset</var> in the list of
     {{ScrollTimeline/scrollOffsets}}, perform the following steps:
-    1.  Let <var>effective offset</var> be a result of applying the procedure
+    1.  Let <var>effective offset</var> be the result of applying the procedure
         to [=resolve a scroll timeline offset=] for <var>scroll offset</var>
         with the <var>is first</var> flag set to <var>first offset</var>.
     1.  If <var>effective offset</var> is null, the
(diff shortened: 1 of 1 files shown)


kevers-google (commented):
Sorry for the tardy response. Overall looks good. Deferring to Rob and Anders for approval.

scroll-animations-1/Overview.bs
```diff
@@ -789,28 +770,66 @@ follows:
...
+        with the <var>is first</var> flag set to <var>first offset</var> and
+        add the resulted value into <var>effective scroll offsets</var>.
+1.  If {{ScrollTimeline/scrollOffsets}} has one element
+    1.  Run the procedure to [=resolve a scroll timeline offset=] for 'auto'
+        with the <var>is first</var> flag set to <var>first offset</var> and
+        add the resulted value into <var>effective scroll offsets</var>.
+    1.  Set <var>first offset</var> to false.
+1.  For each <var>scroll offset</var> in the list of
+    {{ScrollTimeline/scrollOffsets}}, perform the following steps:
+    1.  Let <var>effective offset</var> be a result of applying the procedure
```
andruud, line 797: "Let X be _a_ result of applying Y" is pretty uncommon. (Occurs several times here). 

"Let X be _the_ result of applying Y"?

scroll-animations-1/Overview.bs
```diff
@@ -789,28 +770,66 @@ follows:
...
+1.  Let <var>first offset</var> be true.
+1.  If {{ScrollTimeline/scrollOffsets}} is empty
+    1.  Run the procedure to [=resolve a scroll timeline offset=] for 'auto'
+        with the <var>is first</var> flag set to <var>first offset</var> and
+        add the resulted value into <var>effective scroll offsets</var>.
+    1.  Set <var>first offset</var> to false.
+    1.  Run the procedure to [=resolve a scroll timeline offset=] for 'auto'
+        with the <var>is first</var> flag set to <var>first offset</var> and
+        add the resulted value into <var>effective scroll offsets</var>.
+1.  If {{ScrollTimeline/scrollOffsets}} has one element
```
andruud, line 790: Should this say:

If {{ScrollTimeline/scrollOffsets}} has at least one element?

scroll-animations-1/Overview.bs
```diff
@@ -789,28 +770,66 @@ follows:
...
+1.  Let <var>offset weight</var> be the result of evaluating
+    <code> 1 / (<var>size</var> - 1)</code>.
+1.  Let <var>interval progress</var> be the result of evaluating
+    <code>(<var>offset</var> - <var>start offset</var>) /
+    (<var>end offset</var> - <var>start offset</var>)</code>.
+1.  Return the result of evaluating
+    <code>(<var>offset id</var> - 1 + <var>interval progress</var>) &times;
+    <var>offset weight</var></code>.
+
+Note: The rational behind choosing last matching offset in the array is to
```
andruud, line 831: rational -> rationale

scroll-animations-1/Overview.bs
```diff
@@ -789,28 +770,66 @@ follows:
...
+    1.  Set <var>first offset</var> to false.
+1.  Return <var>effective scroll offsets</var>.
+
+### The progress of a {{ScrollTimeline}} ### {#progress-calculation-algorithm}
+The <dfn>scroll timeline progress</dfn> is calculated by applying the procedure
+to <dfn>calculate scroll timeline progress</dfn> given |offset| as follows:
+
+1.  Let <var>scroll offsets</var> be a result of applying the procedure to
+    [=resolve scroll timeline offsets=] for {{scrollOffsets}}.
+1.  Let <var>offset id</var> correspond to the position of the last offset in
```
andruud, line 813: Suggestion: `<var>offset index</var>`

scroll-animations-1/Overview.bs
```diff
@@ -789,28 +770,66 @@ follows:
...
+    <var>scroll offsets</var> whose value is less than or equal to
+    <var>offset</var> and the value at the following position greater than
```
andruud, line 815: Does this mean that the specified scroll offsets have ordering constraints? If yes, what are they, and when/how are they enforced?

scroll-animations-1/Overview.bs
```diff
@@ -789,28 +770,66 @@ follows:
...
+1.  Let <var>end offset</var> be the value of next offset in <var>scroll
+    offsets</var> after <var>start offset</var>.
+1.  Let <var>size</var> be the number of offsets in <var>scroll offsets</var>.
+1.  Let <var>offset weight</var> be the result of evaluating
+    <code> 1 / (<var>size</var> - 1)</code>.
+1.  Let <var>interval progress</var> be the result of evaluating
+    <code>(<var>offset</var> - <var>start offset</var>) /
+    (<var>end offset</var> - <var>start offset</var>)</code>.
+1.  Return the result of evaluating
+    <code>(<var>offset id</var> - 1 + <var>interval progress</var>) &times;
```
andruud, line 828: > `<var>offset id</var> - 1`

I guess this means that `offset id` starts at `1`? Can't we just specify earlier that it starts at zero?

andruud (commented):
( One month later...: )

scroll-animations-1/Overview.bs
ogerchikov: This describes how to handle single scroll offset - add 'auto' for as a first element, the provided offset will be added as part of the next statement.

scroll-animations-1/Overview.bs
ogerchikov: Thanks! Done.

scroll-animations-1/Overview.bs
ogerchikov: Done.

scroll-animations-1/Overview.bs
ogerchikov: Thanks for pointing this out. It's a copy and paste from the [implementation](https://source.chromium.org/chromium/chromium/src/+/master:cc/animation/scroll_timeline.h;l=149;drc=131697ea2c25234744fe1917571bfa454c2d22cb?q=scroll_timeline&ss=chromium) where offset_id correspond to the end of the matching interval.
I added a comment to [crbug](https://bugs.chromium.org/p/chromium/issues/detail?id=1094014#c7) to update the implementation based on spec.

scroll-animations-1/Overview.bs
ogerchikov: Done.

scroll-animations-1/Overview.bs
ogerchikov: There is no ordering constrains. Explanation why last interval is chosen is explained in the note below the section. I added another note explaining why matching interval will always be found.

scroll-animations-1/Overview.bs
```diff
@@ -967,15 +987,14 @@ instead of document's scrolling Element for `auto`.
 
 'orientation' descriptor determines the scroll timeline's {{orientation}}.
 
-
 <pre class='descdef'>
-  Name: start
+  Name: scrollOffsets
```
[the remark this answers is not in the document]
ogerchikov: How about `scroll-offsets` to match Javascript definition of ScrollTimeline?

ogerchikov:
Thanks @andruud for reviewing! I addressed your comments besides the one which refers to `scrollOffsets` name. Before I make the change I would like to get consensus from other reviewers. Please take a look at the latest patch.

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
andruud: Seems somewhat unnecessary to have the `scroll-` prefix when we're already inside `@scroll-timline`. IIRC we had `scroll-time-range` etc before, but the prefix was dropped.

_But_, if there are reasons for keeping the `scroll-` prefix on `scroll-offsets` specifically, then `scroll-offsets` WFM.

scroll-animations-1/Overview.bs
andruud: Right. I did understand that at some point, then I confused myself.

Then probably this should say:

If {{ScrollTimeline/scrollOffsets}} has _exactly_ one element

Just to make it clear that we're not intending the "at least" meaning. (Which would prepend 'auto' for any non-empty list I guess).

scroll-animations-1/Overview.bs
```diff
@@ -506,9 +484,11 @@ There are two types of scroll timeline offset: [=container-based offset=], and
 offset=] or to [=resolve an element-based offset=] depending on the offset type.
 It is possible for a [=scroll timeline offset=] to be resolved to null.
 
-The <dfn>effective start offset</dfn> is the [=effective scroll offset=] that is
-resolved from {{start}}. The <dfn>effective end offset</dfn> is
-the [=effective scroll offset=] that is resolved from {{end}}.
+The <dfn>effective start offset</dfn> is the value of first offset in
+[=effective scroll offsets=] array or null if the array is empty.
```
flackr, line 488: What if the array has a single value in it? Shouldn't the effective offset be 'auto'?

scroll-animations-1/Overview.bs
```diff
@@ -506,9 +484,11 @@ There are two types of scroll timeline offset: [=container-based offset=], and
...
+
+The <dfn>effective end offset</dfn> is the value of last offset in
+[=effective scroll offsets=] array or null if the array is empty.
```
flackr, line 491: Should an empty array produce effective offsets of [auto, auto]?

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
flackr: There was a concern that offset has a different meaning in animation context, to specify the position of each keyframe. We may in fact want to add offsets as well as scrolloffsets in the future so that you could map scroll offsets to arbitrary animation offsets rather than having them uniformly distributed.

scroll-animations-1/Overview.bs
flackr: Nevermind, on rereading the algorithm for calculating effective scroll offsets, it looks like the one element offsets case becomes two effective scroll offsets.

scroll-animations-1/Overview.bs
```diff
@@ -789,28 +770,72 @@ follows:
...
+    <var>offset weight</var></code>.
+
+Note: This procedure guarantees that given <var>offset</var> belongs to
+scroll range within first and last offsets of <var>scroll offsets</var>.
+This is because the case when <var>offset</var> falls outside of the scroll
+range is handled before calling this procedure, as part of [=current time=]
+calculation.
+
+Note: The rationale behind choosing last matching offset in the array is to
+be consistent with CSS property overrides.
```
flackr, line 838: I think we should link to the rule for finding which KeyframeRule applies in a comma separated list of keyframes with overlap as justification: https://drafts.csswg.org/css-animations-1/#interface-csskeyframesrule-findrule

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
ogerchikov: Changed to scroll-offsets.

scroll-animations-1/Overview.bs
ogerchikov: Done

scroll-animations-1/Overview.bs
```diff
@@ -294,8 +288,7 @@ if (window.matchMedia('(prefers-reduced-motion: no-preference)').matches) {
   var scrollTimeline = new ScrollTimeline({
     source: scrollableElement,
     orientation: "vertical",
-    scrollOffset: '0px',
-    end: '200px'
+    scrollOffsets: '0px', '200px'
```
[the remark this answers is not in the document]
ogerchikov: Done.

scroll-animations-1/Overview.bs
```diff
@@ -156,15 +156,13 @@ Using the CSS markup:
   @scroll-timeline collision-timeline {
     source: selector(#container);
     orientation: block;
-    start:  200px;
-    end: 300px;
+    scrollOffsets:  200px, 300px;
```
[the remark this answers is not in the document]
ogerchikov: Done.

scroll-animations-1/Overview.bs
ogerchikov: Done.

ogerchikov:
Thanks for reviewing! All your comments are addressed in the recent commit.

birtles (approved):
If you've got review from @kevers-google, @andruud, and @flackr then I'm pretty sure you don't need mine! (I did skim it however, and it looked very thorough.)

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
andruud: I see, makes sense.

andruud (approved):

flackr (approved):

foolip:
Was there a Chromium bug filed for this? `CSSScrollTimelineRule` was changed to remove `start` and `end` properties, but those are still in Chromium.

ogerchikov:
@foolip Yes, https://bugs.chromium.org/p/chromium/issues/detail?id=1094014 includes removal of start and end properties.

foolip:
Thanks @ogerchikov! I noticed the removal when updating IDL for https://mdn-bcd-collector.appspot.com/, and wasn't sure if I should add back some custom tests for this or not. I won't then, if it's not going to be left long in Chromium.

PR #5834, opened 2021-01-05
[css-ui-4] Add full <image> support for 'cursor' property

### LINKED ISSUE #5831: [css-ui] Should the cursor property support image-set()?
WebKit's implementation of image-set allows you to use `cursor: -webkit-image-set()`.

There's a few tests in WPT (http://wpt.live/css/css-ui/cursor-image-014.html, http://wpt.live/css/css-ui/cursor-image-015.html), that seem to claim it should work (and it should support full `<image>` syntax). However the spec syntax only has `<url>`: https://drafts.csswg.org/css-ui-4/#cursor

cc @frivoal

SebastianZ:
This change replaces the currently defined `<url>` value supported by the `cursor` property by its superset `<image>`.

Closes #5831

Sebastian

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -601,7 +601,7 @@ user resizing of that element.
 
 <pre class="propdef">
 Name: cursor
-Value: [ [ <<url>> | <<url-set>> ] [&lt;x&gt; &lt;y&gt;]? ]#? <br>
+Value: [ <<image>> [&lt;x&gt; &lt;y&gt;]? ]#? <br>
  [ auto | default | none |<br>
  context-menu | help | pointer | progress | wait | <br>
  cell | crosshair | text | vertical-text | <br>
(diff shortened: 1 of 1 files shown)


frivoal:
I think this PR is premature unless UAs are actually implementing it. We absolutely should make sure that it's allowed, so that they can go in that direction, but making it a requirement is just going to slow down the spec if nobody is actually doing it. Until that changes (and I don't believe it has changed), I'd suggest closing.

SebastianZ:
@emilio just implemented [`image-set()` for `cursor` in Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1695402) and another [issue was filed lately](https://bugzilla.mozilla.org/show_bug.cgi?id=1696314) by the Mozilla team to also support gradients on it. So at least Mozilla has implementation interest, it seems.

Maybe Emilio can provide an official statement here on their plans.

Sebastian

emilio:
We filed that bug just for tracking, but we're not particularly likely to get to it any time soon. I don't see a particular use case for gradients on a cursor and for `$reasons` it was harder to implement. `image-set()` (with urls) seemed worth supporting, and other browsers also support it.

SebastianZ:
Ok, so I close this change then and I'll come up with one that only adds `image-set()`.

Sebastian

tabatkins:
@emilio `image-set()` allows arbitrary `<<image>>`s in its arguments, including gradients. Are y'all only supporting a subset of `image-set()` functionality in this case? If so, maybe we should standardize that as a restricted type (similar to how only fixed-length values are allowed in `repeat(auto-fit)` for grids); if not, I don't think it makes sense to disallow something on its own that you could validly get just by wrapping it in an `image-set()`.

SebastianZ:
@emilio Ping on @tabatkins' question.

SebastianZ:
I close this now due to the lack of implementer support.

Sebastian

PR #5838, opened 2021-01-06
[css-fonts-4] Use “create a potential-CORS request”

sideshowbarker:
This change makes the CSS Fonts spec reference the “create a potential-CORS request” algorithm from the HTML spec.

Otherwise, without this change, the CSS Fonts spec references “potentially CORS-enabled fetch”, which is an obsolete algorithm that no longer exists either in the Fetch spec nor in HTML.

---

Related: https://github.com/whatwg/html/pull/6265

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -24,6 +24,8 @@ spec:css-color-4; type:property; text:color
 spec:css-values; type:value; text:ex
 spec:css22; type:value; for:/; text:block
 spec:html; type:element; text:font
+spec:fetch; type:dfn; for:/; text:request
+spec:fetch; type:dfn; for:/; text:fetch
 </pre>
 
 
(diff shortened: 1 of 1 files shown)


annevk:
Why can this not create a request directly? HTML's abstraction is mostly useful if you have a `crossorigin` attribute, which I doubt is the case here?

sideshowbarker:
> Why can this not create a request directly? HTML's abstraction is mostly useful if you have a `crossorigin` attribute, which I doubt is the case here?

Good point. I'll rewrite it

svgeesus:
> Why can this not create a request directly? HTML's abstraction is mostly useful if you have a crossorigin attribute, which I doubt is the case here?

It is not the case for the CSS `url()` syntax, no.

It will be the case for the planned replacement for `url()` which among other things will deal with origin.

svgeesus:
> Good point. I'll rewrite it

Will wait for your rewrite before merging.

annevk:
@svgeesus even in that case I doubt it would have an actual HTML attribute state, which is what that algorithm expects.

svgeesus:
@sideshowbarker 

> Will wait for your rewrite before merging.

Is the force-push the rewrite, or should I still wait?

sideshowbarker:
Hi @svgeesus 
> Is the force-push the rewrite, or should I still wait?

The force-push is not the rewrite, so please wait. (Sorry for not having gotten back to this yet; I’ll try to do it early this week.)

triple-underscore:
Some of other CSS modules also use “potentially CORS-enabled fetch” algorithm.
Specifically, I found it in css-masking, css-shapes and filter-effects (and css-font-3).
It would be nice if they are updated together.

sideshowbarker:
> Some of other CSS modules also use “potentially CORS-enabled fetch” algorithm.
> Specifically, I found it in css-masking, css-shapes and filter-effects (and css-font-3).
> It would be nice if they are updated together.

OK, thanks for the heads-up — I’ll update those too

sideshowbarker:
OK, I’ve amended the commit and this is now ready for re-review

sideshowbarker:
> Some of other CSS modules also use “potentially CORS-enabled fetch” algorithm.
> Specifically, I found it in css-masking, css-shapes and filter-effects (and css-font-3).
> It would be nice if they are updated together.

I’d prefer to get this one change right first — rather than trying to include patches to multiple other specs in this PR. Once we have the language in this PR right, the language can be re-used for patches to the other specs.

svgeesus (approved):
Excellent, thanks

PR #5839, opened 2021-01-06
[selectors-4][editorial] Remove mention of <link> being a :link match

tabatkins:
Mergeable when <https://github.com/whatwg/html/pull/6269> merges, and not before.


domenic:
The HTML change has landed, so this probably should now too :)

PR #5842, opened 2021-01-07
[css-animations-1] Updating typo in CSSKeyframesRule.findRule

rachelandrew:
I was documenting this over at MDN and spotted that the information for `select` in findRule mentioned the "rule to be deleted", I think this should be "found".


birtles (approved):
Thank you!

PR #5851, opened 2021-01-11
[css-color-4] Fix typos in examples of lab(), lch() and color(lab ...)

weinig:
I noticed a few typos in the example uses of lab(), lch() and color(lab ...) as I was reading through spec.


svgeesus (approved):
Thanks!

PR #5875, opened 2021-01-17
[css-2020] Remove css-transforms-1 from pre-CR exclusions

SelenIT:
Since [css-transforms-1] has been a CR since 2019 and is already listed in the Official Definition, it seems unreasonable to keep it among the pre-CR excections in the 2020 snapshot. I assume this is an editorial fix.


frivoal:
Nice catch, thanks.

PR #5948, opened 2021-02-07
[css-rhythm-1] Collapse img for Bikeshed

nschonni:
Maybe more of a Bikeshed issue, but was outputing as `` before


tabatkins:
Yeah, it's a known issue with the fact that Bikeshed files are Markdown that contains HTML, and `>` on a line starts a blockquote.

nschonni:
Ah, that makes sense

PR #5955, opened 2021-02-08
[css-fonts-4] HTML validation from Bikeshed

nschonni:
- Drop old summary attribute from table
- Trailing paragraphs in OLs
- Nested links

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -6584,12 +6584,12 @@ None.
 
 <h3 id="sp213">How does this specification distinguish between behavior in first-party and third-party contexts?</h3>
 
-<a href="#font-fetching-requirements">For font loads, user agents must use the
-potentially CORS-enabled fetch method defined by the [HTML] specification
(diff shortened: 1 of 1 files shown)


css-fonts-4/Overview.bs
```diff
@@ -6585,7 +6585,7 @@ None.
 <h3 id="sp213">How does this specification distinguish between behavior in first-party and third-party contexts?</h3>
 
 <a href="#font-fetching-requirements">For font loads, user agents must use the
-potentially CORS-enabled fetch method defined by the [[!HTML]] specification
```
svgeesus, line 6588: No, that prevents HTML being a normative reference.
I agree tat the nested link should be fixed, but not like that.

svgeesus (changes requested):

css-fonts-4/Overview.bs
tabatkins: Fixed, I restored the biblio link and just pulled back the `a` to only cover the starting phrase, which is still a reasonable link text I think.

svgeesus (approved):

PR #5958, opened 2021-02-08
typo: Double word elements

nschonni:
Could go a few ways as possessive or singular then plural, but I think these are duplicates


frivoal:
Thanks!

PR #5982, opened 2021-02-11
[css-ruby-1] Rename Issues file to BSI extension

nschonni:
I get a pretty different HTML file generating locally on the Bikeshed repo, but didn't add that to this PR.


frivoal:
This particular issues list is being processed through bin/issuegen.pl (in the same repo), which seems to be doing escaping a bit differently from bikeshed, and is also why the generated file is kind of different. Given that script, I don't think this PR is right (or necessary).

Minimizing this kind of difference between the two tools is probably a good idea over all, but I am unsure how much effort that is worth.

frivoal:
PS: while this particular fix might not be right, thanks for the numerous clean-up PRs. Definitely noticed and appreciated.

nschonni:
No problem. Looking at other files in the repo, they have a `.bsi` or `.txt` extension. Should it maybe be renamed?

nschonni:
@frivoal I went ahead an recycled this PR to rename it to a "BSI" extension.
Downstream, Bikshed won't try to pull it in that way, and it matches the extension that the Perl script is looking for

PR #5991, opened 2021-02-13
[css-fonts-5] Specify the advance-override @font-face descriptor

litherum:
Specify the advance-override @font-face descriptor

cc @xiaochengh @svgeesus


css-fonts-5/Overview.bs
```diff
@@ -168,19 +168,42 @@ For: @font-face
...
 Note: User agents may draw data from different places from the font file as the metric values,
 which results in different text layouts.
 
-When the descriptor value is a percentage, the corresponding metric value is resolved as the
+For the 'ascent-override', 'descent-override' and 'line-gap-override' descriptors, when the
+descriptor value is a percentage, the corresponding metric value is resolved as the
 given percentage multiplied by the used font size. Negative values are invalid at parse time.
 
+For the 'advance-override' descriptor, when the descriptor value is a percentage, the
+spacing between adjacent <a>typographic character units</a> is multiplied by the percentage.
```
svgeesus, line 195: Is it the spacing between, that is multiplied? It seems that it is the total advance width, (which include the part taken up by the glyph) which is multiplied. If the percentage is greater than 100, that will have the effect of increasing the spacing between TCUs.

That probably needs a diagram, (later, not to land this pull request)

css-fonts-5/Overview.bs
```diff
@@ -220,14 +243,16 @@ given percentage multiplied by the used font size. Negative values are invalid a
...
+  advance-override: 109%;
 }
 
 &lt;div style="font-family: cool-web-font, fallback-to-local"&gt;Title goes here&lt;/div&gt;
 &lt;img src="https://example.com/largeimage" alt="A large image that you don't want to shift"&gt;
 </pre>
 
-	The image will not be vertically shifted when the user agent finishes loading and
-	switches to use the web font.
+	The image will not be shifted as much when the user agent finishes loading and
```
svgeesus, line 253: Yes, better to under- than over-promise here

css-fonts-5/Overview.bs
litherum: Fixed in https://github.com/w3c/csswg-drafts/pull/5991/commits/8cd3dd90e1e582dc0fd54c7799845a2588f37e75.

An example with diagrams is in https://github.com/w3c/csswg-drafts/pull/5991/commits/a6e10d5130ff746374ee54fc28857b541037d7e1.

css-fonts-5/Overview.bs
litherum: 👍

svgeesus (approved):

xiaochengh:
(Missed it last Friday due to Chinese New Year...)

Thanks for putting this up!

During the meeting, we said there should be a second parameter to override the advance heights of the glyphs, in case we have vertical text where characters are not rotated.

Do we still need it?

PR #6013, opened 2021-02-18
[css-fonts-5] Remove space in image names

nschonni:
Alternately the spaces could just be `%20` escaped in the `src` values on the img


svgeesus (approved):

litherum:
Why are spaces in filenames a problem?

nschonni:
It's not a terrible problem, but they need to be escaped in the `src` value. But non-spacing is usually better for filesystems globbing

litherum:
I think I disagree with this change, but not enough to act upon that belief... It's fiiiiiiiiine.... 🙁 grump grump I wish we could have nice things

PR #6015, opened 2021-02-18
[css-ruby-1] Replace obsolete TT tag

nschonni:


fantasai:
It's... not code, not really. :/ Idk what tag you use for ASCII art.

nschonni:
Fair, if it doesn't seem semantically clean, the other recommendation from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt is to use a span and font style

fantasai:
Ended up going with SAMP because it's probably the closest...

PR #6020, opened 2021-02-18
Add the second percentage parameter to advance-override

xiaochengh:
Following the previous discussion, this patch adds the second percentage parameter to `advance-override` for upright glyphs in vertical text layout.


xiaochengh:
@litherum @svgeesus PTAL, thanks!

svgeesus (approved):

xiaochengh:
Thanks!

Issue #6032, opened 2021-02-22
[css-highlight-api] In examples 2 and 3, setStart and setEnd fail


ffiori:
In examples 2 and 3 from the [draft](https://www.w3.org/TR/css-highlight-api-1/), the Ranges are built using document.body as start/end node. This node is of Element type, so the offset passed to setStart/setEnd is actually the number of child nodes between the start of document.body and not the number of characters. As a result, the browser fails to execute setStart and setEnd when the offset is greater than 1.

One possible fix for this could be writing document.body.childNodes[0] instead of document.body. In this way the offset is representing the number of characters since the start of the text node as intended.

sanketj:
Good find, and thanks for fixing this!

PR #6033, opened 2021-02-22
[css-highlight-api] Fix examples' Ranges

ffiori:
Resolves issue [6032](https://github.com/w3c/csswg-drafts/issues/6032).

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -308,15 +308,15 @@ Painting</h4>
 					</style>
 					<body>Lorem Ipsum.
 					<script>
-						let text_node = document.body.firstChild;
+						let textNode = document.body.firstChild;
 
 						let r1 = new Range();
-						r1.setStart(text_node, 1);
-						r1.setEnd(text_node, 5);
+						r1.setStart(textNode, 1);
+						r1.setEnd(textNode, 5);
 
 						let r2 = new Range();
-						r2.setStart(text_node, 3);
-						r2.setEnd(text_node, 7);
+						r2.setStart(textNode, 3);
+						r2.setEnd(textNode, 7);
 
 						CSS.highlights.add(new Highlight("sample", r1, r2));
 					</script>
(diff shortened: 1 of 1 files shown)


css-highlight-api-1/Overview.bs
```diff
@@ -309,12 +309,12 @@ Painting</h4>
 					<body>Lorem Ipsum.
 					<script>
 						let r1 = new Range();
-						r1.setStart(document.body, 1);
-						r1.setEnd(document.body, 5);
+						r1.setStart(document.body.childNodes[0], 1);
```
sanketj, line 312: I think it would be simpler to use `document.body.firstChild`.

css-highlight-api-1/Overview.bs
```diff
@@ -309,12 +309,12 @@ Painting</h4>
 					<body>Lorem Ipsum.
 					<script>
 						let r1 = new Range();
```
sanketj, line 311: For clarity, I would recommend making a text node variable and using that in the calls to setStart/setEnd.

sanketj (changes requested):

css-highlight-api-1/Overview.bs
ffiori: Makes sense, I'll change it

css-highlight-api-1/Overview.bs
```diff
@@ -308,13 +308,15 @@ Painting</h4>
 					</style>
 					<body>Lorem Ipsum.
 					<script>
+						let text_node = document.body.firstChild;
```
sanketj, line 311: I think we should use `textNode` instead of `text_node`. That seems to be more aligned with the style I've seen in other specs.

ffiori: Makes sense, there aren't any other long variable names in this draft and I thought it was a spec-independent thing, but your comment is right, I'll change this.

sanketj (approved):

PR #6060, opened 2021-03-02
[css-color-4] Rephrase interpolation to be more generic wrt to interpolation spac

LeaVerou:
Currently, it is phrased as if interpolating in any other color space is an exception to Lab. If we're going to be referencing this for `color-mix()` and the like, it needs to be more generic wrt to interpolation space.


argyleink (approved):

svgeesus (approved):
LGTM

Issue #6064, opened 2021-03-02
[css-color-5] More clarity around the percentage in color-mix()


mirisuzanne:
[The defintition of `color-mix()` in CSS Colors level 5](https://drafts.csswg.org/css-color-5/#color-mix) includes a `<percentage>` value for defining the relative weight of the colors being mixed. That weight is given to one of the two colors, with the remaining percentage given to the other color.

From the examples, I can infer that the given weight is used for the _first_ color, and the remaining percentage is applied to the _second_ color – `color-mix(red 80% blue in lch)` will be 80% red and 20% blue – but I don't think that's clearly stated anywhere. It also seems to contradict this note:

> Note: As a corrollary, 0% just returns the first color converted to the specified colorspace, 100% returns the second color converted to the specified colorspace, and 50% produces an equal mix of both.

I would expect that to be the opposite. 

- If we apply `0%` of the first color, we should get the second one converted to a given colorspace.
- If we apply `100%` of the first color, we should get that same color, but converted.

(I'm not sure if there's also a way to make this more clear in the syntax, but that's a separate issue. I have to look this up every time I use the Sass function as well. My mind refuses to retain this information…)

LeaVerou:
Yup, that's an excellent point. The percentage could be interpreted as "X% of the first color" or "X% along the progression". Currently the grammar is written with the first definition in mind (hence why the percentage cannot be after the second color) but the prose describes the second definition.
Perhaps we should make this more clear with syntax (e.g. with an `of` keyword)

argyleink:
agree that the `Note` is opposite. I've submitted a PR adjusting the note #6065 

thoughts on adding 2 keywords so it's suuuuper clear?.. 

```css
color-mix(0% of red with blue in lch)
color-mix(100% of red with blue in lch)
color-mix(80% of red with blue in lch)
```

Myndex:
The following may be tangential to this thread, but:

0% should always have the same result, regardless of linear or the gamma for gamma encoded space.
100% should also.

However 50% should be expected to be very different depending on the gamma or linearness of the space, and just as an FYI, alpha values are affected by the gamma of the color values, so an alpha of 0.5 effectively has a different perceptual effect for different colorspaces, i.e. if you have a color in linear space with an alpha of 0.5, when converting that color to a gamma encoded space, the alpha also needs to be adjusted/converted.

LeaVerou:
[...]

PR #6065, opened 2021-03-02
Fixes note about 100% and 0% usage with `color-mix()`

argyleink:
[css-color-5] ref #6064


LeaVerou:
This is good, but doesn't entirely correct the issue. There's more prose that needs to be edited. I'm going to push a fix now.

PR #4275, opened 2019-09-04
[css-tables-3] Remove invisible symbols in <'border-spacing'> syntax definition

lahmatiy:
Removed symbols even not visible on GitHub's diff view.
The diff from GitHub Desktop client:
image

PR #6077, opened 2021-03-04
[css-tables-3] Drop zero-width spaces in border-spacing dfn

tidoust:
This addresses the syntax issue raised in #4275 (which got merged then reverted because `<>` have to be doubled in practice). The goal of the update is to make the definition compliant with the [Value Definition Syntax](https://www.w3.org/TR/css-values-4/#value-defs).


FremyCompany:
Looks good to me

Issue #4431, opened 2019-10-17
[css-backgrounds] Make box-shadow a Shorthand Property


Crissov:
- [Level 3 `box-shadow`](https://drafts.csswg.org/css-backgrounds-3/#box-shadow)
- [Level 4 `box-shadow`](https://drafts.csswg.org/css-backgrounds-4/#box-shadow)

```` abnf
box-shadow: none | <shadow>#
<shadow>: <length>{2,4} && inset? && <color>?
````

Since box shadows are frequently altered on hover etc., but not all of its components need to change, can we make them available independently?

```` abnf
box-shadow-color: <color>
box-shadow-offset: <length>{2} ; x/horizontal y/vertical
box-shadow-blur: <length>      ; or 'box-shadow-radius'
box-shadow-spread: <length>    ; or 'box-shadow-distance'
box-shadow: none | <shadow>#
<shadow>: [<'box-shadow-offset'> [<'box-shadow-blur'> <'box-shadow-spread'>?]?] \ 
          && [inset || <'box-shadow-color'>]?
````

tabatkins:
Yeah, we've wanted to do this for a long while, precisely for the reason you give. Just haven't gotten to it yet.

Loirooriol:
I guess `text-shadow` may also need this.

Crissov:
Indeed. Unfortunately, `<shadow>` cannot be unified as far as I can see. 

```` abnf
text-shadow-color: <color>
text-shadow-offset: <length>{2} ; x/horizontal y/vertical
text-shadow-blur: <length>      ; or 'text-shadow-radius'
text-shadow: none | <shadow>#
<shadow>: [<'text-shadow-offset'> <'text-shadow-blur'>?] \ 
          && <'text-shadow-color'>?
````

SebastianZ:
The suggested syntaxes are missing a few things:

1. `box-shadow` takes a _list_ of shadows, so all the longhands also need to take lists, similar to the different `background-*` longhands.
2. There's no longhand for the `inset` keyword.
3. There is no spread radius property defined for `text-shadow` while it was [added to it in Text Decoration 4](https://drafts.csswg.org/css-text-decor-4/#text-shadow-property).
4. The `none` keyword is not part of any longhand. (Or are there any precedent cases for that?)

So regarding those points, I suggest the following:

For [`box-shadow`](https://drafts.csswg.org/css-backgrounds-3/#box-shadow):

```abnf
[...]

PR #6083, opened 2021-03-06
[css-backgrounds-4] Added box-shadow-* longhand properties

SebastianZ:
This change adds the different longhand properties for the `box-shadow` property to the CSS Backgrounds and Borders Module Level 4. This is one part of #4431.

The definition of the `box-shadow` property was taken from [level 3 of the specification](https://drafts.csswg.org/css-backgrounds-3/#box-shadow). Then the different longhand properties were created by splitting the syntax and value descriptions of the `box-shadow` property. Some additional descriptions were added and all the properties placed in a new section "Drop Shadows".

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -690,6 +690,184 @@ The 'border-clip' properties</h3>
 		<p>The fragments are shown in red for illustrative purposes; they should be black in compliant UAs.
 	</div>
 
+<h2 id="misc">Drop Shadows</h2>
+
+<h3 id="box-shadow-color">Coloring shadows: the 'box-shadow-color' property</h3>
+
+<pre class="propdef">
+Name: box-shadow-color
+Value: <<color>>#
+Initial: currentcolor
+Applies to: all elements
+Inherited: no
+Percentages: N/A
+Computed value: list, each item a computed color
+Animatable: by computed value
+</pre>
+
+<p>The 'box-shadow-color' property defines one or more drop shadow colors.
+The property accepts a comma-separated list of shadow colors.
+
+<p>See the section [[css-backgrounds-3#shadow-layers|“Layering, Layout, and
+Other Details”]] for how 'box-shadow-color' interacts with other
+comma-separated drop shadow properties to form each drop shadow
+layer.
+
+<h3 id="box-shadow-offset">Offsetting shadows: the 'box-shadow-offset' property</h3>
+
+<pre class="propdef">
+Name: box-shadow-offset
+Value: none | <<length>>{2}#
+Initial: none
+Applies to: all elements
+Inherited: no
+Percentages: N/A
+Computed value: either 'none' or a list,
+	each item a pair of offsets (horizontal and vertical) from the element‘s box
+Animatable: by computed value
+</pre>
+
+<p>The 'box-shadow-offset' property defines one or more drop shadow offsets.
+The property accepts a comma-separated list of horizontal and vertical offset pairs,
+where both values are described as <<length>> values.
+
+<dl>
+  <dt><dfn id="shadow-offset-x">1st <<length>></dfn>
+  <dd>
+    Specifies the <dfn>horizontal offset</dfn> of the shadow.
+    A positive value draws a shadow that is offset to the right of the box,
+    a negative length to the left.
+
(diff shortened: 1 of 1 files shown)


SebastianZ:
Given the [resolution about linked list-valued properties](https://github.com/w3c/csswg-drafts/issues/7164#issuecomment-1201340918) in #7164, it looks like we can finally move forward with this PR.

@fantasai As Lea seems to be extremely busy at the moment, is it possible that you do the review for this change?

Sebastian

LeaVerou:
It's hard to review this properly because the entire text from backgrounds-3 is brought in and then edited, so it's hard to know what the delta is. The PR as it currently stands includes stuff about drawing shadows, which is not relevant to breaking down the syntax into longhands.

Wrt to the longhands, some thoughts, not very organized:
- `box-shadow-offset` is not consistent naming with any other property that takes a horizontal and vertical offset. 
- We should have separate horizontal and vertical longhands. Not sure if these should be physical (like `background-position`) or logical. I defer to @fantasai on that.
- If we name the offset property `box-shadow-position`, we should name the one for `inset` differently. Perhaps `box-shadow-placement`?
- Yes, CSS is internally inconsistent here. We have conflicting precedents for `*-position` properties: `list-style-position` to follow for `inset | outset` and things like `background-position`. 
- OTOH `background-position` does not specify *offsets* from an initial position, it specifies a whole new position. So maybe that is not a good precedent to follow. What else in CSS specifies offsets? There is `top` and `left`, `translate`, what else?

SebastianZ:
> It's hard to review this properly because the entire text from backgrounds-3 is brought in and then edited, so it's hard to know what the delta is. The PR as it currently stands includes stuff about drawing shadows, which is not relevant to breaking down the syntax into longhands.

I've removed the Background 3 parts now. So it should be easier to review the changes.

> * `box-shadow-offset` is not consistent naming with any other property that takes a horizontal and vertical offset.
> * Yes, CSS is internally inconsistent here. We have conflicting precedents for `*-position` properties: `list-style-position` to follow for `inset | outset` and things like `background-position`.
> * OTOH `background-position` does not specify _offsets_ from an initial position, it specifies a whole new position. So maybe that is not a good precedent to follow. What else in CSS specifies offsets? There is `top` and `left`, `translate`, what else?

There are several reasons why I chose `box-shadow-offset` as name.

1. It defines an offset from the element's box.
2. Backgrounds 3 already calls those values "offsets".
3. It is the controlling property in the linked list-valued properties and therefore it takes the `none` value from the `box-shadow` shorthand. And `box-shadow-offset: none;` fits quite well, in my opinion.

> * We should have separate horizontal and vertical longhands. Not sure if these should be physical (like `background-position`) or logical. I defer to @fantasai on that.

I don't see a strong use case for having separate longhands for the two axes. And it seems @tabatkins [agreed on that on a call](https://github.com/w3c/csswg-drafts/issues/4431#issuecomment-1016728393):

> <emilio> smfr: should box-shadow-offset be further broken down into x/y offsets?
> <emilio> TabAtkins: [meh reaction]

But iff we add them, they have to be physical because the existing offsets in Backgrounds 3 are also physical. We might then add logical counterparts in addition to them, though.

>  * If we name the offset property `box-shadow-position`, we should name the one for `inset` differently. Perhaps `box-shadow-placement`?

I like the suggested name `box-shadow-placement` and I am open to rename it to that or something else but please see [my comment on why I chose that name](https://github.com/w3c/csswg-drafts/issues/4431#issuecomment-1016946929).

Anyway, bikeshedding the names can still be done after landing this patch.

Sebastian

SebastianZ:
@LeaVerou @fantasai This PR is lying around for exactly 1.5 years now. Could you do a quick review of the changes, please?

Sebastian

SebastianZ:
@fantasai @LeaVerou Another quarter has passed. Would it be possible to give this a(nother) review and provide some feedback, please? I'd really like to land this patch.

Sebastian

SebastianZ:
Because there was no further feedback since my update ~5 months ago, @LeaVerou's feedback got addressed by me and to pave the way to split the spec. and also to finally start working on adding the `text-shadow-*` longhands, I merge the changes now.

Discussions on further improvements and on naming the properties can happen in separate issues.

Sebastian

PR #6084, opened 2021-03-08
[css-device-adapt-1] Visual: Remove inline background-color style breaking dark theme

lf-:
Previously the background-color set inline for the pseudocode block resulted in nearly zero contrast between the text and the background. I feel it is preferable to remove the broken inline style and potentially later add a global style to readd a subtle background to all `pre` code blocks if desired.

## Before:

Screenshot_20210308_011832

Screenshot_20210308_011816

## After:

Screenshot_20210308_011729

Screenshot_20210308_011750

### DIFF
--- a/css-device-adapt-1/Overview.bs
+++ b/css-device-adapt-1/Overview.bs
@@ -1024,7 +1024,6 @@ comma as the preferred separator and semicolon as optional:
         white-space: pre;
         margin: 1em;
         padding: 1em;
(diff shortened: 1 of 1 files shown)


tabatkins:
Thank you for this, merged. If you spot more of these manually-specified backgrounds, feel free to file more PRs. ^_^ Either removing them, or adding a `@media (prefers-color-scheme: dark) {...}` variant is fine.

PR #6086, opened 2021-03-08
[css-ui] Use :focus-visible instead of :focus

mrego:
Use :focus-visible instead of :focus in the explanation about when UAs paint outlines.


frivoal:
This is not in text that establishes normative requirements, but text that gives examples of how the property can be used, so I'm deeming this to be an editorial change, and merging it without calling for a group resolution.

PR #6096, opened 2021-03-11
[css-lists] Define reversed counters with automatic initial value

### LINKED ISSUE #4181: [css-lists-3] Should automatic list-item increment adjust for ol[reversed]?
The list-item counter is defined to automatically increment by one on each element with `display: list-item`. (See https://github.com/w3c/csswg-drafts/issues/3686 for how that works.) `<ol reversed>` lists need to have a list-item increment of -1.

Should this be done by magic interactions with the host language, or should the UA style sheet simply include `ol[reversed] > li { counter-increment: list-item -1; }`?

### LINKED ISSUE #4211: [css-lists-3] Setting the start value of a reversed list
Note: We currently have https://github.com/w3c/csswg-drafts/issues/4181 open on reversing the increment in reversed lists.

Currently, the suggestion for mapping HTML reversed list starts to CSS is that HTML injects a `counter-reset` declaration on the OL element as part of the attribute-to-CSS mapping of the HTML presentational hint cascade level. There is, however, some amount of magic: the UA has to calculate the start number and pass it to CSS as an `<integer>`, e.g. `counter-reset: list-item 4` for a 3-item reversed list.

Doing some calculations to figure out the value to pass to CSS isn't particularly unusual: for example, HTML does some special parsing of colors before passing to `color` and `background-color`, and it does some bidi calculations before passing to `direction`.

What is super magical is that these calculations are currently defined to be dependent on CSS styles, not just on the contents of the markup. See [HTML spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element). In particular, [this case](http://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Col%20reversed%3E%0A%20%20%3Cli%3Eone%0A%20%20%3Cli%20style%3D%22display%3A%20none%22%3Enone%0A%20%20%3Cli%3Etwo%20%3Cspan%20style%3D%22display%3A%20list-item%3B%20margin%3A%201em%22%3Ethree%3C%2Fspan%3E%0A%20%20%3Cli%3Efour%0A%3C%2Fol%3E) is very strange, as not only is the `display: none` element skipped, the descendant SPAN is counted (which I personally find very surprising):

```
<ol reversed>
  <li>one
  <li style="display: none">none
  <li>two <span style="display: list-item">three</span>
  <li>four
</ol>
```

Some options going forward include:
- A) We define the preshint mapping as `counter-reset: list-item calc(N+1)` where N is magically calculated as defined currently⁠—counting all `display: list-item` _descendants_, but skipping the contents of any ol/ul/dir/menu elements.
- B) We define the preshint mapping as `counter-reset: list-item calc(N+1)` with some reduced amount of HTML+CSS co-dependence for the counting (assuming there's a definition of that that makes sense and is Web-compatible)?
- C) We define a reverse-counting counter feature, which HTML uses.
- D) We define CSS syntax to enable magic counting of  `display: list-item` descendants while skipping the contents of any ol/ul/dir/menu elements. HTML maps to that syntax.
- E) We define CSS syntax to enable counting, but use a simplified definition of what's counted (Option B). HTML maps to that syntax.
- F) Something else?

zcorpan:
This allows HTML to define the rendering for `<ol reversed>` in terms
of a presentational hint for `counter-reset: reversed(list-item)`
or `<ol reversed start=X>` to `counter-reset: reversed(list-item) X+1`.
See https://github.com/whatwg/html/pull/4816

Fixes #4181
Fixes #4211

### DIFF
--- a/css-lists-3/Overview.bs
+++ b/css-lists-3/Overview.bs
@@ -1142,7 +1142,7 @@ Instantiating Counters</h4>
 
 	When a [=counter=] is [=instantiated=] without an initial value,
 	the user agent must dynamically calculate the initial value at layout-time
-	to be the number of elements and pseudo-elements that are in [=scope=], plus 1.
+	to be the number of elements and pseudo-elements that [=instantiate=] the same counter in the same [=scope=], plus 1.
 
 	Note: Only [=reversed=] counters can be instantiated without an initial value.
 
--- a/css-lists-3/Overview.bs
+++ b/css-lists-3/Overview.bs
@@ -744,7 +744,8 @@ Automatic Numbering With Counters</h2>
 	which are inherited through the document tree in a way similar to inherited property values.
 	[=Counters=] have a <dfn export for="CSS counter">name</dfn> and <dfn export for="CSS counter">creator</dfn>,
 	which identify the counter,
-	and an integer <dfn export for="CSS counter">value</dfn>.
+	a boolean <dfn export for="CSS counter">reversed</dfn> (false by default),
+	and an integer <dfn export for="CSS counter">value</dfn> (optional when the counter is <a>reversed</a>).
 	They are created and manipulated with
 	the <dfn export>counter properties</dfn> 'counter-increment', 'counter-set' and 'counter-reset',
 	and used with the ''counter()'' and ''counters()'' [=functional notations=].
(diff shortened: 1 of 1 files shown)


css-lists-3/Overview.bs
```diff
@@ -1120,9 +1136,16 @@ Instantiating Counters</h4>
...
 			with name |name|,
 			originating element |element|,
-			and initial value |value|
+			reversed being |reversed|,
+			and initial value |value| (if given)
 	</div>
 
+	When a [=counter=] is [=instantiated=] without an initial value,
+	the user agent must dynamically calculate the initial value at layout-time
+	to be the number of elements and pseudo-elements that are in [=scope=], plus 1.
```
zcorpan, line 1145:
```suggestion
	to be the number of elements and pseudo-elements that [=instantiate=] the same counter in the same [=scope=], plus 1.
```

zcorpan:
Tests: https://github.com/web-platform-tests/wpt/pull/28453

Issue #6075, opened 2021-03-04
[css-fonts-5] Add glyph scaling override descriptor to @font-face


fantasai:
In https://github.com/w3c/csswg-drafts/issues/5983#issuecomment-788776441 @jfkthame wrote:

> I think there should be a separate font-size-override that can apply a scale factor to the used font size; I suspect this would often be the better tool to use.

Using a scaling factor to make adjustments for font fallback would avoid some of the more problematic aspects of advance-override (breaking cursive/complex scripts, breaking typographic rhythm by adding uneven spacing, breaking alignment at the edges of paragraphs).

It also provides a long-needed feature--optically matching fonts when mixing font faces--by allowing per-font scaling adjustments at the author's discretion. `font-size-adjust` does this for certain bicameral scripts by using the ex-height metric, but that doesn't work more generally across writing systems as noted in https://github.com/w3c/csswg-drafts/issues/4540 .

The proposal would be to add a descriptor taking a percentage and effectively functions [the same as `font-size-adjust`](https://www.w3.org/TR/css-fonts-4/#font-size-adjust-prop) for a given scale factor. Details to consider [suggestions in brackets to match font-size-adjust]:

- Effect on `em` units? [No effect on `em`.]
- Effect on `ch`/`ic`/`ex` units? [These are supposed to match glyph metrics, so are affected.]
- Effect on baseline table? [Scales the table.]
- Effect on ascent/descent metrics? [Should scale them also.]
- Effect on `ascent-override`/`descent-override`/etc.? [Scales with the scale factor.]
- Allow separate vertical/horizontal scale factors? [No idea. We can start with one and see if independent axes are needed.]

fantasai:
Possible names: `glyph-scale-factor`, `font-size-adjust`, `glyph-size-adjust`, `size-adjust`.

jfkthame:
Thanks for filing this! Mostly agree with the suggested details above, with one doubt:

>     * Effect on `ascent-override`/`descent-override`/etc.? [Scales with the scale factor.]

Not sure about this one. `ascent-override` etc are not derived from the size of the glyphs, they're provided so that the author can *override* the metrics coming from the font and establish known line-spacing metrics that are *not* dependent on the specific font:

> the corresponding metric value is resolved as the given percentage multiplied by the used font size

(from css-fonts [4.11. Default font metrics overriding](https://drafts.csswg.org/css-fonts/#font-metrics-override-desc)).

[...]

PR #6108, opened 2021-03-17
[css-fonts-5] Add size-adjust @font-face descriptor

litherum:
Fixes https://github.com/w3c/csswg-drafts/issues/6075

### DIFF
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -145,7 +145,7 @@ Issue(5892):
 
 <h3 id="font-metrics-override-desc">
 Default font metrics overriding:
-the 'ascent-override', 'descent-override', and 'line-gap-override' descriptors</h3>
+the 'ascent-override', 'descent-override', 'line-gap-override', and 'size-adjust' descriptors</h3>
 
 <pre class='descdef'>
 Name: ascent-override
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
This is exactly what I understood from the description by @fantasai  at today's CSS call.  Nicely done.

jfkthame:
Is there any reason to have a `normal` value for this descriptor, or should we just make its initial value `100%`?

css-fonts-5/Overview.bs
```diff
@@ -168,24 +168,49 @@ For: @font-face
...
+
+None of these descriptors affect the calculation of
+<a href="https://drafts.csswg.org/css-values-4/#font-relative-lengths">Font-relative Lengths</a>.
+None of these descriptors affect the behavior of any of the values of 'line-height' other than
+'normal'. The behavior of <code>line-height: normal</code> may be affected by all of these
+descriptors. 'size-adjust' does not affect the computed font size of any element.
+
+When 'size-adjust' is applied in conjunction with 'ascent-override', 'descent-override',
+or 'line-gap-override' in the same @font-face block, the relevant metrics are multiplied by both
+percentages.
```
jfkthame, line 206: This is a bit misleading, I think. The `{ascent|descent|line-gap}-override` descriptors *don't* multiply "the relevant metrics" by the given percentage; they *replace* the metrics derived from the font with metrics computed from the used font size.

astearns:
@jfkthame since this is merged, could you open a new issue or two with your comments? It will be too easy to lose your feedback otherwise.

css-fonts-5/Overview.bs
```diff
@@ -145,7 +145,7 @@ Issue(5892):
 
 <h3 id="font-metrics-override-desc">
 Default font metrics overriding:
-the 'ascent-override', 'descent-override', and 'line-gap-override' descriptors</h3>
+the 'ascent-override', 'descent-override', 'line-gap-override', and 'size-adjust' descriptors</h3>
```
jfkthame, line 148: I'm not sure `size-adjust` should be grouped under "Default font metrics overriding". It's significantly different from the metrics `...-override` descriptors and probably deserves separate description rather than being mixed in with them.

jfkthame:
> @jfkthame since this is merged, could you open a new issue or two with your comments? It will be too easy to lose your feedback otherwise.

Sorry, will do.

PR #6117, opened 2021-03-18
[cssom] Make CSSStyleRule serialization aware of nesting.

tabatkins:
@emilio review please?

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1523,16 +1523,33 @@ To <dfn export>serialize a CSS rule</dfn>, perform one of the following in accor
 <dl class="switch">
  <dt>{{CSSStyleRule}}
  <dd>
-  The result of concatenating the following:
-  <ol>
-   <li>The result of performing <a>serialize a group of selectors</a> on the rule's associated selectors.
-   <li>The string "<code> { </code>", i.e., a single SPACE (U+0020), followed by LEFT CURLY BRACKET (U+007B),
-   followed by a single SPACE (U+0020).
-   <li>The result of performing <a>serialize a CSS declaration block</a> on the rule's associated declarations.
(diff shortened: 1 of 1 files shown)


cssom-1/Overview.bs
```diff
@@ -1523,17 +1523,34 @@ To <dfn export>serialize a CSS rule</dfn>, perform one of the following in accor
...
+            <li>Append |decls| to |s|
+            <li>Append " }" to |s| (i.e. a single SPACE (U+0020) followed by RIGHT CURLY BRACKET (U+007D)).
+            <li>Return |s|.
+        </ol>
+    <li>Otherwise:
+        <ol>
+            <li>If |decls| is not null, prepend it to |rules|.
+            <li>For each |rule| in |rules|:
+                <ol>
+                    <li>Append a newline followed by two spaces to |s|.
```
emilio, line 1547: A bit weird that we're doing this fake indentation thing. So for nested declarations this would look like:

Source:

```
a {
  b {
    c {
      color: red;
    }
  }
}
```

cssText:

```
a {
  b {
  c { color: red; } }
}
```

Which is a bit off. I _think_ I'd rather avoid newlines completely, but not particularly opposed.

emilio (approved):

cssom-1/Overview.bs
tabatkins: I absolutely agree that this is busted, I'm just copying the existing indentation treatment for the other rules in the spec; @media already shows off that exact same awful behavior, for instance.

I didn't want to try and fix the indentation for *just* this rule and leave the rest busted, so I stuck with the current behavior and assume that all of them can be fixed up at once in a separate commit.

emilio: Sounds good to me, thanks.

Issue #6133, opened 2021-03-24
[css-fonts-5] Description of `size-adjust` descriptor is misleading


jfkthame:
Regarding `size-adjust`, the [current draft](https://drafts.csswg.org/css-fonts-5/#font-metrics-override-desc) says:

> For `size-adjust`, when the descriptor value is a percentage, all metrics values of this font, including glyph advance sizes, are resolved as the given percentage multiplied by the used font size.

This is not correct, I think. It says that every metric -- so that'd be ascent, descent, x-height, cap-height, etc -- as well as every glyph advance will end up with the same value: the used font size multiplied by the `size-adjust` percentage.

I think it should say something more like:

> All metrics values of this font, including glyph advances, are scaled by the given percentage, as are the rendered glyph images. However, the computed font size (and hence units such as `em` that depend on it) are unaffected.

fantasai:
@jfkthame I think this was fixed in https://github.com/w3c/csswg-drafts/commit/df024473bfb1a1e548a3ff42b25d70c92b98e67b#diff-3e391c0014318f36debe3e04cffbfc458ebcae3176bda0b904641e758106bbd1

PR #6137, opened 2021-03-24
[css-fonts-5] Move 'size-adjust' to its own section

### LINKED ISSUE #6110: [css-fonts] Initial value of `size-adjust` descriptor
Is there any reason to have a `normal` value for this descriptor? We could just use `100%` as the initial value.

### LINKED ISSUE #6112: [css-fonts] `size-adjust` should be moved out of the "Default font metrics overriding" section
The new `size-adjust` descriptor is not analogous to the previously-defined `ascent-override` and its friends, and probably shouldn't be grouped in the same section of the spec. Those descriptors *replace* metrics derived from the font with metrics computed from the used font size (as the section title says, they "override" things), while `size-adjust` applies a scale factor to all metrics (and images) derived from the font, but does not "override" them.

The draft text landed in #6108 includes `size-adjust` in the "metrics overriding" section, but then has to call out its different behavior multiple times. I suggest it would be better separated from the metrics overrides and described individually.

xiaochengh:
Fixes #6112. Fixes #6110.

### DIFF
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -139,13 +139,38 @@ Issue(5892):
 </dl>
 
 
+<h3 id="font-size-adjust-desc">
+Default font metrics and glyph size scaling: the 'size-adjust' descriptor
+</h3>
+
+<pre class='descdef'>
+Name: size-adjust
+Value: <<percentage>>
+For: @font-face
+Initial: 100%
+</pre>
+
+The 'size-adjust' descriptor defines a scalar by which glyph outlines and metrics from this font
+are multiplied. Negative values are invalid at parse time.
+
+All metrics values of this font, including glyph advance sizes, are resolved as the given percentage
+multiplied by the used font size. When glyphs are visibly drawn, the visual size of the glyphs is
+also multiplied by the percentage. 'size-adjust' does not affect the computed font size of any element.
+
+Note: Using the 'size-adjust' descriptor can achieve a similar effect as specifying a percentage for
+the 'font-size' property.
+
+Note: 'size-adjust' is applied per-font rather than per-element. Multiple fonts may be used within an
+individual element e.g. for characters not supported by the primary font.
+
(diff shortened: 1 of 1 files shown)


xiaochengh:
@litherum @svgeesus @jfkthame PTAL. Thanks!

css-fonts-5/Overview.bs
```diff
@@ -139,13 +139,38 @@ Issue(5892):
...
+Value: <<percentage>>
+For: @font-face
+Initial: 100%
+</pre>
+
+The 'size-adjust' descriptor defines a scalar by which glyph outlines and metrics from this font
+are multiplied. Negative values are invalid at parse time.
+
+All metrics values of this font, including glyph advance sizes, are resolved as the given percentage
+multiplied by the used font size. When glyphs are visibly drawn, the visual size of the glyphs is
```
jfkthame, line 157: I realize it's basically just being moved here, but this description is wrong; see #6133. But that could be fixed as a separate PR if you don't want to deal with it now.

xiaochengh: I'd like to deal with it in a separate PR. Doing code move and modifications at the same time may easily lose track of things. (Removing `normal` is trivial enough that I'd like to do together...)

svgeesus (approved):
LGTM.

I agree that the move and any subsequent edits should be separate commits.

xiaochengh:
Somehow the check never finishes... Could anyone help me merge it?

svgeesus:
I noticed that the bot seems to be down; I have reported it.

@fantasai has also [made a large PR](https://github.com/w3c/csswg-drafts/pull/6154) which does this move (plus a _bunch_ of other edits) so I suspect, unless review turns up problems, we would be better keeping her PR.

Which has the same IPR bot problem

fantasai:
Closing as fixed via https://github.com/w3c/csswg-drafts/pull/6154

PR #6146, opened 2021-03-25
[css-transforms-2] Add note that computed value of translate has percentages

### LINKED ISSUE #2124: [css-transforms-2] Serialization of the computed value of `translate` property.
I can not find a definition of serialization of individual transform's computed value in the spec[1]

When serializing a 'translate' property with percentage value, such as 'translate: 50%', should we resolve it to px or keep it unresolved?

PS: Currently, both edge and chromium resolve percentage to px.

[1] https://drafts.csswg.org/css-transforms-2/#individual-transforms

dbaron:
While the premise of #2124 was that this was not already defined, I
believe this is already the straightforward result of the definition in
cssom-1.  Thus this adds a note rather than additional normative text.

Fixes #2124.

### DIFF
--- a/css-transforms-2/Overview.bs
+++ b/css-transforms-2/Overview.bs
@@ -611,6 +611,11 @@ If three values are given,
 this specifies a 3d translation,
 equivalent to the ''translate3d()'' function.
 
+Note: The <a>resolved value</a> of the 'translate' property
+is the <a>computed value</a>,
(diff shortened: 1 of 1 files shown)


tabatkins:
Merged, but you don't need to PR and ask for my review when you're an editor, dbaron. ^_^

astearns:
It’s not a bad workflow to have at least one other person look at the changelist, though…

Issue #3305, opened 2018-11-08
[css-transforms-2] Is it necessary to serialize all 3 parameters of translate given the 3rd component is 0px


BorisChiou:
Just a follow-up question of this [issue](https://github.com/w3c/fxtf-drafts/issues/313). The new updated of serialization of `translate` [1]:
- If a 2d translation is specified, the property must serialize with only one or two values (per usual, if the second value is 0px, the default, it must be omitted when serializing).
- If a 3d translation is specified, all three values must be serialized.

This question is only for the case: `translate: 100px 100px 0px` vs `translate: 100px 100px`. IIRC, it seems Gecko and Blink treat the `translate(x, y)` as `translate(x, y, 0)` while converting it into a matrix for computation. (Maybe it will simplify it if it only has 2d components.) Basically, 3d transformation is different from 2d transformation (by using different number of axes), so my question here is, what is the potential side-effects if we serialize `translate: 100px 100px 0px` as `translate: 100px 100px`? If the side-effects is not obvious in this case, why not make it simpler.

[1] https://drafts.csswg.org/css-transforms-2/#individual-transform-serialization

BorisChiou:
cc @emilio

emilio:
cc @tabatkins 

TL;DR is: why is it incorrect to serialize `translate: 10px 0 0;` as `translate: 10px`. I don't get which side effect the spec talks about, at least for translation and scaling I'm pretty sure there's no such side effect.

tabatkins:
`translateZ(0)` absolutely has side effects; it's a classic hack to trigger forcing the element onto its own rendering layer. Iirc all the 3d transforms do so, even if they're "identity" transforms otherwise.

birtles:
As mentioned in https://github.com/w3c/fxtf-drafts/issues/313, that's WebKit/Blink-specific behavior. It's not clear to me that it's necessary to preserve that behavior with the individual transform properties.

birtles:
I guess I'm wondering if we can make Blink _not_ interpret `translate: 10px 10px 0` as a 3D transform? (While leaving `transform: translateZ(0)` as doing so.) I believe the individual transform properties are not shipping yet and are not implemented in WebKit?

Animating from `translate: 10px 10px -10px` to `translate: 10px 10px 10px` shouldn't trigger layer destruction/re-creation since so long as the element is animated, it effectively has `will-change: translate` applied and hence should keep its layer throughout.

tabatkins:
[...]

PR #6147, opened 2021-03-25
[css-transforms-2] Don't maintain state for whether individual transform property values are 2D or 3D.

dbaron:
The group resolved in #3305 that individual transform properties do not
distinguish between values specified using 2D syntax and values
specified using 3D syntax that produce the same matrices.

Fixes the individual transforms part of #3305.

cc @BorisChiou (who isn't in the list of reviewers I can choose from) and also @smfr (who is welcome to review as well, and can let me know if he wants me to wait for his review)


BorisChiou (approved):
LGTM

PR #6153, opened 2021-03-26
[css-ui-4] Added image-set() function to 'cursor' property

### LINKED ISSUE #5831: [css-ui] Should the cursor property support image-set()?

SebastianZ:
This change adds the `image-set()` function to the `cursor` property.

It got [implemented in Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1695402) lately and WebKit has support for a prefixed version of it.

Together with the addition of the function I've also adjusted the syntax a little bit for better readability and fixed some links to CSS Backgrounds 3 because otherwise Bikeshed marks them as fatal errors and doesn't compile the spec.

Closes #5831

Sebastian

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -586,8 +586,9 @@ user resizing of that element.
 
 <pre class="propdef">
 Name: cursor
-Value: [ [ [ <<url>> | <<image-set()>> ] [&lt;x&gt; &lt;y&gt;]?,]* <br>
- [ auto | default | none |<br>
+Value: [ [ <<url>> | <<image-set()>> ] [ &lt;x&gt; &lt;y&gt; ]? , ]* <br>
+ [ <br>
+ auto | default | none | <br>
  context-menu | help | pointer | progress | wait | <br>
  cell | crosshair | text | vertical-text | <br>
  alias | copy | move | no-drop | not-allowed | grab | grabbing | <br>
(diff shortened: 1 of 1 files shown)


tabatkins:
This should just be swapping the `<<url>>` for `<<image>>`, right? There's no reason to limit it to solely `image-set()`, particularly since you can put any `<<image>>` you want *inside of* `image-set()`.

tabatkins:
(See #5834 for the earlier discussion on this.)

SebastianZ:
Based on [Mozilla's implementation issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1696314) I mentioned in the other PR and my tests in Firefox Nightly 88, `image-set()` is actually restricted to URLs for the `cursor` property, while it also takes gradients in other properties.

So I guess a different version of `image-set()` with restricted syntax needs to be defined for this case.

But let's see what Emilio has to say in the other PR.

Sebastian

frivoal:
This has been made redundant by https://github.com/w3c/csswg-drafts/commit/78fb03bd72db009decb0bdad200cd31f9dd9a235

PR #6162, opened 2021-03-30
[css-fonts-5] Don't allow more than 2 values for the font-size @font-face descriptor

litherum:
The prose says what to do if there are exactly 1 or exactly 2 values, but not what to do if there are more. `#` indicates that there can be more.

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -2244,21 +2244,21 @@ Font property descriptors: the 'font-style!!descriptor', 'font-weight!!descripto
 
 	<pre class='descdef'>
 	Name: font-style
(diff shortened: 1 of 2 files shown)


tabatkins:
These aren't the same grammar. `<<number>>#` is comma-separated repetitions, `<<number>> <<number>>?` isn't.

However, I suspect that "no commas" is what's wanted here, right?

svgeesus (changes requested):
Shouldn't this be [{1,2}](https://drafts.csswg.org/css-values-4/#component-multipliers) ?

litherum:
Right. The other descriptors don’t accept commas, so I don’t think this one should either.

svgeesus (approved):

PR #6167, opened 2021-03-31
[css-cascade-5] possible typo

himorin:
possible fixes for new in level 5. (also would like to ask this readings are correct)

(found during, but not related to an i18n review on css-cascase-4/5)

### DIFF
--- a/css-cascade-5/Overview.bs
+++ b/css-cascade-5/Overview.bs
@@ -1234,7 +1234,7 @@ Declaring Cascade Layers: the ''@layer'' rule</h4>
 	If the [=layer name=] matches that of a [=cascade layer=]
 	already defined within this [=origin=] and [=context=]
 	(and the same [=layer=] scope, see [[#nested-layers]]),
-	then its style rules are assigned to the same [=cascade layer=].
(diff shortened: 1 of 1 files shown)


css-cascade-5/Overview.bs
```diff
@@ -1233,8 +1233,8 @@ Declaring Cascade Layers: the ''@layer'' rule</h4>
 	that represents its <dfn>layer name</dfn>.
 	If the [=layer name=] matches that of a [=cascade layer=]
 	already defined within this [=origin=] and [=context=]
-	(and same [=layer=] scope, see [[#nested-layers]]),
-	then its style rules are assigned to that same [=cascade layer=].
+	(and the same [=layer=] scope, see [[#nested-layers]]),
+	then its style rules are assigned to the same [=cascade layer=].
```
tabatkins, line 1237: This "that" is intentional and makes it clear the referent is the preceding cascade layer we mentioned.

```suggestion
	then its style rules are assigned to that same [=cascade layer=].
```

tabatkins (changes requested):

PR #6172, opened 2021-04-01
[css-cascade-5] Begin to define nested layers as lexical issue #6098

### LINKED ISSUE #6098: [css-cascade-5][editorial] Define nested layers as a lexical thing
The current cascade layers spec starts out by explaining that each layer as a single-ident layer name, then that they can be nested, then adds on the ability to reference earlier nested named layers using a conjunctive notation. Probably makes more sense to rewrite this so that layer names consist of a joined list of idents, and nesting blocks joins the name of the parent layer to the child.

mirisuzanne:
Resolves #6098 

- [x] Move the dot-combined identifier option into the original layer-identifier syntax definition
- [x] Need to clarify that layer sorting is grouped by scope, so that `one.one, two, one.two` results in a layer order of `one.one, one.two, two`

### DIFF
--- a/css-cascade-5/Overview.bs
+++ b/css-cascade-5/Overview.bs
@@ -1231,20 +1231,80 @@ Declaring Cascade Layers: the ''@layer'' rule</h4>
 	The optional <dfn for='@layer'><<layer-ident>></dfn>
 	is a list of one or more [=CSS identifiers=] (<<ident>>),
 	joined directly by a full stop (. U+002E) character
-	without any intervening whitespace.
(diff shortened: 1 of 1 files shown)


fantasai:
Hey @mirisuzanne,
I'd had a half-baked patch for this as well, which wasn't good enough to put into a PR. :) Reviewing everything together, Tab and I thought neither one was really quite going in the right direction, so we worked on a new rewrite from scratch and just committed https://github.com/w3c/csswg-drafts/commit/dd1c3c46e64fce78f3106469863b46bd946e649a to fix #6098. We pulled some of your examples, though. So closing out this PR without merging, now that it's fixed.

Issue #6152, opened 2021-03-26
[css-fonts-5] vertical metrics overrides


fantasai:
`ascent-override` and `descent-override` should be defined to apply in the vertical axis only; which brings up the question, do we need to allow an optional second value to handle horizontal-axis overrides for vertical writing modes?

(A missing value should definitely be `normal`, not duplicate the first value though.)

fantasai:
Ah, same question also for the superscript/subscript offsets.

fantasai:
Fixed in https://github.com/w3c/csswg-drafts/commit/f68b5a314e7301fc08d1b96371b719a440660ba9

PR #6154, opened 2021-03-27
[css-fonts-5] Various Edits

fantasai:
Edits for the various metrics descriptors. Commit-by-commit review recommended. (It's not separate PRs because they build on each other. Please don't make me submit them one by one, I'll end up tangling all the branches in git...)

PR #6183, opened 2021-04-06
[css-fonts-5] vertical metrics overrides

litherum:
Fixes https://github.com/w3c/csswg-drafts/issues/6152.

### DIFF
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -149,21 +149,21 @@ the 'ascent-override', 'descent-override', 'line-gap-override', and 'size-adjust
 
 <pre class='descdef'>
 Name: ascent-override
-Value: normal | <<percentage>>
+Value: [ normal | <<percentage>> ]{1,2}
 For: @font-face
 Initial: normal
 </pre>
 
 <pre class='descdef'>
 Name: descent-override
-Value: normal | <<percentage>>
+Value: [ normal | <<percentage>> ]{1,2}
 For: @font-face
 Initial: normal
 </pre>
 
 <pre class='descdef'>
 Name: line-gap-override
-Value: normal | <<percentage>>
+Value: [ normal | <<percentage>> ]{1,2}
 For: @font-face
 Initial: normal
 </pre>
@@ -179,6 +179,11 @@ The 'ascent-override', 'descent-override', and 'line-gap-override' descriptors d
 <a spec="CSS-INLINE-3">ascent metric</a>, <a spec="CSS-INLINE-3">descent metric</a> and
 <a spec="CSS-INLINE-3">line gap metric</a> of the font, respectively. The `size-adjust`
 descriptor defines a scalar by which glyph outlines and metrics from this font are multiplied.
+
+Both values accepted by 'ascent-override', 'descent-override', and 'line-gap-override' measure
+distance in the block direction. When the text is <a href="#typeset-sideways">typeset sideways</a>,
+the first value is used. When the text is <a href="#typeset-upright">typeset upright</a>, the
+second value is used if it is present, and if it isn't present, then the first value is used.
 
 When the descriptor value is 'normal', the corresponding metric value is obtained as-if the
 descriptor was absent from the <code>@font-face</code> block.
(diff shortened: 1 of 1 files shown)


css-fonts-5/Overview.bs
```diff
@@ -149,21 +149,21 @@ the 'ascent-override', 'descent-override', 'line-gap-override', and 'size-adjust
 
 <pre class='descdef'>
 Name: ascent-override
-Value: normal | <<percentage>>
+Value: normal | <<percentage>>{1,2}
```
jfkthame, line 152: Does the `{1,2}` here apply to the entire alternation `normal | <<percentage>>`, or only to the `<<percentage>>` (which is what it looks like to my naïve eye, at least)?

If the latter, I think that's not what we want, as it wouldn't allow things like `ascent-override: normal 75%`.

css-fonts-5/Overview.bs
```diff
@@ -180,6 +180,15 @@ The 'ascent-override', 'descent-override', and 'line-gap-override' descriptors d
 <a spec="CSS-INLINE-3">line gap metric</a> of the font, respectively. The `size-adjust`
 descriptor defines a scalar by which glyph outlines and metrics from this font are multiplied.
 
+The first value to 'ascent-override', 'descent-override', or 'line-gap-override' is used for
+vertical distance with upright glyph orientation (e.g. <a>horizontal writing modes</a> or
+<a>vertical writing modes</a> with <code>text-orientation: upright</code>) and horizontal
```
jfkthame, line 185: In vertical writing modes with `text-orientation: upright`, should the first value be used at all? I guess I assumed that in this case what we refer to as "ascent" etc are the font's vertical-layout metrics, and the override for these is the second value, so the first is simply not relevant.

litherum: Well, with `text-orientation: upright` in vertical writing modes, the glyphs have to be placed vertically somehow... So both values would end up being used. One to place the glyphs one-after-another, and the other to determine horizontal size that affects the line-to-line distance.

jfkthame: Placing the vertical/upright glyphs one after another should normally be based on glyph advances, not on font-wide metrics. It's true that we need a fallback behavior for cases where the font has no vertical metrics (is that explicitly spec'd? need to check...), but I'm not sure the same overrides as are used for normal line-height in horizontal mode should also be used as a form of advance-override in vertical/upright.

litherum: I think you're right that a better way of saying this would be "these metrics measure distance in the block direction". We'd use vertical advances in the inline direction.

css-fonts-5/Overview.bs
```diff
@@ -180,6 +180,11 @@ The 'ascent-override', 'descent-override', and 'line-gap-override' descriptors d
 <a spec="CSS-INLINE-3">line gap metric</a> of the font, respectively. The `size-adjust`
 descriptor defines a scalar by which glyph outlines and metrics from this font are multiplied.
 
+Both values accepted by 'ascent-override', 'descent-override', and 'line-gap-override' measure
+distance in the block direction. When the text is <a href="#typeset-sideways">typeset sideways</a>,
```
litherum, line 184: Aw, crap, this isn't right either. I have to add in a clause "typeset sideways or in horizontal writing mode"

fantasai:
@litherum, can you hang on a minute while I rebase https://github.com/w3c/csswg-drafts/pull/6154 ?

fantasai (changes requested):
I think you want the definition to be overriding the X vs. Y metrics (fundamentally), not as referencing the block/inline axes. And then describing the impact of that to be affecting the block axis only, with the second value only being used in upright vertical typesetting modes. (The actual definition of which metric gets used should be embedded into the part of the spec that uses the metric, rather than in the definition of how it gets overridden.)

Also, you need to be *very clear* that the second value defaults to `normal`, not to duplicating the first value.

fantasai:
Filed a pull request with updated wording as described in https://github.com/w3c/csswg-drafts/pull/6183#pullrequestreview-629402801 at https://github.com/w3c/csswg-drafts/pull/6381

PR #6188, opened 2021-04-07
[css-grid-1][css-grid-2] Remove `grid-row: auto`

SiarheiBobryk:
Although we use `grid-row: auto` in example 24 of the "Automatic
Placement" chapter, I believe it's redundant and we free to remove that.

I'm not sure about the situation regarding this property in the past,
but according to my investigation nowadays we already have implicit
`grid-row: auto` for all grid children for all major browsers: the
Chromium-family, FF, Safari, Edge (legacy)

_Chrome_
chrome

_Firefox_
firefox

_Safari_
safari-1
safari-2

_Edge (legacy)_
edge-1
edge-2

### DIFF
--- a/css-grid-1/Overview.bs
+++ b/css-grid-1/Overview.bs
@@ -2503,13 +2503,11 @@ Implicit Track Sizing: the 'grid-auto-rows' and 'grid-auto-columns' properties</
 				/* Place all labels in the "labels" column and
 				   automatically find the next available row. */
 				grid-column: labels;
-				grid-row: auto;
 			}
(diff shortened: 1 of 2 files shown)


tabatkins:
While it is redundant (the `auto` is the initial value), it's not wrong to do so, and I think it illustrates good practice to give row and column together like in the example. (Without it, it might imply that the row value is coming from another rule, which is not the intent here.) I'd thus prefer to keep the example as-is.

SiarheiBobryk:
Makes sense @tabatkins

PR #6198, opened 2021-04-09
[css-highlight-api-1] Make HighlightRegistry maplike

### LINKED ISSUE #5910: [css-highlight-api] maplike vs setlike
from https://github.com/w3ctag/design-reviews/issues/584#issuecomment-767655588

> @hober and I discussed this today in our VF2F. We noticed that HighlightsRegister is a setlike, but throws when a Highlight object with an existing registered name, which indicates that perhaps maplike would have been more appropriate? It's not ideal developer experience if developers need to iterate the entire data structure before registering their object, to check if an existing one of the same name exists. Maplike would also provide an easy way to get all ranges under a given name.

sanketj:
[css-highlight-api-1] Make HighlightRegistry maplike and remove name property from Highlight.
[css-highlight-api-1] Add note recommending authors to use custom highlight names that are CSS identifiers for easier styling via CSS.
[css-highlight-api-1] Add note recommending authors to only use one name to style a highlight and add an example showcasing why using multiple names is problematic.

Resolves #5910

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -229,6 +229,10 @@ Registering Custom Highlights</h3>
 		The <dfn>custom highlight name</dfn> assigned to a [=custom highlight=] when it is [=registered=]
 		is used to identify the highlight during styling (see [[#styling-highlights]]).
 
+	Note: When registering a [=custom highlight=], authors are recommended to use a
+	[=custom highlight name=] that is a valid CSS [=identifier=]. Using a name that is not a valid
+	identifier can make the highlight hard, and in some cases impossible, to style via CSS.
+
 <h2 id=styling-highlights>
 Styling Custom Highlights</h2>
 
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -103,7 +103,7 @@ Introduction</h2>
 			r.setStart(document.body, 0);
 			r.setEnd(document.body, 2);
 
-			CSS.highlights.add(new Highlight("example-highlight", r));
+			CSS.highlights.set("example-highlight", new Highlight(r));
 			</script>
 		</xmp>
 
@@ -139,7 +139,7 @@ Setting up Custom Highlights</h2>
 <h3 id=creation>
 Creating Custom Highlights</h3>
 
-	A <dfn>custom highlight</dfn> is a named collection of [=ranges=]
+	A <dfn>custom highlight</dfn> is a collection of [=ranges=]
 	representing portions of a document.
 	They do not necessarily fit into the element tree,
 	and can arbitrarily cross element boundaries without honoring its nesting structure.
@@ -160,35 +160,24 @@ Creating Custom Highlights</h3>
 	authors can chose between using {{Range}} objects and {{StaticRange}} objects.
 	See [[#range-invalidation]] for more details about this choice and its implications.
 
-	The <dfn for="custom highlight">name</dfn> of a [=custom highlight=]
-	is represented by its {{Highlight/name}} attribute,
-	which must be a valid <<ident-token>>.
-
 	<xmp class="idl">
 	[Exposed=Window]
 	interface Highlight {
-		constructor(CSSOMString name, AbstractRange... initialRanges);
+		constructor(AbstractRange... initialRanges);
 		setlike<AbstractRange>;
 		attribute long priority;
-		readonly attribute CSSOMString name;
 	};
 	</xmp>
 
 	See [[#priorities]] for more information on the {{Highlight/priority}} attribute.
 
 	<div algorithm="to create a custom highlight">
-		When the <dfn for=Highlight constructor>Highlight(CSSOMString name, AbstractRange... initialRanges)</dfn> constructor is invoked,
+		When the <dfn for=Highlight constructor>Highlight(AbstractRange... initialRanges)</dfn> constructor is invoked,
 		run the following steps:
 
 		<ol>
 			<li>
 				Let |highlight| be the new {{Highlight}} object.
-			<li>
-				If {{name!!argument}} does not [=CSS/parse=] as an <<ident-token>>, then [=throw=] a {{"SyntaxError"}}.
-			<li>
-				Let |nameArg| be the result of [=converted to an ECMAScript value|converting=] {{name!!argument}} to an ECMAScript value.
-			<li>
-				Set |highlight|'s {{Highlight/name}} to |nameArg|
 			<li>
 				Set |highlight|'s {{Highlight/priority}} to <code>0</code>.
 			<li>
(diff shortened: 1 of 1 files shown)


css-highlight-api-1/Overview.bs
```diff
@@ -225,41 +214,45 @@ Registering Custom Highlights</h3>
...
+		with the [=context object=] as the <code>this</code> value,
+		the passed-in [=custom highlight name=] as |keyArg|,
+		and the passed-in highlight as |valueArg|.
+
+		The <dfn>custom highlight name</dfn> assigned to a [=custom highlight=] when it is [=registered=]
+		is used to identify the highlight during styling (see [[#styling-highlights]]).
+
+	Note: It is possible to [=register=] a [=custom highlight=] with any [=custom highlight name=]
+	that is a valid [[webidl#idl-DOMString|DOMString]], but the name must also be a valid CSS
+	<<ident-token>> in order to be used to style the highlight.
```
LeaVerou, line 234: This seems like a footgun, why not restrict these to valid `<ident-token>`s so that the error can be caught at registration time? Is there any other purpose for these highlights other than being styled in CSS?

css-highlight-api-1/Overview.bs
```diff
@@ -225,41 +214,45 @@ Registering Custom Highlights</h3>
...
+
+	Note: It is possible to [=register=] a [=custom highlight=] with more than one [=custom highlight name=].
+	However, using more than one name can lead to the highlight being assigned multiple different sets of styles,
+	without a way to control the stacking order of these styles during [[#painting|painting]].
+	This may be limiting for authors and may cause confusing painting behavior.
+	Therefore, <b>it is recommended that authors only use one name per highlight</b>.
```
LeaVerou, line 240: I'm not sure if we should be recommending against the practice altogether, or just recommending against using **both** names to style the highlight, which is what actually causes trouble. If only one of the names is ever used in the stylesheet, there is no confusion. As I discussed in the thread, I can see valid use cases for registering multiple highlights with the same name by library authors to avoid manually keeping the objects in sync, with the expectation of their authors only actually styling one of these names.
I also think it would be good to include your minimized example code where the cascade doesn't work as expected, as that was quite illuminating in understanding the problem.

css-highlight-api-1/Overview.bs
tabatkins: More importantly, this seems incorrect; you can *write* an <<ident-token>> whose value is *any* arbitrary string, it just might require appropriate use of escapes.  So this Note should be rewritten to instead be a *recommendation* for authors to provide names that are CSS idents, so that it's easier to use them in CSS, rather than implying they can't be referred to at all from CSS.

sanketj: @LeaVerou, @tabatkins: What I had in mind was that this should work similar to id on elements. For example, authors can assign ids to elements that are not ident tokens - ex. `div.setAttribute('id', '+')` - but authors can still target such elements from CSS using escapes - ex. `#\+ { color: red; }`. Similarly, the highlight OM APIs do not need to prevent authors from registering non-ident token names, and the following example should work (ie. 'b' should be red) :

```
<html>
<head>
    <style>
      div::highlight(\+) {
        color: red;
      }
    </style>
</head>
<body>
    <div>abd</div>
    <script>
      let div = document.querySelector('div');
      let textNode = div.firstChild;
      let range = new StaticRange({startContainer: textNode, startOffset: 1, endContainer: textNode, endOffset: 2});
      let highlight = new Highlight(range);
      CSS.highlights.set('+', highlight);
    </script>
</body>
</html>
```
Do you agree with this model, or would you like to see something different?

The rationale behind this note was that I wasn't sure if all character values accepted by DOMString could be represented as an ident token. I couldn't come up with an example of such a case, but it also wasn't clear to me from reading the CSS identifier and DOMString specs that there wasn't a case.
Thinking about this some more though, if there were a case like this then it would affect id and other similar selectors in the same way, and I don't see any note about this in the CSS syntax spec. I also don't see anything about this in the CSS.escape spec either. So I'm starting to think that there really may not be such a case (@tabatkins, that's what you were saying too, right?). If so, I think we can just remove the note.

@LeaVerou, just to respond to your second question as well, I assume you were referring to "highlight names" when you said "highlights"? It's true that highlight names are only used for styling via CSS, but I think it's still better to match the prior art we have here. (The highlight objects themselves are also used for setting priority, and they will also likely become event targets when we work on highlight events in the future.)

css-highlight-api-1/Overview.bs
sanketj: Okay, that works for me. I will update this to reflect "style with one name" and add the example.

css-highlight-api-1/Overview.bs
tabatkins: Since you can always write an ident *solely* as a list of escapes (and you can encode any codepoint in escapes), it's always possible, if sometimes unwieldy, to encode *almost any* DOMString as a CSS ident. (The only exception is that idents can't contain U+0000 NULL, or any surrogate code points. But if you're putting that in your string you've placed your own foot entirely in the barrel of the gun, and it's not our fault you get shot.)

The general policy, from other CSS APIs where we compare or set a CSS ident with a JS string, is to just use the string directly, with no CSS parsing done on it. See <https://drafts.csswg.org/css-conditional/#dom-css-supports>, for example.

So yeah, let authors do what they want here, don't try to limit them. Leave in a note that *warns against* making strings that aren't CSS identifiers (linking `[=CSS/identifier=]`), but no more.

sanketj: Thanks for the information @tabatkins . Makes sense to leave a warning here then. I've made that update. PTAL.

css-highlight-api-1/Overview.bs
sanketj: I have updated this. PTAL.

css-highlight-api-1/Overview.bs
```diff
@@ -225,25 +214,60 @@ Registering Custom Highlights</h3>
...
+	Note: It is possible to [=register=] a [=custom highlight=] with more than one [=custom highlight name=].
+	However, using more than one name to style a highlight will assign the highlight multiple different sets
+	of styles, without a way to control the stacking order of conflicting styles within these sets
+	during [[#painting|painting]]. This may be limiting for authors and may cause confusing painting behavior
+	(see the [[#styling-problems-with-multiple-names-per-highlight|example]] below for more context). Therefore,
+	<b>it is recommended that authors only use one name per highlight during styling</b>.
+	<div class=example id=styling-problems-with-multiple-names-per-highlight>
+		<xmp highlight=html>
+			<style>
+			div::highlight(foo) {
```
LeaVerou, line 245: Do we need this first rule? Don't the last two rules demonstrate the problem even without this?

LeaVerou (approved):
LGTM. See the one minor comment I left.

css-highlight-api-1/Overview.bs
sanketj: Yes, they do. Thanks! Removing the redundant rule.

css-highlight-api-1/Overview.bs
sanketj: @tabatkins Not sure if you wanted to look at this again or not. I'm going to go ahead and merge this. Feel free to let me know if you had any additional feedback here, I can address it with a follow up change.

PR #6219, opened 2021-04-16
[selectors-4] Define a number of new pseudo-classes for matching the state of media elements.

### LINKED ISSUE #3821: [selectors-4] additional resource state pseudo-classes for media elements
In [§11 Resource State Pseudos](https://drafts.csswg.org/selectors/#resource-pseudos) we currently have the `:playing` and `:paused` pseudo-classes, which collectively address the use case of a custom "play/pause" media control which should appear as a play button or as a pause button depending on the current play state of the associated media element. A bunch of us (@beccahughes @mounirlamouri @padenot @jyavenard @eric-carlson @jernoble et al.) would like to address several related use cases with additional resource state pseudo-classes:

* Whether or not the media element is [**muted**](https://html.spec.whatwg.org/multipage/media.html#concept-media-muted). Use case: styling a custom muted control. Proposal: mint a `:muted` pseudo-class.
* Whether or not the media element is currently **stalled** (see several paragraphs in the HTML spec starting at ["The **stall timeout** is a"](https://html.spec.whatwg.org/multipage/media.html#stall-timeout)). Use case: changing the appearance of a loading progress indicator to indicate that loading is currently stalled. Proposal: mint a `:stalled` pseudo-class.
* Whether or not the media element is currently [**seeking**](https://html.spec.whatwg.org/multipage/media.html#seeking). Use case: a custom seek control can reflect the case where a user is seeking using some other, UA-provided control. Proposal: mint a `:seeking` pseudo-class.

Currently, authors of custom media controls have to do some combination of UA-sniffing and other logic from script in order to handle these cases. Ideally, the appearance of such controls would be expressible in CSS, just as play state currently is.

### LINKED ISSUE #3933: [selectors-4] "effective media volume is mutable" pseudo-class for media elements
Split out from #3821.

In [§11 Resource State Pseudos](https://drafts.csswg.org/selectors/#resource-pseudos) we currently have the `:playing` and `:paused` pseudo-classes, which collectively address the use case of a custom "play/pause" media control which should appear as a play button or as a pause button depending on the current play state of the associated media element. In addition to the cases covered in #3821, a bunch of us (@beccahughes @mounirlamouri @padenot @jyavenard @eric-carlson @jernoble et al.) would like to mint a new pseudo-class to expose whether or not setting [the **volume** IDL attribute](https://html.spec.whatwg.org/multipage/media.html#dom-media-volume) would result in the element's [**effective media volume**](https://html.spec.whatwg.org/multipage/media.html#effective-media-volume) actually changing.

With such a pseudo-class, it would be be possible (without script) to show or hide custom volume controls based on whether or not using such a control would actually result in a user-audible change.

I have no specific proposal for a name. Bikeshedding very welcome. @AmeliaBR has [several suggestions](https://github.com/w3c/csswg-drafts/issues/3821#issuecomment-491097390).

hober:
We resolved to add these [in Toronto in June 2019](https://lists.w3.org/Archives/Public/www-style/2019Jul/0002.html).

This is a first draft; I'm sure the text could use some work before merging.

Fixes #3821.
Fixes #3933.

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -35,14 +35,11 @@ spec:css-pseudo-4; type:selector;
 	text: ::after
 	text: ::first-line
 	text: ::first-letter
-</pre>
-<pre class="anchors">
-urlPrefix: https://html.spec.whatwg.org/multipage/; spec: HTML
-    text: effective media volume; url: media.html#effective-media-volume; type: dfn
-    text: media data; url: media.html#media-data; type: dfn
-    text: seeking; url: media.html#seeking; type: dfn
-    text: stall timeout; url: media.html#stall-timeout; type: dfn
-    text: stalled; url: media.html#stall-timeout; type: dfn
+spec:HTML; type:dfn;
+	text: effective media volume
+	text: seeking
+	text: media data
+	text: stall timeout
 </pre>
 
 <style>
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -36,6 +36,14 @@ spec:css-pseudo-4; type:selector;
 	text: ::first-line
 	text: ::first-letter
 </pre>
+<pre class="anchors">
+urlPrefix: https://html.spec.whatwg.org/multipage/; spec: HTML
+    text: effective media volume; url: media.html#effective-media-volume; type: dfn
+    text: media data; url: media.html#media-data; type: dfn
+    text: seeking; url: media.html#seeking; type: dfn
+    text: stall timeout; url: media.html#stall-timeout; type: dfn
+    text: stalled; url: media.html#stall-timeout; type: dfn
+</pre>
 
 <style>
 #selector-examples td:first-child {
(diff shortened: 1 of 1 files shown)


hober:
cc @jernoble

selectors-4/Overview.bs
```diff
@@ -36,6 +36,14 @@ spec:css-pseudo-4; type:selector;
 	text: ::first-line
 	text: ::first-letter
 </pre>
+<pre class="anchors">
+urlPrefix: https://html.spec.whatwg.org/multipage/; spec: HTML
```
tabatkins, line 40: All of these definitions already exist in HTML and are reasonably marked up, they're just not exported.

Please swap to specifying these in link-defaults, and then probably ask HTML to export them.

selectors-4/Overview.bs
```diff
@@ -2403,23 +2411,63 @@ Resource State Pseudos</h2>
 	particularly images/videos,
 	and allow authors to select them based on some quality of their state.
 
-<h3 id="video-state">
-Video/Audio Play State: the '':playing'' and '':paused'' pseudo-classes</h3>
+<h3 id="video-state"> <!-- keeping this ID because Cool URIs Don't Change -->
```
tabatkins, line 2414: Good point, but no need to explain your ID choices. ^_^

selectors-4/Overview.bs
```diff
@@ -2403,23 +2411,63 @@ Resource State Pseudos</h2>
...
 
-	The <dfn>:paused</dfn> pseudo-class represents the same elements,
-	but instead match when the element is <em>not</em> “playing”.
+	The <dfn>:paused</dfn> pseudo-class represents an element
+	that is capable of being “played” or “paused”,
+	when that element is “paused” (i.e. <em>not</em> ”playing”).
 	(This includes both an explicit “paused” state,
 	and other non-playing states like “loaded, hasn't been activated yet”, etc.)
 
+    The <dfn>:seeking</dfn> pseudo-class represents an element
```
tabatkins, line 2431: This spec is tab-indented, please stick with that consistently.

tabatkins (changes requested):
r+ after changes

selectors-4/Overview.bs
hober: Fixed in 59f2bc1. Export of those terms requested in whatwg/html#6603.

selectors-4/Overview.bs
hober: Removed in 6599d0b.

selectors-4/Overview.bs
hober: Fixed in 6599d0b.

tabatkins:
Overall looks good.

Only comment left is that maybe we also want a generic `:media` pseudo-class that matches elements capable of matching any of these pseudos? I suggest this because of the annoyingness of things like `:not(:muted)` causing overselection of everything else on the page; instead, you could write `:media:not(:muted)` and get what you meant.

(I could see us instead recasting the whole thing into a `:media()` that accepts keywords for all the states, but we'd have to add negation to the states as well, and we already have `:playing` and `:paused`, so meh.)

tabatkins (approved):

tabatkins:
r+ for merging what's currently here; we can either address the `:media` question separately or fold it into this, whichever you prefer

hober:
> r+ for merging what's currently here; we can either address the `:media` question separately or fold it into this, whichever you prefer

Let's raise it as a separate / followup issue, since we don't have a WG resolution for it. (Not that I think it'll be hard to get one; you make a compelling argument.) Wanna file the followup issue?

PR #6234, opened 2021-04-22
[mediaqueries-5] Add viewport segment media features

### LINKED ISSUE #5621: [mediaqueries-5] Updating the CSS media feature syntax for foldable & dual-screen devices
My colleague @dlibby- and I had originally [proposed spanning media feature](https://github.com/w3c/csswg-drafts/issues/4736) with a set of enumerated values representing the layout viewport’s relationship to display hardware configuration (`single-fold-vertical`, `single-fold-horizontal`). The original thinking was that screen topologies of new hardware could be expressed by create more enumeration values.

Picture1

The discussion on the previous issue, which provided a few suggestions. Based on those, we’d like the following (based on @fantasai 's proposal) to be considered as alternatives to the previous proposed media feature.

```
display-span-x: <integer[1,inf]>
display-span-y: <integer[1,inf]>
```
The values represent the number of logical displays the viewport spans in the specified direction. A logical display may represent a physical screens, or a more abstract concept like a display region for foldable devices that physically have a single screen. This can be extended to express the intent of the operating system window manager and be used in desktop scenarios as well.

These features are described based on their physical orientation relative to the upper left corner of the layout viewport in order to simplify authors’ reasoning about them, similar to other media features.

we'd like to gather consensus to add these to [mediaqueries-5] to allow Web developers to enhance content for dual-screen and foldable devices, while providing an syntax that can extend to cover multi-screen scenarios.

### Example use-case:

image007

image001-5f5fae6619385

dlibby-:
Add two new media feature that allow for authors to detect situations where the viewport is logically separated based on the window's configuration with respect to display features. This is designed to allow authors to target foldable devices where the viewport spans multiple displays (logical or physical). 

There are features for both the horizontal and vertical directions and are integers so that future form factors (think a tri-fold device) that do not yet exist can be explicitly detected and supported by authors.

 Fixes #5621

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1252,6 +1252,46 @@ Inline-Axis Overflow: the '@media/overflow-inline' feature</h3>
 	and the very concept doesn't seem to make much sense,
 	so there is intentionally no ''paged'' value for '@media/overflow-inline'.
 
+<h3 id='mf-horizontal-viewport-segments'>
+Horizontal Viewport Segments: the '@media/horizontal-viewport-segments' feature</h3>
+
+	<pre class="descdef mq">
+	Name: horizontal-viewport-segments
+	Value: <<integer>>
+	For: @media
+	Type: range
+	</pre>
+
+	The '@media/horizontal-viewport-segments' media feature describes the number of logical segments of
+    the viewport in the horizontal direction. 
+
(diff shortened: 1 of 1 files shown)


dlibby-:
Friendly ping @frivoal @tabatkins

dlibby-:
I'm hoping to make progress on these features over this quarter, so will probably merge this shortly unless there are any objections. The additions are relatively small, so I'm comfortable hashing out feedback via follow up issues.

PR #6239, opened 2021-04-22
[css-ui-4] Define the ''input-security'' property. Fixes #2495.

### LINKED ISSUE #2495: [css-ui] Spec input-security property
An increasing number of sites have switched their login forms from using `<input type=password>` to `<input type=text>` for password input, in order to provide the ability to toggle display of the password text. This has the unfortunate side effect of preventing client-side password/account managers from correctly filling the form. We should spec the ability to toggle display of the contents of `<input type=password>` to help sites avoid breaking such account managers.

WebKit implements `<input type=password>` styling with a `-webkit-text-security` property, for which there are several polyfills out there for other browsers (e.g. noppa/text-security), so perhaps it's the right cowpath to pave here.

It's worth noting that it's sometimes desirable to obscure the text of other inputs—`<input type=tel>`, for instance.

hober:
A [long time ago](https://github.com/w3c/csswg-drafts/issues/2495#issuecomment-380397331) we resolved to add an `input-security : auto | none` property that only applies to `<input type=password>`. Here's a PR that does this, though not very well.

I'll write a corresponding PR for HTML's Rendering section too.

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -101,6 +101,15 @@ urlPrefix: https://html.spec.whatwg.org/multipage/; spec:HTML
 <style>
 .awesome-table td {padding:5px}
 .awesome-table {color:#000;background:#fff;margin: auto;}
+
+.fake-input-type-password {
+	display: inline-block;
+	border: 1px solid black;
+	background: white;
+	text: black;
+	width: 20ch;
+	padding: 0 2px;
+}
 </style>
 
 <h2 id="intro">Introduction</h2>
@@ -2210,7 +2219,7 @@ Obscuring sensitive input: the 'input-security' property</h3>
 Name: input-security
 Value: auto | none
 Initial: ''input-security/auto''
-Applies to: elements that accept sensitive text input, such as <{input/type/password|&lt;input type=password&gt;}>
+Applies to: [=sensitive text inputs=]
 Inherited: no
 Percentages: N/A
 Computed Value: as specified
@@ -2218,12 +2227,16 @@ Canonical order: per grammar
 Animation type: by computed value type
 </pre>
 
-When the purpose of a text input control is to accept sensitive input,
-such as <{input/type/password|&lt;input type=password&gt;}> in HTML,
-user agents obscure the display of the control's content by default
+For the purpose of this specification,
+a <dfn>sensitive text input</dfn> is
+a <a spec=html>mutable</a> form control with textual content whose
+purpose is to accept sensitive input,
+such as <{input/type/password|&lt;input type=password&gt;}>.
+
+By default, user agents obscure the contents of [=sensitive text inputs=]
 in order to prevent onlookers from seeing it.
 Users may wish to temporarily disable this obscuring
-in order to confirm that they've typed their password correctly.
+in order to confirm that they've typed their sensitive information correctly.
 The ''input-security'' property may be used by authors
 to enable or disable this obscuring.
 
@@ -2238,7 +2251,21 @@ to enable or disable this obscuring.
 		so that it cannot be read by the user.
 </dl>
 
-The exact mechanism by which user agents obscure the text in the control is undefined.
+While the exact mechanism by which user agents obscure the text in the control is undefined, user agents typically obscure [=sensitive text inputs=] by replacing each character with some suitable replacement such as U+002A ASTERISK (*) or U+25CF BLACK CIRCLE (●).
+
+<div class="example">
+	For instance, given this style sheet
+	<pre><code class="lang-css">
+	input[type=password] {
+		input-security: auto;
+	}</code></pre>
+	and this HTML
+	<pre><code class="lang-html">
+	&lt;input type=password value=MySecret794>
+	</code></pre>
+	a user agent might render the <{input/type/password|&lt;input type=password&gt;}> like so:
+	<span class=fake-input-type-password>●●●●●●●●●●●</span>
+</div>
 
 <h3 id=control-specific-rules>
 Control Specific Rules</h3>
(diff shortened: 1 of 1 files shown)


css-ui-4/Overview.bs
```diff
@@ -93,6 +93,10 @@ spec:css-color-4; type:property; text:color
 spec:selectors-4; type:selector; text::focus
 spec:selectors-4; type:selector; text::focus-visible
 </pre>
+<pre class=anchors>
+urlPrefix: https://html.spec.whatwg.org/multipage/; spec:HTML
+	text:password; type:attr-value; for:input/type; url: input.html#attr-input-type-password-keyword
+</pre>
```
tabatkins, line 99: I understand that the HTML markup is bad here and you *can't* reasonably use link-defaults, so this is fine for now. When you do the HTML PR, could you submit a fixup for these dfns as well?

css-ui-4/Overview.bs
```diff
@@ -2199,6 +2203,42 @@ that are typical of the element whose appearance it acquires.
...
+	<dd>
+		The UA must not obscure the text in the control,
+		so that it can be read by the user.
+	<dt><dfn>auto</dfn>
+	<dd>
+		The UA should obscure the text in the control,
+		so that it cannot be read by the user.
+</dl>
+
+The exact mechanism by which user agents obscure the text in the control is undefined.
```
tabatkins, line 2240: Keeping it undefined is correct, but this could use an example showing off the dot-replacement used by current UAs.

css-ui-4/Overview.bs
```diff
@@ -2199,6 +2203,42 @@ that are typical of the element whose appearance it acquires.
 	is also not changed by the 'appearance' property.
 </div>
 
+<h3 id="input-security">Obscuring sensitive input: the 'input-security' property</h3>
+
+<pre class=propdef>
+Name: input-security
+Value: auto | none
+Initial: ''input-security/auto''
+Applies to: elements that accept sensitive text input, such as <{input/type/password|&lt;input type=password&gt;}>
```
tabatkins, line 2212: This makes me vaguely uncomfortable. Is it indeed intended that this only works on password inputs, not on text inputs? If so, we should probably be clearer that this is host-language defined. Probably just say "Applies to: [=sensitive-text inputs=]" and define the term in the text as being host-language defined, and either specifying that in HTML only password inputs are sensitive-text inputs or linking to the spot you'll add in HTML that says the same.

css-ui-4/Overview.bs
```diff
@@ -2199,6 +2203,42 @@ that are typical of the element whose appearance it acquires.
...
+Inherited: no
+Percentages: N/A
+Computed Value: the keword ''input-security/auto'' or the keyword ''input-security/none''
```
tabatkins, line 2215: just use "as specified"

css-ui-4/Overview.bs
```diff
@@ -2199,6 +2203,42 @@ that are typical of the element whose appearance it acquires.
 	is also not changed by the 'appearance' property.
 </div>
 
+<h3 id="input-security">Obscuring sensitive input: the 'input-security' property</h3>
```
tabatkins, line 2206: Use the same linebreaking as the rest of the spec, putting the text content on a newline.

tabatkins (changes requested):

css-ui-4/Overview.bs
hober: Will do!

css-ui-4/Overview.bs
hober: > This makes me vaguely uncomfortable. Is it indeed intended that this only works on password inputs, not on text inputs?

That was the resolution. I'm not super enthused about this either—as I said in the original issue, it's sometimes desirable to obscure the text of other inputs, such as `<input type=tel>`. But I figure that's a discussion we can have in a followup issue.

> If so, we should probably be clearer that this is host-language defined. Probably just say "Applies to: [=sensitive-text inputs=]" and define the term in the text as being host-language defined, and either specifying that in HTML only password inputs are sensitive-text inputs or linking to the spot you'll add in HTML that says the same.

Okay, I'll do something like this.

css-ui-4/Overview.bs
hober: Fixed in 1dd4278.

css-ui-4/Overview.bs
hober: Fixed in 1dd4278.

css-ui-4/Overview.bs
hober: I've given this a go in 3f5738a. Please let me know what you think, @tabatkins!

css-ui-4/Overview.bs
hober: I've provided an example in 3f5738a.

css-ui-4/Overview.bs
```diff
@@ -2199,6 +2212,61 @@ that are typical of the element whose appearance it acquires.
...
+Computed Value: as specified
+Canonical order: per grammar
+Animation type: by computed value type
+</pre>
+
+For the purpose of this specification,
+a <dfn>sensitive text input</dfn> is
+a <a spec=html>mutable</a> form control with textual content whose
+purpose is to accept sensitive input,
+such as <{input/type/password|&lt;input type=password&gt;}>.
```
tabatkins, line 2234:
```suggestion
a <dfn export>sensitive text input</dfn> is
a text input whose purpose is to accept sensitive input,
as defined by the host language.
For example, in HTML <{input/type/password|&lt;input type=password&gt;}> is a [=sensitive text input=].
```

tabatkins: (You don't want to say "mutable", readonly inputs aren't mutable but they still hide their text. And generally we just want to be clear this is host-defined.)

tabatkins (changes requested):

tabatkins:
r+ from me after the adjustment, tho I'm not an editor so @frivoal or @tantek will have to give the final nod

PR #6267, opened 2021-05-05
[web-animations-2] Fix WebIDL - no inherit for partial dictionary

tidoust:
Per WebIDL, a partial dictionary declaration cannot inherit from a base dictionary. See Note in:
https://heycam.github.io/webidl/#idl-interfaces

"Note: A partial interface definition cannot specify that the interface inherits from another interface. Inheritance must be specified on the original interface definition."

`ComputedEffectTiming`, defined in Level 1, already inherits from `EffectTiming`, so no need to formulate that again in the partial definition.


birtles:
Thank you!

PR #6271, opened 2021-05-06
[web-animations-2] Make WebIDL valid

tidoust:
Previous fix made WebIDL parse-able: https://github.com/w3c/csswg-drafts/pull/6267

However, the WebIDL remains invalid because `void` has now been replaced by `undefined` in WebIDL and because optional dictionary parameters with no required members need to have a default value. This update fixes these validity issues.


birtles:
Thank you once again!

PR #6276, opened 2021-05-07
[css-nesting-1] demo the result of BEM style nesting

### LINKED ISSUE #3748: [css-nesting] add note explaining that BEM `&__element` class accumulation isn't supported and why
CSS nesting spec
https://drafts.csswg.org/css-nesting-1/

There is a very common method of writing BEM classes among FE developers that looks like this when writing in SCSS:

```scss
.block {
  &__element {
    // .block__element { ... }
  }
}
```

The spec currently doesn't seem to explicitly allow for this style of nesting. Providing an example to mention this method is ok is all that would be needed.

**Update:**

Apparently that syntax isn't actually allowed. 😢

The spec doesn't clearly explain if the syntax is valid or not though.

Please add a clear example stating that this isn't valid and explain why.

argyleink:
[css-nesting-1] fixes #3748

Felt sufficient to demonstrate the result of a BEM style nesting selector, as opposed to a note specifically calling out BEM. The answer is in the examples.

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -243,15 +243,6 @@ Direct Nesting {#direct}
 
 		.foo {
 			color: blue;
-			&__bar { padding: 2ch; }
-		}
-		/* equivalent to
-			 .foo { color: blue; }
-			 __bar.foo { padding: 2ch; }
-		 */
-
-		.foo {
-			color: blue;
 			& { padding: 2ch; }
 		}
 		/* equivalent to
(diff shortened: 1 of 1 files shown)


tabatkins:
I think it's worth pulling this out into a separate example with an explicit call-out, actually. Don't have to mention BEM, but a general reference to "concatenating" that some preprocessors with this feature can do would be useful.

plinss:
@argyleink in the future, please don't create branches in the main repo for PRs. Fork the repo into your personal space and create branches there and PRs from there.

Having additional branches puts an extra workload on the draft server CI system and also can cause issues with the hg/git sync.

argyleink:
well, i'm still in the process of making a pr of this work from a fork.. but i need to fix the commit i added 🤦🏻  i've made it worse trying to delete this branch. i'm sorry!

PR #6289, opened 2021-05-13
[css-images-4] image() fragments example missing semi-colon

jsnkuhn:
https://drafts.csswg.org/css-images-4/#image-notation

Example 5 missing closing semi-colon


tabatkins:
Nothing is "missing" there - the semicolon is required to *separate* properties from each other, but it's not required at the end of a valid property on its own.

jsnkuhn:
I'm aware the semi-colon is unnecessary in this case but examples 1-3 also have individual properties value pairs but do include a closing semi-colon. This led me to believe it was missing from example 5.

PR #6297, opened 2021-05-19
[css-lists] Fix reversed counters to count how many elements increment or set the counter

### LINKED ISSUE #6233: [css-lists] Omitted start value of reversed list isn't calculated right
From https://drafts.csswg.org/css-lists/#instantiating-counters,

> When a counter is instantiated without an initial value, the user agent must dynamically calculate the initial value at layout-time to be the number of elements and pseudo-elements that instantiate the same counter in the same scope, plus 1.

Imagine a simple

```html
<ol reversed>
  <li>foo</li>
  <li>bar</li>
  <li>baz</li>
</ol>
```
There is a singe element instantiating `list-item`: the `<ol>`. The `<li>` don't instantiate it since they only use [`counter-increment`](https://drafts.csswg.org/css-lists/#increment-set)

> The counter-increment and counter-set properties manipulate the value of existing counters. They only instantiate new counters if there is no counter of the given name on the element yet.

But precisely, what we want to count is the number of `<li>`. Or maybe the (negated?) sum of their `counter-increments` if they have been set to a value different than 1 (-1?). Something like that, not the number of instantiations.

zcorpan:
Fixes #6233.

### DIFF
--- a/css-lists-3/Overview.bs
+++ b/css-lists-3/Overview.bs
@@ -1142,7 +1142,29 @@ Instantiating Counters</h4>
 
 	When a [=counter=] is [=instantiated=] without an initial value,
 	the user agent must dynamically calculate the initial value at layout-time
-	to be the number of elements and pseudo-elements that increment or set the same counter in the same [=scope=], plus 1.
+	to be the value returned by the following algorithm:
+
+	1. Let |num| be 0.
+
+	2. Let |first| be true.
+
+	3. For each element or pseudo-element |el| that increments or
+		sets the same counter in the same [=scope=]:
+
+		1. Let |incrementNegated| be |el|'s 'counter-increment' integer value for this counter,
+			multiplied by -1.
+
+		2. If |first| is true,
+			then add |incrementNegated| to |num| and
+			set |first| to false.
+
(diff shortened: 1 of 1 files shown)


css-lists-3/Overview.bs
```diff
@@ -1142,7 +1142,7 @@ Instantiating Counters</h4>
 
 	When a [=counter=] is [=instantiated=] without an initial value,
 	the user agent must dynamically calculate the initial value at layout-time
-	to be the number of elements and pseudo-elements that [=instantiate=] the same counter in the same [=scope=], plus 1.
+	to be the number of elements and pseudo-elements that increment or set the same counter in the same [=scope=], plus 1.
```
Loirooriol, line 1145: Or maybe just the number of list items (plus 1)?

zcorpan: Yeah, that could work too but would make auto-reversed counters only work for list items, not for any custom counter. Though reversed counters on non-lists is probably a *very* rare use case.

Loirooriol: Oh, right

tabatkins (approved):

PR #6300, opened 2021-05-20
[cssom-view] Remove "familiar with" check for move/resize methods

### LINKED ISSUE #1725: [cssom-view]  Stop using "incumbent settings object"?
"Incumbent settings object" is utilized in several places in CSSOM View, although there's a warning in the [HTML Standard](https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects):

> Warning! The incumbent and entry concepts should not be used by new specifications, as they are excessively complicated and unintuitive to work with. We are working to remove almost all existing uses from the platform: see [issue #1430](https://github.com/whatwg/html/issues/1430) for incumbent, and [issue #1431](https://github.com/whatwg/html/issues/1431) for entry.

domenic:
Closes #1725 by removing the usage of incumbent settings object. I could find no evidence of such a check in the Blink, Gecko, or WebKit codebases. Indeed, such a check should not generally be necessary, as (unlike, e.g., window.close()) these methods are not accessible cross-origin.

Also includes various changes for things that have updated in other specs:

* Use "this" instead of "context object"
* Use "string is" instead of "case-sensitive"
* Use "start node" and "start offset" instead of referring to properties of Range (or now, AbstractRange)
* Link to the correct "event loop" definition
* Remove redundant anchor blocks, as these days most definitions are properly exported


tabatkins:
ping @emilio

emilio (approved):

Issue #6175, opened 2021-04-02
[css-contain] What is the migration path for Container Queries?


mirisuzanne:
While at-rules provide their own test for positive support, there is currently no way to test for _lack of_ at-rule support. 

With the introduction of media-queries it became common practice to start with a reasonable fallback outside the query, and add progressive enhancement inside the query. That progressive approach also works with some container-queries ([see Andy Bell's example](https://codepen.io/piccalilli/pen/qBRRErw)), but it falls apart if an author wants to use media-queries as the fallback. Ideally, authors would be able to test for the lack of container-query support, and only apply media-queries when CQ is unsupported (see @eeeps [codepen attempt](https://codepen.io/eeeps/pen/wvgJmXb)).

If all browsers release 1D containment ([or any other new property/value syntax](https://github.com/w3c/csswg-drafts/issues/6174)) at the same time as container queries, Eric's approach should work:

```css
@container {
  /* progressive enhancements */
}

/* query negative support for the related new property/value */
@supports not (contain: inline-size) {
  @media (...) { /* fallback media-queries */ }
}
```

It's not likely that any browser would release `@container` before releasing `inline-size` containment, but it is possible a browser would release 1d containment first. If that is the only new syntax to test, it could result in some false negatives: browsers that support both `inline-size` & `@media`, but not `@container`, would miss out on the fallback media-queries. That's not ideal.

I think the only way to ensure this migration path works smoothly is to include container-query-specific syntax that can be tested by `@supports`. The `inline-size` value might work for that, but only if browsers are careful to implement both features together.

tabatkins:
We've talked about wanting to be able to test for at-rule existence in @supports in the past (I can't find the issues for it right now), but it might very well make sense to just allow a trivial "does this at-keyword look like something you support?" test, like `@supports at-rule(@container) {...}

mirisuzanne:
Both of our solutions exclude the long list of browsers supporting `@media` without support for `@supports`. In your case, browsers would also have to implement `@supports at-rule()` fairly broadly before `@container` is available, or it's not terribly useful. In order to use either pattern, authors would be hiding the media-query fallback from a long list of previously-supported browsers. Adding new support syntax might be a good idea long-term (also for layers and scope?), but in the immediate case it seems like it would make the issue more dramatic?

tabatkins:
True, but fixing it "better" now makes the *next* time this problem comes up less troublesome. ^_^

mirisuzanne:
[...]

PR #6312, opened 2021-05-26
[css-conditional-3] Remove "unknown" value concept

andruud:
The <general-enclosed> production now just produces "false".

See Issue #6175.

### DIFF
--- a/css-conditional-3/Overview.bs
+++ b/css-conditional-3/Overview.bs
@@ -238,7 +238,8 @@ tied to the support for certain features, as is the case for use of new
 layout system features.
 
 The syntax of the condition in the ''@supports'' rule
-is similar to that defined for <<media-condition>> in [[MEDIAQUERIES-4]]:
+is similar to that defined for <<media-condition>> in [[MEDIAQUERIES-4]],
+but without the "unknown" value logic:
 
 * negation is needed so that
 	the new-feature styles and the fallback styles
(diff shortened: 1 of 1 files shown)


andruud:
Plan (from the Blink perspective):

- Optimistically spec that `<general-enclosed>` results in "false".
- Add use-counters to Blink to understand the compat situation. (Wait ~6 months for data to come in).
- Verify that Webkit is willing to make this change.
- If compat situation allows it, make the change in Blink. Otherwise, CSSWG probably needs to revisit the issue, and we'll edit the spec again with some (less desirable) compromise.

Given that interop is [pretty poor currently](https://codepen.io/miriamsuzanne/full/wvJMxjb), I'm optimistic that it will be possible to make this change.

@dbaron, @mirisuzanne

tabatkins (approved):
Looks great, and the plan seems spot-on as well. I won't merge since I'm not an editor, but r+

dbaron (approved):
This looks good to me as well.

I'd note this concept was originally introduced in 206498e6bd16ad8dc00967a26607c72bdcd12dfa, although the commit message doesn't mention it at all.

andruud:
Investigating a bit more, looks like Chromium already behaves according to the new spec text. The reason Chromium appears to evaluate `not fn()` to false in the [demo](url) is because `not fn()` is actually invalid grammar. (`<any-value>`, requires **one or more** tokens).

PR #6361, opened 2021-06-08
[css-color-4] change text of MarkText example to match styling

### LINKED ISSUE #6362: [css-color-4] text of MarkText example reads FieldText instead of MarkText
All of the other examples use text that matches the colour name, except this one. 

https://drafts.csswg.org/css-color/#ex-SystemCombo

image

michaelficarra:
Fixes #6362.

Before:

image

After:

image


svgeesus (approved):
Thanks, my careless copy and paste I suspect

PR #6378, opened 2021-06-14
[css-contain-3] Make the query container an element, ++

andruud:
It is simpler to treat *elements* as query containers, and delay
interaction with the layout box until query-evaluation-time.
This makes it possible to answer the question "is this element a
container?" sooner, which is beneficial for implementing the feature.

It also seems to make more sense given that we're going to have
container features that don't really care about the layout box.

Also:

 - Require layout containment for dimensional queries.
 - Add 'block-size' as part of the computed value.


andruud:
cc @mirisuzanne

mirisuzanne (approved):
These changes make sense to me, thanks.

Issue #6288, opened 2021-05-13
[css-fonts] Clarify/reconsider interaction of new font-size-adjust options with writing modes


jfkthame:
In #6160, we agreed to extend `font-size-adjust` to accept a metric to use as the basis for adjustment, rather than always relying on the ex-height.

Supporting `ex` (the default/existing behavior) and `cap` seems straightforward enough; these would typically be values provided in the font's `OS/2` table.

However, `ch` is trickier, if we assume that it is intended to correspond to [the `ch` unit](https://drafts.csswg.org/css-values-4/#ch), because the Units & Values definition of `ch` requires that it depends on `writing-mode` and `text-orientation`. If the `ch` value for `font-size-adjust` is interpreted similarly, this implies that the same font, with the same `font-size` and `font-size-adjust` properties in effect, will be sized very differently in vertical-upright mode vs horizontal or vertical-sideways.

Consider a typical Latin font where the horizontal-mode `ch` unit may be, for example, 0.5em; however, the vertical-upright `ch` unit is typically more like 1em. If the author uses `font-size-adjust: ch 0.5`, intending to try and maintain a reasonably consistent width of text, then any content with `writing-mode: vertical-rl; text-orientation: upright` will get this font reduced to half of the computed font-size, because its vertical-upright `ch` of 1em will be scaled down to 0.5.

While it could be argued this is "expected" -- it's what the author asked for, maintaining a consistent advance for the characters -- I think it would in fact be surprising and undesirable. I believe the natural expectation of authors will be that for a given font, the combination of font-size and font-size-adjust will give a consistent result, not a radically different used size depending on the writing-mode orientation.

I propose, therefore, that we explicitly define that the `ch` value for `font-size-adjust` refers to *horizontal mode* character advance.

Given that this contrasts with the `ch` unit, perhaps sharing the keyword is in fact unhelpful, and we should expand the `font-size-adjust` value to `ch-width` for clarity.

To allow authors to explicitly request the equivalent adjustment for vertical-upright mode, we could add a vertical-advance version, perhaps called `vch`, or `ch-height` if we expand the horizontal one as suggested above.

I think all the same issues apply to `ic` as well as `ch`, and we should resolve that similarly, probably by explicitly splitting the value into `ic-width` and `ic-height`. (In many cases, horizontal and vertical advances may be the same, such that the issue is irrelevant, but that is not universally the case. And I think having `ic` and `ch` treated *differently* in this regard would be even more confusing.)

[...]

PR #6382, opened 2021-06-15
[css-fonts-5] Extend font-size-adjust

fantasai:
DO NOT SQUASH WHEN MERGING!!!!!!!!!!!!!!!!!!!!


css-fonts-5/Overview.bs
```diff
@@ -93,13 +93,228 @@ Issue(4044):
...
 
-The 'font-size-adjust' property is applied after the 'size-adjust' descriptor.
+	<pre class="propdef">
+	Name: font-size-adjust
+	Value: none | [ ex | cap | ch | ic ]? [ from-font | <<number>> ]
+	Initial: none
+	Applies to: all elements and text
+	Inherited: yes
+	Percentages: N/A
+	Computed value: the keyword ''font-size-adjust/none'', or a map of the specified keywords and their associated <<number>>
```
frivoal, line 103: > a map of…

Is there precedent for the computed value being a map? How is that different from just a tupple of the keyword and the associated number?

css-fonts-5/Overview.bs
```diff
@@ -93,13 +93,228 @@ Issue(4044):
...
+	regardless of the font used.
+
+	Values have the following meanings:
+
+	<dl dfn-type=value dfn-for="font-size-adjust">
+		<dt><dfn>none</dfn>
+		<dd>
+			No special 'font-size' adjustment is applied.
+
+		<dt><dfn>ex | cap | ch | ic</dfn>
```
frivoal, line 136: Should this really be marked up as a `<dfn>`? The actual values are defined further down, and that seems enough.

css-fonts-5/Overview.bs
```diff
@@ -93,13 +93,228 @@ Issue(4044):
...
-The 'font-size-adjust' property is applied after the 'size-adjust' descriptor.
+	<pre class="propdef">
+	Name: font-size-adjust
+	Value: none | [ ex | cap | ch | ic ]? [ from-font | <<number>> ]
+	Initial: none
+	Applies to: all elements and text
+	Inherited: yes
+	Percentages: N/A
+	Computed value: the keyword ''font-size-adjust/none'', or a map of the specified keywords and their associated <<number>>
+	Animation type: by computed value type
```
frivoal, line 104: Is it well defined how one of those animates? Having the keyword animate discretely while the number interpolate doesn't seem it would give particularly interesting results (except in the subset of cases where you're keeping the keyword constant). Or is it all discrete because one component of it is discrete?

Setting things up so that the resulting adjusted font-size is the thing that interpolates smoothly when you go from, say, `ex 0.5` to `cap 0.8` feels like it ought to be doable, but possibly overly complicated given the use cases for animating this, which I suspect are rather limited and gimmicky.

css-fonts-5/Overview.bs
```diff
@@ -93,13 +93,228 @@ Issue(4044):
...
+
+				<dt><dfn>cap</dfn>
+				<dd>
+					Normalize the cap-height of the fonts,
+					using the cap-height by the font size.
+
+				<dt><dfn>ch</dfn>
+				<dd>
+					Normalize the narrow pitch of the fonts,
+					using the advance measure of “0” (ZERO, U+0030)
```
frivoal, line 155: The advance measure of `0` is different depending on the writing mode (general case vs vertical upright). Does this mean that when specified with `ch` font-size-adjust is expect to stretch things by a different amount depending on the writing mode? I suspect not, but then it would be good to be clear that we're talking about the `width` of the glyph, since [advance measure](https://www.w3.org/TR/css-values-4/#length-advance-measure) is defined to vary based on writing modes.

Alternatively, we do want to be writing-mode sensitive, but then `ch` probably should be paired with 2 rather than 1 numbers.

frivoal, line 155: What if the `0` glyph isn't in the font? Presumably you fallback to `0.5em` as is the case with the `ch` unit, but since this is defined without reference to that, it's probably worth spelling out. Or maybe to the pair `0.5em 1em` if we do with two numbers to account for writing modes.

I suspect we don't want to fallback to `0.5em` in general and `1em` in upright vertical if we're using a single number.

css-fonts-5/Overview.bs
```diff
@@ -93,13 +93,228 @@ Issue(4044):
...
+					divided by the font size.
+
+				<dt><dfn>ic</dfn>
+				<dd>
+					Normalize the wide pitch of the font,
+					using the advance measure of “水” (CJK water ideograph, U+6C34)
```
frivoal, line 161: `水` is typically 1em wide *and* 1em tall, but these sizes can differ, which would make the advance measure depend on the writing mode (general case vs vertical upright). Does this mean that when specified with `ic`, `font-size-adjust` is expect to stretch things by a different amount depending on the writing mode? I suspect not, but then it would be good to be clear that we're talking about the `width` of the glyph, since [advance measure](https://www.w3.org/TR/css-values-4/#length-advance-measure) is defined to vary based on writing modes.

Alternatively, we do want to be writing-mode sensitive, but then `ic` probably should be paired with 2 rather than 1 numbers.

frivoal, line 161: What if the 水 glyph isn't in the font? Presumably you fallback to `1em` as is the case with the `ic` unit, but since this is defined without reference to that, it's probably worth spelling out.

fantasai: See https://github.com/w3c/csswg-drafts/issues/6288

fantasai: Good point. I should probably just redefine in terms of the units.

css-fonts-5/Overview.bs
fantasai: hmm, I got this wrong entirely actually. There's only ever one keyword-value pair.

css-fonts-5/Overview.bs
fantasai: It's defining that part of the grammar, isn't it.

css-fonts-5/Overview.bs
fantasai: Probably should append "if the keyword remains constant, otherwise discrete"

css-fonts-5/Overview.bs
fantasai: Actually, that would be bad for ch. Filed issue https://github.com/w3c/csswg-drafts/issues/6384

css-fonts-5/Overview.bs
fantasai: Filed https://github.com/w3c/csswg-drafts/issues/6384 and marked an issue.

fantasai: Can you adapt your comment to https://github.com/w3c/csswg-drafts/issues/6288 ?

css-fonts-5/Overview.bs
fantasai: Done.

css-fonts-5/Overview.bs
fantasai: Fixed.

css-fonts-5/Overview.bs
frivoal: https://github.com/w3c/csswg-drafts/issues/6288#issuecomment-861923340

css-fonts-5/Overview.bs
frivoal: https://github.com/w3c/csswg-drafts/issues/6288#issuecomment-861923340

css-fonts-5/Overview.bs
jfkthame: Having `font-size-adjust` be writing-mode sensitive seems to me unlikely to really be useful/desired, so I feel it would add complexity without adding value.

What could make sense would be to accept explicit vertical-measure values `ch-height` and `ic-height`, which could be used when a design is focused on vertical writing modes.

I would also propose renaming the `ch` and `ic` keywords here to `ch-width` and `ic-width` to make it clear what they represent, as they do *not* correspond directly with the similarly-named units (which *are* writing-mode dependent).

fantasai: @jfkthame Let's tackle this discussion in the issue you filed? :) Rather than here.

svgeesus: If there is no glyph corresponding to "0"?

css-fonts-5/Overview.bs
```diff
@@ -381,8 +603,15 @@ No new security or privacy considerations have been raised on this specification
 <h2 id="acknowledgments">
 Acknowledgments</h2>
 
-The CSS Working group would like to thank:
+	Firstly, the editors would like to thank all of the
+	<a href="https://www.w3.org/TR/css-fonts-4/#acknowledgments">contributors to the previous level of this module</a>.
```
svgeesus, line 607: (Improving the Fonts 4 ack list is on my todo)

svgeesus (approved):
This looks good to me

css-fonts-5/Overview.bs
```diff
@@ -213,6 +213,11 @@ Relative sizing: the 'font-size-adjust' property</h3>
 			text rendered using the first font in the family list
 			will display differently in older user agents
 			that don't support 'font-size-adjust'.
+
+		<dt><dfn>from-font</dfn>
+		<dd>
+			Computes to the <<number>> corresponding to
+			the <a>aspect value</a> of the [=first available font=].
```
shivamidow: What is the actual effect of 'from-font' for the primary font? Maybe, no size adjustment, right? Because m and m' are the same in the equation below.
```
u  =  ( m / m′ ) s
s  =  computed 'font-size-property' value
m  =  metric as specified by the 'font-size-adjust' property
m′ =  metric as specified in the actual font
u  =  adjusted font-size to use
```

jfkthame: Yes, exactly. The intent is that `from-font` makes it easy to leave the primary font untouched, and adjust fallbacks appropriately to match it.

shivamidow: I see. Thanks for the clarification.

PR #6394, opened 2021-06-17
[css-nesting-1] fix nesting example

jonathantneal:
[css-nesting-1] fix the nesting example

This changes adds a missing nesting selector (`&`) to [an example in the CSS Nesting specification](https://drafts.csswg.org/css-nesting-1/#mixing).

The current example, without a nesting selector:

```css
article {
  color: blue;
  :where(&) { color: red; }
}
```

The fixed example, with a nesting selector:

```css
article {
  color: blue;
  &:where(&) { color: red; }
}
```

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -673,7 +673,7 @@ Mixing Nesting Rules and Declarations {#mixing}
 		<xmp highlight=css>
 			article {
 				color: blue;
-				&:where(&) { color: red; }
+				@nest :where(&) { color: red; }
 			}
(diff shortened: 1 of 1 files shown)


tabatkins:
That breaks the point of the example, by raising the specificity again.

Adding an @nest would fix it while preserving the point it's making.

jonathantneal:
Thanks, @tabatkins, this is updated in https://github.com/w3c/csswg-drafts/pull/6394/commits/a06beebc1260bfeb2a2b34e876a9e9939ca61194

PR #6395, opened 2021-06-18
[css-fonts-5] absolutize axis of font-size-adjust metrics #6288

fantasai:
switches `ex | cap | ch | ic` to `ex-height | cap-height | ch-width | ic-width | ic-height`

Possible alternative: `ex | cap | ch | ic` to `ex | cap | ch-width | ic-width | ic-height`


jfkthame (approved):
LGTM, thanks. Although we didn't actually mention changing `ex` and `cap` in the most recent discussion, I think the consistency & clarity is better that way, so I'm in favor of this.

svgeesus (approved):
Yes, this looks good

PR #6304, opened 2021-05-22
[cssom-1] Merge constructable stylesheets into CSSOM (https://lists.w3.org/Archives/Public/www-style/2020Feb/0012.html)

mfreed7:
[cssom-1] Merge constructable stylesheets into CSSOM (https://lists.w3.org/Archives/Public/www-style/2020Feb/0012.html)

This is a first PR to merge the [Constructable Stylesheets proposal](https://wicg.github.io/construct-stylesheets/) into the CSSOM spec. This was resolved in Feb 2020: https://lists.w3.org/Archives/Public/www-style/2020Feb/0012.html. There are several sub-resolutions there, a few of which are handled here:

 - [RESOLVED: CSS() is merged into CSSOM1 (~~either @emilio or @rakina~~ @mfreed7 doing so)](https://lists.w3.org/Archives/Public/www-style/2020Feb/0012.html).
  - [RESOLVED: Remove title and alternate from constructor](https://github.com/WICG/construct-stylesheets/issues/105#issuecomment-577704271).
  - [RESOLVED: Add base URL constructor argument for sole purpose of resolving relative URLs in stylesheet, and location of the stylesheet remains that of the document](https://github.com/WICG/construct-stylesheets/issues/95#issuecomment-577708534).
  - [RESOLVED: Disallow @import in all constructable stylesheet apis with a note that we're doing it to match current state of modules and this might relax in future](https://github.com/WICG/construct-stylesheets/issues/119#issuecomment-588352418). Plus the [modifications discussed in this comment](https://github.com/WICG/construct-stylesheets/issues/119#issuecomment-642300024).

...and a few of which will follow this PR:

  - [RESOLVED: Constructed style sheets to always go after](https://github.com/WICG/construct-stylesheets/issues/93#issuecomment-577713131). (Which spec does this live inside?)
  - Modify URL parsing to take the new `CSSStyleSheet` `base URL` into account. (Suggestions appreciated for where exactly these specs live. These two perhaps? [referrer-policy](https://www.w3.org/TR/referrer-policy/#integration-with-css) [css-values-4](https://drafts.csswg.org/css-values-4/#relative-urls))
  - [Make adopted stylesheets ~~a CSSStyleSheetList and add the appropriate add/remove methods to the document or shadowRoot interface~~ an ObservableArray](https://github.com/WICG/construct-stylesheets/issues/45#issuecomment-825055766).

[...]

PR #6411, opened 2021-06-25
[cssom-1] Refactor CSSStyleSheet constructor and replaceSync to allow steps to be referenced from HTML spec

dandclark:
[cssom-1] Refactor CSSStyleSheet constructor and replaceSync to allow steps to be referenced from HTML spec

The [CSS module scripts](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/css-modules-v1-explainer.md) feature is being specified in HTML using several algorithms defined in cssom-1 on the `CSSStyleSheet` interface. In the review of the CSS module scripts HTML spec PR https://github.com/whatwg/html/pull/4898, there was a [concern ](https://github.com/whatwg/html/pull/4898#discussion_r324088522) that referencing the CSSStyleSheet methods directly implies that if these methods were overridden by user script, the UA should call the user script instead.

This change pulls the steps for the `CSSStyleSheet` constructor and `replaceSynce` method into separate algorithms so that they can be safely referenced from the HTML spec.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -748,24 +748,28 @@ represents a style sheet as defined by the CSS specification. In the CSSOM a
 <dl>
   <dt><dfn constructor for=CSSStyleSheet>CSSStyleSheet(<var>options</var>)</dfn></dt>
   <dd>
-    When called, execute the steps to <a>construct a CSSStyleSheet</a> given <var>options</var> and return the result.
+    When called, execute these steps:
+    <ol>
+      <li>Construct a new {{/CSSStyleSheet}} object <var>sheet</var>.</li>
+      <li>Run the steps to <a>initialize a CSSStyleSheet</a> on <var>sheet</var> given <var>options</var>.
+      <li>Return <var>sheet</var>.</li>
+    </ol>
   </dd>
 
-  <dt>To <dfn export>construct a CSSStyleSheet</dfn></dt> given {{CSSStyleSheetInit}} <var>options</var>, run these steps:
+  <dt>To <dfn export>initialize a {{/CSSStyleSheet}}</dfn></dt> <var>sheet</var> given {{CSSStyleSheetInit}} <var>options</var>, run these steps:
   <dd>
     <ol>
-      <li>Construct a new {{/CSSStyleSheet}} object <var>sheet</var> with the following properties:
       <ul>
-        <li><a spec=cssom>location</a> set to the <a spec=html>base URL</a> of the <a>associated Document</a> for the <a>current global object</a>.
-        <li><a>stylesheet base URL</a> set to the {{CSSStyleSheetInit/baseURL}} attribute value from <var>options</var>.
-        <li>No <a spec=cssom>parent CSS style sheet</a>.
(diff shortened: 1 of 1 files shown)


dandclark:
This refactoring doesn't change any behavior, so not sure if a CSSWG resolution is necessary. Happy to discuss in a meeting if necessary.

@mfreed7, FYI. I'd drafted an earlier version of this at https://github.com/WICG/construct-stylesheets/pull/138 but missed the train on https://github.com/w3c/csswg-drafts/pull/6304. 😊

mfreed7:
This change looks good to me! Thanks for cleaning it up in the process, and sorry for the churn on https://github.com/WICG/construct-stylesheets/pull/138.

mfreed7:
Gentle ping @emilio and @tabatkins - this PR is blocking the [CSS Modules PR](https://github.com/whatwg/html/pull/4898/).

emilio:
Sorry for the lag, I was on vacation until today. Will take a look ASAP, thanks.

emilio:
So I'm curious, why do you need the initialization to be separate from construction? That means that we need to define what a constructed but not initialized stylesheet does. Why can't it just use the CSSStyleSheet constructor?

The "synchronously replace the rules of a CSSStyleSheet" bit looks great though.

dandclark:
> So I'm curious, why do you need the initialization to be separate from construction? That means that we need to define what a constructed but not initialized stylesheet does. Why can't it just use the CSSStyleSheet constructor?

I don't think it's necessary for them to be separate. I've pushed a change to move the creation and initialization into the same set of steps.

Thanks!

cssom-1/Overview.bs
```diff
@@ -748,29 +748,23 @@ represents a style sheet as defined by the CSS specification. In the CSSOM a
...
-    <ol>
-      <li>Construct a new {{/CSSStyleSheet}} object <var>sheet</var>.</li>
-      <li>Run the steps to <a>initialize a CSSStyleSheet</a> on <var>sheet</var> given <var>options</var>.
-      <li>Return <var>sheet</var>.</li>
-    </ol>
+    When called, execute the steps to <a>create a CSSStyleSheet</a> given <var>options</var> and return the result.
   </dd>
 
-  <dt>To <dfn export>initialize a {{/CSSStyleSheet}}</dfn></dt> <var>sheet</var> given {{CSSStyleSheetInit}} <var>options</var>, run these steps:
+  <dt>To <dfn export>create a {{/CSSStyleSheet}}</dfn></dt> given {{CSSStyleSheetInit}} <var>options</var>, run these steps:
```
emilio, line 754: Can we call this algorithm "Create a constructed CSSStyleSheet" rather than "Create a CSSStyleSheet"? We already have https://drafts.csswg.org/cssom/#create-a-css-style-sheet and I found this a bit confusing, wdyt?

emilio (approved):
Other than that nit, it looks basically good to me, thanks!

cssom-1/Overview.bs
dandclark: Sounds good to me -- I've pushed this change. Thanks!

PR #6415, opened 2021-06-30
[selectors-4] Match HTML's use of certain terms defined in it.

hober:
This should not land until https://github.com/whatwg/html/pull/6603 lands in HTML.


hober:
Looks like the HTML side of things landed. @tabatkins, can you review & merge?

PR #6424, opened 2021-07-05
[cssom-1] Move the camel-cased-attribute definition a bit further up.

### LINKED ISSUE #6421: [cssom] "For each CSS property property that is a suppor..."
https://drafts.csswg.org/cssom/#example-a40690cb

> For each CSS property property that is a supported CSS property, the following partial interface applies where camel\-cased attribute is obtained by running the CSS property to IDL attribute algorithm for property\.

It would be nice to have an html anchor for this text, for reference in implementation documentation.

emilio:
Not sure if this is 100% worth it but it should close #6421, and I think
it's better.


svgeesus (approved):
This LGTM lets merge it

Issue #6409, opened 2021-06-24
[css-values-4] Small typo in Absolute-Lengths table


CamdynR:
https://www.w3.org/TR/css-values-4/#absolute-lengths

Under the "equivalence" column in the 'pt' row, I believe it should be 1/72**nd** instead of 1/72**th**

PR #6444, opened 2021-07-13
[css-values-3, css-values-4] Updated the equivalence for points in the absolute-lengths table #6409.

mitp7:
Small typo update to the absolute lengths table where I changed the equivalence in terms of inches given for points. 
Before it was 1pt = 1/72th of 1in, and I believe it should be 1pt=1/72nd of 1in. That section of the page can be found [here for css-values-3](https://www.w3.org/TR/css-values-3/#pt) and [here for css-values-4](https://www.w3.org/TR/css-values-4/#pt) page.


astearns:
Thanks! I’m not sure why the bot ran several times on this.

PR #6472, opened 2021-07-23
[scroll-animations-1] Add predefined `@scroll-timeline` to spec

argyleink:
just like how `css-counter-styles` ships with [predefined sets](https://drafts.csswg.org/css-counter-styles/#predefined-counters) for authors, I think `scroll-animations` could/should as well. This PR starts the conversation by including a very generic and super common scroll-timeline for authors to use `document-timeline`, so they don't have to create it themselves:

```css
@scroll-timeline document-scroll-timeline {}
```

This gives a common name to a (arguably the most common) timeline that authors will make.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -1222,12 +1222,12 @@ state of a scroll-driven animation may be inconsistent in the composited frame.
 <h2 id="predefined-scroll-timelines">
 Predefined Scroll Timelines</h2>
 
-  The following stylesheet fragments provide the normative definition 
-  for predefined scroll timelines that can be used by authors:
+The following stylesheet fragments provide the normative definition 
+for predefined scroll timelines that can be used by authors:
 
-  <pre class='stylesheet'><bdo dir=ltr>
-    @scroll-timeline document-timeline {}
-  </bdo></pre>
+<pre class='stylesheet'><bdo dir=ltr>
+@scroll-timeline document-scroll-timeline {}
+</bdo></pre>
 
 <h2 id="appendix-a-considerations-for-security-and-privacy">Appendix A. Considerations for Security and Privacy</h2>
 
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -1219,6 +1219,15 @@ run), the user agent may likewise choose not to sample scroll-driven animations
 for that composited frame. In such cases, the rendered scroll offset and the
 state of a scroll-driven animation may be inconsistent in the composited frame.
 
+<h2 id="predefined-scroll-timelines">
+Predefined Scroll Timelines</h2>
+
+  The following stylesheet fragments provide the normative definition 
+  for predefined scroll timelines that can be used by authors:
+
+  <pre class='stylesheet'><bdo dir=ltr>
+    @scroll-timeline document-timeline {}
+  </bdo></pre>
 
 <h2 id="appendix-a-considerations-for-security-and-privacy">Appendix A. Considerations for Security and Privacy</h2>
 


scroll-animations-1/Overview.bs
```diff
@@ -1219,6 +1219,15 @@ run), the user agent may likewise choose not to sample scroll-driven animations
 for that composited frame. In such cases, the rendered scroll offset and the
 state of a scroll-driven animation may be inconsistent in the composited frame.
 
+<h2 id="predefined-scroll-timelines">
+Predefined Scroll Timelines</h2>
+
+  The following stylesheet fragments provide the normative definition 
+  for predefined scroll timelines that can be used by authors:
```
flackr, line 1226: nit: this indent looks inconsistent with the rest of the file. I think this shouldn't be indented.

scroll-animations-1/Overview.bs
```diff
@@ -1219,6 +1219,15 @@ run), the user agent may likewise choose not to sample scroll-driven animations
...
+
+  <pre class='stylesheet'><bdo dir=ltr>
+    @scroll-timeline document-timeline {}
```
flackr, line 1229: document-timeline may not be a great name, it sounds a lot like https://www.w3.org/TR/web-animations-1/#the-documents-default-timeline which it's not. What would we call this if we were to add it to the Document object alongside the DocumentTimeline `document.timeline`?

flackr (commented):
This is a good idea, I think we may just need to bikeshed the name a bit.

scroll-animations-1/Overview.bs
argyleink: i updated to be more specific: `document-scroll-timeline`, which could be `document.scrollTimeline` in the DOM. thoughts?

flackr (commented):
This looks good to me. I'm a little uncertain how we make sure that there's no conflict between the UA provided timeline names and the author ones. Do we have examples of predefined named things in other spaces?

argyleink:
yeah, counters: https://www.w3.org/TR/css-counter-styles-3/#predefined-counters

tabatkins:
Yeah, there's several examples. 

* CSS-wide keywords. We try not to expand this set much, and when we do keep them as generic words that are very unlikely to show up in other contexts so we can ban them from being used anywhere else without issue.
* Color spaces for `color()`. Here, the author-defined ones are segregated from CSS-defined by requiring them to be dashed-idents. We're *pretty sure* this list is gonna expand over time, and the most reasonable names to choose are *also* likely names for authors to choose, so keeping them clearly separate was important.
* Counters. Here we just allow the two namespaces to mingle, with author-provided ones overriding CSS-defined ones (except for a handful that we've banned from overriding, for perf or internal-consistency reasons). If I were doing this today I *might* take the `color()` approach and require dashed-idents, but the cultural concerns of minting *some* numbering systems as predefined names while others are dash-prefixed might have stayed my hand.
* Font-face names. The namespaces mingle, but predefined ones are non-overridable. We don't expand the set much, and when we have done so in the recent past we've used names that are very unlikely to collide with existing author-defined names (the `ui-*` names).
* probably a few more, i dunno off the top of my head

So we're all over the board in behavior, and I dunno where we wanna go with this. If we think we'll be adding to the list of predefineds over time, I suspect we want either "mingle, but authors can override names" or "authors must use dashed-idents", because it looks like the names we'd choose are halfway reasonable for authors to choose too. If we think the list will be pretty static, we can possibly go with "mingle, but predefineds are illegal to override".

flackr:
Thanks for the detailed list of options! I'm glad to know that this isn't treading new ground. I suspect that "mingle, but authors can override names" is fine for this. I assume for this option there's nothing to specify as it's implied?

tabatkins:
It'll need to be specified so it's clear, but that's fairly easy text; you can steal from Counter Styles.

flackr:
FYI I think the need for this default scroll timeline may be obsolete with #6674

fantasai:
Closing as no longer relevant. since we redesigned the feature.

Issue #6452, opened 2021-07-14
[web-animations-2] Need to preserve animation progress when setting timeline on animation.


JTensai:
When transitioning between timeline types (time based, progress based) we need to make sure that progress of the animation is preserved.

This can be done by caching the progress by taking the old current time and dividing by the old effect end.

`progress = old_current_time / effect_end`

Then we take that percentage and multiply it by the new effect end in the following cases:

1. timeline is valid && non-monotonically increasing timeline && old play state == paused && old current time is valid
2. timeline is valid && monotonically increasing timeline && old current time is valid && old timeline is non-monotonically increasing timeline.

PR #6475, opened 2021-07-26
[web-animations-2] Preserve animation progress when setting the timeline

kevers-google:
[web-animations-2] Preserve animation progress when setting the timeline

When switching to a non-monotonic timeline on a paused animation, we set the hold time to the previous current time to prevent the animation from visually jumping until the animation is unpaused.

Similarly, when transition from a non-monotonic timeline to a monotonic timeline, we preserved current time to ensure smooth continuity in the animation.

This process worked well, when dealing exclusively with time-based timeline, but is not compatible with progress-based timelines. We address this by preserving previous progress in places where we previously preserved current time.  The change is compatible with time-based animations as the measured progress (current time / end time) is consistent for a time or progress based animation.

Issue #6452

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -213,9 +213,22 @@ follows:
     abort this procedure.
 1.  Let |previous play state| be |animation|'s [=play state=].
 1.  Let |previous current time| be the |animation|'s [=current time=].
-1.  Let |previous progress| = <code>|previous current time| /
-    [=end time=]</code> if |previous current time| is resolved and
-    unresolved otherwise.
+1.  Set |previous progress| based in the first condition that applies:
+    <div class="switch">
+
+    : If |previous current time| is unresolved:
+
+    :: Set |previous progress| to unresolved.
+
+    : If [=end time=] is zero:
+
+    :: Set |previous progress| to zero.
+
+    : Otherwise
+
+    :: Set |previous progress| = <code>|previous current time| /
+    [=end time=]</code>
+    </div>
 1.  Let |from finite timeline| be true if |old timeline| is not null and
     not [=monotonically increasing=].
 1.  Let |to finite timeline| be true if |timeline| is not null and not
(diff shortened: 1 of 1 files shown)


web-animations-2/Overview.bs
```diff
@@ -213,6 +213,9 @@ follows:
     abort this procedure.
 1.  Let |previous play state| be |animation|'s [=play state=].
 1.  Let |previous current time| be the |animation|'s [=current time=].
+1.  Let |previous progress| = <code>|previous current time| /
+    [=end time=]</code> if |previous current time| is resolved and
```
flackr, line 217: What if end time is 0? I think we need to have a special case in order to avoid division by 0.

What if end time is infinity? This will break preserving the current progress for any infinite iteration animation when going between document timelines I think.

See https://drafts.csswg.org/web-animations-1/#end-time

Maybe we should only use the previous progress if we are switching from a progress based to time based or time based to progress based animation, as these are the only two cases where we can't directly assign the progress from the other timeline.

flackr (changes requested):

web-animations-2/Overview.bs
kevers-google: Good catch. Missed a step where previous progress should be set to zero if end time is zero.  The case of end time being infinite should be fine since we don't preserve current time if transitioning between monotonic timelines.  It will simply mean that previous progress will be zero in this case.  Previous progress is only used in two special cases where we want to avoid an animation jump: monotonic to non-monotonic while paused, and non-monotonic to monotonic.

kevers-google: Expanding on my previous comment.  We will see a jump when transitioning from a monotonic to non-monotonic timeline on a paused animation if the previous end time was infinite.  This seems unavoidable.  In fact, we explicitly disallow a progress based animation with zero iteration duration.  This case trips a DOM error if the animation is initially set up with a progress based timeline.  Unclear if it will trip and error downstrream after setting the timeline.  Would like to follow up in a separate PR if additional handling is needed for this case.

flackr (approved):

Issue #6458, opened 2021-07-16
[web-animations-2] Support setting animation currentTime and startTime using CSSNumericValue


JTensai:
The process for setting animation currentTime and startTime needs to be updated to handle CSSNumericValue inputs.

For progress based animations, I suggest we only allow CSSNumericValues that are percentages. which makes setting the time as simple as:
`(new_current_time / 100) * timeline_duration`

There are certain edge cases where we might be able to convert absolute time units to progress based values, but it would never be complete. If the animation is creating using only percentage values for effect iteration duration, then there is no frame of reference to be able to convert an absolute time value into a progress based value. I think it is more straightforward to simply limit the allowed input types depending on the type of timeline being used.

For time based animations we should add support for certain types of CSSNumericValues. At first glance it seems reasonable to support: Number, Milliseconds, Seconds, Integer. Time unit value for Number and Integer would be assumed to be in Milliseconds since that is the time value of absolute units already. Seconds would need to be converted.

Exceptions should be thrown at each point where incompatible input is detected.

PR #6479, opened 2021-07-28
[web-animation-2] Making currentTime and startTime CSSNumerish

kevers-google:
[web-animation-2] Making currentTime and startTime CSSNumerish

The animation attributes currentTime and startTime were nullable doubles and represented times in units of milliseconds.  These units do not align well with progress based animations where it is more natural to work in percentages.

With this change, progress based animations expect and report times as percentages.  When setting current or start time, the value may be expressed as a double with implicit units of milliseconds, or as a CSSNumericValue with duration units ('s' or 'ms') when the animation is associated with a non-monotonic timeline. 

issue #6458

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -323,6 +323,33 @@ descendant effects and custom effects such that the first condition is:
 >     effect</a> or executing any <a>custom effects</a> associated with an
 >     <a>animation effect</a>.
 
+<h4 id='validating-a-css-numberish-time'>Validatring a CSSNumberish time</h4>
+
+The procedure to <dfn>validate a CSSNumberish time</dfn> for an input
+value of <var>time</var> is based on the first condition that matches:
+
+    <div class="switch">
+
+    :    If <em>all</em> of the following conditions are true:
+         * The animation is associated with a [=progress-based timeline=], and
+         * <var>time</var> is not a CSSNumeric value with percent units:
+
+    ::    <a>throw</a> a <span class="exceptionname">TypeError</span>.
+    ::    return false;
+
+    :    If <em>all</em> of the following conditions are true:
+         * The animation is not associated with a [=progress-based timeline=],
+           amd
+         * <var>time</var> is a CSSNumericValue:
+         * The units of time are not <a href="https://www.w3.org/TR/css3-values/#time">duration units</a>.
+
+    ::   <a>throw</a> a <span class="exceptionname">TypeError</span>.
+    ::   return false.
+
+    :    Oterwise
+    ::   return true.
+   </div>
+
 
 <h4 id='setting-the-current-time-of-an-animation'>Setting the current time of an
 Animation</h4>
@@ -341,6 +368,12 @@ an animation, <var>animation</var>, to <var>seek time</var> is as follows:
          <a>throw</a> a <span class=exceptionname>TypeError</span>.
 
     1.   Abort these steps.
+
+1.  Let <var>valid seek time</var> be the result of running the
+    [=validate a CSSNumberish time=] procedure with <var>seek time</var>
+    as the input.
+
+1.  If valid seek time is false, abort this procedure.
 
 1.  Update either <var>animation</var>'s <a>hold time</a> or
     [=start time=] as follows:
@@ -407,6 +440,12 @@ Animation</h4>
 The procedure to <dfn>set the start time</dfn>
 of <a>animation</a>, <var>animation</var>, to <var>new start time</var>,
 is as follows:
+
+1.  Let <var>valid start time</var> be the result of running the
+    [=validate a CSSNumberish time=] procedure with
+    <var>new start time</var> as the input.
+
+1.  If valid start time is false, abort this procedure.
 
 1.  Let <var>timeline time</var> be the current <a>time value</a> of the
     <a>timeline</a> that <var>animation</var> is associated with.
@@ -2004,6 +2043,34 @@ partial interface AnimationTimeline {
 
     </div>
 
+</div>
+
+<h3 id="the-animation-interface">The <code>Animation</code> interface</h3>
+<pre class="idl">
+[Exposed=Window]
+partial interface Animation : EventTarget {
+    attribute CSSNumberish?       startTime;
+    attribute CSSNumberish?       currentTime;
+};
+</pre>
+
+<div class='attributes'>
+:   <dfn attribute for=Animation>startTime</dfn>
+::  The [=start time=] of this animation. When associated with a
+    [=progress-based timeline=], [=start time=] must be returned as a
+    <a href="https://drafts.css-houdini.org/css-typed-om/#cssnumericvalue">
+    CSSNumericValue</a> with percent units. Otherwise [=start time=] must be
+    returned as a double value, representing the time in units of milliseconds.
+    Setting this attribute updates the [=start time=] using the procedure to
+    <a>set the start time</a> of this object to the new value.
+:   <dfn attribute for=Animation>currentTime</dfn>
+::  The <a>current time</a> of this animation. When associated with a
+    [=progress-based timeline=], <a>current time</a> must be returned as a
+    <a href="https://drafts.css-houdini.org/css-typed-om/#cssnumericvalue">
+    CSSNumericValue</a> with percent units. Otherwise <a>current time</a> must
+    be returned as a double value, representing the time in units of milliseconds.
+    Setting this attribute follows the procedure to <a>set the current time</a>
+    of this object to the new value.
 </div>
 
 <h3 id="the-animationeffect-interface">The <code>AnimationEffect</code> interface</h3>


web-animations-2/Overview.bs
```diff
@@ -2006,6 +2045,34 @@ partial interface AnimationTimeline {
 
 </div>
 
+<h3 id="the-animation-interface">The <code>Animation</code> interface</h3>
+<pre class="idl">
+[Exposed=Window]
+partial interface Animation : EventTarget {
+    attribute CSSNumberish?       startTime;
+    attribute CSSNumberish?       currentTime;
+};
```
flackr, line 2054: I think this should be expressed as a delta to the Animation interface from web-animations-1. E.g.

Update the Animation interface to the following:

> Repeat of complete interface with members changed

Alternately, we could try to target the specific members by saying update the startTime and currentTime of the Animation interface to the following:

web-animations-2/Overview.bs
```diff
@@ -2006,6 +2045,34 @@ partial interface AnimationTimeline {
...
+</pre>
+
+<div class='attributes'>
```
flackr, line 2057: As with the interface, please clarify that these are updates to the two members of the Animation interface.

web-animations-2/Overview.bs
```diff
@@ -2006,6 +2045,34 @@ partial interface AnimationTimeline {
...
+:   <dfn attribute for=Animation>startTime</dfn>
+::  The [=start time=] of this animation. When associated with a
+    [=progress-based timeline=], [=start time=] must be returned as a
+    <a href="https://drafts.css-houdini.org/css-typed-om/#cssnumericvalue">
+    CSSNumericValue</a> with percent units. Otherwise [=start time=] must be
+    returned as a double value, representing the time in units of milliseconds.
```
flackr, line 2063: It feels a little out of place to declare in the interface the types that must be returned. I would have expected these types to be the implicit result of starting an animation on or switching an animation to a progress based timeline.

kevers-google: CSSNumerish is naturally ambiguous since it could be a double or a CSSNumericValue, and when expressed as a CSSNumericValue it could have any compatible units (e.g. 's' vs 'ms').  Internally, it doesn't matter how we should express time, but should have a canonical form for API return values. Perhaps this belong in a normalization procedure that is called at the end of "getting the current/start time".

web-animations-2/Overview.bs
```diff
@@ -323,6 +323,33 @@ descendant effects and custom effects such that the first condition is:
 >     effect</a> or executing any <a>custom effects</a> associated with an
 >     <a>animation effect</a>.
 
+<h4 id='validating-a-css-numberish-time'>Validatring a CSSNumberish time</h4>
```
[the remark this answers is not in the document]

web-animations-2/Overview.bs
```diff
@@ -408,6 +441,12 @@ The procedure to <dfn>set the start time</dfn>
 of <a>animation</a>, <var>animation</var>, to <var>new start time</var>,
 is as follows:
 
+1.  Let <var>valid start time</var> be the result of running the
+    [=validate a CSSNumberish time=] procedure with
+    <var>new start time</var> as the input.
+
+1.  If valid start time is false, abort this procedure.
```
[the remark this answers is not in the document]

web-animations-2/Overview.bs
```diff
@@ -342,6 +369,12 @@ an animation, <var>animation</var>, to <var>seek time</var> is as follows:
 
     1.   Abort these steps.
 
+1.  Let <var>valid seek time</var> be the result of running the
+    [=validate a CSSNumberish time=] procedure with <var>seek time</var>
+    as the input.
+
+1.  If valid seek time is false, abort this procedure.
```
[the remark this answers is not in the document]

web-animations-2/Overview.bs
kevers-google: done.

web-animations-2/Overview.bs
kevers-google: Perhaps this should be clarified in an addendum to https://www.w3.org/TR/web-animations-1/#time-value-section.

tabatkins: Yeah, this is why Typed OM never uses CSSNumberish for method return types, and when it uses it for an attribute, it's very explicit about what is returned by the getter.

flackr: Ack, sounds like it is appropriate to be explicit here about the type that should be expected.

web-animations-2/Overview.bs
```diff
@@ -323,6 +323,33 @@ descendant effects and custom effects such that the first condition is:
...
+    :    If <em>all</em> of the following conditions are true:
+         * The animation is associated with a [=progress-based timeline=], and
+         * <var>time</var> is not a CSSNumeric value with percent units:
+
+    ::    <a>throw</a> a <span class="exceptionname">TypeError</span>.
+    ::    return false;
+
+    :    If <em>all</em> of the following conditions are true:
+         * The animation is not associated with a [=progress-based timeline=],
+           amd
```
flackr, line 342: typo

web-animations-2/Overview.bs
```diff
@@ -323,6 +323,33 @@ descendant effects and custom effects such that the first condition is:
...
+         * <var>time</var> is a CSSNumericValue:
+         * The units of time are not <a href="https://www.w3.org/TR/css3-values/#time">duration units</a>.
```
flackr, line 344: Links to other specs should use autolinks https://tabatkins.github.io/bikeshed/#autolink-biblio

E.g. This particular case would be `[[CSS3-VALUES#time|duration units]]`

flackr (changes requested):

web-animations-2/Overview.bs
tabatkins: Well, that would produce a section link. You probably actually want `<a type lt="<time>">duration units</a>`, to link to the definition of those units whereever they may be.

(You *should* be able to write `<<time|duration units>>`, but I haven't yet landed the fix harmonizing the parsing of all linking shorthands, ugh.)

web-animations-2/Overview.bs
kevers-google: fixed

web-animations-2/Overview.bs
```diff
@@ -2006,6 +2045,41 @@ partial interface AnimationTimeline {
...
+</pre>
+
+Replace [=start time=] and [=current time=] attribute descriptions with:
+
+<div class='attributes'>
+ :   <dfn attribute for=Animation>startTime</dfn>
+ ::  Update the description as:
+>    The [=start time=] of this animation. When associated with a
+>    [=progress-based timeline=], [=start time=] must be returned as a
+>    <a href="https://drafts.css-houdini.org/css-typed-om/#cssnumericvalue">
```
flackr, line 2067: Here and below, replace with {{CSSNumericValue}}.

flackr (changes requested):

web-animations-2/Overview.bs

web-animations-2/Overview.bs
```diff
@@ -323,6 +323,33 @@ descendant effects and custom effects such that the first condition is:
...
+    :    If <em>all</em> of the following conditions are true:
+         * The animation is associated with a [=progress-based timeline=], and
+         * <var>time</var> is not a CSSNumeric value with percent units:
+
+    ::    <a>throw</a> a <span class="exceptionname">TypeError</span>.
+    ::    return false;
+
+    :    If <em>all</em> of the following conditions are true:
+         * The animation is not associated with a [=progress-based timeline=],
+           and
```
flackr, line 342: This is an incorrect indent. Bikeshed requires a 4-space (or tab) indent.

web-animations-2/Overview.bs
```diff
@@ -2006,6 +2045,39 @@ partial interface AnimationTimeline {
...
+    attribute CSSNumberish?       currentTime;
+};
+</pre>
+
+Replace [=start time=] and [=current time=] attribute descriptions with:
+
+<div class='attributes'>
+ :   <dfn attribute for=Animation>startTime</dfn>
+ ::  Update the description as:
+>    The [=start time=] of this animation. When associated with a
```
flackr, line 2065: Since this is nested in a list this requires more indentation, The '>' needs to be aligned with Update on the line above. There's an example with delay and endDelay below, e.g.
```
 ::  Update the description as:
     > The [=start time=] of this animation. When associated with a
     > [=progress-based timeline=], [=start time=] must be returned as a
     > {{CSSNumericValue}} with percent units. Otherwise [=start time=] must be
```

web-animations-2/Overview.bs
```diff
@@ -2006,6 +2045,39 @@ partial interface AnimationTimeline {
...
+>    <a>set the start time</a> of this object to the new value.
+ :   <dfn attribute for=Animation>currentTime</dfn>
+ ::  update the description as:
+>    The <a>current time</a> of this animation. When associated with a
+>    [=progress-based timeline=], <a>current time</a> must be returned as a
+>    {{CSSNumericValue}}. with percent units. Otherwise <a>current time</a> must
+>    be returned as a double value, representing the time in units of
+>    milliseconds. Setting this attribute follows the procedure to
+>    <a>set the current time</a> of this object to the new value.
+</div>
```
flackr, line 2079: Newline before </div> to separate it from the block above.

web-animations-2/Overview.bs
```diff
@@ -2006,6 +2045,39 @@ partial interface AnimationTimeline {
...
+ ::  Update the description as:
+>    The [=start time=] of this animation. When associated with a
+>    [=progress-based timeline=], [=start time=] must be returned as a
+>    {{CSSNumericValue}} with percent units. Otherwise [=start time=] must be
+>    returned as a double value, representing the time in units of milliseconds.
+>    Setting this attribute updates the [=start time=] using the procedure to
+>    <a>set the start time</a> of this object to the new value.
+ :   <dfn attribute for=Animation>currentTime</dfn>
+ ::  update the description as:
+>    The <a>current time</a> of this animation. When associated with a
```
flackr, line 2073: Same indentation issues.

flackr (changes requested):

web-animations-2/Overview.bs
tabatkins: The list as a whole is indented incorrectly - for some reason it has a 1-space indent before each item, which doesn't do anything but *does* make the subsequent 4-space indents also misaligned. It *should* be indented one level from the container `div`, and then yeah, the blockquote would need two levels to nest it under the `dd` properly.

Just use tab indents; humans can't reliably space-indent to save their lives. ^_^

flackr: I don't have much experience with tab indenting specs but I'll take your word that it's simpler. I guess it takes away the temptation to use anything less than a full tab indent ^_^. I take it I should update https://github.com/w3c/csswg-drafts/pull/6565#pullrequestreview-744230939 or follow up with another PR to do this for the whole spec and then we can update this to match.

kevers-google (commented):
Indent issues should all be resolved now.  Formatting glitches with dl.switch have also been addressed.

web-animations-2/Overview.bs

web-animations-2/Overview.bs
kevers-google: Done

web-animations-2/Overview.bs
```diff
@@ -6,6 +6,14 @@
...
 }
 
+dl.switch  > [data-md] > p {
+  display: inline;
+}
+
+dl.switch > dt > ul > li {
+  text-indent: 0;
+}
+
```
flackr, line 16: @tabatkins there seems to be issues with the generated bikeshed css when we switched to dl instead of just the div. You can see the issue at this [link](https://drafts.csswg.org/web-animations-2/#setting-the-current-time-of-an-animation#:~:text=Update%20either%20animation%E2%80%99s%20hold%20time%20or%20start%20time%20as%20follows) which @kevers-google is trying to work around here. Do you know if there's something wrong with the spec formatting or generator?

web-animations-2/Overview.bs
```diff
@@ -357,29 +395,37 @@ an animation, <var>animation</var>, to <var>seek time</var> is as follows:
...
+                <var>seek time</var>.
+
+            :   Otherwise,
+            ::  Set <var>animation</var>'s [=start time=] to the result of
+                evaluating
+                <code>|timeline time| - (|seek time| / [=playback rate=])</code>
+                where <var>timeline time</var> is the current <a>time value</a>
+                of <a>timeline</a> associated with <var>animation</var>.
 
     </dl>
```
flackr, line 430: Looks like the indentation level here doesn't match the opening tag.

web-animations-2/Overview.bs
tabatkins: Could you provide more details on what's wrong? I see that the "switch icon" is on its own line (unavoidable given the way it's injected, since the dt contains block elements), but it doesn't look like that's what your CSS is trying to mess with.

flackr: The bullets in the list are also incorrectly indented. See https://drafts.csswg.org/web-animations-1/#setting-the-current-time-of-an-animation for comparison. web-animations-1 seems to be using <div class="switch">: https://github.com/w3c/csswg-drafts/blob/main/web-animations-1/Overview.bs#L957

kevers-google: The switch being on it's own line was resolved by the rule:

dl.switch  > [data-md] > p {
  display: inline;
}

This aligns with a rule we have in web-animations-1, namely:

div.switch > dl > dt > p {
  display: inline;
}

The overlap resulted from the following rules in base.css:

dl.switch > dt {
    text-indent: -1.5em;
}

When applied to a paragraph inside a list item, it creates an overlap between the paragraph and the marker.

web-animation-1 controls the indent within a switch via these rules:

div.switch > dl > dt:before {
  margin-left: -1.3em;
}
div.switch > dl > dt > p {
  display: inline;
}

tabatkins: Oh, huh, I see that `base.css` is doing some weird text-indent stuff. I... don't understand why it's doing that; it appears to only work well on single-line `dt`s, and makes structured ones like this look weird. @fantasai, I don't remember if this style was imported from Bikeshed's stylesheet or done by you, any idea?

web-animations-2/Overview.bs
kevers-google: Fixed.

web-animations-2/Overview.bs
kevers-google: Moving the style changes to a separate PR to unblock the remainder of the changes.

kevers-google: Style changes moved to https://github.com/w3c/csswg-drafts/pull/6655.

flackr (approved):

PR #6483, opened 2021-07-28
[css-nesting-1] Allow conditional nested rules to adopt context

### LINKED ISSUE #5805: [css-nesting] simpler embedding of media queries
The spec has this for how to embed a MQ in a rule:

```
.foo {
  display: grid;

  @media (orientation: landscape) {
    & { 
      grid-auto-flow: column; 
    }
  }
}
```

But the unadorned ampersand and extra set of braces seems extraneous. It is incredibly useful, and much simpler, in SCSS to just write it like this for the commonest uses:

```
.foo {
  display: grid;

  @media (orientation: landscape) {
      grid-auto-flow: column; 
  }
}
```

In fact, embedded the MQ in the rule like that, without extra braces, is one of my favorite things about writing in SCSS. It just feels natural and familiar. Even in old vanilla CSS, I can't tell you the number of times I've forgotten to include the selector inside the MQ, when it was just an MQ for a single rule, written right after the bare rule. When the MQ is inside the rule, instead for the other was around, it is a joy to not have to write another selector and braces inside it. 

You don't have the garden path problem here. If the first non-white-space character after the opening brace of the nested MQ (or after a closing brace of a nested rule within that MQ) is not an ampersand or @ sign, then it is the beginning of a property for a declaration. 

I think there is nothing ambiguous about the following, and it is easier to write and read:

```
.btn {
  text-decoration: none;

  @media (max-width: 768px) {
      display: block; 

      &:hover { 
          box-shadow: 0 0 2px black inset;
      }
  }
}
```

So basically, the `& {` and `}` parts are assumed, as a shortcut for when you don't need any more than that, when you already have the braces from the nested @ rule. This should hold true for other nested @ rules too, such as `@supports`, etc.

argyleink:
[css-nesting-1] Fixes #5805

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -466,11 +466,12 @@ Nesting Conditional Rules {#conditionals}
 -----------------------------------------------
 
 	A style rule can have any number of <a>conditional rules</a> inside of it, of any type,
-	intermixed with any number of declarations, in any order.
+	with rules required to come before declarations (see <a href="#mixing">Mixing Nesting Rules and Declarations</a>).
 
-	The presence of a nested conditional engages the same logic as if ''@nest'' was present.
-	The nested conditional rules must contain a <a>nesting selector</a> and follow the same
-	rules as outlined in <a href="#direct">direct nesting</a>.
+	If there are only rules in the nested conditional, the nesting context is adopted.
+
+	In the presence of a <a>nesting selector</a>, the same
+	rules as outlined in <a href="#direct">direct nesting</a> should be used.
 
 	<div class="example">
 		For example, the following conditional nestings are valid:
@@ -480,9 +481,7 @@ Nesting Conditional Rules {#conditionals}
 			display: grid;
 
 			@media (orientation: landscape) {
-				& {
-					grid-auto-flow: column;
-				}
+				grid-auto-flow: column;
 			}
 		}
 		/* equivalent to
@@ -499,14 +498,10 @@ Nesting Conditional Rules {#conditionals}
 			display: grid;
 
 			@media (orientation: landscape) {
-				& {
-					grid-auto-flow: column;
-				}
+				grid-auto-flow: column;
 
 				@media (min-inline-size > 1024px) {
-					& {
-						max-inline-size: 1024px;
-					}
+					max-inline-size: 1024px;
 				}
 			}
 		}
@@ -551,15 +546,6 @@ Nesting Conditional Rules {#conditionals}
 		But the following are invalid:
 
 		<pre class=lang-css>
-		.foo {
-			display: grid;
-
-			@media (orientation: landscape) {
-				grid-auto-flow: column;
-			}
-		}
-		/* Invalid because there's no nesting selector */
-
 		.foo {
 			color: red;
 


argyleink:
request for `agenda+` because we need help resolving some syntax:

**Current**
```css
.foo {
  display: grid;

  @media (orientation: landscape) {
    & {
      grid-auto-flow: column;
    }
  }
}
```

**Proposed**
```css
.foo {
  display: grid;

  @media (orientation: landscape) {
    grid-auto-flow: column;
  }
}
```

It's an exception to the unambiguous `&` pattern established by the nesting 1 spec but also very popular syntactic sugar.

LeaVerou:
I support this. @tabatkins is it feasible?

tabatkins:
It's definitely feasible spec-wise; it's just carrying the same parsing context of the style rule body into the MQ body.

css-nesting-1/Overview.bs
```diff
@@ -466,11 +466,12 @@ Nesting Conditional Rules {#conditionals}
 -----------------------------------------------
 
 	A style rule can have any number of <a>conditional rules</a> inside of it, of any type,
-	intermixed with any number of declarations, in any order.
+	with rules required to come after declarations (see <a href="#mixing">Mixing Nesting Rules and Declarations</a>).
 
-	The presence of a nested conditional engages the same logic as if ''@nest'' was present.
-	The nested conditional rules must contain a <a>nesting selector</a> and follow the same
-	rules as outlined in <a href="#direct">direct nesting</a>.
+	If there are only rules in the nested conditional, the nesting context is adopted.
```
LeaVerou, line 471: Do you mean:
```suggestion
	If there are only declarations in the nested conditional, the nesting context is adopted.
```
?

But even so, why restrict it that way? if we have mixed rules and declarations, we can still adopt the nesting context. E.g. this:

```css
.fancy-link {
	color: purple;

	@supports (fancy-shadow: purple) {
		color: transparent;
		fancy-shadow: purple;

		&:hover {
			color: mix(purple, white);
		}
	}
}
```

can be sugar for:

```css
.fancy-link {
	color: purple;

	@supports (fancy-shadow: purple) {
		& {
			  color: transparent;
			  fancy-shadow: purple;
	  
			  &:hover {
				  color: mix(purple, white);
			  }
		}
	}
}
```

Even though the conditional rule contains both rules and declarations.

css-nesting-1/Overview.bs
```diff
@@ -466,11 +466,12 @@ Nesting Conditional Rules {#conditionals}
 -----------------------------------------------
 
 	A style rule can have any number of <a>conditional rules</a> inside of it, of any type,
-	intermixed with any number of declarations, in any order.
+	with rules required to come after declarations (see <a href="#mixing">Mixing Nesting Rules and Declarations</a>).
 
-	The presence of a nested conditional engages the same logic as if ''@nest'' was present.
-	The nested conditional rules must contain a <a>nesting selector</a> and follow the same
-	rules as outlined in <a href="#direct">direct nesting</a>.
+	If there are only declarations in the nested conditional, the nesting context is adopted.
```
argyleink, line 471:
```suggestion
	All valid declarations in the nested conditional adopt the nesting context.
```

argyleink: is that clearer @LeaVerou? hopefully it's less restrictive sounding, that nested declarations adopt context and discovered nesting selectors follow the nesting logic path.

LeaVerou: Yes, though I wonder if it's clearer to word this in terms of syntactic sugar, i.e. that if there are any declarations inside the conditional rule, there is an implicit `& { ... }` around all of them.

argyleink: i could add that to the examples where we're saying x is equivalent to y? with nested conditional examples showing the explicit in-between state which includes the `&`? or you think it's better in the text inline here?

Issue #229, opened 2016-06-24
[css-easing-2] Complex easing/timing functions


rachelnabors:
In light of the Webkit team's implementation of `spring()`, it's apparent we need to attend to [the issue of complex timing functions](https://lists.w3.org/Archives/Public/public-fx/2015JulSep/0038.html) sooner rather than later.
# The problem

Designers often need more advanced timing functions than can be described with cubic-beziers. They are not limited to spring functions, either. A common problem is there is no effective way to export a timing graph from Adobe After Effects to a timing function that could be used with CSS or with the Web Animations API. Currently designers have to hack together individual timing functions using CSS animation keyframes, which is impossible to do by hand in all but the most trifling instances.

`spring()` is just a bandaid.
# The solution

We need a format to write functions like `spring()` in, one that we can export to from software like AfterEffects and prototyping tools that have yet to be built.

I am not in a position to propose the technical specifications of this solution. But there are people who have that knowledge. I have invited them (@visiblecode) to share their proposals below.

birtles:
There is some previous discussion of this topic, thread starting here: https://lists.w3.org/Archives/Public/public-fx/2015JulSep/0034.html

visiblecode:
Cubic beziers aren't a bad choice for timing functions. They're very easy to evaluate (just a few additions and multiplications). Also, most animation tools use some form of cubic spline to specify animation paths, which can be converted into cubic bezier easings with good fidelity.

There are two problems with `cubic-bezier()` as it is currently specced, however:

1. You can't specify a spline of multiple segments
2. `cubic-bezier()` is abusing 2d parametric curves to specify 1d functions

The first issue can be addressed by, essentially, just adding more points to the cubic-bezier() syntax. I believe this has already been proposed in various forms.

The second issue is pretty significant, though, and deserves some unpacking. `cubic-bezier()` is specified as a _2d_ bezier, a 2d parametric curve. This has a lot of implications, including:

1. Without additional constraints, the resulting curve isn't guaranteed to have a singular y value for a given x and can't be used to define a mathematical function. This has been dealt with by restricting x to the range [0, 1], but that doesn't actually address the root (no pun) problem -- because a cube root is involved there are still two extra solutions in the complex plane.
2. As mentioned, implementors must solve cube roots to work backwards from x to the implicit t parameter, before they get to the normal (and much simpler/exact!) bezier calculation to evaluate the easing function.
[...]

PR #6533, opened 2021-08-20
[css-easing-1] Some ideas for linear() easing

jakearchibald:
This is an attempt to make some progress on custom easing functions, specifically https://github.com/w3c/csswg-drafts/issues/229#issuecomment-861247074

I've never edited a CSS spec before, so I don't really know what I'm doing. Sorry if this isn't helpful in any way. Here's what I'm trying to do:

**Edit**: I've updated this following discussion below.

There's already a `linear` keyword. I figured we could overload that, so `linear` is a shorthand for `linear()`.

The arguments are a comma-separated groups of:

- Output (number)
- Start input (optional, percentage)
- End input (optional, percentage)

The values work very similar to `linear-gradient`.

`linear(0, 0.7, 0.87, 0.98, 1)` - this creates 5 points:

Screenshot 2021-08-23 at 13 11 21

Which is equivalent to `linear(0 0%, 0.7 25%, 0.87 50%, 0.98 75%, 1 100%)`, as points without a 'start input' are given one automatically (using the same process as linear gradients).

The third value, again similar to gradients, can be used to create an extra point using the same output value. So `linear(0, 0.5 25% 75%, 1)` creates 4 points:

Screenshot 2021-08-23 at 08 52 21

Output values will typically be 0-1, but can be outside of that, as they can with `cubic-bezier`. A likely usage here will be elastic/spring easings.

Similarly, input values will typically be 0%-100%, but can extend beyond that to provide points for edge-cases where easings are chained.

The ends of the graph automatically extend beyond their specified point using their final angle, so the previous easing (`linear(0, 0.5 25% 75%, 1)`) would extend like this:

Screenshot 2021-08-23 at 09 00 32

However, if the final two points are in the same position, the line extends horizontally, so `linear(0, 0.5 25% 75%, 1 100% 100%)` (the difference being the `100% 100%` at the end), would be:

Screenshot 2021-08-23 at 09 02 27

Some edge cases: `linear()` is equivalent to `linear` (as in, an identity function). If only one value is provided, then the output value is always that value, so `linear(0.5)` always outputs `0.5`.

The overall goal is to allow something like:

```css
.whatever {
  animation-timing-function: linear(0, 0.003, 0.013, 0.03, 0.05, 0.08, 0.11, 0.15, 0.2, 0.26, 0.31, 0.38, 0.45, 0.53, 0.62, 0.71, 0.81, 0.9, 0.99, 0.94, 0.89, 0.85, 0.82, 0.79, 0.77, 0.76, 0.752, 0.75, 0.755, 0.77, 0.78, 0.81, 0.84, 0.87, 0.92, 0.97, 0.99, 0.97, 0.95, 0.94, 0.938, 0.94, 0.949, 0.96, 0.99, 0.994, 0.986, 0.985, 0.989, 1 100% 100%);
}
```

…which would graph like this:

Screenshot 2021-08-23 at 11 45 53

And here's a demo of how that would animate: https://static-misc-3.glitch.me/linear-easing/

### DIFF
--- a/css-easing-1/Overview.bs
+++ b/css-easing-1/Overview.bs
@@ -127,12 +127,18 @@ The syntax for specifying an [=easing function=] is as follows:
 
 <h3 id=the-linear-easing-function oldids=linear-timing-function-section>The linear easing function: ''linear()''</h3>
 
-A <dfn export lt="linear easing function|linear timing function">linear easing
-function</dfn> is an easing function that interpolates linearly between its [=linear easing function/points=].
+A <dfn export lt="linear easing function|linear timing function">linear easing function</dfn>
+is an easing function
+that interpolates linearly
+between its [=linear easing function/points=].
 
-A [=linear easing function=] has <dfn for="linear easing function">points</dfn>, a [=/list=] of [=linear easing points=]. Initially a new empty [=/list=].
+A [=linear easing function=] has <dfn for="linear easing function">points</dfn>,
+a [=/list=] of [=linear easing points=].
+Initially a new empty [=/list=].
 
-A <dfn>linear easing point</dfn> is a [=/struct=] that has:
+A <dfn>linear easing point</dfn>
+is a [=/struct=]
+that has:
 
 <dl dfn-for="linear easing point">
 
@@ -145,17 +151,26 @@ A <dfn>linear easing point</dfn> is a [=/struct=] that has:
 
 <section algorithm="to calculate linear output progress">
 
-To <dfn export>calculate linear output progress</dfn> for a given [=linear easing function=] |linearEasingFunction|, and an [=input progress value=] |inputProgress|, perform the following. It returns an [=output progress value=].
+To <dfn export>calculate linear output progress</dfn>
+for a given [=linear easing function=] |linearEasingFunction|,
+and an [=input progress value=] |inputProgress|,
+perform the following.
+It returns an [=output progress value=].
 
-1.   If |linearEasingFunction|'s [=linear easing function/points=] [=list/is empty=], then return |inputProgress|.
+1.   If |linearEasingFunction|'s [=linear easing function/points=] [=list/is empty=],
+     then return |inputProgress|.
 
-1.   If |linearEasingFunction|'s [=linear easing function/points=] [=list/size=] is 1, then return |linearEasingFunction|'s [=linear easing function/points=][0]'s [=linear easing point/output=].
+1.   If |linearEasingFunction|'s [=linear easing function/points=] [=list/size=] is 1,
+     then return |linearEasingFunction|'s [=linear easing function/points=][0]'s [=linear easing point/output=].
 
-1.   Plot |linearEasingFunction|'s [=linear easing function/points=] points on a graph using [=linear easing point/input=] as the x-axis and [=linear easing point/output=] as the y-axis.
+1.   Plot |linearEasingFunction|'s [=linear easing function/points=] points on a graph
+     using [=linear easing point/input=] as the x-axis
+     and [=linear easing point/output=] as the y-axis.
 
 1.   Interpolate the points linearly.
 
-1.   Extend the ends of the graph to infinity using the final angle of the line at either end. If the line starts/ends with two points of the same position, the line should extend along the x-axis.
+1.   Extend the ends of the graph to infinity using the final angle of the line at either end.
+     If the line starts/ends with two points of the same position, the line should extend along the x-axis.
 
 1.   Return the y-axis value corresponding to the x-axis value of |inputProgress|.
 
@@ -166,8 +181,8 @@ To <dfn export>calculate linear output progress</dfn> for a given [=linear easin
 <pre class="prod">
   <dfn>&lt;linear-stop-list></dfn> =
     [ <<linear-hint>>? , <<linear-stop>> ]#
-  <dfn>&lt;linear-stop></dfn> = <<number>> && <<length-percentage>>?
-  <dfn>&lt;linear-hint></dfn> = <<length-percentage>>
+  <dfn>&lt;linear-stop></dfn> = <<number>> && <<percentage>>?
+  <dfn>&lt;linear-hint></dfn> = <<percentage>>
 </pre>
 
 TODO: I want to support syntax like <code>linear(0 0% 0%, 1 100% 100%)</code>, but I can't figure out how it's done. <code>linear-gradient</code> seems to support that syntax but <a href="https://drafts.csswg.org/css-images-3/#typedef-color-stop-list">I can't see it in the grammar</a>.
--- a/css-easing-1/Overview.bs
+++ b/css-easing-1/Overview.bs
@@ -143,11 +143,13 @@ that has:
 <dl dfn-for="linear easing point">
 
 : <dfn>input</dfn>
-:: A number
-: <dfn>output</dfn>
 :: A number or null
 
-      Note: The [=linear easing point/output=] is only null during the [=create a linear easing function=] algorithm.
+      Note: The [=linear easing point/input=] is only null during the [=create a linear easing function=] algorithm.
+
+: <dfn>output</dfn>
+:: A number
+
 
 </dl>
 
--- a/css-easing-1/Overview.bs
+++ b/css-easing-1/Overview.bs
@@ -145,7 +145,9 @@ that has:
 : <dfn>input</dfn>
 :: A number
 : <dfn>output</dfn>
-:: A number
+:: A number or null
+
+      Note: The [=linear easing point/output=] is only null during the [=create a linear easing function=] algorithm.
 
 </dl>
 
@@ -179,13 +181,64 @@ It returns an [=output progress value=].
 <dfn function lt="linear()">linear(<<linear-stop-list>>)</dfn>
 
 <pre class="prod">
-  <dfn>&lt;linear-stop-list></dfn> =
-    [ <<linear-hint>>? , <<linear-stop>> ]#
-  <dfn>&lt;linear-stop></dfn> = <<number>> && <<percentage>>?
-  <dfn>&lt;linear-hint></dfn> = <<percentage>>
+  <dfn>&lt;linear-stop-list></dfn> = [ <<linear-stop>> ]#
+  <dfn>&lt;linear-stop></dfn> = <<number>> && <<linear-stop-length>>?
+  <dfn>&lt;linear-stop-length></dfn> = <<percentage>>{1,2}
 </pre>
 
-TODO: I want to support syntax like <code>linear(0 0% 0%, 1 100% 100%)</code>, but I can't figure out how it's done. <code>linear-gradient</code> seems to support that syntax but <a href="https://drafts.csswg.org/css-images-3/#typedef-color-stop-list">I can't see it in the grammar</a>.
+<section algorithm="to create a linear easing function">
+
+To <dfn>create a linear easing function</dfn>
+from a <<linear-stop-list>> |stopList|,
+perform the following.
+It returns a [=linear easing function=].
+
+1.   Let |function| be a new [=linear easing function=].
+
+1.   Let |largestInput| be negative infinity.
+
+1.   For each |stop| in |stopList|:
+
+     1.   Let |point| be a new [=linear easing point=].
+
+     1.   [=list/Append=] |point| to |function|'s [=linear easing function/points=].
+
+     1.   Set |point|'s [=linear easing point/output=] to |stop|'s <<number>>.
+
+     1.   If |stop| has a <<linear-stop-length>>, then:
+
+          1.   Set |point|'s [=linear easing point/input=] to whichever is greater:
+               |stop|'s <<linear-stop-length>>'s first <<percentage>> as a number,
+               or |largestInput|.
+
+          1.   Set |largestInput| to |point|'s [=linear easing point/input=].
+
+          1.   If |stop|'s <<linear-stop-length>> has a second <<percentage>>, then:
+
+               1.   Let |extraPoint| be a new [=linear easing point=].
+
+               1.   [=list/Append=] |extraPoint| to |function|'s [=linear easing function/points=].
+
+               1.   Set |extraPoint|'s [=linear easing point/output=] to |stop|'s <<number>>.
+
+               1.   Set |extraPoint|'s [=linear easing point/input=] to whichever is greater:
+                    |stop|'s <<linear-stop-length>>'s second <<percentage>> as a number,
+                    or |largestInput|.
+
+               1.   Set |largestInput| to |extraPoint|'s [=linear easing point/input=].
+
+     1.   Otherwise, if |stop| is the first [=list/item=] in |stopList|,
+          then set |point|'s [=linear easing point/input=] to 0.
+
+     1.   Otherwise, if |stop| is the last [=list/item=] in |stopList|,
+          then set |point|'s [=linear easing point/input=] to 1.
+
+1.   For runs of [=list/items=] in |function|'s [=linear easing function/points=] that have a null [=linear easing point/input=],
+     assign a number to the [=linear easing point/input=] by linearly interpolating between the previous and next [=linear easing function/points=] that have a non-null [=linear easing point/input=].
+
+1.   Return |function|.
+
+</section>
 
 <h3 id="the-linear-easing-keyword">The linear easing keyword: ''linear''</h3>
 
--- a/css-easing-1/Overview.bs
+++ b/css-easing-1/Overview.bs
@@ -125,16 +125,62 @@ The syntax for specifying an [=easing function=] is as follows:
   <<step-easing-function>></div>
 
 
-<h3 id=the-linear-easing-function oldids=linear-timing-function-section>The linear easing function: ''linear''</h3>
+<h3 id=the-linear-easing-function oldids=linear-timing-function-section>The linear easing function: ''linear()''</h3>
 
-The <dfn export lt="linear easing function|linear timing function">linear easing
-function</dfn> is an identity function
-meaning that its [=output progress value=] is equal to the
-[=input progress value=] for all inputs.
+A <dfn export lt="linear easing function|linear timing function">linear easing
+function</dfn> is an easing function that interpolates linearly between its [=linear easing function/points=].
 
-The syntax for the [=linear easing function=] is simply the
-<dfn dfn-type=value for=easing-function>linear</dfn> keyword.
+A [=linear easing function=] has <dfn for="linear easing function">points</dfn>, a [=/list=] of [=linear easing points=]. Initially a new empty [=/list=].
 
+A <dfn>linear easing point</dfn> is a [=/struct=] that has:
+
+<dl dfn-for="linear easing point">
+
+: <dfn>position</dfn>
+:: A number
+: <dfn>value</dfn>
+:: A number
+
+</dl>
+
+<section algorithm="to calculate linear output progress">
+
+To <dfn>calculate linear output progress</dfn> for a given [=linear easing function=] |linearEasingFunction|, and an [=input progress value=] |inputProgress|, perform the following steps. They return an [=output progress value=].
+
+1.   If |linearEasingFunction|'s [=linear easing function/points=] [=list/is empty=], the return |inputProgress|.
+
+     Note: This is a special case that creates a pure identity function.
+
+1.   Let |previousPoint| be null.
+
+1.   Let |nextPoint| be null.
+
+1.   [=list/For each=] |point| of |linearEasingFunction|'s [=linear easing function/points=]:
+
+     1.   If |point|'s [=linear easing point/position=] is less than or equal to |inputProgress|, set |previousPoint| to |point|.
+
+     1.   Otherwise, set |nextPoint| to |point| and [=break=].
+
+1.   If |previousPoint| is null, then return |nextPoint|'s [=linear easing point/value=].
+
+1.   If |nextPoint| is null, then return |previousPoint|'s [=linear easing point/value=].
+
+1.   Let |progressBetweenPoints| be (|inputProgress| - |previousPoint|'s [=linear easing point/position=]) / (|nextPoint|'s [=linear easing point/position=] - |previousPoint|'s [=linear easing point/position=]).
+
+1.   Return |previousPoint|'s [=linear easing point/value=] + |progressBetweenPoints| * (|nextPoint|'s [=linear easing point/value=] - |previousPoint|'s [=linear easing point/value=]).
+
+</section>
+
+<dfn function lt="linear()">linear(<<linear-stop-list>>)</dfn>
+
+<pre class="prod">
+  <dfn>&lt;linear-stop-list></dfn> = [ <<linear-stop>> ]#
+  <dfn>&lt;linear-stop></dfn> = <<number>> && <<number>>?
+</pre>
+
+<h3 id="the-linear-easing-keyword">The linear easing keyword: ''linear''</h3>
+
+The <dfn dfn-type=value for=easing-function>linear</dfn> keyword is equivalent to ''linear()''.
 
 <h3 id=cubic-bezier-easing-functions oldids=cubic-bezier-timing-functions>Cubic
 B&eacute;zier easing functions:
--- a/css-easing-1/Overview.bs
+++ b/css-easing-1/Overview.bs
@@ -136,47 +136,41 @@ A <dfn>linear easing point</dfn> is a [=/struct=] that has:
 
 <dl dfn-for="linear easing point">
 
-: <dfn>position</dfn>
+: <dfn>input</dfn>
 :: A number
-: <dfn>value</dfn>
+: <dfn>output</dfn>
 :: A number
 
 </dl>
 
 <section algorithm="to calculate linear output progress">
 
-To <dfn>calculate linear output progress</dfn> for a given [=linear easing function=] |linearEasingFunction|, and an [=input progress value=] |inputProgress|, perform the following steps. They return an [=output progress value=].
+To <dfn export>calculate linear output progress</dfn> for a given [=linear easing function=] |linearEasingFunction|, and an [=input progress value=] |inputProgress|, perform the following. It returns an [=output progress value=].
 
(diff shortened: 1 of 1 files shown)


tabatkins:
Syntax nit: I'd probably use a % for the progress value. That accords better with things like gradients, and makes it *much* clearer which value is progress and which is the value:

`linear(0 0%, 0.1 20%, 0.3 40%, 0.7 60%, 0.8 80%, 1 100%)`

It also means we wouldn't necessarily have to impose an ordering, since the types are distinguishable.

----

I *really* don't like the different behavior between `linear()` and `linear(0,1)`. If we just always made <0 and >1 extend using the first/last segment, you could achieve the clamping behavior with `linear(0 calc(infinity * -1%), 0 0%, 1 100%, 1 calc(infinity * 1%))`. We could probably add some syntax to make that easy and explicit instead, like `linear(0 0% fill, 1 100% fill)`, where the `fill` keyword is only valid on the first/last value, and has the same approximate meaning as `animation-fill-mode`.

-----

Otherwise, big 👍 on the proposal overall. I don't think the spec needs to be *quite* as algorithmic as written here, but the behavior is fine.

okikio:
I personally prefer the `linear-easing(...)` style, it seems easier to understand.

birtles:
As I've said before, I really like this proposal. Thanks for pushing this forward.

I'm a bit dense so I had a bit of trouble following the description in the first comment however because there are many different terms:

* "value" (A)
* "progress position" (B)
* "input progress value" (=A? =B?)
* "position value" (=A? =B?)

I think I might be able to better understand the clamping issue if the terms were more consistent.

okikio:
Yeah, I agree with @birtles. 

If the goal is to enable the creation of complex easings then I feel we should keep the key word `linear-easing(....)` or `custom-easing(....)`, with `linear-easing(0, 0.4, 0.6, ...)` being the basic form and `linear-easing(0 20%, 3 40%, 0.5 80%, 0.6, 0.7, 0.8, 1)` being the advanced mode. I don't think we should complicate the `linear-easing` function with `fillMode`, as it would make the `linear-easing` function the only CSS Easing Function that currently does this and create potentially counter intuitive behaviors, I feel a separate proposal for `fillMode` would be better, so, for example, we could have something like this,
```scss
.div {
    animation-fill-mode: both 0%, forwards 50%, none 80%, backwards 90%;
}
```
As an experiment, I followed the technique @jakearchibald suggested and I was able to create `spring` easing using WAAPI, so, I think the proposal just in the basic form is plenty powerful and versatile, plus together with `progress values as percentages`, it would be even easier to understand what is happening. One thing to note is that `linear-easing` would be pretty difficult to follow if the points of said easing are all over the place, I think some amount of imposed order is beneficial.

jakearchibald:
@tabatkins 

> Syntax nit: I'd probably use a % for the progress value. That accords better with things like gradients, and makes it _much_ clearer which value is progress and which is the value:
> 
> `linear(0 0%, 0.1 20%, 0.3 40%, 0.7 60%, 0.8 80%, 1 100%)`

I had the same thought, but here's why I ended up with two numbers: Both values are the same type, that's how the simple identity version of the `linear` works. So, they could both be %, or both be numbers. I went with numbers because that's what the spec already uses for cubic-bezier and [all the diagrams](https://drafts.csswg.org/css-easing/curve-keywords.svg).

I agree it _looks_ better with different types, but I'm not sure it makes sense. If we go with one value being a number, and the other being a %, I guess we should update the diagrams in the spec too to reflect that.

> It also means we wouldn't necessarily have to impose an ordering, since the types are distinguishable.

This gets really weird when trying to generate auto values for input position. I think it's why gradient syntax doesn't allow it (well, if one percent is less than the previous one, it makes it equal to the previous one). Being consistent with gradient syntax here seems better.

> I _really_ don't like the different behavior between `linear()` and `linear(0,1)`. If we just always made <0 and >1 extend using the first/last segment, you could achieve the clamping behavior with `linear(0 calc(infinity * -1%), 0 0%, 1 100%, 1 calc(infinity * 1%))`. We could probably add some syntax to make that easy and explicit instead, like `linear(0 0% fill, 1 100% fill)`, where the `fill` keyword is only valid on the first/last value, and has the same approximate meaning as `animation-fill-mode`.

We could borrow more from gradient syntax here:

```css
.whatever {
  background: linear-gradient(to right, red 20%, orange 20% 40%, yellow 40% 60%, green 60% 80%, blue 80%);
}
```

Here, `orange 20% 40%` effectively creates two 'orange' points, one at 20% and one at 40%. We could do the same.

```css
.whatever {
  /* No clamping */
  animation-timing-function: linear(0, 1);
  /* Clamping */
  animation-timing-function: linear(0 0 0, 1 1 1);
}
```

The latter creates 4 points, two where the input and output are 0, and two where the input and output are 1. Having two points means the values would 'extend' flatly. This makes it relatively easy to clamp just one side of the easing.

The syntax also allows for creating a 'rest' period during the easing:

```css
.whatever {
  animation-timing-function: linear(0, 1 .25 .75, 0);
}
```

> I don't think the spec needs to be _quite_ as algorithmic as written here

That's my lack of experience with CSS specs showing :smile:

jakearchibald:
@birtles 

> * "value" (A)
> * "progress position" (B)
> * "input progress value" (=A? =B?)
> * "position value" (=A? =B?)

It terms of [these diagrams](https://drafts.csswg.org/css-easing/curve-keywords.svg), a point's 'position' is the x-axis, and a point's 'value' is the y-axis.

Happy to rename these to whatever.

jakearchibald:
@okikio 

> If the goal is to enable the creation of complex easings then I feel we should keep the key word `linear-easing(....)` or `custom-easing(....)`, with `linear-easing(0, 0.4, 0.6, ...)` being the basic form and `linear-easing(0 20%, 3 40%, 0.5 80%, 0.6, 0.7, 0.8, 1)` being the advanced mode

That isn't consistent with gradients. I'm trying to borrow as much as possible from existing concepts.

> I don't think we should complicate the `linear-easing` function with `fillMode`, as it would make the `linear-easing` function the only CSS Easing Function that currently does this and create potentially counter intuitive behaviors, I feel a separate proposal for `fillMode` would be better,

Don't think of this being like fillmode as it exists in other parts of the animation spec, it's pretty different. Does the idea in https://github.com/w3c/csswg-drafts/pull/6533#issuecomment-903086012 make more sense?

birtles:
> @birtles 
> 
> > * "value" (A)
> > * "progress position" (B)
> > * "input progress value" (=A? =B?)
> > * "position value" (=A? =B?)
> 
> It terms of [these diagrams](https://drafts.csswg.org/css-easing/curve-keywords.svg), a point's 'position' is the x-axis, and a point's 'value' is the y-axis.
> 
> Happy to rename these to whatever.

Thanks! I'm just wondering which "input progress value" and "position value" are.

jakearchibald:
Input progress value is an [existing term](https://drafts.csswg.org/css-easing/#input-progress-value). "Position value" isn't in this proposal or the existing spec. Might be easier to make a comment against a particular line in the PR?

birtles:
> Input progress value is an [existing term](https://drafts.csswg.org/css-easing/#input-progress-value). "Position value" isn't in this proposal or the existing spec. Might be easier to make a comment against a particular line in the PR?

Sorry I didn't realise this was a PR! I was just responding to the first comment here which I thought was the proposal and which uses terms like "position value". I'll have to look at the PR next week.

tabatkins:
> I feel we should keep the key word linear-easing(....)

The reason I don't like having `-easing()` in the name is because the existing easing keywords and functions don't use that suffix either. A *custom* easing (as in, backed by JS or something as a Houdini API) would probably need a name like `custom-easing()`, but for the built-ins, `linear-easing()` looks a little weird alongside `linear`, `ease-in`, and `steps()`.

> I don't think we should complicate the linear-easing function with fillModeI don't think we should complicate the linear-easing function with fillMode

I'm not saying it interacts with fill mode. I'm saying that we can use the *analogy* of fill mode for the keyword that extends the linear easing backwards/forwards infinitely. There might be a better keyword, it was just my first suggestion.

> So, they could both be %, or both be numbers. I went with numbers because that's what the spec already uses for cubic-bezier and all the diagrams.

cubic-bezier isn't *quite* doing percentages, tho. (well, all numbers are percentages, just with the decimal point moved, but you know what I mean)  Placing points between "start" and "end" *is* a percentage, tho, and we use that type to indicate the exact same semantic for animations and gradients. We should stick with it, both for consistency *and* readability. (I literally couldn't tell what value was which in your example, and had to read your prose to figure out that you meant for the progress values to be evenly spaced, meaning the .2/.4/.6/.8 must be the progresses.)

> This gets really weird when trying to generate auto values for input position. I think it's why gradient syntax doesn't allow it (well, if one percent is less than the previous one, it makes it equal to the previous one). Being consistent with gradient syntax here seems better.

Sorry, you misunderstood me. Using different types means that ordering *in each pair* doesn't need to be specified - you could write `20% .3` or `.3 20%`. Between pairs we obviously still need ordering, for all the reasons you give.

> [stuff about doubling up progress value]

Yeah that works nicely, and I agree that in general having doubled progresses is useful just as it is for gradients. We'd need to call this technique out explicitly in the spec as a way to get "flat" values at the start/end, but I like when things come for free. ^_^

Note that this makes the values *even less readable* when they're all the same type; percentages are an absolute necessity here.

jakearchibald:
> Placing points between "start" and "end" _is_ a percentage

Isn't the other value also a percentage by that logic? Where 0% is the start value and 100% is the end value.

We could make it like opacity where 0-1 and 0%-100% mean the same thing, and folks could use either. That means you need ordering within the pair though.

> We'd need to call this technique out explicitly in the spec as a way to get "flat" values at the start/end

I don't have a gut feeling for what "expected" behaviour is here, but if folks expect flat at the start/end maybe we should make that the default.

tabatkins:
Yeah like I said, *all* numbers can be percentages, *especially* when they're typically in the [0,1] range.

But we have precedence that progress is written as a % (from both animations and gradients), and forcing it to be a % while the other is a number gives us useful readability benefits, whereas making them both the same type (or allowing both to be either type) makes it *harder* to read.

> I don't have a gut feeling for what "expected" behaviour is here, but if folks expect flat at the start/end maybe we should make that the default.

Problem is that if we do that, then we still need to add a *separate* mechanism to make it extendable so you can get the same behavior as `linear`. Out-of-bounds progress is mostly a corner-case anyway; unless there's a great reason to do one or other specifically, I'd prefer we go with the behavior that gives us the simplest syntax. (Which means, currently, I prefer "extend the first/last segment indefinitely, and you can get flat manually without much effort by just making the first/last segment flat".)

jakearchibald:
Fair points all around, I'm convinced 😀

jakearchibald:
The nice thing about this is you can now explain step() using linear(). You can also create stepped easings that are non-linear.

birtles:
I finally had a look at the PR (sorry, I completely missed that this was a PR the first time around).

First, a minor process point: CSS Easing Level 1 is in CR so this will need to go in Level 2. Tab can tell us how best to proceed with that.

Secondly, an even more pedantic point just to get this out of the way, @fantasai once taught me that it's helpful to break up lines quite a bit, roughly one line per clause so that it makes it easier to follow the blame log later.

e.g. instead of 

```bikeshed
A [=linear easing function=] has <dfn for="linear easing function">points</dfn>, a [=/list=] of [=linear easing points=]. Initially a new empty [=/list=].
```

```bikeshed
A [=linear easing function=] has <dfn for="linear easing function">points</dfn>,
a [=/list=] of [=linear easing points=].
Initially a new empty [=/list=].
```

The long lines also make it a bit hard to review this in GitHub.

On to the content itself:

* I agree with Tab about keeping `linear` and `linear(0,1)` consistent. I think we've generally found when chaining these functions together that extrapolating tends to produce the most natural results. That's why [the algorithm for cubic bézier easing functions](https://drafts.csswg.org/css-easing/#cubic-bezier-algo) tries to find the tangent at the ends and use that for extrapolation. I'm pretty sure it was the Chrome team discovered that works best.

   Doing that would mean, I think, that `linear(0.4 40%, 0.6 60%)` should be the same as `linear(0, 1)`. If you want the straight line version you'd have `linear(0.4, 0.4 40%, 0.6 60%, 0.6)` (assuming the missing position values are filled in using the same approach we use for missing keyframe offsets).

* Speaking of which, does the spec need to define how the second `<<number>>` is calculated when not specified? And the serialization of the function?

* I love that you made the algorithm follow the standard end-point exclusive timing. That is, when two or more points overlap, you made it choose the _last_ point to start for. Thank you!

jakearchibald:
I've updated the initial comment to reflect feedback. I've changed the spec to follow, and made it less algorithmic as @tabatkins suggested. I hope it still makes sense! I guess it would benefit from diagrams like I've included in the OP.

> * Speaking of which, does the spec need to define how the second `<<number>>` is calculated when not specified? And the serialization of the function?

Yes, I'm not really sure how to write that down, but we need something that converts the notation into a set of points. Also, `calculate linear output progress` needs connected up so it's actually used.

I tried to copy the syntax from `linear-gradient`, but I couldn't see where things like `yellow 40% 60%` are allowed. I'm looking at [the syntax](https://drafts.csswg.org/css-images-3/#typedef-color-stop-list), but I can't see how it allows three values

> Secondly, an even more pedantic point just to get this out of the way, @fantasai once taught me that it's helpful to break up lines quite a bit, roughly one line per clause so that it makes it easier to follow the blame log later.

I've given this a go (although I generally find it more of an authoring burden vs any later benefit). However, it isn't the style used throughout the rest of the document.

Afif13:
I recently wrote an article about how to approximate such curves using CSS variables and CSS houdini (wihout any JS): https://css-tricks.com/build-complex-css-transitions-using-custom-properties-and-cubic-bezier/ .. It might be helpful until we have such feature.

dholbert:
Oops, sorry -- I clicked the "review" button in an attempt to read the latest version of the PR, and that seems to have actually been a 1-click button to change the state of things here. (I see "dholbert marked this pull request as ready for review" in the comment history now.)

I don't see an "oops, undo" button, or else I would click it.  Sorry for muddling with things; feel free to revert my marking or whatever's appropriate.

 (I'm an infrequent github user and trip over things now and then.)

tabatkins:
> I tried to copy the syntax from linear-gradient, but I couldn't see where things like yellow 40% 60% are allowed. I'm looking at the syntax, but I can't see how it allows three values.

You want Images 4, not 3: <https://drafts.csswg.org/css-images-4/#color-stop-syntax>. The text is pretty simple:

> A color stop with two positions is equivalent to specifying two color stops with the same color, one for each position.

birtles:
For the unspecified input progress case, I think it would be good to follow the keyframe spacing behavior here, i.e. `linear(0.2, 0.3, 0.5 50%, 0.7 70%, 0.2)` is computed to `linear(0.2 0%, 0.3 25%, 0.5 50%, 0.7 70%, 0.2 100%)`.

Unfortunately the [text for that in Web Animations](https://drafts.csswg.org/web-animations-1/#computed-keyframe-offset) is _very_ algorithmic and we could probably get by with a much more simple prose description for the same here.

Here's an initial (fairly unsuccessful) attempt:

> The input progress value for points without a specified input progress value is computed such that they are spaced equidistantly between points _with_ specified input progress values whilst ensuring that the computed input progress value of the first point in the series of points is the minimum of zero (0) or the first specified input progress value, and the computed input progress value of the last point in the series of points is the maximum of one (1) or the last specified input progress value.

tabatkins:
We can copy the gradient text: https://drafts.csswg.org/css-images-4/#color-stop-fixup

(Notably, we'll still need step 2, because calc() means we can't always detect whether stops are in the correct order at parse-time.)

jakearchibald:
> We can copy the gradient text: https://drafts.csswg.org/css-images-4/#color-stop-fixup

Oh ffs, I totally missed that. I got to https://drafts.csswg.org/css-images-4/#linear-gradients which suggests there's no change in the notation.

jakearchibald:
Ok, I've updated the syntax, and detailed how the syntax becomes a callable function. However, I'm not sure how to hook it all up from here.

birtles:
> However, I'm not sure how to hook it all up from here.

I think we need to create an easing level 2 spec first. Given that level 1 is in CR, I think we could just copy all the text from there into level 1 without bothering with a delta spec. Tab might have a better suggestion though.

torickjdavis:
I'm coming to this thread via the [HTTP 203 Video "Bringing bounce and elastic easing to CSS"](https://www.youtube.com/watch?v=8FuafvJLDpM&ab_channel=GoogleChromeDevelopers). One comment I made there, I thought might have more benefit here:

As an alternative to overloading `linear`, would it make sense to call it `lerp` (linear interpolation)?

SebastianZ:
> As an alternative to overloading `linear`, would it make sense to call it `lerp` (linear interpolation)?

There are two reasons why I wouldn't prefer this. Firstly, abbreviations are very uncommon in CSS. And secondly, the term `lerp` is probaby unknown to most web authors. I'm not sure whether `linear` is the best name for it, though, either.

Sebastian

nohns:
Saw a comment from Viktor Luft on the Youtube video [Bringing bounce and elastic easing to CSS | HTTP 203](https://www.youtube.com/watch?v=8FuafvJLDpM), which made good sense to me:

> I'd call it "linear-spline" instead of "linear" to match the mathematical notation. I really like the idea

Also another comment, from victornbp, highlighted the possibility of adding smoothing to the function, reducing the amount of points need to form beautiful easings:

> I think this could be easily expanded to have smoothed functions, the same way graphing libraries do point interpolation (quadratic, log, etc). This would probably solve long animations without having to have like 500 points.

I think these points worth taking into consideration when moving forward with this proposal. The syntax for smoothing could be something like:

```css
smooth-spline(...)
```

okikio:
> ```css
> animation-timing-function: linear-spline(0, 0.2, 0.5, 0.9, 1.1, .0.95, 1);
> animation-timing-function: smooth-spline(0, 0.2, 0.5, 0.9, 1.1, .0.95, 1);
> ```

I like @nohns idea, I think it makes lots of sense, how does everyone else feel?

mauriciabad:
There's a thing I don't see explained in the spec (I'm not used to `.bs` files, so maybe it's there, but I don't see it):

What happens when 2 points have the same `x` coordinate? 
**e.g. `linear(0.1 50%, 0.9 50%)`** (or even `linear(0.1 50%, 0.1 50%)`)
 
The algorithm only considers the first appearing point in the list, so the example is equal to `linear(0.1 50%)`?

If yes, keep in mind that then, `linear(0.1 50%, 0.9 50%)` becomes that special case where there's only 1 point, but the original list had more than 1 point.

BTW (off-topic), I love this proposal because it is simple to understand and very aligned with the "typical" pure css approach to do complex easings. I wanted custom easings so bad that I created my own tool to make them using keyframes (https://easyeasings.com/). I'm so happy to see progress on this, so I can stop using pesky keyframe animations!

jakearchibald:
The algorithm takes care of this. Values cannot go backwards. If it's less than the previous value, it's clamped to the previous value. It's the same behaviour as linear-gradient.

jakearchibald:
Context for an upcoming working group meeting:

This is a proposal for a way to create custom easings, such as bounce, and elastic.

Easings are created using a syntax very similar to `linear-gradient()`:

```css
.whatever {
  animation-timing-function: linear(0, 0.7, 0.87, 0.98, 1);
}
```

This would create the following easing:

Screenshot 2021-08-23 at 13 11 21

The input values are distributed automatically, but can also be manually specified, as with `linear-gradient`. For example, the above is equivalent to `linear(0 0%, 0.7 25%, 0.87 50%, 0.98 75%, 1 100%)`.

Like `linear-gradient`, output values are linearly interpolated from one value to the next.

By providing enough points, developers can specify easings that are currently impossible, such as bounce.

```css
.whatever {
  animation-timing-function: linear(0, 0.003, 0.013, 0.03, 0.05, 0.08, 0.11, 0.15, 0.2, 0.26, 0.31, 0.38, 0.45, 0.53, 0.62, 0.71, 0.81, 0.9, 0.99, 0.94, 0.89, 0.85, 0.82, 0.79, 0.77, 0.76, 0.752, 0.75, 0.755, 0.77, 0.78, 0.81, 0.84, 0.87, 0.92, 0.97, 0.99, 0.97, 0.95, 0.94, 0.938, 0.94, 0.949, 0.96, 0.99, 0.994, 0.986, 0.985, 0.989, 1 100% 100%);
}
```

…which would graph like this:

Screenshot 2021-08-23 at 11 45 53

The door would remain open for a more advanced syntax that allowed for curves to be specified, hopefully in a way that also works for `linear-gradient`. However, attempts to introduce that kind of complexity initially have stalled previous efforts, eg https://github.com/w3c/csswg-drafts/issues/229#issuecomment-861247074, so this proposal limits itself to linear interpolation between points.

chrishtr:
Agenda+ to discuss and hopefully resolve to add in 2 weeks (can't be an APAC meeting since Jake is in Europe).

birtles:
> Agenda+ to discuss and hopefully resolve to add in 2 weeks (can't be an APAC meeting since Jake is in Europe).

I won't be able to make the telcon (I'm in APAC) but as the only(?) active editor of that spec I'm strongly in support of this proposal. It's something I've been suggesting for years and I'm really grateful Jake has done the hard work of actually turning this into a PR.

flackr:
Not one of the editors, but I just want to say I completely agree with @birtles. This is a valuable addition to the platform and I like the alignment with gradient easing. Thanks Jake!

svgeesus:
> Context for an upcoming working group meeting:

I would hope those examples and diagrams would also make it into the spec?

fantasai:
I think this is a good idea, but it needs to be  PR against css-easing-2, since css-easing-1 is already in CR and widely deployed. (Which also means css-easing-2 needs to exist.)

jakearchibald:
> Which also means css-easing-2 needs to exist.

I don't have experience with level'd specs so I'm not sure what the process is. I guess someone else will set that up, then I'll create a new PR against that spec?

okikio:
> 

Awesome

ydaniv:
I think it would be great to also have `bounce` and `elastic` keywords along with this addition, as an OOTB sensible defaults for the "80%". Kind of like `cubic-bezier()` has its own with `ease` and `ease-*`.
Also since the 2 most probable use cases are `bounce` and `elastic`, and as @jakearchibald noted above, we'll probably start seeing the same value copied all over as a custom property value, so why not have 2 nice looking defaults already spec'd and shipped inside the browsers?

css-easing-1/Overview.bs
```diff
@@ -125,16 +125,126 @@ The syntax for specifying an [=easing function=] is as follows:
...
+
+               1.   Set |extraPoint|'s [=linear easing point/output=] to |stop|'s <<number>>.
+
+               1.   Set |extraPoint|'s [=linear easing point/input=] to whichever is greater:
+                    |stop|'s <<linear-stop-length>>'s second <<percentage>> as a number,
+                    or |largestInput|.
+
+               1.   Set |largestInput| to |extraPoint|'s [=linear easing point/input=].
+
+     1.   Otherwise, if |stop| is the first [=list/item=] in |stopList|,
```
dshin-moz, line 232: Hm, admittedly this is an edge case, but if any of the entries not at either edge contain negative or higher-than-100% percentage values (`linear(0, 0.2 -10%, 0.8 110%, 1.0)`), wouldn't it break the monotonicity?

css-easing-1/Overview.bs
```diff
@@ -125,16 +125,126 @@ The syntax for specifying an [=easing function=] is as follows:
...
+          then set |point|'s [=linear easing point/input=] to 0.
+
+     1.   Otherwise, if |stop| is the last [=list/item=] in |stopList|,
+          then set |point|'s [=linear easing point/input=] to 1.
+
+1.   For runs of [=list/items=] in |function|'s [=linear easing function/points=] that have a null [=linear easing point/input=],
+     assign a number to the [=linear easing point/input=] by linearly interpolating between the previous and next [=linear easing function/points=] that have a non-null [=linear easing point/input=].
+
+1.   Return |function|.
```
dshin-moz, line 241: Serializing back from this returned function may result in a different round trip result: `linear(0 0% 50%, 1 50% 100%)` -> `linear(0 0%, 0 50%, 1 50%, 1 100%)`.  Do we want to specify how the serialization should be carried out?

jakearchibald: What happens with linear-gradient? If there's prescient there, same should happen here.

css-easing-1/Overview.bs
dshin-moz: That's true on the start side - how about the end, `linear([...], 0.8 110%, 1.0)`? Could use `|largestInput|`, though each side behaving differently seems a bit less intuitive (Debatable, since we're already in an edge case).

jakearchibald: I think if the last value doesn't have a position, it should be assumed to be 100%. Does that make sense?

jakearchibald: Actually, let me test gradients…

jakearchibald: https://jsbin.com/mopojul/4/edit?html,css,output

It seems like the first value is assigned 0% if it has no position. However, it seems like the last position will be `Math.max(previousPosition, 100%)`.

My intent will be to spec the same position behaviour as gradients unless there's a good reason to do something different.

css-easing-1/Overview.bs
dshin-moz: `linear-gradient(0deg, red 0% 10%, green, blue 90%);` seems to become `llinear-gradient(0deg, rgb(255, 0, 0) 0%, rgb(255, 0, 0) 10%, rgb(0, 128, 0), rgb(0, 0, 255) 90%);`.
That said, not sure if I 100% agree on keeping it the same as `linear-gradient`:
* There may be a lot more stops being supplied for the easing function in general
* `linear-gradient`'s `function` equivalent is really called once for paint vs. every frame of animation for this easing function. 

So the two options of keeping both presentable & function forms, or calculating the function form every time seem... unattractive.

css-easing-1/Overview.bs
dshin-moz: Sounds reasonable. Thanks!

css-easing-1/Overview.bs
jakearchibald: I agree it's a little weird that syntax unfolds like that, but I think it still makes sense to do the same as gradient.

PR #6534, opened 2021-08-20
[css-highlight-api-1] Specify that custom highlights have no UA styles #6375

### LINKED ISSUE #6375: [css-highlight-api] Need clarification on expected default values for style properties
I didn't find any statement on default values for style properties of ::highlight in the [spec draft](https://drafts.csswg.org/css-highlight-api-1/). There are some default values in the [spec for pseudos](https://drafts.csswg.org/css-pseudo-4/#highlight-ua-styles ) but ::highlight is not in the list. Also in the [spec draft of ::highlight](https://drafts.csswg.org/css-highlight-api-1/) there's no mention of it. 

Should it use no visual effects or should it use for example [Highlight/HighlightText](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/layout/layout_theme.cc;l=593;drc=c06511a314747fab345af58b8523eec1b36caf05;bpv=1;bpt=1) colors instead? I assume it's taking no visual effects as default values (as in when background-color is not specified inside a ::highlight(foo) then the content's background is the same as if there was no ::highlight defined), but maybe it should be explicitly stated?

dandclark:
Per resolution of #6375, specify that:
- UAs do not provide any default styles for custom highlights.
- Custom highlights inherit the styles of their originating element.

Closes #6375.

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -298,8 +298,8 @@ Applicable Properties</h4>
 <h4 id=default-styles>
 Default Styles</h4>
 
-	UAs must not define any styles for [=Custom highlight pseudo-elements=] in the default UA
-	stylesheet. A [=Custom highlight pseudo-element=] inherits the styles of its
+	UAs must not define any styles for [=custom highlight pseudo-elements=] in the default UA
+	stylesheet. A [=custom highlight pseudo-element=] inherits the styles of its
 	[=originating element=].
 
 <h4 id=c-and-h>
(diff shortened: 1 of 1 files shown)


css-highlight-api-1/Overview.bs
```diff
@@ -295,6 +295,13 @@ Applicable Properties</h4>
 	can only be styled with a limited set of properties.
 	See [[css-pseudo-4#highlight-styling]] for the full list.
 
+<h4 id=default-styles>
+Default Styles</h4>
+
+	UAs must not define any styles for [=Custom highlight pseudo-elements=] in the default UA
```
sanketj, line 301: nit: lowercase 'c' in custom

css-highlight-api-1/Overview.bs
```diff
@@ -295,6 +295,13 @@ Applicable Properties</h4>
...
+	stylesheet. A [=Custom highlight pseudo-element=] inherits the styles of its
```
sanketj, line 302: nit: lowercase 'c' in custom

sanketj (approved):

Issue #3526, opened 2019-01-17
[css-ui][html] Layering concerns for 'appearance'


zcorpan:
https://drafts.csswg.org/css-ui-4/#appearance-switching

The button value, if applicable to all elements, probably needs to be specified in css-ui what it does. Other values might be specified in HTML's rendering section.

There are also properties that turn off the appearance (e.g. an author-specified border on a button). In Firefox this applies to appearance values regardless of element, in Chromium this depends on the element. (Specify in HTML?)

@annevk raised in a conversation today that the layering here should be discussed. Can you elaborate here?

cc @bzbarsky @frivoal @fantasai @MatsPalmgren @tkent-google

annevk:
What I'm worried about is that on the HTML side we'll have this thing where we branch on element (and its attributes) and decide on a render algorithm. This render algorithm might then further branch on appearance. On the CSS side we might use appearance to decide on a render algorithm for the remaining elements. That feels a little circular and depending on how you resolve it you either end up with the Chromium or Firefox behavior.

I guess what you want is some kind of algorithm that takes in the element and its set of properties, and then returns what kind of thing to render. However, that means CSS ends up depending on some HTML-specific thing or we'd define the whole thing in HTML.

fantasai:
We don't have to define the whole thing in HTML, we just need to define as much as we can that's consistent in CSS with some explanation of the various classes of exceptions, and then HTML can specify which elements trigger which exceptions. This isn't too different from how we define pseudo-classes.

For convenience in developing the spec, I've suggested we keep the per-element details in a separate chapter of css-ui-4, then once it's stabilized we can move it over to HTML. But `appearance` and its values should all be thoroughly defined in CSS UI.

frivoal:
@zcorpan, @annevk Between the state of the css-ui-4 spec, and the direction the HTML spec is taking with https://github.com/whatwg/html/pull/4857, I think the layering concern question is answered. Not all the pieces have fallen into place yet, but we seem to know where we're going.

Are you ok with closing this issue, or am I missing something that still needs to be discussed here?

annevk:
If someone could summarize the agreed upon layering that'd help. (Or point to where that was discussed.)

zcorpan:
I don't think this is addressed, yet.

zcorpan:
[...]

Issue #356, opened 2016-07-28
[css-ui] appearance: none on <meter>


gregwhitworth:
@LeaVerou pointed out that Chrome 52 now removes the meter visualization for when `-webkit-appearance: none` is set. The spec doesn't seem to state what happens with `<meter>` but we should probably state what occurs. Reading through the spec this seems to be ok based on this text:

> All decorative visual aspects of a form control which are not caused by a CSS style rule must be suppressed when the appearance is changed using appearance, even if the UA’s internal representation for this element was composed of multiple elements or pseudo elements combined together. For example, select is often rendered with an arrow on the right side indicating that the list will be expanded if the element is clicked. If the computed value of appearance is none, this must disappear.

But since `<meter>` is used to present information visually this seems to contradict with this part of the spec:

> However, the UA must preserve aspects of the form control which are necessary to operate the control with its original semantics.

It would be good to get some feedback from other implementers and Google on why they made this change and its potential impact.

[Chrome status of the change](https://www.chromestatus.com/feature/5668635896971264)

RByers:
@tkent-google made this change after discussing it with @frivoal and [posting the plan to www-style](https://lists.w3.org/Archives/Public/www-style/2016Apr/0394.html).  Does the spec need some improvement here?

gregwhitworth:
Yes, a few things, based on my statement above they seem to contradict each other since <meter> is a visual representation, that is essentially its "operation". Yes <meter> can have fallback content, but is this how it is actually used in the wild - even if it's overall usage is low?

Additionally, before making a change that makes them disappear we should probably have a discussion on standardizing the styling of the <meter> parts to provide webdevs with a CSS approach to this rather than force them to use JS for a crossbrowser approach. I'm aware of some `-moz` and `-webkit` based psuedos that allow you to target the inner guts of the meter.

tkent-google:
The motivation of Chrome's behavior change was to fix http://crbug.com/51182.
All of major browsers have similar horizontal bar representation of <meter>. But web authors want to render completely different representation such as rounded speedometer.  We needed a way to stop the horizontal bar representation, and (-webkit-)appearance:none was suitable.

If there are other better idea to provide both of UA-default rendering and full-customization, I'd like to follow it.

LeaVerou:
FWIW, the bug I pointed out was that the `<meter>` disappeared entirely (including its styling) when using `-webkit-appearance: none`, which I believe is undoubtedly a regression.

[...]

PR #6537, opened 2021-08-23
[css-ui-4] Define how to compute the kind of widget to use for an element

howard-e:
[css-ui-4] Define how to compute the kind of widget to use for an element. Relevant to https://github.com/w3c/csswg-drafts/issues/3526#issuecomment-689146382.

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -1943,6 +1943,66 @@ meaning that they are rendered like analogous native widgets or controls of the 
 or with a look and feel not otherwise expressible in CSS.
 It is up to the host language (e.g., HTML [[HTML]]) to define which elements can have <a>native appearance</a>.
 
+The following elements can have a <a>native appearance</a> for the purpose of the 'appearance' property:
+
+<ul>
+    <li><{button}>
+    <li><{input}>
+    <li><{meter}>
+    <li><{progress}>
+    <li><{select}>
+    <li><{textarea}>
+</ul>
+
+The properties that disable <a>native appearance</a> for <a>widgets</a> are:
+
+<ul>
+    <li>'background-color'
+    <li>'border-top-color'
+    <li>'border-top-style'
+    <li>'border-top-width'
+    <li>'border-right-color'
+    <li>'border-right-style'
+    <li>'border-right-width'
+    <li>'border-bottom-color'
+    <li>'border-bottom-style'
+    <li>'border-bottom-width'
+    <li>'border-left-color'
+    <li>'border-left-style'
+    <li>'border-left-width'
+    <li>'border-block-start-color'
+    <li>'border-block-end-color'
+    <li>'border-inline-start-color'
+    <li>'border-inline-end-color'
+    <li>'border-block-start-style'
+    <li>'border-block-end-style'
+    <li>'border-inline-start-style'
+    <li>'border-inline-end-style'
+    <li>'border-block-start-width'
+    <li>'border-block-end-width'
+    <li>'border-inline-start-width'
+    <li>'border-inline-end-width'
+    <li>'background-image'
+    <li>'background-attachment'
+    <li>'background-position'
+    <li>'background-clip'
+    <li>'background-origin'
+    <li>'background-size'
+    <li>'border-image-source'
+    <li>'border-image-slice'
+    <li>'border-image-width'
+    <li>'border-image-outset'
+    <li>'border-image-repeat'
+    <li>'border-top-left-radius'
+    <li>'border-top-right-radius'
+    <li>'border-bottom-right-radius'
+    <li>'border-bottom-left-radius'
+    <li>'border-start-start-radius'
+    <li>'border-start-end-radius'
+    <li>'border-end-start-radius'
+    <li>'border-end-end-radius'
+</ul>
+
 This specification introduces the 'appearance' property
 to provide some control over this behavior.
 In particular, using ''appearance: none'' allows authors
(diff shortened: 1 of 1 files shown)


css-ui-4/Overview.bs
```diff
@@ -2166,6 +2166,129 @@ Document languages should specify
 for each <a>widget</a> that they define
 what needs to be preserved.
 
+The possible kinds of widgets are ''appearance/none'', ''button'', ''searchfield'', ''textfield'',
```
zcorpan, line 2169: These are not CSS property values, and so shouldn't use `''foo''` syntax. They should be defined terms. Something like this:

```
The possible <dfn lt="kind of widget|kinds of widgets">kinds of widgets</dfn> are <dfn for="kind of widget" export>none</dfn>, ...
```

The usages of those terms would then be `[=kind of widget/none=]`

css-ui-4/Overview.bs
```diff
@@ -2166,6 +2166,129 @@ Document languages should specify
...
+''slider-horizontal'', ''checkbox'', ''radio'', ''listbox'', ''menulist'', ''menulist-button'',
+''textarea'', ''progress-bar'', ''meter''.
+
+<div algorithm="compute kind of widget">
+    UAs must run this at [=used value=] time: compute the kind of widget to use for an element, |element|:
+
+    <ol>
+        <!-- none always works -->
+        <li>
+            If |computedAppearance| is ''appearance/none'', then return ''appearance: none''.
```
zcorpan, line 2179: This step needs to be moved down so it's after the step that defines *computedAppearance*. (This was my mistake, I only noticed it now when reviewing this :) )

css-ui-4/Overview.bs
```diff
@@ -2166,6 +2166,129 @@ Document languages should specify
...
+            Let |computedAppearance| be the [=computed value=] of |element|'s 'appearance' property.
+
+        <li>
+            Let |authorProps| be a map of all properties that have a <a>cascaded value</a> for
+            |element|, where the [=Author Origin=], [=Animation Origin=] and [=Transition Origin=]
+            [=Cascade Origins=] are included and the [=User Origin=] and [=UA Origin=] [=cascade origins=]
+            are ignored, where the key is the property name and the value is that <a>cascaded value</a>.
+
+        <li>
+            If |element| can be a ''button'', given |computedAppearance|, then return the appropriate widget with
```
zcorpan, line 2191: My idea here is that "can be a button" should be a defined term in HTML, and linked from CSS UI.

Maybe the term needs to be defined in CSS UI itself, and it can say that it's up to the host language to define which elements "can be a button" (and in practice only HTML will do this).

Something like

```
Host languages can define which elements, if any, <dfn export>can be a button</dfn>, <dfn export>can be a textfield</dfn> ....

Note: HTML defines these in (link...).
```

css-ui-4/Overview.bs
```diff
@@ -2166,6 +2166,129 @@ Document languages should specify
...
+
+        <li>
+            If |element| can be a ''meter'', given |computedAppearance|, then return meter.
+
+        <li>
+            Return ''appearance/none''.
+    </ol>
+
+    The appropriate <a>widget</a>, given a map of properties |authorProps|, and an expected
+    kind of widget |expected|, and a fallback kind of widget |fallback|, is as follows:
```
zcorpan, line 2230: `[=kind of widget=]`

css-ui-4/Overview.bs
```diff
@@ -2166,6 +2166,129 @@ Document languages should specify
...
+
+    <ol>
+        <li>For each |prop| of the properties that disable native appearance for <a>widgets</a>:
```
zcorpan, line 2233: `[=map/For each=]`

`[=properties that disable native appearance for widgets=]`

css-ui-4/Overview.bs
```diff
@@ -2166,6 +2166,129 @@ Document languages should specify
...
+
+            <ol>
+                <li>If |authorProps|[|prop|] exists, and it is not ''revert'', then return |fallback|.
```
zcorpan, line 2236: `[=map/exists=]`

css-ui-4/Overview.bs
```diff
@@ -2166,6 +2166,129 @@ Document languages should specify
...
+            </ol>
+
+        <li>Return |expected|.
+    </ol>
+</div>
+
+The properties that disable <a>native appearance</a> for <a>widgets</a> are:
```
zcorpan, line 2243: `The <dfn export>properties that disable native appearance for widgets</dfn> are:`

zcorpan (changes requested):
Nice @howard-e! Some comments mostly on cross-references and `dfn`s

css-ui-4/Overview.bs
howard-e: I see, got it!

frivoal:
This should be specified, but I am not convinced it be _in the CSS spec_. Not only should the host language own which elements _can be an X_ (as suggested in https://github.com/w3c/csswg-drafts/pull/6537#discussion_r694278766), but it seems to me it should also own the list of widget types elements can be. Otherwise, HTML cannot introduce new types of widgets without first depending on an update in CSS.

css-ui-4/Overview.bs
```diff
@@ -2166,81 +2166,91 @@ Document languages should specify
...
 
     <ol>
         <!-- none always works -->
         <li>
-            If |computedAppearance| is ''appearance/none'', then return ''appearance: none''.
+            Let |computedAppearance| be the [=computed value=] of |element|'s 'appearance' property.
 
         <li>
-            Let |computedAppearance| be the [=computed value=] of |element|'s 'appearance' property.
+            If |computedAppearance| is [=kind of widget/none=], then return [=kind of widget/none=].
```
zcorpan, line 2187: *computedAppearance* is a value of the 'appearance' property

```suggestion
            If |computedAppearance| is ''appearance/none'', then return [=kind of widget/none=].
```

css-ui-4/Overview.bs
```diff
@@ -2166,81 +2166,91 @@ Document languages should specify
...
+            Let |authorProps| be a [=map=] of all properties that have a <a>cascaded value</a> for
             |element|, where the [=Author Origin=], [=Animation Origin=] and [=Transition Origin=]
             [=Cascade Origins=] are included and the [=User Origin=] and [=UA Origin=] [=cascade origins=]
             are ignored, where the key is the property name and the value is that <a>cascaded value</a>.
 
         <li>
-            If |element| can be a ''button'', given |computedAppearance|, then return the appropriate widget with
-            |authorProps| set to |authorProps|, |expected| set to button, and |fallback| set to
-            ''appearance/none''.
+            If |element| can be a button, given |computedAppearance|, then return the appropriate
```
zcorpan, line 2196: Adding a comment here so we don't lose track of it: "can be a button" (and similar below) should be links to a definition

css-ui-4/Overview.bs
```diff
@@ -2166,81 +2166,91 @@ Document languages should specify
...
+            [=kind of widget/meter=].
 
         <li>
-            Return ''appearance/none''.
+            Return [=kind of widget/none=].
     </ol>
 
-    The appropriate <a>widget</a>, given a map of properties |authorProps|, and an expected
-    kind of widget |expected|, and a fallback kind of widget |fallback|, is as follows:
+    The appropriate <a>widget</a>, given a [=map=] of properties |authorProps|, and an expected
```
zcorpan, line 2239: "appropriate widget" should be its own `<div algorithm="appropriate widget">`, and linked to from the algorithm above

zcorpan (changes requested):

zcorpan:
The content now looks good to me! 👍

Reading the generated spec and the surrounding context, I'm not sure the placement is ideal. I think it makes more sense to introduce a new section, before "Effects of appearance on Decorative Aspects of Elements", called something like "Computing the kind of widget".

zcorpan (approved):
LGTM

zcorpan (changes requested):
It turns we have missed to specify steps for "radio" and "checkbox". I think they should be similar to progress-bar/meter, i.e. authorProps doesn't result in 'none' appearance.

Also "listbox" (`<select multiple>`), which I think should be like button

zcorpan (approved):

zcorpan:
https://github.com/whatwg/html/pull/7004 hooks into this from HTML

zcorpan:
`menulist` had regressed in the rewrite, compare with https://github.com/whatwg/html/pull/4857/files#diff-41cf6794ba4200b839c53531555f0f3998df4cbb01a4d5cb0b94e3ca5e23947dR112533-R112536

Fixed in https://github.com/w3c/csswg-drafts/pull/6537/commits/179b8cb3f5e7d83cc83e5712f7685873b429ce3b

zcorpan:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1730147#c5 @emilio wrote

>> For meter, Chromium has the same rendering between the two, while WebKit and Gecko have different rendering. But for progress all browsers seem to agree to have different appearances (the second being like appearance: none).
>
> I think ~all browsers agree on disabling native styling when backgrounds/borders are specified on ~all html-exposed widgets except checkboxes and radios. So I'm not sure why meter would be different, so I think Gecko/WebKit behavior is slightly better.

When I worked on this originally, browsers were less consistent, and the `appearance: none` rendering for `progress` and `meter` in some browsers was *no* style and render the children (like a `span`). There were objections to that behavior, and it has now changed in browsers. But it seems the new normal is to do the *fallback* thing for `progress` and `meter`. See https://github.com/w3c/csswg-drafts/issues/356

zcorpan:
@frivoal does this look OK? We'd like to merge this next week if possible, so it can be [considered for Interop2022](https://github.com/web-platform-tests/interop-2022/issues/11#issuecomment-966632869).

zcorpan:
I realized that there was no mapping from the 'appearance' value to a kind of widget. Now fixed.

Also, the different kinds of widgets now say what the rendering should be, i.e., "checkbox" should render as a checkbox and "button" as a button, etc.

zcorpan:
Open UI discussed this today: https://www.w3.org/2022/02/03-openui-minutes.html

> RESOLUTION: For elements without a concept of native appearance, we should use `appearance: none`.

zcorpan:
@frivoal can you clarify how you envisioned the algorithm would be refactored? Since there's no connection between the term "can be a button" and "kind of widget/button", we'd need to specify that mapping somehow.

zcorpan:
The corresponding PR for HTML https://github.com/whatwg/html/pull/7004 has been reviewed and is ready to merge when this PR is merged.

frivoal:
Alternative refactored version posted here: https://github.com/w3c/csswg-drafts/pull/7224

frivoal:
Closing as retracted, since @zcorpan (and the WG) accepted https://github.com/w3c/csswg-drafts/pull/7224 as a preferable alternative.

PR #6560, opened 2021-08-31
[css-color-5] Resolving color-contrast()

andruud:


andruud:
@argyleink Something like this? Or do we want to eliminate the `color-contrast()` function completely in cases where that's possible? (The same way e.g. `calc(1px + 1%)` computes to `calc(1px + 1%)`, but `calc(1px + 1px)` computes to `2px`).

svgeesus:
> Or do we want to eliminate the color-contrast() function completely in cases where that's possible?

Good question. I can see arguments for eliminating it in the used value (just give the winning color).

Overall this is a good PR (thanks!) and I am tempted to merge it and fix any issues later.

svgeesus (approved):

andruud:
> just give the winning color

Yeah, we should do that. (Done). Having to support "partially" resolved colors would probably just be annoying.

Now I also notice "6.2. Serializing color-contrast()". It's implying that we can select the winning color parse-time, which is not (always) possible. So back to the question about `calc(1px + 1%)` I guess. Do we want to simplify parse-time when possible?

argyleink:
> Do we want to simplify parse-time when possible?

definitely?! if it's possible to select the winning color on parse, then take the early out. otherwise resolve at use time. iiuc?

argyleink (approved):
thanks for this!!

css-color-5/Overview.bs
```diff
@@ -684,10 +684,13 @@ Selecting the most contrasting color: the ''color-contrast()'' function {#colorc
 
 	<h3 id="resolving-contrast">Resolving ''color-contrast()'' values</h3>
 
-	The computed and used value is the specified ''color-contrast()'' function
+	The computed value is the specified ''color-contrast()'' function
 	with each <<color>> parameter resolved according to [[css-color-4#resolving-color-values]],
 	and the keywords ''AA'', ''AA-large'', ''AAA'', ''AAA-large'' replaced with their corresponding numeric value.
 
+	The used value is the winning color
+	resolved according to [[css-color-4#resolving-color-values]].
```
svgeesus, line 692: That implies it is always possible to determine the winning color. Is it? Even with custom properties, calc and so on in the mix?

tabatkins: By used-value time, yes. Custom properties disappear at computed-value time; calc()s disappear, at their latest, at used-value time as well. (Everything *better* be resolvable by then, since CSS doesn't actually *define* anything happening past that, just that the UA can make unstated modifications to obtain the "actual value".)

PR #6565, opened 2021-09-01
Fix bikeshed indentation errors in web-animations-2 diff spec.

flackr:
[web-animations-2] Fix bikeshed indentation errors.

The spec had several incorrect indents that lead to failures running bikeshed on it.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -169,13 +169,13 @@ Add:
 > follows:
 >
 > 1.  If the [=iteration duration=] is auto, then perform the following steps.
->         * Set [=start delay=] and [=end delay=] to 0, as it is not
->           possible to mix time and proportions.
->           <div class='note'>
->           Future versions may allow these properties to be assigned
->           percentages, at which point the delays are only to be ignored if
->           their values are expressed as times and not as percentages.
->           </div>
+>         *   Set [=start delay=] and [=end delay=] to 0, as it is not
+>             possible to mix time and proportions.
+>             <div class='note'>
+>             Future versions may allow these properties to be assigned
+>             percentages, at which point the delays are only to be ignored if
+>             their values are expressed as times and not as percentages.
+>             </div>
 >     Otherwise:
 >          1. Let <var>total time</var> be equal to |end time|
 >          1. Set [=start delay=] to be the result of evaluating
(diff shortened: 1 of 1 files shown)


web-animations-2/Overview.bs
```diff
@@ -169,13 +169,13 @@ Add:
...
->           percentages, at which point the delays are only to be ignored if
->           their values are expressed as times and not as percentages.
->           </div>
+>         *   Set [=start delay=] and [=end delay=] to 0, as it is not
+>             possible to mix time and proportions.
+>             <div class='note'>
+>             Future versions may allow these properties to be assigned
+>             percentages, at which point the delays are only to be ignored if
+>             their values are expressed as times and not as percentages.
+>             </div>
```
tabatkins, line 178:
```suggestion
>     *   Set [=start delay=] and [=end delay=] to 0, as it is not
>         possible to mix time and proportions.
>         
>         Note: Future versions may allow these properties to be assigned
>         percentages, at which point the delays are only to be ignored if
>         their values are expressed as times and not as percentages.
>         
```

web-animations-2/Overview.bs
```diff
@@ -216,18 +216,19 @@ follows:
 1.  Set |previous progress| based in the first condition that applies:
     <div class="switch">
 
-    : If |previous current time| is unresolved:
+    :   If |previous current time| is unresolved:
```
tabatkins, line 219: Indent this list one level so it's properly nested under its `div`; then you won't need the blank line to make it work properly. You can also swap out the `div` for a `dl`, so it'll collapse into the Markdown `dl` in the output rather than remaining as a wrapper.

(Can't do a suggestion for this, since it woudl involve editting unchanged lines.)

web-animations-2/Overview.bs
```diff
@@ -1014,6 +1015,7 @@ Add:
 >     matching condition from below:
 >
 > <div class="switch">
+>
```
tabatkins, line 1018: Same here - indent the markdown list and you won't need these blank lines to hack around the markdown parser. And swap the `div` for a `dl` as well.

web-animations-2/Overview.bs
```diff
@@ -1032,6 +1034,7 @@ Add:
 >     Presently start and end delays are zero until such time as percentage
```
tabatkins, line 1034: Might as well swap this out for a markdown paragraph starting with `Note:`, as I included in an earlier suggestion, too. Then you don't have to worry about the `p`'s contents also being improperly indented. ^_^

tabatkins (changes requested):

web-animations-2/Overview.bs
flackr: For swapping to a dl, I'm not sure how to preserve the switch formatting. In the generated css, `div.switch > dl > dt` gives the custom markers but if i switch to just a dl then of course it doesn't have the wrapping div so the custom markers aren't used.

tabatkins: Hm, the selector should just be keying off of the `switch` class, not the `div` tagname. I'll note, tho, that in the current draft it's not being displayed at all; it's possible that the custom marker is only in Bikeshed's stylesheet and not the W3C stylesheet. ^_^

flackr: That's curious, on https://drafts.csswg.org/web-animations-2/ the styles for switch are specified on the dl element rather than the wrapping div. I guess there's something wrong with my local bikeshed?

tabatkins: Ah indeed, it's on `dl.switch`. Well, my point still stands that you should swap to a wrapper `dl` rather than a wrapper `div` so this actually works the way you intend. ^_^

web-animations-2/Overview.bs
flackr: Done.

web-animations-2/Overview.bs
flackr: I've swapped all of them now ^_^

web-animations-2/Overview.bs
flackr: After fixing all of them it seems like my local bikeshed is generating css for `dl.switch` now 🤷‍♂️

tabatkins (approved):

Issue #4597, opened 2019-12-13
[css-highlight-api] invalidation of static ranges


frivoal:
As far as I am aware, uses of `StaticRange` prior to css-highlight-api were for ranges created by the User Agent and passed to the author.
In css-highlights-api, it's the other way around, which raises (for the first time?) the question of invalidation of static ranges, as they may have become stale by the time the browser needs to use them.

https://drafts.csswg.org/css-highlight-api-1/#range-invalidation defines a way to evaluate that. Is that reasonable? If not, what should we do instead?

dandclark:
The invalidation logic at https://drafts.csswg.org/css-highlight-api-1/#range-invalidation seems reasonable.

In addition to checking whether the start and end node are both [in a document tree](https://dom.spec.whatwg.org/#in-a-document-tree), I wonder if this should check if they are in the *same* document tree.  For example if either the start or end container is adopted into a same-domain iframe document we should probably treat the range as invalid.

Checking that the range doesn't cross the boundary of a contained element per #4598 could also be an addition here.

sanketj:
I have proposed a set of criteria for validating StaticRange here: https://github.com/whatwg/dom/issues/947. These criteria can be used for _any_ StaticRange, not just static ranges used in the highlight API. I think #4598 should be an additional restriction for the highlight API, but not something that applies generally to StaticRange.

annevk:
Why does this not use live ranges internally? Is any implementation actually planning on using static ranges internally?

cc @smaug----

sanketj:
There are certainly benefits to the highlight API using live ranges internally:
- The `Highlight` constructor can be modified to only accept `StaticRange`. The way I see it is that highlights would create and maintain their own "highlight ranges" internally, which are live, and the static ranges passed into the constructor are just inputs for creating these internal ranges.
- We wouldn't need to define any invalidation behavior for the highlight API, and #4598 wouldn't be an issue anymore.

However, I don't think we have answers yet regarding the performance tradeoff - live range maintenance cost vs. static range and highlight validation cost. It would be great to know if Safari ran any numbers on this (since they have an experimental implementation already) and what their findings were. cc: @hober, @megangardner

Regardless of which option is ultimately chosen, we will need to document this in the spec, as there are painting behavior differences depending on which way we go.

annevk:
Are the painting differences elaborated on somewhere?

sanketj:
Consider the example below:

```
<!DOCTYPE html>
<html>
[...]

PR #6584, opened 2021-09-08
[css-highlights-api] StaticRange validity updates

dandclark:
Remove Issue block now that #4597 has been resolved.

Remove start/end offset clipping behavior since we resolved to just use the DOM spec's validity definition.


sanketj (approved):

dandclark:
Hi @frivoal, friendly ping on this as well as #6585, do these updates look good to you given the resolutions from last week?

frivoal (approved):

Issue #520, opened 2016-09-21
[css-fonts-4] [varfont] Supported variation font axes and font features are not discoverable


litherum:
Web authors have requested the ability to know which variation axes and font features are available in the fonts on the page.

Crissov:
This is clearly out of scope for CSS (although I know next to nothing about CSSOM and Houdini).

Additional or changed descriptors for `@font-face` would be a different thing. Authors wouldn’t *know* about available features, but *declare* what to do if they exist.

litherum:
Using a media query for this wouldn't work because media queries have no concept of a specific font face to investigate. Instead, it seems that such a mechanism would have to be in JavaScript.

The canvas text metrics API and the forthcoming Houdini text metrics API expose some other font-driven information. However, these APIs give you information about a line (or lines) of text, and would include things like fallback fonts. Variation axes and features are font-specific, so including fallback fonts isn't the right level of abstraction for solving this problem.

A natural place to put this would be to build off of the CSS Font Loading API. This API already has the concept of an object which represents a font face. In addition, a FontFace even already has a "status" enum property which has a relationship to a specific url() inside the FontFace (so there is precedent for this idea already). Calling the function on an in-flight download would result in an exception or something. Maybe a function can be added to FontFace which returns a new object which has a bunch of properties including font info.

Any thoughts? @tabatkins @bramstein @nattokirai @SergeyMalkin @drott @heycam 

P.S. According to caniuse, Edge hasn't implemented the CSS Font Loading API, so they would probably have to implement at least some of it in order to go with this proposal.

svgeesus:
I agree that extending Font Loading API seems like a natural way to get this discoverability. Although it would be nice to be able to code up a test page in HTML/CSS to explore what some font provides.
Ultimately fonts need to have better documentation.

litherum:
We probably would want something similar for font color palettes.

svgeesus:
> The canvas text metrics API and the forthcoming Houdini text metrics API expose some other font-driven information. However, these APIs give you information about a line (or lines) of text, and would include things like fallback fonts. 

Yes. At the early Houdini meetings the assumption was that Houdini would produce a font metrics API, to query features of a particular font (and with the advent of variation fonts, a particular instance). However, over time that was seen as less important for pollyfiller-writers  than metrics on fonts used together in a text segment, which eventually became the text metrics API.

There is still a need to query an individual font instance.

briansmith:
Consider the following CSS:

[...]

PR #6591, opened 2021-09-10
[css-font-loading] Add IDL for discovery of information about a font

litherum:
This is for https://github.com/w3c/csswg-drafts/issues/520.

The [resolution](https://github.com/w3c/csswg-drafts/issues/520#issuecomment-387799523) is:

> RESOLVED: [add this functionality to] css-font-loading, level TBD

I added this to level 3, because level 4 doesn't exist. If the editor prefers this go in a level 4, that's no problem.

The GitHub thread hasn't come to a conclusion about what should go in the `FontFaceFeatures` interface, so I left it blank for now.

### DIFF
--- a/css-font-loading-3/Overview.bs
+++ b/css-font-loading-3/Overview.bs
@@ -427,6 +427,54 @@ Interaction with CSS’s ''@font-face'' Rule</h3>
 	and <a>CSS-connected</a> to the ''@font-face''.
 	(This will also remove the old and add the new {{FontFace}} objects from any <a>font sources</a> they appear in.)
 
+<h3 id='discovery'>
+Discovery of information about a font</h3>
+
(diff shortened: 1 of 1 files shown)


tabatkins:
Looks good overall. I wonder if we should be using an ObservableArray instead, tho? That would give us all the array methods, which the iterator-based approach doesn't.

litherum:
Is there a such thing as a readonly ObservableArray? Isn't that what FrozenArray is supposed to represent? (WebIDL says "There is no way to represent a constant frozen array value in IDL.")

litherum:
Looks like WebKit doesn't support ObservableArray yet :(

litherum:
This was approved by @tabatkins. From IRC:

[11:48:19]  <TabAtkins> I can go ahead and merge ... or you can, just note that I approved it

PR #6632, opened 2021-09-17
Reset to the @when from the resolved-on proposal

tabatkins:
We resolved to accept the proposal from <https://tabatkins.github.io/specs/css-when-else/>, not a modified version using @if. The UD uses @when for a reason; it's not a trivial edit worthy of being done without discussion.


svgeesus:
Uh, no, we resolved on if/else

It literally says that on the resolution.

But I'm fine to have more discussion (there has already been some)

Issue #6563, opened 2021-08-31
[web-animations-2] Makes finish and cancel event times CSSNumberish


kevers-google:
Per spec, the finished event's currentTime is the current time of the animation, and the finished event's timelineTime is the current time of timeline.  Same story for a cancel event. With https://github.com/w3c/csswg-drafts/pull/6479 (https://github.com/w3c/csswg-drafts/issues/6458), the animations current time becomes CSSNumberish, yet the times in AnimationPlaybackEvent are declared as nullable doubles.  These should be CSSNumberish as well.  This change is backward compatible for monotonic timelines, but is required to support progress-based timelines, where current time must be expressed as a percentage and not a double in ms.

PR #6656, opened 2021-09-21
[web-animations-2] Make times in AnimationPlaybackEvent CSSNumberish

kevers-google:
[web-animations-2] Make times in AnimationPlaybackEvent CSSNumberish

The attributes timelineTime and currentTime in AnimationPlaybackEvent are currently optional doubles. These should be CSSNumberish since timeline.currentTime and animation.currentTime are now CSSNumberish.

https://github.com/w3c/csswg-drafts/issues/6563


web-animations-2/Overview.bs
```diff
@@ -2855,3 +2855,21 @@ Add:
 >     </div>
 >
 >     Note: Need to spec this properly somewhere.
+
+The <code>AnimationPlaybackEvent</code> interface {#the-animationplaybackevent-interface}
```
flackr, line 2859: nit: Can we put this before Model liveness (since this precedes Model liveness in https://drafts.csswg.org/web-animations-1/#the-animationplaybackevent-interface . Also I think this needs to be an h3 to be one of the sections under programming interface.

web-animations-2/Overview.bs
```diff
@@ -2855,3 +2855,21 @@ Add:
...
+    constructor(DOMString type, optional AnimationPlaybackEventInit
+    eventInitDict = {});
+    readonly attribute CSSNumberish? currentTime;
+    readonly attribute CSSNumberish? timelineTime;
+};
+dictionary AnimationPlaybackEventInit : EventInit {
+    CSSNumberish? currentTime = null;
+    CSSNumberish? timelineTime = null;
+};
+</pre>
```
flackr, line 2875: Given that web-animations-1 proceeds to use the type in the attributes and dictionary members we might need to repeat the section here.

flackr (changes requested):

web-animations-2/Overview.bs

web-animations-2/Overview.bs

web-animations-2/Overview.bs
kevers-google: In making the change, I noticed that rules for formatting attributes, methods, et al. where missing.  Copied this rules for web-animations-1.

flackr (approved):

PR #6663, opened 2021-09-22
[css-pseudo] [css-variables] highlight pseudos and custom properties (closes #6264)

### LINKED ISSUE #6264: [css-pseudo] highlight pseudos and custom properties
https://drafts.csswg.org/css-pseudo-4/#highlight-styling

> [ISSUE 1] Are there any other properties that should be included here?

Should custom properties be allowed on highlight pseudos? What about use of cascading variables? If so, does the rule stating that non-inherited properties are to be treated as inherited, apply to [registered custom properties](https://drafts.css-houdini.org/css-properties-values-api-1/#the-propertydefinition-dictionary) where inherit = false?

I suppose these features are ok, given that we’ve completely isolated [highlight inheritance](https://drafts.csswg.org/css-pseudo-4/#highlight-cascade) from originating element inheritance. As a result, defining or changing a custom property value should already be unable to “escape” and affect layout.

delan:
This patch makes edits for #6264:

* adds a section stating that custom properties are allowed on all pseudos
* explicitly clarifies that non-inherited registered customs are also inherited

One thing I’m unsure of… should we restate or refer to this near the lists of applicable properties? Especially with [the one for ::marker](https://drafts.csswg.org/css-lists/#marker-properties), which is in a separate spec, the reader could miss this if they’ve only read the section about a specific pseudo.

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -49,13 +49,6 @@ spec:fill-stroke-3; type:property; text:stroke-width
   and on their syntax and interaction with other <a>selectors</a>,
   see [[!SELECTORS-4]].
 
-<h2 id="styling">Styling Pseudo-elements</h2>
-  [=Custom properties=] [[!CSS-VARIABLES-1]] can be specified on--
-  and inherited into--
-  pseudo-elements,
-  even if they would not otherwise apply
-  due to the pseudo-element having a restricted set of applicable CSS properties.
-
 <h2 id="typographic-pseudos">
 Typographic Pseudo-elements</h2>
 
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -50,9 +50,11 @@ spec:fill-stroke-3; type:property; text:stroke-width
   see [[!SELECTORS-4]].
 
 <h2 id="styling">Styling Pseudo-elements</h2>
-
-  [=Custom properties=] [[!CSS-VARIABLES-1]] are always allowed in pseudo-element styles,
-  even if the pseudo-element has a restricted set of CSS properties applying to it.
+  [=Custom properties=] [[!CSS-VARIABLES-1]] can be specified on--
+  and inherited into--
+  pseudo-elements,
+  even if they would not otherwise apply
+  due to the pseudo-element having a restricted set of applicable CSS properties.
 
 <h2 id="typographic-pseudos">
 Typographic Pseudo-elements</h2>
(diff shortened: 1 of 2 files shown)


css-pseudo-4/Overview.bs
```diff
@@ -49,6 +49,11 @@ spec:fill-stroke-3; type:property; text:stroke-width
   and on their syntax and interaction with other <a>selectors</a>,
   see [[!SELECTORS-4]].
 
+<h2 id="styling">Styling Pseudo-elements</h2>
+
+  [=Custom properties=] [[!CSS-VARIABLES-1]] are always allowed in pseudo-element styles,
+  even if the pseudo-element has a restricted set of CSS properties applying to it.
+
```
Loirooriol, line 56: I think it may be better to explicitly list custom properties in the various lists of allowed properties for pseudo-elements.

And also in https://drafts.csswg.org/css-variables-1/#defining-variables
> Applies to: all elements (Includes ::before and ::after pseudo-elements.)

should probably clarify that all pseudo-elements are included, not just ::before and ::after

delan: i’m happy to move this to the local lists only, if that’s what you mean. note that in the resolution for #6264, @tabatkins seemed to prefer the blanket statement:

> \<bkardell_> TabAtkins: we should probably make a blanket statement that they are valid on all pseudo elements

frivoal: How about this rephrasing?
```suggestion
<h2 id="styling">Styling Pseudo-elements</h2>

  [=Custom properties=] [[!CSS-VARIABLES-1]] can be specified on--
  and inherited into--
  pseudo-elements,
  even if they do not apply
  due to the pseudo-element having a restricted set of applicable CSS properties.
```

I think I like it a little better, because in the original "allowed" is a bit vague. Allowed to do what?

delan: (thanks for the feedback (here and #6665) @frivoal, i’m on leave right now, but will take a look after 4 january)

delan: done in 5b50dee956cb9faec57258bd1e68ddf9e2d9d20f with a minor tweak to “do not apply”, what do you think?

fantasai (changes requested):
The second half looks good.

For the first half, I think it would be better if we put this into the definition of custom properties. E.g.

```diff
- Applies to: all elements
+ Applies to: all elements and all pseudo-elements (including those with restricted property lists)
```

I think that'll be a less awkward place to put it, and makes it an integral part of the definition of exactly the thing it applies to.

delan:
Thanks @fantasai, done in e0adc217f6a193ebb73f2457ba503cab7da84bb3. Do we need to pull Tab in as reviewer for the css-variables part?

tabatkins:
r+ on the Variables edit

fantasai (approved):

Issue #6386, opened 2021-06-16
[css-pseudo] clarify paired-cascading behavior


delan:
https://drafts.csswg.org/css-pseudo-4/#highlight-cascade

> The UA must use its own highlight colors for ::selection only when neither color nor background-color has been specified by the author.

I think this sentence could be worded more precisely:

1. **Does “use” mean [used value](https://www.w3.org/TR/css-cascade-4/#used)?** This is mostly theoretical, because the spec doesn’t expose highlight pseudos to getComputedStyle, but I think it might affect what the devtools story looks like.
2. **Does “highlight colors” refer to a set of properties, and if so, which properties?** My best guess is that color and background-color are included (because those are the ones with defaults in practice), and maybe stroke-color and fill-color too ([wpt](http://wpt.live/css/css-pseudo/textpath-selection-011.html))? But what about text-{decoration,emphasis,shadow}?
3. **Do we really mean “for ::selection” only, or did we intend this for all highlight pseudos?** The note suggests this rule exists for compatibility with existing ::selection impls, but ::target-text often has a default background-color too (see also #6375).
4. **Do we really mean “neither color nor background-color” only, or any of [the applicable properties](https://drafts.csswg.org/css-pseudo-4/#highlight-styling)?** Note that the former would be inconsistent with existing impl behavior. For example, in Chrome and Firefox, setting text-shadow (even to none) also disables the usual default background-color ([example](https://bucket.daz.cat/work/igalia/0/13.html)).
5. **What does “specified by the author” mean?** My best guess is that “specified by the author” means “[cascaded value](https://www.w3.org/TR/css-cascade-4/#cascaded) exists, and its [origin](https://www.w3.org/TR/css-cascade-4/#cascading-origins) was not the user agent”. Not specified as in [specified value](https://www.w3.org/TR/css-cascade-4/#specified), which iiuc always exists due to [defaulting](https://www.w3.org/TR/css-cascade-4/#defaulting) ([to transparent](https://drafts.csswg.org/css-backgrounds-3/#background-color) for background-color).
[...]

PR #6665, opened 2021-09-22
[css-pseudo] clarify paired cascade (#6386)

delan:
This patch makes edits for #6386.

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -758,6 +758,38 @@ Painting the Highlight</h3>
     css/css-pseudo/highlight-painting-003.html
     </wpt>
 
+<h4 id="paired-cascade">
+Highlight Colors</h4>
+
+In [=highlight pseudo-element=] styles,
+'color' and 'background-color' are known as <dfn>highlight colors</dfn>.
+For compatibility reasons,
+UA default highlight colors must only be [=used value|used=]
+when neither of the highlight colors yield a [=specified value=]
+of the [=author origin|author=],
+[=animation origin|animation=],
+or [=transition origin|transition=] origins.
+In other words,
+when setting one custom highlight color,
+UA defaults for the other highlight color are suppressed.
+
+Note: 'stroke-color' and 'fill-color'
+do not suppress the UA default 'background-color' in this way,
+though they continue to affect the foreground color as usual.
+
+For the purposes of this section,
+''unset'' is considered equivalent to a [=specified value=] of ''inherit''
+in the [=declared value|declaring=] origin.
+This makes the value not quite equivalent to
+erasing all [=declared values=] occurring earlier in the cascade,
+as would normally be the case.
+
+On the other hand,
+when a highlight color is ''revert'',
+the origin of the resultant value
+<em>after</em> rolling back the cascade
+is considered by this rule.
+
 <h4 id="highlight-backgrounds">
 Backgrounds</h4>
 


css-pseudo-4/Overview.bs
```diff
@@ -667,12 +689,20 @@ Cascading and Per-Element Highlight Styles</h3>
   for which that element is the <a>originating element</a>.
   When multiple styles conflict,
   the winning style is determined by the cascade.
+
   When any supported property is not given a value by the cascade,
-  it's value is determined by inheritance from
+  or given a value of ''inherit'',
```
fantasai, line 694: Say “''inherit'' or ''unset''” here, and then delete the new paragraph below.

css-pseudo-4/Overview.bs
```diff
@@ -603,10 +603,32 @@ Default UA Styles</h3>
     /* Represent default spelling/grammar error styling in an adjustable way */
     :root::spelling-error { text-decoration-line: spelling-error; }
     :root::grammar-error  { text-decoration-line: grammar-error; }
-    /* Highlight targeted text */
-    :root::target-text    { color: MarkText; background: Mark; }
   </pre>
 
+  Note: the “paired cascade” behavior of [=highlight colors=]
```
fantasai, line 608: We haven't introduced the paired cascading behavior yet, so moving the note that refers to it here is confusing.

css-pseudo-4/Overview.bs
```diff
@@ -603,10 +603,32 @@ Default UA Styles</h3>
...
+  can't be expressed in terms of 'color' and 'background-color'
+  in [=UA style sheet=] rules.
+  However it has been interoperably implemented in browsers
+  and is thus probably a Web-compatibility requirement.
+
+  UAs may define <dfn>default highlight colors</dfn>
+  for each highlight pseudo-element,
+  consisting of a 'color' and a 'background-color',
+  and doing so is recommended
+  for ''::selection'' and ''::target-text''.
```
fantasai, line 619: Is this an RFC2119 RECOMMEND? Or did you mean something different?

css-pseudo-4/Overview.bs
```diff
@@ -758,6 +777,60 @@ Painting the Highlight</h3>
     css/css-pseudo/highlight-painting-003.html
     </wpt>
 
+<h4 id="paired-cascade">
+Highlight Colors</h4>
```
fantasai, line 781: I think this section can also be simplified a bit, but, I'll have to take a pass at it later. :)

css-pseudo-4/Overview.bs
delan: done in 06de02d057ea554fac09643037878d426b9941fb

css-pseudo-4/Overview.bs
delan: by all means!

css-pseudo-4/Overview.bs
delan: yeah, it’s a combination of moving this note up from #highlight-cascade (because it’s about default UA styles) and moving the normative text for paired cascade down to #highlight-painting (because it affects used values), which leaves us with a problem here. i’ve tweaked the wording in fa44a161cdba4f8a4f117f53094478576d76ee32, what do you think?

delan (commented):
Thanks for taking a look at this!

css-pseudo-4/Overview.bs
```diff
@@ -593,7 +601,7 @@ Styling Highlights</h3>
 -->
 
   Note: Historically (and at the time of writing)
-  only 'color' and 'background-color' have been interoperably supported.
+  only the highlight colors have been interoperably supported.
```
frivoal, line 604:
```suggestion
  only the [=highlight colors=] have been interoperably supported.
```

delan: done in 24df479ce2bf25d3090559715a54e7eec31d4025

frivoal:
It seems to me that we might end up needing further tweaks based on https://github.com/w3c/csswg-drafts/issues/6818, but that other than that, this is probably good and merging it would be reasonable.

delan:
Thanks @frivoal! I think we can fairly cleanly leave #6818 for a separate patch, seeing as that issue is about resolved/used values and this patch only affects the cascade sections.

@fantasai did you want to take another pass at this before we merge?

css-pseudo-4/Overview.bs
delan: 4eb95b022c1eebb02bd4969bf723fd8ab4d6825a should make this relationship clearer.

css-pseudo-4/Overview.bs
delan: i’ve rewritten and heavily simplified this section in 86bb487445f17b354230b7832e116e697982d6de.

delan:
@fantasai, @frivoal, some notable improvements to this patch since your last reviews:

* i’ve simplified the origins that affect paired cascade to author only (bfc6498f9fe6342474de16b936810b00e5b0e306)
* i’ve removed the text allowing UA tweaks to author highlight colors (0c46d9d9a15a65f9634cd4c72c6d5dbea27c8c12)
* i’ve defined the inherited value of ‘color’ in the root as ‘currentColor’ (ce778504e1adba403a311094fea9254378b4cf80)
* i’ve tied the default highlight colors for ::selection + ::target-text to system colors (4eb95b022c1eebb02bd4969bf723fd8ab4d6825a)
* i’ve rewritten and heavily simplified the new section about paired cascade itself (86bb487445f17b354230b7832e116e697982d6de)

I believe the new proposed text is clearer than ever, and conveys all of the details I wanted to convey (see pr description for the complete list) in far fewer words. Please take a look!

delan:
@fantasai, @frivoal, #6774 is now resolved, so these changes are ready for review. It looks like fantasai last reviewed 44aa6cede80d6a8fbda5e12b7e443e7c94c8407f while Florian last reviewed 24df479ce2bf25d3090559715a54e7eec31d4025, in case it helps you pick up where you left off and see what has changed.

css-pseudo-4/Overview.bs
```diff
@@ -724,10 +725,24 @@ Default UA Styles</h3>
     /* Represent default spelling/grammar error styling in an adjustable way */
     :root::spelling-error { text-decoration-line: spelling-error; }
     :root::grammar-error  { text-decoration-line: grammar-error; }
-    /* Highlight targeted text */
-    :root::target-text    { color: MarkText; background: Mark; }
   </pre>
 
+  Some highlight pseudo-elements
+  should have <dfn>default highlight colors</dfn>,
+  consisting of a 'color' and a 'background-color'.
```
fantasai, line 732:
```suggestion
  Some [=highlight pseudo-elements=]
  have <dfn>paired default highlight colors</dfn>,
  consisting of a default 'color' and a 'background-color'
  that are either used or overridden together,
  see [[#paired-cascade]].
```

css-pseudo-4/Overview.bs
```diff
@@ -724,10 +725,24 @@ Default UA Styles</h3>
...
+  For ''::selection'' they are
+  ''HighlightText'' and ''Highlight'',
```
fantasai, line 734:
```suggestion
  For ''::selection'' they should correspond to
  ''HighlightText'' and ''Highlight'',
```

css-pseudo-4/Overview.bs
```diff
@@ -724,10 +725,24 @@ Default UA Styles</h3>
...
+  while for ''::target-text'' they are
+  ''MarkText'' and ''Mark''.
```
fantasai, line 736: Did we decide to add paired-cascading behavior to ''::target-text'' also? There's afaik no Web-compatibility requirement for it here, right?

css-pseudo-4/Overview.bs
```diff
@@ -724,10 +725,24 @@ Default UA Styles</h3>
...
+
+  Note: the painting behavior of highlight colors
+  means that default highlight colors
+  can't be expressed in terms of 'color' and 'background-color'
+  in [=UA style sheet=] rules.
```
fantasai, line 741: Can you explain why this wording is better than the original wording?

From the author's perspective (though you might implement it however you want), it's a special cascading rule, not a “painting” behavior, so I think the original wording is more understandable.

Also, you want `'color' and 'background-color' declarations` here.

css-pseudo-4/Overview.bs
```diff
@@ -879,6 +890,43 @@ Painting the Highlight</h3>
     css/css-pseudo/highlight-painting-003.html
     </wpt>
 
+<h4 id="paired-cascade">
+Highlight Colors</h4>
```
fantasai, line 894: This should be a subsection of “Default UA Styles”.

css-pseudo-4/Overview.bs
```diff
@@ -879,6 +890,43 @@ Painting the Highlight</h3>
...
+
+  For compatibility reasons,
+  [=default highlight colors=]
+  must only be [=used value|used=]
+  when neither of the [=highlight colors=]
+  yield a [=cascaded value=]
+  or [=specified value=]
```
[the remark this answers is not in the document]
delan: iirc i was trying to say that paired cascade’s suppression of defaults when authors set highlight colors spreads to descendants. for example, if the author only declares body::selection { color: red; }, then we want p::selection to be red on initial (transparent), not red on default (Highlight), even though there is no author cascaded value, only an author specified value after (implicit) defaulting.

same goes for if the UA or user also declares p::selection { color: inherit; }. in this case, there is no author cascaded value, but there *is* an author specified value after (explicit) defaulting.

hopefully that makes sense, my mental model of css-cascade is pretty new and was only based on reading the spec. either way, i agree that those three words don’t really convey this idea well.

css-pseudo-4/Overview.bs
```diff
@@ -879,6 +890,43 @@ Painting the Highlight</h3>
...
+    in the <code>&lt;em></code> element
+    if given by the author:
+    <pre class="lang-css">
+      /* cascaded value is author origin */
+      em::selection { color: initial; }
+      em::selection { color: inherit; }
+      em::selection { color: unset; }
+      em::selection { color: green; }
+
+      /* specified value is author origin */
```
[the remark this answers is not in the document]
delan: same thing as my big comment about specified value. the first four rules are the author directly setting ‘color’ on the em, so default highlight colors are suppressed due to author cascaded value, while the last rule is the author setting ‘color’ on the parent, so there is no author cascaded value, but they are suppressed due to author specified value.

these comments were only meant to make up for the lack of clarity there, so i’ll remove them.

css-pseudo-4/Overview.bs
```diff
@@ -696,11 +696,12 @@ Styling Highlights</h3>
 
   Issue: Are there any other properties that should be included here?
 
-  The 'color' property specifies the color of both the text
-  and all line decorations (underline, overline, line-through)
-  and emphasis marks ('text-emphasis')
-  applied to the text
-  by the <a>originating element</a> and its ancestors and descendants.
```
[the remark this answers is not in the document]
delan: we already explain this in #highlight-painting (and mention emphasis marks in a note):

> A highlight pseudo-element suppresses the normal drawing of any associated text, and the text decorations (other than shadows) that had been applied to that text. Instead the topmost active highlight overlay redraws that text (and those decorations) over all the highlight overlay backgrounds using that highlight’s own [color](https://drafts.csswg.org/css-color-4/#propdef-color).

i don’t mind redundancy, but the way i interpreted the layout of the highlight pseudos section is that computed-value-time behaviour belonged in #highlight-styling, while used-value-time behaviour belonged in #highlight-painting.

css-pseudo-4/Overview.bs
delan: not formally, and not for web compat, just based on impl opinions from Emilio and i (<https://github.com/w3c/csswg-drafts/issues/6386#issuecomment-957389344>).

css-pseudo-4/Overview.bs
```diff
@@ -696,11 +696,12 @@ Styling Highlights</h3>
...
+  'color' and 'background-color'
+  are known as <dfn>highlight colors</dfn>.
+  Because this concept exists for compatibility reasons,
+  other similar properties
+  like 'fill-color' or 'stroke-color'
+  must not be treated as highlight colors.
```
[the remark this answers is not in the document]
delan: it looks like i moved this definition up in 65597ae6af59211f1c857c6509cf696bcdbf52df. at the time it linked to the section explaining the impl impact, and i wanted to use the term “highlight colors” in the “Historically” note, but neither of those are true now, so i’ll move it back.

css-pseudo-4/Overview.bs
delan: i had to reword it because i removed the preceding sentence

> The UA must use its own highlight colors for [::selection](https://drafts.csswg.org/css-pseudo-4/#selectordef-selection) only when neither [color](https://drafts.csswg.org/css-color-4/#propdef-color) nor [background-color](https://drafts.csswg.org/css-backgrounds-3/#propdef-background-color) has been specified by the author.

which in turn was removed because it’s now explained more thoroughly (and correctly) in #highlight-painting, but i see where you’re coming from. i’ll revert to the old text and correct the thing about ::selection.

css-pseudo-4/Overview.bs
```diff
@@ -696,6 +696,14 @@ Styling Highlights</h3>
 
   Issue: Are there any other properties that should be included here?
 
+  In particular,
+  'color' and 'background-color' are known as <dfn>highlight colors</dfn>,
+  and their [=used values=] are handled specially (see [[#paired-cascade]]).
+  Because this concept exists for compatibility reasons,
+  other similar properties
+  like 'fill-color'
+  must not be treated as highlight colors.
```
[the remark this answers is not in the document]
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
```diff
@@ -696,11 +696,12 @@ Styling Highlights</h3>
 
   Issue: Are there any other properties that should be included here?
 
-  The 'color' property specifies the color of both the text
-  and all line decorations (underline, overline, line-through)
-  and emphasis marks ('text-emphasis')
-  applied to the text
-  by the <a>originating element</a> and its ancestors and descendants.
+  'color' and 'background-color'
+  are known as <dfn>highlight colors</dfn>.
```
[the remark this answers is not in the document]
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
```diff
@@ -714,7 +715,7 @@ Styling Highlights</h3>
 -->
 
   Note: Historically (and at the time of writing)
-  only 'color' and 'background-color' have been interoperably supported.
+  only the highlight colors have been interoperably supported.
```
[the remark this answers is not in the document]
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
```diff
@@ -788,12 +803,19 @@ Cascading and Per-Element Highlight Styles</h3>
...
+
   When any supported property is not given a value by the cascade,
-  its value is determined by inheritance from
+  or given a value of ''inherit'' or ''unset'',
+  its [=specified value=] is determined by inheritance from
   the corresponding <a>highlight pseudo-element</a>
   of its <a>originating element</a>'s parent element
   (regardless of whether that property is an <a>inherited property</a>).
 
+  For the root element,
```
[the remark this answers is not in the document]
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
```diff
@@ -879,6 +890,43 @@ Painting the Highlight</h3>
     css/css-pseudo/highlight-painting-003.html
     </wpt>
 
+<h4 id="paired-cascade">
+Highlight Colors</h4>
+
+  For compatibility reasons,
+  [=default highlight colors=]
+  must only be [=used value|used=]
```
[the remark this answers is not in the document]
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
```diff
@@ -879,6 +890,43 @@ Painting the Highlight</h3>
...
+  when neither of the [=highlight colors=]
+  yield a [=cascaded value=]
+  or [=specified value=]
+  of the [=author origin=].
+  When a highlight color is ''revert'' or ''revert-layer'',
+  the origin <em>after</em> rolling back the cascade
+  determines whether this is the case.
```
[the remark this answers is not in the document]
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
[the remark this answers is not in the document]
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
delan: done in 4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
```diff
@@ -724,10 +732,33 @@ Default UA Styles</h3>
...
     :root::grammar-error  { text-decoration-line: grammar-error; }
-    /* Highlight targeted text */
-    :root::target-text    { color: MarkText; background: Mark; }
   </pre>
 
+  UAs may define <dfn>default highlight colors</dfn>
+  for each highlight pseudo-element,
+  consisting of a 'color' and a 'background-color',
+  and doing so is recommended
+  for ''::selection'' and ''::target-text''.
```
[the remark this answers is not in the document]
delan: i think these tweaks are already true in the current version?

css-pseudo-4/Overview.bs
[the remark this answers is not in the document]
delan: (that said, you’ve made a good point elsewhere about that distinction not being important to authors)

css-pseudo-4/Overview.bs
delan: i think you’re right, but only if we make it clearer for implementors that this is a used-value-time thing that doesn’t affect the computed value. imo it would be enough by linkifying “used” to used value again and not using the term “paired cascade”, despite its historical relevance as what we used in the note and in #6386

delan: done in e94418630480cf5c9ac640b207e446b33e890cc7

css-pseudo-4/Overview.bs
delan: done in 26b6c2b32e30d2ef84cc48365939ce682eb17482

css-pseudo-4/Overview.bs
delan: s/are/should correspond to/ done in https://github.com/w3c/csswg-drafts/commit/4e092c568e8a4247309ffff98fae7b172f2b4067

css-pseudo-4/Overview.bs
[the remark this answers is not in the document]
delan: done in 6a142341ace64adff79c1f622025630fb131b67a. i also eliminated the term “highlight colors” completely, which existed in the current draft but doesn’t really add any value as a separate term to “paired default highlight colors”, since the set of properties will never change (and we’re even adding a note to clarify this).

css-pseudo-4/Overview.bs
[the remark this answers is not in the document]
delan: done in e554b21821e32c57d746e7a70117b5c1aa4848db

css-pseudo-4/Overview.bs
```diff
@@ -724,16 +718,68 @@ Default UA Styles</h3>
     /* Represent default spelling/grammar error styling in an adjustable way */
     :root::spelling-error { text-decoration-line: spelling-error; }
     :root::grammar-error  { text-decoration-line: grammar-error; }
-    /* Highlight targeted text */
-    :root::target-text    { color: MarkText; background: Mark; }
   </pre>
 
+  Some [=highlight pseudo-elements=]
+  should have <dfn>paired default highlight colors</dfn>,
+  consisting of a default 'color' and 'background-color'
```
fantasai, line 725:
```suggestion
  should have <dfn>paired default highlight colors</dfn>--
  a default 'color' and 'background-color'
```

css-pseudo-4/Overview.bs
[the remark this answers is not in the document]
fantasai: OK, in that case maybe make it a Note? That way the information is presented simply here, but the normative text is elsewhere. (You do that by just adding "Note: " at the start of the paragraph.)

css-pseudo-4/Overview.bs
fantasai: > just based on impl opinions from Emilio and i (https://github.com/w3c/csswg-drafts/issues/6386#issuecomment-957389344).

I think this deserves its own issue and WG resolution

css-pseudo-4/Overview.bs
delan: done in 5084397ba8f811b011f8e6680489e7dff884298b

fantasai (approved):

fantasai:
Ok, looks good. Please SQUASH and rebase before merging. :)

delan:
Thank you both for all of your feedback!

PR #6666, opened 2021-09-22
[css-pseudo] text decorations on highlight pseudos (closes #6022)

### LINKED ISSUE #6022: [css-pseudo] questions about text decorations on highlight pseudos
https://drafts.csswg.org/css-pseudo-4/#highlight-text

Given the following content, with both the ::selection and ::target-text highlights marked {between the braces}:

```html
<span class="foo">before {foo</span>
<span class="bar">bar} after</span>
```
```css
.foo, .bar { text-decoration: yellow solid line-through; }
::target-text /* :root or *, doesn’t matter */ { background-color: orange; }
::selection /* :root or *, doesn’t matter */ { background-color: green; }
.foo::selection { color: cyan; text-decoration-color: magenta; /* no text-decoration-line */ }
.bar::target-text { color: xxx; text-decoration: red dotted line-through; }
.bar::selection { color: yyy; text-decoration: blue wavy line-through; }
```

1. What should be painted at “foo” and in what order (topmost last)?
   * a) yellow solid, orange, green, cyan “foo”
   * b) orange, green, cyan “foo”
   * c) orange, green, cyan “foo”, yellow solid
   * d) orange, green, cyan “foo”, cyan solid
   * e) orange, green, cyan “foo”, magenta solid
   * f) something else?
2. What should be painted at “bar” and in what order (topmost last)?
   * a) yellow solid, orange, red dotted, green, yyy “bar”, blue wavy
   * b) orange, red dotted, green, yyy “bar”, blue wavy
   * c) orange, green, yyy “bar”, yellow solid, red dotted, blue wavy
   * d) orange, green, yyy “bar”, yyy solid, red dotted, blue wavy
   * e) orange, green, yyy “bar”, yyy solid, yyy dotted, blue wavy
   * f) something else?

**See also: [resolution](https://github.com/w3c/csswg-drafts/issues/6022#issuecomment-800015071) and [edits](https://github.com/w3c/csswg-drafts/issues/6022#ref-commit-c170d9b), and my [two follow-up questions](https://github.com/w3c/csswg-drafts/issues/6022#issuecomment-800087413), and their [resolution](https://github.com/w3c/csswg-drafts/issues/6022#issuecomment-816229587).**

I thought the answer was 1=b 2=b at first, so that’s how I wrote [the ref](https://wpt.live/css/css-pseudo/highlight-painting-003-ref.html) for [css/css-pseudo/highlight-painting-003.html](https://wpt.live/css/css-pseudo/highlight-painting-003.html) ([permalink](https://crrev.com/c/2670609)), but after reading the spec and more closely, I’m starting to think it’s 1=d 2=d?

I’ve included 1=e above because I stumbled upon [this Chromium patch](https://crrev.com/c/2383212) ([bug 1110393](https://crbug.com/1110393)) today, which changes the color of originating element decorations to the ::selection’s *text-decoration*-color where that highlight is active. I suppose that change was a mistake or from before the current spec, because it sounds like it’s compliant when (and only when) text-decoration-color is currentColor?

Either way, I feel like we can improve the clarity of this section, given that I’m so unsure of the expected behaviour for scenarios like these. For example, there are several instances of “its” and “their” that could be more explicit, and some non-normative examples would be nice.

(cc @frivoal + @mrego)

delan:
This patch makes some further edits for #6022, clarifying that our rules for painting text decorations apply to both line decorations and emphasis marks, and that in effect, decorations will never be obscured by highlight backgrounds.


fantasai (approved):
Looks good, just one thing:

- Don't add the word “Other”, it's “{Text} and {Text Decorations}” not “{Text which is a type of Text Decoration} and {Other Types of Text Decorations}”!
- Squash+Rebase when you merge. (See drop-down next to GH's Merge button.)

delan:
Thanks @fantasai, done in 99cfe2049e01dec4d05146ee8b6617a5c6d18d14. Yeah, that heading would be better clarified by something like “Text, Line Decorations, and Emphasis Marks”, but the body makes the same clarification anyway so it’s not really necessary.

delan:
@fantasai looks like I don’t have write access, could you please merge?

frivoal:
I'm not @fantasai, but I do have merge access :) Merging.

Issue #6398, opened 2021-06-21
[css-contain-3][css-animations][css-transitions] Isolate animation side-effects?


andruud:
(Using "animations" loosely to refer to any of CSS/Web Animations/Transitions/):

There seems to be an underlying idea in the specifications that effects caused by animations should themselves not trigger new animations. We solve this by making things "not animatable", "animation-tainted", etc. With container queries this problem becomes a bit worse, since the animated size of some element can cause the size of a sibling container (for example) to re-evaluate its container query, hence applying a totally different style to its descendants, including new `animation-` properties. Probably we need some adjustment to the specs to define what should happen when such animation properties appear (or change) because of "animated re-evaluation" of container queries.

The css-transitions spec describes some alternate-reality-styles ([before-change style](https://drafts.csswg.org/css-transitions-1/#before-change-style) and [after-change style](https://drafts.csswg.org/css-transitions-1/#after-change-style)) which is used to determine from/to values. This _does_ give a behavior for container queries, but the before/after-change styles now implicitly become before/after-change _layouts_. The before-change style is particularly painful, since we would need to do _layout_ in a world where existing animations are updated to the current time, producing different animation effects, potentially causing a change in container query evaluation results, arbitrarily modifying the "previous" style and layout. This sounds pretty complicated to me, and I thought we should explore alternatives. And even if we want to keep css-transitions as is, we probably do need some update for css-animations.

What if we isolate all side-effects that happen because of stuff that's animating, such that it can't trigger new animations:

 - Let the _base style_ be the style as it would be without _any_ animation effects applied (of any kind), inheriting from the base style of the parent.
 - Let the _before-change style_ be the base style as of the previous style change event, inheriting from the before-change style of the parent. (I.e. just remember the previous base style).
 - Let the _after-change style_ be the base style as of the _current_ style change event, inheriting from the after-change style of the parent.
 - CSS transitions are triggered using the before-change style/after-change style. (Taking into account the current effect value for re-targeting purposes. The spec already covers that, though).
 - CSS animations are triggered using the after-change style.
 - Let the _animated style_ be the style as of the current style change event, _including_ all animation effects.
 - The `animation-`/`transition-` properties are ignored on the _animated style_, hence if they appear/disappear/update because of container queries re-evaluating, then nothing happens. ("Global animation-tainting").

[...]

PR #6672, opened 2021-09-23
[css-contain-3] Add Animated Containers section

andruud:


andruud:
Maybe something like this? (This is also not everything that needs to be edited for #6398, but I think it's the only part that goes into css-contain).

Note: This does not have a resolution backing it yet.

mirisuzanne (approved):
This makes sense to me

css-contain-3/Overview.bs
```diff
@@ -460,6 +461,67 @@ Container Queries: the ''@container'' rule</h2>
 	so it is not always possible to merge the individual <<container-query-list>>s
 	into a single query.
 
+<h3 id="animated-containers">
+Animated Containers</h3>
+
+	A change in the evaluation of a [=container query=] must be treated as a [=style change event=],
```
dbaron, line 467: I think "treated as a" should be "part of a", since there's no requirement that the style change event contain *only* that.

But really, the interesting bit is over in #6688, without which you'd also need to define here that the change in the evaluation of the container query is *not* reflected in the updating of the before-change style to the current time.

css-contain-3/Overview.bs
```diff
@@ -460,6 +461,67 @@ Container Queries: the ''@container'' rule</h2>
...
+			  &lt;div id=container>
+			    &lt;div id=inner>Inner&lt;/div>
+			  &lt;/div>
+			  &lt;div id=sibling>Sibling&lt;/div>
+			&lt;/main>
+		</pre>
+
+	</div>
+
+	This also applies for changes in [=computed values=] caused by [=query length=] units.
```
dbaron, line 522: "This also applies" is a bit vague (and could become worse if more text gets inserted nearby).

Perhaps, instead:

"Changes in computed values caused by query length units must also be part of a style change event"

?

dbaron (approved):
I think this is probably good with these suggested changes, but the interesting part of making this work is in #6688.

css-contain-3/Overview.bs
andruud: Done.

css-contain-3/Overview.bs
andruud: Good point, done.

PR #6673, opened 2021-09-23
[web-animation-2] Resolution of fill=auto for progress-based timelines

kevers-google:
[web-animation-2] Resolution of fill=auto for progress-based timelines

Update the rules for resolving a fill mode of 'auto' in timeline calculations based on discussion in github.com/w3c/csswg-drafts/issues/5223

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -2240,6 +2240,25 @@ partial dictionary OptionalEffectTiming {
     > <a>sequence effect</a>. The |specified end delay| is converted to an
     > <a>end delay</a> following the [=normalize specified timing=] procedure.
 
+:   <dfn>fill</dfn>
+::  Update the rules for resolving the special string value <code>auto</code>
+    in timing calculations as:
+
+    > <dl class="switch">
+    >
+    >     :   If <em>either</em> of the following conditions are true:
+    >         *    the <a>animation effect</a> to which the fill mode is being
+    >              is applied is not a <a>keyframe effect</a>,
+    >         *    the <a>animation effect</a> is associated with a
+    >              [=progress-based timeline=]
+    >     ::  Use <span class="enum-value">both</span> as the
+    >         <a>fill mode</a>.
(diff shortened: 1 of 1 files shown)


web-animations-2/Overview.bs
```diff
@@ -2248,6 +2248,25 @@ partial dictionary OptionalEffectTiming {
...
+::  Update the rules for resolving the special string value <code>auto</code>
+    in timing calculations as:
+
+    > <dl class="switch">
+    >
+    >     :   If <em>either</em> of the following conditions are true:
+    >         *    the <a>animation effect</a> to which the fill mode is being
+    >              is applied is not a <a>keyframe effect</a>,
+    >         *    the <a>animation effect</a> is associated with a
+    >              [=progress-based timeline=]
```
flackr, line 2261: I suspect we might want this to be as narrow as possible. E.g. if the animation active active phase does not touch the end of the timeline it probably shouldn't fill. How about we add something like "and the start delay + active duration = 100%"?

Also, does this need to be fill: both? The start time is inclusive so we only need to extend the end to make it end-inclusive. The complication is of course that we have to flip the fill when the animation direction is "backwards" but this seems not too complicated.

flackr: @birtles FYI

birtles: I think the first condition is in agreement with the updated definition for 'auto' here: https://drafts.csswg.org/web-animations-2/#the-fillmode-enumeration but perhaps we could phrase it to be more generic such that each type of animation effect defines what "auto" means to it? e.g. introduce a term for that? intrinsic fill mode, default fill mode or something?

For the second condition I'm not sure. In general I prefer a wider definition because it feels less magical and more composable if there are fewer rules to consider. Furthermore, "fill: both" would mean we don't need to worry about the reversing case. However, I don't have a good sense for the implications of extending the fill mode to scroll-linked content. Is there any example content (real or artificial) that highlights the difference?

Issue #5394, opened 2020-08-04
[web-animations] Could commitStyles always write the end state of the animation?


zmajeed:
MDN https://developer.mozilla.org/en-US/docs/Web/API/Animation/commitStyles says:

"The commitStyles() method of the Web Animations API's Animation interface commits the end styling state of an animation to the element being animated, even after that animation has been removed. It will cause the end styling state to be written to the element being animated, in the form of properties inside a style attribute."

This is exactly what I want and I like the example in which `commitStyles()` is called right after the animation is created. The expected behavior is very clear since the end state of an animation is well-defined

To my surprise in Chrome 84 this is not the case. Looking at the spec it seems MDN got it wrong:

https://drafts.csswg.org/web-animations/#dom-animation-commitstyles says:

"Writes the current effect values produced by this animation’s animation effects to their corresponding effect targets' inline style using the commit computed styles procedure."

So an animation of `transform: translate(100px), duration: 1000` may record a translation of anywhere between 0 and 100px depending on when `commitStyles()` is called. Sure calling it after the animation finishes insures the end state is written but seems to also require setting `fill: fowards`. It just seems much easier and more intuitive to have it record a predictable state no matter when it's called

Could `commitStyles()` have an option to always record the end state preferably without involving the `fill` property. Even better could it become an animation property and I wouldn't have to use `fill` at all.

birtles:
That's an interesting proposal. It certainly seems feasible to add a flag to `commitStyles` to produce the behavior you describe. Do you mind explaining the particular use case you have for it?

From my understanding, the proposal would be to effectively seek to the target effect end (but without firing events etc.), force the fill mode for all associated effects to `both`, and then apply the procedure to commit computed styles.

Understanding the use cases would give a better idea of whether this should be two separate flags or not (e.g. one to seek to the target end, one to force the fill mode). It would also give a better idea of the priority of this feature.

zmajeed:
My application is a large state graph of vertices and edges. Each state transition animates the entire graph so the destination vertex (new current state) is centered in the view box. The new position of the graph needs to persist. There's a timeline slider to play the transitions from any point forward and backward.

birtles:
It seems like what this case needs is something like:

```js
function finishTransition(anim) {
  // Make sure finish() seeks to the target effect end
  anim.playbackRate = 1;
  anim.finish();
  anim.effect.updateTiming({ fill: 'both' }); // Or 'forwards'
  anim.commitStyles();
[...]

PR #6688, opened 2021-09-27
[css-transitions-1] Define before/after-change style in terms of base

andruud:
- Introduce "base style".
 - Define before/after-change style in terms of that base style
   (at two different points in time).
 - There should be no need for the "completed transition" set anymore,
   since the before/after-change styles on a child are blind to any
   effects from transitions/animations happening on the parent.

See also #6398.


andruud:
This needs a resolution, but proposing an edit ahead of time so we know exactly what to resolve on.

css-transitions-1/Overview.bs
```diff
@@ -432,54 +432,6 @@ Starting of transitions {#starting}
...
-          a <a>completed transition</a> for the same property and element.
-        </span>
-      </p>
-
-      <p>
-        If an element is no longer in the document,
-        implementations must <a>cancel</a> any <a>running transitions</a>
-        on it and remove transitions on it from the
-        <a>completed transitions</a>.
-      </p>
```
dbaron, line 453: You should leave the rule that implementations cancel running transitions when an element is removed from the document (so leave the sentence except for the "and" and what follows).

dbaron (approved):
I think this looks good to me (although it has some compatibility risks... though perhaps low given that current implementations don't agree very much), but I think this either needs working group consensus or at least approval from the other implementors of transitions, in #6398.

It's possible the list could be further simplified at this point, but it will be easier to tell once this lands.

css-transitions-1/Overview.bs
andruud: Oops, absolutely. Thanks.

birtles:
I think this represents a change to case 11 from https://bug1192592.bmoattachments.org/attachment.cgi?id=8843824. That is, I believe with this proposal it would be possible to trigger a transition while an animation is running. Maybe that's ok but it does seem a bit risky in terms of compatibility, since browsers appear to agree on that at the moment.

Also, I'm really rusty on how all this fits together but I believe in Gecko at least SMIL animated styles are written to an override stylesheet separate to the animations level of the cascade whereas this PR seems to assume they are written to the animations level of the cascade.

birtles:
> * You'd think that it's uncommon for authors to specify a transition and an animation on the same property. If you like I can add a (Chrome) use-counter to see how often a would-be transition is "blocked" by a running animation, or
> 
> * We could _explicitly_ specify in https://drafts.csswg.org/css-transitions-1/#starting that a transition should not start if there's currently an animation on the same property. I don't think that behavior _must_ arise from the current definition of before/after-change style?

Thanks. I'm not sure which is preferable. If there's time, adding the use counter would be safest.

If we decide to change this behavior, I guess it would be little odd that transition events are fired even when there's no visual change. Maybe it's ok though.

birtles:
> @birtles I have now added the use-counter to Chrome to investigate. I'll report back in a few months when good data comes in.

Thank you!

andruud:
Yes, but not the update I wanted. The use-counter is [at almost 4%](https://chromestatus.com/metrics/feature/timeline/popularity/4066), which is _way_ higher than expected. I've been waiting for sample pages to come in so I could check if there's any typical usage patterns we could use to argue that the change doesn't matter, but at 4% it might be futile to find the "typical" usage patterns anyway.

So as it stands, the use-counter did not help in making a confident decision here, unfortunately.

@birtles In light of this, do you think we should do the following?

> We could explicitly specify in https://drafts.csswg.org/css-transitions-1/#starting that a transition should not start if there's currently an animation on the same property. I don't think that behavior must arise from the current definition of before/after-change style?

birtles:
> @birtles In light of this, do you think we should do the following?
> 
> > We could explicitly specify in https://drafts.csswg.org/css-transitions-1/#starting that a transition should not start if there's currently an animation on the same property. I don't think that behavior must arise from the current definition of before/after-change style?

I think so. Transition events are quite important to a lot of apps -- often various state changes are keyed off getting a `transitionend` event. I think any change to the circumstances under which transition events are dispatched is risky and we should try to match the current (interoperable) behavior there.

(Anecdotally, many many years ago when working on Firefox OS, a recurring cause for apps breaking was that they didn't receive expected `transitionend` events which is what motivated specifying the `transitioncancel` event.)

andruud:
OK, thanks @birtles. I'll update this PR with new proposed text as soon as I can.

andruud:
Trying to pick this back up (at the request of @dbaron), I have updated this PR:

- Excluded SMIL explicitly from the base style.
- Blocked transitions from starting when they are "affected by animations".
- @flackr's commitStyles() example seemed quite convincing to me, so I added a version of that as an example.

Please review closely.

birtles:
> I started reviewing this today, just briefly, but I'm still trying to get my head around how this will work with #5394 (and whether we can truly do without #11085).

So to capture my working, in #5394 we're proposing that calling `commitStyles` on an animation that doesn't fill and has just finished captures the final value (i.e. ignores the endpoint exclusive timing).

Suppose we have an animation from width 100px to 200px, no fill mode, and we call `commitStyles` as part of its finished event.

Let's assume the unanimated width (i.e. base style width) is 50px (just to make it easier to distinguish the animation style from the base style).

1. `commitStyles` is called.
2. We flush styles to ensure everything is up-to-date.
3. At this point the inline style has yet to be updated so there is no change to the base value (it's still 50px) and no transition is triggered.
4. We calculate the final value of the animation as 200px and update the inline style to 200px.
5. On the next style change event, we notice that the base width has gone from 50px to 200px. Furthermore, during the previous style change event (2~3) `width` was not "affected by an animation" and hence we trigger a transition—which I don't think we intend to.

Does that seem right?

css-transitions-1/Overview.bs
```diff
@@ -920,14 +875,35 @@ Starting of transitions {#starting}
...
+  });
+&lt;/script&gt;
+          </pre>
+          <p>
+            The call to <code>commitStyles()</code>
+            will trigger a <a>style change event</a>,
+            with the <a href="#before-change-style">before-change</a>
+            <code>width</code> being <code>100px</code>,
+            and the <a href="#after-change-style">after-change</a>
+            <code>width</code> being <code>200px</code>.
```
birtles, line 900: I don't think this is right? The style change event occurs before the inline style is updated so the `before-change` and `after-change` width at this point should both be 100px?

andruud:
@birtles Ah, I was not aware that `commitStyles` is specified to "applying any pending style changes" before doing its thing. That is indeed a problem.

Issue #5223, opened 2020-06-16
[scroll-animations-1] Effects for scroll linked animations are not inclusive at max scroll


JTensai:
Effects for scroll linked animations do not handle the inclusive end exception established by #37

When scroll timeline end scroll offset is default, reaching max scroll should result in the effect remaining at its last keyframe value. Instead it finishes. We need a way to allow the effect to detect that it should not finish when its corresponding timeline is at its end point (and satisfies the inclusive requirements).

andruud:
Isn't this handled well by [The phase of a ScrollTimeline](https://drafts.csswg.org/scroll-animations-1/#phase-algorithm) already?

(BTW the linked PR seems irrelevant).

flackr:
Issue #4325 introduces the various issues in this area. The phases solved the issue of the ScrollTimeline becoming inactive, but this particular issue is (I believe) that for the naive case, e.g.

```js
element.animate({transform: ['none', 'translateY(200px)']}, { duration: 10000, timeline: new ScrollTimeline()});
```

When you reach the maximum scroll offset, the timeline produces time value 10000, which means that the animation is in the after phase due to https://www.w3.org/TR/web-animations-1/#animation-effect-phases-and-states

@JTensai I believe your proposal from the other issue was:
> Update animation and animation effect active time and phase calculation to take the timeline state (or hold phase) into account.

At a high level this sounds reasonable to me.

majido:
Given that we need to change the animation and effect timing calculation to change, I would argue it is probably better to no longer have the current special-case logic on ScrollTimeline. Basically just have it so that once timeline reaches its last position it is in "after" phase which is the normal behavior.

Then we just make the keyframe to have a fill forward or act as if it has one. If keyframe already has explicitly a fill value we honor that which allows authors a way to escape this behavior.

Basically I propose something like this:
 - Consider ScrollTimeline with scroll range 100% to be *terminal finite*. (I am sure there are probably better terms for this but I am just using this to just mean that it is not possible to go beyond their last time value).
 - When [setting timeline of an animation](https://drafts.csswg.org/web-animations/#setting-the-timeline), it that timeline is *terminal finite*, then consider that animation to be *prefers fill-forward*.
[...]

PR #6702, opened 2021-09-30
[web-animations-2] Make active boundary times inclusive if timeline duration is finite

kevers-google:
[web-animations-2] Make active boundary times inclusive if timeline duration is finite

Addresses issue #5223. The discussion suggested adding rules for resolving fill "auto".  This approach introduces edge cases as indicated in comments on https://github.com/w3c/csswg-drafts/pull/6673.  In short, resolving fill 'auto' introduces potential surprising behavior when delays or timeline offsets are non-zero.  The before and after phase calculations had explicit rules to make an active boundary exclusive depending on the direction of the animation.  This PR restricts these exclusion rules to non-finite timelines.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -1096,8 +1096,9 @@ with:
 >     <em>and</em> <em>either</em> of the following conditions are met:
 >     1.  the <a>local time</a> is less than the <a>before-active boundary
 >         time</a>, <em>or</em>
->     1.  the <a>animation direction</a> is "backwards" and the
->         <a>local time</a> is equal to the <a>before-active boundary time</a>.
+>     1.  [=timeline duration=] is unresolved, the <a>animation direction</a>
+>         is "backwards" and the <a>local time</a> is equal to the
+>         <a>before-active boundary time</a>.
 
 Replace:
 
@@ -1122,8 +1123,9 @@ with:
 >     <em>and</em> <em>either</em> of the following conditions are met:
 >     1.  the <a>local time</a> is greater than the <a>active-after boundary
 >         time</a>, <em>or</em>
->     1.  the <a>animation direction</a> is "forwards" and the <a>local time</a>
->         is equal to the <a>active-after boundary time</a>.
+>     1.  [=timeline duration=] is unresolved, the <a>animation direction</a> is
+>         "forwards" and the <a>local time</a> is equal to the
+>         <a>active-after boundary time</a>.
 
 <h4 id="fill-modes">Fill modes</h4>
 


web-animations-2/Overview.bs
```diff
@@ -1096,8 +1096,9 @@ with:
 >     <em>and</em> <em>either</em> of the following conditions are met:
 >     1.  the <a>local time</a> is less than the <a>before-active boundary
 >         time</a>, <em>or</em>
->     1.  the <a>animation direction</a> is "backwards" and the
->         <a>local time</a> is equal to the <a>before-active boundary time</a>.
+>     1.  [=timeline duration=] is unresolved, the <a>animation direction</a>
+>         is "backwards" and the <a>local time</a> is equal to the
+>         <a>before-active boundary time</a>.
```
flackr, line 1101: This behavior of making the endpoints inclusive when timeline duration is resolved is a strange special case to have. This will also introduce many of the same edge cases from #6673 even though they will be limited to the endpoint of one animation, e.g.:

```js
element.animate({transform: ['translateY(0px)', 'translateY(10px)']}, {duration: CSS.percent(10), timeline: scrollTimeline, composite: 'add'}); 
element.animate({transform: ['translateY(10px)', 'translateY(50px)']}, {duration: CSS.percent(10), startDelay: CSS.percent(10), timeline: scrollTimeline, composite: 'add'})
```

At time 10% both effects will be active resulting in a transform of 20px. The same thing will happen for back to back effects in a SequenceEffect.

flackr: FWIW, I do like the approach of making the active phase endpoint inclusive more than the auto fill mode, it just needs to only apply when the effect reaches the end of the timeline.

flackr: Perhaps the simplest way to address this is calculate the effective timeline time by undoing the current time calculation, e.g.
effective timeline time = (local time / playback rate) + start_time

And then only use the endpoint inclusive condition if the effective timeline time is 0 (and backwards) or 100% (and forwards)

kevers-google: It appears that this reverse engineering of effective timeilne time is not quite right for sequence effects since local time resets to zero at the start of each effect in the sequence.  Can likely come up with an equation based on inherited time to determine if aligned with the boundary of the timeline.  Just to be clear, are we only applying if scrollOffset = [0%, 100%] (or auto) and not at an arbitrary effective start or end offset?

flackr: local_time is zero, but the start_time would be non-zero wouldn't it? Of course there may be simpler approaches that make it less likely we'll have to deal with equality of values.

Dealing with the actual max scroll offset is a good question, I think ideally we'd only apply this for the 100% / auto case so that stringing together animations on separate scroll timelines which stop when the next one starts also doesn't result in any overlap.

kevers-google: Inferring the timeline time based on the inherited time and start time is only strictly correct for a running animation.  If the animation is paused we cannot tell if there was any lead/lag between the timeline and animation currentTime at the time of the pause.  Added an issue highlighting the problem. This discrepancy would only trigger an unexpected discrepancy for edge cases such as setting current time when pausing a progress-based animation precisely at the timeline boundary.  I don't believe these potential edge cases will be frequently triggered.

flackr (commented):
I think this works but I'd appreciate a second set of eyes on this, @birtles would you be able to take a look?

web-animations-2/Overview.bs
```diff
@@ -1066,13 +1066,36 @@ condition:
...
+    :   If the associated <a>animation</a>'s [=timeline duration=] is
+        unresolved,
+    ::  &rarr; unresolved
+
+    :   If the associated <a>animation</a>'s startTime is unresolved,
+    ::  &rarr; <a>inherited time</a>
+
+    :   Otherwise,
+    ::  <code><a>inherited time</a> / <a>animation</a>'s
+        playbackRate + <a>animation</a>'s startTime</code>
```
birtles, line 1090: Similarly here, for consistency it might be better to use the model terms instead of the interface member names?

birtles (commented):
From what I can tell the approach seems reasonable but I have a few questions about some of the minor details.

web-animations-2/Overview.bs
```diff
@@ -1066,13 +1066,36 @@ condition:
...
+::  Set based on the first matching condition:
+
+    <dl class="switch">
+
+    :   If the associated <a>animation</a>'s [=timeline duration=] is
+        unresolved,
+    ::  &rarr; unresolved
+
+    :   If the associated <a>animation</a>'s startTime is unresolved,
+    ::  &rarr; <a>inherited time</a>
```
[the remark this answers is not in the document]
kevers-google: Greatly simplified the logic here. Normalizing the time to convert to a fraction to be more consistent with other uses of the term 'progress'.

web-animations-2/Overview.bs
kevers-google: Done.

web-animations-2/Overview.bs
```diff
@@ -1066,13 +1066,36 @@ condition:
...
+    :   Otherwise,
+    ::  <code><a>inherited time</a> / <a>animation</a>'s
+        playbackRate + <a>animation</a>'s startTime</code>
+
+    </dl>
+
+Issue: [=effective timeline progress=] is not strictly correct for a paused
+animation if the <a>animation</a>'s currentTime is explicitly set, as this can
+introduce a lead or lag, between the <a>timeline<a>'s currentTime and
+<a>animation</a>'s currentTime.
```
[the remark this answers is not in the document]
kevers-google: Similar to other cases of paused scroll-driven animations going out of sync.

web-animations-2/Overview.bs
```diff
@@ -1096,8 +1119,10 @@ with:
 >     <em>and</em> <em>either</em> of the following conditions are met:
 >     1.  the <a>local time</a> is less than the <a>before-active boundary
 >         time</a>, <em>or</em>
->     1.  the <a>animation direction</a> is "backwards" and the
->         <a>local time</a> is equal to the <a>before-active boundary time</a>.
+>     1.  The <a>animation direction</a> is "backwards", the <a>local time</a>
+>         is equal to the <a>before-active boundary time</a> and
+>         [=effective timeline progress=] is not equal to
+>         <code>CSS.percent(0)</code>.
```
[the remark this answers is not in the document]
kevers-google: Done. Now that effective timeline progress is normalized, it'll be unresolved, or between 0 and 1.

web-animations-2/Overview.bs
```diff
@@ -1066,13 +1066,36 @@ condition:
...
+:   <dfn>effective timeline progress</dfn>
+::  Set based on the first matching condition:
+
+    <dl class="switch">
+
+    :   If <em>all</em> of the following conditions are true:
+        *   the associated <a>animation</a>'s [=timeline duration=] is resolved
+            and non-zero, and
+        *   the associated <a>animation</a>'s [=playback rate=] is 1, and
+        *   the assiciated <a>animation</a>'s [=start time=] is unresolved or 0
```
birtles, line 1085: nit: s/assiciated/associated/

web-animations-2/Overview.bs
```diff
@@ -1066,13 +1066,36 @@ condition:
...
 
+:   <dfn>effective timeline progress</dfn>
+::  Set based on the first matching condition:
+
+    <dl class="switch">
+
+    :   If <em>all</em> of the following conditions are true:
+        *   the associated <a>animation</a>'s [=timeline duration=] is resolved
+            and non-zero, and
+        *   the associated <a>animation</a>'s [=playback rate=] is 1, and
```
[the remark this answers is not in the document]
kevers-google: Was previously trying to avoid some edge cases for paused animations, but in the process, it needlessly restricted check on running animations.  in this version, we are assuming that the timeline and animation are aligned when paused, but otherwise adjust for changes to the playback rate and/or start time.  Not sure of a better option unless storing off a "hold timeline time" instead of the "hold phase".  Having said that, getting the phase at a boundary wrong for a scroll after pause animation is likely not that critical.  Authors can always specify a fill to get the expected behavior if they expect that the scroll-linked animation will be intermittently paused.

In short, generalized "effective timeline progress" and added "at timeline boundary".  The boundary does not discriminate based on direction since the boundary's counterpart is already inclusive.  We are making both endpoints inclusive if "at timeline boundary" is set.

birtles (approved):
This looks really good to me. The only thing I'm not sure about is what the behavior should be when the duration of the timeline is zero. Are we sure it should always return false in that case?

web-animations-2/Overview.bs
```diff
@@ -1066,13 +1066,41 @@ condition:
...
+    1.  Let <var>effective timeline progress</var> be
+        <code><var>effective timeline time</var> / [=timeline duration=]</code>
+
+    1.  If <var>effective timeline progress</var> is 0 or 1, return true,
+        otherwise false.
+
+Issue: This procedure is not strictly correct for a paused
+animation if the <a>animation</a>'s current time is explicitly set, as this can
+introduce a lead or lag, between the <a>timeline<a>'s current time and
+<a>animation</a>'s current time.
```
flackr, line 1102: I'm not sure I understand this issue from the note. It might be helpful to have a specific example, e.g. if a paused animation is set to time 100% the effective timeline time will be Y%.

flackr (approved):
Overall, this looks good to me. I think that the condition should probably be based on whether it's a progress based timeline rather than based on the animation's duration.

birtles:
> 
> 
> I'm not sure that the use case is for a zero-duration animation. I could be convinced either way to make the boundary inclusive or exclusive. My preference is exclusive. The rationale for my preference is that a zero-duration is never active for an animation with a monotonic timeline. You'll be in the after phase if direction is forward, or before phase if direction is backwards. If we make it active, what is the progress as the scroll position changes? Are we stuck at the start or end of the animation? It seems challenging to define "expected" behavior.

Thanks, that makes sense to me.

birtles:
> Overall, this looks good to me. I think that the condition should probably be based on whether it's a progress based timeline rather than based on the animation's duration.

I didn't quite follow this -- I don't think the PR is based on the animation's duration? Or is this just confirming the direction of the PR is correct?

flackr:
> > Overall, this looks good to me. I think that the condition should probably be based on whether it's a progress based timeline rather than based on the animation's duration.
> 
> I didn't quite follow this -- I don't think the PR is based on the animation's duration? Or is this just confirming the direction of the PR is correct?

I was specifically referring to this part:

> If <em>either</em> of the following conditions are true:
>        *   the associated <a>animation</a>'s [=timeline duration=] is
>            unresolved or zero

I think that if we have other fixed duration timelines it doesn't necessarily hold that we would want them to be endpoint inclusive.

birtles:
> I was specifically referring to this part:
> 
> > If _either_ of the following conditions are true:
> > *   the associated animation's [=timeline duration=] is
> > unresolved or zero

Oh, I see, sorry I was looking for a reference to the animation duration, not the animation timeline duration! 😅

> I think that if we have other fixed duration timelines it doesn't necessarily hold that we would want them to be endpoint inclusive.

Yes, that's a good point. That suggests the term should probably be renamed too then, since "timeline boundary" suggests it applies to any timeline. If we're going to scope this to progress-based timelines (or whatever the current term is), I guess this definition should match that (e.g. "at progress timeline boundary" or whatever).

web-animations-2/Overview.bs
```diff
@@ -1073,10 +1073,12 @@ an <a>animation effect</a>.
...
     the <a>animation</a> timeline's [=timeline phase=].
 
-:   <dfn>at timeline boundary</dfn>
+:   <dfn>at progress timeline boundary</dfn>
 ::  Determined using the following procedure:
 
-    1.  If <em>either</em> of the following conditions are true:
+    1.  If <em>any</em> of the following conditions are true:
+        *   the associated <an>animatoin</a> does not have a
+            [=progress-based timeline=], or
```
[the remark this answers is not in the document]
kevers-google: Trying to avoid the awkwardness of using "associated" twice in the same sentence: once for the animation being associated with the effect, and a second time for the timeline being associated with the animation.

https://drafts.csswg.org/web-animations-2/#timelines:txt=timeline%20is%20a%20progress-based%20timeline

web-animations-2/Overview.bs
```diff
@@ -1073,10 +1073,12 @@ an <a>animation effect</a>.
 ::  The associated <a>animation</a>'s <a>hold phase</a> if resolved, otherwise
     the <a>animation</a> timeline's [=timeline phase=].
 
-:   <dfn>at timeline boundary</dfn>
+:   <dfn>at progress timeline boundary</dfn>
 ::  Determined using the following procedure:
 
-    1.  If <em>either</em> of the following conditions are true:
+    1.  If <em>any</em> of the following conditions are true:
+        *   the associated <an>animatoin</a> does not have a
```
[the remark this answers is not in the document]
kevers-google: done

PR #6712, opened 2021-10-04
[web-animations-2]  Update ComputedEffectTiming

kevers-google:
[web-animations-2]  Update ComputedEffectTiming

This PR fixes descriptions in ComputedEffectTiming to remove ambiguity over the return value types. Specifically, effects linked to animations with timelines that have an unresolved duration should return times expressed as doubles in milliseconds and not as CSSNumericValues.  Only progress based animations should return their time values expressed as CSS.percent.  

The progress attribute was previously changed to be of type CSSNumberish but without a description change. This property is a fraction and not a time.  Thus, it should not have been changed.  The change is  reverted in this PR.

A few formatting fixers were applied to clearly show the type for each attribute in the "members" section.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -2448,7 +2448,7 @@ partial dictionary ComputedEffectTiming {
     CSSNumberish         endTime;
     CSSNumberish         activeDuration;
     CSSNumberish?        localTime;
-    CSSNumberish?        progress;
+    double?              progress;
 };
 </pre>
 
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -2448,7 +2448,6 @@ partial dictionary ComputedEffectTiming {
     CSSNumberish         endTime;
     CSSNumberish         activeDuration;
     CSSNumberish?        localTime;
-    double?              progress;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


web-animations-2/Overview.bs
```diff
@@ -2448,16 +2448,17 @@ partial dictionary ComputedEffectTiming {
...
 };
 </pre>
 
 <div class='members'>
 
 :   <dfn dict-member for=ComputedEffectTiming>startTime</dfn>
 ::  The <a lt='animation effect start time'>start time</a> of this
-    <a>animation effect</a> in milliseconds or as a percentage of the
+    <a>animation effect</a> expressed as a double in milliseconds
+    if [=timeline duration=] is unresolved or as a percentage of
```
flackr, line 2459: We seem to be using an unresolved timeline duration as an alternative to associated with a progress based timeline, however the latter is a better reason for the startTime to be expressed as a percentage - whereas if we had a finite time-based timeline I could see the time being expressed either way and most likely using a double for compatibility.

flackr (changes requested):

web-animations-2/Overview.bs
kevers-google: Switched to being explicit about using % only with a progress based timeline.  Switched ordering to avoid negative condition or otherwise, which feels awkward.

Issue #6498, opened 2021-08-04
[css-highlight-api] Figure out how highlights are exposed to the accessibility tree


frivoal:
@atanassov recently reminded us that figure out how highlights are exposed to the accessibility tree. This had been know for a while, but we didn't have an issue I could find, so here's one.

frivoal:
To sighted users, the highlights are only visible through the way they are styled. There is no other way to know what they're for. Maybe the same should be true of the highlights? Getting the highlight itself described through some text sounds like it might be more convenient, but I'm not sure where we'd pull that description from, how to deal with multiple languages, or the potential discrepancies between the descriptive text and what the highlight is actually used for.

astearns:
And whatever gets decided here should likely match what is decided for #6417

frivoal:
How are things like these exposed to the accessibility tree today?
* `::spelling-error`
* `::grammar-error`
* `::target-text`

We don't necessarily have to do the same, but it'd be good to start with an understanding of how these are handled.

ffiori:
#4601 seems to have some related remarks.

dlibby-:
UserAgent detected spelling and grammar errors (i.e. ::spelling-error and ::grammar-error) in Chromium do have a mapping to various AT APIs. These are serialized as part of the accessibility tree and as far as I can tell, this is not part of any of the Aria-related standards. For example on UIA on Windows spelling errors are exposed as a text range with an AnnotationType_SpellingError. Not sure how this would fit into the spec, but perhaps is a path forward (following the UIA example, there is an AnnotationType_Highlight).

For completeness, spelling and grammar errors that are computed by the webpage can be annotated with [aria-invalid](https://www.w3.org/TR/wai-aria/#aria-invalid). The corresponding concept for highlights would probably be the mark role (see https://w3c.github.io/aria/). However, these require additional DOM nodes, which is counter to the purpose of the HighlightAPI.

dandclark:
I agree with @dlibby that for the question of how to expose the highlight ranges to ATs we should be able to follow the lead of how ::spelling-error and ::grammar-error are exposed in Chromium, at least in part.

For expressing the semantics of the highlight ranges, the most flexible option seems to be to add a `title` string attribute to `Highlight`, which would be exposed to ATs. However, as pointed out in https://github.com/w3c/csswg-drafts/issues/4601#issuecomment-806426095, these strings wouldn't be localized. This approach also depends on the author to provide a string with the correct level of descriptiveness.

[...]

PR #6724, opened 2021-10-11
[css-highlight-api-1] Add `type` enum attribute to `Highlight` #6498

dandclark:
Per resolution of #6498, add a `HighlightType` enum attribute to `Highlight`. This will be used to convey the semantics of the `Highlight` for accessibility tools.

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -161,15 +161,24 @@ Creating Custom Highlights</h3>
 	See [[#range-invalidation]] for more details about this choice and its implications.
 
 	<xmp class="idl">
+	enum HighlightType {
+		"highlight",
+		"spelling-error",
+		"grammar-error"
+	};
+
 	[Exposed=Window]
 	interface Highlight {
 		constructor(AbstractRange... initialRanges);
 		setlike<AbstractRange>;
 		attribute long priority;
+		attribute HighlightType type;
 	};
 	</xmp>
 
 	See [[#priorities]] for more information on the {{Highlight/priority}} attribute.
+
+	See [[#highlight-types]] for more information on the {{Highlight/type}} attribute.
 
 	<div algorithm="to create a custom highlight">
 		When the <dfn for=Highlight constructor>Highlight(AbstractRange... initialRanges)</dfn> constructor is invoked,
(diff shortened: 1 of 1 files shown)


css-highlight-api-1/Overview.bs
```diff
@@ -450,6 +461,26 @@ Priority of Overlapping Highlights</h4>
...
+
+	A custom highlight will have a default type of {{HighlightType/highlight}}
+	if its {{Highlight/type}} attribute has not been explicitly set.
+
+	Note: Authors should set a [=custom highlight=]'s {{Highlight/type}} to
+	{{HighlightType/spelling-error}} when that [=custom highlight=] is being used to emphasize
+	misspelled content. Authors should set a [=custom highlight=]'s {{Highlight/type}} to
+	{{HighlightType/grammar-error}} when that [=custom highlight=] is being used to emphasize
+	content that is grammatically incorrect. For all other use cases {{Highlight/type}} should
+	be left as {{HighlightType/highlight}}.
```
sanketj, line 479: Should we also add some comments about what UAs should do with the type?

css-highlight-api-1/Overview.bs
```diff
@@ -450,6 +461,26 @@ Priority of Overlapping Highlights</h4>
 		</div>
 	</div>
 
+<h4 id=highlight-types>
+Highlight types</h4>
+
+	A [=custom highlight=]'s {{Highlight/type}} attribute is used by authors to specify the semantic
+	meaning of the highlight. This allows assistive technologies to include this meaning when
```
sanketj, line 468: I think it is important to explain here why we have chosen this initial set of types. That can also tie into the note on line 481.

css-highlight-api-1/Overview.bs
dandclark: Added some normative text and another Note about this.

sanketj (approved):

dandclark:
@frivoal / @sanketj can this be landed, since we [resolved](https://github.com/w3c/csswg-drafts/issues/6498#issuecomment-937321727) a while back to add this enum?

sanketj:
> @frivoal / @sanketj can this be landed, since we [resolved](https://github.com/w3c/csswg-drafts/issues/6498#issuecomment-937321727) a while back to add this enum?

Yes, LGTM

Issue #6351, opened 2021-06-05
[css-scrollbars] Add `wide` value to `scrollbar-width`


WesleyBranton:
SPEC: https://drafts.csswg.org/css-scrollbars-1/#scrollbar-width

The current spec only allows `auto`, `thin` and `none` for the scrollbar-width CSS property. While both `thin` and `none` are standard across multiple platforms, the `auto` value currently varies depending on the user's platform, as it implements the "default platform scrollbar width".

This creates an inconsistent cross-platform experience, as many versions of both Linux and Mac display a scrollbar that's similar to `thin`. This could create concerns for users with specific accessibility needs.

The specification could be similar to the `thin` specification and read as such:

> Implementations should use wider scrollbars than auto when applicable. This may mean a wide variant of scrollbar provided by the platform, or a custom scrollbar wider than the default platform scrollbar.

frivoal:
While I agree that users can desire wide scrollbars, I don't see in what situation an author would want to control in via stylesheets for specific parts of their site. If users need to be able to say "I want the `auto` value to be wide", that's something they should do in OS or browser settings.

Is there a significant use case for **author** control for wider scrollbars?

frivoal:
Since you seem to be satisfied with the logic in https://github.com/w3c/csswg-drafts/issues/6351#issuecomment-877496097, does closing this wontfix seem fine?

WesleyBranton:
I don't really agree with the logic. The same could be said for the "thin" option. When is there really a case where the author wants to control the scrollbar on a specific part of the website to make it thin?

Just because there may not be a clear use case for it, doesn't mean that there isn't a use case. There's no real necessary use case for the "thin" option either, other than a designer's preference.

Perhaps a website that sells glasses wants to have a larger scrollbar on parts of their website because it makes it more functional for people with vision impairments (their customer base). Perhaps a website designer wants to use a wider scrollbar on a long scrolling section to compensate for the fact that the height of the scrollbar thumb is shorter.

Having a "wide" option offers some functional advantages for designers, as opposed to the "thin" option, which only has cosmetic advantages.

frivoal:
> When is there really a case where the author wants to control the scrollbar on a specific part of the website to make it thin?

> There's no real necessary use case for the "thin" option either, other than a designer's preference.

[...]

PR #6750, opened 2021-10-21
[css-scrollbars] Add notes and advisements to scrollbar-width

frivoal:
This clarifies the purpose of the property, as well as how users (not authors) are ultimately in charge of this UI component.

See #6351

### DIFF
--- a/css-scrollbars-1/Overview.bs
+++ b/css-scrollbars-1/Overview.bs
@@ -402,3 +402,31 @@ or on overflowing elements with scrollbars in the page.</p>
 <p>No.</p>
 </li>
 </ol>
+
+<h2 class="no-num" id="accessibility-considerations">Appendix D. Considerations for accessibility</h2>
+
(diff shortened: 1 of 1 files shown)


css-scrollbars-1/Overview.bs
```diff
@@ -230,6 +257,13 @@ authors should use ''overflow: hidden'' instead of ''scrollbar-width: none''.
 </dd>
 </dl>
 
+<div class=note>Note: Users who find the ''thin'' style of scrollbars unusable
+can include the following rule in their [=user style sheet=]:
+<pre><code class=lang-css>* { scrollbar-width: auto !important; }</code></pre>
```
emilio, line 262: We should probably recommend UAs a way to ignore `scrollbar-width: thin` instead, as this would also remove the effect of `scrollbar-width: none`, which might be undesirable.

emilio (approved):

PR #6795, opened 2021-11-04
[css-mediaqueries-5] nav-controls: Clarify use in a boolean context (closes #6785)

### LINKED ISSUE #6785: [css-mediaqueries-5] nav-controls future proofing concerns (should each control have its own feature?)
Follow-up to #4187 (adding `nav-controls` media feature).

Draft spec text: https://drafts.csswg.org/mediaqueries-5/#nav-controls

This feature seems to work when the only queryable feature is `"back"`, but I don't see how this can be expanded in the future to support querying new features. As an example, in #4187 @grorg suggested a future ability to query for a user-agent-supplied Share button.

With only two values, `none` and `back`, we can clearly distinguish between these two cases. But say we added `share`; how would we distinguish the four cases of "none", "back only", "share only" and "back and share"? I don't believe MQ features can have multiple values (i.e. it's not possible for the UA to set `nav-controls` to {`back`, `share`}, and have the queries `@media (nav-controls: back)` or `@media (nav-controls: share)` both succeed). So, how would we ever add a new queryable control?

_(Please forgive my ignorance if this is actually possible, in which case this issue can be closed.)_

This is touched on by a green box in the spec text, but I don't think the answer is entirely satisfying:

> Note: Theoretically, the two are not strictly equivalent, as there could be new values in a future extension of this media feature other than ['back'](https://drafts.csswg.org/mediaqueries-5/#valdef-media-nav-controls-back) that could match when 'back' doesn’t. In that case, using the ['nav-controls'](https://drafts.csswg.org/mediaqueries-5/#descdef-media-nav-controls) feature in a boolean context could be misleading. However, given that navigation back is arguably the most fundamental navigation operation, the CSS Working Group does not anticipate user interfaces with explicit navigation controls but no back button, so this problem is not expected to occur in practice.

This is basically saying if we add a new control, say, `share`, then:

- `back` would mean "back only"
- `share` would mean "back and share"
- there would be no provision for a "share only" UI, per the above assumption

There are two problems with this:

1. It has a built-in footgun of `@media (nav-controls: back)`. Adding a new control like `share` will immediately break any sites which are (against the advice of the green box) using this query.
2. It can only be extended with one more control. Say we want to add a third control, like `forward`, and assume a back button is always present if there are any controls at all. How will user agents communicate to sites the difference between "back and share", "back and forward" and "back, forward and share"?

Since the intended usage of this is to consider each control in isolation (e.g. hide the in-page back button if the UA provides one, hide the in-page share button if the UA provides one, etc), why not simply provide a separate media feature for each control, with a common prefix?

This implies that instead of `nav-controls`: `none` | `back`, we should have `nav-controls-back`: `<mq-boolean>`. Any new controls would get their own feature, e.g. `nav-controls-share`: `<mq-boolean>`.

Thus the usage changes to
```css
@media (nav-controls-back) {
  #back-button {
    display: none;
  }
}
```

which is easier to read than `@media (nav-controls)` because it's explicit about the fact that we're looking for a back button (noting that `@media (nav-controls: back)` in the current draft is also explicit, but wrong).

mgiuca:
[css-mediaqueries-5] nav-controls: Clarify use in a boolean context (closes #6785)

Non-normative change: Remove the example of using nav-controls in a boolean
context and somewhat confusing note explaining its usage. Replaced with a note
explaining why it is not recommended that sites use nav-controls in a boolean
context.

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -2273,25 +2273,11 @@ Detecting UA-supplied navigation controls: the 'nav-controls' feature</h3>
 			}
 		}</code></pre>
 
-		As this media feature can be used in a [=boolean context=],
-		the same example can be written with shorter syntax:
-
-		<pre><code class=lang-css>
-		@media (nav-controls) {
-			#back-button {
-				display: none;
-			}
-		}</code></pre>
-
-		Note: Theoretically, the two are not strictly equivalent,
-		as there could be new values in a future extension of this media feature
-		other than ''back''
-		that could match when ''back'' doesn't.
-		In that case, using the 'nav-controls' feature in a boolean context could be misleading.
-		However, given that navigation back is arguably the most fundamental navigation operation,
-		the CSS Working Group does not anticipate
-		user interfaces with explicit navigation controls but no back button,
-		so this problem is not expected to occur in practice.
+		Note: This media feature can also be used in a [=boolean context=], e.g.,
+		<code class=lang=css>@media (nav-controls)</code>. However, as there could be new values in a future
+		extension of this media feature other than ''back'', use of the ''nav-controls'' feature in a boolean
+		context could cause the page to incorrectly interpret some other control as a "back" button, and its
+		usage is therefore discouraged.
 	</div>
 
 	Whether [=obviously discoverable=] controls are active does not impact the evaluation of this media feature.


frivoal:
@tabatkins What do you think? I'm sympathetic to this viewpoint, yet I am hesitant to promote more verbose syntax in prevention of a theoretically valid but practically unlikely risk…

mgiuca:
> more verbose syntax

We're talking about the difference between `nav-controls` and `nav-controls: back`. As a site developer, this seems like an easy trade-off to make: 6 more characters to say what you mean, instead of baking a generally-but-not-necessarily-true assumption into your code.

> theoretically valid but practically unlikely

Until we have a second control, this isn't a normative issue. However, let's consider the future. I think there are two directions this could go (normatively) when we do add another control (e.g. `share`):

1. User agents are free to have share controls without a back button. Therefore, from today, sites should never use `nav-controls` in a Boolean context (it is fairly meaningless, since it doesn't imply the existence of any particular control). Instead, they should always write `@media (nav-controls: back)`, and it is misleading to suggest otherwise.
2. `nav-controls` always implies the presence of a back button. Therefore, it is valid to write `@media (nav-controls)` to mean "if there is a back button". In this case, that implicit assumption will need to be made explicit if we ever add a new control; for example, if we add `share`, it will need to be **normatively** specified as "The user agent provides navigation controls, including at least a back button and a share button." (There should be a non-normative note stating that all future values will explicitly specify the inclusion of the back button as well.)

If we're not comfortable committing to having all future controls explicitly including the back button (2), then we should choose 1. I don't think it makes sense to sit on the fence here and say "user agents can have a share button without a back button, but we encourage developers to write code that assumes they won't."

I would prefer 1 because it allows us to give developers the ability to control (via manifest hints) which controls appear. It is perfectly reasonable to want a site that has a share button but not a back button, even if that is unlikely to be the default UA behaviour.

(Ultimately, the non-normative text in the spec doesn't _really_ matter. What matters is what tutorials do and what developers generally write. I can't find any tutorials about this feature at the moment. The reason I'm spending energy debating here is because I suspect when tutorials do come along, they will copy the advice from the spec, and then developers will copy the code from tutorials.)

mgiuca:
This didn't get a lot of traction, possibly because there was no open bug linked to it.

I've filed a new bug: #7288.

PR #6799, opened 2021-11-05
[mediaqueries-4] Make <any-value> optional in <general-enclosed>, remove <ident>

andruud:
This is to make it possible for the following things to match
`<general-enclosed>`:

 - `unknown()`
 - `(unknown)`
 - `(10px < unknown)`


tabatkins:
(We discussed this privately yesterday.) Thanks for the PR, Anders!

PR #6806, opened 2021-11-09
[mediaqueries-4] Lift parens to <media-in-parens>

andruud:
For container queries (css-contain-3), it would be convenient to
be able to reference paren-less `<media-feature>`, since that grammar
wants to wrap the value in a function rather than plain parentheses.

As far as I can tell, `<media-feature>` is only used by `<media-in-parens>`,
so this should be safe.


tabatkins:
Yeah, I like that organization a little better too, keeping all the parens up in the same production.

mirisuzanne (commented):
looks great

Issue #6576, opened 2021-09-03
[css-cascade-5] Cascade Layers need CSSOM support


anttijk:
CSSLayerRule interface needs to be specified.

xiaochengh:
This is not entirely trivial. Since we have two types of `@layer` rules, the empty statement and the block rule, do we want the same interface for both of them, or different interfaces?

I don't have a strong preference. Just putting the options here for discussion.

---

If we use different interfaces, the following seems pretty straightforward:

```idl
[Exposed=Window]
interface CSSLayerStatementRule : CSSRule {
  readonly attribute CSSOMString names;
};

[Exposed=Window]
interface CSSLayerBlockRule : CSSGroupingRule {
  readonly attribute CSSOMString name; // empty string for anonymous layer
};
```

---

If we use the same interface, it would probably be like

```idl
[Exposed=Window]
interface CSSLayerRule : CSSGroupingRule {
  readonly attribute CSSOMString name;
  readonly attribute boolean isEmptyStatement;

  unsigned long insertRule(CSSOMString rule, optional unsigned long index = 0); // Is this the correct syntax for override?
};
```

We need to override CSSGroupingRule.insertRule() so that it throws an exception when called on empty statements.

tabatkins:
I think we should make them separate objects, since they have quite distinct data models - one vs N names, child rules vs not. They're the same name for authoring convenience, but that doesn't need to translate to them being the same object type.

emilio:
I'd prefer not having two interfaces for this, fwiw... Also, for the `@layer a, b, c;` case we need a sequence of names, not just one name, right?

tabatkins:
Yup, that's what I meant by "one vs N names". @xiaochengh's example supports that, it just exposes the N names as a single CSSOMString rather than as a list of them.

anttijk:
Having two separate interfaces might actually make sense as there is nothing that can be shared.

CSSImportRule also needs an extension.

emilio:
Yeah, I guess that's fine / alright, specially given we've given up on making `CSSRule.type` return something meaningful for these.

anttijk:
On the other hand these interfaces are pretty useless anyway so I'm equally fine with a generic object that exposes nothing but cssText and the tree structure.

emilio:
That'd be totally ok for me too.

emilio:
FYI. Not ideal, but I needed _some_ OM support to make DevTools etc work in Firefox.

[...]

PR #6849, opened 2021-12-01
[css-cascade-5] Make CSSImportRule.layerName nullable #6576

xiaochengh:
So that we can distinguish whether an `@import` is unlayered or in an anonymous layer.


xiaochengh:
@mirisuzanne @fantasai @tabatkins Could you review? Thanks!

PR #6874, opened 2021-12-09
[css-syntax-3] Fix serialization matrix related to CDC-token

lahmatiy:
Since double dash `--` is a valid identifier by the spec the `CDC-token` is breaking "round-trip" with parsing now when follows after `#`, `-`, number or `@`:

* `#-->` = `#--` (hash-token) + `>` (delim-token)
* `--->` = `---` (ident-token) + `>` (delim-token)
* `number-->` = `number--` (dimension-token) + `>` (delim-token)
* `@-->` = `@--` (at-keyword-token) + `>` (delim-token)


tabatkins:
Oooh good catch, thanks!

PR #6879, opened 2021-12-11
[css-color-4] fix typo in conversions.js

romainmenke:
[css-color-4] fix typo

Current code behaves like this (which I think is a typo) :

```js
var alpha = 1;
if (alpha = 0) {
 console.log('truthy')
} else {
 console.log(5 / alpha);
}

> Infinity
```


svgeesus:
Typo indeed, thanks!

PR #6969, opened 2022-01-19
[selectors-4] Remove the word "simply"

sheilamoussavi:
Propose removing "simply" as it is not inclusive language.


astearns:
thanks!

PR #6974, opened 2022-01-21
[css-animations-2] Account for `animation-composition` when merging keyframes (closes #6935)

### LINKED ISSUE #6935: [css-animations-2] should `animation-composition` be taken into account when merging keyframes?
The [Keyframes section](https://drafts.csswg.org/css-animations-2/#keyframes) of the [CSS Animations Level 2 spec](https://drafts.csswg.org/css-animations-2/) discusses how `animation-timing-function` applied to individual keyframes should be account for along with the offset to merge keyframes with similar offset and easing.

Should `animation-composition` also be accounted for here such that only keyframes with the same offset, canonical timing function _and_ composite operation be merged?

graouts:
Keyframes should only be merged when they have matching keyframe offset, timing function _and_ composite.

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -259,6 +259,8 @@ and the position of the animation in |element|'s 'animation-name' list,
      repeating the list as necessary as described in
      [[CSS-ANIMATIONS-1#animation-name]].
 
+1.   Let |default composite| be ''replace''.
+
 1.   Find the last ''@keyframes'' at-rule in document order
      with <<keyframes-name>> matching |name|.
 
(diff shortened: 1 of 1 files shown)


graouts:
I have filed https://github.com/web-platform-tests/wpt/pull/32495 with the relevant test changes.

css-animations-2/Overview.bs
```diff
@@ -259,6 +259,8 @@ and the position of the animation in |element|'s 'animation-name' list,
      repeating the list as necessary as described in
      [[CSS-ANIMATIONS-1#animation-name]].
 
+1.   Let |default composite| be ''replace''.
```
birtles, line 262: Should this default to the resolved value of 'animation-composite' for element in the same way we do for the timing function?

birtles: It looks like we we have a test for that behaviour: https://github.com/web-platform-tests/wpt/blob/b5ddd150f64c457a28411545fa6c16685a33bf70/css/css-animations/KeyframeEffect-getKeyframes.tentative.html#L314-L331

graouts: I thought that `animation-composite` should set the effect-wide `composite` operation rather than set a default for keyframes. I think the current behavior for `animation-timing-function` is a mistake since it doesn't let you set an effect-wide easing, and I'd rather we didn't repeat the same mistake.

birtles: Oh, right you are. I completely forgot we had an effect-wide composite operation in the model.

I agree `animation-timing-function`'s behaviour is undesirable. We should find a way to fix that some time (perhaps introducing `animation-effect-timing-function`?).

I suppose even if we were to introduce `animation-effect-timing-function` the way the effect easing applies is different to the way the default composite operation applies so we don't need to worry about making this consistent with that.

birtles (approved):

css-animations-2/Overview.bs
graouts: Yes, `animation-timing-function` has always bugged me. I filed https://github.com/w3c/csswg-drafts/issues/6982 so that we may discuss how to solve this mess.

PR #6975, opened 2022-01-21
[css-animation-2] Setting `animation.effect.composite` should override any further change to `animation-composition`

graouts:
Much like setting other animation or animation effect property using the Web Animations API lead to further changes
to the matching CSS property to not be reflected on the animation or animation effect, changes to `animation-composition`
should no longer be reflected on the animation effect if `animation.effect.composite` is set.

Testing for this has already been landed as part of https://github.com/web-platform-tests/wpt/commit/52c5cb32bb28c90dde71de8f7bbd71ba7eec054f.

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -144,6 +144,11 @@ the changes from the programming interface take precedence as follows:
     However, if the last matching ''@keyframes'' rule is removed
     the animation must still be canceled.
 
+*   After successfully setting the {{KeyframeEffect/composite}} of the
+    {{Animation/effect}} of a {{CSSAnimation}} to some {{CompositeOperation}}
+    other than the original {{CompositeOperation}}, any subsequent change to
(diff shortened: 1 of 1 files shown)


css-animations-2/Overview.bs
```diff
@@ -142,6 +144,11 @@ the changes from the programming interface take precedence as follows:
     However, if the last matching ''@keyframes'' rule is removed
     the animation must still be canceled.
 
+*   After successfully setting the {{KeyframeEffect/composite}} of the
+    {{Animation/effect}} of a {{CSSAnimation}} to some {{CompositeOperation}}
+    other than the original {{CompositeOperation}}, any subsequent change to
```
flackr, line 149: With most of these overrides, setting any value - even the original - means that it is now set from javascript. I think setting the effect is a little different because the original KeyframeEffect by definition still links to the original effect.

TLDR i think even setting the original composite value should track that it has been overridden.

PR #6983, opened 2022-01-25
[css-text-3] The web de-facto requires NULL U+0000 to not be visible

litherum:
Chrome and Firefox make it invisible, and WebKit has compat issues about it.

https://bugs.webkit.org/show_bug.cgi?id=235559

### DIFF
--- a/css-text-3/Overview.bs
+++ b/css-text-3/Overview.bs
@@ -1579,7 +1579,8 @@ White Space Processing &amp; Control Characters</h2>
 	[[CSS2]]
 
 	Control characters ([=Unicode category=] <code>Cc</code>)--
-	other than tabs (U+0009),
+	other than NULL (U+0000),
+	tabs (U+0009),
 	line feeds (U+000A),
 	carriage returns (U+000D)
 	and sequences that form a [=segment break=]--


kojiishi:
Blink renders all characters in the DOM since [crbug.com/530342](https://crbug.com/530342). This change even included default ignorable code points that we have [crbug.com/550275](https://crbug.com/550275).

Blink doesn't do anything special to U+0000 either, but also doesn't do anything special to make sure it's visible when calling the underlying HarfBuzz or platform APIs, so it's possible that the behavior is platform and/or API dependent; i.e., if the platform API requires C-style string, it may disappear.

From a quick test, DOM and canvas `measureText` render U+0000 on Win10, but `don't render on mac.
https://output.jsbin.com/nuqate

jfkthame:
Firefox has an `about:config` setting `layout.css.control-characters.visible` that controls whether control characters are rendered visibly. Currently, this is enabled in Nightly builds, but disabled for the Beta and Release channels.

When `layout.css.control-characters.visible` is set to `true`, U+0000 is rendered as a visible "hexbox" glyph just like other control characters. I'm not aware of any reports of this being noticed in the wild (as has occasionally happened for other control codes), though perhaps shipping it more widely will unearth a few cases.

jfkthame:
> Blink renders all characters in the DOM since [crbug.com/530342](https://crbug.com/530342).

Note that (if I'm understanding correctly) Blink simply passes the control characters through the same rendering pipeline as printable characters, so the result may be visible or not depending on the particular font being used. Some fonts map control-character codepoints to invisible glyphs, and in this case they'll remain invisible in Blink, while other fonts may render a `.notdef` box or even some other graphical representation.

This is not consistent with the [CSS Text requirement](https://drafts.csswg.org/css-text-3/#white-space-processing), which says:

> Control characters ([Unicode category](https://drafts.csswg.org/css-text-3/#unicode-general-category) Cc)—other than tabs (U+0009), line feeds (U+000A), carriage returns (U+000D) and sequences that form a [segment break](https://drafts.csswg.org/css-text-3/#segment-break)—must be rendered as a visible glyph which the UA must synthesize if the glyphs found in the font are not visible

jfkthame:
@litherum Can you clarify what the compat concern here is about? I've tried signing in to iCloud Notes using Firefox Nightly, where the `layout.css.control-characters.visible` setting is `true` (and so any occurrences of U+0000 would appear as hexbox glyphs), but I didn't notice anything untoward. Is there a specific scenario to look at?

litherum:
This came up because the Notes app on iCloud.com assumed that .notdef is invisible and has zero width. I’m assuming this one website is not the only one.

jfkthame:
> This came up because the Notes app on iCloud.com assumes that .notdef is invisible and has zero width. I’m assuming this one website is not the only one.

But where/why does the Notes app try to render `.notdef`? Isn't that just a bug in the iCloud Notes app?

kojiishi:
@jfkthame 
> Note that (if I'm understanding correctly) Blink simply passes the control characters through the same rendering pipeline as printable characters, so the result may be visible or not depending on the particular font being used. Some fonts map control-character codepoints to invisible glyphs, and in this case they'll remain invisible in Blink, while other fonts may render a `.notdef` box or even some other graphical representation.

Correct. Blink has not implemented that part yet, so mac users do not complain about visible L-SEP ([crbug.com/550275](https://crbug.com/550275)) until we implement it.

@litherum 
> This came up because the Notes app on iCloud.com assumes that `.notdef` is invisible and has zero width. I’m assuming this one website is not the only one.

While Blink does not check U+0000, it does check glyph 0, so maybe that's what you observed in icloud.com? The glyph id 0 is `.notdef` as long as the font follows the [Recommendations for OpenType Fonts](https://docs.microsoft.com/en-us/typography/opentype/spec/recom#glyph-0-the-notdef-glyph). Blink has some places that assume all fonts follow this recommendation, such as [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper.cc;l=477?q=%20HarfBuzzShaper::ExtractShapeResults&ss=chromium%2Fchromium%2Fsrc).

fantasai (changes requested):
r- until this is resolved by the WG (and also any such change needs to go into css-text-4 as well). Opened https://github.com/w3c/csswg-drafts/issues/7249 to track it as an issue.

fantasai:
CSSWG rejected this change, see minutes: https://lists.w3.org/Archives/Public/www-style/2022May/0009.html
Closing out.

Issue #2532, opened 2018-04-11
[css-borders-4] Issues with listifying border-color


LeaVerou:
This is related to #1172.

In the [2017 Tokyo F2F we resolved](https://lists.w3.org/Archives/Public/www-style/2017May/0049.html) on listifying `border-color` with the design allowing for future listification of `border-width` and `border-style`.
We decided on `border-color` being the property that sets the number of borders, since a common case is having multiple borders with the same width and we didn't want authors to have to repeat the width of each stripe.

However, this has very unexpected and confusing repercussions. It means that setting `border-color` can increase the total border width. In the following:

```css
border-width: 10px;
border-color: red, black, white;
```

the resulting border width is actually 30px! (since `border-width` is expanded, per usual listification rules) This has never arisen before because none of our other listified properties stack this way, they are all independent layers.

Furthermore, another issue is that listification typically repeats the same syntax that a property already has. Since `border-color` accepts 1-4 colors for each side, this results in very weird listified syntax that inverts the relative power of space and comma to the exact opposite of how it would be expected to work. Look at this code and visual result:

```css
border-color: skyblue orange yellowgreen indianred, black yellow
```

![image](https://user-images.githubusercontent.com/175836/38607092-1198f110-3d78-11e8-9982-ad39e0faf6d4.png)

This is a completely nonsensical way to specify border colors. People think of all colors for a side as a group, they don't group colors by whether they are innermost or outermost.

For these reasons, @fantasai and I are exploring alternative syntaxes that cover most use cases with a more intuitive syntax.

The primary use cases we must definitely cover are:

- Multiple border stripes of equal width should be effortless, with no width duplication.
- Different width stripes with fixed ratios between them (e.g. two 1fr stripes with one 2fr color between them)
- A mix of fixed and flexible width stripes (e.g. two 1px black borders with blue between them that occupies the entire rest of the border). `fr` units deal very nicely with this.
- We don't think different styles are particularly useful here. Almost all use cases we've seen are about solid borders. Therefore, we're ok with having multiple colors for different styles just be defined as a cutout of the solid case.
- Are different colors per side useful?

## Idea 1: Making `border-color` accept a list of color stops

While this may sound like a convenient re-use, it has several issues. Unlike gradients, in borders stripes are more often desired (are gradients even that useful here?) so we'd need to change how the syntax behaves anyway. Re-using a concept from somewhere else (like gradients) but subtly changing it is confusing.
[...]

PR #7029, opened 2022-02-09
[css-images-4][css-backgrounds-4] Added stripes() function

SebastianZ:
This change is meant to convert the [resolution of #2532](https://github.com/w3c/csswg-drafts/issues/2532#issuecomment-402327492) into spec. text.

It adds the `stripes()` function as a generally usable function as discussed in #2532 to CSS Images 4.

Note that I changed the syntax slightly to incorporate the `fr` unit mentioned in @LeaVerou's description. I did that by adding the `<flex>` value defined in CSS Grid 1, though the definition there currently focuses on the usage within grid layout. So its description either might need to be changed/generalized or the unit needs to be defined again within CSS Images 4. I'd prefer to generalize it (and maybe even move it to CSS Values 4).

The `stripes()` function is then used as a replacement for the colors list defined for `border-*-color` in CSS Backgrounds 4.

Reading through the discussions, a few things were still unclear to me or were discussed but not resolved on.

1. What's the behavior when the width of the image the `stripes()` function is used in is smaller or bigger? I've defined that the rest is transparent when the stripes thickness is smaller and that the stripes are clipped when they are bigger than the image, as that seems to be what people were in favor of in the discussion.
2. Should the stripes be painted inwards or outwards? For now I kept the definition of "starting outwards", though to me it seems a bit counter-intuitive that the borders go inwards. Also, the wording may be clarified further.
3. Should there be a description of how stripes (or generally `border-color`) interact with `border-style`? Clipping was mentioned in the discussion, also treating `inset`, `outset`, `groove`, and `ridge` as solid.

Also note that this change is based on #7014, because without those changes Bikeshed won't compile the spec.

Sebastian


svgeesus:
@LeaVerou @fantasai we resolved to add `stripes()` ages ago and the edit was not made, this PR looks like a good way forward?

LeaVerou:
Hi Sebastian,

Thank you for working on this.
It has been a long time, but IIRC the resolution of these discussions, we would define 1 dimensional images, and `stripes()` would just produce a 1 dimensional image. Right now in CSS we have 2d images (gradients, urls etc), and 0 dimensional images (`image(<color>)`). I see that in this PR, `stripes()` is defined as its own type, and can only be used in properties that explicitly allow it. We'd like to be a bit more general than that.

SebastianZ:
Thank you for reviewing the PR! There are obviously some open questions, so let's [continue the discussion in the issue](https://github.com/w3c/csswg-drafts/issues/2532#issuecomment-1040238169) before we continue with this PR.

Sebastian

SebastianZ:
@LeaVerou @tabatkins I went ahead and interpreted your statements in the issue to make the relevant changes. That means, I added `<stripes()>` to `<image>` and changed its description.

I also added it to `outline-color` as previously resolved on.  I've kept the reference to `<flex>` for now. Though as pointed out earlier, I think its definition should be generalized and moved to a more general place to make it clear that it doesn't only apply to grid layout anymore.

I am awaiting your feedback on these changes.

Sebastian

SebastianZ:
As noted by @LeaVerou, I've now put `stripes()` behind a `<1d-image>` type, so it can easily be reused and extended.

Open questions:

* Is it ok to allow different orders in the syntaxes?
* What to do with `<flex>`?

Sebastian

SebastianZ:
>     * Is it ok to allow different orders in the syntaxes?

I used the [double pipe](https://drafts.csswg.org/css-values-4/#comb-any) in the syntaxes, i.e. `<color> || <1d-image>`. This allows the two values to be in any order.

That allows to specify something like `border-color: red stripes(white, silver) stripes(white, silver) yellow lime stripes(black, gray) blue;`.

We might use juxtaposing instead, i.e. `<color> <1d-image>` to force the order of first `<color>` then `<1d-image>`.

>     * What to do with `<flex>`?

As noted in a [previous comment](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1068988402), I reused `<flex>` from CSS Grid 1 because I agree with you that it fits very well this use case.

Though, as it is now used in two completely different places, should it's description be generalized and be moved somewhere else? And if so, should that be done in this PR already?

Sebastian

LeaVerou:
> > ```
> > * Is it ok to allow different orders in the syntaxes?
> > ```
> 
> I used the [double pipe](https://drafts.csswg.org/css-values-4/#comb-any) in the syntaxes, i.e. `<color> || <1d-image>`. This allows the two values to be in any order.
> 
> That allows to specify something like `border-color: red stripes(white, silver) stripes(white, silver) yellow lime stripes(black, gray) blue;`.
> 
> We might use juxtaposing instead, i.e. `<color> <1d-image>` to force the order of first `<color>` then `<1d-image>`.

Oh no, you need to use a single pipe: `[ <<color>> | <<1d-image>> ]`. It's a `<color>` OR `<1d-image>` but not both. It doesn't make sense to have `border-color: red stripes(white, silver) stripes(white, silver) yellow lime stripes(black, gray) blue;` because you can't tell what is for what side.

> > ```
> > * What to do with `<flex>`?
> > ```
> 
> As noted in a [previous comment](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1068988402), I reused `<flex>` from CSS Grid 1 because I agree with you that it fits very well this use case.
> 
> Though, as it is now used in two completely different places, should it's description be generalized and be moved somewhere else? And if so, should that be done in this PR already?

Yeah, it should go to css-values. @fantasai @tabatkins ?

tabatkins:
Yeah we can move `<flex>` out to the general V&U spec now that it's being used in more than just Flexbox.

LeaVerou:
> Yeah we can move `<flex>` out to the general V&U spec now that it's being used in more than just Flexbox. (I'll wait for merging of this PR tho.)

Great, should we call it something else in that case? `<fraction>`?

css-backgrounds-4/Overview.bs
```diff
@@ -185,26 +185,32 @@ Painting Area: the 'background-clip' property</h3>
...
+
+	The stripes defined by <<1d-image>> follow the direction of
+	the side the border is applied on (i.e. horizontally on left and
+	right borders and vertically on top and bottom borders), starting outwards.
+	With that, the corresponding 'border-*-width' property defines the total
+	width of the stripes.
+
+	The stripes are painted above a given <<color>>. This allows the <<color>>
+	value to serve as a fallback, so any transparent stripes let the color
+	shine through.
```
fantasai, line 213: What happens if no color is specified? Do we get transparent or currentColor or...?

css-backgrounds-4/Overview.bs
```diff
@@ -223,7 +229,10 @@ Painting Area: the 'background-clip' property</h3>
 		<pre class=lang-css>
 		.foo {
 			border: 30px solid;
-			border-color: skyblue orange yellowgreen indianred, black yellow;
+			border-top-color: stripes(skyblue, black);
+			border-right-color: stripes(orange, yellow);
+			border-bottom-color: stripes(yellowgreen, black);
+			border-left-color: stripes(indianred, yellow);
```
fantasai, line 235: I believe this example is trying to show off the use of the 'border-color' shorthand, this rewrite into longhands isn't appropriate.

css-backgrounds-4/Overview.bs
```diff
@@ -185,26 +185,32 @@ Painting Area: the 'background-clip' property</h3>
...
+	right borders and vertically on top and bottom borders), starting outwards.
+	With that, the corresponding 'border-*-width' property defines the total
+	width of the stripes.
+
+	The stripes are painted above a given <<color>>. This allows the <<color>>
+	value to serve as a fallback, so any transparent stripes let the color
+	shine through.
 
 	When interpolating between borders with the same number of colors,
 	interpolation is performed individually per color band as <a href="https://www.w3.org/TR/css3-transitions/#animtype-color">color</a>.
```
fantasai, line 216: Interpolation should probably be defined in the definition of the `stripes()` function, rather than here. And it should match what we do for gradients.

css-images-4/Overview.bs
```diff
@@ -72,7 +72,7 @@ Value Definitions {#values}
 	or <a section href="#gradients">gradient notation</a>.
 	Its syntax is:
 
-	<pre class="prod"><dfn>&lt;image></dfn> = <<url>> | <<image()>> | <<image-set()>> | <<cross-fade()>> | <<element()>> | <<gradient>></pre>
+	<pre class="prod"><dfn>&lt;image></dfn> = <<url>> | <<image()>> | <<image-set()>> | <<cross-fade()>> | <<element()>> | <<gradient>> | <<1d-image>></pre>
```
fantasai, line 75: I don't think we have a WG resolution on using 1D images as 2D images, so I wouldn't include that here.

(I also don't think it's a useful idea. If we want stripes() to work in 2D, we should define a 2D stripes function, which can re-orient the stripes as appropriate. It's not worth the implementation/testing/documentation effort to have a stripes() function available in all 2D contexts that can only do horizontal stripes. IMHO)

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,49 @@ Color Stop “Fixup”</h4>
...
+-->
 
 1D Image Values: the ''stripes()'' notation {#stripes}
 ======================================================
 
-	Issue: <a href="https://github.com/w3c/csswg-drafts/issues/2532">Per WG resolution</a>,
-	define the ''stripes()'' function
-	which creates a 1D image for use in borders/outlines.
+	The ''stripes()'' function allows to define a number of stripes of
+	different colors with different widths.
```
fantasai, line 1911:
```suggestion
	The ''stripes()'' function allows defining a one-dimensional image
	consisting of stripes of different colors with different widths.
```

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,49 @@ Color Stop “Fixup”</h4>
...
+
+	The syntax for ''1d-image'' and ''stripes()'' is defined as:
```
fantasai, line 1913:
```suggestion
	The syntax for <<1d-image>> and <<stripes()>> is defined as:
```

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,49 @@ Color Stop “Fixup”</h4>
...
+
+	The function represents a 1-dimensional image generated by a list of colors
+	and an optional thickness for each of them.
+
+	A <<percentage>> or <<flex>> value refers to the total length of the image
+	in the used context.
+	The <<percentage>> must be between ''0%'' and ''100%'' inclusive;
+	any other value is invalid.
+
+	If the thickness of a stripe is omitted, it is interpreted as ''1fr''.
```
fantasai, line 1928: I'd suggest using the DL format we use in property value definitions. Use dfn-for="stripes()".

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,49 @@ Color Stop “Fixup”</h4>
...
+	<pre class=prod>
+		<dfn>&lt;1d-image></dfn> = <<stripes()>>
+		<dfn>stripes()</dfn> = stripes( [ <<color>> && [ <<length-percentage>> | <<flex>>]? ]# )
+	</pre>
+
+	The function represents a 1-dimensional image generated by a list of colors
+	and an optional thickness for each of them.
+
+	A <<percentage>> or <<flex>> value refers to the total length of the image
+	in the used context.
```
fantasai, line 1924: There's no usefulness in fr values being equivalent to percentages. What you want is to treat them similar to grid, i.e. subtract out all the fixed-size stripes and distribute the remaining space accordingly. If the remaining space is not defined, treat it as zero.

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,49 @@ Color Stop “Fixup”</h4>
...
+	If not defined differently by the context the function is used in, the
+	stripes are painted from top to bottom and tiled horizontally. If the size
+	of the painting area exceeds the total thickness of the stripes, the
+	remaining area is painted transparently. I.e. it behaves as if there was an
+	additional last ''transparent'' stripe. If the size of the painting area
+	is smaller than the total thickness of the stripes, the stripes are clipped
+	starting from the last defined color.
+
+	The computed value of this function is the [=specified value=] with all
+	<<color>>s computed '1fr' values removed.
```
fantasai, line 1939: No, we don't want to remove the fr values. :) We just want lengths and colors computed. So the computed value of this function is a list, each item pairing a computed color and either a `<<flex>>` value or a computed `<<length-percentage>>`.

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,49 @@ Color Stop “Fixup”</h4>
...
+██    ██    ██    ██    ██   ██  ██        █       ██    ██  ██     ██
+ ██████     ██    ██     ██ ████ ██        ███████  ██████    ███ ███
+-->
 
 1D Image Values: the ''stripes()'' notation {#stripes}
 ======================================================
 
-	Issue: <a href="https://github.com/w3c/csswg-drafts/issues/2532">Per WG resolution</a>,
-	define the ''stripes()'' function
-	which creates a 1D image for use in borders/outlines.
```
fantasai, line 1908: Start this section by defining what a one-dimension image is. Then mention that the only one-dimensional image type defined in this level is `stripes()`, and go on to define it below.

css-ui-4/Overview.bs
```diff
@@ -287,13 +287,14 @@ Animation type: by computed value
...
 Name: outline-color
-Value: <<color>> | invert
+Value: [ <<color>> || <<1d-image>> ] | invert
 Initial: invert
 Applies to: all elements
 Inherited: no
 Percentages: N/A
 Computed value: The computed value for ''outline-color/invert'' is ''outline-color/invert''.
  For <<color>> values, see [[!CSS-COLOR-4#resolving-color-values]] in [[!CSS-COLOR-4]].
+ <<1d-image>> values are computed as defined in their definition.
```
fantasai, line 297: I suggest using the same sentence construction as the `<<color>>` type, for consistency.

css-backgrounds-4/Overview.bs
```diff
@@ -185,26 +185,32 @@ Painting Area: the 'background-clip' property</h3>
...
 	These properties set the foreground color of the border specified
 	by the 'border-style' properties.
-	If a list of values is provided, the border is split into equal width bands of each color
-	along the direction of the side the border is applied on
-	(i.e. split horizontally on left and right borders and vertically on top and bottom borders),
-	starting outwards.
+
+	The stripes defined by <<1d-image>> follow the direction of
+	the side the border is applied on (i.e. horizontally on left and
+	right borders and vertically on top and bottom borders), starting outwards.
```
fantasai, line 207: Suggested rewrite:

```suggestion
	The stripes defined by <<1d-image>> follow the shape of the border
	on the side to which they apply,
	and are drawn in bands starting from the [=padding edge=] and progressing outwards.
```

css-backgrounds-4/Overview.bs
```diff
@@ -185,26 +185,32 @@ Painting Area: the 'background-clip' property</h3>
...
+	With that, the corresponding 'border-*-width' property defines the total
+	width of the stripes.
+
+	The stripes are painted above a given <<color>>. This allows the <<color>>
```
fantasai, line 211: I'd use `the specified <<color>>` rather than `a given <<color>>`.

fantasai (dismissed):
Overall looks pretty good. Posted some specific things to fix.

Also, the line-breaking convention for our specs is https://rhodesmill.org/brandon/2012/one-sentence-per-line/ (This is why the lines look so ragged, there's a method to the madness.)

(You might also find https://lists.w3.org/Archives/Public/spec-prod/2018OctDec/0011.html a useful read as background context, though nothing else there is particularly relevant to this PR.)

tabatkins:
I've done what turned out to be a decent bit of rearranging and rewording on the Images part. Largely editorial; the only normative changes are to handle sums of flexes < 1 the same way flexbox does, and define the painting area when used as a 2d image.

SebastianZ:
> > > ```
> > > * Is it ok to allow different orders in the syntaxes?
> > > ```
> > 
> > 
> > I used the [double pipe](https://drafts.csswg.org/css-values-4/#comb-any) in the syntaxes, i.e. `<color> || <1d-image>`. This allows the two values to be in any order.
> > That allows to specify something like `border-color: red stripes(white, silver) stripes(white, silver) yellow lime stripes(black, gray) blue;`.
> > We might use juxtaposing instead, i.e. `<color> <1d-image>` to force the order of first `<color>` then `<1d-image>`.
> 
> Oh no, you need to use a single pipe: `[ <<color>> | <<1d-image>> ]`. It's a `<color>` OR `<1d-image>` but not both. It doesn't make sense to have `border-color: red stripes(white, silver) stripes(white, silver) yellow lime stripes(black, gray) blue;` because you can't tell what is for what side.

I deliberately chose to allow both, so `<color>` serves as a fallback, as you noticed in an inline comment. `border-color` is one place where slashes would make it easier to distinguish between the different values like we have it for `grid-area`. So we could also express the syntax as:

```
<<color>>{1,4} | <<color>> <<1d-image>>? [ / <<color>> <<1d-image>>? ]{0,3}
```

That means, authors have to add the slashes when using `<1d-image>`, while the existing `<color>`-only syntax without slashes is still supported as legacy syntax. Applied to my example above, this would then look like

```
border-color: red stripes(white, silver) / yellow stripes(white, silver) / lime stripes(black, gray) / blue;
```

Sebastian

SebastianZ:
> > Yeah we can move `<flex>` out to the general V&U spec now that it's being used in more than just Flexbox. (I'll wait for merging of this PR tho.)
> 
> Great, should we call it something else in that case? `<fraction>`?

As even Tab obviously mixed it up being part of Flexbox and not Grid, I'd say yes. 😉 Though I don't have a strong opinion on that.

And I agree with him that this should be done in a separate PR.

Sebastian

SebastianZ:
> Overall looks pretty good. Posted some specific things to fix.
> 
> Also, the line-breaking convention for our specs is https://rhodesmill.org/brandon/2012/one-sentence-per-line/ (This is why the lines look so ragged, there's a method to the madness.)
> 
> (You might also find https://lists.w3.org/Archives/Public/spec-prod/2018OctDec/0011.html a useful read as background context, though nothing else there is particularly relevant to this PR.)

Thanks for the links! I'll correct the line-breaking.

Sebastian

css-backgrounds-4/Overview.bs
SebastianZ: Good point! I'd say `currentcolor` as it's the initial value. I'll add that for now, though we may still discuss whether that's the best choice.

Sebastian

css-backgrounds-4/Overview.bs
SebastianZ: Ok, I'll change it back to `border-color`.

Sebastian

css-images-4/Overview.bs
SebastianZ: As pointed out in https://github.com/w3c/csswg-drafts/issues/2532#issuecomment-1104836974, the plan is to add 1D images to gradients. So we might end up using that instead of adding it to `<image>` directly.
Let's discuss that in the next call.

Sebastian

css-images-4/Overview.bs
SebastianZ: By saying that `1fr` is removed I wanted to accomodate the serialization rules which say to omit component values for a shorter serialization when they don't change the meaning. And I defined that an omitted thickness is equivalent to `1fr`.
Though I guess it's not necessary to explicitly mention that as that's already part of the serialization rules.

I'll adjust the wording according to your suggestion.

css-images-4/Overview.bs
SebastianZ: Agree. I actually already had something prepared for that before that comment.

Sebastian

SebastianZ:
> I've done what turned out to be a decent bit of rearranging and rewording on the Images part. Largely editorial; the only normative changes are to handle sums of flexes < 1 the same way flexbox does, and define the painting area when used as a 2d image.

I missed that when answering the previous comments. @tabatkins Thank you for the changes! With those you obviously already covered a few things noted by @fantasai.

I'll incorporate the rest of the changes and we can discuss the points that are still open afterwards.

Sebastian

LeaVerou:
> > > > ```
> > > > * Is it ok to allow different orders in the syntaxes?
> > > > ```
> > > 
> > > 
> > > I used the [double pipe](https://drafts.csswg.org/css-values-4/#comb-any) in the syntaxes, i.e. `<color> || <1d-image>`. This allows the two values to be in any order.
> > > That allows to specify something like `border-color: red stripes(white, silver) stripes(white, silver) yellow lime stripes(black, gray) blue;`.
> > > We might use juxtaposing instead, i.e. `<color> <1d-image>` to force the order of first `<color>` then `<1d-image>`.
> > 
> > 
> > Oh no, you need to use a single pipe: `[ <<color>> | <<1d-image>> ]`. It's a `<color>` OR `<1d-image>` but not both. It doesn't make sense to have `border-color: red stripes(white, silver) stripes(white, silver) yellow lime stripes(black, gray) blue;` because you can't tell what is for what side.
> 
> I deliberately chose to allow both, so `<color>` serves as a fallback, as you noticed in an inline comment. `border-color` is one place where slashes would make it easier to distinguish between the different values like we have it for `grid-area`. So we could also express the syntax as:
> 
> ```
> <<color>>{1,4} | [ <<color>>? <<1d-image>>? ]! [ / [ <<color>>? <<1d-image>>? ]! ]{0,3}
> ```
> 
> That means, authors have to add the slashes when using `<1d-image>`, while the existing `<color>`-only syntax without slashes is still supported as legacy syntax. Applied to my example above, this would then look like
> 
> ```
> border-color: red stripes(white, silver) / yellow stripes(white, silver) / lime stripes(black, gray) / blue;
> ```
> 
> Sebastian

What kind of fallback? If `stripes()` is not supported, the entire declaration would be dropped, including the color.
If we want `stripes()` to accept an optional color to paint the stripes over, it should probably be part of the syntax for `stripes()`, not in each individual property's grammar as separate tokens. @fantasai any thoughts on this?

css-images-4/Overview.bs
tabatkins: Computed value is not serialization! The two concepts are completely unrelated, tho they were often conflated in older spec text. In this case, every stripe has a color and a thickness; the fact that a `1fr` thickness is omitted in the serialization falls out automatically from the shortest serialization principle.

tabatkins:
I don't know what the "color to paint the stripes over" is for, tbh.

LeaVerou:
> I don't know what the "color to paint the stripes over" is for, tbh.

No idea, but I assume that's what @SebastianZ intended for the fallback, a color to paint underneath any semi-transparent stripes. FWIW I don't think it's that necessary and complicates the syntax, so I'd be fine removing it.

SebastianZ:
> > I don't know what the "color to paint the stripes over" is for, tbh.
> 
> No idea, but I assume that's what @SebastianZ intended for the fallback, a color to paint underneath any semi-transparent stripes. FWIW I don't think it's that necessary and complicates the syntax, so I'd be fine removing it.

Yes, the meaning of this "fallback color" was to let the color is drawn below any (semi-)transparent stripes. So, basically the same interaction as `background-color` with `background-image`. The idea behind that was that authors might expect that the `<color>` value still has an influence even when using `stripes()`.

Though I'm happy to change the syntaxes to `<color> | `<1d-image>` if we agree that this won't be the case and there are only use cases to have either `<color>` or `stripes()`.

Sebastian

tabatkins:
Oh, I guess my comment I wrote yesterday on the `color || stripes` bit didn't actually end up getting posted - we literally can't do `||` anyway, because it makes `border-color` ambiguous if it has anywhere between 2 and 7 entries. `border-color: red stripes(green, blue)` might be one value (cloned to all four sides) or two values (red top and bottom, stripes left and right). Given that "just color" is allowed today, the only way to do it at all is with `[ <1d-image>? <color> ]+` - that is, a *required* color, with an optional 1d-image that *must* precede it.

But yeah, I don't understand the purpose of having both color and a 1d-image anyway. I understand the analogy you're drawing with background-color vs -image, but there's no use-case for this functionality in borders currently. More generally, we don't have the concept of layered borders like we have layered backgrounds.

SebastianZ:
> Oh, I guess my comment I wrote yesterday on the `color || stripes` bit didn't actually end up getting posted - we literally can't do `||` anyway, because it makes `border-color` ambiguous if it has anywhere between 2 and 7 entries. `border-color: red stripes(green, blue)` might be one value (cloned to all four sides) or two values (red top and bottom, stripes left and right). Given that "just color" is allowed today, the only way to do it at all is with `[ <1d-image>? <color> ]+` - that is, a _required_ color, with an optional 1d-image that _must_ precede it.

You're right. I totally missed that ambiguity. Though I incidentally [brought up a solution for that earlier](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1104969215) by requiring slashes between the values when `<1d-image>` is used.

> But yeah, I don't understand the purpose of having both color and a 1d-image anyway. I understand the analogy you're drawing with background-color vs -image, but there's no use-case for this functionality in borders currently. More generally, we don't have the concept of layered borders like we have layered backgrounds.

With the above said, I agree now to change it to `<color> | <1d-image>`. When use cases for layered borders arise, we can think about that again.

Sebastian

css-backgrounds-4/Overview.bs
SebastianZ: Because we've now defined to have either a color or a 1D image that has transparent black as its default color, there's no need to define this anymore.

Sebastian

css-backgrounds-4/Overview.bs

css-backgrounds-4/Overview.bs
SebastianZ: I've moved it to the "Interpolation" section in CSS Images 4 and tried to align its definition with the one for gradients. Please have a detailed look at it!

Sebastian

css-images-4/Overview.bs
SebastianZ: As discussed, we've split the decision to use 1D images in 2D contexts out of this PR. For reference, see #7241 and #7244.

Sebastian

css-images-4/Overview.bs
SebastianZ: This is covered by Tab's changes.

Sebastian

css-images-4/Overview.bs
SebastianZ: I guess that's covered by Tab's changes already.

Sebastian

css-ui-4/Overview.bs
SebastianZ: I've done that. @fantasai Please have a look whether that's what you had in mind!

Sebastian

css-backgrounds-4/Overview.bs

css-backgrounds-4/Overview.bs
SebastianZ: This is obsolete because we've changed the syntax to use either a color or a 1D image defining transparent black and its default color.

Sebastian

SebastianZ:
I've removed `<1d-image>` from `<image>` as resolved and opened up separate issues for adding 1D images to 2D contexts. And I think I've addressed all remaining comments. So please give this PR another review and let me know if there's anything left!

Sebastian

css-images-4/Overview.bs
SebastianZ: That's already done by the changes made by @tabatkins.

Sebastian

SebastianZ:
@fantasai @LeaVerou @tabatkins Is it possible that you give this another review? I'd really like to finish this.

Sebastian

SebastianZ:
@bert-github I've added you as a reviewer, because you are an editor of CSS Backgrounds 4 and the others seem to be busy. It would be perfect if I got some feedback whether this patch can be merged now or if anything is still missing.

Sebastian

css-images-4/Overview.bs
SebastianZ: I don't think this is still needed but please let me know if it is!

css-ui-4/Overview.bs
SebastianZ: I've fixed the two links now. The one to Color 4 was actually broken.

css-backgrounds-4/Overview.bs
```diff
@@ -185,30 +185,28 @@ Painting Area: the 'background-clip' property</h3>
...
 	Percentages: n/a
-	Computed Value: the computed color
+	Computed Value: the computed color and/or a one-dimensional image function
 	Animation type: see prose
 	</pre>
 
 	<pre class="propdef shorthand">
 	Name: border-color
-	Value: <<color>>#{1,4}
+	Value: [ <<color>> | <<1d-image>> ]{1,4}
```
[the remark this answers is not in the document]
SebastianZ: For reference, here's the suggested syntax again:

```
<<color>>{1,4} | [ <<color>>? <<1d-image>>? ]! [ / [ <<color>>? <<1d-image>>? ]! ]{0,3}
```

And here's an example using that syntax:

```css
border-color: stripes(white, silver) / stripes(white, silver) / stripes(black, gray) / blue;
```

SebastianZ:
As the comment list on this PR got quite long, let me summarize the pending points:

* I've moved the [definition for interpolating `<1d-image>` to Images 4](https://github.com/w3c/csswg-drafts/pull/7029#discussion_r864476143).
* I believe [fantasai's suggestion to add `dfn-for="stripes()"`](https://github.com/w3c/csswg-drafts/pull/7029#discussion_r854572936) is not needed anymore after all the changes.
* I've [previously made a suggestion](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1104969215) to add slashes between the different values in the `border-color` shorthand to align it with other shorthands and make it future prove which I'd like to get feedback on.
* We agreed on moving `<flex>` to Values & Units (and possibly rename it to `<fraction>`) once this PR is merged.

@fantasai @tabatkins Could you please give this another review so the changes can be merged?

Sebastian

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,96 @@ Color Stop “Fixup”</h4>
...
+		the <<flex>> values would dictate.
+		
+		(This is similar to how [=flex layout=] deals with small <<flex>> sums on a line,
+		and ensures smoothly continuous behavior
+		as the <<flex>> values approach zero.)
+	</div>
+
+	The |painting area| is defined by the context in which the ''stripes()'' function is used.
+	If used generically as an <<image>>,
+	the |painting area| is the height of the [=concrete image size=].
```
Loirooriol, line 1976: What does "If used generically as an `<image>`" mean, if  `<1d-image>` is no longer an `<image>`?

SebastianZ: That was just a left-over. I've removed the whole paragraph now as it didn't add any additional information anymore.

Thanks for the hint!

SebastianZ:
I went ahead and merged the changes now as I believe all feedback was addressed and I waited for half a year for review.

I'll create separate issues for my suggestion to change the syntax of `border-color` and for generalizing the `<flex>` value. For reference, see [my previous comment](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1272614499).

If there is still anything left, please let me know and I'll change it.

Sebastian

css-images-4/Overview.bs
```diff
@@ -1897,15 +1894,90 @@ Color Stop “Fixup”</h4>
...
+	The ''stripes()'' function defines a 1d-image
+	as a number of colored stripes with different widths.
+	
+	Each comma-separated entry defines a solid-color stripe,
+	each placed end-to-end on the [=paint line=] in the order given,
+	with the specified <<color>> and thickness.
+	If the thickness is omitted,
+	it defaults to ''1fr''.
+	
+	In each entry, a <<percentage>> is relative to the |painting area|
```
Loirooriol, line 1933: I think this is very confusing, the "painting area" is never defined but since it's an area it seems to be 2D. How to resolve a 1D `<length-percentage>` relatively to a 2D area?

Should this refer to the "length of the paint line" instead?

css-images-4/Overview.bs
Loirooriol: You removed another `used generically as an <<image>>`, but not this one.

css-images-4/Overview.bs
```diff
@@ -2393,6 +2465,39 @@ Interpolating <<gradient>> {#interpolating-gradients}
...
+	Similar to <a lt="interpolating-gradients">gradient images</a>,
+	one-dimensional images can be interpolated directly in CSS transitions and animations,
+	smoothly animating from one image to another.
+	There are only a few restrictions on what images are allowed to be interpolated:
+
+	1. Both the starting and ending image must have the same number of <<color-stripe>>s.
+
+	2. Neither image uses a combination of <<length>>, <<percentage>>, and <<flex>> stripes.
+
+	If the two image satisfy both constraints,
```
Loirooriol, line 2480: If the two image***s***

css-images-4/Overview.bs
```diff
@@ -2393,6 +2465,39 @@ Interpolating <<gradient>> {#interpolating-gradients}
...
+	Note: The abrupt transition at 50% is so that content will not rely on cross-fading,
+	and smarter interpolation rules can be added for this case in the future.
+
+	1. Interpolate each component and stripe of the images independently.
+
+	2. To interpolate a stripe,
+		first match each stripe in the start image
+		to the corresponding stripe at the same index in the end image.
+		Then, for each pair of stripes,
+		interpolate the <a href="https://www.w3.org/TR/css3-transitions/#animtype-length">thickness</a>
```
Loirooriol, line 2498: How to interpolate a `<length-percentage>` with a `<flex>`?

Rather than checking that each image doesn't mix different types, should the requirement be that each pair of matched stripes have the same type?

css-images-4/Overview.bs
Loirooriol: https://github.com/w3c/csswg-drafts/issues/8160

css-images-4/Overview.bs
Loirooriol: https://github.com/w3c/csswg-drafts/issues/8161

css-images-4/Overview.bs
Loirooriol: https://github.com/w3c/csswg-drafts/issues/8162

css-images-4/Overview.bs
Loirooriol: https://github.com/w3c/csswg-drafts/issues/8163

Issue #1031, opened 2017-02-14
[css-contain-2] do we need size containment in a single dimension to enable container queries?


dbaron:
One thing that came up at the [extensible Web summit](https://ti.to/bocoup/extensible-web-summit) in Boston on Friday was a discussion of container queries.

While I'd [previously suggested](https://github.com/ResponsiveImagesCG/container-queries/issues/3#issuecomment-185951645) tying container queries to CSS containment, one thing I realized is that there will be cases where developers want to fix the width from the outside but still allow an auto height, and then do media queries on the container's width.

I haven't thought this through very much, but it seems to me that this may require a concept of layout containment in a single dimension to be exposed from containment.

frivoal:
Specifically, what would be needed is 1d `size` containment. 
https://drafts.csswg.org/css-containment/#containment-size

dbaron:
I think container queries require both [size containment](https://drafts.csswg.org/css-containment/#containment-size) and [layout containment](https://drafts.csswg.org/css-containment/#containment-layout).

frivoal:
I agree. What I meant is that you should be able to apply regular layout containment (which we already have) together with 1d size containment (which we don't) to get the effect you described.

So authors need both, but the WG only needs to add 1d size containment, since we already have the rest.

ausi:
I would really like this feature to help container query scripts to solve the recursion issue. But is size/layout containment in one dimension even possible?

Take, for example, the following structure (assuming that `size-x` would be the value for containment across the X axis):

```html
<style>
	.wrapper { height: 100px; overflow-y: auto }
	.component { contain: size-x }
</style>
<div class="wrapper">
	<div class="component">
		Content that grows in height via a container query script
		when the width gets wider.
	</div>
</div>
```

In an edge case this could still lead to an endless loop I think: The width of `.component` depends on the content-with of `.wrapper` and the content-width of `.wrapper` depends on the height of `.component` (because of the scrollbar). This means that the width of `.component` would still depend on its contents even though `contain: size-x` is set.

gregwhitworth:
@ausi containment only works on the box that it is applied to; so, in your scenario, you're only containing `size-x` for `.component`. Thus whatever `size-x` is defined to do will occur for its content and will then propagate the resulting geometry to `.wrapper`. Regarding scrolling, it this geometry results in the height being over `100px` then you re-layout `.wrapper` with the scrollbar in place (this is how scrollbars work today)

ausi:
[...]

PR #7031, opened 2022-02-10
[css-contain-3] Remove leftover block-size container-type #1031

lilles:
The definition of container-type:block-size was removed when block-size
containment was removed, but the value syntax still had it.


lilles:
This has been fixed in another PR

Issue #7033, opened 2022-02-10
[css-contain-3] Add CSSOM API for CSSContainerRule


lilles:
The spec needs to specify a CSSContainerRule interface for @container rules. It should extend the CSSConditionRule.

tabatkins:
You should be able to rely on it; just copy @supports, yeah.

mirisuzanne:
Drafted. Agenda+ for a resolution on this approach.

anttijk:
Please make any attributes `readonly`.

lilles:
conditionText is resolved to be readonly in issue #6819, but the spec hasn't been changed yet. I added the NeedsEdits label to the issue.

astearns:
I think we could resolve on this approach asynchronously. Once the spec has the `readonly` edit in, I’ll call for objections here and on the mailing list.

mirisuzanne:
Since `conditionText` is defined in a separate spec (css-conditional-3), I don't think the readonly change requires any edits on this `CSSContainerRule` definition?

lilles:
> Since `conditionText` is defined in a separate spec (css-conditional-3), I don't think the readonly change requires any edits on this `CSSContainerRule` definition?

No.

SebastianZ:
Maybe we should also have a getter for the query container. So, depending on what's set for `containerName` and `containerQuery` it returns the related element.

Sebastian

SebastianZ:
In the discussion in #4240, I realized that it may also be useful for authors as well as developer tools to have a getter that exposes the matching query container.

So something like `queryContainer` that uses the name and condition and spits out the element that currently matches them.

Sebastian

PR #7034, opened 2022-02-10
[css-contain-3] Add APIs section with CSSContainerRules #7033

lilles:
Might need to add @container specifics prose for conditionText like we
have for @support, but the syntax is e.g. currently missing
<general-enclosed>.


lilles:
Someone else added this

PR #7038, opened 2022-02-11
[css-fonts-4] Use last key/palette index in override-colors, #7026

### LINKED ISSUE #7026: [css-fonts-4] Inconsistent order of override-colors precedence
https://drafts.csswg.org/css-fonts-4/#override-color

> If the keys of multiple distinct key/value pairs identify the same color index (either by name or by integer), the first key is used for the purposes of rendering. However, for serialization purposes, both key/value pairs are present.

IMO this seems inconsistent with the usual ways of CSS in which later declarations override earlier ones. For the sake of consistency I suggest to change that to the last indexed override's value prevailing.

Was there some context that I missed for why it was defined this way around?

drott:
Improve consistency with other parts of CSS where later declarations
override earlier ones, fixes #7026.


svgeesus (approved):
Agreed

PR #7086, opened 2022-02-24
[css-fonts-5] Fix typo in tech() function

### LINKED ISSUE #6791: [css-fonts-4] Naming font-technology() supports function
In https://www.w3.org/2021/10/20-css-minutes.html#r04 we resolved to add, roughly speaking, a technology() function to the `src` descriptor and an equivalent font-technology() function to `@supports` which would take keywords representing font technology capabilities to check for UA support. We also resolved to open a bikeshedding issue to see if there's a better name for the function, so this is that issue.

drott:
Correct typo techn() -> tech() for font technology function.

Follow up to 102da5ae79, also fixes #6791.

### DIFF
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -357,7 +357,7 @@ according to section [[css-syntax#parse-comma-separated-list-of-component-values
 Then each component value is parsed according to this grammar:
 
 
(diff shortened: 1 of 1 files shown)


svgeesus (commented):
Thanks!

PR #7105, opened 2022-03-06
[css-shapes-2] Add the "rotate" keyword to shape()

### LINKED ISSUE #6925: [css-shapes-2] Ambiguity when creating an arc between angle & ry
See https://drafts.csswg.org/css-shapes-2/#typedef-shape-arc-command:
`<arc-command> = arc <by-to> <coordinate-pair> of <length-percentage>{1,2} [ <arc-sweep> || <arc-size> || <angle> ]`

There could be arguments that are ambiguous at time of parsing between the second `length-percentage` radius and the `angle`.
For example, unclear how to parse:
`arc to 10px 10px of 5px calc(var(--some-number) * 2) cw`

Is the `calc` value the angle or the y radius?

Even if it was possible for the parser to figure this out by following the calc parameters, this is ambiguous to the reader.

Options:
- Make it mandatory to have 2 radii, maybe allowing `auto` or so for the second one to represent symmetrical radii, like:

`arc to 10px 10px of 5px auto calc(var(--some-number)*2) cw // angle`
`arc to 10px 10px of 5px calc(var(--some-number)*2) cw // radius`

- add a keyword or a function(`rotate`?) before the angle, like:
`arc to 10px 10px of 5px rotate(calc(var(--some-number)*2)) cw // angle` or `arc to 10px 10px of 5px rotate calc(var(--some-number)*2) cw // angle`
`arc to 10px 10px of 5px calc(var(--some-number)*2) cw // radius`

I think I prefer the `rotate` as keyword option. Maybe there are more options?

noamr:
Avoids an ambiguity where the second radius and angle can both come from calc

Closes #6925

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -171,7 +171,7 @@ The ''shape()'' Function</h4>
 		<<curve-command>> = curve <<by-to>> <<coordinate-pair>> via <<coordinate-pair>>{1,2}
 		<<smooth-command>> = smooth <<by-to>> <<coordinate-pair>> [via <<coordinate-pair>>]?
 		<<arc-command>> = arc <<by-to>> <<coordinate-pair>> of <<length-percentage>>{1,2}
-		                    [ <<arc-sweep>> || <<arc-size>> || <<angle>> ]?
+		                    [ <<arc-sweep>> || <<arc-size>> || rotate <<angle>> ]?
 		<<arc-sweep>> = cw | ccw
 		<<arc-size>> = large | small
 	</pre>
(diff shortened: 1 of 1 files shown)


noamr:
@tabatkins ?

tabatkins:
I'm not against this in principle (it probably does make things slightly more readable, actually), but what's the ambiguity concern? You can statically determine the type of a calc(); it'll only be a valid radius if the type is {length:1} or {percentage:1}, and a valid angle if the type is {angle:1}.

noamr:
> I'm not against this in principle (it probably does make things slightly more readable, actually), but what's the ambiguity concern? You can statically determine the type of a calc(); it'll only be a valid radius if the type is {length:1} or {percentage:1}, and a valid angle if the type is {angle:1}.

I didn't know that! But in any case it's ambiguous for reading, especially with custom properties

tabatkins:
Yeah that's very fair. And the whole point of this syntax is optimizing for authorability and readability, so yeah, let's go for it.

PR #7138, opened 2022-03-14
Add explainer for CSS-fonts-4 `font-palette`

drott:
As I am filing a FYI tag review, adding an explainer for font-palette.


svgeesus (approved):
Thanks for putting this together!

PR #7159, opened 2022-03-19
[css-fonts-4] Convert @font-palette-values names to <dashed-ident>

devongovett:
Noticed that the names in the examples for `@font-palette-values` and `font-palette` were often not `<dashed-ident>` values as the spec requires. Updated them accordingly.


svgeesus:
Thanks, but we already landed [another PR for the same issue](https://github.com/w3c/csswg-drafts/pull/7168)

PR #7165, opened 2022-03-22
Run PNGs through oxipng

gsnedders:
This essentially takes the PNGs that are shrunk by `oxipng -s -o 3` then re-runs them through `oxipng -s -o max -Z`; the reason to do this subset is to avoid changing too much from much smaller gains, and instead focus on the images that have larger gains (which will almost be due to removing metadata that doesn't affect rendering).


tabatkins (approved):
Looks good from a quick scan of all the images. My colorblindness makes assessment of color quality suspect for the color/gradient ones, but afaict they're unchanged, so :+1:

PR #7175, opened 2022-03-28
[cssom-view-1] Fix remaining IDL instances of "bool"

tidoust:
This is a follow-up to #7162 to fix occurrences of `bool` that were missed in the previous PR.

(Sorry I missed these in the first PR. I checked the resulting IDL this time, we should be all good.)


emilio:
Thanks!

PR #7176, opened 2022-03-28
[css-fonts-4]: Fix font-palette and @font-palette-values should use `<dashed-ident>`

yisibl:


yisibl:
This is the first CSS spec contribution of my life and I'm still late to the party. 😂🤣

svgeesus:
🥇 Please don't let that put you off! 👍

PR #7192, opened 2022-04-01
[css-syntax-3] Fix two typos in `<urange>` parsing algorithm

AtkinsSJ:
Noticed these while implementing this yesterday.


astearns:
Thanks!

PR #7201, opened 2022-04-06
[css-flexbox-1] Add closing tag to the mocked markup

wolewnicki:
The mockup for the flex box demonstration was missing closing tags.

I believe this issue is `non-substantive`


tabatkins:
Thanks, but the code is correct as-is.

Issue #6819, opened 2021-11-16
[css-conditional-3] Setting .conditionText interop is terrible


fantasai:
The CSS Conditional Rules spec specifies that `conditionText` can be set, and if it is grammatical, updates the condition applied to the CSSConditionRule to which it applies.

So far, Firefox mostly supports this correctly on CSSMediaRules, and mostly fails on CSSSupportsRule; and neither Blink nor WebKit seem to support setting conditionText at all. See [testcase](https://github.com/web-platform-tests/wpt/pull/31651) Do we want to file bugs, or redefine conditionText as readonly?

Loirooriol:
I think this should be consistent `CSSStyleRule`'s `selectorText`. I remember getting frustrated at some point because that was readonly on Firefox, but later [it got fixed](https://bugzilla.mozilla.org/show_bug.cgi?id=37468).

tabatkins:
Being able to update the condition for a rule is somewhat useful for tools today, as you can use it to toggle entire blocks of rules easily by just setting the condition to something that's always-true or always-false.  Ultimately we want to solve this use-case *better* with custom MQs, but given that setting the conditionText does work in at least some places today, I think it's reasonable to preserve that if possible.

svgeesus:
Fixed by https://github.com/w3c/csswg-drafts/commit/6c659d6e6b8675a80d4face3a904985605c643e0

cdoublev:
I think the following sentence should be removed in [7.3. The CSSMediaRule interface](https://drafts.csswg.org/css-conditional-3/#the-cssmediarule-interface):

  > Setting the `conditionText` attribute must set the `media.mediaText` attribute on the rule.

PR #7205, opened 2022-04-08
[css-conditional] Make conditionText readonly.

emilio:
As per https://github.com/w3c/csswg-drafts/issues/6819.


svgeesus (approved):
LGTM, as resolved

PR #7221, opened 2022-04-15
[css-pseudo] no decoration propagation on highlights (closes #6829)

### LINKED ISSUE #6829: [css-pseudo] highlights and decoration propagation
When selecting the following ([demo](https://bucket.daz.cat/work/igalia/0/30.html)):

```html
<style>
.example1::selection { text-decoration: blue underline; }
.example1 sup::selection { text-decoration: green overline; }
.example2::selection { text-decoration: #66339980 underline; }
.example3::selection { text-decoration: underline; }
</style>
<span class="example1">Hello, <sup>world</sup>!</span>
<span class="example2">Goodbye, <span>space</span>!</span>
<span class="example3">On my way, <sup>core</sup>!</span>
```

1. **Should “world” have (a) green overline and blue underline, or (b) green overline only?** The underlying question is… do highlights support something like [decoration propagation](https://drafts.csswg.org/css-text-decor-4/#line-decoration)?
2. **Should “space” be (a) underlined twice yielding a darker line, (b) underlined once yielding the same line colour, or (c) not underlined at all?** The underlying question is… if highlights *do* propagate decorations, wouldn’t that interact poorly with [highlight inheritance](https://drafts.csswg.org/css-pseudo-4/#highlight-cascade) treating all properties as inherited?
3. **Should “core” appear to have underlines on (a) superscript baseline *and* normal baseline, (b) superscript baseline only, or (c) normal baseline only?** The underlying question is… if highlights *don’t* propagate decorations, with text instead relying on inheritance to still be decorated in descendants, wouldn’t that be inconsistent with how decorations should generally render in descendants?

We want to be careful not to inadvertently favour the incorrect decoration impl in Blink and WebKit, where descendants paint propagated decorations as if they were specified directly — equivalent to 3(b) in non-highlight content.

(cc @frivoal, @fantasai, @mrego, @MatsPalmgren)

delan:
This patch makes edits for #6829, stating that decorations introduced by highlight pseudos are propagated through inheritance rather than through decorating boxes (the opposite of originating decorations).

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -988,6 +988,23 @@ Text and Text Decorations</h4>
     </div>
   </div>
 
+  While the element’s own
+  <a href="https://www.w3.org/TR/css-text-decor/#line-decoration">line decorations</a>
+  are propagated from
+  containing <a spec=css-text-decor>decorating boxes</a>
+  as usual,
+  decorations introduced by [=highlight pseudo-elements=]
+  apply only to the text
+  directly associated with the [=originating element=]
+  (or intervening [=pseudo-element=] such as ''::first-line'' or ''::first-letter'').
+
+  Note: Because highlight decorations
+  are propagated through inheritance,
+  not through the box tree,
+  they can be declared on the root element
+  and still reliably decorate out-of-flow descendants,
+  which is not the case for non-highlight decorations.
+
 <h4 id="highlight-replaced">
 Replaced Elements</h4>
 
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -988,22 +988,15 @@ Text and Text Decorations</h4>
     </div>
   </div>
 
-  While the element’s own
-  <a href="https://www.w3.org/TR/css-text-decor/#line-decoration">line decorations</a>
-  are propagated from
-  containing <a spec=css-text-decor>decorating boxes</a>
-  as usual,
-  decorations introduced by [=highlight pseudo-elements=]
-  apply only to the text
-  directly associated with the [=originating element=]
-  (or intervening [=pseudo-element=] such as ''::first-line'' or ''::first-letter'').
+  <a href="https://www.w3.org/TR/css-text-decor/#line-decoration">Line decorations</a> introduced by [=highlight pseudo-elements=]
+  apply only to the text associated with their [=originating element=],
+  and are not propagated to descendants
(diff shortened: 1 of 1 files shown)


css-pseudo-4/Overview.bs
```diff
@@ -988,6 +988,23 @@ Text and Text Decorations</h4>
...
 
+  While the element’s own
+  <a href="https://www.w3.org/TR/css-text-decor/#line-decoration">line decorations</a>
+  are propagated from
+  containing <a spec=css-text-decor>decorating boxes</a>
+  as usual,
+  decorations introduced by [=highlight pseudo-elements=]
+  apply only to the text
+  directly associated with the [=originating element=]
+  (or intervening [=pseudo-element=] such as ''::first-line'' or ''::first-letter'').
```
fantasai, line 999: Possible shorter rephrasing?

```suggestion
  <a href="https://www.w3.org/TR/css-text-decor/#line-decoration">Line decorations</a> introduced by [=highlight pseudo-elements=]
  apply only to the text associated with their [=originating element=],
  and are not propagated to descendants
  except via property [=inheritance=] (as described [[#highlight-cascade|above]]).
```

fantasai (changes requested):

fantasai:
Overall looks good, but I have some suggestions for improvement. :)

css-pseudo-4/Overview.bs
```diff
@@ -988,6 +988,16 @@ Text and Text Decorations</h4>
     </div>
   </div>
 
+  <a href="https://www.w3.org/TR/css-text-decor/#line-decoration">Line decorations</a> introduced by [=highlight pseudo-elements=]
+  apply only to the text associated with their [=originating element=],
+  and are not propagated to descendants
+  except via property [=inheritance=] (as described [[#highlight-cascade|above]]).
+
+  Note: Unlike the element’s own decorations,
+  decorations declared on the root element’s highlights
```
fantasai, line 997:
```suggestion
  Note: Unlike the originating element’s own decorations,
  decorations declared on a highlight pseudo-element
```

fantasai: Slight fix, because it's not just the root element's highlight that gets this effect. :)

fantasai:
Feel free to "rebase and merge" once you've fixed the comment above~

css-pseudo-4/Overview.bs
delan: done in 6dbf3d9906f48963aefed24fa7bebd986f9149f6

PR #7222, opened 2022-04-15
[css-pseudo] allow underline-related properties on highlights (closes #7101)

### LINKED ISSUE #7101: [css-pseudo] highlight pseudos and emphasis/underline properties
https://drafts.csswg.org/css-pseudo-4/#highlight-styling

> The following properties apply to the highlight pseudo-elements:
> * […]
> * [text-decoration](https://drafts.csswg.org/css-text-decor-4/#propdef-text-decoration) and its associated properties
> * [text-shadow](https://drafts.csswg.org/css-text-decor-4/#propdef-text-shadow)
> * […]

The current spec text could be interpreted as only allowing ‘text-decoration’ and ‘text-decoration-*’ and ‘text-shadow’, which would include line decorations and text shadows, but not emphasis marks or ‘text-underline-position’ or ‘text-underline-offset’. Should we somehow allow those properties more explicitly?

(@fantasai, @frivoal, @mrego)

delan:
This patch makes edits for #7101, adding ‘text-underline-position’, ‘text-underline-offset’, ~~and ‘text-emphasis-color’~~ to the list of those applicable to highlight pseudos.

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -675,11 +675,7 @@ Styling Highlights</h3>
     css/css-pseudo/active-selection-012.html
     css/css-pseudo/active-selection-031.html
     </wpt>
-    <li>'text-decoration' and its associated properties, including
-    <ul>
-      <li>'text-underline-position'
-      <li>'text-underline-offset'
-    </ul>
+    <li>'text-decoration' and its associated properties (including 'text-underline-position' and 'text-underline-offset')
     <wpt>
     css/css-pseudo/active-selection-004-manual.html
     css/css-pseudo/active-selection-014.html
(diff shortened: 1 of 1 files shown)


css-pseudo-4/Overview.bs
```diff
@@ -675,7 +675,11 @@ Styling Highlights</h3>
     css/css-pseudo/active-selection-012.html
     css/css-pseudo/active-selection-031.html
     </wpt>
-    <li>'text-decoration' and its associated properties
+    <li>'text-decoration' and its associated properties, including
+    <ul>
+      <li>'text-underline-position'
+      <li>'text-underline-offset'
+    </ul>
```
fantasai, line 682: I might suggest doing this as 
```suggestion
    <li>'text-decoration' and its associated properties (including 'text-underline-position' and 'text-underline-offset')
```

so that it reads more as a clarification, and even less as an exhaustive list.

delan: done in 8c15aead880bc2a761dd3629cfe0e3480dd21c32

Issue #7215, opened 2022-04-14
[css-ui-4] appearance-disabling property origins


fantasai:
This is an issue against a detail of the text in https://github.com/w3c/csswg-drafts/pull/6537

> Let |authorProps| be a [=map=] of all properties that have a <a>cascaded value</a> for |element|, where the [=Author Origin=], [=Animation Origin=] and [=Transition Origin=] [=Cascade Origins=] are included and the [=User Origin=] and [=UA Origin=] [=cascade origins=] are ignored, where the key is the property name and the value is that <a>cascaded value</a>.

I don't think it makes sense to include the Transition Origin here. A declaration can only exist in the Transition Origin if there is a related declaration in another origin, as it represents the transition between two declarations outside the Transition Origin. So first, I assert that the Transition Origin doesn't need to be inspected. And second, suppose we were to adopt this spec text and the UA defined a rule like input:focus:hover { border-color: orange; } and gave it a 1s transition. Then for the 1s transition, we'd need to treat the element as having `appearance: none`... which would be absurd.

I think it should be sufficient to inspect the Author Origin and Animation Origin only. And maybe only the Author Origin, as I doubt there's content out there relying on animation rules disabling native `appearance`.

CC @zcorpan @emilio

zcorpan:
cc @andruud @lilles

I see this test http://wpt.live/css/css-ui/appearance-transition.tentative.html which references https://bugs.chromium.org/p/chromium/issues/detail?id=1086732

If I understand correctly, the current chromium behavior is to *not* include the transition and animation origins, but also while a transition or animation is active they don't change whether an author-origin declaration is present. Correct?

andruud:
@zcorpan Yes, a property only counts as "author-styled" if the cascaded value has author origin, and only exactly that origin.

zcorpan:
Thanks, submitted https://github.com/w3c/csswg-drafts/pull/7230

zcorpan:
There was an open PR to test the animation and transition origins at https://github.com/web-platform-tests/wpt/pull/25439

PR #7224, opened 2022-04-18
[css-ui-4] Provide hooks for HTML to define rendering for widgets

frivoal:
@fantasai and I gave a stab at the refactoring of https://github.com/w3c/csswg-drafts/pull/6537 we had been talking about in https://github.com/w3c/csswg-drafts/issues/3526. Goal: avoid having to talk about specific HTML elements in the CSS spec, and in particular, avoid having to have long lists of types of widgets.

Doing that was undesirable because:
* CSS, at least in theory, is not limited to HTML.
* The descriptions of each types of widgets were stubs “a button is a <a>widget</a> which must have the <a>native appearance</a> of a button”,  while the HTML spec held the actual descriptions of the rendering of various elements in its Rendering chapter.
* This created a bunch of unnecessary indirections between the two specs, making both of them longer and harder to read than necessary.

The resulting spec is much more straightforward and much shorter, so we think this is a better approach.
This PR is the CSS part of this refactor. For the HTML part of the refactor, look here: https://github.com/frivoal/html/pull/new/appearance-compute-widget-3-updated-take-2 (Note: this is not yet submitted as a PR, pending CSSWG review). The HTML part is complete in the sense that it works with the CSS part, but like the original PR from @zcorpan, lacks details about the rendering of each type of widgets. I'd expect the HTML folks to keep iterating from there.

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -1999,7 +1999,7 @@ to match the control to platform conventions for the use of the accent color.
 
 <pre class="propdef">
 Name: appearance
-Value: ''appearance/none'' | ''auto'' | ''textfield'' | ''menulist-button'' | <<compat-auto>>
+Value: ''appearance/none'' | ''auto'' | <<compat-auto>> | <<compat-special>>
 Initial: none
 Applies To: all elements
 Inherited: no
@@ -2020,7 +2020,8 @@ This specification introduces the 'appearance' property
 to provide some control over this behavior.
 In particular, using ''appearance: none'' allows authors
 to suppress the <a>native appearance</a> of <a>widgets</a>,
-so that CSS can be used to restyle them.
+giving them a <dfn export>primitive appearance</dfn>
+where CSS can be used to restyle them.
 
 <details class=note>
 	<summary>Note on the history of this feature</summary>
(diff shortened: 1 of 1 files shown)


zcorpan:
Thank you! This indeed seems like a nicer solution. I see that `<input type=search>` with `appearance: textfield` and `<select>`'s devolved state is `menulist-button` in https://github.com/frivoal/html/pull/new/appearance-compute-widget-3-updated-take-2 , so that's good.

From a first read, it looks equivalent except:

- transition origin is excluded (as per https://github.com/w3c/csswg-drafts/issues/7215)
- `revert-level` is handled the same as `revert`. Is this intended to be [`revert-layer`](https://drafts.csswg.org/css-cascade-5/#revert-layer)?

frivoal:
> From a first read, it looks equivalent except:
> *   transition origin is excluded (as per https://github.com/w3c/csswg-drafts/issues/7215)

Correct. Let's get to the bottom of that difference in that issue.

> * revert-level is handled the same as revert. Is this intended to be [revert-layer](https://drafts.csswg.org/css-cascade-5/#revert-layer)?

Oops. Nice catch, thank you. Fixed.

css-ui-4/Overview.bs
```diff
@@ -2193,11 +2200,82 @@ so this specification needs to say something about this.
...
+<h4 id="appearance-disabling-properties">
+Properties Disabling Native Appearance</h4>
+
+Certain properties, when declared in the [=Author Origin=] or [=Animation Origin=],
+will disable the [=native appearance=] of certain [=widgets=].
+Specifically,
+if all the declarations of the [=Author Origin=] and [=Animation Origin=]
+pertaining to such a [=widget=] are cascaded
+and a [=cascaded value=] exists--
+and is not ''revert'' or ''revert-layer''--
```
Loirooriol, line 2213: What about

```css
@layer {
  .widget { border-left-style: solid }
}
@layer {
  .widget { border-left-style: revert-layer }
}
```

The cascaded value is `revert-layer`, but the native appearance should be disabled anyways since the specified value will be `border-left-style: solid`.

frivoal: filed as https://github.com/w3c/csswg-drafts/issues/7252

PR #7230, opened 2022-04-25
[css-ui] Remove Animation Origin for 'appearance' disabling

### LINKED ISSUE #7215: [css-ui-4] appearance-disabling property origins

zcorpan:
Fixes #7215

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -2204,10 +2204,10 @@ as a [=legacy name alias=] of 'appearance'.
 <h4 id="appearance-disabling-properties">
 Properties Disabling Native Appearance</h4>
 
-Certain properties, when declared in the [=Author Origin=] or [=Animation Origin=],
+Certain properties, when declared in the [=Author Origin=],
 will disable the [=native appearance=] of certain [=widgets=].
 Specifically,
-if all the declarations of the [=Author Origin=] and [=Animation Origin=]
+if all the declarations of the [=Author Origin=]
 pertaining to such a [=widget=] are cascaded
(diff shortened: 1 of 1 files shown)


andruud (approved):

fantasai (approved):

frivoal:
Merging in under editorial discretion, as the most engaged people have opined already, and we have data about what's already implemented, so it's more of a bug fix than anything.

Ping to the chairs (cc: @astearns @atanassov ), if you'd like to agenda+ nonetheless.

PR #7284, opened 2022-05-17
[css-text-decor-4] Clarified constraints for thickness length and percentage #7283

### LINKED ISSUE #7283: [css-text-decor-4] Clarify thickness constraints
The current thickness constraints for [length](https://drafts.csswg.org/css-text-decor-4/#valdef-text-decoration-thickness-length) and [percentage](https://drafts.csswg.org/css-text-decor-4/#valdef-text-decoration-thickness-percentage) state that 
> The UA must floor the actual value at one device pixel.

This could be confusing and interpreted as performing a `floor` on the actual calculated value, while it is actually rounded to the nearest integer, with a minimum value of one.
Equivalent to: `x = min(1, round(x))`

tcaptan-cr:
The current sentence could be interpreted as performing a `floor` on the actual calculated value, while it is actually rounded to the nearest integer, with a minimum value of one.
Equivalent to: `x = min(1, round(x))`

Fixes #7283

### DIFF
--- a/css-text-decor-4/Overview.bs
+++ b/css-text-decor-4/Overview.bs
@@ -339,7 +339,8 @@ Text Decoration Line Thickness: the 'text-decoration-thickness' property</h3>
 		<dt><dfn><<length>></dfn>
 		<dd>
 			Specifies the thickness of text decoration lines as a fixed length.
-			The UA must floor the actual value at one device pixel.
+			The UA should round the actual value to the nearest integer,
+			and ensure it is at least one device pixel.
 
 			Note: A length will inherit as a fixed value,
 			and will not scale with the font.
@@ -347,7 +348,8 @@ Text Decoration Line Thickness: the 'text-decoration-thickness' property</h3>
 		<dt><dfn><<percentage>></dfn>
 		<dd>
 			<p>Specifies the thickness of text decoration lines as a percentage of ''1em''.
-			The UA must floor the actual value at one device pixel.
+			The UA should round the actual value to the nearest integer,
+			and ensure it is at least one device pixel.
 
 			Note: A percentage will inherit as a relative value,
 			and will therefore scale with changes in the font as it inherits.
--- a/css-text-decor-4/Overview.bs
+++ b/css-text-decor-4/Overview.bs
@@ -339,7 +339,8 @@ Text Decoration Line Thickness: the 'text-decoration-thickness' property</h3>
 		<dt><dfn><<length>></dfn>
 		<dd>
 			Specifies the thickness of text decoration lines as a fixed length.
-			The UA should round to the nearest positive integer.
+			The UA should round the actual value to the nearest integer device pixel,
+			and ensure it is at least one device pixel.
 
 			Note: A length will inherit as a fixed value,
 			and will not scale with the font.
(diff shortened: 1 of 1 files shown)


css-text-decor-4/Overview.bs
```diff
@@ -339,15 +339,17 @@ Text Decoration Line Thickness: the 'text-decoration-thickness' property</h3>
 		<dt><dfn><<length>></dfn>
 		<dd>
 			Specifies the thickness of text decoration lines as a fixed length.
-			The UA must floor the actual value at one device pixel.
+			The UA should round the actual value to the nearest integer,
```
progers, line 342: WDYT of this rephrasing?
`The UA should round to the nearest positive integer.`

This removes the mention of "device pixel" which is difficult for implementations, and is a little simpler.

tcaptan-cr: Sounds good to me.

dbaron (commented):
I'm a little worried that putting the key bit about flooring in the single word "positive" might hide it from many readers.  I think it's still worth having two separate clauses to mention the round-to-nearest-device-pixel behavior and the minimum-of-one-device-pixel behavior.

Also, you should put "Fixes #7283" in the commit message so that the issue is linked.

progers:
> I'm a little worried that putting the key bit about flooring in the single word "positive" might hide it from many readers. I think it's still worth having two separate clauses to mention the round-to-nearest-device-pixel behavior and the minimum-of-one-device-pixel behavior.
> 

I defer to David about the best way to phrase spec language for clarity, so I'm okay with switching back to the two-clause approach.

Re: mentioning "device pixels". The Chrome implementation doesn't generally have access to pixel snapping at this place in the pipeline. As an example, would spec language referring to pixel snapping here require that we re-snap every frame when under a scale transform animation?

tcaptan-cr:
How does this sound?
> The UA should round the actual value to the nearest integer,
and ensure it has a value of at least 1.

dbaron:
> Re: mentioning "device pixels". The Chrome implementation doesn't generally have access to pixel snapping at this place in the pipeline. As an example, would spec language referring to pixel snapping here require that we re-snap every frame when under a scale transform animation?

We still need to say something about what units the numbers have.  There's no default here.

Are you suggesting that what's really relevant here is *raster pixels*, which are usually but not always device pixels?

progers:
@dbaron, after playing with this a bit, I think I have a better understanding of what _device pixels_ means.

When loading the following example on a highdpi (dsf=2) monitor, the blue and green lines have the same thickness (in Chrome Canary and Firefox), but the red line is thicker. This is because Chrome/Firefox draw the blue and green lines at (1.1 css pixels * dsf (2)) = 2 device pixels, whereas the red line draws at round(1.4 css pixels * dsf (2)) = 3 device pixels.
```
  <div style="position: absolute; top: 25px; left: 5px; width: 20px; height: 1px; background: blue;"></div>
  <span style="color: green; text-decoration: underline; text-decoration-thickness: 1.1px;">&nbsp;&nbsp;</span>
  <span style="color: red; text-decoration: underline; text-decoration-thickness: 1.4px;">&nbsp;&nbsp;</span>
```
Is this dsf behavior what is meant by _device pixels_? If so, I agree that referring to _device pixels_ makes sense. We would not want to round to css pixels (e.g., (round(1.4 css pixels) * dsf (2)) = 2 device pixels for the red line) because that would limit the granularity of the decoration thickness on highdpi monitors.

dbaron:
@progers Yes, that's the core of the reason to refer to device pixels rather than some other length unit like CSS pixels.  However, as you point out in https://github.com/w3c/csswg-drafts/pull/7284#issuecomment-1129354642 , I think this isn't quite right.  I've filed #7287 to improve the situation, but I think given current terminology it's probably good enough to continue to refer to device pixels here (as the existing spec text already does).

tcaptan-cr:
Thank you for the suggestions!

dbaron (approved):

fantasai (approved):

Issue #7203, opened 2022-04-07
[css-contain] <container-name> in @container ambiguities


emilio:
Two thoughts:

 * `<container-name>` kinda conflicts with `<container-condition>`, as `not` can be a valid `<container-name>`, how should `@container not (width > 300px)` be parsed? There's an obvious answer, but probably that means that `not` should not be a valid `<container-name>`?

 * Per spec `<container-name>` includes `none` and multiple `<custom-idents>`, is `@container foo bar (width > 300px)` expected to be valid? I don't see any test for multiple container names in the rule, so just checking whether that's intentional or not.

cc @andruud @mirisuzanne @anttijk

yisibl:
Can the specification add examples or other instructions to further clarify the `not` case?

Implementers are still susceptible to misunderstandings or misinterpretations. see: https://github.com/parcel-bundler/parcel-css/issues/158#issuecomment-1181343944

> That's not what the grammar in the spec says though. I don't see a test that implies it either. It only tests that not is not a valid container name, not that it can't start a condition.

mirisuzanne:
Yes, both of these are valid, with `not` being part of the condition rather than the name. If implementations are getting that wrong, we should at least create a WPT, and file bugs - and potentially clarify the spec. 

```css
@container not (width <= 500px ) { … }
@container name not (width <= 500px) { … }
```

lilles:
Since the spec changed without test changes after the implementation/tests were written, the spec change went unnoticed and the previous syntax is what's currently implemented in both Chrome and Safari.

lilles:
Should all be covered by https://wpt.live/css/css-contain/container-queries/at-container-parsing.html

PR #7292, opened 2022-05-20
[css-contain-3] Exclude 'none' and 'not' from custom-ident. #7203

lilles:
Current proposal to disallow 'not' and 'none' as container-names.


mirisuzanne:
This looks good to me, pending a resolution on the issue. I've marked it agenda+. I think we'd also want the change documented in the changelog.

mirisuzanne (approved):

PR #7294, opened 2022-05-20
[css-contain-3] Disallow not/none for container-name in @-rule. #7203

lilles:
Use a separate definition of container-name in the @container syntax to
disallow 'none' and 'not'.


lilles:
With this PR, only a single container-name is allowed in an `@container` rule. If we should allow multiple names, the spec needs to define how the container-name matching happens with multiple names.

lilles:
This PR also clarifies that only a single container-name is allowed in  @container.

mirisuzanne (approved):

PR #7300, opened 2022-05-23
[css-contain-3] Fix link errors

andruud:
- Make 'fictional tag sequence' link to something.
 - Use 'ident sequence', not 'identifier'.
 - Remove references to `<<size-query>>`, as that no longer exists, and:
 - Also remove references to `<<style-query>>`, to be consistent.
 - '@container/orientation' instead of 'orientation'.
 - Link to `<<'container-type'>>` instead of the now non-existent
   `<<container-type>>`.


css-contain-3/Overview.bs
```diff
@@ -477,7 +477,7 @@ Naming Query Containers: the 'container-name' property</h3>
 		Initial: none
 		Inherited: no
 		Applies to: all elements
-		Computed Value: a set of unique [=identifiers=]
```
tabatkins, line 480: No, this should be identifiers or `<custom-ident>` or whatever; ident sequence described a sequence of characters, not a CSS value.

tabatkins (changes requested):
The two ident changes need to be changed to refer to a CSS Values concept, not a CSS Syntax concept, but the rest look good.

andruud:
@tabatkins Happy with this now?

Issue #6343, opened 2021-06-03
[mediaqueries] display mode media feature


frivoal:
In https://github.com/w3c/manifest/issues/975, @marcoscaceres said:

>While the CSS crew is here, somewhat unrelated, but related, could we move "display mode media feature" to MQ5?

The current spec is over here: https://www.w3.org/TR/appmanifest/#the-display-mode-media-feature
Support according to MDN: 
![Screen Shot 2021-06-03 at 13 39 24](https://user-images.githubusercontent.com/113268/120587887-29689380-c471-11eb-84bd-0ce44999363c.png)

frivoal:
I'd say that we should indeed merge it into the main MQ spec.

marcoscaceres:
Great! Let me know if I can be of help in moving stuff over.

frivoal:
Waiting for clarifications in https://github.com/w3c/manifest/issues/1023

marcoscaceres:
Looking forward on your thoughts on the proposal in w3c/manifest#1023

frivoal:
Publication has happened (since we're no worse off than we were before the transfer), so closing. Will follow up in the other issue.

PR #7307, opened 2022-05-25
[mediaqueries-5] Move the definition of display-mode back to APPMANIFEST. Closes #7306.

### LINKED ISSUE #7306: [mediaqueries-5] Move the definition of "display mode" back to Manifest spec
(Attn: @frivoal @marcoscaceres )

In #6343, two sections of the Web App Manifest spec were moved into the CSS Media Queries 5 spec:

* "Display modes"
* "The `'display-mode'` media feature"

The resulting negative diff on Manifest can be seen here: https://github.com/w3c/manifest/pull/1022/files

I believe the first of these moves was a mistake and would like to move the "display modes" section back to Manifest where it belongs. Moving the "The `'display-mode'` media feature" section to CSSMQ was correct.

(Apologies that I've just noticed this now. It happened in December as I was planning to go on leave and I have been absent since then.)

**Background**

This started from an off-hand comment by @marcoscaceres in https://github.com/w3c/manifest/issues/975:

> While the CSS crew is here, somewhat unrelated, but related, could we move "display mode media feature" to MQ5?

I believe the intention here was to move *just* the "display mode media feature" to MQ5 (which is correct). However what ended up happening was that *both* the "display modes" and "display mode media feature" section got moved. Essentially, the "display mode media feature" was "remotely" injecting new content into the CSS spec, and I believe the intention was always to move it into CSS when it was ready. But I think moving the entire definition of "display mode" was a mistake.

**Proposal**

Move the "display modes" section (including the definition of the four display modes: browser, minimal-ui, standalone and fullscreen) back to the Manifest spec.

Keep the "The `'display-mode'` media feature" section in the CSSMQ spec, referencing the definition of "display mode" in the Manifest spec.

**Rationale**

The "display mode" section, which defines [display mode](https://www.w3.org/TR/mediaqueries-5/#display-mode) and provides the definition of the four display modes (browser, minimal-ui, standalone and fullscreen) is one of the core concepts of the web app manifest. The Web App Manifest spec is concerned with what it means to be a web app, and largely this revolves around being displayed in a non-browser display mode.

The Web App Manifest spec defines the `"display"` member which can be one of the four display modes, and allows the web app to control which display mode it should be presented in. So far, this is all self-contained within the Manifest spec and has nothing to do with CSS whatsoever. The definition of the display modes is critical to the manifest spec (hence the phrase appears 30 times in the spec).

In addition to this, we defined the `display-mode` CSS media query, which allows pages to query which app display mode they are in. While this is a useful feature, it is ancillary. We could delete the display-mode media query and the entire web app system would continue to function, albeit it would be harder for pages to determine which mode they're in.

Having the entire definition of "display mode" be in the CSS spec makes no sense, because it is a manifest concept that happens to have a CSS media query associated with it, not a CSS concept that happens to be referenced by the manifest spec. It doesn't make sense that the Manifest spec defines the "`display`" member by saying "see the CSS spec".

My practical concerns over this are that it will unnecessarily complicate future discussions around display modes. We (Manifest authors) have plans to add at least two new display modes ([tabbed mode](https://github.com/w3c/manifest/issues/737) and [window controls overlay](https://github.com/WICG/window-controls-overlay/blob/main/explainer.md)), and those discussions should take place in the Web Apps WG, around the Manifest spec. We should not be discussing proposals about new Web App functionality in the CSSWG.

mgiuca:
This text was moved out of the Manifest spec into CSS mediaqueries-5 in
w3c/csswg-drafts#6343, along with the display-mode media feature. The
actual definition of display mode belongs in Manifest, while the
display-mode media feature remains here.

Added some more text explaining how the display-mode media feature works
given that this is in a separate spec to where web apps are defined.

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -3372,7 +3372,7 @@ the following changes and additions were made to this module since the
 <a href="https://www.w3.org/TR/2021/WD-mediaqueries-5-20211218/">2021-12-18 Working Draft</a>:
 
 * Moved 'display mode' definition back to [[APPMANIFEST]] ('display-mode' media feature remains
-  here). (See <a href="https://github.com/w3c/csswg-drafts/issues/7306">Issue 7306</a>)
+	here). (See <a href="https://github.com/w3c/csswg-drafts/issues/7306">Issue 7306</a>)
 * Establish a normative reference for [[Display-P3]]
 * Disallow use of ''layer'' as a media type, rather than merely treat it as an unknown one, for compatibility with [=cascade layers=].
 * Clarify intent of 'prefers-reduced-motion'
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1343,6 +1343,16 @@ Display Modes: the ''display-mode'' media feature </h3>
 			that are not fullscreen.
 
 			Corresponds to the [=display mode/browser=] display mode.
+
+		<dt><dfn>picture-in-picture</dfn>
+		<dd>
+			This mode allows users to continue consuming media while they interact
+			with other sites or applications on their device.
+			The web application is displayed in a floating and always-on-top window.
+			A user agent may include other platform specific UI elements,
+			such as "back-to-tab" and "site information" buttons
+			or whatever is customary on the platform and user agent.
+
 	</dl>
 
 	<div class="example">
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1330,7 +1330,7 @@ Display Modes: the ''display-mode'' media feature </h3>
 			The [=display mode/browser=] display mode.
 	</dl>
 
-	This media feature reflects the actual display mode that is being used on the current browsing
+	This media feature reflects the actual [=display mode=] that is being used on the current browsing
 	context, not necessarily the one that was requested in the web app manifest (if any).
 
 	On normal web pages, 'display-mode' will have a value of ''browser''. Only pages appearing in the
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1333,9 +1333,10 @@ Display Modes: the ''display-mode'' media feature </h3>
 	This media feature reflects the actual [=display mode=] that is being used on the current browsing
 	context, not necessarily the one that was requested in the web app manifest (if any).
 
-	On normal web pages, 'display-mode' will have a value of ''browser''. Only pages appearing in the
-	context of an [=installed web application=] can have a different 'display-mode', reflecting the
-	way in which the application window is being presented to the end user.
+	Note: When a page is being displayed in a normal web browser, 'display-mode' will always have a
+	value of ''browser''. Only pages appearing in the context of an [=installed web application=] will
+	have a different 'display-mode', reflecting the way in which the application window is being
+	presented to the end user.
 
 	<details class="note">
 		<summary>The [=display mode/fullscreen=] [=display mode=] is distinct from the [[FULLSCREEN|Fullscreen API]].</summary>
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1306,40 +1306,67 @@ Display Modes: the ''display-mode'' media feature </h3>
 	Type: discrete
 	</pre>
 
-	The 'display-mode' media feature represents the [=display mode=] of the current [=browsing context=].
-	Child browsing contexts reflect the [=display mode=] of their [=top-level browsing context=].
+	The '@media/display-mode' media feature describes the mode in which the current [=browsing context=] is
+	currently being presented to the end user.  In child browsing contexts, the [=display mode=] must
+	match that of the [=top-level browsing context=].
 
-	The values of this feature correspond to the [=display mode|display modes=] defined in
-	[[APPMANIFEST]]:
+	This feature is primarily used to determine which [=display mode=] the user agent has applied to
+	an [=application context=]. As such, the values of this feature correspond to the [=display
+	mode|display modes=] defined in [[APPMANIFEST]]. However, it can also be used in non-application
+	contexts to determine whether the viewport is in fullscreen mode.
+
 
 	<dl dfn-type=value dfn-for="@media/display-mode">
 		<dt><dfn>fullscreen</dfn>
 		<dd>
-			The [=display mode/fullscreen=] display mode.
+			The browsing context is displayed with browser UI elements hidden and takes up the entirety of
+			the available display area. The fullscreen context may have been caused by the [=display
+			mode/fullscreen=] display mode in the [=application manifest=], by the {{requestFullscreen()}}
+			method of the [[FULLSCREEN|Fullscreen API]], or through some other means (such as the user
+			manually activating fullscreen mode using the user agent's built-in controls).
+
+			Corresponds to the [=display mode/fullscreen=] display mode.
 
 		<dt><dfn>standalone</dfn>
 		<dd>
-			The [=display mode/standalone=] display mode.
+			The [=display mode/standalone=] display mode is in use.
 
 		<dt><dfn>minimal-ui</dfn>
 		<dd>
-			The [=display mode/minimal-ui=] display mode.
+			The [=display mode/minimal-ui=] display mode is in use.
 
 		<dt><dfn>browser</dfn>
 		<dd>
-			The [=display mode/browser=] display mode.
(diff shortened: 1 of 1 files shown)


aarongustafson:
The Manifest spec is aligned and can be merged at the same time.

marcoscaceres:
It would be good if we then linked the display-mode values back to the Web Manifest spec (so people can easily look up what each one means). We've updated the other PR to export them.

mgiuca:
Good idea. Is this too verbose? (Explicitly listing each one and linking it to the manifest) This way there is a separate explicit definition of "the standalone display-mode MQ feature" (defined in CSS), distinct from the definition of "the standalone display mode" (defined in Manifest).

Also, I can't really verify that this linking will work... right now when I build it, it links to the CSS spec, but I assume once both of these PRs land, those links will go back to the Manifest spec.

mgiuca:
@frivoal for review, please when you get a chance.

aarongustafson:
In `ImageResource` we just reference the `sizes` attribute in HTML and don't get in to acceptable values and such. I'd leave that the referenced spec to define.

mgiuca:
> In `ImageResource` we just reference the `sizes` attribute in HTML and don't get in to acceptable values and such. I'd leave that the referenced spec to define.

I'm not quite clear what you're proposing? Are you saying I should revert the last commit (disagree with Marcos)? Or some middle ground.

FWIW the intention here is that we're defining two different things: the "display modes" which are the modes you can request in the Manifest, and the "`display-mode` media query values" which are syntactic values that appear in CSS. We should explicitly define the meaning of each of those syntactic words to be the corresponding word in Manifest.

mediaqueries-5/Overview.bs
```diff
@@ -1305,94 +1305,73 @@ Display Modes: the ''display-mode'' media feature </h3>
...
-			The web application is displayed using the platform-specific convention
-			for opening hyperlinks in the user agent
-			(e.g., in a browser tab or a new window).
+			The [=display mode/browser=] display mode.
 	</dl>
 
+	This media feature reflects the actual display mode that is being used on the current browsing
+	context, not necessarily the one that was requested in the web app manifest (if any).
+
+	On normal web pages, 'display-mode' will have a value of ''browser''. Only pages appearing in the
```
marcoscaceres, line 1335: I'm not sure we should say anything about "normal web pages" (unless that's actually a thing?).

mediaqueries-5/Overview.bs
```diff
@@ -1305,94 +1305,73 @@ Display Modes: the ''display-mode'' media feature </h3>
...
 
 		<dt><dfn>browser</dfn>
 		<dd>
-			The web application is displayed using the platform-specific convention
-			for opening hyperlinks in the user agent
-			(e.g., in a browser tab or a new window).
+			The [=display mode/browser=] display mode.
 	</dl>
 
+	This media feature reflects the actual display mode that is being used on the current browsing
```
marcoscaceres, line 1332:
```suggestion
	This media feature reflects the actual [=display mode=] that is being [=applied=] on the current browsing
```

aarongustafson:
> I'm not quite clear what you're proposing? Are you saying I should revert the last commit (disagree with Marcos)? Or some middle ground.
> 
> FWIW the intention here is that we're defining two different things: the "display modes" which are the modes you can request in the Manifest, and the "`display-mode` media query values" which are syntactic values that appear in CSS. We should explicitly define the meaning of each of those syntactic words to be the corresponding word in Manifest.

Sorry, I should have been more detailed. What I was saying was that I'm not sure the values are as necessary in the CSS spec as they should be bound to the values in the Manifest, right? For instance, when `display_override` lands, I’d expect the value that is ultimately applied in the browser would be the one popped through to the media query. When that lands, we wouldn't need to update the Media Query spec to include it if the spec references the acceptable values in the Manifest.

mediaqueries-5/Overview.bs
mgiuca: Yeah I wasn't really sure how to express this, but I am trying to simply say (since we're in the context of CSS and not the manifest) that "normally" (on the "drive-by web") this will be "browser".

I've reworded it and put it in a note. The non-normative context I think allows me to talk about "web browsers" as a special class of user agent, which I wouldn't normally talk about in normative text.

mgiuca:
> Sorry, I should have been more detailed. What I was saying was that I'm not sure the values are as necessary in the CSS spec as they should be bound to the values in the Manifest, right? For instance, when `display_override` lands, I’d expect the value that is ultimately applied in the browser would be the one popped through to the media query. When that lands, we wouldn't need to update the Media Query spec to include it if the spec references the acceptable values in the Manifest.

So what you're saying is that we shouldn't duplicate the list of possible values, and instead just say "it can be any value from the "display mode" list in the manifest"?

I think avoiding that duplication would be ideal, but my feeling from reading the CSSMQ spec is that this is formally defining the CSS syntax and we can't just delete the blue box and put in a hand-wave "see the list from the manifest". So we do need to duplicate the list of possible values. I think it's OK: adding new display modes will be pretty rare (we'll add maybe 2 in the next couple of years) so we will just be sure to add an entry here when we do.

mgiuca:
Hi CSS folks, including @frivoal .

I have _finally_ gotten around to updating this PR with the changes discussed around this time last year. Apologies for the slow turnaround on this - this was disrupted at the start of this year and been on my backburner since then.

I believe my edits have captured the essence of what was requested in the [CSSWG meeting from 17 Nov, 2022](https://github.com/w3c/csswg-drafts/issues/7306#issuecomment-1317378222).

A brief recap:

- In December 2021, the definition of "display mode" and its corresponding values was moved from the Manifest spec to the Media Queries spec. This is because it was defined as both a JSON value in the Manifest and a media query feature.
- I argued that the definition should be in the Manifest spec, as the display mode is a core concept (or rather, _the_ core concept) defined by the web app manifest, and the media query was just a convenience feature for developers. I proposed two PRs: this one, and the [corresponding one in the Manifest spec to put it back](https://github.com/w3c/manifest/pull/1039).
- In November 2022, the CSSWG agreed in principle to move the definition back, but argued that because "fullscreen" is triggered by not only the manifest `"display": "fullscreen"` but also by any other type of fullscreen, it isn't proper to have that display mode defined in Manifest when the media query works independent of the manifest.

So for today's update to the PR, I have:

1. Largely rewritten the PR, so that rather than saying "this MQ reflects the display mode from the Manifest", it's more neutral: it says that it reflects the way the page is being presented to the user, with the Manifest display mode being one of the ways it can be modified, various ways of entering fullscreen being another, and "browser" being the default.
2. Added an explicit definition of "browser" and "fullscreen" modes here in CSS. The "fullscreen" one is neutral as to how it can be triggered (manifest, `requestFullscreen` or user control). The other two modes are defined in terms of the Manifest spec, as they can only be entered from an application context.
3. Added an example.

There are two main aspects that I wish to keep in the Manifest spec:

- The definition of the term "display mode" itself is in the Manifest spec, since that is the core concept which that spec revolves around.
- There is no prose describing any display modes other than "browser" and "fullscreen" (the two modes which can be triggered outside of an application context). This is for future proofing: we intend to add more display modes going forward (two new modes, `window-controls-overlay` and `tabbed` are explicitly planned), and we want the process of adding new display modes to be that you define their meaning in full in the Manifest spec, and then simply add their name to the Media-Queries spec, linking back to the Manifest definition, and not duplicating any prose.

Please take a look and let me know if this is going to be discussed at a CSSWG meeting.

Cheers.

mgiuca:
Hi @frivoal . Friendly ping, would you be able to review the revised pull request or re-raise it at a WG meeting, whatever the process may be? I believe the last update satisfied your feedback.

mgiuca:
Oh no, I just saw the merge conflict. #9920 has added a new value to CSS `display-mode` (but not Manifest `display-mode`). This makes disentangling the two more complex.

I will take another pass at this now.

mgiuca:
Updated. I have reworked things so that we now have a separation between the "display mode" in the manifest (which defines the four display modes) and the values of the "display-mode" MQ (which now has five possible values).

The text in this PR is now clear on the fact that `display-mode` _can_ represent the display mode for an application context (like `standalone` or `minimal-ui`), or it can also reflect other non-application display modes (like `fullscreen` or `picture-in-picture`).

This is a bit trickier now that there isn't a 1:1 mapping between "manifest display modes" and "MQ display modes" but I think this text makes sense. I am similarly making small non-normative changes on the Manifest side so that it's clearer.

@frivoal @tabatkins @beaufortfrancois FYI

marcoscaceres:
@frivoal, @mgiuca, can this be merged yet? What's the latest?

grorg:
Landed this as shown above.

Issue #5922, opened 2021-02-03
[css-color-4] Matrices for sRGB to XYZ and back were (slightly) wrong


svgeesus:
TLDR; very small change, no visible effect, but imprecise round-trip was annoying. Non-normative sample code updated. Paywalled standards are bad and lead to errors.

**What was wrong**

The [normative definition of sRGB in CSS Color 4](https://drafts.csswg.org/css-color-4/#valdef-color-srgb) is correct. In the [non-normative sample code](https://drafts.csswg.org/css-color-4/#color-conversion-code), the matrix to go from linear-light sRGB to CIE XYZ, and the inverse matrix to go from CIE XYZ to linear-light sRGB, was wrong starting at the fourth decimal place.

**Why**

How did this happen? The official definition of sRGB 

**IEC 61966-2-1:1999/AMD1:2003** 
Amendment 1 - Multimedia systems and equipment - Colour measurement and management - Part 2-1: Colour management - Default RGB colour space - sRGB

is a 16 page PDF which you [buy from the IEC webstore for 70 CHF](https://webstore.iec.ch/publication/6168).  I don't have a copy.

The standard is different from the [original definition of sRGB](https://www.w3.org/Graphics/Color/sRGB) because roundoff errors to 4 decimal places in the original proposal meant that the linear and curved parts of the transfer function were not continuous (although the effect made no difference, at 8 bits per component). The IEC standard fixed that.

So, like [others](https://www.image-engineering.de/library/technotes/958-how-to-convert-between-srgb-and-ciexyz) I used the conversion matrices [calculated by Bruce Lindbloom](http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html) in the sRGB sample code.

The impact of this was fairly subtle. If you converted sRGB white, `#FFFFFF` to display-p3 you got `color(display-p3 1.0000557 0.99999258 0.99990441)` which is `color(display-p3 1.000 1.000 1.000)` to four significant figures, but `color(display-p3 1.0001 1.0000 0.9999)` to five significant figures, and this should be exactly `1 1 1` because both sRGB and display-p3 use a D65 white.

This bugged me.

Converting from some other D65-using colorspace, like `color(rec2020 1 1 1)` to display-p3, gave a nice satisfying round-trippable `color(display-p3 1 1 1)`. So this imprecision was related to sRGB.

[...]

PR #7320, opened 2022-05-27
[css-color-4] Use exact values for some matrices in conversions.js

kainino0x:
These color space conversion matrices have exact rational values that
can be computed from the numbers provided in the spec. Using exact
values is more succinct for most of these matrices, and also makes it a
nice reference implementation for other languages. This example code
already uses exact inline formulations for a number of other things,
like D50 and D65 definitions, so this is similar to that.

I'm not really an expert in this, so if this doesn't make sense (e.g. the
source values aren't actually exact to begin with) let me know.
I checked that these ratios generate values very close to the current
ones, usually differing in the last few digits, but double-checking
can't hurt.

I only did the XYZ conversion matrices for srgb, display-p3, a98-rgb,
and rec2020.
- I don't have code to easily compute the D65/D50 conversions or
  OKLab/OKLCH as I was only interested in the predefined color spaces.
- The rational forms of prophoto-rgb's matrices exceed the precision of
  JavaScript math. I could include them as comments though.

Source to compute these: https://github.com/kainino0x/exact_css_xyz_matrices
using this Rust crate: https://crates.io/crates/rgb_derivation
as described for sRGB on this page: https://mina86.com/2019/srgb-xyz-matrix/
but using the numbers from this spec.

I used these in the WebGPU conformance test suite: https://github.com/gpuweb/cts/pull/1089
WebGPU needed only srgb and display-p3, but it was easy to extend to the
other predefined color spaces. (WebGPU may add some of those color
spaces eventually anyway.)


kainino0x:
Diff in decimal format, for easier verification (I have verified it):

`lin_sRGB_to_XYZ`
```diff
-		[ 0.41239079926595934, 0.357584339383878,   0.1804807884018343  ],
+		[ 0.4123907992659595,  0.35758433938387796, 0.1804807884018343  ],
-		[ 0.21263900587151027, 0.715168678767756,   0.07219231536073371 ],
+		[ 0.21263900587151036, 0.7151686787677559,  0.07219231536073371 ],
-		[ 0.01933081871559182, 0.11919477979462598, 0.9505321522496607  ]
+		[ 0.01933081871559185, 0.11919477979462599, 0.9505321522496606  ]
```

`XYZ_to_lin_sRGB`
```diff
-		[  3.2409699419045226,  -1.537383177570094,   -0.4986107602930034  ],
+		[  3.2409699419045213,  -1.5373831775700935,  -0.4986107602930033  ],
-		[ -0.9692436362808796,   1.8759675015077202,   0.04155505740717559 ],
+		[ -0.9692436362808798,   1.8759675015077206,   0.04155505740717561 ],
-		[  0.05563007969699366, -0.20397695888897652,  1.0569715142428786  ]
+		[  0.05563007969699361, -0.20397695888897657,  1.0569715142428786  ]
```

`lin_P3_to_XYZ`
```diff
-		[ 0.4865709486482162,  0.26566769316909306,  0.1982172852343625 ],
+		[ 0.48657094864821626, 0.26566769316909294,  0.1982172852343625 ],
-		[ 0.2289745640697488,  0.6917385218365064,   0.079286914093745  ],
+		[ 0.22897456406974884, 0.6917385218365062,   0.079286914093745  ],
-		[ 0.0000000000000000,  0.04511338185890264,  1.043944368900976  ]
+		[ 0,                   0.045113381858902575, 1.0439443689009757 ]
```

`XYZ_to_lin_P3`
```diff
-		[  2.493496911941425,    -0.9313836179191239, -0.40271078445071684  ],
+		[  2.4934969119414245,   -0.9313836179191236, -0.40271078445071684  ],
-		[ -0.8294889695615747,    1.7626640603183463,  0.023624685841943577 ],
+		[ -0.829488969561575,     1.7626640603183468,  0.02362468584194359  ],
-		[  0.03584583024378447,  -0.07617238926804182, 0.9568845240076872   ]
+		[  0.035845830243784335, -0.07617238926804171, 0.9568845240076873   ]
```

`lin_a98rgb_to_XYZ`
```diff
-		[ 0.5766690429101305,   0.1855582379065463,   0.1882286462349947  ],
+		[ 0.5766690429101308,   0.18555823790654627,  0.18822864623499472 ],
-		[ 0.29734497525053605,  0.6273635662554661,   0.07529145849399788 ],
+		[ 0.29734497525053616,  0.627363566255466,    0.07529145849399789 ],
-		[ 0.02703136138641234,  0.07068885253582723,  0.9913375368376388  ]
+		[ 0.027031361386412378, 0.07068885253582714,  0.9913375368376389  ]
```

`XYZ_to_lin_a98rgb`
```diff
-		[  2.0415879038107465,    -0.5650069742788596,   -0.34473135077832956 ],
+		[  2.041587903810746,     -0.5650069742788596,   -0.3447313507783295  ],
-		[ -0.9692436362808795,     1.8759675015077202,    0.04155505740717557 ],
+		[ -0.9692436362808798,     1.8759675015077206,    0.04155505740717561 ],
-		[  0.013444280632031142,  -0.11836239223101838,   1.0151749943912054  ]
+		[  0.013444280632031024,  -0.11836239223101824,   1.0151749943912054  ]
```

`lin_2020_to_XYZ`
```diff
-		[ 0.6369580483012914,  0.14461690358620832,  0.1688809751641721   ],
+		[ 0.6369580483012913,  0.14461690358620838,  0.16888097516417205  ],
-		[ 0.2627002120112671,  0.6779980715188708,   0.05930171646986196  ],
+		[ 0.26270021201126703, 0.677998071518871,    0.059301716469861945 ],
-		[ 0.000000000000000,   0.028072693049087428, 1.060985057710791    ]
+		[ 0,                   0.028072693049087508, 1.0609850577107909  ]
```

`XYZ_to_lin_2020`
```diff
-		[ 1.7166511879712674,   -0.35567078377639233, -0.25336628137365974 ],
+		[ 1.7166511879712676,   -0.3556707837763924,  -0.2533662813736598  ],
-		[ -0.6666843518324892,   1.6164812366349395,   0.01576854581391113 ],
+		[ -0.666684351832489,    1.616481236634939,    0.01576854581391113 ],
-		[ 0.017639857445310783, -0.042770613257808524, 0.9421031212354738  ]
+		[ 0.017639857445310915, -0.042770613257808655, 0.942103121235474   ]
```

svgeesus:
> These color space conversion matrices have exact rational values that
can be computed from the numbers provided in the spec. Using exact
values is more succinct for most of these matrices, and also makes it a
nice reference implementation for other languages. 

In general using exact rational values is a better approach, agreed.

I am curious where the  [D65 values 0.312713, 0.329016](https://mina86.com/2019/srgb-xyz-matrix/) come from.

About a year ago there was a strong effort to recalculate everything using a [consistent white point](https://drafts.csswg.org/css-color-4/#white-point) everywhere, which reduced residual errors substantially. I am reluctant to redo that without further details.

kainino0x:
> _Edit:_ Although I see [your rust code](https://github.com/kainino0x/exact_css_xyz_matrices/blob/d8171c7f529f8b48ad407c0a85546751add247aa/result.txt#L2) uses the same four-digit D65 values as CSS Color 4.

Yes, I took all of the numbers including D50/D65 from this spec, and then verified that that had the same results (because they had also been used to compute the original matrices).

svgeesus (approved):

ValZapod:
What is this. They must stay float, as defined by the IEC standard.

kainino0x:
I'll admit I'm not at all familiar with these specs. However, this change is only to the sample JavaScript code, and it only writes the same exact floating point numbers in a rational format rather than a decimal format. It's also easy to convert back to decimal format - just paste the code into a JavaScript interpreter.

kainino0x:
If they're incorrect because BT.709 *requires* decimal rounding at certain points of the computation, then that would have already been wrong before my change, but a legitimate concern, I think.

ValZapod:
> However, this change is only to the sample JavaScript code

This actually changes Apple webkit code that autosyncs to this, Google may also look into this:  https://github.com/WebKit/WebKit/commit/0f1dfd0dc15724e2353d460ce846bdea26ec7ab9.

> at all familiar with these specs

Familiarize yourself #7809 and please LGTM my pull request. Hope no one will DCMA standards.

>If they're incorrect because BT.709 requires decimal rounding at certain points of the computation

No rounding, it just says that one of the matrix has infinite precision and cannot be extended (infnite amount of zeroes). 0.2126, 0.7152, 0.0722 must add to 1. I updated wikipedia for to be 100% obvious.

kainino0x:
Interesting. If the values of these matrices are wrong (again, I didn't change them, so if so they were already wrong), then either the derivation is wrong, or the numbers in the CSS spec are wrong and the normative part of the spec also needs to be fixed. I'm just a drive-by contributor, so I'd advise this discussion be continued over in #5922 which is the actual relevant place. I'm happy to help re-derive the rational forms if some spec change is required, though.

PR #7337, opened 2022-06-05
[css-fonts-3] Fix the mistake that the scheme is inconsistent, and improve comments

sxyazi:
Examples of different schemas and different domains are inconsistent with the given assumed URL.

### DIFF
--- a/css-fonts-3/Overview.html
+++ b/css-fonts-3/Overview.html
@@ -3043,7 +3043,7 @@ h2 {
    CORS-enabled fetch method, is defined or required.
 
   <div class=example> For the examples given below, assume that a document is
-   located at <code>https://example.com/page.html</code> and all URL's link
+   located at <code>http://example.com/page.html</code> and all URL's link
    to valid font resources supported by the user agent. Fonts defined with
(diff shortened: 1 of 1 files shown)


svgeesus (commented):
Since example.x is unresolvable, the use of http or https is irrelevant.

I don't see the changes in the comments as being worthwhile, sorry.

sxyazi:
@svgeesus Hi, I think the inconsistent URLs in the examples would confuse the reader (at least for me), and besides it makes no sense to load HTTP resources on an HTTPS page, the security policy would prevent it. Is this aspect of rigor worth considering?

tabatkins:
This is actually legit, since the example straight up doesn't make sense right now - it's claiming an https url is "different scheme" from the main url, which is also https. It looks like someone just hit one of the urls with a regex at some point, or maybe just hypercorrected out of reflex?

Regardless, either the example base url needs to be switched back to http, or the font urls listed below it need to have their schemes swapped to match the comments.

svgeesus:
Dug into the history a bit and yes, originally example.com/page.html was http not https.

Agree that changing this back to http, as the PR does, is the simplest.

svgeesus (approved):

PR #7396, opened 2022-06-21
[css-color-4] Improve 'color' property description.

fantasai:
* Clarify that currentColor resolves the used value.
* Use more standard terminology for describing its effect on text fill color.
* Describe what the value of 'color' does, not its syntax, in its value dfn.
* Shift non-normative text into example.
* Improve quotation marks.
* Adjust punctuation and minor tweaks to wording of note.
* Make the note normative, because it describes a normative effect


svgeesus (approved):
Thanks, nice.

PR #7414, opened 2022-06-24
[css-easing-2] Adding multi-step linear-interpolated easing function

jakearchibald:
This PR defines `linear-spline()`, a way to create custom easings, as discussed in https://github.com/w3c/csswg-drafts/issues/229#issuecomment-861247074.

TODO:

- [x] Plug in "create a linear easing function"
- [ ] Create diagrams to explain the feature
- [ ] Add a note at the start, explaining the feature for developers

### DIFF
--- a/css-easing-2/Overview.bs
+++ b/css-easing-2/Overview.bs
@@ -304,9 +304,11 @@ It returns a [=linear spline easing function=].
 
                1.   Set |extraSerializablePoint|'s [=linear spline easing point/input=] to <<linear-spline-stop-length>>'s second <<percentage>> as a number.
 
-     1.   Otherwise, if |stop| is the first [=list/item=] in |stopList|,
+     1.   Otherwise, if |stop| is the first [=list/item=] in |stopList|, then:
 
-          then set |point|'s [=linear spline easing point/input=] to 0.
+          1.   Set |point|'s [=linear spline easing point/input=] to 0.
+
+          1.   Set |largestInput| to |point|'s [=linear spline easing point/input=].
 
      1.   Otherwise, if |stop| is the last [=list/item=] in |stopList|,
           then set |point|'s [=linear spline easing point/input=] to whichever is greater:
--- a/css-easing-2/Overview.bs
+++ b/css-easing-2/Overview.bs
@@ -7,10 +7,12 @@ Level: 2
 Group: csswg
 ED: https://drafts.csswg.org/css-easing/
 TR: https://www.w3.org/TR/css-easing-2/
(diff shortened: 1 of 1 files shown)


okikio:
image

jakearchibald:
@dshin-moz I've corrected the issues we discussed in the previous PR, and wrote a more solid algorithm for calculating the output for a given input (although you might be ahead of me there).

I've defined how the computed value works, matching gradients, but it seems silly now I've written it down. I'll open an issue to see if we can go for something similar.

jakearchibald:
https://github.com/w3c/csswg-drafts/issues/7415 - to discuss the computed value

dshin-moz:
Since we aren't overloading `linear` anymore, I presume it is no longer the case that `linear-spline()` is a shorthand of `linear`?

jakearchibald:
Yep. Although I'm open to bikeshedding.

jakearchibald:
I guess I need to define that in the parsing

dshin-moz:
> Yep. Although I'm open to bikeshedding.

`linear-spline()` != `linear`, I think that makes sense (A linear spline with 0 entries can't suddenly become `f(x) = x`)

jakearchibald:
Yeah, I kinda liked reusing `linear`, but everyone I told about it was "wtf??" until I explained it was still linear, but between points, and then they were "ohh, ok". So, I don't think it's worth the initial confusion.

birtles:
Yeah, my experience is that short and memorable is better than semantically correct (hence why we should have used `animation-easing` instead of `animation-timing-function`). People learn to associate the slightly shifted meaning very quickly (a kind of semantic bleaching I guess).

I thought using `linear()` was great and `spline` makes me think of curves but that's just me.

jakearchibald:
I'll give it some thought over the weekend. Maybe I changed it too hastily.

jakearchibald:
Opened https://github.com/w3c/csswg-drafts/issues/7419 for the name bikeshedding

jakearchibald:
Seems increasingly likely this will go back to `linear()` https://github.com/w3c/csswg-drafts/issues/7419. I'm about to go on leave for a week, but I'll make the call when I'm back, and we can land this.

plinss:
@jakearchibald Can you please move this PR into a fork of the repo, rather than a branch in the main repo? 

Having branches puts an additional load on our draft server (and when there are history changes on the branch it plays havoc with the git to hg conversion currently used by that server).

(There are improvements planned for the draft server that will make this less of an issue, but we're not there just yet.)

jakearchibald:
https://github.com/w3c/csswg-drafts/pull/7414#issuecomment-1176381134

Issue #6212, opened 2021-04-14
[css-overflow] [css2?] Baseline of an inline-block with overflow:clip.


emilio:
Blink and Gecko differ on what baseline should an element with `display: inline-block` and `overflow: clip`.

Example ([live](http://software.hixie.ch/utilities/js/live-dom-viewer/saved/9184)):

```html
<!doctype html>

Clip:
<br>
<div style="border:1px solid red; display:inline-block">
  <div style="overflow:clip; width:10px; height:10px; background:blue; display:inline-block;">Hallo Welt</div>
</div>

<br>
Visible:
<br>
<div style="border:1px solid red; display:inline-block">
  <div style="width:10px; height:10px; background:blue; display:inline-block;">Hallo Welt</div>
</div>
<br>
Hidden:
<br>
<div style="border:1px solid red; display:inline-block">
  <div style="overflow:hidden; width:10px; height:10px; background:blue; display:inline-block;">Hallo Welt</div>
</div>
```

@bfgeek pointed out that CSS2.1 has:

>  The baseline of an 'inline-block' is the baseline of its last line box in the normal flow, unless it has either no in-flow line boxes or if its 'overflow' property has a computed value other than 'visible', in which case the baseline is the bottom margin edge.

However CSS2 was written before `overflow: clip` was a thing. I intuitively expect `overflow: clip` to behave like `overflow: visible` here (because it's not an scrollable box), but I could see the argument the other way around. Thoughts?

emilio:
An interesting question, if we decide that `clip` and `visible` should behave differently, is what happens if only one of the axes is clipped.

bfgeek:
I somewhat expected blink's behaviour (and have a mild preference for it) as I sort of view `clip` as a better `hidden`, but either way is fine.

Loirooriol:
> what happens if only one of the axes is clipped

Seems a good argument for treating `clip` like `visible`.

tabatkins:
Yeah, since only one axis can be clipped, this really should continue to be treated like `visible`, not like `hidden`. Diverging from that principle will only cause us confusion in the future, I think.

And yes, older text saying "overflow other than visible" just needs to be fixed up, it should always *mean* "scroll container" in modern parlance.

MatsPalmgren:
[...]

PR #7436, opened 2022-06-30
Audit specifications for correct handling of overflow:clip.

dbaron:
This set of changes is the result of auditing all specification
Overview.bs other than CSS2 and css-overflow-* for correct handling of
overflow:clip, since many specifications have historically referenced
concepts like "overflow values other than visible".

It is related to #6212, but doesn't actually fix it, because it's really
only a bug in CSS2.

### DIFF
--- a/css-ui-3/Overview.bs
+++ b/css-ui-3/Overview.bs
@@ -528,7 +528,7 @@ and if so, along which axis/axes.
 Name: resize
 Value: none | both | horizontal | vertical
 Initial: none
-Applies to: elements with 'overflow' other than visible,
+Applies to: elements that are <a>scroll containers</a>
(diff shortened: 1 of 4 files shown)


shawky13 (approved):

shawky13 (approved):

css-ui-3/Overview.bs
```diff
@@ -664,6 +663,7 @@ its <a>end</a> line box edge
 in the inline progression direction of its block container element ("the block")
 that has 'overflow'
 other than ''visible''.
+<!-- REVIEW: I *think* this should not include ''overflow/clip'' -->
```
tabatkins, line 666: Agreed.

tabatkins:
r+ for backgrounds and UI

Issue #5614, opened 2020-10-13
[cssom-view] Separate getBoundingClientRect algorithm from the public API


szager-chromium:
The IntersectionObserver spec refers to "the getBoundingClientRect() algorithm", which isn't actually a thing. We need to split out the algorithm from the public API, so that other specs can refer to the algorithm.

https://github.com/w3c/IntersectionObserver/issues/55

zcorpan:
#7446 fixed this.

PR #7446, opened 2022-07-01
[cssom-view] Add a "get the bounding box" algorithm for getBoundingClientRect() #5614 

miketaylr:
This way other specs like IntersectionObserver can call it.

See https://github.com/w3c/IntersectionObserver/issues/55


miketaylr:
PTAL @emilio

tabatkins (approved):

tabatkins:
I'll let Emilio make the final say as I'm not an editor, just a frequent tweaker, but this looks fine to me.

emilio (approved):

PR #7471, opened 2022-07-07
[web-animations-2] Update to editor spec

kevers-google:
[web-animations-2] Update to editor spec

The patch converts web-animations-2 from an unofficial draft to an editors' draft.  In the process, converted from a delta spec to a full spec since web-animations-1 is sufficiently mature and few additional edits to the v1 spec are expected.  All lint warnings have been resolved.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -1,16 +1,46 @@
 <style>
-.constructors::before, .members::before, .attributes::before {
+.informative-bg {
+  margin: 1.5em 0 1em;
+  padding: 1em;
+  margin-top: 1em;
+  background: #efe;
+  border: green 1px dotted;
+}
+@media (prefers-color-scheme: dark) {
+  .informative-bg {
+    background: rgba(255, 255, 255, .05);
+  }
+}
+
+div.informative-bg *:last-child {
+  margin-bottom: 0;
+}
+
+div.informative-bg p:first-child {
+  margin-top: 0;
+}
+
+div.informative-bg h2,
+div.informative-bg h3,
+div.informative-bg h4 {
+  background: none;
+}
+
+.attributes::before, .methods::before,
+.parameters::before, .exceptions::before,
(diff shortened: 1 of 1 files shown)


birtles:
Thanks for doing this. I'm not entirely sure we're ready to convert from a delta spec yet. We still have a few more outstanding issues on Web Animations and it would be good to see it get to CR first?

flackr:
> We still have a few more outstanding issues on Web Animations and it would be good to see it get to CR first?

Specifically the list of open `label: web-animations-1` issues? i.e. https://github.com/w3c/csswg-drafts/issues?q=is%3Aissue+is%3Aopen+label%3Aweb-animations-1+

Skimming the list, I suspect the resolution of most of these would be relatively minor edits to apply to both web-animations-1 and web-animations-2, though I'm not too familiar with how often this is done for other specs.

kevers-google:
I can look into putting up a fresh PR for an editor's draft that remains a delta spec.  This should go quicker than the full merge.  Once web-animations-2 is ready for a merge, we can look into whether it will be quicker to restart the merge process or walk through the edits and copy them over to this PR.

birtles:
> > We still have a few more outstanding issues on Web Animations and it would be good to see it get to CR first?
> 
> Specifically the list of open `label: web-animations-1` issues? i.e. https://github.com/w3c/csswg-drafts/issues?q=is%3Aissue+is%3Aopen+label%3Aweb-animations-1+

Yes, along with #5394 (which I've just updated the label for now).

> Skimming the list, I suspect the resolution of most of these would be relatively minor edits to apply to both web-animations-1 and web-animations-2, though I'm not too familiar with how often this is done for other specs.

That's true. I've noticed that recently there have been edits to web-animations-1 too by non-editors which were just pushed without a PR so we'll need to make sure others either update both specs or, better, submit PRs so we can check.

birtles:
@kevers-google Can we close this now? I think Web Animations 2 has already been published as a delta ED spec?

PR #7477, opened 2022-07-07
Editorial Refactor of <color> Definition Section

fantasai:
Working my way through the spec top-to-bottom as a new reader, so, I notice these things...

* Split out conceptual definitions from syntactic definitions, pulling the former up into the main section intro.
* Define “color function” properly as a term.
*  Give an actual overview of all the color syntaxes in the intro, so readers have a map of what's coming.
* Push legacy syntax into its own subsection.
* Factor `<alpha-value>` into its own section, rather than as a paragraph of the rgb() section.
* Shift non-normative text about hue angles into a note.

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -760,6 +760,29 @@ Legacy Comma-separated Color Function Syntax</h4>
 	where there is no Web compatibility issue,
 	the <a>legacy color syntax</a> is invalid.
 
+<h3 id="alpha-syntax">
+Representing Transparency: the <<alpha-value>> syntax</h3>
+
+	<pre class="prod">
+	<dfn>&lt;alpha-value></dfn> = <<number>> | <<percentage>>
+	</pre>
+
+	Opacity in CSS is typically represented using the <<alpha-value>> syntax,
+	for example in the 'opacity' property
+	or as the [=alpha component=] in a [=color function=].
+	Represented as a <<number>>, the useful range of the value is ''0''
+	(representing full transparency)
+	to ''1''
+	(representing full opacity).
+	It can also be written as a <<percentage>>,
+	which [=computed value|computes to=]
+	the equivalent <<number>>
+	(''0%'' to ''0'', ''100%'' to ''1'').
+	Unless otherwise specified,
+	an <<alpha-value>> component defaults to ''100%'' when omitted.
+	Values outside the range [0,1] are not invalid,
+	but are clamped to that range when [=computed value|computed=].
+
 <h3 id="hue-syntax">
 Representing Cylindrical-coordinate Hues: the <<hue>> syntax</h3>
 
@@ -1685,7 +1708,6 @@ The RGB functions: ''rgb()'' and ''rgba()''</h3>
 	<pre class='prod'>
 	<dfn>rgb()</dfn> = rgb( [<<percentage>> | none]{3} [ / [<<alpha-value>> | none] ]? ) |
 	  rgb( [<<number>> | none]{3} [ / [<<alpha-value>> | none] ]? )
-	<dfn>&lt;alpha-value></dfn> = <<number>> | <<percentage>>
 	</pre>
 
 	<wpt>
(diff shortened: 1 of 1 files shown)


fantasai:
Working my way through the spec as a new reader, so, I notice these things...

fantasai:
A couple things to be aware of:
- Tried my best, but I'm not entirely sure I got the terminology exactly right.
- I'm not sure this is the right place to put the `<alpha-value>` definition, so maybe you want it somewhere else.

css-color-4/Overview.bs
```diff
@@ -614,89 +614,151 @@ Transparency: the 'opacity' property</h2>
...
+		by hue, whiteness, and blackness
+		using the [[#the-hwb-notation|HWB]] cylindrical coordinate model.
+	* ''lab()'',
+		which specifies a CIELAB color
+		by CIE Lightness and its A- and B-axis hue coordinates
+		(red/green-ness, and yellow/blue-ness)
+		using the [[#cie-lab|CIE LAB rectangular coordinate model]].
+	* ''lch()'' ,
+		which specifies a CIELAB color
+		by CIE lightness, chroma, and hue
```
svgeesus, line 670: Lightness, Chroma and hue

css-color-4/Overview.bs
```diff
@@ -614,89 +614,151 @@ Transparency: the 'opacity' property</h2>
...
+	* ''oklab()'',
+		which specifies an OKLAB color
+		by OKLAB lightness and its A- and B-axis hue coordinates
+		(red/green-ness, and yellow/blue-ness)
+		using the [[#ok-lab|OKLAB]] rectangular coordinate model.
+	* ''oklch()'' ,
+		which specifies an OKLAB color
+		by OKLAB lightness, chroma, and hue
+		using the [[#ok-lab|OKLCH]] cylindrical coordinate model.
+	* ''device-cmyk()'',
```
svgeesus, line 681: we don't have device-cmyk in Color 4 any more, moved to Color 5

css-color-4/Overview.bs
```diff
@@ -614,89 +614,151 @@ Transparency: the 'opacity' property</h2>
...
+	the other [=color functions=] use [=rectangular orthogonal color=] representations.
 
-	For easy reference in other specifications,
-	<dfn export>opaque black</dfn> is defined as the color <nobr>''rgb(0 0 0 / 100%)''</nobr>;
-	<dfn export>transparent black</dfn> is the same color,
-	but fully transparent--
-	i.e. <nobr>''rgb(0 0 0 / 0%)''</nobr>.
-
-	All of the <<color>> syntactic forms
+	All of the <<absolute-color-function>> syntactic forms
```
svgeesus, line 720: Nice

css-color-4/Overview.bs
```diff
@@ -614,89 +614,151 @@ Transparency: the 'opacity' property</h2>
...
+
+<h4 id="color-syntax-legacy">
+Legacy Comma-separated Color Function Syntax</h4>
+
 	For Web compatibility,
-	<<color>> syntactic forms
-	defined in earlier levels of this specification
-	also support
+	the syntactic forms
+	of ''rgb()'', ''rgba()'', ''hsl()'', ''hsla()'', and ''device-cmyk()''
```
svgeesus, line 738: device-cmyk is not in Color 4 (and is not legacy, was not defined in Color 3)

svgeesus (changes requested):

svgeesus (changes requested):
Yes, better to have alpha in its own section rather than hanging off one of the spaces that happens to use it

svgeesus:
Thanks, this is overall a nice improvement. A few small requests for changes.

css-color-4/Overview.bs
fantasai: It's legacy because it's implemented with commas in both PrinceXML and Antenna House, and possibly some other CSS-to-PDF implementations, so there's a compat issue there. (It was specced originally in GCPM, which didn't get very far on the REC track, but there were a number of features that were implemented and deployed in the print world.)

Commented out for now, but, I think we need to keep it as part of this section once it's re-introduced.

fantasai:
> A few small requests for changes.

@svgeesus Done!

css-color-4/Overview.bs
svgeesus: I believe they called it `cmyk()` though

svgeesus (approved):
Looks good!

css-color-4/Overview.bs
svgeesus: Added https://github.com/w3c/csswg-drafts/commit/475eb97836cc101b9c569e5c218caffd5e778e4e

PR #7479, opened 2022-07-08
[cssom-view-1] Fix dfns for events that target VisualViewport

tidoust:
The `resize` and `scroll` event definitions need to be completed to account for the newly introduced `VisualViewport` target.

This update also adjusts the firing phrasing related to `VisualViewport` for consistency with the rest of the spec (other alignments are covered in #7466) and fixes a typo.


bokand:
lgtm, thanks!

emilio (approved):
Thanks!

Issue #7240, opened 2022-04-29
[scroll-animations-1] Rethinking timeline-phase


kevers-google:
At present, timelines have a before, active, and after phase, which can be confused with animation-effect phases.  The before and after timeline-phases are exclusive to scroll-timelines. We could consider restricting the phases to active and inactive and simply allow timeline.currentTime to be outside the bounds of 0 to 100%.  Doing so, may also alleviate some of the special casing that is performed when computing the phase of an animation-effect.  A negative value for current time, regardless of units, would naturally result in the animation-effect phase being 'before' (assuming zero delay).  Conversely, a value after effect end (100% for a progress-based animation) would result in an animation-effect's phase being 'after'.  The special casing can be limited to what happens at the effect boundary, where it is conditionally exclusive with a time-based timeline and inclusive with a scroll-timeline.

flackr:
Thanks for filing this! The timeline phase has added considerable complexity (and a somewhat confusing relation to animation phase) to the spec. In particular it needs to be involved in the animation phase calculations and properly held when we pause animations.

I agree that we can dramatically simplify the state logic in the spec by simply allowing time values to extend beyond the 0 to 100% range. As an additional consequence / feature, this would allow using a negative start / end delay to extend an animation beyond the normal range of a an intersection based timeline (e.g. ViewTimeline).

@birtles is there any reason having time values go outside of this range would be problematic?

flackr:
I believe #2066 was the prior discussion that led to adding timeline phases. It's not clear to me that at the time we considered and rejected simply having the timeline extend beyond its scrollable range. The inactive timeline phase is special, however it also produces an unresolved currentTime so it will also not apply animated effects: https://drafts.csswg.org/web-animations-1/#timeline-timeline-inactive-phase

birtles:
> @birtles is there any reason having time values go outside of this range would be problematic?

I'm afraid I can't remember the full context of this discussion but I can't think of any particular issues at this point.

flackr:
It looks like `phase` was added to the [AnimationTimeline interface](https://www.w3.org/TR/web-animations-1/#the-animationtimeline-interface) but no browser currently exposes it ([demo](https://jsbin.com/diqirib/edit?js,output)).

I propose we remove `phase` from the interface, and remove the before and after phases (which were only used by scroll timelines) replacing them by allowing the current time to progress beyond 0 or 100% which would implicitly put the animation in the correct phase.

fantasai:
[...]

PR #7487, opened 2022-07-11
[web-animations-2] Clean build warnings and remove obsolete hold phase

kevers-google:
[web-animations-2]: Cleanup all build warnings in prep for transition to editor's draft.  As before and after phases for timelines have been deprecated, these are also removed as part of the cleanup.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -53,35 +53,49 @@ Ignored Vars: auto-rewind, current direction, index, initial progress,
 </pre>
 <pre class="anchors">
 urlPrefix: https://drafts.csswg.org/web-animations-1/; type: dfn; spec: web-animations-1
+    text: active-after boundary time
     text: active interval
     text: active phase
     text: active time
     text: after phase
     text: animation
+    text: animation direction
     text: animation playback rate
     text: animation effect
+    text: Apply any pending playback rate
+    text: associated effect end
+    text: before-active boundary time
     text: before flag
     text: before phase
     text: cancel an animation
     text: current iteration
+    text: current ready promise
     text: current time
     text: current finished promise
     text: directed progress
+    text: effective playback rate
     text: effect value
     text: end delay
+    text: end time
     text: fill mode
     text: finished play state
+    text: idle play state
     text: in effect
     text: iteration count
     text: iteration duration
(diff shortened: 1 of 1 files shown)


birtles:
Thanks for doing this. Just checking, but is the hold phase part of the timeline phase removal?

And the addition of the _unlimited current time_ to the definition of **at progress timeline boundary** seems new. Is this intended?

kevers-google:
With Timeline before/after phase being removed from web-animations-1, the hold phase is irrelevant.  Rather than making animation.currentTime unresolved when in the timeline before/after phase, we now allow timeline.currentTime to extend beyond the range of 0 to 100%, and let the animation effect phase take care of the rest.  This plays nicely with animation delays and is a simpler solution. 

Issue tracked here: https://github.com/w3c/csswg-drafts/issues/7240

The "unlimited current time" for the "at progress timeline boundary" calculation is new and intentional.  the problem is that a finished animation has a start time and a hold time so the current time is clamped at end time and it appears that the timeline is at the boundary when it fact the in the after phase.  This problem was previously mitigated by considering the timeline phase, but once the timeline phase is removed this safeguard becomes necessary.

I'm still not overly keen on the method for detecting if at a boundary.  The current method is trying to be overly clever and accommodate the playback rate.  It might simply be better to detect if the timeline is at 0 or 100% and leave it at that.  If the playback rate is not 1, then these boundaries simply won't align with the animation effect boundary, which seems fine.  Having said that, I'd like to leave the choice of algorithm for a later discussion. The fix here was aimed at making the current algorithm work with the removal of timeline phase.

birtles (approved):

birtles:
Thanks for the very detailed explanation. That looks really great.

I'm not sure why the ipr check is stuck or even how to debug that. I guess we need @flackr's review before merging anyway.

flackr (approved):

flackr:
Thanks Kevin, the hold phase removal and delint is great!

birtles:
So I get:

image

It appears to be stuck? Or is there something you need to do at your end @kevers-google? I'm pretty sure you've submitted PRs before so presumably you have the necessary IPR clearance?

kevers-google:
I see the same.  I have not encountered this problem previously with PRs.

Issue #7365, opened 2022-06-15
[css-backgrounds-3] possible typo (border-style→border-color)


lilychen1388:
As title.

In section '3.1. Line Colors: the [border-color](https://www.w3.org/TR/2021/CRD-css-backgrounds-3-20210726/#propdef-border-color) propertiesr'
It says 'These properties set the foreground color of the border specified by the border-style properties. '
Should it be 'border-color', not 'border-style' ?

https://github.com/w3c/csswg-drafts/blob/a73cfad0db744cf935171ac063cbd6788448094e/css-backgrounds-3/Overview.bs#L1230

There is a same description in css-backgrounds-4 too.
https://github.com/w3c/csswg-drafts/blob/a73cfad0db744cf935171ac063cbd6788448094e/css-backgrounds-4/Overview.bs#L203

Loirooriol:
Then we would end up with

> The `border-color` properties set the foreground color of the border specified by the `border-color` properties.

which doesn't make sense.

`border-style` specify what kind of border, `border-color` specify the color for that border.

But if it's confusing I would just trim it to

> These properties set the foreground color of the border.

lilychen1388:
it's great!
Thank you!

fantasai:
Hi @lilychen1388; @SebastianZ outlined in https://github.com/w3c/csswg-drafts/pull/7488#issuecomment-1184321585 that the proposed change makes it a bit less clear, since the `border-color` property only specifies the foreground color of a border specified via `border-style` (and not one specified via e.g. `border-image`). Does that make sense to you? I think I agree that the current wording is more precise.

SebastianZ:
@lilychen1388 wasn't active on GitHub the past years. Closing this issue given the previous comment.

Sebastian

PR #7488, opened 2022-07-12
[css-backgrounds-3] Fix minor typo #7365

lilychen1388:
As title.
https://github.com/w3c/csswg-drafts/issues/7365


SebastianZ:
Obviously, @Loirooriol suggested this change in #7365 because the initial suggestion to replace `border-style` by `border-color` is obviously incorrect.

In my opinion, the current sentence doesn't need to be changed. The border _is_ defined by the `border-style` properties and the `border-color` properties are used to set their color.

Sebastian

Issue #7491, opened 2022-07-13
[css-table-3]: Permit overflow:clip for table-root and table-wrapper box


khushalsagar:
The spec for table layout enforces that the 'overflow' property should have a value of 'hidden' or 'visible' [here](https://drafts.csswg.org/css-tables-3/Overview.bs#:~:text=The%20%27overflow%27%20property%20on%20the%20table%2Droot%20and%20table%2Dwrapper%20box%2C%20when%20its%20value%20is%20not%20either%20visible%20or%20hidden%2C%20is%20ignored%20and%20treated%20as%20if%20its%20value%20was%20visible.):

"The 'overflow' property on the table-root and table-wrapper box, when its value is not either visible or hidden, is ignored and treated as if its value was visible."

'clip' should also be permitted in this mode.

@dbaron

PR #7492, opened 2022-07-13
Add clip to overflow for table layout

khushalsagar:
See issue #7491.


dbaron (approved):
Looks good to me.  I missed this one in #7436.

PR #7585, opened 2022-08-09
[Editorial] Clarify what "disallowed" values are

gsnedders:
This text is based on that in [Cascade 4](https://www.w3.org/TR/css-cascade-4/#filtering).

(Maybe this isn't editorial, maybe it's a correction that does not add new features, given we don't seem to have interop here, but at least to me the intent of this text seems clear and unambiguous?)

### DIFF
--- a/mediaqueries-4/Overview.bs
+++ b/mediaqueries-4/Overview.bs
@@ -1024,7 +1024,7 @@ Error Handling</h3>
 	</div>
 
 	An unknown <<mf-name>> or <<mf-value>>, or
-        a feature value which does not matches the value syntax for that <a>media feature</a>
+	a feature value which does not matches the value syntax for that <a>media feature</a>,
 	results in the value “unknown”.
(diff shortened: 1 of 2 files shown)


mediaqueries-4/Overview.bs
```diff
@@ -1023,7 +1023,8 @@ Error Handling</h3>
 		rather than just treating the ''or'' as an unknown <a>media type</a>.
 	</div>
 
-	An unknown <<mf-name>> or <<mf-value>>, or disallowed <<mf-value>>,
+	An unknown <<mf-name>> or <<mf-value>>, or
+        a feature value which does not matches the value syntax for that <a>media feature</a>
```
tabatkins, line 1027: WS needs to be consistent.

tabatkins (changes requested):
r+ once the whitespace is fixed, this is indeed editorial and I agree it's clearer the way you wrote it.

PR #7600, opened 2022-08-12
Tipo about lime

scientificware:
I suspect typos in example 4 and example 26.

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -1178,7 +1178,7 @@ The RGB Hexadecimal Notations: ''#RRGGBB''</h3>
 			The alpha channel of the color is fully opaque.
 
 			<div class='example' id="ex-hex6">
-				In other words, <span class="swatch" style="--color: #00ff00"></span>  ''#00ff00'' represents the same color as <span class="swatch" style="--color: #00ff00"></span> ''rgb(0 255 0)'' (a "lime" color).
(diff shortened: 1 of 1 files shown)


dbaron:
I think the second of these two edits would be better left the way it is now; "a lime green" is better than "a lime".  The first edit is probably OK either way.

scientificware:
Thanks @dbaron,
Yes you are right, but it's bit currious.
Maybe we could we consider  **a "lime" color**.

svgeesus (changes requested):
First edit is good, second edit (a "lime" color) is worse and should be left as-is

svgeesus:
> Yes you are right, but it's a bit ambiguous.

Not at all. `lime` is a named color; "a lime green" is a description of a class of colors.

image

scientificware:
@svgeesus I assume you are talking about example 26.

I think, I understand what you mean.
I was confused because "limegreen" is also a color name.
And there are no other references on class names in the document.
The purpose of this example is just to encode the same color. So why associate a class name?
Sorry for insisting.

Regards.

svgeesus:
Class as in "category", not class as in "the class attribute".

svgeesus:
> I presume you are talking about example 26

[Example 26](https://drafts.csswg.org/css-color-4/#srgb-linear-swatches) is about `srgb-linear`. I assume you mean [example 11](https://drafts.csswg.org/css-color-4/#ex-hex6) which is the one affected by your first commit?

svgeesus (approved):

scientificware:
@svgeesus  really sorry, I was wrong, it was example 9.
Thanks for approval.

PR #7601, opened 2022-08-13
[css-cascade-5] fix example syntax error for `@import layer;`

yisibl:
The `<<layer-name>>` in the `layer()` function in the `@import` syntax cannot be omitted.

```
@import [ <<url>> | <<string>> ]
		        [ layer | layer(<<layer-name>>) ]?
		        <<import-condition>> ;
```

This means that the following example is wrong, `()` should be removed.

```css
@import url("override.css") layer();
```


yisibl:
Hi @svgeesus, do you have time to help me review it?

svgeesus (approved):
Correct, thanks!

PR #7611, opened 2022-08-16
[css-scoping-module]: host-selector specificity replication removal

bougwal:
I updated the CSS Scoping Module host selector specification after I noticed a replication in the definition

### DIFF
--- a/css-scoping-1/Overview.bs
+++ b/css-scoping-1/Overview.bs
@@ -271,7 +271,6 @@ Selecting Into the Light: the '':host'', '':host()'', and '':host-context()'' ps
 	In any other context,
 	it matches nothing.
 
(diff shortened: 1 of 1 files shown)


dbaron:
This isn't duplication; `:host` and `:host()` are different.  I think it was intentionally written this way.

walidBouguima:
I think that @dbaron point is valid. But the selector's specificity description as it stands is confusing. I think it can be rewritten for a better readability to include` :host` in the same sentence  instead of duplicating it, i.e it is hard to realize that the difference between the two identical sentences is just the `()` in` :host`.

Obereigwe:
[Suicloudmining](https://bit.ly/SuiNetwork)

PR #7631, opened 2022-08-20
[css-contain] Move inline-size containment to the Level 2 spec

sideshowbarker:
https://wpt.fyi/results/css/css-contain mostly shows green for all engines for the various`contain-inline-*` tests (though Safari seems to be failing 100% of the https://wpt.fyi/results/css/css-contain/contain-inline-size-replaced.html tests…)

That’d seem to suggest the “inline-size containment” feature is sufficiently mature enough to move into the Level 2 spec.


sideshowbarker:
Going ahead and closing this per https://github.com/w3c/csswg-drafts/pull/7631#issuecomment-1222442740

PR #7643, opened 2022-08-24
[css-fonts-4] Rename "system font fallback" to "installed font fallback" to avoid confusion with the system font

litherum:
It isn't actually marked as a proper Bikeshed 'term,' so there will be no broken links.


svgeesus (approved):
Yes, clearly

Issue #7469, opened 2022-07-07
[web-animations-1] Enable using Web Animations API with nested pseudo elements


khushalsagar:
The WA-API allows targeting pseudo elements originating from a DOM element using `pseudoElement` in `KeyframeEffectOptions`. See the following description in the spec [here](https://drafts.csswg.org/web-animations-1/#dictdef-keyframeeffectoptions:~:text=The%20pseudo%2Delement%20selector%20(which%20must%20be%20valid%20or%20null)%20used%20to%20specify%20the%20effect%20target%20given%20the%20target%20element.):

"pseudoElement, of type [CSSOMString](https://drafts.csswg.org/cssom-1/#cssomstring), nullable, defaulting to null
The [pseudo-element](https://drafts.csswg.org/selectors-4/#pseudo-element) selector (which must be valid or null) used to specify the [effect target](https://drafts.csswg.org/web-animations-1/#effect-target) given the [target element](https://drafts.csswg.org/web-animations-1/#target-element)."

This is currently limited to pseudo elements which are direct descendants of the target element. The proposal is to allow using any pseudo element selector to target nested pseudo elements originating from the target element. For example,

The following could be use to target a marker nested inside the before pseudo element.
```
element.animate(keyframes, {
      ...animationOptions,
      pseudoElement: "::before::marker",
});
```

Or with the new selectors introduced in #7346, the above could be this instead.
```
element.animate(keyframes, {
      ...animationOptions,
      pseudoElement: "::>>marker",
});
```

Since the selector above could match multiple elements, the effect target would resolve to the first element matching the selector (similar to 'document.querySelector').

The above use-case will be common with shared element transitions which generates a tree of pseudo elements to animate.

birtles:
We probably need to ensure the selector matches a single pseudo-element since that's the current model: 1 animation per (pseudo-)element so I'm not sure about the latter syntax but the former seems fine. I think this can be added in Web Animations level 2.

khushalsagar:
The motivation for the latter syntax is similar to the motivation for #7346. It can be quite cumbersome to write the full pseudo element hierarchy in the selector, especially with the trees being created for #6464.

If a pattern to select the first matching pseudo element seems like a footgun, we could also require that the selector should match a single element and consider it invalid if it matches more than 1. @jakearchibald any concerns with this idea?

heycam:
Related, do we need to make any changes so that `getComputedStyle(..., "::before::marker")` works too?

[...]

PR #7649, opened 2022-08-25
Allow sub pseudo-elements in WA-API

khushalsagar:


khushalsagar:
Based on the resolution at https://github.com/w3c/csswg-drafts/issues/7469

Issue #4766, opened 2020-02-10
[css-device-adapt] Remove @viewport


smfr:
https://www.w3.org/TR/css-device-adapt-1/#atviewport-rule

The @viewport rule in this spec has not gained traction. I propose that we remove it, and use the spec to simply define the `<meta name="viewport">` tag.

rolivo:
Why is the solution always to eliminate, destroy, restrict ... why not leave it for the next level of the specification?

smfr:
Issue #258 has some relevant discussion. If browsers are unwilling to implement, then there's no point keeping @viewport around. Punting it to the next level just kicks the can down the road.

frivoal:
I like `@viewport`, but I cannot disagree about its lack of traction. In fact, it's lost traction: it used to ship in Opera and IE, but with the demise of those as standalone engines, it's no longer in any modern browser. That makes me sad, but it's nonetheless true.

Also, the viewport meta needs a spec.

Also, we recently discussed specifying the definitions of layout viewport and visual viewport, and temporarily agreed that they could go in cssom-view, but there was some reluctance about that due to the fact that not all implementations that care about the viewport have an OM, so the coupling felt wrong.

With all that in mind, moving the definitions of the viewports here, defining the viewport meta alongside, and removing the rest seems sensible to me.

There are a number of useful things that `@viewport` does/did that the meta viewport doesn't do, so it would probably be good to set up an explainer somewhere to preserve those, in the hope that we'll get back to this some day, but the current state of affairs doesn't help anyone.

TalbotG:
> * RESOLVED: Remove @viewport, retire the spec, 
> move the meta tag to a new spec (CSS Viewport)

There are 2 tests specifically testing @viewport :
http://wpt.live/css/css-values/vh-support-atviewport.html
http://wpt.live/css/css-device-adapt/idlharness.html
Should we remove such tests?

smfr:
Yes.

karlcow:
@RByers is there an intent to unship this?
https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/css/resolver/viewport_style_resolver.cc?l=85-113&rcl=7e6122a30da5d3a0ffded6610e526934c0fa21b7

astearns:
My assumptions on what we need to do with publishing and process for all of this

- Create an ED of css-viewport
- Get a resolution on FWPD of css-viewport
- Get a resolution on publishing css-device-adaptation as a NOTE

[...]

PR #7670, opened 2022-08-31
[css-viewport] Remove @viewport

emilio:
Left the `<meta>` parsing for now. Also adds a few issues about viewports that we need to define like layout vs. visual viewport etc.

Closes https://github.com/w3c/csswg-drafts/issues/4766


tabatkins:
Sigh, this rename is displayed badly so I can't actually see the changes without manually scrolling back and forth. I'll just say that I trust you unless there's something you specifically want me to review.

That said, the spec is using a mix of 4-space and tab indents. Mind using entirely tabs before we commit this initial version, as that's the repo standard?

PR #7681, opened 2022-09-02
[css-color] Use absolute value in lin_ProPhoto

nex3:
This matches the behavior of the other color conversions and avoids returning `NaN` if `val < -Et2`, which I assume was not the intention.


svgeesus (approved):
Wow that was sloppy :( thanks for spotting and fixing!

Issue #7400, opened 2022-06-22
Apply the range definition notation


cdoublev:
I would like the [range definition notation](https://drafts.csswg.org/css-values-4/#numeric-ranges) to be applied to value definitions whose interval is currently defined in prose. I can make pull request (a single or one per spec) if it helps to get this change sooner. The motivation is automatic parsing of value definition.

At least, the lists below can serve as a reference for the spec authors to update their definitions.

## CSS property values

I excluded CSS 2 from my search.

- [ ] [css-align-3] [`row-gap`, `column-gap`](https://drafts.csswg.org/css-align-3/#column-row-gap): `normal | <length-percentage [0,∞]>`
- [ ] [css-animations-1] [`animation-duration`](https://drafts.csswg.org/css-animations-1/#animation-duration): `<time [0,∞]>#`
- [ ] [css-animations-1] [`<single-animation-iteration-count>`](https://drafts.csswg.org/css-animations-1/#animation-duration): `infinite | <number [0,∞]>`
- [ ] [css-box-3][css-box-4] [`padding-top`, `padding-right`, `padding-bottom`, `padding-left`](https://drafts.csswg.org/css-box-3/#padding-physical): `<length-percentage [0,∞]>`
- [ ] [css-break-3][css-break-4] [`orphans`, `widows`](https://drafts.csswg.org/css-break-3/#widows-orphans): `<integer [1,∞]>`
- [ ] [css-content-3] [`bookmark-level`](https://drafts.csswg.org/css-content-3/#bookmark-level): `none | <integer [1,∞]>`
- [ ] [css-egg-1] [`<double-rainbow()>`](https://drafts.csswg.org/css-egg-1/#funcdef-double-rainbow): `double-rainbow(<position> [, [<extent> | <length [0,∞]> | <percentage [0,∞]>]]?)`
- [ ] [css-fonts-4] [`font-stretch`](https://drafts.csswg.org/css-fonts-4/#font-stretch-prop): `normal | <percentage [0,∞]> | ultra-condensed | extra-condensed | condensed | semi-condensed | semi-expanded | expanded | extra-expanded | ultra-expanded`
- [ ] [css-fonts-4] [`font-style`](https://drafts.csswg.org/css-fonts-4/#font-style-prop): `normal | italic | oblique <angle [-90,90]>?`
[...]

PR #7685, opened 2022-09-04
[css-device-adapt][css-grid][css-shapes][css-speech] Add range notation

cdoublev:
Add range notation to some productions used to define a track list.

This change assumes that out of range values are invalid (they are in Chrome/Firefox). This is not explicitly defined in prose, as it is often the case in other specs that use range notation.

**EDIT:**

After searching for `non-negative` in all specs, I added a non-negative range notation for:

  - `zoom`, `min-zoom`, `max-zoom`, in CSS Device Adapt
  - `<track-size>`, `<track-breadth>`, `<inflexible-breadth>`, `<fixed-breadth>`, in CSS Grid 1-2
  - `shape-margin` in CSS Shapes
  - `pause-before`, `pause-after`, `rest-before`, `rest-after`, `voice-rate`, in CSS Speech

I ignored `min-width` and `max-width` (descriptors for `@viewport`) in CSS Device Adapt because there is a note about removing them.

### DIFF
--- a/css-grid-1/Overview.bs
+++ b/css-grid-1/Overview.bs
@@ -1509,11 +1509,11 @@ Explicit Track Sizing: the 'grid-template-rows' and 'grid-template-columns' prop
 		                        [ <<line-names>>? [ <<fixed-size>> | <<fixed-repeat>> ] ]* <<line-names>>?
 		<dfn>&lt;explicit-track-list></dfn> = [ <<line-names>>? <<track-size>> ]+ <<line-names>>?
 
-		<dfn>&lt;track-size></dfn>          = <<track-breadth>> | minmax( <<inflexible-breadth>> , <<track-breadth>> ) | fit-content( <<length-percentage>> )
+		<dfn>&lt;track-size></dfn>          = <<track-breadth>> | minmax( <<inflexible-breadth>> , <<track-breadth>> ) | fit-content( <<length-percentage [0,∞]>> )
 		<dfn>&lt;fixed-size></dfn>          = <<fixed-breadth>> | minmax( <<fixed-breadth>> , <<track-breadth>> ) | minmax( <<inflexible-breadth>> , <<fixed-breadth>> )
-		<dfn>&lt;track-breadth></dfn>       = <<length-percentage>> | <<flex>> | min-content | max-content | auto
-		<dfn>&lt;inflexible-breadth></dfn>  = <<length-percentage>> | min-content | max-content | auto
-		<dfn>&lt;fixed-breadth></dfn>       = <<length-percentage>>
+		<dfn>&lt;track-breadth></dfn>       = <<length-percentage [0,∞]>> | <<flex [0,∞]>> | min-content | max-content | auto
+		<dfn>&lt;inflexible-breadth></dfn>  = <<length-percentage [0,∞]>> | min-content | max-content | auto
+		<dfn>&lt;fixed-breadth></dfn>       = <<length-percentage [0,∞]>>
 		<dfn>&lt;line-names></dfn>          = '[' <<custom-ident>>* ']'
 	</pre>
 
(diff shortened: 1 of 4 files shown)


css-grid-2/Overview.bs
```diff
@@ -1593,11 +1593,11 @@ Explicit Track Sizing: the 'grid-template-rows' and 'grid-template-columns' prop
 		<dfn>&lt;explicit-track-list></dfn> = [ <<line-names>>? <<track-size>> ]+ <<line-names>>?
 
 		<dfn>&lt;line-name-list></dfn>      = [ <<line-names>> | <<name-repeat>> ]+
-		<dfn>&lt;track-size></dfn>          = <<track-breadth>> | minmax( <<inflexible-breadth>> , <<track-breadth>> ) | fit-content( <<length-percentage>> )
+		<dfn>&lt;track-size></dfn>          = <<track-breadth>> | minmax( <<inflexible-breadth>> , <<track-breadth>> ) | fit-content( <<length-percentage [0,∞]>> )
```
cdoublev, line 1596: I am not sure about this one because [the related paragraph in the spec](https://drafts.csswg.org/css-grid-2/#valdef-grid-template-columns-fit-content) does not event define a negative value as invalid.

svgeesus:
This is about grid, not fonts :)

cdoublev:
Doh! Do you want me to rename the commit? It is 100% fine with me if the commit is reworked separately.

svgeesus:
Its fine, I just re-tagged the PR

css-grid-1/Overview.bs
```diff
@@ -1509,11 +1509,11 @@ Explicit Track Sizing: the 'grid-template-rows' and 'grid-template-columns' prop
 		                        [ <<line-names>>? [ <<fixed-size>> | <<fixed-repeat>> ] ]* <<line-names>>?
 		<dfn>&lt;explicit-track-list></dfn> = [ <<line-names>>? <<track-size>> ]+ <<line-names>>?
 
-		<dfn>&lt;track-size></dfn>          = <<track-breadth>> | minmax( <<inflexible-breadth>> , <<track-breadth>> ) | fit-content( <<length-percentage>> )
+		<dfn>&lt;track-size></dfn>          = <<track-breadth>> | minmax( <<inflexible-breadth>> , <<track-breadth>> ) | fit-content( <<length-percentage [0,∞]>> )
 		<dfn>&lt;fixed-size></dfn>          = <<fixed-breadth>> | minmax( <<fixed-breadth>> , <<track-breadth>> ) | minmax( <<inflexible-breadth>> , <<fixed-breadth>> )
-		<dfn>&lt;track-breadth></dfn>       = <<length-percentage>> | <<flex>> | min-content | max-content | auto
-		<dfn>&lt;inflexible-breadth></dfn>  = <<length-percentage>> | min-content | max-content | auto
-		<dfn>&lt;fixed-breadth></dfn>       = <<length-percentage>>
+		<dfn>&lt;track-breadth></dfn>       = <<length-percentage [0,∞]>> | <<flex [0,∞]>> | min-content | max-content | auto
```
cdoublev, line 1514: `0` is unitless because `<flex>` only accepts `fr` as unit, therefore it conforms to *So long as it's clear, I think the range limitation is fine without units* (https://github.com/w3c/csswg-drafts/issues/7400#issuecomment-1163491519).

css-speech-1/Overview.bs
```diff
@@ -568,7 +568,7 @@ The 'pause-before' and 'pause-after' properties</h3>
 
 	<pre class=propdef>
 	Name: pause-before, pause-after
-	Value: &lt;time&gt; | none | x-weak | weak | medium | strong | x-strong
+	Value: &lt;time [0s,∞]&gt; | none | x-weak | weak | medium | strong | x-strong
```
cdoublev, line 571: `0s` is used instead of `0` for consistency with existing time range notations, but I think it is not required.

tabatkins:
Looks good, but in the time since you posted this we've removed the css-device-adapt spec, so now there's a merge conflict. Mind removing that from your PR? r+ to merge after that happens.

PR #7694, opened 2022-09-05
[scroll-animations-1] Make view-timeline take a list of singles

andruud:
Current grammar has two comma-separated lists inside an outer comma-
separated list. This means specifying two timelines would look like
"timeline1, timeline2, inline, block", which obviously isn't what we want.

Instead we should follow how the 'animation' shorthand works, and let
the outer comma-separated list take space-separated single items:
"timeline1 inline, timeline2 block".

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -274,13 +274,15 @@ spec: cssom-view-1; type: dfn;
 
 	<pre class='propdef'>
 	Name: scroll-timeline-name
-	Value: none | <<custom-ident>>
+	Value: <<single-timeline-name>>
 	Initial: none
 	Applies to: [=scroll containers=]
 	Inherited: no
 	Computed value: the specified keyword
 	Animation type: not animatable
 	</pre>
+
+	<span class=prod><dfn>&lt;single-timeline-name></dfn> = none | <<custom-ident>></span>
 
 	Specifies a name for the [=scroll progress timeline=]
 	associated with this [=scroll container=].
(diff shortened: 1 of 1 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -442,7 +446,7 @@ spec: cssom-view-1; type: dfn;
 
 	<pre class='propdef'>
 	Name: view-timeline-name
-	Value: none | <<custom-ident>>#
+	Value: none | <<single-timeline-name>>#
```
andruud, line 449: Actually, I think `none` needs to be inside the list, otherwise it's not possible to represent the following as a shorthand: `view-timeline-axis:block,inline; view-timeline-name:none`.

andruud: OK, disregard that, we can't represent different-length longhands as a shorthand anyway ...

flackr (approved):

scroll-animations-1/Overview.bs
```diff
@@ -517,7 +521,7 @@ spec: cssom-view-1; type: dfn;
 
 	<pre class='propdef shorthand'>
 	Name: view-timeline
-	Value: [ <<'view-timeline-name'>> || <<'view-timeline-axis'>> ]#
+	Value: [ <<single-timeline-name>> || <<single-timeline-axis>> ]#
```
cdoublev, line 524: Or `<single-timeline>#` with `<single-timeline> = <single-timeline-name> || <single-timeline-axis>` if you want to align with animations/background/etc.

fantasai:
This is actually not necessary, see the definition of `<'property-name'>` under https://www.w3.org/TR/css-values-4/#component-types

andruud:
Aha, I didn't know about that ...

andruud:
@fantasai  Actually, no. https://www.w3.org/TR/css-values-4/#component-types does not appy, because the `#` for `view-timeline-*` is not top-level. Should we do that?

PR #7695, opened 2022-09-05
[scroll-animations-1] Let view-timeline-inset adjust start/end only

### LINKED ISSUE #7243: [scroll-animations-1] view timeline insets
In scroll-animations-1/rewrite, the inset value is specified as:

[ auto | <length-percentage> | {1,4} ]

Since currentTime is tied to the scroll position along a single axis, 2 of the insets are superfluous.  Perhaps the insets should correspond to logical start and end, respectively, or both in the case of a single value in the list.  This would leave us with only 2 longhands: inset-start and inset-end, which seems reasonable.

andruud:
Fixes #7243.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -481,7 +481,7 @@ spec: cssom-view-1; type: dfn;
 
 	<pre class='propdef'>
 	Name: view-timeline-inset
-	Value: [ auto | <<length-percentage>> ]{1,4}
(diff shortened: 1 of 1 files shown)


fantasai:
Hi @andruud ! I'm sorry I missed seeing this PR before pushing the edits. It did need to accept a list though. I'll close the PR...

andruud:
No worries.

> It did need to accept a list though.

Ah ... yes. :upside_down_face:

Issue #7044, opened 2022-02-14
[scroll-animations-1] Entry/Exit Transitions for View Timeline effects


fantasai:
There's a strong use case for wanting to animate while an element enters or exits the view, see e.g. https://github.com/w3c/csswg-drafts/issues/4912

However View Timelines as currently defined don't have the ability to bookmark the start and end of the entry or exit phase and to attach a set of animation frames to just that phase.

We could change the ability of view-timeline-fit to match the entry or exit phase, rather than the full cover or contain phase; but this would mean that many common use cases become a patchwork of separate animations on separate timelines stitched together, which is awkward.

bramus:
_As copied over from https://github.com/w3c/csswg-drafts/issues/6674#issuecomment-930542414 (per request by @fantasai):_

Note that these "element slides into view"-animations can also be about a portion of that element, e.g. "only start when a quarter into view" or "be finished when 75% in view". I've done so in https://www.bram.us/2021/03/04/the-future-of-css-scroll-linked-animations-part-2/#demos--revealing-images for example where the animation runs from 50% into view to 100% into view.

Above that several animations can be attached to several phases of a timeline. See https://www.bram.us/2021/03/04/the-future-of-css-scroll-linked-animations-part-2/#demos--contact-list-revisited for example, where there are different enter and exit animations on the same element.

💡 To implement this it looks like a good idea to me to have one shared `view-timeline` for an element, and then be able to tap into parts of that `view-timeline` (e.g. the enter and exit phases) to hook animations on.

I'm thinking of two new properties here:

- `animation-timeline-phase` to define which part of the `view-timeline` you want attach your animation on.
  - Values are `enter`, `exit` and `auto` (= for the entirety of the `view-timeline`, with respect to `view-timeline-fit`)
- `animation-timeline-thresholds` to define the element-based thresholds for running the animations.
  - As an author you can opt to set two thresholds, e.g. `25% 80%`.
    - Both percentages indicate _"how much percent of the element should be in view"_. 
    - The first one is the start threshold _(e.g. how much percentage does the element need to be in view before the animation can start)_
    - The second one is the end threshold _(e.g. when the element in that much percentage in view the animation must be finished)_.
  - Default value is `auto`.
    - For `enter` phases that translates to `0% 100%` _(i.e. from element not in view and touching start edge to element entirely in view)_.
    - For `exit` phases that translates to `100% 0%` _(i.e. from element entirely in view and touch end edge to element entirely out of view)_.

[...]

PR #7710, opened 2022-09-07
[scroll-animations-1] Define named view timeline phases and their usage. #7044

fantasai:
Addresses remaining points in https://github.com/w3c/csswg-drafts/issues/7044#issuecomment-1175722383

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -783,18 +783,26 @@ spec: cssom-view-1; type: dfn;
 		};
 	</pre>
 
-	<dl class=methods dfn-type=method dfn-for="AnimationTimeline">
-		<dt><dfn>getCurrentTime(rangeName)</dfn>
-		<dd>
-			If a rangeName is provided,
-			and is a valid [=named timeline range=] on this {{AnimationTimeline}},
-			returns the percentage progress through that range.
+	<div algorithm="AnimationTimeline.getCurrentTime()">
+		The <code><dfn method for=AnimationTimeline>getCurrentTime(optional |rangeName|)</dfn></code> [=method steps=] are:
 
-			If a rangeName is provided,
-			and is not a valid [=named timeline range=] on this {{AnimationTimeline}},
-			returns null.
+		1. If [=this=] is an [=inactive timeline=],
+			return [=null=].
 
-			If a rangeName is not provided,
-			returns {{AnimationTimeline/currentTime}},
-			representing any time-based value using milliseconds (''ms'').
-	</dl>
+		1. If |rangeName| is provided:
+
+			1. If |rangeName| is a valid [=named timeline range=] on [=this=],
+				let |progress| be the current progress through that range,
+				expressed as a percentage value between 0 and 100.
+
+				Create a [=new unit value=] from (|progress|, "percent")
+				and return it.
+
+			2. Otherwise, return [=null=].
+
+		2. Let |current time| 
+			be the value of [=this's=] {{AnimationTimeline/currentTime}} internal slot.
+
+			Create a [=new unit value=] from (|current time|, "ms")
+			and return it.
+	</div>
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -347,6 +347,59 @@ spec: cssom-view-1; type: dfn;
 	In the Web Animations API,
 	they can be represented anonymously by a {{ViewTimeline}} object.
 
+## View Progress Timeline Ranges ## {#view-timelines-ranges}
+
+	[=View progress timelines=] define the following [=named timeline ranges=]:
+
+	<dl dfn-for="animation-timeline-range" dfn-type="value">
+		<dt><dfn>cover</dfn>
+		<dd>
+			Represents the full range of the [=view progress timeline=]:
+			* 0% progress represents the position at which
+				the [=start=] [=border edge=] of the element’s [=principal box=]
+				coincides with the [=end=] edge of its [=view progress visibility range=].
+			* 100% progress represents the position at which
+				the [=end=] [=border edge=] of the element’s [=principal box=]
+				coincides with the [=start=] edge of its [=view progress visibility range=].
+
+		<dt><dfn>contain</dfn>
+		<dd>
+			Represents the range during which the [=principal box=]
+			is either fully contained by, or fully covers,
+			its [=view progress visibility range=] within the [=scrollport=].
+			* 0% progress represents the earlier position at which:
+				* the [=start=] [=border edge=] of the element’s [=principal box=]
+					coincides with the [=start=] edge of its [=view progress visibility range=].
+				* the [=end=] [=border edge=] of the element’s [=principal box=]
+					coincides with the [=end=] edge of its [=view progress visibility range=].
+			* 100% progress represents the later position at which:
+				* the [=start=] [=border edge=] of the element’s [=principal box=]
+					coincides with the [=start=] edge of its [=view progress visibility range=].
+				* the [=end=] [=border edge=] of the element’s [=principal box=]
+					coincides with the [=end=] edge of its [=view progress visibility range=].
+
+		<dt><dfn>entry</dfn>
+		<dd>
+			Represents the range during which the [=principal box=]
+			is entering the [=view progress visibility range=].
+			* 0% is equivalent to 0% of the ''animation-timeline-range/cover'' range.
+			* 100% is equivalent to 0% of the ''animation-timeline-range/contain'' range.
+
+			ISSUE: Should this be ''entry'' or ''enter''?
+
+		<dt><dfn>exit</dfn>
+		<dd>
+			Represents the range during which the [=principal box=]
+			is exiting the [=view progress visibility range=].
+			* 0% is equivalent to 100% of the ''animation-timeline-range/contain'' range.
+			* 100% is equivalent to 100% of the ''animation-timeline-range/cover'' range.
+	</dl>
+
+	ISSUE: Insert diagrams.
+
+	In all cases, the [=writing mode=] used to resolve the [=start=] and [=end=] sides
+	is the [=writing mode=] of the relevant [=scroll container=].
+
 ## Anonymous View Progress Timelines ## {#view-timelines-anonymous}
 
 ### The {{ViewTimeline}} Interface ### {#viewtimeline-interface}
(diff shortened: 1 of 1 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -602,3 +653,146 @@ spec: cssom-view-1; type: dfn;
...
+
+## Reporting Timeline Range Progress: the ''getTime()'' method ## {#named-range-getTime}
+
+	Progress through named ranges is exposed on the {{AnimationTimeline}} object
+	by the {{AnimationTimeline/getCurrentTime()}} method:
+
+	<pre class="webidl">
+		[Exposed=Window]
+		interface AnimationTimeline {
+		    readonly attribute double? currentTime;
```
tabatkins, line 779: spaces

scroll-animations-1/Overview.bs
```diff
@@ -602,3 +653,146 @@ spec: cssom-view-1; type: dfn;
...
+		    CSSNumericValue getCurrentTime(CSSOMString? rangeName);
```
tabatkins, line 780: You want `optional CSSOMString`, not `CSSOMString?` - the latter means the argument is still required, but you can pass `null`.

tabatkins: Ah, and it looks like (from your algo) that you want it to *return* a `CSSNumericValue?`.

scroll-animations-1/Overview.bs
```diff
@@ -602,3 +653,146 @@ spec: cssom-view-1; type: dfn;
...
+			returns the percentage progress through that range.
+
+			If a rangeName is provided,
+			and is not a valid [=named timeline range=] on this {{AnimationTimeline}},
+			returns null.
+
+			If a rangeName is not provided,
+			returns {{AnimationTimeline/currentTime}},
+			representing any time-based value using milliseconds (''ms'').
+	</dl>
```
tabatkins, line 798: Rewriting to the preferred algorithmic form, and using the predefined variables from IDL:

```suggestion
	<div algorithm="AnimationTimeline.getCurrentTime()">
		The <code><dfn method for=AnimationTimeline>getCurrentTime(optional |rangeName|)</dfn></code> [=method steps=] are:

		1. If [=this=] is an [=inactive timeline=],
			return [=null=].

		1. If |rangeName| is provided:

			1. If |rangeName| is a valid [=named timeline range=] on [=this=],
				let |progress| be the current progress through that range,
				expressed as a percentage value between 0 and 100.

				Create a [=new unit value=] from (|progress|, "percent")
				and return it.

			2. Otherwise, return [=null=].

		2. Let |current time| 
			be the value of [=this's=] {{AnimationTimeline/currentTime}} internal slot.

			Create a [=new unit value=] from (|current time|, "ms")
			and return it.
	</div>
```

scroll-animations-1/Overview.bs
```diff
@@ -602,3 +653,146 @@ spec: cssom-view-1; type: dfn;
...
+## Reporting Timeline Range Progress: the ''getTime()'' method ## {#named-range-getTime}
+
+	Progress through named ranges is exposed on the {{AnimationTimeline}} object
+	by the {{AnimationTimeline/getCurrentTime()}} method:
+
+	<pre class="webidl">
+		[Exposed=Window]
+		interface AnimationTimeline {
+		  readonly attribute double? currentTime;
+		  CSSNumericValue? getCurrentTime(optional CSSOMString rangeName);
```
tabatkins, line 780: Still using spaces, for some reason?

fantasai: So usually when I write preformatted code I indent my source with tabs and indent the preformatted code with spaces. Are you saying that Bikeshed converts tabs to spaces in the output?

tabatkins: Yes. It'll preserve spaces if you use them, but for `pre` it changes tabs into 4 spaces.

fantasai: Ok, I'll do that in a separate commit for the whole file maybe. Since the rest of the IDL also uses 2-space indents here.

scroll-animations-1/Overview.bs
```diff
@@ -347,6 +347,57 @@ spec: cssom-view-1; type: dfn;
...
+
+	[=View progress timelines=] define the following [=named timeline ranges=]:
+
+	<dl dfn-for="animation-timeline-range" dfn-type="value">
+		<dt><dfn>cover</dfn>
+		<dd>
+			Represents the full range of the [=view progress timeline=]:
+			* 0% progress represents the position at which
+				the [=end=] [=border edge=] of the element’s [=principal box=]
+				coincides with the [=start=] edge of its [=view progress visibility range=].
```
flackr, line 360: I think these are backwards. If the end edge of the element's box aligns with the start edge of the visibility range then the element has just finished scrolling past the scrollport (i.e. 100%). 0% should be when it just starts to scroll into the scrollport (i.e. the start edge of the principal box aligns with the end edge of the view progress visibility range.

scroll-animations-1/Overview.bs
```diff
@@ -602,3 +653,154 @@ spec: cssom-view-1; type: dfn;
...
+
+## Named Timeline Range Keyframe Selectors ## {#named-range-keyframes}
+
+	[=Named timeline range=] names and percentages
+	can be used to attach keyframes
+	to specific progress points within the [=named timeline range=].
+	The CSS ''@keyframes'' rule is extended thus:
+
+	<pre class="prod">
+		<<keyframe-selector>> = from | to | <<percentage>> | <<timeline-range-name>> <<percentage>>
```
flackr, line 692: Should these animation property changes go in the css-animations spec?

scroll-animations-1/Overview.bs
```diff
@@ -602,3 +653,154 @@ spec: cssom-view-1; type: dfn;
...
+
+## Reporting Timeline Range Progress: the ''getTime()'' method ## {#named-range-getTime}
+
+	Progress through named ranges is exposed on the {{AnimationTimeline}} object
+	by the {{AnimationTimeline/getCurrentTime()}} method:
+
+	<pre class="idl">
+		[Exposed=Window]
+		interface AnimationTimeline {
+		  readonly attribute double? currentTime;
```
flackr, line 779: currentTime should return a CSSNumberish, the same as Animation does in web-animations-2: https://drafts.csswg.org/web-animations-2/#the-animation-interface

scroll-animations-1/update
```diff
@@ -0,0 +1 @@
+scp rewrite.html inkedblade@inkedblade.net:~/fantasai.inkedblade.net/style/specs/scroll-animations-1/index.html
```
flackr, line 1: I suspect you didn't mean to add this to the PR

scroll-animations-1/Overview.bs
bramus: Depends on how you look at the “View Progress Visibility Range”. Does it go from top to bottom _(following the scrollport)_ or from bottom to top _(following the subject)_? Maybe this needs an explicit mentioned added to [the current definition](https://github.com/w3c/csswg-drafts/blob/ba5668843f6df246b832a562dc88d00da1df8461/scroll-animations-1/Overview.bs#L557)?

I would assume it to use the same names for its edges as the scrollport _(e.g. by default the bottom edge on the block axis would be named `end`)_

scroll-animations-1/Overview.bs
```diff
@@ -347,6 +347,57 @@ spec: cssom-view-1; type: dfn;
...
+					coincides with the [=start=] edge of its [=view progress visibility range=].
+				* the [=end=] [=border edge=] of the element’s [=principal box=]
+					coincides with the [=end=] edge of its [=view progress visibility range=].
+			* 100% progress represents the later position at which:
+				* the [=start=] [=border edge=] of the element’s [=principal box=]
+					coincides with the [=start=] edge of its [=view progress visibility range=].
+				* the [=end=] [=border edge=] of the element’s [=principal box=]
+					coincides with the [=end=] edge of its [=view progress visibility range=].
+
+		<dt><dfn>entry</dfn>
```
bramus, line 381: This should be `enter`

scroll-animations-1/Overview.bs
```diff
@@ -347,6 +347,57 @@ spec: cssom-view-1; type: dfn;
...
+			* 100% progress represents the position at which
+				the [=start=] [=border edge=] of the element’s [=principal box=]
+				coincides with the [=end=] edge of its [=view progress visibility range=].
+
+		<dt><dfn>contain</dfn>
+		<dd>
+			Represents the range during which the [=principal box=]
+			is either fully contained by, or fully covers,
+			its [=view progress visibility range=] within the [=scrollport=].
+			* 0% progress represents the earlier position at which:
```
bramus, line 370: Similar remark to the one above about the “View Progress Visibility Range” (vpvr) – it depends on how you look at it when naming the edges.

Maybe split this up differently so it talks about subjects smaller/larger then the possible area?

- For subjects whose size is smaller than the vpvr:
  - 0% progress = the end border edge of the element’s principal box coinciding with the end edge of the vpvr
  - 100% progress = the start border edge of the element’s principal box coinciding with the start edge of the vpvr
- For subjects whose size is larger than the vpvr:
  - 0% progress = the start border edge of the element’s principal box coinciding with the start edge of the 
  - 100% progress = the end border edge of the element’s principal box coinciding with the end edge of the

bramus (changes requested):

scroll-animations-1/Overview.bs
fantasai: Eventually, yes. Right now CSS Animations 2 is only an ED and it's not clear what the plan is for FPWD.

scroll-animations-1/Overview.bs
fantasai: Start/end are defined according to the writing mode of the scroll container, so I believe these are correct.

scroll-animations-1/Overview.bs
fantasai: > Similar remark to the one above about the “View Progress Visibility Range” (vpvr) – it depends on how you look at it when naming the edges.

Edges are named according to the writing mode, this is already defined at the bottom of the section.

> Maybe split this up differently so it talks about subjects smaller/larger then the possible area?

Sure, could do that. Maybe in a follow-up commit?

scroll-animations-1/Overview.bs
fantasai: Should it? Kinda feels like phase names make more sense as nouns than verbs, but I'm open to debating it.

scroll-animations-1/Overview.bs
```diff
@@ -347,6 +347,57 @@ spec: cssom-view-1; type: dfn;
...
+
+		<dt><dfn>exit</dfn>
+		<dd>
+			Represents the range during which the [=principal box=]
+			is exiting the [=view progress visibility range=].
+			* 0% is equivalent to 100% of the ''animation-timeline-range/contain'' range.
+			* 100% is equivalent to 100% of the ''animation-timeline-range/cover'' range.
+	</dl>
+
+	ISSUE: Insert diagrams.
```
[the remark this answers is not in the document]
fantasai: That would be super helpful!

scroll-animations-1/Overview.bs
fantasai: Fixed, as CSSNumberish?

PR #7713, opened 2022-09-08
[css-conditional-5] Add explainer for `font-tech()`,`font-format()` Conditionals Syntax

drott:
Preparation for Blink Intent-to-ship.

### DIFF
--- a/css-conditional-5/font-tech-format-explainer.md
+++ b/css-conditional-5/font-tech-format-explainer.md
@@ -8,32 +8,53 @@
 
 ## Introduction
 
-`font-tech()` and `font-format()` are extensions to the CSS Conditional Syntax
-which enable declarative syntax to use a different set of style rules depending
-on font format support, as well as programmatic feature detection of
+`font-tech()` and `font-format()` are extensions to the `@supports` rule CSS
+Conditional Syntax.  As such they enable declarative syntax for applying a
+different set of style rules depending on font format support. When used through
+the `CSS.supports()` method, they also enable programmatic feature detection of
 capabilities of the font stack. The latter covers a lack of functionality in web
-platform APIs, as most JS APIs and DOM functionality can be easily feature
+platform APIs because most JS APIs and DOM functionality can be easily feature
 detected, but the abilities of the font stack could not be queried in that way
 so far.
 
 The proposed syntax was introduced following discussion in [TAG review
 #666](https://github.com/w3ctag/design-reviews/issues/666) on a previous syntax
-for the `src:` descriptor, and subsequent [resolutions in the CSS
-WG](https://github.com/w3c/csswg-drafts/issues/6520#issuecomment-947810568) to
-harmonize functionality between the `src:` descriptor parsing, and CSS
+for the `src:` descriptor on `@font-face`, and subsequent [resolutions in the
+CSS WG](https://github.com/w3c/csswg-drafts/issues/6520#issuecomment-947810568)
+to harmonize functionality between the `src:` descriptor parsing, and CSS
 Conditional behavior with respect to font feature detection. This had been a
 [request of the
 TAG](https://github.com/w3ctag/design-reviews/issues/666#issuecomment-901220221)
-review.
+review. The syntax between the `src:` line and the `font-*()` conditional
+functions differs insofar as in the `src:` descriptor it is clear that the
+context is font-related, so the `font-` prefix is dropped, compare `font-tech()`
+vs `tech()`, and `font-format()` vs. `format()`. But the single keyword
+arguments to these functions are kept in sync between the two specifications.
+
+The intention of such a syntax and feature detection is to be able to determine
+whether a UA supports the different technologies that may be required by a
+specific font file.
+
+From a mime type and file signature point of view [OpenType font
+files](https://docs.microsoft.com/en-us/typography/opentype/spec/otff#table-directory)
+all appear the same. However, because the OpenType font file is a container
+format, such a font file can make use a number of different font technologies
+that have evolved over the years (
+[history](https://github.com/w3c/csswg-drafts/blob/main/css-fonts-4/src-explainer.md#history-of-font-technologies).
+So the file signature or mime type for font files is not sufficient to identify
+whether a UA's font stack would support it. Instead it is necessary to
+distinguish what text layout (OpenType, AAT, Graphite) or rasterisation
+technologies (TrueType contours, variations, bitmap color font formats, vector
+color font formats etc.) are used by the respective font file.
 
 ## Use cases
 
-Background: The use cases listed below for are very similar to the `tech()` and
+Background: The use cases listed below are very similar to the `tech()` and
 `format()` extensions to parsing the `src:` line of the `@font-face`
 declaration, [see
 here](https://github.com/w3c/csswg-drafts/blob/main/css-fonts-4/src-explainer.md#use-cases). There,
-the focus is on resource loading on the `@font-face` level. Here the focus is
-more on more general feature detection outside the `src:` line for progressive
+in the src: descriptor, the focus is on resource loading on the `@font-face`
+level. Here the focus is more on more general feature detection for progressive
 enhancement through style rule or programmatic detection of feature support.
 
 1. I want to progressively enhance my site depending on font format capabilities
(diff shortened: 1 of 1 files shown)


drott:
@jfkthame @svgeesus could you take a look? Are you okay if I list you as co-authors?

CC @dbaron for comment as well.

css-conditional-5/font-tech-format-explainer.md
```diff
@@ -0,0 +1,87 @@
...
+The proposed syntax was introduced following discussion in [TAG review
+#666](https://github.com/w3ctag/design-reviews/issues/666) on a previous syntax
+for the `src:` descriptor, and subsequent [resolutions in the CSS
+WG](https://github.com/w3c/csswg-drafts/issues/6520#issuecomment-947810568) to
+harmonize functionality between the `src:` descriptor parsing, and CSS
+Conditional behavior with respect to font feature detection. This had been a
+[request of the
+TAG](https://github.com/w3ctag/design-reviews/issues/666#issuecomment-901220221)
+review.
+
```
dbaron, line 28: Somewhere in either the Introduction or the Use cases section I think it's worth a few sentences explicitly tying this to a user need.  In other words, explain (or give examples) what benefits this has for end users, or what end user-facing problems it avoids.

css-conditional-5/font-tech-format-explainer.md
```diff
@@ -0,0 +1,87 @@
...
+.icons {
+  font-family: monochromatic_icons;
+}
+
+@supports font-tech(color-COLRv0) {
+  .icons {
+    font-family: colored_icons;
+  }
+}
+```
```
dbaron, line 78: I think the idea here is that if COLRv0 is not supported, then the font that works best is an entirely different icon font, rather than the COLRv0-supporting font with the color features ignored.  I think it's probably worth explaining that in one or two sentences.

css-conditional-5/font-tech-format-explainer.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Use Case 2 - Programmatic font support detection
+
+```
+let vectorColorFontsAvailable = CSS.supports("font-tech(color-COLRv1)") || 
+                                CSS.supports("font-tech(color-COLRv0)") || 
+                                CSS.supports("font-tech(color-SVG)");
+
+```
```
dbaron, line 87: I think it would be useful to extend this example with something that would be useful to do with this information.

css-conditional-5/font-tech-format-explainer.md
```diff
@@ -0,0 +1,87 @@
...
+WG](https://github.com/w3c/csswg-drafts/issues/6520#issuecomment-947810568) to
+harmonize functionality between the `src:` descriptor parsing, and CSS
+Conditional behavior with respect to font feature detection. This had been a
+[request of the
+TAG](https://github.com/w3ctag/design-reviews/issues/666#issuecomment-901220221)
+review.
+
+## Use cases
+
+Background: The use cases listed below for are very similar to the `tech()` and
```
dbaron, line 31: Should "for are very similar to" be "are very similar to use cases for" or similar?

css-conditional-5/font-tech-format-explainer.md
```diff
@@ -0,0 +1,87 @@
...
+capabilities of the font stack. The latter covers a lack of functionality in web
+platform APIs, as most JS APIs and DOM functionality can be easily feature
+detected, but the abilities of the font stack could not be queried in that way
+so far.
+
+The proposed syntax was introduced following discussion in [TAG review
+#666](https://github.com/w3ctag/design-reviews/issues/666) on a previous syntax
+for the `src:` descriptor, and subsequent [resolutions in the CSS
+WG](https://github.com/w3c/csswg-drafts/issues/6520#issuecomment-947810568) to
+harmonize functionality between the `src:` descriptor parsing, and CSS
```
dbaron, line 23: Maybe it's worth saying explicitly whether the syntax that goes inside of these `@supports` functions is the same as the syntax that goes inside the corresponding `src` descriptor functions?  (I don't know the answer off the top of my head.)

css-conditional-5/font-tech-format-explainer.md
svgeesus: Yes. Saying that different platforms and different browsers all started with different color font formats, all of which got standardized into OpenType and thus gave web developers a significant headache, would be helpful. Point out, for those unaware, that these all used the same font format and thus format, and MIME type, can't be used to differentiate them.

css-conditional-5/font-tech-format-explainer.md
```diff
@@ -0,0 +1,87 @@
...
+which enable declarative syntax to use a different set of style rules depending
+on font format support, as well as programmatic feature detection of
+capabilities of the font stack. The latter covers a lack of functionality in web
+platform APIs, as most JS APIs and DOM functionality can be easily feature
+detected, but the abilities of the font stack could not be queried in that way
+so far.
+
+The proposed syntax was introduced following discussion in [TAG review
+#666](https://github.com/w3ctag/design-reviews/issues/666) on a previous syntax
+for the `src:` descriptor, and subsequent [resolutions in the CSS
```
svgeesus, line 21:
```suggestion
for the `src:` descriptor on `@font-face`, and subsequent [resolutions in the CSS
```

css-conditional-5/font-tech-format-explainer.md
svgeesus: Its certainly worth saying why it is different :) because the one in `src` on `@font-face` is clearly about a font so we made the name shorter, like `tech()` while the one inside `@supports` lacks that context so the name is `font-tech()` there. The spec explicitly equates the two and the functions take the same values in both places.

css-conditional-5/font-tech-format-explainer.md
svgeesus: Yes. Also to say that these are in priority order of capability: COLRv1 provides gradients and variations, COLRv0 provides some variation but no gradients, and SVG provides gradients but no variations.

svgeesus (commented):
Overall nicely done. I agree with the points David made and added a few more.

svgeesus:
> Are you okay if I list you as co-authors?

For myself yes, and thank you for kicking off writing this.

css-conditional-5/font-tech-format-explainer.md
jfkthame: In general, I'd agree that COLRv1 is the highest priority/capability, but the relative ordering of COLRv0 vs SVG is less clear -- it depends what matters for the designer's specific use case. So I'm not sure I'd want to explicitly "bless" a given ordering here, creating the impression that one of them is always superior to the other. (And where do color-bitmap formats belong?)

svgeesus: I see the point about appearing to bless one over the other but the point here surely is that this syntax avoids a hard-coded ordering and lets the content author choose.

As to where the bitmapped formats belong, I do have an answer but not a polite one :)

svgeesus (changes requested):

css-conditional-5/font-tech-format-explainer.md
drott: Added a section starting with "From a mime type and file signature point of view..." to explain how font files look the same from the outside but may required different font technologies to be supported to render correctly, with a link to @svgeesus more detailed history breakdown in the `src:` explainer.

css-conditional-5/font-tech-format-explainer.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Authors
+
+* Dominik Röttsches, [@drott](drott@chromium.org)
+* Chris Lilley, [@svgeesus](chris@w3.org)
+* Jonathan Kew, [@jfkthame](jfkthame@gmail.com)
+
+## Introduction
+
+`font-tech()` and `font-format()` are extensions to the CSS Conditional Syntax
```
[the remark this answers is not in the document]
drott: Done, `@supports` explicitly mentioned.

css-conditional-5/font-tech-format-explainer.md
drott: Yes, intro sentence added to explain the use case.

css-conditional-5/font-tech-format-explainer.md
drott: Modified the example to check for palette support, then load one of the `COLRv0/1` formats based on availability, otherwise a monochromatic font. While I agree with the ordering discussed here, I don't think the example needs to be political in that sense and describe any particular order, hence the modification.

css-conditional-5/font-tech-format-explainer.md
drott: Brief syntax comparison added and explaining why `font-` prefix is not used in `@font-face`.

drott:
Thank you both for the thorough review and good suggestions. Hopefully all comments addressed, please take another look.

svgeesus (approved):
LGTM

Issue #7603, opened 2022-08-14
[css-contain-2] Incomplete event definition for ContentVisibilityAutoStateChanged


tidoust:
Spec says that a `ContentVisibilityAutoStateChanged` event gets fired at an element. But `ContentVisibilityAutoStateChanged` is the name of the interface used by the event, not the name of the event itself. What is the name of the event that fires at the element?

Ideally, the spec would use the [fire an event](https://dom.spec.whatwg.org/#concept-event-fire) concept defined in DOM with wording similar to:

> When the rendering state of an element with [content-visibility: auto](https://drafts.csswg.org/css-contain-2/#propdef-content-visibility) style changes and the element either becomes or stops being [relevant to the user](https://drafts.csswg.org/css-contain-2/#relevant-to-the-user), the user agent must [fire an event](https://dom.spec.whatwg.org/#concept-event-fire) named `contentvisibilityautostatechange` at the `Element` using `ContentVisibilityAutoStateChanged`.

Also, if you're firing a new event on the `Element` interface, that interface should be extended to add an `onevent` event handler IDL attribute to meet the [Always add event handlers](https://w3ctag.github.io/design-principles/#always-add-event-handlers) design principle.

Loirooriol:
I mean, this is not clear from the top comment, but as stated in #7740

> Apart from a few exceptions in old specs, events are always named with verbs in the present tense. Accordingly, this update switches the event type to `contentvisibilityautostatechange` instead of `contentvisibilityautostatechanged` and also renames the IDL interfaces. For what it's worth, this guidance does not yet appear in the TAG's API design principles but is considered for inclusion in https://github.com/w3ctag/design-principles/issues/280

In particular, there is the [`change` event](https://html.spec.whatwg.org/multipage/indices.html#event-change) which is not called `changed`.

svgeesus:
Due to a typo, this is not marked as resolved. From the minutes:

> <fantasai> RESOVLED: Change to present tense for ContentVisiblityAutoStateChanged (event and object)

should be

<fantasai> RESOLVED: Change to present tense for ContentVisiblityAutoStateChanged (event and object)

svgeesus:
Also due to the merged PRs, is this still really **Needs Edits**?

Pushing on this because [RESOLVED: Accept the PR and republishing WD of containment 2](https://github.com/w3c/csswg-drafts/issues/6900)

vmpstr:
I believe this can be closed

[...]

PR #7740, opened 2022-09-14
[css-contain-2] Improve definition of contentvisibilityautostatechange event

tidoust:
A [recent commit](https://github.com/w3c/csswg-drafts/commit/c3d88b28e594a665aca14b6a8fea7014e8dd8430) improved the definition of the event somehow but the link between the `contentvisibilityautostatechanged` event and the `ContentVisibilityAutoStateChanged` is not specified anywhere. This update switches to the usual phrasing used across specs for firing events to make the link between the event type and the interface explicit.

Also, Apart from a few exceptions in old specs, events are always named with verbs in the present tense. Accordingly, this update switches the event type to `contentvisibilityautostatechange` instead of `contentvisibilityautostatechanged` and also renames the IDL interfaces. For what it's worth, this guidance does not yet appear in the TAG's API design principles but is considered for inclusion in https://github.com/w3ctag/design-principles/issues/280

This partially addresses concerns raised in #7603. The spec would still need to define an `oncontentvisibilityautostatechange` event handler IDL attribute, which could perhaps be done by extending the `GlobalEventHandlers` mixin (to target `HTMLElement`, `MathMLElement` and `SVGElement` all at once) or by working with the WHATWG to integrate the change in HTML directly.

### DIFF
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -1690,41 +1690,43 @@ Using ''content-visibility: auto''</h3>
 		will be forced to rely on the ''200px'' height estimate.
 	</div>
 
-Detecting ''content-visibility: auto'' state changes: the <a event>contentvisibilityautostatechanged</a> event {#content-visibility-auto-state-changed}
+Detecting ''content-visibility: auto'' state changes: the <a event>contentvisibilityautostatechange</a> event {#content-visibility-auto-state-changed}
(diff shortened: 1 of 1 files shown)


paulirish:
+1  This name change went under the radar. And Chrome 108 shipped with `contentvisibilityautostatechange`.

Apetree100122 (approved):

svgeesus:
(kicking the bots)

Issue #7439, opened 2022-06-30
[css-values] add 'revert' and 'revert-layer' to list of CSS-wide keywords


dbaron:
css-values [lists](https://www.w3.org/TR/css-values-4/#common-keywords) the CSS-wide keywords.  This definition is important because it has implications for things like what is allowed in [custom identifiers](https://www.w3.org/TR/css-values-4/#custom-idents).

css-cascade-4 [adds the 'revert' keyword](https://www.w3.org/TR/css-cascade-4/#default) and css-cascade-5 [adds the 'revert-layer' keyword](https://www.w3.org/TR/css-cascade-5/#revert-layer).  Should these be added to the list of CSS-wide keywords?

I'd note that Blink already [mostly implements](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/properties/css_parsing_utils.cc;l=3064-3076;drc=901d0221a10d1321a0870b2e07f97839b2f251cc) this change (modulo [a bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1340993)).

I haven't checked other implementations.

tabatkins:
Yes, they should be.

yisibl:
I have a similar issue: https://github.com/w3c/csswg-drafts/issues/7170

PR #7763, opened 2022-09-19
[css-values-4] Add `revert` and `revert-layer` to CSS-wide keywords.

yisibl:
Fixes: #7439

@dbaron @tabatkins PTAL.


fantasai:
We took a different approach in https://github.com/w3c/csswg-drafts/commit/9b95cab15a325abb5934590e85c830811ba1ba70 to avoid having definitions split awkwardly between the specs, so closing out this PR since the issue is now fixed. :)

PR #7801, opened 2022-09-28
Fix the auto-builder redirect pages to maintain the URL fragment

andreubotella:
In the https://w3c.github.io/csswg-drafts/ mirror, if you append a spec's shortname without a level to the URL (https://w3c.github.io/csswg-drafts/css-grid, for example), you land on a page that redirects you to the current work of that spec. However, if you add a fragment to the URL, that fragment goes away in the redirect. This is an issue for MDN, for example.

The fragment goes away because currently redirects happen with meta refresh. This change uses `<noscript>` to make sure such redirects keep working when scripting is disabled, and then does a JS-based redirect that does preserve the fragment.


andreubotella:
@tabatkins PTAL

PR #7820, opened 2022-10-03
[css-grid-2] Remove U+0323 COMBINING DOT BELOW

tgraham-antenna:
Looked like dust on the screen, but then it scrolled with the text.


dbaron:
Looks like css-grid-1 could use the same change.

PR #7823, opened 2022-10-04
[resize-observer] Remove HTML monkey patch.

### LINKED ISSUE #4824: [resize-observer] The monkey patching of the event loop processing model appears to be outdated
The entire section about [HTML event loop processing model](https://drafts.csswg.org/resize-observer-1/#html-event-loop) appears to be outdated.

I think this section is trying to monkey patch step 14 of [updating the rendering](https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering) step in the current processing model.

The spec also mentions steps like "recall styles" and "update layout". Since these aren't concepts formally defined anywhere, it's probably best to consolidate as a single step to generate or update the CSS boxes and their geometry or something instead too.

emilio:
This is now upstream.

Fixes #4824


Loirooriol (approved):
Nice!

PR #7826, opened 2022-10-04
[css-viewport-1] Add 'interactive-widgets' to viewport meta

bokand:
[css-viewport-1] Add 'interactive-widgets' to viewport meta

This describes the proposed `interactive-widgets` property of the viewport meta tag for describing the expected behavior when a virtual-keyboard-like widget is shown.

### DIFF
--- a/css-viewport/Overview.bs
+++ b/css-viewport/Overview.bs
@@ -125,7 +125,7 @@ The recognized properties in the viewport
 	<li><code class="index">minimum-scale</code></li>
 	<li><code class="index">maximum-scale</code></li>
 	<li><code class="index">user-scalable</code></li>
-	<li><code class="index">interactive-widgets</code></li>
+	<li><code class="index">interactive-widget</code></li>
 </ul>
 
 <h3 id="parsing-algorithm">
@@ -261,18 +261,18 @@ as follows:
 
 Issue: Specify extend-to-zoom behavior by the viewport meta tag
 
-<h3 id="interactive-widgets-section">''interactive-widgets''</h3>
+<h3 id="interactive-widget-section">''interactive-widget''</h3>
 
 Issue: Move the definition of ''visual viewport'' from CSSOM-View to this spec.
 
-The <dfn><code>interactive-widgets</code></dfn> property specifies the effect that interactive UI
+The <dfn><code>interactive-widget</code></dfn> property specifies the effect that interactive UI
 widgets have on the page's viewports. It defines whether widgets overlay a given viewport or whether
 the viewport is shrunken so that it remains fully visible while the widget is showing. Interactive
 UI widgets are transient user agent or operating system UI through which a user can provide input.
 
 <div class="note">The most common such UI widget is a virtual keyboard.</div>
 
-The following is a list of valid values for [=interactive-widgets=] and the associated
+The following is a list of valid values for [=interactive-widget=] and the associated
 viewport-resizing behavior:
 
 <dl>
@@ -284,24 +284,24 @@ viewport-resizing behavior:
 		as when <a href="https://www.w3.org/TR/virtual-keyboard/#dom-virtualkeyboard-overlayscontent">
 		VirtualKeyboard.overlaysContent</a> is set to <code>true</code>.
 	</dd>
-	<dt><dfn><code>resize-layout</code></dfn></dt>
+	<dt><dfn><code>resizes-document</code></dfn></dt>
 	<dd>
 		Interactive UI widgets MUST [=resize=] the [=initial viewport=] by the interactive widget.
 	</dd>
 	<div class="note">
 		Since the <a spec="CSSOM-VIEW">visual viewport</a>'s size is derived from the
-		[=initial viewport=], [=resize-layout=] will cause a resize of both the initial and visual
+		[=initial viewport=], [=resizes-document=] will cause a resize of both the initial and visual
 		viewports.
 	</div>
-	<dt><dfn><code>resize-visual</code></dfn></dt>
+	<dt><dfn><code>resizes-visual</code></dfn></dt>
 	<dd>
 		Interactive UI widgets MUST [=resize=] the <a spec="CSSOM-VIEW">visual viewport</a> but MUST
 		NOT [=resize=] the <a spec="CSS-VIEWPORT">initial viewport</a>.
 	</dd>
 </dl>
 
-If no value, or an invalid value, is set for [=interactive-widgets=], the behavior implied by
-[=resize-visual=] is used as the default.
+If no value, or an invalid value, is set for [=interactive-widget=], the behavior implied by
+[=resizes-visual=] is used as the default.
 
 To <dfn lt="resize">resize a viewport by an interactive widget</dfn>, subtract from it the
 intersection of the viewport rect with the widget's OS reported bounding rect. In cases where this
@@ -321,14 +321,14 @@ would result in a non-rectangular viewport, the behavior is user agent defined.
 The virtual-keyboard API provides an imperitive API to apply the [=overlays-content=] behavior via
 the <a href="https://www.w3.org/TR/virtual-keyboard/#dom-virtualkeyboard-overlayscontent">
 virtualKeyboard.overlaysContent</a> attribute. This attribute shadows the value set to
-[=interactive-widgets=], namely:
+[=interactive-widget=], namely:
 
 When <code>virtualKeyboard.overlaysContent</code> is set to <code>true</code>, the UA MUST ignore
-any value set to [=interactive-widgets=] when determining the resizing behavior of interactive
+any value set to [=interactive-widget=] when determining the resizing behavior of interactive
 widgets.
 
 When <code>virtualKeyboard.overlaysContent</code> is set to <code>false</code>, the UA MUST use the
-value set to [=interactive-widgets=], or the default behavior if a value is not set, when
+value set to [=interactive-widget=], or the default behavior if a value is not set, when
 determining the resizing behavior of interactive widgets.
 
 Getting the value of <code>virtualKeyboard.overlaysContent</code> MUST return only the value previously
@@ -336,7 +336,7 @@ set to it.
 
 <div class="note">
 	That is, unless previously set, <code>virtualKeyboard.overlaysContent</code> returns false
-	even if <code>interactive-widgets=overlays-content</code> is set via the <code>&ltmeta&gt</code>
+	even if <code>interactive-widget=overlays-content</code> is set via the <code>&ltmeta&gt</code>
 	tag.
 </div>
 
--- a/css-viewport/Overview.bs
+++ b/css-viewport/Overview.bs
@@ -289,13 +289,13 @@ viewport-resizing behavior:
 		as when {{VirtualKeyboard/overlaysContent|VirtualKeyboard.overlaysContent}} is set to
 		<code>true</code>.
 	</dd>
-	<dt><dfn><code>resizes-document</code></dfn></dt>
+	<dt><dfn><code>resizes-content</code></dfn></dt>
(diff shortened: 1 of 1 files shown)


bokand:
@emilio here's a first crack, WDYT?

One question on naming: "interactive-widget*s*" is plural so implies "resize-visual" rather than "resizes-visual" but the virtual keyboard API has [`overlaysContent`](https://www.w3.org/TR/virtual-keyboard/#dom-virtualkeyboard-overlayscontent) which I wanted to keep consistent. Perhaps `interactive-widget=resizes-visual` would be preferable?

I also had some issues linking to defs in the virtual-keyboard spec, I'll keep looking at that but if you can spot the issue help would be appreciated.

bokand:
Also, unrelated to this specific PR but if no one's signed up for it, I can try and find some time before end of year to specify viewport meta more generally. PLMK

css-viewport/Overview.bs
```diff
@@ -260,6 +261,85 @@ as follows:
...
+
+When <code>virtualKeyboard.overlaysContent</code> is set to <code>true</code>, the UA MUST ignore
+any value set to [=interactive-widgets=] when determining the resizing behavior of interactive
+widgets.
+
+When <code>virtualKeyboard.overlaysContent</code> is set to <code>false</code>, the UA MUST use the
+value set to [=interactive-widgets=], or the default behavior if a value is not set, when
+determining the resizing behavior of interactive widgets.
+
+Getting value of <code>virtualKeyboard.overlaysContent</code> MUST return only the value previously
```
chrishtr, line 334:
```suggestion
Getting the value of <code>virtualKeyboard.overlaysContent</code> MUST return only the value previously
```

css-viewport/Overview.bs
```diff
@@ -260,6 +261,85 @@ as follows:
...
+
+The following is a list of valid values for [=interactive-widgets=] and the associated
+viewport-resizing behavior:
+
+<dl>
+	<dt><dfn><code>overlays-content</code></dfn></dt>
+	<dd>
+		Interactive UI widgets MUST NOT [=resize=] the [=initial viewport=] nor
+		the <a spec="CSSOM-VIEW">visual viewport</a>. The user agent must perform the same steps
+		<!--TODO I can't seem to autolink to overlaysContent, help?-->
```
emilio, line 283: @tabatkins can you help with this?

css-viewport/Overview.bs
```diff
@@ -260,6 +261,85 @@ as follows:
...
+<h4 id="interaction-with-virtualkeyboard-overlayscontent">
+	Interaction with virtualKeyboard.overlaysContent
+</h4>
+
+The virtual-keyboard API provides an imperitive API to apply the [=overlays-content=] behavior via
+the <a href="https://www.w3.org/TR/virtual-keyboard/#dom-virtualkeyboard-overlayscontent">
+virtualKeyboard.overlaysContent</a> attribute. This attribute shadows the value set to
+[=interactive-widgets=], namely:
+
+When <code>virtualKeyboard.overlaysContent</code> is set to <code>true</code>, the UA MUST ignore
```
emilio, line 326: This seems like a rather odd thing (that is, that the default for VirtualKeyboard.overlaysContent is different from the default for the meta).

This might be sensible behavior given the current set of things, but maybe it'd be nice to change that, either by changing the `overlaysContent` default or allowing it to be a tri-state or something...

Can you file an issue about that and reference it here perhaps?

css-viewport/Overview.bs
```diff
@@ -260,6 +261,85 @@ as follows:
...
+<dl>
+	<dt><dfn><code>overlays-content</code></dfn></dt>
+	<dd>
+		Interactive UI widgets MUST NOT [=resize=] the [=initial viewport=] nor
+		the <a spec="CSSOM-VIEW">visual viewport</a>. The user agent must perform the same steps
+		<!--TODO I can't seem to autolink to overlaysContent, help?-->
+		as when <a href="https://www.w3.org/TR/virtual-keyboard/#dom-virtualkeyboard-overlayscontent">
+		VirtualKeyboard.overlaysContent</a> is set to <code>true</code>.
+	</dd>
+	<dt><dfn><code>resize-layout</code></dfn></dt>
```
emilio, line 287: So https://github.com/web-platform-tests/interop-2022-viewport/issues/25 is a discussion about whether the "layout viewport" should be called something else... Maybe we can give it a bit more thought and figure out whether we want to change that? If we do want to change that, it seems using here the layout term would be unfortunate...

emilio (approved):
Let's fix the links before landing.

It seems reasonable. I'm not a fan of the naming but this is much better than virtual-keyboard, and I don't have any better suggestions off the top of my head. One question on the naming of `resize-layout` tho.

css-viewport/Overview.bs
bokand: Hmm, how about "resize-initial" since that's what the spec text refers to

css-viewport/Overview.bs
bokand: I'm not sure I understand what you mean by "default". The default value for `overlaysContent` is `false` and the default for the meta is `resize-visual` which is also not `overlays-content`. That seems consistent to me.

They aren't linked, i.e. if you have `<meta name="viewport" content="interactive-widgets=overlays-content">`, `virtualKeyboard.overlaysContent` will still return false unless the author sets it explicitly. Is that what you're referring to?

bokand:
> One question on naming: "interactive-widgets" is plural so implies "resize-visual" rather than "resizes-visual" but the virtual keyboard API has [overlaysContent](https://www.w3.org/TR/virtual-keyboard/#dom-virtualkeyboard-overlayscontent) which I wanted to keep consistent. Perhaps interactive-widget=resizes-visual would be preferable?

@emilio Thoughts on this? Or do you think it's ok as-is?

css-viewport/Overview.bs
bramus: `resize-initial` seems weird to me, as that reminds me of the initial in ICB, whereas this targets the LVP. Just commented on https://github.com/web-platform-tests/interop-2022-viewport/issues/25 and suggested “Document Viewport” to further specify it without introducing a new ambiguous term. Should suggestion make it in, maybe `resize-document` could be considered?

emilio:
> > One question on naming: "interactive-widgets" is plural so implies "resize-visual" rather than "resizes-visual" but the virtual keyboard API has [overlaysContent](https://www.w3.org/TR/virtual-keyboard/#dom-virtualkeyboard-overlayscontent) which I wanted to keep consistent. Perhaps interactive-widget=resizes-visual would be preferable?
> 
> @emilio Thoughts on this? Or do you think it's ok as-is?

I don't feel strongly either way. If you think resizes- is more consistent that works for me.

css-viewport/Overview.bs
emilio: Ah yeah, you're right, I had misunderstood what overlaysContent=true meant. Nvm then. Making the API support the three states may be worth it but also is kind of orthogonal to this.

css-viewport/Overview.bs
bokand: Well, it does resize the ICB which I think is the most salient part of its behavior (it also happens to resize the fixed viewport).

That said, "Document Viewport" actually sgtm, will reply on that issue.

css-viewport/Overview.bs
bokand: Got some help with this internally - added an anchors block at the top so autolinks to `overlaysContent` are working now.

css-viewport/Overview.bs
bokand: There was a good counter point to "document viewport" on that issue but I also realized that this really does resize more than _just_ this viewport. In skobe's example page on https://github.com/web-platform-tests/interop-2022-viewport/issues/25, this mode will resize both LV and ICB. This really does resize the "initial viewport", though I agree it's not a great author-facing name.

How about we don't make it "viewport" specific and describe it in less precise but terms more understandable to authors? Some other ideas:

* `resizes-content` works well as a contrast to `overlays-content` I think
* `resizes-container`
* `resizes-window`

bramus: No hard feelings about `resizes-content`, although technically your content does not get resized. I mean, your content will still have the same size when the OSK gets shown, unless you have them sized to specific viewport units.

Definitely not a fan of `resizes-container` given we have [Container Queries in CSS](https://drafts.csswg.org/css-contain-3/#container-queries). This would lead to confusion.

Not a huge fan of `resizes-window` as it does not actually resize the app’s window, right? I would think an app _(on mobile)_ to still run at the same size as on launch when the OSK gets shown. Is this the case? [By the looks of it](https://user-images.githubusercontent.com/213073/191926687-57300f61-9aa9-409d-a9e5-a096bc96c19e.png), this is also the behavior on Windows _(at least for Chrome)_.

bokand: Good points about `container` and `window`.

> I mean, your content will still have the same size when the OSK gets shown, unless you have them sized to specific viewport units.

I'd say it has the _potential_ to resize content since it resizes the initial containing block.

@emilio WDYT?

bramus:
A late remark but I’m seeing some potential confusion between the term “Interactive UI widgets” _(used here)_ and “UA interfaces that are dynamically expanded and retracted” _(used in the definition for [the various Viewports and their associated lengths](https://drafts.csswg.org/css-values-4/#viewport-relative-lengths))_. After all, a Virtual Keyboard is also “an interface that is dynamically expanded and retracted” _(albeit from the OS – not the UA)_.

Taking a step back: we approached this from the virtual keyboard and therefore first named the extension `virtual-keyboard` + loaned the value `overlays-content` from the VK API. As it can be more than just the Virtual Keyboard _(can it?)_, it got renamed to `interactive-widgets` and later `interactive-widget` as we continued down that path.

Another approach: what if we called it `resize-behavior`? This because it allows you to define the resize behavior of the viewport(s). Values would be `document-viewport`, `visual-viewport`, and `none` _(former `overlays-content`)_. Spec can then later always find a better term for “Interactive UI widgets” should that be confusing to authors _(which I think it will be)_.

Using `resize-behavior` would also keep the language consistent: all other allowed options of the viewport meta tag (e.g. `initial-scale`) use the viewport as the subject. With `resize-behavior` this is also the case, whereas with `interactive-widget` it is not.

It would also leave the door open for something like `resize-triggers`, should we ever want to allow authors to tweak what exactly can cause a resize.

css-viewport/Overview.bs
emilio: `resizes-content` works for me... It might be worth raising this in the next meeting to get wider feedback tho.

bokand:
> As it can be more than just the Virtual Keyboard (can it?), it got renamed to interactive-widgets and later interactive-widget as we continued down that path.

This is intended to apply only to virtual-keyboard and virtual-keyboard-like UI - I'm not sure if any exists today but something that's similar to today's keyboards but not necessarily exactly a keyboard (e.g. voice interpreter, handwriting panel, etc. that behave equivalently to the "regular" virtual keyboard).

> Another approach: what if we called it resize-behavior? This because it allows you to define the resize behavior of the viewport(s). Values would be document-viewport, visual-viewport, and none (former overlays-content)

My main push back is that this property isn't a general resizing mechanism, it applies only to virtual-keyboards. i.e. none of URL bar, bottom bars, split screen, etc. will respect this property so I think `resize-behavior` will be somewhat misleading.

This addition is mainly intended to pave a path to an interoperable behavior for virtual-keyboards. I think if we want to design a more complete and general API later we can (and should!) still do that.

bokand:
Thanks all, I think that's everything resolved.

css-viewport/Overview.bs
nongaum2112: Can i follow!!

PR #7840, opened 2022-10-06
[css-page-floats-3] Add missing CSS2 'both' keyword to 'clear' grammar

### LINKED ISSUE #7350: [css-page-floats-3][css-box-3] Move definitions of `float` and `clear`
`both`, `both-inline`, `both-block`, are not included in the `value` of the definition table for the [`clear`](https://drafts.csswg.org/css-page-floats-3/#clearing_page_floats) property in CSS Page Floats 3, which is currently the primary `value` definition for `clear`, but are defined in prose below the `clear` definition table.

**EDIT** 

According to this [comment](https://github.com/w3c/csswg-drafts/pull/7840#issuecomment-1270225771), `clear` and `float` should be defined in CSS Box 3 (or *a module for block layout*) therefore I change my demand by updating the title of this issue, and by asking for the following changes in CSS Page Floats 3 (after defining `float` and `clear` in another spec):

```diff
  Name: `float`
- Value: `block-start | block-end | inline-start | inline-end | snap-block | <snap-block()> | snap-inline | <snap-inline()> | left | right | top | bottom | none`
+ New values: `block-start | block-end | inline-start | inline-end | snap-block | <snap-block()> | snap-inline | <snap-inline()> | top | bottom`

  Name: `clear`
- Value: `inline-start | inline-end | block-start | block-end | left | right | top | bottom | none`
+ New values: `inline-start | inline-end | block-start | block-end | top | bottom`
```

cdoublev:
Fixes #7350.

There has been no update to the spec since 1 year but I would hope that at least `both`, which is [defined in CSS 2](https://w3c.github.io/csswg-drafts/css2/#propdef-clear), could be re-included in the `value` definition of `clear`, for backward-compatibility, noting that: 

1. `both-inline` and `both-block` are also defined in prose in CSS Page Floats 3 but are missing in the `value` definition of `clear`
2. `both` is defined in CSS Page Floats 3 as equivalent to `both-inline`, which is defined as equivalent to `inline-start` and `inline-end`

### DIFF
--- a/css-page-floats-3/Overview.bs
+++ b/css-page-floats-3/Overview.bs
@@ -464,7 +464,7 @@ The 'clear' property</h2>
 
   <pre class="propdef">
     Name: clear
-    Value: inline-start | inline-end | block-start | block-end | left | right | top | bottom | none
+    Value: inline-start | inline-end | block-start | block-end | both | left | right | top | bottom | none
     Initial: none
     Applies to: block-level elements, floats, regions, pages
     Inherited: no


dbaron:
I don't think css-page-floats-3 should be presenting itself as the primary definition of the `float` and `clear` properties.  This is a set of additions to those properties.  The primary definitions should be in a different level 3 module:  maybe css-box-3, or maybe a module for block layout.

Issue #7785, opened 2022-09-23
[css-view-transitions-1] Define "active animation" used to signal end of transition


khushalsagar:
Shared Element Transition creates an ephemeral tree of elements with snapshots from incoming and outgoing DOM state. We need to define the lifetime of this tree. Since the purpose of this tree is to permit animations on these elements, we rely on the existence of the animations to decide how long this tree lives.

For declarative animations (CSS transitions, CSS animations and Web Animations) which can be detected by the browser, the [spec] (https://drafts.csswg.org/css-shared-element-transitions-1/#issue-9b6c2555) talks about detecting when any of the generating elements have active animations. We need to define what active means here.

Each of the animation APIs above provide an event that the developer can use to detect if the animation was finished or cancelled:

* CSS Transition : [transitionend](https://drafts.csswg.org/css-transitions/#transitionend) and [transitioncancel](https://drafts.csswg.org/css-transitions/#transitioncancel).
* CSS Animation : [animationend](https://w3c.github.io/csswg-drafts/css-animations-1/#eventdef-globaleventhandlers-animationend) and [animationcancel](https://w3c.github.io/csswg-drafts/css-animations-1/#eventdef-globaleventhandlers-animationcancel).
* Web Animation : finish, cancel and remove events as defined [here](https://w3c.github.io/csswg-drafts/web-animations-1/#animation-playback-event-types).

We want to ensure that these events are fired before assuming that an element has no active animation. This allows the developer to chain animations by waiting for one to finish before triggering the next. The precise proposal is as follows:

* Immediately after the transition elements are generated a [ready promise](https://drafts.csswg.org/css-shared-element-transitions-1/#domtransition-ready-promise) is resolved to notify the developer that these elements can be targeted in script. Wait until this event is fired.
* Then check that none of the generated elements have an animation which is waiting to start, is currently running or is pending dispatch of finished or cancel events. Since checking if an element has an animation waiting to start requires a clean style the check may be deferred until the next rendering opportunity if the style is dirty. Similarly this check runs after dispatch of the finished/cancel animation events.
* Once there is a check when none of the elements have an animation in the state above, the transition is assumed to be finished and generated elements are removed.

I'm not sure if the above can be simplified by running the check at the end of every [update the rendering](https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering) loop. That loop might not be triggered after the animationend event (if there are no new updates to draw?). Ideas welcome on that.

[...]

PR #7857, opened 2022-10-07
Clarify checking for active animations

khushalsagar:
[css-view-transitions-1] Clarify checking for active animations.

See discussion at https://github.com/w3c/csswg-drafts/issues/7785

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -893,15 +893,21 @@ The {{ViewTransition/domUpdated}} [=getter steps=] are to return [=this's=] [=Vi
     To <dfn>update transition DOM</dfn> given a {{ViewTransition}} |transition|:
 
     1. Let |document| be |transition|'s [=relevant global object's=] [=associated document=].
+	
+	1. Let |hasActiveAnimations| be a boolean, initially false. For each [=page-transition pseudo-elements=]
+        associated with |transition|:
+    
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -893,15 +893,21 @@ The {{ViewTransition/domUpdated}} [=getter steps=] are to return [=this's=] [=Vi
...
-        Issue: There needs to be a definition/link for "active animation".
+	
+	1. Let |hasActiveAnimations| be a boolean, initially false. For each [=page-transition pseudo-elements=]
+        associated with |transition|:
+    
+        1. Let |element| be the [=page-transition pseudo-element=].
+    
+        1. For each |animation| with [=effect target=] set to |element| and [=animation timeline=]
+            of type [=document timeline=], set |hasActiveAnimations| to true if any of the following
+            conditions is true:
```
birtles, line 904: Animation's don't have an effect target. That's a property of keyframe effects.

I think we could use wording similar to that for `getAnimations()`, e.g.

```bikeshed
1. For each |animation| that contains at least one [=animation effect=]
   whose [=effect target=] is |element| and
   whose associated [=animation timeline=] is of type [=document timeline=], ...
```

birtles (approved):
Looks good to me with the suggested change applied

css-view-transitions-1/Overview.bs
khushalsagar: Done.

PR #7886, opened 2022-10-16
Make GH Pages deployment create symlinks rather than redirects

sideshowbarker:
This change makes the GH pages deployment create symlinks for all un-leveled shortnames, pointing to the appropriate leveled shortnames — eliminating the need for the web server to do any rewrites, and eliminating the need for any redirects (either server- or client-side). Thus, it replaces the existing build mechanism that generated index.html files containing client-side redirects.

https://sideshowbarker.github.io/csswg-drafts/ has the results of a working deployment of the change in this PR branch.


andreubotella (approved):
LGTM

svgeesus (approved):
LGTM

PR #7890, opened 2022-10-17
Fix description and sample code for "longer"

brianosman:
The previous algorithm was incorrect, producing sweeps that were either too short or too long. Also, the description of the post-conditions was wrong.

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -4709,13 +4709,13 @@ Hue Interpolation</h3>
 <h4 id="hue-longer">
 <dfn export>longer</dfn></h4>
 
-	Angles are adjusted so that |θ₂ - θ₁| ∈ {0, [180, 360)}. In pseudo-Javascript:
+	Angles are adjusted so that |θ₂ - θ₁| ∈ {(-360, -180], [180, 360]}. In pseudo-Javascript:
 		<pre>
 		if (0 < θ₂ - θ₁ < 180) {
-			θ₂ += 360;
+			θ₁ += 360;
 		}
 		else if (-180 < θ₂ - θ₁ <= 0) {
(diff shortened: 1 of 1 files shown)


mysteryDate:
Makes sense to me!

For red (theta = 0) to green (theta = 120), the interpolation should go the long way, around blue. If we're gonna accomplish this by adding 360 to one of the angles, it obviously needs to be theta1.

svgeesus:
@LeaVerou

css-color-4/Overview.bs
```diff
@@ -4709,13 +4709,13 @@ Hue Interpolation</h3>
 <h4 id="hue-longer">
 <dfn export>longer</dfn></h4>
 
-	Angles are adjusted so that |θ₂ - θ₁| ∈ {0, [180, 360)}. In pseudo-Javascript:
+	Angles are adjusted so that |θ₂ - θ₁| ∈ {(-360, -180], [180, 360]}. In pseudo-Javascript:
```
LeaVerou, line 4712:
```suggestion
	Angles are adjusted so that |θ₂ - θ₁| ∈ {[-360, -180], [180, 360]}. In pseudo-Javascript:
```

LeaVerou:
We need to make further edits to fix the normalization algorithm, but since this is an improvement over the current spec, let's merge it.

css-color-4/Overview.bs
brianosman: I don't think it's actually possible to end up with theta2-theta1 == -360, which is why I wrote it the way I did. They can't be in that configuration before the `longer` logic runs, due to the first normalization pass. The first condition is a strict inequality at zero, so if the angles are initially equal, you'll end up in the second case (adding 360 to theta2).

PR #7894, opened 2022-10-17
[css-backgrounds-3] Use new #? multiplier in background

cdoublev:
Non-substantive change.


cdoublev:
I had thought that it could help to simplify the structure of the data representation resulting from automatic parsing against this value definition, but it remains the same so I give up this PR.

Issue #7750, opened 2022-09-15
[css-color-4] Hue interpolation "specified" should be removed


fserb:
Although it sounds like a reasonable and simple idea, the interpolation method called `"specified"` by the spec is actually extremely costly and complex to be implemented by UAs. The reason for it is that is breaks a basic assumption about colors, namely, that they (apart from some loss of precision) can be represented and converted between different systems. 

UAs usually choose an internal color representation (like sRGBA). Hue-based systems (like HSL, HWB) are, in practice, expected to have Hues defined in the `[0, 360)` region. Outside of that region, colors can be converted out of them, but not preserved. To allow internal representation of arbitrary Hues, UAs need to have special code to store those types of color representation throughout their pipelines.

We can see a problem caused by this on the current Safari TP implementation of `hue specified` gradients, where Safari follows the proper algorithm (numeric interpolation of hue values), but fails to preserve their value outside of `mod 360` (since they are internally converted to sRGB), creating a weirdly unspecified behavior, that doesn't address the original spirit of the spec (the ability to have multiple rotations on the hue circle in a single gradient interpolation).

Considering that it does seem like a reasonable request, but also understanding that there's no pressing use case for it and that there's a big implementation complexity, we'd like to request that this small section of the spec to be removed from CSS Color 4 and then re-discussed at a later time.

weinig:
Do you have a test case of the issue with gradients you are describing?

tabatkins:
I don't have access to a Safari to verify this, but the testcase would be to use a "specified" interpolation to go from, say, `hsl(0deg 100% 50%)` to `hsl(720deg 100% 50%)` - this should produce two full sweeps thru the hue wheel, per spec.

weinig:
Oh, interesting, I hadn't realized how unrestrictive the hue notion was. Makes me curious if we are getting serialization wrong as well, as that also ends up being `% 360`.

weinig:
I don't think this will be a problem for WebKit. The current behavior is just a bug, due to a misreading or misunderstanding of the spec on my part. 

I agree this poses a little problem for hsl() and hwb() specifically, as we do currently store them as 8-bit per channel sRGB when no component is "none", but I don't believe that optimization is really doing anything and is more of a historical oddity. We can just as easily move that conversation to serialization time.

I filed https://bugs.webkit.org/show_bug.cgi?id=245552 to track fixing this in WebKit.

weinig:
[...]

PR #7895, opened 2022-10-17
[css-color-4] Define hue interpolation angle precisely, maintain differences of 360deg

fantasai:


css-color-4/Overview.bs
```diff
@@ -4671,16 +4671,25 @@ Interpolating with Alpha</h3>
...
 	(angles in the following are in degrees, but the logic is the same regardless of how they are specified).
 	Specifying a hue interpolation strategy is already part of the <<color-interpolation-method>> syntax
 	via the <<hue-interpolation-method>> token.
 
-	Unless the type of hue interpolation is ''specified'', both angles need to be constrained to [0, 360) prior to interpolation.
-	One way to do this is <code><i>θ</i> = ((<i>θ</i> % 360) + 360) % 360</code>.
+	Unless the type of hue interpolation is ''specified'',
+	both angles and their difference need to be constrained to [0, 360] prior to interpolation.
+	To do this, the minimum number of turns that fit in the lesser angle
+	is added subtracted from both angles,
```
svgeesus, line 4686: added? subtracted?

css-color-4/Overview.bs
```diff
@@ -4671,16 +4671,25 @@ Interpolating with Alpha</h3>
...
+	bringing the lesser angle into the range [0,360);
+	and if the difference between them is greater than 360&deg;
+	then the minimum number of turns to bring the difference into the range [0,360]
+	is further subtracted from the greater angle.
+
+	Note: Differences of 360&deg; are allowed to maintain the originally specified direction of a full turn.
```
svgeesus, line 4692: But once normalized, a difference of 360 is no longer possible. So this has to start with the specified value. See https://github.com/w3c/csswg-drafts/issues/7750

css-color-4/Overview.bs
LeaVerou, line 4686:
```suggestion
	is subtracted from both angles,
```

PR #7898, opened 2022-10-18
[gh-pages] Generate timestamps.json file w/ git timestamps for sources

sideshowbarker:
For the GH Pages deployment, this change generates a timestamps.json file that maps each shortname in the repo to a timestamp representing the time for the most recent commit of corresponding Bikeshed source file for that shortname — enabling any downstream users/consumers of the deployed GH Pages specs to have accurate last-committed data for all sources.

I have immediate use for this timestamp data for the https://html-now.github.io/ dashboard and can imagine that others might end up having good use for the timestamp data as well.

### DIFF
--- a/bin/build-index.py
+++ b/bin/build-index.py
@@ -9,6 +9,7 @@ import os
 import os.path
 import re
 import sys
+import subprocess
 from collections import defaultdict
 
 from html.parser import HTMLParser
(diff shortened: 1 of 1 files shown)


svgeesus (commented):
LGTM

andreubotella (approved):
Could this file be used to set the dates in the spec headers to the last time each spec was updated, rather than to the last time the build was run? Although that could be done by getting the git timestamp for each spec in the build and calling bikeshed with `--md-date=$GIT_TIMESTAMP`, independently of this PR.

LGTM otherwise.

sideshowbarker:
> Could this file be used to set the dates in the spec headers to the last time each spec was updated, rather than to the last time the build was run? Although that could be done by getting the git timestamp for each spec in the build and calling bikeshed with `--md-date=$GIT_TIMESTAMP`, independently of this PR.

Right — if we want to do that (and it sounds like a good idea to me), we can do it in a follow-up PR

sideshowbarker:
Given the LGTMs — and as this only affects the GH Pages deployment — going ahead and merging. But if there’s anything I’ve overlooked about this, we can always revert it if needed, post-merge

PR #7903, opened 2022-10-18
[css-view-transitions-1] Make the ready promise rejection conditional, since it may have already been resolved.

vmpstr:
[css-view-transitions-1] Make the ready promise rejection conditional, since it may have already been resolved.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -850,7 +850,7 @@ The {{ViewTransition/domUpdated}} [=getter steps=] are to return [=this's=] [=Vi
 
     1. Set |document|'s [=document/active DOM transition=] to null.
 
-    1. [=Reject=] |transition|'s [=ViewTransition/ready promise=] with |reason|.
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -850,7 +850,7 @@ The {{ViewTransition/domUpdated}} [=getter steps=] are to return [=this's=] [=Vi
 
     1. Set |document|'s [=document/active DOM transition=] to null.
 
-    1. [=Reject=] |transition|'s [=ViewTransition/ready promise=] with |reason|.
+    1. If |transition|'s [=ViewTransition/ready promise=] has not yet been resolved, [=reject=] it with |reason|.
```
khushalsagar, line 853:
```suggestion
    1. If |transition|'s [=ViewTransition/ready promise=] has not yet been resolved, [=reject=] it with |reason|.
       Note: The ready promise would've been resolved if skipTransition() is called after we start animating.
```

khushalsagar (approved):

tabatkins:
This isn't necessary; rejecting an already-resolved promise is perfectly valid and simply a no-op. The observable behavior is identical both before and after this change.

PR #7966, opened 2022-10-27
[css-nesting-1] Update Overview.bs

romainmenke:
These examples were modified to illustrate that `&` can be omitted but the source was not updated.

_ironically this also illustrates that the shift to proposal 3 also makes it easier to make small mistakes :D_

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -53,10 +53,10 @@ Motivation</h3>
 			table.colortable td {
 				text-align:center;
 			}
-			table.colortable td.c {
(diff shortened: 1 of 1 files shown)


tabatkins:
No, I didn't intend to modify the example. I was just quickly removing some unnecessary &s, to avoid implying they're required, and got a little overzealous. I've fixed this properly.

> ironically this also illustrates that the shift to proposal 3 also makes it easier to make small mistakes :D

Not in particular; you won't normally be taking code written for a "required to start with &" syntax and switching it back to an implicit &.

PR #7985, opened 2022-10-31
[css-images-4] Simplify linear-gradient syntax

cdoublev:
`?` is not required in `[ <angle> | to <side-or-corner> ]? || ...` because *a double bar (`||`) separates two or more options: one or more of them must occur, in any order.*

https://drafts.csswg.org/css-values-4/#component-combinators


cdoublev:
I was confused by this weird trick to make the whole combination optional: `[... || ...]?` seems more simple to understand than `[...]? || ...`.

Issue #6915, opened 2021-12-28
[selectors] `:lang` for documents without content language and for elements of unknown language; consider `:lang("")` over `:not(:lang("*"))`


myfonj:
### Brief questions to answer

1. Is `:lang("*")` really valid selector?  (Safari supports it and Chrome accepted [Issue 1281157](https://bugs.chromium.org/p/chromium/issues/detail?id=1281157) to implement it.)
2. How to address document that failed to define it's content language?
3. Should be HTML `<el lang=""></el>` or equivalent `<el lang></el>` matched by CSS `:lang("")` or even `:lang()`?  (Not yet proposed nor implemented.)
4. Is erroneous "undefined" document language equivalent of `<html lang="">` or is it something different?

### Trivia

1. It seems that there have never been a way for `lang()` functional pseudo-class to precisely target document without defined content language nor element sub-tree set as such with `lang=""` attribute.
2. [Selectors-4 draft introduces wildcard support](https://drafts.csswg.org/selectors-4/#the-lang-pseudo) ([md](https://github.com/w3c/csswg-drafts/blob/da9f4c11ca40e850c22dd8387e4069cb818e702e/selectors-4/Overview.bs#L1896)) in string argument for matching "any" language of given script or region group (like `"*-Latn"` or `"*-ch"`), opening possibility (loophole?) for matching *any **specified** language* value using `:lang("*")` and with conjunction with negation pseudo-class opens possibility to use `:not(:lang("*"))` for targeting elements that belong to "no specified language" from previous point.  While plain `"*"` value is not explicitly mentioned in the draft, this reportedly already works in current Safari. 
3. Undetermined language occur on any HTML document that does not have explicit `lang` attribute and does not come with `content-language` HTTP header or it's "deprecated nasty" `<meta http-equiv>` counterpart.
4. For marking ~~non-linguistic~~ __unknown language__ content it is advised to use `lang=""` attribute. While it is possible to target it with attribute selector (`[lang=""]`), is does not seem like a right tool and introduces nesting / inheritance problems that `lang()` selector was created for.

#### Sample 1: Document without content language

```
HTTP/2 200 OK
[no `content-language: xy,zz` HTTP header here]

<html [no `lang="xy"]` here>
	<head>
		[no `<meta http-equiv="content-language" content="xy,zz">` here]
	</head>
	<body>
		[I want target this document.]
	</body>
</html>
```

#### Sample 2: element with unknown language content

```HTML
<html lang="xy">
[...]

PR #8011, opened 2022-11-03
[selectors-4] Make empty language strings match untagged elements. #6915

fantasai:
Wasn't clear whether the resolution mapped both `:lang()` and `:lang("")` or just `:lang("")` ...


fantasai:
Looks like WebKit and Gecko both treat `:lang()` as invalid so going with that.

myfonj:
Maybe I am reading it wrong, but it seems all information introduced by this merge were effectively obliterated when older #4212 was merged last year, so I see no notion about `:lang("")` vs `:lang("*")` in the current version. Was the conclusion of #6915 invalidated recently?

PR #8021, opened 2022-11-05
[css-images-3] remove && from <linear-color-stop> syntax

12Me21:
w3c/csswg-drafts@c03b2ea changed `<color-stop> = <color> <length-percentage>?` to `<linear-color-stop> = <color> && <length-percentage>?`
I believe the `&&` was a mistake, since `<length-percentage> <color>` doesn't seem to be valid?

Note: This also affects [css-images-4], which uses the same definition.

### DIFF
--- a/css-images-3/Overview.bs
+++ b/css-images-3/Overview.bs
@@ -887,7 +887,7 @@ Color Stop Lists</h4>
 	<pre class=prod>
 		<dfn>&lt;color-stop-list></dfn> =
 			<<linear-color-stop>> , [ <<linear-color-hint>>? , <<linear-color-stop>> ]#
-		<dfn>&lt;linear-color-stop></dfn> = <<color>> && <<length-percentage>>?
+		<dfn>&lt;linear-color-stop></dfn> = <<color>> <<length-percentage>>?
 		<dfn>&lt;linear-color-hint></dfn> = <<length-percentage>>
 	</pre>
 


svgeesus:
@tabatkins could you review, and confirm && was a mistake?

SebastianZ:
For what it's worth, this syntax was introduced many years ago in https://github.com/w3c/csswg-drafts/commit/7bb4fbade2591eb292dc1b916cec8a868c4a3433?diff=unified#diff-f80d8fc0ac03564f1ec01e8503aeb5bea39694d9b20764727225b00510671de3R2249, initially restricted to `<cap-linear-color-stop>`. Later on it was relaxed to apply to any `<linear-color-stop>` in https://github.com/w3c/csswg-drafts/commit/321c3e550f04a683973dec139f277d899463214d#diff-6558f7be1858e8498722b9ac5981271f70d5219829f557f54fa720cf443c3e5bR1681.

Sebastian

cdoublev:
Also for what it is worth, Chrome and FF do not parse position(s) before the color as valid, and if it should be allowed, it may be odd to be able to specify both positions before or after the color, but not a color between them.

tabatkins:
Yeah, reviewing the diff I'm almost certain it was just a mistake while writing it - it *looked* reasonable to use an `&&` in the grammar while I was rewriting. I still think it's probably reasonable to do so, but nobody's followed that in forever and it's not very important, so reverting to the old spec text which matches impls makes sense.

PR #8038, opened 2022-11-07
[css-nesting] fix `@layer` example

romainmenke:
@tabatkins @argyleink 

`@layer` already defines how nesting multiple `@layer` rules works.
In that case sub layers are created.

```css
@layer foo {
  @layer bar {

  }
}

/* equivalent to */

@layer foo.bar {}
```


argyleink (approved):
lgtm

Issue #7085, opened 2022-02-24
[selectors-4] Allow more pseudo-classes following a pseudo-element


cdoublev:
There is already an issue mentioned in [3.6.3. Pseudo-classing Pseudo-elements](https://drafts.csswg.org/selectors/#pseudo-element-states) (I did not find the corresponding issue on GH):

> A [pseudo-element](https://drafts.csswg.org/selectors/#pseudo-element) may be immediately followed by any combination of the [user action pseudo-classes](https://drafts.csswg.org/selectors/#useraction-pseudos) [...]
>
> ISSUE 3 Clarify that [:not()](https://drafts.csswg.org/selectors/#negation-pseudo) and [:is()](https://drafts.csswg.org/selectors/#matches-pseudo) can be used when containing above-mentioned pseudos.

I think the issue can be completed with:

- all pseudo-classes: when following a pseudo-element explicitly defined to be matchable by this pseudo-class (eg. `::shadow:empty`, `::shadow:defined`, `::shadow-child:first-child`, etc)
- `:where()`: when containing user-action pseudo-classes as arguments
- `:defined`: I'm not sure about this one because it is defined as matching *elements that are fully defined, as dictated by the host language*, not as matching *pseudo-element*

Pseudo-class functions which do not match the corresponding (pseudo-element) sub-classing condition could be considered as matching nothing instead of being invalid.

But I wonder if I'm missing something because: 
- `::before:hover` does not work in Chrome/Firefox (not tested in other browsers)
- [MDN browser compat table](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) includes *Pseudo-element support* with support from early versions of all browsers: is it the support of `:hover` *followed by* a pseudo-element, or the other way around? The french version defines that *the `:hover` pseudo-class can follow any pseudo-element*, but not the english version. Non-english versions are often translations of the english text. 
- I could not find any related issues on Chromium Issues or ~~Mozilla Firebug~~ Bugzilla 
- **EDIT:** I missed the related issue on [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1122965) and created an issue on [Chromium Issues](https://bugs.chromium.org/p/chromium/issues/detail?id=1300566)
[...]

PR #8041, opened 2022-11-08
[selectors-4] Allow logical combination pseudos wherever their arguments are themselves valid. #7085

fantasai:


selectors-4/Overview.bs
```diff
@@ -988,6 +992,13 @@ Invalid Selectors and Error Handling</h3>
 <h2 id="logical-combination">
 Logical Combinations</h2>
 
+	Selector logical can be manipulated by
```
Loirooriol, line 995: Should this be "Selector logic"? I don't understand an adjective here.

selectors-4/Overview.bs
```diff
@@ -831,8 +829,14 @@ Pseudo-classing Pseudo-elements</h4>
 		whereas ''::first-line:hover'' only matches if the first line itself is hovered.
 	</div>
 
+	The [=logical combination pseudo-classes=]
+	'':not()'', '':is()'', and '':where()''
+	are also valid immediately following a [=pseudo-element=]
+	provided their arguments all adhere to the above restriction.
```
Loirooriol, line 835: Does this mean that `::before:is(:hover, :first-child)` is completely invalid, or becomes `::before:is(:hover)`?

selectors-4/Overview.bs
fantasai: yes, that was a typo :) I'll fix it

selectors-4/Overview.bs
fantasai: Seems like the latter would be more consistent with standard behavior, so I'll update the PR.

Issue #7749, opened 2022-09-15
[web-animations-2] Prefer phase based start / end delays over converting time based delays


flackr:
In web-animations-2 we defined a procedure for [converting time based animations to non-time based timelines](https://w3c.github.io/csswg-drafts/web-animations-2/#time-based-animation-to-a-proportional-animation) to allow switching the timeline of a regular animation to a progress-based timeline and have it "just work"

The conversion currently takes place if the iteration duration is not auto. I think the developer expectation would reasonably be that we only try to convert a time based animation if all of the fields are time based and that if any of them are not that we would force the time based fields to 0.

Filing this in relation to the handling in our implementation.

flackr:
@kevers-google FYI

flackr:
Since we now have animation-range for the attachment these non-time based timelines, I think it would be much simpler if we dropped all time-based values for their default values when applied to a non-time based timeline. The somewhat common case of assigning a non-time-based timeline on a time-based animation essentially fills the specified animation-range.

Applying this would mean removing the procedure for [converting time based animations to non-time based timelines](https://w3c.github.io/csswg-drafts/web-animations-2/#time-based-animation-to-a-proportional-animation) and instead using any time values as their initial values which would go down the "auto" duration path.

TLDR Proposal is:
Time based specified duration, start delay, and end delay will compute to their initial values ("auto", 0, 0) when applied to an animation using a non-time based timeline.

flackr:
That said, dropping the specified time values means that we would lose the proportional sizes of start delay, end delay and duration in both single effects and in the future for sequence and group effects. Long term it would probably be better to allow developers to define durations / delays applicable to either timeline type by specifying their values in percentages (of the applied animation-range), but the question is do we want to have the complexity of continuing to convert time based values to their relative proportions in the meantime. @birtles Do you think it's valuable to try to convert values specified in one timelines units to another?

birtles:
The main use case I'm aware of is having a scroll-based animation that switches to a time-based one and runs to completion once it passes a certain threshold (the "point of no return"). Whatever we do, I hope that's a relatively easy effect to achieve. I'm fine with deferring features provided we have a concrete proposal about how they will work in future so we don't paint ourselves into a corner.

flackr:
Thanks! In order to make that use case just work I think it's better to stick with the direction we've gone right now where we convert time values to their relative percentages preserving the relative differences in duration / start / end delays.

[...]

PR #8086, opened 2022-11-16
[web-animations-2] Don't convert mixed times and proportions to proportional.

flackr:
[web-animations-2] Don't convert mixed times and proportions to proportional #7749

In supporting the timeline range based start and end delays on animations we should only convert time based animations if all values are time based so as not to clobber any developer set proportional values.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -226,18 +226,12 @@ Append:
 
 > Animation effects associated with a [=progress-based timeline=] require their
 > timing properties to be converted to proportions. The procedure for converting
-> a <dfn lt="time-based animation to a proportional animation"></dfn> is as
+> a <dfn lt="time-based animation to a proportional animation">time-based
+> animation to a proportional animation</dfn> is as
 > follows:
 >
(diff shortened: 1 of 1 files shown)


web-animations-2/Overview.bs
```diff
@@ -226,18 +226,12 @@ Append:
 
 > Animation effects associated with a [=progress-based timeline=] require their
 > timing properties to be converted to proportions. The procedure for converting
-> a <dfn lt="time-based animation to a proportional animation"></dfn> is as
+> a <dfn lt="time-based animation to a proportional animation">time-based
+> animation to a proportional animation</dfn> is as
```
birtles, line 230: Nit: since the definition text appears to be identical to the `lt` text, do we need the `lt` attribute?

flackr: Thanks, good catch, removed the `lt`.

flackr:
Given that we haven't added percentages to delays yet and range targeted start/end times are handled through `animation-range` and we already have a note that if delays support proportions eventually they should be used, I think we don't need any change here right now.

PR #8089, opened 2022-11-16
[web-animations-1] Animations are current if associated with non-increasing timelines.

### LINKED ISSUE #8053: [scroll-animations-1][web-animations-2] getAnimations should return scroll animations which haven't ticked yet.
In web-animations-1, [getAnimations](https://www.w3.org/TR/web-animations-1/#dom-animatable-getanimations) is defined to return the list of [relevant](https://www.w3.org/TR/web-animations-1/#relevant-animation) animations, most notably where the [associated effect](https://www.w3.org/TR/web-animations-1/#animation-associated-effect) is [current](https://www.w3.org/TR/web-animations-1/#current) or [in effect](https://www.w3.org/TR/web-animations-1/#in-effect). The intent is to not need to return animations which will no longer having an affect on the element. Animations which will later have an effect are still considered current.

Scroll-linked animations are not current or in effect due to an inactive time when:
* They have just been constructed and have not sampled the scroll position yet, see [avoiding cycles with layout](https://drafts.csswg.org/scroll-animations-1/#avoiding-cycles). See repro in [crbug.com/1356542](https://bugs.chromium.org/p/chromium/issues/detail?id=1356542)
* The scroller is not currently scrollable

In both of these cases, I think developers would expect the animation to still be considered relevant as it may later have an effect, just as they also expect an animation which has been delayed to be returned by getAnimations.

I propose these animations should be considered to be current, as the [definition for current](https://www.w3.org/TR/web-animations-1/#ref-for-current) suggests it includes any animation which may become in play. This would mean:
* These animations would be listed in getAnimations per [web-animations-1 6.8](https://www.w3.org/TR/web-animations-1/#dom-animatable-getanimations)
* Animated properties will act as if will-change is specified per [web-animations-1 5.6](https://www.w3.org/TR/web-animations-1/#side-effects-section).

Concretely the change could be adding "If has finite timeline is true" to the list of [conditions under which an animation is considered to be current](https://www.w3.org/TR/web-animations-1/#current).

flackr:
[web-animations-1] Animations are current if associated with non-increasing timelines.

If an animation is associated with a non-monotically increasing timeline it may become active in the future. As such it should still be considered current (the same as an animation which may play in the future). This fixes #8053.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2396,8 +2396,9 @@ of the following conditions are true:
 *   the [=animation effect=] is [=associated with an animation=] with
     a [=playback rate=] &lt; 0 and
     the [=animation effect=] is in the [=animation effect/after phase=].
-*   the [=animation effect=] is [=associated with an animation=] with an
-    associated [=timeline=] that is not [=monotonically increasing=].
+*   the [=animation effect=] is [=associated with an animation=] not in the
+    [=play state/idle=] [=play state=] with a non-null associated
+    [=timeline=] that is not [=monotonically increasing=].
 
 An animation effect is <dfn>in effect</dfn> if its <a>active time</a>, as
 calculated according to the procedure in [[#calculating-the-active-time]],
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2395,7 +2395,7 @@ of the following conditions are true:
     the [=animation effect=] is in the [=animation effect/before phase=], or
 *   the [=animation effect=] is [=associated with an animation=] with
     a [=playback rate=] &lt; 0 and
-    the [=animation effect=] is in the [=animation effect/after phase=].
+    the [=animation effect=] is in the [=animation effect/after phase=], or
 *   the [=animation effect=] is [=associated with an animation=] not in the
     [=play state/idle=] [=play state=] with a non-null associated
     [=timeline=] that is not [=monotonically increasing=].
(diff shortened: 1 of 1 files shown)


birtles:
@flackr Is this right? If you cancel an animation associated with a scroll-driven timeline do we really want its effect to be treated as current?

(Also, I guess we should make it clear that an animation without a timeline does not count as an animation "with an
associated timeline that is not monotonically increasing" because I think in some cases we tend to think of the animation as being associated with a null timeline in that case.)

flackr:
Thanks for the review!

> @flackr Is this right? If you cancel an animation associated with a scroll-driven timeline do we really want its effect to be treated as current?

No. Not if canceled. But we do want newly started scroll timelines or scroll timelines which are still playing associated with a scrolling box which isn't currently scrollable to still be treated as current. Looking back at what happens, a played scroll linked animation is given a start time so it will not be idle, unlike a canceled animation so I've included this in the requirements.

> (Also, I guess we should make it clear that an animation without a timeline does not count as an animation "with an associated timeline that is not monotonically increasing" because I think in some cases we tend to think of the animation as being associated with a null timeline in that case.)

Agreed. Added non-null timeline.

web-animations-1/Overview.bs
```diff
@@ -2396,6 +2396,9 @@ of the following conditions are true:
 *   the [=animation effect=] is [=associated with an animation=] with
     a [=playback rate=] &lt; 0 and
     the [=animation effect=] is in the [=animation effect/after phase=].
+*   the [=animation effect=] is [=associated with an animation=] not in the
+    [=play state/idle=] [=play state=] with a non-null associated
+    [=timeline=] that is not [=monotonically increasing=].
```
birtles, line 2401:
```suggestion
    the [=animation effect=] is in the [=animation effect/after phase=], or
*   the [=animation effect=] is [=associated with an animation=] not in the
    [=play state/idle=] [=play state=] with a non-null associated
    [=timeline=] that is not [=monotonically increasing=].
```

birtles (approved):

PR #8093, opened 2022-11-17
[css-view-transitions-1] Clean up use of dynamic styles

### LINKED ISSUE #8069: [css-view-transitions-1] Scope of the "user-agent origin" stylesheet
In view transitions, we [dynamically update styles in the user-agent origin](https://drafts.csswg.org/css-view-transitions-1/#animate-a-view-transition-algorithm). (I'm not happy with the prose of that section, so don't worry too much about that)

In Chrome, there's an instance of the user-agent origin stylesheet per document, so the modifications we make only impact the current document. However, the [cascade spec](https://drafts.csswg.org/css-cascade-5/#cascading-origins) suggest there may only be one user-agent origin stylesheet, therefore the modifications we make in the view transitions spec would impact other documents, which is bad.

My gut instinct is that we should spec a new cascading origin for view transitions, and specify that there's one instance per document, so modifications will only apply to that document. The [cascade spec](https://drafts.csswg.org/css-cascade-5/#cascading-origins) already mentions an "animation origin" and a "transition origin", which feel similar, although I don't see any usage of those in the linked specs.

@mirisuzanne: I think you might know the right way to spec this.

jakearchibald:
Fixes #8069

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -30,6 +30,8 @@ spec:css-images-4; type:function; text:element()
 urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	text: NavigateEvent
 	text: signal; for: NavigateEvent; url: #ref-for-dom-navigateevent-signal①
+urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
+	text: HTML user agent style sheet; url: #the-css-user-agent-style-sheet-and-presentational-hints
 </pre>
 
 <style>
@@ -162,7 +164,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: The value <css>none</css> is invalid as a <<custom-ident>>.
 	</dl>
 
-	The [=document element=] participates in a view transition by default using the following style in the [=user-agent origin=]:
+	The [=document element=] participates in a view transition by default using the following rule,
+	added to the [=HTML user agent style sheet=]:
 
 	```css
 	html {
@@ -180,7 +183,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 # User-agent keyframes # {#ua-keyframes}
 
-	The following <<@keyframes>> are added to the [=user-agent origin=]:
+	The following <<@keyframes>> are added to the [=HTML user agent style sheet=]:
 
 	```css
 	@keyframes -ua-view-transition-fade-out {
@@ -229,7 +232,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	:: This pseudo-element is the grouping container of all the other [=view-transition pseudo-elements=].
 		Its [=originating element=] is the document's [=document element=].
 
-		The following [=user-agent origin=] styles apply to this element:
+		The following is added to the [=HTML user agent style sheet=]:
 
 		```css
 		html::view-transition {
@@ -248,7 +251,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		Its [=originating element=] is the ''::view-transition'' pseudo-element.
 
-		The following [=user-agent origin=] styles apply to this element:
+		The following is added to the [=HTML user agent style sheet=]:
 
 		```css
 		html::view-transition-group(*) {
@@ -264,7 +267,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		<div class="note">
 			The aim of the style is to position the element relative to its ''::view-transition'' parent.
 
-			In addition to above, styles in the [=user-agent origin=] animate this pseudo-element's 'width' and 'height'
+			In addition to above, styles in the [=document/view transition style sheet=] animate this pseudo-element's 'width' and 'height'
 			from the size of the old element's [=border box=] to that of the new element's [=border box=].
 			Also the element's 'transform' is animated from the old element's screen space transform to the new element's screen space transform.
 			This style is generated dynamically since the values of animated properties are determined at the time that the transition begins.
@@ -279,7 +282,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		Its [=originating element=] is the ''::view-transition-group()'' pseudo-element with the same transition-name.
 
-		The following [=user-agent origin=] styles apply to this element:
+		The following is added to the [=HTML user agent style sheet=]:
 
 		```css
 		html::view-transition-image-pair(*) {
@@ -291,7 +294,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		}
 		```
 
-		In addition to above, styles in the [=user-agent origin=] add ''isolation: isolate'' to this pseudo-element
+		Note: In addition to above, styles in the [=document/view transition style sheet=] add ''isolation: isolate'' to this pseudo-element
 		if it has both ''::view-transition-new()'' and ''::view-transition-old()'' as children.
 
 		Note: The aim of the style is to position the element to occupy the same space as its ''::view-transition-group()'' element
@@ -307,7 +310,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		Its [=originating element=] is the ''::view-transition-image-pair()'' pseudo-element with the same transition-name.
 
-		The following [=user-agent origin=] styles apply to this element:
+		The following is added to the [=HTML user agent style sheet=]:
 
 		```css
 		html::view-transition-old(*) {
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -162,7 +164,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: The value <css>none</css> is invalid as a <<custom-ident>>.
 	</dl>
 
-	The [=document element=] participates in a view transition by default using the following style in the [=user-agent origin=]:
+	The [=document element=] participates in a view transition by default using the following rule,
+	added to the [=HTML user agent style sheet=]:
```
jakearchibald, line 168: If the style never changes, and does no harm always-being-there, I've added it to the "HTML user agent style sheet"

css-view-transitions-1/Overview.bs
```diff
@@ -277,9 +292,11 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		}
 		```
 
-		In addition to above, styles in the [=user-agent origin=] add ''isolation: isolate'' to this pseudo-element
+		In addition to above, styles in the [=document/view transition style sheet=] add ''isolation: isolate'' to this pseudo-element
 		if it has both ''::view-transition-new()'' and ''::view-transition-old()'' as children.
 
+		Issue: This reads like a non-normative note, but it's normative. This rule should be specified at the same time as other styles, when the condition above can be determined.
```
jakearchibald, line 298: I can fix this as part of this PR, once I've checked the details.

css-view-transitions-1/Overview.bs
```diff
@@ -420,6 +438,25 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+
+		: <dfn>group styles rule</dfn>
+		:: A {{CSSStyleRule}} or null. Initially null.
+
+		: <dfn>old view-box rule</dfn>
+		:: A {{CSSStyleRule}} or null. Initially null.
+
+		: <dfn>new view-box rule</dfn>
+		:: A {{CSSStyleRule}} or null. Initially null.
+	</dl>
```
jakearchibald, line 458: These definitions let us update/remove rules later.

css-view-transitions-1/Overview.bs
```diff
@@ -430,6 +467,12 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		: <dfn>transition suppressing rendering</dfn>
 		:: a boolean. Initially false.
+
+		: <dfn>view transition style sheet</dfn>
+		:: a [=style sheet=].
+			Initially a new [=style sheet=] in the [=user-agent origin=], ordered after the [=HTML user agent style sheet=].
+
+			Note: This is used to hold dynamic styles relating to transitions.
```
jakearchibald, line 475: This stylesheet is for dynamic styles involved in the transition. It's explicitly linked to a single document, clarifying the scoping.

css-view-transitions-1/Overview.bs
```diff
@@ -627,16 +670,16 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
 			1. Otherwise, if |document|'s [=document/active DOM transition=]'s [=ViewTransition/phase=] is "`animating`",
 				then [=update transition DOM=] for |document|'s [=document/active DOM transition=].
 	</div>
 
-## [=Perform an old capture=] ## {#perform-an-old-capture-algorithm}
+## [=Setup view transition=] ## {#setup-view-transition-algorithm}
 
 	<div algorithm>
-		To <dfn>perform an old capture</dfn> given a {{ViewTransition}} |transition|,
+		To <dfn>setup view transition</dfn> given a {{ViewTransition}} |transition|,
```
jakearchibald, line 682: This algorithm is doing much more than capturing, so I renamed it.

css-view-transitions-1/Overview.bs
```diff
@@ -828,13 +866,13 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 			Note: The ready promise would've been resolved if {{ViewTransition/skipTransition()}} is called after we start animating.
 
-		1. [=promise/React=] to [=ViewTransition/DOM updated promise=]:
+		1. [=promise/React=] to |transition|'s [=ViewTransition/DOM updated promise=]:
```
jakearchibald, line 869: Being specific about which instance of the promise we're talking about

css-view-transitions-1/Overview.bs
```diff
@@ -987,83 +1019,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Issue: Define the algorithm used to clip the snapshot when it exceeds max size.
 	</div>
 
-## [=Animate a view transition=] ## {#animate-a-view-transition-algorithm}
```
jakearchibald, line 990: This algorithm was being called every frame, but the things it does only need to be done once ever (like non-changing UA styles), or once at the start of each transition.

I moved the steps elsewhere.

css-view-transitions-1/Overview.bs
```diff
@@ -987,83 +1019,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+## [=Setup transition pseudo-elements=] ## {#setup-transition-pseudo-elements-algorithm}
```
jakearchibald, line 1022: This algorithm used to be "create transition pseudo-elements", but I added the style rules for things that only need to be done once per transition, so I renamed it to reflect that.

css-view-transitions-1/Overview.bs
```diff
@@ -1101,19 +1080,51 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+						html::view-transition-group(<var>transitionName</var>) {
+							animation-name: -ua-view-transition-group-anim-<var>transitionName</var>;
+						}
+					</pre>
+
+					Note: The above code example contains variables to be replaced.
 	</div>
 
-## [=Style transition pseudo-elements=] ## {#style-transition-pseudo-elements-algorithm}
+## [=Update pseudo-element styles=] ## {#style-transition-pseudo-elements-algorithm}
```
jakearchibald, line 1121: Changed the naming to suggest that it happens more than once.

css-view-transitions-1/Overview.bs
```diff
@@ -1189,6 +1221,27 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+				Issue: There needs to be a definition/link for "remove".
+
+				Issue: There needs to be a definition/link for "associated".
+
+		1. [=list/For each=] |capturedElement| of |transition|'s [=ViewTransition/named elements=]' [=map/values=]:
+
+			1. [=list/For each=] |style| of |capturedElement|'s [=captured element/style definitions=]:
+
+				1. If |style| is not null,
+					then remove |style| from |document|'s [=document/view transition style sheet=].
```
jakearchibald, line 1242: We were missing this bit before, so styles would just hang around forever.

css-view-transitions-1/Overview.bs
```diff
@@ -1088,6 +1041,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+					<!-- deliberately using <pre> so the example can contain <var> references -->
+					<pre highlight="css">
+						html::view-transition-old(<var>transitionName</var>) {
+							object-view-box: <var>oldViewBox</var>;
+						}
+					</pre>
+
+					Note: The above code example contains variables to be replaced.
+
+					Issue: Which of ''xywh()''/''rect()''/''inset()'' should we use?
```
vmpstr, line 1058: Can this be resolved at editor's discretion?

css-view-transitions-1/Overview.bs
```diff
@@ -1101,19 +1070,81 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 				except the [=view-transition pseudo-elements=],
 				are not painted (as if they had visibility: hidden)
 				and do not respond to hit-testing (as if they had pointer-events: none) until |new| exists.
+
+			1. If neither of |capturedElement|'s [=captured element/old image=] or [=captured element/new element=] is null:
```
vmpstr, line 1074: I prefer "If both ... are not null", since it sounds more in the positive? Just a nit though.

css-view-transitions-1/Overview.bs
```diff
@@ -1101,19 +1070,81 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+
+				1. Let |transform| be |capturedElement|'s [=old styles=]'s 'transform' property.
```
vmpstr, line 1076: Where is old styles defined? I know this has moved from elsewhere, but is this correct? We're not using the captured element's style and read its transform property, but rather find the transform it has in screen space and use that as the transform. Similarly for width/height, it could be a flex grow with no explicit width, so what does style width mean here?

css-view-transitions-1/Overview.bs
```diff
@@ -1140,32 +1171,53 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+					<!-- deliberately using <pre> so the example can contain <var> references -->
+					<pre highlight="css">
+						html::view-transition-new(<var>transitionName</var>) {
+							object-view-box: <var>newViewBox</var>;
+						}
+					</pre>
+
+					Note: The above code example contains variables to be replaced.
+
+					Issue: Which of ''xywh()''/''rect()''/''inset()'' should we use?
```
vmpstr, line 1217: ditto

css-view-transitions-1/Overview.bs
jakearchibald: Any preferences?

css-view-transitions-1/Overview.bs
jakearchibald: I've replaced `[=old styles=]` with `[=captured element/old styles=]`. It's set in "setup view transition".

I think the confusion here is it feels like a direct reference to the old styles. Would it be better if the prose was:

> ```
> 1. Let |transform| be |capturedElement|'s [=captured element/old offset transform=].
> ```

css-view-transitions-1/Overview.bs
jakearchibald: I like that idea, but wouldn't it mean a user style like:

```css
::view-transition {
  content: '';
}
```

…would cause the view transition pseudo to be 'always there'?

jakearchibald: I've filed https://github.com/w3c/csswg-drafts/issues/8113 to think more about this.

css-view-transitions-1/Overview.bs
jakearchibald: Settled

vmpstr (approved):

PR #8105, opened 2022-11-19
[web-animations-1] Factor out relevant animation definitions

### LINKED ISSUE #8001: [web-animations-1] Export dfns to expose the effect target of an animation
In the [view transitions spec](https://drafts.csswg.org/css-view-transitions-1/) we want to write prose about the element being animated by a specific animation.

It feels like the right prose would be something like:

> 1. Let effect be animation's associated effect.
> 2. If effect is a keyframe effect, and effect's effect target's target element is…

If that's correct, I propose:

- Export [keyframe effect](https://drafts.csswg.org/web-animations-1/#keyframe-effect).
- Create a proper association between "effect target" and "keyframe effect". A lot of prose right now talks about an effect's effect target, but there's no defined association between the two. This association will be exported.
- Export [effect target](https://drafts.csswg.org/web-animations-1/#effect-target), and its [target element](https://drafts.csswg.org/web-animations-1/#target-element) and [target pseudo-selector](https://drafts.csswg.org/web-animations-1/#target-pseudo-selector).

birtles:
Fixes #8001.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2401,6 +2401,39 @@ An animation effect is <dfn>in effect</dfn> if its <a>active time</a>, as
 calculated according to the procedure in [[#calculating-the-active-time]],
 is <em>not</em> <a>unresolved</a>.
 
+### Relevant animations ### {#relevant-animations-section}
+
+We may define an [=animation=] as being [=relevant=]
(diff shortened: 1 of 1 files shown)


jakearchibald (approved):
LGTM!

flackr (approved):

birtles:
Thanks!

Issue #7788, opened 2022-09-24
[css-view-transitions-1] CSS selector syntax for generated elements and API names


jakearchibald:
Folks at TPAC asked for the pseudo-element names to be reduced so we had a bikeshedding session, and here's what we came up with.

@fantasai @mirisuzanne I'm interested to hear your thoughts on this. We've tried to go for brevity rather than full namespacing.

## "Shared element transition" becomes "View transition"

Rationale:

- "shared element" didn't seem accurate. There's no sharing of elements.
- The DOM changes, and a transition is created from the two states. The actual DOM elements aren't animated, it's just a view of them, so "view transitions".

## `page-transition-tag` becomes<br/>`view-name`

```js
.header {
  /* old */
  page-transition-tag: header;
  /* new */
  view-name: header;
}
```

This causes pseudos to be created so the header can be animated independently during the transition.

Rationale:

- The scope of a transition may not always be the whole "page".
- Other specs use "name" rather than "tag" to set an identifier (`container-name`, `scroll-timeline-name`).
- Using "view" because "view transition".

## `html::page-transition` becomes<br/> `html::view-root`

This is the root of the pseudo-element tree that contains all the transition parts.

Rationale:

- The scope of a transition may not always be the whole "page".
- Using "view" because "view transition".
- "root" because it's the root of the pseudo-element tree.

## `html::page-transition-container(header)` becomes<br/> `html::view(header)`

This is the pseudo-element that represents a view of an element during the transition. In this case it's the view of the 'header'. Developers target this to customize the animation. It's a child of `html::view-root`.

Rationale:

- Each element that's given a `view-name` results in the creation of one of these pseudo-elements. So `view-name: header` results in a `html::view(header)` during the transition.

## `html::page-transition-image-wrapper(header)` becomes<br/> `html::view-image-group(header)`

This wraps the view's images, providing isolation for their blending (to allow a true cross-fade). In this case it's the image-group for the 'header'. It's a child of `html::view`.

Rationale:

- The pseudo elements within are replaced elements, but they behave like images. They have a width, height, and an aspect ratio.

## `html::page-transition-outgoing-image(header)` and<br/> `html::page-transition-incoming-image(header)` become<br/> `html::view-before(header)` and<br/> `html::view-after(header)` respectively

These are the replaced elements that represent the before and after view. In their default animation, they cross-fade.

Rationale:

- Folks had difficulty mapping 'outgoing' and 'incoming'.
[...]

PR #8126, opened 2022-11-22
[css-view-transitions-1] Define pseudo-nesting and naming

### LINKED ISSUE #8113: [css-view-transitions-1] Prose to modify pseudo-trees
When adding/removing regular elements from the DOM, we can use defined terms like [pre-insert](https://dom.spec.whatwg.org/#concept-node-pre-insert), and [remove](https://dom.spec.whatwg.org/#concept-node-remove).

I don't think similar things exist for pseudo-element trees?

Right now we're using language like "append somePesudoElement to someOtherPseudoElement", but the term "append" isn't linked, so it's a bit of a hand-wave.

@fantasai @tabatkins are you aware of any other specs that have ran into this?

If it isn't a problem generally, we could define a "view transition pseudo-element" dfn that has a "pseudo-element children" list, and say that a view transition pseudo-element must render its pseudo-element children in order.

jakearchibald:
This attempts to define how pseudos are built and ordered. It also defines the view-transition-name of the pseudos.

Fixes #8113.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -199,7 +199,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 ## Pseudo-element parents ## {#parent-pseudo}
 
-	A <dfn>pseudo-element parent</dfn> is a mixin that has <dfn for="pseudo-element parent">children</dfn>,
+	A <dfn>pseudo-element parent</dfn> is a mixin for [=tree-abiding pseudo-elements=] that has <dfn for="pseudo-element parent">children</dfn>,
 	a [=/list=] of [=pseudo-elements=].
 	Initially an empty [=/list=].
 
@@ -227,7 +227,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 ## Named view-transition pseudo-elements ## {#named-view-transition-pseudo}
 
-	A <dfn>named view-transition pseudo-element</dfn> is a mixin for [=pseudo-elements=]
+	A <dfn>named view-transition pseudo-element</dfn> is a mixin for [=tree-abiding pseudo-elements=]
 	that also have a <dfn for="named view-transition pseudo-element">view-transition name</dfn>,
 	a string.
 
@@ -249,7 +249,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 ## View transition pseudo-elements ## {#view-transition-pseudos}
 
 	: <dfn>::view-transition</dfn>
-	:: A [=pseudo-element parent=].
+	:: A [=tree-abiding pseudo-element=] that is also
+		a [=pseudo-element parent=].
 		Its [=originating element=] is the document's [=document element=].
 
 		The following is added to the [=HTML user agent style sheet=]:
@@ -270,7 +271,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		</div>
 
 	: <dfn>::view-transition-group( <<pt-name-selector>> )</dfn>
-	:: A [=pseudo-element parent=], and a [=named view-transition pseudo-element=].
+	:: A [=tree-abiding pseudo-element=] that is also
+		a [=pseudo-element parent=], and a [=named view-transition pseudo-element=].
 
 		It is selected from its [=ultimate originating element=], the [=document element=].
 
@@ -299,7 +301,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		</div>
 
 	: <dfn>::view-transition-image-pair( <<pt-name-selector>> )</dfn>
-	:: A [=pseudo-element parent=], and a [=named view-transition pseudo-element=].
+	:: A [=tree-abiding pseudo-element=] that is also
+		a [=pseudo-element parent=], and a [=named view-transition pseudo-element=].
 
 		It is selected from its [=ultimate originating element=], the [=document element=].
 
@@ -318,7 +321,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		Note: This is only ever a [=pseudo-element parent/children|child=] of a ''::view-transition-group()''.
 
 	: <dfn>::view-transition-old( <<pt-name-selector>> )</dfn>
-	:: A [=named view-transition pseudo-element=].
+	:: A [=tree-abiding pseudo-element=] that is also
+		a [=named view-transition pseudo-element=].
 
 		It is selected from its [=ultimate originating element=], the [=document element=].
 
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -199,7 +199,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 ## Pseudo-element root ## {#pseudo-root}
 
-	Issue: This definition should be moved to [[css-pseudo-4]].
+	Note: This is a general definition for trees of pseudo-elements. If other features need this behavior, these definitions will be moved to [[css-pseudo-4]].
 
 	A <dfn>pseudo-element root</dfn> is a type of [=tree-abiding pseudo-element=] that is the [=tree/root=] in a [=tree=] of [=tree-abiding pseudo-elements=],
 	known as the <dfn>pseudo-element tree</dfn>.
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -259,6 +259,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		a [=pseudo-element parent=].
 		Its [=originating element=] is the document's [=document element=].
 
+		Its [=containing block=] is the [=snapshot viewport=].
+
 		The following is added to the [=HTML user agent style sheet=]:
 
 		```css
@@ -422,8 +424,6 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 	- The [=view-transition layer=] paints after the stacking context for the [=document element=] and [=top layer=].
 
-	- The [=view-transition layer=]'s [=containing block=] is the [=snapshot viewport=].
-
 	Note: The intent of the feature is to be able to capture the contents of the page, which includes the top layer elements.
 	In order to accomplish that, the [=view-transition layer=] cannot be a part of the captured top layer context,
 	since that results in a circular dependency.
@@ -499,6 +499,16 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 			Note: This is used to hold dynamic styles relating to transitions.
 	</dl>
+
+## Additions to [=document element=] ## {#additions-to-document-element}
+
+	A [=document element=] additionally has a <dfn for="document element">view-transition root pseudo-element</dfn>.
+	A ''::view-transition'' or null. Initially null.
+
+	When this is a ''::view-transition'', the ''::view-transition'' renders as a child of the [=document element=],
+	and the [=document element=] is its [=originating element=].
+
+	Note: The position of the ''::view-transition'' within the [=document element=] does not matter, as the ''::view-transition'''s [=containing block=] is the [=snapshot viewport=].
 
 # API # {#api}
 
@@ -1054,7 +1064,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 		1. Let |document| be [=this's=] [=relevant global object's=] [=associated document=].
 
-		1. [=list/Append=] |transition|'s [=ViewTransition/transition root pseudo-element=] to this |document|'s [=snapshot viewport=]'s [=pseudo-element parent/children=].
+		1. Set |document|'s [=document element=]'s [=view-transition root pseudo-element=] to |transition|'s [=ViewTransition/transition root pseudo-element=].
 
 		1. [=map/For each=] |transitionName| → |capturedElement| of |transition|'s [=ViewTransition/named elements=]:
 
@@ -1281,8 +1291,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 		1. Let |document| be |transition|'s [=relevant global object's=] [=associated document=].
 
-		1. [=list/Remove=] |transition|'s [=ViewTransition/transition root pseudo-element=]
-			from |document|'s [=snapshot viewport=]'s [=pseudo-element parent/children=].
+		1. Set |document|'s [=document element=]'s [=view-transition root pseudo-element=] to null.
 
 		1. [=list/For each=] |capturedElement| of |transition|'s [=ViewTransition/named elements=]' [=map/values=]:
 
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1282,6 +1282,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 					and |style| is in |document|'s [=document/view transition style sheet=],
 					then remove |style| from |document|'s [=document/view transition style sheet=].
 
+		1. Set |document|'s [=document/show view-transition root pseudo-element=] to false.
+
 		1. Set |document|'s [=document/active DOM transition=] to null.
 	</div>
 
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -212,7 +212,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 	When a [=pseudo-element=] |pseudo| is part of a [=pseudo-element parent=] |parent|'s [=pseudo-element parent/children=],
 	then |pseudo|'s [=originating pseudo-element=] is |parent|.
 
-	If a [=pseudo-element parent=]'s [=pseudo-element parent/children=]'s [=list/size=] is 1, then '':only-child'' selects [=pseudo-element parent/children=][0].
+	If a [=pseudo-element parent=]'s [=pseudo-element parent/children=]'s [=list/size=] is 1, then '':only-child'' selects [=pseudo-element parent/children=][0]. Otherwise, '':only-child'' selects nothing.
+
+	Note: This means that `::view-transition-new(ident):only-child` will only select `::view-transition-new(ident)` if the parent `::view-transitions-image-pair(ident)` only contains one child. As in, there is no `::view-transition-old(ident)`.
 
 	<div algorithm>
 		To get the <dfn for="pseudo-element parent">descendants</dfn> of a [=pseudo-element parent=] |parent|, perform the following steps.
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -197,38 +197,59 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 # Pseudo-elements # {#pseudo}
 
-	The following <dfn export>view-transition pseudo-elements</dfn> represent the various items being animated.
```
jakearchibald, line 200: This dfn was no longer needed.

css-view-transitions-1/Overview.bs
```diff
@@ -197,38 +197,59 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
-	to specify which elements named with 'view-transition-name' are affected.
+	When a [=pseudo-element=] |pseudo| is part of a [=pseudo-element parent=] |parent|'s [=pseudo-element parent/children=],
+	and |parent| is a [=pseudo-element=],
+	then |pseudo|'s [=originating pseudo-element=] is |parent|.
 
-	There can be multiple pseudo-elements of the same type,
-	one for each 'view-transition-name' participating in a transition.
+	<div algorithm>
+		To get the <dfn for="pseudo-element parent">descendants</dfn> of a [=pseudo-element parent=] |parent|, perform the following steps.
+		They return a [=/list=] of [=pseudo-elements=].
```
vmpstr, line 214: I don't think you need this line

vmpstr (commented):
Generally looks good. I'm starting to get a bit worried that this is extremely complicated to read, but that's not due to this PR: it has been building up.

I'll read the spec once its compiled again, but perhaps we can have separate "intro" sections for pseudo element algorithms that just describe the tree and a separate "intro" section for the startViewTransition process and steps that follow. WDYT?

css-view-transitions-1/Overview.bs
jakearchibald: It's the defined format in https://infra.spec.whatwg.org/#algorithm-declaration

css-view-transitions-1/Overview.bs
```diff
@@ -197,38 +197,59 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 
-	The ''::view-transition'' pseudo-element acts as a grouping element for other [=view-transition pseudo-elements=]
-	and has the document's [=document element=] as its [=originating element=].
+	A <dfn>pseudo-element parent</dfn> is a mixin that has <dfn for="pseudo-element parent">children</dfn>,
+	a [=/list=] of [=pseudo-elements=].
+	Initially an empty [=/list=].
 
-	Note: For example, '':root::view-transition'' selector matches this pseudo-element,
-	but ''div::view-transition'' does not.
+	A [=pseudo-element parent=]'s [=pseudo-element parent/children=] are laid out in the order they appear in the [=/list=], according to their styles.
```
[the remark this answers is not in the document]
khushalsagar: I'm worried we'll end up re-defining a lot of DOM concepts if we have to define how a tree of pseudo-elements works. For instance, this still leaves hit-testing in an ambiguous state.

Also, implementation wise a pseudo-element tree works exactly like a [node](https://dom.spec.whatwg.org/#node) tree. I mean if you inserted nodes using the script API in the same order (as defined in this spec), style, layout, paint, hit-testing would all work the same as a script generated DOM tree. The only difference here is that the Node APIs are not exposed to the developer.

So could we just clarify the difference between pseudo-element and node, and then lean on same spec concepts as a node tree for this pseudo-element tree?

css-view-transitions-1/Overview.bs
```diff
@@ -371,6 +397,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 	This means the snapshot canvas size is likely to be consistent for the [=document element=]'s [=captured element/old image=] and [=captured element/new element=].
 
+	The [=snapshot viewport=] is also a [=pseudo-element parent=].
```
khushalsagar, line 400: I didn't follow this. The snapshot viewport is just a conceptual containing block, not a pseudo-element.

css-view-transitions-1/Overview.bs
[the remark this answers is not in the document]
jakearchibald: > So could we just clarify the difference between pseudo-element and node

I don't think that should be this spec's job. Presumably the details around how pseudo-elements handle clicks is already defined for things like `::before` and `::after`.

The additions we need compared to `::before` and `::after`:

- A clear tree structure. Although `::before::marker` already has nesting, it's hand-waved in a way that I don't think it suitable for the level of nesting we have.
- Insert and remove. `::before` and `::after` are kinda 'always there', whereas ours come and go.
- Ordering within the parent. `::before` and `::after` [hand-wave this in a kinda ambiguous way](https://w3c.github.io/csswg-drafts/css-pseudo/#generated-content).

My intent here is that [=pseudo-element parent=]'s [=pseudo-element parent/children=] defines that the pseudos are document children of the parent, in that order.

I don't think we need to define hit testing, because (I assume) the hit testing of pseudo-elements is already defined, and it would be redundant to say that these pseudo-elements behave like pseudo-elements.

css-view-transitions-1/Overview.bs
```diff
@@ -197,38 +197,59 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 # Pseudo-elements # {#pseudo}
 
-	The following <dfn export>view-transition pseudo-elements</dfn> represent the various items being animated.
+## Pseudo-element parents ## {#parent-pseudo}
 
-	The ''::view-transition'' pseudo-element acts as a grouping element for other [=view-transition pseudo-elements=]
-	and has the document's [=document element=] as its [=originating element=].
+	A <dfn>pseudo-element parent</dfn> is a mixin that has <dfn for="pseudo-element parent">children</dfn>,
```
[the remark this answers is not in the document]
jakearchibald: Hmm, I was hoping https://en.wikipedia.org/wiki/Mixin was well-understood. Is there a term that's more familiar? "Trait"?

css-view-transitions-1/Overview.bs
jakearchibald: Yeah, I was unsure about this too. Any feelings on the right terms to use here? When we add a `::view-transition`, what are we adding it to?

css-view-transitions-1/Overview.bs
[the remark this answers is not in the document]
jakearchibald: I could change it to "If something is a <dfn>pseudo-element parent</dfn>, it has…", which avoids "mixin"

css-view-transitions-1/Overview.bs
```diff
@@ -197,38 +197,66 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 
-	There can be multiple pseudo-elements of the same type,
-	one for each 'view-transition-name' participating in a transition.
+	A [=pseudo-element parent=]'s [=pseudo-element parent/children=] defines the nested children of the [=pseudo-element parent=] in document order.
 
-	The <<pt-name-selector>> is defined as follows:
+	When a [=pseudo-element=] |pseudo| is part of a [=pseudo-element parent=] |parent|'s [=pseudo-element parent/children=],
+	then |pseudo|'s [=originating pseudo-element=] is |parent|.
+
+	If a [=pseudo-element parent=]'s [=pseudo-element parent/children=]'s [=list/size=] is 1, then '':only-child'' selects [=pseudo-element parent/children=][0].
```
khushalsagar, line 215: How should only-child work for ::view-transitions whose originating element is html. If we consider ::view-transition to be a child, similar to other DOM children, then should it always match? I can't think of a case where there are no nodes under html. Even if its possible, it'll be simpler if we don't have to support that.

css-view-transitions-1/Overview.bs
```diff
@@ -388,9 +422,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 	- The [=view-transition layer=] paints after the stacking context for the [=document element=] and [=top layer=].
 
-	- The [=view-transition layer=] is positioned at the [=snapshot viewport origin=],
-		and is the same size as the [=snapshot viewport=].
-		As in, it covers the [=snapshot viewport=].
+	- The [=view-transition layer=]'s [=containing block=] is the [=snapshot viewport=].
```
khushalsagar, line 425:
```suggestion
	- The [=containing block=] of this pseudo-element is the [=snapshot viewport=].
```

And move this to the definition for ::view-transition.

css-view-transitions-1/Overview.bs
```diff
@@ -1020,30 +1054,27 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 		1. Let |document| be [=this's=] [=relevant global object's=] [=associated document=].
 
-		1. Let |transitionRoot| be the result of creating a new ''::view-transition'' pseudo-element in |document|'s [=snapshot viewport=].
+		1. [=list/Append=] |transition|'s [=ViewTransition/transition root pseudo-element=] to this |document|'s [=snapshot viewport=]'s [=pseudo-element parent/children=].
```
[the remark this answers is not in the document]
jakearchibald: Ugh, I totally forgot about this. Still leaves us with a hand-wave.

css-view-transitions-1/Overview.bs
jakearchibald: The more I think about it, I think `::view-transition:only-child` should be 'invalid'.

`element:only-child` ignores pseudo-elements, so I don't think there's a sensible behaviour for `::pseudo:only-child` when the pseudo is a child of an element.

css-view-transitions-1/Overview.bs
```diff
@@ -468,6 +500,16 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 
+## Additions to [=document element=] ## {#additions-to-document-element}
+
+	A [=document element=] additionally has a <dfn for="document element">view-transition root pseudo-element</dfn>.
+	A ''::view-transition'' or null. Initially null.
+
+	When this is a ''::view-transition'', the ''::view-transition'' renders as a child of the [=document element=],
+	and the [=document element=] is its [=originating element=].
+
+	Note: The position of the ''::view-transition'' within the [=document element=] does not matter, as the ''::view-transition'''s [=containing block=] is the [=snapshot viewport=].
```
jakearchibald, line 511: @khushalsagar is this a correct statement?

khushalsagar: > A [=document element=] additionally has a <dfn for="document element">view-transition root pseudo-element</dfn>.
  A ''::view-transition'' or null. Initially null.
  When this is a ''::view-transition'', the ''::view-transition'' renders as a child of the [=document element=],
  and the [=document element=] is its [=originating element=].
	
Do we need all this text? I figured its enough to say: [=document element=] is the [=originating element=] for ''::view-transition'' and ''::view-transition'''s [=containing block=] is the [=snapshot viewport=].

What we do need is spec text to say:

- ''::view-transition'' is not clipped by [=document element=].
- Effects on the [=document element=] do not apply to ''::view-transition''.

There is similar spec text for this for top layer elements [here](https://fullscreen.spec.whatwg.org/#new-stacking-layer:~:text=Ancestor%20elements%20with%20overflow%2C%20opacity%2C%20masks%2C%20etc.%20cannot%20affect%20it.).

''::view-transition'' renders as a child of the [=document element=] makes it sound like its the opposite of the above 2 points.

css-view-transitions-1/Overview.bs
khushalsagar: > element:only-child ignores pseudo-elements, so I don't think there's a sensible behaviour for ::pseudo:only-child when the pseudo is a child of an element.

That makes sense. And lines up with @vmpstr's suggestion + the implementation.

css-view-transitions-1/Overview.bs
jakearchibald: I want a thing that can be `::view-transition` or null, so when it's not null, it's rendered, and when it's null, it's not rendered. As in, a way to say when it's in the document or not.

So… what is it 'in' when it's there?

css-view-transitions-1/Overview.bs
```diff
@@ -240,15 +270,17 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+		</div>
 
 	: <dfn>::view-transition-group( <<pt-name-selector>> )</dfn>
-	:: One of these pseudo-elements exists for each 'view-transition-name' in a view transition,
-		and holds the rest of the pseudo-elements corresponding to this 'view-transition-name'.
+	:: A [=tree-abiding pseudo-element=] that is also
+		a [=pseudo-element parent=], and a [=named view-transition pseudo-element=].
 
-		Its [=originating element=] is the ''::view-transition'' pseudo-element.
+		It is selected from its [=ultimate originating element=], the [=document element=].
```
cdoublev, line 283: With this change, ie. without defining `::view-transition` as the originating pseudo-element of `::view-transition-group()`, it is not clear whether the latter is still a valid sub-pseudo-element of the former, ie. whether `::view-transition::view-transition-group(name)` is still valid, even if I do not think there are usefull use cases for such explicit selection (unlike with `::marker`, which selects `::before::marker` and `::after::marker`).

Maybe the intention is precisely to make it invalid?

For CSS parser/preprocessor authors, it makes a difference since invalid sub-pseudo-elements makes the selector invalid. I do not need to be explicitly defined, but I just would like to get a confirmation.

jakearchibald: This definition also appears in this PR:

> When a [=pseudo-element=] |pseudo| is part of a [=pseudo-element parent=] |parent|'s [=pseudo-element parent/children=], then |pseudo|'s [=originating pseudo-element=] is |parent|.

But, it's intended that `::view-transition::view-transition-group(name)` is invalid. We might do something with more tree-like selectors later.

cdoublev: I came accross this note in CSS Lists and Counters: *an originating element that is a pseudo-element needs to be explicitly specified in the selector*. So the examples with `html::view-transition-*()` would be incorrect? `::before::marker` would be selected with `::marker` if this were not required, right?

css-view-transitions-1/Overview.bs
```diff
@@ -197,40 +197,55 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
-	The following <dfn export>view-transition pseudo-elements</dfn> represent the various items being animated.
+## Pseudo-element root ## {#pseudo-root}
 
-	The ''::view-transition'' pseudo-element acts as a grouping element for other [=view-transition pseudo-elements=]
-	and has the document's [=document element=] as its [=originating element=].
+	Issue: This definition should be moved to [[css-pseudo-4]].
 
-	Note: For example, '':root::view-transition'' selector matches this pseudo-element,
-	but ''div::view-transition'' does not.
+	A <dfn>pseudo-element root</dfn> is a type of [=tree-abiding pseudo-element=] that is the [=tree/root=] in a [=tree=] of [=tree-abiding pseudo-elements=],
```
khushalsagar, line 204: So `::view-transition` will be the pseudo-element root. Correct? Can we add some prose to clarify how this tree attaches to the DOM. Rough idea below:

"The ancestor of a pseudo-element root is a [node](https://dom.spec.whatwg.org/#node). This node is the root's originating element and the [ultimate originating element](https://www.w3.org/TR/selectors-4/#ultimate-originating-element) for all pseudo-elements in this tree."

khushalsagar: Actually reading further I'm not sure if pseudo-element root is limited to view-transition or any pseudo-element which has descendants. Can you confirm.

css-view-transitions-1/Overview.bs
```diff
@@ -197,40 +197,55 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
-	to specify which elements named with 'view-transition-name' are affected.
+	The [=pseudo-element tree=] defines the document order of its [=tree/descendant=] [=tree-abiding pseudo-elements=].
 
-	There can be multiple pseudo-elements of the same type,
-	one for each 'view-transition-name' participating in a transition.
+	When a [=pseudo-element=] [=tree/participates=] in a [=pseudo-element tree=],
+	its [=originating pseudo-element=] is its [=tree/parent=].
 
-	The <<pt-name-selector>> is defined as follows:
+	If a [=tree/descendant=] |pseudo| of a [=pseudo-element root=] has no other [=tree/siblings=],
```
khushalsagar, line 212: For a moment I got confused about why we're limiting this to only descendants of root. Now I understand, child is used to refer to pseudo-elements which originate directly from the root and descendant is for any pseudo-element in root's tree.

css-view-transitions-1/Overview.bs
khushalsagar: We explicitly resolved on a syntax which allows selecting these pseudo-elements directly from html, for example `html::view-transition-group(foo)` here: https://github.com/w3c/csswg-drafts/issues/7788#issuecomment-1292336905.

css-view-transitions-1/Overview.bs
```diff
@@ -466,6 +486,13 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 			Initially a new [=style sheet=] in the [=user-agent origin=], ordered after the [=HTML user agent style sheet=].
 
 			Note: This is used to hold dynamic styles relating to transitions.
+
+		: <dfn>view-transition root pseudo-element</dfn>
```
khushalsagar, line 490: nit: Do we need this here? We should be able to refer to this by active DOM transition's transition root pseudo-element.

css-view-transitions-1/Overview.bs
cdoublev: Ok, I got it, sorry. Assuming `:div()` exists, `.foo:div(.bar)` would be equivalent to `.foo div.bar`, but there is no such equivalence with view transition pseudo-elements. The note in CSS List and Counters must not apply to functional pseudo-element.

css-view-transitions-1/Overview.bs
jakearchibald: `::view-transition` is the only root. I don't mind adding that extra paragraph, but the originating element for `::view-transition` is already defined later in the spec.

css-view-transitions-1/Overview.bs
jakearchibald: Yeah, I don't think `::view-transition:only-child` makes sense as a thing. Or, if we wanted it to be a thing, its definition would be more complicated.

css-view-transitions-1/Overview.bs
jakearchibald: That's a very good point. In fact, it's kinda silly that this isn't cleared as part of "Clear view transition". I'll fix that up.

jakearchibald: Oh, now I know why.

We don't add the "view-transition root pseudo-element" at the same time as the active transition. The document's "view-transition root pseudo-element" is used to define when this pseudo element is in the document.

jakearchibald: I'll change it to a boolean to avoid duplicated info

css-view-transitions-1/Overview.bs
```diff
@@ -466,6 +482,14 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 			Initially a new [=style sheet=] in the [=user-agent origin=], ordered after the [=HTML user agent style sheet=].
 
 			Note: This is used to hold dynamic styles relating to transitions.
+
+		: <dfn>show view-transition root pseudo-element</dfn>
```
khushalsagar, line 486: We didn't need the bool earlier because the spec ensured the lifetime of ::view-transition root pseudo-element is when this is true.

Now I see that we generate the ::view-transition pseudo-element as soon as the transition object is created.

css-view-transitions-1/Overview.bs
```diff
@@ -1020,27 +1045,24 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 		1. Let |document| be [=this's=] [=relevant global object's=] [=associated document=].
 
-		1. Let |transitionRoot| be the result of creating a new ''::view-transition'' pseudo-element in |document|'s [=snapshot viewport=].
+		1. Set |document|'s [=show view-transition root pseudo-element=] to true.
```
khushalsagar, line 1048: That's it, why the change from generating the root here to assuming it always exists? I found lifetimes easier to follow when this element was generated here (with rest of the pseudos) and torn down in clear view transition.

css-view-transitions-1/Overview.bs
```diff
@@ -1250,18 +1272,17 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 
 		1. [=list/For each=] |capturedElement| of |transition|'s [=ViewTransition/named elements=]' [=map/values=]:
 
 			1. [=list/For each=] |style| of |capturedElement|'s [=captured element/style definitions=]:
 
 				1. If |style| is not null,
+					and |style| is in |document|'s [=document/view transition style sheet=],
 					then remove |style| from |document|'s [=document/view transition style sheet=].
+
+		1. Set |document|'s [=document/active DOM transition=] to null.
```
khushalsagar, line 1285: Shouldn't we reset [=show view-transition root pseudo-element=] here? Otherwise it starts of as true and the next time [=document/active DOM transition=] is set, it's ::view-transition is displayed before Setup View transition.

css-view-transitions-1/Overview.bs
khushalsagar: Ok. I'm fine with this pattern too.

PR #8170, opened 2022-12-01
[selectors-4] Added `:blank` to at-risk list and removed `:read-write` and `:has()` from it

### LINKED ISSUE #8168: [selectors-4] Update at-risk info for `:has()`, `:read-write()`, and `:blank()`
The features listed as at-risk in the [document status section](https://drafts.csswg.org/selectors-4/#sotd) seem outdated.

Since `:has()` is now implemented in WebKit and Chromium, it probably shouldn't be marked as at-risk anymore.
Furthermore, `:read-write` is implemented in all main browser engines for a long time already. So I guess it can also be removed from the list.
And last but not least, there's a note under `:blank` that it's at-risk, there's [still some discussion around it](https://github.com/w3c/csswg-drafts/issues/1967) and there is no implementation so far, though it is currently _not_ listed in the document status. So I believe it should definitely be added there.

Sebastian

SebastianZ:
What the title says. See #8168 for the reasoning.

Fixes #8168

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -27,6 +27,7 @@ At Risk: the column combinator
 At Risk: the '':read-write'' pseudo-class
 At Risk: the '':has()'' pseudo-class
 At Risk: [=user action pseudo-classes=] applying to non-[=tree-abiding=] [=pseudo-elements=]
+At Risk: the '':blank'' pseudo-class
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -27,6 +27,7 @@ At Risk: the column combinator
 At Risk: the '':read-write'' pseudo-class
 At Risk: the '':has()'' pseudo-class
```
SelenIT, line 28: It seems that `:has()` is still there in the at-risk list after this merge?

Issue #8172, opened 2022-12-01
[css-backgrounds] Does stripes() progress inwards or outwards?


Loirooriol:
https://drafts.csswg.org/css-backgrounds-4/#the-border-color

> The stripes defined by [`<image-1D>`](https://drafts.csswg.org/css-images-4/#typedef-image-1d) follow the shape of the border on the side to which they apply, and are drawn in bands starting from the [padding edge](https://drafts.csswg.org/css-box-4/#padding-edge) and **progressing outwards**.

But then in the example the colors seem to progress inwards??

> `border-color: stripes(skyblue, black) stripes(orange, yellow) stripes(yellowgreen, black) stripes(indianred, yellow);`
> ![](https://drafts.csswg.org/css-backgrounds-4/images/multicolor-border.png)

SebastianZ:
Thanks for the hint! The initial idea obviously was to make them progress inwards. Later on it was decided that progressing outwards is better but the examples weren't updated.

I did that now and also tried to improve the colors a little.

Sebastian

Loirooriol:
TBH I have no idea about what kind of blue `dodgerblue` and `skyblue` are, so the example is not great to visualize whether the progress is inwards or outwards. Maybe add little squares with the colors like in https://drafts.csswg.org/css-color-4/#ex-currentcolor

SebastianZ:
Fair point. I've added the previews in https://github.com/w3c/csswg-drafts/pull/8178. (Hope that works because Bikeshed didn't provide a proper review.)

Sebastian

Loirooriol:
Thanks, but another thing, the left side of the 2nd image seems to use #ff8d8b instead of #cd5c5c (indianred).

PR #8178, opened 2022-12-02
[css-backgrounds-4] Added previews for colors

SebastianZ:
Like in CSS Color 4, this adds previews for the colors in the example related to the `stripes()` function.

This was requested in https://github.com/w3c/csswg-drafts/issues/8172#issuecomment-1335791336.


svgeesus (commented):
Seems correct to me

Issue #7825, opened 2022-10-04
[web-animations-2][css-animations-2] What to do with keyframe offsets outside of animations.


flackr:
In https://github.com/w3c/csswg-drafts/issues/7044 we https://github.com/w3c/csswg-drafts/issues/7044#issuecomment-1175722383 to add phase linked offsets to the animations APIs. This allows tying both the animation start/end and keyframes to particular parts of the overall timeline time. However, this also allows accidentally specifying keyframe offsets which are before 0% or after 100% of the animation, e.g.

```css
@keyframes fade-in-animation {
  enter 0% { opacity: 0; }
  enter 100% { opacity: 1; }
}
.elem {
  animation: fade-in-animation;
  animation-timeline: view-timeline;
  animation-delay: enter 50% enter 100%;
}
```

The above is obviously a mistake, though you could imagine cases if lengths or calculations are allowed (e.g. #7575) we could end up in this situation dynamically in more legitimate incidental cases, e.g.

```css
@keyframes fade-in-animation {
  /* When 20% of the enter phase is
   * less than 50px this will be < 0% progress. */
  enter 20% { opacity: 0; }
  enter 100% { opacity: 1; }
}
.elem {
  animation: fade-in-animation;
  animation-timeline: view-timeline;
  animation-delay: enter 50px enter 100%;
}
```

I can think of 4 options for how to handle this:

1. Treat keyframes resolving outside of 0 - 100% as an error. This is what [is specified in css-animations-1](https://www.w3.org/TR/css-animations-1/#keyframes):
   > Values less than 0% or higher than 100% are invalid and cause their `<keyframe-block>` to be ignored.

   and [is also specified by web-animations-1](https://www.w3.org/TR/web-animations-1/#processing-a-keyframes-argument):
   > Specifying an offset outside the range [0.0, 1.0] will cause a TypeError to be thrown.

   However, if we allow length calculated offsets, it will not be possible at animation construction time to know whether a future size of scrolling box will result in an invalid keyframe offset. Cancelling the animation seems dangerous as this state would often be ephemeral. Naively, we could simply skip the animation when this happens (the same as if the timeline was currently inactive). This could result in sudden flips in visible styles when box sizes pass this threshold.

[...]

PR #8202, opened 2022-12-07
[scroll-animations-1] Define keyframe attachment outside animation's effective range #7825

fantasai:


scroll-animations-1/Overview.bs
```diff
@@ -732,6 +732,9 @@ spec: cssom-view-1; type: dfn;
 	If the timeline does not have a corresponding [=named timeline range=],
 	then any keyframes attached to points on that [=named timeline range=]
 	are ignored.
+	If the attachment points are outside the effective range of the animation
+	(e.g. due to 'animation-delay'),
+	the first/last keyframes of the animation are found by interpolation.
```
flackr, line 737: I don't think there's any change needed here, we should use the specified keyframes as is.

What needs to change is [the text in css-animations-1](https://www.w3.org/TR/css-animations-1/#keyframes) where it says:
> If a 0% or from keyframe is not specified, then the user agent constructs a 0% keyframe using the computed values of the properties being animated. If a 100% or to keyframe is not specified, then the user agent constructs a 100% keyframe using the computed values of the properties being animated.

Instead, this needs to only construct 0% or 100% keyframes for properties which don't have keyframes < 0% or greater than 100%.

fantasai: OK, tried rephrasing

flackr (approved):

Issue #4805, opened 2020-02-25
[css-pseudo-4] Custom state pseudo class proposal


atanassov:
Based on recent [TAG review](https://github.com/w3ctag/design-reviews/issues/428) of the proposed [Custom State Pseudo Class](https://wicg.github.io/custom-state-pseudo-class) feature it was surprising that this topic hasn't been discussed with the csswg yet.

Given Chrome has an [intent to ship](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/dJibhmzE73o) this feature it will be helpful for us to at least discuss it.

SelenIT:
Given that this proposal seems to be custom-element-specific, probably [css-shadow-parts-1] would be a better place for it than [css-pseudo-4]?

heycam:
> it was surprising that this topic hasn't been discussed with the csswg yet

Procedurally, is it incumbent on us as CSSWG members to notice specifications developed in WICG that impact CSS, and to then bring them up for discussion here to provide feedback in that venue?

tabatkins:
No, it's supposed to be part of our process to rope in appropriate experts, and in this case would absolutely be a "file an issue in CSSWG" sort of thing. Failing on our part, sorry about that; I didn't realize we were at the ItS stage yet.

smfr:
How does this relate to https://github.com/w3c/csswg-drafts/issues/3431?

FremyCompany:
(subscribing myself to this issue)

fantasai:
CSSWG minutes are here: https://lists.w3.org/Archives/Public/www-style/2020Mar/0009.html
An important comment was to make sure this feature can handle enums, not just boolean states.

tabatkins:
Web Components f2f minutes here: https://github.com/WICG/construct-stylesheets/issues/45#issuecomment-602756608

Overall there's consensus on implementing `document.adoptedStyleSheets` as an `ObservableArray`.

There also appears to be consensus on moving `document.styleSheets` to a (readonly) `ObservableArray` subclass, so it'll sprout the same array methods.

The contentious question that couldn't be resolved on the call, and which is being left for us to decide, is whether direct assignment of an array is allowed; that is, whether `document.aSS = [ss1, ss2];` works or throws.

Points in favor:
[...]

Issue #7431, opened 2022-06-29
[css-values-3] Restrict none/auto/normal from <custom-ident>


fantasai:
We often use these keywords in grammars, so @tabatkins suggested we blanket-restrict them the same way we do all the global keyboards (initial/inherit/etc.)

tabatkins:
Okay, at least for most of these the CSS is already invalid and/or useless. In particular, the CSS counter `auto` appears to not do anything (it looks like they were intending it to be `none`, maybe?).  I can't find the usage of a `none` grid line on the first page listed, so it must be buried in one of the many, many additional requests. (And haven't looked in the others.) Without seeing it, tho, it really seems weird to have a line named `none`, and I suspect it's an attempt to say `[]` - no grid line names.

So, uh, seems like we're pretty reasonably compatible?

tabatkins:
Agenda+ to discuss this change

tabatkins:
After the discussion today, I no longer think this is needed. The breakage potential seems higher than I initially thought, and the use-cases less important - generally, new APIs should use dashed-ident rather than custom-ident anyway. (And if it's safe to use custom-ident because it wont' mix with other keywords, like grid line names, then it's safe to name them none/auto/normal as well.)

PR #8213, opened 2022-12-09
[selectors-5] Add custom state pseudo-classes

josepharhar:
We resolved to add the custom state pseudo-classes to selectors 5 here: https://github.com/w3c/csswg-drafts/issues/4805


josepharhar:
I'm not sure how to link to the definition which will be added to the HTML spec and I'm not sure how much detail we should add, I started with the bare minimum

selectors-4/deferred-for-level-5
```diff
@@ -127,3 +127,14 @@ The local link pseudo-class '':local-link''</h3>
...
 		Should "null segments" count as matching, or not?
+
+<h3 id="custom-pseudo">Exposing custom state: the '':state()'' pseudo-class</h3>
+
+  The <dfn selector>:state()</dfn> pseudo-class takes an argument and matches
+  custom elements whose exposed custom states include the argument.
+
+  The grammar of the '':state()'' pseudo-class is:
+
+  <pre class=prod>:state( <<string>> )</pre>
```
keithamus, line 138: I think this wants to be 

```suggestion
  <pre class=prod>:state( [<ident> | <dashed-ident>] )</pre>
```

josepharhar: Does `<ident>` not include `<dashed-ident>`?

keithamus: Oh yeah you’re right

josepharhar: Should I change it to `<ident>` then? I'm not an expert on CSS grammar. @dbaron

keithamus: Yes I think it needs to be `<ident>`.

```suggestion
  <pre class=prod>:state( <ident> )</pre>
```

selectors-4/deferred-for-level-5
```diff
@@ -127,3 +127,14 @@ The local link pseudo-class '':local-link''</h3>
...
+<h3 id="custom-pseudo">Exposing custom state: the '':state()'' pseudo-class</h3>
+
+  The <dfn selector>:state()</dfn> pseudo-class takes an argument and matches
+  custom elements whose exposed custom states include the argument.
+
+  The grammar of the '':state()'' pseudo-class is:
+
+  <pre class=prod>:state( <ident> )</pre>
+
+  TODO add some text explaining that this is defined in the HTML spec? The :user-valid pseudo class is like this too: https://github.com/w3c/csswg-drafts/pull/9308
```
keithamus, line 140: I think this can be:

```suggestion
 The exact matching behavior of '':state()'' pseudo-class is defined by the host language. See <a spec=html lt="custom" selector>HTML's definition</a>.
```

selectors-4/deferred-for-level-5
```diff
@@ -127,3 +127,14 @@ The local link pseudo-class '':local-link''</h3>
...
 		Should "null segments" count as matching, or not?
+
+<h3 id="custom-pseudo">Exposing custom state: the '':state()'' pseudo-class</h3>
+
+  The <dfn selector>:state()</dfn> pseudo-class takes an argument and matches
+  custom elements whose exposed custom states include the argument.
+
+  The grammar of the '':state()'' pseudo-class is:
+
+  <pre class=prod>:state( <ident> )</pre>
```
keithamus, line 138: Existing WPTs & impls make this case sensitive so:

```suggestion
  <pre class=prod>:state( <custom-ident> )</pre>
```

dbaron: I suspect we don't want the exclusion of CSS-wide keywords that `<custom-ident>` implies; I think we have at least some amount of consensus that such exclusion was a bad idea in the first place and places that needed it should be using `<dashed-ident>` instead, and that `<custom-ident>` was a mistake.  But on the other hand we should clearly document the case-sensitivity.  I'm not sure what the current best practice is.

(I can't find any reference for this longer than https://github.com/w3c/csswg-drafts/issues/7431#issuecomment-1276732615 but there might be a more detailed explanation somewhere.)

josepharhar: so should we do `<dashed-ident>` for now...? or keep it as `<ident>`?

keithamus: `<dashed-ident>` is definitely wrong as this can take any ident. I suppose `<ident>` is right from a strictly grammar standpoint because, as @dbaron points out, it doesn’t exclude css wide keywords. `<ident>` claims case insensitivity though. Perhaps we could make a new token type here like `<cased-ident>`. Otherwise we should keep it as `<ident>` and either explain (or remove) the implementation handles cases discretely.

josepharhar: I'm not sure what is best here. @tabatkins can you advise?

tabatkins: > `<ident>` claims case insensitivity though.

Where are you seeing that? If there's any such statement it's an error and I can fix it; <https://drafts.csswg.org/css-values/#textual-values> defines `<ident>` as the union of keywords and author-defined identifiers, and since the latter are case-sensitive the `<ident>` must be too, by default.

Anyway, yeah, `<ident>` is the correct production to use here if the value set is *not* dashed and the CSS-wide keyword restriction isn't appropriate (both of which are true here).

keithamus: I can’t find the reference to case insensitivity now. 

The string given to customstateset can be dashed so should this be `<ident> | <dashed-ident>`?

tabatkins: No, `<dashed-ident>` is a subset of `<ident>`, so just `<ident>` is sufficient.

josepharhar: Ok so since i'm currently using `<ident>` can we merge this?

selectors-4/deferred-for-level-5
```diff
@@ -127,3 +127,14 @@ The local link pseudo-class '':local-link''</h3>
 		What about if the document url has less than N segments,
 		and the link is same-page?
 		Should "null segments" count as matching, or not?
+
+<h3 id="custom-pseudo">Exposing custom state: the '':state()'' pseudo-class</h3>
```
dbaron, line 131: I'd suggest `custom-state` as the ID rather than `custom-pseudo`.

selectors-4/deferred-for-level-5
dbaron: I think you should add a sentence saying it's case-sensitive.  (unless @tabatkins disagrees, at least.)

dbaron (commented):
This seems close to ready to merge, though I have a few comments below.

It's also not clear to me how useful this `deferred-for-level-5` file is -- at some point we should have a real draft up on `drafts.csswg.org`.

(Did you make any attempt to run `bikeshed` on what you added here?)

keithamus:
I added it regarding https://github.com/w3c/csswg-drafts/pull/8213#discussion_r1478885668 but that has since been resolved. Apologies for the noise.

selectors-4/deferred-for-level-5
josepharhar: I amended the first sentence to say that the argument is case sensitive

selectors-4/deferred-for-level-5
josepharhar: I changed it to "custom-state"

josepharhar:
> It's also not clear to me how useful this deferred-for-level-5 file is -- at some point we should have a real draft up on drafts.csswg.org.

Good point, I moved it to overview.bs

> (Did you make any attempt to run bikeshed on what you added here?)

I just did and it works

dbaron:
> > It's also not clear to me how useful this deferred-for-level-5 file is -- at some point we should have a real draft up on drafts.csswg.org.
> 
> Good point, I moved it to overview.bs

I don't think this is what we want either -- the resolution was to add to level 5, not to the (substantially more stable) level 4.  We should probably get an actual selectors-5 draft going -- I'm not sure whether it should contain all or some or none of the material currently in the `deferred-for-level-5` file.

(I admit that this makes getting the change landed more complicated!)

josepharhar:
> I don't think this is what we want either -- the resolution was to add to level 5, not to the (substantially more stable) level 4.

Thanks, I put it back in the deferred for level 5 file

> We should probably get an actual selectors-5 draft going -- I'm not sure whether it should contain all or some or none of the material currently in the deferred-for-level-5 file.

What is the process for this? Can I do it in a separate PR?

dbaron:
> What is the process for this? Can I do it in a separate PR?

A separate PR would probably be good.  I think it would make sense to move the deferred-for-level-5 file to be an actual level 5 draft.  (Some of the material might need notes about its current state/status.)

josepharhar:
> A separate PR would probably be good. I think it would make sense to move the deferred-for-level-5 file to be an actual level 5 draft. (Some of the material might need notes about its current state/status.)

Great! Is there anything left to do to merge this?

dbaron (approved):

Issue #7901, opened 2022-10-18
[scroll-animations-1] What's the point of animation-range?


Loirooriol:
https://w3c.github.io/csswg-drafts/scroll-animations/#named-range-animation-declaration

> The [animation-delay](https://w3c.github.io/csswg-drafts/scroll-animations/#propdef-animation-delay) and [animation-range](https://w3c.github.io/csswg-drafts/scroll-animations/#propdef-animation-range) properties are both [shorthands](https://w3c.github.io/csswg-drafts/css-cascade-5/#shorthand-property) that set [animation-delay-start](https://w3c.github.io/csswg-drafts/scroll-animations/#propdef-animation-delay-start) and [animation-delay-end](https://w3c.github.io/csswg-drafts/scroll-animations/#propdef-animation-delay-end) together in a single declaration.

What's the point of having 2 shorthands if both have the same grammar and expand to the same set of longhands in the same way?

@cdoublev also confused in https://github.com/w3c/csswg-drafts/issues/7044#issuecomment-1252092947

@bramus said

> Otoh I do understand the need for animation-range: it is more readable in the content of scroll-linked animations. E.g. `animation-range: exit` is easier to grasp than `animation-delay: exit`.

But then maybe restrict `animation-delay` to `<time>{1,2}#` and `animation-range` to `[ <timeline-range-name> <percentage>? ]{1,2}#` or something?

ydaniv:
Note that the [`animation` shorthand](https://drafts.csswg.org/css-animations-1/#animation)'s second `<time>` value is parsed as `animation-delay`, so unless we add `animation-range` to the definition of [`animation`](https://drafts.csswg.org/css-animations-2/#animation-shorthand) we can't restrict `animation-delay` from accepting range values.
Otherwise we can't declare a scroll animation with `animation`.

But at the same time, any `animation-duration` in a scrub animation (scroll or other) is meaningless. It's always `100%`.
Since I'm not an expert on values/parsing, would it make sense to have an alternative definition of `animation` that contains only ranges and no `<time>`? (@fantasai , @flackr)

Loirooriol:
Once `animation-delay` becomes a shorthand, `animation` will need to expand into the longhands: `animation-delay-start` and `animation-delay-end`. So changing `animation-delay` doesn't have to affect `animation`, it depends on how it's defined.

[...]

PR #8219, opened 2022-12-13
[scroll-animations-1] separate animation-range and animation-delay #7901

fantasai:


scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,79 @@ spec: cssom-view-1; type: dfn;
 
 	A set of animation keyframes can be attached
 	in reference to [=named timeline ranges=]
-	by adjusting the duration of the animation.
+	by restricting the to the specified timeline range.
+	Delays ('animation-start-delay'/'animation-end-delay')
+	are set within this restricted range.
+	If a duration ('animation-duration') is given,
+	it measures from the range-restricted start,
+	and further restricts the range-restricted end.
```
flackr, line 744: It is just used instead rather than replacing it, isn't it? E.g. similar to how width is just used instead of right for absolute positioning regardless of whether that further restricts or expands the width: https://jsbin.com/bekiceg/edit?html,css,output

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,79 @@ spec: cssom-view-1; type: dfn;
...
 
-	The 'animation-delay' and 'animation-range' properties
-	are both [=shorthand property|shorthands=]
-	that set 'animation-delay-start' and 'animation-delay-end'
+	The 'animation-range' property is a [=shorthand property|shorthand=]
+	that sets 'animation-range-start' and 'animation-range-end'
 	together in a single declaration.
-	If the <<'animation-delay-end'>> value is omitted,
+	If the <<'animation-range-end'>> value is omitted,
 	it is set to zero.
```
flackr, line 755: What does zero mean for range? I think we need to use auto and ensure that auto has no effect on time-based animations.

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,79 @@ spec: cssom-view-1; type: dfn;
...
+	'animation-range-start' is set to that name plus ''0%''
+	and 'animation-range-end' is set to that name plus ''100%''.
 
 	<pre class="propdef">
-		Name: animation-delay-start
-		Value: [ <<time>> | <<timeline-range-name>> <<percentage>> ]#
-		Initial: 0s
+		Name: animation-range-start
+		Value: [ normal | <<timeline-range-name>> <<percentage>> ]#
+		Initial: normal
```
flackr, line 763: I think maybe you meant auto here to align with the auto <dfn> below?

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,79 @@ spec: cssom-view-1; type: dfn;
...
+	(i.e. when keyframes attached to 0% progress are mapped)
+	to the specified position on the timeline.
 
-	<dl dfn-for="animation-delay-start, animation-delay-end" dfn-type=value>
-		<dt><dfn><<time>></dfn>
+	<dl dfn-for="animation-range-start" dfn-type=value>
+		<dt><dfn>auto</dfn>
 		<dd>
-			Defines the delay as a duration. See [[CSS-ANIMATIONS-1]].
+			The start of the animation attaches to the start of the timeline.
```
flackr, line 778: Since auto is the default value it will have to have the current behavior for regular time based timelines. That is,

If the timeline is monotonically increasing:
- The start of the animation attaches to the current time of the timeline.
- Otherwise, the start of the animation attaches to the start of the timeline.

This is essentially restating the behavior specified in steps 5-7 of playing an animation:
https://drafts.csswg.org/web-animations-1/#playing-an-animation-section

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,79 @@ spec: cssom-view-1; type: dfn;
...
 
-	Defines a delay between the end of the animation
-	and when it ends executing
-	(i.e. when keyframes attached to 100% progress are mapped).
-	Values are as for 'animation-delay-start',
-	but calculating backwards from the end,
-	i.e. a ''2s'' delay effectively shortens the animation duration,
-	whereas ''-2s'' lengthens it.
+	Shifts the end of the animation
+	(i.e. when keyframes attached to 100% progress are mapped)
```
flackr, line 798: This is true if iteration count = 1 and end delay is 0, but if iteration count > 1 there will be multiple points at which 100% is reached and if end delay is > 0 100% will be reached before animation-range-end. This is the [end time](https://drafts.csswg.org/web-animations-1/#end-time) of the animation.

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,79 @@ spec: cssom-view-1; type: dfn;
...
+	to the specified position on the timeline.
+
+	<dl dfn-for="animation-range-start, animation-range-end" dfn-type=value>
+		<dt><dfn>auto</dfn>
+		<dd>
+			The end of the animation attaches to the end of the animation duration.
```
flackr, line 804: Same comment as for range-start.

scroll-animations-1/Overview.bs
fantasai: Oh, right. That's an error left over from removing the time values of delay, I'll fix it.

scroll-animations-1/Overview.bs
fantasai: I think it would make more sense to take the more restrictive one, since if you set a shorter animation-range-end and a longer duration, it'd seem weird to ignore the animation-range-end?

scroll-animations-1/Overview.bs
fantasai: I meant normal and failed to make it match below. :/ Fixed.

scroll-animations-1/Overview.bs
fantasai: I switched it to “The [=end time=] of the animation is determined as normal.”

scroll-animations-1/Overview.bs
fantasai: OK, I *think* I fixed this.

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,79 @@ spec: cssom-view-1; type: dfn;
 
 	A set of animation keyframes can be attached
 	in reference to [=named timeline ranges=]
-	by adjusting the duration of the animation.
+	by restricting the to the specified timeline range.
```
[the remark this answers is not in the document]
fantasai: Ok, updated terminology. Let me know if I got it right yet. (I also inserted a diagram.)

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,95 @@ spec: cssom-view-1; type: dfn;
 
 	A set of animation keyframes can be attached
 	in reference to [=named timeline ranges=]
-	by adjusting the duration of the animation.
+	by restricting the [=animation active phase=]
+	to the specified timeline range.
+	Delays ('animation-start-delay'/'animation-end-delay')
```
ydaniv, line 741: Shouldn't these be:
```suggestion
	Delays ('animation-delay-start'/'animation-delay-end')
```
?

fantasai: Yes. :/ Switched to “(see 'animation-delay')”.

scroll-animations-1/Overview.bs
ydaniv: I'm confused by the interaction with duration here. Ranges relate to `ViewTimeine`s,  and these don't have a specified duration (unless we specify interaction with `animation-iteration-count`) but rather a duration of `auto` or `100%`, right?
So unless there's some sort of `animation-iteration-count: infinite/auto` i'd expect `animation-duration` to be ignored and be calculated from ranges and delays.

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,96 @@ spec: cssom-view-1; type: dfn;
...
+			The <a spec="web-animations-1">end time</a> of the animation
+			is determined as normal.
+
+		<dt><dfn><<timeline-range-name>> <<percentage>></dfn>
+		<dd>
+			The <a spec="web-animations-1">end time</a> of the animation
+			is at the specified point on the [=timeline=]
+			minus any applicable end delay,
+			or at the end of the animation duration,
+			whichever comes first,.
```
ydaniv, line 827: nit:
```suggestion
			whichever comes first.
```

fantasai: Fixed.

scroll-animations-1/Overview.bs
fantasai: @ydaniv That's another approach we could take: if animation-range-end has a valid non-normal value, ignore the duration. (It seemed the discussion in an earlier issue wanted them to both apply somehow.)

ydaniv: Yes, I thought that's the main use-case, so you're saying range-end could be `normal` and duration in <length> can determine the actual duration, staring from the range-start? 
SGTM, I'd also prefer to respect the range-end and clamp to that

flackr: So taking the shorter of `<range-end> - <range-start>` and `<duration>` still isn't really honoring the request:
- If the range is shorter, you're not using the specified duration
- If the duration is shorter, the animation doesn't end at range-end.

If we don't need to follow the precedent for box positioning, I'd suggest that when a start and end range are both not `normal` we use that instead of the duration.

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,96 @@ spec: cssom-view-1; type: dfn;
...
+	If a duration ('animation-duration') is given,
+	it extends from the end of the start delay
+	until the start of the end delay,
+	and further restricts the range-restricted end.
+
+	<pre class="ascii-art">
+	range start┐             ╺┉┉┉┉active phase┉┉┉┉╸           ┌range end
+	┄┄┄┄┄┄┄┄┄┄┄├─────────────╊━━━━━━━━━━━━━━━━━━━━╉───────────┤┄┄┄┄┄┄┄┄
+               ╶┄start delay┄╴                    ╶┄end delay┄╴
+                             ╶┄┄┄┄┄ duration┄┄┄┄┄┄╴
```
flackr, line 753: I think this diagram might have some alignment issues, this is what I'm seeing locally when viewing the spec:
Screenshot 2023-02-02 at 3 19 44 PM
The start delay, duration and end delay are meant to line up exactly with the ranges and active phases I assume.

scroll-animations-1/Overview.bs
```diff
@@ -736,67 +736,96 @@ spec: cssom-view-1; type: dfn;
...
+	to the specified position on the timeline.
 
-	<dl dfn-for="animation-delay-start, animation-delay-end" dfn-type=value>
-		<dt><dfn><<time>></dfn>
+	<dl dfn-for="animation-range-start" dfn-type=value>
+		<dt><dfn>normal</dfn>
 		<dd>
-			Defines the delay as a duration. See [[CSS-ANIMATIONS-1]].
+			The start of the [=active phase=]
+			is determined as normal.
```
flackr, line 790: I prefer `auto` as it seems more in line with how implicitly computed defaults are specified elsewhere and aligns with `animation-duration: auto`.

I believe how we determine the start time of the animation should be updated in [web-animations-2 3.3.7. Playing an animation](https://drafts.csswg.org/web-animations/#playing-an-animation-section) to account for range start and end. This currently explains how the "normal" behavior gets a 0 start time for scroll driven animations and the current timeline time for time-based animations. The logic there should be updated to account for range start and range end and then the css-animations spec can refer to that procedure to play the animation.

flackr (approved):
Sorry for the delay. I don't know why but this review doesn't show up in PR's awaiting my review (https://github.com/w3c/csswg-drafts/pulls?q=is%3Apr+is%3Aopen+user-review-requested%3A%40me) so i have to search for it every time.

Overall I think this is looking really close now and if you want to move my particular comments to followup issues that's fine with me.

scroll-animations-1/Overview.bs
ydaniv: +1

scroll-animations-1/Overview.bs
fantasai: I went with `normal` because (unlike duration) we aren't magically computing a range: if the range is not otherwise defined, the timeline range is the whole timeline, as normal.

scroll-animations-1/Overview.bs
fantasai: OK, I'll open a new issue about this. -> https://github.com/w3c/csswg-drafts/issues/8405

scroll-animations-1/Overview.bs
fantasai: Filed as https://github.com/w3c/csswg-drafts/issues/8406

scroll-animations-1/Overview.bs
fantasai: Tabs vs spaces issue. :/ Fixed now.

Issue #5332, opened 2020-07-16
[css-position-3] Circular definition of fixed positioning containing block


zjopy:
(For simplicity, I'll assume continuous media below.)

The definition of the fixed positioning containing block in [css-position-3](https://drafts.csswg.org/css-position-3/#fixed-positioning-containing-block) reads

> If no ancestor establishes [an fixed positioning containing block], the fixed positioning containing block is [...] the viewport.

This definition is circular. You can't define "fixed positioning containing block" assuming the existence of it in the preceeding condition "If no ancestor establishes [an fixed positioning containing block]".

This contradicts the definition above, which [says](https://drafts.csswg.org/css-position-3/#valdef-position-fixed)

> [...] a fixed positioning containing block (the viewport in continuous media, or the page area in paged media)

IMO the term "fixed positioning containing block" is a needless complication for defining the containing block of fixed positioned elements. The containing block for fixed positioned elements should be defined directly as viewport, instead of as "fixed positioning containing block" which is then defined as viewport. It introduces an additional term that just complicates the definition. Further, this term contains the very words that you are trying to define - "containing block" - which is either a logical error or at least bad practice.

It's as if you define the word "car" as "an object with a car door" and then leave out the definition of "car door".

Loirooriol:
I guess it should just be

> The containing block is established by the nearest ancestor box that establishes a<del>n</del> fixed positioning containing block [...]
> If no ancestor establishes one, the <del>fixed positioning</del> containing block is: [...]

zjopy:
Yes, that would fix the circularity.

Although then the [comment](https://drafts.csswg.org/css-position-3/#valdef-position-fixed) in the parentheses would need to be removed as well, since the fixed positioning containing block is *not* the viewport.

tabatkins:
You're right that the parenthetical was incorrect; I've amended it to say "usually".

However, the definition of FPCB is not circular. The reference in question is to "establishes a fixed positioning containing block", which is a term of art that some properties invoke; this section defines what the FPCB is when this happens (a rectangle determined identically to the abspos containing block on the element). If no element "establishes a FPCB", then the definition defines how to calculate the FPCB from the viewport/page instead. That's a well-founded definition.

PR #8246, opened 2022-12-20
[css-position-3] Move definition of "absolute positioning containing block"

mikke89:
The term "absolute positioning containing block" (APCB) previously linked to a usage of the term, rather than its definition. Instead, with this change, it now links to the description of when an APCB is established. This should remove some confusion.

This confusion is partially discussed in [#5332](https://github.com/w3c/csswg-drafts/issues/5332), where the term "fixed positioning containing block" (FPCB) is also discussed. Since there is no normative description of anything that establishes an FPCB (only a vague note), an equivalent change cannot be made for FPCB without some larger changes.


fantasai:
Nice, thanks. :)

PR #8254, opened 2022-12-22
[cssom-view-1] Extend 'scroll into view' steps to Ranges

bokand:
[cssom-view-1] Extend 'scroll into view' steps to Ranges

Generalizes The `scroll an element into view` algorithm to work also work for a Range as well as an Element. This is trivial as the only use of Element is to call `getBoundingClientRect` which Range also defines.

Additionally exports this algorithm for use in https://github.com/whatwg/html/issues/8282.

Nit: also fixes erroneous reference to a non-existent `box` variable where `scrolling box` was intended.


bokand:
@smfr or @emilio PTAL

smfr (approved):

Issue #7387, opened 2022-06-19
[css-flexbox][css-grid] Providing authors with a method of opting into following the visual order, rather than logical order


rachelandrew:
**A proposal to provide a method in CSS for authors to opt in to following the visual order when in a flex or grid context.**

The [grid](https://drafts.csswg.org/css-grid/#order-accessibility) and [flexbox](https://drafts.csswg.org/css-flexbox-1/#flow-order) specifications detail that reordering caused by these methods is only visual, and note that use of order or grid placement to do logical ordering is non-conforming.

With care from the author in creating a good document, this in general works well for screen reader users. Authors sometimes also use this to improve the experience for screenreader users. For example, by being able to show something visually near the top of the page but, because it would be repetitive for a screen reader user, physically locate it in elsewhere in the document.

### Visual/tab order disconnect for keyboard users

In most cases, the document order is the correct logical order for screen reader users who are not able to see the visual display at all. However, if the author has reordered content using grid or flexbox, then the tab order of the document for keyboard users who are referencing the visual display can become disconnected, as it follows the document order.

Some use of grid layout in particular can cause reordering without input from the author. For example using `grid-auto-flow: dense`. In this case a screen reader user is likely well served by the document order, the keyboard navigator may experience unexpected jumps around the component as items appear out of logical order.

Example: https://codepen.io/rachelandrew/pen/NWbLRQE 

Authors also want to be able to reorder content at different breakpoints, to provide the best presentation of the layout for different amounts of screen real estate. In this case again, it is likely that by starting with a sensible document source, a screen reader user will not be affected. A keyboard navigator could experience big jumps around the layout if, for example, an item that logically appears near the top of the source is placed in the footer.

Example: https://codepen.io/rachelandrew/pen/xxRJBGz 

### Impact on screen reader users if the author assumes the visual order is the logical order

Screen reader users can be impacted by the issue **if the author assumes the visual order is the logical order**. For example, someone building a site using a visual design tool that lets them drag elements about and places them using grid layout. In this scenario the author may never see the document order. If that tool does not also modify the source to reflect the visual layout, the screen reader user is going to hear things in a different order than the author intended.

In addition, tools that use the CSS order to infer logical order, will affect screen reader and keyboard navigators alike. Examples:

- [Twitter thread](https://twitter.com/jlengstorf/status/1531506923811483648) (the demo uses order to re-order things by interest, it was pointed out this may have accessibility issues)
[...]

PR #8257, opened 2022-12-23
[css-display-4] Define 'reading-order: auto' #7387

fantasai:


css-display-4/Overview.bs
```diff
@@ -1149,10 +1149,22 @@ Reading Order: the 'reading-order' property</h3>
...
-	It takes a single <dfn value for=reading-order><<integer>></dfn> value,
-	which specifies which ordinal group the item belongs to.
-	Sibling elements are ordered starting from the lowest numbered ordinal group and going up;
-	elements with the same ordinal group are keep the order they appear in the source document.
+	This is their <dfn>reading and navigation order</dfn>.
+
+	<dl dfn-type=value dfn-for=reading-order>
+		<dt><dfn>auto</dfn>
+		<dd>
+			Same as ''0'' except that in randomizing formatting contexts
```
Loirooriol, line 1157: I wonder if people may want `auto || <integer>` instead of restricting to 0.

css-display-4/Overview.bs
```diff
@@ -1149,10 +1149,22 @@ Reading Order: the 'reading-order' property</h3>
...
+			(which are only reasonable to apply to inherently unordered content)
+			the reading order is matched to the layout order.
```
Loirooriol, line 1159: This doesn't explain the relative order of items with `auto` and `0`.

css-display-4/Overview.bs
```diff
@@ -1176,6 +1188,39 @@ Reading Order: the 'reading-order' property</h3>
...
+
+			<div class="example">
+				For example, [=grid items=] are laid out in a ''grid-auto-flow: column dense'' grid,
+				they are first re-ordered to [=order-modified document order=]
+				before applying the [[css-grid-1#auto-placement-algo]].
+				Among the subset of these items whose ''reading-order'' is ''0'' or ''auto'',
+				those with ''reading-order: auto'' are shifted
+				to the latest position in the [=reading and navigation order=]
+				that is also after any items with ''reading-order: 0'' or ''reading-order: auto''
+				that came earlier in the [=grid-modified document order=]
```
Loirooriol, line 1219: What's the grid-modified document order? Grid is 2D, so I don't really understand how to map it to a 1D sequence.

css-display-4/Overview.bs
fantasai: That's an interesting idea...

css-display-4/Overview.bs
fantasai: That's what the cross-reference on the next line is for...?

css-display-4/Overview.bs
fantasai: This is a defined term in the Grid spec.

css-display-4/Overview.bs
```diff
@@ -1176,6 +1188,39 @@ Reading Order: the 'reading-order' property</h3>
 
 	ISSUE: Is this the most appropriate interaction with with <code>tabindex</code>?
 
+<h4 id="randomizing-layout">
+Randomizing Layout Methods and ''reading-order: auto''</h4>
+
+	<dfn>Randomizing formatting contexts</dfn>
+	are formatting contexts in which which can indiscriminately scramble the layout order,
```
tabatkins, line 1195:
```suggestion
	are formatting contexts which can indiscriminately scramble the layout order,
```

css-display-4/Overview.bs
tabatkins, line 1157:
```suggestion
			Same as ''0'' except that in [=randomizing formatting contexts=]
```

css-display-4/Overview.bs
```diff
@@ -1163,6 +1198,39 @@ Reading Order: the 'reading-order' property</h3>
...
+
+	<dfn>Randomizing formatting contexts</dfn>
+	are formatting contexts that can indiscriminately scramble the layout order,
+	as listed below.
+	Since use of such methods indicates that the underlying order of the elements
+	is not important,
+	the UA will match the [=reading and navigation order=] of the elements
+	to match the layout order of their boxes as follows:
+
+	<dl>
```
rachelandrew, line 1212: What else will be included here? Masonry for grid seems likely. What about flex-direction: *-reverse?

rachelandrew:
@fantasai I've been discussing this with @chrishtr and @bfgeek and I wanted to bring that back to this PR.

TL:DR is that we are wondering if we only need `reading-order:auto`  (on the children) and if there are reasons to do `reading-order: <integer>` at all, or initially. Do compelling enough use cases exist to have that level of control or do people really just want to follow the layout?

We also wonder what will happen with row and column reversing in flexbox as people may expect different behavior there.

Longer notes below:

My original suggestion was a switch on the grid/flex container. In that scenario all of the following would follow the grid or flex modified order:

- flex layout with order
- flex layout with *-reverse
- grid layout doing placement with lines/template areas
- grid layout with order
- "randomized grid layouts" with dense packing or masonry

In the existing edits plus this PR, the `reading-order: auto` value is applied to the children, and only covers:

- "randomized grid layouts" with dense packing (and I'm assuming masonry)

To do the following you would need `reading-order: <integer>` on the children:

- flex layout with order
- flex layout with *-reverse
- grid layout doing placement with lines/template areas
- grid layout with order

There are definitely use cases for people wanting to follow the layout created by placement (usually because they want different layouts at different responsive breakpoints), so if we only do auto as specced, we don't have a method to achieve those use cases, and if we go for the switch on the container that likely stops us doing this version later.

An alternative, however, would be if the only value for `reading-order` was `auto` (applied to the children) and `auto` applied in all layout types, not just the randomized ones. As that shouldn't roadblock us from adding the integer value in the future.

Reading order and layout order being separated, is a use case. For example, [this](https://twitter.com/hovhaDovah/status/1531681075562520579?s=20&t=IW1Qq-7k-daMcKb4LkYU6Q), but I think in that case, if we went with auto applied everywhere that doesn't need to affect layout-order with no reading-order.

I feel that having the property on the children is better than my idea of a switch on the parent (though I guess we could then add a switch on the parent that sets them all as a group, as with align-items), because it allows for more flexibility, but I think we could get away with that initially being auto only.

fantasai:
@rachelandrew One of my goals here was to make it impossible for an author to apply `* { use-layout-order-as-speech-order }` and thereby convince themselves that a) they don't need to manage source order correctly and b) they've solved accessibility properly for the whole page. That was my primary objection to your original proposal, and why I proposed something different that requires the author to consider what they're doing for each level of layout more explicitly.

css-display-4/Overview.bs
fantasai: Only randomized orders, so masonry might (does it actually randomize?) but flex-direction: *-reverse wouldn't.

rachelandrew:
> One of my goals here was to make it impossible for an author to apply `* { use-layout-order-as-speech-order }` and thereby convince themselves that a) they don't need to manage source order correctly and b) they've solved accessibility properly for the whole page. 

I get that, however it does seem odd to be making people do more work just in case they do something bad. Having `reading-order: <integer>` is only useful in those cases where the developer doesn't want the document order OR the grid/flex-modified order. I can't think of any good cases for that, so we are adding this additional complexity purely to make it harder for developers to do a thing they want to do. Do we do this anywhere else?

I think it would be worth bringing this issue back to the WG, my comment above and @fantasai's response explain the issue reasonably well I think.

chrishtr:
Agenda+ to consider @rachelandrew 's proposed simplification of the proposal. I agree with her arguments.

Loirooriol:
I still don't understand what's supposed to happen in grid. Grid is 2D, this is 1D, it's not straightforward.

"relative order of an item" is not well defined when there are so many different order to choose from. "grid placement order" makes no sense when grid placement has different steps that iterate different sets of items in different orders. I don't understand "grid layout doing placement with lines/template areas". "grid modified order" is not a single thing, the major axis needs to be specified.

Issue #6950, opened 2022-01-14
[css-text-4] Make ideograph-alpha and ideograph-numeric part of text-spacing: normal


hfhchan:
The baseline behaviour of text-spacing should always turn on ideograph-alpha and ideograph-numeric when in the horizontal writing direction. In the vertical direction, it should be turned on when the western text is rotated[^1] and turned off if the western text is not[^2].

Microsoft Word has ideograph-alpha/ideograph-numeric turned on by default. The iOS UI also has this spacing, but I'm not sure if it's done on the rendering level, or spaces are injected during string interpolation.

On other platforms, it is not supported out of the box. Authors end up adding spaces themselves inconsistently, leading to horrible UIs like this:
<img width="818" alt="image" src="https://user-images.githubusercontent.com/8191296/149442034-483ad4c1-3a78-4d25-9e03-d0ce855d00ad.png">

CSS should support this turned on as the default.

There may also need to be an additional property for authors to specify the width of the inserted space they want.  The default 1/4 space is fine.

[^1]: See [CLREQ Figure 8](https://www.w3.org/TR/clreq/images/en/latin-90-clockwise.svg).
[^2]: See [CLREQ Figure 7](https://www.w3.org/TR/clreq/images/en/latin-one-by-one.svg).

frivoal:
There's a double question here:
* what is good behavior
* what is compatible with existing content

I believe you're right about what's good, but given that it is not the historical behavior and that plenty of content was written assuming it wouldn't happen, this is likely not to be compatible.

We could maybe write an informative note giving a suggestion of an algorithm for the UA-dependent `auto` value, which could do what you suggest, including the dependencies on writing-modes, but I suspect we cannot make that behavior the default. If you'd like to insist on it, it would take some evidence that this is web compatible.

hfhchan:
I tried imagining scenarios where changing to more desirable defaults would make it incompatible with existing content, but I cannot come up with much.

Content which already has extra space characters injected wouldn't be affected, because ideograph-alpha and ideograph-numeric do not insert additional spaces between characters and spaces.

One case which may potentially be affected would be designs that emulate a terminal with mixed western/CJK text.

[...]

PR #8262, opened 2022-12-29
[css-text-4] interscript text-spacing as default

fantasai:
This pull request makes several changes in response to https://github.com/w3c/csswg-drafts/issues/6950:

- Adds `ideograph-alpha` and `ideograph-numeric` to the initial value, `normal`.
- Adds a UA stylesheet rule to disable `text-spacing` in `pre`, `code`, and `tt` to preserve their monospace behavior.
- Specifies that non-zero margin/border/padding inhibits autospacing.
- Clarifies that intervening characters such as spaces also inhibit autospacing.
- Defines the interscript autospace to be 1/8 of the ideographic advance (`0.125ic`), which is the smallest amount in the common-use range. The primary justification for using 1/8 instead of 1/6 or 1/4 is that, since we're trying to make this the default, it minimizes the layout impact on existing pages. (Note that Antenna House uses a 1/4 default.)

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -7991,7 +7991,7 @@ Character Class Spacing: the 'text-spacing' property</h3>
 		<dt><dfn>normal</dfn>
 		<dd>
 			Specifies the baseline behavior,
-			equivalent to ''space-start trim-end trim-adjacent''.
+			equivalent to ''space-start trim-end trim-adjacent ideograph-alpha ideograph-numeric''.
 
 		<dt><dfn>none</dfn>
 		<dd>
@@ -9291,6 +9291,9 @@ Default UA Stylesheet</h2>
 
 			/* do not allow white space to collapse in textarea */
 			textarea { text-space-collapse: preserve !important; }
+
+			/* preserve character grid in preformatted text */
+			pre, code, kbd, samp, tt { text-spacing: none; }
 		</pre>
 	</div>
 	<!--
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -8107,12 +8107,14 @@ Character Class Spacing: the 'text-spacing' property</h3>
 		<dt><dfn>ideograph-alpha</dfn>
 		<dd>
 			Creates extra spacing between runs of
-			<a>ideographs</a> and <a>non-ideographic letters</a>.
+			<a>ideographs</a> and <a>non-ideographic letters</a>,
+			see [[#inter-script-spacing]].
 
 		<dt><dfn>ideograph-numeric</dfn>
 		<dd>
 			Creates extra spacing between runs of
-			<a>ideographs</a> and <a>non-ideographic numerals</a> glyphs.
+			<a>ideographs</a> and <a>non-ideographic numerals</a>,
+			see [[#inter-script-spacing]].
 
 		<dt><dfn>punctuation</dfn>
 		<dd>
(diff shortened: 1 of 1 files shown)


fantasai:
CC @MurakamiShinyu (wanted to request your review, but your account wasn't linked up to the repository...)

css-text-4/Overview.bs
```diff
@@ -9291,6 +9291,9 @@ Default UA Stylesheet</h2>
 
 			/* do not allow white space to collapse in textarea */
 			textarea { text-space-collapse: preserve !important; }
+
+			/* preserve character grid in preformatted text */
+			pre, code, tt { text-spacing: none; }
```
frivoal, line 9312: Here are some extract from the HTML UA stylesheet:
https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3 :
```css
listing, plaintext, pre, xmp {
  font-family: monospace; white-space: pre;
}
```
https://html.spec.whatwg.org/multipage/rendering.html#phrasing-content-3 :
```css
code, kbd, samp, tt { font-family: monospace; }
```
This makes me think that the selector here should be expanded a little, from `pre, code, tt` to `listing, plaintext, pre, xmp, code, kbd, smp, tt`. Some of these tags are obsolete and only in the HTML UA stylesheet for compat reasons, but `tt` is one of those, and was already included, so we should probably go with the whole set.

css-text-4/Overview.bs
```diff
@@ -8153,6 +8149,19 @@ Character Class Spacing: the 'text-spacing' property</h3>
...
+Inter-script Spacing</h4>
+
+	The ''ideograph-alpha'' and ''ideograph-numeric'' values
+	introduce spacing at the boundary between particular classes of characters
+	when they are directly adjoining on a line,
+	i.e. without any intervening non-zero [=margin=], [=border=], or [=padding=]
+	or intervening characters (such as a quotation mark or a space).
+
+	Note: Spacing conventions vary, but values typically range from 1/4ic to as low as 1/8ic,
+	with 1/4ic being particularly common in monospace contexts
```
frivoal, line 8164: "particularly common in monospace contexts" may be true, though I don't know it personally.

However, an explanation I have more often heard is that 1/4em was common in metal type letterpress and movable type printing (presumably because spacer blocks thinner than that would have been too fiddly to deal with).

css-text-4/Overview.bs
```diff
@@ -8153,6 +8149,19 @@ Character Class Spacing: the 'text-spacing' property</h3>
...
+	and 1/6ic being more common in proportional typesetting.
```
frivoal, line 8165: I'd go with "1/6ic or thinner" rather than "1/6ic". Then 1/8 value that CSS ends up picking is on the thinner side, but not outside of common practice.

css-text-4/Overview.bs
```diff
@@ -8157,10 +8157,17 @@ Inter-script Spacing</h4>
 	when they are directly adjoining on a line,
 	i.e. without any intervening non-zero [=margin=], [=border=], or [=padding=]
 	or intervening characters (such as a quotation mark or a space).
+	The amount of space introduced by these keywords is 1/8 of the CJK advance measure,
+	i.e ''0.125ic''.
 
 	Note: Spacing conventions vary, but values typically range from 1/4ic to as low as 1/8ic,
 	with 1/4ic being particularly common in monospace contexts
 	and 1/6ic being more common in proportional typesetting.
+	However because these spaces are inserted by default
```
frivoal, line 8166: If you adopt the suggestion made in https://github.com/w3c/csswg-drafts/pull/8262/commits/aa252b9100d76945ae040e6c970780cc80d1f251#r1059249648, you could/should drop the "However" from this sentence.

frivoal (approved):
I made a few comments which I think would improve on this PR. I don't think they're blocking, so you can merge as is if you want to as far as I am concerned (and either deal with these later, or disagree), but I would suggest incorporating them now.

css-text-4/Overview.bs
fantasai: tt is obsolete, but it's at least standard (part of HTML4). I added kbd and samp, but hesitant to add some of the others. Once things stabilize a bit, though, we should definitely add a PR for the HTML spec.

css-text-4/Overview.bs
fantasai: Adjusted the note to reference metal type

PR #8294, opened 2023-01-09
[css-view-transitions-1] Fix play state tracking for animations

khushalsagar:
Animations targeting pseudo-elements are considered active, for the purpose of finishing the transition, if they are in running or paused state. See resolution here for details: https://github.com/w3c/csswg-drafts/issues/7785#issuecomment-1269115123.

"paused" was incorrectly changed to idle in the following commit: https://github.com/w3c/csswg-drafts/commit/5151ae80775a862c0edb5ebb35653bc47a55f906.

This change fixes that.


jakearchibald:
Hah, not sure what I was thinking when I made that change. Thanks for fixing it.

PR #8306, opened 2023-01-12
[css-view-transitions-1] Skip transition if viewport changes. Fixes #8045

### LINKED ISSUE #8045: [css-view-transitions-1] Handling snapshot viewport sizes changes during transitions
The [snapshot viewport](https://github.com/w3c/csswg-drafts/pull/8037) size would change by:

- The user or [script](https://developer.mozilla.org/en-US/docs/Web/API/Window/resizeTo) resizing the window.
- A mobile device changing orientation.

This impacts a transition if it happens:

- Between capturing the old and new states.
- During the transition animation.

Options:

## Skip the transition

A resize creates situations the developer may not have predicted and catered for. For example, they may have assumed that their site header would be the same size in the old and new states. A resize can break that assumption.

Given that the transition is an enhancement, and the user is performing an 'interrupting' action by resizing the window, skipping the transition seems reasonable.

## Update the animation state and continue

All 'new' elements will be updated with their new size, which in turn will update all animations in progress. This includes the transition container representing the root element - this will be the only case where this element animates width & height.

jakearchibald:


css-view-transitions-1/Overview.bs
```diff
@@ -806,6 +814,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 						Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
 
+					1. If |transition|'s [=ViewTransition/initial snapshot viewport size=] is not equal to the [=snapshot viewport size=],
```
khushalsagar, line 817: Not sure if we need to specify exact when we do this check, or we can just add it to the step that's setting it:

```
1. Set |transition|'s [=ViewTransition/initial snapshot viewport size=] to the [=snapshot viewport size=]. If this size changes after this point, skip the transition.
```

I'm fine with this too.

khushalsagar (approved):

css-view-transitions-1/Overview.bs
jakearchibald: I thought about this, but doing it properly would be more complicated.

Changes to the snapshot viewport happen outside the page, so to do things on the page, it would need to queue a task. Because a task is queued, we can't guarantee the same view transition is active compared to when the task was queued, so that will need to be verified before skipping the transition.

PR #8317, opened 2023-01-16
[css-view-transitions-1] Fix refs

jakearchibald:
Bikeshed changed databases, and we ended up with a couple of new warnings/errors.

There's still one remaining new error:

```
LINE ~371: Spec-section autolink tried to link to non-existent 'css-viewport' spec:
[[css-viewport#interactive-widget-section]]
```

I don't think this is our fault, I think it's something missing from the database (cc @tabatkins)


khushalsagar (approved):

jakearchibald:
Merging this despite the error, because it improves on the current state.

Issue #8070, opened 2022-11-14
[css-view-transitions-1] Positioning of ::view-transition (the root pseudo)


jakearchibald:
In https://github.com/w3c/csswg-drafts/issues/7859 we resolved on the concept of a "snapshot viewport", which would be consistent regardless of things like URL bar, virtual keyboard, and scrollbars, meaning animations would be stable, even if those things only existed on one side of the transition.

<img alt="" src="https://raw.githubusercontent.com/w3c/csswg-drafts/2f8e60e1e2469852f9610c1a37158d65cef718c7/css-view-transitions-1/desktop-browser-snapshot-viewport.svg" width="500">

<img alt="" src="https://raw.githubusercontent.com/w3c/csswg-drafts/2f8e60e1e2469852f9610c1a37158d65cef718c7/css-view-transitions-1/phone-browser-snapshot-viewport.svg" width="200">

That means that `::view-transition` would fill this snapshot viewport throughout the transition.

We see two options to achieve this:

## Option 1: Position `::view-transition` per frame

`::view-transition` would use regular `position: fixed` positioning. `inset: 0` would only fill the layout viewport, which is often smaller than the snaphshot viewport. So, to fix this, negative `inset` values would be used to make the element fill the snapshot viewport. These values would be updated per frame, to ensure the positioning is correct during any state changes (eg url bar hiding/showing).

This would be developer-visible via `getComputedStyle`, and would behave differently if the position was set to a non-fixed value.

In browsers that use a compositor to manage `position: fixed` items, there's a worry that there may be rendering instability in frames where the compositor is 'ahead' of the main thread's updates to the `inset`.

## Option 2: Layout `::view-transition` within the "snapshot viewport"

The `::view-transition` would have `position: fixed; inset: 0`. However, we would use spec prose to say this pseudo is laid out in the snapshot viewport, so `position: fixed; inset: 0` would cause `::view-transition` to fill the snapshot viewport, rather than the layout viewport.

This wouldn't have the compositor issues. However, it's perhaps a bit weird for this one pseudo to appear to treat layout differently. Is there anything else that does something similar?

---

We're leaning towards "option 2".

bokand:
Assuming we go with option 2, a related issue is what happens if author style does something else, e.g. `position: absolute`. Should that work and what does it mean? Is it possible to say author style cannot affect the position of this element? Or else, can we spec that this lives in its own formatting context with absolute & fixed containing blocks being the "snapshot viewport"?

[...]

Issue #7859, opened 2022-10-07
[css-shared-element-transitions-1] Define transitions over changing viewport


bokand:
Due to various browser UI, a transition can take place where the source and destination pages have a different size initial-containing-block and fixed/layout viewport. Some examples of UI widgets that cause this:

- Mobile URL bar (and related hide-on-scroll browser bars)
- Virtual Keyboard
- Desktop browser scrollbars

For example, most mobile browsers will force a hidden URL bar to be shown when a cross-document navigation occurs. In the other direction, a showing virtual-keyboard will be hidden when a navigation occurs. We should define exactly how transitioning elements are captured, sized, and positioned in these cases to ensure a seamless transition between the two states (i.e. elements should not "jump"). Here's what we're currently implementing which we can use as a starting point for spec text, thoughts welcome.

### Snapshot Viewport

We define the concept of the "snapshot viewport". This is a rect that is a consistent size and position regardless of what UI is currently shown. For example, on mobile browsers, the snapshot viewport overlays any shown UI like the URL bar and virtual keyboard:

![The snapshot viewport overlaying UI in a mobile browser](https://cdn.glitch.global/e61d77ef-db24-481a-87dc-f282612eb078/SET%20-%20Snapshot%20Viewport%20(1).png?v=1665177867435)

On desktop browsers, a common UI widget requiring attention is scrollbars (in particular, left-side vertical scrollbars in RTL writing modes since it shifts the content origin). The snapshot viewport is overlays and fills the full content area:

![The snapshot viewport underlaying scrollbars in a desktop browser](https://cdn.glitch.global/e61d77ef-db24-481a-87dc-f282612eb078/SET%20-%20Snapshot%20Viewport%20Scrollbar%20(1).png?v=1665178343787)

### Positioning

The snapshot viewport serves as the common coordinate space in which transitions take place. Since each shared element's transition is [positioned by its `::page-transition-container`](https://drafts.csswg.org/css-view-transitions-1/#selectordef-page-transition-container-pt-tag-selector), and this element is positioned absolutely within the `::page-transition`, we make `::page-transition` match the snapshot viewport rect:

```CSS
html::page-transition {
  position: fixed;

  /* Dynamically generated based on whether UI is showing or not. */
  left: -15px; /* left-side vertical scrollbar */
  top: -56px; /* mobile URL bar */
  width: 412px;
  height: 1000px;
}
```

When capturing a shared element's "screen space position", use the position relative to the snapshot viewport:

[...]

PR #8324, opened 2023-01-18
[css-view-transitions-1] Replace 'snapshot viewport' with 'snapshot root'

bokand:
[css-view-transitions-1] Replace 'snapshot viewport' with 'snapshot root'

"Viewport" is already a heavily overloaded term and this rect isn't really a "viewport" in most senses so rename it to "snapshot root" to avoid confusion.

Also some minor improvements around its origin and how transforms are computed.

See also: #8070, #7859


css-view-transitions-1/Overview.bs
```diff
@@ -363,39 +363,39 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 		<figcaption>
-			An example of the [=snapshot viewport=] on a desktop OS.
+			An example of the [=snapshot root=] on a desktop OS.
 			This includes the scrollbars, but does not include the URL bar, as web content never appears in that area.
 		</figcaption>
 	</figure>
 
 	This means the snapshot canvas size is likely to be consistent for the [=document element=]'s [=captured element/old image=] and [=captured element/new element=].
 
-	The <dfn>snapshot viewport origin</dfn> refers to the start of the block and inline axes of the [=snapshot viewport=].
```
bokand, line 396: This is used only in reference to transforms which AFAIK are always based on x,y running positive starting from the top-left.

css-view-transitions-1/Overview.bs
```diff
@@ -1210,7 +1210,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 				1. Set |height| to the current height of |capturedElement|'s [=new element=]'s [=border box=].
 
-				1. Set |transform| to a transform that maps the |capturedElement|'s [=new element=]'s [=border box=] from document origin to its quad from the [=snapshot viewport origin=].
```
bokand, line 1213: I'm a bit confused by what was meant here so maybe I'm just misunderstanding but "its quad from the snapshot viewport origin" is what we're trying to compute here so it seems a bit circular.

I also think the "border box from document origin" is wrong if we were to load the page scrolled (since the snapshot viewport is effectively fixed but the document origin is pushed up by the scroll)...instead we want to go from the snapshot rect origin to "element's border box relative to the snapshot rect origin".

bokand:
@vmpstr @khushalsagar as FYI

css-view-transitions-1/Overview.bs
```diff
@@ -1210,7 +1210,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+				1. Set |transform| to a transform that maps the [=snapshot root origin=] to the top-left corner of the |capturedElement|'s [=new element=]'s [=border box=].
```
jakearchibald, line 1213: I'm trying to think of clearer ways to word this. What about:

```
1. Set |transform| to a transform that would place |capturedElement|'s [=new element=]'s [=border box=] in its current visual position, from the [=snapshot root origin=].
```

Although I'm happy to go with what you have here.

jakearchibald (approved):
LGTM. One comment about a line I struggled to get my head around, but that might just be me.

bokand:
Thanks for the review!

PR #8326, opened 2023-01-18
[css-view-transitions-1] Add manual spec link for CSS-Viewport

bokand:
[css-view-transitions-1] Add manual spec link for CSS-Viewport

It appears CSS-Viewport isn't being exported to SpecRef due to not having a published TR. Add a manual link to css-view-transitions for now to fix linking errors.


bokand:
CSS-Viewport appears to link now without this change so lets abandon it.

tabatkins:
Yup, got that fixed over in Reffy for an unrelated reason. ^_^

PR #8330, opened 2023-01-18
[css-view-transitions-1] Rename domUpdated to updateCallbackDone

### LINKED ISSUE #8144: [css-view-transitions-1] domUpdated isn't the right term
Apart from a couple of events that are prefixed with `DOM*`, we don't really expose DOM as a term to web content and I think we should keep it that way.

Either keep this somewhat more abstract or go with `nodeTreeChanged` or some such.

cc @jakearchibald @domenic

vmpstr:
[css-view-transitions-1] Rename domUpdated to updateCallbackDone

Fixes #8144


jakearchibald:
Feels like we should also rename the callback as part of this.

vmpstr:
Done. PTAL

jakearchibald (approved):

PR #8331, opened 2023-01-18
[css-grid-1][css-grid-2][editorial] Fix 'auto' sizing function link in layout-algorithm

citriqa:
Without the 'grid-template-rows' prefix this link points at 'auto' in the context of the grid item placement properties, but like its neighbors it is actually meant to refer to the track sizing function.


fantasai:
Merged, thanks very much for catching that!

PR #8337, opened 2023-01-20
[css-view-transitions-1] Expanded intro section

jakearchibald:
Expanded the intro section. See inline notes:


css-view-transitions-1/Overview.bs
```diff
@@ -61,82 +86,346 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+	These captures are used in a tree of pseudo-elements
+	(detailed in [[#view-transition-pseudos]]),
+	where the old state cross-fades with the new state,
+	while animating from the old to new size and position.
+
+	The developer chooses which elements are captured independently,
+	meaning they can be animated independently,
+	using the 'view-transition-name' CSS property.
+
+## Lifecycle ## {#lifecycle}
```
jakearchibald, line 115: Added a high-level lifecycle. Feels like it gives a good overview before diving into the spec proper?

css-view-transitions-1/Overview.bs
```diff
@@ -61,82 +86,346 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+
+	1. <code>viewTransition.{{ViewTransition/ready}}</code> fulfills.
+
+	1. Pseudo-elements animate until finished.
+
+	1. Transition pseudo-elements removed.
+
+	1. <code>viewTransition.{{ViewTransition/finished}}</code> fulfills.
+
+	<div id="phases-diagram">
```
jakearchibald, line 146: A step by step diagram! I was originally going to make this a video, but being able to step back & forward seems useful. It also means we can change it without having to create a whole new video.

css-view-transitions-1/Overview.bs
```diff
@@ -61,82 +86,346 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+			});
+			previous.addEventListener('click', async () => {
+				const slide = iframe.contentDocument.querySelector('spec-slide');
+				await slide.previous();
+				updateButtons(slide);
+			});
+		</script>
+	</div>
+
+## Transitions as an enhancement ## {#transitions-as-enhancements}
```
jakearchibald, line 178: We already had a section titled this, but I felt it went a bit beyond describing how transitions are just an enhancement. It had lots of good stuff around the structure of the pseudos, so I moved that into the section about the pseudos.

css-view-transitions-1/Overview.bs
```diff
@@ -61,82 +86,346 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+	If the DOM change should also be skipped, then that should be handled by another feature.
+	<code>{{NavigateEvent|navigateEvent}}.{{NavigateEvent/signal}}</code> is an example of a feature developers could use to handle this.
+
+	Although the transition API allows DOM changes to be asynchronous via the {{UpdateCallback}},
+	the transition API is not responsible for queuing or otherwise scheduling the DOM changes beyond the scheduling needed for the transition itself.
+	Some asynchronous DOM changes can happen concurrently (e.g if they're happening within independent components),
+	whereas others need to queue, or abort an earlier change.
+	This is best left to a feature or framework that has a more holistic view of the application.
+
+## Examples ## {#examples}
```
jakearchibald, line 201: Added a sequence of examples similar to https://developer.chrome.com/docs/web-platform/view-transitions/, with videos.

css-view-transitions-1/Overview.bs
```diff
@@ -61,82 +86,346 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
-	which itself is attached to the [=document element=] of the page.
+		<figure>
+			<video src="diagrams/videos/bad-sidebar.mp4" style="aspect-ratio: 1404/738" class="main-example-video" controls muted loop playsinline></video>
+		</figure>
 
-	Each of the groups is comprised of up to four pseudo elements:
+		In this case, things would look better if the sidebar was static if it was in both the "old" and "new" states.
+		Otherwise, it should animate in or out.
 
-	: Group
```
jakearchibald, line 99: This content isn't gone, I moved it into the pseudo-element section.

css-view-transitions-1/Overview.bs
```diff
@@ -166,15 +455,6 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 	</dl>
 
-	The [=document element=] participates in a view transition by default using the following rule,
-	added to the [=HTML user agent style sheet=]:
-
-	```css
-	html {
-		view-transition-name: root;
-	}
-	```
```
jakearchibald, line 176: I felt this was kinda hidden away, so I moved it into the UA styles section (which used to be the UA keyframes section)

css-view-transitions-1/Overview.bs
```diff
@@ -183,9 +463,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 	This may be confusing, since the elements themselves are not necessarily referring to the same object,
 	but it is a useful model to consider them to be visual states of the same conceptual page entity, that we happen to call "element".
 
-# User-agent keyframes # {#ua-keyframes}
+# User-agent styles # {#ua-styles}
```
jakearchibald, line 466: Although this section doesn't contain all of the UA styles (feels like a lot of them are better off in the pseudo-element section), it gives a heading for readers to see, and the note in this section directs them to other UA styles.

css-view-transitions-1/Overview.bs
```diff
@@ -313,17 +646,31 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 		that is also a [=named view-transition pseudo-element=],
 		and [=tree/participates=] in a [=pseudo-element tree=].
 
 		It is selected from its [=ultimate originating element=], the [=document element=].
 
-		It is a [=replaced element=] displaying the old element's snapshot image.
-		It has [=natural dimensions=] equal to the snapshot's size.
+		It is a [=replaced element=], with [=natural dimensions=] equal to the content's size.
+
+		Note: The image content is set and updated in [=setup transition pseudo-elements=] and [=update pseudo-element styles=].
```
jakearchibald, line 673: I made the language here a bit more generic, so it works for old & new.

css-view-transitions-1/Overview.bs
```diff
@@ -334,6 +681,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 			inline-size: 100%;
 			block-size: auto;
 
+			animation-name: -ua-view-transition-fade-out;
```
jakearchibald, line 684: Heh, we forgot this, or it got lost along the way.

css-view-transitions-1/Overview.bs
```diff
@@ -417,8 +775,6 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 	since that results in a circular dependency.
 	Instead, this stacking context is a sibling of other page contents.
 
-	Issue: Create a diagram to explain these stacking contexts.
```
jakearchibald, line 420: Feels like the snapshot root diagrams take care of this?

css-view-transitions-1/Overview.bs
```diff
@@ -534,60 +890,6 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		1. Return |transition|.
 	</div>
 
-	<div class=example>
```
jakearchibald, line 537: Removed this in favour of the flow of examples in the intro

css-view-transitions-1/Overview.bs
```diff
@@ -344,14 +692,24 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+			position: absolute;
+			inset-block-start: 0;
+			inline-size: 100%;
+			block-size: auto;
+
+			animation-name: -ua-view-transition-fade-in;
+			animation-duration: inherit;
+			animation-fill-mode: inherit;
+		}
+		```
```
jakearchibald, line 710: Felt like it was worth specifying this in full, especially due to `-ua-view-transition-fade-in`.

khushalsagar (commented):
I just looked at the changes to Overview.bc. Trust you got the rest right. :)

css-view-transitions-1/Overview.bs
```diff
@@ -61,82 +86,346 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+	as the structure of the DOM is compromised for a purely-visual effect.
+
+	View transitions avoid this troublesome in-between state by allowing the DOM to switch between states synchronously,
+	then performing a customizable visual transition between the two states in another layer,
+	using a static capture of the old state, and a live capture of the new state.
+
+	These captures are used in a tree of pseudo-elements
+	(detailed in [[#view-transition-pseudos]]),
+	where the old state cross-fades with the new state,
+	while animating from the old to new size and position.
```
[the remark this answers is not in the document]
jakearchibald: Done!

css-view-transitions-1/Overview.bs
jakearchibald: Hah, I still hate the line break thing, but it's better than 80 char breaks.

css-view-transitions-1/Overview.bs
```diff
@@ -61,82 +86,346 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+		```css
+		::view-transition-image-pair(root) {
+			isolation: auto;
+		}
+
+		::view-transition-old(root),
+		::view-transition-new(root) {
+			animation: none;
+			mix-blend-mode: normal;
+			display: block;
```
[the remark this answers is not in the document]
jakearchibald: Ohh, good catch. Adding `display:block` to images is a bit of a habit of mine, but it isn't needed here since they're abs-pos. I'll remove this line.

css-view-transitions-1/Overview.bs
```diff
@@ -195,8 +475,22 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 		from { opacity: 0; }
 	}
+
+	html {
+		view-transition-name: root;
+	}
 	```
 
+	<div class=note>
+		<<@keyframes>> with a name beginning `-ua-` are scoped to the [=HTML user agent style sheet=].
```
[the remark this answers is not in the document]
jakearchibald: I'll tweak this a bit, but this isn't a normative definition. Unfortunately the normative definition is limited to that github issue for now.

css-view-transitions-1/Overview.bs
```diff
@@ -283,19 +619,16 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		}
```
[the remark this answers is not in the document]
jakearchibald: Good catch, but as you say, this needs a CSSWG issue, so we'll deal with it after this PR.

css-view-transitions-1/Overview.bs
```diff
@@ -313,17 +646,31 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+
+		A [=tree-abiding pseudo-element=]
 		that is also a [=named view-transition pseudo-element=],
 		and [=tree/participates=] in a [=pseudo-element tree=].
 
 		It is selected from its [=ultimate originating element=], the [=document element=].
 
-		It is a [=replaced element=] displaying the old element's snapshot image.
-		It has [=natural dimensions=] equal to the snapshot's size.
+		It is a [=replaced element=], with [=natural dimensions=] equal to the content's size.
```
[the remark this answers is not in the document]
jakearchibald: Done!

jakearchibald:
@khushalsagar I think I've addressed the feedback

css-view-transitions-1/Overview.bs
```diff
@@ -195,8 +478,24 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+
+	html {
+		view-transition-name: root;
+	}
 	```
 
+	<div class=note>
+		<<@keyframes>> with a name beginning `-ua-` are scoped to the user agent style sheet,
+		meaning that when `-ua-view-transition-fade-out` is used in a regular style sheet as an animation name,
+		it does not refer to the above definition.
```
khushalsagar, line 490: The second resolution on the issue is opposite: "The -ua- names *are* usable in author/user stylesheets."

We specifically discussed whether the keyframes in UA stylesheets should be usable in author stylesheets and the answer was yes. This means there can be a collision and the last rule parsed wins. The assumption was that the "-ua-" prefix would avoid the collisions.

jakearchibald: ohhhhh, I hadn't realised that at all. Thank you for the correction.

khushalsagar:
LGTM, just the one comment to make sure we're on the same page about UA keyframes.

css-view-transitions-1/Overview.bs
jakearchibald: I removed the note entirely, since it doesn't need clarifying, since it behaves as normal.

Issue #7655, opened 2022-08-29
[css-align-3][css-tables-3] Define last baseline for tables.


bfgeek:
"The maximum distance between the top of the cell box and the baseline over all cells that have 'vertical-align: baseline' is used to set the baseline of the row. If a row doesn’t have any cell that has 'vertical-align: baseline', the baseline of that row is the bottom content edge of the lowest cell in the row.

The baseline of a table-root is the baseline of its first row, if any. Otherwise, it is the bottom content edge of the table-root."

This definition for first baseline is good! - We also need to define the last baseline behaviour.

Something like:
"The last baseline is defined as:
 - The maxiumum distance between the top of the cell box and the last baseline over all the cells that have 'vertical-align: baseline' is used to set the baseline of the row.
 -  If the row doesn't have any cell that has 'vertical-align: basleine', the baseline of that row is the bottom content edge of the lowest cell in the row.

The last baseline of a table-root is the baseline of its last row, if any."

Ian

tabatkins:
This should use the same logic as we decide in #7641, wrt using cells with last-baseline alignment, and maybe falling back to first-baseline (if we decide to do that for flexbox) before falling back to a synthesized baseline.

Also it should be checking for `align-content: last baseline`, not `vertical-align: baseline`; see <https://www.w3.org/TR/css-align-3/#distribution-block> which defines that `align-content: normal` *defers to* `vertical-align` but other values override that behavior.

bfgeek:
testcase showing lowest content edge baseline: https://www.software.hixie.ch/utilities/js/live-dom-viewer/?saved=10716

fantasai:
@tabatkins and I have drawn up a PR for this at https://github.com/w3c/csswg-drafts/pull/8341

Issue #7641, opened 2022-08-23
[css-align-3][css-flexbox-1] How to determine the last baseline of a flex container with different alignment groups.


bfgeek:
See: https://www.software.hixie.ch/utilities/js/live-dom-viewer/?saved=10608
```html
<!DOCTYPE html>
<div style="display: inline-flex; align-items: last baseline;">
  <div style="background: lime; font-size: 30px;">line1<br>line2</div>
  <div style="display: flex;">
    <div style="font-size: 30px; align-self: last baseline; background: cyan;">line1<br>line2<br>line3</div> 
    <div style="font-size: 80px; align-self: baseline; background: hotpink;">line1<br>line2</div> 
  </div>
</div>
```
https://drafts.csswg.org/css-flexbox-1/#flex-baselines

Should the last baseline of the "lime" box be aligned with the last-baseline of the "cyan" box, or the "hotpink" box?

The rules just look at if something "participates in baseline alignment" so by that rule it should be the "hotpink" box.
But likely should be the "cyan" box.

tabatkins:
Yeah, I *think* this is just a bit of behavior that slipped thru the edits when we upgraded from "baseline" to "first/last baseline". I suspect we should prefer taking from the last-baseline element when determining the inner flexbox's last baseline, and mutatis mutandis for the first.

This does mean that the last baseline can potentially be higher than the first baseline, tho (say, both items are single-line, and the last-baseline one has a smaller font-size). Is that ok?

bfgeek:
> Is that ok?
I think its ok - but reserve the right to complain later if it isn't. :)

tabatkins:
Okay, so suggested edits are to modify Step 1 of <https://drafts.csswg.org/css-flexbox-1/#flex-baselines> to first look at items with the same baseline alignment preference as the requested baseline, falling back to *any* baseline-aligned item if there aren't any matching.

bfgeek:
I think I have a slight preference for look at items which match the baseline alignment preference - then fallback to any item.

Otherwise we have three(!) levels of fallback - e.g.
1. Baseline alignment preference.
2. Any baseline alignment.
3. Any.
vs.
1. Baseline alignment preference.
2. Any.

The two stage fallback would be web compatible with what we have today.

fantasai:
[...]

PR #8341, opened 2023-01-20
[css-align-3][css-flexbox-1][css-grid-1] Improve container baseline finding algo #7655 #7641

fantasai:
This pull request covers the issues #7655 #7641 including the following resolutions (see [minutes](https://lists.w3.org/Archives/Public/www-style/2022Oct/0013.html)):

* RESOLVED: When taking the baseline of a row of items, we check a shared first baseline, then a shared last baseline, then the first baseline of the first item. (And vice versa for last baselines.)

* RESOLVED: Rowspanning cells participate only in first baseline alignment of their first row, and last baseline alignment in the last row.

* RESOLVED: Table baselines match grid, per details above.

**OPEN ISSUES**
* There was some confusion over whether we resolved to take the baseline from the first item or the highest item, see [comments](https://github.com/w3c/csswg-drafts/issues/7641#issuecomment-1288415323). Firefox and WebKit use the first, Chrome NG uses the highest.
* If we are indeed using the highest, what happens if there's a tie? (Unclear if there are cases where this matters, but maybe someday there will be?) We wrote in to use order-modified document order for Flex and grid placement order for Grid.

### DIFF
--- a/css-align-3/Overview.bs
+++ b/css-align-3/Overview.bs
@@ -2218,7 +2218,8 @@ Determining the Baselines of a Box</h2>
 		<dt>tables
 		<dd>
 			The first/last [=baseline set=] of a table box
-			is the first/last [=baseline set=] of its first/last row.
+			is the first/last [=baseline set=] of its first/last row
+			(after any reordering of [=table row groups=]).
 
 			When finding the first/last baseline set of an inline-block,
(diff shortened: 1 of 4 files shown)


bfgeek:
This seems good.

bfgeek (approved):

PR #8367, opened 2023-01-27
[css-images-4] Fix basic syntax of gradients

### LINKED ISSUE #7986: [css-images-4] Relax color interpolation method syntax in gradients
Please forgive me if some related discussions exist somewhere but I did not find them in #6667 or #5833. I would like to know why the color interpolation method cannot be declared between the gradient line parameters (I am not sure this is the correct term).

For example:

  - current: `conic-gradient([[from <angle>]? [at <position>]?] || <color-interpolation-method>, <angular-color-stop-list>)`
  - instead of `conic-gradient([from <angle> || at <position> || <color-interpolation-method>]?, <angular-color-stop-list>)`

From the perspective of the authoring experience, `conic-gradient(at center in srgb from 45deg, ...)` does not seem even more meaningless (no criticism here) than `conic-gradient(at center in srgb, ...)` for me, and the syntax is easier to read.

cdoublev:
Fixes #7986.

The current syntax makes the color stop list optional, which is obviously unexpected.

This PR also aligns CSS Images 4 with CSS Images 3 regarding `<size>` renamed to `<rg-size>` (resolving a conflict with `<ray()>`) and `<ending-shape>` renamed to `<rg-ending-shape>` (for consistency, I guess).

I am not sure if `a? b? || c` is different than `[a? b?]! || c` so I prefered the first option for simplicity.

### DIFF
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -18,7 +18,7 @@ Abstract: This module contains the features of CSS level 4 relating to the <<ima
 
 Issue Tracking: Tracker http://www.w3.org/Style/CSS/Tracker/products/27
 Previous Version: https://www.w3.org/TR/2017/WD-css-images-4-20170413/
-Ignored Terms: <offset>, background positioning area, border image area, <meetorslice>, <ending-shape>, Map, <image>, invalid image, invalid images, concrete object size, linear-gradient(), radial-gradient(), default object size, CSS
+Ignored Terms: <offset>, background positioning area, border image area, <meetorslice>, Map, <image>, invalid image, invalid images, concrete object size, linear-gradient(), radial-gradient(), default object size, CSS
 Ignored Vars: H, P
 Include Can I Use Panels: yes
 Default Highlight: css
@@ -1198,7 +1198,7 @@ Linear Gradients: the ''linear-gradient()'' notation {#linear-gradients}
 
 	<pre class=prod>
 		<dfn>linear-gradient()</dfn> = linear-gradient(
-			[ <<angle>> | to <<side-or-corner>> ]? || <<color-interpolation-method>>,
+			[ [ <<angle>> | to <<side-or-corner>> ] || <<color-interpolation-method>> ]? ,
 			<<color-stop-list>>
 		)
 		<dfn>&lt;side-or-corner></dfn> = [left | right] || [top | bottom]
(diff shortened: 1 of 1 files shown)


tabatkins:
> I am not sure if a? b? || c is different than [a? b?]! || c so I prefered the first option for simplicity.

They're not equivalent (the first is allowed to produce nothing, the second isn't), but for the purpose of this spec what you did was correct.

cdoublev:
Right, and actually I would have preferred the other syntax for two reasons.

First, the trailing `?` in `[[[<rg-ending-shape> || <rg-size>]? [at <position>]?] || <color-interpolation-method>]?` is useless, which does not seem right to me.

Second, the result is not consistent with the result from parsing other gradients.

`conic-gradient(red, cyan)`, `linear-gradient(red, cyan)`, `radial-gradient(red, cyan)`, produce the following pseudo-representations:

```
ConicGradient(
  Omitted('[[[from <angle>]? [at <position>]?] || <color-interpolation-method>]')`,
  ColorStopList(Keyword('red'), Keyword('cyan')),
)
LinearGradient(
  Omitted('[[<angle> | to <side-or-corner>] || <color-interpolation-method>]')`,
  ColorStopList(Keyword('red'), Keyword('cyan')),
)
RadialGradient(
  List(
    List(Omitted('<rg-ending-shape> || <rg-size>'), Omitted('at <position>')), 
    Omitted('<color-interpolation-method>')
  ),
  ColorStopList(Keyword('red'), Keyword('cyan')),
)
```

---

I let you see if the following asides/ideas are worth creating a new issue:

**Aside 1:**

`<rg-ending-shape>` could be replaced by `<basic-shape>` to allow new shapes of radial gradients.

`circle` or `ellipse` (possibly qualified with `<extent-keyword>`) would be aliases of `circle(100% at center center)` and `ellipse(100% 100% at center center)`.

**Aside 2:**

`<rg-size>` is `<rg-extent-kewyord>` + `<length-percentage>` for `radial-gradient()`, and is `<rg-extent-kewyord>` + `sides` for [`<ray()>`](https://drafts.fxtf.org/motion-1/#funcdef-offset-path-ray).

I think the related productions would be better defined without `<size>`, whose semantic seems too broad to me:

```
<radial-gradient()> = radial-gradient( [ ... || [ <extent-keyword> | <length> | <length-percentage> ] ] ... )
<ray()> = ray( [ ... && [ <extent-keyword> | sides ] && ... ] )

<extent-keyword> = closest-corner | closest-side | farthest-corner | farthest-side
```

PR #8390, opened 2023-02-01
[scroll-animations-1] Add self keyword

### LINKED ISSUE #8227: [scroll-animations-1] Allow Anonymous Scroll Progress Timelines to target self
When creating an Anonymous Scroll Progress Timelines usning `scroll()` you can target either the `root` or the `nearest` scroller. 

Currently `nearest` is defined as:

> Specifies to use the nearest ancestor scroll container. _(Default.)_

This does not allow authors to create an Anonymous Scroll Progress Timeline that uses the element itself, as `nearest` only looks to ancestors. E.g. the code below does not work when `.container` is the scroller

```css
.container {
  animation: 1s bg-scale both linear;
  animation-timeline: scroll(nearest inline);
}
```

As a result, authors must use a named Scroll Progress Timeline instead:

```css
.container {
  animation: 1s bg-scale both linear;
  scroll-timeline: --self inline;
  animation-timeline: --self;
}
```

This seems like a bit a hassle. Can we adjust `nearest` so that it starts looking from the targeted element itself first, and only then to ancestors? Or maybe introduce a new keyword `self` should altering `nearest` introduce some nasty side-effects?

bramus:
Fixes #8227.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -144,7 +144,7 @@ spec: cssom-view-1; type: dfn;
 	<pre class="prod">
 		<<scroll()>> = scroll( [ <<scroller>> || <<axis>> ]? )
 		<dfn noexport><<axis>></dfn> = block | inline | vertical | horizontal
-		<dfn noexport><<scroller>></dfn> = root | nearest
+		<dfn noexport><<scroller>></dfn> = root | nearest | self
 	</pre>
 
 	By default,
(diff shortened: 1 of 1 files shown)


plinss:
@bramus In the future, please for the repo and create branches and PRs from your private fork rather than in the main repo. Thanks.

bramus:
@plinss Oops, my bad. Unintentionally pushed to wrong origin. Thanks for notifying me.

bramus:
Closing this PR, removing the erroneously created branch, and filing a new PR from my personal fork instead.

PR #8395, opened 2023-02-03
[css-anchor-1] Fix syntax of a few anchor-* properties

xiaochengh:
- Fix the missing keyword `implicit` in `anchor-scroll`'s syntax
- Change the syntax of `anchor-default` from `<<anchor-element>>` to `none | <<dashed-ident>>`, since the `implicit` keyword doesn't make sense here


xiaochengh:
@tabatkins

tabatkins:
Talked with @xiaochengh offline; the first change is good, the second wasn't. I'll close this and just handle the first change myself.

PR #8418, opened 2023-02-08
[css-multicol-1] Define `column-count` with animation type

cdoublev:
Replaces `Animatable` by `Animation Type`, as defined in [Web Animations 1](https://drafts.csswg.org/web-animations-1/#animating-properties):

`w3c/reffy` (spec crawler) does not normalize `Animatable` to `Animation Type` therefore users have to check both `propDef.animationType` and `propDef.animatable`, which is not great.


rachelandrew (approved):
Thank you!

PR #8419, opened 2023-02-08
[css-speech-1] Define all properties as not animatable

cdoublev:


cdoublev:
With some hindsight, I am not sure all properties in CSS Speech should be considered as `not animatable` or `discrete`. Therefore I close this issue because I am only interested in defining `not animatable` properties.

tabatkins:
We only rarely define things as "not animatable", when they'd actually cause problems (or at least excessive complexity) to animate. I don't think anything in Speech qualifies.

PR #8422, opened 2023-02-08
[css-nav-1] Define properties with animation type

cdoublev:

### DIFF
--- a/css-nav-1/Overview.bs
+++ b/css-nav-1/Overview.bs
@@ -1345,6 +1345,7 @@ Name: spatial-navigation-contain
 Value: auto | contain
 Initial: auto
 Inherited: no
+Animation Type: discrete
 </pre>
 
 <dl dfn-for=spatial-navigation-contain dfn-type=value>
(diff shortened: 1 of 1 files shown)


tabatkins:
I presume these should be discretely animatable.

cdoublev:
Would they accept `mix()`?

> A [mix()](https://drafts.csswg.org/css-values-4/#funcdef-mix) notation is invalid if either of its [`<declaration-value>`](https://drafts.csswg.org/css-syntax-3/#typedef-declaration-value)s are invalid when interpreted as a whole value for the property the function is used in, or if the property using it is [not animatable](https://drafts.csswg.org/web-animations-1/#not-animatable).

tabatkins:
I'm not sure I understand the purpose of the question. `mix()` just relies on the animation concept; if you can animate, you can mix().

cdoublev:
Forget it, I was trying to give an argument for *not animatable* but `mix()`, based on wrong thinking, indeed.

I am not sure exactly when a property should be *not animatable*, looking at the current list of *not animatable* properties.

Do you want me to set *Animation type* to `discrete` or close this PR and discuss the matter in a dedicated issue?

tabatkins:
Properties are only "not animatable" if they actually *cause problems* to animate. The animation-* properties are the obvious example, due to circularity, but a handful of other properties have been special-cased due to impl reasons. But absent any special reason, *every* property is animatable, just discretely if there's nothing else reasonable to do; that's why the Animations spec has that default in the first place.

Go ahead and just change these to discrete.

PR #8427, opened 2023-02-08
[css-exclusions-1] Define properties with animation type

cdoublev:


cdoublev:
With some hindsight, I am not sure properties in CSS Exclusions should be considered as `not animatable` or `discete`. Therefore I close this issue because I am only interested in defining `not animatable` properties.

PR #8435, opened 2023-02-09
[css-contain-3] Explicitly define `container` as not animatable

cdoublev:
The `Animation type` of `animation`, `transition`, `scroll-timeline`, which are shorthands whose sub-properties are all `not animatable`, is explicitly defined as `not animatable` instead of `see individual properties`, like `container`.

In my opinion, `see individual properties` should be used only when some/all sub-properties are animatable.


tabatkins:
Actually `animation`/etc should be changed the other way. Shorthands *do not have* animation behavior, only longhands do. (Same as how they don't have inheritance behavior, or initial values.)

PR #8436, opened 2023-02-09
[scroll-animations-1] Define `view-timeline` and `animation-range-*` as not animatable

cdoublev:
This PR replaces the default `Animation Type: see individual properties` by an explicit `Animation Type: not animatable` for the `view-timeline` and `animation-range` shorthands.

The `Animation type` of `animation`, `transition`, `scroll-timeline`, which are shorthands whose sub-properties are all `not animatable`, is explicitly defined as `not animatable` instead of `see individual properties`, like `view-timeline` and `animation-range`.

In my opinion, `see individual properties` should be used only when some/all sub-properties are animatable.

This PR also replaces `Animatable` with `Animation type` for `animation-range-*`, as required by [Web Animations](https://drafts.csswg.org/web-animations-1/#animating-properties):

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -762,6 +762,7 @@ spec: cssom-view-1; type: dfn;
 	<pre class="propdef shorthand">
 		Name: animation-range
 		Value: [ <<'animation-range-start'>> <<'animation-range-end'>>? | <<timeline-range-name>> ]#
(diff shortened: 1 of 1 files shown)


tabatkins:
Closing for the same reason as <https://github.com/w3c/csswg-drafts/pull/8435#issuecomment-1426311628>

PR #8437, opened 2023-02-09
[scroll-animations-1] Fix `<keyframe-selector>`

cdoublev:
This PR adds a range restriction [0, 100] on `<percentage>` of `<keyframe-selector>`, [as defined in CSS Animations 1](https://drafts.csswg.org/css-animations-1/#typedef-keyframe-selector), with a minor simplification of its syntax.


cdoublev:
Closing in favor of #8438 in which I suggest `[from | to] <timeline-range-name>? | <timeline-range-name>? <percentage [0,100]>` to allow `from timeline-1 { ... } to timeline-2 {}`.

PR #8454, opened 2023-02-15
[css-view-transitions-1] Fix some promise timing, and avoid duplicate unhandled rejections

jakearchibald:
In the current spec, this will cause three unhandled rejections:

```js
const transition = document.startViewTransition(() => {
  doesNotExist();
});
```

As it will cause `transition.updateCallbackDone`, `transition.ready`, and `transition.finished` to reject with the same error.

This PR _always_ marks `transition.finished` as handled, as it only ever rejects for the same reason as  `transition.updateCallbackDone`. It marks `transition.ready` as handled if it's rejected with the same reason as `transition.updateCallbackDone`.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1451,6 +1451,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 			with the result of [=reacting=] to |callbackPromise|:
 
 			- If the promise was fulfilled, then return undefined.
+
+			Note: Since the rejection of |callbackPromise| isn't explicitly handled here,
+			if |callbackPromise| rejects, then |transition|'s [=ViewTransition/update callback done promise=] will reject with the same reason.
 	</div>
 
 ## [=Skip the view transition=] ## {#skip-the-view-transition-algorithm}
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1015,9 +1015,17 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		:: a {{Promise}}.
 			Initially [=a new promise=] in [=this's=] [=relevant Realm=].
 
+			Note: These promises are immediately created,
+			so rejections will cause {{unhandledrejection}}s unless they're [=mark as handled|handled=],
+			even if the getters such as {{updateCallbackDone}} are not accessed.
+
 		: <dfn>finished promise</dfn>
 		:: a {{Promise}}.
-			Initially [=a new promise=] in [=this's=] [=relevant Realm=].
+			Initially [=a new promise=] in [=this's=] [=relevant Realm=],
+				[=marked as handled=].
+
+			Note: This is [=marked as handled=] to prevent duplicate {{unhandledrejection}}s,
+			as this promise only ever rejects along with the [=update callback done promise=].
 
 		: <dfn>transition root pseudo-element</dfn>
 		:: a ''::view-transition''.
@@ -1134,6 +1142,11 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 					1. If |transition|'s [=ViewTransition/phase=] is "`done`", then return.
 
 						Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
+
+					1. [=Mark as handled=] |transition|'s [=ViewTransition/ready promise=].
+
+						Note: |transition|'s [=ViewTransition/update callback done promise=] will provide the {{unhandledrejection}}.
+						This step avoids a duplicate.
 
 					1. [=Skip the view transition=] |transition| with |reason|.
 
@@ -1424,19 +1437,20 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 		1. [=Assert=]: |transition|'s [=ViewTransition/phase=] is [=phases/before=] "`update-callback-called`".
 
-		1. Let |callbackPromise| be [=a new promise=] in |transition|'s [=relevant Realm=].
+		1. Let |callbackPromise| be null.
 
-			* If |transition|'s [=ViewTransition/update callback=] is null, then resolve |callbackPromise|.
+		1. If |transition|'s [=ViewTransition/update callback=] is null,
+			then set |callbackPromise| to [=a promise resolved with=] undefined,
+			in |transition|'s [=relevant Realm=].
 
-			* Otherwise, let |callbackPromise| be the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
+		1. Otherwise, set |callbackPromise| to the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
 
 		1. Set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
-		1. [=promise/React=] to |callbackPromise|:
+		1. [=Resolve=] |transition|'s [=ViewTransition/update callback done promise=]
+			with the result of [=reacting=] to |callbackPromise|:
 
-			* If |callbackPromise| was resolved, then [=resolve=] |transition|'s [=ViewTransition/update callback done promise=].
-
-			* If |callbackPromise| was rejected with reason |r|, then [=reject=] |transition|'s [=ViewTransition/update callback done promise=] with |r|.
+			- If |callbackPromise| was resolved, then return undefined.
 	</div>
 
 ## [=Skip the view transition=] ## {#skip-the-view-transition-algorithm}
@@ -1458,17 +1472,16 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 		1. Set |transition|'s [=ViewTransition/phase=] to "`done`".
 
-		1. If |transition|'s [=ViewTransition/ready promise=] has not yet been resolved, [=reject=] it with |reason|.
+		1. [=Reject=] |transition|'s [=ViewTransition/ready promise=] with |reason|.
 
-			Note: The ready promise would've been resolved if {{ViewTransition/skipTransition()}} is called after we start animating.
+			Note: The [=ViewTransition/ready promise=] may already be resolved at this point,
+			if {{ViewTransition/skipTransition()}} is called after we start animating.
+			In that case, this step is a no-op.
 
-		1. [=promise/React=] to |transition|'s [=ViewTransition/update callback done promise=]:
+		1. [=Resolve=] |transition|'s [=ViewTransition/finished promise=] with the result of [=reacting=] to |transition|'s [=ViewTransition/update callback done promise=]:
 
-			* If |transition|'s [=ViewTransition/update callback done promise=] was resolved,
-				then [=resolve=] |transition|'s [=ViewTransition/finished promise=].
-
-			* If |transition|'s [=ViewTransition/update callback done promise=] was rejected with reason |reason|,
-				then [=reject=] |transition|'s [=ViewTransition/finished promise=] with |reason|.
+			- If |transition|'s [=ViewTransition/update callback done promise=] was resolved,
+				then return undefined.
 	</div>
 
 ## [=Capture the image=] ## {#capture-the-image-algorithm}
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1015,7 +1015,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		:: a {{Promise}}.
 			Initially [=a new promise=] in [=this's=] [=relevant Realm=].
 
-			Note: These promises are immediately created,
+			Note: The [=ready promise=] and [=update callback done promise=] are immediately created,
 			so rejections will cause {{unhandledrejection}}s unless they're [=mark as handled|handled=],
 			even if the getters such as {{updateCallbackDone}} are not accessed.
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -897,7 +897,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 			{{UpdateCallback|updateCallback}} is called asynchronously, once the current state of the document is captured.
 			Then, when the promise returned by {{UpdateCallback|updateCallback}} fulfills, the new state of the document is captured.
 
-			{{UpdateCallback|updateCallback}} is _always_ called, even if the transition cannot happen (e.g. due to duplicate `view-transition-name` values).
+			{{UpdateCallback|updateCallback}} is *always* called, even if the transition cannot happen (e.g. due to duplicate `view-transition-name` values).
```
jakearchibald, line 900: Seems like bikeshed's markdown parser doesn't support emphasis with `_`. Just a formatting fix.

css-view-transitions-1/Overview.bs
```diff
@@ -1424,19 +1437,20 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
-		1. Let |callbackPromise| be [=a new promise=] in |transition|'s [=relevant Realm=].
+		1. Let |callbackPromise| be null.
 
-			* If |transition|'s [=ViewTransition/update callback=] is null, then resolve |callbackPromise|.
+		1. If |transition|'s [=ViewTransition/update callback=] is null,
+			then set |callbackPromise| to [=a promise resolved with=] undefined,
+			in |transition|'s [=relevant Realm=].
 
-			* Otherwise, let |callbackPromise| be the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
+		1. Otherwise, set |callbackPromise| to the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
```
jakearchibald, line 1446: No normative change here, just using more standard spec prose.

css-view-transitions-1/Overview.bs
```diff
@@ -1424,19 +1437,20 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 		1. Set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
-		1. [=promise/React=] to |callbackPromise|:
-
-			* If |callbackPromise| was resolved, then [=resolve=] |transition|'s [=ViewTransition/update callback done promise=].
+		1. [=Resolve=] |transition|'s [=ViewTransition/update callback done promise=]
+			with the result of [=reacting=] to |callbackPromise|:
 
-			* If |callbackPromise| was rejected with reason |r|, then [=reject=] |transition|'s [=ViewTransition/update callback done promise=] with |r|.
+			- If |callbackPromise| was resolved, then return undefined.
```
jakearchibald, line 1453: This is a normative change, but it doesn't change behaviour. It protects us from possible race conditions in future.

css-view-transitions-1/Overview.bs
```diff
@@ -1424,19 +1437,20 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 		1. Set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
-		1. [=promise/React=] to |callbackPromise|:
-
-			* If |callbackPromise| was resolved, then [=resolve=] |transition|'s [=ViewTransition/update callback done promise=].
+		1. [=Resolve=] |transition|'s [=ViewTransition/update callback done promise=]
+			with the result of [=reacting=] to |callbackPromise|:
 
-			* If |callbackPromise| was rejected with reason |r|, then [=reject=] |transition|'s [=ViewTransition/update callback done promise=] with |r|.
+			- If the promise was fulfilled, then return undefined.
```
jakearchibald, line 1453: This is technically a normative change, but it doesn't change any currently observable behaviour. It protects us from possible race conditions in future.

css-view-transitions-1/Overview.bs
```diff
@@ -1458,17 +1472,15 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+			if {{ViewTransition/skipTransition()}} is called after we start animating.
+			In that case, this step is a no-op.
 
-			* If |transition|'s [=ViewTransition/update callback done promise=] was resolved,
-				then [=resolve=] |transition|'s [=ViewTransition/finished promise=].
+		1. [=Resolve=] |transition|'s [=ViewTransition/finished promise=] with the result of [=reacting=] to |transition|'s [=ViewTransition/update callback done promise=]:
 
-			* If |transition|'s [=ViewTransition/update callback done promise=] was rejected with reason |reason|,
-				then [=reject=] |transition|'s [=ViewTransition/finished promise=] with |reason|.
+			- If the promise was fulfilled, then return undefined.
```
jakearchibald, line 1483: Again, technically a normative change, but no observable behaviour change.

css-view-transitions-1/Overview.bs
```diff
@@ -1135,9 +1143,14 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 
+					1. [=Mark as handled=] |transition|'s [=ViewTransition/ready promise=].
+
+						Note: |transition|'s [=ViewTransition/update callback done promise=] will provide the {{unhandledrejection}}.
+						This step avoids a duplicate.
+
 					1. [=Skip the view transition=] |transition| with |reason|.
 
-				* If the promise was resolved, then:
+				* If the promise was fulfilled, then:
```
jakearchibald, line 1153: https://jakearchibald.com/2014/resolve-not-opposite-of-reject/ 😄

css-view-transitions-1/Overview.bs
```diff
@@ -1015,9 +1015,17 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 		:: a {{Promise}}.
 			Initially [=a new promise=] in [=this's=] [=relevant Realm=].
 
+			Note: These promises are immediately created,
+			so rejections will cause {{unhandledrejection}}s unless they're [=mark as handled|handled=],
```
vmpstr, line 1019: Can you clarify this a bit: Which rejections will cause unhandledrejection? Is it that if this promise is rejected before it is handled, then it will cause an unhandledrejection? 

Also "these promises": that's the ready promise and the update callback done promise, but not the finished promise, right?

css-view-transitions-1/Overview.bs
```diff
@@ -1015,9 +1015,17 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+			even if the getters such as {{updateCallbackDone}} are not accessed.
+
 		: <dfn>finished promise</dfn>
 		:: a {{Promise}}.
-			Initially [=a new promise=] in [=this's=] [=relevant Realm=].
+			Initially [=a new promise=] in [=this's=] [=relevant Realm=],
+				[=marked as handled=].
+
+			Note: This is [=marked as handled=] to prevent duplicate {{unhandledrejection}}s,
+			as this promise only ever rejects along with the [=update callback done promise=].
```
vmpstr, line 1028: This would only be a duplicate if update callback done promise is not handled. If it is handled, but the finished promise is not handled, then should that still cause an unhandledrejection?

In other words, I feel like we shouldn't try too hard to prevent duplicate unhandledrejections, since these are different promises that would both be not handled if rejected

css-view-transitions-1/Overview.bs
```diff
@@ -1135,9 +1143,14 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
 
 						Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
 
+					1. [=Mark as handled=] |transition|'s [=ViewTransition/ready promise=].
```
vmpstr, line 1146: Likewise here, it seems that this is only a duplicate if update callback done promise is not explicitly handled? Or am I misunderstanding?

css-view-transitions-1/Overview.bs
```diff
@@ -1424,19 +1437,20 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
 
-			* Otherwise, let |callbackPromise| be the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
+		1. Otherwise, set |callbackPromise| to the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
 
 		1. Set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
-		1. [=promise/React=] to |callbackPromise|:
-
-			* If |callbackPromise| was resolved, then [=resolve=] |transition|'s [=ViewTransition/update callback done promise=].
+		1. [=Resolve=] |transition|'s [=ViewTransition/update callback done promise=]
```
vmpstr, line 1450: I feel like this is an extremely subtle point which I wouldn't have known if you didn't link to the your 2014 article.

"resolve the promise with the result of reacting to callbackPromise". This does, in fact, mean reject the promise if callbackPromise was rejected, right?

I'm not sure what race conditions you're referencing below, but I think it would be better to be explicit here "reject if rejects, resolve if resolved" type of language

css-view-transitions-1/Overview.bs
```diff
@@ -1458,17 +1472,15 @@ urlPrefix: https://html.spec.whatwg.org/multipage/rendering.html; type: dfn;
...
+		1. [=Reject=] |transition|'s [=ViewTransition/ready promise=] with |reason|.
 
-		1. [=promise/React=] to |transition|'s [=ViewTransition/update callback done promise=]:
+			Note: The [=ViewTransition/ready promise=] may already be resolved at this point,
+			if {{ViewTransition/skipTransition()}} is called after we start animating.
+			In that case, this step is a no-op.
 
-			* If |transition|'s [=ViewTransition/update callback done promise=] was resolved,
-				then [=resolve=] |transition|'s [=ViewTransition/finished promise=].
+		1. [=Resolve=] |transition|'s [=ViewTransition/finished promise=] with the result of [=reacting=] to |transition|'s [=ViewTransition/update callback done promise=]:
```
vmpstr, line 1481: Similar comment here

css-view-transitions-1/Overview.bs
jakearchibald: > Can you clarify this a bit: Which rejections will cause unhandledrejection? Is it that if this promise is rejected before it is handled, then it will cause an unhandledrejection?

Pretty much. If a promise is rejected, but not handled before the next task, it becomes an unhandled rejection. I think it's a kinda rough edge of promises tbh https://jakearchibald.com/2023/unhandled-rejections/.

> Also "these promises": that's the ready promise and the update callback done promise, but not the finished promise, right?

Yeah. This note is too vague. I'll fix that.

css-view-transitions-1/Overview.bs
jakearchibald: Imagine code like this:

```js
async function attemptTransition() {
  const transition = document.startViewTransition(callback);

  try {
    await transition.updateCallbackDone;
  } catch (err) {
    reportBadError(err);
    return;
  }

  try {
    await transition.ready;
    // Do some JS animations here
  } catch (err) {
    reportSkippedTransition(err);
    return;
  }

  try {
    await transition.finished;
    // Do some cleanup
  } catch (err) {
    reportCleanupError(err);
    return;
  }
}
```

If the update callback throws, the above code will result in two identical unhandled rejections. Although the `updateCallbackDone` promise is handled, the function exits, leaving `ready` and `finished` unhandled. That doesn't seem useful, and annoying to work around. So, my goal in this PR is to avoid unhandled rejections on `ready` and `finished` when the rejection would be the same as `updateCallbackDone`.

Note that as soon as the developer does:

```js
await transition.finished;
```

…they're creating a new unhandled promise. This PR only impacts cases where the developer doesn't do anything with the promises.

css-view-transitions-1/Overview.bs
jakearchibald: > "resolve the promise with the result of reacting to callbackPromise". This does, in fact, mean reject the promise if callbackPromise was rejected, right?

Yeah, the process of [react](https://webidl.spec.whatwg.org/#dfn-perform-steps-once-promise-is-settled) is very similar to `promise.then()`, as in, if you don't react to the reject case, then the rejection is passed through.

I'll add a note to make this clear.

The only reason the fulfilled case is handled is to ensure that `updateCallbackDone` fulfills with `undefined` rather than whatever the callback returns.

jakearchibald: > I'm not sure what race conditions you're referencing below

We didn't have any in the spec right now, but since our intention was to have `updateCallbackDone` follow `callbackPromise`, except resolving with `undefined`, it seems better to spec it with that intent.

Here's the difference between the two patterns: https://jsbin.com/butulov/edit?js,console

css-view-transitions-1/Overview.bs
jakearchibald: I'll add a note here too.

css-view-transitions-1/Overview.bs
vmpstr: I was imagining the case where updateCallbackDone is listened to by some app state monitor, to make sure that all of the proper dom modifications take place, whereas finished is listened to by some logging / metrics subsystem to just report failures. In that case, suppose that updateCallbackDone rejects and is handled appropriately by the app state monitor. However, the logging system doesn't handle it (maybe as an oversight) but it does not result in unhandled rejection because updateCallbackDone handled it.

I think my case is fairly contrived, but I'm just wary of handling rejections for promises that seem different enough.

WDYT?

jakearchibald: Yeah, I see what you're saying. But it doesn't feel that different to what would happen if both systems were listening to `updateCallbackDone`, which I think it more likely in this case (`finished` only ever rejects if `updateCallbackDone` rejects).

- The app state handles `updateCallbackDone`.
- The logging system intends to handle `updateCallbackDone`, but fails to for some code path.

There's no unhandled rejection from `updateCallbackDone`, because the app state already handled it.

I'll ask Domenic for a second opinion on this though.

Tangent: I kinda think unhandled rejections in general are a bad design. They seem to be built around single consumer.

jakearchibald: Domenic likes the pattern in this PR (he did a similar thing in the navigation API) https://github.com/w3c/csswg-drafts/issues/8455#issuecomment-1436832150.

But I'm not opposed to take it to the group internally, or CSSWG, if you want more eyes on it.

PR #8461, opened 2023-02-16
[css-cascade-6] Disallow pseudo-elements in @scope prelude

### LINKED ISSUE #7382: [css-cascade-6] Disallow pseudo-elements in @scope's prelude
I think we should make it a parse error to use pseudo-element selectors in `@scope`'s prelude (like we already do for e.g. `:is/:where`). It's a significant engineering effort to support properly, and it doesn't seem very useful. You could reference the pseudo-scoping-root itself with `:scope`, but not much else.

We should disallow it, at least until there is a point in allowing it.

andruud:
Resolves #7382.


css-cascade-6/Overview.bs
```diff
@@ -553,7 +556,10 @@ Changes since the 21 December 2021 First Public Working Draft</h3>
 	Significant changes since the
 	<a href="https://www.w3.org/TR/2021/WD-css-cascade-6-20211221/">21 December 2021 First Public Working Draft</a> include:
 
-	* Removed selector scoping notation.
+	* Disallowed [=pseudo-elements=] in the ''@scope'' prelude.
+		(<a href="https://github.com/w3c/csswg-drafts/issues/7382">Issue 7382</a>)
+
+  * Removed selector scoping notation.
```
mirisuzanne, line 562:
```suggestion
* Removed selector scoping notation.
```

css-cascade-6/Overview.bs
```diff
@@ -476,6 +476,9 @@ Scoping Styles: the ''@scope'' rule</h4>
 	are [=scoped selectors|scoped by=]
 	the selectors of the outer one.
 
+	[=Pseudo-elements=] cannot be [=scoping roots=] or [=scoping limits=];
```
mirisuzanne, line 479: Should this move up into `2.5 Scoped Styles` where `<scope-start>` and `<scope-end>` are defined? Merging as-is, but might want to come back and move this, so making note of it…

mirisuzanne (approved):

Issue #7349, opened 2022-06-11
[css-cascade-6] add a :stylesheet selector to match the current stylesheet


faceless2:
This is derived from both https://github.com/w3c/csswg-drafts/issues/3547 and (in particular) https://github.com/w3c/csswg-drafts/issues/6606, but I've opened a new issue as it's a more specific proposal.

Suggestion is: add a `:stylesheet` pseudo-class selector which functions like `:scope` but matches the element that declared the rule - ie the `<style>` or `<link>`element that created the stylesheet. For programatically created stylesheets, it would match nothing.

* I suspect this would be very easy to implement as the rule should already have a reference to the stylesheet that declared it.
* it would mean the functionality of the old `<style scoped>` syntax could be reproduced. This has been pretty heavily discussed in 3547, it was clearly popular, and while `@scope` is good it doesn't(*) provide a direct replacement for that functionality. `:stylesheet` would do that:

```css
<style>
@scope (* > :stylesheet) {
...
}
</style>
```

(* you could give the `<style>` an id and reference that from the `@scope` rule, so it's technically possible; just not very flexible)

tabatkins:
I think you need `@scope (:has(> :stylesheet)) {...}`, but yeah, that'd work.

Anyway, huh. Slightly funky idea, but I suppose it works and satisfies a need. I'd be fine with it.

mirisuzanne:
It seems to me like all the use-cases for this actually require selecting the parent of the style sheet, which @tabatkins achieves by using the `:has()` selector. I don't know any reason to select the stylesheet itself. Is there a reason we would do this, rather than something like `:stylesheet-parent` or something similar that directly targets the parent?

I suppose this allows you to more generically define _which ancestor_ of the style sheet to select?

faceless2:
> Is there a reason we would do this, rather than something like :stylesheet-parent or something similar that directly targets the parent?

Not really, to be honest. For the use-case I had in mind for this, `:stylesheet-parent` is fine. Only being able to select the parent would remove the ability to distinguish between two sibling stylesheets, and apply selectors to their siblings (for example) - but I don't have a use-case that would require that

mirisuzanne:
Yeah, `:stylesheet` would allow more flexibility around where the stylesheet lives in relation to the scoping element. That might be useful? But it would then require more complex selectors, and flexibility comes with more potential for unclear relationships. 

Agenda+ to hopefully resolve on:

[...]

PR #8463, opened 2023-02-17
[css-cascade-6] Add implicit scopes

### LINKED ISSUE #6606: [css-cascade-6] A way for <style> elements to scope to their parent element
https://drafts.csswg.org/css-cascade-6/#scoped-styles

The [explainer](https://css.oddbird.net/scope/explainer/#references--acknowledgements) references this issue: [Please bring back scoped styles](https://github.com/w3c/csswg-drafts/issues/3547).

Either there's something missing from the spec or I'm missing it, which is a way to have a `<style>` element that gets scoped to its own parent element.

```html
<div>
  <style scoped>p { color: red; }</style>
  <p>red</p>
</div>
```

`<style scoped>` would guarantee that the color only applied to children of the parent div.

I could do it by adding a unique ID or attribute:

```html
<div id="foo">
  <style>#foo p { color: red; }</style>
  <p>red</p>
</div>
```

… this is basically what Vue does at build time. But I want the same effect without a build step!

I think we'd want something like `@scope(:style-parent)`.

```html
<div>
  <style>@scope(:style-parent) { p { color: red; } }</style>
  <p>red</p>
</div>
```

Otherwise, even with `@scope` back in play, I'd still have to use a build tool (or something) to generate an ID for the upper bound of the scope.

andruud:
Resolves #6606 #7349


andruud:
@mirisuzanne Please have a look.

I felt that I had to reorganize a bit here to make a natural place for "implicit" scopes to live.

Description of changes:

 - A scope should probably have a _single_ scoping root. However:
 - A single `@scope` rule can produce multiple (explicit) scopes. Or a single implicit scope.
 - Removed the circular or near-circular sentence "selectors are scoped to the given scope, with the :scope element being the scoping root" which Emilio complained about at some point (somewhere).
 - Avoid selector-4's definition of "scoped" for now, because it does not clarify what it means by "descendant", and does not account for scoping limits.
 - Beautify the grammar part of the spec.

Hopefully it's more or less to your liking.

css-cascade-6/Overview.bs
```diff
@@ -351,17 +351,49 @@ Scoping Styles: the ''@scope'' rule</h4>
 
 	The syntax of the ''@scope'' rule is:
 
-	<pre class='prod'>
-	@scope (<<scope-start>>) [to (<<scope-end>>)]? {
+	<pre class="prod def">
+	@scope [ (<<scope-start>>) [to (<<scope-end>>)]? ]? {
```
mirisuzanne, line 355: I think we should still allow scoping limits, even when the root is implicit?

```suggestion
	@scope [(<<scope-start>>)]? [to (<<scope-end>>)]? {
```

css-cascade-6/Overview.bs
```diff
@@ -351,17 +351,49 @@ Scoping Styles: the ''@scope'' rule</h4>
...
+
+	* For each element matched by <<scope-start>>,
+		create a [=scope=] using that element as the [=scoping root=].
+	* For each [=scope=] created by the above:
+		* Collect all elements that are [=in scope=]
+			and that match <<scope-end>>,
+			using the [=scoping root=] as the '':scope'' element, then
+		* Set those elements as the [=scoping limits=].
+
+	[=Pseudo-elements=] cannot be [=scoping roots=] or [=scoping limits=];
```
mirisuzanne, line 378: After moving this above in the previous PR, I moved it back down! 😅

mirisuzanne (commented):
@andruud I think it's a big improvement, thanks. I suggested a change to the syntax production, allowing lower boundaries when the root is implicit. Otherwise, it looks good to me.

css-cascade-6/Overview.bs
andruud: :-)

andruud:
> @andruud I think it's a big improvement, thanks. I suggested a change to the syntax production, allowing lower boundaries when the root is implicit.

Ah, right, that didn't occur to me. But then we need another fixup: PR: https://github.com/w3c/csswg-drafts/pull/8466

PR #8468, opened 2023-02-17
[css-cascade-6] Add CSSScopeRule

andruud:


mirisuzanne:
I don't know if I'm the best to review this, since I'm not an editor on CSSOM - but from an `@scope` perspective, it looks good to me.

mirisuzanne (approved):

tabatkins (approved):

PR #8471, opened 2023-02-19
[scroll-animations-1] Add self keyword

### LINKED ISSUE #8227: [scroll-animations-1] Allow Anonymous Scroll Progress Timelines to target self

bramus:
Fixes #8227.

This PR replaces #8390 which was created from the wrong branch.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -187,9 +187,9 @@ spec: cssom-view-1; type: dfn;
 
 		<dt><dfn>self</dfn>
 		<dd>
-			Specifies to use the element’s own [=principal box=] as the
-			[=scroll container=]. If the [=principal box=] is not a 
-			[=scroll container=], or if there is no [=scrollable overflow=], 
+			Specifies to use the element’s own [=principal box=] as the [=scroll container=].
+			If the [=principal box=] is not a [=scroll container=], 
+			or if there is no [=scrollable overflow=], 
 			then the [=scroll progress timeline=] is [=inactive timeline|inactive=].
 	</dl>
 
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -187,7 +187,10 @@ spec: cssom-view-1; type: dfn;
 
 		<dt><dfn>self</dfn>
 		<dd>
-			Specifies to use the element’s own box as the [=scroll container=].
+			Specifies to use the element’s own [=principal box=] as the
+			[=scroll container=]. If the [=principal box=] is not a 
+			[=scroll container=], or if there is no [=scrollable overflow=], 
+			then the [=scroll progress timeline=] is [=inactive timeline|inactive=].
 	</dl>
 
 	Each use of ''scroll()'' corresponds to its own instance of {{ScrollTimeline}}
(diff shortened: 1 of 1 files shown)


bramus:
Thanks for the feedback @fantasai; greatly appreciate it. Did some changes which should address your remarks. Happy to take more pointers, as this is all fairly new to me.

fantasai:
Looks pretty good! The one other thing to fix is the line breaks.
Read https://rhodesmill.org/brandon/2012/one-sentence-per-line/ ; then take a look at the rest of the document to get a sense of how we're formatting the source.

bramus:
Interesting link. Makes sense, instead of cutting off mid-sentence when the line length limit is almost reached. Think I got it right now.

flackr:
> You need to define what happens when the element's own box isn't a [=scroll container=].

I would have thought these cases are covered by the text in the [ScrollTimeline interface](https://drafts.csswg.org/scroll-animations-1/#scrolltimeline-interface):
> If the [source](https://drafts.csswg.org/scroll-animations-1/#dom-scrolltimeline-source) of a [ScrollTimeline](https://drafts.csswg.org/scroll-animations-1/#scrolltimeline) is an element whose [principal box](https://drafts.csswg.org/css-display-4/#principal-box) does not exist or is not a [scroll container](https://drafts.csswg.org/css-overflow-3/#scroll-container), or if there is no [scrollable overflow](https://drafts.csswg.org/css-overflow-3/#scrollable-overflow), then the [ScrollTimeline](https://drafts.csswg.org/scroll-animations-1/#scrolltimeline) is [inactive](https://drafts.csswg.org/web-animations-1/#inactive-timeline).

Do we define this for [scroll-timeline-name](https://drafts.csswg.org/scroll-animations-1/#scroll-timeline-name) as well since that can target a non scroll container?

scroll-animations-1/Overview.bs
```diff
@@ -181,6 +181,13 @@ spec: cssom-view-1; type: dfn;
 		<dt><dfn>root</dfn>
 		<dd>
 			Specifies to use the document viewport as the [=scroll container=].
+
+		<dt><dfn>self</dfn>
+		<dd>
+			Specifies to use the element’s own [=principal box=] as the [=scroll container=].
+			If the [=principal box=] is not a [=scroll container=], 
+			or if there is no [=scrollable overflow=],
```
flackr, line 189: I think this condition at least should already be covered by the definition for [Scroll Progress Timelines](https://drafts.csswg.org/scroll-animations-1/#scroll-timelines):

> If the 0% position and 100% position coincide (i.e. the scroll container has no overflow to scroll), the timeline is [inactive](https://drafts.csswg.org/web-animations-1/#inactive-timeline).

flackr (approved):

scroll-animations-1/Overview.bs
fantasai: I think you're right, and this line should be removed.

```suggestion
```

bramus:
@fantasai I’ve committed the suggested change.

fantasai (approved):

PR #8480, opened 2023-02-21
[various] Replace optional group with implicitly optional comma

cdoublev:
Editorial change replacing explicit optional groups with implicitly optional commas, ie. `... [ , ... ]?` with `... , ...?`.

I would appreciate this change because parsing against these two syntaxes with the CSS parser I am using, produces two different results (ie. `List(..., optionalList(comma, ...))` vs. `List(..., comma, ...)`), and the second is easier to handle.

I searched for all occurrences with `/\[\s*,.+\]\s*\?/`. I did not replace the result found in the value definition of [`copy-into`](https://drafts.csswg.org/css-gcpm-4/#propdef-copy-into) (CSS GCPM 4, not ready for implementation): the whole alternation can be omitted, which seems to be a bug.

  > **Name:** `copy-into`
  > **Value**: `none | [ [ <custom-ident> <content-level>] [, <custom-ident> <content-level>]* ]?`

It should probably be `none | [ <custom-ident> <content-level> ]#`.

### DIFF
--- a/css-conditional-values-1/Overview.bs
+++ b/css-conditional-values-1/Overview.bs
@@ -180,7 +180,7 @@ to different values based on certain conditions.
 
 <pre class="prod def">
 
-<<if()>> = if( <<condition>>, <<consequent>> [, <<antecedent>>]?)
+<<if()>> = if( <<condition>>, <<consequent>>, <<antecedent>>? )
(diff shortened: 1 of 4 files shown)


tabatkins:
> It should probably be `none | [ <custom-ident> <content-level> ]#`.

Looks like it, yeah.

PR #8489, opened 2023-02-23
[css-values-4] Section 6.1: Removed the word "unit" to avoid confusion

facorread:
Non-substantive contribution for grammar

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -1742,7 +1742,7 @@ Font-relative Lengths: the ''em'', ''rem'', ''ex'', ''rex'', ''cap'', ''rcap'', 
 	</figure>
 
 	<dl export dfn-type=value dfn-for="<length>">
-		<dt><dfn id="em" lt="em">em>
+		<dt><dfn id="em">em</dfn>
 		<dd>
 			Equal to the computed value of the 'font-size' property of the element on which it is used.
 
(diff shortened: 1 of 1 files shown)


tabatkins:
I'm not sure what the motivation behind this change is. The spec reads fine as-is.

(You've also made several additional changes - the very first edit deletes part of the end tag, you've removed some caniuse attributes, etc.)

facorread:
Honestly I am an outsider to all of this. Here is a screenshot of the standard today:

image

The screenshot shows that the definitions do not have the word ' unit' inside the single quotes before section 6.1.1; the confusion I was alluding to is that it makes it seem viable to write

```css
h1 {
    font-size: 1.2em unit
}
```

Similarly, I inferred that the `caniuse` values in the diff were derived from those typos.

Please let me know if these changes are introducing more problems than they fix.

tabatkins:
If you want to move the word "unit" out of the `dfn` element, so it won't be inside the generated quotes, that would be fine with me.

facorread:
I am happy to move the word "unit" out of `dfn`; I am not sure if there is some sort of standard way of building these titles. In another screenshot, the definitions at Section 7 come without "units" or any other word accompanying them whatsoever:

image

tabatkins:
Ah, fair. Might as well be consistent. Okay, well fix the markup error your very first edit introduces, and put the caniuse attribute back, and I can accept this. You can go ahead and remove the `lt` attributes as well, since they'll be matching the text content and thus are unnecessary.

Issue #4441, opened 2019-10-22
[css-transitions-2] Start transitions on discrete animation types


flackr:
One of the goals of web-animations is to [allow developers to define custom transitions](https://github.com/w3c/csswg-drafts/issues/2068#issuecomment-349194890). In theory, this would let a developer define a transition for property types which are animation type Discrete, however css-transitions-1 defines discrete animations as not transitionable:

https://drafts.csswg.org/css-transitions-1/#starting
> When comparing the before-change style and after-change style for a given property, the property > values are transitionable if they have an animation type that is neither not animatable nor discrete.

This issue is to explore the possibility of allowing developers to create custom transitions for such properties. Since a lot of content today likely relies on the current behavior (`transition-property: all` is a very common pattern), any solution would likely have to preserve the existing visual behavior

### Option 1 - Use custom easing on discrete transitions

We could define transitions on discrete properties to create a transition using an easing type which immediately assumes the end value. This would preserve the existing visual behavior while allowing developers to listen for and modify the transition.

However, if we assume that a lot of existing comment has transition: all >0 duration, doing all of the work to create and run these transitions may be prohibitively expensive.

### Option 2 - Start and cancel discrete transitions

Start the transition animation, and immediately cancel it, such that a developer could access it and play it / modify it from a transitionstart event?

birtles:
Yes, it would be nice to support this.

Apart from perf considerations, I'm a bit concerned about compatibility with option 1. If scripts are, for example, waiting for transitionend events and then asserting their `propertyName` is such-and-such, then firing transitions for other events could lead to breakage.

Option 2 would still result in extra `transitionrun` and `transitioncancel` events but I expect far less content is actually listening to them so they are less likely to produce breakage.

What about not allowing `all` or any shorthand to ever match discrete properties but still creating transitions for them (using the behavior from option 1) if they are explicitly listed? e.g. `transition-property: all, overflow-x`?

I suppose that runs the risk that if a longhand is promoted to a shorthand we might need to special case it so it keeps working (depending on how we define a discrete property). Then again, any content listening for `transitionrun` events on the (now) shorthand name, would already break since the UA would start firing events on the corresponding longhand names instead.

The other option that comes to mind is simply to add extra syntax for this, be it an extra property (`transition-property-set: discrete-too`) or extra syntax attached to property names.

TheBigApple:
Flackr,
[...]

PR #8520, opened 2023-03-03
[css-transitions-1] Make discrete properties transition at 50%; exclude them from 'all'. #4441

fantasai:


css-transitions-2/Overview.bs
```diff
@@ -72,12 +72,29 @@ spec:web-animations-1; type:dfn;
...
 name</dfn> of a transition is the name of the longhand sub-property for which
 the transition was generated (e.g. 'border-left-width').
 
+ISSUE: Update the defining instance of [=transitionable=]
+once it is ported from Level 1.
+
+In CSS Transitions Level 2,
+property values are [=transitionable=]
+<em>unless</em> they have an [=animation type=] that is [=not animatable=],
+i.e. [=discrete=] values <em>are</em> [=transitionable=],
```
dbaron, line 90: Maybe make this its own sentence and make it a note?

css-transitions-2/Overview.bs
```diff
@@ -72,12 +72,29 @@ spec:web-animations-1; type:dfn;
 
 ## The 'transition-property' Property ## {#transition-name-property}
 
+ISSUE: Insert text from [[css-transitions-1#transition-property-property]]
+and backport the following paragraph.
+
 Although the 'transition-property' may specify shorthand properties and the
 ''all'' keyword, individual transitions are generated for each longhand
 sub-property that is animatable. The <dfn>expanded transition property
```
dbaron, line 80: I think this looks like a previous edit that may have been intended to partly fix this issue.  However, I think this paragraph needs modification to match the resolution -- shorthands should consider all "transitionable" or "animatable" properties (I think), whereas "all" should reject both properties that are "not animatable" or whose animation type is "discrete".

On the other hand, I think perhaps level 2 should really get rid of the "transitionable" definition since it just means not "not animatable", and just use "not animatable" as needed.

css-transitions-2/Overview.bs
```diff
@@ -72,12 +72,29 @@ spec:web-animations-1; type:dfn;
...
+
+In CSS Transitions Level 2,
+property values are [=transitionable=]
+<em>unless</em> they have an [=animation type=] that is [=not animatable=],
+i.e. [=discrete=] values <em>are</em> [=transitionable=],
+and flip at 50% progress.
+However, because authors will rarely want [=discrete=] values to transition at 50%
+(and for compatibility with CSS Transitions Level 1)
+the ''transition-property/all'' keyword
+expands only to all [=transitionable=] properties
```
dbaron, line 95: I think this should probably be its own paragraph (since it's specifically about changing the definition of `all`), and perhaps should be more clearly split into (1) a change to the description of `all` and (2) the motivation for that change.

dbaron (commented):
I think this is mostly good, although there are a number of things that are awkward about doing this as a delta spec and it's not completely clear how diff-like to make some of the edits.

I have a few comments below.

In addition to those comments, I think you should specifically patch all occurences of "transitionable" in the "Starting of transitions" section, which does need to be patched to make this part of the detailed algorithm.  I think that may need a term (I have mixed feelings about whether to re-use "transitionable") that considers whether the relevant part of `transition-property` was `all` or not.

css-transitions-2/Overview.bs
dbaron: (OK, github reviews doesn't actually associate my comment with the text I selected, so I was referring to the bit starting with "i.e.".)

css-transitions-2/Overview.bs
fantasai: I'm just adding an issue here to clarify that this section is incomplete; I'm not changing the existing paragraph that follows the issue, which is defining a term used elsewhere in the spec. I don't see any reason why that paragraph would need edits here.

css-transitions-2/Overview.bs
fantasai: Done.

css-transitions-2/Overview.bs
fantasai: Rewrote it differently, but kept it normative because the bit about transitioning at 50% seemed worth stating explicitly.

fantasai:
>  I think you should specifically patch all occurences of "transitionable" in the "Starting of transitions" section, which does need to be patched to make this part of the detailed algorithm

I'm patching the definition of "transitionable" here; I don't see that, with that change, anything in "Starting of transitions" needs edits?

css-transitions-2/Overview.bs
dbaron: Yeah, I know you weren't editing it... but I was suggesting that you should.  That existing paragraph (and I didn't check why it was inserted... but I'm somewhat curious) states pretty clearly that `all` and shorthands work the same way.

dbaron:
> I'm patching the definition of "transitionable" here; I don't see that, with that change, anything in "Starting of transitions" needs edits?

With your patch to "transitionable", the detailed algorithm in "Starting of transitions" will start transitions for discretely-animatable properties when `all` is specified for `transition-property`, which is incorrect.

fantasai:
> With your patch to "transitionable", the detailed algorithm in "Starting of transitions" will start transitions for discretely-animatable properties when all is specified for transition-property, which is incorrect.

I don't see where it says that. It doesn't mention the `all` keyword in that section at all, afaict. And the section we're editing here now says very clearly that discretely-animatable properties are not included in `all`. So I don't see how you're arriving at this conclusion.

css-transitions-2/Overview.bs
fantasai: It says that each sub-property of a shorthand of the `all` keyword generates its own transition. But we also say that the `all` keyword doesn't map to discretely-animatable properties (unlike the `all` shorthand itself), so discretely-animatable longhands aren't part of `all` and therefore aren't addressed by this paragraph afaict.

dbaron:
> I don't see where it says that. It doesn't mention the `all` keyword in that section at all, afaict. And the section we're editing here now says very clearly that discretely-animatable properties are not included in `all`. So I don't see how you're arriving at this conclusion.

OK, after rereading, I agree.

dbaron (approved):
OK, looks good.  Thanks.

fantasai:
OK, let me know if there's anything else to fix, or feel free to rebase+merge if it looks good to you.

dholbert:
I have an editorial nit with the new text here which I filed as https://github.com/w3c/csswg-drafts/issues/8761

dholbert:
I also have one question about various tests that were updated to test this change but I think are mistaken; see https://github.com/web-platform-tests/wpt/issues/39871

(Alternately, maybe the tests are right and the spec-text just needs clarification?)

PR #8526, opened 2023-03-04
[scroll-animations-1] Fixed `view()` grammar related to view timeline inset

### LINKED ISSUE #8519: [css-animations-2][scroll-animations-1] Is `view()` meant to allow multiple `view-timeline-inset` values?
The syntax of the `view()` function allows [`<'view-timeline-inset'>`](https://drafts.csswg.org/scroll-animations-1/#propdef-view-timeline-inset) values. And the syntax for `view-timeline-inset` is defined as a comma-separated list of values.

So, as it stands, the syntax currently allows values like `view(block 100px 200px, 10% 20%)`.

I assume `view()` was intended to only take _one_ set of view timeline insets. Correct?

If so, this could be solved by introducing a new `<single-view-timeline-inset>` data type that refers to `[ auto | <length-percentage> ]{1,2}`. And that data type could then be used for `view-timeline-inset` as `<single-view-timeline-inset>#` and for `<view()>` as `view( [ <axis> || <single-view-timeline-inset> ]? )`.

Sebastian

SebastianZ:
`view()` only refers to a single view timeline inset, so its grammar was changed to reflect that.

Fixes #8519

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -446,7 +446,7 @@ spec: cssom-view-1; type: dfn;
 	Its syntax is
 
 	<pre class="prod">
-		<<view()>> = view( [ <<axis>> || <<'view-timeline-inset'>> ]? )
+		<<view()>> = view( [ <<axis>> || <<single-view-timeline-inset>> ]? )
 	</pre>
 
 	By default,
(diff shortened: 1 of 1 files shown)


fantasai:
Oh. I forgot, the `<'property-name'>` notation already excludes top-level commas. (Because otherwise we have exactly this problem everywhere.) https://www.w3.org/TR/css-values-4/#component-types

Issue #8204, opened 2022-12-07
[scroll-animations-1] Define how the `source` member of a `ScrollTimeline` corresponding to a `scroll()` timeline is updated


birtles:
https://drafts.csswg.org/scroll-animations-1/#dom-scrolltimelineoptions-source

I might be misunderstanding how all this works but if I define an anonymous scroll timeline in CSS using the `scroll()` notation, then inspect the corresponding `Animation` object's `timeline` I assume I can inspect the `ScrollTimeline.source` to see what element `root` or `nearest` resolved to.

If I then modify the CSS of an ancestor element such that `root` or `nearest` resolve to a different element, is the same `ScrollTimeline` object updated? If so do we need to specify a procedure for updating the source of a `ScrollTimeline`? Or is a new `ScrollTimeline` minted? In either case we'd need to define it because the result would be observable.

fantasai:
This is a good question, and imho it would make the most sense to mint a new `ScrollTimeline`. @flackr ?

bramus:
Hmm, tricky. I guess there’s something to say for both, but leaning towards updating only the `source` as it essentially still is the same `scroll()` definition.

ydaniv:
Changing `source` will also change `currentTime` and may render the timeline _active/inactive_. Not sure I understand the specifics of minting a new `ScrollTimeline`, but we'll need to reflect more change than just the `source`.

fantasai:
@bramus That also brings up the question of, if I have multiple `scroll()` functions and a named timeline all referencing the same scroll container, do they share the same identity?

flackr:
FYI @kevers-google

In the implementation, we currently use the same timeline instance and it resolves the actual source and returns that from the instance - but it is impossible to construct a JS scroll timeline which refers to the nearest ancestor. From the JS side this doesn't seem too different from how animation timing is updated from CSS except when you updateTiming from the JS side overriding it. Each ScrollTimeline is currently a different instance, especially as different elements have different reference elements in the internal scroll timeline details. The owning element concept also feels similar: https://drafts.csswg.org/css-animations-2/#owning-element - an internal detail that affects the JS behavior but is not exposed.

flackr:
My proposal would be that we spec the chromium behavior. That similar to an animation object, a timeline object is created for the element which defines it and that same object is used until an explicit style change causes us to create a new one.

[...]

PR #8541, opened 2023-03-07
[scroll-animations-1] source is updated upon request and in sync with currentTime

fantasai:
See #8204

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -280,6 +280,9 @@ spec: cssom-view-1; type: dfn;
 
 	A {{ScrollTimeline}}’s {{AnimationTimeline/duration}} is 100%.
 
+	The values of {{ScrollTimeline/source}} and {{AnimationTimeline/currentTime}}
+	are both computed when either is requested or updated.
+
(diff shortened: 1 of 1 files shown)


flackr (changes requested):
These changes look good, but in the original issue you had also assumed that we minted a new scrolltimeline each time the subject changed https://github.com/w3c/csswg-drafts/issues/8204#issuecomment-1349919173 and we decided in the discussion that it was not a new one and had an identity consistent with animation creation. Does this also need to be clarified?

flackr:
Here's the relevant text from the css-animations spec:
https://drafts.csswg.org/css-animations/#interface-animationevent:~:text=If%20a%20match%20is%20found%2C%20the%20existing%20animation%20is%20updated

> Changes to the animation-name update existing animations by iterating over the new list of animations from last to first, and, for each animation, finding the last matching animation in the list of existing animations. If a match is found, the existing animation is updated using the animation properties corresponding to its position in the new list of animations, whilst maintaining its current playback time as described above.

flackr:
Oh n/m I didn't catch that you committed the identity thing separately. This looks good.

flackr (approved):

PR #8553, opened 2023-03-09
[scroll-animations-1] Progress percentages must be between 0% and 100%

### LINKED ISSUE #8552: [scroll-animations-1] Clarify that progress percentages must be between 0% and 100%
While checking https://github.com/w3c/csswg-drafts/issues/5203#issuecomment-1461962937, I noticed the scroll-animations-1 spec doesn’t mention that the percentages for the timeline ranges must be between 0% and 100%. 

Furthermore the [original `<keyframe-selector>` definition from css-animations-1](https://drafts.csswg.org/css-animations-1/#typedef-keyframe-selector) seems to have inadvertently been relaxed as the allowed range for `<percentage>` is no longer mentioned.

- css-animations-1:

    ```
    <keyframe-selector> = from | to | <percentage [0,100]>
    ```
- scroll-animations-1:

    ```
    <keyframe-selector> = from | to | <percentage> | <timeline-range-name> <percentage>
    ```

I think we should explicitly mention the allowed range.

bramus:
Fixes #8552.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -761,12 +761,12 @@ spec: cssom-view-1; type: dfn;
 	The CSS ''@keyframes'' rule is extended thus:
 
 	<pre class="prod">
-		<<keyframe-selector>> = from | to | <<percentage [0,100]>> | <<timeline-range-name>> <<percentage [0,100]>>
+		<<keyframe-selector>> = from | to | <<percentage [0,100]>> | <<timeline-range-name>> <<percentage>>
 	</pre>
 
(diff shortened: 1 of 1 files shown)


fantasai (changes requested):
Since we're considering allowing out-of-range percentages in https://github.com/w3c/csswg-drafts/issues/8552 I don't think we should be merging this PR.

If you only add [0,100] to the first instance of `<<percentage>>`, though, that would be acceptable because raw percentages must be between 0%-100%.

bramus:
Yeah, I was a bit too hasty with this PR. Good to read out-of-range percentages are allowed after all :)

fantasai (approved):

PR #8594, opened 2023-03-15
[css-cascade-4] Replace @import supports() example

CanadaHonk:
Current example does not work (legacy UAs do not support @import supports(), so will never be imported), replaced with one based on font-tech(color-COLRv1) instead (open to other suggestions).

### DIFF
--- a/css-cascade-4/Overview.bs
+++ b/css-cascade-4/Overview.bs
@@ -221,14 +221,11 @@ Conditional ''@import'' Rules</h3>
 	(even if it is loaded through some other link).
 
 	<div class="example">
-		The following rule illustrates how an author can provide fallback rules for legacy user agents
-		without impacting network performance on newer user agents:
+		The following rule illustrates how an author can provide rules for modern user agents
+		which support newer web features, for example, COLRv1 fonts:
 
 		<pre class='lang-css'>
-		@import url("fallback-layout.css") supports(not (display: flex));
-		@supports (display: flex) {
-		  ...
-		}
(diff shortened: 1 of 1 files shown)


fantasai (changes requested):
I think you need to keep the “without impacting network performance” bit, though.

CanadaHonk:
Whoops, sorry meant to keep that in, fixed!

css-cascade-4/Overview.bs
```diff
@@ -221,14 +221,12 @@ Conditional ''@import'' Rules</h3>
 	(even if it is loaded through some other link).
 
 	<div class="example">
-		The following rule illustrates how an author can provide fallback rules for legacy user agents
-		without impacting network performance on newer user agents:
+		The following rule illustrates how an author can provide rules for modern user agents
+		which support newer web features and without impacting network performance on unsupported user agents.
+		In this example, a CSS file which loads COLRv1 fonts is only fetched when it are supported:
```
fantasai, line 226: A few grammar fixes and rewrapping per https://rhodesmill.org/brandon/2012/one-sentence-per-line/ :)

```suggestion
		The following rule illustrates how an author can provide rules
		for modern user agents that support newer web features 
		without impacting network performance on older user agents.
		In this example, a CSS file which loads COLRv1 fonts
		is only fetched when COLRv1 is supported:
```

fantasai (changes requested):

css-cascade-4/Overview.bs
CanadaHonk: Fixed with suggested change, thanks!

Issue #8592, opened 2023-03-15
[css-borders-4] `box-shadow` needs to define coordinating list base property


Loirooriol:
https://drafts.csswg.org/css-backgrounds-4/#misc just keeps saying

> See the section [“Layering, Layout, and Other Details”](https://w3c.github.io/csswg-drafts/css-backgrounds-3/#shadow-layers) for how [...] interacts with other comma-separated drop shadow properties to form each drop shadow layer.

But the link points to L3, where `box-shadow` is a longhand.

This section needs to define these box shadow properties as a [coordinating list property group](https://drafts.csswg.org/css-values/#coordinating-list-property) and choose which one is the [coordinating list base property](https://drafts.csswg.org/css-values/#coordinating-list-base-property).

PR #8605, opened 2023-03-16
[css-backgrounds-4] Fixed and added definitions for box-shadow-offset: none

### LINKED ISSUE #8567: [css-backgrounds-4] Allow repeating `none` in `box-shadow-offset`
`none, none` is currently valid in [`box-shadow`](https://drafts.csswg.org/css-backgrounds-4/#propdef-box-shadow) but not in [`box-shadow-offset`](https://drafts.csswg.org/css-backgrounds-4/#propdef-box-shadow-offset).

`animation`, `transition`, `scroll-timeline`, `view-timeline`, accept declaring `none` with other values and repeating it in a [coordinating value list](https://drafts.csswg.org/css-values-4/#coordinated-value-list). In #8440, I suggested restricting `none` in these properties, before considering some (contrived) use cases.

```css
.box {
  box-shadow: 0px 0px, black none 1px;
}
.box:hover {
  box-shadow-offset: 1px 1px, 2px 2px;
}
```

Nit: stacking `{1,2}` and `#` does not seem invalid but some spec readers may think `[ <length>{1,2} ]#` is clearer.

SebastianZ:
I've added the definition for the `none` value of `box-shadow-offset` and corrected the computed value of that property to account for the `none` value now being part of the shadow list.

This is meant to fix #8567.

Sebastian

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -902,10 +902,11 @@ Inherited: no
 Percentages: N/A
 Computed value: see individual properties
 Animation type: by computed value,
-    appending blank shadows (''transparent 0 0 0 0'')
-    with a corresponding ''box-shadow-position/inset'' keyword as needed
-    to match the longer list
-    if the shorter list is otherwise compatible with the longer one
+	letting ''box-shadow-offset/none'' create a blank shadow
+	(''transparent 0 0 0 0'') when interpolated with non-'none' values
+	and appending blank shadows with a corresponding
+	''box-shadow-position/inset'' keyword as needed to match the longer list
+	if the shorter list is otherwise compatible with the longer one
 </pre>
 
   <p>The 'box-shadow' property attaches one or more drop-shadows to the box.
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -768,16 +768,24 @@ Initial: none
 Applies to: all elements
 Inherited: no
 Percentages: N/A
-Computed value: either 'none' or a list,
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
 Animation type: by computed value
 </pre>
 
 <p>The 'box-shadow-offset' property defines one or more drop shadow offsets.
-The property accepts a comma-separated list of horizontal and vertical offset pairs,
+The property accepts a comma-separated list.
+Each item in that list can either accept the ''box-shadow-offset/none'' value,
+which indicates no shadow, or a pair of horizontal and vertical offsets,
 where both values are described as <<length>> values.
 
 <dl>
+  <dt><dfn id="shadow-offset-none">none</dfn>
+  <dd>
+    No box shadow will be created.
+	Any other box shadow properties specified for this box shadow have no
+	effect.
+
   <dt><dfn id="shadow-offset-x">1st <<length>></dfn>
   <dd>
     Specifies the <dfn>horizontal offset</dfn> of the shadow.
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -775,7 +775,7 @@ Animation type: by computed value
 
 <p>The 'box-shadow-offset' property defines one or more drop shadow offsets.
 The property accepts a comma-separated list.
-Each item in that list can either accept the ''box-shadow-offset/none'' value,
+Each item in that list can either be the ''box-shadow-offset/none'' value,
 which indicates no shadow, or a pair of horizontal and vertical offsets,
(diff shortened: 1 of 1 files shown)


css-backgrounds-4/Overview.bs
Loirooriol, line 1: `box-shadow` should also change to "Computed value: see individual properties".

BTW, "Animation type" should also be "see individual properties", and the animation logic should go to the right property, but I guess that depends on #8592.

css-backgrounds-4/Overview.bs
```diff
@@ -771,16 +771,24 @@ Initial: none
...
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
 Animatable: by computed value
 </pre>
 
 <p>The 'box-shadow-offset' property defines one or more drop shadow offsets.
-The property accepts a comma-separated list of horizontal and vertical offset pairs,
+The property accepts a comma-separated list.
+Each item in that list can either accept the ''box-shadow-offset/none'' value,
```
Loirooriol, line 781:
```suggestion
Each item in that list can either be the ''box-shadow-offset/none'' value,
```

css-backgrounds-4/Overview.bs
SebastianZ: I've changed the computed value and removed the now incorrect note for "Animatable" about "treating `none` as a zero-item list".

I kept the animation logic at the shorthand for now and will fix it in the patch for #8592.

I also kept "Animatable" for now and replace it in another patch by "Animation type" as it's unrelated to this one.

css-backgrounds-4/Overview.bs

css-backgrounds-4/Overview.bs
```diff
@@ -895,12 +903,9 @@ Initial: none
...
 Inherited: no
 Percentages: N/A
-Computed value: either the keyword ''box-shadow-offset/none'' or
-    a list, each item consisting of four absolute lengths
-    plus a computed color and optionally also a ''box-shadow-position/inset'' keyword
+Computed value: see individual properties
 Animatable: by computed value,
-    treating ''box-shadow-offset/none'' as a zero-item list
-    and appending blank shadows (''transparent 0 0 0 0'')
+    appending blank shadows (''transparent 0 0 0 0'')
```
Loirooriol, line 908: Just dropping "treating `none` as a zero-item list" would change the behavior.

I think that `box-shadow-offset` should treat `none` as `0 0` when interpolated with non-`none` values.

SebastianZ: The idea was to align it with `animation-name`, i.e. a value of `none` cancels out this one shadow within the list.

And I think it is better to treat `none` as-is, meaning the interpolation between `none` and non-`none` values is discrete and the shadow appears or disappears.

Sebastian

Loirooriol: Well this would break https://software.hixie.ch/utilities/js/live-dom-viewer/saved/11442 which works with interoperability, so you need CSSWG resolution to change it.

css-backgrounds-4/Overview.bs
```diff
@@ -895,15 +903,13 @@ Initial: none
...
-    treating ''box-shadow-offset/none'' as a zero-item list
-    and appending blank shadows (''transparent 0 0 0 0'')
-    with a corresponding ''box-shadow-position/inset'' keyword as needed
-    to match the longer list
-    if the shorter list is otherwise compatible with the longer one
+	letting ''box-shadow-offset/none'' create a blank shadow
+	(''transparent 0 0 0 0'') when interpolated with non-'none' values
+	and appending blank shadows with a corresponding
+	''box-shadow-position/inset'' keyword as needed to match the longer list
+	if the shorter list is otherwise compatible with the longer one
```
Loirooriol, line 912: I don't think this works.

Let's say we interpolate from `box-shadow-offset: none; box-shadow-color: #f00` to `box-shadow-offset: 10px 10px; box-shadow-color: #f00`.

`box-shadow-offset` interpolates `none` as `0 0`, so at 50% we have `5px 5px`. But what about `box-shadow-color`? Its computed value was always `#f00` (`none` was only changing the used value into `transparent`), so we get `#f00`, but current browsers say `rgba(255, 0, 0, 0.5)`.

So I guess that either:

 - `box-shadow-offset: none` should affect the computed values of other properties (even though coordinated value lists were only supposed to affect each other at used value time I believe).
 - And interpolation with `box-shadow-offset: none` should not just treat it as `0 0` but somehow affect the used value of other properties at used value time.
 - We get rid of this magical behavior, as you initially wanted.

SebastianZ:
Since we split borders and box decorations out into CSS Borders 4, I'll close this PR and set up a new one to avoid rebasing issues.

Sebastian

SebastianZ:
For reference, the new PR is #9096.

PR #8613, opened 2023-03-18
[css-backgrounds-4] Defined `box-shadow-*` as coordinating list property group

### LINKED ISSUE #8592: [css-borders-4] `box-shadow` needs to define coordinating list base property

SebastianZ:
Copied over the section from Level 3 about "Layering, Layout, and Other Details" of box shadows and defined the longhands as coordinating list property group with `box-shadow-offset` being the coordinating list base property.

I've also move the "Animatable" definition to `box-shadow-offset`. Though I have to admit, I am not completely sure whether it fits there.
Maybe there should be a new section "Animation of Drop Shadows" explaining this in more detail?

This patch is based on the changes of #8605 to avoid any merge conflicts. So the other one should be merged first.

This is meant to fix #8592.

Sebastian

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -771,16 +771,24 @@ Initial: none
 Applies to: all elements
 Inherited: no
 Percentages: N/A
-Computed value: either 'none' or a list,
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
 Animatable: by computed value
 </pre>
 
 <p>The 'box-shadow-offset' property defines one or more drop shadow offsets.
-The property accepts a comma-separated list of horizontal and vertical offset pairs,
+The property accepts a comma-separated list.
+Each item in that list can either accept the ''box-shadow-offset/none'' value,
+which indicates no shadow, or a pair of horizontal and vertical offsets,
 where both values are described as <<length>> values.
 
 <dl>
+  <dt><dfn id="shadow-offset-none">none</dfn>
+  <dd>
+    No box shadow will be created.
+	Any other box shadow properties specified for this box shadow have no
+	effect.
+
   <dt><dfn id="shadow-offset-x">1st <<length>></dfn>
   <dd>
     Specifies the <dfn>horizontal offset</dfn> of the shadow.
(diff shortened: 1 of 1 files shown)


css-backgrounds-4/Overview.bs
```diff
@@ -771,16 +771,28 @@ Initial: none
...
-Computed value: either 'none' or a list,
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
-Animatable: by computed value
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
+Animatable: by computed value,
+    appending blank shadows (''transparent 0 0 0 0'')
+    with a corresponding ''box-shadow-position/inset'' keyword as needed
+    to match the longer list
+    if the shorter list is otherwise compatible with the longer one
```
Loirooriol, line 780: I guess that this should actually be split and put the relevant part to the corresponding longhand.
So `box-shadow-offset` will append `0 0`, `box-shadow-color` will append `transparent`, etc.

Also, to preserve the old behavior, I guess that `none` should be treated as `0 0` when interpolated with a non-`none` value?

css-backgrounds-4/Overview.bs
```diff
@@ -771,16 +771,28 @@ Initial: none
 Applies to: all elements
 Inherited: no
 Percentages: N/A
-Computed value: either 'none' or a list,
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
-Animatable: by computed value
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
+Animatable: by computed value,
```
cdoublev, line 776: Can you please use `Animation type`, as required in [Web Animations](https://drafts.csswg.org/web-animations-1/#animation-type)? 🙏 Otherwise `w3c/reffy` would extract rules with fields mixing `animationType`/`animatable`.

cdoublev: Ah sorry! https://github.com/w3c/csswg-drafts/pull/8605#discussion_r1141162192

SebastianZ:
Closing this PR in favor of #9186. The rest of the patch here should already be in the spec. now.

Sebastian

PR #8614, opened 2023-03-18
[css-images-4] Fixed interpolation of `stripes()` with different types

### LINKED ISSUE #8163: [css-images] `stripes()` interpolation makes no sense
https://drafts.csswg.org/css-images-4/#interpolating-1d-images

> There are only a few restrictions on what images are allowed to be interpolated:
> 1. Both the starting and ending image must have the same number of [`<color-stripe>`](https://drafts.csswg.org/css-images-4/#typedef-color-stripe)s.
> 2. Neither image uses a combination of [`<length>`](https://drafts.csswg.org/css-values-4/#length-value), [`<percentage>`](https://drafts.csswg.org/css-values-4/#percentage-value), and [`<flex>`](https://drafts.csswg.org/css-grid-2/#typedef-flex) stripes.
> 
> [...] Then, for each pair of stripes, interpolate the [thickness](https://www.w3.org/TR/css3-transitions/#animtype-length) and [color](https://www.w3.org/TR/css3-transitions/#animtype-color) independently.

How do you interpolate from `stripes(#000 10px, #111 20px)` to `stripes(#222 10fr, #333 20fr)`?
This is invalid, right? `stripes(#111 calc(5px + 5fr), #222 calc(10px + 10fr))`

I think that restriction 2 should be changed. Instead of checking that each `<1d-image>` does not mix different types, it should check that each pair of interpolated thicknesses does not mix different types.

SebastianZ:
This changes the interpolation of `stripes()` values to look at each pair of thicknesses instead of disallowing mixing values completely.

This fixes #8163.

Sebastian

### DIFF
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -2420,7 +2420,8 @@ Interpolating ''stripes()'' {#interpolating-stripes}
 
 	1. Both the starting and ending image must have the same number of <<color-stripe>>s.
 
-	2. Neither image uses a combination of <<length>>, <<percentage>>, and <<flex>> stripes.
+	2. Each pair of interpolated thicknesses must be of the same type,
+		i.e. both must either be of type <<length-percentage>>, or <<flex>>.
 
 	If the two images satisfy both constraints,
 	they must be interpolated as described below.


css-images-4/Overview.bs
```diff
@@ -2420,7 +2420,8 @@ Interpolating ''stripes()'' {#interpolating-stripes}
 
 	1. Both the starting and ending image must have the same number of <<color-stripe>>s.
 
-	2. Neither image uses a combination of <<length>>, <<percentage>>, and <<flex>> stripes.
+	2. Each pair of interpolated thicknesses must be of the same type,
+		i.e. both must either be of type <<length>, <<percentage>>, or <<flex>>.
```
Loirooriol, line 2424: Missing `>` in `<<length>`.

And the grammar uses `<length-percentage> | <flex>`, so it should probably be

```suggestion
		i.e. both must either be of type <<length-percentage>> or <<flex>>.
```

(which implies that you lengths and percentages can be mixed).

SebastianZ: Changed. And as noted, I'll open a new issue to discuss mixing _all_ values.

Loirooriol (approved):

Issue #7589, opened 2022-08-10
[web-animations-2] Web animations API for specifying timeline phases in animation options.


flackr:
As part of #7044 we [resolved](https://github.com/w3c/csswg-drafts/issues/7044#issuecomment-1175722383):
* Add ability to specify start and end delays in terms of position in a named phase position (keyword + percentage), e.g. animation-delay: contain 0%; animation-duration: contain 100% attaches the animation between the 0% and 100% points in the contain phase. (See also https://github.com/w3c/csswg-drafts/issues/4342].)
* Add shorthand that expands a phase name to 0% start delay and 100% end delay. E.g. animation-phase: contain expands to animation-delay: contain 0%; animation-end-delay: contain 100%.

These resolutions specify how in CSS you can target an animation to a particular phase of the view timeline. However, we also need a way to do the same from the web animations API. In particular the options passed for the timing of web animations are [EffectTiming](https://www.w3.org/TR/web-animations-1/#dictdef-effecttiming) and [OptionalEffectTiming](https://www.w3.org/TR/web-animations-1/#dictdef-optionaleffecttiming).

Here are a few options I can think of:
1. The simplest would be to extend the `delay` and `endDelay` options to also accept a string specification equivalent to the CSS counterpart (e.g. `delay: "enter 20%"`).
2. Extend `delay` and `endDelay` to also accept an object type representing the combination of phase and time, e.g. `delay: {phase: "enter", offset: CSS.percentage(20)}`.
3. Add new options, e.g. `startTime` / `endTime` or `phase` for specifying the "phase", "time" pairs. We would then have to decide what to do when `delay` and `endDelay` are also specified.

kevers-google:
Leaning towards option 1 as it seems easier for the web developer (less verbose than option 2 and avoids potential confusion due to over-specification with option 3).

ydaniv:
Also leaning towards 1, as authors expect the css animations and web animations APIs to reflect each other as much as possible.

flackr:
Option 1 certainly seems the most straightforward, though option 2 is more consistent with the direction we've been trying to move in terms of having full rich types for the things that are expressed as strings in CSS.

Regardless of option (1) or (2), we will still have similar over-specification issues if we also allow specifying non-"auto" durations e.g. #4342. I think we'll just have to choose a precedent - probably try to align this with what happens when you specify left, right and width on an element size.

flackr:
@birtles @graouts FYI

[...]

Issue #7637, opened 2022-08-23
[css-animations-2][web-animations-2] API For specifying phase linked offsets in keyframes


flackr:
In #7044 we [resolved](https://github.com/w3c/csswg-drafts/issues/7044#issuecomment-1175722383) to add phase linked offsets to the animations APIs. We discussed a simple CSS syntax such as the following:

```css
@keyframes fade-in-out-animation {
  enter 0% { opacity: 0; }
  enter 100% { opacity: 1; }
  exit 0% { opacity: 1; }
  exit 100% { opacity: 0; }
}
```

We didn't talk about a syntax for the web-animations API, however #7589 touches on an API for the similar `delay` and `endDelay` properties so we should probably do the same here. Concretely this means whichever format we choose for #7589 (i.e. String or Object), that could be specified for `offset` in the [Keyframe dictionary](https://drafts.csswg.org/web-animations-1/#processing-a-keyframes-argument:~:text=WebIDL%2Dlike%20definition%3A-,dictionary%20Keyframe,-%7B%0A%20%20%20%20//%20...%20property%2Dvalue%20pairs) and should likely be returned by [getKeyframes](https://drafts.csswg.org/web-animations-1/#dom-keyframeeffect-getkeyframes).

PR #8630, opened 2023-03-22
[web-animations-2] Add range offset APIs

fantasai:
See https://github.com/w3c/csswg-drafts/issues/7589


fantasai:
First question: Did we decide to do https://github.com/w3c/csswg-drafts/issues/7589#issuecomment-1307384718 ?

flackr:
> First question to @flackr: Did we decide to do [#7589 (comment)](https://github.com/w3c/csswg-drafts/issues/7589#issuecomment-1307384718) ?

Yes, I think we're good with this. Note that this type will be for keyframe offsets. The delay and endDelay will remain as is (not referring to ranges) and we'll add rangeStart and rangeEnd to KeyframeAnimationOptions.

flackr:
> Yes, I think we're good with this. Note that this type will be for keyframe offsets. The delay and endDelay will remain as is (not referring to ranges) and we'll add rangeStart and rangeEnd to KeyframeAnimationOptions.

I'm somewhat confused by whether CSSNumericValue and TimelineRangeOffset are indeed [distinguishable types](https://www.w3.org/TR/WebIDL-1/#dfn-distinguishable). I think it may be okay as interface is distinguishable from interface as long as no platform object implements both right?

tabatkins:
Interface types are specifically *not* fancy destructuring, those actually check that it's the correct object type - vended by the correct window object, even! Dictionary is just similar to `sequence<>` or `record<>` in that it's fancy destructuring.

fantasai:
Alright, this is ready for review. :) I tried my best to guess what the right thing to do was for https://github.com/w3c/csswg-drafts/issues/7637 ; not sure if I got all the details right.

web-animations-2/Overview.bs
```diff
@@ -2937,6 +2937,27 @@ new SequenceEffect(
...
+a {{DOMString}} is parsed into a {{TimelineRangeOffset}}
+according to the syntax of the ''@keyframes'' prelude,
+and the resulting {{TimelineRangeOffset}} or {{CSSNumericValue}}
+is interpreted as defined for ''@keyframes''.
+
+If the {{DOMString}} fails to parse as a valid ''@keyframes'' prelude,
+then the keyframe is invalid,
+and will throw a <span class=exceptionname>TypeError</span>
+when [[web-animations-1#processing-a-keyframes-argument|processed]]
+(same as if the <code>doube</code> value was outside the [0,1] range).
```
birtles, line 2959:
```suggestion
(same as if the <code>double</code> value was outside the [0,1] range).
```

Although actually I think we probably don't need this parenthetical?

web-animations-2/Overview.bs
```diff
@@ -2937,6 +2937,27 @@ new SequenceEffect(
...
+The <code>double?</code> type of the <code>offset</code> key
+in <code>ComputedKeyframe</code>,
+<code>BaseComputedKeyframe</code>,
+<code>BaseKeyframe</code>,
+<code>BasePropertyIndexedKeyframe</code>,
+etc.
+is replaced by <code>(CSSNumberish? or TimelineRangeOffset or DOMString)</code>,
+where <code>double</code> continues to be interpreted as a percentage,
+a {{DOMString}} is parsed into a {{TimelineRangeOffset}}
+according to the syntax of the ''@keyframes'' prelude,
```
birtles, line 2951: Is there a link to the syntax of the @​keyframes prelude so I can check what that is?

web-animations-2/Overview.bs
```diff
@@ -3028,10 +3049,40 @@ with which it is associated is updated.
...
-<div class="methods">
+Objects that may be the target of an {{KeyframeEffect}} object implement
+the {{Animatable}} interface mixin.
+
+<pre class='idl'>
+interface mixin Animatable {
+    Animation           animate(object? keyframes,
+                                optional (unrestricted double or KeyframeAnimationOptions) options = {});
+    sequence&lt;Animation&gt; getAnimations(optional GetAnimationsOptions options = {});
+};
```
birtles, line 3063: This seems to be duplicating text from level 1 as opposed to being a delta. Is that intentional?

web-animations-2/Overview.bs
```diff
@@ -3028,10 +3049,40 @@ with which it is associated is updated.
...
+    (TimelineRangeOffset or CSSNumericValue or DOMString) rangeStart = "normal";
+    (TimelineRangeOffset or CSSNumericValue or DOMString) rangeEnd = "normal";
+};
 
+dictionary GetAnimationsOptions {
+    boolean subtree = false;
+};
+</pre>
+
+ISSUE: Should rangeStart/rangeEnd also accept CSSNumericValue directly (for cases that don't want a rangeName)?
```
birtles, line 3083: I'm not sure what this refers to. It looks like the definitions of `rangeStart` / `rangeEnd` already to accept a `CSSNumericValue`?

web-animations-2/Overview.bs
```diff
@@ -3043,6 +3094,41 @@ with which it is associated is updated.
 
 </div>
 
+<div class="members">
+:   <dfn dict-member for=KeyframeAnimationOptions>id</dfn>
+::  The string to assign to the generated {{Animation}}'s
+    {{Animation/id}} attribute.
```
birtles, line 3100: Again, this part seems to be duplicating level 1 instead of being a delta. Is that intended?

web-animations-2/Overview.bs
```diff
@@ -3043,6 +3094,41 @@ with which it is associated is updated.
...
+:   <dfn dict-member for=KeyframeAnimationOptions>rangeStart</dfn>
+::  An optional value which, if present,
+    specifies the start of the <a>animation</a>’s [=timeline attachment range=].
+    A {{DOMString}} value is parsed as an 'animation-range-end' value
+    to yield a {{TimelineRangeOffset}} {{TimelineRangeOffset/rangeName}}
+    and {{TimelineRangeOffset/offset}};
+    a {{CSSNumericValue}} is interpreted as a {{TimelineRangeOffset}}
+    with that {{TimelineRangeOffset/offset}} and a null {{TimelineRangeOffset/rangeName}}.
+    The {{TimelineRangeOffset/rangeName}} and {{TimelineRangeOffset/offset}}
+    are the interpreted as defined for 'animation-range-end'.
```
birtles, line 3116: I don't follow what this sentence means (or really this paragraph). What does interpreted mean?

web-animations-2/Overview.bs
fantasai: Added it to help link the two concepts and where they occur in the processing model, since we're not yet incorporating the full text of the processing rules. I agree the parenthetical shouldn't be needed after merging in WA1, though.

Lmk if I should remove it; I don't mind too much either way, but I'm somewhat inclined to keep it until then, since I'm worried someone could come up a "but when is it thrown" type question otherwise...

web-animations-2/Overview.bs
fantasai: Replaced with `<keyframe-selector>`, which is more precise. :)

web-animations-2/Overview.bs
fantasai: Yes, I thought it would be less confusing...

web-animations-2/Overview.bs
fantasai: Thanks! Removed; that was leftover from a previous version. ^_^;;

web-animations-2/Overview.bs
fantasai: It means follow the definitions in https://drafts.csswg.org/scroll-animations-1/#animation-range-start for turning the rangeName and offset into a point on the timeline. I didn't want to inline those instructions, because they're changing as we adjust the syntax, defaulting rules, and capabilities of `animation-range`.

web-animations-2/Overview.bs
birtles: Sure, let's leave it until merging with WA1 then.

web-animations-2/Overview.bs
birtles: Thank you!

web-animations-2/Overview.bs
birtles: I think we should stick to being a delta if possible otherwise it's going to be really hard to know what the intentional changes are when we go to merge it 😬

web-animations-2/Overview.bs
birtles: For better or worse this spec is pretty procedural, especially from "Programming interface" onwards, so that you can mostly just convert spec prose into implementation code. I'm just wondering what an implementer is supposed to do with "interpreted as".

From the point of view of this interface, as long as `rangeName` and `rangeOffset` get set to the right values, I think we're done. How those values are applied goes somewhere else (ideally somewhere in Web Animations which might then refer to definitions in scroll animations).

web-animations-2/Overview.bs
flackr: Yeah I think we only need to say that we add rangeStart and rangeEnd to KeyframeAnimationOptions here, similar to how e.g. web-animations-2 adds iterationComposite to KeyframeEffectOptions https://www.w3.org/TR/web-animations-2/#the-keyframeeffectoptions-dictionary

web-animations-2/Overview.bs
fantasai: > How those values are applied goes somewhere else (ideally somewhere in Web Animations which might then refer to definitions in scroll animations).

Well, yeah, but that needs to be defined _somewhere_, and we don't have the conceptual infrastructure yet to reference it here. I can move the interpretation out of the definition here, and put it in a separate paragraph, but I can't update Web Animations 2 to define the attachment range concepts without a significant amount of additional editing work in sections that have not yet been ported over.

web-animations-2/Overview.bs
fantasai: Done.

web-animations-2/Overview.bs
```diff
@@ -3028,21 +3049,51 @@ with which it is associated is updated.
...
+partial dictionary KeyframeAnimationOptions : KeyframeEffectOptions {
+    (TimelineRangeOffset or CSSNumericValue or DOMString) rangeStart = "normal";
+    (TimelineRangeOffset or CSSNumericValue or DOMString) rangeEnd = "normal";
+};
+</pre>
 
+<div class="members">
+:   <dfn dict-member for=KeyframeAnimationOptions>rangeStart</dfn>
+::  An optional value which, if present,
+    specifies the start of the <a>animation</a>’s [=timeline attachment range=].
```
flackr, line 3073: [=timeline attachment range=] doesn't link when I check this out locally. Is it just a stale index or does it need to specify this comes from scroll-animations?

web-animations-2/Overview.bs
```diff
@@ -3028,21 +3049,51 @@ with which it is associated is updated.
...
+<pre class='idl'>
+dictionary TimelineRangeOffset {
+  CSSOMString? rangeName;
+  CSSNumericValue offset;
+};
 
-    > If this object is the [=effect target/target element=] of two or more
-    > <a>animation effects</a> which are associated with the
-    > same <a>animation</a>, the corresponding {{Animation}}
-    > object will still only appear in the returned list once.
```
flackr, line 3042: This seems to have been unintentionally removed.

web-animations-2/Overview.bs
birtles: Thanks, yeah, I think we should keep the description here just to how the value gets set. If we don't yet have anywhere to park the description of how those values are applied, we could add it here as a note of some sort indicating that it eventually needs to live somewhere else.

web-animations-2/Overview.bs
```diff
@@ -3028,21 +3049,51 @@ with which it is associated is updated.
...
+partial dictionary KeyframeAnimationOptions : KeyframeEffectOptions {
+    (TimelineRangeOffset or CSSNumericValue or DOMString) rangeStart = "normal";
+    (TimelineRangeOffset or CSSNumericValue or DOMString) rangeEnd = "normal";
+};
+</pre>
 
+<div class="members">
+:   <dfn dict-member for=KeyframeAnimationOptions>rangeStart</dfn>
+::  An optional value which, if present,
```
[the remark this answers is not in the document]
fantasai: Heh, I guess I'm thinking about it in CSS terms, where "optional" means you don't have to pass it in explicitly because we default it. :)

Other thing we forgot: this also needs to accept CSSKeywordValue, in order to represent "normal" in TypedOM form.

web-animations-2/Overview.bs
fantasai: It'll eventually link up: WA is going to need to set up the conceptual framework for attachment ranges (which Scroll Animations and CSS Animations 2 will link to), and that will define this term. But I wanted to get these bits of the API edits in first, since they're more concrete.

web-animations-2/Overview.bs
fantasai: Thanks, fixed.

web-animations-2/Overview.bs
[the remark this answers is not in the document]
fantasai: Fixed.

web-animations-2/Overview.bs
flackr: No problem, sounds good.

flackr (approved):

Issue #5771, opened 2020-12-08
[css-highlight-api] i18n self review


frivoal:
Short i18n review checklist is [here](http://w3c.github.io/i18n-activity/reviews/shortchecklist)

1. [x] _If the spec (or its implementation) contains any natural language text that will be read by a human (this includes error messages or other UI text, JSON strings, etc, etc),_

    It does not.
    
2. [x] _If the spec (or its implementation) allows content authors to produce typographically appealing text, either in its own right, or in association with graphics._

    It does not introduce any new typographical capabilities. It does allow applying a restricted subset of CSS's usual features to arbitrary ranges within the document. No i18n related issue is though to arise from that, and in any case, this merely builds up on built-in highlight pseudos as found in https://drafts.csswg.org/css-pseudo-4/#highlight-pseudos. Should there be any issue (and I don't think there is), they likely stem from that underlying technology rather than from this extension of it.

3. [X] _If the spec (or its implementation) allows the user to point into text, creates text fragments, concatenates text, allows the user to select or step through text (using a cursor or other methods), etc._

    It does not.

    However, it allows the *author* to do point to text fragments. It uses the pre-existing [Range object](https://dom.spec.whatwg.org/#range) from the DOM specification for that. Authors should be careful when doing so to avoid accidentally putting fragment boundaries in the middle of grapheme clusters. The Range object allows for that, but admittedly does not facilitate it. If it would be deemed useful to give authors advice about careful usage of the Range object to avoid getting into difficult i18n situations, it would seem more appropriate to include that advice in the upstream DOM specification.

4. [X] _If the spec (or its implementation) allows searching or matching of text, including syntax and identifiers_

    It does not.

5. [X] _If the spec (or its implementation) sorts text_

    It does not.

6. [X] _If the spec (or its implementation) captures user input_

    It does not.

7. [X] _If the spec (or its implementation) deals with time in any way that will be read by humans and/or crosses time zone boundaries_

    It does not.

8. [X] _If the spec (or its implementation) allows any character encoding other than UTF-8._

    It does not.

9. [X] _If the spec (or its implementation) defines markup._

    It does not.

10. [X] _If the spec (or its implementation) deals with names, addresses, time & date formats, etc_

    It does not.

11. [X] _If the spec (or its implementation) describes a format or data that is likely to need localization._

    It does not.

12. [X] _If the spec (or its implementation) makes any reference to or relies on any cultural norms_

    It does not.

sanketj:
[...]

PR #8647, opened 2023-03-27
[css-highlight-api-1] Add i18n warning about placement of range endpoints in text

sanketj:
Address #5771

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -162,6 +162,11 @@ Creating Custom Highlights</h3>
 	authors can chose between using {{Range}} objects and {{StaticRange}} objects.
 	See [[#range-invalidation]] for more details about this choice and its implications.
 
+	Note: When creating {{Range}} objects for use in [=custom highlights=], it is suggested
+	that authors avoid placing [=range=] endpoints in the middle of characters that are
+	comprised of multiple unicode codepoints. Doing so can create undesirable highlighting
+	effects, ex. highlighting a base character but not its combining mark.
+
 	<xmp class="idl">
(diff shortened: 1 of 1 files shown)


css-highlight-api-1/Overview.bs
```diff
@@ -162,6 +162,11 @@ Creating Custom Highlights</h3>
 	authors can chose between using {{Range}} objects and {{StaticRange}} objects.
 	See [[#range-invalidation]] for more details about this choice and its implications.
 
+	Note: When creating {{Range}} objects for use in [=custom highlights=], it is suggested
+	that authors avoid placing [=range=] endpoints in the middle of characters that are
+	comprised of multiple unicode codepoints. Doing so can create undesirable highlighting
+	effects, ex. highlighting a base character but not its combining mark.
+
```
aphillips, line 169: This should probably link the i18n-glossary and use our more formal terminology. Also note that our original ask has to do with quirks of DOM being encoded in UTF-16. Perhaps:

```suggestion
Note: When creating {{Range}} objects for using in [=custom highlights=], authors should avoid placing [=range=] endpoints in the middle of a [=grapheme=], such as when a visible unit of text is comprised of multiple [=Unicode code points=]. Doing so can create undesirable highlighting effects, such as highlighting only part of a Indic syllable. In addition, care should be exercised to avoid placing an endpoint in the middle of a [=supplementary character=].
```

aphillips (changes requested):
Thanks for this change. I made some terminology suggestions.

sanketj:
@aphillips Any more feedback on this?

aphillips (approved):
Looks great! Thank you!

Issue #7748, opened 2022-09-15
[scroll-animations-1] JS API for view-timeline-inset


flackr:
In CSS we have [view-timeline-inset](https://w3c.github.io/csswg-drafts/scroll-animations-1/#view-timeline-inset) to allow reducing the portion of the scroll area which would be considered visible, however the [JS ViewTimeline constructor](https://w3c.github.io/csswg-drafts/scroll-animations-1/#viewtimeline-interface) does not have the ability to set the same property.

I propose extending the [ViewTimelineOptions dictionary](https://w3c.github.io/csswg-drafts/scroll-animations-1/#dictdef-viewtimelineoptions) with a property that allows setting the one or two valued inset. The equivalent property in IntersectionObserver is a DOMString [rootMargin](https://w3c.github.io/IntersectionObserver/#dictdef-intersectionobserverinit) which is a space separated list of offsets. We could do the same here, but call it inset (to be consistent with the CSS property), or we could accept an array of CSSNumeric values.

Option 1: DOMString

```
dictionary ViewTimelineOptions {
  Element subject;
  ScrollAxis axis = "block";
  DOMString inset = "0px 0px";
};
```

Option 2: CSSNumeric

```
dictionary ViewTimelineOptions {
  Element subject;
  ScrollAxis axis = "block";
  sequence<CSSNumericValue> inset = [CSS.px(0), CSS.px(0)];
};
```

I think there's a slight advantage to accepting the string syntax in that:
* It allows specifying the `auto` value from view-timeline-inset, and
* You can pass the computed value of view-timeline-inset, e.g. `new ViewTimeline({subject: el, inset: getComputedStyle(subject).viewTimelineInset})`

However, passing and parsing a string seems slightly inconsistent with the direction of other JS API's where we've decided to go in the direction of richer types, e.g. https://github.com/w3c/csswg-drafts/issues/7589.

fantasai:
Do we want to allow both versions of input? Is that a thing we can do?

birtles:
Allowing both definitely seems attractive. Should `ViewTimeline` also expose these values as (readonly?) attributes? Perhaps using the typed version?

flackr:
> Allowing both definitely seems attractive.

Allowing both sounds reasonable to me.

> Should `ViewTimeline` also expose these values as (readonly?) attributes? Perhaps using the typed version?

Yes, I think it should.

fantasai:
[...]

PR #8648, opened 2023-03-27
[scroll-animations-1] Add inset option to ViewTimeline constructor #7748

fantasai:
See [proposal](https://github.com/w3c/csswg-drafts/issues/7748#issuecomment-1306097703), [WebIDL draft](https://github.com/w3c/csswg-drafts/issues/7748#issuecomment-1307381553), and [CSSWG resolution](https://github.com/w3c/csswg-drafts/issues/7748#issuecomment-1325424372).

Note that with the given syntax, it's not possible to ask for `auto` via the TypedOM version of the API. We could do that also, but we'd have to add some validation steps to make sure the keyword is in fact `auto`. (That said, this _is_ the initial value; it could just be omitted.)

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -534,6 +534,7 @@ spec: cssom-view-1; type: dfn;
 		dictionary ViewTimelineOptions {
 		  Element subject;
 		  ScrollAxis axis = "block";
+		  (DOMString or sequence&lt;(CSSNumericValue or CSSKeywordValue)>) inset = "auto";
 		};
 
 		[Exposed=Window]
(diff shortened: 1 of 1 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -530,6 +530,7 @@ spec: cssom-view-1; type: dfn;
 		dictionary ViewTimelineOptions {
 		  Element subject;
 		  ScrollAxis axis = "block";
+		  (DOMString or sequence&lt;CSSNumericValue>) inset = "auto auto";
```
flackr, line 533: Minor nit: "auto" expands to "auto auto" per https://drafts.csswg.org/scroll-animations-1/#view-timeline-inset . Might that be a simpler default?

>  If the second value is omitted, it is set to the first.

scroll-animations-1/Overview.bs
```diff
@@ -605,15 +606,20 @@ spec: cssom-view-1; type: dfn;
 			1. Set the {{ScrollTimeline/source}} of |timeline|
 				to the {{ViewTimeline/subject}}’s
 				nearest ancestor [=scroll container=] element.
+
+			1.  If a {{DOMString}} value is provided as an |inset|,
+				parse it as a <<'view-timeline-inset'>> value;
+				if a sequence is provided,
+				the first value represents the start inset
+				and the second value represents the end inset.
```
flackr, line 614: What if it's a sequence of length one? Should we follow the same convention as we use for parsing that a single value expands to both? What about a sequence of length 0? Should we throw a TypeError?

flackr:
> Note that with the given syntax, it's not possible to ask for `auto` via the TypedOM version of the API. We could do that also, but we'd have to add some validation steps to make sure the keyword is in fact `auto`. (That said, this _is_ the initial value; it could just be omitted.)

I believe the plan was to allow the keyword `auto` to be specified instead of a CSSNumericValue for either inset.

scroll-animations-1/Overview.bs
fantasai: Agree that repeating would be nice, and throwing a TypeError for anything else.

scroll-animations-1/Overview.bs
fantasai: Sgtm

fantasai:
@flackr OK, updated accordingly.

PR #8667, opened 2023-03-31
[scroll-animations-1] Add note about horizontal and vertical axis still using logical direction

### LINKED ISSUE #8665: [scroll-animations-1][css-writing-modes-4] Direction of horizontal / vertical axes
It's unclear to me whether the direction horizontal and vertical increase in is defined. From [css-writing-modes-4](https://w3c.github.io/csswg-drafts/css-writing-modes-4/#horizontal-dimension):

> The physical dimensions are width and height, which correspond to measurements along the x-axis (horizontal dimension) and y-axis (vertical dimension), respectively.

This clarifies that horizontal is always the x-axis, but there was some confusion when implementing scroll-timeline whether an animation on the horizontal axis should flip or not when in an RTL scroller.

There's effectively two options (for brevity only explaining horizontal):
1. Horizontal is equivalent to inline in a horizontal writing mode and block in a vertical writing mode, but follows the respective flow direction. This means that in RTL horizontal starts at the right and increases in the left direction.
2. Horizontal always starts at the left and increases to the right.

Whichever is the chosen one, I also wanted to make sure it's clear from the spec.

# Affected specs

* Affects animation direction in [scroll-animations-1](https://drafts.csswg.org/scroll-animations-1/)
* May affect [css-scroll-snap-1](https://w3c.github.io/csswg-drafts/css-scroll-snap/#properties-on-the-scroll-container)

# Examples

## Parallax

Transform (and I believe background-position) currently can only be expressed in physical axes (See #4350). Developers may expect to be able to use `horizontal` / `vertical` so that the scroll / view timeline is also physical and work with both transform:
```css
@keyframes slide {
  from { transform: translateX(0); } /* unnecessary but included for clarity. */
  to   { transform: translateX(200px); }
}
. parallax-transform {
  animation-name: slide;
  animation-timeline: scroll(horizontal);
  position: absolute;
  left: 0; /* starts anchored to left edge */
}
```

And background-position:
```css
@keyframes bg-slide {
  from { background-position: left center; }
  to   { background-position: right center; }
}
. parallax-transform {
  animation-name: bg-slide;
  animation-timeline: scroll(horizontal);
  background-image: url('parallax-background.webp');
}
```

flackr:
Fixes #8665

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -215,8 +215,9 @@ spec: cssom-view-1; type: dfn;
 	even if multiple elements use ''scroll()'' to refer
 	to the same [=scroll container=] with the same arguments.
 
-	Note: Progress in the horizontal or vertical axis still follows the logical scroll direction
-	which may flip depending on writing mode.
+	Note: Progress is in reference to the [=scroll origin=],
+	which can flip depending on [=writing mode=],
+	even when ''scroll-timeline-axis/horizontal'' or ''scroll-timeline-axis/vertical'' is specified.
 
 ### The {{ScrollTimeline}} Interface ### {#scrolltimeline-interface}
 
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -215,6 +215,9 @@ spec: cssom-view-1; type: dfn;
 	even if multiple elements use ''scroll()'' to refer
 	to the same [=scroll container=] with the same arguments.
 
+	Note: Progress in the horizontal or vertical axis still follows the logical scroll direction
+	which may flip depending on writing mode.
+
 ### The {{ScrollTimeline}} Interface ### {#scrolltimeline-interface}
 
 	<pre class="idl">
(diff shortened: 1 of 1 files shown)


ydaniv:
For me, in general, this follows expected behavior, so a general SGTM.

flackr:
@fantasai, does this seem reasonable to add?

fantasai (changes requested):
- s/may/can/ since we're not granting RFC2119 permission here. :)
- Move the note up to right after the DL, since it's a user-facing note clarifying the stuff in the DL and the following two paragraphs are rather less interesting to users.
- This brings up an interesting question about how we measure progress for `align-content: end`...

scroll-animations-1/Overview.bs
```diff
@@ -215,6 +215,9 @@ spec: cssom-view-1; type: dfn;
 	even if multiple elements use ''scroll()'' to refer
 	to the same [=scroll container=] with the same arguments.
 
+	Note: Progress in the horizontal or vertical axis still follows the logical scroll direction
+	which may flip depending on writing mode.
```
fantasai, line 219: Possible rewrite:

```suggestion
	Note: Progress is in reference to the [=scroll origin=],
	which can flip depending on [=writing mode=],
	even when ''scroll-timeline-axis/horizontal'' or ''scroll-timeline-axis/vertical'' is specified.
```

flackr:
Thanks for the review, I like your edit and moved the note just after the dl as suggested.

fantasai (approved):
Wfm, but I think you need to rebase over the latest spec. :/

Issue #8226, opened 2022-12-15
[scroll-animation-1] Define scroll linked animation behavior in printing documents


hiikezoe:
What most of users expect in printing documents is, I think, printing out the document as it is when the user printed out.

So basically any scroll linked animations should be output as it is, other then animations tied to the root scroll container in the top level document because there's not scroll position for the root scroller I think.

That's said, it's just my opinion. The spec needs to clarify it.

bramus:
When printing I would assume there is no scroll estate, so the logic from https://github.com/w3c/csswg-drafts/issues/7778#issuecomment-1341822948 would apply in that case.

hiikezoe:
@bramus what do you mean by "no scroll estate"? IIRC Chrome does preserve each scroll position for sub scroll containers in printing document.

hiikezoe:
I suppose you meant for the root scroll container cases?

bramus:
> I suppose you meant for the root scroll container cases?

Definitely for the root scroller as printing is paged media without a viewport.

For non-root scrollers I hadn’t really checked what printing does. In Chrome it prints the contents at the current scroll position like you mentioned. With that given, it makes sense to print the scroll-linked element at a certain position.

Maybe the progression of the animation could be snapshotted and then be re-applied when printing? E.g. say you’ve scrolled halfway and the scroll-linked animation is at 50% because of that. When printing, then use that 50% as the animation progress instead of trying to recalc the whole lot based on the printed box’s size.

flackr:
I think the most natural thing to do is treat the root as a scroller that is fully in view with no overscroll, and inner scrollers as their current position, which is the original position put forward by @hiikezoe and clarified by @bramus for what that means https://github.com/w3c/csswg-drafts/issues/8226#issuecomment-1353002054

flackr:
I believe the required changes for this issue are all landed with c6c859862d018ab9d5b998cf51dc75c933505c96.

Issue #7778, opened 2022-09-22
[scroll-animations-1] currentTime when scroll range is 0 (again)


bramus:
In https://github.com/w3c/csswg-drafts/issues/7401 we resolved on making the timeline inactive in the absence of a scroll range.

Playing around with the implementation in Chrome Canary _(with Experimental Web Platform Features flag)_, I’m starting to doubt this resolution.

- In https://codepen.io/bramus/pen/KKoGNye?editors=1010 it shows a progress bar at the top. As you scroll down, the progress bar grows from 0% width to 100%.
- https://codepen.io/bramus/pen/YzLxJZe/f0eefc95ba1d631d2dc081ea60ac9f0b is a fork from the demo above. Only difference is that it uses short content. As there is no scroll to happen, the progress bar remains at 0% even though you are at “the end of the content”.

From a UX perspective, one could argue if there’s a need to show a full progress bar in the second case or not. However, I can imagine other situations – e.g. where something fades in as you scroll down – that would need to be visible in case of no scroll range.

Maybe we should revisit?

bramus:
💡 Maybe `animation-fill-mode` could affect what needs to be done?

ydaniv:
Yes, actually that's correct!
That's currently what all libraries do, and it also corresponds better to time-based animations with duration of 0.

flackr:
Which end of the animation would you fill? An animation linked to a 0 scroll range is conceptually both before the beginning (first keyframe) and after the end (last keyframe).

bramus:
Last

flackr:
I don't agree with the expectation here. When a scroller shrinks from scrollable to not scrollable the animation will arbitrarily jump to the end even if it was previously at 0 progress (and is still at 0 scrollTop / scrollLeft), which to me seems surprising. I'm not sure how you can decide that one end of the scroll is the preferred state. You can set the style you want when there is 0 scroll range by setting the underlying style (not in the animation).

Making this configurable by animation-fill-mode is I think not feasible. The timeline current time is an input to the animation, and each associated animation can have a different fill mode. I suppose we could always have the timeline output 100% if there is no scroll, but I worry that this may be overindexing on your read progress indicator use case.

ydaniv:
> When a scroller shrinks from scrollable to not scrollable the animation will arbitrarily jump to the end even if it was previously at 0 progress (and is still at 0 scrollTop / scrollLeft), which to me seems surprising

[...]

PR #8673, opened 2023-04-03
[scroll-animations-1] Scroll timelines are inactive in paged media. #8226

fantasai:


scroll-animations-1/Overview.bs
```diff
@@ -210,6 +210,11 @@ spec: cssom-view-1; type: dfn;
 			then the [=scroll progress timeline=] is [=inactive timeline|inactive=].
 	</dl>
 
+	References to the [=root element=] propagate to the document viewport
+	(which functions as its [=scroll container=]).
+	However in [=paged media=],
+	the [=root element=]’s [=scroll progress timeline=] is [=inactive timeline|inactive=].
```
flackr, line 216: This implies that we have a single scroll progress timeline for anything that references the root (whether explicitly or nearest), but they do have different identities. How about something like:

However in [=paged media=],
[=scroll progress timelines=] referring to the document viewport are [=inactive timeline|inactive=]

scroll-animations-1/Overview.bs
```diff
@@ -498,6 +503,8 @@ spec: cssom-view-1; type: dfn;
 
 	If the 0% position and 100% position coincide
 	(i.e. the denominator in the [=current time=] formula is zero),
+	or if there is no corresponding [=scroll offset=]
```
flackr, line 506: Doesn't no corresponding scroll offset imply this happens when there's no scrollable overflow - which is not strictly true? I was thinking we could just say the [=view progress visibility range=] is the scrollable overflow size and let the rest fall out from the math. That said, I'm okay with making timelines looking at the root scroll on printed media blanket inactive I just don't want this to apply to other cases where we have well established behaviors as the overflow becomes entirely visible, e.g. demo in https://github.com/w3c/csswg-drafts/issues/7778#issuecomment-1310360733

fantasai: > Doesn't no corresponding scroll offset imply this happens when there's no scrollable overflow - which is not strictly true?

I was expecting that any scroll container has to have a scroll offset, even if it's always zero.

> I was thinking we could just say the [=view progress visibility range=] is the scrollable overflow size and let the rest fall out from the math.

It becomes a little bit vague what that means for paged media, especially if the content is longer than the root element itself.

fantasai:
@flackr OK, I posted two commits into the PR, with two possible options for wording. Lmk what you think, if you prefer the first commit or the second one. :)

flackr:
I think I prefer the second. I'm a little torn because the general concept of "paged media" doesn't have the same constraints as "print" but they seem to be a bit conflated. I.e. An ebook reader could be considered paged (I think) but *does* actually have a changing scroll progress and you may well want to use that to update a progress indicator with a scroll driven animation.

flackr:
> But I totally agree with you that it could make sense for a dynamic paged medium to have an active timeline. Should we open an issue on that? I think there's some other considerations, though, since a number of scroll animations are designed in a way that only makes sense on continuous media (see e.g. various NY Times articles), and if we default to applying those to dynamic paged media those devices could end up with a lot of unreadable pages. :/

I filed #8688 to discuss :-)

Issue #7759, opened 2022-09-17
[scroll-animations] Broader scope of scroll timelines 


fantasai:
From #7047, it might be worth looking into the ability to split the declaration of a timeline (together with its scoping) from its actual attachment to a scroll container.

This would allow authors, for example, to declare a name on a subtree and make it available to all descendants of that subtree, and attach it to a scroll container that is a descendant within the subtree. (At the top level, declaring the name on the root element would make it global.)

Maybe something like `scroll-timeline-attachment: local | defer | ancestor | closest` where:

- `local` has the current behavior of binding the name to this element’s scroll container
- `defer` declares and scopes the name, but does not bind it to a scroll container
- `ancestor` looks up the ancestor chain for a matching timeline name and attaches to that instance; failing a match, declares it locally
- `closest` looks back up the tree including previous siblings for a matching timeline name (same lookup as animation-timeline), and attaches to the first matching instance; failing a match, declares it locally

flackr:
Can we use this proposal to simplify the lookup of animation-timeline even further? I.e. if we can declare names on an ancestor and bind them later then we don't need the previous sibling lookup right? Just to make sure I understand this proposal fully is this something like:

```html
<style>
#parent {
  scroll-timeline-name: scroller;
  scroll-timeline-attachment: defer;
}
#scroller {
  scroll-timeline-name: scroller;
  scroll-timeline-attachment: ancestor;
}
#animated {
  animation-timeline: scroller;
}
</style>

<div id="parent">
  <div>
    <div id="animated"></div>
  </div>
  <div id="scroller"></div>
</div>
```

fantasai:
Agenda+ to discuss @flackr's suggestion to narrow the default scope to ancestors.

flackr:
@andruud any concerns with the proposal to instead allow pre-declaring a scroll/view timeline in an ancestor allowing the use of scrollers that have not been discovered yet? I imagine this may introduce some complexity to the implementation but it is certainly nice from an authoring POV, and would mean that we could eliminate the sibling scope visibility since authors could always move the declaration of the name to an ancestor element instead.

bramus:
In true staircase wit, thinking things over last night, I would like to reconsider – at least discuss things a bit more.

As it stands now, with `scroll-timeline-attachment`, an author needs to do three adjustments to hook onto a scroller that’s not a parent:

1. Find a common ancestor and duplicate the `scroll-timeline-name` there
[...]

PR #8680, opened 2023-04-04
[scroll-animations-1] Describe scroll/view-timeline-attachment

andruud:
Issue #7759.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -313,7 +313,7 @@ spec: cssom-view-1; type: dfn;
 	and then referenced by name.
 	See [[#timeline-scope]].
 
-	Named scroll progress timelines are declared in the [=coordinated value list=]
+	Named [=scroll progress timelines=] are declared in the [=coordinated value list=]
 	constructed from the [=longhands=] of the 'scroll-timeline' [=shorthand property=],
 	which form a [=coordinating list property group=]
 	with 'scroll-timeline-name' as the [=coordinating list base property=].
@@ -325,15 +325,20 @@ spec: cssom-view-1; type: dfn;
 	Name: scroll-timeline-name
 	Value: none | <<custom-ident>>#
 	Initial: none
-	Applies to: [=scroll containers=]
+	Applies to: all elements
 	Inherited: no
 	Computed value: the keyword ''scroll-timeline-name/none'' or a list of <<custom-ident>>s
 	Animation type: not animatable
 	</pre>
 
 	Specifies names for the [=scroll progress timelines=]
-	associated with this [=scroll container=].
-	The axis for this timeline is given by 'scroll-timeline-axis'.
+	declared by this element.
+
+	On boxes that are not [=scroll containers=],
+	only names with a 'scroll-timeline-attachment'
+	of ''scroll-timeline-attachment/defer''
+	are considered for [[#timeline-scope|scoping and lookup]];
+	any other name declarations are ignored.
 
 ### Axis of a Scroll Progress Timeline: the 'scroll-timeline-axis' property ### {#scroll-timeline-axis}
 
@@ -347,15 +352,13 @@ spec: cssom-view-1; type: dfn;
 	Animation type: not animatable
 	</pre>
 
-	Specifies an axis for each [=scroll progress timeline=]
-	associated with this [=scroll container=].
-	The name for this timeline is given by 'scroll-timeline-name'.
+	Specifies an axis for each named [=scroll progress timeline=]
+	sourced from this [=scroll container=].
 
 	Values are as defined for ''scroll()''.
 
-	Note: This property has no effect on [=scroll progress timelines=]
-	that are declared on this element
-	but whose [=attachment=] is ''scroll-timeline-attachment/defer''.
+	Note: This property has no effect when combined
+	with a 'scroll-timeline-attachment' of ''scroll-timeline-attachment/defer''.
 
 ### Attachment of a Scroll Progress Timeline: the 'scroll-timeline-attachment' property ### {#scroll-timeline-attachment}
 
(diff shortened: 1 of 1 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -353,20 +354,113 @@ spec: cssom-view-1; type: dfn;
...
+		}
+		</pre>
+	</div>
+
 ### Scroll Timeline Shorthand: the 'scroll-timeline' shorthand ### {#scroll-timeline-shorthand}
 
 	<pre class='propdef shorthand'>
 	Name: scroll-timeline
-	Value: [ <<'scroll-timeline-name'>> <<'scroll-timeline-axis'>>? ]#
+	Value: [ <<'scroll-timeline-name'>> [ <<'scroll-timeline-axis'>> | <<'scroll-timeline-attachment'>> ] ? ]#
```
flackr, line 448: This makes it impossible for the descendent to declare the axis unfortunately. In order to avoid an axis specified on the deferred attachment It would have to be split up into that value which I'm not sure is something we do. e.g.

```
Value: [ <<'scroll-timline-name'>> 'defer' | <<'scroll-timline-name'>> <<'scroll-timeline-axis'>>? 'ancestor'? ]#
```

I'm fine with just doing the simple thing and having them all in the shorthand if this isn't speccable.

bramus:
Looks good. Thank you, Anders.

scroll-animations-1/Overview.bs
andruud: Ah, yes, good point. Fixed (did the simple thing for now).

Although something like what you said does seem speccable. "Splitting" the value to achieve something special doesn't seem too far off from what we're doing in e.g. https://drafts.csswg.org/css-text-4/#white-space-property.

So maybe we could do:  `<<'scroll-timeline-name'>> [ defer | [ <<'scroll-timeline-axis'>> ancestor? ] ]?`

But maybe we leave this for a separate PR, for maximum bikeshedding.

scroll-animations-1/Overview.bs
```diff
@@ -353,18 +354,105 @@ spec: cssom-view-1; type: dfn;
...
+		}
+
+		#root #target {
+			animation: anim auto;
+			animation-timeline: mytimeline;
+		}
+
+		#root #target + #scroller {
+			scroll-timeline-name: mytimeline;
+		  scroll-timeline-attachment: ancestor;
```
flackr, line 439: nit: inconsistent indentation in this code sample.

flackr (approved):
Looks good to me

scroll-animations-1/Overview.bs
andruud: Thanks, done.

scroll-animations-1/Overview.bs
```diff
@@ -149,9 +149,10 @@ spec: cssom-view-1; type: dfn;
 	(<var>[=scrollable overflow=] size</var> &minus;
 		<var>[=scroll container=] size</var>)
 
-	If the 0% position and 100% position coincide
-	(i.e. the denominator in the [=timeline/current time=] formula is zero),
-	the timeline is [=inactive timeline|inactive=].
+	The timeline is [=inactive timeline|inactive=] when:
+	- The 0% position and 100% position coincide
+	(i.e. the denominator in the [=timeline/current time=] formula is zero), or
+	- The timeline has no [=current attachment=].
```
fantasai, line 155: This section (see section title) is about calculating progress, not about when the timeline is inactive.

scroll-animations-1/Overview.bs
```diff
@@ -695,18 +788,43 @@ spec: cssom-view-1; type: dfn;
...
-	Value: [ <<'view-timeline-name'>> <<'view-timeline-axis'>>? ]#
+	Value: [ <<'view-timeline-name'>> [ <<'view-timeline-axis'>> || <<'view-timeline-attachment'>> ]? ]#
 	Applies to: all elements
 	</pre>
 
 	This property is a [=shorthand=] for setting
-	'view-timeline-name' and 'view-timeline-axis'
-	in a single declaration.
-	It does not set 'view-timeline-inset'.
+	'view-timeline-name', 'view-timeline-axis' and 'view-timeline-attachment'
```
fantasai, line 826: Oxford comma

fantasai (changes requested):
@andruud I know I'm asking for a bit of a conceptual rewrite... if you prefer for me to draft it up for you, just lmk, and I'll add it as a commit on this branch. (I was planning to work on this today anyway.)

scroll-animations-1/Overview.bs
```diff
@@ -353,18 +354,105 @@ spec: cssom-view-1; type: dfn;
...
+	Computed value: a list of the keywords specified
+	Animation type: not animatable
+	</pre>
+
+	Specifies attachment behavior for each [=scroll progress timeline=]
+	associated with this element.
+
+	A scroll timeline has a set of <dfn>attachments</dfn>,
+	that determine which [=scroll container=] and {{ScrollTimeline/axis}} to use
+	when calculating the [progress of the timeline](#scroll-timeline-progress).
```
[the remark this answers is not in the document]
andruud: Right, it seemed logical to me that `scroll-timeline-*` would actually define and create scroll timelines (regardless of their nature), and then `scroll-timeline-attachment` would describe how these timelines attach together (it's what it sounds like, isn't it?).

Let me give you some minimal background about this solution, _not_ to "prove" that "scope the name" isn't a viable way to spec (maybe it is), but to help you write the best possible spec. :-)

Suppose that we had elements A, B and C that appear in that order in the tree.

```
  A {
    scroll-timeline: mytimeline defer;
  }
  A B {
    animation: anim auto;
    animation-timeline: mytimeline;
  }
  A B + C {
    scroll-timeline: mytimeline ancestor;
  }
```

 - First we resolve the style for A. In "scope the name", this yields no timeline.
 - Then we resolve the style for B. Doing so creates CSS animations that have no timeline, since we can't find "mytimeline".
 - Then we resolve the style for C. This creates "mytimeline", which is now reachable via A.

But animations from B already decided that they have no timeline, and have no memory or knowledge that they should _now_ be connected to some timeline after all. At least, they don't without additional spec concepts, and explicitly deferred timelines was an attempt at specifying such a concept!

My support for this feature was predicated on attachment hackery being an [aspect of the timeline (comment)](https://github.com/w3c/csswg-drafts/issues/7759#issuecomment-1413582984), and I still think this is the only realistically possible way to do it. In particular, animations should be allowed to remain agnostic of said hackery without having its timeline replaced in some awkward "fixup" step.

andruud:
> if you prefer for me to draft it up for you, just lmk, and I'll add it as a commit on this branch

@fantasai If you are dead set on "scope a name", then yes please, as I'm not sure how to actually spec that. No point in me trying to guess what you would have written. :-)

See other comment, though.

scroll-animations-1/Overview.bs
andruud: Done.

scroll-animations-1/Overview.bs
```diff
@@ -496,9 +584,10 @@ spec: cssom-view-1; type: dfn;
 		the [=scroll offset=] corresponding to
 		the end of the ''animation-timeline-range/cover'' range
 
-	If the 0% position and 100% position coincide
-	(i.e. the denominator in the [=timeline/current time=] formula is zero),
-	the timeline is [=inactive timeline|inactive=].
+	The timeline is [=inactive timeline|inactive=] when:
+	- The 0% position and 100% position coincide
+	(i.e. the denominator in the [=timeline/current time=] formula is zero), or
+	- The timeline has no [=current attachment=].
```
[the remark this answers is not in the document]
andruud: Done.

scroll-animations-1/Overview.bs
andruud: Done.

scroll-animations-1/Overview.bs
```diff
@@ -695,18 +788,43 @@ spec: cssom-view-1; type: dfn;
...
+	Value: [ <<'view-timeline-name'>> [ <<'view-timeline-axis'>> || <<'view-timeline-attachment'>> ]? ]#
 	Applies to: all elements
 	</pre>
 
 	This property is a [=shorthand=] for setting
-	'view-timeline-name' and 'view-timeline-axis'
-	in a single declaration.
-	It does not set 'view-timeline-inset'.
+	'view-timeline-name', 'view-timeline-axis' and 'view-timeline-attachment'
+	in a single declaration. It does not set 'view-timeline-inset'.
```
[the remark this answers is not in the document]
andruud: Done.

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
andruud: > Also, multiple attachments should not be a fundamental data structure in CSS, they're an error condition. You can implement error handling data structures however you want internally, but let's not write the spec as if they're fundamentally how this works.

Done, FWIW.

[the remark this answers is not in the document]
andruud: I'm stating the obvious maybe, but having deferred names affect timeline lookups (i.e. affect what animation-timeline sees) creates a circular dependency within the same frame.

```
<A>
  <B>
  </B>
</A>
```

```
@keyframes anim {
  from { --x:0; }
  to { --x:0; }
}

A {
  --x:1;
  scroll-timeline: timeline defer;
  animation: anim auto;
  animation-timeline: timeline;
}

@container style(--x:1) {
  B {
    scroll-timeline: timeline ancestor;
  }
}
```

If the computed value of `--x` on `A` is `1`, then timeline on `B` should have been reachable via the deferred name on `A`, and `--x` on `A` must be `0`.

(This is a problem with or without style queries, but style queries makes it more obvious).

I said earlier that maybe there is a way to make this work as "scopes as a name", but I'm increasingly convinced that it's not.

scroll-animations-1/Overview.bs
```diff
@@ -662,6 +743,10 @@ spec: cssom-view-1; type: dfn;
 	Specifies an axis for each named [=view progress timeline=]
 	associated with this [=scroll container=].
 
+	The [=used value=] of 'view-timeline-axis'
+	is the axis of the timeline's [=attachment=],
+	or the initial value if there is no [=attachment=].
```
fantasai, line 748: Same as above.

scroll-animations-1/Overview.bs
```diff
@@ -353,18 +353,99 @@ spec: cssom-view-1; type: dfn;
...
+		  to { color: green; }
+		}
+
+		#root {
+		  scroll-timeline-name: mytimeline;
+		  scroll-timeline-attachment: defer;
+		}
+
+		#root #target {
+		  animation: anim auto;
```
fantasai, line 426:
```suggestion
		  animation: anim;
```

We're still discussing whether `auto` is a valid keyword inside `animation`, but it's definitely the initial value for `animation-duration` so leaving it off will work.

scroll-animations-1/Overview.bs
```diff
@@ -353,18 +353,99 @@ spec: cssom-view-1; type: dfn;
...
+
+		<pre class=lang-css>
+		@keyframes anim {
+		  from { color: red; }
+		  to { color: green; }
+		}
+
+		#root {
+		  scroll-timeline-name: mytimeline;
+		  scroll-timeline-attachment: defer;
```
fantasai, line 422: Do we want to use the shorthand syntax in the examples? Since that is usually the best practice.

fantasai (changes requested):
Some review comments on what you've got so far.

fantasai:
Pushed a commit with some minor clean-ups. (I'll try to figure out filing a PR against a PR for the name-scoping bit, though, so you can review it. :)

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
fantasai: I'm not sure how changing how we describe it affects this case? In your text, the timeline is inactive until it's bound; doesn't that mean the keyframes aren't applied?

fantasai:
@andruud I opened https://github.com/andruud/csswg-drafts/pull/1 to recast this in terms of a name binding. There are some additional wording fixes there, so if we end up going with this base version I'll need to port them over; but I didn't want to force-rebase your branch over a new main, so... it's a bit messy on that point. :/

fantasai:
CC @tabatkins for opinions...

scroll-animations-1/Overview.bs
andruud: Yeah, makes sense. Not sure why I used longhands.

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
andruud: In your PR, deferred names _don't_ affect timeline lookups, so it's fine. I didn't know you were just going to change how we describe it. :-)

> In your text, the timeline is inactive until it's bound; doesn't that mean the keyframes aren't applied?

In the first frame, the timeline is inactive because it doesn't have a state yet. This state is updated at a specific point in the frame (before the style update), and we're already beyond that point for the _current_ frame. When we resolve the style for B in this frame, timelines attach to each other, but it has no immediate affect on any timeline state. The timeline on A is still inactive.

In the _next_ frame, the state of all scroll timelines are updated. In this moment (only in this moment), the attachments matter, because we use the attached timeline to produce the state. In the subsequent style resolution, the animation applies this time, and we get `--x:0`. This removes the timeline on B, leaving the deferred timeline with no attachments. This (again) does not affect the state of the timeline this frame. The computed value of `--x` is still `0`.

In the next frame after that, the timeline once again becomes inactive, because it has no attachments at the time of the state update. So we actually switch between the two states every frame. But there is no circularity _within the same frame_, as a lookup-affecting version of "scopes a name" would have caused.

I'm not sure if that answers the question ...

(Note that I'm ignoring our handling of [initially stale](https://drafts.csswg.org/scroll-animations-1/#initially-stale) timelines in the above, so what would _really_ happen would require an even longer explanation).

andruud:
Thanks @fantasai, that looks good to me if we can eliminate the circularity there. :-)

scroll-animations-1/Overview.bs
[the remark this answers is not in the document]
fantasai: > In your PR, deferred names don't affect timeline lookups, so it's fine.

Yeah, because you said that would be bad. :) My first instinct was to have them be invisible, because in CSS we usually ignore things that are invalid (treat them as if they didn't exist)...

> I'm not sure if that answers the question ...

It does. What a terrible scenario.

fantasai:
@andruud I committed a bunch of editorial improvements to try to address @tabatkins's review that it wasn't very understandable. :) Hoping this is good to land now (but we should DEFINITELY squash the history into a single co-authored commit).

andruud:
@fantasai This is very clean now to my eyes. (But of course I'm tainted by already knowing how it should work :-)).

> DEFINITELY squash

For sure. :-)

fantasai:
@andruud OK, merged. Thanks for working on this with me!

PR #8704, opened 2023-04-11
[scroll-animations-1] Also rerun style/layout for changed timeline ranges

### LINKED ISSUE #8694: [scroll-animations-1] Handling changed size/position in current frame
In issue #5261 we defined a mechanism for ensuring that scroll and view timelines that were discovered during style and layout were given an opportunity to produce a current time for the frame in which they were first rendered, however with [view timeline ranges](https://www.w3.org/TR/scroll-animations-1/#view-timelines-ranges) being usable from both `animation-range` and from keyframes, the particular progress of an animation depends on layout.

For example, consider the following ([demo](https://jsbin.com/pudexek/edit?html,css,output)):
```html
<style>
.subject {
  view-timeline: subject;
}

.shifted {
  top: 120vh;
}
.grown {
  height: 300px;
}

@keyframes slide {
  entry 0% {transform: translateY(100vh);}
  exit 0% {transform: translateY(0);}
}

#animated {
  animation: slide linear;
  animation-timeline: subject;
}
</style>
<div class="subject"></div>
<div id="animated"></div>  
<button class="shift">Shift</button>
<button class="grow">Grow</button>
<script>
document.querySelector('.shift').addEventListener('click', () => {
  document.querySelector('.subject').classList.toggle('shifted');
});

document.querySelector('.grow').addEventListener('click', () => {
  document.querySelector('.subject').classList.toggle('grown');
});
</script>
```

When the `shifted` or `grown` classes are added (e.g. by clicking the buttons) we need to update the offset of the `exit 0%` keyframe as well as invalidate the animation and calculate the new position for the `#animated` box.

We have two options:
1. Allow animation to be out of date on the frame in which geometry changes occur. Schedule another frame and allow for some flicker.
2. Detect changes in the geometry and allow rerunning style and layout. We could do this in a similar way to how we re-run the lifecycle for newly detected animations (see #5261).

I think that unless an animation caused the animation to change I would not expect a flicker when a style change results in the subject moving. As such, I suggest pursuing option 2. More concretely, after running resize observers (to take into account developer derived sizes), we would need to check if the animation constraints have changed (in addition to checking for new scroll timelines as we already do) and if so rerun style and layout to incorporate the updated animations. This is not ideal for performance but hopefully will not happen often.

flackr:
Fixes #8694.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -835,12 +835,20 @@ spec: cssom-view-1; type: dfn;
 	as step 1 of [=update animations and send events=].
 
 	During step 7.14.1 of the [HTML Processing Model](https://html.spec.whatwg.org/multipage/webappapis.html#processing-model-8),
-	any created [=scroll progress timelines=] or [=view progress timelines=] are collected into an <dfn export>initially stale</dfn> set.
-	After step 7.14.1 if there are any [=initially stale=] timelines, they update their current time,
-	the set of [=initially stale=] timelines is cleared and
+	any created [=scroll progress timelines=] or [=view progress timelines=] are collected into a <dfn export>stale timelines</dfn> set.
+	After step 7.14 if any timelines' [=named timeline ranges=] have changed,
+	these timelines are added to the [=stale timelines=] set.
+	If there are any [=stale timelines=], they now update their current time and associated ranges,
+	the set of [=stale timelines=] is cleared and
 	we run and we run an additional step to recalculate styles and update layout.
-	As we only gather stale timelines during the first style and layout calculation,
+
+	Note: We check for layout changes after dispatching ResizeObservers
+	intentionally to take into account programatically sized elements.
+
+	Note: As we only gather stale timelines during the first style and layout calculation,
 	this can only directly cause one additional style recalculation.
+	Other APIs which require another update should be checked in the same step
+	and be updated at the same time.
 
 	Note: Without this additional round of style and layout,
 	[=initially stale=] timelines would remain stale
@@ -856,7 +864,7 @@ spec: cssom-view-1; type: dfn;
 	In other words, [=initially stale=] timelines are visible as such
 	through those APIs.
 
-	If the step 7.14 of the [HTML Processing Model](https://html.spec.whatwg.org/multipage/webappapis.html#processing-model-8)
+	If the final style and layout update
 	would result in a change of time for any [=scroll progress timelines=] or
 	[=view progress timelines=], they will not be re-sampled to reflect the
 	new offset until the next update of the rendering.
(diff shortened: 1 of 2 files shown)


scroll-animations-1/Overview.bs
```diff
@@ -835,12 +835,20 @@ spec: cssom-view-1; type: dfn;
...
+	these timelines are added to the [=stale timelines=] set.
+	If there are any [=stale timelines=], they now update their current time and associated ranges,
+	the set of [=stale timelines=] is cleared and
 	we run and we run an additional step to recalculate styles and update layout.
-	As we only gather stale timelines during the first style and layout calculation,
+
+	Note: We check for layout changes after dispatching ResizeObservers
+	intentionally to take into account programatically sized elements.
+
+	Note: As we only gather stale timelines during the first style and layout calculation,
```
kevers-google, line 848: Also worth pointing out that the reason for only gathering stale timelines during the first style+layout cycle is to avoid an infinite style update loop. It is still possible for the style to be stale at the end of the second cycle, if the animation is changing the height or width of the source or subject of a scroll-driven animation.  This can lead to animation jitter.  It is safe however, for the animation to change the size via a transform, since using non-transformed values for determination of the animation's progress.

flackr: Isn't this infinite loop concern covered by the first paragraph? And the paragraph on line 867 points out that they will be stale until the next frame.

kevers-google: ack

scroll-animations-1/Overview.bs
```diff
@@ -835,12 +835,20 @@ spec: cssom-view-1; type: dfn;
...
-	the set of [=initially stale=] timelines is cleared and
+	any created [=scroll progress timelines=] or [=view progress timelines=] are collected into a <dfn export>stale timelines</dfn> set.
+	After step 7.14 if any timelines' [=named timeline ranges=] have changed,
+	these timelines are added to the [=stale timelines=] set.
+	If there are any [=stale timelines=], they now update their current time and associated ranges,
+	the set of [=stale timelines=] is cleared and
 	we run and we run an additional step to recalculate styles and update layout.
-	As we only gather stale timelines during the first style and layout calculation,
+
+	Note: We check for layout changes after dispatching any {{ResizeObserver}}s intentionally
```
andruud, line 845: ... but this means that ResizeObserver sees an irrelevant layout state vs. what we actually intend to render?

flackr: That's correct. It's a tradeoff I think. Either:
* Run before ResizeObserver, then any changes to sizes of content made by ResizeObserver may not be handled by scroll driven animations, but ResizeObserver will see layout driven animation results.
* Run after ResizeObserver, then layout animations may not be fully up to date when calling ResizeObserver, but view timelines will be able to observe resizeobserver derived sizes.

Given that ResizeObservers are designed for establishing the size of elements, I feel like view timelines should be updated to reflect the final resize observer output. The converse, animating layout firing ResizeObservers in response to scroll seems like an anti-pattern.

flackr: We're going to want to use the same mechanism and invalidation for anchor positioning. Chatting with @bfgeek we thought after ResizeObserver would be more in line with the use cases.

Issue #8487, opened 2023-02-22
[css-animations-2][scroll-animations-1] Avoid merging named timeline range keyframes which compute to the same offset


flackr:
In [css-animations-2 3. Keyframes step 6.4](https://drafts.csswg.org/css-animations-2/#keyframes), two keyframes with the same keyframe offset, timing function and composite will be merged into the same keyframe. We may need to call out that keyframes which compute to the same offset but have different [named timeline range offsets](https://drafts.csswg.org/scroll-animations-1/#named-range-keyframes) should not be merged just as we don't merge keyframes with different easing functions or composite values.

Merging such cases would end up dynamically merging different groups of keyframes depending on the specific layout which feels unnecessarily expensive to keep up to date and odd from a developer perspective that the keyframe count would vary dynamically based on layout.

PR #8705, opened 2023-04-11
Rewrite Keyframe Processing

### LINKED ISSUE #8507: [css-animations-2][scroll-animations-1] Keyframe ordering and ordering requirements
# Overview

We need to define the order of keyframes from CSS animations and any requirements when constructing JS animations. Prior to [timeline range keyframe selectors](https://www.w3.org/TR/scroll-animations-1/#named-range-keyframes) we have the following rules:

[css-animations-2 Section 3. Keyframes](https://drafts.csswg.org/css-animations-2/#keyframes) performs a stable sort by offset (step 6), and merges keyframes with matching keyframe offset, timing function and composite (step 6.4). The result is exposed via the [web-animations-1 getKeyframes](https://drafts.csswg.org/web-animations-1/) method.

[web-animations-1 Section 6.6.3. Processing a keyframes argument](https://drafts.csswg.org/web-animations-1/#processing-a-keyframes-argument) procedure to [process a keyframes argument](https://drafts.csswg.org/web-animations-1/#process-a-keyframes-argument) step 6 throws a TypeError if the supplied offsets of the keyframes are not sorted.

With [timeline range keyframe selectors](https://www.w3.org/TR/scroll-animations-1/#named-range-keyframes) the order of keyframes is not known until after layout and may change with changes to the layout. As such, we have to decide how to handle this.

I wrote a short demo to show the observed keyframes at https://jsbin.com/zofukin/edit?html,css,js,output

# Proposal

For css animations:
1. just as differences in composite and the timing function do not allow two keyframe blocks to be merged, we should consider named range offsets to only be mergeable with the exact same specified named range offset.
2. Keyframes should continue to be sorted by a stable sort, though it will need to be a sort of their computed offsets.
3. Where do we sort keyframes whose offsets are not computable? They could be kept in their stable position ignoring offset or they could be effectively removed not appearing in getKeyframes. Naively, I'd propose keeping them in their stable sorted position and visible in the API.

For web animations, we never modify the passed in keyframes. They are returned in the order specified with specified offsets. We can't enforce the ordering requirement as strictly. We could do one of the following:
- Only require ordering of keyframes with overall percentage offsets, allow other keyframes in any order.
- Require ordering of keyframes within each named offset range. E.g. two keyframes on the `cover` range must be ordered with respect to each other. Though this gets more complicated when you consider that offsets will support calculations in the future.

I lean towards the first since calculations will also break the ability to ensure that order is correct when specified.

fantasai:
This aims to fix #8507 and #8487 as well as make the section more correct overall and easier to read, understand, and reference.

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -247,133 +247,102 @@ from its <a>owning element</a> always causes it to enter (or remain) in the
 <a>idle play state</a>.
 
 <h2 id="keyframes">
-Declaring Keyframes</h2>
+Assembling Keyframes</h2>
+
+<h3 id="keyframe-rules">
+Declaring Keyframes: the ''@keyframes'' rule</h3>
+
+See [[css-animations-1#keyframes]].
 
 <h3 id="keyframe-processing">
 Processing Keyframes</h3>
 
-For a given target (pseudo-)element, |element|, an animation name, |name|,
-and the position of the animation in |element|'s 'animation-name' list,
-|position|,
-[=keyframe=] objects are generated as follows:
+For each animation effect defined by the <var>N</var>th item
+in the [=coordinated value list=] of the 'animation-*' properties
+on target (pseudo-)element |element|,
+its associated [=keyframes=] are generated as follows:
 
-1.   Let |default timing function| be
-     the timing function at position |position|
-     of the [=resolved value=] of the 'animation-timing-function' for |element|,
-     repeating the list as necessary as described in
-     [[CSS-ANIMATIONS-1#animation-name]].
+1.  <strong>Set Defaults:</strong>
 
-1.   Let |default composite| be ''replace''.
+    *   Let |default timing function| be the corresponding [=computed value=]
+        of 'animation-timing-function' on |element|.
 
-1.   Find the last ''@keyframes'' at-rule in document order
-     with <<keyframes-name>> matching |name|.
+    *   Let |default composite| be ''replace''.
 
-     If there is no ''@keyframes'' at-rule
-     with <<keyframes-name>> matching |name|,
-     abort this procedure.
-     In this case no animation is generated,
-     and any existing animation matching |name| is canceled.
+    *   Let |keyframes| be an empty sequence of [=keyframe=] objects,
+        each possessing a
+        |keyframe offset|,
+        |keyframe timing function|,
+        |keyframe composite|,
+        and |keyframe values|.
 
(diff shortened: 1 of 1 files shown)


css-animations-2/Overview.bs
```diff
@@ -170,7 +170,7 @@ is <code>elem</code>. <code>animation</code> is disassociated from
 <pre class="example lang-javascript">
 elem.style.animation = 'spin 1s';
 let animation = elem.getAnimations()[0]; // animation's owning element is elem
-elem.style.animation = ''; // animation no longer has an owning element
+elem.style.animation = ''; // animation no longer has an owning element <!-- '' (Bikeshed syntax highlighting fix)-->
```
flackr, line 173: Bikeshed syntax highlighting fix? Is this a bikeshed cheat code or a reminder?

css-animations-2/Overview.bs
```diff
@@ -247,133 +247,133 @@ from its <a>owning element</a> always causes it to enter (or remain) in the
...
+in the [=coordinated value list=] of the 'animation-*' properties
+on target (pseudo-)element |element|,
+its associated [=keyframes=] are generated as follows:
+
+1.  <strong>Set Defaults:</strong>
+
+    *   Let |default timing function| be the corresponding value
+        of 'animation-timing-function'.
+
+    *   Let |default composite| be ''replace''.
```
flackr, line 270: This should come from the `animation-composition` property.

flackr (commented):
Overall looks good, just a few questions / nits.

css-animations-2/Overview.bs
fantasai: Oh, I should remove that! My syntax highlighter was treating the whole file as being part of a string starting on that line...

css-animations-2/Overview.bs
fantasai: Done. (Added a new commit for that, since this was the original state.)

css-animations-2/patch
```diff
@@ -0,0 +1,33 @@
...
++          to the front of the list,
++          and perform a stable sort on these
++          by the offset specified in the keyframe selector.
++     1.   Leave all other keyframe blocks in the order specified.
++
++1.   Iterate over the keyframe blocks in reverse
++     from the order determined in the previous step,
+      applying the following steps:
+ 
+      1.   Let |keyframe offset| be the value of the keyframe selector
```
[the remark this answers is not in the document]
fantasai: Yes, probably while I was trying to rewrite history. >_<;; Fixed.

PR #8711, opened 2023-04-13
[cssom-1] Merge CSSImportRule layer extensions from cascade-5

CanadaHonk:
Merged the CSSImportRule extension for `layerName` into cssom-1 from cascade-5, as recommended by @emilio.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2045,6 +2045,7 @@ interface CSSImportRule : CSSRule {
   readonly attribute USVString href;
   [SameObject, PutForwards=mediaText] readonly attribute MediaList media;
   [SameObject] readonly attribute CSSStyleSheet styleSheet;
+  readonly attribute CSSOMString? layerName;
 };
 </pre>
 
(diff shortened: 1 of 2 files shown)


cssom-1/Overview.bs
```diff
@@ -2059,6 +2060,10 @@ attribute of the associated <a>CSS style sheet</a>.
 
 The <dfn attribute for=CSSImportRule>styleSheet</dfn> attribute must return the associated <a>CSS style sheet</a>.
 
+The <dfn attribute for=CSSImportRule>layerName</dfn> attribute may return the layer name declared in the at-rule itself,
```
emilio, line 2063: Let's keep the `[=layer name=]` link. Also, not `may`, should be a `must`.

emilio (commented):
Thanks! I think this is good. @mirisuzanne @tabatkins any objections to this?

cssom-1/Overview.bs
CanadaHonk: Done, thanks!

emilio:
@svgeesus same question about the ipr bot :)

Issue #6429, opened 2021-07-07
[css-display] Why is display listed as not animatable instead of animation type: discrete?


LeaVerou:
https://www.w3.org/TR/css-display-3/#the-display-properties

I thought we were going for discrete for all properties that can't be intelligently interpolated, and that "not animatable" was reserved primarily for animation properties and the like?

Especially about `display`, there are plenty of use cases where authors want to finish an animation with `display: none`, and this prevents them from doing so.

flackr:
`display` does affect animations in the same way as animation properties. Specifically from https://www.w3.org/TR/css-animations-1/#animations

> Setting the display property to none will terminate any running animation applied to the element and its descendants. If an element has a display of none, updating display to a value other than none will start all animations applied to the element by the animation-name property, as well as all animations applied to descendants with display other than none.

This means that if a CSS animation animated to `display: none` it would immediately cancel the animation, losing the `display: none` which would restart the animation.

LeaVerou:
Ah, bummer. "Not animatable" makes total sense then.

Loirooriol:
Just for reference, at some point `display` was animatable on Firefox, but that could freeze the browser. https://bugzil.la/1264396

flackr:
I wonder if we could support only specific values of display, e.g. anything but `none` on css animations, and possibly anything on web animations since their lifetime is not controlled by the display property. Then you could create an animation which maintains `display: block` until it finishes or use a web animation.

tabatkins:
Having only some values animateable would be a new concept, but yes, that would work and probably be useful.

flackr:
Reopening this to consider allowing animations of display to prevent the element from immediately computing to `display: none`.

In particular, my straw-man proposal is the following:
1. Animation keyframes may specify display values which are not none. `display: none` is dropped from keyframes.
2. Display is interpolated like visibility, i.e. it will prefer the non-none value.

With these two points, we shouldn't have any circularity. An animation cannot itself produce `display: none` since it can't specify it (rule 1). Transitions to `display: none` would also work since the `display: none` wouldn't apply until after the transition finished (rule 2).

With this, developers could do things like:
```css
.hide {
  transition: opacity 200ms, display 200ms;
  display: none;
  opacity: 0;
}
```

Or:
```css
@keyframes slideaway {
  to { transform: translateY(40px); opacity: 0; display: block; }
}

.hide {
  animation: slideaway 200ms;
  display: none;
}
```

[...]

PR #8713, opened 2023-04-13
[css-animations-1] Animations are only canceled if base and computed display are none.

flackr:
See #6429

### DIFF
--- a/css-animations-1/Overview.bs
+++ b/css-animations-1/Overview.bs
@@ -181,11 +181,17 @@ CSS Animations Model</h2>
 
 	</div>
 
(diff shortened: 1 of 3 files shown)


flackr:
This is a rough attempt at specifying the proposed behavior in #6429. Perhaps it should be part of the deltas in css-animations-2?

fantasai:
Yeah, I think it should go into css-animations-2.

flackr:
> Yeah, I think it should go into css-animations-2. `display` isn't animatable in L1.

Done. This also fits better as animations are canceled / started by the display of their owning element - a concept which is more fleshed out in css-animations-2.

fantasai (approved):

PR #8737, opened 2023-04-19
[css-contain-3] Make content-visibility animatable.

### LINKED ISSUE #8627: [css-contain-4] Make content-visibility animatable
[css-contain Section 4](https://w3c.github.io/csswg-drafts/css-contain/#content-visibility) currently defines the content visibility property to be Animation type: not animatable. Most likely this was done for simplicity since like display it can affect which animations are running. However, as with display where we are better refining the ways in which it is safe to animate (See #6429), we should figure out in which ways content-visibility can be animated.

In particular, just like display it should be possible for an animation to extend the application of `content-visibility: visible` in order to allow content to stay visible while an animation is completing. This would be useful for elements like `<details>` whose contents are `content-visibility: hidden` when closed.

flackr:
Allow animating content-visibility as animation type discrete with special interpolation to avoid hidden values similar to visibility. Fixes #8627

### DIFF
--- a/css-contain-3/Overview.bs
+++ b/css-contain-3/Overview.bs
@@ -1119,6 +1119,8 @@ Changes since the 18 August 2022 Working Draft</h3>
 	* Correct typo in 'container-type' syntax, to clarify that ''normal'' cannot
 		be combined with other values.
 		(<a href="https://github.com/w3c/csswg-drafts/issues/7669">Issue 7669</a>)
+	* Make the 'content-visibility' property animatable.
+		(<a href="https://github.com/w3c/csswg-drafts/issues/8627">Issue 8627</a>)
 
 <h3 id="changes-2021-12">
 Changes since the 21 December 2021 First Public Working Draft</h3>


mirisuzanne:
This generally makes sense to me. The linked resolution had a third aspect, and it's not clear to me if it needs to be captured here: 

> Follow the same pattern as display for ensuring that css animations cannot deactivate themselves. Currently following that pattern would mean treating auto and hidden as revert.

flackr:
We're shifting away from that requirement for display as fewer special cases is simpler. See https://github.com/w3c/csswg-drafts/issues/6429#issuecomment-1503849651

Also I realized after making the recommendation for content-visibility that the same logic does not apply. `content-visibility` affects the visibility of its content, but the element itself remains visible (runs animations) even if set to `hidden`.

mirisuzanne (approved):

flackr:
> Want to add a line in the changelog about this?

Thanks for the reminder, done.

PR #8739, opened 2023-04-19
[css-syntax-3] small wording change

### LINKED ISSUE #6978: Difference between css stylesheet and document
I see both being used in the spec but I can't tell the difference

TKDev7:
The term document usually refers to markup-based resources. I think stylesheet fits better here. What do you think @tabatkins ?

Fixes #6978


tabatkins:
I don't think the distinction you're trying to draw actually exists. Calling a stylesheet a document seems fine to me.

PR #8740, opened 2023-04-20
[css-display-3] small editorial changes

TKDev7:
I believe CSS works with document trees and not the source document itself. I am not sure if I am being correct here or not.

### DIFF
--- a/css-display-3/Overview.bs
+++ b/css-display-3/Overview.bs
@@ -75,7 +75,7 @@ Introduction</h2>
 
 	<p><em>This section is normative.</em>
 
-	CSS takes a <a>document tree</a> organized as
+	CSS takes a source document organized as
 	a <dfn id="element-tree" oldids="css-element-tree" export for=CSS lt="element tree">tree</dfn>
 	of <dfn id="elements" oldids="css-element" for=CSS lt="element">elements</dfn>
 	(which can contain a mix of other [=CSS/elements=] and [=text nodes=])


css-display-3/Overview.bs
```diff
@@ -75,15 +75,15 @@ Introduction</h2>
 
 	<p><em>This section is normative.</em>
 
-	CSS takes a source document organized as
+	CSS takes a <a>document tree</a> organized as
```
fantasai, line 78: This change makes the sentence somewhat redundant? We _are_ conceptually working with a source document, even if the source document has its own processing model and processed into an _element tree_ by the time CSS looks at it. And that's what this sentence is trying to explain.

So I would not make this change.

css-display-3/Overview.bs
```diff
@@ -75,15 +75,15 @@ Introduction</h2>
...
 	a <dfn id="element-tree" oldids="css-element-tree" export for=CSS lt="element tree">tree</dfn>
 	of <dfn id="elements" oldids="css-element" for=CSS lt="element">elements</dfn>
 	(which can contain a mix of other [=CSS/elements=] and [=text nodes=])
 	and <dfn id="text-nodes" oldids="css-text-node" export for=CSS lt="text node">text nodes</dfn>
 	(which can contain text),
 	and renders it onto a <a href="https://www.w3.org/TR/CSS2/intro.html#canvas">canvas</a>
 	such as your screen, a piece of paper, or an audio stream.
-	Although any such source document can be rendered with CSS,
+	Although any such document tree can be rendered with CSS,
```
fantasai, line 86: We could change this to “element tree”, which is defined in the previous sentence. But we can't change this to “document tree” which is referring to the DOM specifically.

fantasai (changes requested):

css-display-3/Overview.bs
fantasai: Looks like this all needs a bit of clean-up... the term “element tree” didn't have a definition before, so there wasn't anything to anchor the terminology before. :/

> I don’t think DOMs are element trees since they contain comment nodes. i believe they are more of document trees.

CSS operates on element trees, fundamentally, and wrt the DOM, it essentially operates on an element-tree projection of the DOM (ignoring comment nodes, etc.). That's the concept that this paragraph is defining.

This sentence specifically is about how you can apply CSS to an element tree that *isn't* derived from the DOM, so we definitely shouldn't be using the term “document tree” here.

Issue #3103, opened 2018-09-13
[cssom-1] Indicating when a style sheet is script-blocking is duplicated several times


domfarolino:
In the HTML Standard's [Interactions of styling and scripting](https://html.spec.whatwg.org/multipage/semantics.html#interactions-of-styling-and-scripting) section we have the definition of the "style sheet ready" flag for CSS style sheets. We also have a list of conditions which, if met, make a style sheet a [script-blocking](https://html.spec.whatwg.org/multipage/semantics.html#a-style-sheet-that-is-blocking-scripts) one. Not too long ago this section was reformatted, and another condition has [been added](https://github.com/whatwg/html/pull/2984).

The "style sheet ready" flag is seemingly never used or set, and we're going to change (and maybe rename?) this to be a flag that makes a style sheet that was previously script-blocking, no longer script-blocking. When searching around for usages of the flag, I noticed that some text at the bottom of both the [xml-stylesheet process instruction](https://drafts.csswg.org/cssom/#requirements-on-user-agents-implementing-the-xml-stylesheet-processing-instruction) and [`Link` header requirements](https://drafts.csswg.org/cssom/#requirements-on-user-agents-implementing-the-http-link-header) sections mention the conditions that make a style sheet a script-blocking one. This seems like normative information that is duplicated from the HTML Standard, and can easily get out of sync (like with the adding of more conditions as we've seen).

Would the editors accept a PR from me that replaces these paragraphs with something like a "Note" linking to the appropriate section in the HTML Standard so this logic might stay in one place?

/cc @emilio @annevk

annevk:
I think we'd need to make changes on both sides if we want to reuse HTML's definition. E.g., the xml-stylesheet processing instruction is not an element whereas HTML only talks about elements.

I do agree it makes sense though, especially as they are already out-of-sync.

emilio:
> Would the editors accept a PR from me that replaces these paragraphs with something like a "Note" linking to the appropriate section in the HTML Standard so this logic might stay in one place?

Please do! Otherwise let me know if you don't have the cycles and I'll do the edits myself. Thanks!

domfarolino:
@annevk I think I understand the changes you refer to on the HTML side, but could you clarify what changes would need made elsewhere? Are you saying we should add text in the CSSOM spec and wherever xml-stylesheet processing is that says, "In the context of a web page, these style sheets can be script-blocking [HTML], given the conditions defined in that standard" or something?

[...]

PR #8759, opened 2023-04-24
Remove duplicate script-blocking style sheet requirement

domfarolino:
Per https://github.com/w3c/csswg-drafts/issues/3103#issuecomment-421261334, this PR removes a reference to HTML's "script-blocking style sheet" concept from https://drafts.csswg.org/cssom/#requirements-on-user-agents-implementing-the-xml-stylesheet-processing-instruction.

PTAL @annevk. Btw do you have any opinions on what we should do about the same reference to the same term at the bottom of https://drafts.csswg.org/cssom/#requirements-on-user-agents-implementing-the-http-link-header? These days HTML defines the `Link` header processing model, and as per https://html.spec.whatwg.org/#table-link-relations we only support `preconnect` and `preload`, not what CSSOM-1 defines here (and this is noted by the section's "Note yet widely implemented" call-out). IMO it seems best to expand this PR to also remove (at the very least, scoped to this PR) that reference from that section and let HTML work it out if it chooses to.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1549,7 +1549,6 @@ must be run:
      </ol>
    </ol>
  </ol>
-
 </div>
 
 ### Requirements on user agents Implementing the HTTP Link Header ### {#requirements-on-user-agents-implementing-the-http-link-header}
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -71,6 +71,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: semantics.html
         type: dfn
             text: a style sheet that is blocking scripts
+            text: contributes a script-blocking style sheet
+            for: Document
+                text: script-blocking style sheet set
 urlPrefix: https://dom.spec.whatwg.org/#concept-
     type: dfn
         text: dispatch; url: event-dispatch
@@ -1541,6 +1544,11 @@ must be run:
      </ol>
    </ol>
  </ol>
+
+If <var>node</var> <a>contributes a script-blocking style sheet</a>, then user agents must <a
+for=list>append</a> <var>node</var> to its <a>node document</a>'s <a for=Document>script-blocking
+style sheet set</a>.
+
 </div>
 
 ### Requirements on user agents Implementing the HTTP Link Header ### {#requirements-on-user-agents-implementing-the-http-link-header}
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1543,15 +1543,6 @@ must be run:
  </ol>
 </div>
 
-A style sheet referenced by an <a>xml-stylesheet processing instruction</a> using the rules in this section, in the context of
-the {{Document}} of an <a>XML parser</a> is said to be
-<a>a style sheet that is blocking scripts</a> if the <code>ProcessingInstruction</code>
-<a>node</a> was created by that {{Document}}'s parser, and the style sheet was
-enabled when the node was created by the parser, the last time the
-<a>event loop</a> reached step 1, the node was in that Document,
-and the user agent hasn't given up on loading that particular style sheet
-yet. A user agent may give up on such a style sheet at any time.
-
 ### Requirements on user agents Implementing the HTTP Link Header ### {#requirements-on-user-agents-implementing-the-http-link-header}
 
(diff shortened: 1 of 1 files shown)


annevk:
We cannot remove this text without equivalent text that `style` and `<link rel=stylesheet>` have (which also seems like it could be abstracted more, but I haven't delved into it too deeply):

> If element [contributes a script-blocking style sheet](https://html.spec.whatwg.org/#contributes-a-script-blocking-style-sheet), [append](https://infra.spec.whatwg.org/#set-append) element to its [node document](https://dom.spec.whatwg.org/#concept-node-document)'s [script-blocking style sheet set](https://html.spec.whatwg.org/#script-blocking-style-sheet-set).

I think the `Link` header part can be removed in its entirety. I would do that in a separate PR that references https://github.com/whatwg/html/issues/8741.

domfarolino:
Sure, that sounds good. Do you have a preference as to whether that text should be _inside_ the algorithm steps (I guess close to the end like in HTML) or as this spec had it, outside the algorithm in a normative block below?

domfarolino:
That sounds ideal. If it can happen outside of this PR (if that's what you're suggesting), then would you mind reviewing this for now? (I can't formally add anyone as a reviewer I guess...)

annevk:
As clarified via chat this is blocked on the new changes being pushed.

cssom-1/Overview.bs
```diff
@@ -71,6 +71,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: semantics.html
         type: dfn
             text: a style sheet that is blocking scripts
```
domfarolino, line 73: This is still used by the `Link` section, so we can probably remove it once we address https://github.com/w3c/csswg-drafts/pull/8759#issuecomment-1523008142.

cssom-1/Overview.bs
```diff
@@ -1541,16 +1544,12 @@ must be run:
...
-<a>node</a> was created by that {{Document}}'s parser, and the style sheet was
-enabled when the node was created by the parser, the last time the
-<a>event loop</a> reached step 1, the node was in that Document,
-and the user agent hasn't given up on loading that particular style sheet
-yet. A user agent may give up on such a style sheet at any time.
+If <var>node</var> <a>contributes a script-blocking style sheet</a>, then user agents must <a
+for=list>append</a> <var>node</var> to its <a>node document</a>'s <a for=Document>script-blocking
+style sheet set</a>.
+
+</div>
```
domfarolino, line 1552: Wanted to keep this text inside the algorithm just so the variable highlighting works, but more comfortable following the structure of this spec _before_ this PR, where this text is outside the actual numbered steps.

annevk (approved):
This doesn't look great I'd say, but I guess it's an improvement.

domfarolino:
Yeah, I kinda agree. Mostly my priority here is closing out https://github.com/w3c/csswg-drafts/issues/3103 since it seems bad to have duplicate out-of-sync definitions of the same criteria for script-blocking-ness. Is there anything you'd prefer me to do in this PR to make this all more satisfiable? I guess it's probably better to modify the processing model directly to append the style sheet to the script-blocking style sheet set before we fetch, and the in the response handler remove it from the set etc., but I'm not sure if that matches what you think would be the ideal long-term state for all of this anyways given https://github.com/w3c/csswg-drafts/pull/8759#issuecomment-1531779996.

annevk:
Looking at everyone that invokes "create a CSS style sheet" it seems to me this could be added to "add a CSS style sheet" and removed from all the callers of "create a CSS style sheet".

Those algorithms could also use a lot of cleanup in terms of making their arguments more explicit. And not having this weird "environment encoding" callback setup.

Also, https://html.spec.whatwg.org/#contributes-a-script-blocking-style-sheet has to be cleaned up as it doesn't account for `<?xml-stylesheet?>` at the moment.

domfarolino:
I've added the check to "add a CSS style sheet". If you think this looks good I'll make PRs to HTML and https://w3c.github.io/webvtt/#ref-for-create-a-css-style-sheet to remove the reference to the "contributes ..." condition since it appears inside "add", which indeed is called by "create".

annevk (commented):
Yeah this looks okay.

emilio (approved):
Looks good, thanks!

PR #8763, opened 2023-04-24
[scroll-animations-1] Add auto duration and animation ranges to explainer

flackr:
This adds sections explaining the addition of automatically fitting an auto duration to a scroll timeline or view timeline and the introduction of a ranges section to explain animation ranges.

### DIFF
--- a/scroll-animations-1/EXPLAINER.md
+++ b/scroll-animations-1/EXPLAINER.md
@@ -355,23 +355,77 @@ controlled, and inspected with existing methods.
 
 Some of the complexities that come from this decision are:
 
-* **Time vs Scroll Offset**: The concept of **time value** (exposed in
-  Milliseconds units) were central to the Web Animation model.  Adding
-  ScrollTimeline requires updating the animation timing model to allow
-  for other unit types. We decided to go with percentages as they had a
-  clean direct mapping to animation progress (which is already a proportion).
+#### Time vs Scroll Offset
 
-* **Exclusive end ranges**: In Web Animations, ranges have exclusive ends to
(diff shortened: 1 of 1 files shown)


scroll-animations-1/EXPLAINER.md
```diff
@@ -355,23 +355,77 @@ controlled, and inspected with existing methods.
...
+  animation-timeline: view();
+}
+```
+
+The animation's start time will be implicitly set to the start of the enter timeline range,
+and the animation's "auto" duration will intrinsically resolve to the length of the enter timeline range.
+
+#### Exclusive end ranges
+
+In Web Animations, ranges have exclusive ends to
```
kevers-google, line 416: This might need more clarification.  Only one of the boundaries of the active range is exclusive based on the playback rate.  

Not clearly stated why it is problematic for ScrollTimelines.

Would the following be clearer:

An [exception](...) is made for a ScrollTimeline with an active range that fills the scroll range to avoid the effect from becoming inactive at the scroll boundary. Since unable to scroll past the boundary no special accommodation is needed to facilitate animations intersecting at the boundary.

kevers-google (changes requested):

scroll-animations-1/EXPLAINER.md
flackr: Done, I've updated this section using similar text to what you recommended. Your link didn't seem to work, but I instead linked to the spec section on [progress timeline boundary](https://www.w3.org/TR/web-animations-2/#at-progress-timeline-boundary).

kevers-google: My link was just a  ... placeholder  :)

Issue #8666, opened 2023-03-31
[scroll-animations-1] Physical axis names mismatch


fantasai:
`scroll-snap-type` uses `x` and `y`, but `scroll-timeline-axis` uses `horizontal` and `vertical`. We should get these aligned. :)

fantasai:
Proposal: rename `horizontal` and `vertical` keywords in Scroll Animations to `x` and `y` to match Scroll Snap (`scroll-snap-type`) and Overflow (`overflow-x/y`).

birtles:
Seems reasonable to me.

flackr:
Sounds good to me.

astearns:
Looks like enough consensus to me :) 

The CSSWG will automatically accept this resolution one week from now if no objections are raised here. Anyone can add an emoji to this comment to express support. If you do not support this resolution, please add a new comment.

Proposed Resolution: Rename horizontal and vertical keywords in Scroll Animations to x and y to match Scroll Snap (scroll-snap-type) and Overflow (overflow-x/y).

astearns:
**Resolved**: Rename horizontal and vertical keywords in Scroll Animations to x and y to match Scroll Snap (scroll-snap-type) and Overflow (overflow-x/y).

PR #8774, opened 2023-04-27
[scroll-animations-1] Rename physical axes to x/y

andruud:
#8666


andruud:
@fantasai Please have a look.

The keyword order most consistent with scroll snap would be `x | y | block | inline`, but I figured the initial value should come first ...

bramus (approved):

PR #8779, opened 2023-04-29
[anchor-position] Allow anchoring to abspos anchors

### LINKED ISSUE #8165: [css-anchor-1] Can't anchor one popover to another
Right now we can't anchor one [popover](https://github.com/whatwg/html/pull/8221) to another, because:
- Popovers are in the document's [top layer](https://fullscreen.spec.whatwg.org/#top-layer), which means they are absolutely positioned with either viewport or ICB as the containing block
- Then none of them can be an [acceptable anchor](https://drafts4.csswg.org/css-anchor-1/#acceptable-anchor-element) for another

This means we can't support some very important use cases like nested popover menus.

Not sure exactly which spec to fix (top level, popover or CSS anchor). One possible idea to fix it in CSS is: maybe we can take tree ordering into consideration? If two abspos elements are in the same containing block, the second can anchor to the first.

cc @mfreed7

xiaochengh:
Fixes #8165

Add the idea of https://github.com/w3c/csswg-drafts/issues/8165#issuecomment-1499517290 to spec text.

@tabatkins

### DIFF
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -611,37 +611,20 @@ might not be capable of anchoring the positioned element.
 <div algorithm="acceptable anchor element">
 	An element |el| is a <dfn export>acceptable anchor element</dfn>
 	for an [=absolutely positioned=] element |query el|
-	if any of the following are true:
+	if all of the following are true:
 
-	* |query el| is in a higher [=root layer=] than |el|.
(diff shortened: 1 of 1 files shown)


TheZinna:
> Fixes #8165
> 
> Add the idea of https://github.com/w3c/csswg-drafts/issues/8165#issuecomment-1499517290 to spec text.
> 
> @tabatkins

PR #8814, opened 2023-05-07
[css-view-transitions-1] Rename snapshot-root to snapshot containing block

### LINKED ISSUE #8505: [css-view-transitions-1] Enforce ::view-transition to be fixed position
The `::view-transition` pseudo-element is positioned relative to the snapshot root outlined in the spec [here](https://drafts.csswg.org/css-view-transitions-1/#pseudo:~:text=Its%20containing%20block%20is%20the%20snapshot%20root.) irrespective of the value of its `position` property. This was resolved in #7859.

It is also meant to be the containing block for its child pseudo-elements, the `::view-transition-group`, clarified in the spec [here](https://drafts.csswg.org/css-view-transitions-1/#pseudo:~:text=This%20element%20provides%20a%20containing%20block%20for%20all%20%3A%3Aview%2Dtransition%2Dgroup()%20pseudo%2Delements.). This is done by UA CSS setting its `position` property to `fixed`.

It's unclear what should happen if developer CSS modifies the `position` value for this element. We don't have any developer use-case where that's needed. Also the transform value computed by the UA and set on `::view-transition-group` pseudo-elements is in the snapshot root's coordinate space. This relies on these elements being positioned relative to `::view-transition` which itself uses snapshot root as its containing block.

The proposal is to state that the `position` value for `::view-transition` always computes to `fixed`. I'd also be open to using `absolute` since its containing block is not the ICB and the desired behaviour is to make it a containing block for descendants. This is similar to what's done for top layer elements, spec'd [here](https://fullscreen.spec.whatwg.org/#new-stacking-layer:~:text=If%20its%20specified%20position%20property%20is%20not%20absolute%20or%20fixed%2C%20it%20computes%20to%20absolute.).

noamr:
[css-view-transitions-1] Rename snapshot-root to snapshot containing block and determine that the snapshot CB is an absolute/fixed positioning containing block for its descendants.

Based on [this resolution](https://github.com/w3c/csswg-drafts/issues/8505#issuecomment-1505587046).
Closes #8505

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -621,16 +621,13 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 ### <dfn>::view-transition</dfn> ### {#::view-transition}
 
-	<div class=note>This element provides a containing block for all ''::view-transition-group()'' pseudo-elements.</div>
+	<div class=note>This element serves as a parent for all ''::view-transition-group()'' pseudo-elements.</div>
 
 	A [=tree-abiding pseudo-element=] that is also
 	a [=pseudo-element root=].
 	Its [=originating element=] is the document's [=document element=].
 
 	Its [=containing block=] is the [=snapshot containing block=].
-
-	The [=snapshot containing block=] is considered to be an [=absolute positioning containing block=]
-	and a [=fixed positioning containing block=] for ''::view-transition'' and its descendants.
 
 	The following is added to the [=global view transition user agent style sheet=]:
 
@@ -818,6 +815,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	The <dfn>snapshot containing block origin</dfn> refers to the top-left corner of the [=snapshot containing block=].
 
 	The <dfn>snapshot containing block size</dfn> refers to the width and height of the [=snapshot containing block=] as a [=/tuple=] of two numbers.
+
+	The [=snapshot containing block=] is considered to be an [=absolute positioning containing block=]
+	and a [=fixed positioning containing block=] for ''::view-transition'' and its descendants.
 
 ## The [=view-transition layer=] stacking layer ## {#view-transition-stacking-layer}
 
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -782,7 +782,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 	The initial phase is the first item in the sequence.
 
-## The snapshot containing block ## {#snapshot-root-concept}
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -782,39 +785,39 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 	The initial phase is the first item in the sequence.
 
-## The snapshot root ## {#snapshot-root-concept}
+## The snapshot containing block ## {#snapshot-root-concept}
```
khushalsagar, line 788:
```suggestion
## The snapshot containing block ## {#snapshot-containing-block-concept}
```

css-view-transitions-1/Overview.bs
```diff
@@ -627,7 +627,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	a [=pseudo-element root=].
 	Its [=originating element=] is the document's [=document element=].
 
-	Its [=containing block=] is the [=snapshot root=].
+	Its [=containing block=] is the [=snapshot containing block=].
+
+	The [=snapshot containing block=] is considered to be an [=absolute positioning containing block=]
```
khushalsagar, line 632: @bokand @bfgeek do you mind reviewing this? This is the spec text which defines the containing block for view-transition pseudo-elements. If `::view-transition` is set to `position: static` by the author then it, and any descendant which would've used `::view-transition` as its containing block, should instead use the snapshot containing block.

khushalsagar (approved):
LGTM % the point which can use David or Ian's review.

css-view-transitions-1/Overview.bs
bokand: Would it make sense to put this bit ("the SCB is considered to be an APCB and a FPCB") down with the definition of Snapshot Containing Block?

Also, above:

```
<div class=note>This element provides a containing block for all ''::view-transition-group()'' pseudo-elements.</div>
```

This depends on the style given to `::view-transition` - perhaps it should instead say "This element serves as a parent for all `::view-transition-group()` pseudo-elements"?

bokand (approved):
lgtm but @bfgeek can better judge correctness

css-view-transitions-1/Overview.bs
noamr: Fixed, thanks!

css-view-transitions-1/Overview.bs
khushalsagar: Did you mean to resolve this after applying the suggestion? :)

khushalsagar:
Just the one comment about heading name, otherwise LGTM.

css-view-transitions-1/Overview.bs
noamr: Oops, yes

PR #8819, opened 2023-05-08
[css-view-transitions-1] Clarify that observable image size refers to the border-box, and observed natural size is not affected by hw constraints

### LINKED ISSUE #8561: [css-view-transitions-1] Define behaviour when capturing image for a large element.
The default behaviour for View Transitions when capturing the image of an element is to paint its entire ink overflow rectangle. But there are cases where this is not possible. A common one is the element exceeding max texture size supported by the hardware on that device. It can technically be done (by tiling the image) but is complicated. It's also likely we'll hit cases where painting the whole element will have significant memory overhead and clipping the snapshot makes sense.

We want the snapshot to include *at least* the intersection of the element's ink overflow rectangle and the snapshot root (a.k.a the visible viewport). Other than that, it's unclear how prescriptive the spec needs to be here. The spec could recommend that for the area outside the visible viewport, the implementation should capture the subset closest to the viewport.

This issue is tracked [here](https://drafts.csswg.org/css-view-transitions-1/#issue-c9dce14f) in the spec.

### LINKED ISSUE #8597: [css-view-transitions-1] Exposing ink overflow rect bounds to script
The steps for capturing an element's image during View Transitions involve painting the element and all its descendants in their entirety such that the produced image's bounds are equal to the element's [ink overflow rectangle](https://www.w3.org/TR/css-overflow-3/#ink-overflow-rectangle) bounds. The spec text is [here](https://drafts.csswg.org/css-view-transitions-1/#capture-the-image-algorithm).

The `::view-transition-group` pseudo-element which displays the image has bounds equal to the captured element's border-box. Since the image's natural size is equal to the element's ink overflow area but it needs to be drawn in an element which is smaller than that, we use `object-view-box` to ensure the correct subset paints into the element.

So for example if you have a 100x100 box with a shadow that paints 10px outside this box:

- The ink overflow rect is [-10, -10] 120X120.
- The natural size of the produced image will be 120x120.
- The `::view-transition-group` for this will be sized to 100x100.
- The `object-view-box` set on this group will be `inset (10px, 10px, 10px, 10px)`.

Given that script can read the `object-view-box` on the group, it allow authors to query the value for ink overflow rectangle computed by the UA.

There was an additional risk of fingerprinting if `object-view-box` was affected by where the UA decided to clip the element's painting. But with the resolution in #8561 that is not an issue. Clipping the element's painting is an internal detail, the natural size of the image and its `object-view-box` value is always as described above.

It came up during the discussion in #8561 that its odd that the computed ink overflow rectangle is being exposed to authors. Since there are cases where it is technically infinite and left undefined, see spec text [here](https://www.w3.org/TR/css-overflow-3/#ink-overflow:~:text=Since%20some%20effects%20in%20CSS%20(for%20example%2C%20the%20blurs%20in%20text%2Dshadow%20%5BCSS%2DTEXT%2D3%5D%20and%20box%2Dshadow%20%5BCSS%2DBACKGROUNDS%2D3%5D%2C%20which%20are%20theoretically%20infinite)%20do%20not%20define%20what%20visual%20extent%20they%20cover%2C%20the%20extent%20of%20the%20ink%20overflow%20is%20undefined.). So now authors will see a value for it based on the UA's implementation.

Questions:
- Is the above ok as-is? The value is exposed to script.
- If the value shouldn't be exposed to script, how do we set that up. We can't just elide it from getComputedStyle. An author could do `object-view-box: unset` and learn the value based on how it affects the size of the replaced element displaying this image.

@chrishtr @fantasai FYI.

noamr:
Ink-overflow is an implementation detail, and up to the UA to apply. object-view-box doesn't directly affect the ink-overflow rect. Replaced mentioned of `ink-overflow rect` with `border box`, and added note about how ink-overflow plays into this.

Also, texture size constraints should not affect the observed natural size of the captured image. Instead, implementations
should capture at least the visible part of the element (downscaled if they must).

See [resolution 1](https://github.com/w3c/csswg-drafts/issues/8597#issuecomment-1488952275) and [resolution 2](https://github.com/w3c/csswg-drafts/issues/8561#issuecomment-1470414852).

Closes #8561 
Closes #8597

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1500,7 +1500,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Assert=]: |transition|'s [=ViewTransition/phase=] is not "`done`".
 
-		1. If |transition|'s [=ViewTransition/phase=] is [=phases/before=] "`update-callback-called`", then [=call the update callback=] of |transition|.
+		1. If |transition|'s [=ViewTransition/phase=] is [=phases/before=] "`update-callback-called`", then [=queue a global task=] on the [=DOM manipulation task source=],
+			given |transition|'s [=relevant global object=], to [=call the update callback=] of |transition|.
 
 		1. Set [=document/transition suppressing rendering=] to false.
 
@@ -1591,7 +1592,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		* Implementations may clip the rendered contents if the [=ink overflow rectangle=] exceeds some [=implementation-defined=] maximum.
 			However, the captured image should include, at the very least, the contents of |element| that intersect with the [=snapshot root=].
-			Implementations may adjust the rasterization quality to account for elements with a large [=border box=] that are transformed into view.
+			Implementations may adjust the rasterization quality to account for elements with a large [=ink overflow area=] that are transformed into view.
 
 		* [=list/For each=] |descendant| of [=shadow-including descendant=] {{Element}} and [=pseudo-element=] of |element|,
 			if |descendant| has a [=computed value=] of 'view-transition-name' that is not ''view-transition-name/none'',
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1790,6 +1790,8 @@ Changes from <a href="https://www.w3.org/TR/2022/WD-css-view-transitions-1-20221
 * Add support for :only-child. See <a href="https://github.com/w3c/csswg-drafts/issues/8057">issue 8057</a>.
 * Add concept of a tree of pseudo-elements under [=pseudo-element root=]. See <a href="https://github.com/w3c/csswg-drafts/issues/8113">issue 8113</a>.
 * When skipping a transition, the {{UpdateCallback}} is called in own task rather than synchronosly. See <a href="https://github.com/w3c/csswg-drafts/issues/7904">issue 7904</a>
+* When capturing images, at least the in-viewport part of the image should be captured, downscale if needed. See <a href="https://github.com/w3c/csswg-drafts/issues/8561">issue 8561</a>.
+* Applying the [=ink overflow=] to the captured image is implementation defined, and doesn't affect the image's [=natural size=]. See <a href="https://github.com/w3c/csswg-drafts/issues/8597">issue 8597</a>.
 
 <h3 id="changes-since-2022-10-25">
 Changes from <a href="https://www.w3.org/TR/2022/WD-css-view-transitions-1-20221025/">2022-10-25 Working Draft (FPWD)</a>
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1570,7 +1570,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 				following the [=capture rendering characteristics=].
 
 			1. Return the portion of the canvas that includes |element|'s [=ink overflow rectangle=] as an image.
-				The [=natural size=] of the image must be |element|'s [=border box=]'s [=size=].
+				The [=natural size=] of the image must be |element|'s [=border box=]'s [=size=],
+				and its origin corresponds to the |element|'s [=border box=]'s origin.
 
 	</div>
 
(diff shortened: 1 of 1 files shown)


jakearchibald:
But the ink overflow does impact the texture size, so it does matter when the UA is looking to limit the texture size.

noamr:
@jakearchibald ok clarified it for now, and will revise the whole thing with #8561.

css-view-transitions-1/Overview.bs
```diff
@@ -1580,7 +1583,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	<div algorithm>
 		The <dfn>capture rendering characteristics</dfn> are as follows:
 
-		* The origin of |element|'s [=ink overflow rectangle=] is anchored to canvas origin.
+		* The origin of |element|'s [=border box=] is anchored to canvas origin.
```
khushalsagar, line 1586: I think the existing spec text in capture rendering characteristics doesn't need to change. It's meant to theoretically explain how the image would be generated if no clipping was required and the painted output should cover the element's ink overflow rectangle. So we'd allocate a canvas equal to ink overflow bounds with the origin of the canvas being the ink overflow rectangle's origin.

What you have here should go into step 8.6.2. We can delete the "computing the interest rectangle" algorithm altogether and also steps 2 and 3 in 8.6.2. Instead add what you have here:

* The UA may clip the rendered output from step 8.6.2.1 but it should include what's in the visible viewport and the rasterization quality should be optimal for the scale at which this element renders in the DOM (this is the point that we can discuss tomorrow).
* The returned image has a natural size equal to the element's border box and its origin maps to the border box's origin in the rendered output in step 1. The fact that the underlying pixel data may be smaller or larger than this is a UA detail. The Note below covers this well.

css-view-transitions-1/Overview.bs
```diff
@@ -1500,7 +1500,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Assert=]: |transition|'s [=ViewTransition/phase=] is not "`done`".
 
-		1. If |transition|'s [=ViewTransition/phase=] is [=phases/before=] "`update-callback-called`", then [=call the update callback=] of |transition|.
+		1. If |transition|'s [=ViewTransition/phase=] is [=phases/before=] "`update-callback-called`",
```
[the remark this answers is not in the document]
noamr: oops

css-view-transitions-1/Overview.bs
noamr: Thanks, applied this, please look if this is what you meant.

css-view-transitions-1/Overview.bs
```diff
@@ -1589,6 +1588,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		* Effects on the element, such as 'opacity' and 'filter' are applied to the capture.
 
+		* Implementations may clip the rendered contents if the [=ink overflow rectangle=] exceeds some [=implementation-defined=] maximum.
+			However, the captured image should include, at the very least, the contents of |element| that intersect with the [=viewport=].
```
khushalsagar, line 1592: s/viewport/snapshot root (or snapshot containing block).

css-view-transitions-1/Overview.bs
```diff
@@ -1589,6 +1588,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+			Implementations may adjust the rasterization quality to account for elements with a large [=border box=] that are transformed into view.
```
khushalsagar, line 1593: Better to say large ink overflow bounds instead of border box. If you have an element with a massive overflow but small box, same policy applies.

css-view-transitions-1/Overview.bs
```diff
@@ -1569,10 +1569,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 				over an infinite transparent canvas,
 				following the [=capture rendering characteristics=].
 
-			1. Let |interestRectangle| be the result of [=computing the interest rectangle=] for |element|.
+			1. Return the portion of the canvas that includes |element|'s [=ink overflow rectangle=] as an image.
+				The [=natural size=] of the image must be |element|'s [=border box=]'s [=size=].
```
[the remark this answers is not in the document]

css-view-transitions-1/Overview.bs

css-view-transitions-1/Overview.bs

khushalsagar (approved):

khushalsagar (approved):

PR #8820, opened 2023-05-08
Fix Github Pages redirects

andreubotella:
Rather than building the specs with bikeshed using a text macro that makes them redirect to drafts.csswg.org, this PR instead uses a simple Python script to turn every *.bs file into an HTML redirect, and then also creates unversioned redirects for every spec.

This should fix the current issues with Github Pages.

@sideshowbarker @tabatkins


bin/build-github-pages-redirects.py
```diff
@@ -0,0 +1,47 @@
...
+paths_to_build = set(["/", "/css-snapshot/"])
+
+for file in glob.glob("**/*.bs"):
+    if file.endswith("/Overview.bs"):
+        folder = "/" + file.removesuffix("Overview.bs")
+        paths_to_build.add(folder)
+
+        # If the folder ends with a version number, remove it to build
+        # the unversioned spec as well.
+        # We make sure to exclude css-20XX snapshots.
```
andreubotella, line 20: Oh, I just noticed that https://drafts.csswg.org/css/ does serve the 2020 snapshot (not the current one). Should https://w3c.github.io/csswg-drafts/css redirect to drafts.csswg.org as well, then?

bin/build-github-pages-redirects.py
```diff
@@ -0,0 +1,47 @@
...
+
+    with open(dest, mode="w", encoding="UTF-8") as f:
+        f.write(
+            f"""
+<!doctype html>
+<meta charset=utf-8>
+<title>Redirecting to drafts.csswg.org...</title>
+<meta http-equiv=Refresh content="0; url='https://drafts.csswg.org{path}'">
+<script>
+window.location.href = "https://drafts.csswg.org{path}";
```
sideshowbarker, line 44: What about URLs with fragment IDs? https://w3c.github.io/csswg-drafts/css-backgrounds/#the-background, for example.

bin/build-github-pages-redirects.py
```diff
@@ -0,0 +1,47 @@
...
+for path in paths_to_build:
+    dest = "./build" + path
+    if dest.endswith("/"):
+        dest += "index.html"
+
+    # Make sure the folder exists
+    os.makedirs(os.path.dirname(dest), exist_ok=True)
+
+    with open(dest, mode="w", encoding="UTF-8") as f:
+        f.write(
```
sideshowbarker, line 37: Quite possibly I am completely missing something with regard to how the build and proxying work, but:

Given that drafts.csswg.org is now proxying w3c.github.io/csswg-drafts, isn’t this just going to result in the contents of every `index.html` file for every spec on drafts.csswg.org — both the versioned and unversioned specs — being the meta-refresh+window.location.href redirect below, rather than the actual spec contents?

With the change made in this PR, I can’t see any step where the spec contents are actually being built anywhere at all any longer; specifically, Bikeshed seems to be never getting run anywhere any longer. And even if it were first getting run at some point before this step, then this step would cause whatever `index.html` that Bikeshed had generated to be overwritten by the `index.html` that this step is writing.

As I mentioned earlier, it’s quite possible that I’m missing something — but as far as I understand, Bikeshed is not being run at all any longer in the environment for the drafts.csswg.org server, and instead all that’s happening is that the drafts.csswg.org server is proxying whatever has been built and published on the w3c.github.io/csswg-drafts server. But with this change, what’s on the w3c.github.io/csswg-drafts server would now end up being entirely these meta-refresh+window.location.href redirect files, rather than being and Bikeshed output.

PR #8827, opened 2023-05-11
[css-view-transitions-1] When skipping transitions, call the DOM-update callback from a task

### LINKED ISSUE #7904: [css-view-transitions-1] Should the DOMChangeCallback be posted as a task from skip transition steps?
Step 4 in [skip the page transition algorithm](https://drafts.csswg.org/css-view-transitions-1/#skip-the-page-transition-algorithm) says that the DOM change callback has to be invoked synchronously. However, the skip the page transition algorithm sometimes runs as part of the update the rendering, and we really want to avoid running (new) script in those steps. 

We should make it asynchronous. The question is should we just make it asynchronous for all places that run the skip the page transition algorithm, or should it be conditional?

I'm leaning towards just make it all asynchronous, but I'm unsure if there are cases where synchronous invocation is important.

/cc @khushalsagar @jakearchibald

noamr:
Amend assert to allow calling the callback from a "done" state

The "finished" promise is bound to the update callback promise, so it doesn't need to change (the update callback is async anyway).

See [resolution](https://github.com/w3c/csswg-drafts/issues/7904#issuecomment-1468794641).

Closes #7904

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1468,7 +1468,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		even if the transition is [=Skip the view transition|skipped=].
 		The reasons for this are discussed in [[#transitions-as-enhancements]].
 
-		1. [=Assert=]: |transition|'s [=ViewTransition/phase=] is [=phases/before=] "`update-callback-called`".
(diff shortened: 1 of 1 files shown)


khushalsagar (approved):
Can you add a line to the change list at the end of the spec, describing this edit.

noamr:
> Can you add a line to the change list at the end of the spec, describing this edit.

Done

khushalsagar (approved):
LGTM again.

noamr:
> LGTM again.

Thanks! Do you have merge permissions or do we need to tag someone else?

Issue #7882, opened 2022-10-13
[css-view-transitions-1] Handling cases where transition constraints are broken mid transition


khushalsagar:
View transitions require a bunch of constraints to be satisfied by tagged elements listed [here](https://drafts.csswg.org/css-view-transitions-1/#perform-an-outgoing-capture):

```
If any of the following is true:

usedTransitionTags [contains](https://infra.spec.whatwg.org/#list-contain) transitionTag.

element is not element’s [root](https://dom.spec.whatwg.org/#concept-tree-root) and element does not have [layout containment](https://drafts.csswg.org/css-contain-2/#layout-containment).

element is not element’s [root](https://dom.spec.whatwg.org/#concept-tree-root) and element allows [fragmentation](https://drafts.csswg.org/css-break-4/#fragmentation).
```

This state can also change mid transition. The simplest thing is to skip the transition in this case.

khushalsagar:
Summary from offline discussion: It makes sense to abort the transition if per element constraints are broken, i.e., no layout containment or the element allows fragmentation. But detecting if the computed value of tags on all DOM elements has changed (so we now have duplicate tags etc.) adds more computation work. And that work would be to detect a developer bug and abort the transition. So not worth it.

**Proposed Resolution**: If constraints on elements participating in a transition become unsatisfied after a transition is running, the transition is aborted.

khushalsagar:
One constraint (which I missed above) is that an element participating in a transition must generate a box (can't be `display: none`).

When walking through the DOM to identify named elements, we ignore DOM elements which don't generate a box (as if they had `view-transition-name:none`). See spec text [here](https://drafts.csswg.org/css-view-transitions-1/#perform-an-old-capture-algorithm): "If transitionName is [none](https://drafts.csswg.org/css-view-transitions-1/#valdef-view-transition-name-none), or element is [not rendered](https://www.w3.org/TR/css-images-4/#element-not-rendered), then [continue](https://infra.spec.whatwg.org/#iteration-continue)."

A change in this state (such that the element no longer generates a box) should also skip the transition.

jakearchibald:
Another option for an element changing to `display: none` (or being removed from the DOM):

If there's a `::view-transition-old`: Remove the `::view-transition-new` from that pair, and set the end position & size to the captured element's position & size. This means the animation would continue as if there was no captured element in the new state.

If there isn't a `::view-transition-old`: Remove the `::view-transition-group`.

[...]

PR #8831, opened 2023-05-12
[css-view-transitions-1] Fragmented elements don't participate in view transitions.

### LINKED ISSUE #8339: [css-view-transitions-1] Capturing fragmented elements
[View transitions](https://drafts.csswg.org/css-view-transitions-1/#capture-the-image-algorithm) specify that the image should be captured at the same size as in the document

> Render element and its [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant), at the same size it appears in its [node document](https://dom.spec.whatwg.org/#concept-node-document), over an infinite transparent canvas, following the [capture rendering characteristics](https://drafts.csswg.org/css-view-transitions-1/#capture-rendering-characteristics).

It's unclear what to do if the element is fragmented. Is this supposed to capture the union of the fragment sizes/positions or something else?

I can't think of any good example where a developer would want to transition a fragmented element, so my proposal is to treat an element that is fragmented as invalid element to share in a view transition and skip the transition.

noamr:
Skip fragmented elements when preparing transition, and add note about looking at this again in the future.

See [resolution](https://github.com/w3c/csswg-drafts/issues/8339#issuecomment-1470268830).
Closes #8339

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1612,8 +1612,11 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |hasActiveAnimations| be a boolean, initially false.
 
-		1. [=list/For each=] |element| of |transition|'s [=ViewTransition/transition root pseudo-element=]'s [=tree/inclusive descendants=]:
+		1. [=map/For each=] |transitionName| → |capturedElement| of |transition|'s [=ViewTransition/named elements=],
+			if |capturedElement| has more than one [=box-fragment=],
+			then [=skip the view transition=] for |transition|, and return.
 
+		1. [=list/For each=] |element| of |transition|'s [=ViewTransition/transition root pseudo-element=]'s [=tree/inclusive descendants=]:
 			1. For each |animation| whose [=timeline=] is a [=document timeline=] associated with |document|,
 				and contains at least one [=animation/associated effect=] whose [=effect target=] is |element|,
 				set |hasActiveAnimations| to true if any of the following conditions is true:
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1612,10 +1612,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |hasActiveAnimations| be a boolean, initially false.
 
-		1. [=map/For each=] |transitionName| → |capturedElement| of |transition|'s [=ViewTransition/named elements=],
-			if |capturedElement| has more than one [=box-fragment=],
-			then [=skip the view transition=] for |transition|, and return.
-
 		1. [=list/For each=] |element| of |transition|'s [=ViewTransition/transition root pseudo-element=]'s [=tree/inclusive descendants=]:
 			1. For each |animation| whose [=timeline=] is a [=document timeline=] associated with |document|,
 				and contains at least one [=animation/associated effect=] whose [=effect target=] is |element|,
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1243,19 +1243,17 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 			1. If any [=flat tree=] ancestor of this |element| [=skips its contents=], then [=continue=].
 
+			1. If |element| is not |element|'s [=tree/root=] and |element| allows [=fragmentation=], then [=continue=].
```
khushalsagar, line 1246:
```suggestion
			1. If |element| has more than one [=box-fragment=], then [=continue=].
```
root doesn't need to be special cased, we ignore it if its fragmented just the same as other elements. And the resolution was to do this if the box is fragmented during the transition as opposed to whether it can be fragmented.

khushalsagar, line 1246: We need this in [Handle transition frame](https://drafts.csswg.org/css-view-transitions-1/#handle-transition-frame) as well. Iterate through all the elements and check if any of them is fragmented. If so, skip.

The idea being that we ignored fragmented elements when looking for elements with a view-transition-name. But if an element became fragmented during the animation, we skip the transition.

noamr: I didn't see any of this in the resolution. The resolution was that fragmented elements are ignored, not that they cancel the transition. Am I missing a conversation? Do we need to deal with this at all after elements were captured?

khushalsagar: This was part of option 1 which we were asking the group to resolve on:

"When constraints are broken, transitions skip to the end (https://github.com/w3c/csswg-drafts/issues/7882).
Sub-option: Is the constraint broken if an element fragments, or if an element can fragment?"

We have a similar pattern with elements having a box. Safely ignore an element which has a view-transition-name but doesn't generate a box: https://drafts.csswg.org/css-view-transitions-1/#capture-old-state-algorithm:~:text=for%20element.-,If%20transitionName%20is%20none%2C%20or%20element%20is%20not%20rendered%2C%20then%20continue.,-If%20any%20of

But because elements in the new DOM are live, this state can change for them while the transition is running. We treat them as developer errors and abort the transition: https://drafts.csswg.org/css-view-transitions-1/#update-pseudo-element-styles:~:text=If%20any%20flat%20tree%20ancestor%20of%20capturedElement%E2%80%99s%20new%20element%20skips%20its%20contents%2C%20or%20capturedElement%E2%80%99s%20new%20element%20is%20not%20rendered%2C%20then%20return%20failure.

So the logic is that if a DOM element was not fragmented when the transition started but becomes fragmented during the animation, then the safest way to handle it is to abort the transition.

css-view-transitions-1/Overview.bs
```diff
@@ -1614,8 +1612,11 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |hasActiveAnimations| be a boolean, initially false.
 
-		1. [=list/For each=] |element| of |transition|'s [=ViewTransition/transition root pseudo-element=]'s [=tree/inclusive descendants=]:
+		1. [=map/For each=] |transitionName| → |capturedElement| of |transition|'s [=ViewTransition/named elements=],
```
khushalsagar, line 1615: Sorry for the confusion but now that dug through the spec to answer your question, step 8.8.3.1 [here](https://drafts.csswg.org/css-view-transitions-1/#style-transition-pseudo-elements-algorithm:~:text=If%20any%20flat%20tree%20ancestor%20of%20capturedElement%E2%80%99s%20new%20element%20skips%20its%20contents%2C%20or%20capturedElement%E2%80%99s%20new%20element%20is%20not%20rendered%2C%20then%20return%20failure.) seems like a better spot for this. We already check all the other per element constraints which abort the transition here.

PR #8847, opened 2023-05-17
[css-contain-2] Elements stay relevant to the user during a view transition

### LINKED ISSUE #8060: [css-contain-4] c-v: auto elements painted into snapshots for View Transitions
This was a part of #7874 but wasn't resolved at the last meeting where this issue was discussed. The edit also needs to go into the contain spec so filing another issue.

If an element is participating in a transition, any element in its sub-tree (which has c-v: auto) should become relevant to the user. This makes sense given that the named element will be painted into a snapshot. This snapshot's position and scale is animated during the transition (by the pseudo-element its drawn in). So all its content (including the c-v: auto elements) could be in the viewport.

Given that the snapshot for all non-root elements paints the entire DOM tree underneath, it'll make all c-v: auto elements underneath a non-root named element relevant to the user. It's only the root (for which the snapshot is clipped) which can have c-v: auto elements underneath that are clipped out of the snapshot and hence don't need to be made relevant to the user.

Proposed Resolution: An element with content-visibility: auto is relevant to the user if it is painted into self or an ancestor's snapshot during a view transition.

@vmpstr @frivoal fyi.

noamr:
Closes #8060

See [resolution](https://github.com/w3c/csswg-drafts/issues/8060#issuecomment-1411189366).

### DIFF
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -1681,6 +1681,10 @@ Suppressing An Element's Contents Entirely: the 'content-visibility' property {#
 			content-visibility/content-visibility-with-top-layer-005.html
 			content-visibility/content-visibility-with-top-layer-006.html
 			</wpt>
+
+		* The element or one of its ancestors participates in an ongoing {{ViewTransition}},
+			and the element was [=relevant to the user=] when {{Document/startViewTransition()}} was called.
+
 	</div>
 
 


noamr:
@khushalsagar @tabatkins I'm not sure if `css-contain-2` is the right place to put this, as it's already published? css-contain-3 doesn't have a relevant-to-the-user section and css-contain-4 doesn't exist.

css-contain-2/Overview.bs
```diff
@@ -1681,6 +1681,10 @@ Suppressing An Element's Contents Entirely: the 'content-visibility' property {#
 			content-visibility/content-visibility-with-top-layer-005.html
 			content-visibility/content-visibility-with-top-layer-006.html
 			</wpt>
+
+		* The element or one of its ancestors participates in an ongoing {{ViewTransition}},
```
khushalsagar, line 1685: This can be simplified to: "The element's [involved in a view transition](https://drafts.csswg.org/css-view-transitions-1/#involved-in-a-view-transition) bit is true".

That is meant to track the exact state we need here and makes it more explicit when this state changes.

vmpstr, line 1685: This isn't quite the right text for the resolution.

The resolution states that the element is relevant to the user if it has a (flat tree) **descendant** that is participating in an ongoing ViewTransition.

noamr: Right, will reword.

css-contain-2/Overview.bs
```diff
@@ -1681,6 +1681,10 @@ Suppressing An Element's Contents Entirely: the 'content-visibility' property {#
 			content-visibility/content-visibility-with-top-layer-005.html
 			content-visibility/content-visibility-with-top-layer-006.html
 			</wpt>
+
+		* The element has a [=flat tree=] descendant that is [=involved in a view transition=],
+			and the element was [=relevant to the user=] when the transition started.
```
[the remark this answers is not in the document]
noamr: Done

Issue #8727, opened 2023-04-17
[css-anchor] Anchor positioning should be logical first, not physical


jensimmons:
I started reading the specification for Anchor Positioning, and I was surprised to see it littered with examples that refer to `top`, `left`, etc. 

In 2023, as we design a new way to do layout, it should be _logical_ in the way it describes things. Not _physical_. 

Especially since this is a way to layout popovers in a dynamic fashion — where the designer is being suggestive, not being controlling. "Please try to put this popover over here if it fits, or over there if it doesn't." Here and there should refer to block and inline start, and end. Not top, bottom, left and right.

tabatkins:
Nothing about the spec is physical-only. The `<anchor-side>` keywords allow both physical and logical options; they can be used in any inset property, both `top` and `inline-start`.

Some of the example are written with physical properties, but that's just because they're shorter. I'd be happy to switch some of them to logical to ensure it's even more obvious this feature can be used either way, tho.

jensimmons:
That's great to hear. Yes, I think it'd be wise to make most if not almost all of the examples use logical options. I know it does make the code a bit longer, and even a bit slower to understand... but the web industry _is_ adopting logical thinking and replacing physical layout, but by bit. We want to encourage people to do so. And new layout mechanisms like Anchor Positioning are the perfect opportunity to say "This is how this works. Logical. (oh, I guess if you have to use physical, ok, whatever, we won't break that... but everyone out there teaching this, teach it using logical syntax!"

PR #8856, opened 2023-05-18
[css-anchor-position-1] Update the anchor() math example

kizu:
[css-anchor-1] Update the anchor() math example

I was re-reading the specs, and noticed that one of the examples seemed to promise a fix for one of the use-cases I was wondering what could be the best fix for — the one where we can use an extensive calculation to get a centered message.

When trying it in the Canary, I noticed that the code for the example is not optimal, so I did update the example.

Here is a CodePen demonstrating the example with my changes (using the calculations instead off the missing `abs()`): https://codepen.io/kizu/pen/wvYYLGr?editors=1100

The changes:

1. Because the example does not contain `display: flex` or similar, the `justify-content` does not do anything. So, in order to make things work — to center the element properly — we can use `auto` margin.
2. We can also demonstrate the `anchor-default` instead of using the explicit anchor name in the calculation. Potentially, this can make it easier to later reuse this as a part of a position-fallback.
3. I did recently read the https://github.com/w3c/csswg-drafts/issues/8727 issue, and I mostly agree with it, so I took this chance to replace the physical keywords with logical (I like how when used inside `anchor()` there is no need to think of which axis we're thinking about, and this potentially could be used to also separate these parts and reuse for different axis easier).

### DIFF
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -239,17 +239,17 @@ as if by an additional ''translate()'' transform.
 	<pre highlight=css>
 	.centered-message {
 		position: fixed;
-		max-width: max-content;
-		justify-content: center;
-
-		--center: anchor(--x 50%);
+		anchor-default: --x;
+		max-inline-size: max-content;
+		margin-inline: auto;
+		--center: anchor(50%);
 		--half-distance: min(
 			abs(0% - var(--center)),
 			abs(100% - var(--center))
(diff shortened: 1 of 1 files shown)


kizu:
Closing as outdated and my solution not being perfect (`margin: auto` can result in negative margins, resulting in weird behavior), in favor of trying to find a good general solution: https://github.com/w3c/csswg-drafts/issues/8979

Issue #8174, opened 2022-12-02
[selectors-4] Add syntax to establish before-change style for css-transitions on new elements.


flackr:
Transitions are ergonomic and easy to use for changes to property values when there is an existing style to transition from. However, when an element is first added to the page or made not `display: none` it has no previous [before-change](https://www.w3.org/TR/css-transitions-1/#before-change-style) style so developers have to use script to switch from the initial frame state and some state which specifies the post transition style, e.g.

```js
elem.classList.add('visible'); // Removes display: none, but keeps opacity: 0 and other before styles.
elem.offsetTop; // Force style recalc
elem.classList.add('visiblestyle'); // Adds properties to transition to.
```

It would be great if you could do this directly from CSS perhaps via a pseudoclass selector. E.g.

```css
dialog {
  transition: opacity 200ms;
  opacity: 1; /* Not necessary as it is the default but added for illustration. */
}
dialog:initial {
  opacity: 0;
}
```

This `:initial` pseudo-class would match on the initial frame containing a particular element and be removed immediately after style recalc. This does mean we would have to repeat style on that initial frame - hopefully we could target only elements affected by `:initial` selectors.

This would bring parity with what you can do via css animations, e.g.:
```css
@keyframes fade-in {
  from { opacity: 0; }
}
dialog {
  transition: opacity 200ms;
  animation: fade-in 200ms;
  opacity: 1; /* Not necessary as it is the default but added for illustration. */
}
```

Though thinking ahead, I've also heard frustration around the css animation above animating when you load the page - so we should consider where there's a way for developers to differentiate loading vs post load and only apply animations (and initial transition styles) for new elements.

flackr:
> Though thinking ahead, I've also heard frustration around the css animation above animating when you load the page - so we should consider where there's a way for developers to differentiate loading vs post load and only apply animations (and initial transition styles) for new elements.

I filed #8175 to talk through this particular area separately.

lilles:
It could work if :initial is matching if an element had no style (first frame or display:none) before the element style recalc is done for an element and reapply without :initial immediately. I'm not 100% sure about the consequences where we are looking at multiple passes like for container queries. Whether a multi-pass layout is stateless.

Since the before-change style would be based on the previous frame's style, or perhaps the :initial based styles, we probably just need to have a single :initial style pass even if the style would flip back and forth between display:none for the same update.

andruud:
[...]

PR #8876, opened 2023-05-25
[css-transitions-2] Add spec for @starting-style #8174

lilles:
[css-transitions-2] Add spec for @starting-style #8174

### DIFF
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -180,8 +180,8 @@ elements that change from not [=being rendered=] to being rendered. To allow
 for that, this specification introduces ''@starting-style''.
 
 The <dfn at-rule id="at-ruledef-starting-style">@starting-style</dfn> rule is a
-grouping rule where the rules inside it is used to establish styles to
-transition from if the previous [=style change event=] did not establish a
+grouping rule.  The style rules inside it are used to establish styles to
+transition from, if the previous [=style change event=] did not establish a
 [=before-change style=] for the element whose styles are being computed.
 
 Note: This means that ''@starting-style'' rules only apply to some elements
@@ -195,8 +195,8 @@ a [=before-change style=] for a given [=style change event=], the
 with the [=after-change style=] to start transitions
 ([[css-transitions-1#starting]]).
 
-The rules inside ''@starting-style'' cascade as any other grouping rules without
-introducing any new ordering to the cascade, which means rules inside
+The rules inside ''@starting-style'' cascade as any other grouped style rules
+without introducing any new ordering to the cascade, which means rules inside
 ''@starting-style'' do not necessarily win over those outside.
 
 Style rules in ''@starting-style'' do not apply to [=after-change style=].
@@ -204,13 +204,13 @@ Thus, the presence of matching rules in ''@starting-style'' can cause
 transitions to occur on elements that otherwise could not have transitions
 because they lack a [=before-change style=].
 
-''starting style'' inherits from the parent's [=after-change style=] just like
+[=Starting style=] inherits from the parent's [=after-change style=] just like
 [=after-change style=] does.
 
 <div class=example>
 
-  Transitioning background-color of an h1 element from transparent to green
-	from initial rendering:
+	The 'background-color' of an <code>h1</code> element can be transitioned
+	from transparent to green when it is initially rendered:
 
 	<pre class=lang-css>
 	h1 {
@@ -238,8 +238,8 @@ because they lack a [=before-change style=].
 </div>
 
 <div class=example>
-
-	Transitioning an element with opacity in and out of display:none:
+	The 'opacity' of an element can be transitioned when the element changes
+	to or from ''display: none'':
 
 	<pre class=lang-css>
 	#target {
@@ -258,16 +258,17 @@ because they lack a [=before-change style=].
 	</pre>
 
 	The display is transitioning to allow for an opacity transition before
-	flipping from <code>display:block</code> to <code>display:none</code>.
+	flipping from ''display:block'' to ''display:none''.
 
-	Setting ''@starting-style'' opacity to 0 means the element will
-	transition opacity from 0 to 1 when inserted into the tree or when
-	the <code>hidden</code> class flips ''display'' from
-	''display/none''</code> to ''display/block'' as the target element does
-	not already have a [=before-change style=] in those cases.
+	Specifying ''opacity: 0'' in the ''@starting-style'' rule means the
+	element will transition opacity from ''0'' to ''1'' when inserted into
+	the tree or when the <code>hidden</code> class flips 'display' from
+	''display/none'' to ''display/block'' as the target element does not
+	already have a [=before-change style=] in those cases.
 
-	The opacity set to 0 in <code>#target.hidden</code> is for transitioning
-	opacity from 1 to 0 when the <code>hidden</code> class is added.
+	Specifying ''opacity: 0'' in the <code>#target.hidden</code> rule makes
+	'opacity' transition from ''1'' to ''0'' when the <code>hidden</code>
+	class is added.
 </div>
 
 Global, name-defining at-rules such as ''@keyframes'', ''@font-face'', and
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
 incremented) <a>current transition generation</a> is stored as the
 transition's <dfn>transition generation</dfn>.
 
+## Defining [=before-change style=]: the ''@starting-style'' rule
+
+Transitions can only start during a [=style change event=] for elements which
+has a defined [=before-change style=] established by the previous
+[=style change event=], which means that newly inserted elements or elements
+that change from not [=being rendered=] to being rendered do not start
+transitions. Still, doing transitions starting from the initial rendering is a
+valid use case.
+
+The <dfn at-rule id="at-ruledef-starting-style">@starting-style</dfn> rule is a
+conditional group rule with the implicit condition that it matches if the
+previous [=style change event=] did not establish a [=before-change style=] for
+the element whose styles are being computed. This is used to establish a
+[=before-change style=] for the current [=style change event=].
+
+If any rules in ''@starting-style'' are applied, the resulting style must be
+used as the [=before-change style=] for the current [=style change event=] and
+the computed styles must be re-computed without ''@starting-style'' rules
+applied.
+
+<div class=example>
+
+  Transitioning background-color of an h1 element from transparent to green
+  from initial rendering:
+
+  <pre class=lang-css>
+  h1 {
+    transition: background-color 1.5s;
+    background-color: green;
+  }
+  @starting-style {
+    h1 {
+      background-color: transparent;
+    }
+  }
+  </pre>
+
+  Conditional rules can be used with CSS Nesting:
+
+  <pre class=lang-css>
+  h1 {
+    transition: background-color 1.5s;
+    background-color: green;
+    @starting-style {
+      background-color: transparent;
+    }
+  }
+  </pre>
+</div>
+
+Global, name-defining at-rules such as ''@keyframes'', ''@font-face'', and
+''@layer'' are not constrained by ''@starting-style''.
+
+### The <code>CSSStartingStyleRule</code> interface
+
+The {{CSSStartingStyleRule}} interface represents a ''@starting-style'' rule.
+
+<pre class='idl'>
+[Exposed=Window]
+interface CSSStartingStyleRule : CSSConditionRule {
+};
+</pre>
+
+<dl class='idl-attributes'>
+  <dt><code>conditionText</code> of type <code>CSSOMString</code>
+      (CSSStartingStyle-specific definition for attribute on CSSConditionRule)
+  <dd>The <code>conditionText</code> attribute (defined on the
+      <code>CSSConditionRule</code> parent rule), on getting, must return the
+      empty string.
+</dl>
+
+
 # Application of transitions # {#application}
 
 ## Animation composite order ## {#animation-composite-order}
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -180,27 +180,32 @@ elements that change from not [=being rendered=] to being rendered. To allow
 for that, this specification introduces ''@starting-style''.
 
 The <dfn at-rule id="at-ruledef-starting-style">@starting-style</dfn> rule is a
-grouping rule where the rules inside it apply if the previous
-[=style change event=] did not establish a [=before-change style=] for the
-element whose styles are being computed. This is used to establish a
-[=before-change style=] for the current [=style change event=].
+grouping rule where the rules inside it is used to establish styles to
+transition from if the previous [=style change event=] did not establish a
+[=before-change style=] for the element whose styles are being computed.
 
 Note: This means that ''@starting-style'' rules only apply to some elements
 during a computed style update, namely elements that were not rendered or part
 of the DOM during the previous [=style change event=].
 
-If any style rules in ''@starting-style'' are applied, the resulting style must
-be used as the [=before-change style=] for the current [=style change event=].
-Style rules in ''@starting-style'' do not apply directly to
-[=after-change style=]. Thus, the presence of matching rules in
-''@starting-style'' can cause transitions to occur on elements that otherwise
-could not have transitions because they lack a before-change style.
+Define <dfn>starting style</dfn> for an element as the [=after-change style=]
+with ''@starting-style'' rules applied in addition. If an element does not have
+a [=before-change style=] for a given [=style change event=], the
+[=starting style=] is used instead of the [=before-change style=] to compare
+with the [=after-change style=] to start transitions
+([[css-transitions-1#starting]]).
 
-Note: Since ''@starting-style'' is applied to create a [=before-change style=]
-during the same [=style change event=] as it is being used for triggering
-transitions, it inherits from the parent element's [=after-change style=], even
-if the parent establishes a [=before-change style=] through
-''@starting-style''.
+The rules inside ''@starting-style'' cascade as any other grouping rules without
+introducing any new ordering to the cascade, which means rules inside
+''@starting-style'' do not necessarily win over those outside.
+
+Style rules in ''@starting-style'' do not apply to [=after-change style=].
+Thus, the presence of matching rules in ''@starting-style'' can cause
+transitions to occur on elements that otherwise could not have transitions
+because they lack a [=before-change style=].
+
+''starting style'' inherits from the parent's [=after-change style=] just like
+[=after-change style=] does.
 
 <div class=example>
 
@@ -257,7 +262,12 @@ if the parent establishes a [=before-change style=] through
 
 	Setting ''@starting-style'' opacity to 0 means the element will
 	transition opacity from 0 to 1 when inserted into the tree or when
-	flipping from <code>display:none</code> to <code>display:block</code>.
+	the <code>hidden</code> class flips ''display'' from
+	''display/none''</code> to ''display/block'' as the target element does
+	not already have a [=before-change style=] in those cases.
+
+	The opacity set to 0 in <code>#target.hidden</code> is for transitioning
+	opacity from 1 to 0 when the <code>hidden</code> class is added.
 </div>
 
 Global, name-defining at-rules such as ''@keyframes'', ''@font-face'', and
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -185,22 +185,22 @@ grouping rule where the rules inside it apply if the previous
 element whose styles are being computed. This is used to establish a
 [=before-change style=] for the current [=style change event=].
 
-Note that this means that ''@starting-style'' rules only applies to some
-elements during a computed style update, namely elements that were not rendered
-or part of the DOM during the previous [=style change event=].
+Note: This means that ''@starting-style'' rules only apply to some elements
+during a computed style update, namely elements that were not rendered or part
+of the DOM during the previous [=style change event=].
 
 If any style rules in ''@starting-style'' are applied, the resulting style must
 be used as the [=before-change style=] for the current [=style change event=].
 Style rules in ''@starting-style'' do not apply directly to
-[=after-change style=].
+[=after-change style=]. Thus, the presence of matching rules in
+''@starting-style'' can cause transitions to occur on elements that otherwise
+could not have transitions because they lack a before-change style.
 
-<div class="note">
-Since ''@starting-style'' is applied to create a [=before-change style=] during
-the same [=style change event=] as it is being used for triggering
+Note: Since ''@starting-style'' is applied to create a [=before-change style=]
+during the same [=style change event=] as it is being used for triggering
 transitions, it inherits from the parent element's [=after-change style=], even
 if the parent establishes a [=before-change style=] through
 ''@starting-style''.
-</div>
 
 <div class=example>
 
@@ -243,7 +243,6 @@ if the parent establishes a [=before-change style=] through
 		display: block;
 		opacity: 1;
 		@starting-style {
-			display: none;
 			opacity: 0;
 		}
 	}
@@ -252,6 +251,13 @@ if the parent establishes a [=before-change style=] through
 		opacity: 0;
 	}
 	</pre>
+
+	The display is transitioning to allow for an opacity transition before
+	flipping from <code>display:block</code> to <code>display:none</code>.
+
+	Setting ''@starting-style'' opacity to 0 means the element will
+	transition opacity from 0 to 1 when inserted into the tree or when
+	flipping from <code>display:none</code> to <code>display:block</code>.
 </div>
 
 Global, name-defining at-rules such as ''@keyframes'', ''@font-face'', and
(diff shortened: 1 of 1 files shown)


tabatkins:
I think the language about the global at-rules needs to be tightened up a bit. We don't have a good template text to work from yet, I'll see what I can put together.

Also, this uses a different (third!) indentation style from the rest of the document. Please use tabs for indentation, as is used elsewhere in the spec text.

lilles:
> I think the language about the global at-rules needs to be tightened up a bit. We don't have a good template text to work from yet, I'll see what I can put together.

I basically copied that from the container queries spec. Also, @scope needs the same specification. Wonder if we should have a common spec saying something about conditional rules that evaluate in an element context do not restrict those tree-scope-global rules.

> Also, this uses a different (third!) indentation style from the rest of the document. Please use tabs for indentation, as is used elsewhere in the spec text.

Using tabs in the second commit. How does it look now?

tabatkins:
> Wonder if we should have a common spec saying something about conditional rules that evaluate in an element context do not restrict those tree-scope-global rules.

Yeah, we should, now that it's getting carried to multiple places. Don't worry about it for here, tho.

> How does it look now?

r+

tabatkins:
(But since I'm not a Transitions editor I'll leave it to @birtles or @dbaron to actually merge.)

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
...
+		background-color: green;
+		@starting-style {
+			background-color: transparent;
+		}
+	}
+	</pre>
+</div>
+
+Global, name-defining at-rules such as ''@keyframes'', ''@font-face'', and
+''@layer'' are not constrained by ''@starting-style''.
```
dbaron, line 221: I think this could be clearer:  "are not constrained" is a little confusing.  I think it means that they're allowed inside `@starting-style`, and if they're present, it's just as if they were outside of `@starting-style`.

(Is that actually the desired behavior, or would it be better to disallow them?)

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
...
+
+### The <code>CSSStartingStyleRule</code> interface
+
+The {{CSSStartingStyleRule}} interface represents a ''@starting-style'' rule.
+
+<pre class='idl'>
+[Exposed=Window]
+interface CSSStartingStyleRule : CSSConditionRule {
```
dbaron, line 229: Would it make more sense to inherit directly from `CSSGroupingRule` and thus not have `conditionText`?

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
...
 transition's <dfn>transition generation</dfn>.
 
+## Defining [=before-change style=]: the ''@starting-style'' rule
+
+Transitions can only start during a [=style change event=] for elements which
+has a defined [=before-change style=] established by the previous
+[=style change event=], which means that newly inserted elements or elements
+that change from not [=being rendered=] to being rendered do not start
+transitions. Still, doing transitions starting from the initial rendering is a
+valid use case.
```
dbaron, line 177: This is making factual statements that become untrue as a result of the addition of `@starting-style`.  I think it would be useful to more clearly separate (a) a summary of the rules as of CSS-TRANSITIONS-1 and (b) a description of how this specification (a delta spec) modifies those rules.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
...
+The <dfn at-rule id="at-ruledef-starting-style">@starting-style</dfn> rule is a
+conditional group rule with the implicit condition that it matches if the
+previous [=style change event=] did not establish a [=before-change style=] for
+the element whose styles are being computed. This is used to establish a
+[=before-change style=] for the current [=style change event=].
+
+If any rules in ''@starting-style'' are applied, the resulting style must be
+used as the [=before-change style=] for the current [=style change event=] and
+the computed styles must be re-computed without ''@starting-style'' rules
+applied.
```
dbaron, line 188: I think it would be clearer to replace the part of this sentence after the "and" with a separate sentence saying that `@starting-style` does not affect the after-change style.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
...
+[=style change event=], which means that newly inserted elements or elements
+that change from not [=being rendered=] to being rendered do not start
+transitions. Still, doing transitions starting from the initial rendering is a
+valid use case.
+
+The <dfn at-rule id="at-ruledef-starting-style">@starting-style</dfn> rule is a
+conditional group rule with the implicit condition that it matches if the
+previous [=style change event=] did not establish a [=before-change style=] for
+the element whose styles are being computed. This is used to establish a
+[=before-change style=] for the current [=style change event=].
```
dbaron, line 183: I think it's probably worth adding an informative note here that this means that a style computation for an element that does use `@starting-style` rules usually inherits from (maybe not directly, but somewhere in its ancestor chain) from an element that doesn't use `@starting-style` rules because it had a before-change style already.  (I assume this is how we want this to work?)

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
...
+	<pre class=lang-css>
+	h1 {
+		transition: background-color 1.5s;
+		background-color: green;
+		@starting-style {
+			background-color: transparent;
+		}
+	}
+	</pre>
+</div>
```
dbaron, line 218: I think it would be useful to also have a more complex example that does a color transition both when an element changes from `display:none` to `display:block` and also reverses that color transition when it changes from `display: block` to `display:none`.  This would involve transitions on both the relevant color property and on `display`, and involve putting the `transparent` both in the `@starting-style` and in the rule that sets `display: none`.  I think this is useful because it shows what is likely to be a very common usage pattern.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,78 @@ Each time a new transition is generated, the current value of the (already
...
+
+Transitions can only start during a [=style change event=] for elements which
+has a defined [=before-change style=] established by the previous
+[=style change event=], which means that newly inserted elements or elements
+that change from not [=being rendered=] to being rendered do not start
+transitions. Still, doing transitions starting from the initial rendering is a
+valid use case.
+
+The <dfn at-rule id="at-ruledef-starting-style">@starting-style</dfn> rule is a
+conditional group rule with the implicit condition that it matches if the
```
dbaron, line 180: This "if" is doing a lot of work, and it might be worth explaining a little more clearly that this rule matches for some style computations but not others, across a set of style computations happening at the same time.

css-transitions-2/Overview.bs
lilles: That sounds reasonable to me.

css-transitions-2/Overview.bs
lilles: The reasoning for allowing them is that it is consistent with the specs for @container and @scope. If we wanted to disallow them, it would make sense to change the @container and @scope specs too?

css-transitions-2/Overview.bs
lilles: Yes, and also that the before-change style that is set from @starting-style actually inherits from the after-change style of the parent, even if the parent has @starting-style applied. Does that make sense?

css-transitions-2/Overview.bs
lilles: Not sure exactly how you'd want it. Tried to improve it.

css-transitions-2/Overview.bs
lilles: Done

css-transitions-2/Overview.bs
lilles: Tried a change to the wording now.

css-transitions-2/Overview.bs
lilles: Modified.

css-transitions-2/Overview.bs
```diff
@@ -169,23 +169,38 @@ transition's <dfn>transition generation</dfn>.
...
+grouping rule where the rules inside it apply if the previous
+[=style change event=] did not establish a [=before-change style=] for the
+element whose styles are being computed. This is used to establish a
 [=before-change style=] for the current [=style change event=].
 
-If any rules in ''@starting-style'' are applied, the resulting style must be
-used as the [=before-change style=] for the current [=style change event=] and
-the computed styles must be re-computed without ''@starting-style'' rules
-applied.
+Note that this means that ''@starting-style'' rules only applies to some
```
dbaron, line 188: applies -> apply

Also, this (one) paragraph should probably also be a `<div class=note>`... although it's possible bikeshed lets you do that by just starting it with "Note:".  (If it does, maybe do that to the paragraph two paragraphs later as well.)

css-transitions-2/Overview.bs
```diff
@@ -169,23 +169,38 @@ transition's <dfn>transition generation</dfn>.
...
+
+If any style rules in ''@starting-style'' are applied, the resulting style must
+be used as the [=before-change style=] for the current [=style change event=].
+Style rules in ''@starting-style'' do not apply directly to
+[=after-change style=].
+
+<div class="note">
+Since ''@starting-style'' is applied to create a [=before-change style=] during
+the same [=style change event=] as it is being used for triggering
+transitions, it inherits from the parent element's [=after-change style=], even
```
dbaron, line 200: I don't think before-change style inheriting from parent's after-change style is correct per the current css-transitions-1 spec... although it's possible there are unlanded changes that @andruud was working on that would make it correct??  (But I think the historic change was perhaps in the other direction?)

css-transitions-2/Overview.bs
```diff
@@ -169,23 +169,38 @@ transition's <dfn>transition generation</dfn>.
...
-If any rules in ''@starting-style'' are applied, the resulting style must be
-used as the [=before-change style=] for the current [=style change event=] and
-the computed styles must be re-computed without ''@starting-style'' rules
-applied.
+Note that this means that ''@starting-style'' rules only applies to some
+elements during a computed style update, namely elements that were not rendered
+or part of the DOM during the previous [=style change event=].
+
+If any style rules in ''@starting-style'' are applied, the resulting style must
+be used as the [=before-change style=] for the current [=style change event=].
```
dbaron, line 193: I think it's worth saying explicitly that the presence of matching rules in a `@starting-style` rule causes transitions to occur on elements that otherwise couldn't have transitions because they lack a before-change style.

css-transitions-2/Overview.bs
dbaron: I don't think that's correct (see comment below).

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,108 @@ Each time a new transition is generated, the current value of the (already
...
+		@starting-style {
+			background-color: transparent;
+		}
+	}
+	</pre>
+</div>
+
+<div class=example>
+
+	Transitioning an element with opacity in and out of display:none:
```
dbaron, line 237: Perhaps it would be useful to have a few sentences of explanation on why there are two `display: none; opacity: 0;` pieces and what each one does.

css-transitions-2/Overview.bs
lilles: This is about computing before-change style in the style change event as the after-change style, which is a new thing, so I don't understand how that would strictly be in conflict with existing specs?

You would not be able to completely create the before-change style as of the previous style change event as the state of the elements involved has already changed (for selector matching, for instance).

dbaron: OK... I agree it is somewhat different.  That said, I think there were good reasons (related to container queries) for the setup that I *think* is reflected in the current spec, and I don't completely remember what they are, and I hope @andruud does.

I also don't understand what would be hard about creating the before-change style for the parent -- presumably if the parent had needed to start a transition, it would have used that before-change style, so it should still be around.  Or am I missing something here?

lilles: That would generally require two computed style trees to be kept alive, the new and the old one. This is an implementation detail, but what happens in Blink for container queries is that we temporarily store the before-change style on the first resolve if we risk doing multiple style passes because of multiple layout passes for a query container so that we use the original before-change style and not update the before-change style for every layout pass. However, we only do that when inside a size container and the style may be affected by animations/transitions.

If @starting-style should inherit from before-change style from the previous style-change if it existed, consider the following case:

```html
<div id="a">
  <div id="b">
    <div id="c"></div>
  </div>
</div>
```

Say that #b was display:none in the previous style-change, and we have @starting-style rules for #c. The before-change style for #a is available from the previous style-change, but we have no idea what the style for #b would have been if it was created and stored for the previous style-change event.

Then, if the before-change style for #c should inherit the before-change style from #b, we have no better idea than to use the after-change style for #b as any state for #b may have changed since the last style-change event. We could inherit before-change style from #a and create two separate styles for #b, one before-change style, in case some descendant needs it to inherit from when starting transitions, and one for the after-change style. 

Now, I think that is a pretty big extra burden on an implementation, and still while inheritance may make the before-change style include more styles from the previous style-change event, it still will be heavily influenced by the current style-change event due to selectors matching the current DOM state.

css-transitions-2/Overview.bs
lilles: Done

css-transitions-2/Overview.bs
lilles: Done

css-transitions-2/Overview.bs
andruud: The unlanded change that I never managed to follow up on, stated roughly:

 - The _base style_ is the style as it would be without animations/transitions/SMIL, inheriting from the base style of the parent.
 - The before/after-change style is just the base style observed at different points in time. Hence "cross-inheritance" between before/after-change styles is indeed not a thing in this model.

The reason for proposing that model is that the current spec (if I recall correctly) defines the before-change style to be the before-change base style _plus_ animations/transitions/etc updated to the _current time_. Say we were animating a width for example, and had some CQs (ultimately) affected by that width. At the moment we're calculating the before-change style, the "change" has already happened. The entire DOM could have been mutated heavily. We are no longer able to evaluate the CQs in a world where that DOM mutation _didn't_ happen, because we don't keep around multiple worlds.

This was already a hard problem pre-CQ, because the spec required to keep around multiple "responsive" style worlds: e.g. a font-size animation would affect any em units found in the before-change (base) style (Blink never implemented this). With CQ this extends to all of layout, making it probably too hard to be ever worth considering.

---

@lilles' before-change style inheriting from an after-change style seemed really weird to me at first as well. But he had a good point that the starting-style is produced matching against a current (i.e. after-change) DOM.

Possibly we should avoid saying that @starting-style affects the before-style change itself, and instead spec that:

 - The _starting style_ is the before-change style if it exsts, otherwise produced from @starting-style and inheriting from whatever makes sense.
 - A transition is between the starting style and the after-change style.

dbaron: So I think the *intent* of the current model is that the updating of the before-change style doesn't involve any changes to rule matching -- it just involves taking the already-matched animations/transitions/etc. and updating the time-based parts of them to the current time.  Maybe that isn't stated explicitly, but it's possibly it simplifies the CQ problem.  Though I'm not sure if it's relevant here.

It is indeed true that in the case where the before-change style comes from `@starting-style` rather than from a previous style change event, it is based on matching against the current DOM rather than a previous DOM.  But I think that's fine.

Perhaps the spec should be clearer about this, but I don't think it's possible for an element to both (a) have a before-change style from a previous style change event *and* (b) have its after-change style inherit from a different element than its before-change style.  This is because (and I'm not sure the spec is clear about this) a change in position in the DOM is effectively a removal and re-addition to the document, which is equivalent to not being rendered in the previous style change event.

Given those two things (which you should feel free to disagree with if you think they're wrong -- although I do think they're correct), it's not clear to me why the argument about changes in the DOM is relevant to this question.  I think it's fundamentally about what should happen to the before-change style (in particular, what it inherits from) at the boundary where an element that was previously not being rendered is the child of an element that *was* previously rendered.

It's not clear to me whether introducing a new "starting style" term would help.  If introducing a new term makes things clearer, then doing so seems fine.  But I think the main thing I'm asking about here is what the correct behavior is rather than about how to explain it.

css-transitions-2/Overview.bs
dbaron: OK, but I think it's probably still worth explaining what the two `opacity: 0` declarations do -- and how one of them applies to the transition when `class=hidden` is removed and the other applies to the transition when `class=hidden` is added.

dbaron (commented):
Sorry for the delay getting back to you -- should be mostly back in my normal work location this week after visiting a bunch of folks in my team (3907km away) last week.

When you reply, I should catch the email -- but if you hit the "re-request review" button next to my username, that will slightly lower the risk that I miss it.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,114 @@ Each time a new transition is generated, the current value of the (already
...
+[=style change event=] did not establish a [=before-change style=] for the
+element whose styles are being computed. This is used to establish a
+[=before-change style=] for the current [=style change event=].
+
+Note: This means that ''@starting-style'' rules only apply to some elements
+during a computed style update, namely elements that were not rendered or part
+of the DOM during the previous [=style change event=].
+
+If any style rules in ''@starting-style'' are applied, the resulting style must
+be used as the [=before-change style=] for the current [=style change event=].
```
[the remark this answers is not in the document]
lilles: My intent is that #b should not transition color for that change. As I mentioned in another comment, that you disagreed with, my intent is that starting style for an element inherits from the after-change style of the parent, regardless of whether any @starting-style rules applied to any of the ancestors.

Perhaps that's not consistent with how before-change style behaves, but as mentioned in another comment, the main reason I want to do it that way is that otherwise we need to keep two full style trees during style recalc and create separate @starting-style based computed style  instances for all elements (that did not already have a before-change style) in case there is some arbitrary descendant that can start a transition.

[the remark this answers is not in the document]
lilles: Correction: we don't need to keep the whole tree alive, but the child needs to know what the before-change style of the parent is. That's currently generally lost in the current Blink implementation, apart from some container queries cases, but it still means that we need to compute two sets of computed styles for subtrees of an element that didn't have a before-change style and matches any @starting-style rules making it different than the after-change style.

The worst case for that is:

```css
html {
  color: orange;
  @starting-style {
    color: pink;
  }
}

#some-descendant {
  transition: color 1s;
}
```

We then need to compute two sets of computed styles for every element on the initial rendering in case there is a descendant that relies on the root element's starting style.

css-transitions-2/Overview.bs
lilles: The CQ problem is about multiple passes of computing styles happening during the same style change event because we can have multiple layout passes for a given container with different size for each pass, but for transitions it is an atomic event. Otherwise the number of passes for the layout engine would be observable through transitions. I don't think that is relevant for @starting-style as such.

I think @andruud's point about matching the current DOM was about attribute and state changes, not about parent/child mutations.

css-transitions-2/Overview.bs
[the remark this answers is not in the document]
lilles: @dbaron If I read you correctly, I think where we do not agree on the intent is about inheritance. I don't think that is covered by discussions and resolutions in the working group, so it probably makes sense to formulate this in a new github issue and involve the working group to get a resolution on that. Does that sound good to you?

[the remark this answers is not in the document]
dbaron: I think your comment at https://github.com/w3c/csswg-drafts/pull/8876#discussion_r1236644072 is convincing enough that I'm OK with your current approach to inheritance, assuming it's specified clearly enough (see my comments at the start of this thread at https://github.com/w3c/csswg-drafts/pull/8876#discussion_r1235303098 ).

css-transitions-2/Overview.bs
dbaron: OK, given the explanation in https://github.com/w3c/csswg-drafts/pull/8876#discussion_r1236644072 I'm OK with this approach.

css-transitions-2/Overview.bs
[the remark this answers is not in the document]
lilles: I've tried to rework the spec text now.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+of the DOM during the previous [=style change event=].
+
+Define <dfn>starting style</dfn> for an element as the [=after-change style=]
+with ''@starting-style'' rules applied in addition. If an element does not have
+a [=before-change style=] for a given [=style change event=], the
+[=starting style=] is used instead of the [=before-change style=] to compare
+with the [=after-change style=] to start transitions
+([[css-transitions-1#starting]]).
+
+The rules inside ''@starting-style'' cascade as any other grouping rules without
```
dbaron, line 198: I think this "grouping rules" should probably be "grouped style rules".  (I think the "grouping rule" refers to the `@starting-style` rule and not the rules inside of it.)

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+introducing any new ordering to the cascade, which means rules inside
+''@starting-style'' do not necessarily win over those outside.
+
+Style rules in ''@starting-style'' do not apply to [=after-change style=].
+Thus, the presence of matching rules in ''@starting-style'' can cause
+transitions to occur on elements that otherwise could not have transitions
+because they lack a [=before-change style=].
+
+''starting style'' inherits from the parent's [=after-change style=] just like
```
dbaron, line 207: This `''starting style''` should be `[=starting style=]`, I think.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+[=after-change style=] does.
+
+<div class=example>
+
+  Transitioning background-color of an h1 element from transparent to green
```
dbaron, line 212: Indent looks wrong here.  Also `h1` should be `<code>h1</code>` and `background-color` should be `'background-color'`.  And this should probably be a complete sentence, such as "The background-color of an h1 element can be transitioned from transparent to green when it is initially rendered:".

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+		@starting-style {
+			background-color: transparent;
+		}
+	}
+	</pre>
+</div>
+
+<div class=example>
+
+	Transitioning an element with opacity in and out of display:none:
```
dbaron, line 242: Like the previous example, this could probably be something like:

`The 'opacity' of an element can be transitioned when the element changes to or from ''display: none''.`

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+		}
+	}
+	#target.hidden {
+		display: none;
+		opacity: 0;
+	}
+	</pre>
+
+	The display is transitioning to allow for an opacity transition before
+	flipping from <code>display:block</code> to <code>display:none</code>.
```
dbaron, line 261: Both of these `code` elements should be written with `''...''` quotes.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+
+	Setting ''@starting-style'' opacity to 0 means the element will
+	transition opacity from 0 to 1 when inserted into the tree or when
+	the <code>hidden</code> class flips ''display'' from
+	''display/none''</code> to ''display/block'' as the target element does
```
dbaron, line 266: Remove this extra `</code>`.

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+	#target.hidden {
+		display: none;
+		opacity: 0;
+	}
+	</pre>
+
+	The display is transitioning to allow for an opacity transition before
+	flipping from <code>display:block</code> to <code>display:none</code>.
+
+	Setting ''@starting-style'' opacity to 0 means the element will
```
dbaron, line 263: This seems a little too short, and I'd also prefer to avoid the word "set", so I'd say: `Specifying ''opacity: 0'' in the ''@starting-style'' rule means`...

(Also note additional quoting of CSS properties and values.)

css-transitions-2/Overview.bs
```diff
@@ -167,6 +167,124 @@ Each time a new transition is generated, the current value of the (already
...
+	transition opacity from 0 to 1 when inserted into the tree or when
+	the <code>hidden</code> class flips ''display'' from
+	''display/none''</code> to ''display/block'' as the target element does
+	not already have a [=before-change style=] in those cases.
+
+	The opacity set to 0 in <code>#target.hidden</code> is for transitioning
+	opacity from 1 to 0 when the <code>hidden</code> class is added.
```
dbaron, line 270: Like the previous paragraph, I'd rewrite as `Specifying ''opacity: 0'' in the <code>#target.hidden</code> rule makes 'opacity' transition from ''1'' to ''0'' when`...

css-transitions-2/Overview.bs
dbaron: OK, mostly looks good, but I left a few comments.

dbaron (approved):
OK, thanks for all the iteration on this.  I think this mostly looks good now, but I've left a bunch of comments on details of wording and markup.  This looks good to me with those comments addressed.

PR #8896, opened 2023-06-01
Editorial (first pass)

fantasai:
Comments and explanations in the commit messages... feel free to ask me for explanations if there's anything here that isn't self-evident.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -12,7 +12,8 @@ Work Status: exploring
 Editor: Tab Atkins-Bittner, Google, http://xanthir.com/contact/, w3cid 42199
 Editor: Jake Archibald, Google, w3cid 76394
 Editor: Khushal Sagar, Google, w3cid 122787
-Abstract: This module defines the View Transition API, along with associated properties and pseudo-elements.
+Abstract: This module defines the View Transition API, along with associated properties and pseudo-elements,
+	which allows developers to create animated visual transitions representing changes in the document state.
 Markup Shorthands: css yes, markdown yes
 </pre>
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -705,31 +768,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
-	<div class=note>
-
-		This element is a replaced element that produced the visual representation of the “old” state taken from user-agent provided snapshots.
+### View Transition Old State Image: the ''::view-transition-old()'' pseudo-element ### {#::view-transition-old}
 
-		This is only ever a [=tree/child=] of a ''::view-transition-image-pair()'',
-		never has any [=tree/children=],
-		and is omitted if there's no “old” state to represent.
+	The <dfn>::view-transition-old()</dfn> [=pseudo-element=]
+	is an empty [=named view transition pseudo-element=]
```
khushalsagar, line 774: Does "empty" here imply that it has no children?

css-view-transitions-1/Overview.bs
```diff
@@ -705,31 +768,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
-	that is also a [=named view-transition pseudo-element=],
-	and [=tree/participates=] in a [=pseudo-element tree=].
-
-	It is selected from its [=ultimate originating element=], the [=document element=].
-
-	It is a [=replaced element=], with [=natural dimensions=] equal to the content's size.
-
-	Note: The image content is captured in [=capture the image=],
+	Note: The content and [=natural dimensions=] of the image are captured in [=capture the image=],
 	then set and updated in [=setup transition pseudo-elements=] and [=update pseudo-element styles=].
```
khushalsagar, line 787:
```suggestion
	then set in [=setup transition pseudo-elements=].
```

For the old pseudos, we capture and set it once. Only the new pseudos need an update since their image is live.

css-view-transitions-1/Overview.bs
```diff
@@ -464,7 +485,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 # CSS properties # {#css-properties}
 
-## 'view-transition-name' ## {#view-transition-name-prop}
+## Tagging Individually Transitioning Subtrees: the 'view-transition-name' property ## {#view-transition-name-prop}
```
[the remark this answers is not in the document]
fantasai: I think it still works.

css-view-transitions-1/Overview.bs
fantasai: yes, that was the intention

css-view-transitions-1/Overview.bs
```diff
@@ -476,27 +497,35 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+	An element so captured is animated independently of the rest of the page.
 
 	<dl dfn-type=value dfn-for=view-transition-name>
 		: <dfn>none</dfn>
-		:: The [=/element=] will not participate in a view transition.
+		:: The [=/element=] will not participate independently in a view transition.
 
 		: <dfn><<custom-ident>></dfn>
-		:: The [=/element=] can participate in a view transition,
+		:: The [=/element=] will participate in a view transition,
```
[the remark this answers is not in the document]
fantasai: I think it needs to be a bit more forceful than that, since *unless* some particular exceptions are triggered, this property actually causes that participation: that's fundamentally what it does.

We could add some comments about what can cause an element to be excluded, but I want to make it clear up front what the property does.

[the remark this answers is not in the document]
fantasai: Made some tweaks, lmk what you think.

css-view-transitions-1/Overview.bs
```diff
@@ -476,27 +498,40 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
-			as either an old or new [=/element=],
-			with a <dfn dfn for>view transition name</dfn> equal to the <<custom-ident>>'s value.
+		:: The [=/element=] participates independently in a view transition--
+			as either an old or new [=/element=]--
+			with the specified [=view transition name=].
 
-			Note: The value <css>none</css> is invalid as a <<custom-ident>>.
+			The value <css>none</css> is excluded from <<custom-ident>> here.
+
+			Note: If this name is not unique
```
khushalsagar, line 518: It might be better to link to the relevant algorithm here. Something like:

```
Note: The presence of a `view-transition-name` doesn't guarantee that an element will participate independently in a view-transition. The property is ignored if an element skips its contents, has more than one box fragment, or is not rendered.
Also, if this name is not unique ...
See capture the old state/capture the new state algorithm for the precise conditions.
```

khushalsagar (approved):

css-view-transitions-1/Overview.bs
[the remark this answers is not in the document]
khushalsagar: Left a suggestion for improving the note below.

css-view-transitions-1/Overview.bs
fantasai: Added a cross-reference, good idea.

PR #8901, opened 2023-06-01
[css-view-transitions-1][css-viewport] Resolve a few missing types

noamr:
Use explicit properties in the struct for old element styles, and export interactive-widgets

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -852,9 +852,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		: <dfn>old transform</dfn>
 		:: a <a data-xref-type="css-type">&lt;transform-function&gt;</a>, initially the [=identity transform function=].
 
-		: <dfn>old object-view-box</dfn>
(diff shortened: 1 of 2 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -833,28 +830,36 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+		: <dfn>old height</dfn>
+		:: an {{unrestricted double}}, initially zero.
 
-			Issue: The type of "image" needs to be linked or defined.
+		: <dfn>old transform</dfn>
+		:: a <a data-xref-type="css-type">&lt;transform-function&gt;</a>, initially the [=identity transform function=].
 
-		: <dfn>old styles</dfn>
-		:: a set of styles or null. Initially null.
+		: <dfn>old object-view-box</dfn>
```
khushalsagar, line 855: Actually, we should remove this. Based on the resolution at https://github.com/w3c/csswg-drafts/issues/8597, UA doesn't add a `object-view-box` style anymore. It's supposed to be done transparently by the implementation.

noamr: Right. removed all viewbox stuff.

khushalsagar (approved):

Issue #8542, opened 2023-03-07
[css-contain] content-visibility: auto visibility check timing needs details


emilio:
https://drafts.csswg.org/css-contain/#cv-notes has:

> The initial determination of visibility for [content-visibility: auto](https://drafts.csswg.org/css-contain/#propdef-content-visibility) must happen in the same frame that determined an existence of a new content-visibility: auto element.

When and how, exactly? This is not really implementable without more details. Since the visibility check is usually done with an intersection observer, but intersection observer callbacks happen async per spec (https://w3c.github.io/IntersectionObserver/#queue-intersection-observer-entry-algo), how is this supposed to happen in the same frame?

Does that mean that computing the intersections of the frame can change layout, and thus can change whether other stuff intersects? I hope not...

cc @chrishtr @asurkov @mrobinson @Loirooriol

Loirooriol:
@vmpstr How does Blink do that?

vmpstr:
It is only the initial observation that has to be synchronous. This is to avoid a flash of "skipped" content for onscreen elements with c-v auto. This does indeed mean that sizes can change and other elements may become intersecting... I'm unsure why this is bad other than the inefficiency it can introduce on some specifically crafted pages. The common case here is one repetition of style/layout.

In Blink, we use IntersectionObserver (internal class) which implements all the same behaviours as the spec for IntersectionObserver with two exceptions: the notification delivery happens synchronously, and the time at which we determine intersections is within the lifecycle (right before ResizeObserver steps). When we get the notification callback, we check if the context it affects has been notified before. (context is a class that implements content visibility logic). If it has been notified, we schedule a new animation frame at which to deliver this notification. If it has not been notified, then we determine the c-v state synchronously and, if the skipped state changed, repeat the lifecycle phases

emilio:
> I'm unsure why this is bad other than the inefficiency it can introduce on some specifically crafted pages. The common case here is one repetition of style/layout.

Well, depending on when this is defined to happen (which right now is "somewhere, somehow"), it can cause bogus `ResizeObserver` or `IntersectionObserver` notifications, for example (if we define it to happen at the regular "intersection observer" step).

It seems at the time blink does it prevents that, but it also causes other issues like scroll events not being able to see the unskipped content etc. Which is probably fine, if it is well defined.

[...]

PR #8904, opened 2023-06-01
[css-contain-2] Define proximity to the viewport

vmpstr:
[css-contain-2] Define proximity to the viewport

This doesn't have a resolution per se (and likely needs one), but the resolution in https://github.com/w3c/csswg-drafts/issues/8542#issuecomment-1542482803 led to the following PR https://github.com/whatwg/html/pull/9312. That PR references proximity to the viewport, since that's a needed part of the algorithm. However, it wasn't clear where to define those terms and it seems like css-contain-2 is the right spot

Also, not sure if this is better suited for css-contain-3 instead.

### DIFF
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -1654,7 +1654,7 @@ Suppressing An Element's Contents Entirely: the 'content-visibility' property {#
 		* The element's [=proximity to the viewport=] is not determined:
 			In this state, the computation to determine the element's [=proximity to the viewport=] has not been done.
 		
-		By default, the element's [=proximity to the viewport=] is not determined.
+		Initially, the element's [=proximity to the viewport=] is not determined.
 		The element switches away from not determined state at most once during its lifetime.
 		Only elements with ''content-visibility: auto'' determine their [=proximity to the viewport=].
 		However, once determined, the state can no longer become not determined.
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -1967,7 +1967,7 @@ Restrictions and Clarifications {#cv-notes}
 	of the frame that renders the effects of the change
 	have run.
 	Specifically, such changes will take effect between steps 13 and 14
-	of <a href="https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering">Update the Rendering</a> step of the Processing Model
+	of [=update the rendering=] step of the Processing Model
 	(between “run the animation frame callbacks”
 	and “run the update intersection observations steps”).
 
@@ -1975,7 +1975,7 @@ Restrictions and Clarifications {#cv-notes}
 		Determining the viewport intersection of the element
 		can be done with an internal version of an IntersectionObserver.
 		However, since the observations from this are dispatched at
-		step 14 of Update the Rendering,
+		step 14 of [=update the rendering=],
 		any changes to the [=skipped contents|skipped=] (and thus painted) state
 		will not be visible to the user until the next frame's processing.
 		For this reason, updating the [=skipped contents|skipped=] state,
@@ -1992,8 +1992,8 @@ Restrictions and Clarifications {#cv-notes}
 4. Elements with ''content-visibility: auto''
 	that have not determined [=proximity to the viewport=]
 	must determine their proximity to the viewport
-	in the next <a href="https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering">Update the Rendering</a> cycle. 
-	The effect of this determination must be reflected in the visual update which results from this Update the Rendering cycle.
+	in the next [=update the rendering=] cycle. 
+	The effect of this determination must be reflected in the visual update which results from this [=update the rendering=] cycle.
 	
 	ISSUE: Reference the HTML spec algorithm here, when it's available.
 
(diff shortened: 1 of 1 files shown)


tabatkins:
Hm, it says your proximity can never become undetermined, but is that true? If the element becomes `display: none`, wouldn't it go back to being undetermined?

vmpstr:
> Hm, it says your proximity can never become undetermined, but is that true? If the element becomes `display: none`, wouldn't it go back to being undetermined?

The way we currently have it, the visibility being determined (synchronously or not) is a property of an element, not its box. So yeah, currently it can never become undetermined for the lifetime of the element, but that's something we can change if it makes more sense for that. Although tying this to the box is likely more complicated -- kind of like what we did with contain-intrinsic-size auto last remembered size

css-contain-2/Overview.bs
```diff
@@ -1654,6 +1647,32 @@ Suppressing An Element's Contents Entirely: the 'content-visibility' property {#
...
 			even if the element itself is still off-screen.
 
+
+		* The element is <dfn>far away from the viewport</dfn>:
+			In this state, the element's proximity to the viewport has been computed and is not [=close to the viewport=].
+		
+		* The element's [=proximity to the viewport=] is not determined:
+			In this state, the computation to determine the element's [=proximity to the viewport=] has not been done.
+		
+		By default, the element's [=proximity to the viewport=] is not determined.
```
dbaron, line 1657: I wonder if "Initially" or "When an element is created" would be clearer than "By default"

css-contain-2/Overview.bs
```diff
@@ -2280,6 +2303,7 @@ This appendix is <em>informative</em>.
 Changes from <a href="https://www.w3.org/TR/2022/WD-css-contain-2-20220917/">2022-09-17 Working Draft</a>
 </h3>
 
+* defined [=proximity to the viewport=] and used it in the clarification 4.
```
dbaron, line 2306: What's "the clarification 4"?

css-contain-2/Overview.bs
```diff
@@ -1970,9 +1989,13 @@ Restrictions and Clarifications {#cv-notes}
...
 	</div>
 
-4. The initial determination of visibility for ''content-visibility: auto'' must
-    happen in the same frame that determined an existence of a new
-    ''content-visibility: auto'' element.
+4. Elements with ''content-visibility: auto''
+	that have not determined [=proximity to the viewport=]
+	must determine their proximity to the viewport
+	in the next <a href="https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering">Update the Rendering</a> cycle. 
+	The effect of this determination must be reflected in the next visual update.
```
dbaron, line 1996: This *seems* to be saying something different from what the old spec text was saying, although I'm not sure.  (It seems like it's saying it's deferred one update cycle now, whereas it was immediate before.)  Is a difference intended?

dbaron (commented):
Sorry for the delay getting to this.  I think this mostly looks good, modulo a few comments below.

But, like @tabatkins, I wonder if keeping the state determined when an element becomes `display:none` is the right thing to do.  It's possible other implementors may have feedback on what works here; maybe @emilio has an opinion?

css-contain-2/Overview.bs
vmpstr: "next" is ambiguous here I think. I've changed the wording a bit. The idea is that the element could be created at any point in time, the very next Update the Rendering should determine its proximity. The visual update that results from this "update the rendering" should include the effects of the proximity change to relevance to the user.

I'll update the text

css-contain-2/Overview.bs
vmpstr: There's a section called "Restrictions and Clarifications" which has some items listed, I'll update this to link to it.

css-contain-2/Overview.bs
vmpstr: Done.

css-contain-2/Overview.bs
```diff
@@ -1950,7 +1969,7 @@ Restrictions and Clarifications {#cv-notes}
 	Specifically, such changes will take effect between steps 13 and 14
 	of <a href="https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering">Update the Rendering</a> step of the Processing Model
```
dbaron, line 1970: possibly better to link to https://html.spec.whatwg.org/multipage/C#update-the-rendering since it would survive a change of the `multipage/` splitting.

css-contain-2/Overview.bs
```diff
@@ -1970,9 +1989,13 @@ Restrictions and Clarifications {#cv-notes}
...
 	</div>
 
-4. The initial determination of visibility for ''content-visibility: auto'' must
-    happen in the same frame that determined an existence of a new
-    ''content-visibility: auto'' element.
+4. Elements with ''content-visibility: auto''
+	that have not determined [=proximity to the viewport=]
+	must determine their proximity to the viewport
+	in the next <a href="https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering">Update the Rendering</a> cycle. 
+	The effect of this determination must be reflected in the visual update which results from this Update the Rendering cycle.
```
dbaron, line 1996: Probably good to link this "Update the Rendering" too.

And maybe it should use a bikeshed link-definition at the top rather than using `<a href=''>...</a>` syntax.

dbaron (approved):
A few minor comments, but I guess this looks good to me.  But probably good to give some other folks a chance to give feedback, particularly on the `display:none` question.

css-contain-2/Overview.bs
vmpstr: Turns out it's exported! I've changed all of these to `[=update the rendering=]`

Issue #8917, opened 2023-06-05
[css-view-transitions-1] Clarifications on stacking contexts


fantasai:
https://drafts.csswg.org/css-view-transitions/#view-transition-stacking-layer

> Its parent stacking context is the root stacking context.

“root stacking context” isn't a defined term in Appendix E.
Did you mean “root element’s stacking context” or something else?

> If the ::view-transition pseudo-element exists, a new stacking context is created for the document element and the document's top layer. The view-transition layer is a sibling of this stacking context.

It's not clear to me what structure you're creating here, is it

- root stacking context
  - document element stacking context
  - top layer stacking context
  - view transition stacking context

or 

- root stacking context
  - document element and top layer stacking context
  - view transition stacking context

?

> The view-transition layer paints after the stacking context for the document element and document's top layer. This includes the filters and effects that are applied to the document element.

Is this trying to say 
* a) the filters and effects apply to everything together, or
* b) they apply to the document element and top layer, and the view transition layer is above these filters and effects

?

fantasai:
Side note that we don't anywhere define that `::view-transition` paints into this newly defined layer?

khushalsagar:
We took the term "root stacking context" from the fullscreen spec. I realize now that its not well defined in that spec either : https://drafts.csswg.org/css-position-4/#document-top-layer:~:text=Its%20parent%20stacking%20context%20is%20the%20root%20stacking%20context.

Not sure if that concept should go into the top layer spec or in Appendix E. Latter makes more sense, its supposed to elaborately define stacking contexts. :)

> It's not clear to me what structure you're creating here

The second one:

- root stacking context
  - document element and top layer stacking context
  - view transition stacking context

Generally the stacking context for "document element and top layer stacking context" is the root of the stacking context tree but we create a new one during the transition. Since the root element's image comes from the root stacking context. Amongst other things, its related to the question below.

> Is this trying to say

Option b, they apply to document element and top layer but not to view transition layer. We paint the stacking context for "document element and top layer" into a canvas to generate the image for the root element and then feed that image into the view-transition-old/new elements.

It's an existing pattern to lift some effects (background color and effects like filter/opacity) from the root element onto the canvas which paints both the root element + top layer elements. So we have to capture the painting of this stacking context to make sure the image for the "root" is correct. It can be a bit odd that `view-transition-name` is on the root element while in practice it takes the output of the stacking context which includes the root element + top layer stuff. But its building on the same pattern we do for these effects.

[...]

PR #8927, opened 2023-06-07
[css-view-transitions-1] Clarify stacking context model #8917

fantasai:
See https://github.com/w3c/csswg-drafts/issues/8917 The revised definition looks less formal, but afaict it actually hooks up more properly.

THIS PATCH BUILDS ON https://github.com/w3c/csswg-drafts/pull/8918 in order to avoid merge conflicts, and therefore incorporates that entire PR. See last commit for this change.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -770,25 +770,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 ## View Transition Painting Order ## {#view-transition-stacking-layer}
 
 	This specification introduces a new stacking layer, the [=view transition layer=],
-	to the <a href="https://www.w3.org/TR/CSS22/zindex.html">Elaborate Description of Stacking Contexts</a>.
+	to the end of the painting order established in
+	<a href="https://www.w3.org/TR/CSS22/zindex.html">CSS2&sect;E Elaborate Description of Stacking Contexts</a>. [[!CSS2]]
 
-	The ''::view-transition'' pseudo-element generates a new stacking context
-	called the <dfn>view transition layer</dfn>
-	with the following characteristics:
-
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -770,25 +770,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
-		The [=view transition layer=] is a sibling of this stacking context.
-
-	- The [=view transition layer=] paints after the stacking context for the [=document element=] and [=Document/top layer=].
-		This includes the filters and effects that are applied to the [=document element=].
+	to the end of the painting order established in
+	<a href="https://www.w3.org/TR/CSS22/zindex.html">CSS2&sect;E Elaborate Description of Stacking Contexts</a>. [[!CSS2]]
+
+	The ''::view-transition'' pseudo-element generates a new stacking context,
+	called the <dfn>view transition layer</dfn>,
+	which paints after all other content of the document
```
khushalsagar, line 778: One point which I want to make sure is obvious is that the root and top layer elements don't paint during the regular DOM if they're "captured in a view transition", a.k.a, computed value of view-transition-name on the root element is not none.

The spec text here is supposed to cover that: https://drafts.csswg.org/css-view-transitions-1/#setup-transition-pseudo-elements-algorithm:~:text=The%20new%20element,until%20new%20exists. If you have any better ideas to include that here (even in a normative note), would be good!

khushalsagar (approved):

khushalsagar:
This PR had everything from the previous one but I assumed https://github.com/w3c/csswg-drafts/pull/8927/commits/dda58ee55eb5840f92581ff4cf12bea03fc086cd is the only new commit? I just reviewed this commit.

Issue #7814, opened 2022-09-30
[css-view-transitions-1] Should isolation and plus-lighter blending be applied conditionally


khushalsagar:
The default animation in shared element transitions can have either of the following trees per tag:

## Cross-fade
```
::page-transition-container
|_::page-transition-image-wrapper
    |_::page-transition-incoming-image
    |_::page-transition-outgoing-image
```

There is visual state for an element in both outgoing and incoming DOM. The 2 have a cross-fade and need `isolation: isolate` on the wrapper and `mix-blend-mode: plus-lighter` on both the images to the blending right.

## Enter animation
```
::page-transition-container
|_::page-transition-image-wrapper
    |_::page-transition-incoming-image
```

There is visual state only for an element in the incoming page which fades in.

## Exit animation
```
::page-transition-container
|_::page-transition-image-wrapper
    |_::page-transition-outgoing-image
```

There is visual state only for an element in the outgoing page which fades out.

It's unclear how best to handle `isolation` and `mix-blend-mode` for the 3 cases:

1. Technically we don't need the `isolation` and `mix-blend-mode` properties in enter/exit animations. These were also causing sub-optimal rendering (in Blink) because the fact that they are no-op isn't always optimized out. So we could omit adding them in these 2 modes.
2. Always add them to these elements to ensure developers see a consistent style. Optimize out their effect in the implementation.

Option 1 is always better perf wise though it can be optimized if needed. I'd err for that unless there is a good reason to have option 2.

jakearchibald:
Summary for CSSWG:

We use `mix-blend-mode: plus-lighter` on the old & new views, plus `isolation: isolate` on the image-pair, in order to create a true cross-fade.

Isolation isn't free performance-wise.

Some transition groups only have an old view, or only new view. This happens when an element with a particular `view-transition-name` only exists on one side of the transition. Isolation is unnecessary in these cases.

## Option 1: Apply isolation & blend mode styles only when both old & new views are present

This would be observable in `getComputedStyle`.

It would also result in a different rendering if the developer gives the image-pair a background color, since some contents will plus-lighter blend with that background, and others wouldn't. However, if you set a background on the image-pair, it like that you'd want to set blending to normal anyway.

We already apply lots of styles dynamically in a view transition, such as width, height, transform etc of transition groups.

## Option 2: Optimise internally for cases where isolation and blending is unnecessary

[...]

PR #8948, opened 2023-06-12
[css-view-transitions-1] Resolve open issues.

khushalsagar:
This change fixes the following issues inlined into the spec:

* The link for "paint order" doesn’t seem right. Is there a more canonical definition?

The DOM order of `::view-transition-group` pseudo-elements needs to match the paint order of the corresponding DOM elements. This is why elements are traversed in paint order when adding them to named elements. The definition of paint order used here is correct.

* Isolation and the dynamic setting of blending is only necessary...

This was resolved in https://github.com/w3c/csswg-drafts/issues/7814.

* Refactor this so the algorithm takes a set of elements that will be captured...

This was a problem because there was no way to refer to an element participating in a transition. Presence of view-transition-name was not enough. Now there's a bool on element to refer to this state.

* Specify what happens with mix-blend-mode.

Adds spec text to clarify this.

The issues remaining are related to web animations, will take those up in a follow up PR.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1284,9 +1284,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			and has a [=node document=] equal to to |document|,
 			in [paint order](https://drafts.csswg.org/css2/#painting-order):
 
-			Issue: The link for "paint order" doesn't seem right.
-				Is there a more canonical definition?
-
 			1. If any [=flat tree=] ancestor of this |element| [=skips its contents=], then [=continue=].
 
 			1. If |element| has more than one [=box fragment=], then [=continue=].
@@ -1341,9 +1338,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		1. [=list/For each=] |element| of every [=/element=] that is [=/connected=],
 			and has a [=node document=] equal to to |document|,
 			in [paint order](https://drafts.csswg.org/css2/#painting-order):
-
-			Issue: The link for "paint order" doesn't seem right.
-				Is there a more canonical definition?
 
 			1. If any [=flat tree=] ancestor of this |element| [=skips its contents=], then [=continue=].
 
@@ -1475,9 +1469,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 					Note: The above code example contains variables to be replaced.
 
 					Note: ''mix-blend-mode: plus-lighter'' ensures that the blending of identical pixels from the old and new images results in the same color value as those pixels, and achieves a "correct" cross-fade.
-
-					Issue: Isolation and the dynamic setting of blending is only necessary to get the right cross-fade between new and old image pixels.
-					Would it be simpler to always add it and try to optimize in the implementation?
 	</div>
 
 ## [=Call the update callback=] ## {#call-dom-update-callback-algorithm}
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1616,14 +1607,16 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
-
-			Issue: Refactor this so the algorithm takes a set of elements that will be captured. This centralizes the logic for deciding if an element should be included or not.
-
-			Issue: Specify what happens with 'mix-blend-mode'.
+		
+		* 'mix-blend-mode' on the element is ignored during the capture.
+		
+			Note: 'mix-blend-mode' defines how the element is composited into its parent stacking context.
+				Since the element skips painting into its stacking context, as defined above, 'mix-blend-mode' is a no-op.
+				The blending used to composite the image is defined by the 'mix-blend-mode' on the replaced pseudo-element drawing this image.
```
fantasai, line 1619: I think this should get its own commit.  I also don't think I understand what it's saying...

Let's say I have a view transition capturing the root and element foo, and foo is semi-transparent with a non-normal `mix-blend-mode`. And let's say I define a very basic animation, e.g. sliding the old root out to the right and element foo out to the left to reveal the new document underneath.

Wouldn't ignoring `mix-blend-mode` on the element mean that as soon as the transition starts, the blending of foo against its background would change, resulting in a visible change in the color of foo?

khushalsagar: Cool. Let me take this in a follow up. Lmk if you think we should resolve on an issue based on the clarification below.

> Wouldn't ignoring mix-blend-mode on the element mean that as soon as the transition starts, the blending of foo against its background would change, resulting in a visible change in the color of foo?

You're right about the change in blending here. We could make it work by copying over the `mix-blend-mode` from the element to the replaced pseudo where the image will be drawn. But it only works if the element's closest ancestor that is also being captured into an image is the root. For example, if you have a tree like this:

```
root
|_B mix-blend-mode: multiply;
   |_C mix-blend-mode: multiply;
```

The blending order in the DOM is : Blend(root, Blend(B, C). If B and C are being captured for a transition, then the pseudo tree is:

```
::view-transition-group(root)
|_::view-transition-image-pair(root)
    |_::view-transition-old(root)
    |_::view-transition-new(root)
::view-transition-group(B)
|_::view-transition-image-pair(B)
    |_::view-transition-old(B)
    |_::view-transition-new(B)
::view-transition-group(C)
|_::view-transition-image-pair(C)
    |_::view-transition-old(C)
    |_::view-transition-new(C)
```

root, B and C become siblings so the blending order is Blend(Blend(root, B), C) which is not the same as what's in the DOM.

We have the same problem with a bunch of other properties that depend on the DOM's hierarchy. For example, as soon as the transition starts you'll stop being clipped by an ancestor which is participating in a transition. [Nested transition groups](https://github.com/WICG/view-transitions/blob/main/explainer.md#nested-transition-groups) is the principled fix. Authors would opt into this mode in which case group pseudo for C will be nested inside group pseudo for B which itself will be nested inside group pseudo for root.

fantasai: What you're saying makes sense, but I'm not clear on what the right way forward is here. :) So let's open a new issue with your explanation and go from there. My first impression is that honoring mix-blend-mode on the pseudo-element is both the right thing to do here and also the right thing to do in the nested case, i.e. they should be consistent.

fantasai (approved):

css-view-transitions-1/Overview.bs
khushalsagar: Done. Issue is here: https://github.com/w3c/csswg-drafts/issues/8962.

Issue #6530, opened 2021-08-20
[css-animations-2] Should the initial value for animation-duration be auto?


flackr:
[css-animations-1](https://drafts.csswg.org/css-animations/#animation-duration) defines the initial value of animation-duration to be 0s. However, the web-animations specification [allows for the 'auto' keyword](https://drafts.csswg.org/web-animations-1/#dom-effecttiming-duration) which is useful for scroll-animations and group animations where the duration should be [determined automatically](https://drafts.csswg.org/web-animations-2/#duration).

Since the computed value of 'auto' for regular animations is 0s, it seems to me that we could safely change the initial value of animation-duration to be 'auto', and this would allow using a scroll-timeline animation or possibly a group effect animation without having to explicitly set the auto duration.

@birtles WDYT?

flackr:
The one place I can think of where this would be web visible today is the getTiming() method if you were to call getAnimations() on a fill: forward css animation. e.g. https://jsbin.com/kawubir/edit?css,js,output

birtles:
If I understand correctly, this is about introducing an `auto` keyword to the CSS syntax for `animation-duration` and mapping that to an `'auto'` value in Web Animations. I think that sounds reasonable but, like you mentioned, there is some Web compat risk.

flackr:
@andruud noted this would also be exposed in `getComputedStyle` since we wouldn't have determined it to be 0 yet. This raises the risk of having it be the initial value as usage of `getComputedStyle` is much higher. Typically, developers would usually be reading the duration of elements animations which have durations, but if they did `let duration = parseFloat(getComputedStyle(elem).duration)` this would result in NaN for `"auto"` instead of 0 from `"0s"` :(.

Proposals:

1. We should definitely support specifying the `auto` value in css-animations for animation-duration.

2. I also think ergonomically it would be a much better initial value:
   * Time-based durations don't make sense for scroll linked animations and having to specify `auto` or a time value and rely on the [conversion from a time based animation to a progress based one](https://drafts.csswg.org/web-animations-2/#animations) is an awkward requirement, e.g. `animation: keyframes scroll()` should "just work".
   *  It results in the same duration for time-based animations.

   If we decide that we have to keep `0s` as the initial value we could update the [conversion from a time based animation to a progress based one](https://drafts.csswg.org/web-animations-2/#animations) to effectively convert `0s` animations to 100% as well, but this feels a bit more surprising especially if combined with proportion based delays.

@fantasai FYI

[...]

PR #8952, opened 2023-06-12
[css-animations-2] Make animation-duration:auto resolve to 0s for legacy compat #6530

fantasai:
See https://github.com/w3c/csswg-drafts/issues/6530


birtles:
Looks good to me but I should probably defer to someone more familiar with `animation-timeline` to review. (Also, I assume someone is making WPT for this case.)

andruud:
> I assume someone is making WPT 

Yes, I'll do that.

flackr:
Looks good to me. This resolves the most important blocking case that when neither `animation-timeline` nor `animation-duration` are set we get `0s` in the computed style.

PR #8954, opened 2023-06-13
[css-view-transitions-1][editorial] Minor tweaks to API prose

fantasai:


css-view-transitions-1/Overview.bs
```diff
@@ -934,18 +942,18 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+	The {{ViewTransition}} interface represents and controls a single same-document [=view transition=],
+	i.e. a transition where the starting and ending document are the same,
 	possibly with changes to the document's DOM structure.
 
 	<dl class="domintro non-normative">
 		: <code>{{ViewTransition|viewTransition}}.{{ViewTransition/updateCallbackDone}}</code>
 		:: A promise that fulfills when the promise returned by {{UpdateCallback|updateCallback}} fulfills, or rejects when it rejects.
 
-			The View Transition API wraps a DOM change and creates a transition.
+			(The View Transition API wraps a DOM change and creates a visual transition.
```
khushalsagar, line 953:
```suggestion
			The View Transition API wraps a DOM change and creates a visual transition.
```

Did you mean to put brackets here? Seems odd. I'm guessing the intent was to clarify that this text is just an explanation, nothing functional. So a note makes more sense.

khushalsagar (approved):
Just the one comment about brackets on some text which didn't look intentional. Assuming those go away or it turns into a note, LGTM.

css-view-transitions-1/Overview.bs
fantasai: Yes, it felt a little weird reading this paragraph because it doesn't feel like it's really part of the definition. A note might indeed make more sense

PR #8956, opened 2023-06-13
[css-view-transitions-1] Algorithms Section: various minor editorial fixes

fantasai:
It's probably easiest to review commit by commit. This PR does three things:
* Renames “transition suppressing rendering” to “rendering suppression for view transitions” which is longer but more understandable and also more clearly a noun. Another option might be “transition rendering suppression”... I'm not entirely sure of the nuance we're going for, but “transition suppressing rendering” definitely feels awkward.
* Renames “show view-transition root pseudo-element” to “show view transition tree”; because (afaict) fundamentally this is about displaying the entire tree, not just the root of it
* Makes a bunch of minor fixes to prepositions and markup

This PR builds on top of https://github.com/w3c/csswg-drafts/pull/8954

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -953,7 +953,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: The View Transition API wraps a DOM change and creates a visual transition.
 			However, sometimes you don't care about the success/failure of the transition animation,
 			you just want to know if and when the DOM change happens.
-			{{ViewTransition/updateCallbackDone}} is for that use-case.
+			{{ViewTransition/updateCallbackDone}} is for that use-case.)
 
 		: <code>{{ViewTransition|viewTransition}}.{{ViewTransition/ready}}</code>
 		:: A promise that fulfills once the pseudo-elements for the transition are created,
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1075,11 +1088,11 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 			Note: This is used to hold dynamic styles relating to transitions.
 
-		: <dfn>show view-transition root pseudo-element</dfn>
+		: <dfn>show view transition tree</dfn>
 		:: A boolean. Initially false.
 
 			When this is true, [=this=]'s [=active view transition=]'s [=ViewTransition/transition root pseudo-element=] renders as a child of [=this=]'s [=document element=],
-			and [=this=]'s [=document element=] is its [=originating element=].
+			with [=this=]'s [=document element=] is its [=originating element=].
```
khushalsagar, line 1095:
```suggestion
			with [=this=]'s [=document element=] as its [=originating element=].
```

Current sentence looks grammatically incorrect.

khushalsagar (approved):
The last commit was on another PR which I already reviewed so skipped adding the same comments here.

PR #8965, opened 2023-06-14
[web-animations-1] Improvements to term definitions and cross-linking

fantasai:
Going through the spec and cleaning up things that tripped me up...

Might be easier to review by commit.
Some of the commit messages have detailed explanations, hope that helps...

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -85,9 +85,9 @@ Former Editor: Alex Danilo 31960, Google, adanilo@google.com
 Former Editor: Tab Atkins 42199, Google, jackalmage@gmail.com
 Abstract: This specification defines a model for synchronization and timing
           of changes to the presentation of a Web page.
-          This specification also also defines an application programming interface
+          It also defines an application programming interface (API)
           for interacting with this model
-          and it is expected that further specifications will define
+          and it is expected that other specifications will define
           declarative means for exposing these features.
 
 !Participate: <a href="https://github.com/w3c/csswg-drafts/tree/main/web-animations-1">Fix the text through GitHub</a>
@@ -222,34 +222,38 @@ Introduction {#introduction}
 <em>This section is non-normative</em>
 
   This specification defines a model
+  (the <dfn export>Web Animations model</dfn>)
   for supporting animation and synchronization on the Web platform.
   It is intended that other specifications will build on this model
   and expose its features through declarative means.
+
   In addition, this specification also defines
   a programming interface to the model
+  (the <dfn export>Web Animations <abbr title="Application Programming Interface">API</abbr></dfn>,
+    defined in [[#programming-interface]])
   that may be implemented by user agents that provide support for scripting.
 
 Use cases {#use-cases}
 ---------
 
-  The Web Animations model is intended to provide the features necessary
+  The [=Web Animations model=] is intended to provide the features necessary
   for expressing CSS Transitions [[CSS-TRANSITIONS-1]],
   CSS Animations [[CSS-ANIMATIONS-1]], and
   SVG [[SVG11]].
-  As such, the use cases of Web Animations model is the union of use cases
-  for those three specifications.
+  As such, the use cases of the [=Web Animations model=] are
+  the union of use cases for those three specifications.
 
-  The use cases for the programming interface include the following:
+  The use cases for the [=Web Animations API=] include the following:
 
   : Inspecting running animations
   ::
       Often Web applications need to wait for certain animated effects to complete
       before updating some state.
-      The programming interface in this specification allows such applications
+      The [=Web Animations API=] allows such applications
       to wait for all currently running animation to complete,
       regardless of whether they are defined
       by CSS Transitions, CSS Animations, SVG animations,
-      or created directly using the programming interface.
+      or created directly using the [=Web Animations API=].
 
       <div class=example>
         <pre highlight=javascript>
@@ -296,7 +300,7 @@ Use cases {#use-cases}
       in terms of how they are represented in the CSS cascade
       and the performance optimizations that are possible,
       such as performing the animation on a separate thread.
-      Using the Web Animations programming interface,
+      Using the [=Web Animations API=],
       it is possible to create animations from script
       that have the same behavior and performance characteristics
       as declarative animations.
@@ -312,7 +316,7 @@ Use cases {#use-cases}
   ::
       In a complex application, it can be difficult
       to determine how an element arrived in its present state.
-      The Web Animations programming interface can be used
+      The [=Web Animations API=] can be used
       to inspect running animations to answer questions such as,
       “Why is the opacity of this element changing?”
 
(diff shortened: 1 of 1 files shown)


web-animations-1/Overview.bs
```diff
@@ -391,43 +395,44 @@ Relationship to other specifications {#relationship-to-other-specifications}
...
-  This specification begins by defining an abstract model for animation.
-  This is followed by a programming interface
+  This specification begins by defining an abstract model for animation,
+  the [=Web Animations model=].
+  This is followed by a programming interface, [=Web Animations API=],
   defined in terms of the abstract model.
-  The programming interface is defined in terms of the abstract model
-  and is only relevant to user agents that provide scripting support.
+  The programming interface is only relevant
+  to user agents that provide scripting support.
```
fantasai, line 435: Remove redundant part of the sentence (see previous sentence)

web-animations-1/Overview.bs
```diff
@@ -85,9 +85,9 @@ Former Editor: Alex Danilo 31960, Google, adanilo@google.com
 Former Editor: Tab Atkins 42199, Google, jackalmage@gmail.com
 Abstract: This specification defines a model for synchronization and timing
           of changes to the presentation of a Web page.
-          This specification also also defines an application programming interface
+          It also also defines an application programming interface (API)
```
birtles, line 88: Can we s/also also/also/ at the same time?

web-animations-1/Overview.bs
```diff
@@ -459,39 +459,41 @@ Specification conventions {#spec-conventions}
...
   The roles of these pieces are as follows:
 
-  : Timing model
+  : <dfn id="timing-model-dfn">timing model</dfn>
   ::
       Takes a moment in time and converts it to a proportional distance
       within a single iteration of an animation,
-      called the <em>iteration progress</em>.
-      The <em>iteration index</em> is also recorded
+      called the <dfn>iteration progress</dfn>.
```
birtles, line 471: The definition for iteration progress should remain in 4.11, not here.

web-animations-1/Overview.bs
```diff
@@ -459,39 +459,41 @@ Specification conventions {#spec-conventions}
...
+      The <dfn>iteration index</dfn> is also recorded
```
birtles, line 472: This should link to the definition of the current iteration.

birtles (changes requested):
I really want to keep the definition of iteration progress at the point where we define how it is calculated. As a Web developer and implementer it's much more helpful to know what the value is and how it is calculated, rather than its significance in the model.

The rest looks great.

web-animations-1/Overview.bs
```diff
@@ -629,13 +637,11 @@ Timing model characteristics {#timing-model-overview}
 Time values {#time-value-section}
 -----------
 
-  Timing is based on a hierarchy of time relationships between timing nodes.
-  Parent nodes provide timing information to their child nodes
-  in the form of [=time values=].
-
-  A <dfn>time value</dfn> is a real number which nominally represents
+  A <dfn>time value</dfn> in the [=Web Animations model=]
+  is a value that serves the purpose of synchronization.
```
birtles, line 641: "serves the purpose of synchronization" ?

I'm not sure what this is saying. Would you care to elaborate?

birtles:
Sorry for the dribs and drabs of comments here, I'm reviewing each commit separately (currently progress: 3 of 6).

birtles (commented):
This fourth commit is really great. Thank you.

birtles:
Ok, I've finished reviewing all the commits. They are really great. The details commit messages were very helpful. The only thing I feel strongly about is the definition of iteration progress. Everything else is fantastic. Thank you!

web-animations-1/Overview.bs
fantasai: Yes, lol. I can't believe I missed that.

web-animations-1/Overview.bs
fantasai: It's a bit weird, I agree; I pulled this definition from below, see 

> ```
> A <dfn export>timeline</dfn> provides a source of [=time values=]
>  for the purpose of synchronization.
> ```

The problem I'm trying to solve here is giving “time value” a definition that works equally well in WA2 where we have non-time-based “time values” ...

web-animations-1/Overview.bs
fantasai: OK, I didn't catch that they're the same concept. In that case, can we fold the names together and use “iteration index” throughout?

web-animations-1/Overview.bs
fantasai: Ok, so reviewing the change... I disagree. :) I don't mind adding a cross-reference, but I want the DFN here for two reasons:

1. It's the actual defining instance of the term. 4.11 says how to calculate it (sorta), but doesn't say what it *is*. 

2. For exactly that reason, 4.11 is confusing by itself. I land there and I don't understand what I'm looking at. I keep rewinding up the call stack, but I'm deep in the middle of a bunch of calculations for which I have no context. If it's a link back to the core definition, at least I can click on it to understand what's happening.

Looking through the spec, moving the DFN to 4.9 might make sense; it would require some tweaking of the prose to introduce an actual definition, but it seems to have appropriate context for one. Wdyt?

birtles: Ok, maybe I'm overreacting after encountering so many cases where CSS spec definitions just throw a term out there without actually defining it.

My understanding was this section was defining the timing model. The appearance of _iteration progress_ was just to aid that description. The actual definition of the iteration should come later.

I'm not sure about 4.9 (the current draft has 4.9 as "4.9 Direction control" which I'm pretty sure it not what you meant). I could see "4.5 Animation effects" as a possibility, since the iteration progress is a property of an animation effect.

Alternatively, we could just beef up the description of the iteration progress in 4.11. For example, for the _overall progress_ we have:

> The **overall progress** describes the number of iterations that have completed (including partial iterations) and is defined as follows:

web-animations-1/Overview.bs
birtles: It's exposed in the API as `currentIteration` so can we use "current iteration" instead? Or "current iteration index"?

web-animations-1/Overview.bs
birtles: Yeah, I can see that. It's tricky. It's almost like the animations equivalent of a length. Maybe what you have is fine.

web-animations-1/Overview.bs
fantasai: > Ok, maybe I'm overreacting after encountering so many cases where CSS spec definitions just throw a term out there without actually defining it.

That seems like an editorial error that we should fix. :)

> I'm not sure about 4.9 (the current draft has 4.9 as "4.9 Direction control" which I'm pretty sure it not what you meant).

I was referencing https://www.w3.org/TR/web-animations-1/#core-animation-effect-calculations so actually, I guess it's 4.8 :)

So I think what I'll do is leave the DFN where it is for now, and then maybe switch it up once I get something to anchor it in maybe around 4.8. (since I'm going top down, essentially)

web-animations-1/Overview.bs
fantasai: Sure, works for me. I'll try to layer that in then.

web-animations-1/Overview.bs
fantasai: Ok, we'll roll with this for now, and if anyone comes up with something better we can swap it in. :)

web-animations-1/Overview.bs
birtles: Ok, fair enough. If you wanted to drop in a note so we don't forget to move the definition, that would be nice. I'd like to make sure we define the term properly in a suitable place, instead of it just being a side-effect of another definition.

birtles (approved):

Issue #8030, opened 2022-11-07
[css-ui-4] Nested `<url-set>` should be invalid


cdoublev:
I think nested `<url-set>` cannot be valid because CSS Images forbids it.

> `<url-set>` is a limited version of `image-set()`, where only `<url>` and `<url-set>` can be used in the part of the `<image-set-option>` sub-production which would normally allow `<image>`.

https://drafts.csswg.org/css-ui-4/#typedef-cursor-url-set

> The `image-set()` function can not be nested inside of itself, either directly or indirectly (as an argument to another `<image>` type).

https://drafts.csswg.org/css-images-4/#funcdef-image-set

frivoal:
Agreed. Closed with https://github.com/w3c/csswg-drafts/commit/887025da9e52a8f19961fb9e298fda9961821f4b

PR #8969, opened 2023-06-15
[css-ui-4] <url-set> cannot be nested

### LINKED ISSUE #7388: [css-ui-4][css-images-4] `<string>` in `<image-set()>` and `<url-set>`
[CSS UI 4](https://drafts.csswg.org/css-ui-4/#typedef-cursor-url-set) defines `<url-set>` as *a limited version of `image-set()`, where only `<url>` and `<url-set>` can be used in the part of the `<image-set-option>` sub-production which would normally allow `<image>`*.

[CSS Images 4](https://drafts.csswg.org/css-images-4/#funcdef-image-set) defines that *each `<string>` inside `image-set()` represents a `<url>`*.

Is `image-set()` invalid if `<string>` does not represent a valid URL? If yes, I suggest *each `<string>` inside `image-set()` should represents a [valid URL string](https://url.spec.whatwg.org/#valid-url-string) otherwise the function is invalid*.

Is `<string>` valid in `<url-set>`?

I believe there is a small typo in the description of the part of `<image-set-option>` that corresponds to `type(<string>)`, in *an `<image-set-option>` like `url("picture.png") 1x format("image/jpeg")` is valid, [...]*: `format()` should be `type()`.

cdoublev:
Fixes #7388 and #8030.

I came back to these issues after several months and the corrections suggested in this PR definitely make sense to me.

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -673,8 +673,8 @@ Values have the following meanings:
 				support <<image>> which is a superset.
 
 				<dfn type>&lt;url-set></dfn> is a limited version of ''image-set()'',
-				where only <<url>> and <<url-set>> can be used
-				in the part of the <<image-set-option>> sub-production which would normally allow <<image>>.
+				where only <<url>> can be used
+				in the part of the <<image>> sub-production.
 
 				The UA must support the following image file formats:
 


frivoal:
Thanks for filing this. I agree with the problem you are aiming to fix. In the end I went with an slightly different editorial take (https://github.com/w3c/csswg-drafts/commit/887025da9e52a8f19961fb9e298fda9961821f4b), but thank you for getting me going.

Issue #7193, opened 2022-04-01
[css-text-4] Don't provide a language parameter for word-boundary-detection


r12a:
2.2.1. Detecting Word Boundaries: the [word-boundary-detection](https://www.w3.org/TR/css-text-4/#propdef-word-boundary-detection) property

>auto([<lang>](https://www.w3.org/TR/css-text-4/#typedef-word-boundary-detection-lang))
>    This value directs the user agent to perform language-specific content analysis to determine where to insert [virtual word boundaries](https://www.w3.org/TR/css-text-4/#virtual-word-boundary).
>
> [`<lang>`](https://www.w3.org/TR/css-text-4/#typedef-word-boundary-detection-lang) must be a valid CSS [`<ident>`](https://www.w3.org/TR/css-values-4/#typedef-ident) or [`<string>`](https://www.w3.org/TR/css3-values/#string-value). It represents an IETF BCP 47 language range (see [[BCP47]](https://www.w3.org/TR/css-text-4/#biblio-bcp47)). If the UA does not support word-boundary detection for all languages represented by the specified range, that specified value is invalid (and will cause the declaration to be ingored).

Fantasai provided some additional explanation about this feature, which explains why a range of language tags would be used:

> I think the idea of this parameter was that it enables word-boundary-detection for the specified language(s) only; other languages are not affected. What language the element is actually in is what determines the language used for the word-boundary-detection. For example, if I have a trilingual document in English, Japanese, and French, if I set word-boundary-detection: auto(ja) then it will enable detection for Japanese paragraphs only. If there are no Japanese paragraphs, it won't have any effect.

The i18n WG thinks that the choice of whether or not to apply the word boundary detection algorithm should be set by applying the word-boundary-detection styling to the relevant content. The language information used should be that provided by the `lang` attribute, and not supplied as a parameter with this property value.

We don't think the content author is able to guess what languages are supported by the user agent, so it doesn't seem useful to make them specify language in the property value.  We think that the approach currently described in the spec also requires the content author to have a level of understanding about language tagging that is too high (see the examples about Cantonese). 

We think that there should also be a simple recommendation that user agents SHOULD NOT apply a boundary detection algorithm to text in a language for which the algorithm is not defined (modulo decisions wrt dialect support).

FremyCompany:
(Question: would it be reasonable to enforce user agents who do not support a language to insert word boundaries chars everywhere a break is allowed normally, to use as a fallback?)

litherum:
[...]

PR #8974, opened 2023-06-16
[css-text-4] Replace `word-boundary-detection: auto()` with `word-break: phrase`

litherum:
From https://github.com/w3c/csswg-drafts/issues/7193, this replaces `word-boundary-detection: auto()` with `word-break: phrase`.

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -4187,7 +4187,7 @@ Breaking Rules for Letters: the 'word-break' property</h3>
 
 	<pre class="propdef">
 	Name: word-break
-	Value: normal | keep-all | break-all | break-word | auto
+	Value: normal | keep-all | break-all | break-word | phrase
 	Initial: normal
 	Applies to: text
 	Inherited: yes
@@ -4457,19 +4457,25 @@ Breaking Rules for Letters: the 'word-break' property</h3>
 			and is also useful for mixed-script text where CJK snippets are mixed
 			into another language that uses [=spaces=] for separation.
 
-		<dt><dfn>auto</dfn>
+		<dt><dfn>phrase</dfn>
 		<dd>
 			This value directs the user agent to perform language-specific content analysis
-			to determine where to insert [=virtual word boundaries=].
+			to prioritize keeping natural phrases (of multiple words) together
+			to determine [=soft wrap opportunities=]
+			and [=virtual word boundaries=].
 
             User agents may activate
             language-specific content analysis
             in response to user preferences.
             User agents with this behavior must do this
-            by setting the [=declared value=] of 'word-break' to ''word-break/auto''
+            by setting the [=declared value=] of 'word-break' to ''word-break/phrase''
             in the [=User Origin=].
             User agents that do not support the [=User Origin=]
             should use the [=User-Agent Origin=] instead.
+
+            Note: This means that web content can detect whether or not this feature is
+            enabled by calling getComputedStyle(), even if the user agent enabled this
+            feature by default.
 	</dl>
 
 	<div class=advisement>
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -992,7 +992,7 @@ Detecting Word Boundaries: the 'word-boundary-detection' property</h4>
 
 	<pre class="propdef">
 	Name: word-boundary-detection
-	Value: normal | manual | auto(<<lang>>)
+	Value: normal | manual
 	Initial: normal
 	Applies to: text
 	Inherited: yes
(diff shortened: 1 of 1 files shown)


css-text-4/Overview.bs
```diff
@@ -4603,8 +4456,35 @@ Breaking Rules for Letters: the 'word-break' property</h3>
 			(which uses [=spaces=] between words),
 			and is also useful for mixed-script text where CJK snippets are mixed
 			into another language that uses [=spaces=] for separation.
+
+		<dt><dfn>auto</dfn>
+		<dd>
+			This value directs the user agent to perform language-specific content analysis
```
kojiishi, line 4462: How about making the name more explicit? e.g., naming the value `phrase` and define it "to prioritize keeping natural phrases (of multiple words) together" (borrowed from the [text in LDML](https://www.unicode.org/reports/tr35/#:~:text=Prioritize%20keeping%20natural%20phrases%20(of%20multiple%20words)%20together%20when%20breaking%2C%20used%20in%20short%20text%20like%20title%20and%20headline))? Other value name is fine too, I just wonder if it's better to be more explicit.

In Thai, both word-boundary and phrase-boundary require language-specific content analysis. Today, by default, ICU uses dictionary-based (may switch to ML-based in future) content analysis to determine word boundaries. This behavior is for `word-break: normal`, not for this new value.

On the other hand, there are requests for [BudouX](https://github.com/google/budoux) to support Thai to break lines only at natural phrase (of multiple words) boundaries, not at every word boundary.

css-text-4/Overview.bs
```diff
@@ -4603,8 +4456,35 @@ Breaking Rules for Letters: the 'word-break' property</h3>
...
+			to determine where to insert [=virtual word boundaries=].
```
kojiishi, line 4463: Minor, but how about saying "to determine soft wrap opportunities and virtual word boundaries"?

I suppose you mean this value has full control to determine where soft wrap opportunities are, but "determine where to insert" may be a bit ambiguous about whether it can remove exisitng soft wrap opportunities or not.

css-text-4/Overview.bs
```diff
@@ -4603,8 +4456,35 @@ Breaking Rules for Letters: the 'word-break' property</h3>
...
+
+            User agents may activate
+            language-specific content analysis
+            in response to user preferences.
+            User agents with this behavior must do this
+            by setting the [=declared value=] of 'word-break' to ''word-break/auto''
+            in the [=User Origin=].
```
kojiishi, line 4470: Probably due to my English skill, sorry, but I failed to understand what this sentence emeans. Could you consider rephrasing?

kojiishi (commented):
Thank you for the PR, this looks great! Some comments:

css-text-4/Overview.bs
litherum: Oh, this just means "if a browser turns this on by default, they should make `getComputedStyle()` reflect the fact that the feature is enabled"

litherum: I'll add a note which should hopefully make this more clear.

kojiishi:
This looks great to me, thank you so much. Would you try to get a resolution at WG?

litherum:
> This looks great to me, thank you so much. Would you try to get a resolution at WG?

I added Agenda+ to https://github.com/w3c/csswg-drafts/issues/7193.

css-text-4/Overview.bs
```diff
@@ -4603,8 +4456,41 @@ Breaking Rules for Letters: the 'word-break' property</h3>
...
+		Manual analysis of the content can be more reliable than UA heuristics.
+		For best results, authors who can perform this analysis are encouraged to markup their documents
+		using <{wbr}> or U+200B
+		to exhaustively indicate word boundaries.
+
+		Authors who prepare their content in this manner
+		should not rely on the initial value, and
+		should explicitly specify ''word-break'' on the relevant parts of the content,
+		in order to override a potential ''word-break: auto''
+		in the [=User Origin=] or [=User-Agent Origin=].
```
fantasai, line 4491: I believe this is no longer relevant. (The previous model relied on specifying `word-break: keep-all` in order to work.)

css-text-4/Overview.bs
```diff
@@ -4603,8 +4456,41 @@ Breaking Rules for Letters: the 'word-break' property</h3>
...
+            Note: This means that web content can detect whether or not this feature is
+            enabled by calling getComputedStyle(), even if the user agent enabled this
+            feature by default.
 	</dl>
 
+	<div class=advisement>
+		Manual analysis of the content can be more reliable than UA heuristics.
+		For best results, authors who can perform this analysis are encouraged to markup their documents
+		using <{wbr}> or U+200B
+		to exhaustively indicate word boundaries.
```
fantasai, line 4485: This should probably add that UAs should respect such WBRs as more authoritative than their automatic detection.

css-text-4/Overview.bs
```diff
@@ -4603,8 +4456,41 @@ Breaking Rules for Letters: the 'word-break' property</h3>
...
+			to prioritize keeping natural phrases (of multiple words) together
+			to determine [=soft wrap opportunities=]
+			and [=virtual word boundaries=].
+
+            User agents may activate
+            language-specific content analysis
+            in response to user preferences.
+            User agents with this behavior must do this
+            by setting the [=declared value=] of 'word-break' to ''word-break/phrase''
+            in the [=User Origin=].
```
fantasai, line 4472: This bit should go into a separate section about accessibility tweaks that the UA can provide, not here in the core definition (where it interrupts understanding the property).

css-text-4/Overview.bs
```diff
@@ -4603,8 +4456,41 @@ Breaking Rules for Letters: the 'word-break' property</h3>
...
+            User agents that do not support the [=User Origin=]
+            should use the [=User-Agent Origin=] instead.
```
fantasai, line 4474: I don't think we need to say anything about what happens when a UA is non-conformant to css-cascade.

frivoal:
addressed by https://github.com/w3c/csswg-drafts/pull/9242

PR #8976, opened 2023-06-16
[css-color-5] move `color()` parameters description out of `<xyz-params>` bullet point

myakura:
https://drafts.csswg.org/css-color-5/#relative-color-function

`<ident>`, `<dashed-ident>`, `<number>`, `<percentage>`, and `<alpha-value>` should not be tied to `<xyz-params>` but are in the current draft.


svgeesus:
Sorry I just saw this. However the problem you point out has since been fixed as part of a [subsequent commit.](https://github.com/w3c/csswg-drafts/commit/20b67150ea8553bd11d33fd2ae463d08654f4216)

PR #8985, opened 2023-06-19
[web-animations-1] Restructure and Improve DFNs in Animation Effects section

fantasai:
Continuing from https://github.com/w3c/csswg-drafts/pull/8965 ... This refactors the Animation Effects section of the Web Animations model.

Have not yet worked through the progress calculation section (other than teasing out things which should go ahead of it), so haven't dealt with https://github.com/w3c/csswg-drafts/pull/8965#discussion_r1233105532, but it's almost 2am so submitting everything I've got for above it since it's a pretty coherent set of changes as-is. :)

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2504,8 +2504,12 @@ Animation effects {#animation-effects}
 
 ### Local time ### {#local-time-section}
 
-  The <dfn>local time</dfn> of an [=animation effect=] at a given moment
-  is based on the first matching condition from the following:
+  The <dfn>local time</dfn> of an [=animation effect=]
+  represents the [=animation/current time=]
+  provided by its [=associated animation=],
+  determining the progress of the [=animation effect=].
+  At a given moment,
+  it is based on the first matching condition below:
 
   <div class=switch>
 
@@ -2519,44 +2523,74 @@ Animation effects {#animation-effects}
 
   </div>
 
-### Time Spaces ### {#iteration-time-space}
+### Time spaces ### {#effect-time-spaces}
 
 <div class=informative-bg>
 <em>This section is non-normative</em>
 
   In Web Animations all times are relative to some point of reference.
-  These different points of reference produce different <em>time spaces</em>.
+  These different points of reference produce different <dfn>time spaces</dfn>.
 
   This can be compared to coordinate spaces
   as used in computer graphics.
-  The zero time of a time space is analogous
+  The zero time of a [=time space=] is analogous
   to the origin of a coordinate space.
 
-  We can describe animations that repeat
-  as establishing a new time space
-  each time the animation repeats:
-  the <em>iteration time space</em>.
+  Within the [=Web Animations model=]
+  we define progress of an [=animation effect=] on the basis of its [=active time=],
+  which is a time relative to the beginning of its [=active interval=]--
+  thus a time in the [=active time space=].
 
-  <em>Iteration time space</em> is a time space
-  whose zero time is the beginning
-  of an animation effect's current iteration.
-
-  Within the Web Animations model we also refer to [=active time=],
-  which is a time relative to the beginning of the active interval.
-  This time space, however, is internal to the model
+  Note: This time space is internal to the model
   and not exposed in the programming interface or in markup.
+
+  We can further describe animations that repeat
+  as establishing a new [=time space=]
+  each time the animation repeats:
+  the [=iteration time space=].
+
+  The [=animation effect=] thus translates the [=animation/current time=]
+  provided by the [=associated animation=]
+  through a series of [=time spaces=]:
+  * the <dfn noexport>local time space</dfn>,
+      relative to the [=start time=] of the [=associated animation=],
+      yielding the [=local time=]
+  * the <dfn noexport>active time space</dfn>,
+      relative to the start of the [=active interval=],
+      yielding the [=active time=]
+  * the <dfn noexport>iteration time space</dfn>,
+      relative to the start of the [=animation effect=]’s current iteration,
+      yielding the <dfn>iteration time</dfn>
 
   These time spaces are illustrated below.
 
   <figure>
     <img src="images/time-spaces.svg" width="600"
-         alt="A comparison of local time, active time, and iteration time.">
+         alt="A comparison of local time, active time, and iteration time.
+              Local time crosses zero at the start time of the animation,
+              and rises linearly.
+              Active time is zero until the start of the active interval,
+              at which point it rises linearly to the end time,
+              beyond which it remains at its maximum value.
+              Iteration time is likewise zero until the start of the active interval,
+              and rises linearly,
+              but restarts at zero at the start of every new iteration,
+              and remains at its halfway point from the end time onward.">
     <figcaption>
-      A comparison of [=local time=], [=active time=], and iteration time
+      A comparison of [=local time=], [=active time=], and [=iteration time=]
       for an animation with an [=iteration duration=] of 1s
       and an [=iteration count=] of 2.5.
+      The [=animation/start time=] defined by the [=animation=]
+      and [=animation effect/end time=] defined by the [=animation effect=]
+      are also annotated.
     </figcaption>
   </figure>
+
+  In addition to these time spaces
+  we can also refer to the <dfn noexport>document time space</dfn>,
+  which is time space of the [=time values=]
+  of the [=default document timeline=] of the {{Document}}
+  of the [=current global object=].
 
   Note: While the time spaces themselves are not bounded,
   Web Animations defines [=active time=] and the [=iteration progress=]
@@ -2566,12 +2600,6 @@ Animation effects {#animation-effects}
   the procedure for calculating the [=active time=]
   defined in [[#calculating-the-active-time]]
   will never return a negative value.
-
-  In addition to these time spaces
-  we can also refer to the <em>document time space</em>,
-  which is time space of the [=time values=]
-  of the [=default document timeline=] of the {{Document}}
-  of the [=current global object=].
 </div>
 
 ### The active interval, duration, and delays ### {#the-active-interval}
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2502,79 +2502,67 @@ Animation effects {#animation-effects}
   Subsequent levels of this specification
   will define further types of [=animation effects=].
 
-### The active interval ### {#the-active-interval}
+### The active interval, duration, and delays ### {#the-active-interval}
 
-<div class=informative-bg>
-
-  The period that an [=animation effect=] is scheduled to run
-  is called its [=active interval=].
+  [=Animation effects=] define an <dfn>active interval</dfn>,
+  which is the period of time during which the effect
+  is scheduled to produce its effect
+  (excepting [=fill modes=],
+  which apply outside the [=active interval=]).
   Each [=animation effect=] has only one such interval.
 
-  The lower bound of the [=active interval=] typically corresponds
-  to the [=start time=] of
-  the [=animation=] associated with this [=animation effect=],
-  but can be shifted by a [=start delay=] on the [=animation effect=].
-
-  The upper bound of the interval is determined
-  by the [=active duration=].
-
-  The relationship between
-  the [=start time=], [=start delay=], and [=active duration=]
-  is illustrated below.
-
-  <figure>
-    <img src="images/active-interval-examples.svg" width="600"
-        alt="Examples of the effect of the start dela
-             on the endpoints of the active interval">
-    <figcaption>
-      Examples of the effect of the [=start delay=]
-      on the endpoints of the [=active interval=].<br>
-      (a) An animation effect with no delay;
-      the [=start time=] and beginning of the [=active interval=]
-      are coincident.<br>
-      (b) An animation effect with a positive delay;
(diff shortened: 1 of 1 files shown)


fantasai:
I guess a somewhat relevant question here is whether we want to rename “timing function” to “easing function” while we're at it.

Obereigwe:
[Suicloudmining](https://bit.ly/SuiNetwork)

web-animations-1/Overview.bs
```diff
@@ -2465,123 +2465,203 @@ Animation effects {#animation-effects}
...
+### Local time ### {#local-time-section}
 
-  This specification defines a single type of [=animation effect=]:
-  [=keyframe effects=].
-  Subsequent levels of this specification
-  will define further types of [=animation effects=].
+  The <dfn>local time</dfn> of an [=animation effect=]
+  represents the [=animation/current time=]
+  provided by its [=associated animation=],
+  driving the progress of the [=animation effect=].
```
birtles, line 2510: I don't like the `diving the progress of the animation effect` part so much. I don't think we normally use dynamic language like "driving" so much, at least not in this part of the spec. Could we just drop this final clause? Or alternatively change it to something a bit more static like "...provided by its associated animation, which forms the input for determining the current progress of the animation effect"?

web-animations-1/Overview.bs
```diff
@@ -2465,123 +2465,203 @@ Animation effects {#animation-effects}
...
+  ::
+      the [=local time=] is the [=animation/current time=] of the [=animation=].
+
+  : Otherwise,
+  ::
+      the [=local time=] is [=unresolved=].
+
+  </div>
+
+### Time Spaces ### {#effect-time-spaces}
```
birtles, line 2526: Nit: capitalization

```suggestion
### Time spaces ### {#effect-time-spaces}
```

web-animations-1/Overview.bs
```diff
@@ -2465,123 +2465,203 @@ Animation effects {#animation-effects}
...
+  as used in computer graphics.
+  The zero time of a [=time space=] is analogous
+  to the origin of a coordinate space.
 
-  The upper bound of the interval is determined
-  by the [=active duration=].
+  Within the [=Web Animations model=]
+  we define progress of the [=animation effect=] in terms of the [=active time=],
+  which is a time relative to the beginning of the [=active interval=]--
+  thus a time in the [=active time space=].
```
birtles, line 2542: Progress is ultimately based on the transformed time space which is closer to iteration time than active time. Can we change this paragraph and the following note to "For example, we can define times relative to the active interval as being in active time space.  This time space, however, is internal to the model and not exposed in the programming interface or in markup." (I don't think the extra highlighting produced by making it a separate "Note:" is warranted here.)

web-animations-1/Overview.bs
```diff
@@ -2465,123 +2465,203 @@ Animation effects {#animation-effects}
...
-  is said to be <dfn lt="associated with an animation">associated</dfn>
-  with that animation.
+### Types of animation effects ### {#types-of-animation-effects}
+
+  This specification defines a single type of [=animation effect=]:
+  [=keyframe effects=].
+  Subsequent levels of this specification
+  will define further types of [=animation effects=].
+
+### Associated Animation and Timeline ### {#animation-effects-and-animations}
```
birtles, line 2490: Nit: capitalization

```suggestion
### Associated animation and timeline ### {#animation-effects-and-animations}
```

web-animations-1/Overview.bs
```diff
@@ -2465,123 +2465,203 @@ Animation effects {#animation-effects}
...
+### The active interval, duration, and delays ### {#the-active-interval}
+
+  [=Animation effects=] define an <dfn>active interval</dfn>,
   which is the period of time during which the effect
-  is scheduled to produce its effect--
-  with the exception of [=fill modes=],
-  which apply outside the [=active interval=].
+  is scheduled to produce its effect
+  (excepting [=fill modes=],
+  which apply outside the [=active interval=]).
```
birtles, line 2611: I wonder if this should be a parenthetical. I think the information is pretty central to the definition?

web-animations-1/Overview.bs
```diff
@@ -2465,123 +2465,203 @@ Animation effects {#animation-effects}
...
 
   The lower bound of the [=active interval=]
-  is defined by the [=start delay=].
-
-  The <dfn>start delay</dfn> of an [=animation effect=]
-  is a signed offset from the [=start time=] of the [=animation=]
-  with which the animation effect is associated.
+  by default corresponds to the [=start time=] of the [=associated animation=],
+  but can be shifted by the <dfn>start delay</dfn>,
+  which is a signed offset from the [=start time=] of the [=animation=].
```
birtles, line 2617: The original definition here is not quite right. We really should be defining the active interval in terms of local time (since nested effects have active intervals determined by _their_ local time).

So I perhaps something like: "The lower bound of the active interval is defined by the start delay which is a value in local time space defaulting to zero."

web-animations-1/Overview.bs
```diff
@@ -2465,123 +2465,203 @@ Animation effects {#animation-effects}
...
 
-  The length of the [=active interval=] is called the [=active duration=],
+  The length of the [=active interval=] is called the <dfn>active duration</dfn>,
   the calculation of which is defined in [[#calculating-the-active-duration]].
+  Its length determines the upper bound of the [=active interval=].
```
birtles, line 2621: Perhaps we should just bring the the `#calculating-the-active-duration` section inline here?

As an implementer, when I go to the definition of "active duration", I want to see how to calculate (implement) it. Not its significance in the model.

web-animations-1/Overview.bs
```diff
@@ -2788,23 +2868,70 @@ Animation effects {#animation-effects}
...
+      [=fill mode/forwards=] fill behavior is used.
+
+  </dl>
 
-  The normative definition of these modes
+  The normative effect of these modes
   is incorporated in the calculation of the [=active time=]
   in [[#calculating-the-active-time]].
 
+  Some examples of the these [=fill modes=] are illustrated below.
```
birtles, line 2914: Should this part onwards be in a non-normative block?

web-animations-1/Overview.bs
```diff
@@ -3154,8 +3165,88 @@ Repeating {#repeating}
...
+  can create a sense of momentum and produce a more natural effect.
+  The <a href="https://www.w3.org/TR/css-easing/">CSS Easing Functions Module</a>
+  defines [=timing functions=] for this purpose.
+  [[!CSS-EASING-1]]
+
+  [=Animation effects=] each have
+  an <dfn for="animation effect">effect timing function</dfn>,
+  which is a [=timing function=]
+  used to transform progress across the entirety of each iteration.
+  See [[#calculating-the-transformed-progress]].
```
birtles, line 3236: I really like what you've done with this section, including adding this definition. Thanks!

web-animations-1/Overview.bs
```diff
@@ -3154,8 +3165,88 @@ Repeating {#repeating}
...
+  is the [=linear timing function=].
+
+  Note: Specific [[#types-of-animation-effects|types of animation effects]]
+  may provide additional time transformations.
+  For example, [=keyframe effects=] can provide
+  [=keyframe-specific timing functions=]
+  that can apply a [=timing function=]
+  specifically between two adjacent keyframes.
+
+Calculating Progress {#core-animation-effect-calculations}
```
birtles, line 3248: Nit: capitalization

```suggestion
Calculating progress {#core-animation-effect-calculations}
```

web-animations-1/Overview.bs
```diff
@@ -3154,8 +3165,88 @@ Repeating {#repeating}
...
+--------------------
 
 ### Overview ### {#animation-effect-calculations-overview}
```
birtles, line 3251: I think the caption on the big diagram here needs to be updated.

It currently has:

> The first step, calculating the local time is described in § 4.6.3 Local time. Steps 2 to 4 in the diagram are described in the following sections. Steps 5 and 6 are described in § 4.7.6 Calculating the directed progress and § 4.7.7 Calculating the transformed progress respectively.

But I believe that steps 2 to 6 are now all covered in the following sections.

birtles (commented):
This looks really great. It makes a lot of sense to me.

I've left some comments about some of the earlier sections that you may not agree with so I'd like to hear your thoughts on those. Apart from that, it's great. Thank you!

web-animations-1/Overview.bs
fantasai: How about just changing "driving" to "determining"? :)

web-animations-1/Overview.bs
fantasai: Hah. I'm so used to using title case. :)

web-animations-1/Overview.bs
fantasai: How about tweaking the wording to “we define progress of an [=animation effect=] on the basis of its [=active time=]”?

Part of what I was trying to do here was build up a mental model of how these time spaces build on each other. Your example works fine for merely illustrating what a time space is. But doesn't help us understand how the various time spaces build on each other.

Also, keep in mind that the animation effect encompasses the concepts of iteration and iteration direction. So if you lay out the animation effect with all its iterations and time transformations onto a line, progress through the animation effect is progress through all that. It's not progress through the keyframe stack, which is indeed based on the transformed progress / iteration progress.

web-animations-1/Overview.bs
fantasai: I definitely debated doing that. :) The main downside to it is that we haven't yet defined repetitions. But if you're OK with inlining the calculation here and pointing people forward for the definitions, I'm happy to do it. I agree it feels more natural.

web-animations-1/Overview.bs
fantasai: It's pretty clearly described as an example, so I don't think it's necessary, unless you really want a green box here.

web-animations-1/Overview.bs
fantasai: OK done. That prompted me to restructure the active interval section also. See the latest two new commits.

web-animations-1/Overview.bs
birtles: Sounds good!

web-animations-1/Overview.bs
birtles: Yeah, my concern is that while "progress" is a general term I tend to equate it with "iteration progress" since that's the real star of the show. Iteration progress is what [`progress`](https://drafts.csswg.org/web-animations-1/#dom-computedeffecttiming-progress) in the API refers to and it's the value that ties the timing model and animation model together.

But in the specific section you're editing (the timing model section), we talk about different kinds of progress like "overall progress" and "transformed progress" so I think what you have makes sense. I think the tweak you proposed here ("...on the basis of...") makes it clearer still. Thanks!

web-animations-1/Overview.bs
birtles: Ok, I trust your instincts for this.

web-animations-1/Overview.bs
birtles: What about just:

> The lower bound of the animation interval is determined by its start delay which defaults to zero.

The normative definition of how all that works is the calculation of [active time](https://drafts.csswg.org/web-animations-1/#active-time).

If we really want to make it more descriptive, we could add an (effectively non-normative) explanation saying something like: "When the start delay is zero, the lower bound of the active interval coincides with the start time of the associated animation." or something to that effect. Then we'd simply drop that sentence in level 2.

web-animations-1/Overview.bs
```diff
@@ -2617,9 +2617,21 @@ Animation effects {#animation-effects}
   which is a signed offset from the [=start time=] of the [=animation=].
 
   The length of the [=active interval=] is called the <dfn>active duration</dfn>,
-  the calculation of which is defined in [[#calculating-the-active-duration]].
+  and is determined by the [=iteration duration=] and [=iteration count=].
```
birtles: Should this somehow refer to the below equation rather than it just appearing unannounced? e.g.

"...and is determined by the iteration duration and iteration count as defined below:"

web-animations-1/Overview.bs
birtles: Thanks! I left a comment on the first commit.

web-animations-1/Overview.bs
birtles: I am pretty partial to a bit of green but you know better than I do what's standard practice in these specs so if you think it's fine without, I'm sure it's fine.

web-animations-1/Overview.bs
fantasai: We'd need to introduce a concept of the animation effect start time (distinct from the animation start time) for Level 2. I think we *should* do that, but how about tackling that in a separate PR?

For reference, the text that's in this PR is

> The lower bound of the active interval by default corresponds to the start time of the associated animation, but can be shifted by the start delay, which is a signed offset from the start time of the animation.

And it is consolidated from the existing sentences:

> The lower bound of the active interval is defined by the start delay.
>
> The start delay of an animation effect is a signed offset from the start time of the animation with which the animation effect is associated.
> 
> The lower bound of the active interval typically corresponds to the start time of the animation associated with this animation effect but may be shifted by a start delay on the animation effect.

birtles: Ok, let's go with what you have and fix it in level 2.

birtles (approved):

Issue #8407, opened 2023-02-07
[css-contain] Behavior of slightly offscreen content having`content-visibility:auto` when UA margin around the viewport is 0


jihyerish:
In the description about the [element being relevant](https://w3c.github.io/csswg-drafts/css-contain-2/#relevant-to-the-user),
there is the recommended UA margin around the viewport to use for deciding whether the element is on-screen or not.

> NOTE: This margin is meant to allow the user agent to begin preparing for an element to be in the viewport soon. A margin of 50% is suggested as a reasonable default.

But if the **margin is set to 0%** and the content having `content-visibility:auto` is slightly offscreen,
then the content looks flashing.

Here is the test code shows the behavior above:
* Sample code [(Live demo)](https://jihyerish.github.io/demos/content-visibility/content-visibility-auto-relevant-content-margin.html)
```html
...
<style>
.outer {
    position: absolute;
    left: -1px;
    background: black;
    padding-right: 10px;
    padding-bottom: 10px;
}

.inner {
    content-visibility: auto;
    contain-intrinsic-size: 50px 50px;
    background: red;
}
</style>

<div class="outer">
  <div class="inner" id="target">  
  </div>
</div>

...
```
This happens because when IntersectionObserver for `content-visibility` triggers, the content isn't considered relevant at first. Then `contain-intrinsic-size` applies to the content and it makes the element relevant to the user.
Thus becoming relevant to user, the content has its size as 0x0 again. 
This causes the content not being considered relevant again and continue the loop, and as a result it looks like the content is flashing.

When I check this case for both chrome and firefox after setting the UA margin as 0,
the results are the same. The content is flashing.

> _NOTE: How to modify UA margin 0_
> * Chromium: 
>   - change `kViewportMarginPercentage` value as 0 in "third_party/blink/renderer/core/display_lock/display_lock_document_state.h"
> * Gecko:
>   - go to "about:config" in firefox
>   - change `layout.css.content-visibility-relevant-content-margin` value as 0

I guess the expected result of this case is that the content shows the stable layout (without flashing).
But maybe what happens now is the reasonable result.
What would be the appropriate behavior of this?

(This is related to gecko bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1804761)

Loirooriol:
[...]

PR #8989, opened 2023-06-20
[css-sizing-4] Add contain-intrinsic-size: auto none syntax

vmpstr:
[css-sizing-4] Add contain-intrinsic-size: auto none syntax

This addresses one of components of the following resolution:
https://github.com/w3c/csswg-drafts/issues/8407#issuecomment-1440466558


vmpstr:
Please take a look.

css-sizing-4/Overview.bs
```diff
@@ -569,6 +569,15 @@ Overriding Contained Intrinsic Sizes: the 'contain-intrinsic-*' properties</h3>
 			Otherwise,
 			its [=explicit intrinsic inner size=]
 			is the specified <<length>>.
+
+		: <dfn>auto none</dfn>
+		:: For a given axis, if the element has a [=last remembered size=],
+			and is currently [=skipping its contents=],
+			its [=explicit intrinsic inner size=] for the axis
+			is the [=last remembered size=].
```
Loirooriol, line 577: To avoid repeating this, it may be better to define the property with

```
Value: auto? [none | <<length>>]
```

and then just define the `auto [none | <<length>>]` cases together.

Loirooriol (approved):
LGTM, but you should probably get the approval of an editor.

css-sizing-4/Overview.bs
```diff
@@ -543,32 +543,22 @@ Overriding Contained Intrinsic Sizes: the 'contain-intrinsic-*' properties</h3>
...
-	Values are defined as:
-
 	<dl dfn-type=value dfn-for="contain-intrinsic-width, contain-intrinsic-height, contain-intrinsic-block-size, contain-intrinsic-inline-size, contain-intrinsic-size">
-		: <dfn>none</dfn>
-		:: The corresponding axis does not have an [=explicit intrinsic inner size=].
+		: <dfn>auto</dfn>? [ <dfn>none</dfn> | <<length>> ]
+		:: If ''contain-intrinsic-size/auto'' is specified
+			and the element has a [=last remembered size=]
+			and is currently [=skipping its contents=],
+			its [=explicit intrinsic inner size=] is the [=last remembered size=].
```
tabatkins, line 551:
```suggestion
			its [=explicit intrinsic inner size=] in the corresponding axis
			is the [=last remembered size=] in that axis.
			
			Note: This occurs, for example,
			when an element with ''content-visibility: auto''
			is off-screen.
```

css-sizing-4/Overview.bs
```diff
@@ -543,32 +543,22 @@ Overriding Contained Intrinsic Sizes: the 'contain-intrinsic-*' properties</h3>
...
 
-		: <dfn><<length>></dfn>
-		:: The corresponding axis has an [=explicit intrinsic inner size=] of the specified <<length>>.
+			Otherwise, if ''contain-intrinsic-size/none'' is specified,
+			the corresponding axis does not have an [=explicit intrinsic inner size=].
+
+			Otherwise,
+			the corresponding axis has an [=explicit intrinsic inner size=] of the specified <<length>>.
```
tabatkins, line 557:
```suggestion
			Otherwise,
			the corresponding axis either doesn't have an [=explicit intrinsic inner size=]
			(if ''contain-intrinsic-size/none'' is specified)
			or has an [=explicit intrinsic inner size=] of the specified <<length>>.
```

css-sizing-4/Overview.bs
```diff
@@ -628,6 +618,10 @@ Last Remembered Size</h4>
 	offering the performance benefits of [=size containment=]
 	while <em>probably</em> sizing accurately to its contents.
 
+	Note: ''contain-intrinsic-size: auto'' use in this text is a shorthand representation of
```
tabatkins, line 621: Prefer deleting this note and just changing  "The ''contain-intrinsic-size: auto'' value" above to instead say "The ''contain-intrinsic-size/auto'' keyword of 'contain-intrinsic-size'"

tabatkins (changes requested):
A few nits, otherwise R+

vmpstr:
PTAL

PR #8991, opened 2023-06-20
[scroll-animations-1] Time-based delays get convered to their respective proportions #7749

bramus:
There still was a leftover paragraph from an earlier version of the spec that mentioned time-based `animation-delay`s do not apply. This is not the case, as per https://github.com/w3c/csswg-drafts/issues/7749#issuecomment-1505563335 resolution.


bramus:
@fantasai @flackr Could you give this PR a review? Thanks!

flackr (approved):
Looks good to me.

PR #9005, opened 2023-06-23
[css-anchor-position-1] Add CSSOM interfaces

xiaochengh:


xiaochengh:
@tabatkins

Obereigwe:
[suicloudmining](https://bit.ly/SuiNetwork)

tabatkins:
Are we sure we want this rule to be readonly? The `@try` rule isn't readonly (you can adjust the style), so it feels a little weird for `@position-fallback` to be.

If it's not readonly, we can probably just inherit from `CSSGroupingRule` and add the `name` attribute.

xiaochengh:
OK, I'll make them mutable (aka I'll make a closer copy of `CSSKeyframesRule`).

> If it's not readonly, we can probably just inherit from CSSGroupingRule and add the name attribute.

I think inheritance from `CSSGroupingRule` is more appropriate if the child rules can be of any type.

If only one type of child rules are allowed, then it should inherit from `CSSRule` (again using `CSSKeyframesRule` as an example).

tabatkins:
>  (aka I'll make a closer copy of CSSKeyframesRule).

Nah CSSKeyframesRule has a *weird* set of operations. This is why I'm suggested CSSGroupingRule instead.

> I think inheritance from CSSGroupingRule is more appropriate if the child rules can be of any type.
> 
> If only one type of child rules are allowed, then it should inherit from CSSRule (again using CSSKeyframesRule as an example).

Nothing about CSSGroupingRule suggests it's specifically for multiple child rule types. The "insert a rule" operation checks if the rule type is allowed in that context.

If they'd been created in the opposite order, we've have CSSKeyframesRule inherit from CSSGroupingRule rather than use some weird bespoke operations.

PR #9014, opened 2023-06-28
[css-anchor-position-1] Change implicit anchor element example

josepharhar:
I am writing an HTML spec PR to define the anchor attribute, which sets the implicit anchor element. That PR doesn't have any references to or integrations with the popover attribute, and I got feedback that the implicit anchor element seems popover-specific due to the example: https://github.com/whatwg/html/pull/9144#discussion_r1244592176

This PR changes the example to talk about the anchor attribute rather than the popover API. I'm not sure if the popover API is supposed to set an implicit anchor element... maybe @xiaochengh knows?

### DIFF
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -160,9 +160,11 @@ a particular element is an <dfn>implicit anchor element</dfn>
 for a given positioned element.
 
 <p class=example>
-	For example, the HTML spec defines an "anchor" attribute which allows an
-	element to declare what element it is anchored to. This makes the declared
-	element the [=implicit anchor element=] for the element with the attribute.
+	For example, the HTML spec <a
+	href="https://whatpr.org/html/9144/dom.html#the-anchor-attribute">defines an
+	"anchor" attribute</a> which allows an element to declare what element it is
+	anchored to. This makes the declared element the [=implicit anchor element=]
+	for the element with the attribute.
 
 [=Implicit anchor elements=] can be referenced
 with the ''implicit'' keyword,
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -160,9 +160,9 @@ a particular element is an <dfn>implicit anchor element</dfn>
 for a given positioned element.
 
 <p class=example>
-	For example, the HTML spec defines an attribute which allows an element to
-	declare what element it is anchored to. This makes the declared element the
-	[=implicit anchor element=] for the element with the attribute.
+	For example, the HTML spec defines an "anchor" attribute which allows an
+	element to declare what element it is anchored to. This makes the declared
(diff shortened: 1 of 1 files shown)


josepharhar:
@tabatkins Want to take a look?

xiaochengh:
> I'm not sure if the popover API is supposed to set an implicit anchor element...

I don't think it should.

The current PR looks more appropriate.

css-anchor-position-1/Overview.bs
```diff
@@ -160,10 +160,9 @@ a particular element is an <dfn>implicit anchor element</dfn>
 for a given positioned element.
 
 <p class=example>
-	For example, the Popover API allows a popover
-	to declare what element is anchoring it.
-	This makes the declared element the [=implicit anchor element=]
-	for the popover.
+	For example, the HTML spec defines an "anchor" attribute which allows an
```
xiaochengh, line 163: Maybe add a link to the HTML spec?

xiaochengh: How about linking it to https://whatpr.org/html/9144/dom.html#the-anchor-attribute for now, and then change it when the HTML PR is merged?

josepharhar: done

css-anchor-position-1/Overview.bs
```diff
@@ -160,10 +160,11 @@ a particular element is an <dfn>implicit anchor element</dfn>
...
 <p class=example>
-	For example, the Popover API allows a popover
-	to declare what element is anchoring it.
-	This makes the declared element the [=implicit anchor element=]
-	for the popover.
+	For example, the HTML spec <a
+	href="https://whatpr.org/html/9144/dom.html#the-anchor-attribute">defines an
+	"anchor" attribute</a> which allows an element to declare what element it is
+	anchored to. This makes the declared element the [=implicit anchor element=]
+	for the element with the attribute.
```
tabatkins, line 167:
```suggestion
	For example, the HTML spec defines an <{html-global/anchor}> attribute
	which allows an element to declare what element it is anchored to. 
	This makes the declared element the [=implicit anchor element=]
	for the element with the attribute.
```

tabatkins:
When I initially wrote it the assumption was that the anchor attribute was *part of* the Popover API. If it's been separated, it makes sense to rewrite to refer to this properly, so yeah, merged. (Is there no way for the JS API to trigger an implicit anchor anymore?)

And I fixed the link to just be what will *eventually* link to HTML after the PR lands; it's okay for it to fail for a little bit rather than try to remember to change the link later.

xiaochengh:
> Is there no way for the JS API to trigger an implicit anchor anymore?

There is:

```js
element.anchorElement = anotherElement;
element.anchorElement; // returns anotherElement
```

tabatkins:
I knew about that one, I just thought there was a way to trigger it from the opening call.

Issue #72, opened 2015-10-29
Replace "Animatable" with "Animation type"


birtles:
As discussed: https://lists.w3.org/Archives/Public/www-style/2015Oct/0226.html (item 8)

This will probably involve editing all the CSS specs that use this line and updating terms in CSS transitions too.

birtles:
Need to be careful to mark the 'display' attribute as having 'Animation type: none' when doing this. See [Mozilla bug 1264396](https://bugzilla.mozilla.org/show_bug.cgi?id=1264396) for some background.

xfq:
Related: #3198

birtles:
As pointed out by @xfq this has been done in #3198 and others.

Issue #2751, opened 2018-06-09
[css-logical] Logical properties should be animatable like the physical ones


Loirooriol:
I don't see why this shouldn't be possible:

```css
#element {
  animation: size 1s infinite alternate;
}
@keyframes size {
  from { inline-size: 100px }
  to { inline-size: 200px }
}
```

I expect it to animate just like if I used `width`. But the [spec](https://drafts.csswg.org/css-logical/#dimension-properties) says

> Animation type: discrete

On Chrome it's effectively discrete, on Firefox it doesn't animate at all. https://jsfiddle.net/45e6okrm/

birtles:
From memory, we began implementing this in Firefox but never finished it. It's complicated by the fact that you can also animate the `writing-mode` etc. so that part-way through the animation the mapping for the properties changes, e.g.

```css
#element {
  margin: 50px 100px;
  animation: yer 5s;
}
@keyframes yer {
  to { margin-inline-start: 200px; writing-mode: vertical-lr; }
}
```

As a result you can't just do the mapping to physical properties when you set up the animation. Instead you need to re-evaluate the mapping whenever the _animated_ value of the `writing-mode`, `direction` etc. change.

upsuper:
> It's complicated by the fact that you can also animate the writing-mode etc. 

If that's where the difficulty is from, maybe we should put `writing-mode` and `direction` into the category of proeprties that is not animatable. I don't see how animating them can be useful, and I guess it makes sense to get rid of the animatability of them if that interferes with other stuff.

birtles:
(I should clarify that responding to changes in the _animated_ value is more difficult than responding to other changes in the computed value.

In order to avoid triggering transitions due to animations, UAs need to distinguish between non-animated style updates and animated style updates. In Gecko we do this using a separate style resolution step for animations and I suspect in order to do transitions properly with regards to inheritance, other UAs will need to do something similar, if they don't already. As a result, while non-animation related changes will get picked up naturally during the animation styling, changes due to animation itself need special handling.

We already have a similar situation with `font-size` and custom properties but those only affect the property values where as this affects the property name mappings.)

fantasai:
@upsuper Currently `writing-mode`, `text-orientation`, and `direction` are all listed as not animatable. (By contrast, `unicode-bidi` is listed as discretely animatable.)

fantasai:
[...]

PR #9019, opened 2023-06-30
Convert remaining Animatable lines to Animation type lines.

dbaron:
[css-backgrounds-4][css-box-3][css-color-hdr][css-module][css-writing-modes-3][css-writing-modes-4] Convert remaining "Animatable" lines to "Animation type" lines.

See #72.

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -266,7 +266,7 @@ The 'border-top-radius', 'border-right-radius',
 		Inherited: no
 		Percentages: Refer to corresponding dimension of the <a>border box</a>.
 		Computed value: see individual properties
-		Animatable: see individual properties
+		Animation type: see individual properties
 	</pre>
 
 	<p>The 'border-*-radius' shorthands set the two 'border-*-*-radius'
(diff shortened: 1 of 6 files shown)


svgeesus (commented):
This all looks good, but what does **Animatable: n/a** mean? Is it the same as **not animatable**?

yarusome:
This [resolution](https://github.com/w3c/csswg-drafts/issues/2751#issuecomment-402604609) _implies_ that `writing-mode` should be "not animatable", but the final resolution says differently from the proposed one.

dbaron:
> This all looks good, but what does **Animatable: n/a** mean? Is it the same as **not animatable**?

The "n/a" values are for a propdef table that's for an alias rather than a real property.  (Perhaps aliases should be more like shorthands and say "see aliased property" or something like that.)

> This [resolution](https://github.com/w3c/csswg-drafts/issues/2751#issuecomment-402604609) _implies_ that `writing-mode` should be "not animatable", but the final resolution says differently from the proposed one.

Based on that resolution I agree that like writing-mode, direction, and unicode-bidi should all be not animatable rather than discrete.  But I don't see anything in this PR that's changing those properties.

svgeesus (approved):
LGTM

Issue #8684, opened 2023-04-04
[css-view-transitions-2] MPA: What happens if there's a cross-origin redirect?


jakearchibald:
Previous discussion: https://github.com/WICG/view-transitions/issues/200

If you navigate from origin A to origin B, which redirects back to A, it seems bad to expose that there was no in-between page by allowing a transition to run or not.

So, if there's a cross-origin redirect, it feels like we should disallow transitions. We can lift the behaviour in future if we feel it's safe.

tabatkins:
Agreed.

noamr:
I don't think there is any disagreement here. Can we close this issue? It's in the explainer.

jakearchibald:
As long as folks remember to put it in a spec

noamr:
Speaking with @annevk about this, one valid reason to disallow transitions on cross-origin redirects though is so that the user doesn't feel that things are "smoothed out", they should probably be aware somehow that there was a redirect, even if it's minor like a flash-of-white etc.

If this is the reason to disallow VT on XO-redirect, it's pretty clear to spec, and wouldn't apply to history traversals as the redirect doesn't happen then and it's ok to keep them smooth.

noamr:
Closing, as this is in the draft spec. https://drafts.csswg.org/css-view-transitions-2/#document-setup-outbound-cross-document-view-transition

PR #9022, opened 2023-07-02
[css-view-transitions-1][css-view-transitions-2] Initial draft for cross-document view transitions

noamr:
This lays the foundations for cross-document view transitions.
Specifically:
- The opt-in
- The reveal event

See https://github.com/WICG/view-transitions/blob/main/cross-doc-explainer.md

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1048,6 +1048,12 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: This is used to detect changes in the [=snapshot containing block size=],
 			which causes the transition to [=skip the view transition|skip=].
 			[Discussion of this behavior](https://github.com/w3c/csswg-drafts/issues/8045).
+
+		: <dfn>old state captured steps</dfn>
+		:: An algorithm accepting nothing, or null.
+			Initially null.
+
+			Note: this is used for cross-document view transitions.
 	</dl>
 
 	The {{ViewTransition/finished}} [=getter steps=] are to return [=this's=] [=ViewTransition/finished promise=].
@@ -1204,7 +1210,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Capture the old state=] for |transition|.
 
-			If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
+		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
+
+		1.If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
 			and return.
 
 		1. Set |document|'s [=document/rendering suppression for view transitions=] to true.
@@ -1217,7 +1225,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 					although the render steps in the HTML spec act as if it's synchronous.
 
 			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then abort these steps.
-
+0
 				Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
 					The [=skip the view transition=] steps [=call the update callback=],
 					ensuring the |transition|'s [=ViewTransition/update callback=] is always called.
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1572,6 +1572,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Assert=]: |transition|'s [=ViewTransition/phase=] is not "`done`".
 
+		1. If |transition|'s [=ViewTransition/phase=] is "`pending-capture`" and |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
+
 		1. If |transition|'s [=ViewTransition/phase=] is before "`update-callback-called`", then [=queue a global task=] on the [=DOM manipulation task source=],
 			given |transition|'s [=relevant global object=], to [=call the update callback=] of |transition|.
 
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1230,6 +1230,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 					although the render steps in the HTML spec act as if it's synchronous.
 
 			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then abort these steps.
+
 				Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
 					The [=skip the view transition=] steps [=call the update callback=],
 					ensuring the |transition|'s [=ViewTransition/update callback=] is always called.
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1225,7 +1225,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 					although the render steps in the HTML spec act as if it's synchronous.
 
 			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then abort these steps.
-0
 				Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
 					The [=skip the view transition=] steps [=call the update callback=],
 					ensuring the |transition|'s [=ViewTransition/update callback=] is always called.
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1056,6 +1056,11 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: this is used for cross-document view transitions.
 	</dl>
 
+	A {{ViewTransition}} must never have both an [=ViewTransition/update callback=] and a [=ViewTransition/old state captured steps=].
+
+	Note: [=ViewTransition/update callback=] is optionally set for same-document view transitions,
+		and [=ViewTransition/old state captured steps=] is set for cross-document view transitions.
+
 	The {{ViewTransition/finished}} [=getter steps=] are to return [=this's=] [=ViewTransition/finished promise=].
 
 	The {{ViewTransition/ready}} [=getter steps=] are to return [=this's=] [=ViewTransition/ready promise=].
@@ -1208,7 +1213,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |document| be |transition|'s [=relevant global object's=] [=associated document=].
 
-		1. [=Capture the old state=] for |transition|.
+		1. [=Capture the old state=] for |transition|. If that fails, set |transition|'s [=ViewTransition/phase=] to "`done`".
 
 		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
 
@@ -1572,8 +1577,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Assert=]: |transition|'s [=ViewTransition/phase=] is not "`done`".
 
-		1. If |transition|'s [=ViewTransition/phase=] is "`pending-capture`" and |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
-
 		1. If |transition|'s [=ViewTransition/phase=] is before "`update-callback-called`", then [=queue a global task=] on the [=DOM manipulation task source=],
 			given |transition|'s [=relevant global object=], to [=call the update callback=] of |transition|.
 
@@ -1596,6 +1599,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: Since the rejection of |transition|'s [=ViewTransition/update callback done promise=] isn't explicitly handled here,
 			if |transition|'s [=ViewTransition/update callback done promise=] rejects,
 			then |transition|'s [=ViewTransition/finished promise=] will reject with the same reason.
+
+		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call |transition|'s [=ViewTransition/process old state captured=].
 	</div>
 
 ## [=Capture the image=] ## {#capture-the-image-algorithm}
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1213,12 +1213,12 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |document| be |transition|'s [=relevant global object's=] [=associated document=].
 
-		1. [=Capture the old state=] for |transition|. If that fails, set |transition|'s [=ViewTransition/phase=] to "`done`".
+		1. [=Capture the old state=] for |transition|.
+
+			If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
+			and return.
 
 		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
-
-		1. If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
-			and return.
 
 		1. Set |document|'s [=document/rendering suppression for view transitions=] to true.
 
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1212,7 +1212,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
 
-		1.If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
+		1. If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
 			and return.
 
 		1. Set |document|'s [=document/rendering suppression for view transitions=] to true.
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -106,7 +106,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	A successful cross-document view transition goes through the following phases:
 
 	1. The user navigates, by clicking a link, submitting a form, traversing history using the
-		browser API, etc.
+		browser UI, etc.
 
 	1. Once it's time to [=unload=] the old document, if the navigation is [=same origin=]
 		and the old {{Document}} has opted in to cross-document view-transitions, the old state is captured.
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -11,7 +11,7 @@ TR: https://www.w3.org/TR/css-view-transitions-2/
 Work Status: exploring
 Editor: Noam Rosenthal, Google, w3cid 121539
 Editor: Khushal Sagar, Google, w3cid 122787
-Editor: Vladimir Levin, Google
+Editor: Vladimir Levin, Google, w3cid 75295
 Editor: Tab Atkins-Bittner, Google, http://xanthir.com/contact/, w3cid 42199
 Abstract: This module defines how the View Transition API works with cross-document navigations.
 Markup Shorthands: css yes, markdown yes
@@ -20,15 +20,16 @@ Markup Shorthands: css yes, markdown yes
 <pre class=link-defaults>
 spec:css-view-transitions-1;
 	text: active view transition; type: dfn;
+	text: activate view transition; type: dfn;
 	text: skip the view transition; type: dfn;
 	text: ViewTransition; type: interface;
 	text: named elements; for: ViewTransition; type: dfn;
 	text: update callback done promise; for: ViewTransition; type: dfn;
 	text: initial snapshot containing block size; for: ViewTransition; type: dfn;
-	text: capture the old state; type: dfn;
 	text: activate view transition; type: dfn;
 	text: captured elements; type: dfn;
 	text: updateCallbackDone; type: property; for: ViewTransition;
+	text: phase; type: property; for: ViewTransition;
 	text: call the update callback; type: dfn;
 spec:dom; type:dfn; text:document
 spec:css22; type:dfn; text:element
@@ -118,7 +119,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	1. Right before the new {{Document}} has the first [=rendering opportunity=], its state is captured as
 		the "new" state.
 
-	1. An event named [[#reveal-event|`reveal`]] is fired on the new {{Document}}, with a `viewTransition` property,
+	1. An event named {{PageRevealEvent|reveal}} is fired on the new {{Document}}, with a `viewTransition` property,
 		which is a {{ViewTransition}} object. This {{ViewTransition}}'s <code>{{ViewTransition/updateCallbackDone}}</code> is already resolved,
 		and its [=captured elements=] are populated from the old {{Document}}.
 
@@ -150,7 +151,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		- Opt-in to auto-view-transitions in both pages.
 		- Pass the click location to the new document, e.g. via {{WindowSessionStorage/sessionStorage}}.
-		- Intercept the {{ViewTransition}} object in the new document, using the [[#reveal-event|`reveal` event]].
+		- Intercept the {{ViewTransition}} object in the new document, using the {{PageRevealEvent|reveal event}}.
 
 		In both pages:
 		```css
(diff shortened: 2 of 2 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+			is being [=Document/reactivated=], in which case we don't need to check for cross-origin redirects.
+
+		1. If |oldDocument| does not [=opt in to cross-document view transitions=], then return.
+
+			Note: we don't know yet if |newDocument| has opted in, as it might not be parsed yet.
+			We check the opt-in for |newDocument| when it is [=reveal document|revealed=].
+
+		1. Let |transition| be a new {{ViewTransition}} object in |oldDocument|'s [=relevant Realm=].
+
+		1. [=Capture the old state=] for |transition|. If this fails, return.
```
khushalsagar, line 329: A few notes that need some more thinking:

- As above, this needs to be called from within a rendering opportunity. Might just set it as the "active view transition" and then let the "perform pending transition operations" take care of doing the capture. It calls "setup view transition" for the "active view transition" pending-capture. The algorithm can split in "setup view transition" where we queue a task to run the update callback today. It can check if this is a cross-doc VT and call some other algorithm in L2?
- Since the capture is async, technically script can call startViewTransition while it is running. Currently new transition always cancels existing transition but in this case it makes sense for the cross-doc one to win. So this [algorithm](https://www.w3.org/TR/css-view-transitions-1/#ViewTransition-prepare) will need to change.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+
+			Note: implementations might need to [=queue a task=] in order to [=capture the old state=].
+			However, this should not have web-observable side-effects.
+
+		1. [=Queue a global task=] on the [=DOM manipulation task source=],
```
khushalsagar, line 334: Not sure about a task here referencing the new Document here? It feels like we should stash the old doc's transition object somewhere and then as a part of creating the new Document create the VT object which copies over state from it (as in the step below).

We basically want an async phase between deciding that the navigation will be committed but before the unload steps are run. This global task should unblock running the unload steps.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+			given |newDocument|'s [=relevant global object=], to perform the following step:
+
+			1. Set |newDocument|'s [=active view transition=] to a new {{ViewTransition}} in |newDocument|'s [=relevant Realm=],
+				whose [=ViewTransition/named elements=] is |transition|'s [=ViewTransition/named elements=],
+				[=ViewTransition/update callback done promise=] is [=a promise resolved with=] undefined,
+				[=ViewTransition/initial snapshot containing block size=] is |transition|'s [=ViewTransition/initial snapshot containing block size=],
+				and whose [=ViewTransition/is cross-document=] is true.
+
+			Note: this task would run before any script on the new {{Document}}.
```
khushalsagar, line 343: This wouldn't be true for pre-render/BFCache. So remove it?

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+
+		1. [=Capture the old state=] for |transition|. If this fails, return.
+
+			Note: implementations might need to [=queue a task=] in order to [=capture the old state=].
+			However, this should not have web-observable side-effects.
+
+		1. [=Queue a global task=] on the [=DOM manipulation task source=],
+			given |newDocument|'s [=relevant global object=], to perform the following step:
+
+			1. Set |newDocument|'s [=active view transition=] to a new {{ViewTransition}} in |newDocument|'s [=relevant Realm=],
```
khushalsagar, line 337: We need more precise timing for when this is done on the new Document, both for a new nav vs pre-render/BFCache restore.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+				whose [=ViewTransition/named elements=] is |transition|'s [=ViewTransition/named elements=],
+				[=ViewTransition/update callback done promise=] is [=a promise resolved with=] undefined,
+				[=ViewTransition/initial snapshot containing block size=] is |transition|'s [=ViewTransition/initial snapshot containing block size=],
```
khushalsagar, line 340: The phase also needs to start with "update-callback-called" on the new Document?

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+## Monkey patches to HTML ## {#monkey-patch-to-html}
+
+	<div algorithm="monkey patch to unload">
+		Run the following step before <a href="https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-documents:update-the-visibility-state">updating the visibility state</a> in the [=unload=] steps:
+
+		1. If |newDocument| is given, then [=setup outbound cross-document view transition=] given |oldDocument| and |newDocument|.
+	</div>
+
+	<div algorithm="monkey patch to render blocking">
+		Run the following step at the end of [=unblock rendering=]:
```
khushalsagar, line 274: "unblock rendering" is per element. We need a different link here. There is also subtlety about whether this event fires after the "load" events on all resources which were render blocking.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+		Run the following step in [=Document/reactivate=], before step 4 (querying for [=page showing=]):
+
+		1. [=reveal document|reveal=] |document|.
+	</div>
+
+	<div algorithm="page reveal">
+	To <dfn>reveal {{Document}}</dfn> |document|:
+		1. Assert: |document|'s [=page showing=] is false.
+
+		1. If |document| is [=render-blocked=] then return.
```
khushalsagar, line 291: Why this check? We call this function when render blocking is finished (or timed out). So assert that we're not render-blocked?

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+
+		1. Fire a new event named <code>reveal</code> on |document|'s [=relevant global object=],
```
khushalsagar, line 293: Should we check the new Document opt-in before dispatching the reveal event?

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+				whose [=ViewTransition/named elements=] is |transition|'s [=ViewTransition/named elements=],
+				[=ViewTransition/update callback done promise=] is [=a promise resolved with=] undefined,
+				[=ViewTransition/initial snapshot containing block size=] is |transition|'s [=ViewTransition/initial snapshot containing block size=],
+				and whose [=ViewTransition/is cross-document=] is true.
+
+			Note: this task would run before any script on the new {{Document}}.
+	</div>
+
+	<div algorithm>
+		To get the <dfn>inbound cross-document view-transition</dfn> for a {{Document}} |document|:
```
khushalsagar, line 347: Why do we need a new variable? Shouldn't this just use "active view transition". This raises the question of what happens if author calls startVT before the cross-doc transition is done. I feel like we should give the same-doc VT priority in the new Doc but this deserves an issue.

Overall this means we can't support concurrent transitions which are scoped to the same element. @vmpstr and I discussed this and align on that restriction. MPA and SPA transitions are scoped to the root element today. This restriction will get relaxed with scoped transitions, where an author could do a same-doc transition scoped to body for example.

css-view-transitions-2/Overview.bs
noamr: > A few notes that need some more thinking:
> 
> * As above, this needs to be called from within a rendering opportunity. Might just set it as the "active view transition" and then let the "perform pending transition operations" take care of doing the capture. It calls "setup view transition" for the "active view transition" pending-capture. The algorithm can split in "setup view transition" where we queue a task to run the update callback today. It can check if this is a cross-doc VT and call some other algorithm in L2?

sure.

> * Since the capture is async, technically script can call startViewTransition while it is running. Currently new transition always cancels existing transition but in this case it makes sense for the cross-doc one to win. So this [algorithm](https://www.w3.org/TR/css-view-transitions-1/#ViewTransition-prepare) will need to change.

Will do.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+		1. Assert: this is called when |oldDocument| is [=unloading=].
+
+		1. If |oldDocument|'s [=active view transition=] is not null,
+			then [=skip the view transition|skip=] |oldDocument|'s [=active view transition=]
+			with an "{{AbortError}}" {{DOMException}} in |oldDocument|'s [=relevant Realm=].
+
+		1. If |oldDocument|'s [=environment settings object/origin=] is not [=same origin=] as
+			|newDocument|'s [=environment settings object/origin=] then return.
+
+		1. If |newDocument| [=was created via cross-origin redirects=] is true
```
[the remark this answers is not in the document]
noamr: OK this is important and I gave this a lot of thought.

For an A->B->C case, the previous condition would be true anyway, so we would return early. What matters here is an A->B->A redirect ("confused agent"). For that, see the next line and note. In a BFCache restore/prerender reactivation, `|newDocument|'s [=latest entry=]` is not null so we would ignore the cross-origin redirect. 

The restriction on cross-origin redirects only applies *when the redirect is happening*, meaning in the initial navigation. This is to prevent a situation where the view transition obscures the fact that there was a redirect.

From that point on, in BFCache restores and in prerender activations, the cross-origin redirects are no longer a factor. This is simply a navigation between same-origin documents.

css-view-transitions-2/Overview.bs
noamr: We also call it on reactivation, which could happen when the document is still render blocked (e.g. you activate a prerendered page really quickly before it's unblocked). To clarify, I moved this line into the reactivation bit.

css-view-transitions-2/Overview.bs
noamr: It's OK that it's called per element. Why do we need a different link?
Regarding the subtlety: right, it should be unblocked after the load event. This is already the case for styles, we should change the order for scripts so that this is called after the load/error events.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+	A {{ViewTransition}} additionally has:
+	<dl dfn-for=ViewTransition>
+		: <dfn>is cross-document</dfn>
+		:: a boolean, initially false.
+	</dl>
+
+## Monkey patches to HTML ## {#monkey-patch-to-html}
+
+	<div algorithm="monkey patch to unload">
+		Run the following step before <a href="https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-documents:update-the-visibility-state">updating the visibility state</a> in the [=unload=] steps:
```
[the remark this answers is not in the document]
noamr: Yes. It requires some work, on it.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+
+	<pre class='descdef'>
+	Name: same-origin
+	For: @auto-view-transition
+	Value: enabled | disabled
+	Initial: disabled
+	</pre>
+
+	The '<dfn for="@auto-view-transition">same-origin</dfn>' property opts in to automatically performing a view transition when performing a [=same origin=] navigation.
+	It needs to be enabled both in the old document (when unloading) and in the new document (when ready to render).
```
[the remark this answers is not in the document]
noamr: > We need to be precise about the timing of when the property is read and applied by the browser.
> 
> * On the old Document, the state is read after the `navigate` event is fired. I'm not sure what we should do if the author toggles it to disabled between this event and unloading. Let's sort out that detail on an issue, seems like an edge case we can leave undefined.

I think it's important actually. Let me try to spec it.

> * On the new Document, the property is read when the Document is not render-blocked which should be immediately before the `reveal` event. The ViewTransition object on that event would be null if the new Document doesn't opt-in at this point. 

> Again, not sure what we should do if the author toggles it in the `reveal` event or anytime after. The easy thing seems to be to not read the value after, author can use VT object to skip in the event if they need to.

Yes. Setting the opt-in in/after reveal would opt-in for outbound transitions but would be too late for the inbound one. I think that's ok.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+Status: WD
+Group: csswg
+Date: 2023-05-30
+Prepare for TR: yes
+ED: https://drafts.csswg.org/css-view-transitions-2/
+TR: https://www.w3.org/TR/css-view-transitions-2/
+Work Status: exploring
+Editor: Noam Rosenthal, Google, w3cid 121539
+Editor: Khushal Sagar, Google, w3cid 122787
+Editor: Tab Atkins-Bittner, Google, http://xanthir.com/contact/, w3cid 42199
```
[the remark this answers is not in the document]
noamr: Sure! @vmpstr what's your w3cid?

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+
+	View Transitions, as specified in [css-view-transitions-1], is a feature that allows developers
+	to create animated transitions between visual states of the [=/document=].
+
+	Level 2 extends that specification, by adding the necessary API and lifecycle to enable
+	transitions across a same-origin cross-document navigation.
+
+
+## Lifecycle ## {#lifecycle}
+
```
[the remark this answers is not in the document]
noamr: Done

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+Note: this should go in the HTML spec. See [Issue 9315](https://github.com/whatwg/html/issues/9315).
+
+<xmp class=idl>
+		[Exposed=Window]
+		interface PageRevealEvent : Event {
+			readonly attribute ViewTransition? viewTransition;
+		};
+</xmp>
+
```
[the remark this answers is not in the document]
noamr: Done

css-view-transitions-1/Overview.bs
```diff
@@ -122,63 +118,51 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 	*This section is non-normative.*
 
-	This specification introduces a DOM API and associated CSS features
-	that allow developers to create animated visual transitions,
-	called <dfn export>view transitions</dfn>
-	between different states of a [=/document=].
+	View Transitions is a feature that allows developers to create animated transitions between visual states of the [=/document=].
```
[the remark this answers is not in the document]
noamr: Submitted by default

css-view-transitions-1/Overview.bs
```diff
@@ -12,8 +12,7 @@ Work Status: exploring
 Editor: Tab Atkins-Bittner, Google, http://xanthir.com/contact/, w3cid 42199
 Editor: Jake Archibald, Google, w3cid 76394
 Editor: Khushal Sagar, Google, w3cid 122787
-Abstract: This module defines the View Transition API, along with associated properties and pseudo-elements,
-	which allows developers to create animated visual transitions representing changes in the document state.
+Abstract: This module defines the View Transition API, along with associated properties and pseudo-elements.
```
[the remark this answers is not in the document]
noamr: Oops.

css-view-transitions-1/Overview.bs
```diff
@@ -122,63 +118,51 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+## Separating transitions from DOM updates ## {#separating-transitions}
 
-	Traditionally, creating a visual transition between two document states
-	required a period where both states were present in the DOM at the same time.
-	In fact, it usually involved creating a specific DOM structure
-	that could represent both states.
-	For example, if one element was “moving” between containers,
-	that element often needed to exist outside of either container for the period of the transition,
-	to avoid clipping from either container or their ancestor elements.
+	Traditionally, creating a transition between two states involves a period where both states are present in the DOM at the same time.
```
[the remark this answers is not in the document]
noamr: Oopd

css-view-transitions-2/Overview.bs
noamr: Added a monkey patch to script processing. The reveal event will also be called after the load/error events (this is already true for styles)

css-view-transitions-2/Overview.bs
noamr: Revised this.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,373 @@
...
+		tab-size: 2;
+	}
+</style>
+<script async type="module" src="diagrams/resources/scaler.js"></script>
+
+# Introduction # {#intro}
+
+	*This section is non-normative.*
+
+	View Transitions, as specified in [css-view-transitions-1], is a feature that allows developers
```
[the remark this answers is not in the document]
noamr: Fixed

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,400 @@
...
+			then [=skip the view transition|skip=] |oldDocument|'s [=active view transition=]
+			with an "{{AbortError}}" {{DOMException}} in |oldDocument|'s [=relevant Realm=].
+
+			Note: this means that any running transition would be skipped when the document is ready
+			to unload.
+
+		1. Let |outboundTransition| be a new {{ViewTransition}} object in |oldDocument|'s [=relevant Realm=],
+			whose [=ViewTransition/process old state captured=] is set to the following steps:
+
+			1. |outboundTransition|'s [=ViewTransition/phase=] is not "`pending-capture`", then call |onReady| and return.
```
khushalsagar, line 352:
```suggestion
			1. If |outboundTransition|'s [=ViewTransition/phase=] is not "`pending-capture`", then call |onReady| and return.
```

When would this happen? It seems like the state here must either be "pending-capture", if the state was captured as expected.

It could be "done" if something caused us to skip the transition in the meanwhile. Is that the case you're handling? If so, mind framing it as:

1. If outboundTransition's phase is done, then call onready and return.

1. Assert transition's phase is pending-capture.

This is also worth leaving an issue about. I think the only situation where this can happen if a call to startViewTransition is made before the unload steps resume. And I'm not sure if we should be skipping the cross-doc transition capture in favour of a same-doc transition.

css-view-transitions-1/Overview.bs
```diff
@@ -1204,7 +1210,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Capture the old state=] for |transition|.
 
-			If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
+		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
```
khushalsagar, line 1213: Let's also call this from [skip](https://drafts.csswg.org/css-view-transitions-1/#skip-the-view-transition-algorithm). Or assert that this is not set in that algorithm. It'll depend on if we find a case where the cross-doc transition can be skipped on the old Document. Worth leaving an issue about.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,400 @@
...
+
+			1. If |oldDocument| does not [=opt in to cross-document view transitions|opts in to cross-document view transitions=],
```
khushalsagar, line 354: Why do we need a check here? Didn't we already do it before initiating the capture?

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,400 @@
...
+					and whose [=ViewTransition/is cross-document=] is true.
+
+				1. [=Call the update callback=] for |newDocument|'s [=active view transition=].
+
+				1. Call |onReady|.
+
+			Note: |outboundTransition| is not exposed to JavaScript, it is used only for capturing
+			the state of the old document.
+
+		1. Set |oldDocument|'s [=active view transition=] to |outboundTransition|.
```
khushalsagar, line 370: Can you leave a note here: "The [view transition](https://drafts.csswg.org/css-view-transitions-1/#view-transitions) process continues in [setup view transition](https://drafts.csswg.org/css-view-transitions-1/#setup-view-transition), via [perform pending transition operations](https://drafts.csswg.org/css-view-transitions-1/#perform-pending-transition-operations), which is called in [§ 7.2 Monkey patches to rendering](https://drafts.csswg.org/css-view-transitions-1/#monkey-patch-to-rendering-algorithm)."

Same as what's in startViewTransition() algorithm.

css-view-transitions-2/Overview.bs
khushalsagar: I mistook "inbound cross-document view-transition" to be a variable instead of algorithm. My bad, this LGTM.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,400 @@
...
+	</div>
+
+	<div algorithm>
+		To get the <dfn>inbound cross-document view-transition</dfn> for a {{Document}} |document|:
+
+		1. If |document| does not [=opt in to cross-document view transitions=], then return null.
```
khushalsagar, line 376: I don't think we should have a check for this here. If there is an "active view transition" whose "is cross document" is true then that's always returned. The places where the opt in is checked should be explicit and we should clear the "active view transition" at those places if needed.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,400 @@
...
+
+		1. Let |transition| be |document|'s [=active view transition=].
+
+		1. If |transition| is null or |transition|'s [=ViewTransition/is cross-document=] is false,
```
khushalsagar, line 380: nit: Do you mind using "is inbound cross-document"? Just to make it explicit that the bit is only set for the VT object on the new Document.

css-view-transitions-2/Overview.bs
noamr: Oh yes, there was an early return in `reveal a document` that checked this and I removed it by mistake.

css-view-transitions-2/Overview.bs
noamr: Fixed the wording.
Regarding `done`, even if a same-doc transition will not be automatically skipped, someone could run:
`startViewTransition().skipTransition()`, no?

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,400 @@
...
+
+	The '<dfn for="@auto-view-transition">same-origin</dfn>' property opts in to automatically performing a view transition when performing a [=same origin=] navigation.
+	It needs to be enabled both in the old document (when unloading) and in the new document (when ready to render).
+
+	<dl dfn-type=value dfn-for="same-origin">
+		: <dfn>disabled</dfn>
+		:: There will be no transition.
+
+		: <dfn>enabled</dfn>
+		:: The transition will be enabled if the navigation is same-origin, without cross-origin
```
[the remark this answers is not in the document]
noamr: Linking https://github.com/w3c/csswg-drafts/issues/8684 instead, to continue the discussion here.

css-view-transitions-1/Overview.bs
noamr: Done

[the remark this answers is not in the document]
khushalsagar: Hmmm, I copied the text over to the bikeshed tool to confirm. Figured that will show me what the generated spec will look like when this lands. But we can fix this in a follow up if needed.

css-view-transitions-1/Overview.bs
```diff
@@ -1565,6 +1572,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Assert=]: |transition|'s [=ViewTransition/phase=] is not "`done`".
 
+		1. If |transition|'s [=ViewTransition/phase=] is "`pending-capture`" and |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
```
khushalsagar, line 1575: Shouldn't the phase value be updated before calling "process old state captured"? Otherwise that algorithm will incorrectly assume that the capture was successful.

Doing this as the last step is a good idea.

css-view-transitions-1/Overview.bs
```diff
@@ -1048,6 +1048,12 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: This is used to detect changes in the [=snapshot containing block size=],
 			which causes the transition to [=skip the view transition|skip=].
 			[Discussion of this behavior](https://github.com/w3c/csswg-drafts/issues/8045).
+
+		: <dfn>old state captured steps</dfn>
+		:: An algorithm accepting nothing, or null.
+			Initially null.
```
khushalsagar, line 1054: Can you add normative text saying only one of "update callback" or "old state captured steps" should be set?

And a note to clarify that the callback is optionally set for same-doc transitions and "old state captured steps" is always set if this is on the old doc for a cross-document transition.

css-view-transitions-1/Overview.bs
khushalsagar, line 1213: We need to inform the "process old state captured algorithm" if there is a failure in "capture the old state".  You could set the phase to done for it.

css-view-transitions-2/Overview.bs
```diff
@@ -304,22 +308,26 @@ The <dfn attribute for=PageRevealEvent>viewTransition</dfn> [=getter steps=] are
...
 	To <dfn>reveal {{Document}}</dfn> |document|:
 		1. Assert: |document|'s [=page showing=] is false.
 
-		1. Fire a new event named <code>reveal</code> on |document|'s [=relevant global object=],
-			using {{PageRevealEvent}}.
+		1. If |document| is [=render-blocked=], then return.
 
 		1. Let |transition| be the result of getting the [=inbound cross-document view-transition=] for |document|.
 
+		1. If |document| does not [=opt in to cross-document view transitions=], then [=skip the view transition|skip=] |transition| and set |transition| to null.
```
khushalsagar, line 322: nit: Do we need a null check for `transition` here? Also, I think you want say set document's active view transition to null. Mapping this to code, `transition` is now a local variable in this algorithm so resetting it won't reset the member variable on Document.

css-view-transitions-2/Overview.bs
```diff
@@ -0,0 +1,400 @@
...
+	1. The user navigates, by clicking a link, submitting a form, traversing history using the
+		browser UI, etc.
+
+	1. Once it's time to [=unload=] the old document, if the navigation is [=same origin=]
+		and the old {{Document}} has opted in to cross-document view-transitions, the old state is captured.
+
+	1. Right before the new {{Document}} has the first [=rendering opportunity=], its state is captured as
+		the "new" state.
+
+	1. An event named {{PageRevealEvent|reveal}} is fired on the new {{Document}}, with a `viewTransition` property,
```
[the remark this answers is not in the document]
khushalsagar: Eh, not seeing the ordering change? Just to clarify I mean, "Right before the new {{Document}} has the first [=rendering opportunity=], its state is captured as
		the "new" state." needs to be after firing the reveal event.

khushalsagar (approved):

css-view-transitions-2/Overview.bs
noamr: skip transition already sets the active view transition to null.
Fixed the nit.

css-view-transitions-2/Overview.bs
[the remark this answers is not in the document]
noamr: Oops, fixed

css-view-transitions-1/Overview.bs
noamr: Fixed

css-view-transitions-1/Overview.bs
noamr: Done

css-view-transitions-1/Overview.bs
noamr: Done

PR #9024, opened 2023-07-02
[css-pseudo-4][editorial] Fixed typo in example

### LINKED ISSUE #9023: [css-pseudo] [css-pseudo-4] Draft text: First-Line Text example missing 'styled'
In the 20221230 Working Draft https://www.w3.org/TR/css-pseudo-4/ (minor typo)
In the example for First-Line Text,
the base HTML text is:
  ... first line will be styled by the ‘::first-line’ pseudo-element.
In the 2 examples for page width impact,
the word 'styled' is omitted from the example:
  ... first line will be by the ‘::first-line’ pseudo-element.

SebastianZ:
This adds the missin word "styled" to the examples.

Fixes #9023

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -94,7 +94,7 @@ First-Line Text: the ''::first-line'' pseudo-element</h3>
     <pre class="figure">
       THIS IS A SOMEWHAT LONG HTML PARAGRAPH THAT
       will be broken into several lines. The
-      first line will be by the ‘::first-line’
+      first line will be styled by the ‘::first-line’
       pseudo-element. The other lines will be
       treated as ordinary lines in the paragraph.
     </pre>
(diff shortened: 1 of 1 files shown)


LarryLACa:
Flowing the inserted text at the right margin was tricky.
because of the long hyphenated terms like first-line  and pseudo-element,
which may or may not break and wrap on the right at the '-',
when pushed to the right by the 'styled' insertion flow.

The #9024 fix, inserting 'styled' and extending that line on the right 
without page width wrap, seems to extend the line 
past the right margin implied by the preceding lines.

Can you give me a link to the updateddraft spec file, 
so we can see the end result visually?

LarryLACa:
I found the Editors Draft page:
  https://drafts.csswg.org/css-pseudo-4/
with the displayed visuals of Example 2.
The repaired 'be styled by' lines are too long.
Their right margin wrap is inconsistent with the preceding lines.

See my rework here:
  https://github.com/w3c/csswg-drafts/compare/main...LarryLACa:csswg-drafts:patch-1

SebastianZ:
You're right. I didn't check the line lengths of the examples.

I've merged your patch in #9031. Thanks a lot, @LarryLACa!

Sebastian

PR #9036, opened 2023-07-05
[mediaqueries-5] Add `prefers-shapes` media query

### LINKED ISSUE #7406: Media query to prefer shapes over colours
Thank you for the recent additions of

- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

However, as a colourblind person, I wish websites would use shapes over colours. For example, it is common to use green for success and red for failure. The two most common types of colourblindness are related to green/red confusion. People are directly harmed when only color is used as differentiation.

Ideally colour is used only as a secondary styling concern. However, sometimes using shapes is an afterthought and isn't fit the original design.

Video games often address colourblindness issues by adding the ability to use shapes/patterns as additional differentiation.

Examples of "prefer shapes":
- https://ablegamers.org/unlockingaccessibilitypuzzlegames/
- https://gameaccessibilityguidelines.com/faster-than-light-colourblind-mode/
- https://gameaccessibilityguidelines.com/hue-colorblind-mode/
- https://gameaccessibilityguidelines.com/two-dots-colorblind-mode/
- https://wearecolorblind.com/examples/trello-colorblind-friendly-mode/
- https://www.kodable.com/learn/making-kodable-colorblind-accessible/
- https://wearecolorblind.com/examples/fodmap-app/
- https://wearecolorblind.com/examples/trainyard/

Other examples when the community suggests how to fix colour issues:
- https://www.reddit.com/r/AmongUs/comments/ix21ch/my_idea_for_a_colorblind_mode_the_pattern_is_over/
- https://www.reddit.com/r/ColorBlind/comments/izarr3/among_us_color_blind_concept/

As an aside, speaking from someone who has a rather several case of colourblindness, those "colourblind filters" video games try to make never work for me. I always need shapes. For me personally, the only two colours that are blindingly different is bright/saturated red/blue.

lukewarlow:
Fixes  #7406

This is used to indicate that a user has a preference for conveying information through more than just colors alone.

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -157,8 +157,9 @@ Prefers-* Media Features Security and Privacy</h3>
 	User agents and developers implementing this
 	specification need to be aware of this vector and take it
 	into consideration when deciding whether to use the feature.
-	Specifically `prefers-reduced-motion`, `prefers-color-scheme` and
-	`prefers-reduced-data` are currently of concern for exploitation.
+	Specifically `prefers-reduced-motion`, `prefers-color-scheme`,
+	`prefers-reduced-data`, `prefers-reduced-transparency` and `prefers-shapes` are currently of concern for
+	exploitation.
 	</div>
 
 <!--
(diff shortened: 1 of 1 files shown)


lukewarlow:
Not been approved by the CSSWG yet but it seems like a good candidate for addition so I thought I'd go ahead and make an attempt at writing a spec PR for it. Apologies if this not wanted.

Issue #8913, opened 2023-06-03
[css-anchor-position-1] Reusing the implicit anchor name on the nested elements


kizu:
- Anchor positioning spec: https://drafts.csswg.org/css-anchor-position-1/
- Anchoring in the OpenUI popover explainer: https://open-ui.org/components/popover.research.explainer/#anchoring

We currently can use the [implicit anchor element](https://drafts.csswg.org/css-anchor-position-1/#implicit-anchor-element) provided by the `anchor` attribute associated with some `id`. This would be mainly used for popovers (though, at least the current implementation in Chrome Canary allows this attribute and its implicit anchor on any element).

The issue is: we currently seem to only be able to use this implicit anchor on the element itself, but we don't have access to it in any other way?

What we could want is to somehow access this anchor from elements or pseudo-elements nested inside our popover (or another element with `anchor` on it). This can be useful when we'd want to use this element as the tooltip arrow or another connector that should span from the popover's anchor to the popover itself.

Here is an implementation of how this should work: https://codepen.io/kizu/pen/qBJQpjm?editors=1100 — using both the `anchor` attribute, and the explicitly set `anchor-name`s.

(note that there is currently a bug where we _have_ to use the `anchor` attribute even if we do not use the implicit anchor and override it with the `anchor-default` — https://bugs.chromium.org/p/chromium/issues/detail?id=1451095)

In the example above we'd want to have an element that is nested inside the popover and which spans from it to the button it is anchored with via an `anchor` attribute.

If we'd use an additional element, we could add the same `anchor` attribute, but that would mean duplicating things.

Some ideas of possible solutions:

- When we define the implicit anchor, make it somehow inherit to all the elements inside?
- Introduce a new value for the [anchor-element](https://drafts.csswg.org/css-anchor-position-1/#typedef-anchor-element)?
- Something else?

xiaochengh:
An easy fix, at least for your use case, would be saying a `::before` or `::after` pseudo element (which are the only pseudo elements that can create an OOF box) has the same implicit anchor element as the originating element.

---

Or if we want any child element to be able to use its parent's implicit anchor, then this may be solved by @tabatkins's proposal in #8181

The proposal says whenever we have an anchor element reference (either by implicit anchor, `anchor-default` or an anchor name), its computed value should be the actual anchor element instead of just a name.

Then we can do something like

```
.parent { anchor-default: implicit; } /* This should compute to the implicit anchor element */
[...]

PR #9037, opened 2023-07-05
[css-anchor-position-1] Add implicit anchors to pseudo elements

xiaochengh:
For #8913

Allows `::before`, `::after` and `::backdrop` to use originating element's implicit anchor element

@tabatkins  PTAL


tabatkins:
We *might* want to extend this to all tree-abiding pseudos later, but for now I'm happy with this list.

PR #9040, opened 2023-07-06
[css-anchor-position-1] Hide `anchor-scroll` as an internal detail

### LINKED ISSUE #8675: [css-anchor-1] The `anchor-scroll` property name is confusing/overloaded
"CSS anchor-scroll" and "CSS scroll anchoring" are completely irrelevant things, but the names are almost identical, which may lead to confusions.

More notes/clarifications are needed. Or if possible, rename the property.

xiaochengh:
Fixes #8675

As discussed, there's no need to expose `anchor-scroll` as a separate property and provide opt-out. So we turn it into a internal detail of computing the snapshotted scroll offset.

### DIFF
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -457,7 +457,7 @@ assuming that all [=scroll containers=]
 between the [=target anchor element=]
 and the positioned element's [=containing block=]
 are scrolled to their initial scroll position
-(but see 'anchor-scroll').
+(but see [[#scroll]]).
 
 If the [=target anchor element=] is [=fragmented=],
 the axis-aligned bounding rectangle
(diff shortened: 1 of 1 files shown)


xiaochengh:
@tabatkins ptal

css-anchor-position-1/Overview.bs
```diff
@@ -667,37 +658,16 @@ This means a positioned element
...
-		Selects a [=target anchor element=]
-		the same as ''anchor()'',
-		which will be compensated for in positioning and fallback.
-</dl>
-
-Note: When the element uses 'anchor-default'
-or has an [=implicit anchor element=],
-authors can often avoid explicitly setting an 'anchor-scroll' value
-because the initial value is ''anchor-scroll/default''.
+is only anchoring to elements in the same [=scroll container=],
```
tabatkins, line 664: Hm, this is a new restriction. Do you mean something like "as long as all the [=target anchor elements=] of [=anchor functions=] used on the element have the same nearest ancestor [=scroll container=]"?

xiaochengh: I didn't mean to add a new restriction, but just trying to reword the existing statement and make it more precise: the compensation works well only if all anchors are in the same scroller.

The existing statement says it works well only if there's only one anchor, which isn't entirely correct.

tabatkins: Ah, then it shouldn't be stated here in the list of restrictions, but rather in a Note.

xiaochengh: Actually, I don't see a very straightforward condition for this compensation to work well. Even if there's only one anchor, it can be e.g.

```css
top: calc(anchor(bottom) * 0.9)
```

which will appear detached with the anchor during scrolling.

I'll just remove this entirely.

PR #9073, opened 2023-07-16
[css-view-transitions-1][css-view-transitions-2] Fix some refs and add security section

noamr:


css-view-transitions-2/Overview.bs
```diff
@@ -414,3 +411,25 @@ The <dfn attribute for=PageRevealEvent>viewTransition</dfn> [=getter steps=] are
...
 <h2 id="sec" class="no-num">Security Considerations</h2>
+
+To prevent cross-origin issues, at this point cross-document view transitions can only be enabled for
+same-origin navigations. As discussed in <a href="https://github.com/WICG/view-transitions/issues/200">WICG/view-transitions#200</a>,
+this still presents two potential threats:
+
+1. The <a data-xref-type="http-header">Cross-Origin-Opener-Policy</a> of both documents might be different.
+	This can cause a situation where a {{Document}} that is [=environment settings object/cross-origin isolated capability|cross-origin isolated=]
+	can read image data from a document that is not cross-origin isolated. To mitigate this, implementations should
+	keep the image data in an isolated memory space (e.g. a separate process), not directly accessible to the new {{Document}}.
```
khushalsagar, line 422: This requirement is already there because of cross-origin iframes being a part of captured images: https://drafts.csswg.org/css-view-transitions-1/#sec. So we can point to that to argue why we're not veto-ing transitions between Documents with different COOP policies.

css-view-transitions-2/Overview.bs
```diff
@@ -414,3 +411,25 @@ The <dfn attribute for=PageRevealEvent>viewTransition</dfn> [=getter steps=] are
...
+
+1. A same-origin navigation might still occur via a cross-origin redirect, e.g. <code>https://example.com</code>
+	links to <code>https://auth-provider.com/</code> which redirects back to <code>https://example.com/loggedin</code>.
```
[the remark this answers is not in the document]
noamr: Added a note

css-view-transitions-1/Overview.bs
```diff
@@ -1049,17 +1049,17 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			which causes the transition to [=skip the view transition|skip=].
 			[Discussion of this behavior](https://github.com/w3c/csswg-drafts/issues/8045).
 
-		: <dfn>old state captured steps</dfn>
+		: <dfn>process old state captured</dfn>
```
khushalsagar, line 1052: We need to update the call-sites which invoke this algorithm too.

khushalsagar (approved):

css-view-transitions-1/Overview.bs
noamr: They're all updated in this PR.

Issue #8805, opened 2023-05-05
[css-view-transitions-2] Script event on new Document for cross-document ViewTransitions


khushalsagar:
View Transition API needs to provide an event on the new Document to enable customization of the transition based on the URL for the old Document and its state when the transition was initiated. For example, the old Document could have changed from user interaction or the transition could depend on which click initiated the navigation.

The proposed IDL is as follows:

```
[Exposed=Window]
interface ViewTransitionBeforeAnimationEvent : Event {
  // The URL being navigated from,.
  readonly attribute USVString url;

  // Opaque contextual information passed from the old Document.
  // This must be a serializable object : https://developer.mozilla.org/en-US/docs/Glossary/Serializable_object.
  attribute any context;

  // The transition object associated with this navigation.
  ViewTransitionOnNavigation transition;
};
```

The [ViewTransition](https://drafts.csswg.org/css-view-transitions-1/#the-domtransition-interface) interface is also split into a base IDL with functionality common to same-document and cross-document transitions.

```
[Exposed=Window]
interface ViewTransitionBase {
  readonly attribute Promise ready;
  readonly attribute Promise finished;
  undefined skipTransition();
};

[Exposed=Window]
interface ViewTransition : ViewTransitionBase {
   readonly attribute Promise updateCallbackDone;
};

[Exposed=Window]
interface ViewTransitionOnNavigation : ViewTransitionBase {
};
```

The following is sample code using this event:

```js
document.addEventListener("viewtransitionbeforeanimation", (event) => {
  // Cancel the transition (based on old URL) if needed.
  if (shouldNotTransition(event.url)) {
    event.preventDefault();
    return;
  }

  const transition = event.transition;
  const info = event.info;

  // Add render-blocking resources to delay the first paint and transition
  // start. This can be customized based on the old Document state when the
  // transition was initiated.
  markRenderBlockingResources(info);

  // The `ready` promise resolves when the pseudo-elements have been generated
  // and can be used to customize animations via script.
  transition.ready.then(() => {
    document.documentElement.animate(...,
       {
         // Specify which pseudo-element to animate
         pseudoElement: "::view-transition-new(root)",
       }
    );

    // Remove viewTransitionNames tied to this transition.
    thumbnail.style.viewTransitionName = "none";
  });
  
  // The `finished` promise resolves when all animations for the transition are
  // finished or cancelled and the pseudo-elements have been removed.
  transition.finished.then(() => { ... });
});
```

[...]

PR #9090, opened 2023-07-19
[css-view-transitions-2] Apply resolutions from F2F

### LINKED ISSUE #8784: [css-view-transitions-2] Should ViewTransitions be triggered for reloads?
A reload is technically a cross-document same-origin navigation, but its unclear whether authors want a ViewTransition in such cases. One of the sites we're working with mentioned they don't want any animation on reloads.

Note that since the UA CSS does a cross-fade between root images using plus-lighter, that animation is a no-op if there is no change in visual content. But if the site overrides this for a different default (like a slide), they'll need to detect reloads to get back the no-op cross-fade.

We can make this configurable so the author can decide whether reloads from the current Document trigger a ViewTransition.

### LINKED ISSUE #8889: [css-view-transitions-2] security/privacy considerations with cross-origin css
#8048 and linked feature suggest to expose certain information about the navigation (navigation type, previous/next URL) to CSS, e.g. in the form of media queries.

Some security concerns about it that should be discussed and resolved:
As a general rule we don't consider 3rd party CSS as "safe", it can already exfiltrate a lot of information and alter the behavior of the page. However, it is still handled differently than scripts, e.g. `style-src` and `script-src` are separate directives.

When exposing navigation type and previous/next URL to CSS, we give CSS direct observation into the cross-document world. 
The threat model: An existing 3p CSS would now be able to exfiltrate (e.g. by an invisible background image) information about the referrer of its host document.

We should resolve how much this should be a constraint and affect our design, if at all. If we decide that this is not a problem, great!

One potential mitigation would be to have some kind of HTML opt-in to make CSS "navigation aware", to avoid making these changes something that affects existing sites.

noamr:
- Add wording in security about 3p CSS
- Fire reveal event before rAF, inside update the rendering

Closes #8889 (see [resolution](https://github.com/w3c/csswg-drafts/issues/8889#issuecomment-1640752547))
See #8805 (see [partial resolution](https://github.com/w3c/csswg-drafts/issues/8805#issuecomment-1640688426))

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -36,7 +36,6 @@ spec:css-view-transitions-1;
 spec:dom; type:dfn; text:document
 spec:css22; type:dfn; text:element
 spec:html
-	text: current session history entry; type: dfn;
 	text: latest entry; type: dfn;
 	text: was created via cross-origin redirects; type: dfn;
 	text: run the animation frame callbacks; type: dfn;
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -282,37 +292,26 @@ The <dfn attribute for=PageRevealEvent>viewTransition</dfn> [=getter steps=] are
 		Prepend a step at the beginning of the task [=queue a global task|queued=] on |navigable|'s [=active window=]
 		when <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-the-history-step">applying the history step</a> (14.11.1, <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#updating-the-traversable:queue-a-global-task-3">here</a>):
 
-		If |changingNavigationContinuation| update-only is false, then [=setup outbound cross-document view transition=] given |oldDocument|, |newDocument| and the remaining steps and return from these steps.
+		If |changingNavigationContinuation| update-only is false and |targetEntry| is not |navigable|'s [=current session history entry=], then [=setup outbound cross-document view transition=] given |oldDocument|, |newDocument| and the remaining steps and return from these steps.
```
khushalsagar, line 295: > |targetEntry| is not |navigable|'s [=current session history entry=]

This is to veto reloads right? Can you add a note with a link to the issue where this was resolved. Also, the wording here sounds like it would apply to "history.replace" but I might be misunderstanding this.

khushalsagar (approved):

css-view-transitions-2/Overview.bs
noamr: OK removed from this PR. merge?

PR #9093, opened 2023-07-19
[css-view-transitions-1] Remove resolved issues

khushalsagar:


css-view-transitions-1/Overview.bs
```diff
@@ -1684,8 +1684,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			then skip painting |descendant|.
 
 			Note: This is necessary since the descendant will generate its own snapshot which will be displayed and animated independently.
-
-		Issue: Specify what happens with 'mix-blend-mode'.
```
vmpstr, line 1688: We need to add text to copy mix-blend-mode, right? Or is that in a separate PR?

khushalsagar: Already added in https://github.com/w3c/csswg-drafts/pull/9089/files.

PR #9096, opened 2023-07-19
[css-borders-4] Fixed and added definitions for box-shadow-offset: none

### LINKED ISSUE #8567: [css-backgrounds-4] Allow repeating `none` in `box-shadow-offset`

SebastianZ:
I've added the definition for the `none` value of `box-shadow-offset` and corrected the computed value of that property to account for the `none` value now being part of the shadow list.
The `box-shadow` shorthand was adjusted accordingly.

Fixes #8567

Sebastian

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -633,9 +633,8 @@ where both values are described as <<length>> values.
 <dl>
   <dt><dfn id="shadow-offset-none">none</dfn>
   <dd>
-    No box shadow will be created.
-    Any other box shadow properties specified for that shadow in the list
-    of box shadows have no effect.
+    The shadow will not be rendered.
+    The values of other box shadow properties corresponding to this shadow have no effect.
 
   <dt><dfn id="shadow-offset-x">1st <<length>></dfn>
   <dd>
@@ -762,8 +761,10 @@ Animation type: see individual properties
   <p>Each shadow is given as a <<spread-shadow>>,
   outlining the 'box-shadow-offset', and optional values for the 'box-shadow-blur',
   'box-shadow-spread', 'box-shadow-color', and 'box-shadow-position'.
-  If the offset is ''box-shadow-offset/none'',
-  ''box-shadow-color'' is treated as ''transparent''.
+  Omitted lengths are '0';
+  omitted colors default to ''transparent'' when the specified offset is ''box-shadow-offset/none''
+  and to ''currentcolor'' otherwise.
+
   <pre class=prod>
   <dfn><<spread-shadow>></dfn> = <<'box-shadow-color'>>? &amp;&amp; [ <<'box-shadow-offset'>> [ <<'box-shadow-blur'>> <<'box-shadow-spread'>>? ]? ] &amp;&amp; <<'box-shadow-position'>>?</pre>
 
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -617,16 +617,24 @@ Initial: none
 Applies to: all elements
 Inherited: no
 Percentages: N/A
-Computed value: either 'none' or a list,
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
 Animation type: by computed value
 </pre>
 
 <p>The 'box-shadow-offset' property defines one or more drop shadow offsets.
-The property accepts a comma-separated list of horizontal and vertical offset pairs,
+The property accepts a comma-separated list.
+Each item in that list can either be the ''box-shadow-offset/none'' value,
+which indicates no shadow, or a pair of horizontal and vertical offsets,
 where both values are described as <<length>> values.
 
 <dl>
+  <dt><dfn id="shadow-offset-none">none</dfn>
+  <dd>
+    No box shadow will be created.
+    Any other box shadow properties specified for this box shadow have no
+    effect.
+
   <dt><dfn id="shadow-offset-x">1st <<length>></dfn>
   <dd>
     Specifies the <dfn>horizontal offset</dfn> of the shadow.
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -749,14 +749,12 @@ Initial: none
 Applies to: all elements
 Inherited: no
 Percentages: N/A
-Computed value: either the keyword ''box-shadow-offset/none'' or
-    a list, each item consisting of four absolute lengths
-    plus a computed color and optionally also a ''box-shadow-position/inset'' keyword
+Computed value: see individual properties
 Animation type: by computed value,
-    treating ''box-shadow-offset/none'' as a zero-item list
-    and appending blank shadows (''transparent 0 0 0 0'')
-    with a corresponding ''box-shadow-position/inset'' keyword as needed
-    to match the longer list
+    letting ''box-shadow-offset/none'' create a blank shadow
+    (''transparent 0 0 0 0'') when interpolated with non-'none' values
+    and appending blank shadows with a corresponding
+    ''box-shadow-position/inset'' keyword as needed to match the longer list
     if the shorter list is otherwise compatible with the longer one
 </pre>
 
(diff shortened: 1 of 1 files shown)


Loirooriol (changes requested):

css-borders-4/Overview.bs
```diff
@@ -741,14 +749,12 @@ Initial: none
 Applies to: all elements
 Inherited: no
 Percentages: N/A
-Computed value: either the keyword ''box-shadow-offset/none'' or
-    a list, each item consisting of four absolute lengths
-    plus a computed color and optionally also a ''box-shadow-position/inset'' keyword
+Computed value: see individual properties
 Animation type: by computed value,
```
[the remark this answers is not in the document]
SebastianZ: How about now?

css-borders-4/Overview.bs
```diff
@@ -617,16 +617,29 @@ Initial: none
...
 Inherited: no
 Percentages: N/A
-Computed value: either 'none' or a list,
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
-Animation type: by computed value
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
+Animation type: by computed value,
+    letting ''box-shadow-offset/none'' create a blank shadow
+    (''transparent 0 0 0 0'') when interpolated with non-'none' values
```
Loirooriol, line 624: I don't think that works, the computed value of a property can affect another, but if the computed value at the midpoint between `none` and `10px 10px` is `5px 5px`, then `5px 5px` shouldn't by itself force `box-shadow-color` to get a `transparent`.

As I said, I think handling `transparent` in the shorthand expansion will suffice to keep the old behavior.

css-borders-4/Overview.bs
```diff
@@ -617,16 +617,29 @@ Initial: none
...
+    and appending blank shadows with a corresponding
```
Loirooriol, line 625: I think this should actually be handled in each longhand, i.e. the animation type of `box-shadow-offset` should just add `none` values, `box-shadow-blur` add `0`, etc.

css-borders-4/Overview.bs
```diff
@@ -617,16 +617,29 @@ Initial: none
...
+Each item in that list can either be the ''box-shadow-offset/none'' value,
+which indicates no shadow, or a pair of horizontal and vertical offsets,
 where both values are described as <<length>> values.
 
 <dl>
+  <dt><dfn id="shadow-offset-none">none</dfn>
+  <dd>
+    No box shadow will be created.
+    Any other box shadow properties specified for this box shadow have no
+    effect.
```
Loirooriol, line 641: Maybe make it clearer that this only refers to the current shadow? That is, if `none` is mixed with other values, the others can still create shadows.

Loirooriol (changes requested):

css-borders-4/Overview.bs
SebastianZ: Ok, I've moved the handling of `transparent` to the shorthand expansion now.

css-borders-4/Overview.bs
SebastianZ: I tried to clarify now that `none` only affects the current shadow.

SebastianZ:
While I was at it, I also removed "either the `none` value, which indicates no shadows, or a comma-separated list of shadows, " from the `box-shadow` shorthand as it isn't correct anymore.

css-borders-4/Overview.bs
```diff
@@ -617,16 +617,26 @@ Initial: none
...
 Inherited: no
 Percentages: N/A
-Computed value: either 'none' or a list,
-	each item a pair of offsets (horizontal and vertical) from the element‘s box
-Animation type: by computed value
+Computed value: list, each item either 'none' or a pair of offsets
+  (horizontal and vertical) from the element‘s box
+Animation type: by computed value,
+  treating ''box-shadow-offset/none'' as ''0 0''
+  when interpolated with non-'none' values.
```
Loirooriol, line 624: This needs to say that if the lists have different lengths, `none` values are added to the short list until the have the same length.

Something analogous for the other longhands. Possibly consider a new animation type like "defaultable list" or something, since there already is a "repeatable list".

css-borders-4/Overview.bs
```diff
@@ -617,16 +617,26 @@ Initial: none
...
+Each item in that list can either be the ''box-shadow-offset/none'' value,
+which indicates no shadow, or a pair of horizontal and vertical offsets,
 where both values are described as <<length>> values.
 
 <dl>
+  <dt><dfn id="shadow-offset-none">none</dfn>
+  <dd>
+    No box shadow will be created.
+    Any other box shadow properties specified for that shadow in the list
+    of box shadows have no effect.
```
Loirooriol, line 638: Still seems confusing. Maybe something like

```suggestion
    The shadow will not be rendered.
    The values of other box shadow properties corresponding to this shadow have no effect.
```

Probably @fantasai can come up with something better.

css-borders-4/Overview.bs
```diff
@@ -741,25 +751,19 @@ Initial: none
...
-  or a comma-separated list of shadows, ordered front to back.
+  The property accepts a comma-separated list of shadows,
+  ordered front to back.
 
   <p>Each shadow is given as a <<spread-shadow>>,
   outlining the 'box-shadow-offset', and optional values for the 'box-shadow-blur',
   'box-shadow-spread', 'box-shadow-color', and 'box-shadow-position'.
-  Omitted lengths are 0; omitted colors default to the ''currentcolor'' value.
+  If the offset is ''box-shadow-offset/none'',
+  ''box-shadow-color'' is treated as ''transparent''.
```
Loirooriol, line 766: "Treated as" sounds like changing the used value of `box-shadow-color`, which can be a possible approach, but the shorthand wouldn't be the correct place for it, and it would need more complication than that.

What I was thinking (maybe it's not the best behavior, but it's simple to specify for now, better approaches can be considered in a follow-up) is that the specified value of `box-shadow-color` would be `transparent`, so no "treating as" necessary.

Don't remove the default length either.

```suggestion
  Omitted lengths are 0;
  omitted colors default to the ''transparent'' value when the specified offset is ''none''
  and to ''currentcolor'' otherwise.
```

Loirooriol (changes requested):

css-borders-4/Overview.bs
SebastianZ: Note that I aligned that with the [`none` value for `animation-name`](https://drafts.csswg.org/css-animations-1/#valdef-animation-name-none).
And I believe it's clear enough that this means that other shadows in the list won't be affected by that.

css-borders-4/Overview.bs
SebastianZ: Pushed this change with a slight adjustment.

css-borders-4/Overview.bs
SebastianZ: Ok, added your phrashing, though both seem equally good to me.

css-borders-4/Overview.bs
SebastianZ: I meant to add the everything related to the coordinating list property group in https://github.com/w3c/csswg-drafts/pull/8613.

I was also thinking about the animation type, as "repeatable list" doesn't fit here. @fantasai Any ideas?

Loirooriol (approved):
I guess this is fine as long as you address interpolation of lists with different lengths in a follow-up.

I also plan to file an issue to discuss other possible ways of affecting the color when interpolating from a `none` offset to something else.

Issue #8857, opened 2023-05-19
[css-transitions-2] Put discrete transitions behind new syntax for compatibility


flackr:
# Problem

@josepharhar implemented #4441 and discovered some compat issues. While it seems to be true that developers don't set up transitions on properties that only have discrete values, there are many cases where developers have transitions set up on properties which sometimes have discrete values. E.g. going to/from `auto` currently results in a discrete interpolation. This means that changes to/from auto currently do not start transitions but with transitions of discrete properties they now do.

Note: We want to support smooth interpolation to/from auto (See #626). If/when we add this, if it just starts intepolating to/from auto and thus starting transitions we would run into the same set of compat issues.

## Example issues

@josepharhar looked through 30 affected sites and found 3 regressions:
[stackoverflow.com](http://stackoverflow.com/)
With discrete transitions for width and height on page load the orange text doesn’t show up for a frame or two:
![image3](https://github.com/w3c/csswg-drafts/assets/366761/3351b136-716d-4787-873c-f3513ccf7e16)

[hpe.com](http://hpe.com/)
With discrete transitions for max-height, the accordions at the bottom don’t slide closed but instead suddenly close with a delay after clicking:
![image1](https://github.com/w3c/csswg-drafts/assets/366761/86497992-9c48-411f-8a4b-128006d2089b)

[fender.com](http://fender.com/)
With discrete transitions for top, the menu bar at the top of the page doesn’t stick to the top of the page but instead teleports to the top after a delay when scrolling:
![image2](https://github.com/w3c/csswg-drafts/assets/366761/fb2362e4-56e2-4220-ad4b-1ea2ca99e9ca)

# Solutions

From what we have seen so far, the compat issues are only for properties which currently sometimes have interpolable transitions (e.g. length properties which are interpolable when specified but not when `auto`). This means that technically the only transitions which need to be avoided may be these. However, it seems ergonomically awkward to make the developer need to think about whether the property is sometimes interpolable and sometimes not. As such, I think we should come up with a new syntax for opting in to discrete transitions. A side benefit, is that explicitly specifying discrete also provides a way to opt in to discrete transitions on `all` without needing a keyword like [discrete-too](https://github.com/w3c/csswg-drafts/issues/4441#issuecomment-545190019).

For comparison, the code necessary to animate out a dialog (animating display and overlay discretely, and opacity) is shown.

Some options (vote for all that you think would be reasonable):

[...]

PR #9101, opened 2023-07-21
[css-transitions-2] Add transition-behavior property

josepharhar:
Fixes https://github.com/w3c/csswg-drafts/issues/8857

### DIFF
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -147,7 +147,7 @@ The syntax for specifying 'transition-behavior' is as follows:
 
 When ''transition-behavior-value/normal'' is specified, transitions will not be
 started for discrete properties, only for interpolable properties. When
-''allow-discrete'' is specified, animations will be started for discrete
+''allow-discrete'' is specified, transitions will be started for discrete
 properties as well as interpolable properties.
 
 ## The 'transition' Shorthand Property ## {#transition-shorthand-property}
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -103,6 +103,9 @@ Note: This is for compatibility with CSS Transitions Level 1
 Note this means that the ''tranisition-property/all'' keyword here
 maps to a subset of the 'all' [=shorthand property=]’s [=longhand properties=].
 
+Note: When values with a [=discrete=] [=animation type=] are transitioned, they
+flip at 50% progress.
+
 ## The 'transition-duration' Property ## {#transition-duration-property}
 
 The 'transition-duration' property specifies the [=iteration duration=]
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -145,7 +145,7 @@ The syntax for specifying 'transition-behavior' is as follows:
 
 <div class=prod><dfn><<transition-behavior-value>></dfn> = normal | allow-discrete</pre>
 
-When ''transition-behavior-value/normal'' is specified, animations will not be
+When ''transition-behavior-value/normal'' is specified, transitions will not be
 started for discrete properties, only for interpolable properties. When
 ''allow-discrete'' is specified, animations will be started for discrete
 properties as well as interpolable properties.
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -94,15 +94,6 @@ In CSS Transitions Level 2, when comparing the [=before-change style=] and
  [=animation type=] that is [=discrete=].</li>
 </ol>
 
-However, the ''transition-property/all'' keyword
-continues to expand only to all [=transitionable=] properties
-whose [=animation type=] is not [=discrete=].
-
-Note: This is for compatibility with CSS Transitions Level 1
-(and because authors will rarely want [=discrete=] properties to transition at 50% by default).
-Note this means that the ''tranisition-property/all'' keyword here
-maps to a subset of the 'all' [=shorthand property=]’s [=longhand properties=].
-
 Note: When values with a [=discrete=] [=animation type=] are transitioned, they
 flip at 50% progress.
 
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -155,7 +155,7 @@ properties as well as interpolable properties.
 The syntax for specifying 'transition' is as follows:
 
 <div class="prod">
-  <dfn type id="single-transition">&lt;single-transition&gt;</dfn> = <<transition-behavior-value>> || [ ''transition-property/none'' | <<single-transition-property>> ] || <<time>> || <<easing-function>> || <<time>>
+  <dfn type id="single-transition">&lt;single-transition&gt;</dfn> = [ ''transition-property/none'' | <<single-transition-property>> ] || <<time>> || <<easing-function>> || <<time>> || <<transition-behavior-value>>
 </div>
 
 
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -129,7 +129,7 @@ started or not for discrete properties.
 
 <pre class="propdef">
   Name: transition-behavior
-  Value: normal | allow-discrete
+  Value: <<transition-behavior-value>>#
   Initial: normal
   Applies to: All elements
   Inherited: no
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -152,7 +152,7 @@ properties as well as interpolable properties.
 
 ## The 'transition' Shorthand Property ## {#transition-shorthand-property}
 
-The syntax for specifying 'transition' is as follows:
+The syntax for specifying an item in the 'transition' shorthand is as follows:
 
 <div class="prod">
   <dfn type id="single-transition">&lt;single-transition&gt;</dfn> = [ ''transition-property/none'' | <<single-transition-property>> ] || <<time>> || <<easing-function>> || <<time>> || <<transition-behavior-value>>
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -141,9 +141,7 @@ started or not for discrete properties.
 
 The syntax for specifying 'transition-behavior' is as follows:
 
-<div class="prod">
-  <dfn type id="transition-behavior-value">&lt;transition-behavior-value&gt;</dfn> = [ 'normal' | 'allow-discrete' ]
-</div>
+<div class=prod><dfn><<transition-behavior-value>></dfn> = normal | allow-discrete</pre>
 
 When 'normal' is specified, animations will not be started for discrete
 properties, only for interpolable properties. When 'allow-discrete' is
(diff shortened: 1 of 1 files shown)


css-transitions-2/Overview.bs
```diff
@@ -120,6 +122,42 @@ the reversing shortening factor.
...
 of the transition's associated [=animation effect=].
 
+## The 'transition-behavior' Property ## {#transition-behavior-property}
+
+The 'transition-behavior' property specifies whether transitions will be
+started or not for discrete properties.
+
+<pre class="propdef">
+  Name: transition-behavior
+  Value: normal | allow-discrete
```
dbaron, line 132: This needs to allow a list of values, not just a single value.  To do that you should probably make this `<transition-behavior-value>#`.  See https://drafts.csswg.org/css-backgrounds-3/#background-attachment for an example of what this should look like (and probably example bikeshed syntax too).

css-transitions-2/Overview.bs
```diff
@@ -120,6 +122,42 @@ the reversing shortening factor.
...
+  Percentages: N/A
+  Computed value: as specified
+  Canonical order: per grammar
+  Animation type: not animatable
+</pre>
+
+The syntax for specifying 'transition-behavior' is as follows:
+
+<div class="prod">
+  <dfn type id="transition-behavior-value">&lt;transition-behavior-value&gt;</dfn> = [ 'normal' | 'allow-discrete' ]
```
dbaron, line 145: This could be a bit simpler, see the source of css-backgrounds-3 for an example.  Though I admit it matches the complexity in css-transitions-1...

css-transitions-2/Overview.bs
```diff
@@ -88,7 +88,9 @@ In CSS Transitions Level 2,
 property values are [=transitionable=]
 <em>unless</em> they have an [=animation type=] that is [=not animatable=].
 Values with a [=discrete=] [=animation type=] <em>are</em> [=transitionable=],
-and flip at 50% progress (<var>p</var> = 0.5).
+and flip at 50% progress (<var>p</var> = 0.5) if the 'transition-behavior' is
```
dbaron, line 91: To match the terminology for the correct item in the list in css-transforms-1, this should probably refer to the "matching transition behavior".

css-transitions-2/Overview.bs
```diff
@@ -88,7 +88,9 @@ In CSS Transitions Level 2,
 property values are [=transitionable=]
 <em>unless</em> they have an [=animation type=] that is [=not animatable=].
```
dbaron, line 89: I think this paragraph should be rewritten a bit more fully to reflect the new approach.

css-transitions-2/Overview.bs
```diff
@@ -120,6 +122,42 @@ the reversing shortening factor.
...
+properties, only for interpolable properties. When 'allow-discrete' is
+specified, animations will be started for discrete properties as well as
+interpolable properties.
+
+## The 'transition' Shorthand Property ## {#transition-shorthand-property}
+
+The syntax for specifying 'transition' is as follows:
+
+<div class="prod">
+  <dfn type id="single-transition">&lt;single-transition&gt;</dfn> = <<transition-behavior-value>> || [ ''transition-property/none'' | <<single-transition-property>> ] || <<time>> || <<easing-function>> || <<time>>
```
dbaron, line 158: I think it's probably best to put the `<<transition-behavior-value>>` last; the order does affect the order of serialization that the spec requires.

css-transitions-2/Overview.bs
flackr: It might be simpler to leave out the "and flip at 50% progress" as this is already defined in web-animations-1. The important part is whether they are transitionable.

If you look at the [definition from css-transitions-1](https://www.w3.org/TR/css-transitions-1/#transitionable):
> When comparing the [before-change style](https://www.w3.org/TR/css-transitions-1/#before-change-style) and [after-change style](https://www.w3.org/TR/css-transitions-1/#after-change-style) for a given property, the property values are transitionable if they have an [animation type](https://drafts.csswg.org/web-animations-1/#animation-type) that is neither [not animatable](https://drafts.csswg.org/web-animations-1/#not-animatable) nor [discrete](https://drafts.csswg.org/web-animations-1/#discrete).

We could update this to something like:

When comparing the [before-change style](https://www.w3.org/TR/css-transitions-1/#before-change-style) and [after-change style](https://www.w3.org/TR/css-transitions-1/#after-change-style) for a given property, the property values are transitionable if:
- They have an [animation type](https://drafts.csswg.org/web-animations-1/#animation-type) that is neither [not animatable](https://drafts.csswg.org/web-animations-1/#not-animatable) nor [discrete](https://drafts.csswg.org/web-animations-1/#discrete), or
- The transition-behavior is allow-discrete and they have an [animation type](https://drafts.csswg.org/web-animations-1/#animation-type) that is [discrete](https://drafts.csswg.org/web-animations-1/#discrete).

css-transitions-2/Overview.bs
flackr: We might want to leave the normal value out of the shorthand - it is the initial value so unnecessary to be specified, and it would prevents its specification in other longhands in the future.

dbaron: I think leaving `normal` out of the shorthand is an interesting idea -- but I also think it's not something the WG has done before for any shorthand, so if we think it's worth doing, I think we should bring it up explicitly to the WG to see if other folks have feedback on the idea.

dbaron: I realize that `transition-property` takes `<custom-ident>` values, which means that the spec should probably also have wording excluding the two new keywords from that `<custom-ident>` as described in https://drafts.csswg.org/css-values-4/#identifier-value , although that change does have (low, I think) compatibility risk.

Or, alternatively, we could choose *not* to follow that advice and exclude the keywords, in which case we probably *do* want to leave this in the first position.

css-transitions-2/Overview.bs
josepharhar: done, thanks!

css-transitions-2/Overview.bs
josepharhar: Thanks rob! I replaced this paragraph with your text

css-transitions-2/Overview.bs
josepharhar: I rewrote the whole thing as per rob's suggestion

css-transitions-2/Overview.bs
josepharhar: > I think it's probably best to put the <<transition-behavior-value>> last; the order does affect the order of serialization that the spec requires.

So instead of doing `transition: allow-discrete display 2s` it should be `transition: display 2s allow-discrete`? Now that the value is called "allow-discrete" I'm more ok with this. When it was just "discrete" or "animatable" I felt more strongly that it should come first. What do you think rob?

> the spec should probably also have wording excluding the two new keywords from that `<custom-ident>`

If we put the keyword at the end, then this wouldn't be needed right?

dbaron: I don't think the ordering is relevant.  This affects the parsing rules -- and effectively without that exclusion the spec would need prose rules for how to distinguish the values, given that https://drafts.csswg.org/css-values-4/#component-combinators explicitly says that the ordering around `||` does *not* matter.  (This implies, I think, that it doesn't imply any rule about breaking ties during parsing when more than one option of an `||` matches.  Though maybe that implication wasn't intended.  But I think it does imply that it's a specification error if more than one option around an `||` matches and the spec doesn't say how to resolve the conflict.)

josepharhar: > given that https://drafts.csswg.org/css-values-4/#component-combinators explicitly says that the ordering around || does not matter

Oh wow, so `transition: 2s color` works the same as `transition: color 2s`!? It looks like this is true after I tried a basic test in chrome.

I moved it to the end of the list

css-transitions-2/Overview.bs
```diff
@@ -120,6 +124,40 @@ the reversing shortening factor.
...
+<div class=prod><dfn><<transition-behavior-value>></dfn> = normal | allow-discrete</pre>
+
+When ''transition-behavior-value/normal'' is specified, animations will not be
+started for discrete properties, only for interpolable properties. When
+''allow-discrete'' is specified, animations will be started for discrete
+properties as well as interpolable properties.
+
+## The 'transition' Shorthand Property ## {#transition-shorthand-property}
+
+The syntax for specifying 'transition' is as follows:
```
dbaron, line 155:
```suggestion
The syntax for specifying an item in the 'transition' shorthand is as follows:
```

css-transitions-2/Overview.bs
```diff
@@ -120,6 +124,40 @@ the reversing shortening factor.
...
+  Computed value: as specified
+  Canonical order: per grammar
+  Animation type: not animatable
+</pre>
+
+The syntax for specifying 'transition-behavior' is as follows:
+
+<div class=prod><dfn><<transition-behavior-value>></dfn> = normal | allow-discrete</pre>
+
+When ''transition-behavior-value/normal'' is specified, animations will not be
```
dbaron, line 148:
```suggestion
When ''transition-behavior-value/normal'' is specified, transitions will not be
```

css-transitions-2/Overview.bs
```diff
@@ -120,6 +124,40 @@ the reversing shortening factor.
...
+started for discrete properties, only for interpolable properties. When
+''allow-discrete'' is specified, animations will be started for discrete
```
dbaron, line 150:
```suggestion
''allow-discrete'' is specified, transitions will be started for discrete
```

css-transitions-2/Overview.bs
```diff
@@ -84,11 +84,15 @@ the transition was generated (e.g. 'border-left-width').
 ISSUE: Update the defining instance of [=transitionable=]
 once it is ported from Level 1.
 
-In CSS Transitions Level 2,
-property values are [=transitionable=]
-<em>unless</em> they have an [=animation type=] that is [=not animatable=].
-Values with a [=discrete=] [=animation type=] <em>are</em> [=transitionable=],
-and flip at 50% progress (<var>p</var> = 0.5).
```
dbaron, line 91: I think it's worth adding this back as a note.  In bikeshed that should be doable by making it its own paragraph beginning with "Note: ", though I think in full it probably needs to be something like: "Note: When values with a [=discrete=] [=animation type] are transitioned, they flip at 50% progress."

dbaron (commented):
This basically looks good, just a few comments:

css-transitions-2/Overview.bs
josepharhar: done, thanks!

josepharhar:
In my last commit ce5d6d3 I removed a note about transition:all which is no longer true based on the way I implemented the new CSS property: https://chromium-review.googlesource.com/c/chromium/src/+/4616155/14/third_party/blink/web_tests/external/wpt/css/css-transitions/all-with-discrete.tentative.html

@dbaron do you think this is ok?

dbaron (approved):
Good catch.  I agree that `all` combined with `allow-discrete` should actually allow discrete transitions!

PR #9105, opened 2023-07-22
[css-page-floats-3]: editorial: add all clear values to summary

keithamus:
[css-page-floats-3] 

The page-floats-3 spec does not enumerate all possible clear values in the property summary. The description below the summary details the available properties `both-inline`, `both-block` and `both`, but these aren't in the summary.

### DIFF
--- a/css-page-floats-3/Overview.bs
+++ b/css-page-floats-3/Overview.bs
@@ -464,7 +464,7 @@ The 'clear' property</h2>
 
   <pre class="propdef">
     Name: clear
-    Value: inline-start | inline-end | block-start | block-end | left | right | top | bottom | none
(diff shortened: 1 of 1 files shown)


keithamus:
Looks like this relates to https://github.com/w3c/csswg-drafts/issues/7350. I skimmed the comments, I'm not sure what the ideal resolution would be here, but this bit me as I implemented `clear` from this spec.

fantasai (approved):
Looks like an oversight indeed.

We don't yet have a CSS Block Layout module, we're still relying on CSS2.1 for that. (Early drafts of it were named css-box-3, but that stopped being maintained a couple decades ago, and it's been repurposed as the CSS Box Model module.)

Issue #4559, opened 2019-12-04
[css-values] Proposal: add sibling-count() and sibling-index()


argyleink:
#### Problem
Currently **we can query** a child based on it's [child index position](https://www.w3.org/TR/selectors-4/#the-nth-child-pseudo) or query on children length (with some [complex syntax](https://github.com/w3c/csswg-drafts/issues/1176)), but **we can't use** the index or length as values in our styles. Feels like they're known, but not accessible. 

so we end up doing stuff like this:
```css
.el {
  --delay: calc(var(--i, 1) * 400ms);
  animation: fadeIn 1000ms var(--delay) forwards;
  width: 100px;
  height: 100px;
  background: darkorchid;
  margin: 20px;
  opacity: 0;
}

.el:nth-child(2) {
  --i: 2;
}

.el:nth-child(3) {
  --i: 3;
}

@keyframes fadeIn {
  100% { opacity: 1; }
}
```
> above source from [Codepen](https://codepen.io/michellebarker/pen/VwYYpqw)

<br>

## Proposal
2 new functions: `sibling-count()` and `sibling-index()`. These should report values based on element count, not node count.

### `sibling-count()`
This function returns the total length of sibling elements as a number, similar to `node.childElementCount` ([docs](https://www.w3schools.com/jsref/prop_element_childelementcount.asp)) or `node.children.length` but accessible from a sibling. Consider this like a child asking "how many siblings they have". 

##### example usage
> Dynamic background color that changes based on total children in the list
```css
ul > li {
  background-color: hsl(sibling-count() 50% 50%);
}
```

<br>

### `sibling-index()`
This function returns the contextual child index as a number. Similar to the value queried with `nth-child`, this would be a get() call for the contextual child's index in the tree. Consider this like a child asking "what position am I in this family".

##### example usage
> Dynamic background color that changes based on child index position, resulting in a stepped gradient effect
```css
ul > li {
  background-color: hsl(calc(sibling-index() * 10) 50% 50%);
}
```

<br>

#### All Together
> Example code is a slightly modified version of [Jake Archibald's comment here](https://github.com/w3c/csswg-drafts/issues/1869#issuecomment-549136307) 👍
```css
ul > li {
  /* Stagger from start to end */
  animation-delay: calc(sibling-index() * 100ms);

  /* Stagger from end to start */
  animation-delay: calc(sibling-count() - sibling-index() * 100ms);
}
```

<br><br>

#### Use Cases
[...]

PR #9134, opened 2023-08-01
[css-values-5] Introduce `sibling-index()` and `sibling-count()` (#4559)

bramus:
A first, and possible naive, attempt at speccing `sibling-index()` and `sibling-count()`. Very much open to feedback on how to do this properly.

### DIFF
--- a/css-values-5/Overview.bs
+++ b/css-values-5/Overview.bs
@@ -945,8 +945,47 @@ Generating/Caching Random Values: the <<random-caching-options>> value</h4>
 		</pre>
 	</div>
 
+<!--
+████████ ██       ████████ ██     ██ ████████ ██    ██ ████████        ██████   ███████  ██     ██ ██    ██ ████████ ████ ██    ██  ██████  
+██       ██       ██       ███   ███ ██       ███   ██    ██          ██    ██ ██     ██ ██     ██ ███   ██    ██     ██  ███   ██ ██    ██ 
+██       ██       ██       ████ ████ ██       ████  ██    ██          ██       ██     ██ ██     ██ ████  ██    ██     ██  ████  ██ ██       
+██████   ██       ██████   ██ ███ ██ ██████   ██ ██ ██    ██          ██       ██     ██ ██     ██ ██ ██ ██    ██     ██  ██ ██ ██ ██   ████
+██       ██       ██       ██     ██ ██       ██  ████    ██          ██       ██     ██ ██     ██ ██  ████    ██     ██  ██  ████ ██    ██ 
+██       ██       ██       ██     ██ ██       ██   ███    ██          ██    ██ ██     ██ ██     ██ ██   ███    ██     ██  ██   ███ ██    ██ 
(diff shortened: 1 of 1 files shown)


css-values-5/Overview.bs
```diff
@@ -945,8 +945,47 @@ Generating/Caching Random Values: the <<random-caching-options>> value</h4>
...
+
+<h3 id=random>
+Getting the number of siblings an element is part of: the ''sibling-count()'' function</h3>
+
 
+	The <dfn>sibling-count()</dfn> function resolves to the total number of
+	<a>inclusive siblings</a> an element is part of. That is, it resolves to 
+	the total number of direct children the element’s parent element contains.
 
+Issue: Describe when it actually resolves. Computed Value? Used Value?
```
andruud, line 988: Definitely resolves computed-value-time, no reason to wait any longer.

css-values-5/Overview.bs
```diff
@@ -945,8 +945,47 @@ Generating/Caching Random Values: the <<random-caching-options>> value</h4>
...
+	allow authors to take these factors into account in ''calc()'' expressions.
+
+	With these, authors can create things like staggered animation delays
+	or easily position elements on a circle.
+
+<h3 id=random>
+Getting the position of an element within its siblings: the ''sibling-index()'' function</h3>
+
+	The <dfn>sibling-index()</dfn> function resolves to the position
+	of an element within its group of <a>inclusive siblings</a>.
```
andruud, line 975: Might want to clarify that it's the inclusive _flat tree_ siblings, which is what we want (I assume).

css-values-5/Overview.bs
bramus: Added in af5ce47

css-values-5/Overview.bs
bramus: Clarified in 30de15b, but not sure if my wording is 100% correct.

lilles: We don't use flat tree order for :nth-child(). I think we want :nth-*() and sibling-index() to be consistent?

andruud: That's true.

css-values-5/Overview.bs
```diff
@@ -945,8 +945,51 @@ Generating/Caching Random Values: the <<random-caching-options>> value</h4>
...
+
+<h2 id=element-counting>
+Counting Elements</h2>
+
+	Authors often want to apply element specific styles
+	based on the position of the element within the tree.
+
+	The ''sibling-index()'' and ''sibling-count()'' functions	
+	(the <dfn export for=CSS>element counting functions</dfn>)
+	allow authors to take these factors into account in ''calc()'' expressions.
```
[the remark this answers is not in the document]
tabatkins: (Just a light rewrite to avoid implying that these are only valid in calc().)

tabatkins (changes requested):
A couple of nits. Also, I'd add a few small examples to show them off.

Loirooriol:
With 762eacee436fdcb1b11e7e482e6cdb3fee161317, I guess this can be closed.

bramus:
Yes, thanks for closing this one @Loirooriol. I was preoccupied with Other Important Things™ + some time OOO, so never got round to reviewing the suggested changes in this thread.

PR #9135, opened 2023-08-01
[css-view-transitions-1] Refactor VT old capture algorithm

vmpstr:
[css-view-transitions-1] Refactor VT old capture algorithm

We need to correctly set `captured in a view transition` before capturing the image, since the capture checks the boolean's value for descendant elements.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1314,6 +1314,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |usedTransitionNames| be a new [=/set=] of strings.
 
+		1. Let |captureElements| be a new [=/list=] of elements.
+
 		1. Let |document| be |transition|'s [=relevant global object's=] [=associated document=].
 
 		1. Set |transition|'s [=ViewTransition/initial snapshot containing block size=] to the [=snapshot containing block size=].
@@ -1338,6 +1340,16 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			1. If |usedTransitionNames| [=list/contains=] |transitionName|, then return failure.
 
 			1. [=set/Append=] |transitionName| to |usedTransitionNames|.
+
+			1. Set |element|'s [=captured in a view transition=] to true.
+
+			1. [=list/Append=] |element| to |captureElements|.
+
+				<div class="note">The algorithm continues in a separate loop to ensure that
+				[=captured in a view transition=] is set on all elements participating in
+				this capture before it is read by future steps in the algorithm.</div>
+
+		1. [=list/For each=] |element| in |captureElements|:
 
 			1. Let |capture| be a new [=captured element=] struct.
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1314,6 +1314,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |usedTransitionNames| be a new [=/set=] of strings.
 
+		1. Let |captureElements| be a new [=/list=] of elements.
+
 		1. Let |document| be |transition|'s [=relevant global object's=] [=associated document=].
 
 		1. Set |transition|'s [=ViewTransition/initial snapshot containing block size=] to the [=snapshot containing block size=].
```
khushalsagar, line 1321: While we're here, can you add a note in the loop below. Something like:

```
We iterate in paint order (back to front) to ensure that this order is cached in |namedElements|.
This defines the DOM order for ::view-transition-group pseudo-elements, such that the element at the back of the paint stack is the first child of ::view-transition.
```

css-view-transitions-1/Overview.bs
```diff
@@ -1364,6 +1376,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			1. Set |capture|'s [=captured element/old mix-blend-mode=] to the [=computed value=] of 'mix-blend-mode' on |element|.
 
 			1. Set |namedElements|[|transitionName|] to |capture|.
```
khushalsagar, line 1378: We probably need to redefine `transitionName` in this loop. Can simply be the computed value for `view-transition-name` on element.

css-view-transitions-1/Overview.bs
vmpstr: Done.

css-view-transitions-1/Overview.bs
vmpstr: Done.

css-view-transitions-1/Overview.bs
```diff
@@ -1322,6 +1324,12 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			and has a [=node document=] equal to to |document|,
 			in [paint order](https://drafts.csswg.org/css2/#painting-order):
 
+			<div class=note>We iterate in paint order to ensure that this order is
```
khushalsagar, line 1327: nit: New line for each logical line.

css-view-transitions-1/Overview.bs
```diff
@@ -1339,6 +1347,16 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 			1. [=set/Append=] |transitionName| to |usedTransitionNames|.
 
+			1. Set |element|'s [=captured in a view transition=] to true.
+
+			1. [=list/Append=] |element| to |captureElements|.
+
+				<div class="note">The algorithm continues in a separate loop to ensure that
```
khushalsagar, line 1354: Same here.

khushalsagar (approved):

PR #9147, opened 2023-08-03
[css-images-4] Apply #8021 to gradient definitions in CSS Images 4

cdoublev:
It prevents switching positions of a color and its position in a gradient color stop.

This change was already applied in CSS Images 3 via #8021.


svgeesus:
@fantasai @tabatkins @LeaVerou looks like the ipr check is broken, which means I can't approve it.

PR #9181, opened 2023-08-11
[scroll-animations-1][web-animations-2]  Deferred start time

kevers-google:
[scroll-animations-1][web-animations-2] Deferred start time

When using a scroll-driven animation the start time of the animation is auto-aligned with the animation range boundary is not explicitly set.


web-animations-2/Overview.bs
```diff
@@ -800,6 +802,20 @@ The procedure to [=pause an animation=] needs to refer not only to the
 <ins cite="#amend-CE">Likewise, the procedure to [=pause an animation=] needs to include scheduling
 a task for updating [=custom effects=].</ins>
 
+Update the constraints for setting the seek time to only apply when using a
+monotonic timeline. A [=scroll-driven animation=] needs to defer setting the
+hold time until the animation range has been computed.
```
flackr, line 807: I think this needs to be more specific. For example, you probably mean to update step 14 of the procedure to pause an animation https://drafts.csswg.org/web-animations-1/#pause-an-animation such that we have a pending pause task until this additional constraint is satisfied.

web-animations-2/Overview.bs
```diff
@@ -2250,18 +2266,20 @@ Update the attribute type for currentTime.
 
 <h3 id="the-animation-interface">The <code>Animation</code> interface</h3>
 
-Update the startTime and currentTime of the Animation interface as follows:
+Update the startTime and currentTime of the Animation interface, and add
+rangeStatrt and rangeEnd as follows:
```
flackr, line 2270: s/rangeStatrt/rangeStart

web-animations-2/Overview.bs
```diff
@@ -2250,18 +2266,20 @@ Update the attribute type for currentTime.
...
 
 <pre class="idl">
 [Exposed=Window]
 partial interface Animation {
     attribute CSSNumberish?       startTime;
     attribute CSSNumberish?       currentTime;
+    attribute (TimelineRangeOffset or CSSNumericValue or CSSKeywordValue or DOMString)rangeStart;
```
flackr, line 2277: nit: Space after closing bracket before rangeStart.

web-animations-2/Overview.bs
```diff
@@ -2287,6 +2305,21 @@ descriptions with:
...
     > <a>set the current time</a> of this object to the new value.
 
+Add:
+
+> :   <dfn attribute for=Animation>rangeStart</dfn>
+> ::  Specifies the start of the <a>animation</a>’s [=animation attachment range=].
+>     Setting the attribute follows the same rules as the KeyframeAnimationOption
+>     <a href="#dom-keyframeanimationoptions-rangestart">rangeStart</a>.
+>     When reading the attribute, the returned value is either a
+>     {{TimelineRangeOffset}} or the {{DOMString}} normal.
```
flackr, line 2315: nit: Put normal in quotes, "normal".

flackr (changes requested):

scroll-animations-1/Overview.bs
```diff
@@ -145,9 +161,9 @@ spec: cssom-view-1; type: dfn;
 
 	Progress (the [=timeline/current time=]) for a [=scroll progress timeline=]
 	is calculated as:
-	<var>[=scroll offset=]</var> &div;
-	(<var>[=scrollable overflow=] size</var> &minus;
-		<var>[=scroll container=] size</var>)
+	<var ignore=''>[=scroll offset=]</var> &div;
+	(<var ignore=''>[=scrollable overflow=] size</var> &minus;
+		<var ignore=''>[=scroll container=] size</var>)
```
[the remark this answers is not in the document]
kevers-google: Suppressing bikeshed warnings about a local variable only being used once.  The use of var makes the links italicized, but otherwise serves no purpose.

web-animations-2/Overview.bs

scroll-animations-1/Overview.bs
```diff
@@ -895,9 +911,8 @@ spec: cssom-view-1; type: dfn;
 	any created [=scroll progress timelines=] or [=view progress timelines=] are collected into a <dfn export>stale timelines</dfn> set.
 	After step 7.14 if any timelines' [=named timeline ranges=] have changed,
 	these timelines are added to the [=stale timelines=] set.
-	If there are any [=stale timelines=], they now update their current time and associated ranges,
-	the set of [=stale timelines=] is cleared and
-	we run and we run an additional step to recalculate styles and update layout.
+	If there are any [=stale timelines=], they now update their current time and associated ranges, the set of [=stale timelines=] is cleared and
```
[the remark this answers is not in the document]
kevers-google: reverted change.

scroll-animations-1/Overview.bs
```diff
@@ -32,19 +38,29 @@ spec:web-animations-1;
...
 	type:method; text:requestAnimationFrame()
 spec: cssom-view-1; type: dfn;
 	text: overflow direction;
 	text: css layout box
+spec:css-writing-modes-4; type: dfn;
+	text:start
+	text:end
+spec:infra; type:dfn; text:user agent
+spec:selectors-4; type:dfn; text:selector
+spec:web-animations-2: type:dfn; text:auto-aligned start time
```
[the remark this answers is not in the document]
kevers-google: Obsolete.

scroll-animations-1/Overview.bs
```diff
@@ -938,13 +953,50 @@ spec: cssom-view-1; type: dfn;
...
+
+		1.  Set [=start time=] to |start offset| if [=effective playback rate=]
+			&ge; 0, and |end offset| otherwise.
+
+		1.  Clear [=hold time=].
+
+	Note: The [=start time=] is auto-aligned unless [=animation/current time=] or
+	[=start time=] are explicitly set via the web-animations API, which makes the
+	[=start time=] sticky.  Playing an animation implicitly resets the stickiness
+	of the [=start time=].
```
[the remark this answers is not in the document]
kevers-google: Move the procedure for updating an auto-aligned start time to web-animations-2 as a subsection under playing-an-animation.

scroll-animations-1/Overview.bs
```diff
@@ -917,9 +932,9 @@ spec: cssom-view-1; type: dfn;
 	in the rendered output.
 
 	Note: This section has no effect on forced style and layout
-	calculations triggered by {{Window/getComputedStyle()|getComputedStyle()}} or similar.
-	In other words, [=initially stale=] timelines are visible as such
-	through those APIs.
+	calculations triggered by {{Window/getComputedStyle()|getComputedStyle()}} or
+	similar. In other words, [=stale timelines|initially stale=] timelines are
+	visible as such through those APIs.
```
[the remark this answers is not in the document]
kevers-google: done

scroll-animations-1/Overview.bs
```diff
@@ -938,13 +953,50 @@ spec: cssom-view-1; type: dfn;
 	and the state of a scroll-driven animation
 	may be inconsistent in the composited frame.
 
-# Privacy Considerations
+## Updating an auto-aligned start time ## {#updating-auto-aligned-start-time}
 
-  There are no known privacy impacts of the features in this specification.
+	When updating timeline current time, the [=start time=] of any attached
```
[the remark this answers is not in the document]
kevers-google: Moved the process for deferred start time of an animation to a subsection under playing-an-animation. Fixed links between scroll-animations-1 and web-animations-2.  Event processing loop now links to auto-aligned start time procedure in web-animations-2 and procedure in web-animations-2 links back to scroll-animations-1 for when the procedure is called.

scroll-animations-1/Overview.bs
```diff
@@ -938,13 +953,50 @@ spec: cssom-view-1; type: dfn;
...
-  There are no known security impacts of the features in this specification.
+		1. If the animation does not have an [=auto-aligned start time=], abort
+			this procedure.
+
+		1. If [=play state=] is idle or [=hold time=] is resolved and
+			[=play state=] is not finished, abort this procedure.
+
+		1. Let |start offset| be the resolved scroll offset as a percentage
+			corresponding to the start of the [=animation attachment range=], or
+			0% if unresolved.
```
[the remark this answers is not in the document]
kevers-google: Added a step above to abort if the timeline is inactive (in web-animations-2).

web-animations-2/Overview.bs
```diff
@@ -388,15 +390,38 @@ After the step to assign <var>new effect</var> as <var>animation</var>'s
...
 >     each <a>inclusive descendant</a> of
 >     the animation's <a>associated effect</a>
 >     including rendering the first frame of any <a>keyframe
 >     effect</a><ins cite="#amend-CE"> or executing any <a>custom effects</a> associated with an
 >     <a>animation effect</a></ins>.
+>
+> *   the animation is associated with a [=timeline=]
+>     that is not [=inactive timeline|inactive=].
+>
+> *   the animation's [=hold time=] or [=animation/start time=] is resolved.
```
flackr, line 420: Arguably knowing the hold time and start time is one of the requirements for the first point - i.e. we can't "render" the first frame until we know this. But I suppose it doesn't hurt to call out explicitly :-).

web-animations-2/Overview.bs
```diff
@@ -388,15 +390,38 @@ After the step to assign <var>new effect</var> as <var>animation</var>'s
...
+>
 > *   the user agent has completed any setup required to begin the playback of
 >     each <a>inclusive descendant</a> of
 >     the animation's <a>associated effect</a>
 >     including rendering the first frame of any <a>keyframe
 >     effect</a><ins cite="#amend-CE"> or executing any <a>custom effects</a> associated with an
 >     <a>animation effect</a></ins>.
+>
+> *   the animation is associated with a [=timeline=]
+>     that is not [=inactive timeline|inactive=].
```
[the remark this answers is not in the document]
kevers-google: Agree that this should be a separate issue.  Unrelated to deferred start time.  Would you like me to file the issue?

web-animations-2/Overview.bs
kevers-google: ack

web-animations-2/Overview.bs
```diff
@@ -791,6 +809,36 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
...
+
+    1.  If the timeline is inactive, abort this procedure.
+
+    1.  If [=play state=] is idle, abort this procedure.
+
+    1.  If [=play state=] is paused, and [=hold time=] is resolved,
+        abort this procedure.
+
+    1.  Let |start offset| be the resolved scroll offset as a percentage
+        corresponding to the start of the [=animation attachment range=],
```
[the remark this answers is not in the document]

scroll-animations-1/Overview.bs
```diff
@@ -896,8 +913,7 @@ spec: cssom-view-1; type: dfn;
 	After step 7.14 if any timelines' [=named timeline ranges=] have changed,
 	these timelines are added to the [=stale timelines=] set.
 	If there are any [=stale timelines=], they now update their current time and associated ranges,
-	the set of [=stale timelines=] is cleared and
-	we run and we run an additional step to recalculate styles and update layout.
+	the set of [=stale timelines=] is cleared and we run an additional step to recalculate styles and update layout.
```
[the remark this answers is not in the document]
kevers-google: done.

flackr (approved):

fantasai (changes requested):
Hi @kevers-google, when you land this PR can you flatten all the clean-up commits into one commit (with an appropriate commit message) and keep the substantive changes separate in their own commit?

kevers-google:
Link / markup fixes pulled out into separate PR and merged.

web-animations-2/Overview.bs
```diff
@@ -138,10 +138,15 @@ urlPrefix: https://webidl.spec.whatwg.org/; type: dfn; spec: webidl
...
     text: execution contexts
+urlPrefix: https://drafts.csswg.org/scroll-animations-1/; type: dfn; spec: scroll-animations-1
+    text: html-processing-model-event-loops
 </pre>
 
 <pre class=link-defaults>
 spec:webidl; type:dfn; text:throw
+spec:scroll-animations-1; type:dfn;
+    text:scroll-driven animation
+    text:html-processing-model-event-loop
```
fantasai, line 149: What are you trying to do here? “scroll-driven animation” should already be exported, and “html-processing-model-event-loop” is not a term, it's an anchor that you can reference via `[[scroll-animations-1#html-processing-model-event-loop|linking text here, or drop for heading substitution]]`

fantasai (commented):
Still have a few stray markup changes, look for "switch" (x2) and "constructors".

web-animations-2/Overview.bs
kevers-google: Removed.

kevers-google:
Believe I've reverted the remaining formatting changes (constructor typo and 2 div / dl format). Moved to separate formatting PR.

fantasai (approved):
I defer to @flackr on the content of this PR, but thanks for cleaning up the commits. :)

At some point we'll want to update the WD also, so I suggest thinking about what remains before you're ready to publish.

web-animations-2/Overview.bs
```diff
@@ -805,6 +849,62 @@ The procedure to [=pause an animation=] needs to refer not only to the
...
+
+With:
+
+> If |seek time| is [=unresolved|resolved=],
+>
+>   <dl class=switch>
+>
+>       : If |has finite timeline| is true,
+>       ::
+>           Set the |auto align start time| flag to true.
```
flackr, line 880: This seems odd. If auto align start time was false, wouldn't we want to continue using the modified start time? Doing this for non-aligned animations could result in a big jump to the time it pauses at even if there was no layout change.

I'm not sure I fully understand how non aligned start times work in this ecosystem - it seems like we should have a well defined way to preserve a non-aligned start time but also have simple ways to re-align it.

kevers-google: Good catch, the condition for setting the flag should have required that start and hold time were both unresolved.  Refactored the section a bit to avoid nesting conditionals.  Basically if seek is set and the timeline is monotonic, we set the hold time to the seek time. If both hold and start time are unresolved, we must have a non-monotonic timeline and should set the auto align start time flag. 

Non-aligned start times are used after explicitly setting the start or current time. So if we set the current time on a play or pause pending animation, the current time will be preserved when the animation resolves its ready promise.
Playing an animation explicitly resets the "stickiness" as otherwise we have no way of doing so short of cancelling the animation. Pause pending animations only require an auto-aligned start time to compute the correct hold time. If we already have a time when pause-pending, we simply hold that time.

web-animations-2/Overview.bs
```diff
@@ -871,19 +871,12 @@ Replace:
...
+> |has finite timeline| is false,
+>    *  Set |animation|'s [=hold time=] to |seek time|.
 >
->       : Otherwise,
->       ::
->           Set |animation|'s [=hold time=] to |seek time|.
->
->   </dl>
+> If the animation's [=hold time=] and [=start time=] are unresolved,
+>    * Set the |auto align start time| flag to true.
```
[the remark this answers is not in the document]
kevers-google: You might have set the animation to the paused state via the animation-play-state property or could have explicitly called pause() after creating the animation.  If not a finite timeline, the hold time will be set to the seek time, and thus this condition will not apply.  It is not possible to have all three of the seek, start, and hold times being unresolved.  This is why the handling of monotonic timelines was placed upstream of this check.  If reversed, we would need to add the additional check for a non-monotonic timeline.

web-animations-2/Overview.bs
```diff
@@ -871,19 +871,12 @@ Replace:
...
 
-> If |seek time| is [=unresolved|resolved=],
->
->   <dl class=switch>
->
->       : If |has finite timeline| is true,
->       ::
->           Set the |auto align start time| flag to true.
+> If |seek time| is [=unresolved|resolved=], and
+> |has finite timeline| is false,
```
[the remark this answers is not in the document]
kevers-google: No we shouldn't set the hold time. Seek time is only set if current time is unresolved (https://www.w3.org/TR/web-animations-1/#pausing-an-animation-section), which means that hold time and start time are both unresolved.  If there is a start time, then we'll set the hold time when the ready promise is resolved.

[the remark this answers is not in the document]
kevers-google: When starting a scroll-driven animation in the paused state, we wait for a deferred start time during timeline-validation and then compute a hold time from the start time when running the pending-pause task.

web-animations-2/Overview.bs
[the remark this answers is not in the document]
flackr: Filed https://github.com/w3c/csswg-drafts/issues/9256

web-animations-2/Overview.bs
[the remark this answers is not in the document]
flackr: I there may have been a misunderstanding, my comment is about the `|has finite timeline|` condition (the line on which this comment is attached to). This current spec update says the following:

> If seek time is [resolved](https://drafts.csswg.org/web-animations-1/#unresolved), and has finite timeline is false,
>
> * Set animation’s [hold time](https://drafts.csswg.org/web-animations-1/#animation-hold-time) to seek time.
>
> If the animation’s [hold time](https://drafts.csswg.org/web-animations-1/#animation-hold-time) and [start time](https://drafts.csswg.org/css-transitions-1/#transition-start-time) are unresolved,
>
> * Set the auto align start time flag to true.

If we have a resolved seek time, shouldn't we set the hold time regardless?

web-animations-2/Overview.bs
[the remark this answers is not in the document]
flackr: I understand now - this ensures that monotonic timelines will have one of hold time or start time resolved by this step.

web-animations-2/Overview.bs
flackr: I think it'd be reasonable to require canceling the animation to reset the start time stickiness. It seems odd that calling pause, then play could change the start time.

kevers-google: Updated playing-an-animation to only set the |auto align start time| flag to true if |previous current time| is unresolved.
Updated setting-the-timeline to set |auto align start time| to true if |to finite timeline| is true and |from finite timeline| is false.  This also handles transitioning from a  null to a scroll timeline.

web-animations-2/Overview.bs
```diff
@@ -805,6 +850,55 @@ The procedure to [=pause an animation=] needs to refer not only to the
...
+>       ::
+>           Set |animation|'s [=hold time=] to |seek time|.
+>
+>   </dl>
+
+With:
+
+> If |seek time| is [=unresolved|resolved=], and
+> |has finite timeline| is false,
+>    *  Set |animation|'s [=hold time=] to |seek time|.
```
flackr, line 877: Actually, I'm still confused about this. This will set the hold time to 0 or 100% if we pause with an unresolved current time. When would we want this behavior?

kevers-google: A time based animation starting from an idle state. The [==unresolved\resolved=] label is likely the source of confusion.  The state is actually resolved, but links to text describing when the state would be unresolved.

web-animations-2/Overview.bs
```diff
@@ -318,7 +322,6 @@ follows:
```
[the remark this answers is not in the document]
kevers-google: I cannot see where your links are pointing to.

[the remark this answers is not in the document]
kevers-google: Only one of the two links is working for me.  

Referring to the handling of start time in the set timeline procedure, we have a few options:
* When switching timelines, we always set auto align start time.  This means that setting current time than then switching timelines won't preserve current time.  Though as the units of time are changing, all we would be able to preserve is progress.
* When switching timelines, we preserve progress if auto-align is false, or the new timeline is time-based, otherwise hands off.  This means that we should either make auto-alignment 3 state (true|false|unset) and only set to true if in the unset state, or always set auto-align to true when calling play regardless of the timeline type (it simply won't get used while time-based). Then when we switch to a scroll-based animation, the flag will already be set if current time or start time have not been called.

web-animations-2/Overview.bs
kevers-google: Removed seek time from pausing-an-animation for better alignment with playing-an-animation.

web-animations-2/Overview.bs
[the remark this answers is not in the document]
kevers-google: Cleaned up the procedure for setting the timeline based on offline discussion.

web-animations-2/Overview.bs
```diff
@@ -856,27 +835,87 @@ hold time until the animation range has been computed.
...
+>            : Otherwise,
+>            ::
+>                Set [=hold time=] to |animation|'s [=associated effect end=].
+>
+>            </dl>
+>
+>        </dl>
 >
-> If the animation's [=hold time=] and [=start time=] are unresolved,
+>    1. If |has finite timeline| = true,
```
flackr, line 917: nit: s/=/is

web-animations-2/Overview.bs
```diff
@@ -304,39 +304,18 @@ follows:
...
-                set |auto align start time| to true.
-            1.  Let |seek time| be zero if [=playback rate=] &ge; 0, and
-                |animation|'s <a>associated effect end</a> otherwise.
-            1.  Update the animation based on the first matching condition if any:
+            1.  set |auto align start time| to true.
+            1.  If |previous play state| is "paused" and
+                |previous progress| is resolved:
 
-                <dl class="switch">
+                1.  Set [=animation/start time=] to unresolved.
```
flackr, line 311: Shouldn't we set the start time to unresolved when we set auto align start time to true regardless of the previous play state, as once we switch timelines, deciding to auto align the start time we don't know what the start time will be, right?

kevers-google: Doing so would affect the value of anim.currentTime.  Not that currentTime is that relevant if in the process of switching timelines and needing a new start time.  My initial gut feeling was that rather than having a currentTime that can potentially become null, we should hold the last known value of currentTime until we've established a new deferred start time.  Having said that a play-pending animation will have a null value of currentTime until start time is resolved rather than 0 or effect-end that it would have if time-based.  In short, I can be convinced either way.

kevers-google: A finished animation would also briefly appear to be paused if we unconditionally reset the start time if auto-aligning the start time unless we also clear the hold time.  Note that if paused we will set the hold time based on the previous progress so likely OK to clear both start and hold time.  Then we are in a consistent state with playing an animation.

web-animations-2/Overview.bs

web-animations-2/Overview.bs
kevers-google: Fixed.

Made resetting the start time unconditional after setting auto-aligned start time.
Also clear hold time to avoid changing the play state to paused and added scheduling a pending play task if "running" or "finished".

web-animations-2/Overview.bs
flackr: Thanks this reads much more clearly now.

flackr (approved):
Thanks for all the changes. I think this is looking good.

PR #9186, opened 2023-08-14
[css-borders-4] Defined `box-shadow-*` as coordinating list property group

### LINKED ISSUE #8592: [css-borders-4] `box-shadow` needs to define coordinating list base property

SebastianZ:
Copied over the section from Level 3 about "Layering, Layout, and Other Details" of box shadows and defined the longhands as coordinating list property group with `box-shadow-offset` being the coordinating list base property.

This is meant to fix #8592 and supersedes #8613.

Sebastian

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -773,6 +773,37 @@ Animation type: see individual properties
   <pre class=prod>
   <dfn><<spread-shadow>></dfn> = <<'box-shadow-color'>>? &amp;&amp; [ <<'box-shadow-offset'>> [ <<'box-shadow-blur'>> <<'box-shadow-spread'>>? ]? ] &amp;&amp; <<'box-shadow-position'>>?</pre>
 
+<h3 id="shadow-layers">
+Layering, Layout, and Other Details</h4>
+
+  Drop shadows are declared in the [=coordinated value list=]
+  constructed from the 'box-shadow-*' properties,
+  which form a [=coordinating list property group=]
+  with 'box-shadow-offset' as the [=coordinating list base property=].
+  See [[css-values-4#linked-properties]].
+
+  <p>The shadow effects are applied front-to-back:
+  the first shadow is on top and the others are layered behind.
+  Shadows do not influence layout and may overlap (or be overlapped by)
+  other boxes and text or their shadows.
+  In terms of stacking contexts and the painting order,
+  the <i>outer box-shadows</i> of an element are drawn immediately below the background of that element,
+  and the <i>inner shadows</i> of an element are drawn immediately above the background of that element
+  (below the borders and border image, if any).
+
+  <p>If an element has multiple boxes, all of them get drop shadows,
+  but shadows are only drawn where borders would also be drawn;
+  see 'box-decoration-break'.
+
+  <p>Shadows do not trigger scrolling or increase the size of the scrollable area.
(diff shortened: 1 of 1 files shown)


cdoublev (approved):

css-borders-4/Overview.bs
```diff
@@ -773,6 +773,37 @@ Animation type: see individual properties
...
+  <p>The shadow effects are applied front-to-back:
+  the first shadow is on top and the others are layered behind.
+  Shadows do not influence layout and may overlap (or be overlapped by)
+  other boxes and text or their shadows.
+  In terms of stacking contexts and the painting order,
+  the <i>outer box-shadows</i> of an element are drawn immediately below the background of that element,
+  and the <i>inner shadows</i> of an element are drawn immediately above the background of that element
+  (below the borders and border image, if any).
+
+  <p>If an element has multiple boxes, all of them get drop shadows,
```
Loirooriol, line 794: I think this is wrong: only the principal box should get the shadow. But if the principal box gets broken into multiple fragments, then all of them get drop shadows (according to `box-decoration-break`).

Loirooriol (commented):
This is just importing from L3, so feel free to land as-is and fix problems later. But I think this should refer to fragments instead of boxes.

Also, reminder to fix the animation types to handle lists (https://github.com/w3c/csswg-drafts/pull/9096#discussion_r1278062829)

css-borders-4/Overview.bs
SebastianZ: Note that I just copied that from [CSS Backgrounds 3](https://drafts.csswg.org/css-backgrounds-3/#shadow-layers).

And isn't that, what the sentence is saying? Or do you critize that it's not referring to the definitions of "principle box" and "fragments"?

Loirooriol: The text probably uses the old CSS2 terminology. Now boxes and fragments are different things.

For example, `display: block ruby` generates a principal block container box, and an inline-level ruby container inside it. Only the block container should get a shadow, as you can test on Firefox:

```html
<ruby style="display: block ruby; box-shadow: 0 0 5px">
  <rb>振</rb><rb>り</rb><rb>仮</rb><rb>名</rb>
  <rp>(</rp><rt>ふ</rt><rt>り</rt><rt>が</rt><rt>な</rt><rp>)</rp>
</ruby>
```

Anyways, as I said, I think it's fine to land as-is since the problem already exists in L3.

SebastianZ: I see. So I think you should open a separate issue (or directly create a PR) to discuss and fix this.

Loirooriol: https://github.com/w3c/csswg-drafts/issues/9286

PR #9189, opened 2023-08-15
[scroll-animations-1] Add view-timeline-inset to shorthand

### LINKED ISSUE #8926: [scroll-animations-1] view-timeline-inset as part of view-timeline shorthand?
There's been an issue marked in the draft for a long time, which is whether `view-timeline-inset` should be a part of the `view-timeline` shorthand or not.

The key question here should be, do we want casual uses of `view-timeline` to reset the inset if it has been set already, or do we want it to cascade independently?

_Note: For `text-decoration`, for example, we allow it to reset most of the line properties (line type, style, color, thickness) but intentionally don't reset the position or offset so that flipping decorations on doesn't change the positioning._

andruud:
Resolves #8926.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -736,16 +736,13 @@ spec: cssom-view-1; type: dfn;
 
 	<pre class='propdef shorthand'>
 	Name: view-timeline
-	Value: [ <<'view-timeline-name'>> <<'view-timeline-axis'>>? ]#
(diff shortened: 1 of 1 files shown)


andruud:
@fantasai It would be nice to spec this ASAP, since `view-timeline` recently shipped in Chrome, and it's just going to get harder to change the longer we wait.

Not sure if this is the grammar we want. :-)

flackr (approved):
Looks good to me.

PR #9215, opened 2023-08-22
[css-conditional-5] Define the block contents of `@container` with `<rule-list>`

cdoublev:
`<block-contents>` seems confusing to me because declarations are allowed only when `@container` is nested (at any level) in a style rule.

It is also inconsistent with other conditional rules (`@media` and `@supports`), whose block contents is defined with `<rule-list>`.

 > All conditional group rules are defined to take a `<rule-list>` in their block

https://drafts.csswg.org/css-conditional-3/#contents-of


tabatkins:
Agreed, it should be consistent with how we define `@media`/etc - `<rule-list>` in the official grammar, and let Nesting override that when nested (it specifies that the `<rule-list>` turns into a `<block-contents>` in this case).

PR #9219, opened 2023-08-22
[web-animations-1] Define 'start time' for animation effects

fantasai:
See https://github.com/w3c/csswg-drafts/pull/8985#discussion_r1300440338

In Level 2, we can change just the definition of "start time" to handle sequencing, and those adjustments will flow through the rest of the definitions appropriately.


birtles:
I was pretty opposed to introducing two kinds of start times but I notice we already have a concept of animation effect _end_ time so I guess this makes sense.

Issue #8292, opened 2023-01-09
[css-text-4] Glyphs may collide by "Fullwidth Punctuation Collapsing"


kojiishi:
I'd like to bring up a discussion by sharing our experience on implementing [a tool](https://github.com/googlefonts/chws_tool) that adds the functionality of "[Fullwidth Punctuation Collapsing]" in fonts as the OpenType `chws` feature.

This worked quite well for Noto CJK, and the fonts are shipping as part of Android and ChromeOS for months.

When we tried to support some other fonts, there were some cases where glyphs collide. We started adding an "exception list" for some fonts. When we tried to support more fonts, a good number of collisions were found. As we tested, the exception list grew longer and longer. Some of them are from font creators not knowing which glyphs must have enough spacing. Some of them are intentional, such as curly braces in extra bold fonts not being able to fit in half-width.

Given the experience, we implemented the automatic glyph bounding box detection in the tool. The currnet implementation is rather simple, just exclude the code point if its glyph does not have expected spacing.

Using `halt` is another option, but there are many fonts that don't have `halt`. I guess there are multiple ways to resolve this, including not to do anything but let authors turn the feature off for such fonts or for specific pairs, but I'd like to hear what people thinks.

/cc @fantasai @litherum @frivoal @MurakamiShinyu 

[Fullwidth Punctuation Collapsing]: https://drafts.csswg.org/css-text-4/#fullwidth-collapsing

PR #9225, opened 2023-08-23
[css-text-4] Clarify fonts without enough blank may not be trimmed

kojiishi:
To address the risk mentioned at https://github.com/w3c/csswg-drafts/issues/8292, this change allows UA to not to trim when there's not enough blank.


kojiishi:
Hey @fantasai, I heard from @chrishtr that you agree UA may not trim when doing so can cause excessive kerning. Can you review this change and see if this describes it?

kojiishi:
Reworded a bit, hope this is better English ;-)

css-text-4/Overview.bs
```diff
@@ -8337,6 +8337,13 @@ Fullwidth Punctuation Collapsing</h4>
 	the font has proportional ideographs
 	and the fullwidth advance width cannot be reliably determined by measuring glyphs.
 
+	Some fonts have fullwidth punctuation characters
+	whose blank are too small to trim (kern.)
+	UA may not trim (kern)
```
frivoal, line 8342:
```suggestion
	UA may choose not to trim (kern)
```

Otherwise "may not" is ambiguous, and could be seen as meaning the same as "must not"

kojiishi: Done, thank you for the review.

Issue #8564, opened 2023-03-12
[color-5] Serialization of percentages in color-mix()


tiaanl:
In reference to this section:
https://github.com/w3c/csswg-drafts/blob/43b12e47663cff4cead0d187086dcf9ea8f33626/css-color-5/Overview.bs#L2585

The spec doesn't mention anything about serializing the incoming color-mix(), only the result.  For color-mix() itself the tests suggest that the percentages on both sides should only be omitted if the author omitted them, otherwise they should be present.

@emilio mentioned that there is also the [serializing-css-values](https://drafts.csswg.org/cssom/#serializing-css-values) section that suggests the percentages could be omitted if they are obvious.

So for example `color-mix(in hsl, red 60%, blue 40%)` should serialize to `color-mix(in hsl, red 60%, blue)`.  This contradicts what is currently in the tests.

As all other browsers (except firefox) is following the current wpt tests, I think we should maybe explicitly mention this in the spec?

emilio:
I think we should fix the test instead FWIW. cc @svgeesus

svgeesus:
The spec should clearly state what happens, even if it is just the logical application of a principle, and should have an example. I'm sure that lack of this clarity explains the divergent implementations on that point.

I agree that the tests do not follow that, and should be corrected.

svgeesus:
So I just added this

```html
The serialization of the specified value of a ''color-mix()'' function
is the string "color-mix(in ",
followed by the specified <<color-space>> in all-lowercase,
followed by ", ",
followed by the first specified color,
followed by a space,
followed by the normalized first percentage,
followed by ", ",
followed by the second specified color,
followed by ")".

Following the principle of shortest serialization,
and given that the normalized percentages are serialized,
the normalized second percentage is omitted.
```

but see a problem: if the normalized percentages add to less than 100%, how do we serialize the alpha multiplier (which cannot be expressed directly in the syntax, only indirectly.

Si I am thinking we need to retain the as-specified (un-normalized) percentages, if they are both given and if they add to less than 100%.

svgeesus:
@tiaanl @emilio any thoughts? The only way I see to retain the alpha modifier is to retain the un-normalized percentages; we can do that just where they add to less than 100%, or do it always.

emilio:
[...]

PR #9231, opened 2023-08-23
[css-color-5] Serialization of color-mix edge case with not normalized values including 50% #8564

AndresRPerez12:
With the current instructions in the spec for serializing `color-mix`, values of `50%` are incorrectly excluded from the serialization when the specified percentages don't add up to 100% and need normalization.

The proposed wording makes it more clear when to exclude these percentage values for both the first and the second color. The first percentage is excluded only if both are `50%`, and the second one only when normalization is not required and therefore the information from the first color is enough.


svgeesus (approved):
Thanks!

PR #9216, opened 2023-08-22
[css-text-4] Replace `word-boundary-detection: auto()` with `word-break: auto-phrase`

kojiishi:
This PR takes the https://github.com/w3c/csswg-drafts/pull/8974 by @litherum, addresses the @fantasai's feedback, and the WG resolutions at https://github.com/w3c/csswg-drafts/issues/7193#issuecomment-1663101848.

PR #9242, opened 2023-08-25
[css-text-4] auto-phrase

frivoal:
Recast `word-boundary-detection: auto()` as `word-break: auto-phrase`, as well as `word-boundary-detection: manual` as `word-break: manual`.

This allows retiring the `word-boundary-detection` property.

Also, the `word-boundary-expansion` property needed to gain its own `auto-phrase` value, since it could no longer depend on `word-boundary-detection: auto()`.

Since there's no longer a pair of `word-boundary-*` properties, there's no need to keep the two names related, and the original name was quite a mouthful. `word-space-transform` is simpler, and better show-cases the similarities with `text-transform`.

Supporting sections had to be reorganized as well to keep the whole thing coherent.

This addresses https://github.com/w3c/csswg-drafts/issues/7193 (and https://github.com/w3c/i18n-activity/issues/1454). It builds upon and is meant to replace https://github.com/w3c/csswg-drafts/pull/9216 (which itself builds upon and replaces https://github.com/w3c/csswg-drafts/pull/8974).

(co-authored with @fantasai)

----

Note: the `<wpt>` annotations have not been fully updated yet. Ignore them for now: I'll update the tests and the corresponding `<wpt>` annotations in the spec once we agree on the behavior described in this PR.

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -946,7 +946,7 @@ Mapping Rules</h4>
 	The mappings for [=small Kana=] to [=full-size Kana=] are defined in [[#small-kana]].
 
 
-<h3 id=word-space-transform>
+<h3 id=word-space-transform oldids="word-boundary-expansion">
 Expanding Between Words: the 'word-space-transform' property</h3>
 
 	<pre class="propdef">
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -946,11 +946,11 @@ Mapping Rules</h4>
 	The mappings for [=small Kana=] to [=full-size Kana=] are defined in [[#small-kana]].
 
 
-<h3 id=word-boundary-expansion>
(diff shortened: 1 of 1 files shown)


frivoal:
This PR moves quite a bit of text around, so it's not too easy too review as a source diff. Here are:
- [a built version](https://lists.w3.org/Archives/Public/www-archive/2023Aug/att-0002/Overview.html)
- [a diff against the main branch](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flists.w3.org%2FArchives%2FPublic%2Fwww-archive%2F2023Aug%2Fatt-0003%2FOverview.html&doc2=https%3A%2F%2Flists.w3.org%2FArchives%2FPublic%2Fwww-archive%2F2023Aug%2Fatt-0002%2FOverview.html)

kojiishi (approved):
LGTM, thanks!

PR #9250, opened 2023-08-28
[css-contain-2] Removed "Twitter" from CSS Contain 2

OddDev:
Twitter was rebranded to "X". With the latest development I think that a more persistent platform like Mastodon is a better example for this document.  Thus changed all occurrences of Twitter to Mastodon.

### DIFF
--- a/css-contain-2/Overview.bs
+++ b/css-contain-2/Overview.bs
@@ -1830,7 +1830,7 @@ Using ''content-visibility: auto''</h3>
 	it's best to use at a reasonably fine granularity.
 
 	<div class=example>
-		For example, on Twitter,
+		For example, on Mastodon,
 		applying ''content-visibility: auto'' to the entire timeline
 		wouldn't accomplish much--
 		it's always on-screen,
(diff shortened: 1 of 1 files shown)


OddDev:
User affiliation information is not needed, as this is editorial.

tabatkins:
I think it's fine as-is; Twitter is still correct historically (and will likely be in the future, imo). I don't think we need to follow the deranged descent-into-madness of the company's current owner.

OddDev:
Thanks for taking the time to review this! 😊 Very honored to interact with you. 

Can you elaborate on this a little? I might miss some insights in guidelines etc. The work is already done and appears - in my eyes - to be an improvement regardless, no? 😬

>  I don't think we need to follow the deranged descent-into-madness of the company's current owner.

That's what I was trying to do with this PR actually. I didn't want an "implicit ad" for a platform that is currently in an "interesting" state. 😅 
From my naive standpoint it appears to be the other way around. 

Thanks again for taking the time! You rock. 🚀🙌

Issue #7542, opened 2022-07-28
[css-ui] ? Allow <textarea> to be sized by contents.


bfgeek:
See: https://github.com/whatwg/html/issues/6807 for more context about the problem we'd like to solve. The TL;DR is that we'd like to allow the `<textarea>` element to be sized to its contents a little easier. I.e.

![textarea-1](https://user-images.githubusercontent.com/992712/166292770-8659fed3-4717-4d65-a0ad-80ce4c9b7e25.gif)

We explored different solutions in the whatwg thread. Folks seem to prefer a CSS solution to this problem - hence this issue!

One way to achieve the "minrows"/"maxrows" in that issue is to use the "lh" unit. (Fortunately the "lh" unit works exactly the same way textareas are implemented today - in that they use the first available font to calculate the line-height).

E.g. 
```
textarea {
  min-height: 2lh;
  max-height: 5lh;
}
```

The only thing we are missing is the ability to use the "actual"[1] intrinsic size of the textarea. There are a few different potential solutions here.

Today the intrinsic-size of a textarea will (roughly) read the "rows" attribute, and multiply this by the "lh" unit. We need a switch to instead read the "rows" attribute to be based on the actual number of lines. One potential solution is:

`textarea-rows: auto | <number>`

If "auto" was set the user-agent would read the number of lines, then multiply this by "lh" to get the intrinsic block-size for the text-area.

We could then map the rows attribute to a presentation style hint to this property.

cc/ @tabatkins @lilles

Loirooriol:
Why not `height: max-content`? See https://drafts.csswg.org/css-sizing-3/#intrinsic-sizes

> Although the [`auto`](https://drafts.csswg.org/css-sizing-3/#valdef-width-auto) size of text input controls such as HTML’s `<input type=text>` and `<textarea>` elements is typically a fixed size, the contents of such elements can be used to determine a content-based [intrinsic size](https://drafts.csswg.org/css-sizing-3/#intrinsic-size), as for non-replaced block containers. The [`min-content`](https://drafts.csswg.org/css-sizing-3/#valdef-width-min-content) and [`max-content`](https://drafts.csswg.org/css-sizing-3/#valdef-width-max-content) keywords of the [sizing properties](https://drafts.csswg.org/css-sizing-3/#sizing-property) thus represent content-based sizes for form controls which render their value as text contained within their box, allowing such controls to size to fit their visible contents similarly to regular non-replaced elements.

[...]

PR #9251, opened 2023-08-28
[css-ui-4] Add 'field-sizing' property

tkent-google:
See https://github.com/w3c/csswg-drafts/issues/7542

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -102,6 +102,24 @@ urlPrefix: https://html.spec.whatwg.org/multipage/; spec:HTML
 .awesome-table td {padding:5px}
 .awesome-table {color:#000;background:#fff;margin: auto;}
 
+.fake-textarea {
+	background: white;
+	border: inset;
+	color: black;
+	display: inline-block;
+	font-family: monospace;
+	padding: 0 2px;
+}
+.fake-textarea.auto {
+	height: 2lh;
+	overflow: auto;
+	resize: both;
+	width: 10em;
+}
+.fake-textarea.normal {
+	width: 10em;
+}
+
 .fake-input-type-password {
 	display: inline-block;
 	border: 1px solid black;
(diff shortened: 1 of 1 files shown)


tkent-google:
My plan is that more detailed behavior is described in the HTML standard.

tabatkins:
Per <https://github.com/w3c/csswg-drafts/issues/7542#issuecomment-1685352461>, the names "normal" and "auto" aren't very illuminating here. (And, notably, as defined here they appear to be the exact opposite of what I expected them to be - I expected "normal" to be the legacy half-replaced behavior, while "auto" was the new 'act like other elements' behavior. That's a good argument for the names being bad!)

Maybe we should go with `legacy | auto` then? `auto` still isn't any clearer, but `legacy` clearly points to "how form elements *used to* size", so `auto` must be the new behavior. `legacy | normal` would also be fine.

tkent-google:
@tabatkins Let's discuss keywords in https://github.com/w3c/csswg-drafts/issues/7542

tkent-google:
Updated this PR for `fixed | content`.
https://github.com/w3c/csswg-drafts/issues/7542#issuecomment-1747803316

@frivoal would you take a loot at this please?

frivoal (approved):

css-ui-4/Overview.bs
```diff
@@ -2452,6 +2470,60 @@ Effects of 'appearance' on Semantic Aspects of Elements</h4>
...
+	Percentages: N/A
+	Computed Value: as specified
+	Canonical order: per grammar
+	Animation type: discrete
+	</pre>
+
+	For the purpose of this specification,
+	an <dfn export>element with default preferred size</dfn> is an element
+	of which size is fixed regardless of its content size though its
+	[=preferred size property=] is 'auto'.
```
frivoal, line 2491:
```suggestion
	whose [=intrinsic size=] is fixed regardless of the size of its content.
```

css-ui-4/Overview.bs
```diff
@@ -2452,6 +2470,60 @@ Effects of 'appearance' on Semantic Aspects of Elements</h4>
...
+	The host language defines which elements are applicable to it.
+	For example, in HTML <{textarea}> is an [=element with default preferred size=].
+
+	<dl dfn-type=value dfn-for=field-sizing>
+		<dt><dfn>fixed</dfn>
+		<dd>
+			The UA must apply the default preferred size to
+			an element if it is an
+			[=element with default preferred size=] and its
+			[=preferred size property=] value is 'auto'.
```
frivoal, line 2501:
```suggestion
			For [=element with default preferred size=],
			the UA must apply set the [=intrinsic size=]
			to the default preferred size defined by the host language for that element.
```

css-ui-4/Overview.bs
```diff
@@ -2452,6 +2470,60 @@ Effects of 'appearance' on Semantic Aspects of Elements</h4>
...
+			Otherwise UA must behave the same as ''field-sizing/content''.
+		<dt><dfn>content</dfn>
+		<dd>
+			The UA must not apply the default preferred size to an element.
```
frivoal, line 2505:
```suggestion
			The UA must determine the element's [=intrinsic size=] based on its content,
			and must ignore any default preferred size defined by the host language for that element.
```

frivoal (changes requested):
Unless I am missing something, it seems to me that we'd be better off expressing the difference between the two modes in terms of intrinsic size rather than talking directly about the preferred size properties. That gets the desired effect into all aspects of sizing, which is my understanding of the groups resolution.

I've made suggested edits implementing that change. Let me know if you think these work, or if I missed subtlety you were deliberately introducing by using the proposed wording.

tkent-google:
> I've made suggested edits implementing that change. 

Thanks! I accepted all your suggestions.

tkent-google:
@frivoal  would you take a loot at this again please?

frivoal (approved):

Issue #6641, opened 2021-09-20
[css-pseudo] Custom properties on :root


andruud:
It's a common pattern to define all the primitives in your style guide as "global" (/ attempted global) custom properties on :root, e.g.

```
:root {
  --pale-blue: <etc>;
  --light-blue: <etc>;
  /* ... hundreds ... */
}
```

If I understand the highlight inheritance model correctly, such global custom properties would not be available on `::selection` (etc), unless you modify the custom-property-defining rule to something like `:root, :root::selection {}`. In principle what the author would be doing is creating two separate mega-blobs of custom properties. This may or may not be easy to optimize depending on how the usage patterns will actually end up.

Authors can use `@property`  w/ initial value to address it. Is that enough, or we do something else about this?

@delan @tabatkins

delan:
(summary for meeting)

> Same context as the previous issue (#6774). Frontend frameworks often define a whole bunch of custom properties for things like colour palettes, on the root originating element, which used to be available to highlights “for free”. But now they won’t be, and authors will need to define those properties on both the originating element and each highlight pseudo.

andruud:
Author-defined `env()` could be something to consider for this.

tabatkins:
We've discussed using variable values in other contexts that don't inherit from root as well (like in @font-palette, for example).

Having an author-defined env() might work, but would require an author to duplicate values across properties and env(), or else decide for each whether they want to express it as a var or an env().

The idea that's been presented a few times is to special-case properties defined on the root, and make them available more globally, so any context that's not an element with :root as an ancestor can use var() to obtain the root values of the custom property. Would this be reasonable impl-wise?

andruud:
> Having an author-defined env() might work, but would require an author to duplicate values across properties and env(), or else decide for each whether they want to express it as a var or an env().

Yeah, I see how that would be annoying.

OK, then maybe we can specify that `var()` in non-element contexts use the initial value of custom properties, and then provide a new way of specifying those initial values (if `@property` is too verbose).

> Would [root special case] be reasonable impl-wise?

[...]

PR #9279, opened 2023-08-31
[css-pseudo-4] Define the behavior when custom properties must be taken from the root

schenney-chromium:
[css-pseudo-4] Define the behavior when custom properties must be taken from the root

In https://github.com/w3c/csswg-drafts/issues/6641 it was resolved that "highlight pseudos inherit across the highlight tree, and the root highlight inherits custom props from the root element". Here we add spec language to explain the behavior, with an example.

A WPT will be added when the chromium change that implements the behavior is landed.


schenney-chromium:
This is my first ever spec edit. And feedback is very welcome.

schenney-chromium:
Ping!

astearns (approved):
LGTM. Thanks!

css-pseudo-4/Overview.bs
```diff
@@ -936,6 +936,49 @@ Cascading and Per-Element Highlight Styles</h3>
   css/css-pseudo/cascade-highlight-002.html
   </wpt>
 
+<h4 id="highlight-custom-properties">
+Custom Properties in Highlight Styles</h3>
```
fantasai, line 940: I would merge this section with 3.5. Cascading and Per-Element Highlight Styles

css-pseudo-4/Overview.bs
```diff
@@ -936,6 +936,49 @@ Cascading and Per-Element Highlight Styles</h3>
...
+
+  A [=custom property=] used in highlight styles uses the
+  first custom property value found in the highlight cascade.
```
fantasai, line 943: This is confusing. Just add custom properties to the list in 3.2. Styling Highlights.

css-pseudo-4/Overview.bs
```diff
@@ -936,6 +936,49 @@ Cascading and Per-Element Highlight Styles</h3>
...
+  However, if a custom property value cannot be resolved
+  through the highlight cascade, the property value is taken from
+  the root element.
```
fantasai, line 946: Instead of this, I would take this sentence:

>  Additionally, for highlight pseudo-elements originating from the root element, the inherited value of color is currentColor, not the initial value.

Split it out into a list item like this:

> Additionally, for [=highlight pseudo-elements=] originating from the [=root element=]:
> * the inherited value of 'color' is ''currentColor'', not the [=initial value=].
> * the value of [=custom properties=] inherit from the [=root element=].

fantasai (changes requested):
Thanks for the PR. Here's comments. :)

css-pseudo-4/Overview.bs
```diff
@@ -696,6 +696,8 @@ Styling Highlights</h3>
     <wpt>
     css/css-pseudo/textpath-selection-011.html
     </wpt>
+
+    <li>A [=custom property=] may be defined and/or used inside a highlight pseudo-element.
```
fantasai, line 700:
```suggestion
    <li>[=custom properties=] [[!CSS-VARIABLES-1]]
```

A property applying means you can declare or use it, custom properties aren't different from others in this respect.

schenney-chromium:
OK. I fixed a couple of typos and made the latest requested change. Thanks again.

Next up, in another PR, will be the changes about font properties from the originating element.

schenney-chromium:
@fantasai I know you must be super busy, but I think this is ready to go.

schenney-chromium:
Ping!

fantasai (approved):
Looks great! Thanks a lot!

p.s. I am nearly impossible to reach through GH notifications (someday, it's my dream, I will figure this out, but it hasn't happened yet), but don't be shy about emailing me.

Issue #8922, opened 2023-06-06
[css-fonts-5] Proposal: Animating font-palette


tursunova:
### Overview

We propose the animation of font-palette property. Currently [CSS Fonts 4 spec](https://drafts.csswg.org/css-fonts/#font-palette-prop) defines the animation behaviour of this property as discrete, which is insufficient to achieve a smooth transition between two selected palettes. 

For motivation, examples and more information, see [Explainer: Animating font-palette](https://docs.google.com/document/d/1XMTrKH003KBOes6hxzI-3E7LTwp5YwFC-rnzoFpFrfw/edit?usp=sharing).

### Proposal

The [font-palette property accepts](https://drafts.csswg.org/css-fonts/#font-palette-prop) either `@font-palette-values` identifier or keyword. Currently there are no [existing animation types](https://www.w3.org/TR/web-animations/#animation-type) to represent animation of font-palette property. For this we propose the following modifications to the font-palette property specification:

```
Value:		        normal | light | dark | <palette-identifier> | <palette-mix()>
Animation type: 	by mix value
```

### Animation type:   by mix value

Since there are no [existing animation types](https://www.w3.org/TR/web-animations/#animation-type) to represent the interpolated value of font-palette property at the specific time, we propose the animation type “by mix value”. For that, we introduce a new font-palette function `palette-mix()` to be accepted by the font-palette property. Function `palette-mix()` is used to represent the interpolated value of font-palette property during transition / animation progress. 

### Mixing Palettes: the ‘palette-mix()’ Function

This function, modeled after the [color-mix() function](https://drafts.csswg.org/css-color-5/#color-mix), takes two `<font-palette>` specifications and returns the result of mixing them, performing the interpolation in a given [\<color-interpolation-method\>](https://drafts.csswg.org/css-color-5/#color-interpolation-method), by a specified amount:

```
palette-mix() = palette-mix(<color-interpolation-method> , [ [normal | light | dark | <palette-identifier> | <palette-mix()>] && <percentage [0,100]>? ]#{2})
```

Percentages not adding up to 100% follow the  rules for [percentage normalization of the color-mix function](https://drafts.csswg.org/css-color-5/#color-mix-percent-norm).

### Palette interpolation
A font-palette property interpolates by interpolating each color from the palette the same way it is done for other color properties, [see section 12 Color Interpolation of ](https://drafts.csswg.org/css-color-4/#interpolation)CSS Color Level 4:

[...]

PR #9282, opened 2023-08-31
[css-fonts-4] Animating palettes

drott:
Draft wording to address #8922, created as a PR against CSS Fonts 4 to 
make it easier to see differences. I would expect this to be an addition to CSS Fonts 5.

Define `palette-mix()` function, interpolation and percentage normalization behavior
by help of referencing CSS Color Module 5.


drott:
@tabatkins @svgeesus Could I ask you to review this draft? Our hope is that this is something we can present as a as basis for discussion and proposal for resolution at TPAC. Thanks in advance.

svgeesus:
On it.

css-fonts-4/Overview.bs
```diff
@@ -7058,6 +7064,25 @@ Animation type: discrete
...
+        identified by <<palette-identifier>>, by the palette keywords
+        ''font-palette/light'' or ''font-palette/dark'',
+        or by another <<palette-mix()>> value.
+
+        The glyphs of the font must be rendered with a palette that is created by
+        interpolating for each palette color index between the matching color indices
+        of the first and second palette specified as arguments,
+        recursively resolving <<palette-mix()>> functions if needed.
+
+        For each palette color, the interpolation follows
```
svgeesus, line 7079: Does this just mean that it uses the [&lt;color-interpolation-method>](https://drafts.csswg.org/css-color-4/#color-interpolation-method) token and in general follows CSS Color 4 [Color Interpolation](https://drafts.csswg.org/css-color-4/#interpolation) or is there a tighter dependency on `color-mix()`, for example normalization of percentage values?

css-fonts-4/Overview.bs
```diff
@@ -7058,6 +7064,25 @@ Animation type: discrete
...
+        the rules of the <<color-mix()>> function.
+
+        When not specified otherwise the <<color-interpolation-method>> is
+        <code>in ''oklab''</code> as recommended by [[css-color-5#interpolation-space]].
+
+        Percentages are normalized according to [[css-color-5#color-mix-percent-norm]].
```
svgeesus, line 7085: Ah okay.

css-fonts-4/Overview.bs
```diff
@@ -7068,7 +7093,39 @@ Note: The names 'light' and 'dark' describe the <i>background</i> the color pale
...
+    }
+}
+p {
+     font-family: Nabla;
+     animation: animate-palette 1.4s infinite alternate linear;
+}
+    </pre>
+
+	<figure>
+		<img src="images/nabla-animated.webp" alt="The word Animate animated between a pink and a yellow palette."/>
```
svgeesus, line 7126: Nice example

svgeesus (approved):
This all looks great, very clear, lets discuss and get a resolution at TPAC.

svgeesus:
I appreciate the suggestion to put this in Fonts 5, but wonder if it would be clearer in Fonts 4 to keep all the changes in one place and avoid duplication. I could go either way.

css-fonts-4/Overview.bs
```diff
@@ -7058,6 +7064,25 @@ Animation type: discrete
 	    be a modification of an existing palette already present in the font,
 	    or be an entirely new palette.
 
+	<dt><dfn type><<palette-mix()>></dfn>
+	<dd>
+        This value defines a position in the interpolation between two palette values,
+        identified by <<palette-identifier>>, by the palette keywords
+        ''font-palette/normal'', ''font-palette/light'', ''font-palette/dark'',
```
drott, line 7071: FYI @tursunova - `normal` now added, thanks for pointing that out.

css-fonts-4/Overview.bs
drott: Yes, as you noticed, percentage normalization specified  below.

css-fonts-4/Overview.bs
```diff
@@ -6955,13 +6955,14 @@ as well as the colored versions.
...
-Value: normal | light | dark | <<palette-identifier>>
+Value: normal | light | dark | <<palette-identifier>> | <<palette-mix()>>
 Initial: normal
 Applies to: all elements and text
 Inherited: yes
 Percentages: N/a
-Computed value: specified keyword or identifier
-Animation type: discrete
+Computed value: specified keyword, identifier or <<palette-mix()>> function.
+    <<palette-mix>> must be simplified to a single keyword or identifier if resulting palette is equivalent.
```
drott, line 6964: @lilles Thanks for the feedback, now updated.

drott:
> I appreciate the suggestion to put this in Fonts 5, but wonder if it would be clearer in Fonts 4 to keep all the changes in one place and avoid duplication. I could go either way.

Agree, I would also prefer an edit to Fonts 4 if possible.

Issue #9237, opened 2023-08-24
[css-viewport-1] Bring the segments property into viewport 


darktears:
The `segments` property allows developers to target dual screens and foldable devices by giving information about the segments (or zones) when the browser window is spanned across the screen(s). There is a good set of pictures to demonstrate what we are trying to achieve [here](https://github.com/foldable-devices/demos).

Work was started by Microsoft in WICG to bring a JavaScript API, link [here](https://wicg.github.io/visual-viewport/#dom-visualviewport-segments). The proposed specification adds a `segments` property to the `visualViewport` object. `segments` are only defined when the browser window is maximized. They are also defined in CSS pixels. When the browser window state changes (maximized/resized) one needs to listen for the `resize` events to query the updated `segments`.

Please note that the `segments` property is intended mostly to mirror in JavaScript what’s being available in CSS (see [MQs](https://www.w3.org/TR/mediaqueries-5/#mf-horizontal-viewport-segments)). 

Both APIs, CSS and JavaScript went through TAG reviews [here](https://github.com/w3ctag/design-reviews/issues/689) and [here](https://github.com/w3ctag/design-reviews/issues/690). 

In terms of implementation status, Android and Windows are supported in Chromium based browser (with some caveats). Microsoft Edge does ship the CSS and the JavaScript APIs enabled by default on all platforms. Samsung Internet also is shipping both APIs by default on Android (the only OS they target).

tabatkins:
This seems like a pretty reasonable addition to me. As far as I can tell it's just exposing directly in JS what is already exposed to CSS via the `env()` [viewport segment variables](https://drafts.csswg.org/css-env-1/#viewport-segments), so I can't see any possible concern with this.

tabatkins:
Is this issue asking for review of the `segments` property? Or of the `VisualViewport` class more generally (defined in <https://wicg.github.io/visual-viewport/#dom-visualviewport-segments>). Or is it asking for that spec to be folded into CSSOM? Just unclear what the request for this WG is here.

darktears:
> This seems like a pretty reasonable addition to me. As far as I can tell it's just exposing directly in JS what is already exposed to CSS via the `env()` [viewport segment variables](https://drafts.csswg.org/css-env-1/#viewport-segments), so I can't see any possible concern with this.

You're right, that's exactly this.

darktears:
[...]

PR #9285, opened 2023-08-31
[cssom-view] Add segments property to visual viewport (#9237)

darktears:
[cssom-view]  Add segments property to visual viewport (#9237)

This property exposes in JS what is already exposed to CSS via the env() [viewport segment variables](https://drafts.csswg.org/css-env-1/#viewport-segments). TAG review [here](https://github.com/w3ctag/design-reviews/issues/689).

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1806,6 +1806,8 @@ interface VisualViewport : EventTarget {
 
   readonly attribute double scale;
 
+  readonly attribute FrozenArray&lt;DOMRect> segments;
+
   attribute EventHandler onresize;
   attribute EventHandler onscroll;
   attribute EventHandler onscrollend;
@@ -1855,6 +1857,19 @@ The <dfn attribute for=VisualViewport>scale</dfn> attribute must run these steps
     and abort these steps.
 1. If there is no output device, return 1 and abort these steps.
 1. Otherwise, return the <a>visual viewport</a>'s <a>scale factor</a>.
+
+The segments is an array of DOMRects that represent the dimensions of each existing viewport segment.
+Additonal details about the defintion of a viewport segment can be found here : [[css-env-1#viewport-segments]].
+
+The segments attribute must run these steps:
+1. If the <a>visual viewport</a>'s <a for=visualviewport>associated document</a> is not <a>fully active</a>, return 0.
+2. Returns null if called from within an iframe context and abort these steps.
+3. Returns null if there is only a single viewport segment and abort these steps.
+4. Otherwise, return the <a>visual viewport</a>'s segments array.
+
+Note: Based on the data returned for each viewport segment, developers will be able to infer the number of hinges available as well as the hinge orientation.
+The browser window can be moved/resized such that the number of segments and/or their dimensions change.
+In these cases, resize events will fire at which point authors should re-query this property, as the returned FrozenArray is just a snapshot of the current state.
 
 <dfn attribute for=VisualViewport>onresize</dfn> is the <a>event handler IDL attribute</a> for the <a event>resize</a> event.
 
@@ -1980,6 +1995,9 @@ Changes {#changes}
 
 This section documents some of the changes between publications of this specification. This section is not exhaustive. Bug fixes and editorial changes are
 generally not listed.
+
+<h3 id='changes-from-2022-07-07' class=no-num>Changes From 14 September 2023</h3>
+* Introduced the {{VisualViewport/segments}} property in the {{VisualViewport}} API.
 
 <h3 id='changes-from-2022-07-07' class=no-num>Changes From 07 July 2022</h3>
 * Introduced the {{VisualViewport}} API and related concepts


bokand (approved):
Not an expert on segments but the text lgtm generally.

cssom-view-1/Overview.bs
```diff
@@ -1856,6 +1858,21 @@ The <dfn attribute for=VisualViewport>scale</dfn> attribute must run these steps
...
+The <dfn attribute for=VisualViewport>segments</dfn> is an array of DOMRects that represent the dimensions of each existing viewport segment.
+
+A viewport segment represents the region of the visual viewport that resides on a separate display or logical display region.
+When the visual viewport spans across some number of physical features of a device, these features in turn convey a logical separation of the space available for content.
+
+The segments attribute must run these steps:
+1. If the <a>visual viewport</a>'s <a for=visualviewport>associated document</a> is not <a>fully active</a>, return 0.
+2. Returns null if called from within an iframe context and abort these steps.
+3. Returns null if there is only a single viewport segment and abort these steps.
+4. Otherwise, return the <a>visual viewport</a>'s segments array.
```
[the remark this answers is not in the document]
darktears: I'm not sure what you mean but line 1861 defines what the array is. If you mean by "linking" like jumping into the spec I think I can't use <dfn attribute for=VisualViewport>segments</dfn> again without getting a warning.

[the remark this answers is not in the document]
bokand: Ah I see. You could link to it with <a> but that seems rather self-referential, I was more thinking about some spec text that explains how these DOMRects are defined but I'm not sure that exists today.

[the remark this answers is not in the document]
darktears: We could link [this](https://www.w3.org/TR/geometry-1/#DOMRect)?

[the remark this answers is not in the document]
bokand: That defines DOMRect interface; I mean: what values do we put to the `x`, `y`, `width`, `height` for each rect in `segments`? I know we say they "represent the dimensions of each existing viewport segment" but that's a bit hand wavy. i.e. what is a "viewport segment"? Is it affected by pinch-zoom? Overlay keyboards? etc.

Perhaps we could link "viewport segment" to https://drafts.csswg.org/css-env-1/#viewport-segments? I think it'd explain that this is the same thing as the env variables and if we want to make their definition more precise in the future (w.r.t my questions above) we could then do so in one place for both `visualViewport.segments` and the `env` variables.

darktears:
@bokand what do you think about the latest change?

bokand:
lgtm!

darktears:
Feel free to merge. I'm not sure if additional reviews are required.

bokand:
Hmm, I'm not an editor so not sure if that's ok - perhaps @smfr or @emilio could take a look?

darktears:
Yes, I think we should wait their feedback :)

darktears:
@smfr @emilio any comments?

smfr:
I have a few thoughts, mainly about potential for user fingerprinting based on how they generally use a foldable/multi-screen device.

First, on a device that's foldable but with no hinge gap (i.e. continuous foldable screen), does the number of segments change when the screen goes between some folded state and unfolded flat? At what fold angle does this occur? If so, this might be used to identify users based on their habits of using the device fully unfolded or partially folded. This may also be used to estimate whether the user is paying full attention to the screen (fully unfolded) or is using it in a more glanceable fashion (partially folded), which may provide a signal to malicious content about when it can put things on the screen. Combined with device orientation data, this could be used to obscure activity on the less visible screen segment.

Second, there are potentially two additional means by which screen segment data could leak per-user fingerprintable data. If an OS allows the user to position a window  freely in a way that spans screens, the segments API is way to leak the window position (the segment geometry would change based on window position). This is akin to desktop browsers exposing `window.screenX`/`screenY` which can be very user-specific. In addition, even with non-draggable windows but modifiable browser UI including toolbars, the segments would leak information about the UI/toolbar geometry.

darktears:
The number of segments changes when the screen goes between the folded state and unfolded flat indeed. That's the essence of the API in the first place which is the same goal than the CSS MQs added in [MQ Level 5](https://www.w3.org/TR/mediaqueries-5/#mf-horizontal-viewport-segments). Here we're just basically providing the same functionality to JavaScript developers where CSS isn't a good fit (e.g. Canvas based content). 

I think we need to take a step back here. User is expected to pay full attention when the device is in a folded state. The entire idea of the form factor is to provide bigger screen estate for content when the device is half folded and the keyboard detached. Even on smaller foldable form factors like smartphones typically content author will display a split ux. Have a look [here](https://github.com/foldable-devices/demos) to see various pictures of the device in action.

Regarding the potential fingerprinting, on Chromium browser segments are only returned when the browser window is in fullscreen or maximized, this is an implementation choice which we can decide to standardize or not. We did not hear from content author the need to get segments when the window isn't maximized and floating around. As a clarification the browser window is not resized when changing state (e.g. folded to unfolded) but could be if the keyboard is docked for example.

I want to emphasize that the segments are defined in client coordinates (so adjusted to zoom factor and so forth) therefore relative to the viewport. I'm not sure how much information they would "leak" provided that say top segment + bottom segment will be equivalent to the inner height/width of the viewport. Could you use these APIs to fingerprint that it's a folded device? Yes sure but again today you can do that by checking the window dimensions because they have a not so common sizes (including for example Android where the dimensions are off for a tablet and a typical phone). Can you use these APIs to fingerprint a specific foldable model? Likely no because many of them share the same panel dimensions/sizes (phone or not) but again even if one had an exotic panel, they add nothing more than what I could already derive from the viewport size.

Thanks for your comments. I'll be happy to clarify more if you need so.

cssom-view-1/Overview.bs
```diff
@@ -1806,6 +1806,8 @@ interface VisualViewport : EventTarget {
 
   readonly attribute double scale;
 
+  readonly attribute FrozenArray&lt;DOMRect> segments;
```
emilio, line 1809: I don't think you need the `&lt;`?

emilio: I don't think a `FrozenArray` can be used tho, right? A `FrozenArray` is supposed to be immutable (https://webidl.spec.whatwg.org/#idl-frozen-array), but this is mutable right?

emilio: I think it should just be a `sequence`?

cssom-view-1/Overview.bs
```diff
@@ -1856,6 +1858,19 @@ The <dfn attribute for=VisualViewport>scale</dfn> attribute must run these steps
...
+
+The segments attribute must run these steps:
+1. If the <a>visual viewport</a>'s <a for=visualviewport>associated document</a> is not <a>fully active</a>, return 0.
+2. Returns null if called from within an iframe context and abort these steps.
+3. Returns null if there is only a single viewport segment and abort these steps.
+4. Otherwise, return the <a>visual viewport</a>'s segments array.
+
+Note: Based on the data returned for each viewport segment, developers will be able to infer the number of hinges available as well as the hinge orientation.
+The browser window can be moved/resized such that the number of segments and/or their dimensions change.
+In these cases, resize events will fire at which point authors should re-query this property, as the returned FrozenArray is just a snapshot of the current state.
```
emilio, line 1872: It'd be good to note here that these are returned in un-zoomed CSS pixels. It's a bit weird that these are part of the VisualViewport API even though they don't have anything to do with it right? In particular, they don't change when zooming and panning, right?

Or do those environment variables change when zooming and panning? That seems unfortunate if so.

cssom-view-1/Overview.bs
darktears: > I don't think you need the `&lt;`?

I get an error otherwise.

darktears: > I don't think a `FrozenArray` can be used tho, right? A `FrozenArray` is supposed to be immutable (https://webidl.spec.whatwg.org/#idl-frozen-array), but this is mutable right?

It's frozen because it's not changing, just like any other property of visualviewport. You're expected to listen for some events to query the new state (e.g. resize). If just the posture is changed then [Device Posture API](https://www.w3.org/TR/device-posture/) has an event listener specifically for that. Both APIs are complimentary.

darktears: > I think it should just be a `sequence`?

Probably a good idea since it can be zero length.

fantasai (commented):
Adding new features (or substantially changing existing ones) requires CSSWG approval, not just editor approval... it doesn't seem like https://github.com/w3c/csswg-drafts/issues/9237 has that?

cssom-view-1/Overview.bs
emilio: > It's frozen because it's not changing, just like any other property of visualviewport. You're expected to listen for some events to query the new state (e.g. resize). If just the posture is changed then [Device Posture API](https://www.w3.org/TR/device-posture/) has an event listener specifically for that. Both APIs are complimentary.

Well but that means that they do change (when the event fires) right? or what am I missing?

darktears: Here is a concrete example: initial page load and the device is used flat (or a laptop posture with the keyboard docked) the `segments` property will be 0 meaning an array of 0 length. 

If I change the posture to folded (or detach the keyboard) then the events will fire and the `segments` property will have an array of 2 segments representing the two sides beside the fold/hinge. 

Just like the other properties of the visual viewport segments they represent the given state at the time `window.visualViewport` was queried.

emilio: Right, I get that, the point is that my understanding is that `FrozenArray`s aren't supposed to change, ever, not even in reactions to an event. @domenic is that right?

domenic: FrozenArray properties can change. The contents of the array cannot, but a new array can start being returned.

domenic: Note that it is invalid for attributes to be sequences; that is part of why FrozenArray exists.

cssom-view-1/Overview.bs
darktears: They don't change when zooming or panning. But unless I don't understand your point, I'm not sure why they have "anything to do" with the viewport? For example, on a symmetrical folding device which is in a folded state, they're helping developers to implement say a split UX/dual pane view in their app/websites which is basically dividing the viewport in half logically and they need the segments to do so. But again, maybe I misinterpreted your comment.

darktears: @emilio : Take a look [here](https://github.com/foldable-devices/demos). You can see in the pictures what the API is trying to achieve in JS, similar to what the CSS API is doing.

aykutbulut:
Privacy section in the [explainer](https://github.com/WICG/visual-viewport/blob/gh-pages/segments-explainer/SEGMENTS-EXPLAINER.md) indicates that segments will return null when called from within an iframe and explains how this prevents using the API as a fingerprinting vector from cross origin iframes.

A concern I have is accessing the segments through [CSS media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries#targeting_media_features). Having CSS API in iframes might enable this type of fingerprinting. Is disabling CSS media queries for the segments feasible?

darktears:
> Privacy section in the [explainer](https://github.com/WICG/visual-viewport/blob/gh-pages/segments-explainer/SEGMENTS-EXPLAINER.md) indicates that segments will return null when called from within an iframe and explains how this prevents using the API as a fingerprinting vector from cross origin iframes.
> 
> A concern I have is accessing the segments through [CSS media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries#targeting_media_features). Having CSS API in iframes might enable this type of fingerprinting. Is disabling CSS media queries for the segments feasible?

Do we have precedence in doing that? I couldn't find anything in CSS specifications where certain CSS features are restricted to iframes, especially in the MQ/CSS env context but the domain is vast so I could have been missing something. But again, the real question is whether the fingerprinting is really that big of an issue if segments are just subdivisions of the total viewport size. Foldables already have exotic viewport size when unfolded so they can be identified already but as more devices are in the market, targeting a specific device is becoming harder as they share panel dimensions so segments definition as well. Yes, it does add a bit of entropy but nothing that couldn't be derived from the viewport size today (or even combining with other APIs)

aykutbulut:
I am not aware of such a precedence. I agree with your arguments.

cssom-view-1/Overview.bs
```diff
@@ -1981,6 +1996,9 @@ Changes {#changes}
 This section documents some of the changes between publications of this specification. This section is not exhaustive. Bug fixes and editorial changes are
 generally not listed.
 
+<h3 id='changes-from-2022-07-07' class=no-num>Changes From 14 September 2023</h3>
```
dbaron, line 1999: The ID here should match the title.

(It's also not clear to me why there are so many changes sections in this spec.  I'd normally expect one changes section for each new publication to https://www.w3.org/TR/cssom-view-1/ .)

darktears:
Close this PR per CSS WG resolution where we want the `segments` property to belong on a new field in window.

PR #9296, opened 2023-09-03
[css-backgrounds-3][css-borders-4] Fixed definition for where box shadows apply

### LINKED ISSUE #9286: [css-backgrounds-3][css-borders-4] Only the principal box should get a shadow
https://drafts.csswg.org/css-backgrounds-3/#shadow-layers, https://drafts.csswg.org/css-borders-4/#shadow-layers

> If an element has multiple boxes, all of them get drop shadows, but shadows are only drawn where borders would also be drawn; see [`box-decoration-break`](https://drafts.csswg.org/css-break-4/#propdef-box-decoration-break).

This is wrong: only the [principal box](https://drafts.csswg.org/css-display-3/#principal-box) should get the shadow. If that box has multiple fragments, then all of them get shadows, according to `box-decoration-break`.

SebastianZ:
This corrects the description for where box shadows are applied.

Fixes #9286

### DIFF
--- a/css-backgrounds-3/Overview.bs
+++ b/css-backgrounds-3/Overview.bs
@@ -28,6 +28,7 @@ spec:css2; type:value; text:inline-table
 spec:css2; type:value; text:table-cell
 spec:css2; type:property; text:overflow
 spec:css2; type:value; text:visible
+spec:css-break-4; type:dfn; text:fragment
 spec:css-color-4; type:property; text:color
 spec:css-color-4; type:value; text:transparent
 spec:selectors-3; type:selector; text: ::first-letter
@@ -2512,7 +2513,8 @@ Layering, Layout, and Other Details</h4>
   and the [=inner shadows=] of an element are drawn immediately above the background of that element
   (below the borders and border image, if any).
 
-  <p>If an element has multiple boxes, all of them get drop shadows,
+  <p>Drop shadows are only applied to the [=principal box=].
+  If that box has multiple [=fragments=], then all of them get shadows,
   but shadows are only drawn where borders would also be drawn;
   see 'box-decoration-break'.
 
@@ -2591,6 +2593,8 @@ Changes since the 14 February 2024 Candidate Recommendation Snapshot</h3>
     to include “Animation Type” and “Logical Property Group”.
   <li>Streamlined property grammar definitions using the latest
     [=CSS/value definition syntax=].
+  <li>Fixed the definition for where box shadows apply.
+    (<a href="https://github.com/w3c/csswg-drafts/issues/9286">Issue 9286</a>)
 </ul>
 
 
--- a/css-backgrounds-3/Overview.bs
+++ b/css-backgrounds-3/Overview.bs
@@ -2513,10 +2513,9 @@ Layering, Layout, and Other Details</h4>
   and the [=inner shadows=] of an element are drawn immediately above the background of that element
   (below the borders and border image, if any).
 
-  <p>Drop shadows are only applied to the [=principal box=].
-  If that box has multiple [=fragments=], then all of them get shadows,
-  but shadows are only drawn where borders would also be drawn;
-  see 'box-decoration-break'.
+  <p>Unless otherwise specified, drop shadows are only applied to the [=principal box=].
+  If the affected box has multiple fragments,
+  the shadows are applied as specified in 'box-decoration-break'.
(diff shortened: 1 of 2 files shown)


Loirooriol (commented):
So I was expecting the behavior to be the principal box since that's the simple choice, but I have tried a table:

```html
<!DOCTYPE html>
<table style="outline: 3px solid cyan; box-shadow: 0 0 0 6px magenta">
  <caption>Tablecaption</caption>
  <tr><td>cell</td></tr>
</table>
```

Both Firefox and Chromium apply the shadow to the table grid box, not to the principal table wrapper box.

So maybe it should be discussed whether this should change? Or say that it's the principal box unless otherwise defined by the relevant spec (https://www.w3.org/TR/CSS21/tables.html#model would imply table grid box, and seemingly https://drafts.csswg.org/css-tables/#global-style-overrides too)

I can't find it now, but I remember some discussion about defaulting properties on a table to apply to the principal wrapper box, with some exceptions that apply to the grid box. Now the default is the grid box, which seems a bit weird.

fantasai:
@Loirooriol Can you drop that comment into the issue?

css-backgrounds-3/Overview.bs
```diff
@@ -2504,7 +2505,8 @@ Layering, Layout, and Other Details</h4>
   and the <i>inner shadows</i> of an element are drawn immediately above the background of that element
   (below the borders and border image, if any).
 
-  <p>If an element has multiple boxes, all of them get drop shadows,
+  <p>Drop shadows are only applied to the [=principal box=].
+  If that box has multiple [=fragments=], then all of them get shadows,
```
fantasai, line 2509: This is incorrect. How box-shadow is handled on fragments depends on `box-decoration-break`, see https://www.w3.org/TR/css-break-4/#break-decoration

Loirooriol: Isn't this addressed in the next sentence?

>  but shadows are only drawn where borders would also be drawn;
>  see 'box-decoration-break'.

fantasai: No, it's not that we have multiple shadows some parts of which aren't drawn. box-decoration-break determines whether the entire box gets shadowed and then sliced or the fragments get shadowed. If you offset the shadow by more than the width of the first fragment and then blur it, for example, the first fragment won't have a shadow attached, but there will be a shadow as wide as the combined fragments, blurred as a single unit, attached to the second fragment.

fantasai, line 2509:
```
  <p>Unless otherwise specified, drop shadows are only applied to the [=principal box=].
  If the affected box has multiple fragments,
  the shadows are applied as specified in 'box-decoration-break'.
```

css-backgrounds-3/Overview.bs
```diff
@@ -2599,6 +2601,8 @@ Changes since the 14 February 2024 Candidate Recommendation Snapshot</h3>
     (<a href="https://github.com/w3c/csswg-drafts/issues/8496">Issue 8496</a>)
   <li>Fixed an error in the Computed Value line of 'background-image'.
     (<a href="https://github.com/w3c/csswg-drafts/issues/8604">Issue 8604</a>)
+  <li>Fixed the definition for where box shadows apply.
+    (<a href="https://github.com/w3c/csswg-drafts/issues/9286">Issue 9286</a>)
```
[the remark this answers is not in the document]
SebastianZ: Not sure whether @fantasai's comment was related to this, though I went ahead and fixed this obvious mistake.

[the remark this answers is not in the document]
fantasai: No, I was commenting on the actual spec text.

Loirooriol (commented):
You didn't add the "Unless otherwise specified" into css-background-3. There is a suggested wording from fantasai.

css-backgrounds-3/Overview.bs
SebastianZ: Done.

SebastianZ:
> You didn't add the "Unless otherwise specified" into css-background-3. There is a suggested wording from fantasai.

I made up for that now.

Sebastian

SebastianZ:
@Loirooriol @fantasai Can you please have another look?

Sebastian

Loirooriol (approved):

Issue #9257, opened 2023-08-28
[selectors-4] Should :user-valid and :user-invalid apply to form and fieldset elements?


josepharhar:
The selectors-4 spec says this here after the definition of :user-valid and :user-invalid: https://drafts.csswg.org/selectors-4/#issue-df919919
> Clarify that this (and [:invalid](https://drafts.csswg.org/selectors-4/#invalid-pseudo)/[:valid](https://drafts.csswg.org/selectors-4/#valid-pseudo)) can apply to form and fieldset elements.

I made a quick demo here: https://jsfiddle.net/jarhar/Lohxsz2n/30/
It looks like :valid and :invalid apply to form elements, whereas :user-valid and :user-invalid don't apply to forms. This behavior is the same across chrome, firefox, and safari.

Should we standardize this behavior which happens to already be interoperable? Or should we make changes to make :user-valid and :user-invalid apply to forms and fieldsets?

@nt1m

nt1m:
The newly added HTML spec currently explicitly does not apply the :user-* pseudos on fieldset / form, I haven't really thought too much about this personally.

If we really wanted to define those pseudos on fieldset / form, they could be defined as the equivalent of `:has(:user-invalid)` or `:has(:user-valid)` for the respective elements. From an user experience perspective that probably makes sense, though status quo is fine too.

cc @pxlcoder @emilio Any opinions?

emilio:
I think I have a slight preference for making `form` / `fieldset` never match these. The code tracking these in gecko is not amazing, I'm not excited about making it more complicated, and I've never seen a use case for these... But if there are use cases that need this I'd be ok supporting these.

I don't think it's fine to make these effectively `:has(:user-valid)` / `:has(:user-invalid)`, however. That allows a form / fieldset to match _both_ at the same time. Probably should be:

 * `:user-invalid` matches if there's at least one `:user-invalid` associated element underneath.
 * `:user-valid` matches if there's at least one `:user-valid` associated element underneath _and_ there are no `:invalid` elements underneath or so (so `:valid:has(:user-valid)`?)

emilio:
I think they don't match per spec already right? https://html.spec.whatwg.org/#selector-user-valid mentions it only matches `input`, `textarea` and `select` elements.

josepharhar:
Yeah, we just need to do something about this note in the CSS spec: https://drafts.csswg.org/selectors-4/#issue-df919919
[...]

PR #9308, opened 2023-09-05
[selectors-4] Replace :user-valid/:user-invalid definition with HTML spec

josepharhar:
Fixes https://github.com/w3c/csswg-drafts/issues/9257

As suggested here, this definition is not up to par with the HTML definition and should be replaced by it: https://github.com/w3c/csswg-drafts/issues/9257#issuecomment-1705426225


selectors-4/Overview.bs
```diff
@@ -2813,52 +2813,12 @@ The Optionality Pseudo-classes: '':required'' and '':optional''</h4>
...
-		&lt;form>
-		  &lt;label>
-		    Volume:
-		    &lt;input name='vol' type=number min=0 max=10 value=11>
-		  &lt;/label>
-		  ...
-		&lt;/form>
-		</pre>
-	</div>
+  The '':user-valid'' and '':user-invalid'' pseudo-classes are defined in HTML here: https://html.spec.whatwg.org/#selector-user-valid
```
emilio, line 2816: I think I'd probably do `<a href="https://html.spec.whatwg.org/#selector-user-valid">in HTML</a>.` or so (or using some bikeshed magic to get that), rather than dropping the raw URL there.

@tabatkins or @fantasai should look at this as editors, but lgtm for the general change, we shouldn't have two incompatible definitions of these.

emilio (approved):
Not sure editorially what'd be the best way to do this but the intention of the change looks good. @fantasai / @tabatkins could you take a look?

selectors-4/Overview.bs
nt1m: I suspect we'll want language like "are defined by the host language" along with a vague description of what the selectors are about.

josepharhar: Ok, I brought back the summary paragraph and said that its defined by the host language instead of html

selectors-4/Overview.bs
```diff
@@ -2818,47 +2818,13 @@ The User-interaction Pseudo-classes: '':user-valid'' and '':user-invalid''</h4>
...
-		  &lt;label>
-		    Volume:
-		    &lt;input name='vol' type=number min=0 max=10 value=11>
-		  &lt;/label>
-		  ...
-		&lt;/form>
-		</pre>
-	</div>
+	The '':user-valid'' and '':user-invalid'' pseudo-classes are defined by the
+	host language.
```
nt1m, line 2822: Could be good to mention "Exact semantics of ", but I'll leave the nitpicking up to the editors :)

tabatkins, line 2822:
```suggestion
	The exact matching behavior of '':user-valid'' and '':user-invalid'' 
	are defined by the host language.
	See, for example, <a spec=html lt=":user-invalid" selector>HTML's definition</a>.
```

selectors-4/Overview.bs
```diff
@@ -2818,47 +2818,13 @@ The User-interaction Pseudo-classes: '':user-valid'' and '':user-invalid''</h4>
...
-	interacted again with the form element.
-
-	User-agents may allow them to match such elements at other times,
-	as would be appropriate for highlighting an error to the user.
-	For example, a UA may choose to have '':user-invalid'' match an
-	'':invalid'' element once the user has typed some text into it and
-	changed the focus to another element, and to stop matching only after
-	the user has successfully corrected the input.
-
-	<div class='example'>
```
tabatkins, line 2838: I'd leave this example in, it's a useful thing for the reader and matches what HTML defines.

tabatkins (changes requested):
Dropped a suggestion, and think we should restore the example, but after that r+

selectors-4/Overview.bs
josepharhar: ok, I made it say "the exact semantics of :user-valid ..."

josepharhar: Nevermind, I'm going with tabs suggestion for this sentence

selectors-4/Overview.bs
josepharhar: ok, i put the example back in

josepharhar:
@tabatkins can we merge this?

fantasai:
I think this approach doesn't do a good job of keeping a baseline definition for host languages that don't define their own rules. Let me see if I can come up with a sort of halfway step here...

josepharhar:
> Let me see if I can come up with a sort of halfway step here...

Happy to take any suggestions or a new PR

dbaron:
So I think the existing definition that this is removing is a bit of a mess:  it seems to talk about relationships between some of the other pseudo-classes (`:invalid`, `:out-of-range`, `:required`) that should perhaps be defined in the sections for those pseudo-classes but probably shouldn't be defined here.

I think the essence of these pseudo-classes is that they are subsets of `:valid` and `:invalid` that only match following certain user interactions that should call attention to the validity/invalidity, such as the user interacting with the control or the user attempting to submit the form.  HTML mandates both of these; the text in CSS mandated only the second (submission) but not the first (interaction).  The text in CSS also mandated a behavior that the HTML spec *doesn't* have, which is that they stop matching when the user interacts with the control again.

So I think perhaps the key things to add back to address @fantasai's concern are:
* making the parallel with `:valid` and `:invalid` more explicit again (but hopefully dropping the muddled bit about the other pseudos)
* explicitly mentioning submitting the form as a type of (or alternative to) significant user interaction with the control

Hopefully we're ok with dropping the bit about no longer matching when the user returns to the control, given that HTML doesn't say that.  (Or do any engines implement that?)

josepharhar:
> Hopefully we're ok with dropping the bit about no longer matching when the user returns to the control, given that HTML doesn't say that. (Or do any engines implement that?)

Yeah I don't think this is a thing. I think that the only case we start matching these and then stop matching them is when the form is reset.

> other pseudo-classes (:invalid, :out-of-range, :required) that should perhaps be defined in the sections for those pseudo-classes but probably shouldn't be defined here.

> making the parallel with :valid and :invalid more explicit again (but hopefully dropping the muddled bit about the other pseudos)

Ok, I added a paragraph to the :valid and :invalid sections. How does it look?

> explicitly mentioning submitting the form as a type of (or alternative to) significant user interaction with the control

I added a paragraph for this. How does it look?

PR #9320, opened 2023-09-07
[css-gcpm-3] Remove duplicate definitions of <string()>

cdoublev:
`<string()>` is defined with the same value definition in [CSS Content 3](https://drafts.csswg.org/css-content-3/#funcdef-string) and [CSS GCPM 3](https://drafts.csswg.org/css-gcpm-3/#funcdef-string).

This PR removes the definition in CSS GCPM 3.


tabatkins:
(Merging this, even tho I'm not the editor, as this affects data outside of CSS and should be resolved quickly.)

PR #9323, opened 2023-09-07
[css-gcpm-4] Fix <content-level> definition

### LINKED ISSUE #2919: [css-gcpm] "content-list" should rather be "<content-level>"
The [`copy-into` property definition](https://drafts.csswg.org/css-gcpm-4/#propdef-copy-into) references `<content-level>` defined shortly after. However, the [actual definition of `content-level`](https://drafts.csswg.org/css-gcpm-4/#content-list) uses `content-list` instead of `content-level`.

Also, for consistency with other specs, the left-side of the equation should rather be enclosed in `<>`, i.e. `<content-level>`.

cdoublev:
Fixes #2919.

I removed [`<attr()>`](https://drafts.csswg.org/css-values-5/#attr-notation) from its definition because it can be used at any place in a property value.

### DIFF
--- a/css-gcpm-4/Overview.bs
+++ b/css-gcpm-4/Overview.bs
@@ -65,10 +65,10 @@ Animation type: discrete
 
 The 'copy-into' property contains one or more pairs, each consisting of a custom identifier followed by a content-level keyword describing how to construct the value of the named content fragment.
 
-''content-level'' expands to the following values:
+<<content-level>> expands to the following values:
 
 <pre class="prod">
-	<dfn id="content-list">content-list</dfn> = element | content | text | attr(&lt;identifier>) | counter() | counters()
+	<dfn>&lt;content-level></dfn> = element | content | text | <<counter>>
 </pre>
 
 
@@ -77,7 +77,7 @@ The 'copy-into' property contains one or more pairs, each consisting of a custom
 <dd>the entire element is copied into the named content fragment</dd>
 
 <dt>contents</dt>
-<dd>only the element’s contents are copied into the named content fragment. This is the default if ''content-level'' is not specified.</dd>
+<dd>only the element’s contents are copied into the named content fragment. This is the default if <<content-level>> is not specified.</dd>
 
 <dt>text</dt>
 <dd>only the element’s text (including normally collapsed white space) is copied into the named content fragment.</dd>


fantasai (changes requested):
Don't remove attr(). We can't rely on global attr() here, since the generated content properties are defined earlier than global attr()'s acceptance (which is still not very accepted!)

cdoublev:
Do you think it can be updated to `<attr()>` rather than preserving `attr(<identifier>)`?

[`<identifier>`](https://drafts.csswg.org/css2/#value-def-identifier) is only used in CSS 2 for [`content`](https://drafts.csswg.org/css2/#propdef-content), [`counter-reset`](https://drafts.csswg.org/css2/#propdef-counter-reset), [`counter-increment`](https://drafts.csswg.org/css2/#propdef-counter-increment).

It seems inappropriate for `counter-increment` and `counter-reset` because [CSS Lists 3](https://drafts.csswg.org/css-lists-3/#propdef-counter-increment) and browsers do not accept CSS-wide keywords or `default` (even when followed by an explicit integer), as well as in [`counter()`](https://drafts.csswg.org/css-lists-3/#typedef-counter).

CSS Content 3 redefines [`content`](https://drafts.csswg.org/css-content-3/#propdef-content) without `attr(<identifier>)` and implicitly accepts the global `<attr()>`, and `string-set` accepts `<attr()>` in CSS GCPM 3.

---

Similarly, do you think it would be fine to replace `<content-list>` by `<content-list> | <attr()>` in the value definition of [`string-set`](https://drafts.csswg.org/css-gcpm-3/#propdef-string-set) in CSS GCPM 3, and to remove its production rule of `<content-list>`, so that it references [the production rule in CSS Lists 3](https://drafts.csswg.org/css-content-3/#typedef-content-content-list)? The goal is to remove duplicate `<content-list>` production rules (cf. #9318).

Note that it would mean that `string-set` accepts eg. `<image>` as `<content-list>`, whereas it is not accepted in CSS GCPM 3.

cdoublev:
I think all the changes suggested in this PR and the comment above are applied in https://github.com/w3c/csswg-drafts/commit/017d0727b2e6aba4ca4fd44ec4b7bf8c4cc84bf6, except for the unrelated change to `<content-list>` in CSS GCPM 3, for which I will later create a dedicated issue.

Issue #9239, opened 2023-08-24
[css-anchor-position-1] Snapshotted scroll offset needs to consider both axis separately


xiaochengh:
(Originally filed as a [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1473388))

From the current spec, an absolutely positioned element `query el` has a snapshotted scroll offset if:
> - there is a [target anchor element](https://drafts.csswg.org/css-anchor-position-1/#target-anchor-element) given the [default anchor specifier](https://drafts.csswg.org/css-anchor-position-1/#default-anchor-specifier) of query el
> - and at least one [anchor()](https://drafts.csswg.org/css-anchor-position-1/#funcdef-anchor) function on query el refers to the same target anchor element

Proposal: See https://github.com/w3c/csswg-drafts/issues/9239#issuecomment-1694009908

~~I propose removing the second condition, because:~~
1. ~~This condition is complicated to check. We need to go through both the base style and all the fallback styles~~
2. ~~It doesn't work with the new `anchor-center` alignment keyword, which can be used alone without any explicit `anchor()` functions~~
   - ~~This also suggests that the condition isn't future-proof. If we introduce new anchoring syntax in the future, this condition has to grow more complicated~~
3. ~~Having a valid default anchor without actually positioning it against the anchor is a weird/incorrect usage. I don't see much value guarding against such usage~~

tabatkins:
> Having a valid default anchor without actually positioning it against the anchor is a weird/incorrect usage. I don't see much value guarding against such usage

I don't quite agree. A final fallback might reasonably be to just position in screen space, for example, without reference to any anchor at all, or position relative to a different anchor. Having the element still get shifted to match the scroll offset of the default anchor element seems weird/bad, and `anchor-default` isn't settable from the fallback styles so the author wouldn't have any way to stop that from happening.

That said, I sympathize with the "complicated to check", and you're right that anchor-center positions relative to the anchor without any anchor() functions at all, so *something* definitely has to change.

* The minimal change is to just generalize the condition to check for the element referring to the default anchor element in any way (either via anchor() or via anchor-center alignment, or any future ways we add).
[...]

PR #9347, opened 2023-09-13
[css-anchor-position-1] Make snapshotted scroll offset per-axis

xiaochengh:
For #9239 

Define whether there's a snapshotted scroll offset on each axis separately, and then revise the concept of snapshotted scroll snapshot and how it's used accordingly.

### DIFF
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -737,7 +737,7 @@ we define:
 	* At least one ''anchor()'' function on |query el|'s
 		used [=inset properties=] in the axis
 		refers to a [=target anchor element=]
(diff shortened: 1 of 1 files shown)


xiaochengh:
Wait, this doesn't seem to work for the multiple anchor case.

Say the element uses anchor `--a` in the x axis and anchor `--b` in the y axis, and both anchors are in the same scroller. We still want to do scroll adjustments in both axes, but there's only one default anchor.

I think we should refine it into: do adjustment in an axis if there's reference to any anchor element in the same scroller (but not any descendant scroller) as the default anchor in that axis.

xiaochengh:
Fixed it, PTAL.

Issue #918, opened 2017-01-13
Preserve-3d + backface visibility semantics need to be clarified


smfr:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=23015

smfr:
We talked about this at the Fukuoka 2019 F2F. Outcome of that discussion was that this should be specified in terms of "pseudo stacking contexts" which need to be defined in another spec (a painting spec).

In the meantime, css-transforms-2 certainly needs to more clearly specify backface-visibility:hidden. Matt, would you like to take a stab at this?

chrishtr:
There are two aspects to this issue: preserve-3d and backface-visibility.

For backface-visibility, there was a resolution earlier this week that `backface-visibility:hidden` should also apply to all descendants that are not pseudo stacking contexts [1].


[1] I think we should define a pseudo stacking context as meeting one of the following criteria:

chrishtr:
For preserve-3d: we could use the same concept.

Proposal: only elements that induce pseudo stacking contexts can cause flattening.

Thus this example would preserve 3d across the `#middle` div:

```
<div style="transform-style: preserve-3d">
  <div id=middle>
    <div style="transform: rotateX(100px)"></div>
  </div>
</div>
```

But not:

```
<div style="transform-style: preserve-3d">
  <div id=middle style='position: relative'>
    <div style="transform: rotateX(100px)"></div>
  </div>
</div>
```

This will probably improve interop, makes it a bit easier to structure content in some cases, and may be more consistent to reason about if we do the same thing for `backface-visibility: hidden`.

chrishtr:
Now for perspective CSS:

Blink and WebKit both use the enclosing stacking context ancestor that is composited. Therefore, behavior is a side-effect of compositing. :(

If an element has a 3D transform, then its nearest DOM ancestor with perspective is also composited (and because it has perspective it is also a stacking context). However, there may be an intermediate stacking ancestor that is below the perspective and above the transformed element.

Example 1 (perspective is applied to `#inner` in Blink and WebKit:


```
<!doctype HTML>
<div id=p style='perspective: 100px'>
  <div style="position: relative; will-change: transform">
    <div id=target style='transform: translateZ(-200px); background: lightblue; width: 100px; height: 100px'>
  </div>
</div>
```
Example 2 (no perspective applied in Blink (still does in WebKit, not sure why), only because of compositing.


```
<!doctype HTML>
<div id=p style='perspective: 100px'>
[...]

PR #9348, opened 2023-09-13
[css-transforms-2] Make backface-visibility:hidden create a stacking context and containing block when participating in a 3D rendering context.

mattwoodrow:
The first (simpler) half of what was discussed in https://github.com/w3c/csswg-drafts/issues/918.

Will try write some new tentative tests for this too.


sergioandrade (approved):

dbaron (approved):
Looks good to me.  Probably good to mention #918 in the commit message.

PR #9374, opened 2023-09-18
[css-inline-3][css-page-3][css-text] Enforce non-negative value from …

cdoublev:
Adds `[0,∞]` to CSS types in contexts that require a non-negative value.

### DIFF
--- a/css-inline-3/Overview.bs
+++ b/css-inline-3/Overview.bs
@@ -964,7 +964,7 @@ Line Spacing: the 'line-height' property</h3>
 
 	<pre class="propdef">
 	Name: line-height
-	Value: normal | <<number>> | <<length-percentage>>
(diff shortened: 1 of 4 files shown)


frivoal (approved):

frivoal:
Looks fine to me, and it's OK IPR-wise, as this is an editorial change.
@fantasai, as editor of css-inline and css-page, can you confirm this is fine by you?

cdoublev:
Please let me know if you think it is ok to add `[1,∞]` to `fixed <integer>?` in [`system`](https://drafts.csswg.org/css-counter-styles-3/#descdef-counter-style-system) for `@counter-style`. The integer is the required number of counter symbols that the appropriate descriptor must have.

This is not defined in prose, whereas this PR only includes range additions (`[0,∞]`) that are already defined in prose for now.

fantasai (approved):

PR #9376, opened 2023-09-19
[css-fonts-4] Make color interpolation method optional in palette-mix()

cdoublev:
`<color-interpolation-method>` is defined as optional in prose for [`<palette-mix()>`](https://drafts.csswg.org/css-fonts-4/#typedef-font-palette-palette-mix):

  > When not specified otherwise the <color-interpolation-method> is in oklab as recommended by CSS Color 5 § 7.1 Color Space for Interpolation.

But not in its value definition:

  > `palette-mix(<color-interpolation-method> , [ [normal | light | dark | <palette-identifier> | <palette-mix()> ] && <percentage [0,100]>? ]#{2})`

I suppose this is an oversight.

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -7065,7 +7065,7 @@ Animation type: by computed value
 With the <code>palette-mix()</code> function defined as follows:
 
 <pre class="prod def" nohighlight>
-<dfn>palette-mix()</dfn> = palette-mix(<<color-interpolation-method>> , [ [normal | light | dark | <<palette-identifier>> | <<palette-mix()>> ] && <<percentage [0,100]>>? ]#{2})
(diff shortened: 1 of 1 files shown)


svgeesus:
I think the prose is incorrect, because in discussions it was clear that `palette-mix()` was closely based on `color-mix()` where the `color-interpolation-method` is _not_ optional.

But I think there will be ongoing discussion on this, so let's leave this PR open while that happens.

svgeesus:
Further discussion aligned `palette-mix()` with `color-mix()`, so the color interpolation method is (per grammar) not optional.

Closing this PR therefore and I will remove the prose mention of it being optional.

svgeesus:
Fixed by https://github.com/w3c/csswg-drafts/commit/1a415a2daa3edfc0f902c2f80cde3cfcaae09563

Issue #4951, opened 2020-04-15
[css-sizing-3] Mapping HTML’s IMG aspect-ratio behavior into CSS.


fantasai:
HTML accepted to map the `width` and `height` attributes of the IMG element (and some others) to an aspect ratio for the purpose of rendering in https://github.com/whatwg/html/pull/4952 . Currently this is specified as providing an “intrinsic aspect ratio” until the image loads and is able to provide its own. Since the current `aspect-ratio` property doesn't have a way to indicate "use this value until the element can provide its own", the behavior can't be mapped directly to CSS.

But we do prefer having rendering rules mappable to CSS, so the question is how should we make this possible? The two questions are:

1. What syntax to use to indicate "here's an aspect ratio to use but only until the thing loads and provides its own info"?
2. How do we handle the nuances of `box-sizing`? Currently an explicit aspect ratio (`<ratio>`) uses the specified box, whereas an intrinsic one (`auto`) uses content-box always. Probably the IMG mapping should continue to use content-box always. Is this a function of the above-mentioned syntax, or an extra switch?

Wrt 1, I suggest just allowing the `auto` and `<ratio>` values to be combined. (The current syntax is `auto | <ratio>`, this would make it `auto || <ratio>` with the combined value treating `<ratio>` as fallback only.)

Wrt 2, we could consider one or both of:
- Combining `auto` and `<ratio>` triggers using content-box always.
- Add an optional `box-sizing` keyword into the syntax, e.g. `auto || <ratio> || <visual-box>`.

My inclination is to do both.

PR #9384, opened 2023-09-20
[css-sizing-4] Fix typo

cdoublev:
[Resolution](https://github.com/w3c/csswg-drafts/issues/4951#issuecomment-621337849)

  > RESOLVED: Use the `<aspect-ratio>||auto` syntax

[Commit](https://github.com/w3c/csswg-drafts/commit/2ccdc2921ce66217d60fd29653dd3901c856f810) for the current value:

  > **Value:** `auto || <<ratio>>`
  >
  > **`auto && <ratio>`**: [...]

https://drafts.csswg.org/css-sizing-4/#propdef-aspect-ratio

### DIFF
--- a/css-sizing-4/Overview.bs
+++ b/css-sizing-4/Overview.bs
@@ -211,7 +211,7 @@ Preferred Aspect Ratios: the 'aspect-ratio' property</h3>
 			If the <<ratio>> is [=degenerate ratio|degenerate=],
 			the property instead behaves as ''aspect-ratio/auto''.
 
-		<dt><dfn>auto && <<ratio>></dfn>
+		<dt><dfn>auto || <<ratio>></dfn>
 		<dd>
 			If both ''aspect-ratio/auto'' and a <<ratio>> are specified together,
 			the [=preferred aspect ratio=] is the specified ratio


css-sizing-4/Overview.bs
```diff
@@ -211,7 +211,7 @@ Preferred Aspect Ratios: the 'aspect-ratio' property</h3>
 			If the <<ratio>> is [=degenerate ratio|degenerate=],
 			the property instead behaves as ''aspect-ratio/auto''.
 
-		<dt><dfn>auto && <<ratio>></dfn>
+		<dt><dfn>auto || <<ratio>></dfn>
```
Loirooriol, line 214: Wouldn't this imply that this case also describes lone `auto` and `<ratio>`?

cdoublev: Yes, please see the very first next sentence: `&&` requires them together (in any order) unconditionnally.

Loirooriol: I mean, lone `auto` and `<ratio>` are already defined above, so they shouldn't be included here. This case is just for the combination of them

cdoublev: Oh, now I see. I assume the authors do not mind using a value definition to name a term representing a subset of the property value definition. 

Sorry, I was not reading carefully and I did not expected it, obviously.

PR #9385, opened 2023-09-20
[css-view-transitions-2] Add transition types

### LINKED ISSUE #8960: [css-view-transitions-1] Possible perf footgun when styling pseudo-tree based on state
Currently in order to change the style of the generated pseudo-tree based on state of the DOM one has to use a `:has()` selector on the `html` element to capture that state and then chain the pseudo-elements from there to style them.
I have discussed this with @una, @lilles, and @andruud at CSS Day and they said that using a `:has()` on an element so high up the DOM tree may be a big perf issue with rendering possibly taking a few seconds in some cases.

Here is a codepen with a simple example: https://codepen.io/ydaniv/pen/QWJjgQp

Basically it has this:
```html
<main>
  <div id=source></div>
  <div id=target></div>
</main>
```

And styles:
```css
/* This is our source of trouble */
html:has(.switch)::view-transition-group(circle) {
  animation-duration: 1.2s;
}

#source {
  view-transition-name: circle;
 
 .switch & {
    display: none;
  }
}

#target {
  display: none;

  .switch & {
    display: block;
  }
}
```

And the JS for triggering VT is:

```javascript
document.startViewTransition(() => {
    main.classList.toggle('switch');
});
```

My concern is that this may become a common pattern and cause big frustration, probably even more so users than to authors.

It's possible to work around this by lifting all such state up to the `html` element and try to mitigate this with communicating this to authors.
However, I think it's worth trying to find another solution for avoiding it altogetehr.

noamr:
As per [resolution at TPAC](https://github.com/w3c/csswg-drafts/issues/8960#issuecomment-1719695337), adding a concept of "active types" to a view transition, with an `active-view-transition` pseudo-class that matches the root element if it has an active view transition that matches the list of active types.

To set the active types, `startViewTransition` now accepts options, like `startViewTransition(update, types: ["slide", "reverse"]})`

Closes #8960

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -221,7 +221,6 @@ It has the following syntax definition:
 An ''::active-view-transition()'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=] |viewTransition|, for which any of the following are true:
 
 * The <<vt-type-selector>> is ''*''
-* |viewTransition|'s [=ViewTransition/active types=] is null
 * |viewTransition|'s [=ViewTransition/active types=] [=list/contains=] at least one of the <<custom-ident>> values of the <<vt-type-selector>>.
 
 
@@ -299,11 +298,13 @@ The <dfn attribute for=PageRevealEvent>viewTransition</dfn> [=getter steps=] are
 ### {{Document/startViewTransition(options)}} Method Steps ### {#ViewTransition-start-with-options}
 
 	<div algorithm="start-vt-with-options">
-		The [=method steps=] for <dfn method for=Document>startViewTransition(|options|)</dfn> are as follows:
+		The [=method steps=] for <dfn method for=Document>startViewTransition(|callbackOrOptions|)</dfn> are as follows:
 
-		1. Let |viewTransition| be the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |options|'s {{StartViewTransitionOptions/update}}.
+		1. If |callbackOrOptions| is an {{UpdateCallback}}, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions| and return the result.
 
-		1. Set |transition|'s [=ViewTransition/active types=] to |options|'s {{StartViewTransitionOptions/types}}.
+		1. Let |viewTransition| be the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions|'s {{StartViewTransitionOptions/update}}.
+
+		1. Set |transition|'s [=ViewTransition/active types=] to |callbackOrOptions|'s {{StartViewTransitionOptions/types}}.
 
 		1. Return |viewTransition|.
 	</div>
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -324,20 +324,20 @@ The <dfn attribute for=PageRevealEvent>viewTransition</dfn> [=getter steps=] are
 
 		partial interface Document {
 
-			ViewTransition startViewTransition(optional (UpdateCallback or StartViewTransitionOptions)? callbackOrOptions = null);
+			ViewTransition startViewTransition((UpdateCallback or StartViewTransitionOptions?) callbackOptionsOrNull);
 		};
 	</xmp>
 
 ### {{Document/startViewTransition(options)}} Method Steps ### {#ViewTransition-start-with-options}
 
 	<div algorithm="start-vt-with-options">
-		The [=method steps=] for <dfn method for=Document>startViewTransition(|callbackOrOptions|)</dfn> are as follows:
+		The [=method steps=] for <dfn method for=Document>startViewTransition(|callbackOptionsOrNull|)</dfn> are as follows:
 
-		1. If |callbackOrOptions| is an {{UpdateCallback}}, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions| and return the result.
+		1. If |callbackOptionsOrNull| is an {{UpdateCallback}} or null, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions| and return the result.
 
 		1. Let |viewTransition| be the result of running [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions|'s {{StartViewTransitionOptions/update}}.
 
-		1. Set |transition|'s [=ViewTransition/active types=] to |callbackOrOptions|'s {{StartViewTransitionOptions/types}}.
+		1. Set |transition|'s [=ViewTransition/active types=] to |callbackOptionsOrNull|'s {{StartViewTransitionOptions/types}}.
 
 		1. Return |viewTransition|.
 	</div>
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -333,9 +333,9 @@ The <dfn attribute for=PageRevealEvent>viewTransition</dfn> [=getter steps=] are
 	<div algorithm="start-vt-with-options">
 		The [=method steps=] for <dfn method for=Document>startViewTransition(|callbackOptionsOrNull|)</dfn> are as follows:
 
-		1. If |callbackOptionsOrNull| is an {{UpdateCallback}} or null, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions| and return the result.
+		1. If |callbackOptionsOrNull| is an {{UpdateCallback}} or null, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOptionsOrNull| and return the result.
 
-		1. Let |viewTransition| be the result of running [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions|'s {{StartViewTransitionOptions/update}}.
+		1. Let |viewTransition| be the result of running [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOptionsOrNull|'s {{StartViewTransitionOptions/update}}.
 
 		1. Set |transition|'s [=ViewTransition/active types=] to |callbackOptionsOrNull|'s {{StartViewTransitionOptions/types}}.
 
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -223,6 +223,39 @@ An ''::active-view-transition()'' pseudo-class matches the [=document element=] 
 * The <<vt-type-selector>> is ''*''
 * |viewTransition|'s [=ViewTransition/active types=] [=list/contains=] at least one of the <<custom-ident>> values of the <<vt-type-selector>>.
 
+<div class=example>
+For example, the developer might start a transition in the following manner:
+```js
+document.startViewTransition({update: updateTheDOMSomehow, types: ["slide-in", "reverse"]});
+```
+
+This will activate any of the following '::active-view-transition()'' selectors:
+```css
+:root:active-view-transition(slide-in) {}
+:root:active-view-transition(reverse) {}
+:root:active-view-transition(slide-in, reverse) {}
+:root:active-view-transition(slide-in, something-else) {}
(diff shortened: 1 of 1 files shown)


noamr:
@khushalsagar @vmpstr ping?

css-view-transitions-2/Overview.bs
```diff
@@ -205,7 +205,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		```
 	</div>
 
+# Pseudo-classes # {#pseudo-classes}
 
+## Active View Transition Pseudo-class '':active-view-transition()''' ## {#the-active-view-transition-pseudo}
+
+The <dfn id='active-view-transition-pseudo'>:active-view-transition(* | <<transition-types>>)</dfn> pseudo-class represents
```
khushalsagar, line 212:
```suggestion
The <dfn id='active-view-transition-pseudo'>:active-view-transition(* | <<transition-types>>)</dfn> pseudo-class applies to
```

css-view-transitions-2/Overview.bs
```diff
@@ -205,7 +205,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+the root element of the document, if it has a matching [=active view transition=].
```
khushalsagar, line 213: Can you use semantic line breaks : https://rhodesmill.org/brandon/2012/one-sentence-per-line/. We went with that for the L1 spec.

css-view-transitions-2/Overview.bs
```diff
@@ -205,7 +205,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+It has the following syntax definition:
+
+<pre class=prod>
+	:active-view-transition(<<vt-type-selector>>)
+	<<vt-type-selector>> = '*' | <<custom-ident>>#
```
khushalsagar, line 218: This can be a comma separated list as per the resolution. This grammer doesn't capture that?

css-view-transitions-2/Overview.bs
```diff
@@ -205,7 +205,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+</pre>
+
+An ''::active-view-transition()'' [=selector=] matches the corresponding [=document element=] if its
```
khushalsagar, line 221:
```suggestion
An ''::active-view-transition()'' [=selector=] matches the [=document element=] if its
```

Corresponding to? I didn't follow that and also seems like its not necessary.

css-view-transitions-2/Overview.bs
```diff
@@ -264,6 +279,33 @@ Note: this event is fired when [=reveal document|revealing a document=].
...
+		dictionary StartViewTransitionOptions {
+			UpdateCallback? update = null;
+			sequence<DOMString>? types = null;
+		};
+
+		partial interface Document {
+			ViewTransition startViewTransition(optional StartViewTransitionOptions options);
+		};
+
+		callback UpdateCallback = Promise<any> ();
```
khushalsagar, line 294: The L1 spec already defines it: https://drafts.csswg.org/css-view-transitions-1/#callbackdef-updatecallback.

css-view-transitions-2/Overview.bs
noamr: The `#` means comma-separated

css-view-transitions-2/Overview.bs
noamr: Removed

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs
```diff
@@ -264,6 +278,31 @@ Note: this event is fired when [=reveal document|revealing a document=].
...
+		partial interface Document {
+			ViewTransition startViewTransition(optional StartViewTransitionOptions options);
+		};
+	</xmp>
+
+### {{Document/startViewTransition(options)}} Method Steps ### {#ViewTransition-start-with-options}
+
+	<div algorithm="start-vt-with-options">
+		The [=method steps=] for <dfn method for=Document>startViewTransition(|options|)</dfn> are as follows:
+
```
vmpstr, line 298: Given the comment above, these steps should also handle both situations by deferring to L1's algorithms as appropriate

css-view-transitions-2/Overview.bs
```diff
@@ -205,7 +205,21 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+The <dfn id='active-view-transition-pseudo'>:active-view-transition(* | <<transition-types>>)</dfn> pseudo-class applies to the root element of the document, if it has a matching [=active view transition=].
+It has the following syntax definition:
+
+<pre class=prod>
+	:active-view-transition(<<vt-type-selector>>)
+	<<vt-type-selector>> = '*' | <<custom-ident>>#
+</pre>
+
+An ''::active-view-transition()'' [=selector=] matches the [=document element=] if its
+[=node document=] has an [=active view transition=] whose [=ViewTransition/active types=] is null or
```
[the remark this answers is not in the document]
noamr: How's the new revision?

css-view-transitions-2/Overview.bs
```diff
@@ -205,6 +206,23 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+
+<pre class=prod>
+	:active-view-transition(<<vt-type-selector>>)
+	<dfn>&lt;vt-type-selector></dfn> = '*' | <<custom-ident>>#
+</pre>
+
+An ''::active-view-transition()'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=] |viewTransition|, for which any of the following are true:
+
+* The <<vt-type-selector>> is ''*''
+* |viewTransition|'s [=ViewTransition/active types=] is null
```
vmpstr, line 224: Can you file a separate issue for this? We haven't explicitly discussed this, and my intuition is that it shouldn't match if type-selector is non-`*` and `active types` is null

noamr: You're right. I think this line should go, will open an issue.

css-view-transitions-2/Overview.bs
```diff
@@ -205,6 +206,22 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+
+<pre class=prod>
+	:active-view-transition(<<vt-type-selector>>)
+	<dfn>&lt;vt-type-selector></dfn> = '*' | <<custom-ident>>#
+</pre>
+
+An ''::active-view-transition()'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=] |viewTransition|, for which any of the following are true:
+
+* The <<vt-type-selector>> is ''*''
+* |viewTransition|'s [=ViewTransition/active types=] [=list/contains=] at least one of the <<custom-ident>> values of the <<vt-type-selector>>.
```
khushalsagar, line 224: nit: Add a non-normative note with a code example after this?

css-view-transitions-2/Overview.bs
```diff
@@ -264,6 +281,34 @@ Note: this event is fired when [=reveal document|revealing a document=].
...
+
+	<xmp class=idl>
+		dictionary StartViewTransitionOptions {
+			UpdateCallback? update = null;
+			sequence<DOMString>? types = null;
+		};
+
+		partial interface Document {
+
+			ViewTransition startViewTransition(optional (UpdateCallback or StartViewTransitionOptions)? callbackOrOptions = null);
```
khushalsagar, line 294: Do we need this IDL to say "callback or options"? L1 declares the API which only takes a callback : https://drafts.csswg.org/css-view-transitions-1/#additions-to-document-api. So if in L2 the IDL adds a new API which is

```
startViewTransition(StartViewTransitionOptions options = null);
```

Would that be equivalent?

css-view-transitions-2/Overview.bs
```diff
@@ -264,6 +281,34 @@ Note: this event is fired when [=reveal document|revealing a document=].
...
+	</xmp>
+
+### {{Document/startViewTransition(options)}} Method Steps ### {#ViewTransition-start-with-options}
+
+	<div algorithm="start-vt-with-options">
+		The [=method steps=] for <dfn method for=Document>startViewTransition(|callbackOrOptions|)</dfn> are as follows:
+
+		1. If |callbackOrOptions| is an {{UpdateCallback}}, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions| and return the result.
+
+		1. Let |viewTransition| be the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOrOptions|'s {{StartViewTransitionOptions/update}}.
```
khushalsagar, line 305: The above step says "run the method steps" while this one says, "Let |viewTransition| be the method steps for ...". Why the difference? I was expecting that we'd run the method steps unconditionally and = set the active types (if there is any) after that.

css-view-transitions-2/Overview.bs
```diff
@@ -264,6 +281,34 @@ Note: this event is fired when [=reveal document|revealing a document=].
...
+
+		1. Set |transition|'s [=ViewTransition/active types=] to |callbackOrOptions|'s {{StartViewTransitionOptions/types}}.
```
khushalsagar, line 307: These also need to be reset, likely when the finish promise is dispatched or sooner if we're skipping the transition.

I'm not sure how to integrate new functionality like this with algorithms which are in the L1 spec...

noamr: Why does that matter? That transition is no longer going to be active.

css-view-transitions-2/Overview.bs
noamr: You can't have overloads like this in WebIDL (or in javascript, for that matter). The new method shadows the previous one in a way, but we make sure it's compatible.
Adding `startViewTransition(StartViewTransitionOptions options = null)` would shadow the L1 function and not allow using a callback argument.

css-view-transitions-2/Overview.bs
noamr: It's equivalent to the following code:
```js
if (typeof callbackOrOptions === "function")
  return startViewTransitonInternal(callback);

const viewTransition = startViewTransitionInternal(callbackOrOptions.update);
viewTransition.activeTypes = callbackOrOptions.types;
return viewTransition;
```

Sure we could do something like:
```js
const update = typeof callbackOrOptions === "function" ? callbackOrOptions : callbackOrOptions.update;
const viewTransition = startViewTransitionInternal(update);
if (callbackOrOptions?.types)
  viewTransition = callbackOrOptions.types;
```

But I'm not sure it's an improvement.

css-view-transitions-2/Overview.bs
noamr: Done

css-view-transitions-2/Overview.bs
vmpstr: So, I was just playing around with the implementation and I think this is also not allowed due to the last point here:
https://webidl.spec.whatwg.org/#idl-nullable-type

> The [inner type](https://webidl.spec.whatwg.org/#dfn-inner-type) must not be:
> ...
> * a [union type](https://webidl.spec.whatwg.org/#dfn-union-type) that itself [includes a nullable type](https://webidl.spec.whatwg.org/#dfn-includes-a-nullable-type) or has a dictionary type as one of its [flattened member types](https://webidl.spec.whatwg.org/#dfn-flattened-union-member-types).

I don't quite understand the non-normative justification for this rule, but I also think that we can just make the union non-nullable?

vmpstr: Specifically something like this might work
```
ViewTransition startViewTransition((UpdateCallback or StartViewTransitionOptions?) callbackOrOptionsOrNull);
```

css-view-transitions-2/Overview.bs
khushalsagar: Oh yea, my bad.

css-view-transitions-2/Overview.bs
khushalsagar: Ack. This was probably related to the comment below. I incorrectly assumed the transition types are a global state, so we should be updating it if a second transition runs with a null types list. But since its stored on the transition itself, it doesn't matter.

khushalsagar (approved):
LGTM % vmpstr.

vmpstr (approved):
lgtm

PR #9387, opened 2023-09-20
[css-fonts-4] Define src with <font-src-list>

### LINKED ISSUE #7632: [css-fonts-4] Define `value` of `src` with a valid value definition
The `value` field of [`src`](https://drafts.csswg.org/css-fonts-4/#descdef-font-face-src) is defined with `see prose`. Would it be possible to use a valid value definition according the CSS value definition syntax?

I know that `src` must be parsed in a [specific way](https://drafts.csswg.org/css-fonts-4/#font-face-src-parsing), similarly to `<media-query-list>` and `<forgiving-selector-list>`. So I suggest to set its `value` with something like `<font-src-list>` that would reference this specific procedure, which should take advantage of [*parse a comma-separated list according to a CSS grammar*](https://drafts.csswg.org/css-syntax-3/#parse-comma-list) with `<font-src>` as the CSS grammar.

~~Also, I think it could be helpfull for CSS authors to remove invalid items from the serialization, like invalid media query or selector.~~ *edit:* invalid selectors in `<forgiving-selector-list>` and invalid/unknown media queries in `<media-query-list>` are now preserved - https://github.com/w3c/csswg-drafts/issues/8356#issuecomment-1642840607, https://github.com/w3c/csswg-drafts/issues/7595#issuecomment-1284268421.

cdoublev:
Fixes #7632.

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -2440,7 +2440,7 @@ The <dfn id="at-font-face-rule">''@font-face''</dfn> rule</h3>
 
 	<pre class='descdef'>
 	Name: src
-	Value: see prose
+	Value: <<font-src-list>>
 	For: @font-face
 	Initial: N/A
 	</pre>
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
Thanks!

Issue #7582, opened 2022-08-09
[css-transforms-2] Potentially invalid syntax for `perspective()`


pyoor:
The CSS Transforms 2 spec currently defines the `perspective()` function as:  
`perspective( <length [0,∞]> | none )`

In the syntax above, the functional notation accepts one of two arguments (using the single bar operator).  Now, I can't find anything specifically in css-values-5 that forbids this, but I believe that these arguments should be wrapped in a group as in:
`perspective( [ <length [0,∞]> | none ] )`

Could some please provide clarification on if this actually is an issue?

https://github.com/w3c/csswg-drafts/blob/77f40b2fa25dcb36dda6915c3eb2571b57405881/css-transforms-2/Overview.bs#L1003

tabatkins:
Yup, that's an error.

cdoublev:
Is it a syntax style error, or more? Where is this style defined?

- [CSS Overflow 4](https://drafts.csswg.org/css-overflow-4/#funcdef-text-overflow-fade) has `fade( <length> | <percentage> )`
- [Filter Effects 1](https://drafts.fxtf.org/filter-effects-1/#funcdef-filter-drop-shadow) has `drop-shadow( <color>? && <length>{2,3} )`

tabatkins:
The [Syntax extension of the CSS grammar](https://drafts.csswg.org/css-syntax/#rule-defs) defines how at-rules and functions (and other things) are written in the grammar. It doesn't have any special rules for functions (or anything else) to automatically create groups in the grammar.

Perhaps it should, tho? Hm.

cdoublev:
> It doesn't have any special rules for functions (or anything else) to automatically create groups in the grammar.

What are the cases (other than function arguments) where a group should be created automatically? Do these rules also explain why a group is created for the types specific to `font-variant-alternates` in the value definition of [`font-variant`](https://drafts.csswg.org/css-fonts-4/#font-variant-prop)?

I guess the reason for creating a group that is not strictly required, is to make a grammar parser produce a nested list with the corresponding value(s). From this perspective, it would have been nice to create groups for the types specific to `font-variant-ligatures`, `font-variant-numeric`, and `font-variant-east-asian`. It would make shorthand expansion easier, imo.

I'm sorry for this noise on a closed issue. Please tell me if I should create a new issue.

Related: #2921 

```
  normal 
| none 
| [
       # font-variant-ligatures
[...]

PR #9394, opened 2023-09-21
[css-transforms] Minor syntax rewrite

cdoublev:
- add [missing](https://github.com/w3c/csswg-drafts/issues/7582#issuecomment-1210838228) group around function arguments
- use multipliers instead of repeating symbols

### DIFF
--- a/css-transforms-1/Overview.bs
+++ b/css-transforms-1/Overview.bs
@@ -912,31 +912,31 @@ A percentage for vertical translations is relative to the height of the [=refere
 ----------------------
 
 <dl dfn-for=transform>
-	: <span class='prod'><dfn>matrix()</dfn> = matrix( <<number>>#{6} )</span>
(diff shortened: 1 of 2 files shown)


tabatkins:
You don't need a group around a function's arguments. In this particular case it's becuase there's nothing to group in the first place, but more generally [function syntax implicitly groups its arguments](https://drafts.csswg.org/css-values/#component-functions). (This is a relatively recent addition to the grammar, reflecting what a lot of people were accidentally doing anyway.)

I also think it's fine to repeat arguments a little, so long as it's not a clarity issue. It's easier to read `<length> , <length>` than `<length>{1,2}`, in my opinion. So either way is correct, but none of the examples here look like they'd particularly *benefit* from changing.

cdoublev:
I missed this change, sorry. Actually it is `<length>#{1,2}`, which requires to know the meaning of `#`,  instead of `<length> , <length>?`, which requires to know about the comma-elision rule, but I admit the latter is still easier to read. Thanks for the feedback.

PR #9414, opened 2023-09-26
[cssom] Add a “disable a CSS style sheet” definition

sideshowbarker:
See also https://github.com/whatwg/html/pull/9798

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1197,6 +1197,19 @@ To <dfn export>remove a CSS style sheet</dfn>, run these steps:
  <!-- XXX does anything need to happen wrt alternate style sheets? what if the last style sheet with the preferred style sheet set name is removed? -->
 </ol>
 
+To <dfn export>disable a CSS style sheet</dfn>, run these steps:
+
+<ol>
+ <li>Remove the <a>CSS style sheet</a> from the list of <a>document or shadow root CSS style sheets</a>.
+
+ <li>Set the <a>CSS style sheet</a>'s <a for=CSSStyleSheet>parent CSS style sheet</a>,
+ <a for=CSSStyleSheet>owner node</a> and <a for=CSSStyleSheet>owner CSS rule</a> to null.
+
+ <!-- "associated CSS style sheet" is defined in terms of owner node, so we don't need to set it to null explicitly -->
+
+ <li>Set the <a>disabled flag</a>.
+</ol>
+
 A <dfn export>persistent CSS style sheet</dfn> is a
 <a>CSS style sheet</a> from the <a>document or shadow root CSS style sheets</a>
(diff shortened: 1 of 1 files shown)


zcorpan:
How is it different from [removing](https://drafts.csswg.org/cssom/#remove-a-css-style-sheet ) (and then creating a new sheet when enabling again)?

In WebKit it seems the re-enabled sheet is the same object, but in Chrome and Firefox it's a new object. Demo for http://software.hixie.ch/utilities/js/live-dom-viewer/ : 
```html
<!DOCTYPE html>
<link rel=stylesheet href="data:text/css,html{background:yellow} " id=x> 
<script>
onload = () => {
var sheet = x.sheet;
w(document.styleSheets.length)
x.disabled = true;
w(sheet.ownerNode)
w(document.styleSheets.length)
x.disabled = false;
w(sheet === x.sheet);
}
</script>
```

cssom-1/Overview.bs
```diff
@@ -1197,6 +1197,19 @@ To <dfn export>remove a CSS style sheet</dfn>, run these steps:
  <!-- XXX does anything need to happen wrt alternate style sheets? what if the last style sheet with the preferred style sheet set name is removed? -->
 </ol>
 
+To <dfn export>disable a CSS style sheet</dfn>, run these steps:
+
+<ol>
+ <li>Remove the <a>CSS style sheet</a> from the list of <a>document or shadow root CSS style sheets</a>.
+
+ <li>Set the <a>CSS style sheet</a>'s <a for=CSSStyleSheet>parent CSS style sheet</a>,
+ <a for=CSSStyleSheet>owner node</a> and <a for=CSSStyleSheet>owner CSS rule</a> to null.
```
emilio, line 1206: How can a top level sheet have an owner CSS rule? I think that's not necessary.

cssom-1/Overview.bs
```diff
@@ -1197,6 +1197,19 @@ To <dfn export>remove a CSS style sheet</dfn>, run these steps:
...
+
+ <!-- "associated CSS style sheet" is defined in terms of owner node, so we don't need to set it to null explicitly -->
+
+ <li>Set the <a>disabled flag</a>.
```
emilio, line 1210: Probably `to true`?

emilio (changes requested):
So this doesn't match how browsers behave with `<link rel=stylesheet disabled>` right? In particular that doesn't mess with the `disabled` flag at all.

Maybe call this "remove a CSS style sheet" or so?

sideshowbarker:
Yeah, I’m going ahead and closing 

> How is it different from [removing](https://drafts.csswg.org/cssom/#remove-a-css-style-sheet) (and then creating a new sheet when enabling again)?

I can see now that it’s not different from removing, so I’m going ahead and closing this PR.

PR #9426, opened 2023-09-29
[css-view-transitions-2] Use expected phrasing and markup for event firing

dontcallmedom:


noamr (commented):
This is going to be upstreamed to HTML soon and will be deleted from here... Also I wouldn't add a `dfn` in that way and it's being renamed...

tidoust:
Event definition and firing now upstreamed to HTML. Taking the liberty to close as no longer applicable.

PR #9439, opened 2023-10-04
[css-color-5] Define light-dark().

### LINKED ISSUE #7561: [css-color] Add a function to allow authors to specify colors reacting to `color-scheme`
Some UIs require rendering a light button in some places, but dark in another. It'd be nice if you could switch between these by using `color-scheme`, like you can with native widgets. However, right now there's no way to do this other than exclusively using system colors.

Chromium has `-internal-light-dark(<light-color>, <dark-color>)` for this. However `color-scheme` was designed a bit more generically than `light` / `dark`, so maybe we need a `schemed-color(<custom-ident> <color>, <fallback-color>)` or so, where the `<custom-ident>` is the scheme name, and the `<color>` is the color for that scheme. So `-internal-light-dark(red, blue)` would become `schemed-color(light red, blue)` or so.

This would allow authors to create widgets that properly react to `color-scheme`.

emilio:
Closes #7561

### DIFF
--- a/css-color-5/Overview.bs
+++ b/css-color-5/Overview.bs
@@ -34,6 +34,8 @@ spec:css-color-4; type:value; text:srgb
 spec:css-color-4; type:dfn; text:modern color syntax
 spec:css-color-4; type:value; text:alpha-value
 spec:css-values-5; type:value; text:color
+spec:css-color-adjust-1; type:value; text:light
+spec:css-color-adjust-1; type:value; text:dark
 </pre>
 <link rel="stylesheet" href="style.css" />
 <style>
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
Thanks!

css-color-5/Overview.bs
```diff
@@ -2468,6 +2470,22 @@ or any other color or monochrome output device which has been characterized.
 			<li>alpha is the same as the input color
 		</ul>
 
+<h2 id="light-dark">
+	Reacting to the used color-scheme: the ''light-dark()'' Function
+</h2>
+
+	System colors have the ability to react to the current used ''color-scheme'' value.
```
SebastianZ, line 2477: This should be "currently" instead of "current".

Sebastian

Issue #8207, opened 2022-12-08
[css-grid-3] Masonry - align-tracks / justify-tracks potentially not desirable for accessibility.


bfgeek:
This feature of css-grid-3 allows developers to align tracks individually. However this creates a disconnect pretty quickly with the "visual order" and the "box order" of the document.

Example 9 in the spec demonstrates this well:
https://drafts.csswg.org/css-grid-3/#masonry-axis-alignment

The tab order for above would be 1,2,3,4... etc. However the visual order for that example definitely not that :).

The use-cases I've seen here typically align masonry axis tracks together rather than individually.

Something to lean on here might be something similar to how flexbox works in that you can start the masonry algorithm for the "end" side, instead of the "start" (e.g. row-reverse and friends).

Note this problem is different to alignment on individual grid/flex items, as for those layout modes alignment happens within a line/track, and the visual order still matches the box order.

PR #9444, opened 2023-10-06
[css-grid-3] Remove align-tracks and justify-tracks

lukewarlow:
Based on https://github.com/w3c/csswg-drafts/issues/8207#issuecomment-1747805578


fantasai (changes requested):
I think we need to keep the paragraph about baseline alignment in the grid axis, and say something about baseline alignment in the masonry axis...

lukewarlow:
I believe this change has already been made to the spec so will close this for now.

Issue #9473, opened 2023-10-16
[css-syntax-3] The `+` in `<an+b>` confuses parsing tools


tidoust:
The [`<an+b>` type](https://drafts.csswg.org/css-syntax-3/#the-anb-type) is defined with a production rule that starts with `<an+b> = ...`. This type is now referenced from the definition of [`:nth()`](https://drafts.csswg.org/css-gcpm/#selectordef-nth) in the CSS Generated Content for Paged Media Module.

Problem is that the `+` confuses grammar parsers, including [CSSTree](https://github.com/csstree/csstree/issues/263) and the highlighting mechanism in Bikeshed (which creates a link to the definition of `+`), as they parse `+` as a [component value multiplier](https://drafts.csswg.org/css-values-3/#component-multipliers).

The [Value Syntax Definition](https://drafts.csswg.org/css-values-3/#component-types) is not entirely clear about what characters can be used to define basic data types between `<` and `>` but it says:
> Other non-keyword literal characters that may appear in a component value, such as “+”, must be written enclosed in single quotes.

I think that's the only occurrence in CSS specs of a construct that parsers find problematic. Tools could be fixed of course, I'm wondering whether it would be ok to defined the type as `<an-plus-b>` instead.

(If so, I'm happy to prepare a PR to update the specs that define or reference the type)

cdoublev:
I think those tools should be able to parse any [`<ident>`](https://drafts.csswg.org/css-syntax-3/#ref-for-typedef-ident-token%E2%91%A0) between `<` and `>`, like in the [`syntax` descriptor](https://drafts.css-houdini.org/css-properties-values-api-1/#the-syntax-descriptor).

https://drafts.css-houdini.org/css-properties-values-api-1/#consume-a-data-type-name

tabatkins:
I think it's fine to just rename it to `<an-b>`, honestly. The plus isn't necessary.

cdoublev:
I do not see any mention in CSS GCPM 4: are you are thinking to `:nth-of-page(n)`? I am not sure if the spec authors expect `<integer>` or `<an-b>`. There are mentions in CSS GCPM 3 and CSS Overflow 4 though.

This PR might also be an opportunity to replace *"the An+B notation"* with *"the `<an-b>` type"*, and `An+B` with `<an-b>` in value definitions, to fix #9494.

fantasai:
[...]

PR #9480, opened 2023-10-17
[css-syntax-3] Rename `<an+b>` to `<an-b>`

tidoust:
See context in #9473. Rename is aimed at avoiding interpretation of `+` as a component value multiplier by parsing tools. The ID of the definition does not change.

The rename is also done in the definition of the `serialize an <an+b> value` algorithm, preserving the previous ID since the algorithm was exported (and referenced by other specs).

Once construct has been renamed, the following specs will need to be slightly updated as well (they either reference `<an+b>` or `serialize an <an+b> value`):
- cssom-1
- css-gcpm-4
- selectors-4

I'll prepare a follow-up PR for these updates once the cross-references database gets the new definitions.

### DIFF
--- a/css-syntax-3/Overview.bs
+++ b/css-syntax-3/Overview.bs
@@ -3271,20 +3271,20 @@ Informal Syntax Description</h3>
 
 
 <h3 id="the-anb-type">
-The <code>&lt;an+b></code> type</h3>
+The <<an-b>> type</h3>
 
 	The <var>An+B</var> notation was originally defined using a slightly different tokenizer than the rest of CSS,
 	resulting in a somewhat odd definition when expressed in terms of CSS tokens.
 	This section describes how to recognize the <var>An+B</var> notation in terms of CSS tokens
-	(thus defining the <var>&lt;an+b></var> type for CSS grammar purposes),
+	(thus defining the <<an-b>> type for CSS grammar purposes),
 	and how to interpret the CSS tokens to obtain values for <var>A</var> and <var>B</var>.
 
-	The <var>&lt;an+b></var> type is defined
+	The <<an-b>> type is defined
 	(using the <a href="https://www.w3.org/TR/css3-values/#value-defs">Value Definition Syntax in the Values &amp; Units spec</a>)
 	as:
 
 	<pre class='prod'>
-		<dfn id="anb-production">&lt;an+b></dfn> =
+		<dfn id="anb-production">&lt;an-b></dfn> =
 		  odd | even |
 		  <var>&lt;integer></var> |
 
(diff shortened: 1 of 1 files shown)


cdoublev:
I do not see any mention in CSS GCPM 4: are you are thinking to `:nth-of-page(n)`? I am not sure if the spec authors expect `<integer>` or `<an-b>`. There are mentions in CSS GCPM 3 and CSS Overflow 4 though.

This PR might also be an opportunity to replace *"the An+B notation"* with *"the `<an-b>` type"*, and `An+B` with `<an-b>` in value definitions, to fix #9494.

tidoust:
> I do not see any mention in CSS GCPM 4: are you are thinking to `:nth-of-page(n)`? I am not sure if the spec authors expect `<integer>` or `<an-b>`. There are mentions in CSS GCPM 3 and CSS Overflow 4 though.

Sorry, `css-gcpm-4` is a typo, it should have been `css-gcpm-3`.

I didn't include CSS Overflow 4 because, while it mentions `an+b`, it does not reference the `<an+b>` type, and the mentions are only there to alert about open issues in any case.

> This PR might also be an opportunity to replace _"the An+B notation"_ with _"the `<an-b>` type"_, and `An+B` with `<an-b>` in value definitions, to fix #9494.

I leave that up to spec editors, but how would that change help with #9494, which is more about updating Selectors 4 to use more formal definitions, and reference `<an-b>`, for pseudo-class selectors?

cdoublev:
> I didn't include CSS Overflow 4 because, while it mentions an+b, it does not reference the `<an+b>` type [...]

Yeah, I realized it afterwards.

<details>
  <summary>I am also fine with making changes for <code>css-gcpm-4</code> and <code>selectors-4</code> in another PR.</summary>

Selectors 4

  - `:nth-child(An+B [of S]?)` should be `:nth-child(<an-b> [of <complex-real-selector-list>]?)`
  - `:nth-last-child(An+B [of S]?)` should be `:nth-last-child(<an-b> [of <complex-real-selector-list>]?)`
  - `:nth-of-type(An+B)` should be `:nth-of-type(<an-b>)`
  - `:nth-last-of-type(An+B)` should be `:nth-of-type(<an-b>)`
  - `:nth-col(An+B)` should be `:nth-col(<an-b>)`
  - `:nth-last-col(An+B)` should be `:nth-last-col(<an-b>)`

CSS GCPM 4

  - `:nth-of-page(n)` should be `:nth-of-page(<an-b>)` (to be confirmed)
</details>

tabatkins:
Yes, go ahead and change the selectors references.

Ignore GCPM 4, I'm not sure what's intended (and the spec isn't maintained anyway).

fantasai:
Marking as a draft pending resolution of https://github.com/w3c/csswg-drafts/issues/9473#issuecomment-1819819647

tidoust:
Taking the liberty to close the PR as no longer applicable. The underlying problem is fixed. The type was updated in `css-syntax-3` with a different name [`<a-n-plus-b>`](https://drafts.csswg.org/css-syntax/#the-anb-type).

Issue #9501, opened 2023-10-20
[css-text] The definition of ideographs includes punctuation marks


xfq:
https://drafts.csswg.org/css-text-4/#ideographs

The current definition of [ideographs](https://drafts.csswg.org/css-text-4/#ideographs) includes two Chinese punctuation marks, U+16FE2 and U+16FE3 (see the [UnicodeSet](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5B%5B%5Cu3041-%5Cu30FF%5D%26%5B%3Agc%21%3D%2F%5EP%2F%3A%5D%5D%5B%5Cu31C0-%5Cu31FF%5D%5B%3Asc%3DHan%3A%5D%5D&g=sc&i=gc)), but no other punctuation marks. These characters are not used in modern Chinese, but they are used by scholars for the textual processing, electronic interchange, and publication of ancient Chinese texts.

As an example, if the end of a section/clause/sentence is Western text, then the spacing between the Western text and the U+16FE2	punctuation will become larger per [text-autospace](https://drafts.csswg.org/css-text-4/#text-autospace-property), which is not the result users want to see.

There should be some examples of classical Chinese mixed with Western texts in the early days of Republic China, but it will take some time to find specific examples.

fantasai:
So probably the definition of "ideographs" should include only Letters, Numbers, Symbols, and Marks (excluding Punctuation, Spaces, and Control characters). Does that sound right?

frivoal:
@xfq https://github.com/w3c/csswg-drafts/pull/9503 tries to address this (and a some other things), can you have a look?

Clqsin45:
A similar case might be using Bopomofo directly in text ( non-annotation usage).

I think according to https://www.w3.org/International/clreq/#chinese_and_western_mixed_text_composition, no spacing should be inserted between Bopomofo and Chinese characters, as Bopomofo should be non-western characters, but the current version is asking to do so, and it should be addressed by the PR.(?)

frivoal:
@Clqsin45 I've posted an answer to your point above into https://github.com/w3c/csswg-drafts/issues/9471 This is a related topic, but I felt it might be better to keep this issue about punctuation, and that issue about classes of letters we might have put in the wrong category

Issue #9471, opened 2023-10-16
[css-text][text-autospace] Is halfwidth Kana "non-ideographic letters"?


kojiishi:
From the [8.5.2. Text Spacing Character Classes](https://drafts.csswg.org/css-text-4/#text-spacing-classes), halfwidth Kana belongs to "non-ideographic letters". This means that there will be an auto-space between ideographs and halfwidth Kana. This seems like an overlook.

The question is whether they should be "ideograph" or not. Terminology-wise, Kana is derived from Han, so it's closer to ideograph than to non-ideograph. Behavior-wise, I don't think authors expect auto-space between ideograph/fullwidth Kana and halfwidth Kana.

So I think we should classify them not to belong to either "ideographs" nor "non-ideographic letters". I'll check with JLTF about this separately.

kojiishi:
[Japanese discussion here](https://lists.w3.org/Archives/Public/public-i18n-japanese/2023OctDec/0160.html)

kojiishi:
3 opinions in the Japanese discussion, all agreed:
* No auto-space between halfwidth Kana and "ideograph".
* No auto-space between halfwidth Kana and "non-ideographic letters/numerals".

frivoal:
@kojiishi  https://github.com/w3c/csswg-drafts/pull/9503 tries to address this (and a some other things), can you have a look?

nt1m:
cc @fantasai @vitorroriz for WebKit input.

frivoal:
As pointed out in https://github.com/w3c/csswg-drafts/issues/9501#issuecomment-1779704918, we might have a related question about Bopomofo. By the way, how about Hangul? Maybe it should be treated similarly to Bopomofo?
Should these be made to fit in the `ideographs` category, or `non-ideographic letters`, or neither like half-width katakana?

What about other east asian scripts, such as:
* https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:Script_Extensions=Khitan_Small_Script:]
* https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:Script=Tangut:]
* https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:Script_Extensions=Nushu:]

All these have East_Asian_Width set to 'Wide', but not 'FullWidth'

kojiishi:
> how about Hangul?

[...]

PR #9503, opened 2023-10-20
[css-text-4] Refine definition of ideographs

frivoal:
See #9501 and #9471

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -10720,14 +10720,11 @@ Text Spacing Character Classes</h4>
 
 	<dl>
 		<dt><dfn>ideographs</dfn>
-		<dd>Includes all [=typographic character units=] [[CSS-TEXT-3]] whose base character is listed below:
-			<ul>
-				<li>All characters in the range of U+3041 to U+30FF,
-					except those that belong to Unicode Punctuation [P*] [=general category=].
-				<li>CJK Strokes (U+31C0 to U+31EF).
-				<li>Katakana Phonetic Extensions (U+31F0 to U+31FF).
-				<li>All characters that have the Han [=script property=].
-			</ul>
+		<dd>Includes all [=typographic character units=] [[CSS-TEXT-3]]
+		whose base character:
+		* belongs to Unicode Letters [L*], Mark [M*], Symbols [S*], or Numbers [N*] [=general category=], and
+		* has the Han, Hiragana, or Katakana [=script property=], and
+		* is not categorized as East Asian Halfwidth (H) by [[!UAX11]]
 
 		<dt><dfn>non-ideographic letters</dfn>
 		<dd>
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -10664,13 +10664,12 @@ Inter-script Spacing</h4>
 
 	<dl>
 		<dt><dfn>ideographs</dfn>
-		<dd>Includes all [=typographic character units=] [[CSS-TEXT-3]] whose base character is listed below:
+		<dd>Includes all [=typographic character units=] [[CSS-TEXT-3]]
+		whose base character:
 			<ul>
-				<li>All characters in the range of U+3041 to U+30FF,
-					except those that belong to Unicode Punctuation [P*] [=general category=].
-				<li>CJK Strokes (U+31C0 to U+31EF).
-				<li>Katakana Phonetic Extensions (U+31F0 to U+31FF).
-				<li>All characters that have the Han [=extended script property=].
+				<li>belongs to Unicode Letters [L*], Mark [M*], Symbols [S*], or Numbers [N*] [=general category=], and
+				<li>has the Han, Hiragana, or Katakana [=extended script property=], and
+				<li>is not categorized as East Asian Halfwidth (H) by [[!UAX11]]
 			</ul>
 
 		<dt><dfn>non-ideographic letters</dfn>


css-text-4/Overview.bs
```diff
@@ -10720,14 +10720,11 @@ Text Spacing Character Classes</h4>
...
-				<li>All characters in the range of U+3041 to U+30FF,
-					except those that belong to Unicode Punctuation [P*] [=general category=].
-				<li>CJK Strokes (U+31C0 to U+31EF).
-				<li>Katakana Phonetic Extensions (U+31F0 to U+31FF).
-				<li>All characters that have the Han [=script property=].
-			</ul>
+		<dd>Includes all [=typographic character units=] [[CSS-TEXT-3]]
+		whose base character:
+		* belongs to Unicode Letters [L*], Mark [M*], Symbols [S*], or Numbers [N*] [=general category=], and
+		* has the Han, Hiragana, or Katakana [=script property=], and
```
kojiishi, line 10726: * [Characters removed by this definition](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5Cu3041-%5Cu30FF%5D%5B%5Cu3041-%5Cu30FF%5D-%5B%3Agc%3DP%3A%5D-%5B%3Asc%3DHira%3A%5D-%5B%3Asc%3DKana%3A%5D-%5B%3Asc%3DHrkt%3A%5D&g=&i=gc%2Csc%2Cea)
* [Characters added by this definition](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Asc%3DHira%3A%5D%5B%3Asc%3DKana%3A%5D%5B%3Asc%3DHrkt%3A%5D-%5B%3Aea%3DH%3A%5D-%5B%5Cu3041-%5Cu30FF%5D-%5B%5Cu3041-%5Cu30FF%5D&g=&i=gc%2Csc%2Cea)

From these two lists, I think:
* Adding Kana extensions, Hentaigana, etc. is good.
* U+30FC KATAKANA-HIRAGANA PROLONGED SOUND MARK should not be removed.
* Should Circled Katakana probably be removed?
* Squared Katakana words, I'm not sure. Ask at JLTF?

css-text-4/Overview.bs
```diff
@@ -10720,14 +10720,11 @@ Text Spacing Character Classes</h4>
...
+		* is not categorized as East Asian Halfwidth (H) by [[!UAX11]]
```
kojiishi, line 10727: This looks good. I think we also need to exclude EAW=H from "non-ideographic letters/numerics" too, right?

css-text-4/Overview.bs
frivoal: > U+30FC KATAKANA-HIRAGANA PROLONGED SOUND MARK should not be removed.

In css-text-4, we have this:
> [script property](http://www.unicode.org/reports/tr24/#Values)
    Defined in Unicode Standard Annex #24 [[UAX24]](https://drafts.csswg.org/css-text-4/#biblio-uax24) and given as the Script property in the [Unicode Character Database](https://www.unicode.org/reports/tr44/) [[UAX44]](https://drafts.csswg.org/css-text-4/#biblio-uax44). (**UAs must include any ScriptExtensions.txt assignments in this mapping**.)

Since U+30FC has the ScriptExtensions property set to Hiragana and Katakana, it is not be removed.

But I think that our defintion of "script property" being not just the Unicode script property but also the Script Extensions property is too easy to miss. Maybe, editorially, we should rename it to "script properties" or "script related properties".

> Should Circled Katakana probably be removed?

No strong opinion, from me. Why do you think they should be removed?

css-text-4/Overview.bs
frivoal: seems reasonable, yes.

css-text-4/Overview.bs
kidayasuo: I agree with @kojiishi that the character U+30FC KATAKANA-HIRAGANA PROLONGED SOUND MARK should not be removed. It is cl-10 in JLReq and its behavior for spacing is the same as Hiragana, Katakana and Kanji.

I believe circled Katakana should be treated the same as other circled letters, however it is currently a part of Kanji cl-19. We'll discuss this at JLReq TF.

Squared Katakana words is postfixed abbreviations cl-13 in JLReq. It is a part of symbols in the [new simplified proposal](https://github.com/w3c/jlreq/blob/gh-pages/docs/spacing_property/spacing_property.md). We'll revisit this in JLReq TF also.

frivoal: > the character U+30FC KATAKANA-HIRAGANA PROLONGED SOUND MARK should not be removed

It is not being removed, so we're in agreement.

kojiishi: > Since U+30FC has the ScriptExtensions property set to Hiragana and Katakana, it is not be removed.

The ScriptExtensions property is a different property from the Script property. If you want to use that, it's better to be explicit. [UAX#24 3.2](https://unicode.org/reports/tr24/#Assignment_ScriptX_Values) says it has a list only when the list is "a small, enumerable list," so it's a "be careful when you must rely on it" type of the property.

From implementers' point of view, it's a lot more expensive performance-wise. It's (probably) the only property that return a list. Great if we can avoid that when possible.

For this specific case, if the [Unicode Blocks](https://www.unicode.org/Public/UNIDATA/Blocks.txt) works, that's helpful. Is that ok?

css-text-4/Overview.bs
kidayasuo: I compared the proposed definition with "[spacing property](https://github.com/w3c/jlreq/blob/gh-pages/docs/spacing_property/spacing_property.md)" which is under development for jlreq-d. Differences are shown below:

Characters that are in css 'ideograph' but excluded from 'J' class in spacing_property
1. ㋐ CIRCLED KATAKANA (sc=Katakana, gc=So)
2. ㌁ SQUARE KATAKANA and HIRAGANA (sc=Katakana/Hiragana, gc=So)
3. Han radicals (sc=Hani, gc=So)
4. U+16FF0 VIETNAMESE ALTERNATE READING MARK CA and U+16FF1 VIETNAMESE ALTERNATE READING MARK NHAY (gc=Mc; sc=Hani; ea=W)
5. U+16FE3 OLD CHINESE ITERATION MARK (gc=Lm; sc=Hani; ea=W)

Characters that are excluded from css 'ideograph' but in 'J' class in spacing_property
6. 〆、〼 (gc=Lo; sc=Common; ea=W)
7. U+30FC KATAKANA-HIRAGANA PROLONGED SOUND MARK,  U-3031-3034 VERTICAL KANA REPEAT MARKs  (gc=Lm; sc=Common; ea=W)
8. Idegrahpic characters other than Han Ideograph (Ideographic=Yes, sc=Kits/Nshu/Tang)

Here are comments on each of these based on the discussions done when we developed spacing_property.

1. We believe circled characters should be removed from CSS 'Ideograph'. They were a part of cl-19 (Kanji) in JLReq, but it was agreed to treat all circled, double circled, negative circled and square letters as symbols that do not generate aki space with any other characters (class 'O' in the spacing_property). The reasoning is that they are enclosed symbols more than they are numbers, katanaka, kanji or alphabet (e.g, ①Ⓐ㉄㉮㊀).
2. We believe squared Katakana / Hiragana should be removed from CSS 'Ideograph' for the same reason as 1.
3. We believe radicals should stay in 'Ideograph', i.e as-is. We want to update spacing_property to match it.
4. We do not have enough knowledge to tell if it should be or should not be a part of 'Ideograph'. Since they are used in combination with base Kanji, they do not need to be included perhaps?
5. We believe this character should stay in 'Ideograph', i.e as-is. We want to update spacing_property to match it.
6. Probably 〆〼 should be added to 'Ideograph'. We do not have a strong opinion on them.
7. As agreed, the PROLONGED SOUND MARK should be added. The same goes to U-3031-3034 VERTICAL KANA REPEAT MARKs.
8. Probably all ideographic characters should be added to 'Ideograph' because they all originated from Han ideograph. At the same time we do not have a strong opinion on them.

 I might provide updates as I just asked JLReq TF if they have further comments.

kidayasuo: Regarding excluding East Asian Halfwidth (H), I believe we can just say "and is categorised as East Asian Wide (W)" because all characters in this category is "ea=W".  What do you think?

kidayasuo: If we are to update [class "J" of the spacing_property](https://github.com/w3c/jlreq/blob/gh-pages/docs/spacing_property/spacing_property.md#j-アルファベットとの間に空間を必要とする文字--仮名や漢字) with above recommendations incorporated, it would look like:
[[[[:sc=Hiragana:][:sc=Katakana:][:sc=Common:]][:ideographic:]]&[:gc=L:]&[:ea=W:]]
[[[:sc=Hani:]]&[[:gc=L:][:gc=Nl:][:gc=So:]]]

kojiishi: @kidayasuo Your comments look great, but it also looks like the ideal list is more complicated than I would like to add to a CSS spec. How about doing this in 2 steps:
1. We go with some simple class for now, so that browsers can ship earlier.
2. Someone voluteers to add the property to Unicode, so that the future spec can switch to it.

For the purpose of 1, I'm fine with the current definition, or add/modify some more, but I wish to keep it simple.

Another possibility for 1: @nt1m @vitorroriz Is it possible for you to provide the list used by iOS/macOS? If it's simple enough, it might be a good candidate for the purpopse of 1 above.

@kidayasuo Are you or someone at JLTF willing to volunteer talking to Unicode?

> Regarding excluding East Asian Halfwidth (H), I believe we can just say "and is categorised as East Asian Wide (W)" because all characters in this category is "ea=W". What do you think?

Whether "ea=W" includes "ea=F" or not is a bit ambiguous. The [UAX#11](https://unicode.org/reports/tr11/) reads to me it does, but ICU doesn't. I think it's safer to be explicit.

/cc @Clqsin45

kidayasuo: > * We go with some simple class for now, so that browsers can ship earlier.

How simpler should it be?

For most practical uses I think the proposed changes + excluding ones below would be sufficient. 

> ㋐ CIRCLED KATAKANA (sc=Katakana, gc=So)
> ㌁ SQUARE KATAKANA and HIRAGANA (sc=Katakana/Hiragana, gc=So)

nt1m: cc @fantasai

kojiishi: @kidayasuo and I will work on a proposal to Unicode. This will be 2 steps; one to reach a consensus to define a new property, then on the data. Hopefully the first step shouldn't be too hard.

css-text-4/Overview.bs
fantasai: @kojiishi That's what we did originally, but when Unicode updated they added letters outside the original blocks. So from the spec point of view, it's better to use the script properties. But from an implementation point of view, you probably want to implement it as unicode blocks.

kojiishi: As @xfq reported somewhere, the proposal to the Unicode is in progress. I'd like to see how it goes.

I agree that the maintenance when Unicode updates is a concern for us. I think we're in consensus to try to avoid/minimize it, only slightly different opinions on how.

Allow me to confirm, here we're talking about the script extension property, not the script property, correct? I'm assuming you meant so when you wrote "the script properties" but please let me know if I read you wrong.

First, let me correct my previous comment, it wasn't correct. It was based on the past feedback from Myles and Jonathan. But after talking to UTC experts, they preferred script extension over blocks. This is part because while the script extension property isn't easy for browsers to use in a performant way, it's easier for ICU. If we were going to implement this in ICU, it's possible for ICU to support CSS-defined derived properties, as they did for the line break iterators, but Unicode is more natural place.

Another point. CSS defining simple derived properties, combination of only one or two properties, is fine, but when it goes beyond certain complexity, the risk of overfitting increases, and thus the possibility of maintenance to adopt new Unicode versions will increase, so we may not be achieving our original goal. IIRC this was the primary reason for UAX#50, feedback from both CSS (John) and Unicode (Eric).

Seeing @frivoal's improved proposal and other feedback, I think this properly is likely to exceed the limit. If UTC can accept the proposal, I think it's the best way to go.

css-text-4/Overview.bs
```diff
@@ -11049,11 +10736,6 @@ Text Spacing Character Classes</h4>
 				<li>is categorized as East Asian Fullwidth (F) by [[!UAX11]].
 				<li>is upright in vertical text flow using the 'text-orientation' property
 					or the 'text-combine-upright' property.
```
tats-u, line 10738: #9979 #9980

```html
				<li>has the Hangul [=script property=].</li>
```

or

```html
				<li>has the Hangul or Bopomofo [=script property=].</li>
```

frivoal:
While we should still pursue https://github.com/w3c/csswg-drafts/issues/11013 and basing all of this on top of UTR59 eventually, for now it hasn't happened yet, so I have rebased this PR on top of the latest draft, in particular in order to benefit from https://github.com/w3c/csswg-drafts/commit/961075636a7da7b46c3e6faf7d33bcec10d3b48f.

Not saying this PR is ready, but at least this should reduce confusion, and maybe it can serve as a comparison point.

Issue #7442, opened 2022-07-01
[css-scroll-snap-2] snapChanged and snapChanging events are confusing and incomplete


dontcallmedom:
https://drafts.csswg.org/css-scroll-snap-2/#snap-events seems two describe  two new events, `snapChanged` and `snapChanging`.

Confusingly, it describes them as "event handler" with an event type of `scroll`, but I'm not sure what this is supposed to mean.

This misses an IDL fragment that describes `on…` event handler (presumably on the `HTMLElement` interface, but maybe `GlobalEventHandlers` if the events are supposed to bubble).

The vast majority of event names are purely lowercase, so the event should probably be `snapchanged` and `snapchanging`.

bakura10:
Adding on that, JS seems to always have used present tense (input events: change, blur), page events (pagehide), window events (resize)... would it be more sense to replace "snapchange" by "snapchanged" ? I agree snapchanged better reflect what it happens, but for consistency reason I think standard events should follow standard casing _and_ naming logic.

zcorpan:
The table seems to use the same structure as in https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements%2C-document-objects%2C-and-window-objects but confusingly defines "snapChanged" and "snapChanging" as an event handlers for `scroll` events (on which objects?). i.e.: 

```
obj.snapChanged = e => console.assert(e.type === "scroll")
```

It looks like implementation has started in Chromium, so addressing the naming at least would be good.
https://bugs.chromium.org/p/chromium/issues/detail?id=1456339
https://bugs.chromium.org/p/chromium/issues/detail?id=1494892

zcorpan:
The implemented `SnapEvent` interface is also not specified.

flackr:
@DavMila is working on improving the spec for these on https://github.com/w3c/csswg-drafts/pull/9515

zcorpan:
Why are new events needed, vs adding the snap information to the existing `scroll` and `scrollend` events?

DavMila:
For snapchanging, I'd say I think it's better as a separate event form `scroll` so a developer doesn't have to have an event listener run on every scroll event  just to know whether something snap-related is happening.
For snapchanged, I think modifying scrollend would work except that snapchanged could be invoked by a layout change as well.
(I acknowledge that spec still doesn't say much about these events.)

zcorpan:
OK, thanks.

[...]

PR #9515, opened 2023-10-23
[css-scroll-snap-2] Add SnapEvent definition

DavMila:
SnapEvent defines the interface of scroll-snap-related JavaScript events, snapchanged and snapchanging.

### DIFF
--- a/css-scroll-snap-2/Overview.bs
+++ b/css-scroll-snap-2/Overview.bs
@@ -474,14 +474,6 @@ Snap Events {#snap-events}
 	1. Run the steps to <a>update snapchanged targets</a> for |snapcontainer|.
 
 
-	<h4>Dispatching Pending Snap Events</h4>
-	When asked to <dfn export>dispatch pending snap events</dfn> for a {{Document}},
-	|doc|, run these steps:
-
-	1. Run the steps to <a>dispatch pending snapchanging events</a> for |doc|.
-	1. Run the steps to <a>dispatch pending snapchanged events</a> for |doc|.
-
-
 SnapEvent interface
 -------------------
 
--- a/css-scroll-snap-2/Overview.bs
+++ b/css-scroll-snap-2/Overview.bs
@@ -322,7 +322,7 @@ Snap Events {#snap-events}
 	</li>
 	<li> if, after a [[css-scroll-snap-1#re-snap|layout change]], the element to
 		which a snap container is snapped to changes, regardless of whether there is
-		a change in scroll position	after the layout change.
+		a change in scroll position after the layout change.
 	</li>
 	</ol>
 
@@ -335,9 +335,10 @@ Snap Events {#snap-events}
 
 	Each
 	<a spec=css-scroll-snap lt="scroll snap container">snap container</a> has
-	one <dfn export>snapchangedTargetBlock</dfn> (which is either null or an
-	{{Element}}) and one <dfn export>snapchangedTargetInline</dfn> (which is either
-	null or an {{Element}}) in the block and inline axes respectively.
+	one <dfn export>snapchangedTargetBlock</dfn> and one
+	<dfn export>snapchangedTargetInline</dfn> in the block and inline axes
+	respectively, each of which can either be null if the container is not
+	snapped in that axis or the {{Element}} to which the container is snapped.
 
 	When asked to <dfn export for=Document>update snapchanged targets</dfn>
 	for a <a spec=css-scroll-snap lt="scroll snap container">snap container</a>,
@@ -423,9 +424,10 @@ Snap Events {#snap-events}
 
 	Each
 	<a spec=css-scroll-snap lt="scroll snap container">snap container</a> has
-	one <dfn>snapchangingTargetBlock</dfn> (which is either null or an {{Element}})
-	and one <dfn>snapchangingTargetInline</dfn> (which is either null or an
-	{{Element}}) in the block and inline axes respectively.
+	one <dfn>snapchangingTargetBlock</dfn> 
+	and one <dfn>snapchangingTargetInline</dfn>in the block and inline axes
+	respectively, each of which can either be null if the container is not
+	snapping in that axis or the {{Element}} to which the container is snapping.
 
 	When asked to <dfn export for=Document>update snapchanging targets</dfn>
 	for a <a spec=css-scroll-snap lt="scroll snap container">snap container</a>,
--- a/css-scroll-snap-2/Overview.bs
+++ b/css-scroll-snap-2/Overview.bs
@@ -326,9 +326,6 @@ Snap Events {#snap-events}
 	</li>
 	</ol>
 
-	Although, snapping occurs when a snap container is first laid out, {{snapchanged}}
-	should not be triggered by this initial layout.
-
 	Scrolling operations always lead to {{scrollend}} events being fired. If a
 	scrolling operation also results in a {{snapchanged}} event being fired, the
 	{{snapchanged}} event should be fired before the {{scrollend}} event.
@@ -351,51 +348,61 @@ Snap Events {#snap-events}
 		associated with |snapcontainer|.
 	1. Let <var>inlineSnapchangingTarget</var> be the
 		<a>snapchangingTargetInline</a> associated with |snapcontainer|.
-	1. If <var>blockTarget</var> is not the same element as
-		<var>blockSnapchangingTarget</var>:
+	1. If <var>blockTarget</var> is not the same element as <var>blockSnapchangingTarget</var> or
+			<var>inlineTarget</var> is not the same element as <var>inlineSnapchangingTarget</var>:
 		1. Set the <a>snapchangedTargetBlock</a> associated with |snapcontainer| to
 			<var>blockSnapchangingTarget</var>.
-		1. If |snapcontainer| is not already in <var>doc</var>'s
-			<a spec=cssom-view lt="pending snapchanged event targets">pending snapchanged event targets</a>.
-			1. Append |snapcontainer| to <var>doc</var>'s
-				<a spec=cssom-view lt="pending snapchanged event targets">pending snapchanged event targets</a>.
-	1. If <var>inlineTarget</var> is not the same element as
-			<var>inlineSnapchangingTarget</var>:
 		1. Set the <a>snapchangedTargetInline</a> associated with |snapcontainer| to
 			<var>inlineSnapchangingTarget</var>.
-		1. If |snapcontainer| is not already in <var>doc</var>'s
-			<a spec=cssom-view lt="pending snapchanged event targets">pending snapchanged event targets</a>.
-			1. Append |snapcontainer| to <var>doc</var>'s
-				<a spec=cssom-view lt="pending snapchanged event targets">pending snapchanged event targets</a>.
+		1. Fire a {{SnapEvent}}, |snapevent|, named {{snapchanged}} at <var>target</var>
+			 and let |snapevent|'s {{SnapEvent/snapTargetBlock}} and
+			 {{SnapEvent/snapTargetInline}} attributes be the <a>snapchangedTargetBlock</a> and the
+			 <a>snapchangedTargetInline</a>, respectively, that are associated with <var>target</var>.
 
-<h4 id="snapchanging"> snapchanging </h4>
-	{{snapchanging}} is dispatched when the element to which a
-	snap container would snap (in either axis) during a scrolling operation changes.
+	Note: When snapping occurs on a scroller (either due to a layout change or a
+	 scrolling operation) the <a>snapchangingTargetBlock</a> and <a>snapchangingTargetInline</a>
+	 associated with that scroller are updated and represent the current snap targets
+	 of that scroller. This allows the <a>update snapchanged targets</a> algorithm
+	 to use these elements to determine whether a {{snapchanged}} event should be fired.
+
+	<h4 id="snapchanging"> snapchanging </h4>
+	{{snapchanging}} is dispatched:
+	* during a scrolling operation, if the element to which a
+		 <a spec=css-scroll-snap lt="scroll snap container">snap container</a> would
+		 <a spec="css-scroll-snap-1" lt="scroll snap">snap</a> (in either axis) changes, or
+	* if a [[css-scroll-snap-1#re-snap|layout change]] occurs such that a {{snapchanged}} event
+		 is to be dispatched. In this case, as with the scrolling case, the {{snapchanging}} event
+		 should be dispatched before the {{snapchanged}} event.
+
 	A scrolling operation may animate towards a particular position (e.g.
 	scrollbar arrow clicks, arrow key presses, "behavior: smooth" programmatic
 	scrolls) or may directly track a user's input (e.g. touch scrolling, scrollbar
-	dragging).
-	* In the former case, the user agent should evaluate whether snapchanging is
-		occurring based on the
-		<a spec="css-scroll-snap-1" lt="scroll snap position">snap position</a>
-		that will eventually be snapped to after the scroll animation's target offset
-		is reached.
-	* In the latter case, the user agent should evaluate whether snapchanging
-		should be triggered based on the
-		<a spec="css-scroll-snap-1" lt="scroll snap position">snap position</a> that
-		would be snapped to if the scroll is ended at the current scroll position
-		(determined by the user's input).
+	dragging). In either case, the user agent [[css-scroll-snap-1#choosing|chooses]] an
+	<dfn>eventual snap target</dfn> in each axis to which the scroller will
+	<a spec="css-scroll-snap-1" lt="scroll snap">snap</a> after the scrolling operation
+	reaches its intended scroll position.
+	* In the former case, the intended scroll position is the scroll animation's
+		 target scroll offset.
+	* In the latter case, the intended scroll position is the current scroll offset as
+		 determined by the user's input.
 
 	{{snapchanging}} aims to let the web page know, as early as possible,
 	that the scrolling operation will result in a change in the element the snap
-	container is snapped to.
-
-	{{snapchanging}} events should fire before {{scroll}} events.
+	container is snapped to. The user agent should evaluate whether to trigger
+	{{snapchanging}} based on the	<a>eventual snap target</a> to which the scroller would
+	<a spec="css-scroll-snap-1" lt="scroll snap">snap</a> were the scrolling operation
+	to reach its intended scroll position.
 
 	Note: Since snapchanging gives the web page hints about future snapping,
 	the snapping hinted at by a snapchanging event may not materialize since it
 	will be	possible for subsequent scrolling input to further alter the snap
-	container's scroll position and result in a different eventual snap position.
+	container's scroll position and result in a different eventual snap target.
+
+
+	{{snapchanging}} events should fire before {{scroll}} events.
+
+	Each {{Document}} has an associated list of
+	<dfn export for=Document>pending snapchanging event targets</dfn>, initially empty.
 
 	Each
 	<a spec=css-scroll-snap lt="scroll snap container">snap container</a> is
@@ -407,35 +414,52 @@ Snap Events {#snap-events}
 	|snapcontainer|, run these steps:
 
 	1. Let <var>doc</var> be |snapcontainer|'s associated {{Document}}.
-	1. Let <var>scrollPosition</var> be the scroll offset resulting from applying
-		all pending scroll updates to |snapcontainer|.
 	1. Let <var>blockTarget</var> be the <a>snapchangingTargetBlock</a> that is
 		associated with |snapcontainer|.
 	1. Let <var>inlineTarget</var> be the <a>snapchangingTargetInline</a> that is
 		associated with |snapcontainer|.
-	1. Let <var>newBlockTarget</var> be the element that the UA would snap
-		|snapcontainer| to from <var>scrollPosition</var> (or null if it would not snap
-		to any element) along the block axis.
-	1. Let <var>newInlineTarget</var> be the element that the UA would snap
-		|snapcontainer| to from <var>scrollPosition</var> (or null if it would not snap
-		to any element) along the inline axis.
-	1. If <var>newBlockTarget</var> is not the same element as
-		<var>blockTarget</var>:
+	1. Let <var>newBlockTarget</var> be the <a>eventual snap target</a> that |snapcontainer| will snap
+		to along the block axis or null if it would not snap to any element.
+	1. Let <var>newInlineTarget</var> be the <a>eventual snap target</a> that |snapcontainer| will snap
+		to along the inline axis or null if it would not snap to any element.
+	1. If <var>newBlockTarget</var> is not the same element as <var>blockTarget</var>:
 		1. Set the <a>snapchangingTargetBlock</a> associated with |snapcontainer| to
 			<var>newBlockTarget</var>.
 		1. If |snapcontainer| is not already in <var>doc</var>'s
-			<a spec=cssom-view lt="pending snapchanging event targets">pending snapchanging event targets</a>,
+			<a>pending snapchanging event targets</a>,
 			1. Append |snapcontainer| to <var>doc</var>'s
-				<a spec=cssom-view lt="pending snapchanging event targets">pending snapchanging event targets</a>
-	1. If <var>newInlineTarget</var> is not the same element as
-		<var>inlineTarget</var>:
+				<a>pending snapchanging event targets</a>
+	1. If <var>newInlineTarget</var> is not the same element as <var>inlineTarget</var>:
 		1. Set the <a>snapchangingTargetInline</a> associated with |snapcontainer| to
 			<var>newInlineTarget</var>.
 		1. If |snapcontainer| is not already in <var>doc</var>'s
-			<a spec=cssom-view lt="pending snapchanging event targets">pending snapchanging event targets</a>,
+			<a>pending snapchanging event targets</a>,
 			1. Append |snapcontainer| to <var>doc</var>'s
-				<a spec=cssom-view lt="pending snapchanging event targets">pending snapchanging event targets</a>.
+				<a>pending snapchanging event targets</a>.
 
+
+	<h4 id="snap-events-on-layout-changes"> Snap Events due to Layout Changes </h4>
+	When a <a spec=css-scroll-snap lt="scroll snap container">snap container</a>,
+	|snapcontainer|, [[css-scroll-snap-1#re-snap|re-snaps]], run these steps:
+
+	1. Let <var>newBlockTarget</var> be the element that |snapcontainer| has
+		 <a spec="css-scroll-snap-1" lt="scroll snap">snapped</a>  to
+		 in the block axis or null if it did not snap to any element.
+	1. Let <var>newInlineTarget</var> be the element that |snapcontainer| has
+		 <a spec="css-scroll-snap-1" lt="scroll snap">snapped</a> to
+		 in the inline axis or null if it did not snap to any element.
+	1. Let <var>snapchangingBlock</var> be the <a>snapchangingTargetBlock</a> associated
+		with |snapcontainer|.
+	1. Let <var>snapchangingInline</var> be the <a>snapchangingTargetInline</a> associated
+		with |snapcontainer|.
+	1. If <var>newBlockTarget</var> is not the same element as <var>snapchangingBlock</var> or
+		 <var>newInlineTarget</var> is not the same element as <var>snapchangingInline</var>:
+		1. Set the <a>snapchangingTargetBlock</a> associated with |snapcontainer| to <var>newBlockTarget</var>.
+		1. Set the <a>snapchangingTargetInline</a> associated with |snapcontainer| to <var>newInlineTarget</var>.
+		1. Fire a {{SnapEvent}}, |snapevent|, named {{snapchanging}} at |snapcontainer| and let
+			 |snapevent|'s {{SnapEvent/snapTargetBlock}} and {{SnapEvent/snapTargetInline}}
+			 attributes be <var>newBlockTarget</var> and <var>newInlineTarget</var>, respectively.
+		1. Run the steps to <a>update snapchanged targets</a> for |snapcontainer|.
 
 SnapEvent interface
 -------------------
--- a/css-scroll-snap-2/Overview.bs
+++ b/css-scroll-snap-2/Overview.bs
@@ -330,6 +330,9 @@ Snap Events {#snap-events}
 	scrolling operation also results in a {{snapchanged}} event being fired, the
 	{{snapchanged}} event should be fired before the {{scrollend}} event.
 
+	Each {{Document}} has an associated list of
+	<dfn export for=Document>pending snapchanged event targets</dfn>, initially empty.
+
 	Each
 	<a spec=css-scroll-snap lt="scroll snap container">snap container</a> is
 	associated with at most one <dfn export>snapchangedTargetBlock</dfn> and at most one
@@ -358,16 +361,25 @@ Snap Events {#snap-events}
 			<var>inlineSnapchangingTarget</var>.
 		1. Set <var>snap targets changed</var> to true.
 	1. If <var>snap targets changed</var> is true:
-		1. Fire a {{SnapEvent}}, |snapevent|, named {{snapchanged}} at <var>target</var>
-			 and let |snapevent|'s {{SnapEvent/snapTargetBlock}} and
-			 {{SnapEvent/snapTargetInline}} attributes be the <a>snapchangedTargetBlock</a> and the
-			 <a>snapchangedTargetInline</a>, respectively, that are associated with <var>target</var>.
+		1. If |snapcontainer| is not already in <var>doc</var>'s
+			<a>pending snapchanged event targets</a>:
+			1. Append |snapcontainer| to <var>doc</var>'s
+				<a>pending snapchanged event targets</a>.
 
 	Note: When snapping occurs on a scroller (either due to a layout change or a
 	 scrolling operation) the <a>snapchangingTargetBlock</a> and <a>snapchangingTargetInline</a>
 	 associated with that scroller are updated and represent the current snap targets
 	 of that scroller. This allows the <a>update snapchanged targets</a> algorithm
 	 to use these elements to determine whether a {{snapchanged}} event should be fired.
+
+	When asked to <dfn>dispatch pending snapchanged events</dfn> for a {{Document}},
+		<var>doc</var>, run these steps:
+	1. For each item <var>target</var> in |doc|'s <a>pending snapchanged event targets</a>:
+		1. Fire a {{SnapEvent}}, |snapevent|, named {{snapchanged}} at <var>target</var>
+			and let |snapevent|'s {{SnapEvent/snapTargetBlock}} and
+			{{SnapEvent/snapTargetInline}} attributes be the <a>snapchangedTargetBlock</a> and the
+			<a>snapchangedTargetInline</a>, respectively, that are associated with <var>target</var>.
+	1. Empty <var>doc</var>'s <a>pending snapchanged event targets</a>.
 
 	<h4 id="snapchanging"> snapchanging </h4>
 	{{snapchanging}} is dispatched:
@@ -415,17 +427,14 @@ Snap Events {#snap-events}
 
 	When asked to <dfn export for=Document>update snapchanging targets</dfn>
 	for a <a spec=css-scroll-snap lt="scroll snap container">snap container</a>,
-	|snapcontainer|, run these steps:
+	|snapcontainer|, given an {{Element}} newBlockTarget and an {{Element}}
+	newInlineTarget, run these steps:
 
 	1. Let <var>doc</var> be |snapcontainer|'s associated {{Document}}.
 	1. Let <var>blockTarget</var> be the <a>snapchangingTargetBlock</a> that is
 		associated with |snapcontainer|.
 	1. Let <var>inlineTarget</var> be the <a>snapchangingTargetInline</a> that is
 		associated with |snapcontainer|.
-	1. Let <var>newBlockTarget</var> be the <a>eventual snap target</a> that |snapcontainer| will snap
-		to along the block axis or null if it would not snap to any element.
-	1. Let <var>newInlineTarget</var> be the <a>eventual snap target</a> that |snapcontainer| will snap
-		to along the inline axis or null if it would not snap to any element.
 	1. If <var>newBlockTarget</var> is not the same element as <var>blockTarget</var>:
 		1. Set the <a>snapchangingTargetBlock</a> associated with |snapcontainer| to
 			<var>newBlockTarget</var>.
@@ -441,8 +450,16 @@ Snap Events {#snap-events}
 			1. Append |snapcontainer| to <var>doc</var>'s
 				<a>pending snapchanging event targets</a>.
 
+	When asked to <dfn>dispatch pending snapchanging events</dfn> for a {{Document}},
+		<var>doc</var>, run these steps:
+	1. For each item <var>target</var> in |doc|'s <a>pending snapchanging event targets</a>:
+		1. Fire a {{SnapEvent}}, |snapevent|, named {{snapchanging}} at <var>target</var>
+			and let |snapevent|'s {{SnapEvent/snapTargetBlock}} and
+			{{SnapEvent/snapTargetInline}} attributes be the <a>snapchangingTargetBlock</a> and the
+			<a>snapchangingTargetInline</a>, respectively, that are associated with <var>target</var>.
+  1. Empty <var>doc</var>'s <a>pending snapchanging event targets</a>.
 
-	<h4 id="snap-events-on-layout-changes"> Snap Events due to Layout Changes </h4>
+	<h4 id="snap-events-on-layout-changes">Snap Events due to Layout Changes </h4>
 	When a <a spec=css-scroll-snap lt="scroll snap container">snap container</a>,
 	|snapcontainer|, [[css-scroll-snap-1#re-snap|re-snaps]], run these steps:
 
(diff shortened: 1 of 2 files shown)


DavMila:
Hi, @flackr , ptal at this addition of SnapEvent definition to the scroll-snap-2 spec.

PS: I don't seem to be able to add reviewers so I thought I'd just ping you.

css-scroll-snap-2/Overview.bs
```diff
@@ -279,13 +279,35 @@ Snap Events {#snap-events}
...
+			[Exposed=Window]
+			interface SnapEvent : Event {
+				readonly attribute NodeList snapTargets;
+			};
+</pre>
+
+<dl>
+		<dt><code>SnapEvent . snapTargets</code></dt>
+		<dd>
+			A list of elements that a scroller is considered snapped to on either axis.
```
flackr, line 307: This could use more details. In particular,

What is the order of the elements?

If you look at [getAnimations](https://drafts.csswg.org/web-animations-1/#dom-animatable-getanimations) it says:
> The returned list is sorted using the composite order described for the associated [animations](https://drafts.csswg.org/web-animations-1/#concept-animation) of effects in [§ 5.4.2 The effect stack](https://drafts.csswg.org/web-animations-1/#the-effect-stack).

Where the composite order is defined in:
https://drafts.csswg.org/web-animations-1/#animation-composite-order

I suggest defining an order similar to the [css animation composite order](https://drafts.csswg.org/css-animations-2/#animation-composite-order) which is basically [tree order](https://dom.spec.whatwg.org/#concept-tree-order) with a defined ordering for pseudo-elements.

What does it mean that a scroller is considered snapped to a particular element?

For this, you probably want to refer to being in a valid snap position for the particular element where this is defined in css-scroll-snap-1. You should also mention how for snapchanging it is referring to the eventual scroll position, not the position that is currently scrolled to.

css-scroll-snap-2/Overview.bs
```diff
@@ -279,13 +279,35 @@ Snap Events {#snap-events}
...
+				<th><dfn event>snapchanged</dfn>
 				<td>{{scroll!!event}}
 			<tr>
-				<th><dfn event>snapChanging</dfn>
+				<th><dfn event>snapchanging</dfn>
 				<td>{{scroll!!event}}
 	</table>
 
+	{{ snapchanged }} will indicate that the set of snap areas to which a snap container is snapped has changed.
+	{{ snapchanging }} will hint at the set of snap areas to which a snap container would snap during a scroll.
```
flackr, line 290: When are these events dispatched?

css-scroll-snap-2/Overview.bs
```diff
@@ -279,13 +279,35 @@ Snap Events {#snap-events}
 				<th>Event handler event type
 		<tbody>
 			<tr>
-				<th><dfn event>snapChanged</dfn>
+				<th><dfn event>snapchanged</dfn>
 				<td>{{scroll!!event}}
 			<tr>
-				<th><dfn event>snapChanging</dfn>
+				<th><dfn event>snapchanging</dfn>
 				<td>{{scroll!!event}}
```
flackr, line 286: Should we expand the table to include the targets and descriptions for these events similar to https://drafts.csswg.org/cssom-view-1/#eventdef-document-scroll or perhaps like https://w3c.github.io/uievents/#events-wheel-types ?

I think event handler event type is meant to be the type of the event which I believe should be the new type you've defined below.

flackr (commented):
I'm not an

css-scroll-snap-2/Overview.bs
DavMila: Thanks for the links. I think it suffices to just specify that the list is sorted in tree order so I've included that comment. One point of consideration was pseudo-elements: tree order does not seem to cover pseudo-elements, but since there are no elements for pseudo-elements, pseudo elements would not be included in snapTargets which is a list of elements.

I have also added a reference to the css-scroll-snap-1 spec for clarity on the criteria for populating snapTargets.

css-scroll-snap-2/Overview.bs
DavMila: Thanks for pointing out this gap. snapchanged should fire at the end of a scroll (or style/layout change that triggers snapchanged) and before a scrollend event. snapchanging should fire during a scroll, before scroll events. I've included these details in the spec.

css-scroll-snap-2/Overview.bs
DavMila: Good suggestion. The table is now expanded to include "Interface", "Targets", & "Description" columns. The Targets for snapchanged is "scroll containers" and not snap container only because of the one case where snapchanged happens because of a style change from non-none scroll-snap-type to none, where technically, the event target is no longer a snap container. (Happy to change it if making this distinction isn't helpful).

css-scroll-snap-2/Overview.bs
```diff
@@ -262,7 +262,7 @@ Snap Events {#snap-events}
 ████████    ███    ████████ ██    ██    ██     ██████
 -->
 
-{{snapChanged}} and {{snapChanging}} {#snapchanged-and-snapchanging}
+{{snapchanged}} and {{snapchanging}} {#snapchanged-and-snapchanging}
```
dontcallmedom, line 265: see [discussion using past tense for the event name](https://github.com/w3c/csswg-drafts/issues/7442#issuecomment-1214626815)

css-scroll-snap-2/Overview.bs
zcorpan: For `scroll` and `scrollend`, it's defined here:

https://drafts.csswg.org/cssom-view-1/#scrolling-events

which integrates closely with HTML's event loop:

https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering

If new events are to be fired at `scroll`-event timing, it should be in cssom-view's algo so that it's well-defined when they fire. But also see https://github.com/w3c/csswg-drafts/issues/7442#issuecomment-1824851458

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+			from the snap area it most recently snapped to in the block axis, or
+		</li>
+		<li>	the snap area it is snapped to in the inline axis is different
+			from the snap area it most recently snapped to in the inline axis.
+		</li>
+		</ul>
+	</li>
+	<li> (for snap containers with ''scroll-snap-type/proximity'' strictness) at the end of a scroll in which
+		the snap container scrolls and settles into (or out of) a
+		snap area's range of proximity snap positions in either axis.
```
flackr, line 323: I don't understand, isn't this the same case as the first one, i.e. after the scrolling operation is completed we determine that a new snap point or no snap point is snapped?

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+	{{snapchanged}} indicates that the snap area to which a snap container is snapped along either axis has changed.
+	{{snapchanged}} should be dispatched:
+
+	<ol>
+	<li>
+		at the end of a scrolling operation, after the user agent has completed its
+		snapping, if either:
+		<ul>
+		<li>	the snap area which the snap container is snapped to (TODO: link spec on how the target is picked) in the block axis is different
+			from the snap area it most recently snapped to in the block axis, or
```
flackr, line 314: Might want to be explicit that for proximity strictness this may be no snapped area.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+		</tbody>
 	</table>
 
+	<h4 for="snapchanged" id="snapchanged"> snapchanged </h4>
+	{{snapchanged}} indicates that the snap area to which a snap container is snapped along either axis has changed.
+	{{snapchanged}} should be dispatched:
+
+	<ol>
+	<li>
+		at the end of a scrolling operation, after the user agent has completed its
```
flackr, line 310: This could probably refer to the completed term in cssom-view, https://drafts.csswg.org/cssom-view/#scroll-completed . Note that we should also specify the relative order with respect to scroll and scrollend events, which is further reason to make this part of cssom-view as suggested by @zcorpan here: https://github.com/w3c/csswg-drafts/pull/9515#discussion_r1403693509

We could for now add it as an appendix, similar to scroll-animation additions to the other animations specs: https://www.w3.org/TR/scroll-animations-1/#timeline-name-scope

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+			from the snap area it most recently snapped to in the inline axis.
+		</li>
+		</ul>
+	</li>
+	<li> (for snap containers with ''scroll-snap-type/proximity'' strictness) at the end of a scroll in which
+		the snap container scrolls and settles into (or out of) a
+		snap area's range of proximity snap positions in either axis.
+	</li>
+	<li> if there is a change to a snap container's style such that it goes from having a
+		non-'none' value for scroll-snap-type to having a 'none' value or vice versa.
```
flackr, line 326: This should have well specified timing, maybe this occurs as part of "run the scroll steps" as well? https://drafts.csswg.org/cssom-view/#document-run-the-scroll-steps . I'm not sure if we would expect snap events for a non snapping scroller.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+		the snap position that will eventually be snapped to after the scroll animation's
+		target offset is reached.
+	* In the latter case, the user agent should evaluate whether snapchanging should
+		be triggered based on the snap position that would be snapped to if
+		the scroll is ended at the current scroll position (determined by the user's input).
+
+	Note that since snapchanging gives the web page hints about future snapping,
+	the snapping hinted at by a snapchanging event may not materialize since it will be
+	possible for subsequent scrolling input to further alter the snap container's
+	scroll position and result in a different eventual snap position.
```
flackr, line 359: I suspect it may be clearer to explain the semantics and timing of snapchanging and snapchanged together as an algorithm. E.g. I think we expect that snapchanging is fired before snapchanged, but also that both will be fired in some cases, e.g. if a scroll operation ends unexpectedly early we might fire snapchanging and snapchanged immediately for the current snap location.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+	that the scrolling operation will result in a change in the snap area
+	the snap container is snapped to.
+	* In the former case, the user agent should evaluate whether snapchanging is occurring based on
+		the snap position that will eventually be snapped to after the scroll animation's
+		target offset is reached.
+	* In the latter case, the user agent should evaluate whether snapchanging should
+		be triggered based on the snap position that would be snapped to if
+		the scroll is ended at the current scroll position (determined by the user's input).
+
+	Note that since snapchanging gives the web page hints about future snapping,
```
flackr, line 356: There is a bikeshed specific note syntax: https://speced.github.io/bikeshed/#notes-etc

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+	<h4 for="snapchanged" id="snapchanged"> snapchanged </h4>
+	{{snapchanged}} indicates that the snap area to which a snap container is snapped along either axis has changed.
+	{{snapchanged}} should be dispatched:
+
+	<ol>
+	<li>
+		at the end of a scrolling operation, after the user agent has completed its
+		snapping, if either:
+		<ul>
+		<li>	the snap area which the snap container is snapped to (TODO: link spec on how the target is picked) in the block axis is different
```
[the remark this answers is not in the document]
DavMila: I meant for it to be a TODO until the resolution from the related [issue](https://github.com/w3c/csswg-drafts/issues/9622) about picking snap targets is included in the scroll-snap-1 spec but I  think I can just link to: https://drafts.csswg.org/css-scroll-snap-1/#choosing.

css-scroll-snap-2/Overview.bs
DavMila: It is, but I thought it was worth being explicit about the specific case of proximity strictness scrollers. I think this is the same thing you're addressing in your other comment?
Perhaps it doesn't need to be its own <li>. I'll combine it with the first list item.

css-scroll-snap-2/Overview.bs
DavMila: Agreed, I'll combine it with the first <li>.

css-scroll-snap-2/Overview.bs
DavMila: I've specified in the snapchanged and snapchanging sections that snapchanging events should be fired before scroll events and that snapchanged events should be fired before scrollend events. I've also added Appendix which describes an algorithm where snapchanging should be fired during scrolling and snapchanged should be fired after scrolling is completed and referenced scroll-completed from cssom-view.

css-scroll-snap-2/Overview.bs
DavMila: I've added Appendix B which indicates that snapchanging will fire during scrolling and snapchanged will fire after scrolling is completed.

css-scroll-snap-2/Overview.bs
DavMila: Thanks, I've adopted this.

tabatkins:
@emilio (or @smfr) could you review the additions to the CSSOM View spec here? We want to make sure we're not monkey-patching awkwardly.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,17 +275,125 @@ Snap Events {#snap-events}
...
+				<td>Fired at the scrolling element or {{Document}} at the end of a scroll (before a {{scrollend}} event)
+					or after a <a href="https://drafts.csswg.org/css-scroll-snap-1/#re-snap">layout snap</a>
+					if the element that the scrolling element or Document is snapped to changed.</td>
+			</tr>
 			<tr>
-				<th><dfn event>snapChanging</dfn>
-				<td>{{scroll!!event}}
+				<th><dfn for="snapchanging" event>snapchanging</dfn></th>
+				<td>{{SnapEvent}}</td>
+				<td>snap containers</td>
```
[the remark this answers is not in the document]
DavMila: Yes, changed, thanks.

css-scroll-snap-2/Overview.bs
DavMila: I've added an "Appendix B" section as well as changes to the [Events](https://drafts.csswg.org/cssom-view/#events) section of cssom-view-1 which integrates with the [scroll steps](https://drafts.csswg.org/cssom-view/#document-run-the-scroll-steps).

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,129 @@ Snap Events {#snap-events}
...
+			</tr>
 			<tr>
-				<th><dfn event>snapChanging</dfn>
-				<td>{{scroll!!event}}
+				<th><dfn for="snapchanging" event>snapchanging</dfn></th>
+				<td>{{SnapEvent}}</td>
+				<td>scroll containers</td>
+				<td>Fired at the scrolling element or {{Document}} during scrolling (before a {{scroll}} event),
+					if the scrolling would lead to a change in the
+					element that the scrolling element or Document is snapped to.</td>
```
flackr, line 299: I think we want to explicitly call out that it is also fired if it's different than the last time the event was fired. E.g. if you scroll over a new potential target and scroll back, it should fire another event to tell you you are back on the original element even though its the same as the element the document is snapped to.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,129 @@ Snap Events {#snap-events}
...
+	{{snapchanging}} aims to let the web page know, as early as possible,
+	that the scrolling operation will result in a change in the element the snap
+	container is snapped to.
+
+	{{snapchanging}} events should fire before {{scroll}} events.
+
+	Note: Since snapchanging gives the web page hints about future snapping,
+	the snapping hinted at by a snapchanging event may not materialize since it
+	will be	possible for subsequent scrolling input to further alter the snap
+	container's	scroll position and result in a different eventual snap position.
```
flackr, line 363: Remove extra space?

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,129 @@ Snap Events {#snap-events}
...
+			<div dfn-type=attribute class=attributes dfn-for="SnapEvent">
+		: <dfn>snapTargetInline</dfn>
+		::
+			The element that the snap container is snapped to in the inline axis
+			at the <a spec="css-scroll-snap-1" lt="scroll snap position">snap position</a>
+			for the associated snap event.
+			</div>
+
+		For {{snapchanged}} events, the snap position is the position already
+		realized by	the snap container after a scroll snap. For {{snapchanging}}
```
flackr, line 394: Looks like some extra space snuck in?

css-scroll-snap-2/Overview.bs
```diff
@@ -470,3 +480,114 @@ Physical Longhands for 'scroll-start-target' {#scroll-start-target-longhands-phy
...
+		 to <var>newBlockTarget</var>.
+	2. Append the snap container to <var>doc</var>'s
+		 <a>pending snapchanging event targets</a> 
+7. If <var>newInlineTarget</var> is not the same element as
+	 <var>inlineTarget</var>:
+
+	1. Set <var>doc</var>'s <a>snapchangingTargetInline</a> for the snap container
+		 to <var>newInlineTarget</var>.
+	2. Append the snap container to <var>doc</var>'s pending snapchanging event
+		 targets.
```
flackr, line 532: I think you only want to append the snap container to the pending snapchanging event targets list once. Note that cssom-view says to only do this if the target is not already in the pending scroll event targets.

css-scroll-snap-2/Overview.bs
```diff
@@ -470,3 +480,114 @@ Physical Longhands for 'scroll-start-target' {#scroll-start-target-longhands-phy
 	</pre>
 
 	...
+
+Appendix B: Snap Events Timing {#snap-events-timing}
+==================================
+
+This appendix describes the algorithm UAs should follow to dispatch {{snapchanged}}
```
flackr, line 487: This is a patch to cssom-view scrolling, correct? https://www.w3.org/TR/cssom-view-1/#scrolling-events Can you make it explicit that we are patching those algorithms, and specifically that they run before the related steps mentioned there.

I think ... though @tabatkins can confirm, that it may be preferred to list all modifications to cssom-view-1 here so that they can be reviewed as part of css-scroll-snap-2.

cssom-view-1/Overview.bs
```diff
@@ -1899,6 +1899,81 @@ When asked to <dfn export for=Document>run the resize steps</dfn> for a {{Docume
...
+
+    Each {{Document}} associates each of its
+    <a spec=css-scroll-snap lt="scroll snap container">snap containers</a>
+    with at most one <dfn>snapchangingTargetBlock</dfn> and at most one
+    <dfn>snapchangingTargetInline</dfn> in the block and inline axes respectively.
+
+    Each {{Document}} associates each of its
+    <a spec=css-scroll-snap lt="scroll snap container">snap containers</a>
+    with at most one <dfn>snapchangedTargetBlock</dfn> and at most one
+    <dfn>snapchangedTargetInline</dfn> in the block and inline axes respectively.
```
emilio, line 1914: Duplicated paragraph?

cssom-view-1/Overview.bs
```diff
@@ -1899,6 +1899,81 @@ When asked to <dfn export for=Document>run the resize steps</dfn> for a {{Docume
...
+            to <var>newBlockTarget</var>.
+        1. Append the snap container to <var>doc</var>'s
+            <a>pending snapchanging event targets</a>
+
+    1. If <var>newInlineTarget</var> is not the same element as
+        <var>inlineTarget</var>:
+        1. Set <var>doc</var>'s <a>snapchangingTargetInline</a> for the snap container
+            to <var>newInlineTarget</var>.
+        1. Append the snap container to <var>doc</var>'s
+            <a>pending snapchanging event targets</a>
```
emilio, line 1947: So if both inline and block change, you append it twice, which means that you fire two snapchanging events, that seems wrong?

cssom-view-1/Overview.bs
```diff
@@ -1899,6 +1899,81 @@ When asked to <dfn export for=Document>run the resize steps</dfn> for a {{Docume
...
+        1. Set <var>doc</var>'s <a>snapchangedTargetBlock</a> for the snap container
+            to <var>blockSnapchangingTarget</var>.
+        2. Append the snap container to <var>doc</var>'s
+            <a>pending snapchanged event targets</a>.
+    1. If <var>inlineTarget</var> is not the same element as
+            <var>inlineSnapchangingTarget</var>:
+        1. Set <var>doc</var>'s <a>snapchangedTargetInline</a> for the snap container
+            to <var>inlineSnapchangingTarget</var>.
+        2. Append the snap container to <var>doc</var>'s
+            <a>pending snapchanged event targets</a>.
```
emilio, line 1976: Same issue here, I believe.

emilio, line 1976: Same issue here, I believe.

emilio:
> @emilio (or @smfr) could you review the additions to the CSSOM View spec here? We want to make sure we're not monkey-patching awkwardly.

It seems fine with stuff fixed (I just noticed that @flackr had caught some of the issues I pointed out too, whoops).

cssom-view-1/Overview.bs
emilio: Ah, not duplicated (changed/changing). I think these could hang off the snap container itself tho, right? As in, the document doesn't _really_ do anything here.

css-scroll-snap-2/Overview.bs
DavMila: Yes, in fact I think the case you describe is actually all we need - what I had before ("a change in the	element that the scrolling element or Document is snapped to.")  implies that snapchanging would keep firing as long as you are scrolling over a target that isn't the last snapped-to target, which isn't the intent. I've adjusted the wording.

css-scroll-snap-2/Overview.bs
DavMila: Seems I had a tab in there, fixed.

css-scroll-snap-2/Overview.bs
DavMila: Yes, fixed.

css-scroll-snap-2/Overview.bs
DavMila: Yes, I've added wording to make it clear that the snap container should only be added to the list if it's not already there.

css-scroll-snap-2/Overview.bs
DavMila: I've included text saying that that section is "a patch to [scrolling events](https://www.w3.org/TR/cssom-view-1/#scrolling-events)"  and pointed out within the algorithm that the steps firing snap events should be run before the steps firing the scroll/scrollend events.
All the text added to cssom-view can be found in this appendix as well but I'm not sure if there is a particular format for representing a "patch"?

cssom-view-1/Overview.bs
DavMila: Yes it can be written without document, thanks for the suggestion. I've rewritten these without document.

cssom-view-1/Overview.bs
DavMila: Yes I've corrected this to say that it should be appended only if it isn't already in the list.

cssom-view-1/Overview.bs
DavMila: Thanks, addressed as above.

DavMila: Thanks, addressed as above.

cssom-view-1/Overview.bs
DavMila: And thanks for taking a look!

css-scroll-snap-2/Overview.bs
flackr: Linking to the relevant spec and explaining at which step the new algorithms run is sufficient. It doesn't need to be a literal patch. That said, if you are adding directly to cssom-view in this PR then we don't need the appendix section here.

> All the text added to cssom-view can be found in this appendix as well

I think we should only add this text in one place. Either here or in cssom-view. We shouldn't repeat the algorithm in both places as there should be one place in which it is defined.

I think the simplest would be to have a single step in cssom-view run an algorithm in this spec. This is similar for example to how [update animations and send events is defined in web-animations-1](https://www.w3.org/TR/web-animations-1/#update-animations-and-send-events) but [called from the html spec](https://html.spec.whatwg.org/#:~:text=For%20each%20doc%20of%20docs%2C%20update%20animations%20and%20send%20events%20for%20doc%2C%20passing%20in%20relative%20high%20resolution%20time%20given%20frameTimestamp%20and%20doc%27s%20relevant%20global%20object%20as%20the%20timestamp%20%5BWEBANIMATIONS%5D)

DavMila: That makes sense, I've taken out the appendix section and put the algorithms in the snapchanged and snapchanging sections and linked to them from the relevant cssom-view steps.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,209 @@ Snap Events {#snap-events}
...
+	associated with at most one <dfn>snapchangingTargetBlock</dfn> and at most one
+	<dfn>snapchangingTargetInline</dfn> in the block and inline axes respectively.
+
+	When asked to <dfn export for=Document>update snapchanging targets</dfn>
+	for a <a spec=css-scroll-snap lt="scroll snap container">snap container</a>,
+	|snapcontainer|, run these steps:
+
+	1. Let <var>doc</var> be |snapcontainer|'s associated {{Document}}.
+	1. Let <var>scrollPosition</var> be the scroll offset resulting from applying
+		all pending scroll updates to |snapcontainer|.
```
flackr, line 411: I don't think "pending scroll updates" is quite correct as they're not pending yet. I think you should define a term such as "eventual snap target" above where you describe the cases of scrolling operations, and then refer to that term here.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,209 @@ Snap Events {#snap-events}
...
+	1. Let <var>inlineTarget</var> be the <a>snapchangingTargetInline</a> that is
+		associated with |snapcontainer|.
+	1. Let <var>newBlockTarget</var> be the element that the UA would snap
+		|snapcontainer| to from <var>scrollPosition</var> (or null if it would not snap
+		to any element) along the block axis.
+	1. Let <var>newInlineTarget</var> be the element that the UA would snap
+		|snapcontainer| to from <var>scrollPosition</var> (or null if it would not snap
+		to any element) along the inline axis.
+	1. If <var>newBlockTarget</var> is not the same element as
+		<var>blockTarget</var>:
```
flackr, line 423: I think that you'll need to either set snapChangingTargetBlock on first layout or carve out an exception for the first layout to avoid it dispatching a snapchanging event. That said, It may actually be better for developers if we dispatched a snapchanging / snapchanged if a snap target is initially snapped to instead of no target. Otherwise, without adding an API to get the initial snap target it may be hard for a developer to know what it is.

cssom-view-1/Overview.bs
```diff
@@ -1908,15 +1907,23 @@ Each {{Document}} has an associated list of <dfn for=Document>pending scroll eve
 
 Each {{Document}} has an associated list of <dfn for=Document>pending scrollend event targets</dfn>, initially empty.
 
+Each {{Document}} has an associated list of <dfn export for=Document>pending snapchanging event targets</dfn>, initially empty.
+
+Each {{Document}} has an associated list of <dfn export for=Document>pending snapchanged event targets</dfn>, initially empty.
```
flackr, line 1912: I think it might be good to move everything over to scroll-snap-2 and only have cssom-view invoke methods over in scroll-snap-2 at the appropriate times rather than having some of the algorithm in each place.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,209 @@ Snap Events {#snap-events}
...
+			1. Append |snapcontainer| to <var>doc</var>'s
+				<a spec=cssom-view lt="pending snapchanged event targets">pending snapchanged event targets</a>.
+	1. If <var>inlineTarget</var> is not the same element as
+			<var>inlineSnapchangingTarget</var>:
+		1. Set the <a>snapchangedTargetInline</a> associated with |snapcontainer| to
+			<var>inlineSnapchangingTarget</var>.
+		1. If |snapcontainer| is not already in <var>doc</var>'s
+			<a spec=cssom-view lt="pending snapchanged event targets">pending snapchanged event targets</a>.
+			1. Append |snapcontainer| to <var>doc</var>'s
+				<a spec=cssom-view lt="pending snapchanged event targets">pending snapchanged event targets</a>.
```
flackr, line 369: It took me some careful reading to work out why we can use the blockSnapChangingTarget here. Can you add a note below this explaining that the snapchanging inline and block targets are updated before snapchanged and as such represent the actual new snap target?

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,244 @@ Snap Events {#snap-events}
...
+	1. Let <var>newInlineTarget</var> be the element that |snapcontainer| has
+		 <a spec="css-scroll-snap-1" lt="scroll snap">snapped</a> to
+		 in the inline axis or null if it did not snap to any element.
+	1. Let <var>blockTarget</var> be the <a>snapchangingTargetBlock</a> associated
+		with |snapcontainer|.
+	1. Let <var>inlineTarget</var> be the <a>snapchangingTargetInline</a> associated
+		with |snapcontainer|.
+	1. Let <var>snap targets changed</var> be a boolean flag that is initially false.
+	1. If <var>newBlockTarget</var> is not the same element as <var>blockTarget</var>:
+		1. Set the <a>snapchangingTargetBlock</a> associated with |snapcontainer| to
```
flackr, line 461: This looks like it's mostly repeating the snapchanging steps, I think we could invoke the update snapchanging targets algorithm. We can introduce an argument to the algorithm if needed to handle any necessary differences in behavior.

cssom-view-1/Overview.bs
```diff
@@ -1929,6 +1932,14 @@ Whenever a <a>visual viewport</a> gets scrolled (whether in response to user int
...
 When asked to <dfn export for=Document>run the scroll steps</dfn> for a {{Document}} <var>doc</var>, run these steps:
 
+1. For each item <var>target</var> in <var>doc</var>'s
+    <a spec="css-scroll-snap-2" lt="pending snapchanging event targets">pending snapchanging event targets</a>:
+
+    1. Fire a {{SnapEvent}}, |snapevent|, named <a spec=css-scroll-snap-2 event>snapchanging</a> at <var>target</var> and let |snapevent|'s {{SnapEvent/snapTargetBlock}} and
+        {{SnapEvent/snapTargetInline}} attributes be the <a spec=css-scroll-snap-2 lt="snapchangingTargetBlock">snapchangingTargetBlock</a> and the
+        <a spec=css-scroll-snap-2 lt="snapchangingTargetInline">snapchangingTargetInline</a>, respectively, that are
+        associated with <var>target</var>.
+1. Empty <var>doc</var>'s <a spec="css-scroll-snap-2" lt="pending snapchanging event targets">pending snapchanging event targets</a>.
```
flackr, line 1942: These steps should be moved to an algorithm in css-scroll-snap-2 as well, invoked from here.

css-scroll-snap-2/Overview.bs
DavMila: Done. I've introduced parameters newBlockTarget and newInlineTarget. For scrolling operation, they are set to the eventual snap targets. For layout changes they are set to the current snap targets.

cssom-view-1/Overview.bs
DavMila: Good suggestion. Done.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,246 @@ Snap Events {#snap-events}
...
+		describes the method a UA follows when choosing	between elements which are
+		<a spec="css-scroll-snap-1" lt="scroll snap area">snap areas</a>.
+	</li>
+	<li> if there is a change to a snap container's style such that it goes from
+		having a non-'none' value for [[css-scroll-snap-1#scroll-snap-type|scroll-snap-type]]
+		to having a 'none' value or vice versa.
+	</li>
+	<li> if, after a [[css-scroll-snap-1#re-snap|layout change]], the element to
+		which a snap container is	snapped to changes, regardless of whether there is
+		a change in scroll position	after the layout change.
```
flackr, line 325: nit: It seems like you've accidentally introduced some tabs in the middle of these sentences.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,246 @@ Snap Events {#snap-events}
...
+	scrolling operation also results in a {{snapchanged}} event being fired, the
+	{{snapchanged}} event should be fired before the {{scrollend}} event.
+
+	Each {{Document}} has an associated list of
+	<dfn export for=Document>pending snapchanged event targets</dfn>, initially empty.
+
+	Each
+	<a spec=css-scroll-snap lt="scroll snap container">snap container</a> is
+	associated with at most one <dfn export>snapchangedTargetBlock</dfn> and at most one
+	<dfn export>snapchangedTargetInline</dfn> in the block and inline axes respectively.
```
flackr, line 339: It might be cleaner to say each snap container has a snapChangedTargetBlock and snapChangedTargetInline which may be null if not snapped, rather than saying associated with at most one which implies a set-like behavior, especially also since you set these to single elements in the algorithm.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,246 @@ Snap Events {#snap-events}
...
+				<a>pending snapchanging event targets</a>.
+
+	When asked to <dfn>dispatch pending snapchanging events</dfn> for a {{Document}},
+		<var>doc</var>, run these steps:
+	1. For each item <var>target</var> in |doc|'s <a>pending snapchanging event targets</a>:
+		1. Fire a {{SnapEvent}}, |snapevent|, named {{snapchanging}} at <var>target</var>
+			and let |snapevent|'s {{SnapEvent/snapTargetBlock}} and
+			{{SnapEvent/snapTargetInline}} attributes be the <a>snapchangingTargetBlock</a> and the
+			<a>snapchangingTargetInline</a>, respectively, that are associated with <var>target</var>.
+  1. Empty <var>doc</var>'s <a>pending snapchanging event targets</a>.
```
flackr, line 460: Looks like the indentation is off here?

cssom-view-1/Overview.bs
```diff
@@ -1929,12 +1938,14 @@ Whenever a <a>visual viewport</a> gets scrolled (whether in response to user int
...
 When asked to <dfn export for=Document>run the scroll steps</dfn> for a {{Document}} <var>doc</var>, run these steps:
 
+1. Run the steps to <a spec="css-scroll-snap-2">dispatch pending snapchanging events</a> for <var>doc</var>.
 1. For each item <var>target</var> in <var>doc</var>'s <a>pending scroll event targets</a>,
     in the order they were added to the list, run these substeps:
 
     1. If <var>target</var> is a {{Document}}, <a>fire an event</a> named <a event>scroll</a> that bubbles at <var>target</var>.
     1. Otherwise, <a>fire an event</a> named <a event>scroll</a> at <var>target</var>.
 1. Empty <var>doc</var>'s <a>pending scroll event targets</a>.
+1. Run the steps to <a spec="css-scroll-snap-2">dispatch pending snapchanged events</a> for <var>doc</var>.
```
flackr, line 1948: I wonder if there can be ordering issues if we always dispatch snapchanged after snapchanging. I'm thinking of cases where you may have a snapchanged queued for a finished scroll or layout re-snap, followed by subsequent snapchanging for a newly started scroll in which case dispatching the snapchanged afterwards implies that the snapchanging operation is complete. Thoughts?

flackr (commented):
Overall this is looking good, just a few nits and one question about event ordering which maybe we could leave as an open question.

css-scroll-snap-2/Overview.bs
DavMila: Yes, deleted.

css-scroll-snap-2/Overview.bs
DavMila: Good suggestion. I've adopted it.

css-scroll-snap-2/Overview.bs
DavMila: Yes, fixed

DavMila (commented):
Fixed the spaces issues and the "at most one" wording.

cssom-view-1/Overview.bs
flackr: To avoid lumping too much complexity into this one PR, perhaps this is worth a followup issue / PR (and maybe a test case)?

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,248 @@ Snap Events {#snap-events}
...
+	{{snapchanging}} events should fire before {{scroll}} events.
+
+	Each {{Document}} has an associated list of
+	<dfn export for=Document>pending snapchanging event targets</dfn>, initially empty.
+
+	Each
+	<a spec=css-scroll-snap lt="scroll snap container">snap container</a> has
+	one <dfn>snapchangingTargetBlock</dfn> (which is either null or an {{Element}})
+	and one <dfn>snapchangingTargetInline</dfn> (which is either null or an
+	{{Element}}) in the block and inline axes respectively.
```
flackr, line 428: nit: I think you only need to say this once. E.g.

	Each
	<a spec=css-scroll-snap lt="scroll snap container">snap container</a> has
	a <dfn>snapchangingTargetBlock</dfn> a <dfn>snapchangingTargetInline</dfn>
	for the block and inline axes respectively which can either be null if not currently snapping in that axis, or an
	{{Element}} to which the scroller is snapping.

css-scroll-snap-2/Overview.bs
flackr: There's still a tab on line 325

cssom-view-1/Overview.bs
DavMila: Sounds good, I'd prefer that too.

flackr (approved):
Overall this looks good to me. I think there may still be an ordering issue but we can pursue fixing that separately.

css-scroll-snap-2/Overview.bs
DavMila: Thanks!

css-scroll-snap-2/Overview.bs
DavMila: Oops, fixed.

DavMila (commented):
Fixed the space issue and eliminated superfluous wording in snapchang{ed,ind}Target{Block,Inline} definitions.

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,250 @@ Snap Events {#snap-events}
...
-				readonly attribute boolean smoothlyScrolled;
-			};
 
-			[Exposed=Window]
-			interface SnapTargetList {
-			    readonly attribute SnapTargetArray x;
-			    readonly attribute SnapTargetArray y;
+	<h4 for="snapchanged" id="snapchanged"> snapchanged </h4>
+	{{snapchanged}} indicates that the snap area to which a snap container is snapped along either axis has changed.
+	{{snapchanged}} should be dispatched:
```
tabatkins, line 306:
```suggestion
	{{snapchanged}} is dispatched:
```

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,250 @@ Snap Events {#snap-events}
...
+		1. If |snapcontainer| is not already in <var>doc</var>'s
+			<a>pending snapchanged event targets</a>:
+			1. Append |snapcontainer| to <var>doc</var>'s
+				<a>pending snapchanged event targets</a>.
+
+	Note: When snapping occurs on a scroller (either due to a layout change or a
+	 scrolling operation) the <a>snapchangingTargetBlock</a> and <a>snapchangingTargetInline</a>
+	 associated with that scroller are updated and represent the current snap targets
+	 of that scroller. This allows the <a>update snapchanged targets</a> algorithm
+	 to use these elements to determine whether a {{snapchanged}} event should be fired.
```
tabatkins, line 375:
```suggestion
	scrolling operation) the <a>snapchangingTargetBlock</a> and <a>snapchangingTargetInline</a>
	associated with that scroller are updated and represent the current snap targets
	of that scroller. This allows the <a>update snapchanged targets</a> algorithm
	to use these elements to determine whether a {{snapchanged}} event should be fired.
```

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,250 @@ Snap Events {#snap-events}
...
+	* during a scrolling operation, if the element to which a
+		 <a spec=css-scroll-snap lt="scroll snap container">snap container</a> would
+		 <a spec="css-scroll-snap-1" lt="scroll snap">snap</a> (in either axis) changes, or
+	* if a [[css-scroll-snap-1#re-snap|layout change]] occurs such that a {{snapchanged}} event
+		 is to be dispatched. In this case, as with the scrolling case, the {{snapchanging}} event
+		 should be dispatched before the {{snapchanged}} event.
+
+	A scrolling operation may animate towards a particular position (e.g.
+	scrollbar arrow clicks, arrow key presses, "behavior: smooth" programmatic
+	scrolls) or may directly track a user's input (e.g. touch scrolling, scrollbar
```
tabatkins, line 397:
```suggestion
	A scrolling operation might animate towards a particular position (e.g.
	scrollbar arrow clicks, arrow key presses, "behavior: smooth" programmatic
	scrolls) or might directly track a user's input (e.g. touch scrolling, scrollbar
```

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,250 @@ Snap Events {#snap-events}
...
+
+	{{snapchanging}} aims to let the web page know, as early as possible,
+	that the scrolling operation will result in a change in the element the snap
+	container is snapped to. The user agent should evaluate whether to trigger
+	{{snapchanging}} based on the	<a>eventual snap target</a> to which the scroller would
+	<a spec="css-scroll-snap-1" lt="scroll snap">snap</a> were the scrolling operation
+	to reach its intended scroll position.
+
+	Note: Since snapchanging gives the web page hints about future snapping,
+	the snapping hinted at by a snapchanging event may not materialize since it
```
tabatkins, line 415:
```suggestion
	the snapping hinted at by a snapchanging event might not materialize since it
```

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,250 @@ Snap Events {#snap-events}
...
+	{{snapchanging}} aims to let the web page know, as early as possible,
+	that the scrolling operation will result in a change in the element the snap
+	container is snapped to. The user agent should evaluate whether to trigger
+	{{snapchanging}} based on the	<a>eventual snap target</a> to which the scroller would
+	<a spec="css-scroll-snap-1" lt="scroll snap">snap</a> were the scrolling operation
+	to reach its intended scroll position.
+
+	Note: Since snapchanging gives the web page hints about future snapping,
+	the snapping hinted at by a snapchanging event might not materialize since it
+	will be	possible for subsequent scrolling input to further alter the snap
```
tabatkins, line 416:
```suggestion
	will be possible for subsequent scrolling input to further alter the snap
```

css-scroll-snap-2/Overview.bs
```diff
@@ -275,119 +275,250 @@ Snap Events {#snap-events}
...
+	<a spec="css-scroll-snap-1" lt="scroll snap">snap</a> were the scrolling operation
+	to reach its intended scroll position.
+
+	Note: Since snapchanging gives the web page hints about future snapping,
+	the snapping hinted at by a snapchanging event might not materialize since it
+	will be possible for subsequent scrolling input to further alter the snap
+	container's scroll position and result in a different eventual snap target.
+
+
+	{{snapchanging}} events should fire before {{scroll}} events.
```
tabatkins, line 420:
```suggestion
	{{snapchanging}} events are fired before {{scroll}} events.
```

PR #9516, opened 2023-10-23
[css-values-4] Centralized and generalized definition of fr unit in CSS Values 4

### LINKED ISSUE #8027: [css-values-4] Generalize definition of `<flex>`
With the [addition of the `stripes()` function](https://github.com/w3c/csswg-drafts/pull/7029), [`<flex>` values](https://w3c.github.io/csswg-drafts/css-grid-1/#typedef-flex) are now used outside of the Grid spec.

So as discussed in the PR linked above, the definition of `<flex>` should be generalized and moved to the Values & Units spec. It might also be renamed to `<fraction>` as [suggested](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1104458371) by @LeaVerou.

Sebastian

SebastianZ:
As the `fr` unit is now used in several different places, I moved it to CSS Values 4.

To make it clearer that we're talking about fractions of leftover space, the data type also got renamed to `<fraction>`.

This fixes #8027.

Sebastian


css-grid-1/Overview.bs
```diff
@@ -1511,7 +1511,7 @@ Explicit Track Sizing: the 'grid-template-rows' and 'grid-template-columns' prop
 
 		<dfn>&lt;track-size></dfn>          = <<track-breadth>> | minmax( <<inflexible-breadth>> , <<track-breadth>> ) | fit-content( <<length-percentage [0,∞]>> )
 		<dfn>&lt;fixed-size></dfn>          = <<fixed-breadth>> | minmax( <<fixed-breadth>> , <<track-breadth>> ) | minmax( <<inflexible-breadth>> , <<fixed-breadth>> )
-		<dfn>&lt;track-breadth></dfn>       = <<length-percentage [0,∞]>> | <<flex [0,∞]>> | min-content | max-content | auto
+		<dfn>&lt;track-breadth></dfn>       = <<length-percentage [0,∞]>> | <<fraction [0,∞]>> | min-content | max-content | auto
```
cdoublev, line 1514: I think you can remove `[0,∞]` since `<fraction>` is defined to only accept non-negative values, like [`<resolution>`](https://drafts.csswg.org/css-values-4/#resolution) eg. in [`resolution`](https://drafts.csswg.org/mediaqueries-5/#descdef-media-resolution).

css-grid-1/Overview.bs
```diff
@@ -1876,45 +1875,6 @@ Flexible Lengths: the ''fr'' unit</h4>
 	Each column or row's share of the <a>leftover space</a> can be computed as the column or row's
 	<code>&lt;flex> * &lt;leftover space> / &lt;sum of all <a>flex factors</a>&gt;</code>.
```
cdoublev, line 1876: I think `&lt;flex>` should be `&lt;fraction>`

css-grid-1/Overview.bs
```diff
@@ -4902,6 +4858,10 @@ Changes since the <a href="https://www.w3.org/TR/2020/CRD-css-grid-1-20201218/">
 				instead use the <a>grid area</a> determined in [[#abspos-items]].</del>
 			</blockquote>
 
+		<li id="change-2020-fr-unit">
+			Moved the definition of <<fraction>> to [[css-values-4#fractions]],
```
cdoublev, line 4862: I think it should be *replaced `<flex>` with `<fraction>` and moved its definition to ...*.

css-values-4/Overview.bs
```diff
@@ -2918,6 +2918,75 @@ Combination of <<position>></h4>
...
+██████   ████████  ██     ██  ██          ██     ██  ██     ██ ██ ██ ██
+██       ██   ██   █████████  ██          ██     ██  ██     ██ ██  ████
+██       ██    ██  ██     ██  ██    ██    ██     ██  ██     ██ ██   ███
+██       ██     ██ ██     ██   ██████     ██    ████  ███████  ██    ██
+-->
+
+<h3 id="fractions">
+Flexible lengths: the <<fraction>> type</h3>
+
+	Fraction values are non-negative <<dimension>>s denoted by
```
cdoublev, line 2934: I would start with *Fraction values are denoted by `<fraction>`, and indicates a value that is some fraction of another reference value...*, borrowing the definition of [`<percentage>`](https://drafts.csswg.org/css-values-4/#percentages), then by emphasizing the difference between `<fraction>` and `<percentage>`.

Then I would say: *A fraction is a [dimension](https://drafts.csswg.org/css-values-4/#dimension)*, borrowing the definition of `<length>`.

css-images-4/Overview.bs
```diff
@@ -1941,7 +1941,7 @@ Color Stop “Fixup”</h4>
 	<pre class=prod>
 		&lt;image-1D> = <<stripes()>>
 		<<stripes()>> = stripes( <<color-stripe>># )
-		<<color-stripe>> = <<color>> && [ <<length-percentage>> | <<flex>> ]?
+		<<color-stripe>> = <<color>> && [ <<length-percentage>> | <<fraction>> ]?
```
cdoublev, line 1944: Unrelated to this PR but iirc, `0` et `∞` are always unambiguous and do not need a following unit, however `100` requires to be followed by `%` and `<length-percentage>` requires `[0,∞]`.

Simplified:

  > **`<length-percentage [0,∞]`**
  >
  > Negative values are invalid.
  >
  > Percentage thicknesses are relative to the *total width*.

cdoublev: Never mind for the unitless `100`: it is unambiguous in `<percentage>`.

css-values-4/Overview.bs
cdoublev: And finally with *the allowed range of `<fraction>` values always excludes negative values, in addition to any explicit ranges that might be specified*, borrowing the definition of `<resolution>`.

css-grid-1/Overview.bs
LeaVerou, line 4862:
```suggestion
			Replaced <<flex>> with <<fraction>> and moved the definition to [[css-values-4#fractions]],
```

css-values-4/Overview.bs
LeaVerou: Technically the math *could* support negative fractions, though not sure there are enough use cases to be worth the complexity.

tabatkins: I'm not sure what the meaning of a negative fraction would be; even if there is a consistent "just follow the numbers" behavior one could assign to it, I wouldn't want to make it valid without a reasonable argument that it's actually doing something useful and understandable.

css-grid-1/Overview.bs
SebastianZ: Done.

css-grid-1/Overview.bs
SebastianZ: Right. Done.

css-grid-1/Overview.bs
SebastianZ: Yep. Changed.

css-values-4/Overview.bs
SebastianZ: I've rephrased the description.

@LeaVerou Please open a separate issue for discussion negative fractions if you see a use case for them.

The change here is only meant to centralize and generalize the definition for the existing usages.

css-images-4/Overview.bs
SebastianZ: Please open an issue so this can be discussed separately! FWIW, I think it's fine as it is. Values greater than 100% don't make much sense as a 100% wide color band already takes up the total width.

css-grid-1/Overview.bs
SebastianZ: Done.

css-grid-2/Overview.bs
```diff
@@ -5503,6 +5459,11 @@ Changes since the <a href="https://www.w3.org/TR/2020/CRD-css-grid-2-20201218/">
 				<del>However, if the <a>grid container</a> parent is also the generator of the absolutely positioned element's <a>containing block</a>,
 				instead use the <a>grid area</a> determined in [[#abspos-items]].</del>
 			</blockquote>
+
+		<li id="change-2020-fr-unit">
+			Moved the definition of <<fraction>> to [[css-values-4#fractions]],
```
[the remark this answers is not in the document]
SebastianZ: Done.

cdoublev (approved):

SebastianZ:
@tabatkins @fantasai @LeaVerou Can I get a review for this, please?

Sebastian

css-grid-1/issues-cr-2016.html
```diff
@@ -292,7 +292,7 @@ <h1>CSS Grid Layout Level 1 Disposition of Comments for 2016-09-29 CR</h1>
 <span>Resolved: <a href='https://lists.w3.org/Archives/Public/www-style/2017Apr/0018.html'>https://lists.w3.org/Archives/Public/www-style/2017Apr/0018.html</a> <a href="https://www.w3.org/Mail/flatten/index?subject=https%3A%2F%2Flists.w3.org%2FArchives%2FPublic%2Fwww-style/2017Apr/0018.html"><abbr title="Thread">&#x2208;</abbr></a></span></pre>
 <pre class=' a' id='issue-32'>
 <span>Issue 32. <a href='#issue-32'>#</a></span>
-<span>Summary:  Computed value for single value &lt;flex>?</span>
+<span>Summary:  Computed value for single value &lt;fraction>?</span>
```
fantasai, line 295: Don't change descriptions of past discussions!

fantasai (changes requested):
I would not make changes to css-grid-1 or css-grid-2, since they're in CR already; and if we're introducing a generic definition, it should go into css-values-5, because it's new and hasn't quite stabilized yet.

I'm also unsure about `<fraction>` vs `<flex>`; we've been using `<flex>` for awhile now because it helps tie the idea to the way flexbox works (which was the goal)... it's short for “flex fraction” in that spec.  But that's just bikeshedding. :)

css-grid-1/issues-cr-2016.html
SebastianZ: Search & replace mistake. Thanks for noticing! Fixed now.

SebastianZ:
> I would not make changes to css-grid-1 or css-grid-2, since they're in CR already; and if we're introducing a generic definition, it should go into css-values-5, because it's new and hasn't quite stabilized yet.

Ok, I reverted the changes to css-grid-1 and css-grid-2 and moved the definition to css-values-5.

> I'm also unsure about `<fraction>` vs `<flex>`; we've been using `<flex>` for awhile now because it helps tie the idea to the way flexbox works (which was the goal)... it's short for “flex fraction” in that spec. But that's just bikeshedding. :)

While Flexbox refers to different types of "flex fractions", `<flex>` is _not_ part of Flexbox. It is part of Grid, which calls it "flex fractions".

The [confusion with Flexbox](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1104451018) is a great indicator that it should be renamed (which was [initially suggested](https://github.com/w3c/csswg-drafts/pull/7029#issuecomment-1104458371) by @LeaVerou).

Also, there might be other use cases in which the relation to flexible lengths could be hindering. E.g. it could be used to indicate fractions of angles, times or frequencies at some point.

Sebastian

css-values-5/Overview.bs
```diff
@@ -111,6 +111,80 @@ Request URL Modifiers</h4>
...
+	Its unit identifier is <dfn value for="<flex>" lt="fr|fr unit">fr</dfn>.
+
+	These values specify
+	<dfn export lt="flexible length">flexible lengths</dfn>.
+	Each length defines a portion of a given space
+	which is specified by the context the type is used in.
+
+	The allowed range of <<fraction>>> values
+	<em>always</em> excludes negative values,
+	in addition to any explicit ranges that might be specified.
```
fantasai, line 144: Let's actually define what this thing does...

```suggestion
Proportional Fractions: the <<fraction>> type</h3>

	<dfn id=flex-fraction>Proportional fractions</dfn> are denoted by <dfn id="flex-value">&lt;fraction></dfn>,
	which is <<dimension>> with the <dfn value for="<flex>" lt="fr|fr unit">fr</dfn>unit.
	They represent a fraction of a reference value
	(typically a length)
	as a proportion in relation to other [=flex fractions=] referencing that same value,
	i.e. a value of ''<var>N</var>fr'' represents <var>N</var> / max(<var>sum of all ''fr'' values</var>, 1).
	For example, leftover space in [=grid layout=]
	is distributed to a ''1fr'' track and a ''2fr'' track in a 1:2 ratio.

	The allowed range of <<fraction>>> values
	<em>always</em> excludes negative values,
	in addition to any explicit ranges that might be specified.
```

SebastianZ: Thanks! Reads better now.

SebastianZ:
@tabatkins @fantasai @LeaVerou Could you please do another review? Would be great to finish this up to avoid further merge conflicts and to open up usages of the `fr` unit in more places.

Sebastian

cdoublev:
Fwiw, `<fraction>` needs to be listed as a [base type](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-base-type) in order to be [valid](https://drafts.csswg.org/css-values-4/#calc-type-checking) in math functions, but it seems to be used in properties that are not implemented, so it may be ok to leave it without a note that it should be treated the same as `<flex>` in math functions, I guess.

SebastianZ:
Thank you for the hint, @cdoublev! `<flex>` got _renamed_ to `<fraction>` in this PR besides being generalized. So CSS Typed OM also needs to be adjusted. Therefore, I've created https://github.com/w3c/css-houdini-drafts/pull/1119 now.

Sebastian

cdoublev:
From an implemetation point of view, parsing `<fraction>` as a `<flex>`, which will remain in CSS Grid 1-2, iiuc, is a simple solution that does not require updating CSS Values and CSS Typed OM, until `<flex>` is replaced CSS Grid 3.

PR #9523, opened 2023-10-25
[css-view-transitions-2] Apply new @view-transition rule syntax

### LINKED ISSUE #9383: [css-view-transitions-2] Decide on naming for navigation-triggered view-transition rule
For context, see [TPAC discussion](https://github.com/w3c/csswg-drafts/issues/8048#issuecomment-1719751618) as well of the rest of the conversation at #8048.

The resolution was that we would have an at-rule that defines the relationship between a navigation and a view-transition - for start, it would allow you to specify that any same-origin navigation results in triggering a view transition.
The final naming was left open.

Any naming proposal would have to:
- convey that this is about navigation-triggered view-transitions
- start as "same-origin", in a way that's extendible later to a URL, a pair of URLs, or some other form of conditioning

Current suggestions (but open for more!):

1. `@view-transition same-origin { trigger: navigation }`
2. `@view-transition-behavior same-origin { trigger: navigation }`
3. `@navigation-behavior same-origin { view-transition: trigger }`
4. `@nav-transition same-origin { behavior: trigger }`
5. `@view-transition navigation(same-origin) { behavior: trigger }`
6. `@auto-view-transition same-origin { trigger: navigation }`

My personal favorite which I've put in the [draft spec](https://github.com/w3c/csswg-drafts/pull/9382) is (3), as it allows extending in the future to other "navigation behaviors" if we so choose, while not being too generic, and tells the story of `if (navigation) { startViewTransition()}`.

noamr:
- Use the `@view-transition { navigation: auto | none; type: ... }` syntax as per the resolution
- `type: none` means that type is unspecified, needs a resolution
- reloads are ignored as per previous resolution
- Renamed the JS param "type" instead of "types"
- Currently types are resolved both at old and new document separately, this needs a resolution.

Closes #9383

See resolution: https://github.com/w3c/csswg-drafts/issues/9383#issuecomment-1779644377

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -118,7 +118,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	1. Once it's time to [=unload=] the old document, if the navigation is [=same origin=]
 		and the old {{Document}} has opted in to cross-document view-transitions, the old state is captured.
 
-	1. An event named {{RevealEvent|reveal}} is fired on the new {{Document}}, with a `viewTransition` property,
+	1. An event named {{PageRevealEvent|reveal}} is fired on the new {{Document}}, with a `viewTransition` property,
 		which is a {{ViewTransition}} object. This {{ViewTransition}}'s <code>{{ViewTransition/updateCallbackDone}}</code> is already resolved,
 		and its [=captured elements=] are populated from the old {{Document}}.
 
@@ -153,7 +153,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		- Opt-in to navigation-triggered view-transitions in both pages.
 		- Pass the click location to the new document, e.g. via {{WindowSessionStorage/sessionStorage}}.
-		- Intercept the {{ViewTransition}} object in the new document, using the {{RevealEvent}}.
+		- Intercept the {{ViewTransition}} object in the new document, using the {{PageRevealEvent}}.
 
 		In both pages:
 		```css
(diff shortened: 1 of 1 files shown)


noamr:
cc @bokand

css-view-transitions-2/Overview.bs
```diff
@@ -263,72 +263,81 @@ document.startViewTransition({update: updateTheDOMSomehow});
...
+			redirects, and is not a {{NavigationType/reload}}.
 	</dl>
 
+## The [=@view-transition/type=] descriptor ## {#view-transition-type-descriptor}
 
+	<pre class='descdef'>
+	Name: type
+	For: @view-transition
+	Value: <<custom-ident>>+
+	Initial: n/a
```
vmpstr, line 309: In `@font-face`, there are  coupe of properties that specify N/A as initial value, but I think the reason they can do that is that the `@font-face` rule is invalid unless those properties are specified (`font-family` and `src`). The other properties can be omitted so they have an initial value (like `font-weight`). 

Here, `type` is an optional thing, so I think it needs to have an initial value. WDYT? Maybe we can make Value be something like `none | <<custom-ident>>+`. It makes it a bit awkward for JS since we should then make `none` an invalid type

But all of this is a very special `@-rule` stuff, not the regular cascade, so maybe this is ok too

vmpstr (approved):

css-view-transitions-2/Overview.bs
vmpstr: Discussed offline, I think this is fine.

css-view-transitions-2/Overview.bs
```diff
@@ -405,44 +416,57 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
 
-		1. Fire a new event named <code>reveal</code> on |document|'s [=relevant global object=],
-			using {{RevealEvent}}.
+		1. Let |matchingRule| be the last ''@view-transition'' rule in |document| which has a ''@view-transition/navigation'' descriptor whose [=computed value=] is ''@view-transition/navigation/auto''.
 
-		1. If |transition| is not null, then [=activate view transition|activate=] |transition|.
+		1. If |matchingRule| is not found, then return "<code>not found</code>".
 
-		1. Set |document|'s [=document/ready to render fired=] to true.
+		1. If |matchingRule| contains a ''@view-transition/type'' descriptor, then return a [=list=] of strings corresponding to that descriptor's [=computed value=].
```
khushalsagar, line 458: @noamr an empty list vs null has the same effect right? Based on my reading [here](https://drafts.csswg.org/css-view-transitions-2/#ref-for-viewtransition-active-types:~:text=viewTransition%E2%80%99s%20active%20types%20contains%20at%20least%20one%20of%20the%20%3Ccustom%2Dident%3E%20values%20of%20the%20%3Cvt%2Dtype%2Dselector%3E.) it does so just confirming.

css-view-transitions-2/Overview.bs
```diff
@@ -405,44 +416,57 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
-	When {{Document}} |document| is <dfn>ready to render</dfn>:
-		1. If |document|'s [=document/ready to render fired=] is true, then return.
+	<div algorithm="page reveal">
+		To <dfn>reveal</dfn> {{Document}} |document|:
+			1. If |document|'s [=document/page pagereveal fired=] is false, then:
+				1. Let |transition| be the result of [=activating cross-document view-transition=] for |document|.
 
-		1. Let |transition| be the result of getting the [=inbound cross-document view-transition=] for |document|.
+				1. [=Fire an event=] named <code>pagereveal</code> at |document|'s [=relevant global object=],
+					using {{PageRevealEvent}}, with [=PageRevealEvent/view transition=] initialized to .
```
khushalsagar, line 442: I think you meant "initialized to the [=inbound cross-document view-transition=]".

css-view-transitions-2/Overview.bs
noamr: Yes, though there is no way to specify an empty list.

css-view-transitions-2/Overview.bs
noamr: Oops, should be |transition|. Will follow up.

PR #9527, opened 2023-10-26
[css-view-transitions-2] A few minor edits

### LINKED ISSUE #8888: [css-view-transitions-2] Hold paint of old Document until new Document draws a frame
During a cross-document navigation, there is a duration when the active Document associated with the browsing context (tab) has been swapped. But there is no visual content from the new Document because it hasn't produced a frame yet. The browser has a few options for what to paint during this duration:

1. Keep the last frame from the old Document onscreen. This has security implications if the navigation is cross-origin.
2. Clear the screen immediately and paint any UA defined color.
3. Display cached rendering of the new Document (if available). This is possible for session history navigations where the browser can cache the rendering of a Document when navigating away from it.

View Transitions require 1 since by design we want the new Document to define the transition animations. The new Document starts by displaying images captured from the old Document's last frame. This is done via the pseudo DOM defined [here](https://drafts.csswg.org/css-view-transitions-1/#setup-transition-pseudo-elements-algorithm). Until the new Document has produced a frame, the browser needs to paint the last frame from the old Document to avoid visual glitches.

There are also no security implications to this since View Transitions are limited to same-origin.

Proposed Resolution: During a cross-document View Transition, the browser must continue to paint the last frame from the old Document until the new Document produces its first frame.

The behavior for such navigations when there is no View Transition is left to the UA (same as current behaviour).

@smaug---- @noamr

noamr:
* Specify paint holding
* Fix missing |transition| var
* Allow an empty list in the type descriptor

Closes #8888

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -524,7 +524,9 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
 
 			Note: The process continues in [=setup view transition=], via [=perform pending transition operations=].
 
-		1. The user agent should display the currently displayed frame until |newDocument| is [=revealed=], or until its [=active view transition=]'s [=ViewTransition/phase=] is "`done`".
+		1. The user agent should display the currently displayed frame until either:
+			* |newDocument| is [=revealed=]
+			* its [=active view transition=]'s [=ViewTransition/phase=] is "`done`".
 
 			Note: this is to ensure that there are no unintended flashes between displaying the old and new state, to keep the transition smooth.
 	</div>
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -174,7 +174,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		In the new page:
 		```js
 		// This would run both on initial load and on reactivation from BFCache.
-		addEventListener("reveal", async event => {
+		addEventListener("pagereveal", async event => {
 			if (!event.viewTransition)
 				return;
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -523,6 +520,8 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
 		1. Set |oldDocument|'s [=active view transition=] to |outboundTransition|.
 
 			Note: The process continues in [=setup view transition=], via [=perform pending transition operations=].
+
+		1. The user agent should display the currently displayed frame until |newDocument| is [=revealed=], or until its [=active view transition=]'s [=ViewTransition/phase=] is "`done`".
```
khushalsagar, line 524: Trying to reason about why we need "or until its [=active view transition=]'s [=ViewTransition/phase=] is "`done`". Is this for the case where the transition is skipped before reveal? I think that can only happen in the reveal algorithm [here](https://drafts.csswg.org/css-view-transitions-2/#document-reveal:~:text=Let%20transition%20be%20the%20result%20of%20activating%20cross%2Ddocument%20view%2Dtransition%20for%20document.) anyway which is right before we hit "revealed".

khushalsagar, line 524: Can you add a non-normative note here clarifying why this is needed.

noamr: If the document navigates away or an SPA nav is initiated.

khushalsagar:
Leaving a note about activating the transition after dispatching the reveal event. So authors have a script hook to set up VT names.

css-view-transitions-2/Overview.bs
noamr: Done

khushalsagar: Ah ok cool. Slight suggestion on wording:

The user agent should display the currently displayed frame until either:
- |newDocument| is [=revealed=]
- its [=active view transition=]'s [=ViewTransition/phase=] is "`done`"

Non-normative note to document the cases you mentioned for the second condition.

khushalsagar (approved):

css-view-transitions-2/Overview.bs
khushalsagar: LGTM % this comment.

PR #9549, opened 2023-11-01
[cssom-view-1] Add more checkVisibility dictionary properties

### LINKED ISSUE #9474: [cssom-view] API to expose skipped content state
Currently there is no straightforward way to query the skipped content state, despite browsers having this info internally.

In the context of web extensions, it would be useful if you could query that state without invoking the layout of the element, which takes away from the content-visibility optimizations.

@vmpstr @rwlbuis

### LINKED ISSUE #9487: [cssom-view] checkVisibility options have inconsistent naming schemes
We have checkOpacity and checkVisibilityCSS. Can we add a CSS suffix to `checkOpacity` ? Or rename to: `checkOpacityProperty` / `checkVisibilityProperty`

cc @josepharhar @vmpstr @smfr

vmpstr:
[cssom-view-1] Add more checkVisibility dictionary properties

Fixes #9474
Fixes #9487

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1161,6 +1161,9 @@ dictionary ScrollIntoViewOptions : ScrollOptions {
 dictionary CheckVisibilityOptions {
     boolean checkOpacity = false;
     boolean checkVisibilityCSS = false;
+    boolean contentVisibilityAuto = false;
+    boolean opacityProperty = false;
+    boolean visibilityProperty = false;
 };
 
 partial interface Element {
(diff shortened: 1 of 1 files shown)


nt1m:
@vmpstr Can you rename the checkContentVisibilityAuto option from: https://github.com/web-platform-tests/wpt/pull/42829 as well? I added these tests only tentatively.

vmpstr:
Yep, doing that as part of the implementation. Thank you for pointing this out

nt1m:
Adding @emilio & @smfr since they are spec editors.

cssom-view-1/Overview.bs
```diff
@@ -1246,15 +1249,25 @@ Note: The {{DOMRect}} object returned by {{Element/getBoundingClientRect()}} is
     1. If an ancestor of |this| in the [=flat tree=]
         has ''content-visibility: hidden'',
         return false.
-    1. If the {{CheckVisibilityOptions/checkOpacity}} dictionary member of |options|
-        is true,
+    1. If either the {{CheckVisibilityOptions/checkOpacity}}
+        or the {{CheckVisibilityOptions/opacityProperty}}
```
nt1m, line 1253: I would list `opacityProperty` first, since it's the new name.

cssom-view-1/Overview.bs
```diff
@@ -1246,15 +1249,25 @@ Note: The {{DOMRect}} object returned by {{Element/getBoundingClientRect()}} is
...
+        dictionary members of |options|
+        are true,
         and |this|, or an ancestor of |this| in the [=flat tree=],
         has a computed 'opacity' value of ''0'',
         return false.
-    1. If the {{CheckVisibilityOptions/checkVisibilityCSS}} dictionary member of |options|
-        is true,
+    1. If either the {{CheckVisibilityOptions/checkVisibilityCSS}}
+        or the {{CheckVisibilityOptions/visibilityProperty}}
```
nt1m, line 1260: ditto

cssom-view-1/Overview.bs
```diff
@@ -1161,6 +1161,9 @@ dictionary ScrollIntoViewOptions : ScrollOptions {
 dictionary CheckVisibilityOptions {
     boolean checkOpacity = false;
```
smfr, line 1162: Please add a `Note` that says that `checkVisibilityCSS` and `checkOpacity` are deprecated.

cssom-view-1/Overview.bs
vmpstr: Done.

cssom-view-1/Overview.bs
vmpstr: Done.

cssom-view-1/Overview.bs
vmpstr: We haven't resolved to deprecate these. I've added a note that says they are historical names that have aliases that match the new naming scheme. I suspect we want to revisit at CSSWG if we want to deprecate the names though.

emilio (approved):

cssom-view-1/Overview.bs
emilio: I think it's fine to call them deprecated IMO. I don't feel strongly either way tho. Assuming @smfr is fine with the current wording feel free to merge.

PR #9552, opened 2023-11-02
[selectors-4] Replace <an+b> with <an-b>

### LINKED ISSUE #9494: [css-selectors-4] Use actual types in nth pseudo class selectors
The syntax for nth pseudo class selectors currently use a simplified syntax which later references actual types in the prose.  It would be simpler and better for tooling which parses this data to use the actual types.

For example: 
`:nth-child(An+B [of S]? )`

Would be represented as:
`:nth-child(<an-b> [of <complex-real-selector-list>]? )`

If this is acceptable I'd be happy to supply a patch.

See also https://github.com/w3c/csswg-drafts/issues/9473.

cdoublev:
Fixes #9494. 

Depends on #9480.

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -2957,7 +2957,7 @@ Child-indexed Pseudo-classes</h3>
 	If the selector list is omitted,
 	it defaults to ''*|*''.
 
-	The <var>An+B</var> notation and its interpretation
+	Note: The <var>An+B</var> notation and its interpretation
 	are defined in [[css-syntax-3#anb-microsyntax]];
 	it represents any index <var>i</var> = <var>A</var><var>n</var> + <var>B</var>
 	for any non-negative integer <var>n</var>.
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -3045,14 +3043,14 @@ Child-indexed Pseudo-classes</h3>
...
 '':nth-last-child()'' pseudo-class</h4>
 
-	The <dfn id='nth-last-child-pseudo' lt=":nth-last-child()">:nth-last-child(<var>An+B</var> [of <var>S</var>]? )</dfn>
+	The <dfn id='nth-last-child-pseudo' lt=":nth-last-child()">:nth-last-child(<<an-b>> [of <<complex-real-selector-list>>]?)</dfn>
 	pseudo-class notation represents elements that
 	are among <var>An+B</var>th elements
 	from the list composed of
-	their <a>inclusive siblings</a> that match the <a>selector list</a> <var>S</var>,
+	their <a>inclusive siblings</a> that match the <a>selector list</a>,
+	which is a <<complex-real-selector-list>>,
```
fantasai, line 3051:
```suggestion
	their <a>inclusive siblings</a> that match the <a>selector list</a> <<complex-real-selector-list>>,
```

or 

```suggestion
	their <a>inclusive siblings</a> that match the <<complex-real-selector-list>>,
```

as mentioned above.

fantasai (changes requested):
A few comments but also a question about bikeshedding: as we're getting further away from the actual form of the notation "2n+1", it's getting more confusing to refer to the notation. An+B was somewhat understandable, because it pattern-matches against the notation. But `<an-b>` just seems like weird 1970s computer garbage naming.

Some possible replacements:
* `<An-B>` at least helps it parse as three components (a, n, b), rather than two (an, b).
* `<ith>` or `<index>` or something like that
* other ideas?

UPDATE: Filed the bikeshedding into https://github.com/w3c/csswg-drafts/issues/9473#issuecomment-1819819647

cdoublev:
Marking this PR as a draft while there are discussions to rename `<an-b>`.

cdoublev:
a9805fb

PR #9553, opened 2023-11-02
[css-color-4] Fix LCH alpha interpolation example

waacton:
The example LCH interpolation with premultiplied alpha at https://www.w3.org/TR/css-color-4/#ex-premultiplied-lch un-premultiplies the hue component, which I believe is wrong according to:

> To obtain a color value from a premultipled color value ... each component which had been premultiplied is divided by the interpolated alpha value.


svgeesus:
Oh gosh, yes! Thanks for spotting!

waacton:
I've not contributed before, so apologies if I've missed any other steps I needed to have taken!

svgeesus (approved):

PR #9570, opened 2023-11-08
[css-shapes-2] Make shape-padding consistent with shape-margin

### LINKED ISSUE #7352: [css-shapes-2] Value definition of `shape-padding` is missing `none`
The initial value of [`shape-padding`](https://drafts.csswg.org/css-shapes-2/#shape-padding-property) is `none` but it is missing from its `value` definition.

> **Name:** `shape-padding`
> **Value:** `<length>`
> **Initial:** `none`

cdoublev:
Fixes #7352.

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -637,13 +637,20 @@ The 'shape-padding' Property</h4>
 
 	<pre class='propdef'>
 		Name: shape-padding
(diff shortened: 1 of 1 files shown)


fantasai (approved):
I think maybe it would be a good idea to also update the `shape-margin` definition in this spec to match the `shape-margin` in shapes-1. :)

PR #9575, opened 2023-11-08
[css-view-transitions-1] Update spec for resolved issues

khushalsagar:


svgeesus:
You are an editor. If the issues are resolved you can commit directly.

PR #9589, opened 2023-11-10
[css-values-5] Minor syntax fixes

### LINKED ISSUE #8384: [css-values-5] Minor problems on `toggle()` and `random-*()` productions
**[`toggle()`](https://drafts.csswg.org/css-values-5/#funcdef-toggle)**

`toggle()` cannot be extracted by `w3c/reffy` because it is not defined in a production rule. I do not know if you want to declare it as a function or type. I do not know what difference it makes.

---

**[`random-item()`](https://drafts.csswg.org/css-values-5/#funcdef-random-item)**

I think `<any-value>` is inappropriate in `<random-item()>` and should be replaced with `<declaration-value>?`, because `<any-value>` allows top-level `;` and because custom properties accept empty inputs (like in `<mix()>`, see #8556):

`<random-item()> = random-item( <random-caching-options> ';' <any-value> [ ';' <any-value> ]* )`

---

**[`random()`](https://drafts.csswg.org/css-values-5/#funcdef-random)**

Finally, the trailing `;` in the production for `<random()>` seems to be a typo:

  > `<random()> = random( <random-caching-options>? , <calc-sum>, <calc-sum>, [by <calc-sum>]? );`

cdoublev:
Fixes #8384.

- defines `toggle()` in a production rule
- removes the trailing semicolon in `random()` production
- replaces `<any-value> ; ...`/`<declaration-value> ; ...` with `<declaration-value>? ; ...` since `<any-value>` eats semicolons, and custom properties accept empty inputs

### DIFF
--- a/css-values-5/Overview.bs
+++ b/css-values-5/Overview.bs
@@ -596,7 +596,9 @@ Toggling Between Values: ''toggle()''</h3>
 
 	The syntax of the ''toggle()'' expression is:
 
-	<pre>toggle( <<whole-value>> [ ';' <<whole-value>> ]+ )</pre>
+	<pre class=prod>
+		<dfn><<toggle()>></dfn> = toggle( <<whole-value>> [ ';' <<whole-value>> ]+ )
+	</pre>
 
 	Note: This [=functional notation=] uses semicolons to separate arguments
 	rather than the more typical comma
@@ -980,7 +982,7 @@ Generating a Random Numeric Value: the ''random()'' function</h3>
 	optionally limiting the possible values to a step between those limits:
 
 	<pre class=prod>
-	&lt;random()> = random( <<random-caching-options>>? , <<calc-sum>>, <<calc-sum>>, [by <<calc-sum>>]? );
+	&lt;random()> = random( <<random-caching-options>>? , <<calc-sum>>, <<calc-sum>>, [by <<calc-sum>>]? )
 
 	<dfn><<random-caching-options>></dfn> = <<dashed-ident>> || per-element
 	</pre>
@@ -1142,7 +1144,7 @@ Picking a Random Item From a List: the ''random-item()'' function</h3>
 	from among its list of items.
 
 	<pre class=prod>
-	&lt;random-item()> = random-item( <<random-caching-options>> ';' <<any-value>> [ ';' <<any-value>> ]* )
+	&lt;random-item()> = random-item( <<random-caching-options>> ';' <<declaration-value>>? [ ';' <<declaration-value>>? ]* )
 	</pre>
 
 	The <em>required</em> <<random-caching-options>>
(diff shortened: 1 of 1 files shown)


css-values-5/Overview.bs
```diff
@@ -549,7 +549,7 @@ Selecting the First Supported Value: the ''first-valid()'' notation</h3>
 	as the whole value of the property it's used in.
 
 	<pre class=prod>
-		<dfn>&lt;first-valid()></dfn> = first-valid( <<declaration-value>> [ ';' <<declaration-value>> ]* )
+		<dfn>&lt;first-valid()></dfn> = first-valid( <<declaration-value>>? [ ';' <<declaration-value>>? ]* )
```
tabatkins, line 552: Nah, this was intended. The only place an empty value is valid is in untyped custom properties, and you can't actually use first-valid() in such properties (it's just seen as part of the substitution value, not evaluated).

tabatkins (changes requested):
r+ after the first-valid() change is reverted

css-values-5/Overview.bs
cdoublev: > you can't actually use `first-valid()` in such properties

I assume this will be added to the spec later, and that *untyped custom properties* means custom properties not registered with `@property`, or registered with the universal `syntax` value. 

Does it mean that `first-valid()` must be invalid for the `initial-value` of custom properties declared with the universal syntax?

  > (it's just seen as part of the substitution value, not evaluated)

Sorry, I do not understand.

tabatkins: I mean that in `--foo: first-valid(foo, bar);`, the UA does not see a `first-valid()` function there. It just sees several uninterpreted tokens. The only syntax that is actually processed in an unregistered custom property is `var()`. The `first-valid()` won't be processed until it's substituted into either a real property or a registered custom property with a non-universal grammar. (Which is precisely the value here - we lost the ability to do `prop: old-val; prop: new-val;` if you're using vars - `--foo: old-val; --foo: new-val; prop: var(--foo);` doesn't work. But `--foo: first-valid(new-val, old-val); prop: var(--foo);` works like the first example.)

tabatkins: `first-valid()` is perfectly fine for use in for custom properties; that's its *primary use-case*, in fact. It's just not useful *for custom properties themselves*, as in, for selecting the first valid that is valid *for the custom property itself*.

cdoublev: Ah yes, it will be processed when substituting a property or a typed custom property. I clearly get the point about its usage now, thanks. I was bothered and stuck by the idea that an empty value cannot be a *first valid* value of an untyped custom property, whereas it is valid value according to its grammar.

I assume that untyped custom properties declared with `first-valid()` should always serialize unresolved, whereas other properties declared with `first-valid()` should serialize with the first valid value resolved at computed value time.

tabatkins: Correct.

PR #9602, opened 2023-11-15
[css-view-transitions-2] Tighten view-transition rule resolving

### LINKED ISSUE #9534: [css-view-transitions-2] resolve on descriptor/parameter names
The current [editor's draft](https://drafts.csswg.org/css-view-transitions-2/) defines a few new names.
We should resolve whether they're the right names or should be changed.

Current spec:
```css
/* css */
@view-transition {
  navigation: auto | none;
  type: foo bar;  
}

:root:active-view-transition(foo) {
  img.hero {
    view-transition-name: hero;
  }
}
```

```js
/* js */
document.startViewTransition({
  update: () => { ... },
  type: ["foo", "bar"]
});

viewTransitionRule.type // DOMTokenList
```

**Proposed Resolution**
- The navigation-type descriptor is `navigation`, and its current possible values are `auto` and `none`.
- The type descriptor is `type`. Its default value is an empty list. There is no `none` value or so.
- The JS parameter for `startViewTransition` as well as the property name in the CSSOM rule can be `typeList`, to be consistent with `classList`, as it receives some sort of iterable. Alternatively we can keep it as `type` or use `types`.
- Bonus: decide whether the types can be regular idents vs. using dashed-idents for future compatibility with keywords.

noamr:
- `types: none` should resolve to having no types
- `navigation: none` should always skip transition.

Closes #9534

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -305,8 +305,8 @@ Note: as per default behavior, the ''@view-transition'' rule can be nested insid
 	<pre class='descdef'>
 	Name: type
 	For: @view-transition
-	Value: <<custom-ident>>*
-	Initial: an empty list
+	Value: none | <<custom-ident>>*
+	Initial: none
 	</pre>
 
 	The '<dfn for="@view-transition">type</dfn>' descriptor sets the [=ViewTransition/active types=] for the transition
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -453,11 +453,19 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
 	<div algorithm>
 		To get the <dfn>resolve @view-transition rule</dfn> for a {{Document}} |document|:
 
-		1. Let |matchingRule| be the last ''@view-transition'' rule in |document| which has a ''@view-transition/navigation'' descriptor whose [=computed value=] is ''@view-transition/navigation/auto''.
+		1. Let |matchingRule| be the last ''@view-transition'' rule in |document|.
```
khushalsagar, line 456: I tripped on this for a moment, figured we need to filter rules based on which ones apply to the navigation. But right now all rules would apply since its a binary opt-in.

khushalsagar (approved):

css-view-transitions-2/Overview.bs
vmpstr: I was wondering about the same thing. The only thing I'm not sure of is how "reloads" are skipped. Is that documented elsewhere in this spec?

bokand: Right now we don't have to do any matching logic since there's just `auto` and `none`. Since `none` "always applies" (as per discussion on chat), and `auto` covers everything but reloads, they basically both always apply and we break the tie by picking the last one. In a reload it doesn't matter since there's currently no way to turn it on?

Once we break apart `auto` into more keywords and add `reload` this will need to be more involved.

> The only thing I'm not sure of is how "reloads" are skipped. Is that documented elsewhere in this spec?

[Here](https://drafts.csswg.org/css-view-transitions-2/#setting-up-the-view-transition-in-the-old-document:~:text=If%20navigationType%20is%20reload%2C%20then%20return)? Though that'll have to change once we add a `reload` type.

noamr: Yes exactly

PR #9604, opened 2023-11-16
[css-cascade-5] page and view-transition at rules should cascade

### LINKED ISSUE #9601: [css-view-transitions-2] @view-transition resolution algorithm should cascade
The @view-transition resolution algorithm is currently [specifies](https://drafts.csswg.org/css-view-transitions-2/#resolving-the-view-transition-rule):

> To get the resolve @view-transition rule for a [Document](https://dom.spec.whatwg.org/#document) document:
> 1. Let matchingRule be the last [@view-transition](https://drafts.csswg.org/css-view-transitions-2/#at-view-transition-rule) rule in document which has a navigation descriptor whose [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value) is [auto](https://drafts.csswg.org/css-view-transitions-2/#valdef-view-transition-navigation-auto).
> 2. If matchingRule is not found, then return "not found".
> 3. Return a [list](https://infra.spec.whatwg.org/#list) of strings corresponding to that descriptor’s [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value).

But I think it needs to account for the CSS cascade. Take this example:

```html
<style>
  @view-transition {
    navigation: auto;
    type: foo;
  }

  @import url('style.css') layer(A);
</style>
```
```css
  /*style.css*/
  @view-transition {
    navigation: auto;
    type: bar;
  }
```

I would expect the `foo` transition to be resolved, despite the fact that the `bar` transition is "last".

I think the way to do this will be to keep a map of all encountered `@view-transition` rules, keyed by their matcher descriptors (currently just `navigation` but later including `from`, `to`, etc. with succeeding levels of the cascade overriding a entries where the matchers are equal.

---

While we're here:

> Return a [list](https://infra.spec.whatwg.org/#list) of strings corresponding to that descriptor’s [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value).

I'm assuming "that descriptor" is `navigation`?

noamr:
Mention that `@view-transition` and `@page` at-rules respect CSS layer when computing order/specificity.

Closes #9601

### DIFF
--- a/css-cascade-5/Overview.bs
+++ b/css-cascade-5/Overview.bs
@@ -1253,6 +1253,11 @@ Cascade Layers</h3>
 	that are defined inside [=cascade layers=]
 	also use the layer order when resolving name collisions.
 
+	Other [=at-rules=] which rely on [=cascading=]
+	such as ''@page'' or ''@view-transition''
+	that are defined inside [=cascade layers=]
+	also use the layer order when determining order and [=specificity=].
+
 	<div class="example">
 		For example,
 		authors could override the animation from a framework,


fantasai:
Drafted an alternative that uses a bit more general wording at https://github.com/w3c/csswg-drafts/pull/13798

@tabatkins @mirisuzanne Thoughts?

PR #9608, opened 2023-11-17
[css-view-transitions-2] UA may apply implementation-defined timeout o cross-document view transition

### LINKED ISSUE #9155: [css-view-transitions-2] Abort transition if the navigation takes too long
This came up on the [TAG review](https://github.com/w3ctag/design-reviews/issues/851#issuecomment-1663075082). If the navigation is taking longer than a certain threshold, it's better to just show the new content directly instead of running an animation. The exact time can be left to the UA.

noamr:
As per CSSWG resolution.
Closes #9155

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -517,16 +517,24 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
 					We explicitly clear it here since the old Document may be cached by the UA.
 
 			1. [=Queue a global task=] on the [=DOM manipulation task source=] given |newDocument|'s [=relevant global object=],
-				to perform the following step:
+				to perform the following steps:
 
-				1. Let |newDocument|'s [=active view transition=] be a new {{ViewTransition}} in |newDocument|'s [=relevant Realm=],
+				1. Let |inboundTransition| be a new {{ViewTransition}} in |newDocument|'s [=relevant Realm=],
 					whose [=ViewTransition/named elements=] is |outboundTransition|'s [=ViewTransition/named elements=],
 					[=ViewTransition/initial snapshot containing block size=] is |outboundTransition|'s [=ViewTransition/initial snapshot containing block size=],
 					and whose [=ViewTransition/is inbound cross-document transition=] is true.
 
-				1. [=Call the update callback=] for |newDocument|'s [=active view transition=].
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -517,17 +517,25 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
-				1. Let |newDocument|'s [=active view transition=] be a new {{ViewTransition}} in |newDocument|'s [=relevant Realm=],
+				1. Let |inboundTransition| be a new {{ViewTransition}} in |newDocument|'s [=relevant Realm=],
 					whose [=ViewTransition/named elements=] is |outboundTransition|'s [=ViewTransition/named elements=],
 					[=ViewTransition/initial snapshot containing block size=] is |outboundTransition|'s [=ViewTransition/initial snapshot containing block size=],
 					and whose [=ViewTransition/is inbound cross-document transition=] is true.
 
-				1. [=Call the update callback=] for |newDocument|'s [=active view transition=].
+				1. Let |newDocument|'s [=active view transition=] be |inboundTransition|.
+
+				1. [=Call the update callback=] for |inboundTransition|.
```
vmpstr, line 529: I know this is the same as above, but is this just to resolve the promises since there is no update callback specified above when we declare what `inboundTransition` is

css-view-transitions-2/Overview.bs
```diff
@@ -517,17 +517,25 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
 
 				1. Call |onReady|.
 
+				1. At any given time, the UA may decide to skip |inboundTransition|, e.g. after an implementation-defined timeout, by running the following steps:
```
vmpstr, line 533: We likely need a better term than "At any given time". For example, it can't do it during layout for example, so it's only at a time that a script task can run? I'm not sure if idle time is the right thing or if there is a better term

noamr: Script tasks can run during layout. But we can make it so that the steps queue a task.

css-view-transitions-2/Overview.bs
noamr: Yes

vmpstr (approved):
new wording lgtm

PR #9609, opened 2023-11-17
[css-view-transitions-2] cross-document transition preceeds startViewTransition on outgoing page

### LINKED ISSUE #9512: [css-view-transitions-2] Starting a same-doc view transition while a cross-doc view transition is pending
The current behavior in the spec is that starting any view transition cancels the active one, which means that the following edge cases might have unexpected results:

- A capture is initiated for a document navigation and script initiates a transition before the document unloads. should we ignore `document.startViewTransition()` calls during a navigation initiated transition?

- On the new Document, script initiates transition before the navigation initiated transition could finish (before `pagereveal`). Given that the navigation initiated transition will be before the first frame of the new Document was rendered, should we ignore it and let the MPA transition go through?

@khushalsagar @bokand @vmpstr

noamr:
When starting a transition in an old page and there's a capture ongoing for a cross-document view transition, the programatically-created transition is automatically skipped, calling its callback and resolving the promises on the next task. The `ready` promise is rejected.

As per resolution.
Closes #9512


css-view-transitions-1/Overview.bs
vmpstr, line 1: Can you add the changes made here into the appendix as well "changes since <date>" thing at the bottom

vmpstr (approved):

PR #9611, opened 2023-11-17
[css-view-transitions-1] Refactor skip during cross-doc transition on old Document

khushalsagar:


css-view-transitions-1/Overview.bs
```diff
@@ -949,22 +949,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |document| be [=this's=] [=relevant global object's=] [=associated document=].
 
-		1. If |document|'s [=auto-skip view transitions=] is true, then:
-
-		1. [=Queue a global task=] on the [=DOM manipulation task source=],
-			given |transition|'s [=relevant global object=],
-			to execute the following steps:
-
-			1. [=Call the update callback=] for |transition|.
```
vmpstr, line 958: We still need this, no?

noamr (approved):
Much better, thanks!

css-view-transitions-1/Overview.bs
noamr: I think skip transition calls this

khushalsagar: skip the view transition is supposed to handle that here: https://drafts.csswg.org/css-view-transitions-1/#skip-the-view-transition:~:text=If%20transition%E2%80%99s%20phase%20is%20before%20%22update%2Dcallback%2Dcalled%22.

But I just realized we're not initializing phase.

PR #9615, opened 2023-11-20
[css-anchor-position-1] Re-create `anchor()` function definition

tidoust:
The definition was inadvertently dropped in:
https://github.com/w3c/csswg-drafts/commit/471d90fc0d0063bc267540266bb7adacbf013df2

Without the definition, Bikeshed links to the definition in the TR version of the spec, but that definition will disappear next time the spec is updated.


tabatkins:
Ah whoops, thanks!

PR #9620, opened 2023-11-22
[css-logical] Add `block-start` and `block-end` values to `caption-side` property

ramiy:
While documenting the `caption-side` property on MDN we noticed that the flow-relative `block-start` and `block-end` values are not defined by any spec, although they are mentioned in the spec.

- [Cascading Style Sheets Level 2 Revision 2](https://drafts.csswg.org/css2/#propdef-caption-side) spec defines the `top` | `bottom` | `inherit` values.
- [CSS Logical Properties and Values Level 1](https://drafts.csswg.org/css-logical/#caption-side) spec defines the `inline-start` | `inline-end` values.

The `block-start` and `block-end` values are mentioned but not really defined in the spec.

### DIFF
--- a/css-logical-1/Overview.bs
+++ b/css-logical-1/Overview.bs
@@ -204,11 +204,11 @@ Logical Values for the 'caption-side' Property</h3>
 
   <pre class="propdef partial">
     Name: caption-side
-    New values: inline-start | inline-end
+    New values: inline-start | inline-end | block-start | block-end
     Computed value: specified keyword
   </pre>
 
-  These two values are added only for implementations that support
+  These two inline values are added only for implementations that support
   ''caption-side/left'' and ''caption-side/right'' values for 'caption-side'.
   The ''caption-side/left'' and ''caption-side/right'' values themselves
   are defined to be <a>line-relative</a>.


css-logical-1/Overview.bs
```diff
@@ -204,11 +204,11 @@ Logical Values for the 'caption-side' Property</h3>
 
   <pre class="propdef partial">
     Name: caption-side
-    New values: inline-start | inline-end
+    New values: inline-start | inline-end | block-start | block-end
     Computed value: specified keyword
   </pre>
 
-  These two values are added only for implementations that support
+  These two inline values are added only for implementations that support
```
SebastianZ, line 211: The two inline values...

SebastianZ (changes requested):
Logical Properties 1 says this:

> The existing `top` and `bottom` values are idiosyncratically redefined as assigning to the block-start and block-end sides of the table, respectively.

So those actually are flow-relative, even when they refer to physical sides.

Though I think it makes sense to add them as aliases to the existing `top` and `bottom` keywords. Therefore, I have created #9623 to discuss their addition.

Sebastian

Loirooriol (changes requested):
This needs to be discussed in https://github.com/w3c/csswg-drafts/issues/9623

But anyways, if we end up adding `block-star` and `block-end`, it doesn't make sense to condition them on supporting `left` and `right`.

css-logical-1/Overview.bs
ramiy: It refers to `inline-start` and `inline-end`.

Loirooriol: Very confusing to start with "these" if you are only referring to some of them, something like this is clearer:

```suggestion
  The ''caption-side/inline-start'' and ''caption-side/inline-end'' values
  are added only for implementations that support
```

ramiy: Much better, thank you.

ramiy:
[Cascading Style Sheets (CSS) Level 2](https://drafts.csswg.org/css2/#caption-position) - defines only `top` | `bottom` | `inherit`

image


[CSS Logical Properties and Values Level 1](https://drafts.csswg.org/css-logical/#caption-side) - defines only `inline-start` | `inline-end`

image

No spec defines `block-start` and `block-end`.

PR #9630, opened 2023-11-23
[css-view-transitions-1] Add a note about why view-transition-name is animatable

### LINKED ISSUE #9616: [css-values-4] Advanced simplification of calc-operator nodes
Maybe returning a result *in its canonical unit* should not be mandatory?

  > 4. If `root` is an operator node that’s not one of the calc-operator nodes, and all of its calculation children are numeric values with enough information to compute the operation root represents, return the result of running `root`’s operation using its children, **expressed in the result’s canonical unit**.

https://drafts.csswg.org/css-values-4/#simplify-a-calculation-tree

It prevents simplifying `min(1em)` as `calc(1em)` (related: #9559), `clamp(0em, 2em, 1em)` as `calc(1em)`, `round(1.1em, 1em)` as `calc(1em)`, etc.

Chrome/FF currently simplify `min(max(1%, -1%))` as `max(1%, -1%)`. There does not seem to be any corresponding WPT tests. Support for other math functions is still low so I have not tested them.

noamr:
As per resolution
Closes #9616

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -519,6 +519,11 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	Animation type: discrete
 	</pre>
 
+	Note: though 'view-transition-name' is [=discrete|discretely animatable=], animating it doesn't
+	affect the running view transition. Rather, it's a way to set its value in a way that can change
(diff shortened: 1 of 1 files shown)


khushalsagar:
Just make sure bikeshed is happy before merging. :)

Issue #9233, opened 2023-08-24
[css-text-decor] [css-pseudo] text-decoration-line:: spelling-error inside ::spelling-error pseudo (and grammar)


schenney-chromium:
It was resolved in https://github.com/w3c/csswg-drafts/issues/7522 to ignore all other text-decoration properties when the `text-decoration-line`property has value `spelling-error` or `grammar-error`.

>When ''spelling-error'' or ''grammar-error'' apply, the user agent <em>must</em> disregard the other <a>sub-properties</a> of 'text-decoration', as well any other properties typically affecting the appearance of line decorations (such as 'text-underline-position', 'color', 'stroke', or 'fill') when rendering these decorations.

A side effect of this change is that
```
::spelling-error { text-decoration-color: blue; }
```
has no effect because the default `text-decoration-line` for `::spelling-error` is `spelling-error` (as one would hope).

Same applies to `::grammar-error`.

I propose that we accept this as working as intended, and update the spec to clearly say that `::spelling-error` will only respect text decoration properties when `text-decoration-line` is present with a value other than `spelling-error` or `grammar-error`. 
```
::spelling-error {
   text-decoration-line: underline;
   text-decoration-style: wavy;
   text-decoration-color: blue;
}
```

schenney-chromium:
@fantasai is the intent to say, yes, this is working as intended and update the spec as necessary?

schenney-chromium:
PR to do the edit: https://github.com/w3c/csswg-drafts/pull/9633

PR #9633, opened 2023-11-25
[css-pseudo-4] Clarify text decorations in ::spelling-error and ::grammar-error

schenney-chromium:
Issue #9233
Add a note for authors on the requirements for custom text-decorations inside ::spelling-error pseudo (and grammar).


fantasai (changes requested):
This is way too long for the problem it's solving.

fantasai:
I switched up the rules to use `text-decoration` instead of `text-decoration-line` to kind-of nudge people to use the shorthand also when trying to overwrite it. We could add a note, but this behavior should be very explicitly specified in the definition of these values; so I'm not sure how much we really need a note in this appendix here. But definitely we should keep it minimal if we add anything.

schenney-chromium:
I think something is needed because authors are quite likely to overlook this, just setting a `text-decoration-color` and expecting the UA rendered dots or squiggles to pick up the color, at least given the way Chrome renders error markers.

I simplified it to be more direct and moved it up to the section discussing properties in highlight pseudos.

This is basically a bandaid on the disconnect between `::spelling-error` defined here and `text-decoration-line: spelling-error` defined in the text-decor spec.

Issue #7693, opened 2022-09-05
[cssom-view] No event to track window position


frivoal:
[The window object exposes](https://drafts.csswg.org/cssom-view/#extensions-to-the-window-interface) attributes to find the position (and size) of the window. However, there are no corresponding event, so anyone interested in tracking these must do so by actively polling, which is wasteful and/or laggy, depending on how frequently you poll.

A use case for doing this is when a parent page pops up an child window, and wants to keep it positioned to a particular spot relative to the parent window. That's currently doable, but it takes active polling. Being able to switch to an event would be good.

We already do have the resize event on the Window object, but that only solves half the problem, as there's no equivalent for the position.

So, can we get

```js
window.addEventListener('move', callback);
```
to go along with

```js
window.addEventListener('resize', callback);
```
?

----
cc: @emilio @mrego

Loirooriol:
AFAIK  the `resize` event can easily be awful for perf since it's sync and fires lots of times during the resize.
It seems `move` would have the same problem. An async observer may be a better idea?

emilio:
Resize doesn't fire sync afaik (though it fires very frequently). In any case the other potential issue I see is that the window resize event fires everywhere (like iframes) but we probably don't want that here. Resize reflects inner size, but it seems here you are interested in outer window position (not inner one).

frivoal:
An async observer would be equaly fine for the usecase, and preferable if that enables better performance.

The point about iframes is interesting. I suppose tracking the inner position would be a security issue as it would leak information about the parent content, but I don't think tracking the outer window position would be an issue, would it? If it is, I suppose it's an acceptable restriction for this to just not work from within iframes.

Loirooriol:
I guess in iframes it should just track changes to the `screenLeft` or `screenTop` of the iframe's contentWindow? Which seem to be the same values as for the top window.

alice:
> A use case for doing this is when a parent page pops up an child window, and wants to keep it positioned to a particular spot relative to the parent window. 

I had a play with doing something like this, and discovered that not being able to make an always-on-top popup window (is there a way to do this that I've missed) makes the tracking popup case pretty limited in usefulness. (It might be useful if you have a popup that doesn't overlap its parent, I guess?)

https://people.igalia.com/alice/test/popup/

Also, writing the logic to position the popup relative to the window was *extremely* involved, and it still only sort of works, and seems to be subtly different between browsers and operating systems.

How realistic is a tracking popup window as a use case?

[...]

PR #9664, opened 2023-11-30
[cssom-view] Add 'onmove' event handler for the Window object

javifernandez:
Define a partial interface WindowEventHandller to integrate the changes that would eventually be moved to the html spec. A new element 'onmove' is added to this interface. 

See cssom-view [issue 7693](https://github.com/w3c/csswg-drafts/issues/7693) for details.


frivoal:
CSSWG resolved to give up on this feature. Closing. https://github.com/w3c/csswg-drafts/issues/7693#issuecomment-2842594228

PR #9676, opened 2023-12-04
[css-view-transitions-1] Defer restoring persisted state until after capturing old state

noamr:
This fixes an issue that the spec restores the persisted state (e.g. scroll position) too eagerly, e.g. when calling `document.startViewTransition` inside a `popstate` handler, not giving the UA an opportunity to first capture the old state.

This exports an algorithm that should be called from the HTML spec when ready to automatically restore persisted state, e.g. after `popstate`. If the algorithm returns true, proceed as normal, otherwise the state will be restored after capturing the old state.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -47,6 +47,9 @@ spec:css-display-4; type: dfn; text:invisible;
 urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	text: NavigateEvent
 	text: signal; for: NavigateEvent; url: #ref-for-dom-navigateevent-signal①
+urlPrefix: https://html.spec.whatwg.org/multipage/; spec: html; type: dfn;
+	text: restore persisted state; url: #restore-persisted-state
+	text: latest entry; url: #latest-entry
 </pre>
 
 <script async type="module" src="diagrams/resources/scaler.js"></script>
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -47,6 +47,9 @@ spec:css-display-4; type: dfn; text:invisible;
 urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	text: NavigateEvent
 	text: signal; for: NavigateEvent; url: #ref-for-dom-navigateevent-signal①
+urlPrefix: https://html.spec.whatwg.org/multipage/; spec: html; type: dfn;
+	text: restore persisted state; url: #restore-persisted-state
+	text: latest entry; url: #latest-entry
```
domenic, line 52: Instead of poking into HTML's externals, try to get https://github.com/whatwg/html/pull/9996 to export the appropriate parts.

(I'm a bit scared of exporting "latest entry" though, since it's a pretty low-level detail... document's node navigable's active session history entry would be a more public API. Maybe it's the same thing in this case?)

css-view-transitions-1/Overview.bs
```diff
@@ -1094,6 +1097,18 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
 			Note: this is used for cross-document view transitions.
+
+		: <dfn>persisted state restoration</dfn>
+		:: One of the following, initially "`defer`":
+
+			1. "`defer`".
+			1. "`pending`".
+			1. "`immediate`".
+
+			Note: this is used to defer automatic restoration of persisted state, e.g. scroll position,
```
khushalsagar, line 1108: Can you add what immediate implies as well?

khushalsagar (commented):
Do you have the html spec PR which will use this? Would help to review that together with this.

noamr:
> Do you have the html spec PR which will use this? Would help to review that together with this.

https://github.com/whatwg/html/pull/9996

css-view-transitions-1/Overview.bs
```diff
@@ -1940,6 +1978,7 @@ Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230
 * Copy `color-scheme` from DOM element to ''::view-transition-group()''. See <a href="https://github.com/w3c/csswg-drafts/issues/9276">issue 9276</a>.
 * Expose [=auto-skip view transition=] for a {{Document}}, to allow having outbound cross-document transitions preceed programmatic view transiitons. see <a href="https://github.com/w3c/csswg-drafts/issues/9512">issue 9512</a>.
 * Add a note about why 'view-transition-name' should be animatable.
+* Export an algorithm and add mechanism to defer restoring scroll/persisted state.
```
khushalsagar, line 1981: Is there a link to an html or csswg issue where this was discussed?

PR #9688, opened 2023-12-08
[css-color-4] Fixed a few issues regarding hue interpolation

SebastianZ:
This fixes the issues I outlined in https://github.com/w3c/csswg-drafts/commit/7148b31cbc862faba3c1e6f6dcd481d2a06d47b7.

Sebastian

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -1085,12 +1085,12 @@ any achromatic color will have a [=powerless=] hue component.
 		to achieve an effect where you only <em>want</em> 
 		to interpolate certain components of a color.
 
-		For example, to animate a color to "grayscale", no matter what the color is,
+		For example, to interpolate a color to "grayscale", no matter what the color is,
 		one can interpolate it with ''oklch(none 0 none)''.
 		This will take the hue and lightness from the starting color,
-		but animate its chroma down to 0,
+		but interpolate its chroma down to 0,
 		rendering it into an equal-lightness gray
-		with a steady hue across the whole animation.
+		with a steady hue across the whole interpolation.
 
 		Doing this manually would require 
 		matching the hue and lightness of the starting color explicitly.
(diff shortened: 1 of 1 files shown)


svgeesus:
As I said at the time, I disagree that changing "animated" to "interpolated" helps.

We are talking about interpolating two colors A and B. At the same time, B is being animated such that the hue angle between A and B passes through 180 degrees.

Saying "if B is being interpolated" is tautological, of course it is being interpolated. Saying "if it is also being animated" makes it clearer what is happening..

svgeesus:
(The proposed re-arrangements of the equations are fine, btw)

SebastianZ:
> As I said at the time, I disagree that changing "animated" to "interpolated" helps.

Though when talking about "animating hues", you are disregarding interpolations in gradients, filters, or when mixing colors. Or can the word animation be used in those contexts as well and it is just my non-native English interpretation that it is wrong in that case?

Sebastian

svgeesus:
> Though when talking about "animating hues", you are disregarding interpolations in gradients, filters, or when mixing colors. 

No, not at all. I get the sense we are talking past each other here; I'm not really understanding your point and you are not really understanding mine.

SebastianZ:
> > Though when talking about "animating hues", you are disregarding interpolations in gradients, filters, or when mixing colors.
> 
> No, not at all. I get the sense we are talking past each other here; I'm not really understanding your point and you are not really understanding mine.

Ok, so let me try to articulate it differently. Is it ok to use the word "animate" when talking about interpolating a value in gradients or mixing colors? I.e. are "animate" and "interpolate" synonyms in that context? Because I associate the word "animate" or "animation" with some kind of movement.

Sebastian

svgeesus:
Okay, that tells me that what I was trying to convey hasn't come across. No, I am not using them as synonyms.

SebastianZ:
> Okay, that tells me that what I was trying to convey hasn't come across. No, I am not using them as synonyms.

Ok, so you intentionally used "animate". Then it's obviously just me misunderstanding the meaning of that word in that context. So I reverted the changes related to that and only kept the ones we agreed on.

Sebastian

svgeesus (approved):

PR #9692, opened 2023-12-10
[css-2023] Possible errata in four specification names

BartolomeSintes:
In [CSS 2023](https://www.w3.org/TR/css-2023/) sections 2.1, 2.2 and 2.3 the names of four specifications do not match with the name that appear in the specifications themselves. 
- CSS User Interface Module Level 3 -> CSS Basic User Interface Module Level 3 
- CSS Text Decoration Level 3 -> CSS Text Decoration Module Level 3 
- CSS Masking Level 1 -> CSS Masking Module Level 1
- CSS Transitions Level 1 -> CSS Transitions

### DIFF
--- a/css-2023/Overview.bs
+++ b/css-2023/Overview.bs
@@ -463,7 +463,7 @@ Modules with Rough Interoperability</h3>
 	We hope to incorporate them into the [[#css-official|official definition of CSS]] in a future snapshot.
 
 	<dl>
-		<dt><a href="https://www.w3.org/TR/css-transitions-1/">CSS Transitions</a>
+		<dt><a href="https://www.w3.org/TR/css-transitions-1/">CSS Transitions Level 1</a>
 		[[CSS-TRANSITIONS-1]]
 		and <a href="https://www.w3.org/TR/css-animations-1/">CSS Animations Level 1</a>
 		[[CSS-ANIMATIONS-1]].
--- a/css-2023/Overview.bs
+++ b/css-2023/Overview.bs
@@ -307,7 +307,7 @@ Classification of CSS Specifications</h2>
 		<dd>
 			Introduces a flexible linear layout model for CSS.
 
-		<dt><a href="https://www.w3.org/TR/css-ui-3/">CSS User Interface Module Level 3</a>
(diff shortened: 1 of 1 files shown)


dbaron:
I think this may be a sign that we should be more consistent with the specification titles used in the specifications.  I've updated `css-transitions-1` to include "Level 1" in the title.  (My personal preference is that we use "Module" less, and current specifications appear more mixed on whether or not to use "Module".)

BartolomeSintes:
Indeed, most of the specifications begin with CSS in their title, except Media Queries (both [Level 3](https://www.w3.org/TR/mediaqueries-3/) and [Level 4](Level)), [Selectors](https://www.w3.org/TR/selectors-3/), [Web Animations](https://www.w3.org/TR/web-animations-1/), [Resize Observer](https://www.w3.org/TR/resize-observer-1/) and [Filter Effects](https://www.w3.org/TR/filter-effects-1/).
By the way, apart from CSS Transitions, Level is not included in the title of the following specifications: [Style attributes](https://www.w3.org/TR/css-style-attr/), [Speech](https://www.w3.org/TR/css-speech-1/), [Animations](https://www.w3.org/TR/web-animations-1/) and [Resize Observer](https://www.w3.org/TR/resize-observer-1/).

dbaron:
I think all those cases omitting "CSS" are intentionally different, since those specifications are either things (like selectors and media queries) that can be used outside of CSS or APIs (like Resize Observer) that aren't part of the CSS language despite being related to CSS.

css-2023/Overview.bs
```diff
@@ -463,7 +463,7 @@ Modules with Rough Interoperability</h3>
 	We hope to incorporate them into the [[#css-official|official definition of CSS]] in a future snapshot.
 
 	<dl>
-		<dt><a href="https://www.w3.org/TR/css-transitions-1/">CSS Transitions Level 1</a>
+		<dt><a href="https://www.w3.org/TR/css-transitions-1/">CSS Transitions</a>
```
svgeesus, line 466:
```suggestion
		<dt><a href="https://www.w3.org/TR/css-transitions-1/">CSS Transitions Level 1</a>
```

svgeesus (approved):

Issue #9442, opened 2023-10-05
[css-zoom?] Zoom and replaced element intrinsic dimensions


emilio:
The zoom property has an effect on intrinsic element dimensions ([live](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cstyle%3E%0Ap%20%7B%20zoom%3A%202%20%7D%0A%3C%2Fstyle%3E%0A%3Cp%3E%3Cimg%20src%3Dimage%3E), [another demo](http://crisal.io/tmp/zoom-image-size.html)):

```html
<!DOCTYPE html>
<style>
p { zoom: 2 }
</style>
<p><img src=image>
```

That is fine and all, because it matches what you'd get if you specified the size with `width` and `height` (because those would go through the attribute mapping stuff etc).

But I don't think it's explained/defined in https://github.com/atanassov/css-zoom/ or so. What other side-effects does Zoom have? Does it apply to all replaced elements? Presumably it should, since their intrinsic size works in CSS pixels as well?

cc: https://github.com/w3c/csswg-drafts/issues/5623 @chrishtr @lilles @smfr @tabatkins @atanassov @zcorpan

chrishtr:
It does apply to all replaced elements. There is currently a bug that zoom from a document doesn't follow through to iframes, but I plan to fix it soon.

zcorpan:
Also needs to affects size of background images for web compat: https://bugzilla.mozilla.org/show_bug.cgi?id=390936#c106

Issue #9398, opened 2023-09-22
[css-zoom?] Zoom and CSSOM


emilio:
Zoom has some interesting behavior for the CSSOM, since it undoes some of its effects, but only in some APIs.

The original zoom draft by @atanassov makes sense to me:

 * You want all `getComputedStyle` things to be unzoomed, because you want to preserve `div.style.foo = getComputedStyle(div).foo` round-tripping.
 * You _might_ want some of the sizes to be unzoomed.
 * But you never want positions to be unzoomed, that makes no sense because the positions are meaningless if you cross a zoom boundary and you unzoom them.

However Blink and WebKit seem to unzoom in things like `getBoundingClientRect`, `ResizeObserver` rects, and so on. I don't think those make much sense for the same reason zooming positions doesn't make much sense.

My personal take is that, in general, the less APIs that account for zoom by unzooming, the better.

cc: https://github.com/w3c/csswg-drafts/issues/5623 @chrishtr @lilles @smfr @tabatkins @atanassov

chrishtr:
> However Blink and WebKit seem to unzoom in things like `getBoundingClientRect`, `ResizeObserver` rects, and so on. I don't think those make much sense for the same reason zooming positions doesn't make much sense.

I agree that it doesn't make sense and should change. My comment [here](https://github.com/w3c/csswg-drafts/issues/5623#issuecomment-1644208565) says basically the same thing you're saying here.

Issue #9397, opened 2023-09-22
[css-zoom?] Zoom and inheritance. 


emilio:
Zoom has some interesting behavior that makes no sense to me ([live](https://www.software.hixie.ch/utilities/js/live-dom-viewer/saved/12028)):

```
<!doctype html>
<div style="font-size: 16px; width: 100px; line-height: 16px;">
  <div style="background-color: green; font-size: inherit; line-height: inherit; width: inherit; height: 100px; zoom: 2;">What size am I?</div>
</div>
```

That for some reason ends up with:

 * `line-height` of `16px`
 * `font-size` of `32px`
 * `width` of `100px`
 * `height` of `200px`

That just makes no sense. This happens because the way `zoom` is implemented in browsers is by mutating the computed value, so inheritance doesn't account for zoom since it copies the un-zoomed computed value.

But then somehow `font-size` (but not `line-height`) is special, and gets recomputed on zoom changes??

cc: #5623 @chrishtr @lilles @smfr @tabatkins @atanassov

SelenIT:
Chrome 116 (M1 Mac) reports the following computed styles of this element:
<img width="368" alt="Screenshot 2023-09-25 at 00 24 35" src="https://github.com/w3c/csswg-drafts/assets/282754/3b97e7a0-1793-4cce-883a-8de9d85443ee">
It seems that font-size is unchanged, but all pixel values related to the parent (width and inherited line-height) get divided by the zoom factor. Since the "parent pixel" equals to half of the "zoomed pixel" of the element, treating all such values consistently makes some sense to me.

lilles:
To confirm this is not line-height specific I tried inheriting border-left-width explicitly, which also ends up with not applying zoom in Chrome.

lilles:
Implicit inheritance of line-height also doesn't work.

lilles:
And getComputedStyle() un-zooms based on the effective zoom which means it returns 8px for the inherited line-height/width etc. in the case above.

I'm not too concerned about fixing this for explicit inheritance of non-inherited properties in Blink, but adjusting inherited properties when the effective zoom changes will probably be expensive. Hopefully, changing the zoom property won't happen on many elements.

emilio:
Yeah, my preference would be not to adjust on _any_ inheritance (so, removing the font-size special-case).

chrishtr:
[...]

PR #9699, opened 2023-12-12
[css-viewport] [cssom-view-1] Add a definition of the `zoom` CSS property

### LINKED ISSUE #9644: [css-viewport] [css-zoom] Should zoom affect iframes?
Issue #9442 is about replaced element intrinsic dimensions, but what about zoom of iframe contents?

Right now, the CSS implementation in Blink at least does not affect iframes. I think we should have zoom affect at least same-origin iframes, since I predict developers would want that behavior for ease of styling. For cross-origin iframes, we could not do it if it's judged to be a privacy issue? (Or the developer could opt-in via something like #1771).

@emilio @zcorpan

chrishtr:
This PR resolves #9644, #9442, #9398 and #9397.

In particular, it defines zoom visual semantics as follows:
* A non-negative number that applies to elements and descendants. Nested zooms multiply. (The "effective zoom" is the product of the zooms of all ancestors.)
* Non-auto and non-percentage CSS lengths, including inherited lengths, are multiplied by effective zoom.
* Replaced elements' intrinsic sizes are resized according to effective zoom.
* Nested iframes are resized by the effective zoom of their containing `<iframe>` element

CSSOM:
* `getBoundingClientRect`, `getClientRects` and `IntersectionOserver` return rects that include zoom
* All other CSSOM APIs return values that do not include zoom
* `offsetParent` has a new quirk (stopping at ancestors with non-default `zoom`) that is analogous to what CSS transform does
* `offsetTop` excludes zoom of self, just like CSS transform
* `devicePixelRatio` is affected by zoom inherited from an ancestor frame

Web compat analysis: the above is what is already implemented in WebKit and Chromium browsers, except:
* `getBoundingClientRect`, `getClientRects` and `IntersectionOserver` are unzoomed
* Inherited properties other than `font-size` are unaffected
* Nested frames are not zoomed

Blink plans to make the above 3 compat-affecting changes soon, and I expect them to be web compatible.

A working design doc is [here](https://docs.google.com/document/d/1AcnDShjT-kEuRaMchZPm5uaIgNZ4OiYtM4JI9qiV8Po/edit).


css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
 }
 </pre>
 
+<h2 id='zoom-property'>
+The 'zoom' property</h2>
+
+An element becomes zoomed when the 'zoom' property has a positive computed value
+different than 1. 'zoom' affects computed property values and inherited
+property values. The computed value of 'zoom' is applied as a scalar to
+the used value &lt;length&gt;s CSS properties for the element and
```
fantasai, line 365: The correct Bikeshed markup is `<<length>>`. Same for percentage etc. throughout

css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
 }
 </pre>
 
+<h2 id='zoom-property'>
+The 'zoom' property</h2>
+
+An element becomes zoomed when the 'zoom' property has a positive computed value
```
fantasai, line 362: I would suggest adopting https://rhodesmill.org/brandon/2012/one-sentence-per-line/ and maybe also consider https://lists.w3.org/Archives/Public/spec-prod/2018OctDec/0011.html (last section)

css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
...
+Inherited: no
+Percentages: Refer to all &lt;length&gt; property values
+Media: visual
+Computed value: as specified
+</pre>
+
+The values of this property have the following meanings:
+
+<dl dfn-for="wrap-flow" dfn-type=value>
+  <dt><dfn lt="'zoom'!!'number'">number</dfn>
```
fantasai, line 399: Leave out the lt and put `<<number>>`. But fix the dfn-for to say "zoom"!

css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
...
+  <dd>
+    Positive floating point number indicating a zoom factor. Numbers smaller
+    than 1.0 indicate a "zoom out" or minification effect, while numbers
+    greater than 1.0 indicate a "zoom in" or magnification effect.
+
+  <dt><dfn lt="'zoom'!!&lt;percent&gt;">percent</dfn>
```
fantasai, line 405: same

css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
...
+.label {
+    background: lightgrey;
+    padding: 1em;
+    text-align: center;
+}
+&lt;/style&gt;
+</pre>
+
+  Illustration of the before and after hover state of the message box element.
+  <img src="css_zoom_hover_example.png" />
```
fantasai, line 441: This needs alt text.

css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
...
+The <dfn>effective zoom</dfn> of an element is the product of its computed
+value of 'zoom' and all ancestors' computed values of 'zoom'.
+
+The <dfn>scaled</dfn> value of a CSS length is the used value of that length;
+in particular it includes zoom.
+
+The <dfn>unscaled</dfn> value of a CSS length relative to an element is the
+<a>scaled</a> value divided by the element's <a>effective zoom</a>.
+
+<div class="note">Zoom applied to an iframe multiplies all its CSS lengths
```
fantasai, line 480: If it's a one-paragraph note, you can just prepend it with "Note: " and Bikeshed will handle the markup.

css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
...
+during layout, but is otherwise invisible to authors except via the observable
+width and height of the viewport in CSS units, and via
+{{Window/devicePixelRatio}}. In other words, the <a>effective zoom</a> of
+elements in an iframe are 1 unless 'zoom' is set on them directly or their
+ancestor in the same document.
+</div>
+
+<h3 id='dom--cssom-interactions'>
```
fantasai, line 488: I'd shorten this to `om-interactions`. Or maybe call it `zoom-om`.

css-viewport/Overview.bs
```diff
@@ -356,6 +356,154 @@ only the value previously set to it.
...
+</div>
+
+The <dfn>effective zoom</dfn> of an element is the product of its computed
+value of 'zoom' and all ancestors' computed values of 'zoom'.
+
+The <dfn>scaled</dfn> value of a CSS length is the used value of that length;
+in particular it includes zoom.
+
+The <dfn>unscaled</dfn> value of a CSS length relative to an element is the
+<a>scaled</a> value divided by the element's <a>effective zoom</a>.
```
fantasai, line 478: This works, but Bikeshed cross-references also work with `[=this=]` which I think Tab has been migrating towards...

fantasai (commented):
Quick pass on markup.

css-viewport/Overview.bs
tabatkins: I'm curious what spec you got this pattern from, @chrishtr, because it's an *extremely* old pattern that triggered some special behavior in the pre-Bikeshed processor (which Bikeshed now does by default). If there's still some of this hanging around the repo we need to fix it.

But yeah, fix your `dfn-for` attribute too. ^_^

css-viewport/Overview.bs
tabatkins: Yeah, plain dfn links should ideally be marked up as as `[=foo=]` for consistency.

css-viewport/Overview.bs
chrishtr: Done. As for where I got this old syntax from, it's because I started out with the text from Rossen's unofficial spec draft on his personal repo, which was from many years ago.

chrishtr:
> Quick pass on markup.

Thanks!

css-viewport/Overview.bs
```diff
@@ -356,6 +360,149 @@ only the value previously set to it.
...
+		Numbers smaller than 1.0 indicate a "zoom out" or minification effect,
+		while numbers greater than 1.0 indicate a "zoom in" or magnification effect.
+
+	<dt><dfn><<percentage>></dfn>
+	<dd>
+		Positive floating point number,
+		followed by a percentage character ("%") which indicates a zoom factor multiplied by 100.
+</dl>
+
+Negative or zero values for 'zoom' are illegal.
```
zcorpan, line 415: Negative value drops the declaration in WebKit and Chromium, but 0 doesn't.

http://software.hixie.ch/utilities/js/live-dom-viewer/saved/12260

Maybe instead zero should be defined to act as if it was 1 (or have no effect).

chrishtr: Good catch. Done.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
...
+
+<div class="example">
+	Example of the 'zoom' property applied during hover for magnification effect.
+
+	<pre>
+		&lt;div class="messageBox"&gt;
+			&lt;div class="label"&gt;Text of the label&lt;/div&gt;
+		&lt;/div&gt;
+
+		&lt;style type="text/css"&gt;
```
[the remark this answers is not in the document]
chrishtr: Done

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
...
+</div>
+
+<div class="example">
+	Example of replaced elements. In this example,
+	the image and iframe will be twice as large as their default sizing.
+
+	<pre>
+	&lt;div style="zoom: 2"&gt;
+		&lt;img src=""&gt;
+		&lt;iframe src=""&gt;
```
zcorpan, line 474:
```suggestion
		&lt;img src="a" alt=""&gt;
		&lt;iframe src="b"&gt;&lt;/iframe&gt;
```

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
 }
 </pre>
 
+<h2 id='zoom-property'>
+	The 'zoom' property
+</h2>
+
+An element becomes zoomed when the 'zoom' property has a positive computed value different than 1.
```
emilio, line 367: Or any flat tree ancestor?

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
...
+'zoom' affects computed property values and inherited property values.
+The computed value of 'zoom' is applied as a scalar to the used value of <<length>> CSS properties for the element and its descendants
```
emilio, line 369: We need to apply the "effective zoom" instead, not the computed value of zoom.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
...
+(including inherited properties such as 'line-height' or 'font-size'),
+plus intrinsic sizing of all replaced elements,
+background images,
+and nested frames (except for fenced frames [[!FENCED-FRAME]]),
+resulting in a magnification or minification effect.
+
+Nested values of 'zoom' multiply, resulting in additional scaling of <<length>> values.
```
emilio, line 376: Ok I guess you kinda fix it up here, but that's unfortunate.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
...
+
+The values of this property have the following meanings:
+
+<dl dfn-for="zoom" dfn-type=value>
+	<dt><dfn><<number>></dfn>
+	<dd>
+		Positive floating point number indicating a zoom factor.
+		Numbers smaller than 1.0 indicate a "zoom out" or minification effect,
+		while numbers greater than 1.0 indicate a "zoom in" or magnification effect.
+		A 0 value is treated as if it was 1.
```
emilio, line 408: Maybe note that this is a compatibility quirk.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
...
+
+	<dt><dfn><<percentage>></dfn>
+	<dd>
+		Positive floating point number,
+		followed by a percentage character ("%") which indicates a zoom factor multiplied by 100.
+		A 0 percentage is treated as if it was 100%.
```
emilio, line 414: Same here.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,151 @@ only the value previously set to it.
...
+
+<pre class="propdef">
+	Name: zoom
+	Value: <<number>> || <<percentage>>
+	Initial: 1
+	Applies to: all <<length>> property values of block-level elements
+	Inherited: no
+	Percentages: Refer to all <<length>> property values
+	Media: visual
+	Computed value: as specified
```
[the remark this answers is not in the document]
emilio: Right, this needs to convert percents to number.

cssom-view-1/Overview.bs
```diff
@@ -1190,7 +1190,7 @@ partial interface Element {
 The <dfn method for=Element>getClientRects()</dfn> method, when invoked, must return the result of the following algorithm:
 
 1. If the element on which it was invoked does not have an associated [=CSS/box=] return an empty {{DOMRectList}} object and stop this algorithm.
-1. If the element has an associated <a>SVG layout box</a> return a {{DOMRectList}} object containing a single {{DOMRect}} object that describes the bounding box of the element as defined by the SVG specification, applying the <a>transforms</a> that apply to the element and its ancestors.
+1. If the element has an associated <a>SVG layout box</a> return an [=unscaled=] {{DOMRectList}} object containing a single {{DOMRect}} object that describes the bounding box of the element as defined by the SVG specification, applying the <a>transforms</a> that apply to the element and its ancestors.
```
emilio, line 1193: Shouldn't this return scaled lengths?

cssom-view-1/Overview.bs
```diff
@@ -1141,7 +1141,7 @@ aborting on the first step that returns a value:
     1. Return the {{DOMRect}} object in <var>list</var> at index 0.
 1. If <a>caret node</a> is a text entry widget that is a replaced element,
     and that is in the document,
-    return a {{DOMRect}} object for the caret in the widget
+    return an [=unscaled=] {{DOMRect}} object for the caret in the widget
```
emilio, line 1144: This seems like it should return scaled lengths too, isn't these global client coordinates like getBoundingClientRect()?

cssom-view-1/Overview.bs
```diff
@@ -1596,11 +1597,11 @@ partial interface HTMLImageElement {
 };
 </pre>
 
-The <dfn attribute for=HTMLImageElement>x</dfn> attribute, on getting, must return the x-coordinate of the left <a>border edge</a> of the
+The <dfn attribute for=HTMLImageElement>x</dfn> attribute, on getting, must return the [=scaled=] x-coordinate of the left <a>border edge</a> of the
 first [=CSS/box=] associated with the element, relative to the <a>initial containing block</a> origin, ignoring any
 <a>transforms</a> that apply to the element and its ancestors, or zero if there is no [=CSS/box=].
```
emilio, line 1602: Is this the only case where we ignore transforms but not zoom out of curiosity? (Seems fine but...)

cssom-view-1/Overview.bs
```diff
@@ -1620,7 +1621,7 @@ containing a list of {{DOMRect}} objects in content order that matches the follo
 
 * For each element selected by the range, whose parent is not selected by the range, include the border areas returned by invoking {{Element/getClientRects()}} on the element.
 * For each {{Text}} node selected or partially selected by the range (including when the
-    boundary-points are identical), include a {{DOMRect}} object (for the part that is selected, not
+    boundary-points are identical), include an [=unscaled=] {{DOMRect}} object (for the part that is selected, not
```
emilio, line 1624: This seems like it should be scaled, it returns global coordinates.

emilio (changes requested):

css-viewport/Overview.bs
chrishtr: I added a parenthetical above, does that resolve the concern? Happy to rephrase to be more clear if you have a suggestion..

cssom-view-1/Overview.bs
chrishtr: Oops, fixed.

cssom-view-1/Overview.bs
chrishtr: Yes, that was a mistake. Fixed.

cssom-view-1/Overview.bs
chrishtr: Yes it is, to my knowledge. Unfortunate, but obscure and doesn't matter much.

css-viewport/Overview.bs
chrishtr: Oh you're referring to the computed value vs effective zoom...

chrishtr: I added a sentence about the used value, LMK if that helps.

cssom-view-1/Overview.bs
```diff
@@ -1141,7 +1141,7 @@ aborting on the first step that returns a value:
     1. Return the {{DOMRect}} object in <var>list</var> at index 0.
 1. If <a>caret node</a> is a text entry widget that is a replaced element,
     and that is in the document,
-    return a {{DOMRect}} object for the caret in the widget
+    return n [=scaled=] {{DOMRect}} object for the caret in the widget
```
zcorpan, line 1144:
```suggestion
    return a [=scaled=] {{DOMRect}} object for the caret in the widget
```

svgeesus:
@emilio how close is this to being merged? If it can get merged by Tues 23 I can include it in the Thurs 25 FPWD.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,157 @@ only the value previously set to it.
 }
 </pre>
 
+<h2 id='zoom-property'>
+	The 'zoom' property
+</h2>
+
+An element becomes zoomed when the 'zoom' property has a positive computed value different than 1 (or when a flat tree ancestor has zoom).
+'zoom' affects computed property values and inherited property values.
```
lilles, line 368: The zoom/effective-zoom doesn't affect any computed property values?

If I understand the expected effects correctly, it's a lot easier for me to think of it as an effect on the CSS reference pixel on an element basis. Then it doesn't really change cascading/ inheritance?

css-viewport/Overview.bs
```diff
@@ -356,6 +360,157 @@ only the value previously set to it.
...
+
+	The {{Element/getBoundingClientRect}},
+	{{Element/getClientRects}},
+	and {{IntersectionObserver}} APIs must return rects with [=scaled=]
+	lengths.
+
+	All other APIs related to element geometries must return [=unscaled=] lengths.
+	This is explained in detail in [[cssom-view#extensions-to-the-htmlelement-interface]].
+
+	The {{Window/devicePixelRatio}} of a frame is multiplied by the [=effective zoom=] inherited by its parent frame.
```
lilles, line 511: So that means the effective zoom of the iframe element affects the CSS reference pixel of the iframe document. I guess that means getBoundingClientRect() etc on elements inside the iframe will return px sizes for that new reference pixel?

css-viewport/Overview.bs
chrishtr: > The zoom/effective-zoom doesn't affect any computed property values?

It does. It just doesn't affect the serialized form returned by getComputedStyle().

> If I understand the expected effects correctly, it's a lot easier for me to think of it as an effect on the CSS reference pixel on an element basis.

You could think of it that way...

> Then it doesn't really change cascading/ inheritance?

Defining it that way might simplify the definition somewhat, but it's already only a couple of sentences. Is there something very unclear about how I've written it?

lilles: > > The zoom/effective-zoom doesn't affect any computed property values?
> 
> It does. It just doesn't affect the serialized form returned by getComputedStyle().

I don't think so. Then it would affect the output from typed OM, and I don't think we want that.

In Chrome's implementation the effective zoom factor is applied to px values on the ComputedStyle objects, and it also includes the device to css pixel ratio baked in. That's an implementation detail. It doesn't mean computed values are affected by the device pixel ratio.

> > If I understand the expected effects correctly, it's a lot easier for me to think of it as an effect on the CSS reference pixel on an element basis.
> 
> You could think of it that way...
> 
> > Then it doesn't really change cascading/ inheritance?
> 
> Defining it that way might simplify the definition somewhat, but it's already only a couple of sentences. Is there something very unclear about how I've written it?

I find the =scale=/ =unscale= concept a bit hard to verify.

chrishtr: > I don't think so. Then it would affect the output from typed OM, and I don't think we want that.
> 
> In Chrome's implementation the effective zoom factor is applied to px values on the ComputedStyle objects, and it also includes the device to css pixel ratio baked in. That's an implementation detail. It doesn't mean computed values are affected by the device pixel ratio.

Yes...maybe I am not understanding what you are asking. Maybe I should say "used style" instead of "computed style"?

css-viewport/Overview.bs
lilles: Makes sense.

css-viewport/Overview.bs
lilles: Used values are what the resolved values refer to for e.g. the width property, so that wouldn't work directly without also changing the definition of resolved values in CSSOM. I'm not even sure actual values include the device pixel ratio. CSS has a concept of device pixels used e.g. in https://drafts.csswg.org/css-values-4/#ref-for-device-pixel

Maybe reference pixel isn't the right term here, but need to use the CSS px to device pixels somehow.

chrishtr: Ok I read through the specs. I have concluded that it should apply to the used value but just before layout; spec text adjusted accordingly. Why:

* The CSS reference pixel is for defining the ratio between a generic CSS pixel (before any zoom is applied) and device pixels. This concept is really a post-layout one, and also is a global concept and not specific to an element. For these reasons, we can't amend that.

* As you said, modifying computed styles would affect CSSOM APIs by default, so we don't want that if possible.

* The actual size is post-layout, which we can't use for that reason, because zoom affects layout.

* The used size is where units like `auto` get resolved to numbers via layout. I placed zoom at the start of the used size calculation. We could in principle define yet another stage between computed and used just for zoom, but that seems like overkill.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,165 @@ only the value previously set to it.
...
+
+<div class="note">
+	The [=effective zoom=] of an element in a nested frame may be a value other than 1 even if 'zoom' is never set on an element in that frame. This can be observed by authors via APIs such as {{Window/devicePixelRatio}} and {{Element/getBoundingClientRect}}.
+</div>
+
+<h3 id='zoom-om'>
+	DOM and CSSOM interaction
+</h3>
+
+	Computed style APIs (i.e., all values returned by {{getComputedStyle()}}) that are non-auto and non-percentage lengths must be [=unscaled=].
```
lilles, line 509: This covers the resolved used value lengths, which is great.

css-viewport/Overview.bs
```diff
@@ -356,6 +360,165 @@ only the value previously set to it.
...
+	<pre>
+	&lt;div style="zoom: 2"&gt;
+		&lt;img src="..."&gt;
+		&lt;iframe src="..."&gt;&lt;/iframe&gt;
+	&lt;div&gt;
+	</pre>	 
+</div>
+
+The <dfn>effective zoom</dfn> of an element is the product of its computed
+value of 'zoom' and all ancestors' computed values of 'zoom'.
```
chrishtr, line 494:
```suggestion
value of 'zoom' and all flat tree ancestors' computed values of 'zoom'.
```

Issue #9720, opened 2023-12-18
[css-values-4] Serialization of Functional Notations


fantasai:
Tab and I noticed that we don't rigorously define the serialization of functional notations, so we're proposing to add a generic definition to CSS Values L4.

The non-controversial aspects are:

- Serialize the arguments per their individual grammars.
- Commas serialize with a following space.
- Space-separate adjacent tokens (except between a value and a following comma)

The one area where we have very little agreement is

- Whether to omit optional values in specified value serialization.

Generally-speaking, we omit optional components in computed values. This is an important principle of CSSOM serialization for computed values. But we are very inconsistent for specified values--this is extra confused by the fact that multiple "serialization" sections in our specs don't distinguish between specified and computed values and just assume there's only one serialization.

fantasai:
Specified vs computed is defined here, fwiw: https://www.w3.org/TR/css-cascade-5/#value-stages

svgeesus:
Nice, I'm comparing with the serialization stuff in Color 4 and 5 and Fonts 4 to pick up any differences.

PR #9721, opened 2023-12-18
[css-values-4] Define generic serialization for functional notations.…

fantasai:
See issue https://github.com/w3c/csswg-drafts/issues/9720


css-values-4/Overview.bs
```diff
@@ -2928,6 +2928,57 @@ Functional Notations</h2>
...
 
+<h3 id="fuctional-serialization">
+Serialization of Functional Notations</h3>
+
+	<div algorithm="serialize a functional notation">
+		Unless otherwise specified,
+		to <dfn export>serialize a functional notation</dfn>:
+
+		1. Let |s| be a string containing
+			the function's name in lowercase,
```
svgeesus, line 2939:
```suggestion
			the function's name in <a href="https://infra.spec.whatwg.org/#ascii-lowercase">ASCII lowercase</a>,
```

css-values-4/Overview.bs
```diff
@@ -2928,6 +2928,57 @@ Functional Notations</h2>
...
+			followed by “(” (U+0028 LEFT PARENTHESIS).
+
+		2. Append to |s|
+			the serializations of the arguments of the [=functional notation=]
+			per their individual grammars,
+			in the order the grammars are written in,
+			joining space-separated tokens with a single space,
+			and following each serialized comma with a single space (U+0020 SPACE).
```
svgeesus, line 2947: Is it possible to describe here that other tokens like `;`, `/` and so on are also followed by a single space? Or is there no consistency in what we have now?

css-values-4/Overview.bs
fantasai: I don't think that we actually need to specify this detail because the function names are already limited to ASCII.

css-values-4/Overview.bs
fantasai: I think that makes sense.

Issue #9009, opened 2023-06-27
[css-shapes-1] Simplify the animation type of `shape-outside`


yarusome:
The animation type of [`shape-outside`](https://drafts.csswg.org/css-shapes-1/#shape-outside-property) is "as defined for &lt;basic-shape&gt;, otherwise discrete", but the [`clip-path`](https://drafts.fxtf.org/css-masking/#the-clip-path) property in a similar situation has its animation type as "by computed value". [Web Animations](https://drafts.csswg.org/web-animations-1/#animating-properties) seems to have also covered the case needed by `shape-outside` in "by computed value", so it may be OK to simplify the animation type of `shape-outside` as "by computed value".

astearns:
The "By Computed Value" section in Web Animations doesn’t fit one of the discrete cases for shape-outside: we cannot animate between two shapes from image, so that has to be discrete (even though the computed values have the same number of components and are the same type)

Issue #5160, opened 2020-06-04
[css-shapes-1] Update propdef to apply to initial letter boxes


fantasai:
Since we're applying 'shape-outside' and 'shape-margin' to initial letter boxes, probably should include that in the Applies-to propdef (at-risk).

PR #9722, opened 2023-12-18
[css-shapes-1] propdef fixes

fantasai:
See https://github.com/w3c/csswg-drafts/issues/5160

### DIFF
--- a/css-shapes-1/Overview.bs
+++ b/css-shapes-1/Overview.bs
@@ -944,7 +944,7 @@ Float Area Shape: the 'shape-outside' property</h3>
 		Name: shape-outside
 		Value: none | [ <<basic-shape>> || <<shape-box>> ] | <<image>>
 		Initial: none
-		Applies to: floats
+		Applies to: floats and [=initial letter boxes=]
 		Inherited: no
 		Computed value: as <a href="#basic-shape-computed-values">defined</a> for <<basic-shape>> (with <<shape-box>> following, if supplied); else the computed <<image>>; else the keyword as specified
 		Animation type: as <a href="#basic-shape-interpolation">defined</a> for <<basic-shape>>, otherwise discrete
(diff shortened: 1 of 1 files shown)


astearns:
Could you add this to the shape-margin propdef table, too?

fantasai:
@astearns Done.

css-shapes-1/Overview.bs
```diff
@@ -944,10 +944,10 @@ Float Area Shape: the 'shape-outside' property</h3>
 		Name: shape-outside
 		Value: none | [ <<basic-shape>> || <<shape-box>> ] | <<image>>
 		Initial: none
-		Applies to: floats
+		Applies to: floats and [=initial letter boxes=]
 		Inherited: no
 		Computed value: as <a href="#basic-shape-computed-values">defined</a> for <<basic-shape>> (with <<shape-box>> following, if supplied); else the computed <<image>>; else the keyword as specified
-		Animation type: as <a href="#basic-shape-interpolation">defined</a> for <<basic-shape>>, otherwise discrete
+		Animation type: by computed value; see <a href="#basic-shape-interpolation">interpolation</a> of <<basic-shape>>
```
astearns, line 950: I think we need the discrete portion for shapes from image and/or combos of images/basic-shapes etc.

(thought I’d already left this comment, but now I don’t see it)

fantasai: I think that's covered here: https://www.w3.org/TR/web-animations/#by-computed-value

astearns: Image to image has the same type and number of components, so does not fit the requirements for discrete animation there.

fantasai: Oh, right. I forgot that doesn't animate here the same way as normally. I'll drop that commit.

You might want to close https://github.com/w3c/csswg-drafts/issues/9009 as rejected.

Issue #7390, opened 2022-06-20
[css-shapes-1] Define `<basic-shape>` with CSS value definition syntax


cdoublev:
I would like to get [`<basic-shape>`](https://drafts.csswg.org/css-shapes-1/#typedef-basic-shape) defined using the CSS value definition syntax so that it can be be extracted by `w3c/reffy`, similarly as [`<color>`](https://drafts.csswg.org/css-color-4/#typedef-color) is defined with its sub-productions.

> `<basic-shape> = <basic-shape-rect> | <circle()> | <ellipse()> | <polygon()> | <path()>`

cdoublev:
`<basic-shape>` is defined in CSS Shapes 1 with the following functions:

  - `circle(<shape-radius>? [at <position>]?)`
  - `ellipse([<shape-radius>{2} ]? [at <position>]?)`
  - `inset(<length-percentage>{1,4} [round <'border-radius'>]?)`
  - `polygon(<'fill-rule'>? , [<length-percentage> <length-percentage>]#)`
  - `rect([<length-percentage> | auto ]{4} [round <'border-radius'>]?)`
  - `xywh(<length-percentage>{2} <length-percentage [0,∞]>{2} [round <'border-radius'>]?)`
  - `path(<'fill-rule'>? , <string>)`

CSS Shapes 2 extends this list with `shape(<'fill-rule'>? from <coordinate-pair>, <shape-command>#)`.

---

`<basic-shape>` is used in: 

  - CSS Masking: [`clip-path`](https://drafts.fxtf.org/css-masking-1/#propdef-clip-path)
  - Motion: [`offset-path`](https://drafts.fxtf.org/motion-1/#propdef-offset-path)
  - CSS Shapes 1: [`shape-outside`](https://drafts.csswg.org/css-shapes-1/#propdef-shape-outside)
  - CSS Shapes 2: [`shape-inside`](https://drafts.csswg.org/css-shapes-2/#propdef-shape-inside)
  - CSS Round Display 1: [`shape-inside`](https://drafts.csswg.org/css-round-display-1/#propdef-shape-inside)
  - SVG 2: [`shape-inside`](https://svgwg.org/svg2-draft/text.html#TextShapeInside), [`shape-substract`](https://svgwg.org/svg2-draft/text.html#TextShapeSubtract)

For `clip-path`, CSS Masking links `<basic-shape>` to CSS Shapes 1 and does not restrict its value.

[...]

Issue #9053, opened 2023-07-10
[css-shapes-1] Incorrect conversion from `xywh()` to the equivalent `inset()` function, for computed values


BorisChiou:
Per the spec definition of the [computed values in [css-shapes-1]](https://drafts.csswg.org/css-shapes-1/#basic-shape-computed-values), we have to compute `xywh()` and `rect()` into the equivalent `inset()` function. There is a note in the spec:
>  Given rect(t r b l), the equivalent function is inset(t calc(100% - r) calc(100% - b) l). Given xywh(x y w h), the equivalent function is inset(y calc(x + w) calc(y + h) x).

It seems the equivalent `inset()` function for `xywh()` is not correct to me. Per definition of `inset()`, the four `<length-percentage>`s define the position of the top, right, bottom, and left edges of a rectangle, respectively, as insets from the corresponding edges of the reference box. Therefore, the right edge should be something like `100% - (x + w)` because it is the offset from the right edge. Similar for the bottom edge. It should be `100% - (y + h)`.

There are two possible ways to fix this:
1. compute the equivalent `rect()` function, so we don't have to change the formula, and just change the function name.
2. still compute to equivalent `inset()` function, but update the note to something like: Given `xywh(x y w h)`, the equivalent function is `inset(y calc(100% - x - w) calc(100% - y - h) x)`.

I may use the 2nd way as the current proposal and add it to the wpt for now.

cc @tabatkins

tabatkins:
Oh whoops, indeed, I got that conversion wrong. `inset(y calc(100% - x - w) calc(100% - y - h) x)` is obviously the right answer.

progers:
I just ran into this when working on these tests. We need to update the spec note ([here](https://drafts.csswg.org/css-shapes-1/#basic-shape-computed-values)) to mention the correct computed inset value for xywh.

@nt1m FYI, since you are also actively working in this area.

Issue #8695, opened 2023-04-06
[css-shapes-1] Optional values shouldn't be serialized


fantasai:
Currently the CSS Shapes draft [specifies](https://www.w3.org/TR/css-shapes-1/#basic-shape-serialization):

> To serialize the [<basic-shape>](https://www.w3.org/TR/css-shapes-1/#typedef-basic-shape) functions, serialize as per their individual grammars, in the order the grammars are written in, avoiding calc() expressions where possible, avoiding calc() transformations, omitting components when possible without changing the meaning, joining space-separated tokens with a single space, and following each serialized comma with a single space.

This seems mostly fine, except there's a couple problems:

1. There's an _example_ that says:

   > Omitting components means that some default values do not show up in the serialization. But since [<position>](https://drafts.csswg.org/css-values-4/#typedef-position) always uses the 2- or 4-value form, a default <position> is not omitted.

   which is nonsense, because omitting a default `at <position>` is required by the normative text above.

   This example is then tested in WPT, which shows that we (unfortunately) have interop on the example rather than on the normative text. However, having special rules that go against our general serialization principles without good reason is an anti-pattern. We should be consistent and allow `at <position>` to be omitted from the serialization just like any other optional component.

2. The guidance about “avoiding calc() expressions where possible, avoiding calc() transformations” conflicts with our more generally-applicable rules about serializing `<position>` and `calc()` values, which distinguish specified-value and computed-value serializations in very particular ways.

Proposal:
1. Fix the examples and tests, and file implementation bugs, allowing `at <position>` to be omitted, consistent with our general serialization principles.
2. Remove guidance about `calc()` serialization, allowing the spec to defer to the more specific serialization rules in css-values-4.

tabatkins:
+1 to both of these

BorisChiou:
+1 to the proposal to omit the `at <position>`. This is especially necessary to motion path because `offset-path` doesn't use the standard default for `at <position>` if the component is omitted.

tabatkins:
We definitely need to omit the `at <position>` when it was omitted by the author, for offset-path to work correctly, but we also need to *preserve* the `at <position>` even if it specifies the default value, for the same reason - `circle(at center)` works differently than `circle()` when used as an offset path, because the latter will use 'offset-position' for the default instead.

emilio:
@tabatkins It's a bit weird to have a magic value by omission, I don't think we have many precedents for such thing? None come to mind right now but I haven't tried too hard...

[...]

Issue #824, opened 2016-12-21
[css-shapes-1] ellipse() grammar gratuitously inconsistent with radial-gradient()


tabatkins:
The [`ellipse()` grammar defined in Shapes](https://drafts.csswg.org/css-shapes-1/#funcdef-ellipse) was clearly designed to *resemble* [the grammar of radial-gradient()](https://drafts.csswg.org/css-images-3/#funcdef-radial-gradient), but it's different in small but important ways, for no apparent reason.

In particular, the ellipse part of the `radial-gradient()` grammar is: `<extent-keyword> | <length-percentage>{2}`.  You either say something like "farthest-side", which defines an ellipse, or give lenpers for both axises.

On the other hand, the `ellipse()` function says `[ <extent-keyword> | <length-percentage> ]{2}?` - for *each axis*, you can either provide a keyword or a lenper. So you can do something like `ellipse(farthest-side closest-side)`, to make its width the amount necessary to hit the farthest horizontal side, and its height the amount necessary to hit the closest vertical side.  But you *can't* do `ellipse(closest-side)`.

Given the seniority of the `radial-gradient()` syntax, and the lack of major issues with it, I think we should converge on that as the grammar for `ellipse()`.

AmeliaBR:
There's an additional inconsistency that also applies to the `circle()` function: only `-side` keywords are allowed, not `-corner` keywords.

Was there any reason for not allowing `circle()` and `ellipse()` shape functions to use `closest-corner` and `farthest-corner` to define radii?

astearns:
@AmeliaBR `closest-corner` and `farthest-corner` did not seem necessary for `shape-outside` and `clip-path`, as using those for radii often creates a shape that extends outside the margin box. I'm not against allowing them for consistency's sake, but I don't know that they would get much use.

AmeliaBR:
@astearns  The reference box for sizing a clip-path or shape-outside doesn't need to be the margin-box. A circle sized to the content box's corners could still fit within the margin box. And the shape functions may be adopted for other uses.

But regardless of the strength of use cases, I agree with Tab's original point that consistency is preferred.

astearns:
I should have addressed this a few years ago. Sorry for the delay.

Adding the *-corner keywords should be fine.

[...]

PR #9723, opened 2023-12-18
Shapes edits

tabatkins:
Fixing up a few long-pending Shapes edits for #9053, #8695, #824, and a Guillaume complaint (about the basic shape grammar definitions not being in a prod) that I can't find the issue number for right now.

### DIFF
--- a/css-shapes-1/Overview.bs
+++ b/css-shapes-1/Overview.bs
@@ -340,13 +340,15 @@ Supported Shapes</h3>
 			[ round <<'border-radius'>> ]?
 		)
 
+		<dfn>&lt;basic-shape-rect></dfn> = <<inset()>> | <<rect()>> | <<xywh()>>
+
 		<<circle()>> = circle(
-			<<shape-radius>>?
+			<<radial-size>>?
 			[ at <<position>> ]?
 		)
 
 		<<ellipse()>> = ellipse(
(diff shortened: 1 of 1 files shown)


cdoublev:
Will fix https://github.com/w3c/csswg-drafts/issues/7390, thanks!

css-shapes-1/Overview.bs
```diff
@@ -423,106 +457,64 @@ Supported Shapes</h3>
...
-					the center of the circle.
-					This defaults to center if omitted.
-				</li>
-			</ul>
-		</dd>
-		<dt><dfn>ellipse()</dfn> =
-			ellipse( [ <<shape-radius>>{2} ]? [ at <<position>> ]? )
-		</dt>
+			* The <<radial-size>> argument defines
+				the circle's radius.
```
astearns, line 463: Was there a reason for omitting “Negative values are invalid” here and for ellipse()?

astearns (approved):
One nit about removing prose for disallowing negative values (only a nit because it’s expressed in the syntax)

css-shapes-1/Overview.bs
tabatkins: Already defined as part of the production, which has bounds on the `<length-percentage>`. It didn't need to be reproduced.

PR #9749, opened 2023-12-23
[css-images-4][editorial] Improved examples for stripes() function

SebastianZ:
I've made the changes I've mentioned in #9719. That is, SVGs are used instead of PNGs, the file names have been aligned with the rest of the assets and the alternative text has been removed because it's unnecessary because the description of the images is right above them.

Sebastian


svgeesus (approved):
LGTM

fantasai (approved):
I would leave the height as 1em so it scales with text. Otherwise looks good.

francescorn:
I am wondering why the stripes() examples specifically should be svgs @SebastianZ @fantasai @svgeesus, the other images are still pngs in the draft. I volunteer to convert them from png to svg.

SebastianZ:
> I would leave the height as 1em so it scales with text. Otherwise looks good.

I've changed them to be 1em high, though note that this is counter-productive when users have chosen a different font size than 16px because the images then get stretched vertically which let's the text "transparent" appear stretched as well.

Sebastian

SebastianZ:
> I am wondering why the stripes() examples should be svgs, the other images are pngs in the draft.

I only did that for the `stripes()` example images because they were added lately. The others should also benefit from being replaced by SVGs.

Sebastian

PR #9764, opened 2024-01-04
[css-view-transitions-1] Treat `auto` as an invalid value for `view-transition-name`

### LINKED ISSUE #9639: [css-view-transitions-1] Disallow the name `auto` as `view-transition-name`
Currently the name `auto` is allowed as `view-transition-name`. Given the talk about view-transition "classes" in #8319. I think it would be good to resolve on reserving the "auto" keyword so it can be leveraged if needed.

I'm opening this as a separate issue since this is probably something we want to change in VT level 1 rather than level 2.

@khushalsagar @fantasai @tabatkins @noamr

noamr:
Keeping `auto` as reserved for future purposes, so for now it is not a valid ident for `view-transition-name`.

Closes #9639

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1943,7 +1943,7 @@ Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230
 * Copy `color-scheme` from DOM element to ''::view-transition-group()''. See <a href="https://github.com/w3c/csswg-drafts/issues/9276">issue 9276</a>.
 * Expose [=auto-skip view transition=] for a {{Document}}, to allow having outbound cross-document transitions preceed programmatic view transiitons. see <a href="https://github.com/w3c/csswg-drafts/issues/9512">issue 9512</a>.
 * Add a note about why 'view-transition-name' should be animatable.
-* `view-transition-name: auto` should be an invalid value.
+* `view-transition-name: auto` should be an invalid value. See <a href="https://github.com/w3c/csswg-drafts/issues/9639">issue 9639</a>.
 
 <h3 id="changes-since-2022-05-25">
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1940,6 +1943,7 @@ Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230
 * Copy `color-scheme` from DOM element to ''::view-transition-group()''. See <a href="https://github.com/w3c/csswg-drafts/issues/9276">issue 9276</a>.
 * Expose [=auto-skip view transition=] for a {{Document}}, to allow having outbound cross-document transitions preceed programmatic view transiitons. see <a href="https://github.com/w3c/csswg-drafts/issues/9512">issue 9512</a>.
 * Add a note about why 'view-transition-name' should be animatable.
+* `view-transition-name: auto` should be an invalid value.
```
khushalsagar, line 1946: Mind linking to the relevant issue.

css-view-transitions-1/Overview.bs
```diff
@@ -534,6 +534,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		: <dfn>none</dfn>
 		:: The [=/element=] will not participate independently in a view transition.
 
+		: <dfn>auto</dfn>
+		:: This should be treated as an invalid value, despite being a valid <<custom-ident>>.
```
khushalsagar, line 538:
```suggestion
		:: This should be treated as an invalid value, despite being a valid <<custom-ident>>. The property value remains unchanged.
```

Maybe that's implicit when we say "invalid value". Wasn't sure.

khushalsagar (approved):

PR #9767, opened 2024-01-04
[css-images-4] Described how to interpolate between fr units and lengths or percentages

### LINKED ISSUE #8622: [css-backgrounds-4][css-grid] Allow transitions between different types within `stripes()`
In #8140 @fantasai changed the computed value of [`stripes()`](https://drafts.csswg.org/css-images-4/#funcdef-stripes) to say

> The computed value of this function is an ordered list of stripes, each given as a computed color and a thickness represented either a `<flex>` value or a computed `<length-percentage>` value.

The reason behind that was to make it work for inherited properties.

Though that seems to come at the cost of not being able to transition between values.

At the same time, [CSS Grid defines](https://drafts.csswg.org/css-grid-1/#typedef-flex) this:

> `<flex>` values are not `<length>`s (nor are they compatible with `<length>`s, like some `<percentage>` values), so they cannot be represented in or combined with other unit types in `calc()` expressions.

Both result in disallowing transitions between values of different types.

In #8163 I provided an example of how those types could be computed. That would allow transitioning between them and animating them.

Taken from that issue:

> How do you interpolate from `stripes(#000 10px, #111 20px)` to `stripes(#222 10fr, #333 20fr)`?

Possible answer: With a total width of 60px, the transition would be from `stripes(#000 10px, #111 20px, transparent 30px)` to `stripes(#222 20px, #333 40px, transparent 0px)`. For 120px total width it would be from `stripes(#000 10px, #111 20px, transparent 90px)` to `stripes(#222 40px, #333 80px, transparent 0px)`. 

Also `stripes(#111 calc(5px + 5fr), #222 calc(10px + 10fr))` could be made valid. With a total length of 60px again, it would be equivalent to `stripes(#111 calc(5px + 15px), #222 calc(10px + 30px))`. And assuming a total width of 120px, it would be the same as `stripes(#111 calc(5px + 35px), #222 calc(10px + 70px))`.

Sebastian

SebastianZ:
Defined how transitions between different types in `stripes()` are handled.

Fixes #8622

Sebastian

### DIFF
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -2471,24 +2471,13 @@ Interpolating ''stripes()'' {#interpolating-stripes}
 	Similar to gradients,
 	two ''stripes()'' can be interpolated,
 	allowing for smooth animations from one image to another.
-	There are only a few restrictions on what ''stripes()'' are allowed to be interpolated:
-
-	1. Both the starting and ending image must have the same number of <<color-stripe>>s.
-
-	2. Each pair of interpolated thicknesses must be of the same type,
-		i.e. both must either be of type <<length-percentage>>, or <<flex>>.
-
-	If the two images satisfy both constraints,
+	There is only one restriction on what ''stripes()'' are allowed to be interpolated.
+	Both the starting and ending image must have the same number of <<color-stripe>>s.
+	If the two images satisfy this constraint,
 	they must be interpolated as described below.
-	If they fail the second one only,
-	they must be abruptly transitioned at 50%
-	(unless otherwise specified by a future specification).
-	If they fail the first constraint,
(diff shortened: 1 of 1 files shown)


SebastianZ:
@tabatkins @fantasai Friendly ping to review this change. The PR is already open for more than half a year.

SebastianZ:
@fantasai @tabatkins @LeaVerou Friendly ping to have a look at this PR, so we can move the `stripes()` function forward.

Sebastian

fantasai (approved):
Sorry, I didn't notice the GH notification. In general, I'm not really tracking them (even though I know I should >_<;;) so feel free to poke me on other channels.

PR #9773, opened 2024-01-08
[css-view-transitions-2] Specify 'view-transition-class' and corresponding algorithms

### LINKED ISSUE #8319: [css-view-transitions-2] Creating 'classes' of transition groups
https://codepen.io/jaffathecake/pen/VwBprqL

In this example there are lots of boxes. Each has a unique `view-transition-name`, but each animates in the same way.

In the demo, I'm able to use selectors like `::view-transition-group(*)` to target all of them, but that's only because they're the only thing animating. That's unlikely to be the case in the real world.

It'd be nice to have a way to style the animation of all of the boxes at once.

Some ideas:

---

```css
.box-1 {
  view-transition-name: box-1;
  view-transition-class: box;
}

.box-2 {
  view-transition-name: box-2;
  view-transition-class: box;
}

::view-transition-group(*, box) {
  /* … */
}
```

The pseudo-class selector is `::view-transition-group(view-transition-name, view-transition-class)`. Unlike `view-transition-name`, many elements can be given the same `view-transition-class`.

This style of selector would work on the other pseudo-elements too, such as `::view-transition-new(view-transition-name, view-transition-class)`.

---

```css
::view-transition-group(*):element-class(box) {
  /* … */
}
```

(I can't think of a good name for this. I know the current name is bad)

`:element-class(ident)` matches a view transition pseudo, if either the associated outgoing or incoming element has/had a class of `ident`.

We could also have `:old-element-class(ident)` and `:new-element-class(ident)` to work with classes specifically on the outgoing or incoming element.

noamr:
A `view-transition-class` property allows specifying view-transition pseudo-elements that apply to multiple participating elements without having to replicate them.

In this PR:
- Specifying `view-transition-class`
- Extending the pseudo-element syntax to support `::view-transition-foo(name.class)`
- Extending the capture algorithms and data structure to capture the classes
- Added a simple example for using classes

Based on [this resolution](https://github.com/w3c/csswg-drafts/issues/8319#issuecomment-1876155920).
Closes #8319

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -269,7 +269,7 @@ for multiple elements without having to replicate the corresponding pseudo-eleme
 
 	<dl dfn-type=value dfn-for=view-transition-class>
 		: <dfn>none</dfn>
-		:: No class would apply to the [=/element=], and pseudo-elements would have to match its 'view-transition-name'.
+		:: No class would apply to the [=named view-transition pseudo-elements=] generated for this element.
 
 		: <dfn><<custom-ident>>*</dfn>
 		:: All of the specified <<custom-ident>> values (apart from <css>none</css>) apply as classes for the [=/element=].
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -209,7 +209,7 @@ spec:infra; type:dfn; text:list
 ### 'view-transition-class' ### {#vt-class-example}
 
 view-transition-class provides a way to use the same style
-for multiple elements without having to replicate the corresponding pseudo-elements.
+for multiple view transition pseudo elements without having to replicate the corresponding pseudo-elements.
 
 	<div class="example">
 	This example creates a transition with each box participating under its own name, while applying
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -208,7 +208,7 @@ spec:infra; type:dfn; text:list
 
 ### 'view-transition-class' ### {#vt-class-example}
 
-view-transition-class provides a way to use the same view-transition style
+view-transition-class provides a way to use the same style
 for multiple elements without having to replicate the corresponding pseudo-elements.
 
 	<div class="example">
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -204,7 +204,7 @@ spec:infra; type:dfn; text:list
 
 ### 'view-transition-class' ### {#vt-class-example}
 
-In addition to cross-document view-transitions, this document specifies a way to use the same view-transition style
+view-transition-class provides a way to use the same view-transition style
 for multiple elements without having to replicate the corresponding pseudo-elements.
 
 	<div class="example">
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -32,6 +32,8 @@ spec:css-view-transitions-1;
 	text: call the update callback; type: dfn;
 	text: perform pending transition operations; type: dfn;
 	text: setup view transition; type: dfn;
+	text: additional capture steps; type: dfn;
+	text: named view transition pseudo-element; type: dfn;
 spec:dom; type:dfn; text:document
 spec:css22; type:dfn; text:element
 spec:selectors-4; type:dfn;
@@ -124,6 +126,8 @@ spec:infra; type:dfn; text:list
 	1. From this point forward, the transition continues as if it was a same-document transition, as per [=activate view transition=].
 
 ## Examples ## {#examples}
+
+### Cross-document view-transitions ### {#cross-doc-example}
 
 	<div class=example>
 		To generate the same cross-fade as in the first example [[css-view-transitions-1#examples]],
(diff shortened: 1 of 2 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -386,6 +494,27 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
+## Additions to capture algorithms: ## {#additions-to-capture-algorithms}
+<div algorithm="capture old classes">
+When capturing the old state of the view transition, perform the following step when iterating on |captureElements| given |element| and |capture|:
+
+	1. Set |capture|'s [=captured element/class list=] to the [=computed value=] of |element|'s 'view-transition-class'.
+
+</div>
+
+<div algorithm="capture new classes">
+When capturing the new state of the view transition, perform the following step when iterating on |captureElements| given |element| and |namedElements|:
```
khushalsagar, line 513: I suppose the same thing here for this loop: https://www.w3.org/TR/css-view-transitions-1/#capture-new-state-algorithm:~:text=set%20of%20strings.-,For%20each%20element%20of%20every%20element%20that%20is%20connected%2C%20and%20has%20a%20node%20document%20equal%20to%20to%20document%2C%20in%20paint%20order%3A,-If%20any%20flat.

But I also don't want to explode the number of sub-algorithms in the L1 spec. So if you have any other idea to better link to the exact spot that's good to. Another suggestion is, "Do this after step 4.7 of https://www.w3.org/TR/css-view-transitions-1/#capture-the-new-state". If you like that, I'm happy with the same for above as well.

khushalsagar (commented):
I'd appreciate @tabatkins's review on this too. I feel like a good chunk of the VT spec is like html, so adding to the existing algorithms is hard.

If there's an existing style CSS specs use for this kind of addition to algorithms between levels of a module, would be good to stay consistent with that.

noamr:
> I'd appreciate @tabatkins's review on this too. I feel like a good chunk of the VT spec is like html, so adding to the existing algorithms is hard.
> 
> If there's an existing style CSS specs use for this kind of addition to algorithms between levels of a module, would be good to stay consistent with that.

Yea doing this in a monkey-patchy way didn't feel exactly right.

css-view-transitions-2/Overview.bs
```diff
@@ -206,6 +209,83 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+	Percentages: n/a
+	Computed Value: as specified
+	Animation type: discrete
+	</pre>
+
+	The 'view-transition-class' works alongside 'view-transition-name', and is read at the same time
+	'view-transition-name' is read. But unlike 'view-transition-name', 'view-transition-class' doesn't
+	have to be unique - an element can have several view-transition classes, and the same 'view-transition-class'
+	can apply to multiple elements. While 'view-transition-name' is used to match between the element
+	in the old state with its corresponding element in the new state, 'view-transition-class' is used
```
vmpstr, line 273: I'd add a sentence to say explicitly that `view-transition-class` on its own does not affect the structure of the view transition pseudo element tree constructed.

I know this is what this says, but perhaps repeating in a different way is worthwhile even if non-normatively

css-view-transitions-2/Overview.bs
```diff
@@ -206,6 +209,83 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+	only to apply styles using the view-transition pseudo-elements
+	(''::view-transition-group()'', ''::view-transition-image-pair()'', ''::view-transition-old()'', ''::view-transition-new()'').
+
+	<dl dfn-type=value dfn-for=view-transition-class>
+		: <dfn>none</dfn>
+		:: No class would apply to the [=/element=], and pseudo-elements would have to match its 'view-transition-name'.
+
+		: <dfn><<custom-ident>>*</dfn>
+		:: All of the specified <<custom-ident>> values (apart from <css>none</css>) apply as classes for the [=/element=].
```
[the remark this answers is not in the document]
vmpstr: I don't think we should... I don't see how class being auto would be a useful reserved keyword since the class doesn't affect anything we do, just allow more selection

css-view-transitions-2/Overview.bs
```diff
@@ -206,6 +209,83 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
+
+			Note: If the same 'view-transition-name' is specified for an element both in the old and new states of the transition,
```
vmpstr, line 284: This needs to be normative. Or rather, maybe move the note below to where the capture algorithm is as non-normative?

css-view-transitions-2/Overview.bs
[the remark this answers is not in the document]
vmpstr: This needs to be defined a bit better I think, using `[=captured element/class list=]` perhaps, or reference the algorithm below.

css-view-transitions-2/Overview.bs
noamr: That's monkey-patchy, I don't like that... let's see what Tab has to say.

css-view-transitions-2/Overview.bs
noamr: This is an informative note that described what the normative algorithms already do.

css-view-transitions-2/Overview.bs
```diff
@@ -199,6 +206,84 @@ spec:infra; type:dfn; text:list
 		```
 	</div>
 
+### 'view-transition-class' ### {#vt-class-example}
+
+view-transition-class provides a way to use the same view-transition style
```
vmpstr, line 211:
```suggestion
view-transition-class provides a way to use the same style
```

css-view-transitions-2/Overview.bs
```diff
@@ -199,6 +206,84 @@ spec:infra; type:dfn; text:list
...
+for multiple elements without having to replicate the corresponding pseudo-elements.
```
vmpstr, line 212:
```suggestion
for multiple view transition pseudo elements without having to replicate the corresponding pseudo-elements.
```

css-view-transitions-2/Overview.bs
```diff
@@ -199,6 +206,84 @@ spec:infra; type:dfn; text:list
...
+	Note that 'view-transition-class' by itself doesn't mark an element for capturing, it is only used as an additional
+	way to style an element that already has a 'view-transition-name'.
+
+	<dl dfn-type=value dfn-for=view-transition-class>
+		: <dfn>none</dfn>
+		:: No class would apply to the [=named view transition pseudo-elements=] generated for this element.
+
+		: <dfn><<custom-ident>>*</dfn>
+		:: All of the specified <<custom-ident>> values (apart from <css>none</css>) are applied when used in [=named view transition pseudo-element=] selectors.
+			<css>none</css> is an invalid <<custom-ident>> for 'view-transition-class', even when combined with other <<custom-ident>>s.
```
vmpstr, line 280: Just to clarify (for the impl sake). If none appears as one of the tokens, then the whole value is invalid, right? Or do we just skip over the none value and still use the other values? We prototyped the former

vmpstr (commented):
This looks good to me, but let's make the changes to make this a diff spec as we agreed in the call today

noamr:
> This looks good to me, but let's make the changes to make this a diff spec as we agreed in the call today

Already done.

css-view-transitions-2/Overview.bs
noamr: The whole value is invalid.

vmpstr: 👍

vmpstr (approved):

PR #9774, opened 2024-01-08
[css-view-transitions-1] Refactor timings of `updateCallbackDone` and related

### LINKED ISSUE #9762: [css-view-transitions-1] Clarify timing of `updateCallbackDone`
Currently `updateCallbackDone` is a reaction to the provided `update` callback (in `startViewTransition(update)`),
and activating the view transition is a reaction to `updateCallbackDone`.

This makes it so that author-provided `updateCallbackDone` reactions are *usually* called before capturing the new state, but sometimes after. This can create a confusion and subtle bugs.

I found that `updateCallbackDone` is most useful in its usual behavior, when all the author reactions are flushed and only then the new state is captured - this allows using this callback as a way to add new state changes on top of whatever is in the update callback 

Proposing to slightly change the wording so that it's clear that there's a microtask checkpoint between resolving `updateCallbackDone` and activating the view transition (capturing the new state).

noamr:
Note: this change is not a behavior change, but rather a clarification of existing expected behavior.

- Instead of reacting to the update callback's promise in several places, we react to it once with fulfill/reject steps and continue based on the state, all inside the "call the update callback" algorithm.
- Clarify that in the successful case, `updateCallbackDone` is always resolved after the new state capture but before the `ready` promise. (This is current behavior, but not explicit in the current spec).
- Fix a bug where we were resolving a promise with the result of reacting to a different promise, which is wrong use of those operations.
- Use normative language for user-agent transition timeout
- Remove redundant early returns for `done` phase.

Closes #9762

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1611,7 +1611,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		1. Let |fulfillSteps| be to following steps:
 			1. [=Activate view transition|Activate=] |transition|.
 
-			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then [=resolve=] |transition|'s [=ViewTransition/update callback done promise=] with undefined.
+			1. [=Resolve=] |transition|'s [=ViewTransition/update callback done promise=] with undefined.
+
+				Note: This would be a no-op if the previous step already resolved the promise.
 
 		1. Let |rejectSteps| be the following steps given |reason|:
 			1. [=Reject=] |transition|'s [=ViewTransition/update callback done promise=] with |reason|.
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1264,14 +1264,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			given |transition|'s [=relevant global object=],
 			to perform the following steps:
 
+				Note: A task is queued here because the texture read back in [=capturing the image=] may be async,
+					although the render steps in the HTML spec act as if it's synchronous.
+
 			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then abort these steps.
 
 				Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
 
 			1. [=call the update callback=].
-
-				Note: A task is queued here because the texture read back in [=capturing the image=] may be async,
-					although the render steps in the HTML spec act as if it's synchronous.
 	</div>
 
 	<div algorithm>
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1262,7 +1262,13 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. [=Queue a global task=] on the [=DOM manipulation task source=],
 			given |transition|'s [=relevant global object=],
-			to [=call the update callback=].
+			to perform the following steps:
+
+			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then abort these steps.
+
+				Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
+
+			1. [=call the update callback=].
 
 				Note: A task is queued here because the texture read back in [=capturing the image=] may be async,
 					although the render steps in the HTML spec act as if it's synchronous.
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1633,13 +1602,39 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. If |transition|'s [=ViewTransition/phase=] is not "`done`", then set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
-		1. [=Resolve=] |transition|'s [=ViewTransition/update callback done promise=]
-			with the result of [=reacting=] to |callbackPromise|:
+		1. Let |fulfillSteps| be to following steps:
+			1. [=Activate view transition|Activate=] |transition|.
 
-			- If the promise was fulfilled, then return undefined.
+			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then [=resolve=] |transition|'s [=ViewTransition/update callback done promise=] with undefined.
```
khushalsagar, line 1608: Resolving multiple times is a no-op right?

css-view-transitions-1/Overview.bs
```diff
@@ -1262,43 +1262,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
 			given |transition|'s [=relevant global object=],
-			to execute the following steps:
+			to [=call the update callback=].
 
 				Note: A task is queued here because the texture read back in [=capturing the image=] may be async,
 					although the render steps in the HTML spec act as if it's synchronous.
-
-			1. If |transition|'s [=ViewTransition/phase=] is "`done`", then abort these steps.
-
-				Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
```
khushalsagar, line 1272: We still need to handle this case IIUC. With this PR, we queue a task to call the update callback. Before that task could run, author can call skipTransition() via script. skipTransition() will see that the phase is before "update-callback-called" and also post a task to call the update callback. And call the update callback always invokes the update callback, even if the phase is done because it assumes that we're in the skip case.

css-view-transitions-1/Overview.bs
noamr: When would we resolve multiple times?

css-view-transitions-1/Overview.bs
noamr: Fair enough, reinstated this one.

css-view-transitions-1/Overview.bs
noamr: Oh I guess if setting up the pseudo-elements fail... yea, https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise-resolve-functions #5

We might not even need the `if`, we can simply resolve and rely on that behavior. WDYT?

css-view-transitions-1/Overview.bs
```diff
@@ -1262,43 +1262,16 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
...
-
-					1. [=Mark as handled=] |transition|'s [=ViewTransition/ready promise=].
-
-						Note: |transition|'s [=ViewTransition/update callback done promise=] will provide the {{unhandledrejection}}.
-						This step avoids a duplicate.
-
-					1. [=Skip the view transition=] |transition| with |reason|.
-
-				* If the promise was fulfilled, then [=activate view transition|activate=] |transition|.
+				Note: A task is queued here because the texture read back in [=capturing the image=] may be async,
```
khushalsagar, line 1273: Do you mind restoring the position of this note? Nicer to have this next to where we post the task. Otherwise its part of this sub-point which is less clearer.

css-view-transitions-1/Overview.bs
```diff
@@ -1633,13 +1608,41 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. If |transition|'s [=ViewTransition/phase=] is not "`done`", then set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
-		1. [=Resolve=] |transition|'s [=ViewTransition/update callback done promise=]
-			with the result of [=reacting=] to |callbackPromise|:
+		1. Let |fulfillSteps| be to following steps:
```
khushalsagar, line 1611:
```suggestion
		1. Let |fulfillSteps| be the following steps:
```

khushalsagar (approved):
Thanks for fixing the spec here!

css-view-transitions-1/Overview.bs

PR #9779, opened 2024-01-10
[scroll-animations-1] Add `all` as a value for `timeline-scope`

### LINKED ISSUE #9158: [scroll-animations-1] Add a way for timeline-scope to pull up all contained timelines
_(Branching off from https://github.com/w3c/csswg-drafts/issues/8915 here, where @flackr suggested `timeline-scope: all` on the call)_

This proposal is to add a way for `timeline-scope` to pull up all contained timelines using a single property value.

Use-case is when you have a set of elements that need to sync to another set (with the same size) of elements. Think of navigation indicators, typically implemented using a “scrollspy”. Authors can nowadays do this using Scroll-Driven Animations _(see [this demo](https://kizu.dev/scroll-driven-animations/#solution-using-timeline-scope) by @kizu)_.

```css
.parent > nav {
  li {
   animation: activate linear both;
  }

  li:nth-child(1) { animation-timeline: --tl-1; }
  li:nth-child(2) { animation-timeline: --tl-2; }
  li:nth-child(3) { animation-timeline: --tl-3; }
}

.parent > .content {
  div:nth-child(1) { view-timeline: --tl-1; }
  div:nth-child(2) { view-timeline: --tl-2; }
  div:nth-child(3) { view-timeline: --tl-3; }
}
```

To pull up the visibility of the contained timelines up to a shared parent, authors need to set the `timeline-scope` property’s value to include those names. This can be quite repetitive and is brittle to maintain.

```css
.parent {
  timeline-scope: --tl-1, --tl-2, --tl-3, --tl-4, …, --tl-n; /* 👈 Ugh! */
}
```

An easier way out, would be to allow a keyword such as `all` as the value for `timeline-scope`. That way, all timelines in view of that element would be pulled up.

```css
.parent {
  timeline-scope: all;
}
```

This addition would simplify all this for authors.

_(The rest of the code – the first snippet posted – can be simplified using the proposal in https://github.com/w3c/csswg-drafts/issues/9141)_

bramus:
Fixes #9158.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -1211,7 +1211,7 @@ spec:selectors-4; type:dfn; text:selector
 
 	<pre class="propdef">
 		Name: timeline-scope
-		Value: none | <<dashed-ident>>#
+		Value: none | all | <<dashed-ident>>#
 		Initial: none
 		Applies to: all elements
 		Inherited: no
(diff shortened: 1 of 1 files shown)


bramus:
@fantasai PTAL.

bramus:
Or maybe @flackr or @andruud can LGTM this one?

flackr (approved):
This looks good to me.

PR #9783, opened 2024-01-10
[css-color-4] ActiveCaption is a background.

emilio:
And thus shouldn't match CanvasText, but Canvas.

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -6567,7 +6567,7 @@ Appendix A: Deprecated CSS System Colors</h2>
 		<dd>Active window border. Same as ''ButtonBorder''.
 
 		<dt><dfn>ActiveCaption</dfn>
-		<dd>Active window caption. Same as ''CanvasText''.
+		<dd>Active window caption. Same as ''Canvas''.
 
 		<dt><dfn>AppWorkspace</dfn>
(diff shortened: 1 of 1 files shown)


svgeesus:
Is it? The few hits on "Active window caption" seem to suggest that it is the text (and thus, the text color) on the Windows title bar:

 - https://www.vbforums.com/showthread.php?37466-Getting-Active-Window-Caption
 - https://forum.ozgrid.com/forum/index.php?thread/47281-show-text-in-active-window-caption/

emilio:
@svgeesus it is. The color pairs are:

 * ActiveCaption / CaptionText
 * InactiveCaption / InactiveCaptionText

emilio:
@svgeesus the origin of these colors is win32, and the right docs are [here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsyscolor). Quoting:

> #### COLOR_ACTIVECAPTION
>
> Active window title bar. The associated foreground color is COLOR_CAPTIONTEXT.
>
> ---
>
> #### COLOR_CAPTIONTEXT
>
> Text in caption, size box, and scroll bar arrow box. The associated background color is COLOR_ACTIVECAPTION.

svgeesus (approved):
Thanks for the win32 reference!

PR #9818, opened 2024-01-18
[css-view-transitions-2] Change the short name to avoid MDN errors

noamr:


css-view-transitions-1/Overview.bs
```diff
@@ -1113,7 +1113,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 ### {{ViewTransition/skipTransition()}} Method Steps ### {#ViewTransition-skipTransition}
 
-	<div algorithm>
+	<div algorithm="dom-viewtransition-finished">
```
khushalsagar, line 1116:
```suggestion
	<div algorithm="dom-viewtransition-skipTransition">
```

This suffix seems better because the algorithm describes the steps for this API?

PR #9819, opened 2024-01-18
[css-view-transitions-2] Separte MPA transition start from check

noamr:
In prepartion for the HTML integration and the `pageconceal` event, the check for whether cross-document view transition should take place is to be called twice from the HTML spec:
- once when initiating the navigation, to check if a `pageconceal` event should be blocking
- once right before firing `pageconceal`.

So separating the algorithm and exporting the check.

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -434,43 +434,34 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
 ### Setting up the view-transition in the old {{Document}} ###  {#setup-old-document-vt}
 
 	<div algorithm>
-		To check if a <dfn export>should navigation trigger a cross-document view-transition?</dfn> given
-		an [=origin=] |oldOrigin|, an [=origin=] |newOrigin|, a boolean |newDocumentWasCreatedWithCrossOriginRedirects|, a {{NavigationType}} |navigationType|, and a boolean |isBrowserUINavigation|:
+		To check if a <dfn export>navigation can trigger a cross-document view-transition?</dfn> given
+		an [=origin=] |oldOrigin|, an [=origin=] |newOrigin|, a boolean |navigationHasCrossOriginRedirects|, a {{NavigationType}} |navigationType|, and a boolean |isBrowserUINavigation|:
+
+			Note: this is called during navigation, potentially [=in parallel=], for documents that have opted-in to view-transitions using the ''@view-transition'' rule.
 
 		1. If the user agent decides to display an [=implementation-defined=] navigation experience, e.g. a gesture-based transition for a back navigation,
 			the user agent may ignore the author-defined view transition. If that is the case, return false.
 
 		1. If |navigationType| is {{NavigationType/reload}}, then return false.
 
-		1. If ||newDocumentWasCreatedWithCrossOriginRedirects|, then return false.
-
 		1. If |isBrowserUINavigation| is true, and |navigationType| is {{NavigationType/push}} or {{NavigationType/replace}}, then return false.
 
 		1. If |oldOrigin| is not [=same origin=] as |newOrigin| then return false.
 
-		1. If |newDocument| [=was created via cross-origin redirects=] is true
-			and |newDocument|'s [=latest entry=] is null, then return false.
-
-			Note: A document with a non-null [=latest entry=]
-			is being [=Document/reactivated=], in which case we don't need to check for cross-origin redirects.
+		1. If |navigationHasCrossOriginRedirects| is true, then return false.
 
 		1. Return true.
 	</div>
 	<div algorithm>
 		To <dfn export>setup cross-document view-transition</dfn> given a {{Document}} |oldDocument|,
-		a {{Document}} |newDocument|, a {{NavigationType}} |navigationType|, a boolean |isBrowserUINavigation|, and |onReady|, which is an algorithm accepting nothing:
+		a {{Document}} |newDocument|, and |onReady|, which is an algorithm accepting nothing:
 
-		1. [=Resolve @view-transition rule=] for |oldDocument| and let |resolvedRule| be the result.
+		1. Let |resolvedRule| be the result of [=Resolve @view-transition rule|resolving the @view-transition rule=] for |oldDocument|.
 
-		1. If |resolvedRule| is "<code>skip transition</code>", then return false.
+		1. If |resolvedRule| is "<code>skip transition</code>", then return null.
 
 			Note: We don't know yet if |newDocument| has opted in, as it might not be parsed yet.
 			We check the opt-in for |newDocument| when we fire the {{Window/pagereveal}} event.
-
-		1. If the result of calling [=should navigation trigger cross-document view-transition?=]
-			given |oldDocument|'s [=Document/origin=], |newDocument|'s [=Document/origin=],
-			|newDocument|'s [=was created via cross-origin redirects=], |navigationType|, and |isBrowserUINavigation| is false,
-			then return.
 
 		1. If |oldDocument|'s [=active view transition=] is not null,
 			then [=skip the view transition|skip=] |oldDocument|'s [=active view transition=]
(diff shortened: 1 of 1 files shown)


khushalsagar:
Still need to review the PR but one quick comment based on this statement.

> once when initiating the navigation, to check if a pageconceal event should be blocking

The check should still be at 1 place after we have the final response. Because in the future it'll be based on the post redirect URL. We'll just run it in the browser (or parent process in Gecko?) to decide whether `pageconceal` needs to happen before VT capture steps.

noamr:
> Still need to review the PR but one quick comment based on this statement.
> 
> > once when initiating the navigation, to check if a pageconceal event should be blocking
> 
> The check should still be at 1 place after we have the final response. Because in the future it'll be based on the post redirect URL. We'll just run it in the browser (or parent process in Gecko?) to decide whether `pageconceal` needs to happen before VT capture steps.

I hope this would become clearer with the HTML PR :) but in the meantime:

We need to resolve the opt-in twice: once when the navigation is initiated and send that boolean to the browser process, and once right before `pageconceal`, given that it was blocking due to the opt-in being there when the navigation was initiated. We cannot read the opt-in directly in the browser process.

khushalsagar:
> once when the navigation is initiated and send that boolean to the browser process, and once right before pageconceal, given that it was blocking due to the opt-in being there when the navigation was initiated. We cannot read the opt-in directly in the browser process.

Specs shouldn't have a concept of browser vs renderer process, this should be an implementation detail. This is the flow of events:

1. Navigation is initiated.
2. We have the final network response (in the browser process).

If we're evaluating the opt-in after #2 but *before* dispatching any events to the page, which implementation option we choose below wouldn't be web observable:

- The browser sends an IPC to the renderer to evaluate the opt-in.
- The renderer replicates the renderer's CSS state in the browser.

Do you see any issue with this?

noamr:
> > once when the navigation is initiated and send that boolean to the browser process, and once right before pageconceal, given that it was blocking due to the opt-in being there when the navigation was initiated. We cannot read the opt-in directly in the browser process.
> 
> Specs shouldn't have a concept of browser vs renderer process, this should be an implementation detail. 

Not exactly. Much of the navigation flow occurs "in parallel", which means in a different thread/process, in this case that's what represents the browser process. So if during an "in parallel" flow we want to read something from the document (like run an algorithm that reads styles from current CSS), we have to queue a task on that document (which would be equivalent to an IPC call).

> This is the flow of events:
> 
> 1. Navigation is initiated.
> 2. We have the final network response (in the browser process).
> 
> If we're evaluating the opt-in after 2) but _before_ dispatching any events to the page, which implementation option we choose below wouldn't be web observable:
> 
> * The browser sends an IPC to the renderer to evaluate the opt-in.
> * The renderer replicates the renderer's CSS state in the browser.
> 
> Do you see any issue with this?

The issue is that this IPC means an unnecessary round-trip in case there is no opt-in, or in case the opt-in was already there when we started the navigation. The way I've specified it in the PR is that we read it when starting the navigation, without that extra round-trip after the final response.

css-view-transitions-2/Overview.bs
```diff
@@ -434,32 +434,44 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
 ### Setting up the view-transition in the old {{Document}} ###  {#setup-old-document-vt}
 
 	<div algorithm>
-		To <dfn export>setup cross-document view-transition</dfn> given a {{Document}} |oldDocument|,
-		a {{Document}} |newDocument|, a {{NavigationType}} |navigationType|, a boolean |isBrowserUINavigation|, and |onReady|, which is an algorithm accepting nothing:
+		To check if a <dfn export>should navigation trigger a cross-document view-transition?</dfn> given
```
khushalsagar, line 437:
```suggestion
		To check if a <dfn export>navigation can trigger a cross-document view-transition?</dfn> given
```

css-view-transitions-2/Overview.bs
```diff
@@ -434,32 +434,44 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
+		an [=origin=] |oldOrigin|, an [=origin=] |newOrigin|, a boolean |newDocumentWasCreatedWithCrossOriginRedirects|, a {{NavigationType}} |navigationType|, and a boolean |isBrowserUINavigation|:
 
 		1. If the user agent decides to display an [=implementation-defined=] navigation experience, e.g. a gesture-based transition for a back navigation,
-			the user agent may ignore the author-defined view transition. If that is the case, return.
+			the user agent may ignore the author-defined view transition. If that is the case, return false.
+
+		1. If |navigationType| is {{NavigationType/reload}}, then return false.
 
-		1. If |navigationType| is {{NavigationType/reload}}, then return.
+		1. If ||newDocumentWasCreatedWithCrossOriginRedirects|, then return false.
```
khushalsagar, line 445:
```suggestion
		1. If |newDocumentWasCreatedWithCrossOriginRedirects| is true, then return false.
```

css-view-transitions-2/Overview.bs
```diff
@@ -434,32 +434,44 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
 
-		1. If |isBrowserUINavigation| is true, and |navigationType| is {{NavigationType/push}} or {{NavigationType/replace}}, then return.
+		1. If |isBrowserUINavigation| is true, and |navigationType| is {{NavigationType/push}} or {{NavigationType/replace}}, then return false.
 
-		1. If |oldDocument|'s [=environment settings object/origin=] is not [=same origin=] as
-			|newDocument|'s [=environment settings object/origin=] then call |onReady| and return.
+		1. If |oldOrigin| is not [=same origin=] as |newOrigin| then return false.
 
 		1. If |newDocument| [=was created via cross-origin redirects=] is true
```
khushalsagar, line 451: There is an early return above for the "created with cross-origin redirects" case so we'll never hit this.

css-view-transitions-2/Overview.bs
```diff
@@ -434,32 +434,44 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
+		1. Return true.
+	</div>
+	<div algorithm>
+		To <dfn export>setup cross-document view-transition</dfn> given a {{Document}} |oldDocument|,
+		a {{Document}} |newDocument|, a {{NavigationType}} |navigationType|, a boolean |isBrowserUINavigation|, and |onReady|, which is an algorithm accepting nothing:
+
 		1. [=Resolve @view-transition rule=] for |oldDocument| and let |resolvedRule| be the result.
 
-		1. If |resolvedRule| is "<code>skip transition</code>", then call |onReady| and return.
+		1. If |resolvedRule| is "<code>skip transition</code>", then return false.
```
khushalsagar, line 465: Is this algorithm expected to provide a boolean return value? This is the only line which is doing that. If so, can you add a note to clarify what the return value means and make sure rest of the early returns are correct.

css-view-transitions-2/Overview.bs
khushalsagar, line 437: Also a note would be nice to clarify that this algorithm is to check whether a navigation qualifies for a ViewTransition assuming the Document has opted-in.

css-view-transitions-2/Overview.bs
```diff
@@ -434,32 +434,44 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
 
 			Note: We don't know yet if |newDocument| has opted in, as it might not be parsed yet.
 			We check the opt-in for |newDocument| when we fire the {{Window/pagereveal}} event.
 
+		1. If the result of calling [=should navigation trigger cross-document view-transition?=]
```
khushalsagar, line 470: It looks like the spot in the html spec which calls "setup cross-document view-transition" only does this after checking "should navigation trigger cross-document view-transition".  Its being used to set potentiallyTriggerViewTransition. So do we need this here?

If you do remove this, also remove all the arguments needed for this.

css-view-transitions-2/Overview.bs
khushalsagar: Ah, the return value is expected to be a VT object based on the call site in the html spec.

css-view-transitions-2/Overview.bs
noamr: Not exactly, but I'll move this bit to the HTML spec PR. A navigation was created with a cross-origin redirects if:
- the document was created with cross-origin redirects
- this is not a BFCache restore

css-view-transitions-2/Overview.bs
noamr: Right, fixed

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs
noamr: You're right! removed.

khushalsagar (approved):

PR #9821, opened 2024-01-19
[cssom-1] Return declarations in specified order from parse a CSS dec…

cdoublev:
The attribute change steps for `Element.style`, and the procedure for creating a new `CSSStyleDeclaration`, are receiving declarations from *parse a CSS declaration block*, in undefined order.

https://drafts.csswg.org/cssom-1/#css-declaration-blocks

  > A CSS declaration block has these attribute change steps [...]:
  >
  > [...]
  >
  > 5. Otherwise, let the declarations be the result of parse a CSS declaration block from a string `value`.

  > When a CSS declaration block object is created, then:
  >
  > [...]
  >
  > If `value` is not `null`, let the declarations be the result of parse a CSS declaration block from a string `value`.

Declarations should be in specified order, which I propose to apply in *parse a CSS declaration block*.


cdoublev:
Actually [*consume a block's contents*](https://drafts.csswg.org/css-syntax-3/#consume-a-blocks-contents) (CSS Syntax) now filters invalid declarations in the context, so it should also be responsible of applying the specified order, and the steps of *parse a CSS declaration block* should only be returning the declarations from the result of *parse a block's contents*.

I do not know how to create a link to *specified order* in CSSOM from CSS Syntax, so I mark this PR as a draft.

PR #9831, opened 2024-01-22
[cssom-1] Avoid validating declarations twice

cdoublev:
[Consume a block's contents](https://drafts.csswg.org/css-syntax-3/#consume-a-blocks-contents) (CSS Syntax) now filters invalid declarations in the context, so it does not need to be performed in [*parse a CSS declaration block*](https://drafts.csswg.org/cssom-1/#parse-a-css-declaration-block) anymore.

Related: #9821


cdoublev:
The suggested edit is now incomplete, since *parse a block's contents* new only returns rules. This would probably be best fixed in a larger edit (for *parse a CSS rule*, *parse a CSS value*, etc.

PR #9885, opened 2024-02-01
[css-view-transitions-2] Fix "<pt-class-selector>" syntax

tidoust:
The syntax used parentheses for grouping, instead of brackets, see: https://drafts.csswg.org/css-values-4/#component-combinators

(Buggy syntax introduced in #9773)


noamr (approved):
Thanks!

Issue #9886, opened 2024-02-01
[css-view-transitions-2] Update non-normative notes once `pageconceal` or so is upstreamed


noamr:

khushalsagar:
This text could also reference explicit pageconceal and pagereveal events: https://drafts.csswg.org/css-view-transitions-2/#:~:text=(when%20unloading)%20and%20in%20the%20new%20document%20(when%20reveal).

khushalsagar:
remove https://drafts.csswg.org/css-view-transitions-2/#:~:text=boolean%2C%20initially%20false.-,should%20a%20cross%2Ddocument%20transition%20take%20precedent%3F%20See%20%239512,-active%20types

khushalsagar:
Add a non-normative note for algorithms invoked by the html spec. For example: https://drafts.csswg.org/css-view-transitions-2/#resolve-view-transition-rule-algo

khushalsagar:
This should be in L2: https://drafts.csswg.org/css-view-transitions-2/#resolve-view-transition-rule-algo:~:text=Set%20oldDocument%E2%80%99s%20auto%2Dskip%20view%20transitions%20to%20true.

khushalsagar:
Move this to L2: https://drafts.csswg.org/css-view-transitions-1/#viewtransition-process-old-state-captured

khushalsagar:
"Call onReady" should be the last step of https://drafts.csswg.org/css-view-transitions-2/#setup-cross-document-view-transition

fantasai:
(p.s. @noamr please do me a favor and label your issues :)

noamr:
> (p.s. @noamr please do me a favor and label your issues :)

Ack

noamr:
I believe all of this is addressed now.

Issue #9850, opened 2024-01-24
[css-view-transitions-1] [css-view-transitions-2] consider a living-standard model for view-transitions


noamr:
The delta-spec model has made adding features to css-view-transitions-2 a bit awkward and counterproductive. 
See https://github.com/w3c/csswg-drafts/pull/9773 for example.
The more we advance the feature, the more this is likely to make that problem more difficult.

Suggesting instead to have a single level for CSS view-transitions, and keep things in an editor draft, moving to /TR/ once they're stable, but without levels. (Is that the model for css-syntax?)

We should still be careful about feature-detection, to make sure subfeatures are properly detected (e.g. you don't assume `view-transition-class` is supported just because `view-transition-name` is supported).

Alternate proposals for cleaner models also appreciated.

/cc @tabatkins @khushalsagar @vmpstr @nt1m

khushalsagar:
Another example of awkwardness in the spec is [process old state captured](https://drafts.csswg.org/css-view-transitions-1/#viewtransition-process-old-state-captured). It's a callback used to inject code for cross-document transitions (which is the L2 spec) in the middle of an algorithm defined by the L1 spec.

I've seen a lot of value in having an ED to incubate a new feature. It helps in iterating as we work through all the issues associated with a feature. The awkwardness is also reasonable in the interim state. But if we have to actually take L2 through the standards track, it becomes messy. So I'd prefer if we can use the ED to incubate but move features to a living spec once all associated issues have been resolved.

Drawing from the html spec as an example (which is a living spec), IIRC the navigation API started off in a rough ED like spec and was then moved to the html spec in 1 PR? @domenic on that.

Also @astearns @atanassov in case there's other CSS specs we can learn from.

vmpstr:
+1 to rethinking the levels here and I support more of a living standard type of model: work in ED and republish occasionally?

Other than the awkwardness of authoring the spec text, I also find it confusing to think about which features go into which level. For instance, view transition types, view transition classes, and cross-document view transitions are all L2, but that's almost coincidental since types and classes need to integrate with same-document view transitions which is L1. 

If we publish L2, then would things like scope transitions need to go into L3? Seems somewhat arbitrary and confusing.

nt1m:
I think the levels model might be fine here? but maybe instead of having level 2 patch level 1, we could make level 2 re-define algorithms from level 1.

noamr:
[...]

PR #9892, opened 2024-02-02
[css-view-transitions-2] Editorial: refactor L2 concerns out of L1

noamr:
- L2 redefines appropriate algorithms and additional fields instead of relying on L1
- moved `onReady` to correct place, and renamed `proceedWithNavigation`.

See #9886, based on [this resolution](https://github.com/w3c/csswg-drafts/issues/9850#issuecomment-1919627409).

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1100,8 +1100,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			[Discussion of this behavior](https://github.com/w3c/csswg-drafts/issues/8045).
 	</dl>
 
-	A {{ViewTransition}} must never have both an [=ViewTransition/update callback=] and a [=ViewTransition/process old state captured=].
-
 	The {{ViewTransition/finished}} [=getter steps=] are to return [=this's=] [=ViewTransition/finished promise=].
 
 	The {{ViewTransition/ready}} [=getter steps=] are to return [=this's=] [=ViewTransition/ready promise=].
@@ -1669,8 +1667,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: Since the rejection of |transition|'s [=ViewTransition/update callback done promise=] isn't explicitly handled here,
 			if |transition|'s [=ViewTransition/update callback done promise=] rejects,
 			then |transition|'s [=ViewTransition/finished promise=] will reject with the same reason.
-
-		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call |transition|'s [=ViewTransition/process old state captured=].
 	</div>
 
 ## [=Capture the image=] ## {#capture-the-image-algorithm}
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1949,7 +1949,7 @@ Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230
 * Refactor algorithm to clarify timing, especially of `updateCallbackDone. See <a href="https://github.com/w3c/csswg-drafts/issues/9762">issue 9762</a>.
 * Add animation-delay inherit to UA stylesheet rules for (::view-transition) -image-pair, -old, and -new. See <a href="https://github.com/w3c/csswg-drafts/issues/9817">issue 9817</a>.
 * Auto-skip animation when document is hidden. See <a href="https://github.com/w3c/csswg-drafts/issues/9543">issue 9543</a>.
-* Remove references to cross-document view-transitions, to keep the L1 spec clean.
+* Remove references to cross-document view-transitions, to keep the L1 spec clean. See <a href="https://github.com/w3c/csswg-drafts/issues/9886">Issue 9886</a>.
 
 <h3 id="changes-since-2022-05-25">
 Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230525/">2022-05-25 Working Draft</a>
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -956,6 +956,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |document| be [=this's=] [=relevant global object's=] [=associated document=].
 
+<<<<<<< HEAD
 		1. If |document|'s [=Document/visibility state=] is "<code>hidden</code>",
 			then [=skip the view transition|skip=] |transition| with an "{{InvalidStateError}}" {{DOMException}},
 			and return.
@@ -964,6 +965,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			then [=skip the view transition|skip=] |transition| with an "{{InvalidStateError}}" {{DOMException}},
 			and return.
 
+=======
+>>>>>>> 2b700bdce (Move all cross-doc references to L2)
 		1. If |document|'s [=active view transition=] is not null,
 			then [=skip the view transition|skip that view transition=]
 			with an "{{AbortError}}" {{DOMException}} in [=this's=] [=relevant Realm=].
@@ -1095,18 +1098,9 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: This is used to detect changes in the [=snapshot containing block size=],
 			which causes the transition to [=skip the view transition|skip=].
 			[Discussion of this behavior](https://github.com/w3c/csswg-drafts/issues/8045).
-
-		: <dfn export>process old state captured</dfn>
-		:: An algorithm accepting nothing, or null.
-			Initially null.
-
-			Note: this is used for cross-document view transitions.
 	</dl>
 
 	A {{ViewTransition}} must never have both an [=ViewTransition/update callback=] and a [=ViewTransition/process old state captured=].
-
-	Note: [=ViewTransition/update callback=] is optionally set for same-document view transitions,
-		and [=ViewTransition/process old state captured=] is set for cross-document view transitions.
 
 	The {{ViewTransition/finished}} [=getter steps=] are to return [=this's=] [=ViewTransition/finished promise=].
 
@@ -1158,9 +1152,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			with [=this=]'s [=document element=] is its [=originating element=].
 
 			Note: The position of the [=ViewTransition/transition root pseudo-element=] within the [=document element=] does not matter, as the [=ViewTransition/transition root pseudo-element=]'s [=containing block=] is the [=snapshot containing block=].
-
-		: <dfn export>auto-skip view transitions</dfn>
-		:: A boolean. Initially false.
 	</dl>
 
 ### Additions to Elements ### {#elements-concept}
@@ -1258,8 +1249,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 			If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
 			and return.
-
-		1. If |transition|'s [=ViewTransition/process old state captured=] is not null, then call [=ViewTransition/process old state captured=] and return.
 
 		1. Set |document|'s [=document/rendering suppression for view transitions=] to true.
 
@@ -1960,6 +1949,8 @@ Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230
 * Refactor algorithm to clarify timing, especially of `updateCallbackDone. See <a href="https://github.com/w3c/csswg-drafts/issues/9762">issue 9762</a>.
 * Add animation-delay inherit to UA stylesheet rules for (::view-transition) -image-pair, -old, and -new. See <a href="https://github.com/w3c/csswg-drafts/issues/9817">issue 9817</a>.
 * Auto-skip animation when document is hidden. See <a href="https://github.com/w3c/csswg-drafts/issues/9543">issue 9543</a>.
+* Remove references to cross-document view-transitions, to keep the L1 spec clean.
+
 <h3 id="changes-since-2022-05-25">
 Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230525/">2022-05-25 Working Draft</a>
 </h3>
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -560,20 +560,20 @@ Prepend this to the [=Perform pending transition operations=] algorithm given a 
 
 		1. Assert: |document|'s [=active view transition=]'s  [=ViewTransition/phase=] is "`pending-capture`".
 
+		1. Let |viewTransitionParams| be null;
+
 		1. [=Capture the old state=] for |transition|.
 
-		1. If this succeeded, then set |document|'s [=inbound view transition params=] to a new [=view transition params=] whose
+		1. If this succeeded, then set |viewTransitionParams| to a new [=view transition params=] whose
 			[=view transition params/named elements=] is a [=map/clone=] of |transition|'s [=ViewTransition/named elements=],
 			and whose [=view transition params/initial snapshot containing block size=] is |transition|'s [=ViewTransition/initial snapshot containing block size=].
-
-			Note: The inbound transition is activated after the dispatch of {{Window/pagereveal}} to ensure mutations made in this event apply to the captured new state.
 
 		1. [=Clear view transition=] |transition|.
 
 			Note: The ViewTransition object on the old Document should be destroyed after its state has been copied to the new Document below.
 				We explicitly clear it here since the old Document may be cached by the UA.
 
-		1. Call |transition|'s [=outbound post-capture steps=].
+		1. Call |transition|'s [=outbound post-capture steps=] given |viewTransitionParams|.
 
 
 ## Monkey patches to HTML ## {#monkey-patch-to-html}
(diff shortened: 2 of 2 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -627,21 +627,21 @@ When capturing the old or new state for an element, perform the following steps
 
 				1. Set |inboundTransition|’s [=ViewTransition/phase=] to "`update-callback-called`".
 
-				1. Call |onReady|.
```
khushalsagar, line 630: I was mistaken! We do need the callback to be invoked here. The reason being that the callback executes unload steps for the old Document and those need to be deferred until we have captured it.

noamr: Yea they need to be in the `old state captured steps` but not in the steps we post as a task to the new doc

noamr: Oops, yes, moved it to the "old state captured steps"

css-view-transitions-2/Overview.bs
```diff
@@ -598,20 +615,19 @@ When capturing the old or new state for an element, perform the following steps
...
 
 		1. Let |outboundTransition| be a new {{ViewTransition}} object in |oldDocument|'s [=relevant Realm=],
-			whose [=ViewTransition/active types=] is |resolvedRule|, and whose [=ViewTransition/process old state captured=] is set to the following steps:
+			whose [=ViewTransition/active types=] is |resolvedRule|.
 
-			Issue: should we check for the opt-in again, in case there was a CSSOM change in a requestAnimationFrame callback?
+		1. Run the following steps when [=Perform pending transition operations|performing pending transition operations=]:
 
-			1. If |outboundTransition|'s [=ViewTransition/phase=] is "`done`", then call |onReady| and return.
+			1. If |outboundTransition|'s [=ViewTransition/phase=] is "`done`", then call |proceedWithNavigation| and return.
```
khushalsagar, line 622: This won't be needed because "skip the view transition" clears the active view transition.

css-view-transitions-2/Overview.bs
```diff
@@ -598,20 +615,19 @@ When capturing the old or new state for an element, perform the following steps
...
 
 			1. Assert: |outboundTransition|'s  [=ViewTransition/phase=] is "`pending-capture`".
 
-			1. [=Clear view transition=] |outboundTransition|.
+			1. Set |oldDocument|'s [=auto-skip view transitions=] to false.
```
khushalsagar, line 626: Should also do this in the skip case. In case author calls skip transition and page enters BFCache.

css-view-transitions-2/Overview.bs
```diff
@@ -598,20 +615,19 @@ When capturing the old or new state for an element, perform the following steps
...
 
-				Note: The ViewTransition object on the old Document should be destroyed after its state has been copied to the new Document below.
-					We explicitly clear it here since the old Document may be cached by the UA.
+			1. [=Capture the old state=] for |outboundTransition|.
```
khushalsagar, line 628: We need to handle the fact that capture is async, see note [here](https://drafts.csswg.org/css-view-transitions-1/#setup-view-transition:~:text=A%20task%20is%20queued%20here%20because%20the%20texture%20read%20back%20in%20capturing%20the%20image%20may%20be%20async%2C%20although%20the%20render%20steps%20in%20the%20HTML%20spec%20act%20as%20if%20it%E2%80%99s%20synchronous.).

I think that means setting the rendering suppression bit to true here and unsetting both rendering suppression + auto-skip VT in the task queued below?

css-view-transitions-2/Overview.bs
```diff
@@ -598,20 +615,19 @@ When capturing the old or new state for an element, perform the following steps
...
 
-			1. Set |oldDocument|'s [=auto-skip view transitions=] to false.
+			1. If failure is returned, call |proceedWithNavigation| and return.
```
khushalsagar, line 630: Simpler to call skip the VT here? Assuming you update the skip algorithm for the case I mentioned above.

css-view-transitions-2/Overview.bs
khushalsagar: So there's 2 options for calling "proceed with navigation":

- At the end of the rendering opportunity which we're caching, the code in this PR. This is running on the [rendering task source](https://html.spec.whatwg.org/#rendering-task-source).
- At the end of the task we post to copy all the state to the new Document here. This task is running on the [dom manipulation task source](https://html.spec.whatwg.org/#dom-manipulation-task-source).

I left a comment to understand which task source these steps should run on. But another issue I see here is, what is the ordering guarantee between these task sources? Because the task which sets up the VT object on the new Document must run before any of its rendering opportunities. If these task queues are ordered then this might be ok because we queue the task to copy the state *before* invoking "proceedWithNavigation". Please leave a note to document this assumption if so.

Since the above is quite subtle, I'd be curious on what's the advantage on invoking proceedWithNavigation sooner than the task being queued here. Doing it after initializing the VT object on the new Document seems more robust.

css-view-transitions-2/Overview.bs
```diff
@@ -598,20 +615,19 @@ When capturing the old or new state for an element, perform the following steps
 			Note: this means that calling {{Document/startViewTransition()}} while capturing the old document for a cross-document view-transition would run the callback but skip the animation.
 
 		1. Let |outboundTransition| be a new {{ViewTransition}} object in |oldDocument|'s [=relevant Realm=],
-			whose [=ViewTransition/active types=] is |resolvedRule|, and whose [=ViewTransition/process old state captured=] is set to the following steps:
+			whose [=ViewTransition/active types=] is |resolvedRule|.
 
-			Issue: should we check for the opt-in again, in case there was a CSSOM change in a requestAnimationFrame callback?
+		1. Run the following steps when [=Perform pending transition operations|performing pending transition operations=]:
```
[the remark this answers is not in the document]
khushalsagar: Since we're dealing with task sources in this spec, good to document in the "setup-cross-doc-view-transition-on-old-doc" algo that it runs on the rendering task source.

css-view-transitions-2/Overview.bs
```diff
@@ -577,7 +594,7 @@ When capturing the old or new state for an element, perform the following steps
 	</div>
 	<div algorithm>
 		To <dfn export>setup cross-document view-transition</dfn> given a {{Document}} |oldDocument|,
```
khushalsagar, line 596: Which task source is this algo called from? Can you leave a note.

css-view-transitions-2/Overview.bs
khushalsagar, line 630: "proceedWithNavigation" will be invoked from the rendering task source if done from "perform pending transition operations". And whatever task source script runs in if done via `viewTransition.skipTransition()`.

Also, in the case where we don't even set up an outbound transition, "proceedWithNavigation" is invoked synchronously at the html call-site. So now it will run on whatever source "setup cross-document view-transition" is called from.

It seems odd if the navigation steps are running on different task sources in the 3 cases. Can you document the API contract on where navigation steps should be run, based on what happens in the non-VT case. And ensure we consistently use the same task source for the VT case?

css-view-transitions-2/Overview.bs
noamr: "Proceed with navigation" doesn't run on a task source, as those steps don't run for a document (renderer). it invokes a callback that adds steps to the traversal queue (the browser process).

The task that runs on the new document and sets up the new VT Object is defined here, and runs on the DOM manipulation task source.

I think the issue is that there might be a rendering opportunity before this task is run. That's indeed a problem, I'll refactor it to be special steps that run when the document is activated.

css-view-transitions-2/Overview.bs
noamr: This is handled in the HTML spec. The callback that's passed here queues steps on the traversal queue (browser process), so it doesn't matter who calls it.

css-view-transitions-2/Overview.bs
```diff
@@ -448,15 +449,33 @@ Note: as per default behavior, the ''@view-transition'' rule can be nested insid
...
 		1. If |callbackOptions| is not provided, then run the [=method steps=] for {{Document/startViewTransition()}} and return the result.
 
-		1. If |callbackOptions| is an {{UpdateCallback}}, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOptions| and return the result.
+		1. Let |viewTransition| be null.
 
-		1. Let |viewTransition| be the result of running [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOptions|'s {{StartViewTransitionOptions/update}}.
+		1. If |callbackOptions| is an {{UpdateCallback}}, then set |viewTransition| to the result of running the  [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOptions|.
 
-		1. Set |viewTransition|'s [=ViewTransition/active types=] to |callbackOptions|'s {{StartViewTransitionOptions/type}}.
+		1. Otherwise:
```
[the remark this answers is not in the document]
noamr: That's not consistent with spec syntax

css-view-transitions-2/Overview.bs
noamr: Why are task sources important here? It's an HTML concern

css-view-transitions-2/Overview.bs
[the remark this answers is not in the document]
noamr: It's not though... it's running on the navigation and traversal task source.
Ack regarding calling the callback when skipped.

noamr:
@khushalsagar I've moved things around a bit, take a look, I think it addresses your concerns.

css-view-transitions-2/Overview.bs
```diff
@@ -593,50 +620,57 @@ When capturing the old or new state for an element, perform the following steps
...
 			1. Assert: |outboundTransition|'s  [=ViewTransition/phase=] is "`pending-capture`".
 
-			1. [=Clear view transition=] |outboundTransition|.
+			1. [=Capture the old state=] for |outboundTransition|.
 
-				Note: The ViewTransition object on the old Document should be destroyed after its state has been copied to the new Document below.
-					We explicitly clear it here since the old Document may be cached by the UA.
+			1. If this succeeded, then:
+
+				1. Let |clonedNamedElements| be |outboundTransition|'s [=ViewTransition/named elements=] to a [=map/clone=] of |outboundTransition|'s [=ViewTransition/named elements=].
```
khushalsagar, line 636: This and the step below, I think you mean to initialize the variable and then set it. "Let |clonedNamedElements| be a [=ViewTransition/named elements=]. Set |clonedNamedElements| to a [=map/clone=] of |outboundTransition|'s [=ViewTransition/named elements=]".

css-view-transitions-2/Overview.bs
khushalsagar: Reminder for this. We need to set rendering suppression for view transitions to true here if capture old state succeeds and post a task to run the success steps which also reset this bit. This is to allow async capture.

css-view-transitions-2/Overview.bs
```diff
@@ -593,50 +620,57 @@ When capturing the old or new state for an element, perform the following steps
...
+
+				1. Let |initialSnapshotContainingBlockSize| to |outboundTransition|'s [=ViewTransition/initial snapshot containing block size=].
+
+				1. Set |newDocument|'s [=inbound view transition steps=] to the following steps:
```
khushalsagar, line 640: I actually find it cleaner to stash initialSnapshotContainingBlockSize and clonedNamedElements on the Document and inline these steps into "resolve cross-document view-transition", where they run. Instead of binding them to a callback and then stashing that on the new Document.

css-view-transitions-2/Overview.bs
```diff
@@ -593,50 +656,15 @@ When capturing the old or new state for an element, perform the following steps
...
-					
-					Note: The |inboundTransition| is activated after the dispatch of {{Window/pagereveal}} to ensure mutations made in this event apply to the captured new state.
-
-				1. At any given time, the UA may decide to skip |inboundTransition|, e.g. after an [=implementation-defined=] timeout.
-					To do so, the UA should [=queue a global task=] on the [=DOM manipulation task source=] given |newDocument|'s [=relevant global object=] to perform the following step:
-						If |transition|'s [=ViewTransition/phase=] is not "`done`", then [=skip the view transition=] |transition| with a "{{TimeoutError}}" {{DOMException}}.
-
-			Note: |outboundTransition| is not exposed to JavaScript, it is used only for capturing
-			the state of the old document.
+			1. Call |proceedWithNavigation|.
```
khushalsagar, line 667: nit: Update the note below, "The process continues in [=perform pending transition operations=]."

css-view-transitions-2/Overview.bs
```diff
@@ -656,20 +684,38 @@ When capturing the old or new state for an element, perform the following steps
...
 
-		1. If |transition| is null or |transition|'s [=ViewTransition/is inbound cross-document transition=] is false,
-			then return null.
+		1. If |document|'s [=active view transition=] is not null, then return null.
 
-			Note: |transition|'s [=ViewTransition/is inbound cross-document transition=] would be false if a same-document
-			transition was started before the {{Window/pagereveal}} event was fired.
+			Note: this means that starting a same-document transition before revealing the document would cancel a pending cross-document transition.
+
+		1. Let |transition| be a new {{ViewTransition}} in |document|'s [=relevant Realm=],
```
khushalsagar, line 699: nit: Should we move this after resolving the rule below? Other |transition| is unused.

css-view-transitions-2/Overview.bs
```diff
@@ -515,6 +553,29 @@ When capturing the old or new state for an element, perform the following steps
 	Note: This is written in a monkey-patch manner, and will be merged into the algorithm once the L1 spec graduates.
 </div>
 
+## Addition to pending transition operations ## {#additions-to-pending-transition-operation}
```
khushalsagar, line 556: We need a similar block for "skip the view transition algorithm". To add a step to the end of that algorithm which invokes "[=outbound post-capture steps=]".

css-view-transitions-2/Overview.bs
```diff
@@ -515,6 +553,29 @@ When capturing the old or new state for an element, perform the following steps
...
+		1. If this succeeded, then set |viewTransitionParams| to a new [=view transition params=] whose
+			[=view transition params/named elements=] is a [=map/clone=] of |transition|'s [=ViewTransition/named elements=],
+			and whose [=view transition params/initial snapshot containing block size=] is |transition|'s [=ViewTransition/initial snapshot containing block size=].
+
+		1. [=Clear view transition=] |transition|.
+
+			Note: The ViewTransition object on the old Document should be destroyed after its state has been copied to the new Document below.
+				We explicitly clear it here since the old Document may be cached by the UA.
+
+		1. Call |transition|'s [=outbound post-capture steps=] given |viewTransitionParams|.
```
khushalsagar, line 576: This needs to be done in a posted task because "Capture the old state" can be async. So we set "rendering suppression" to true, post a task which resets that bit and then invoke [=outbound post-capture steps=]. In that task, also check in case the phase for the transition is done (skip might have been called).

The L1 spec does the same in the task posted [here](https://drafts.csswg.org/css-view-transitions-1/#setup-view-transition-algorithm:~:text=A%20task%20is%20queued%20here%20because%20the%20texture%20read%20back%20in%20capturing%20the%20image%20may%20be%20async%2C%20although%20the%20render%20steps%20in%20the%20HTML%20spec%20act%20as%20if%20it%E2%80%99s%20synchronous.).

noamr: The L1 spec in this case is a bit dodgy... posting a task doesn't guarantee anything wrt that async capture. We need to think about this further.

css-view-transitions-2/Overview.bs
noamr: Added a reaction to `finished promise`.

css-view-transitions-2/Overview.bs
```diff
@@ -484,19 +498,43 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
-
-		Issue: should a cross-document transition take precedent? See [#9512](https://github.com/w3c/csswg-drafts/issues/9512)
+		: <dfn>outbound post-capture steps</dfn>
+		:: Null or a set of steps, initially null.
 
 		: <dfn>active types</dfn>
 		:: Null or a [=list=] of strings, initially null.
 	</dl>
 
+### [=Captured elements=] ### {#captured-elements}
```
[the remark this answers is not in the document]
noamr: Oops

css-view-transitions-2/Overview.bs
```diff
@@ -446,15 +449,26 @@ Note: as per default behavior, the ''@view-transition'' rule can be nested insid
...
 
-		1. If |callbackOptions| is not provided, then run the [=method steps=] for {{Document/startViewTransition()}} and return the result.
+		1. If |this|'s [=active view transition=] is not null and its [=outbound post-capture steps=] is not null,
+			then:
 
-		1. If |callbackOptions| is an {{UpdateCallback}}, then run the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOptions| and return the result.
+			1. Let |preSkippedTransition| be a new {{ViewTransition}} in |this|'s [=relevant realm=].
 
-		1. Let |viewTransition| be the result of running [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |callbackOptions|'s {{StartViewTransitionOptions/update}}.
+			1. [=Skip the view transition|Skip=] |preSkippedTransition| with an "{{InvalidStateError}}" {{DOMException}}.
```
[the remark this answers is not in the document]
noamr: I still don't understand why we need rendering suppression. Is that for that unspecified async capture?

css-view-transitions-2/Overview.bs
noamr: Done

css-view-transitions-1/Overview.bs
```diff
@@ -1676,8 +1656,6 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			Note: Since the rejection of |transition|'s [=ViewTransition/update callback done promise=] isn't explicitly handled here,
 			if |transition|'s [=ViewTransition/update callback done promise=] rejects,
 			then |transition|'s [=ViewTransition/finished promise=] will reject with the same reason.
-
```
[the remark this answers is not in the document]
khushalsagar: Resolving based on offline conversation. We'll take up the async capture in a follow up PR.

css-view-transitions-2/Overview.bs
```diff
@@ -593,54 +660,23 @@ When capturing the old or new state for an element, perform the following steps
...
-					To do so, the UA should [=queue a global task=] on the [=DOM manipulation task source=] given |newDocument|'s [=relevant global object=] to perform the following step:
-						If |transition|'s [=ViewTransition/phase=] is not "`done`", then [=skip the view transition=] |transition| with a "{{TimeoutError}}" {{DOMException}}.
+		1. Set |oldDocument|'s [=active view transition=] to |outboundTransition|.
 
-			Note: |outboundTransition| is not exposed to JavaScript, it is used only for capturing
-			the state of the old document.
+			Note: The process continues in [=perform pending transition operations=].
 
-		1. Set |oldDocument|'s [=active view transition=] to |outboundTransition|.
+		1. React to |outboundTransition|'s [=ViewTransition/finished promise=]:
```
khushalsagar, line 677: I really don't like this. We're just making our spec unnecessarily complicated by relying on promises for browser steps. I also suggested adding an "addition to skip steps" style of update for this because that's what we'll do when the specs are eventually merged.

khushalsagar (approved):

PR #9895, opened 2024-02-02
[css-view-transitions-1] Exclude auto as custom ident for view-transition-name

cdoublev:
Currently, below `view-transition-name`, defined with `none | auto | <custom-ident>`:

  > `<custom-ident>`: [...] The value `none` is excluded from `<custom-ident>` here.

https://drafts.csswg.org/css-view-transitions-1/#valdef-view-transition-name-custom-ident

This PR adds `auto` as an excluded `<custom-ident>` value.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -542,7 +542,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			as either an old or new [=/element=]--
 			with the specified [=view transition name=].
 
-			The value <css>none</css> is excluded from <<custom-ident>> here.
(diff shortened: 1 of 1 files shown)


nt1m (approved):
I made some extra changes, but looks good otherwise, thanks!

cdoublev:
Out of curiosity, what is the reason to exclude `auto` if is not a valid keyword for the property?

nt1m:
> Out of curiosity, what is the reason to exclude `auto` if is not a valid keyword for the property?

We were thinking of potentially reserving this keyword for in case we'd want to auto-generate names, but in general there's no real reason to name a view transition "auto".

Issue #9368, opened 2023-09-17
[css-overflow-4] line-clamp `max-lines` - support `auto` value to fit container 


yairEO:
# Feature request

#### Spec links
- https://drafts.csswg.org/css-overflow-4/#propdef--webkit-line-clamp
- https://drafts.csswg.org/css-overflow-4/#propdef-line-clamp

Currently `-webkit-line-clamp` property only supports an integer number which represents the maximum 
number of lines to possibly render before truncating the text with an ellipsis.

I have a use-case where I need an dynamic value for the `-webkit-line-clamp` so it matches the 
number of lines which can fit in the container (which is being truncated), assuming the container 
has a dynamic height.

Also important - nothing should be painted past the ellipsis point (unlike today when the container's height is set to a value greater than the height of the truncated lines)

### Example:

https://jsbin.com/kovuwow/edit?html,css,output

```html

<div>
  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt 
  ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco 
  laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in 
  voluptate velit esse cillum dolore eu fugiat nulla
</div>
```

```css
div {
  display: -webkit-box;
  -webkit-line-clamp: auto; /* 👈 keyword to let line-clamp according to the container */
  -webkit-box-orient: vertical;
  overflow: hidden;
  resize: auto;
}
```

If you want `line-clamp` to (show N lines that can) fit the container element, it is currently impossible:

https://github.com/w3c/csswg-drafts/assets/845031/72b47c23-f015-4883-bba4-8ed02db2900d

### Javascript-based solution
I've hacked this using (a lot of) javascript: https://codepen.io/vsync/pen/eYbGgGL 

https://github.com/w3c/csswg-drafts/assets/845031/2553fbc1-3f83-41c9-9e9c-ca03f0a9a013

----

Related *StackOverflow* questions:
- https://stackoverflow.com/q/51816862/104380
- https://stackoverflow.com/q/71145324/104380

SebastianZ:
[...]

PR #9896, opened 2024-02-02
[css-view-transitions-2] Require one or more custom ident for view-transition-class and type

cdoublev:
The value definitions of [`view-transition-class`](https://drafts.csswg.org/css-view-transitions-2/#propdef-view-transition-class) and [`type`](https://drafts.csswg.org/css-view-transitions-2/#descdef-view-transition-type) currently allow an empty value with `<custom-ident>*`, which means zero or more `<custom-ident>`:

  > **Name:** `view-transition-class`
  > **Value:** `none | <custom-ident>*`

See [this discussion](https://github.com/w3c/csswg-drafts/issues/9368#issuecomment-1907299467) about accepting an empty value in the property value range.


noamr:
Hi, do you mind rebasing? I'll be happy to merge it in (for classes, types are already sorted)

noamr (approved):

noamr (approved):

noamr:
@tabatkins mind marking this as non-substantive for IPR?

astearns:
@cdoublev could you contact me? (email is in my profile, or on irc, or whatever else)

I’d like to see if we can remove the constant IPR checks we have to deal with on your PRs

PR #9900, opened 2024-02-02
[css-view-transitions-2] Only accept declarations in `@view-transitions`

cdoublev:
In #9383, which introduced `@view-transitions`, I do not see examples with `@view-transitions` (or with the other rule names) containing nested rules, so I assume this is not planned in the future, and that the basic syntax of its block contents can be defined with `<declaration-list>` instead of `<declaration-rule-list>`.

https://drafts.csswg.org/css-view-transitions-2/#view-transition-grammar


noamr (approved):
LGTM, thanks!

PR #9902, opened 2024-02-02
[css-sizing-4] Define stretch sizing properly, rather than (incorrectly) indirecting thru stretch alignment.

tabatkins:
The old spec said "do what stretch self-alignment does", with some caveats for handling layout modes where that doesn't apply. But that gives bad behavior - it means that auto margins win over stretch sizing (since they win over alignment), when the opposite should occur.

### DIFF
--- a/css-sizing-4/Overview.bs
+++ b/css-sizing-4/Overview.bs
@@ -826,37 +826,24 @@ Extrinsic Size Determination</h2>
 <h3 id="stretch-fit-sizing">
 Stretch-fit Sizing: filling the containing block</h3>
 
-	Stretch-fit sizing tries to set the box’s used size
-	to the length necessary
-	to make its outer size as close to filling the [=containing block=] as possible
-	while still respecting the constraints imposed by min-height/min-width/max-height/max-width.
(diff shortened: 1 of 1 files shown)


tabatkins:
@bfgeek Does this look reasonable?

css-sizing-4/Overview.bs
```diff
@@ -826,37 +826,24 @@ Extrinsic Size Determination</h2>
...
-	in [=formatting contexts=] and axes in which the relevant [=self-alignment property=] does not apply
-	(such as the block axis in Block Layout, or the main axis in Flex Layout),
-	in cases where a percentage size in that axis would resolve to a definite value,
-	a [=stretch-fit size=]
-	causes the box to attempt to fill its containing block--
-	behaving as ''100%''
-	but applying the resulting size to its margin box
-	instead of the box indicated by 'box-sizing'.
+	[=Stretch-fit sizing=] indicates the size necessary to make the box’s [=outer size=]
+	fill the [=containing block=].
```
Loirooriol, line 830: What if the box is a block-level that establishes an independent formatting context, and there are floats to avoid?
`justify-self` aligns in the float-shrunken containing block.

Issue #8967, opened 2023-06-15
[css-fonts-5] font-size-adjust and font metrics overrides


shivamidow:
Related spec: https://drafts.csswg.org/css-fonts-5/#valdef-ascent-overridedescriptor-percentage

The spec document describes the percentage value for the font metrics overrides as follows.
> '\<percentage\>'
>    The corresponding metric is replaced by the given percentage multiplied by the used font size.

That description sounds a bit obscure to me. What is the used font size? Does it mean the specified size by `font-size` or  the computed size or the actual size to render (i.e., effective size) or something else?

The spec also needs to clarify how the font size should be adjusted when font metrics overrides and font-size-adjust are simultaneously applied. For instance,

```css
@font-face {
  font-family: TestFont;
  src: serif;
  ascent-override: 100%;
  descent-override: 50%;
}

.size-adjusted {
  font: 20px TestFont;
  font-size-adjust: 2.0;
}
```

When I tested, Gecko applied the metric override after font-size-adjust. So, Chromium is following the Gecko's behavior [1], but I hope the spec would be improved here.

Proposed solution:
Change the wording like the following;
`The corresponding metric is replaced by the given percentage multiplied by the effective font size.`
or
`The corresponding metric is replaced by the given percentage multiplied by the font size after adjustment.`

[1] https://chromium-review.googlesource.com/c/chromium/src/+/4305200

shivamidow:
> Related, but I think we discussed this: Does the spec already clarify the case of how the `size-adjust` `@font-face` descriptor interacts with `font-size-adjust`?

I think so. [The level 5 draft](https://www.w3.org/TR/css-fonts-5/#font-metrics-override-desc) mentions
> The font-size-adjust property is applied after the size-adjust descriptor.
> Note: The consequence of applying font-size-adjust after size-adjust is that size-adjust appears to have no effect.

> 
> PS:
> 
> > When I tested, Gecko applied the metric override after font-size-adjust.
> 
> Do you mean by that, the font size is first scaled (up, 2.0x in your example), then he ascent/descent overrides are applied?

Yes. [Here is the test.](http://wpt.live/css/css-fonts/font-size-adjust-metrics-override.html)

svgeesus:
Sorry to have ignored this issue; I try to keep on top of issues tagged `css-fonts-4` but don't check the `css-fonts-5` tag often enough.

> Proposed solution:
> Change the wording like the following;
[...]

PR #9915, opened 2024-02-07
[css-fonts-5] font-size-adjust precedes font metrics overrides #8967

shivamidow:
[css-fonts-5] font-size-adjust precedes font metrics overrides #8967


svgeesus (approved):
Thanks!

Issue #8799, opened 2023-05-05
[web-animations-2] Progress APIs


flackr:
In #8669, #8765 and #8201 there seems to be a common desire to have more convenient APIs for getting some representation of progress. Currently you can get the progress of the current iteration of an effect using the [getComputedTiming()](https://www.w3.org/TR/web-animations-1/#dom-animationeffect-getcomputedtiming) API as follows:

```js
let progress = animation.effect.getComputedTiming().progress;
```

You can also get the overall effect progress by doing a calculation such as:
```js
let progress = animation.currentTime / animation.effect.getComputedTiming().endTime;
```

However, with several of the other linked issues it seems that it would be convenient for developers to have some form of access to the progress through the attached animation range, or to get that progress directly from the timeline.

We could add a convenience currentProgress helper alongside the currentTime API on [Animation](https://www.w3.org/TR/web-animations-1/#the-animation-interface):
```
attribute double? currentProgress;
```
This could return the progress of currentTime through the entire range before which it is considered complete, e.g. calculated as follows:

currentProgress = [currentTime](https://www.w3.org/TR/web-animations-1/#animation-current-time) / [effect endTime](https://www.w3.org/TR/web-animations-1/#end-time)

Note that endTime is infinite when either iterations or duration is infinite. In these cases currentProgress should probably be null or undefined?

Open question - should it be read only or should we allow setting the progress which would be equivalent to setting currentTime to currentProgress * endTime?

As per discussion on #8669 a progress reported on animation should be mostly independent of the effect timing, so we calculate it as the progress from [startTime](https://www.w3.org/TR/web-animations-1/#dom-animation-starttime) to the time when the animation would resolve its finished promise. This would intentionally not match the progress through the developer specified keyframes.

@birtles @bramus @fantasai 

I also propose that given the uncertainty around what's expected from the `getCurrentTime` API on timeline, that we remove it until we have a better understanding of the specific use cases that developers need it for. Perhaps the animation progress API above will address some of them where it seems developers were trying to work this out through the timeline.

birtles:
Overall it seems reasonable to me. I think defining it in terms of `startTime` and `endTime` is a nice way of keeping it at arms length from effect timing.

> Note that endTime is infinite when either iterations or duration is infinite. In these cases currentProgress should probably be null or undefined?

Mathematically, should it be zero? Not sure if that would be the most useful result, however.

[...]

Issue #8765, opened 2023-04-24
[scroll-animations-1][web-animations-2] getCurrentTime is self-inconsistent wrt representing time


fantasai:
In order to represent time in relation to named timeline ranges, we have two options:

- Represent as absolute progress from the start of the range, in units consistent with the overall timeline (i.e. equivalent to currentTime).
- Represent as percentage progress between the start and end of the range.

If we consider the default behavior as returning the time with respect to a timeline range that encompasses the entire timeline, `getCurrentTime()` switches between the two modes depending on whether you specify a range name or not. (!)

This doesn't feel unnatural to scroll timelines, because their “absolute timeline units” are percentages of the timeline (which itself has a finite range)... We would probably have noticed the discrepency sooner if they used length-based time. :/

But it runs into problems if we extend the named timeline concept to other types of timelines, such as time-based timelines. These bring in (potentially) two differences:

- Their absolute time is not a percentage, its an absolute unit.
- They might possibly not be finite?

I think we might want to rethink this API. Some possible options:
* Restrict (and define) it as returning percentage progress through a range, rather than also being able to return the global `currentTime`, and name it accordingly as @birtles suggests in #8201.
* Define two APIs, one that returns percentage progress and one that returns absolute progress (i.e. units compatible with `currentTime`) in relation to a named range.
* Define a single API that can return either percentage progress or absolute progress depending on its arguments.

I'm not sure which direction we want to go... @bramus, thoughts?

bramus:
I can see that as an author you’d sometimes want to know the relative “time” _(i.e. a percentage)_ and sometimes you’d want the absolute “time”. We could go with two different methods `getRelative…` and `getAbsolute…` for this, or add an extra argument to the property bag of the existing one. Undecided that the default argument value – `relative` or `absolute` – should be in the latter case.

Side note: seeing that I put time between quotes in the previous paragraph and @fantasai used the word “progress” – and I also hinted at it [here](https://github.com/w3c/csswg-drafts/issues/8201#issuecomment-1493397265) – maybe it’s better to rename the method to `getProgress()`?

fantasai:
@bramus ... so maybe we want parallel `getCurrentTime` and `getCurrentProgress` functions where the former is in .currentTime units and the latter is in percentages (always) and returns null on non-finite ranges?

[...]

Issue #8669, opened 2023-04-02
[scroll-animations-1] currentTime for animations attached to a timeline attachment range


bramus:
While playing with ViewTimelines and trying to derive how much of an `animation-range` had already progressed, it surprised me that reading `currentTime` always returned the progress relative to the entire ViewTimeline range.

To demonstrate, I have [a demo up on CodePen](https://codepen.io/bramus/full/RwYOYeO/b972db85414fe9f8ef17e6c1b304c1e8) which you can **check using the latest Chrome Canary**. The demo tries to resprent the progression within the currently set animation-range for the ViewTimeline and has these controls:
- Using the radio buttons it’s possible to change the `animation-range`
- Using the _“set range-start to 20% and range-end to 80%“_ checkbox it’s possible to set the range-start and range-end percentages to 20% and 80% respectively. When not checked, the default values of 0% and 100% are used.
- The “force main thread animation“ checkbox is added to work around https://crbug.com/1421690

On load, the `animation-range` is set to `entry 20% entry 80%`.

The demo outputs the `currentTime` in three possible ways:

1. Directly read it from the animation.

    ```js
    let progress = animation.currentTime;
    ```

    This doesn’t work when set to a value other than `cover`. This because italways returns the progress measured against the entire range of the ViewTimeline (which coincides with `cover`)

2. Read it from the range

    ```js
    let progress = animation.timeline.getCurrentTime(animation.rangeStart.rangeName).value;
    ```

    Using `getCurrentTime[ForRange]` doesn’t cut it here things here, as it only covers animations that have the same rangeName in `animation-range-start` and `animation-range-end`. Above that, it also only works for when the entire range is covered (from 0% to 100%).

3. Read from the animation’s effect

    As a workaround, I turned to reading the animation’s effect, which did the trick:
    
    ```js
    // This works
    let progress = animation.effect.getComputedTiming().progress * 100;
    if (animation.playState === 'finished') progress = 100;
    ```

    This, however, felt a lot like jumping through hoops to be honest.

To simplify things for authors, I would like to propose that `currentTime` _(the method used in the first approach)_ returns the progress relative to the timeline attachment range.

- If one explicitly sets the `animation-range` to, for example, `entry 50% cover 100%` the `currentTime` should be relative to that set _(custom)_ range.
- If one sets no `animation-range`, the progress should be relative to the entire range, which for View Timelines equals to `cover`

[...]

Issue #8201, opened 2022-12-07
[scroll-animations-1] Animation.getCurrentTime is easily confused with Animation.currentTime


birtles:
I think it's confusing to have both `currentTime` and `getCurrentTime`.

Would it be possible to:

1. ~~Move this to the `ScrollTimeline` interface,~~ (Update: I see now that this is proposed to be used for `DocumentTimeline` too--I still don't quite follow the use cases but assuming we want to do that, I guess it should stay on `AnimationTimeline`)
2. Rename it to `getCurrentTimeForRange`, _and_
3. Make the `rangeName` argument non-optional

?

fantasai:
`getCurrentTime` is essentially a replacement for `.currentTime` that returns the same concept, but has more expressivity:  So I think it makes sense to keep the same name.

birtles:
> `getCurrentTime` is essentially a replacement for `.currentTime` that returns the same concept, but has more expressivity: So I think it makes sense to keep the same name.

I don't think we shouldn't do that. We'll confuse authors unnecessarily. We can't deprecate `currentTime` and for the overwhelming majority of use cases it's the right choice. We should be encouraging users to continue using `currentTime` for the sake of compatibility.

For users that need a time for a range, there's `getCurrentTimeForRange` for which the function name spells out what the string argument represents. (If we really must use `getCurrentTime` then it needs to take a property bag, not a string.)

birtles:
I'm afraid I'm still opposed to the current naming.

I don't think ranges are a fundamental concept. The overwhelming majority usage of the API will continue to be for time-based animations and I think we should bias towards that.

- Adding `getCurrentTime` makes using the API harder in the common case. When your editor suggests both `currentTime` and `getCurrentTime` how do you know which one to choose? You have to look them up to find out what their (subtle) distinction is.
- Generally, authors should prefer using `currentTime` because it is more backwards compatible.

If we _must_ have `getCurrentTime` because we expect it to be useful for non-range cases then we really should use a property bag. Doing so is necessary to:

- Make the code readable. How on earth could you guess that the single string argument to `getCurrentTime` represents a range otherwise?
- Make the API extensible. If we're going with something as generic as "get current time", then we'll no doubt want to add further conditions to that query in future and having an API that switches between positional arguments and property bags is an unnecessary wart.

Now that TS/JS language servers are able to automatically refactor positional arguments into property bags, property bags have become much more common so I believe this approach is also idiomatic.

[...]

PR #9937, opened 2024-02-12
[web-animations-2] Specify Animation progress API

DavMila:
This is to specify the [agreed upon](https://github.com/w3c/csswg-drafts/issues/8799#issuecomment-1561576205) animation.progress API.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -262,9 +262,6 @@ Append:
 >              *   Set [=iteration duration=] = [=intrinsic iteration duration=]
 >          Otherwise:
 >              *   Set [=iteration duration=] = |specified iteration duration|
->
-> An animation's <dfn export for=animation>progress</dfn> is the ratio of its
->   [=animation/current time=] to its associated effect's [=end time=].
 
 ### Setting the timeline of an animation ### {#setting-the-timeline}
 
@@ -974,27 +971,31 @@ Add the following sentence:
 
 ### Calculating the progress of an animation ### {#the-progress-of-an-animation}
 
-  The [=animation/progress=] of an [=animation=] is calculated as follows:
+  An animation's <dfn export for=animation>progress</dfn> is the ratio of its
+   [=animation/current time=] to its [=associated effect end=].
 
-  <dl class=switch>
+  <div algorithm="calculate animation progress">
+    The [=animation/progress=] of an [=animation=], |animation|, is calculated as follows:
 
-  : If <em>any</em> of the following are true:
-    * the animation does not have an [=animation/associated effect=], or
-    * the animation's associated effect's [=end time=] is infinite, or
-    * the animation's associated effect's [=end time=] is zero, or
-    * the animation's [=animation/current time=] time is an [=unresolved=] time value,
-  ::
-      The animation's [=animation/progress=] is null.
-  : Otherwise,
-  ::
-    <blockquote>
-      <code><a for="animation">progress</a> = [=animation/current time=] / <var>effect end time</var>
-      </code>
-    </blockquote>
+    <dl class=switch>
 
-    Where |effect end time| is the current [=end time=] of the associated effect.
+    : If <em>any</em> of the following are true:
+      * |animation| does not have an [=animation/associated effect=], or
+      * |animation|'s [=associated effect end=] is zero, or
+      * |animation|'s [=animation/current time=] is an [=unresolved=] time value,
+    ::
+        |animation|'s [=animation/progress=] is null.
+    : If |animation|'s [=associated effect end=] is infinite,
+        |animation|'s [=animation/progress=] is zero.
+    : Otherwise,
+    ::
+      <blockquote>
+        <code><a for="animation">progress</a> = [=animation/current time=] / |animation|'s [=associated effect end=]
+        </code>
+      </blockquote>
 
-  </dl>
+    </dl>
+  </div>
 
 
 <h3 id="animation-effects">Animation effects</h3>
@@ -2473,8 +2474,8 @@ Add:
 >     {{TimelineRangeOffset}} or the {{DOMString}} "normal".
 >
 > :  <dfn attribute for=Animation>progress</dfn>
-> ::  Specifies the progress of the <a>animation</a> as a proportion of its
->     [=animation/associated effect=]'s [=end time=].
+> ::  Specifies the [=animation/progress=] of the <a>animation</a> as a proportion of its
+>     [=associated effect end=].
 </div>
 
 <h3 id="the-animationeffect-interface">The <code>AnimationEffect</code> interface</h3>
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -981,16 +981,23 @@ Add the following sentence:
 
     : If <em>any</em> of the following are true:
       * |animation| does not have an [=animation/associated effect=], or
-      * |animation|'s [=associated effect end=] is zero, or
       * |animation|'s [=animation/current time=] is an [=unresolved=] time value,
     ::
         |animation|'s [=animation/progress=] is null.
+    : If |animation|'s [=associated effect end=] is zero,
+    ::
+        : If |animation|'s [=animation/current time=] is negative,
+        :: |animation|'s [=animation/progress=] is zero.
+
+        : Otherwise,
+        :: |animation|'s [=animation/progress=] is one.
     : If |animation|'s [=associated effect end=] is infinite,
+    ::
         |animation|'s [=animation/progress=] is zero.
     : Otherwise,
     ::
       <blockquote>
-        <code><a for="animation">progress</a> = [=animation/current time=] / |animation|'s [=associated effect end=]
+        <code><a for="animation">progress</a> = min(max([=animation/current time=] / |animation|'s [=associated effect end=], 0), 1)
         </code>
       </blockquote>
 
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -262,6 +262,9 @@ Append:
 >              *   Set [=iteration duration=] = [=intrinsic iteration duration=]
 >          Otherwise:
 >              *   Set [=iteration duration=] = |specified iteration duration|
+>
+> An animation's <dfn export for=animation>progress</dfn> is the ratio of its
+>   [=animation/current time=] to its associated effect's [=end time=].
 
 ### Setting the timeline of an animation ### {#setting-the-timeline}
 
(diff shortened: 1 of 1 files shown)


web-animations-2/Overview.bs
```diff
@@ -969,6 +972,30 @@ Add the following sentence:
...
+### Calculating the progress of an animation ### {#the-progress-of-an-animation}
+
+  The [=animation/progress=] of an [=animation=] is calculated as follows:
+
+  <dl class=switch>
+
+  : If <em>any</em> of the following are true:
+    * the animation does not have an [=animation/associated effect=], or
+    * the animation's associated effect's [=end time=] is infinite, or
+    * the animation's associated effect's [=end time=] is zero, or
```
birtles, line 984: I think we can replace "the animation's associated effect's [=end time=]" with "the animation's [=associated effect end=]".

web-animations-2/Overview.bs
```diff
@@ -969,6 +972,30 @@ Add the following sentence:
...
 
+### Calculating the progress of an animation ### {#the-progress-of-an-animation}
+
+  The [=animation/progress=] of an [=animation=] is calculated as follows:
+
+  <dl class=switch>
+
+  : If <em>any</em> of the following are true:
+    * the animation does not have an [=animation/associated effect=], or
+    * the animation's associated effect's [=end time=] is infinite, or
```
birtles, line 983: What's the thinking behind making infinite → null as opposed to zero? I think zero might be more useful overall since it represents where the animation is actually up to.

web-animations-2/Overview.bs
```diff
@@ -969,6 +972,30 @@ Add the following sentence:
 
 </div>
 
+### Calculating the progress of an animation ### {#the-progress-of-an-animation}
+
+  The [=animation/progress=] of an [=animation=] is calculated as follows:
```
birtles, line 977: Should we introduce a variable `animation` here? i.e. something like:

```bs
The [=animation/progress=] of an [=animation=], |animation|, is calculated as follows:
```

Then make the later references to `animation` be `|animation|` and annotate the block with `algorithm` per https://speced.github.io/bikeshed/#var-and-algorithms ?

web-animations-2/Overview.bs
```diff
@@ -262,6 +262,9 @@ Append:
 >              *   Set [=iteration duration=] = [=intrinsic iteration duration=]
 >          Otherwise:
 >              *   Set [=iteration duration=] = |specified iteration duration|
+>
+> An animation's <dfn export for=animation>progress</dfn> is the ratio of its
+>   [=animation/current time=] to its associated effect's [=end time=].
```
birtles, line 267: This needs to refer to the actual definition of the value. I know @fantasai and I talked about this a bit when she fixed up the spec.

When an implementer or developer clicks on a link to `progress` they should be able to get to the algorithm that defines it fairly easily, not just an abstract description of what it is.

I'd suggest moving the definition of `progress` to the same section as the algorithm. If for some reason that's not possible, we should link to the algorithm from here.

web-animations-2/Overview.bs
```diff
@@ -969,6 +972,30 @@ Add the following sentence:
...
+  ::
+      The animation's [=animation/progress=] is null.
+  : Otherwise,
+  ::
+    <blockquote>
+      <code><a for="animation">progress</a> = [=animation/current time=] / <var>effect end time</var>
+      </code>
+    </blockquote>
+
+    Where |effect end time| is the current [=end time=] of the associated effect.
```
birtles, line 995: Again we could replace `effect end time` with `|animation|'s [=associated effect end=]`.

web-animations-2/Overview.bs
```diff
@@ -2443,6 +2471,10 @@ Add:
 >      <a href="#dom-keyframeanimationoptions-rangeend">rangeEnd</a>.
 >     When reading the attribute, the returned value is either a
 >     {{TimelineRangeOffset}} or the {{DOMString}} "normal".
+>
+> :  <dfn attribute for=Animation>progress</dfn>
+> ::  Specifies the progress of the <a>animation</a> as a proportion of its
+>     [=animation/associated effect=]'s [=end time=].
```
birtles, line 2477: This should link to the algorithm that defines how to calculate the progress.

birtles (commented):
Looks good but I suggested a few tweaks.

web-animations-2/Overview.bs
DavMila: Thanks for the suggestion. Moving it to the algorithm section seems fine, done.

web-animations-2/Overview.bs
DavMila: That sounds goo. Done.

web-animations-2/Overview.bs
DavMila: Done.

web-animations-2/Overview.bs
DavMila: Done.

web-animations-2/Overview.bs
```diff
@@ -969,6 +969,34 @@ Add the following sentence:
...
+   [=animation/current time=] to its [=associated effect end=].
+
+  <div algorithm="calculate animation progress">
+    The [=animation/progress=] of an [=animation=], |animation|, is calculated as follows:
+
+    <dl class=switch>
+
+    : If <em>any</em> of the following are true:
+      * |animation| does not have an [=animation/associated effect=], or
+      * |animation|'s [=associated effect end=] is zero, or
```
birtles, line 984: Thanks for making these edits. Two follow-up questions that occurred to me:

1. If the associated effect end is zero, should the progress be 1 (or 0 if the current time is negative)? It seems odd that it's unresolved here but maybe I'm just not remembering how all this works?
2. Just to check, should we be clamping the progress to [0, 1]? I assume we shouldn't but I guess it depends on how we expect this property to be used? The answer to this will likely affect how we define the behavior for zero-duration animations.

flackr: > Thanks for making these edits. Two follow-up questions that occurred to me:
> 
> 1. If the associated effect end is zero, should the progress be 1 (or 0 if the current time is negative)? It seems odd that it's unresolved here but maybe I'm just not remembering how all this works?

I agree, it more accurately reflects the animated progress to return a value here.

> 2. Just to check, should we be clamping the progress to [0, 1]? I assume we shouldn't but I guess it depends on how we expect this property to be used? The answer to this will likely affect how we define the behavior for zero-duration animations.

It seems like it's more useful to allow progresses outside of the [0, 1] range, as especially for 0 it matters to know whether the effect is active or not. That said, allowing values outside of [0, 1] makes it more complicated for how to handle the 0 duration case. Perhaps we should use +/- infinity depending on whether you're before or after the start time?

@bramus @fantasai any thoughts here?

web-animations-2/Overview.bs
```diff
@@ -969,6 +969,34 @@ Add the following sentence:
...
+      * |animation|'s [=animation/current time=] is an [=unresolved=] time value,
+    ::
+        |animation|'s [=animation/progress=] is null.
+    : If |animation|'s [=associated effect end=] is infinite,
+        |animation|'s [=animation/progress=] is zero.
```
flackr, line 989: Can the author tell if we're active (after the start time) or inactive (before the start time)?

web-animations-2/Overview.bs
flackr: Thinking further on this, it may be simpler to restrict to [0, 1] and encourage authors to combine with the phase of the animation which seems like another thing that's not too easy to get today.

birtles: Yeah, I think you're right. Ultimately, though, it probably depends on how we expect the API to be used.

We _could_ mimic the behaviour of `progress` on `ComputedEffectTiming` and use the fill mode and phase of the root animation effect to make `progress` be `unresolved` when the root effect is not in effect but that might be a bit of a layering violation?

For reference, I believe `ComputedEffectTiming.progress` is mostly in the range [0, 1] but can extend outside that range due to easing functions.

I guess it comes back to reviewing these issues and seeing what would work best in each of those cases (quoting #8799):

> In https://github.com/w3c/csswg-drafts/issues/8669, https://github.com/w3c/csswg-drafts/issues/8765 and https://github.com/w3c/csswg-drafts/issues/8201 there seems to be a common desire to have more convenient APIs for getting some representation of progress.

DavMila:
> I've reviewed this PR and I think it looks good. Thanks @DavMila for your work on this.
> 
> My only outstanding question is [from here](https://github.com/w3c/csswg-drafts/pull/9937#discussion_r1491825782):
> 
> > We _could_ mimic the behaviour of `progress` on `ComputedEffectTiming` and use the fill mode and phase of the root animation effect to make `progress` be `unresolved` when the root effect is not in effect but that might be a bit of a layering violation?
> > ...
> > I guess it comes back to reviewing these issues and seeing what would work best in each of those cases (quoting #8799):
> > > In #8669, #8765 and #8201 there seems to be a common desire to have more convenient APIs for getting some representation of progress.
> 
> Does anyone know any specific intended uses of this API that would help determine if the progress should be unresolved when the animation is in the before/after phase and not filling?

 (not exactly a use case but just a thought that might be worth considering): `currentTime` seems to remain resolved regardless of the fill mode.
I made a [codepen](https://codepen.io/awogbemila/pen/eYXwvdw) where you can run the animation and display `currentTime` afterwards by clicking "Paste CurrentTime". This example shows that `currentTime` remains resolved even with `fill: "none"`.

birtles (approved):

flackr:
> I've reviewed this PR and I think it looks good. Thanks @DavMila for your work on this.
> 
> My only outstanding question is [from here](https://github.com/w3c/csswg-drafts/pull/9937#discussion_r1491825782):
> 
> > We _could_ mimic the behaviour of `progress` on `ComputedEffectTiming` and use the fill mode and phase of the root animation effect to make `progress` be `unresolved` when the root effect is not in effect but that might be a bit of a layering violation?
> > ...
> > I guess it comes back to reviewing these issues and seeing what would work best in each of those cases (quoting #8799):
> > > In #8669, #8765 and #8201 there seems to be a common desire to have more convenient APIs for getting some representation of progress.
> 
> Does anyone know any specific intended uses of this API that would help determine if the progress should be unresolved when the animation is in the before/after phase and not filling?

I left a comment on the original issue https://github.com/w3c/csswg-drafts/issues/8799#issuecomment-1971476352 to get a resolution here. I suppose we can land the clamping behavior and change it if we decide differently there.

web-animations-2/Overview.bs
```diff
@@ -969,6 +969,41 @@ Add the following sentence:
...
+
+        : Otherwise,
+        :: |animation|'s [=animation/progress=] is one.
+    : If |animation|'s [=associated effect end=] is infinite,
+    ::
+        |animation|'s [=animation/progress=] is zero.
+    : Otherwise,
+    ::
+      <blockquote>
+        <code><a for="animation">progress</a> = min(max([=animation/current time=] / |animation|'s [=associated effect end=], 0), 1)
```
flackr, line 1000: I think the 0 and 1 are backwards here. It looks like the 1 is passed to min and the 0 to max, meaning that the progress will always be min(progress, 1) or just 1.

flackr (approved):

Issue #9543, opened 2023-10-31
[css-view-transitions-1] Handle startVT/cross-doc transitions when Document is hidden


khushalsagar:
If `startViewTransition` is called while a Document is hidden. Or the document is hidden at any point during the transition's lifecycle, we don't do anything special. We should probably explicitly skip the transition if the Document is hidden.

Not sure what to do with iframes. The UA can stop rendering opportunities if the iframe is far enough for the viewport. Should we assume its hidden?

The spec way of when a Document is hidden can be based on [this](https://html.spec.whatwg.org/#update-the-rendering:~:text=The%20step%20labeled%20Rendering%20opportunities%20prevents%20the%20user%20agent%20from%20updating%20the%20rendering%20when%20it%20is%20unable%20to%20present%20new%20content%20to%20the%20user%20(there%27s%20no%20rendering%20opportunity).) : The step labeled Rendering opportunities prevents the user agent from updating the rendering when it is unable to present new content to the user. If the Document is in this state, skip all ViewTransitions?

noamr:
What would happen given the current spec is that the transition would be delayed until the page is visible or the timeout passes. Perhaps for iframe this is OK, and for explicitly hidden pages we can skip immediately?

khushalsagar:
> What would happen given the current spec is that the transition would be delayed until the page is visible or the timeout passes.

Right now it depends on when the page is hidden. If its before rendering has been frozen then we'll keep waiting for the next frame. If its during rendering frozen then we timeout. And if its after the DOM has been constructed then it will wait until all animations are finished. I'm not sure how we deal with animation progress for hidden Documents but if that's checked when the Document is made visible again then likely we'll discover the transition to be finished then?

> Perhaps for iframe this is OK, and for explicitly hidden pages we can skip immediately?

Sounds reasonable. If the Document is hidden, that probably means the user doesn't care to see an animation when its made visible again. We can render the new DOM directly.

For the iframe case, browsers can stop rendering the iframe if its far off from the viewport. That means sometimes authors might see a timeout error if the iframe goes offscreen right after the transition was initiated. Its rare enough that we can ignore this case. But this was why I was leaning towards using a "rendering opportunity" based definition. So we skip for any case where the Document is not going to produce any more frames.

khushalsagar:
**Proposed Resolution**
[...]

PR #9942, opened 2024-02-13
[css-view-transitions-1] Auto-skip a view-transition if document is hidden

noamr:
See [resolution](https://github.com/w3c/csswg-drafts/issues/9543#issuecomment-1939242246).

Note: this only covers [css-view-transitions-1], will follow up on cross-document view transitions.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -956,8 +956,12 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |document| be [=this's=] [=relevant global object's=] [=associated document=].
 
+		1. If |document|'s [=Document/visibility state=] is "<code>hidden</code>",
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -956,8 +956,12 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |document| be [=this's=] [=relevant global object's=] [=associated document=].
 
+		1. If |document|'s [=Document/visibility state=] is "<code>hidden</code>",
```
khushalsagar, line 959: We need to do the same thing if the state flips to hidden while a transition was running too.

khushalsagar (approved):

css-view-transitions-1/Overview.bs
noamr: Yes, that would be in the HTML spec.

PR #9949, opened 2024-02-13
[css-view-transitions-1] Export an algorithm to skip a transition in a hidden document

noamr:
Together with https://github.com/whatwg/html/pull/10137
See #9543

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1662,6 +1662,17 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			then |transition|'s [=ViewTransition/finished promise=] will reject with the same reason.
 	</div>
 
+## View transition page-visibility change steps ## {#page-visibility-change-steps}
+	<div algorithm>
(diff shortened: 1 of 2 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1662,6 +1662,17 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			then |transition|'s [=ViewTransition/finished promise=] will reject with the same reason.
 	</div>
 
+## View transition page-visibility change steps ## {#page-visibility-change-steps}
+	<div algorithm>
+	The <dfn export>view transition page-visibility change steps</dfn> given {{Document}} |document| are:
+
+	1. If |document|'s [=Document/visibility state=] is "<code>hidden</code>",
```
khushalsagar, line 1669: With my code hat on, should we also assert that if the visibility state change is "visible" then there is no active-view-transition. :)

khushalsagar (approved):

css-view-transitions-1/Overview.bs
noamr: Added an assert.

PR #9954, opened 2024-02-14
[css-view-transitions-2] Resolve outbound transition promises after document unloads

### LINKED ISSUE #9888: [css-view-transitions-2] Dealing with promises on old VT object
What should happen with all the promises once capture is done.

noamr:
See [this resolution](https://github.com/w3c/csswg-drafts/issues/9888#issuecomment-1939367607)
Closes #9888

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -20,7 +20,6 @@ Markup Shorthands: css yes, markdown yes
 spec:css-view-transitions-1;
 	text: active view transition; type: dfn;
 	text: capture the old state; type: dfn;
-	text: clear view transition; type: dfn;
 	text: activate view transition; type: dfn;
 	text: skip the view transition; type: dfn;
 	text: ViewTransition; type: interface;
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -582,13 +581,28 @@ Prepend this to the [=Perform pending transition operations=] algorithm given a
 			[=view transition params/named elements=] is a [=map/clone=] of |transition|'s [=ViewTransition/named elements=],
 			and whose [=view transition params/initial snapshot containing block size=] is |transition|'s [=ViewTransition/initial snapshot containing block size=].
 
-		1. [=Clear view transition=] |transition|.
+		1. Call |transition|'s [=outbound post-capture steps=] given |viewTransitionParams|.
```
khushalsagar, line 584: Why defer this until visibility flip instead of here itself? But reviewing this reminded me of another subtlety that needs rendering suppression! We also do it so that the old Document doesn't draw another frame after we have cached a frame here.

So I think we can just call skip the transition here itself. And maybe add to the issue above that rendering suppression bit is also needed for the reason above.

noamr: It's to avoid resolving those promises in the `visibilitychange` event that comes after this. Though I guess instead we can call these steps after `visibilitychange`.

noamr: When you resolve a promise from platform code it invokes a microtask checkpoint.

khushalsagar: Sorry then I'm confused about why you said, "It's to avoid resolving those promises in the visibilitychange event". Can you clarify again why you want this to be deferred until `visibilitychange`?

That said, now I recall why we shouldn't resolve the promises here at least. Its the last step in update the rendering and we shouldn't be executing any script callbacks here. Good to drop a note about this.

We could post a task to do this. Is your worry that this task could run after the task which runs the visibilitychange handlers? Is there a reason that would be an issue? I can be ok with this timing, just want to understand the rationale for it.

noamr: > Sorry then I'm confused about why you said, "It's to avoid resolving those promises in the visibilitychange event". Can you clarify again why you want this to be deferred until `visibilitychange`?

Because we resolved on those promises triggering only after BFCache restore. `visibilitychange` will fire in the old document after `pagehide`.

> 
> That said, now I recall why we shouldn't resolve the promises here at least. Its the last step in update the rendering and we shouldn't be executing any script callbacks here. Good to drop a note about this.

This is not the last step in update the rendering. This is after calling `pagehide`.

> We could post a task to do this. Is your worry that this task could run after the task which runs the visibilitychange handlers? Is there a reason that would be an issue? I can be ok with this timing, just want to understand the rationale for it.

That's what I'm doing in this patch, in the visibility change step. but that task needs to be queued after we queue the `pagehide` task and that's not guaranteed if we post it here..

khushalsagar: > Because we resolved on those promises triggering only after BFCache restore.

Why do we have to defer it until after BFCache restore? I don't see what would go wrong even if the promises resolve before the visibilitychange handlers run (so before the Document enters BFCache).

> Its the last step in update the rendering and we shouldn't be executing any script callbacks here. Good to drop a note about this.

I should probably clarify what I meant here. We intentionally tried to place the point at which we run the pending transform operations [late in the update the rendering loop](https://html.spec.whatwg.org/#update-the-rendering:~:text=For%20each%20doc%20of%20docs%2C%20perform%20pending%20transition%20operations%20for%20doc). This was so we can ensure we capture the painted output of all script callbacks that run during update the rendering. And for that reason, we also ensure not to invoke script callbacks here which could potentially mutate the DOM. Its also trips multiple assumptions in the rendering stack to allow script callbacks this late in the rendering update.

khushalsagar: > I'm ok with that in that case we don't need new page visibility steps. We should have WPTs that show that promises resolve after pagehide and before visibilitychange.

The WPTs should assert that the promises must have been resolved before `pageshow` when the page is restored from BFCache. That's guaranteed if we're posting a task here to skip this transition.

> OK, how does that relate to this patch?

My preference would've been to synchronously skip after we've copied over the captured state within "perform pending transition operations" itself. But that's not possible, we have to post a task, and I wanted to clarify why. On that note, the async task should check that the transition state is not "done" before invoking the skip algorithm.

khushalsagar (approved):

PR #9995, opened 2024-02-25
[cssom-1][css-color-4][css-color-5][css-color-6] Describe CSSOM serialization in terms of declared values rather than specified values, fix #9993

### LINKED ISSUE #9993: [cssom-1][css-color-4] "specified value" here would better be reworded as declared value
[Serializing CSS Values](https://www.w3.org/TR/2021/WD-cssom-1-20210826/#serializing-css-values) mentions specified, resolved, computed, and used values for serialization. The specified values are mentioned here with values from CSSStyleSheet or ElementCSSInlineStyle in mind. In css-cascade-5 terminology, however, it corresponds with declared values rather than specified values.

One important difference is with CSS-wide keywords. CSS-wide keywords, such as `inherit`, are serialized as they are within CSSOM APIs like `cssText`. This is unintuitive if they are called specified values, as they should have been resolved to other values in specified values (in the sense of css-cascade-5).

The same argument applies to [Resolving \<color\> values](https://www.w3.org/TR/2024/CRD-css-color-4-20240213/#resolving-color-values).

qnighy:
This PR rephrases `specified value` as `declared value`, where it involves CSSOM value serialization. Fixes #9993.

The fact that the CSS-wide keywords are kept as-is is already [covered in serialize-values tests](https://github.com/web-platform-tests/wpt/blob/f57f833daf407ab46840ec52c0a8de8ee02e319a/css/cssom/serialize-values.html#L369-L372).

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -5694,7 +5694,7 @@ Resolving sRGB values</h3>
 
 	If the sRGB color was explicitly specified by the author as a [=named color=],
 	or as a [=system color=],
-	the [=specified value=] is that named or system color, converted to 
+	the [=declared value=] is that named or system color, converted to 
 	<a href="https://infra.spec.whatwg.org/#ascii-lowercase">ASCII lowercase</a>.
 	The computed and used value
 	is the corresponding sRGB color,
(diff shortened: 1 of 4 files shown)


svgeesus (approved):

svgeesus:
The ipr bot seems to be stuck. Perhaps closing and re-opening will fix it?

PR #9997, opened 2024-02-26
[css-text-4] Add note about deviations from UAX-14

### LINKED ISSUE #9432: [css-text] line-break: strict should use UAX#14 rule?
Mozilla is working to replace old line breaking code with [UAX#14 ](https://www.unicode.org/reports/tr14/) compatible line breaking rules. Although Blink and WebKit use ICU4C to implement line breaking, both uses different rules with Unicode Standard Annex (See https://bugzilla.mozilla.org/show_bug.cgi?id=1848049 for details) for ASCII characters when checking behavior.

According to [code comment](https://searchfox.org/wubkat/rev/cd25edd92284ea5ea247483e66b404c0774949b2/Source/WebCore/rendering/BreakLines.cpp#53), it seems to be for a compatibility of old browsers. So when using ASCII character, it doesn't match with UAX#14. Example, https://wpt.fyi/results/css/css-text/i18n/css3-text-line-break-baspglwj-026.html?label=experimental&label=master&aligned is failed on WebKit and Blink now.

So I think that we should use UAX#14 rules if `line-break: strict`. It is clear for line breaking compatibility for future. How do you think?

birtles:
Fixes #9432

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -7450,6 +7450,33 @@ Line Breaking Strictness: the 'line-break' property</h3>
 		(ตัวอย่าง·การ·เขียน·ภาษา·ไทย).
 	</div>
 
+	<div class=note>
+		Note: While user agents may refer to [[!UAX14]] as a starting point for
+		their line-breaking implementation, the following deviations may be
+		desirable for maximum interoperability with existing implementations:
+
+		- Not introducing a line break opportunity between
+			U+0021 (Exclamation Mark, <code>!</code>) and
+			a letter ([=Unicode general category=] <code>L</code>).
+			This prevents a break in the string “!important”.
+		- Not introducing a line break opportunity between
+			U+002F (Solidus, <code>/</code>) and
+			a letter ([=Unicode general category=] <code>L</code>).
+			This prevents a break in dates such as “23/Jan/2024”.
+		- Not introducing a line break opportunity between
+			U+007C (Vertical Line, <code>|</code>) and
+			a letter ([=Unicode general category=] <code>L</code>).
+		- Not introducing a line break opportunity between
+			U+002D (Hyphen-Minus `-`) and
+			a digit ([=Unicode general category=] <code>Nd</code>)
+			if the codepoint prior to the hyphen was _not_ a letter or digit
+			([=Unicode general category=] <code>L</code> or <code>Nd</code>).
+			This prevents breaking after the minus sign before a number
+			such as in “-13”
+			whilst allowing breaks after the hyphen in “ABCD-1234” and “1234-5678”
+			which may appear in long URLs, for example.
+	</div>
+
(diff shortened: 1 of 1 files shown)


css-text-4/Overview.bs
```diff
@@ -7450,6 +7450,33 @@ Line Breaking Strictness: the 'line-break' property</h3>
 		(ตัวอย่าง·การ·เขียน·ภาษา·ไทย).
 	</div>
 
+	<div class=note>
+		Note: While user agents may refer to [[!UAX14]] as a starting point for
+		their line-breaking implementation, the following deviations may be
```
frivoal, line 7455: Avoid RFC2119 in non-normative text

```suggestion
		their line-breaking implementation, the following deviations could be
```

frivoal:
This is editorial, but could you consider reformatting this to use Semantic Linefeeds, as described in https://rhodesmill.org/brandon/2012/one-sentence-per-line/, to match the source formatting style of the rest of the document?

frivoal:
@birtles your GitHub account doesn't seem to be associated with your W3C account. At least I think that's why the IPR check fails. Could you double check on that?

birtles:
@frivoal Huh that's odd. It's always worked for me in the past. I checked my W3C account and it has the correct GitHub account there. I'm still in the W3C org on GitHub too. I suspect it's a bug in the IPR check task. It looks like it hasn't run?

frivoal:
Let me try closing and reopening the PR to see if that wakes up the bot

birtles:
> Let me try closing and reopening the PR to see if that wakes up the bot

Thanks! That seems to have done the trick.

Sorry, I missed your comment about semantic line breaks. (I have my email filtering out all CSS notifications that don't mention animations or transitions 😅). Is there still something that needs to be done there? I thought it was already using semantic line breaks for the most part. Is it just the first sentence that needs better breaking?

frivoal:
> Is it just the first sentence that needs better breaking?

Yes, the rest is indeed fine, but the first sentence could be wrapped better (e.g., after "starting point", after the comma, after desirable).

birtles:
Thanks! Done!

css-text-4/Overview.bs
```diff
@@ -7450,6 +7450,34 @@ Line Breaking Strictness: the 'line-break' property</h3>
 		(ตัวอย่าง·การ·เขียน·ภาษา·ไทย).
 	</div>
 
+	<div class=note>
+		Note: While user agents may refer to [[!UAX14]] as a starting point
```
frivoal, line 7454:
```suggestion
		Note: While user agents can refer to [[!UAX14]] as a starting point
```

frivoal:
@fantasai how does this look to you?

fantasai (approved):
Looks good to me. Squash-merge if @jfkthame approves. :)

jfkthame:
LGTM 👍

PR #10009, opened 2024-02-28
[CONTRIBUTING.md] Clarify that PRs should originate from your own fork

bramus:
Minor update to clarify that PRs should originate from your own fork. This is something I had to find out when filing one of my first PRs – I had created a branch in the  `w3c/csswg-drafts` repo, upon which I was told that this isn’t how things are done here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,10 +49,11 @@ In general, you should not directly commit changes to a draft unless you are an
 Editor of that draft, or have their explicit permission. If you are not an 
 Editor of a draft, but wish to contribute changes, the best practice is to either 
 work directly with an Editor to review proposed text, or file your proposal as a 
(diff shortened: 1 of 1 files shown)


svgeesus:
It was much more of an issue when the GitHub to Mercurial two-way bridge was still an important part of some people's workflow because any branch here would break things. AFAIK that is no longer the case.

svgeesus (approved):
Seems ok to me, though the rationale is mainly historical at this point.

frivoal (approved):

tabatkins:
Yeah, maintaining branches either here or in your personal fork are both fine. The main benefit of keeping them in your own fork is it keeps our list of branches smaller, so you don't have as many things cluttering up the branches list. (In other words, I'm fine with keeping this advice here.)

PR #10048, opened 2024-03-07
[css-overflow-3] Clarify that scripts can also scroll a scroll container

tcaptan-cr:
This clarification was precipitated by an  [intersection observer issue](https://github.com/w3c/IntersectionObserver/issues/521) in order to avoid confusion over who can scroll the overflow area of a scroll container.

### DIFF
--- a/css-overflow-3/Overview.bs
+++ b/css-overflow-3/Overview.bs
@@ -321,9 +321,8 @@ Scrollable Overflow</h3>
 Scrolling Overflow</h3>
 
 	A box’s [=overflow=] can be visible or clipped.
-        CSS also allows a box to be a <dfn export>scroll container</dfn>
-	with a <a>scrollable overflow area</a> whose clipped parts can be
-        scrolled into view via user interaction or script.
+	CSS also allows a box to be a <dfn export>scroll container</dfn>
+	that allows clipped parts of its scrollable overflow area to be scrolled into view.
 	The visual viewport of a <a>scroll container</a>
 	(through which the <a>scrollable overflow area</a> can be viewed)
 	coincides with its padding box,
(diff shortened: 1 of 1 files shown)


tcaptan-cr:
Please take a look @fantasai & @szager-chromium.

dbaron (commented):
I think this rewording reads a little bit too much like it's trying to define "scrollable overflow area", which it isn't.  There are also things that are neither clearly user scrolls nor page scripts that can cause scrolling, such as UA features like find-in-page being used by the user.

This might actually be a good use case for the passive voice, since it lets us avoid answering the question of who's doing the scrolling.  e.g., "that allows clipped parts of its scrollable overflow area to be scrolled into view".

(Also, it looks like this section uses tabs for indentation; please don't convert just a few lines to spaces.)

tcaptan-cr:
> I think this rewording reads a little bit too much like it's trying to define "scrollable overflow area", which it isn't. There are also things that are neither clearly user scrolls nor page scripts that can cause scrolling, such as UA features like find-in-page being used by the user.
> 
> This might actually be a good use case for the passive voice, since it lets us avoid answering the question of who's doing the scrolling. e.g., "that allows clipped parts of its scrollable overflow area to be scrolled into view".
> 
> (Also, it looks like this section uses tabs for indentation; please don't convert just a few lines to spaces.)

That sounds good to me. Thank you David!
I updated it based on your feedback.

dbaron (approved):

fantasai:
Wording looks good, but y'all dropped the cross-link! Fixed in 047ab497b51af55ec748ac34c3dbe362fcabc24f

tcaptan-cr:
> Wording looks good (thanks for the review @dbaron!), but y'all dropped the cross-link! Fixed in [047ab49](https://github.com/w3c/csswg-drafts/commit/047ab497b51af55ec748ac34c3dbe362fcabc24f) :)

Thanks!

PR #10061, opened 2024-03-12
[css-text] Add missing space

j9t:
Minor change.

(Would be open to prepare additional corrections, e.g., to ensure consistency in headings case and some such.)


astearns:
thanks

Issue #6642, opened 2021-09-20
[css-color] Add OKLab, OKLCH


svgeesus:
For the background and explanation of _why_ to add this feature, please see [this Color Workshop presentation](https://www.w3.org/Graphics/Color/Workshop/slides/talk/lilley) which has a video presentation, plus associated slides and transcript. You can just skim the slides/transcript if you are pressed for time.

Having done so, and given the rationale presented, the specific proposal is to add [OKLab](https://bottosson.github.io/posts/oklab/) as follows: _Edit: items re-ordered so highest priority is first._

**CSS Color 4**

1.  in the gamut mapping section, require gamut reduction to use OKLCH chroma reduction and deltaE OK, rather than CIE LCH chroma reduction and deltaE 2000. This is both better _and_ faster. _If we don't do this, I am not confident of being able to write a satisfactory gamut mapping section in CSS Color 4._ See [examples below](https://github.com/w3c/csswg-drafts/issues/6642#issuecomment-942343016)
2. in the interpolation section, for non-legacy color formats, if the host syntax does not specify a colorspace, change `lab` to `oklab` as the default. _If we don't do this, color interpolation in gradients, animations and transitions will not be as good, and will sometimes give surprising hue shifts._
3.  add `oklab()` and `oklch()` (as well as the existing `lab()` and `lch()` which are still useful) and add their implementation to the sample code. _Not doing this means we have a colorspace of proven usefulness, implemented inside the browser, but we deny users access to it._

**CSS Color 5**

4. add `oklab` and `oklch` as [defined color spaces](https://www.w3.org/TR/css-color-5/#color-spaces) which can be used in `color-mix()`. Depends on 3.
5. add `oklab()` and `oklch()`  to Relative Color Syntax. Depends on 3.

@argyleink @una @tabatkins @LeaVerou

svgeesus:
Note - the gamut mapping section does not yet exist, because gamut mapping in CIE LCH gave unacceptable results in the 270-330 degree hue range and also because deltaE 2000 is complex, but is needed for good results and deltaE 76 is fast but not good enough.

Using OKLCH and deltaEOK (which is the simple root-rum-of-squares) gives better results and is computationally simple.

argyleink:
lgtm, nice attention to detail

facelessuser:
[...]

PR #10063, opened 2024-03-12
[css-color-4] Update ΔEOK to be even more perceptually uniform

foolip:
In a discussion with @bottosson, I learned that he considers the scale of the _a_ and _b_ coordinates a mistake in the design of Oklab. Orthogonality between _L_, _a_, and _b_ was a design goal, but the scale of the coordinates was not carefully chosen. They should be roughly doubled to make a change in _a_ or _b_ as noticeable as the same change in _L_.

I don't think this is documented anywhere, so I'm hoping for confirmation here. Perhaps I've also misunderstood.

If correct, then it's too late to change the coordinates, but ΔEOK should probably take this into account. An update to the [preceding prose](https://drafts.csswg.org/css-color-4/#color-difference-OK) would also be needed, but I'm leaving that until the math itself is confirmed to be correct, in case it's not.

### DIFF
--- a/css-color-4/deltaEOK.js
+++ b/css-color-4/deltaEOK.js
@@ -9,7 +9,7 @@ function deltaEOK (reference, sample) {
     let [L1, a1, b1] = reference;
 	let [L2, a2, b2] = sample;
 	let ΔL = L1 - L2;
-	let Δa = a1 - a2;
-	let Δb = b1 - b2;
+	let Δa = 2 * (a1 - a2);
+	let Δb = 2 * (b1 - b2);
 	return Math.sqrt(ΔL ** 2 + Δa ** 2 + Δb ** 2);
 }


bottosson:
This was discussed previously back in 2021: https://github.com/w3c/csswg-drafts/issues/6642#issuecomment-945096257

The linked posted and the subsequent ones provide more detail.

foolip:
Thank you @bottosson! Quoting what I think are the most important bits.

First @bottosson said:

> I unfortunately didn't spend that much time calculating and validating that scaling factor when I first derived Oklab since I was mostly focused on the orthogonality between L, C and h (and I didn't expect it to become so widespread so quickly), and it seems like it is off by quite a bit.
>
> I've recently done some tests with color distance datasets as implemented in [Colorio](https://github.com/nschloe/colorio)
and on both the [Combvd dataset](https://www.osapublishing.org/josaa/abstract.cfm?uri=josaa-25-7-1828) and the [OSA-UCS dataset](https://www.osapublishing.org/josa/abstract.cfm?uri=josa-64-12-1691) a scale factor of slightly more than 2 for a and b would give the best results (2.016 works best for Combvd and 2.045 for the OSA-UCS dataset).

Then @svgeesus said:

> - scaling OKLab a & b relative to L solely based on deltaE 2000 comparisons on the neutral axis picks the point of maximum CIELAB a & b asymmetry
> - scaling based on some dataset of color pairs will depend on the distribution of mean chroma in that data set

Finally @bottosson said:

> Small scale color distances show very strong elongation of spheres into ellipses:
> image
>
> [...]
>
> It is also possible to much closer match the small color difference dataset as well, by introducing chroma compression, similar to what both CIEDE2000 and Ciecam02/16-UCS do, but that also adds quite a bit of complexity and assumptions on viewing conditions, this performs worse for tasks such as color interpolation

The discussion was in the context of changing Oklab, and that's no longer possible. For computing color distance, however, what's the takeaway from this? Is the "right" scale definitely something close to 2, or is "chroma compression" necessary to avoid the "very strong elongation of spheres into ellipses"?

svgeesus:
> For computing color distance, however, what's the takeaway from this? Is the "right" scale definitely something close to 2, or is "chroma compression" necessary to avoid the "very strong elongation of spheres into ellipses"?

The drawbacks of CAM16-like chroma compression (complexity, makes it worse for tasks like color interpolation (gradients color-mix()) seem to argue for the simpler uniform scaling.

I would like to see some experimental implementation and, ideally, results on other datasets including those with higher mean chroma, before a decision is made.

svgeesus (commented):
Needs more discussion and experience before this change is merged

foolip:
> I would like to see some experimental implementation

Do you mean implementation of this color distance method, or something that demonstrates that 2 is roughly the right number?

svgeesus:
I mean something that demonstrates that 2 is the right number, for all real physical colors. We don't want to be changing the scaling factor again in a couple of years.

Sadly since @bottosson commented in 2021, **colorio** has become [**coloria**](https://github.com/coloria-dev/coloria) and now [requires a $50 per year license](https://github.com/coloria-dev/coloria/issues/207#issuecomment-1450344188).

svgeesus:
We have now tracked down the COMVB dataset (which is actually a merger of four earlier datasets). No-one seems to have the [Xu et. al WCG dataset](https://www.researchgate.net/publication/349331250_Testing_uniform_colour_spaces_using_colour_differences_of_wide_colour_gamut).

> On the right number, I found it compelling that two different datasets gave very similar numbers, but don't know what's in those datasets.

It is, so around 2 would be a good starting point, but they are all standard-gamut colors, no wide gamut colors.

About changing the definition of DeltaEOK:

>  I think the only use of it is in https://drafts.csswg.org/css-color-4/#css-gamut-mapping, a

The only use of it *in CSS*, yes. CSSWG does not own DeltaE OK. 

> Are there others users of ΔEOK already that need to be aware of the change?

Certainly. And such a change would need to be made with @bottosson involvement and, ideally, announcing that change. I would also suggest calling it something like `DeltaE OK2` to make it an opt-in rather than a possibly surprising change as people using it update their libraries.

svgeesus:
> If we're not going to adjust for chroma compression with a non-linear function, is anchoring to "all possible surface colors" perhaps better than "all physically real colors"?

The Web is typically not consumed as printed media or physical objects. P3 screens are now widespread, and even sRGB has colors outside Pointer's gamut. 

Here is BT Rec.709 which has he same primaries as sRGB 

figure9_sRGB and pointer CIE1976

and here is DCI P3 which has the same primaries as Display P3

figure15_DCI-P3 and pointer CIE1976

svgeesus:
Here is the coverage of COMBVD (in black) compared to the two Xu et al WCG datasets (in red and green). Gamut of sRGB and P3 as guides.

image

svgeesus:
I compared DeltaE vs DeltaE2 on the Sharma color-difference dataset

<cite>The CIEDE2000 Color-Difference Formula: Implementation Notes, Supplementary Test Data, and Mathematical Observations</cite>, G. Sharma, W. Wu, E. N. Dalal, <em>Color Research and Application</em>, vol. <strong>30</strong>. No. 1, pp. 21-30, February 2005

scatter-plot(1)

foolip:
Very cool, @svgeesus, putting the theory to the test. I'm not sure how to read the figure, does it reveal which method is "more perceptual" or "better" in some sense?

bottosson:
> Certainly. And such a change would need to be made with @bottosson involvement and, ideally, announcing that change. I would also suggest calling it something like DeltaE OK2 to make it an opt-in rather than a possibly surprising change as people using it update their libraries

I think it makes sense to look at creating a "DeltaE OK2" since the error is as large as it is. I'm happy to announce the new distance once settled.

Regarding wide gamut colors: This will end up with some kind of tradeoff, if not introducing chroma compression. Without chroma compression tuning for differences close to a,b=0,0 will result in a different scale than tuning for higher chroma colors, and I don't think introducing chroma compression makes sense, it goes against the simplicity of Oklab.

I'm not convinced that it is best to tune for some kind of average across all possible colors rather than to just focus on more commonly occurring colors. Even if displays eventually reach the limits of the gamut of possible colors, the most common colors, and the ones where color distance will matter the most will be commonly occurring colors, those we see all the time in photos, graphical design etc, real world materials etc.

foolip:
@svgeesus would you like me to send a PR adding ΔEOK2 as a new definition instead?

svgeesus:
The PR is trivial and is not the limiting factor here.

@bottosson I started to investigate Macadam limits to generate a set of real-world (surface) colors. I was planning to generate perturbations in, say, CIECAM16 and then compare the deltaE 2000 of those pairs with the deltaOK and proposed deltaOK2; assumption being that a higher correlation meant it was a better distance metric. Does that seem a good approach?

bottosson:
> The PR is trivial and is not the limiting factor here.
> 
> @bottosson I started to investigate Macadam limits to generate a set of real-world (surface) colors. I was planning to generate perturbations in, say, CIECAM16 and then compare the deltaE 2000 of those pairs with the deltaOK and proposed deltaOK2; assumption being that a higher correlation meant it was a better distance metric. Does that seem a good approach?

Yeah, should work well to check the overall performance. 

If you want to tune the constant rather than just check it, you need quite careful with the selection of test colors. Shouldn't matter much if just validating though.

dom1n1k:
> it's too late to change the coordinates, but ΔEOK should probably take this into account

I know, it's debatable. But I disagree. I think we need to modify the coordinates.

OKLab was launched just 3-4 years ago. But we have at the moment:
1. Several matrices variants
2. Unclear white point
3. Two deltaE formulas

This is very confusing. What is correct? What is actual? What is standard?
There are already many uncertainties, but still relatively small legacy.

I guess we need official "good" OKLab v2.
Which will include all the improvements and will be well tested.
And let the current version be deprecated.

svgeesus:
> Yeah, should work well to check the overall performance.
> 
> If you want to tune the constant rather than just check it, you need quite careful with the selection of test colors. Shouldn't matter much if just validating though.

I (finally) got around to checking the overall performance, and OK2 is clearly better on a synthetic dataset (normally distributed base colors in CAM16 JMh restricted to BT.2020 gamut, normally distributed small perturbations for the comparison colors)

> ΔEOK2 — ΔEOK with the Oklab a and b axes scaled by 2 — agrees with ΔE2000 noticeably better than ΔEOK does (r² 0.722 vs 0.642 on 5000 BT.2020 color pairs). Scaling the a/b axes by 2 is a measurable improvement, supporting Ottosson's recommendation.

[Write up and code available if anyone wants to play with it](https://github.com/svgeesus/deltaE-OK2/blob/main/README.md)

svgeesus:
@bottosson do the results I got on this synthetic dataset align with your findings on Combvd  and OSA-UCS datasets? Mine is intended to be wider gamut than those, while staying firmly in the realm of physically realizable colors.

@foolip I will leave some time for comments but all being well, would then like to 

-  [x] add the ΔEOK2 code to CSS Color 4 (separate from the existing ΔEOK)
-  [x] recommend using ΔEOK2 over ΔEOK
-  [ ] check some of the gamut mapping code to see if using ΔEOK2 gives better results (and if so, change those algorithms to use it)

See https://github.com/w3c/csswg-drafts/commit/af8338edf72e154f05b57d4c199906358e684db0

foolip:
That's awesome, thanks @svgeesus! Nice to also see the lightness fix in ΔEOKr2. Is this the same lightness as on Okhsl, or derived independently?

facelessuser:
I believe details are here: https://github.com/w3c/csswg-drafts/issues/14207, but yes, it was the same lightness as specified in OkLrab, which is used in Okhsl. You could certainly derive a lightness closer, but that is not what was done.

foolip:
Since the spec changes have been made, I'll close this PR.

PR #10072, opened 2024-03-13
[css-view-transitions-2] Apply resolutions for view-transition types

### LINKED ISSUE #9526: [css-view-transitions-2] Behavior of mismatching types between old and new document
As per https://github.com/w3c/csswg-drafts/pull/9523, both documents can specify "types" for the cross document view-transition, which affects the `:active-view-transition` selector.
Note that these types apply both at the old and new document, as they can affect (for example) which elements are captured.

When capturing the old document, only types specified on that document can apply.
On the new document, we can do one of the following:
1. apply only types specified in the new document (override)
2. apply both the types in the old and new documents (union)
3. apply only the types that are in both documents (intersection)
4. fail the transition if the types don't match

The current PR specified (1), the advantage being that it's the most flexible and least opinionated - if developers want to have different types in both documents they can do it, and if they want them to be strictly the same they can restrict it themselves, there is no technical reason to have this restriction or to force a union.

There are also advantages to the other approaches, perhaps mandating the same types would lead to design that is less error-prone.

### LINKED ISSUE #9542: [css-view-transitions-2] Allow changing `types` using JS after (MPA) transition started?
Currently we're only allowing to set the transition types at `startViewTransition`, and they can get changed when revealing the new document in a cross-document view transition.

But perhaps we can allow the author to read or change the types programmatically, as a `DOMTokenList` on the VT object?
Something like this:

```js
document.addEventListener("pagereveal", event => {
  const {viewTransition} = event;
  const hasSlideIn = viewTransition.typeList.has("slide-in");
  if (heroImage.loaded) {
     viewTransition.typeList.toggle("expand-hero", true);
  }
});
```

Note that this is possible to do today using the CSSOM, but not dynamically while the VT is already running.

### LINKED ISSUE #9626: [css-view-transitions-2] Allow empty :active-view-transition() ?
In general, if there's reasonable default behavior, we allow omitting arguments, so... should `:active-view-transition()` be equivalent to `:active-view-transition(*)`?

### LINKED ISSUE #9972: [css-view-transitions-2] Rename `:active-view-transition()` pseudo-class
The current name is a bit confusing and makes it very tempting to put a view-transition name instead of type as argument. This is the type of thing that can typically confuse beginners very easily.

I'd like to suggest renaming to something like `:view-transition-type()` or such.

cc @vmpstr @noamr @khushalsagar

noamr:
- Make types mutable using ViewTransition.typeList https://github.com/w3c/csswg-drafts/issues/9542#issuecomment-1994829341

- Separate `:active-view-transition` and `:active-view-transition-type` See https://github.com/w3c/csswg-drafts/issues/9972#issuecomment-1994889933

- Add notes about types being bound to one document See https://github.com/w3c/csswg-drafts/issues/9526#issuecomment-1994844676

Closes #9972
Closes #9542
Closes #9526
Closes #9626

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -290,39 +290,51 @@ for multiple view transition pseudo elements without having to replicate the cor
 
 # Pseudo-classes # {#pseudo-classes}
 
-## Active View Transition Pseudo-class '':active-view-transition()''' ## {#the-active-view-transition-pseudo}
+## Active View Transition Pseudo-class '':active-view-transition()'' ## {#the-active-view-transition-pseudo}
 
-The <dfn id='active-view-transition-pseudo'>:active-view-transition(<<vt-type-selector>>)</dfn> pseudo-class applies to the root element of the document, if it has a matching [=active view transition=].
+The <dfn id='active-view-transition-pseudo'>:active-view-transition</dfn> pseudo-class applies to the root element of the document, if it has an [=active view transition=].
 It has the following syntax definition:
 
 <pre class=prod>
-	:active-view-transition(<<vt-type-selector>>)
-	<dfn>&lt;vt-type-selector></dfn> = '*' | <<custom-ident>>#
+	:active-view-transition
 </pre>
 
-The [=specificity=] of an '':active-view-transition()'' is one pseudo-class selector if it has value is ''*'', and two if it has any other value.
+The [=specificity=] of an '':active-view-transition'' is one pseudo-class selector.
 
-An '':active-view-transition()'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=] |viewTransition|, for which any of the following are true:
+An '':active-view-transition'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=].
 
-* The <<vt-type-selector>> is ''*''
-* |viewTransition|'s [=ViewTransition/active types=] [=list/contains=] at least one of the <<custom-ident>> values of the <<vt-type-selector>>.
+## Active View Transition Type Pseudo-class '':active-view-transition-type()'' ## {#the-active-view-transition-type-pseudo}
+
+The <dfn id='active-view-transition-type-pseudo'>:active-view-transition-type()</dfn> pseudo-class applies to the root element of the document, if it has a matching [=active view transition=].
+It has the following syntax definition:
+
+<pre class=prod>
+	:active-view-transition-type(<<custom-ident>>#)
+</pre>
+
+The [=specificity=] of an '':active-view-transition-type()'' is two pseudo-class selectors.
+
+An '':active-view-transition-type()'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=],
+whose [=ViewTransition/active types=] [=list/contains=] at least one of the <<custom-ident>> arguments.
+
+## Examples ## {#active-view-transition-pseudo-examples}
 
 <div class=example>
 For example, the developer might start a transition in the following manner:
 ```js
-document.startViewTransition({update: updateTheDOMSomehow, types: ["slide-in", "reverse"]});
+document.startViewTransition({update: updateTheDOMSomehow, type: ["slide-in", "reverse"]});
 ```
 
-This will activate any of the following ':active-view-transition()'' selectors:
+This will activate any of the following ':active-view-transition-type()'' selectors:
 ```css
-:root:active-view-transition(slide-in) {}
-:root:active-view-transition(reverse) {}
-:root:active-view-transition(slide-in, reverse) {}
-:root:active-view-transition(slide-in, something-else) {}
-:root:active-view-transition(*) {}
+:root:active-view-transitiontype(slide-in) {}
+:root:active-view-transition-type(reverse) {}
+:root:active-view-transition-type(slide-in, reverse) {}
+:root:active-view-transition-type(slide-in, something-else) {}
+:root:active-view-transition {}
 ```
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -290,39 +290,51 @@ for multiple view transition pseudo elements without having to replicate the cor
...
 
-This will activate any of the following ':active-view-transition()'' selectors:
+This will activate any of the following ':active-view-transition-type()'' selectors:
 ```css
-:root:active-view-transition(slide-in) {}
-:root:active-view-transition(reverse) {}
-:root:active-view-transition(slide-in, reverse) {}
-:root:active-view-transition(slide-in, something-else) {}
-:root:active-view-transition(*) {}
+:root:active-view-transitiontype(slide-in) {}
```
khushalsagar, line 330:
```suggestion
:root:active-view-transition-type(slide-in) {}
```

css-view-transitions-2/Overview.bs
```diff
@@ -473,7 +488,7 @@ Note: as per default behavior, the ''@view-transition'' rule can be nested insid
 
 		1. Let |viewTransition| be the result of running the [=method steps=] for {{Document/startViewTransition(updateCallback)}} given |updateCallback|.
 
-		1. If |callbackOptions| is a {{StartViewTransitionOptions}}, then set |viewTransition|'s [=ViewTransition/active types=] to |callbackOptions|'s {{StartViewTransitionOptions/type}}.
+		1. If |callbackOptions| is a {{StartViewTransitionOptions}}, then [=list/extend=] |viewTransition|'s [=ViewTransition/active types=] with |callbackOptions|'s {{StartViewTransitionOptions/type}}.
```
khushalsagar, line 491: nit: At every instance active types is updated, its value is overwritten completely instead of appending. Can we say that instead of extend to make that obvious?

css-view-transitions-2/Overview.bs
```diff
@@ -529,13 +544,23 @@ The {{CSSViewTransitionRule}} represents a ''@view-transition'' rule.
...
 
+	The {{ViewTransition}} interface is extended as follows:
+
+<pre class='idl'>
+partial interface ViewTransition {
+	attribute DOMTokenList typeList;
+};
+</pre>
+
+	The {{ViewTransition/typeList}} [=getter steps=] are to return [=this=]'s [=ViewTransition/active types=].
```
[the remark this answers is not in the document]
noamr: No, it's a readonly `DOMTokenList`

css-view-transitions-2/Overview.bs
```diff
@@ -628,7 +653,7 @@ Prepend this to the [=Perform pending transition operations=] algorithm given a
 
 		1. If |typesDescriptor|'s [=computed value=] is ''@view-transition/type/none'', then return a [=/list=] « ».
 
-		1. Return a [=list=] of strings corresponding to |typesDescriptor|'s [=computed value=].
+		1. Return a [=/set=] of strings corresponding to |typesDescriptor|'s [=computed value=].
```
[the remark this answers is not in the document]

css-view-transitions-2/Overview.bs
noamr: It's the proper way to copy a list into an existing `set` (the token set). How would you reword it?

css-view-transitions-2/Overview.bs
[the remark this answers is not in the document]
noamr: ... made it readonly, that was indeed missing

css-view-transitions-2/Overview.bs
noamr: Actually let me try something.

css-view-transitions-2/Overview.bs
```diff
@@ -290,39 +290,51 @@ for multiple view transition pseudo elements without having to replicate the cor
...
-	:active-view-transition(<<vt-type-selector>>)
-	<dfn>&lt;vt-type-selector></dfn> = '*' | <<custom-ident>>#
+	:active-view-transition
 </pre>
 
-The [=specificity=] of an '':active-view-transition()'' is one pseudo-class selector if it has value is ''*'', and two if it has any other value.
+The [=specificity=] of an '':active-view-transition'' is one pseudo-class selector.
 
-An '':active-view-transition()'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=] |viewTransition|, for which any of the following are true:
+An '':active-view-transition'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=].
```
khushalsagar, line 304: nit: when it's Document has a non-null ....

css-view-transitions-2/Overview.bs
```diff
@@ -290,39 +290,51 @@ for multiple view transition pseudo elements without having to replicate the cor
...
+The <dfn id='active-view-transition-type-pseudo'>:active-view-transition-type()</dfn> pseudo-class applies to the root element of the document, if it has a matching [=active view transition=].
+It has the following syntax definition:
+
+<pre class=prod>
+	:active-view-transition-type(<<custom-ident>>#)
+</pre>
+
+The [=specificity=] of an '':active-view-transition-type()'' is two pseudo-class selectors.
+
+An '':active-view-transition-type()'' pseudo-class matches the [=document element=] when it has an non-null [=active view transition=],
```
khushalsagar, line 317: Same

css-view-transitions-2/Overview.bs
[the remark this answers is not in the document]
khushalsagar: Ah ok. So the attribute is readonly but you can mutate the DOMTokenList.

khushalsagar (approved):

PR #10078, opened 2024-03-14
[css-view-transitions-2] Update the algorithm to check opt-in on outbound transition

noamr:
Instead of relying on the source snapshot params in HTML, we update a flag every time something relevant changes, and HTML can read that flag.

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -450,6 +450,11 @@ Note: as per default behavior, the ''@view-transition'' rule can be nested insid
 	Note: the [=@view-transition/type=] descriptor only applies to the {{Document}} in which it is defined.
 	The author is responsible for using their chosen set of types in both documents.
 
+## Responding to changes in the ''@view-transition'' rule ## {#respond-to-rule-changes}
+
+When the ''@view-transition'' rule changes for {{Document}} |document|, [=update the opt-in state for outbound transitions=] given |document|.
+
+
 # API # {#api}
 
 ## Additions to {{Document}} ## {#additions-to-document-api}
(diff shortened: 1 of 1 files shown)


noamr:
This is required for https://github.com/whatwg/html/pull/10178

css-view-transitions-2/Overview.bs
```diff
@@ -450,6 +450,11 @@ Note: as per default behavior, the ''@view-transition'' rule can be nested insid
 	Note: the [=@view-transition/type=] descriptor only applies to the {{Document}} in which it is defined.
 	The author is responsible for using their chosen set of types in both documents.
 
+## Responding to changes in the ''@view-transition'' rule ## {#respond-to-rule-changes}
+
+When the ''@view-transition'' rule changes for {{Document}} |document|, [=update the opt-in state for outbound transitions=] given |document|.
```
khushalsagar, line 455: Should we leave a note here to the effect of, "The UA is responsible for deciding when style resolution is executed to keep the opt-in state in sync with style mutations".

css-view-transitions-2/Overview.bs
```diff
@@ -736,6 +747,10 @@ Prepend this to the [=Perform pending transition operations=] algorithm given a
 
 		1. [=Assert=]: |document| is [=fully active=].
 
+		1. [=Assert=]: |document| [=has been revealed=] is true.
```
khushalsagar, line 750: Do we still need to export this algorithm? It looks like for the new Document we're calling it within this spec in "resolve cross-document view-transition" and html should now use the `can initiate outbound view transition` bit.

noamr: Export which algorithm?

css-view-transitions-2/Overview.bs
noamr: I don't think that's necessary.

css-view-transitions-2/Overview.bs
khushalsagar: "resolve cross-document view-transition" shouldn't need to be visible outside this spec now. Can fix after the html rewrite lands.

khushalsagar (approved):

css-view-transitions-2/Overview.bs
noamr: It's still called from page reveal

noamr: What we don't need is "Resolve @view-transition rule", it's already not exported.

noamr: I renamed it, I think it's clearer now

PR #10113, opened 2024-03-21
chore: typo in css-scrollbars

MrHBS:
@frivoal I think you meant to say paint here?


frivoal:
I did. Thanks for spotting this.

Issue #4957, opened 2020-04-16
[css-align-3] What is supposed to happen to abspos in an end-aligned scroll container?


tabatkins:
Consider this scenario:

```
<container style="display:block; height:100px; overflow:auto; position:relative;">
  <child style="height:200px;"></child>
  <abspos style="position:absolute; left:0; top:0; height:500px;"></abspos>
</container>
```

Layout's well-defined so far. Scrollable overflow area is 500px tall, initial scroll position is 0, etc.

Now set "align-content: end;" on the container.

The rules in the spec say that, since the container is a scroll container, you do *not* move the alignment subject at all. Instead, you adjust the initial scroll position of the scroll container *such that* the alignment subject is aligned *relative to the scrollport* the same that it would be aligned relative to the container if it weren't scrollable. The container is thus scrolled 100px down, so the bottom of the child is aligned with the bottom of the scrollport.

In other words, if you toggle 'overflow' on and off, the content should sit perfectly still, while remaining fully accessible in the scrollable case (that is, never getting shifted into the unscrollable area).

And this goal is achieved! For the in-flow content, at least.  It does *not* work for the abspos. Because the abspos is positioned relative to the scrollable overflow area, it jumps as you toggle overflow on and off, and the container is either scrolled or not.  This violates the original design goal.

I only see one possible fix: Abspos descendants are shifted by content alignment, so they maintain the same positioning relative to the alignment subject. (Increasing the scrollable overflow area, but I think never pushing anything into the unscrollable zone.)

Alternately, we can just ignore it, and accept that the original design goal only applies to in-flow content, and the relative positioning of in-flow and abspos content can change depending on 'overflow'.

Thoughts?

MatsPalmgren:
> `<container style="display:block; height:100px; overflow:auto; position:relative;">`
 
Did you mean `display:grid` perhaps?

>   `<child style="height:200px;"></child>`

Otherwise, `<child>` is an inline and `height` does not apply.

tabatkins:
I meant display:block, but pretend these are all blocks. ^_^

fantasai:
Testcase to try to demonstrate what we're talking about: http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8069

tabatkins:
[...]

PR #10139, opened 2024-03-25
[css-align-3][css-overflow-3] Content alignment on scroll containers extends scrollable area #4957

fantasai:
See #4957.

Review questions:
* Should we collapse the concepts of "scroll origin position" and "initial scroll position" now that they coincide even in the presence of alignment? If so, which word should we use?
* I think we can close the two issues marked in https://drafts.csswg.org/css-overflow-3/#scrolling ?
* Having the[ "The scrollable overflow area is the union of:" list](https://drafts.csswg.org/css-overflow-3/#scrollable) before the [definition of key terms like “scroll position”](https://drafts.csswg.org/css-overflow-3/#scrolling) seems wrong. Should I pull it out into its own subsection? Where?

### DIFF
--- a/css-align-3/Overview.bs
+++ b/css-align-3/Overview.bs
@@ -897,7 +897,7 @@ Content Distribution for Scroll Containers</h5>
 <h5 id=auto-safety-position>
 Self-Alignment for Absolutely Positioned Boxes</h5>
 
-	For [=absolutely positioned=] [=alignment subjects=] 
+	For [=absolutely positioned=] [=alignment subjects=]
 	whose relevant [=self-alignment property=] is not ''justify-self/normal'',
 	the default [=overflow alignment=] behavior is as follows:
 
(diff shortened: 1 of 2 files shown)


tabatkins:
> Should we collapse the concepts of "scroll origin position" and "initial scroll position" now that they coincide even in the presence of alignment? If so, which word should we use?

Yes, we should collapse. For the term, I forget whether we're actually adjust the scroll origin (so 0 is at the end, or whatever) or if we're just starting it out at a non-zero value. Which is the case, we should lean toward that term imo.

> I think we can close the two issues

Is the baseline one answered? I don't see anything in the edits apparently addressing that, but maybe it falls out of them in a more general way.

> Having the "The scrollable overflow area is the union of:" list before the definition of key terms like “scroll position” seems wrong. Should I pull it out into its own subsection? Where?

Can you just swap the sections? It's already practically the entire subsection anyway, I don't think dropping it into a further subsection would really help.

fantasai:
Rebased. Note: Given the addition of `scroll-initial-target`, I think we need to distinguish the [=scroll origin position=] and the [=initial scroll position=].

fantasai:
Split out the ordering question into https://github.com/w3c/csswg-drafts/issues/11895

I'm going to merge this now.

PR #10166, opened 2024-04-01
[css-color-4] Fix a typo (profoto-rgb -> prophoto-rgb)

nex3:


nex3:
Yep, just figured that out :B

PR #10174, opened 2024-04-04
[css-scrollbars] Remove example about IE

frivoal:
While the example isn't inherently wrong, it's of little relevance to present-day authors, as Internet Explorer hasn't been supported for years, and barely has any usage anymore.


tantek (approved):
Good catch.

PR #10181, opened 2024-04-08
[css-view-transitions-1] Use snapshot containing block for new root

### LINKED ISSUE #10177: [css-view-transitions-1] What is the size of the group pseudo for the root element?
From @mattwoodrow from investigating https://bugs.webkit.org/show_bug.cgi?id=272186:

Chrome seems to be setting the size of their ::view-transition-group element to that of the snapshot containing block. That seems wrong, but visually what you’d expect to happen.

Capture the image algorithm has a special case for sizing the image when capturing the root, I suspect Update pseudo-element styles should have the equivalent wording for setting the width/height style properties.

noamr:
This aligns with current behavior, but was overlooked in the spec.

Closes #10177

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1847,7 +1847,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 					Note: Other rendering constraints are enforced via |capturedElement|'s [=new element=] being [=captured in a view transition=].
 
 				1. Let |newRect| be [=snapshot containing block=] if |capturedElement| is the [=document element=],
-					otherwise, |capturedElement's [=border box=].
+					otherwise, |capturedElement|'s [=border box=].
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1846,11 +1846,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 					Note: Other rendering constraints are enforced via |capturedElement|'s [=new element=] being [=captured in a view transition=].
 
-				1. Set |width| to the current width of |capturedElement|'s [=new element=]'s [=border box=].
+				1. Let |newRect| be [=snapshot containing block=] if |capturedElement| is the [=document element=],
+					otherwise, |capturedElement's [=border box=].
```
khushalsagar, line 1850:
```suggestion
					otherwise, |capturedElement|'s [=border box=].
```

Issue #9786, opened 2024-01-11
[css-view-transitions-1] What is the snapshot containing block for iframes?


khushalsagar:
The spec [here](https://drafts.csswg.org/css-view-transitions-1/#capture-rendering-characteristics-algorithm:~:text=However%2C%20the%20captured%20image%20should%20include%2C%20at%20the%20very%20least%2C%20the%20contents%20of%20element%20that%20intersect%20with%20the%20snapshot%20containing%20block.) says that if an element is too large for its entire image to be captured, implementations can clip the image as long as the contents in the snapshot containing block are captured.

But for definition of snapshot containing block is ambiguous for iframes: "all areas of the window that could potentially display page content". The impact it has on clipping of images aside, I think it should be the iframe's ICB. @bokand can explain this better.

But this means we need a better way to decide how to clip the image if the transition is happening in an iframe. The goal with the SCB recommendation was so we at least capture the onscreen content the user is seeing but the whole iframe could be offscreen. I'm thinking the following:

- If the iframe is completely offscreen, just skip the transition. Why animate instead of directly flipping to the new state.
- If its onscreen (fully or partially) use the SCB of the top level frame to decide what should be captured in the images.

khushalsagar:
> Why not just do the simple thing and apply the clip in the same way we do for non-root elements? i.e. take the above definition of the snapshot containing block for an iframe and also clip the root to it.

We already clip the root to SCB, see [here](https://drafts.csswg.org/css-view-transitions-1/#capture-rendering-characteristics-algorithm:~:text=Render%20the%20region%20of%20document%20(including%20its%20canvas%20background%20and%20any%20top%20layer%20content)%20that%20intersects%20the%20snapshot%20containing%20block). This was based on the assumption that SCB is always a reasonable size. That's true for main frames but not iframes since there the SCB can be any bounds chosen by the parent frame.

khushalsagar:
> This information might also need to cross process boundaries so could be out of date. It feels rather complex for something that seems to me like an extreme edge case?

FWIW the info has to be available to cross-process iframes because of IntersectionObserver (which needs to know whether elements in an iframe are visible to the user). But its a fair point that it can be stale. For VT the latency in updating it will be visually noticeable to the user. If this all sounds too complex we can just say that implementations can skip the transition if SCB exceeds an implementation-defined maximum.

khushalsagar:
[...]

PR #10188, opened 2024-04-09
[css-view-transitions-1] More precisely define snapshot containing block in subframes #9786

bokand:
[css-view-transitions-1] More precisely define snapshot containing block in subframes #9786

The intent of the current definition is that the snapshot containing block in a child frame should be a rect at the origin, the same size as the frame, including any scrollbars.

Initial containing block is misleading in 2 ways - it doesn't include scrollbars in its size and its position changes as the page scrolls so a strict reading would make the snapshot containing block's position dependent on the scroll offset.

I think [viewport](https://www.w3.org/TR/CSS22/visuren.html#viewport) is the better definition here but I believe that excludes scrollbars as well so we explicitly union it with [scrollbar gutters](https://drafts.csswg.org/css-overflow/#scrollbar-gutter-property).

Also, refer to [child navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#child-navigable) rather than iframes since there's no need to limit this to just iframes (e.g. `<fencedframes>`? `<frame>`?)

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -41,6 +41,7 @@ spec:css22; type:dfn; text:element
 spec:css-break-4; type:dfn; text:fragment
 spec:css-viewport; type:dfn; text:interactive-widget;
 spec:css-display-4; type: dfn; text:invisible;
+spec:css2; type:dfn; text:viewport
 </pre>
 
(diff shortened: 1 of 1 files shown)


khushalsagar (approved):
@nt1m fyi.

Issue #10083, opened 2024-03-14
[css-pseudo][css-shadow-parts] Define "part-like pseudo-element" concept


tabatkins:
In the WHATNOT joint session with CSSWG today, we resolved that the new `details` pseudos would indeed be exposed as normal UA pseudo-elements (rather than `::part()`s), but that we'd define a new concept of "part-like pseudo-elements" that grants the pseudo the abilities of ::part() that we want to preserve.

Namely:
* a part-like pseudo has the ::part() blocklist of a few pseudo-classes (the tree-structural ones) and allows all other pseudo-elements and pseudo-classes, rather than the normal pseudo *allow*list of a few pseudo-classes (the logical-combination and user-action ones) and banning all the rest.
* a part-like pseudo has a well-defined inheritance relationship with the originating element and other part-like pseudos, rather than the normal "always inherit from the originating element" behavior.
* the `exportparts` attribute will gain a syntax for exporting part-like pseudos, details TBD
* we categorize future pseudo-elements as "part-like" when appropriate, and do a review of existing pseudos for acceptability of retroactively labeling them "part-like" and giving them these new powers

--------

IIRC, the only real point of contention left is that, for normal parts, you *only* have access to the parts declared or exported by the component itself. If you expose a sub-component as a part, you *do not* get access to its parts too (that is, you can't ever write `my-el::part(my-foo)::part(their-bar)`), specifically so we don't expose the fact that the part is implemented as a nested component rather than plain DOM. But parts *do* provide access to their pseudo-elements (you *can* write `my-el::part(my-foo)::before`), because that's useful and many pseudo-elements aren't remotely appropriate to be exported as parts themselves.

But these two points are now in conflict. If you *can* export a part-like pseudo to be one of your component's parts, but you can *also* access it as a pseudo-element, it means you can spell the same selector two distinct ways: `my-el::part(my-content)` and `my-el::part(my-details)::details-content`.

This *does* expose the fact that a part is a particular built-in element, but that's a fact we already explicitly exposed, so it's not a new information leak. (Any element-specific pseudo-classes also expose this.) 

It also means there's two ways to refer to the same element, but that's generally true of selectors, so I don't find this a particularly compelling problem.

I suggest, then, that we simply accept this as a possibility. But we *could* try to work around it in a few ways:

[...]

PR #10199, opened 2024-04-11
Define "part-like pseudo-element", from #10083.

tabatkins:
@fantasai review?

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -1147,6 +1147,20 @@ Tree-Abiding Pseudo-elements</h2>
   non-inheritable properties take their <a>initial values</a> as usual.
   [[CSS-CASCADE-4]]
 
+<h3 id=partlike>
+Part-Like Pseudo-Elements</h3>
+
+  A subset of [=tree-abiding pseudo-elements=],
+  the <dfn>part-like pseudo-elements</dfn>,
+  have slightly stronger requirements:
+  they act as if they have a well-defined location in the document tree.
+  This enables them to interact with some other platform features
+  as if they were real elements.
+
+  Unless otherwise specified,
+  any CSS property that applies to elements
+  applies to [=part-like pseudo-elements=].
+
 <h3 id="generated-content">
 Generated Content Pseudo-elements: ''::before'' and ''::after''</h3>
 
@@ -1165,10 +1179,9 @@ Generated Content Pseudo-elements: ''::before'' and ''::after''</h3>
     immediately after the <a>originating element</a>'s actual content.
   </dl>
 
-  These pseudo-elements can be styled
-  exactly like any normal document-sourced element in the document tree;
-  all properties that apply to a normal element
-  likewise apply to ''::before'' and ''::after''.
+  Both ''::before'' and ''::after''
+  are [=part-like pseudo-elements=];
+  there is no restriction on what properties apply to them.
 
   <div class="example">
     For example, the following rule inserts the string “Note: ”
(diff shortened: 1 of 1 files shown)


css-pseudo-4/Overview.bs
```diff
@@ -1147,6 +1147,20 @@ Tree-Abiding Pseudo-elements</h2>
...
 
+<h3 id=partlike>
+Part-Like Pseudo-Elements</h3>
+
+  A subset of [=tree-abiding pseudo-elements=],
+  the <dfn>part-like pseudo-elements</dfn>,
+  have slightly stronger requirements:
+  they act as if they have a well-defined location in the document tree.
+  This enables them to interact with some other platform features
+  as if they were real elements.
```
fantasai, line 1158: What's an example of this, other than their full participation in CSS layout?

css-pseudo-4/Overview.bs
```diff
@@ -1147,6 +1147,20 @@ Tree-Abiding Pseudo-elements</h2>
   non-inheritable properties take their <a>initial values</a> as usual.
   [[CSS-CASCADE-4]]
 
+<h3 id=partlike>
+Part-Like Pseudo-Elements</h3>
+
+  A subset of [=tree-abiding pseudo-elements=],
+  the <dfn>part-like pseudo-elements</dfn>,
```
fantasai, line 1154: Not sure I like this name. Can we pick something that isn't a reference to `::part()`?

css-pseudo-4/Overview.bs
```diff
@@ -1147,6 +1147,20 @@ Tree-Abiding Pseudo-elements</h2>
   non-inheritable properties take their <a>initial values</a> as usual.
   [[CSS-CASCADE-4]]
 
+<h3 id=partlike>
+Part-Like Pseudo-Elements</h3>
+
```
fantasai, line 1152: I think maybe we can inline the definition into the Tree-Abiding Pseudo-elements section, it feels a bit odd maybe as its own little section given that the next section is both Tree-Abiding and Part-Like?

fantasai (changes requested):

css-pseudo-4/Overview.bs
tabatkins: Like being used in `exportparts` attribute.

css-pseudo-4/Overview.bs
tabatkins: My only issue with that is that I'd want both names to be in the title. But if that's okay, then yeah, let's just fold them together.

css-pseudo-4/Overview.bs
tabatkins: Yup, would love a different name, but having a problem coming up with one. Problem is coming up with something that's not *just* a synonym of "abiding". We could try to skirt a little farther in the "Tree-X" namespace - tree-located? tree-positioned? tree-structural?- to indicate that not only do they abide by the tree structure *generally* (insofar as having a parent/child relationship) but actually have a definite *position* in that tree (and thus know their sibling relationships too).

Or we could go further afield. Semi-real? Pseudo-real?

tabatkins: "Tree-structural" is actually really grabbing me, given its parallel with the tree-structural pseudo-classes.

fantasai: Well `::marker` has a definite position in the tree, but you're not including it in this set of things afaict. So it's not about it's tree position, it's about the fact that generates a box that behaves exactly like any other element's generated box.

css-pseudo-4/Overview.bs
```diff
@@ -1140,23 +1140,35 @@ Security and Privacy Considerations</h3>
...
+    <xmp class=html>
+      <template id=custom-element-template>
+        <p exportparts="::before : preceding-text">
+          You can style my ::before pseudo-element
+          by using ::part(preceding-text), too!
+      </template>
+    </xmp>
+  </div>
+
```
fantasai, line 1171: This isn't helping to illustrate pseudo-elements that "interact with some other platform features as if they were real elements". It's illustrating a real element acting like a pseudo-element (by being selected by one), which is kindof the opposite!

tabatkins: I think you're misunderstanding the example, because it illustrates the exact opposite. `::part(preceding-text)` is selecting a pseudo-element (the `p::before` pseudo).

css-pseudo-4/Overview.bs
tabatkins: I was under the impression that the *reason* ::marker didn't act like ::before/etc is *because* we specifically hadn't nailed down its exact tree position. If it *does* have an exact position, then *what* is the reasoning behind limiting it? (Besides applying a few one-offs, like disallowing it from being `display: list-item` or something; that's fine for this new category of pseudos to do.)

tabatkins:
Merging for now, we can figure out what the better term is later.

Issue #9932, opened 2024-02-09
[cssom-view] Specify document.caretPositionFromPoint in shadow DOM scenario


siliu1:
Current spec is vague about the behavior in shadow DOM: https://www.w3.org/TR/cssom-view-1/#dom-document-caretpositionfrompoint

Gecko's current behavior in shadow DOM:
The returned `CaretPosition`'s `offset` is the text offset of the hit tested text node and the `offsetNode` depends on the mode of the shadow tree:
  1. If the tree is created by user agent such as input element, the node is the shadow host.
  2. Otherwise, the node is the hit tested text node even if the tree is closed.

Gecko's existing behavior can address most of the need for web developers. Should we specify the behavior?

Or are there other alternatives that we should consider:
  1. `caretPositionFromPoint` do not pierce shadow DOM. Instead, put the API on `DocumentOrShadowRoot`.
  2. support closed shadow root with a new parameter: `document.caretPositionFromPoint(x, y, closedShadowRoot);`

sanketj:
@zcorpan @fantasai @tabatkins Curious if any of you have any thoughts on how we can better define the behavior of `caretPositionFromPoint` for shadow DOM scenarios?

siliu1:
Asked opinions from web developer's point of view and got some feedback. Web developer's basic requirement is to be able to get a caret position inside shadow roots. The Firefox approach (piercing through the shadow boundary) is acceptable. In fact, sticking to the already existing behavior would allow web developer to simplify the implementation. 

Given that, should we specify the already existing Firefox approach and put it in spec?

siliu1:
Based on all feedback, the best approach seems to be the `getComposedRanges()` style which takes in a set of `ShadowRoots` and `document.caretPositionFromPoint(ShadowRoot ...shadowRoots)` can return position in the given shadow roots.

The alternative approach is to put the API on `DocumentOrShadowRoot`. But the con to this approach is that the usage might be cumbersome to get a position inside nested shadow roots (need to call the API on each level of shadow root).

Web developer needs to adjust the current usage of current API to get position in a shadow root. The existing behavior violates the boundary of closed shadow tree so we should consider fixing it as well.

siliu1:
Created PR https://github.com/w3c/csswg-drafts/pull/10200 to update the spec. @zcorpan would you mind taking a look? Thank you!

zcorpan:
The PR was merged, closing.

PR #10200, opened 2024-04-11
[cssom-view] add `shadowRoots` parameter to `caretPositionFromPoint` API.

siliu1:
Per https://github.com/w3c/csswg-drafts/issues/9932, add `shadowRoots` parameter to `caretPositionFromPoint` API.

Implementation bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1889503
https://issues.chromium.org/issues/332774620
https://bugs.webkit.org/show_bug.cgi?id=172137

WPT test coverage will be added subsequently.

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1070,16 +1070,16 @@ result of running these steps:
         <dt><a>caret range</a>
         <dd>null
     </dl>
-1. Otherwise, return a <a>caret position</a> where its properties are set after running these steps:
+1. Otherwise:
 
-    1. Let <a>caret range</a> to be a collapsed {{Range}} object for the position where the text insertion point indicator whould have been inserted when applying
+    1. Let <var>caretRange</var> to be a collapsed {{Range}} object for the position where the text insertion point indicator would have been inserted when applying
         the <a>transforms</a> that apply to the descendants of the <a>viewport</a>.
-    1. Let <var>startNode</var> be the [=range/start node=] of the <a>caret range</a>, and let <var>startOffset</var> be the [=range/start offset=] of the <a>caret range</a>.
+    1. Let <var>startNode</var> be the [=range/start node=] of the <var>caretRange</var>, and let <var>startOffset</var> be the [=range/start offset=] of the <var>caretRange</var>.
     1. While <var>startNode</var> is a [=node=], <var>startNode</var>'s [=tree/root=] is a [=shadow root=], and <var>startNode</var>'s [=tree/root=] is not a [=shadow-including inclusive ancestor=] of any of <var>shadowRoots</var>, repeat these steps:
         1. Set <var>startOffset</var> to [=tree/index=] of <var>startNode</var>'s [=tree/root=]'s [=host=].
         1. Set <var>startNode</var> to <var>startNode</var>'s [=tree/root=]'s [=host=]'s [=tree/parent=].</li>
-    1. Set <a>caret node</a> in <a>caret position</a> to <var>startNode</var>, and set <a>caret offset</a> in <a>caret position</a> to <var>startOffset</var>.
-    1. Set <a>caret range</a>'s [=range/start node=] and [=range/end node=] to <var>startNode</var>, and set <a>caret range</a>'s [=range/start offset=] and [=range/end offset=] to <var>startOffset</var>.
+    1. Set <var>caretRange</var>'s [=range/start node=] and [=range/end node=] to <var>startNode</var>, and set <a>caret range</a>'s [=range/start offset=] and [=range/end offset=] to <var>startOffset</var>.
+    1. Return a <a>caret position</a> whose <a>caret node</a> is set to <var>startNode</var>, <a>caret offset</a> is set to <var>startOffset</var>, and <a>caret range</a> is set to <var>caretRange</var>.
 
 Note: The specifics of hit testing are out of scope of this
 specification and therefore the exact details of
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1072,14 +1072,16 @@ result of running these steps:
     </dl>
 1. Otherwise:
 
-    1. Let <var>caretRange</var> to be a collapsed [=range=] object for the position where the text insertion point indicator would have been inserted when applying
+    1. Let <var>caretPosition</var> be a <a for=/>tuple</a> consisting of a <var>caretPositionNode</var> (a <a for=/>node</a>) and a <var>caretPositionOffset</var> (a non-negative integer) for the position where the text insertion point indicator would have been inserted when applying
         the <a>transforms</a> that apply to the descendants of the <a>viewport</a>.
-    1. Let <var>startNode</var> be the [=range/start node=] of the <var>caretRange</var>, and let <var>startOffset</var> be the [=range/start offset=] of the <var>caretRange</var>.
+    1. Let <var>startNode</var> be the <var>caretPositionNode</var> of the <var>caretPosition</var>, and let <var>startOffset</var> be the <var>caretPositionOffset</var> of the <var>caretPosition</var>.
     1. While <var>startNode</var> is a [=node=], <var>startNode</var>'s [=tree/root=] is a [=shadow root=], and <var>startNode</var>'s [=tree/root=] is not a [=shadow-including inclusive ancestor=] of any of <var>shadowRoots</var>, repeat these steps:
         1. Set <var>startOffset</var> to [=tree/index=] of <var>startNode</var>'s [=tree/root=]'s [=host=].
         1. Set <var>startNode</var> to <var>startNode</var>'s [=tree/root=]'s [=host=]'s [=tree/parent=].</li>
-    1. Set <var>caretRange</var>'s [=range/start node=] and [=range/end node=] to <var>startNode</var>, and set <a>caret range</a>'s [=range/start offset=] and [=range/end offset=] to <var>startOffset</var>.
-    1. Return a <a>caret position</a> whose <a>caret node</a> is set to <var>startNode</var>, <a>caret offset</a> is set to <var>startOffset</var>, and <a>caret range</a> is set to <var>caretRange</var>.
+    1. Return a <a>caret position</a> with its properties set as follows:
+        1. <a>caret node</a> is set to <var>startNode</var>.
+        1. <a>caret offset</a> is set to <var>startOffset</var>.
+        1. <a>caret range</a> is set to a collapsed {{Range}} object which [=range/start node=] and [=range/end node=] are <var>startNode</var>, and which [=range/start offset=] and [=range/end offset=] are <var>startOffset</var>.
 
 Note: The specifics of hit testing are out of scope of this
 specification and therefore the exact details of
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1070,16 +1070,16 @@ result of running these steps:
         <dt><a>caret range</a>
         <dd>null
     </dl>
-1. Otherwise, return a <a>caret position</a> where the
-    <a>caret range</a> is a collapsed
-    {{Range}} object for the position
-    where the text insertion point indicator would have been inserted when applying the <a>transforms</a> that apply to the descendants of the
-    <a>viewport</a>, and the other properties are set after running these steps:
+1. Otherwise, return a <a>caret position</a> where its properties are set after running these steps:
+
+    1. Let <a>caret range</a> to be a collapsed {{Range}} object for the position where the text insertion point indicator whould have been inserted when applying
+        the <a>transforms</a> that apply to the descendants of the <a>viewport</a>.
     1. Let <var>startNode</var> be the [=range/start node=] of the <a>caret range</a>, and let <var>startOffset</var> be the [=range/start offset=] of the <a>caret range</a>.
     1. While <var>startNode</var> is a [=node=], <var>startNode</var>'s [=tree/root=] is a [=shadow root=], and <var>startNode</var>'s [=tree/root=] is not a [=shadow-including inclusive ancestor=] of any of <var>shadowRoots</var>, repeat these steps:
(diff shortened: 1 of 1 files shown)


cssom-view-1/Overview.bs
```diff
@@ -1074,17 +1074,12 @@ result of running these steps:
     <a>caret range</a> is a collapsed
```
dandclark, line 1074: Do we want to bubble the `caret range` out of the shadow roots as well? Otherwise when [getClientRect()](https://drafts.csswg.org/cssom-view/#dom-caretposition-getclientrect) is called on the `CaretPosition`, the `DOMRect`s will correspond to the position inside the shadow root but `caret node` and `caret offset` will be at the adjusted position outside the roots.

siliu1: Yes. we should bubble the `caret range` out of the shadow roots. Updated the spec in next iteration. Thank you!

cssom-view-1/Overview.bs
```diff
@@ -1070,21 +1070,16 @@ result of running these steps:
...
-    <dl>
-        <dt><a>caret node</a>
-        <dd>The [=range/start node=]
-            of the <a>caret range</a>.
-
-        <dt><a>caret offset</a>
-        <dd>The [=range/start offset=] of
-            the <a>caret range</a>.
-    </dl>
+1. Otherwise, return a <a>caret position</a> where its properties are set after running these steps:
```
dandclark, line 1073:
```suggestion
1. Otherwise, return a <a>caret position</a> whose properties are set after running these steps:
```

cssom-view-1/Overview.bs
```diff
@@ -1070,21 +1070,16 @@ result of running these steps:
...
+
+    1. Let <a>caret range</a> to be a collapsed {{Range}} object for the position where the text insertion point indicator whould have been inserted when applying
```
dandclark, line 1075:
```suggestion
    1. Let <a>caret range</a> be a collapsed {{Range}} object for the position where the text insertion point indicator would have been inserted when applying
```

cssom-view-1/Overview.bs
```diff
@@ -1070,21 +1070,16 @@ result of running these steps:
...
+        the <a>transforms</a> that apply to the descendants of the <a>viewport</a>.
+    1. Let <var>startNode</var> be the [=range/start node=] of the <a>caret range</a>, and let <var>startOffset</var> be the [=range/start offset=] of the <a>caret range</a>.
+    1. While <var>startNode</var> is a [=node=], <var>startNode</var>'s [=tree/root=] is a [=shadow root=], and <var>startNode</var>'s [=tree/root=] is not a [=shadow-including inclusive ancestor=] of any of <var>shadowRoots</var>, repeat these steps:
+        1. Set <var>startOffset</var> to [=tree/index=] of <var>startNode</var>'s [=tree/root=]'s [=host=].
+        1. Set <var>startNode</var> to <var>startNode</var>'s [=tree/root=]'s [=host=]'s [=tree/parent=].</li>
+    1. Set <a>caret node</a> in <a>caret position</a> to <var>startNode</var>, and set <a>caret offset</a> in <a>caret position</a> to <var>startOffset</var>.
+    1. Set <a>caret range</a>'s [=range/start node=] and [=range/end node=] to <var>startNode</var>, and set <a>caret range</a>'s [=range/start offset=] and [=range/end offset=] to <var>startOffset</var>.
```
dandclark, line 1082: nit: it might be more straightforward to have a `<var>caretRange</var>` that's declared similarly to `<var>startNode</var>` and `<var>startOffset</var>`, and have that be the thing that is manipulated in these steps. Then you could have a single last step that would set all 3 things on the `CaretPosition`:
`1. Set the <a>caret position</a>'s <a>caret node</a> to <var>startNode</var>, <a>caret offset</a> to <var>startOffset</var>, and <a>caret range</a> to <var>caretRange</var>.`

dandclark: Going further with that approach, you could eliminate the earlier bit that says "...return a <a>caret position</a> where its properties are set after running these steps:", and instead have the last step be:

`1. Return a <a>caret position</a> whose <a>caret node</a> is set to <var>startNode</var>, <a>caret offset</a> is set to <var>startOffset</var>, and <a>caret range</a> is set to <var>caretRange</var>.`

dandclark (approved):
I suggested some refactoring/wording nits but functionally this LGTM.

cssom-view-1/Overview.bs
siliu1: Updated in next iteration. Now the single last step sets all three things on `CaretPosition`.

cssom-view-1/Overview.bs
```diff
@@ -1070,21 +1070,16 @@ result of running these steps:
...
-        <dd>The [=range/start node=]
-            of the <a>caret range</a>.
-
-        <dt><a>caret offset</a>
-        <dd>The [=range/start offset=] of
-            the <a>caret range</a>.
-    </dl>
+1. Otherwise:
+
+    1. Let <var>caretRange</var> to be a collapsed {{Range}} object for the position where the text insertion point indicator would have been inserted when applying
```
sanketj, line 1075: Do browsers really need to use a live range object to implement this algorithm? At least the non-shadow DOM case seems doable with live ranges in Chromium. https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/document.cc;l=1787;drc=9e110c64fee9c391021dc81944912f906e6ba5d5;bpv=1;bpt=1

sanketj: Are you trying to refer to the concept of range instead? https://dom.spec.whatwg.org/#concept-range

cssom-view-1/Overview.bs
```diff
@@ -1070,21 +1070,16 @@ result of running these steps:
...
+        the <a>transforms</a> that apply to the descendants of the <a>viewport</a>.
+    1. Let <var>startNode</var> be the [=range/start node=] of the <var>caretRange</var>, and let <var>startOffset</var> be the [=range/start offset=] of the <var>caretRange</var>.
+    1. While <var>startNode</var> is a [=node=], <var>startNode</var>'s [=tree/root=] is a [=shadow root=], and <var>startNode</var>'s [=tree/root=] is not a [=shadow-including inclusive ancestor=] of any of <var>shadowRoots</var>, repeat these steps:
+        1. Set <var>startOffset</var> to [=tree/index=] of <var>startNode</var>'s [=tree/root=]'s [=host=].
+        1. Set <var>startNode</var> to <var>startNode</var>'s [=tree/root=]'s [=host=]'s [=tree/parent=].</li>
+    1. Set <var>caretRange</var>'s [=range/start node=] and [=range/end node=] to <var>startNode</var>, and set <a>caret range</a>'s [=range/start offset=] and [=range/end offset=] to <var>startOffset</var>.
+    1. Return a <a>caret position</a> whose <a>caret node</a> is set to <var>startNode</var>, <a>caret offset</a> is set to <var>startOffset</var>, and <a>caret range</a> is set to <var>caretRange</var>.
```
sanketj, line 1082: Could we create a "collapsed" `CaretPosition` upfront, instead of a range, and adjust the `offsetNode` and `offset` as needed?

cssom-view-1/Overview.bs
sanketj, line 1075: The [caret range](https://drafts.csswg.org/cssom-view/#caret-range) concept doesn't seem particularly well defined in the spec. What's the idea and is it really necessary?

cssom-view-1/Overview.bs
```diff
@@ -1011,7 +1011,7 @@ Note: Some non-conforming implementations are known to return 32 instead of 24.
 partial interface Document {
   Element? elementFromPoint(double x, double y);
   sequence&lt;Element> elementsFromPoint(double x, double y);
-  CaretPosition? caretPositionFromPoint(double x, double y);
+  CaretPosition? caretPositionFromPoint(double x, double y, ShadowRoot... shadowRoots);
```
sanketj, line 1014: Are we also working on adding WPT coverage for the shadow tree cases?

sanketj:
nit: Could you add links to all the implementation bugs in the PR description?

sanketj (changes requested):

cssom-view-1/Overview.bs
siliu1: Yes. I am working on adding WPT coverage.

cssom-view-1/Overview.bs
siliu1: The [caret range](https://drafts.csswg.org/cssom-view/#caret-range) is used in [`CaretPosition::getClientRect`](https://www.w3.org/TR/cssom-view/#dom-caretposition-getclientrect.). Implementation details can be to create the live range on demand when the `getClientRect` method is invoked.

siliu1: Yes. live range is not needed in the algorithm. Will update in next iteration.

cssom-view-1/Overview.bs
siliu1: Updated to use general range concept instead of a live range in latest iteration.

sanketj:
> nit: Could you add links to all the implementation bugs in the PR description?

Could you also add the link for the WebKit implementation bug? If one doesn't exist, we should file one.

cssom-view-1/Overview.bs
sanketj: If you a PR open, could you add a link to it in this PR's description? If not, you can mention there that WPT coverage will be added subsequently.

cssom-view-1/Overview.bs
sanketj: Isn't just using `CaretPosition` simpler? A range has two endpoints, which both need to be described like above, even though only one is really needed. Using a position concept would make this simpler.

sanketj (changes requested):

siliu1:
> > nit: Could you add links to all the implementation bugs in the PR description?
> 
> Could you also add the link for the WebKit implementation bug? If one doesn't exist, we should file one.

Added WebKit implementation bug for the API.

cssom-view-1/Overview.bs
siliu1: I think they are about the same in terms of expression simplicity. We still need to create a `Range` object for `caret range` in the last step.

I've updated the spec in latest iteration to let a position instead of a range to indicate text insertion point in first step.

cssom-view-1/Overview.bs
```diff
@@ -1070,21 +1070,18 @@ result of running these steps:
...
+        the <a>transforms</a> that apply to the descendants of the <a>viewport</a>.
+    1. Let <var>startNode</var> be the <var>caretPositionNode</var> of the <var>caretPosition</var>, and let <var>startOffset</var> be the <var>caretPositionOffset</var> of the <var>caretPosition</var>.
+    1. While <var>startNode</var> is a [=node=], <var>startNode</var>'s [=tree/root=] is a [=shadow root=], and <var>startNode</var>'s [=tree/root=] is not a [=shadow-including inclusive ancestor=] of any of <var>shadowRoots</var>, repeat these steps:
+        1. Set <var>startOffset</var> to [=tree/index=] of <var>startNode</var>'s [=tree/root=]'s [=host=].
+        1. Set <var>startNode</var> to <var>startNode</var>'s [=tree/root=]'s [=host=]'s [=tree/parent=].</li>
+    1. Return a <a>caret position</a> with its properties set as follows:
+        1. <a>caret node</a> is set to <var>startNode</var>.
+        1. <a>caret offset</a> is set to <var>startOffset</var>.
+        1. <a>caret range</a> is set to a collapsed {{Range}} object which [=range/start node=] and [=range/end node=] are <var>startNode</var>, and which [=range/start offset=] and [=range/end offset=] are <var>startOffset</var>.
```
sanketj, line 1084: Should this be removed now?

sanketj (changes requested):

cssom-view-1/Overview.bs
siliu1: Do you mean to remove the `caret range` from the algorithm? I think it's better to keep it in the spec even if we don't need to create the range right away. We should create the live range on demand, similar to https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/cssom/caret_position.cc;l=37

sanketj: Yes, okay to take care of this as a follow up issue, but can we add an issue into the current spec for this, so that implementors know that this may change.

siliu1: Filed https://github.com/w3c/csswg-drafts/issues/10230.

sanketj: Can we also add a section in the spec here that links to the issue? Like what we have here: https://www.w3.org/TR/css-highlight-api-1/#events

siliu1: Done.

sanketj (approved):

zcorpan:
As far as I can tell this works like in https://w3c.github.io/selection-api/#dom-selection-getcomposedranges as @mfreed7 suggested in https://github.com/w3c/csswg-drafts/issues/9932#issuecomment-2035004366 and per the [CSSWG resolution](https://github.com/w3c/csswg-drafts/issues/9932#issuecomment-2035793983).

cc @emilio

zcorpan (approved):

cssom-view-1/Overview.bs
```diff
@@ -1148,6 +1145,8 @@ aborting on the first step that returns a value:
 
 Note: This {{DOMRect}} object is not <a spec=html>live</a>.
 
+Issue(10230): Consider removing <a>caret range</a> concept from <dfn>caret position</dfn> interface.
```
emilio, line 1148: This introduces a build error fwiw. caret position is defined twice.

emilio (commented):
Looks generally fine.

cssom-view-1/Overview.bs
emilio: Fix for that in https://github.com/w3c/csswg-drafts/pull/10285

PR #10227, opened 2024-04-18
[css-scroll-snap-2] Specify scroll-start-target

DavMila:
This adds details to the scroll-snap-2 spec[1] to specify scroll-start-target.

[1] https://drafts.csswg.org/css-scroll-snap-2


css-scroll-snap-2/Overview.bs
```diff
@@ -193,26 +193,117 @@ The 'scroll-start' property {#scroll-start}
...
+				to the result of running the steps to <a>get initial scroll offset</a> for
+				|scrollcontainer| with the axis parameter as “block” and the target parameter
+				as |blocktarget|.
+		1. If |inlinetarget| is not null:
+			1. Set the <a>inline axis</a> component of |scrollcontainer|’s <a>initial scroll position</a>
+				to the result of running the steps to <a>get initial scroll offset</a> for
+				|scrollcontainer| with the axis parameter as “inline” and the target parameter
+				as |inlinetarget|.
+
+		The steps to <dfn>get initial scroll offset</dfn> for a <a>scroll container</a> |scrollcontianer|,
```
flackr, line 227: I think this should be calling into a common algorithm used by the spec rules for [scrolling an element into view](https://drafts.csswg.org/cssom-view/#scroll-a-target-into-view) as we would expect this to behave the same position as you would get by calling scrollIntoView.

css-scroll-snap-2/Overview.bs
```diff
@@ -193,26 +193,117 @@ The 'scroll-start' property {#scroll-start}
...
+		with respect to a particular axis is an element or pseudo-element whose {{scroll-start-target}}
+		property for that axis is non-''none'' and whose nearest <a>scroll container</a> is
+		|scrollcontainer|. Where multiple such elements or pseudo-elements exist, user-agents
+		should select the one which comes first in <a href="https://www.w3.org/TR/dom/#concept-tree-order">tree order</a>. Where no such element
+		or pseudo-element exists, |scrollcontainer|’s <a>initial scroll target</a> in
+		that axis is null.
+
+		Except for [[#cases-where-scroll-start-target-is-inactive|cases where scroll-start-target is inactive]],
+		user-agents should use the <a>initial scroll target</a> for a <a>scroll container</a> |scrollcontainer|
+		to determine the <a>initial scroll position</a> of |scrollcontainer| along each
```
flackr, line 208: I think you'll need to amend css-overflow-3 to explain how this interacts with the other ways initial scroll position can be determined: https://drafts.csswg.org/css-overflow-3/#scrolling

I suspect that a specified target takes precedence over the `align-content` and `justify-content` changes to initial scroll position.

css-scroll-snap-2/Overview.bs
```diff
@@ -193,26 +193,117 @@ The 'scroll-start' property {#scroll-start}
...
+		</pre>
+
+		Values are for the {{scroll-start-target}} shorthand have the same definitions
+		as values for the [[#scroll-start-target-longhands|longhands]].
+
+	<h4 id="scroll-start-target-with-place-content">Interaction with <a>place-content</a></h4>
+
+	If the <a>initial scroll position</a> indicated by <a spec="css-align" lt="place-content">place-content</a> differs
+	from that indicated by {{scroll-start-target}}, user-agents should give preference
+	to the latter.
```
flackr, line 298: Shouldn't this take precedence over place-content? It seems like `scroll-start-target` is a more precise location whereas place-content can only determine whether scrolling should start at the beginning or end of a scroller.

css-scroll-snap-2/Overview.bs
```diff
@@ -193,26 +193,117 @@ The 'scroll-start' property {#scroll-start}
...
-  </dl>
+	<h4 id="initial-scroll-target">Initial scroll target</h4>
+
+		The <dfn>initial scroll target</dfn> of a <a>scroll container</a> |scrollcontainer|
+		with respect to a particular axis is an element or pseudo-element whose {{scroll-start-target}}
+		property for that axis is non-''none'' and whose nearest <a>scroll container</a> is
+		|scrollcontainer|. Where multiple such elements or pseudo-elements exist, user-agents
+		should select the one which comes first in <a href="https://www.w3.org/TR/dom/#concept-tree-order">tree order</a>. Where no such element
+		or pseudo-element exists, |scrollcontainer|’s <a>initial scroll target</a> in
+		that axis is null.
```
flackr, line 204: I think you'll also need to define the interaction with the [try to scroll to the fragment](https://html.spec.whatwg.org/#try-to-scroll-to-the-fragment) step of navigation. In particular, this is the part that deals with cases where the element to scroll to is not in the initial render. I think we'll want this to also detect cases where a scroll-start-target was not originally present but shows up in the document.

css-scroll-snap-2/Overview.bs
```diff
@@ -193,26 +193,117 @@ The 'scroll-start' property {#scroll-start}
...
-    <dd>Element is not a ''scroll-start-target''.
-    <dt><dfn>auto</dfn>
-    <dd>Element is used to calculate the ''scroll-start'' position, 
-    taking into account ''scroll-padding'' or ''scroll-margin'' ,
-    same as a ''scroll-snap'' target.
-  </dl>
+	<h4 id="initial-scroll-target">Initial scroll target</h4>
+
+		The <dfn>initial scroll target</dfn> of a <a>scroll container</a> |scrollcontainer|
+		with respect to a particular axis is an element or pseudo-element whose {{scroll-start-target}}
```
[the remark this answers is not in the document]
argyleink: i've been asking the community and hunting for examples, here's the current thoughts:
1. many examples of wanting to set the block/inline scroll start targets can be achieved by specifying a single intersecting element. but, knowing which element that is this at this intersection might not be known.
2. consider the example where a scroll-start-target is desired on x but not y. so even tho it's a scroller in both axes, it'd be desirable to only scroll-start one axis at a provided element. like a table and wanting to scroll-start at a row and not having it move the x scroll position.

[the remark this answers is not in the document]
DavMila: Thanks for looking around on this Adam. I was wondering, if an author wants to scroll to whatever element is at a particular position, then I guess, in a sense, it's not so much the element they want to scroll to as it is the position. In this case, perhaps an acceptable pattern would be to put a placeholder element at that location and have it be the scroll-start-target. I guess there are multiple ways of constructing a grid and I'm not yet sure if this would be a feasible/acceptable approach to all(any?) of them.

css-scroll-snap-2/Overview.bs
DavMila: I think that's what I'm saying? the former would be the "initial scroll position indicated by place-content" and the latter would be the initial scroll position "indicated by scroll-start-target" and user agents should give preference to the latter.

css-scroll-snap-2/Overview.bs
DavMila: I think this would still be covered by what you pointed out earlier: scroll-start-target only determines what the initial scroll position should be. So it's implied that a scroll-start-target that shows up later only changes the initial scroll position of some affected scroller but not (necessarily) its current scroll position, similar to place-content affecting the initial scroll position. Or perhaps you mean that all this is somewhat too implicit and needs to be called out directly?

css-scroll-snap-2/Overview.bs
```diff
@@ -193,26 +193,117 @@ The 'scroll-start' property {#scroll-start}
...
+		property for that axis is non-''none'' and whose nearest <a>scroll container</a> is
+		|scrollcontainer|. Where multiple such elements or pseudo-elements exist, user-agents
+		should select the one which comes first in <a href="https://www.w3.org/TR/dom/#concept-tree-order">tree order</a>. Where no such element
+		or pseudo-element exists, |scrollcontainer|’s <a>initial scroll target</a> in
+		that axis is null.
+
+		Except for [[#cases-where-scroll-start-target-is-inactive|cases where scroll-start-target is inactive]],
```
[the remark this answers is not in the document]
DavMila: Thanks, I've removed this.

css-scroll-snap-2/Overview.bs
DavMila: I think we intend for scroll-start-target to only affect its nearest scroll ancestor. Invoking "[scrolling an element into view](https://drafts.csswg.org/cssom-view/#scroll-a-target-into-view)" as written would result in the scrolling of all ancestor scrolling boxes. I've modified it to define an optional _nearest_only_ parameter for this algorithm which will default to false.

css-scroll-snap-2/Overview.bs
DavMila: Yes, I've made a change in css-overflow-3.

DavMila (commented):
Made suggested changes, ptal, thanks.

css-scroll-snap-2/Overview.bs
```diff
@@ -196,90 +196,31 @@ The 'scroll-start-target' property {#scroll-start-target}
...
-			Inherited: no
-			Logical property group: scroll-start-target
-			Percentages: n/a
-			Computed Value: either of the keywords "none" or "auto"
-			Animation type: not animatable
-		</pre>
+		1. Let |target| be the <a>initial scroll target</a> for |scrollcontainer|.
+		1. If |target| is null, return.
+		1. Run the steps to <a spec="cssom-view-1" lt="scroll a target into view">scroll |target| into view</a>
+			with |behavior| set to "auto", |block| set to "start", |inline| set to "nearest" and
```
flackr, line 215: So I think we want to share like 95% of the algorithm, except for the last step where it says:

> 6. [Perform a scroll](https://drafts.csswg.org/cssom-view/#perform-a-scroll) of box to position, element as the associated element and behavior as the scroll behavior.

As we don't want this to be an actual scroll, but rather just the initial position of the scroller. Note it can actually perform a scroll if the scroll target turns up late in the loading process where that's unavoidable.

As such, we should pull out all of the steps which determine the target position into an algorithm we can invoke here to calculate the initial scroll position. I think this might also resolve the issue of needing to only scroll the nearest scroller without needing to add the flag.

css-scroll-snap-2/Overview.bs
```diff
@@ -288,21 +229,34 @@ The 'scroll-start-target' property {#scroll-start-target}
...
-	document’s “:target”, {{scroll-start-target}} is inactive for that <a>scroll container</a>.
+	to the position indicated by {{scroll-start-target}}.
+
+	<h4 id="scroll-start-fragment-navigation">Interaction with Fragment Navigation</h4>
+
+	<a href="https://html.spec.whatwg.org/#try-to-scroll-to-the-fragment">
+	Scrolling to the fragment for a Document</a> might set the <a>scroll position</a> of a
+	<a>scroll container</a>. This might <a spec="cssom-view-1" lt="scroll an element">scroll</a>
+	the scroll container away from its <a>initial scroll position</a>.
+	It might occur that an element which is a {{scroll-start-target}}	for that <a>scroll container</a>
```
flackr, line 252: Replace tab character after `{{scroll-start-target}}` with a space.

css-scroll-snap-2/Overview.bs
```diff
@@ -288,21 +229,34 @@ The 'scroll-start-target' property {#scroll-start-target}
...
+	is loaded into the document after such a
+	<a href="https://html.spec.whatwg.org/#try-to-scroll-to-the-fragment">
+	scroll to the fragment</a> has occurred. In such an instance, the
+	{{scroll-start-target}} does set the <a>initial scroll position</a> of the
+	<a>scroll container</a>	but does not affect the current <a>scroll position</a> of the
+	<a>scroll container</a> i.e. does not cause it to be scrolled in response to
+	the late arrival of the {{scroll-start-target}}.
```
flackr, line 259: Sorry I think this is a little vague and should be rewritten to say when this will scroll to the scroll-start-target. I.e. if a scroll-start-target is discovered and we have not scrolled to a fragment. We don't need to  mention the initial scroll position part, since scrolling to a fragment doesn't change the initial position but just scrolls away from the initial position, but we do want scroll-start-target to scroll to late discovered targets as long as we haven't done a fragment scroll.

flackr (changes requested):

css-scroll-snap-2/Overview.bs
```diff
@@ -196,90 +196,31 @@ The 'scroll-start-target' property {#scroll-start-target}
...
-		or pseudo-element exists, |scrollcontainer|’s <a>initial scroll target</a> in
-		that axis is null.
+		or pseudo-element exists, |scrollcontainer|’s <a>initial scroll target</a> is null.
 
-		Except for [[#cases-where-scroll-start-target-is-inactive|cases where scroll-start-target is inactive]],
-		user-agents should use the <a>initial scroll target</a> for a <a>scroll container</a> |scrollcontainer|
-		to determine the <a>initial scroll position</a> of |scrollcontainer| along each
-		axis by running the steps to <a>apply initial scroll targets</a>
+		User-agents should use the <a>initial scroll target</a> for a <a>scroll container</a>
+		to determine its <a>initial scroll position</a> by running the steps to
```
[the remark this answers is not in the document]
DavMila: Sounds good. The section "scroll-start-target-with-place-content" covers that it overrides place-content which is the shorthand for align-content and justify-content.

css-overflow-3/Overview.bs
```diff
@@ -340,9 +340,16 @@ Scrolling Overflow</h3>
 	is typically dependent on the [=scroll container=]’s [=writing mode=],
 	and, unless otherwise specified,
 	coincides with its [=scroll origin position=].
-	However, the 'align-content' and 'justify-content' properties [[!CSS-ALIGN-3]]
-	can be used to change the [=initial scroll position=],
-	see [[css-align-3#overflow-scroll-position]].
+	However, some CSS properties can be used to change the initial position. In decreasing order of
+	precedence, these are:
```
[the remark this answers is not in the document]
DavMila: Sounds good, done.

css-scroll-snap-2/Overview.bs
DavMila: Sounds good, I've taken a stab at extracting the portion of the algorithm that determines the position into a set of steps to "determine the scroll-view position". Lmk wyt.

css-scroll-snap-2/Overview.bs
DavMila: Done.

css-scroll-snap-2/Overview.bs
DavMila: Okay, I've tried to make it a bit clearer by pointing to the html spec on [updating the document](https://html.spec.whatwg.org/#updating-the-document) to cover arrivals of scroll-start-targets after the first layout of a scroll container.

cssom-view-1/Overview.bs
```diff
@@ -1443,14 +1443,14 @@ The <dfn attribute for=Element>currentCSSZoom</dfn> attribute must return the
 
 <h3 id=element-scrolling-members>{{Element}} Scrolling Members</h3>
 
-To <dfn export>scroll a target into view</dfn> <var>target</var>, which is an <a for="/">Element</a> or <a>Range</a>,
+
+To <dfn export>determine the scroll-view position</dfn> of a <var>target</var>, which is an <a for="/">Element</a> or <a>Range</a>,
```
flackr, line 1447: nit: I think you meant scroll-into-view position?

css-overflow-3/Overview.bs
```diff
@@ -340,8 +340,7 @@ Scrolling Overflow</h3>
 	is typically dependent on the [=scroll container=]’s [=writing mode=],
 	and, unless otherwise specified,
 	coincides with its [=scroll origin position=].
-	However, the 'align-content' and 'justify-content' properties [[!CSS-ALIGN-3]]
-	can be used to change the [=initial scroll position=],
+	However, certain CSS properties can be used to change the [=initial scroll position=],
```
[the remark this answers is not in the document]
DavMila: Good idea. Done

cssom-view-1/Overview.bs
DavMila: Yes, thanks. Done.

PR #10240, opened 2024-04-22
[css-syntax-3] Do not consume a qualified rule block twice

cdoublev:
[*Consume a qualified rule*](https://drafts.csswg.org/css-syntax-3/#consume-qualified-rule) consumes two blocks in a row:

  > Otherwise, consume a block from `input`, and assign the results to `rule`’s lists of declarations and child rules. 
  > 
  > Consume a block from `input`, and let `child rules` be the result.

I assume it is a typo from https://github.com/w3c/csswg-drafts/commit/e5547b96f5de6fb5a68d050f42d562c448b99d0b.

### DIFF
--- a/css-syntax-3/Overview.bs
+++ b/css-syntax-3/Overview.bs
@@ -2753,11 +2753,6 @@ Consume a qualified rule</h4>
 			</details>
 
 			Otherwise, [=consume a block=] from |input|,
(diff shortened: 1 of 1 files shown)


tabatkins:
Yup, thanks, I'd added the new text below the existing text, and then forgot to remove it.

Issue #9745, opened 2023-12-21
[css-pseudo-4] Enabling carousel design patterns in CSS


flackr:
Carousels are an often used design pattern on the web. They are used in a variety of contexts, from product listing pages to slideshow like content. OpenUI has [explored a range of carousel designs](https://open-ui.org/components/carousel.research/), showing that the specific layout and appearance can vary dramatically. They are also provided by many frameworks as components, however implementing a carousel correctly is complicated and often results in inconsistent and sometimes inaccessible implementations.

There are a [variety of problems being solved by carousels](https://css.oddbird.net/overflow/explainer/), which we believe could be provided by a set of CSS features. Developers could then combine these CSS features to create the various designs. CSS-only component libraries could be built to further simplify this process.

I'd like to share a strawman set of feature proposals that enable the creation of a variety of carousel patterns out of plain HTML lists using CSS. I've put up an [explainer](https://github.com/flackr/carousel/) with polyfills and demos (e.g. see the [full carousel](https://flackr.github.io/carousel/examples/carousel/image/)) where you can see the individual features and how they can be combined to make a full fledged CSS carousel. I'd like to introduce the problem, features and get feedback on the overall direction and where to take the discussion.

Explainer: https://github.com/flackr/carousel/

I'm adding this under css-pseudo-4 for now, but likely the features here would span css-grid-2, css-scroll-snap-2, and possibly other specs.

yisibl:
Does the existing design consider such a drag-and-drop switching mode?

https://github.com/w3c/csswg-drafts/assets/2784308/b7af07df-77ea-4181-82ff-469c367bb5f0

tabatkins:
Another chunk of conversation logs from the end of the session:

<details>
```

15:26 <@TabAtkins> wendyreid: I noticed in a lot of examples - and i've also struggled with this for devs - I haven't seen a lot where each item in the carousel contains a bunch of other content
15:26 <@TabAtkins> wendyreid: Like say each has an image, text, and a call to action button
15:26 <@TabAtkins> wendyreid: but you also want item-by-item interaction as well as interaction in each
15:26 <CurtBellew> present+
15:26 <@TabAtkins> wendyreid: so in this example (image with label) what if there's a button inside of each, too
15:27 <bkardell_> q+
15:27 — Zakim sees wendyreid, bkardell_ on the speaker queue
15:27 <@TabAtkins> flackr: this is just content inside a scrollable element, we're not treading new ground
[...]

PR #10243, opened 2024-04-23
[css-overflow-5] Define scroll-markers

flackr:
Defines scroll-markers properties, pseudo-elements, scrolltarget attribute and associated behaviors for #9745.


flackr:
I'm sure there's a lot of important details missing here, but it might be useful to have something written down to point to and work out the details over PRs and issues. WDYT?

tabatkins:
(ping @dbaron @fantasai @frivoal )

css-overflow-4/Overview.bs
```diff
@@ -2118,13 +2178,75 @@ showing in which case space is reserved for the <a>scrollbar gutter</a>.
...
+		1. Set <var>selected</var> to <var>target</var>.
+		1. Let <var>targetPosition</var> be the position that would be scrolled to if we scroll <var>target</var> into view.
+		1. If <var>targetPosition</var>'s scroll block offset is less than or equal to <var>position</var>'s scroll block offset, break.
+		1. If <var>targetPosition</var>'s scroll inline offset is less than or equal to <var>position</var>'s scroll inline offset, break.
+	1. Let <var>marker</var> be the first control in <var>markers</var> whose ''HTMLButtonElement.scrollTargetElement'' is <var>selected</var>.
+	1. Set the 'checked' state of <var>marker</var> to true.
+
+<h4 id="scroll-target-focus">Focus behavior</h4>
+
+	A ''scroll marker control'' is only focusable if it is 'checked'.
```
tabatkins, line 2236: It should be styleable with `:checked` as well, yeah?

tabatkins, line 2236: Is there any way to detect which button or pseudo-element is checked besides running a `querySelector()`? (I don't know if there needs to be, just curious.)

tabatkins (changes requested):

css-overflow-4/Overview.bs
flackr: Yes, I've added a reference to `'':checked''` (https://drafts.csswg.org/selectors-3/#checked) from the scrolltarget section where it explains the checked state.

flackr: We could always add this. It may make sense to find the currently active marker from the scroll container since that is what forms the exclusive group.

flackr (commented):
Thanks for the review @tabatkins!

css-overflow-4/Overview.bs
```diff
@@ -285,6 +286,65 @@ Expanding Clipping Bounds: the 'overflow-clip-margin-*' properties</h3>
...
+		<dt><dfn>after</dfn>
+		<dd>
+			The [=scroll container=] generates a ''::scroll-markers'' pseudo-element
+			whose box is an immediate following sibling to its [=originating element=].
+
+	</dl>
+
+<h4 id="fragment-pseudo-element">The ''::scroll-markers'' pseudo-element</h4>
+
+	The <dfn selector>::scroll-markers</dfn> pseudo-element box
```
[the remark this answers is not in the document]
tabatkins: Probably, yeah, just for clarity.

css-overflow-4/Overview.bs
tabatkins: Yeah, if we provide the API at all I think it would be on the scroll container.

css-overflow-4/Overview.bs
```diff
@@ -285,6 +286,65 @@ Expanding Clipping Bounds: the 'overflow-clip-margin-*' properties</h3>
...
+	to ensure that the generation of scroll marker pseudo-elements does not invalidate the layout of the site:
+
+	<pre class="lang-css">
+		/* The generation of ::scroll-marker pseudo-elements cannot invalidate layout outside of this pseudo-element. */
+		::scroll-markers { contain: size !important; }
+	</pre>
+
+<h4 id="fragment-pseudo-element">The ''::scroll-marker'' pseudo-element</h4>
+
+	When the computed 'content' value of an element is not 'none',
```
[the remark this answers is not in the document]
tabatkins: I think you missed the nature of my question. I can't tell, from this spec text, what causes a ::scroll-marker to be created and linked with a certain element, or not created.

css-overflow-4/Overview.bs
```diff
@@ -2118,13 +2178,80 @@ showing in which case space is reserved for the <a>scrollbar gutter</a>.
...
+
+	The <dfn element-attr for=button>scrolltarget</dfn> attribute turns a <{button}> element into a [=scroll marker control=].
+	This takes the ID of the element to target as its value.
+
+	The <dfn attribute for=HTMLButtonElement lt=scrollTargetElement>HTMLButtonElement.scrollTargetElement</dfn> instance property
+	gets and sets the element being interacted with by the control button.
+	This is the JavaScript equivalent of the <{button/scrolltarget}> HTML attribute.
+
+	A 'button' with a non-null {{HTMLButtonElement/scrollTargetElement}}
+	represents a <dfn export>scroll marker control</dfn> that forms a <dfn>scroll marker group<dfn> for its nearest [=scroll container=]
```
tabatkins, line 2193:
```suggestion
	represents a <dfn export>scroll marker control</dfn> that forms a <dfn>scroll marker group</dfn> for its nearest [=scroll container=]
```

css-overflow-4/Overview.bs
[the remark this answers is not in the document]
flackr: Apologies, I was following the style of https://www.w3.org/TR/css-pseudo-4/#generated-content but accidentally referred to the content of the element, not the peudo-element. I meant to refer to the pseudo-element's computed content value.

css-overflow-4/Overview.bs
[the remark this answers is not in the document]
flackr: Done.

fantasai:
> I'm sure there's a lot of important details missing here, but it might be useful to have something written down to point to and work out the details over PRs and issues. WDYT?

I agree, but the plan was to put it into Overflow L5 along with the current fragmentation appendix, not to patch Overflow L4. :) https://github.com/w3c/csswg-drafts/issues/9745#issuecomment-1944824869

Wrt the content of this, I think it needs more explanation of the use cases it's trying to solve (and the use case it's not trying to solve). For my part, I don't understand why we aren't trying to solve the "highlight the relevant table of contents entry" item along with this, for example.

The part about only the currently-active scroll marker being focusable is also weird. How do you activate a scroll marker control with the keyboard if it's not focusable?

flackr:
> > I'm sure there's a lot of important details missing here, but it might be useful to have something written down to point to and work out the details over PRs and issues. WDYT?
> 
> I agree, but the plan was to put it into Overflow L5 along with the current fragmentation appendix, not to patch Overflow L4. :) [#9745 (comment)](https://github.com/w3c/csswg-drafts/issues/9745#issuecomment-1944824869)

Sorry about that, this was me misreading the resolution notes. I've created css-overflow-5, moving the fragmentation appendix as we agreed, and moved this over there.

> Wrt the content of this, I think it needs more explanation of the use cases it's trying to solve (and the use case it's not trying to solve). For my part, I don't understand why we aren't trying to solve the "highlight the relevant table of contents entry" item along with this, for example.

We are. The `scroll-marker` for the current item can be styled with `:checked`: https://github.com/w3c/csswg-drafts/pull/10243/files#diff-fdbff78a41c06b32b51fc29f900df82da5efa11aa9e2e3394047d39c889569f9R896

> The part about only the currently-active scroll marker being focusable is also weird. How do you activate a scroll marker control with the keyboard if it's not focusable?

The current algorithm will always produce a selected marker: https://github.com/w3c/csswg-drafts/pull/10243/files#diff-fdbff78a41c06b32b51fc29f900df82da5efa11aa9e2e3394047d39c889569f9R932

I agree if there were not an active marker we'd want to do something similar to a radio group where the first one can get focused.

flackr:
Merging and moving discussion to followup issues per discussion with @fantasai.

PR #10261, opened 2024-04-26
[cssom-view] Fix offsetParent to refer to effective zoom as a stopping point

### LINKED ISSUE #10252: [cssom-view] offsetParent zoom handling seems wrong.
https://drafts.csswg.org/cssom-view/#dom-htmlelement-offsetparent has:

> The element has a non-default used value of [zoom](https://drafts.csswg.org/css-viewport-1/#propdef-zoom).

But afaict that's not what's implemented in engines. What's implemented is "The element has a different used value of "zoom".

Is my read of the WebKit / Blink code correct @chrishtr? If so we should update the spec. We have tests that depend on that difference.

chrishtr:
Fixes #10252.

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1555,7 +1555,7 @@ The <dfn attribute for=HTMLElement>offsetParent</dfn> attribute must return the 
 1. Let <var>ancestor</var> be the parent of the element in the <a>flat tree</a> and repeat these substeps:
     1. If <var>ancestor</var> is <a>closed-shadow-hidden</a> from the element and its computed value of the 'position' property is ''position/fixed'', terminate this algorithm and return null.
     1. If <var>ancestor</var> is not <a>closed-shadow-hidden</a> from the element and satisfies at least one of the following, terminate this algorithm and return <var>ancestor</var>.
-        * The ancestor is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
+        * <var>ancestor</var> is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
         * The element has a different [=effective zoom=] than <var>ancestor</var>
         * It is <a>the <code>body</code> element</a>.
         * The computed value of the 'position' property of the element is ''static'' and the ancestor is one of the following <a>HTML elements</a>: <code>td</code>, <code>th</code>, or <code>table</code>.
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1556,7 +1556,7 @@ The <dfn attribute for=HTMLElement>offsetParent</dfn> attribute must return the 
     1. If <var>ancestor</var> is <a>closed-shadow-hidden</a> from the element and its computed value of the 'position' property is ''position/fixed'', terminate this algorithm and return null.
     1. If <var>ancestor</var> is not <a>closed-shadow-hidden</a> from the element and satisfies at least one of the following, terminate this algorithm and return <var>ancestor</var>.
         * <var>ancestor</var> is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
-        * The element has a different [=effective zoom=] than <var>ancestor</var>
+        * The element has a different [=effective zoom=] than <var>ancestor</var>.
         * It is <a>the <code>body</code> element</a>.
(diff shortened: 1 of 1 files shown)


cssom-view-1/Overview.bs
```diff
@@ -1556,7 +1556,7 @@ The <dfn attribute for=HTMLElement>offsetParent</dfn> attribute must return the
     1. If <var>ancestor</var> is <a>closed-shadow-hidden</a> from the element and its computed value of the 'position' property is ''position/fixed'', terminate this algorithm and return null.
     1. If <var>ancestor</var> is not <a>closed-shadow-hidden</a> from the element and satisfies at least one of the following, terminate this algorithm and return <var>ancestor</var>.
         * The element is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
```
emilio, line 1558: Pre-existing, but this is a bit confusing, because `the element` above means "the element you call offsetParent on", but here means `<var>ancestor</var>`.

Maybe change it to `* <var>ancestor</var> is a containing block ...` to be explicit?

cssom-view-1/Overview.bs
```diff
@@ -1556,7 +1556,7 @@ The <dfn attribute for=HTMLElement>offsetParent</dfn> attribute must return the
...
-        * The element has a non-default used value of 'zoom'.
+        * The element has a different [=effective zoom=] than its parent.
```
emilio, line 1559: This is not right, is it? You want something like: ` * <var>ancestor</var> has a different [=effective zoom=] than the element.` Or ` * The element has a different [=effective zoom=] than <var>ancestor</var>`

emilio (approved):
Looks good with it clarified a bit.

cssom-view-1/Overview.bs
chrishtr, line 1559:
```suggestion
        * The element has a different [=effective zoom=] than <var>ancestor</var>
```

cssom-view-1/Overview.bs
chrishtr, line 1558:
```suggestion
        * The ancestor is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
```

cssom-view-1/Overview.bs
```diff
@@ -1555,8 +1555,8 @@ The <dfn attribute for=HTMLElement>offsetParent</dfn> attribute must return the
 1. Let <var>ancestor</var> be the parent of the element in the <a>flat tree</a> and repeat these substeps:
     1. If <var>ancestor</var> is <a>closed-shadow-hidden</a> from the element and its computed value of the 'position' property is ''position/fixed'', terminate this algorithm and return null.
     1. If <var>ancestor</var> is not <a>closed-shadow-hidden</a> from the element and satisfies at least one of the following, terminate this algorithm and return <var>ancestor</var>.
-        * The element is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
-        * The element has a non-default used value of 'zoom'.
+        * The ancestor is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
```
chrishtr, line 1558:
```suggestion
        * <var>ancestor</var> is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
```

cssom-view-1/Overview.bs
```diff
@@ -1555,8 +1555,8 @@ The <dfn attribute for=HTMLElement>offsetParent</dfn> attribute must return the
 1. Let <var>ancestor</var> be the parent of the element in the <a>flat tree</a> and repeat these substeps:
     1. If <var>ancestor</var> is <a>closed-shadow-hidden</a> from the element and its computed value of the 'position' property is ''position/fixed'', terminate this algorithm and return null.
     1. If <var>ancestor</var> is not <a>closed-shadow-hidden</a> from the element and satisfies at least one of the following, terminate this algorithm and return <var>ancestor</var>.
-        * The element is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
-        * The element has a non-default used value of 'zoom'.
+        * <var>ancestor</var> is a containing block of absolutely-positioned descendants (regardless of whether there are any absolutely-positioned descendants).
+        * The element has a different [=effective zoom=] than <var>ancestor</var>
```
chrishtr, line 1559:
```suggestion
        * The element has a different [=effective zoom=] than <var>ancestor</var>.
```

PR #10270, opened 2024-04-30
[css-spec-viewport-1] updated the values for zoom property

dletorey:
I feel that the `values` for CSS `zoom` property are in need of updating to reflect the actual values being used in browser. 

I could of course be totally wrong here, new to this.

[zoom property on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/zoom)

[zoom property on drafts CSSWG](https://drafts.csswg.org/css-viewport/#zoom-property)

### DIFF
--- a/css-viewport/Overview.bs
+++ b/css-viewport/Overview.bs
@@ -398,7 +398,7 @@ Note: 'zoom' does not affect or prevent 'transform' scaling.
 
 <pre class="propdef">
 	Name: zoom
-	Value: <<number [0,∞]>> || <<percentage [0,∞]>>
+	Value: normal | reset | <<number [0,∞]>> || <<percentage [0,∞]>>
 	Initial: 1
 	Applies to: all <<length>> property values of all elements
 	Inherited: no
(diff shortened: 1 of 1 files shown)


fantasai (changes requested):
It looks like all browsers do support `normal` as equal to `1`, so maybe we should add it to the spec as computing to `1`. [[testcase](https://www.software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0AZOOM%0A%3Cdiv%20style%3D%22zoom%3A%202%22%3E%0A%20%20ZOOM%0A%20%20%3Cdiv%20style%3D%22zoom%3A%202%3B%20zoom%3A%20normal%22%3EZOOM%3C%2Fdiv%3E%0A%3C%2Fdiv%3E)]

`reset` doesn't seem to be supported by anyone though [[testcase](https://www.software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0AZOOM%0A%3Cdiv%20style%3D%22zoom%3A%202%22%3E%0A%20%20ZOOM%0A%20%20%3Cdiv%20style%3D%22zoom%3A%202%3B%20zoom%3A%20reset%22%3EZOOM%3C%2Fdiv%3E%0A%3C%2Fdiv%3E)], and from the MDN data seems to have been removed. So we probably shouldn't add it.

lilles:
Tests already cover the resolution:

https://wpt.fyi/results/css/css-viewport/zoom/parsing/zoom-computed.html
https://wpt.fyi/results/css/css-viewport/zoom/parsing/zoom-valid.html

PR #10295, opened 2024-05-08
[css-cascade-6] Clarify circular dependency (In wording) in finding scoping limits.

dshin-moz:
Finding scoping limits mention in-scope elements, which in turn mentions scoping limits, which in turn etc etc. Just a wording issue, since it can be inferred that the scoping limit portion is ignored at this time.

Still, I think this is helps reduce confusion on first-read.

### DIFF
--- a/css-cascade-6/Overview.bs
+++ b/css-cascade-6/Overview.bs
@@ -539,7 +539,8 @@ Identifying Scoping Roots and Limits</h4>
 	: Finding any [=scoping limits=]
 	:: For each [=scope=] created by a [=scoping root=],
 		its [=scoping limits=] are set to all elements
-		that are [=in scope=] and that match <<scope-end>>,
+		that are [=in scope=] (Only in terms of [=scoping roots=])
+		and that match <<scope-end>>,
 		interpreting '':scope'' and ''&''
 		exactly as in [=scoped style rules=].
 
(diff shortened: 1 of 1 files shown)


css-cascade-6/Overview.bs
```diff
@@ -535,18 +531,16 @@ Identifying Scoping Roots and Limits</h4>
...
+		then the [=scoping root=] is the [=root=] of the containing [=node tree=].)
 		Any '':scope'' or ''&'' selectors in <<scope-start>>
 		are interpreted as defined for its outer context.
 
 	: Finding any [=scoping limits=]
 	:: For each [=scope=] created by a [=scoping root=],
 		its [=scoping limits=] are set to all elements
-		that are [=in scope=] and that match <<scope-end>>,
+		that are [=in scope=] (Only in terms of [=scoping roots=])
+		and that match <<scope-end>>,
```
tabatkins:
```suggestion
		that are [=descendants=] of the [=scoping root=]
		and that match <<scope-end>>,
```

tabatkins:
I inlined the desired half of the definition instead. 

Looking over it with fresh eyes, it seemed that "in scope" was probably also the wrong definition *anyway*, since it includes the scoping root, and I don't think we want scoping limits to be able to match the scoping root itself? That would be easy to hit if you're trying to style a nestable component with `@scope (.widget) to (.widget)`, and would produce a scope with a single element in it, which doesn't seem particularly useful. In the rare case you actually *do* just want one element like that, just using `@scope (.foo) { :scope {...}}` works just fine, with no need for a limit.

So as part of inlining it, I also changed "inclusive descendant of the scoping root" to just "descendant of the scoping root".

PR #10301, opened 2024-05-09
[css-viewport] Add clarification that path() contents are zoomed

### LINKED ISSUE #10126: [css-viewport] CSS zoom should affect clip-path: path() strings.
Right now per spec it doesn't, but that doesn't match implementations, and it makes sense for it to affect it.

We have tests for this (css/css-masking/clip-path/clip-path-path-with-zoom.html etc), so I think just a tweak to the spec should be enough, wdyt @chrishtr?

chrishtr:
Fixes #10126.

### DIFF
--- a/css-viewport/Overview.bs
+++ b/css-viewport/Overview.bs
@@ -369,7 +369,7 @@ only the value previously set to it.
 An element becomes zoomed when the 'zoom' property has a positive computed value different than 1
 (or when a flat tree ancestor has zoom).
 
-To apply zoom, the [=used value=] of a CSS property (including values inside of [=path=] strings) is pre-multiplied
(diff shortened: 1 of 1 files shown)


css-viewport/Overview.bs
```diff
@@ -369,7 +369,7 @@ only the value previously set to it.
 An element becomes zoomed when the 'zoom' property has a positive computed value different than 1
 (or when a flat tree ancestor has zoom).
 
-To apply zoom, the [=used value=] of a CSS property is pre-multiplied
+To apply zoom, the [=used value=] of a CSS property (including values inside of [=path=] strings) is pre-multiplied
```
chrishtr, line 372:
```suggestion
To apply zoom, the [=used value=] of a CSS property (including values inside of [=path()=] strings) is pre-multiplied
```

emilio (approved):

Issue #10145, opened 2024-03-26
[css-view-transitions-1] Should view transition names be tree scoped?


khushalsagar:
The spec currently traverses all DOM elements to look for view transition names, see text [here](https://drafts.csswg.org/css-view-transitions-1/#capture-old-state-algorithm:~:text=block%20size.-,For%20each%20element%20of%20every%20element%20that%20is%20connected%2C%20and%20has%20a%20node%20document%20equal%20to%20to%20document,-%2C%20in%20paint%20order). This should probably be tree scoped similar to other CSS naming conventions like [scroll timeline](https://github.com/w3c/csswg-drafts/issues/8135) and I'm guessing anchor positioning (@tabatkins ?).

@nt1m fyi.

noamr:
+1 to view transition names being tree-scoped.

nt1m:
More work in the implementation but makes sense to me.

shonya3:
View-transition names have been working seamlessly inside shadow trees until now. After updating Chrome to version 126, view-transition-name no longer pierces the shadow tree. 
My entire project is built with custom elements (CE), including individual elements, pages, and the root. 
This change has significantly degraded the user experience of my website, and I am currently at a loss for how to address this issue.

vmpstr:
Perhaps we should follow [Scroll Timeline](https://drafts.csswg.org/scroll-animations/#other-specs) example here which treats everything as a flat tree:

> Like most operations in CSS besides [selector](https://drafts.csswg.org/selectors-4/#selector) matching, features in this specification operate over the [flattened element tree](https://drafts.csswg.org/css-scoping-1/#flat-tree).

In ViewTransitions, we limit `view-transition-name` in [this algorithm](https://drafts.csswg.org/css-view-transitions-1/#capture-old-state-algorithm), but in the same algorithm use flat tree ancestors, which seems contradictory.

vmpstr:
I'm not sure I understand the distinction. Transitions can run outside of the root element, but they can't reference things within the shadow DOM. Running a transition within the shadow DOM would only solve some cases, but not all of them. For example, icon to detailed view morph wouldn't be possible if the icon is a part of a custom web component.

nt1m:
If `view-transition-name: shadow` is applied to a shadow DOM element, having it exposed to the root element through `::view-transition-group(shadow)` and friends, is a violation of shadow DOM should work. The shadow DOM element and the root element are in different tree scopes.

bramus:
[...]

PR #10306, opened 2024-05-10
[css-view-transitions-1] Ignore names from shadow DOM

khushalsagar:
Applies resolution at https://github.com/w3c/csswg-drafts/issues/10145.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1404,7 +1404,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		1. Let |usedTransitionNames| be a new [=/set=] of strings.
 
 		1. [=list/For each=] |element| of every [=/element=] that is [=/connected=],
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1320,7 +1320,8 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 		1. Set |transition|'s [=ViewTransition/initial snapshot containing block size=] to the [=snapshot containing block size=].
 
 		1. [=list/For each=] |element| of every [=/element=] that is [=/connected=],
-			and has a [=node document=] equal to to |document|,
+			and has a [=node document=] equal to |document|,
+			and has a [=tree context=] equal to |document|,
```
nt1m, line 1324: Doesn't "capture the new state" also have a similar loop that should use the same check?

khushalsagar: Nice catch, done!

Issue #10230, opened 2024-04-19
[cssom-view] Remove `caret range` concept from `CaretPosition` interface


siliu1:
Current spec for `CaretPosition` interface has a `caret range` concept which is a live range. The `caret range` is only used to get the client rect of the `CaretPosition`. We should avoid creating the live range upfront because:

  1. the live range is not used at all except for getting client rect.
  2. Maintain live range can be expensive during DOM mutation.
  3. No browsers implemented the algorithm by creating live range upfront. (Firefox and Chromium both create a live range only when `CaretPosition::getClientRect()` is invoked.)
 

Propose to remove `caret range` concept from `CaretPosition` interface.

sanketj:
Completely agree that caret position shouldn't be live or backed by a live range, because of the perf concerns you called out. I think the spec should be updated to explicitly mention that.

annevk:
Yeah, I think the current definition is broken as someone could mutate the range and that would impact the result, which isn't what you want to happen.

smaug----:
yeah, the spec doesn't really make sense. Gecko does internally create a temporary Range just to call getBoundingClientRect and after that the Range is released, so it won't observe mutations to DOM.

I think this stuff comes initially from https://github.com/w3c/csswg-drafts/commit/532723f180343305397c6ede736854cb90ea6e8e

I don't see that being discussed in
https://www.w3.org/Bugs/Public/buglist.cgi?product=CSS&component=CSSOM%20View&resolution=---
So perhaps it came from the mailing list.

Here https://lists.w3.org/Archives/Public/www-style/2009Nov/0244.html
I didn't try to find the relevant TPAC meeting minutes.

Oh, and looks like I complained about the range already 13 years ago :) https://lists.w3.org/Archives/Public/www-style/2011May/0051.html

siliu1:
Thanks for the input. It seems that @smaug---- from Mozilla, @annevk from Apple, and @sanketj and myself from Microsoft/Chromium all agree that `CaretPosition` should NOT be backed by a live range.

@astearns Is it possible to get to resolution of this issue without bringing up to CSSWG meeting? Thanks.

siliu1:
cc'ing Googler @mfreed7.

mfreed7:
> cc'ing Googler @mfreed7.

LGTM. Incidentally, I'm also in favor of going back in time and making **every** range a static range, but that's a tougher lift.

astearns:
[...]

PR #10307, opened 2024-05-10
[cssom-view] Remove `caret range` concept from `CaretPosition` interface.

siliu1:
Per #10230, `CaretPosition` backed by live range can cause issues. We should remove the `caret range` concept from `CaretPosition` interface.

This is also a long-standing issue. https://lists.w3.org/Archives/Public/www-style/2011May/0051.html

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1070,9 +1070,6 @@ result of running these steps:
         <dt><a>caret offset</a>
         <dd>The amount of 16-bit units to the left of where the
         text insertion point indicator would have inserted.
-
-        <dt><a>caret range</a>
-        <dd>null
     </dl>
 1. Otherwise:
 
@@ -1085,7 +1082,6 @@ result of running these steps:
     1. Return a <a>caret position</a> with its properties set as follows:
         1. <a>caret node</a> is set to <var>startNode</var>.
         1. <a>caret offset</a> is set to <var>startOffset</var>.
-        1. <a>caret range</a> is set to a collapsed {{Range}} object which [=range/start node=] and [=range/end node=] are <var>startNode</var>, and which [=range/start offset=] and [=range/end offset=] are <var>startOffset</var>.
 
 Note: The specifics of hit testing are out of scope of this
 specification and therefore the exact details of
@@ -1118,7 +1114,7 @@ in that the latter can return a <code>frameset</code> element.
 <h3 id=the-caretposition-interface>The {{CaretPosition}} Interface</h3>
 
 A <dfn>caret position</dfn> gives the position of a text insertion point indicator. It always has an associated
-<dfn>caret node</dfn>, <dfn>caret offset</dfn>, and <dfn>caret range</dfn>. It is represented by a {{CaretPosition}} object.
+<dfn>caret node</dfn>, and <dfn>caret offset</dfn>. It is represented by a {{CaretPosition}} object.
 
 <pre class=idl>
 [Exposed=Window]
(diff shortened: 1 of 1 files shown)


cssom-view-1/Overview.bs
```diff
@@ -1118,7 +1114,7 @@ in that the latter can return a <code>frameset</code> element.
 <h3 id=the-caretposition-interface>The {{CaretPosition}} Interface</h3>
 
 A <dfn>caret position</dfn> gives the position of a text insertion point indicator. It always has an associated
-<dfn>caret node</dfn>, <dfn>caret offset</dfn>, and <dfn>caret range</dfn>. It is represented by a {{CaretPosition}} object.
+<dfn>caret node</dfn>, and <dfn>caret offset</dfn>. It is represented by a {{CaretPosition}} object.
```
kbabbitt, line 1117:
```suggestion
<dfn>caret node</dfn> and <dfn>caret offset</dfn>. It is represented by a {{CaretPosition}} object.
```

cssom-view-1/Overview.bs
```diff
@@ -1136,20 +1132,17 @@ The <dfn attribute for=CaretPosition>offset</dfn> attribute must return the <a>c
...
 1. If <a>caret node</a> is a text entry widget that is a replaced element,
     and that is in the document,
     return a [=scaled=] {{DOMRect}} object for the caret in the widget
     as represented by the <a>caret offset</a> value.
     The <a>transforms</a> that apply to the element and its ancestors are applied.
-1. Return null.
+1. Otherwise:
 
-Note: This {{DOMRect}} object is not <a spec=html>live</a>.
+    1. Let <var>caretRange</var> be a collapsed {{Range}} object which [=range/start node=] and [=range/end node=] are set to <a>caret node</a>, and which [=range/start offset=] and [=range/end offset=] are set to <a>caret offset</a>.
```
kbabbitt, line 1142:
```suggestion
    1. Let <var>caretRange</var> be a collapsed {{Range}} object whose [=range/start node=] and [=range/end node=] are set to <a>caret node</a>, and whose [=range/start offset=] and [=range/end offset=] are set to <a>caret offset</a>.
```

dandclark (approved):

kbabbitt (approved):

cssom-view-1/Overview.bs
```diff
@@ -1136,20 +1132,17 @@ The <dfn attribute for=CaretPosition>offset</dfn> attribute must return the <a>c
...
     The <a>transforms</a> that apply to the element and its ancestors are applied.
-1. Return null.
+1. Otherwise:
 
-Note: This {{DOMRect}} object is not <a spec=html>live</a>.
+    1. Let <var>caretRange</var> be a collapsed {{Range}} object whose [=range/start node=] and [=range/end node=] are set to <a>caret node</a>, and whose [=range/start offset=] and [=range/end offset=] are set to <a>caret offset</a>.
+    1. Return the {{DOMRect}} object which is the result of invoking the {{Range/getBoundingClientRect()}} method on <var>caretRange</var>.
 
-Issue(10230): Consider removing <a>caret range</a> concept from <a>caret position</a> interface.
+Note: This {{DOMRect}} object is not <a spec=html>live</a>.
```
sanketj, line 1145: Should we expand on this and clarify that the `CaretPosition` itself, not just the `DOMRect` is not live?

zcorpan: Sure.

zcorpan (approved):

cssom-view-1/Overview.bs
siliu1: Done.

cssom-view-1/Overview.bs
```diff
@@ -1136,20 +1134,17 @@ The <dfn attribute for=CaretPosition>offset</dfn> attribute must return the <a>c
...
 1. If <a>caret node</a> is a text entry widget that is a replaced element,
     and that is in the document,
     return a [=scaled=] {{DOMRect}} object for the caret in the widget
     as represented by the <a>caret offset</a> value.
     The <a>transforms</a> that apply to the element and its ancestors are applied.
-1. Return null.
+1. Otherwise:
 
-Note: This {{DOMRect}} object is not <a spec=html>live</a>.
+    1. Let <var>caretRange</var> be a collapsed {{Range}} object whose [=range/start node=] and [=range/end node=] are set to <a>caret node</a>, and whose [=range/start offset=] and [=range/end offset=] are set to <a>caret offset</a>.
```
sanketj, line 1144: This seems okay for now given this is how both Firefox and Chrome/Edge implement this. However, it seems less than ideal to have to create a range in order to get a bounding client rect. That range will still need to be maintained across mutations until it is garbage collected, which may not happen immediately. Is there a different option we can pursue that doesn't use a live range at all?

sanketj (approved):

songjianet (approved):

cssom-view-1/Overview.bs
siliu1: Thanks. I am only aware of such way to get DOMRect from (node, offset). Totally agree that we should update the spec if we have better way to do that.

PR #10309, opened 2024-05-11
[css-font-loading-3][editorial] Reference WebIDL BufferSource

ADKaster:
Rather than defining a custom BinaryData typedef, the FontSource interface should just use the BufferSource definition from WebIDL.

https://webidl.spec.whatwg.org/#BufferSource

I came across this implementing this API for the Ladybird Browser. Let me know if the issue is not as editorial as I think and needs an issue/meeting topic :).


ADKaster:
I have an affiliation set for the WhatWG repos (HTML, etc), but it doesn't looks like you guys are using the same system. Looks like I'll have to wait for someone to tell me how to fill out my organization on the w3c site for this to go through.

ADKaster:
Aha, we're not a W3c participating member (yet), so I created an account as myself

PR #10367, opened 2024-05-24
[css-mixins] Add CSS Functions and Mixins specification

andruud:
To begin with, only `@function` is described.

#9350

### DIFF
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -20,6 +20,14 @@ spec:css-syntax-3; type:dfn; text:descriptor;
 </pre>
 
 <!-- Big Text: intro
+
+████ █    █▌ █████▌ ████▌   ███▌
+ ▐▌  █▌   █▌   █▌   █▌  █▌ █▌  █▌
+ ▐▌  ██▌  █▌   █▌   █▌  █▌ █▌  █▌
+ ▐▌  █▌▐█ █▌   █▌   ████▌  █▌  █▌
+ ▐▌  █▌  ██▌   █▌   █▌▐█   █▌  █▌
+ ▐▌  █▌   █▌   █▌   █▌ ▐█  █▌  █▌
+████ █▌   ▐▌   █▌   █▌  █▌  ███▌
 -->
 
 Introduction {#intro}
@@ -162,6 +170,13 @@ during the [=substitute a dashed function|substitution=] of a <<dashed-function>
 
 <!-- Big Text: using
 
+█▌  █▌  ███▌  ████ █    █▌  ███▌
+█▌  █▌ █▌  █▌  ▐▌  █▌   █▌ █▌  █▌
+█▌  █▌ █▌      ▐▌  ██▌  █▌ █▌
+█▌  █▌  ███▌   ▐▌  █▌▐█ █▌ █▌ ██▌
+█▌  █▌     █▌  ▐▌  █▌  ██▌ █▌  █▌
+█▌  █▌ █▌  █▌  ▐▌  █▌   █▌ █▌  █▌
+ ███▌   ███▌  ████ █▌   ▐▌  ███▌
 -->
 
 Using Custom Functions {#using-custom-functions}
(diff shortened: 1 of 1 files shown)


css-mixins-1/Overview.bs
```diff
@@ -0,0 +1,538 @@
...
+		return failure.
+	2. For each value |parameter| in |function|'s [=function parameter|parameters=],
+		let |argument| be the corresponding value in |dashed function|'s argument list
+		at the same index:
+		* If |argument| does not exist,
+			set |argument| to the [=guaranteed-invalid value=].
+		* If |argument| is the [=guaranteed-invalid value=],
+			and |parameter| has a [=default value=],
+			set |argument| to that [=default value=].
+		* If |parameter| has a [=parameter type|type=],
```
tabatkins, line 264: This just turns the passed value into the guaranteed-invalid value if it doesn't match the type. If there's a default value, do we want to resolve it to that instead? (I'm really not sure.)

tabatkins (approved):
Got one question, but it's probably better as an issue anyway. Looks good, imo!

tabatkins:
Unless @mirisuzanne has any complaints, I think we should merge this and then bring it up to the WG for review.

css-mixins-1/Overview.bs
andruud: I _think_ what you suggest is more consistent, since not existing at all (which is where `<guaranteed-invalid>` came from in the first place) gives you the default value.

Although it's possible to argue both ways. If we take the default value, then it's like a function-wide fallback, and we don't need to duplicate the same fallback (potentially) in every `var()`. However, then we also can't use fallback A in one place, and fallback B in another place. The solution to this might be that authors simply _don't_ provide a default when they want per-`var()` fallbacks.

andruud: Done.

css-mixins-1/Overview.bs
```diff
@@ -0,0 +1,577 @@
...
+The <dfn>@function</dfn> Rule {#function-rule}
+----------------------------------------------
+
+The ''@function'' rule defines a [=custom function=],
+and its syntax is:
+
+<pre class="prod def" nohighlight>
+@function <<function-name>> [( <<parameter-list>> )]?
+	[using (<<dependency-list>>)]?
+	[returns type(<<string>>)]?
```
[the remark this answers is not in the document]
andruud: Or, maybe we can keep `type()`, but make it optional if there's a single [syntax component](https://drafts.css-houdini.org/css-properties-values-api-1/#syntax-component). For example:

Valid:
```
@function (--a <length>) { ... }
@function (--a <color>) { ... }
@function (--a <color>+) { ... }
```

Not valid:
```
@function (--a <number> | <percentage>) { ... }
```

However, valid:
```
@function (--a type(<number> | <percentage>)) { ... }
```

svgeesus:
@mirisuzanne does this look good to merge?

css-mixins-1/Overview.bs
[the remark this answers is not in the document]
andruud: For example, something like: https://github.com/andruud/csswg-drafts/pull/2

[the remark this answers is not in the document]
mirisuzanne: Yeah, I'm fine with any of these approaches to simplifying the types. The linked PR makes sense to me.

css-mixins-1/Overview.bs
```diff
@@ -0,0 +1,577 @@
...
+
+<dfn><<function-name>></dfn> = <<dashed-ident>>
+<dfn><<parameter-list>></dfn> = <<parameter>>#
+<dfn><<dependency-list>></dfn> = <<parameter>>#
+<dfn><<parameter>></dfn> = <<custom-property-name>> type(<<string>>)? [ : <<declaration-value>> ]?
+</pre>
+
+The name of the resulting [=custom function=] is given by the <<function-name>>,
+the [=function parameters=] are optionally given by <<parameter-list>>,
+the [=function dependencies=] are optionally  given by <<dependency-list>>,
```
mirisuzanne, line 116:
```suggestion
the [=function dependencies=] are optionally given by <<dependency-list>>,
```

mirisuzanne (approved):
This looks like a good start to me, setting aside the tangential/theoretical nature of `@nest`. Thank, @andruud!

PR #10392, opened 2024-06-05
[css-anchor-position-1] Fix markup

dontcallmedom:
`</di>` instead of `</div>` makes the gist of the spec appears under a note markup


dontcallmedom:
done separately

PR #10427, opened 2024-06-11
[css-box-4] Allowed combination of block and inline keywords in margin-trim

### LINKED ISSUE #7884: [css-box-4] Longhand values of `margin-trim` allow more combinations than shorthand values
With the change of #6643 we can now have different combinations of `block-*` and `inline-*` values for the `margin-trim` property.

The current syntax is:

```
none | block | inline | [ inline-start || inline-end || block-start || block-end ]
```

This syntax allows combinations of `*-start` and `*-end` values that for both block and inline directions. Currently, those cannot be represented by the shorthand values `block` and `inline`.

To allow to express those cases in a more condense way, I suggest to allow to combine the longhand and shorthand values.

The syntax would then be:

```
none |
[ block || inline ] |
[ block || inline-start || inline-end ] |
[ inline || block-start || block-end ] |
[ inline-start || inline-end || block-start || block-end ]
```

Sebastian

SebastianZ:
This adds the change for the [resolution for allowing the combination of `block` and `inline` keywords in `margin-trim`](https://github.com/w3c/csswg-drafts/issues/7884#issuecomment-2160955876).

Fixes #7884.

Sebastian

### DIFF
--- a/css-box-4/Overview.bs
+++ b/css-box-4/Overview.bs
@@ -373,7 +373,7 @@ Margins at Container Edges: the 'margin-trim' property {#margin-trim}
 
 	<pre class="propdef">
 		Name: margin-trim
-		Value: none | block | inline | [ block-start || inline-start || block-end || inline-end ]
+		Value: none | [ block || inline ] | [ block-start || inline-start || block-end || inline-end ]
 		Initial: none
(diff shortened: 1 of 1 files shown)


SebastianZ:
As this is a simple syntax change we resolved on, I am going to merge this.

Sebastian

PR #10430, opened 2024-06-12
[css-color-4] `<hue>` is already normalized

cdoublev:
I propose to remove the `<hue>` normalization in range [0,360) from the HSL to RGB algorithm. I think this is supposed to already have been applied ~~at parse time~~ before, as defined in [4.3. Representing Cylindrical-coordinate Hues: the `<hue>` syntax](https://drafts.csswg.org/css-color-4/#hue-syntax):

  > This number is normalized to the range [0,360).

Furthermore, this is not applied in other conversion algorithms (HWB to RGB, LCH to LAB, etc).


svgeesus (approved):
Agreed, it is already mentioned.

Issue #10433, opened 2024-06-12
[css-containment] Reorganizing the Containment specs


frivoal:
It seems to me that Container Queries would be more suited to be a standalone spec. It has significant normative dependencies onto CSS-Contain, but isn't about defining containment itself.

I would suggest:
* moving Container Queries to a separate spec, either standalone (probably), or as part of one of the newer levels of `css-conditional`
* moving `inline-size` containment from level 3 to level 2
(this makes contain-3 empty)

While we're at it, I wonder if something similar should happen to `content-visibility` being pulled into its own spec. The same arguments seem to apply, although maybe not as strongly. Alternatively, `content-visibility` is what becomes level 3, and we let level 2 be only about containment in the strict sense (it would differ from level 1 by adding `style` and `inline-size` containment).

I think that would help with stabilizing the core part of containment faster, and would reflect the de-facto division of labor better.

Co-editors (@mirisuzanne @tabatkins @vmpstr ), what do you think?

frivoal:
Btw, https://lists.w3.org/Archives/Public/www-style/2023Nov/0004.html resolved that we should start a css-contain-4 spec for State Container Queries. This document has not been created yet, but if we go ahead with what I proposed above, that would become Level n+1 of wherever we move Container Queries to.

mirisuzanne:
I think it's strange that we are alternating levels for fairly distinct features - so yes, I agree it would be helpful to split CQs into a different module.

tabatkins:
I don't have a strong opinion either way here. Splitting out CQs into their own track sounds reasonable.

frivoal:
There is currently no `css-contain-4` spec, but we do have it as a label for features deferred from level 3. Should I rename that label to `css-conditional-6`?

frivoal:
Should we republish css-contain-3 as a [discontinued draft](https://www.w3.org/policies/process/#abandon-draft), now that it's empty?

mirisuzanne:
Should we also update the working draft of css-conditional-5, so that this change doesn't result in container queries reverting from WD to ED?

mirisuzanne:
@astearns I think we need two publishing resolutions here. It's a bit strange that in the meantime, Container Queries have been entirely removed from any public working draft.

svgeesus:
> Should we also update the working draft of css-conditional-5,

[...]

PR #10447, opened 2024-06-14
[css-contain][css-conditional] Move container queries from contain-3 to conditional-5

frivoal:
See https://github.com/w3c/csswg-drafts/issues/10433

This PR makes no change to the normative text, just moves it, and adjusts the relevant abstracts/introductions/changelogs/…


css-conditional-5/Overview.bs
```diff
@@ -64,6 +94,11 @@ Introduction</h2>
 	such that only the first matching rule in a [=conditional rule chain=]
 	is applied.
 
+	It also adds Container Queries.
```
dbaron, line 97: I would suggest changing the order of the description of the three new pieces (font technology queries, `@when`/`@else`, and Container Queries) of the introduction.  Specifically:

- I think the bit about querying for supported font technologies should be last, since it's probably the smallest both in scope and size.

- I think the addition of container queries could *optionally* be before the introduction of `@when` and `@else` since I think CQ is a more significant feature but `@when`/`@else` could perhaps be described as having broader scope over the features of the module, so I could see describing these in either order.

css-conditional-5/Overview.bs
```diff
@@ -336,4 +1241,40 @@ Additions since Level 4</h3>
 		<li>Added ''@when'' and ''@else''.
 		<li>Extended [=supports queries=] to express font capabilities
 			via ''font-tech()'' and ''font-format()''.
+		<li>Moved Container Queries from [[CSS-CONTAIN-3] to this specification.
+			(See also the [[CSS-CONTAIN-3#changes]] for more information
+			on the evolution of this feature.)
 	</ul>
+
+<h2 class="no-num" id="acknowledgments">Acknowledgments</h2>
```
dbaron, line 1249: I think you should probably remove me and Elika from this list since editors shouldn't acknowledge themselves.

css-contain-3/Overview.bs
```diff
@@ -1044,8 +120,6 @@ Changes from <a href="https://www.w3.org/TR/css-contain-2/">CSS Containment Leve
 	* Defines the terms, properties, units, and at-rule
 		needed for [=Container Queries=]
 
-ISSUE: [[css-contain-2#changes]]
-
 <h2 class="no-num" id="acknowledgments">Acknowledgments</h2>
```
dbaron, line 123: I have mixed feelings about retaining the Acknowledgments section here.  It seems maybe OK for now, except might make less sense once the draft is repurposed for something else.

dbaron (approved):
Looks good to me.  A few comments below (which others may have thoughts on as well).

css-conditional-5/Overview.bs
```diff
@@ -336,4 +1241,40 @@ Additions since Level 4</h3>
...
+	Amelia Bellamy-Royds,
+	Anders Hartvoll Ruud,
+	Brian Kardell,
+	Chris Coyier,
+	Christopher Kirk-Nielsen,
+	David Herron,
+	Elika J. Etemad (fantasai),
+	Eric Portis,
+	Ethan Marcotte,
+	Geoff Graham,
```
dbaron, line 1262: Also, I suspect @frivoal should probably be added to the Acknowledgments section (as a now-non-editor).

css-contain-3/Overview.bs
frivoal: I was wondering about that too, and for now erred on the side of not deleting acknowledgements, as doing that can be seen as crass if not fully justified. As I did include them in the css-conditional-5 spec, maybe that's enough to remove it from css-contain-3 already?

css-conditional-5/Overview.bs
frivoal: Done

css-conditional-5/Overview.bs
frivoal: done

svgeesus (approved):
I agree with the comment from @dbaron about ordering. That aside, this all looks good.

dbaron (approved):

PR #10476, opened 2024-06-20
[css-pseudo-4] Simplified example of `::details-content`

yisibl:
1. By depending on the animation rules defined in https://drafts.csswg.org/css-contain-3/#content-visibility-animation the step-end and step-start values are not needed.

2. By depending on the change in 
https://github.com/whatwg/html/pull/10265#issuecomment-2145837051 
and
whatwg/html@fb3033a
the display: block is no longer needed.

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -1328,14 +1328,12 @@ Expandable contents of details element: the ''::details-content'' pseudo-element
 
     <pre class="lang-css">
 details::details-content {
-  display: block;
   opacity: 0;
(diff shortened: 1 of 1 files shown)


yisibl:
@dbaron PTAL, this is based on your submission: https://github.com/w3c/csswg-drafts/commit/ff353bd7ae2898f2e939149c2de77ea687a73eae

dbaron (approved):
Indeed.  This looks good -- I had too many places to try to make a similar change, and I didn't catch this one.

(Hopefully not needing `display: block` will stick, but it's possible that compatibility might require that we go back to requiring the `display: block`.)

PR #10482, opened 2024-06-21
[cssom] [css-page] Add orientation to CSSPageDescriptors

AlaskanEmily:
I found this was missing while implementing `CSSPageDescriptors` in Gecko.

`@page{ orientation: ... }` Is generally treated the same as `@page{ size: ... }` in this way and definitely is a descriptor on page-rules, so it should be included in `CSSPageDescriptors`.


emilio (approved):

emilio:
> [emilio](https://github.com/emilio) marked as non substantive for IPR from ash-nazg.

I did this because it's just an oversight but lmk if that's not right. Emily has linked since their W3C account but for some reason the bot still says the contrary...

PR #10496, opened 2024-06-25
[css-mixins-1] Use the <syntax> Production for param types

andruud:


andruud:
This is an "optional add-on" to the first css-mixins-1 text which makes typed parameters a lot less noisy (`--f type("<length>)` vs `--f <length>`). Please have a look.

It should probably eventually go into either css-properties-values-api or css-values, though.

css-mixins-1/Overview.bs
```diff
@@ -98,27 +101,46 @@ The ''@function'' rule defines a [=custom function=],
...
-<dfn><<parameter>></dfn> = <<custom-property-name>> type(<<string>>)? [ : <<declaration-value>> ]?
+<dfn><<type>></dfn> = <<syntax-component>> | type( <<syntax>> )
+<dfn><<parameter>></dfn> = <<custom-property-name>> <<type>>? [ : <<declaration-value>> ]?
 </pre>
 
 The name of the resulting [=custom function=] is given by the <<function-name>>,
 the [=function parameters=] are optionally given by <<parameter-list>>,
-the [=function dependencies=] are optionally given by <<dependency-list>>,
-and the [=return type=] is optionally given by the type string following the "returns" keyword.
+the [=function dependencies=] are optionally  given by <<dependency-list>>,
```
mirisuzanne, line 120: Sneaky little extra space, won't give up

```suggestion
the [=function dependencies=] are optionally given by <<dependency-list>>,
```

css-mixins-1/Overview.bs
```diff
@@ -575,3 +597,90 @@ Issue: Should we indeed use ''@nest'' for this purpose?
...
+	</pre>
+	The above, when parsed as a <<syntax>>,
+	would accept <<percentage>> values,
+	<<number>> values,
+	as well as the keyword <code>auto</code>.
+</div>
+
+<div class='example'>
+	<pre class='lang-css'>
+	red | &lt;color&gt;
```
mirisuzanne, line 666: Is the order important for this example? I don't think it should be, but not sure if we need to clarify that.

mirisuzanne (approved):
This looks good to me generally. Left a couple comments/questions.

css-mixins-1/Overview.bs
andruud: The order matters when the thing using the syntax definition (and the values produced using that syntax definition) wants to act differently on identifiers vs colors. For example, a registered custom property with `"red | <color>"` would compute `red` to `red`, but `blue` to `rgb(0, 0, 255)`.

For custom functions I don't think it actually matters, but I wrote this section with the intent to eventually move it somewhere more general.

mirisuzanne (approved):

Issue #6985, opened 2022-01-25
[css-scroll-snap] Proposing `:snapped`: exposing private snapped item browser state for developers and designers


argyleink:
📄  [Explainer](https://github.com/argyleink/ScrollSnapExplainers/blob/main/css-snap-target/readme.md)

Currently, [IntersectionObserver](https://snap-gallery.netlify.app/horizontal/target/) and or [hacks](https://snap-gallery.netlify.app/horizontal/smart/) (like making gap `100vmin` in a frame so JS can check to see which element is still within the viewport..) are ways to try and derive this, but it's not usually performant, timed properly or accurate. This proposal seeks to remove the hacks and inaccurate algorithms. The browser has an item target it snapped to, developers should have access to that element.

<br>

For example, it's nice to outline the snapped element to help reinforce the state of the scroller:

```css
:snapped-inline {
  outline: 3px solid hotpink;
}
```

<br>

or to lift a snapped header when snapped:

```css
section:snapped-y > header {
  box-shadow: 0 .5em 1em .5em lch(5% 5% 200);
}
```

<br>

There's a lot more use cases and details about the proposal (cyclical resolution, full set of pseudo selectors, etc) in the [Explainer](https://github.com/argyleink/ScrollSnapExplainers/blob/main/css-snap-target/readme.md)

Related: #5979 

Would love to present the details of the explainer to the CSSWG for feedback 🙂

frivoal:
I think there's no question that it'd be useful. The question, to me, is all about feasibility.

From the explainer:

> We are OK letting :snapped cycle like :hover, as long as it's only once per lifecycle update. We believe developers will be responsible with this, as they have with :hover.

What does that mean exactly? I suppose (especially given that @tabatkins is a co-author) that you're aware of https://wiki.csswg.org/faq#selectors-that-depend-on-layout

The `:hover` loop goes through hit-testing, and thus skipping hit-testing on relayout and waiting for most movement solves it. In particular, you don't particularly have to care that the loop is or isn't caused by using `:hover` itself. Just skip hit-testing in general when doing relayout, and you're good. Actual implementations may be more subtle, but conceptually that's sufficient, and not a source of meaningful breakage.

But snapping loops don't go through hit testing, so the thing you'd have to skip would be something else.

I presume you mean that you wouldn't resnap based on relayout, but that seems bad, at least in general. It would solve the problem, but cause other breakage.

```css
.container { scroll-snap-type: block mandatory; }
[...]

PR #10505, opened 2024-06-26
[css-scroll-snap-2] Drop scroll-start and add issue to :snapped

DavMila:
Towards resolving [6985](https://github.com/w3c/csswg-drafts/issues/6985#issuecomment-2166073296) this patch removes scroll-start from the spec and notes that :snapped is being replaced.

### DIFF
--- a/css-scroll-snap-2/Overview.bs
+++ b/css-scroll-snap-2/Overview.bs
@@ -140,10 +140,9 @@ Initial scroll target</h4>
 	</div>
 
 
-<h4 id="scroll-start-target-shorthand">
-The scroll-start-target shorthand</h4>
+<h4 id="scroll-start-target-propdef">scroll-start-target Property Definition</h4>
 
-	<pre class="propdef shorthand">
+	<pre class="propdef">
 		Name: scroll-start-target
 		Value: [ none | auto ]
 		Initial: ''none''
--- a/css-scroll-snap-2/Overview.bs
+++ b/css-scroll-snap-2/Overview.bs
@@ -39,12 +39,13 @@ Introduction {#intro}
 	<em>This is currently a delta spec over Scroll Snap 1.</em>
 
 	Scroll experiences don't always start at the beginning. Interactions with
-	carousels, swipe controls, and listviews often start somewhere in the middle,
-	and each require JavaScript to set this position on page load.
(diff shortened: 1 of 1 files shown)


css-scroll-snap-2/Overview.bs
```diff
@@ -137,93 +108,6 @@ Motivating Examples {#examples}
...
-Interaction when 'display' is toggled</h4>
-
-	Same behavior that animations follow with [[#first-layout]].
-
-<h4 id="rtl">
-Interaction with RTL and LTR</h4>
-
-	Logical properties are offered for length offsets that should be flow relative. Furthermore, the ''end'' and ''start'' keywords are always logical.
-
 The 'scroll-start-target' property {#scroll-start-target}
```
flackr, line 111: scroll-start-target says it is a shorthand but I think that's no longer true since becoming a single property, right?

DavMila: Yes that's correct. Good catch.

css-scroll-snap-2/Overview.bs
```diff
@@ -70,36 +71,6 @@ First Layout {#first-layout}
 Motivating Examples {#examples}
 ===============================
 
-		<div class="example">
-				A carousel that starts in the middle:
```
flackr, line 74: I think this is still a good example, it just requires that we add a class to the middle item in the carousel which sets scroll-start-target. Can you update the example instead?

DavMila: Sounds good, done.

flackr (approved):

tabatkins (approved):

css-scroll-snap-2/Overview.bs
cdoublev: Remnants of `scroll-start-target-*` longhands are still [there](https://github.com/w3c/csswg-drafts/blob/0ac5d9da9e408afe237aae6ecdd8e4cbc6741fd6/css-scroll-snap-2/Overview.bs#L176).

DavMila: Thanks for pointing this out. Created https://github.com/w3c/csswg-drafts/pull/10554 to remove them.

PR #10519, opened 2024-07-02
[cssom] Throw on empty 'nested declarations rule' insertion

andruud:
The current text expects 'parse a CSS declaration block' to return a syntax error, which it never actually returns.

Follow-up to 7d8f87c8b8de9b7dd64719aeac18004ea76fdad0.


emilio (approved):
r=me, can you file an issue to get this change sanity-checked by the group (ideally reference it from the commit message but I guess not a huge deal)?

PR #10528, opened 2024-07-04
[css-view-transitions-1] Fix tree-scoping from element-based to name-based.

### LINKED ISSUE #10145: [css-view-transitions-1] Should view transition names be tree scoped?

noamr:
The previous fix for using the tree context made it inconsistent with how shadow DOM styling works for things like anchor positioning, and made it so that e.g. `::part` cannot set a `view-transition-name`.

Refactored to make the `view-transition-name` a tree-scoped name, rather than check the tree context of the element.

This aligns better with the [existing resolution](https://github.com/w3c/csswg-drafts/issues/10145#issuecomment-2100931679) and should not require a new resolution.

Closes #10145

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1970,6 +1970,7 @@ Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230
 * Use snapshot containing block when capturing new state for document element. See <a href="https://github.com/w3c/csswg-drafts/issues/10177">issue #10177</a>.
 * Fix algorithm for dispatching updateDOMCallback promise.
 * Scope view transition names to matching tree context. See <a href="https://github.com/w3c/csswg-drafts/issues/10145">issue 10145</a>.
+* Fix scoping to match name instead of element. See <a href="https://github.com/w3c/csswg-drafts/issues/10145">issue 10145</a>.
 
 <h3 id="changes-since-2022-05-25">
 Changes from <a href="https://www.w3.org/TR/2023/WD-css-view-transitions-1-20230525/">2022-05-25 Working Draft</a>
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -538,6 +538,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			as either an old or new [=/element=]--
 			with the specified [=view transition name=].
 
+			Each [=view transition name=] is a [=tree-scoped name=].
+
+			Note: Since currently only document-scoped view transitions are supported, only view transition names that are associated with the document are respected.
+
 			The values <css>none</css> and <css>auto</css> are excluded from <<custom-ident>> here.
 
 			Note: If this name is not unique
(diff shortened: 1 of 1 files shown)


noamr:
@nt1m can you check if this sits well with you, and if it requires changes to the WebKit implementation?
If you or @fantasai thinks this needs an additional CSSWG resolution I'm happy to add it to the agenda.

khushalsagar (approved):
Add a line to the appendix for changes below. LGTM otherwise. Thanks!

noamr:
> Add a line to the appendix for changes below. LGTM otherwise. Thanks!

Done

noamr:
> @nt1m can you check if this sits well with you, and if it requires changes to the WebKit implementation? If you or @fantasai thinks this needs an additional CSSWG resolution I'm happy to add it to the agenda.

WPT: https://github.com/web-platform-tests/wpt/pull/47006

nt1m:
Not sure the spec change is worded quite right, if you have:

```
<style>
::part(party) {
  view-transition-name: document-scoped;
}
</style>
#shadow-root
    <style>
      div {
        width: 100px;
        height: 100px;
        view-transition-name: shadow-scoped !important;
      }
    </style>
    <div part="party"></div>
```

I expect the effective view transition name of the `<div>` to be `shadow-scoped`, and `document-scoped` being ignored. 

Since `shadow-scoped` is scoped to the shadow root and `document-scoped` ignored, the `<div>` wouldn't participate in the view transition.

The spec change suggests otherwise, although I think the Chromium change is correct.

Anyway it would be good to add coverage for that case + capturing ::before/::backdrop pseudo-elements (where we should use the originating element's scope) in those situations as well.

nt1m:
Fwiw, the change is fairly easy to adopt on WebKit's side: https://github.com/WebKit/WebKit/pull/30501

noamr:
> Not sure the spec change is worded quite right, if you have:
> 
> ```
> <style>
> ::part(party) {
>   view-transition-name: document-scoped;
> }
> </style>
> #shadow-root
>     <style>
>       div {
>         width: 100px;
>         height: 100px;
>         view-transition-name: shadow-scoped !important;
>       }
>     </style>
>     <div part="party"></div>
> ```
> 
> I expect the effective view transition name of the `<div>` to be `shadow-scoped`, and `document-scoped` being ignored.
> 
> Since `shadow-scoped` is scoped to the shadow root and `document-scoped` ignored, the `<div>` wouldn't participate in the view transition.

Hmm right, so the wording should be something along the line of the implementation, where we get the computed value for `view-transition-name` and check its associated context. Will revise and add a WPT for this (and for the other cases). Thanks!

nt1m (approved):
New version looks good, thanks!

Issue #10512, opened 2024-06-30
[resize-observer] Clarify external spec integrations


johannesodland:
The Intersection Observer specification has a section on [External Spec Integrations](https://www.w3.org/TR/intersection-observer/#external-spec-integrations) that makes the integration with the [HTML Event Loops Processing Model](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model) clear. 

I have not been able to find a similar reference in the [Resize Observer specification](https://drafts.csswg.org/resize-observer/). 

Should a similar section be added to the Resize Observer specification, to make it clear that the algorithms are called during the HTML event loop?

johannesodland:
Hope it’s ok that I tag you here, @Loirooriol and @emilio

emilio:
I mean, I think the only reason that's the case in intersection observer is because it used to be a monkey-patch... But sure I guess a note mentioning it's used from the HTML event loop is fine? I'd take such a PR unless Oriol objects :)

Loirooriol:
Sounds ok to me

johannesodland:
> I mean, I think the only reason that's the case in intersection observer is because it used to be a monkey-patch... But sure I guess a note mentioning it's used from the HTML event loop is fine? I'd take such a PR unless Oriol objects :)

I've attempted to add a note in  [#10512](https://github.com/w3c/csswg-drafts/pull/10533).

PR #10533, opened 2024-07-05
[resize-observer-1] Add note about invocations from update the rendering loop in the html spec #10512

johannesodland:
Adding a short note to clarify which of the algorithms are called from the [HTML Event Loops Processing Model](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model).

### DIFF
--- a/resize-observer-1/Overview.bs
+++ b/resize-observer-1/Overview.bs
@@ -393,6 +393,11 @@ Content rect for the <a>SVGGraphicsElement</a>s without CSS layout boxes is a re
 
 <h3 id="algorithms">Algorithms</h3>
 
+<p class="note">
+The following algorithms are invoked as a part of [=update the rendering=] in the html spec:
+<a>gather active resize observations at depth</a>, <a>has active resize observations</a>, <a>has skipped resize observations</a>, <a>broadcast active resize observations</a> and <a>deliver resize loop error notification</a>.
+</p>
+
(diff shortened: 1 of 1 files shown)


resize-observer-1/Overview.bs
```diff
@@ -393,6 +393,11 @@ Content rect for the <a>SVGGraphicsElement</a>s without CSS layout boxes is a re
 
 <h3 id="algorithms">Algorithms</h3>
 
+<p class="note">
+The following algorithms are invoked as a part of [=update the rendering=] in the html spec:
+<a>gather active resize observations at depth</a>, <a>has active resize observations</a>, <a>has skipped resize observations</a>, <a>broadcast active resize observations</a> and <a>deliver resize loop error notification</a>.
+</p>
```
fantasai: More diff-friendly spacing maybe?

```suggestion
Note: The following algorithms are invoked as a part of [=update the rendering=] in the html spec:
<a>gather active resize observations at depth</a>,
<a>has active resize observations</a>,
<a>has skipped resize observations</a>,
<a>broadcast active resize observations</a>
and <a>deliver resize loop error notification</a>.
```

fantasai: (No comment on whether to add this... that's up to the editors I think.)

emilio (approved):
Seems fine to me!

Issue #10415, opened 2024-06-10
[css-color-5] Missing forms to use to serialize color-mix and relative color


cdoublev:
CSS Colors 5 defines the absolute color function to use to serialize [`color-mix()`](https://drafts.csswg.org/css-color-5/#serial-color-mix) or a [relative color](https://drafts.csswg.org/css-color-5/#serial-relative-color) as a component of a resolved value, which depends on the color space specified in `color-mix()` or of the origin color.

But it does not say anything for `display-p3`, `a98-rgb`, `prophoto-rgb`, `rec2020`, or a custom color space.

I suspect the requirement should be:

  > The form used depends on the color space [specified with "in" | of the ~~origin~~ relative color]: `lab`, `oklab`, `lch`, `oklch`, serialize with the corresponding function, otherwise with `color()` and the corresponding color space (all sRGB functions use the `srgb` color space).

**edit:** I replaced *origin color* with *relative color*, because I think `rgb(from lab(...) ...)` should serialize with `color(srgb ...)`, as hinted by the relative color, not with `lab(...)`, as hinted by the origin color.

svgeesus:
> **edit:** I replaced _origin color_ with _relative color_, because I think `rgb(from lab(...) ...)` should serialize with `color(srgb ...)`, as hinted by the relative color, not with `lab(...)`, as hinted by the origin color.

Yes, indeed.

> But it does not say anything for `display-p3`, `a98-rgb`, `prophoto-rgb`, `rec2020`, or a custom color space.

I agree those should be added (originally, mixing was restricted to a subset of color spaces)

PR #10535, opened 2024-07-06
[css-color-5][editorial] Fix serialization form of mixed colors (#10415)

cdoublev:
#10415

### DIFF
--- a/css-color-5/Overview.bs
+++ b/css-color-5/Overview.bs
@@ -2903,6 +2903,22 @@ depends on the color space specified with "in":
 		<td>color(srgb-linear r g b) </td>
 	</tr>
 	<tr>
+		<td>display-p3</td>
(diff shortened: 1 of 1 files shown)


svgeesus:
No, that looses information. Although the table is verbose it does clarify for example what happens with `in hwb` and `in xyz` which your replacement wording does not.

cdoublev:
> it does clarify for example what happens with `in hwb`

Sorry, I wrongly thought `srgb`/`hsl`/`hwb` were a single color space with different scales, but it seems they are all individual color spaces that belong to the RGB *color model*.

I suggested a generic definition for easier maintenance and also to define the form to serialize `color()` with a `<custom-color-space>`. Do you want to add it to the table?

svgeesus (approved):
Thanks!

Issue #10514, opened 2024-07-01
[cssom] Element (or Node) should contain the properties offset... (offsetParent, offsetLeft, ...)


jogibear9988:
# Problem

I'm writing a Designer Application, where I show overlays for all Elements (like the Dev Tools). For this to work, I need to calculate the the corners of the rect of an Element. This works using getBoundingClientRect, but only if the Element is not transformed. If the Element for example is rotated, this will not return the points of the elemnt, but the outer rect. And also a elemnt could be inside of another transformed element (and so on).
I know calculate the combined transformation of the elemnt, but for this I need the position of the element inside of it's container. This works on HTMLElements (caus they have offsetLeft and offsetTop), but not on SVG or MathML Elements.

As also MathML and SVG Elements could be aligned inside of HTMLElements, they should also contain these Properties. And also, as they also could be positioned absolutely, they should contain the "offsetParent", to get to wich they are positioned to.

Additional, as also a TextNode could be positioned inside of a container, I think it should also get the "offset..." Properties. So maybe it should be added to the "Node" Interface insted of the "Element" interface.

jogibear9988:
see: https://github.com/whatwg/html/issues/10445

emilio:
Does `GeometryUtils.getBoxQuads` give you what you want? Seems like it would do exactly what you want?

jogibear9988:
maybe, I do not find any browser wich implements this API: getBoxQuads

jogibear9988:
seems only firefox has an implementation behind a flag: https://caniuse.com/mdn-api_document_getboxquads

jogibear9988:
Polyfills for Geometry API:

https://github.com/trusktr/geometry-interfaces
https://github.com/maxmaximov/getBoxQuads

They do not work with transformed elements. (Cause that is the hard part) (My Designer Code could work with transformed HTMLElements (https://github.com/node-projects/web-component-designer/blob/92baadd12ebc6821e2bbaac6b4728a6baa49ef0b/packages/web-component-designer/src/elements/helper/TransformHelper.ts#L148C21-L148C49)
Problem is, I need the offset in the container wich I do not get for SVG and MathML (and textNodes)

chrishtr:
> I'm writing a Designer Application, where I show overlays for all Elements (like the Dev Tools).

"DevTools overlay" does seem like a use case for `GeometryUtils.getBoxQuads` that can't be done in other ways, because of transforms...otherwise anchor positioning would work it seems.

[...]

PR #10538, opened 2024-07-08
[cssom-view-1] Added definition of getBoxQuads()

### LINKED ISSUE #10537: [css-omview] update description of getBoxQuads API
The getBoxQuads API is missing lot of its description.

Issue in Firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1107559

Description 
https://hacks.mozilla.org/2014/03/introducing-the-getboxquads-api/

jogibear9988:
Fixes #10537

A linked issue:
#10514

A firefox issue with a little bit of description:
https://bugzilla.mozilla.org/show_bug.cgi?id=1107559

A polyfill for the API
https://github.com/jogibear9988/getBoxQuadsPolyfill


jogibear9988:
One issues atm:
My polyfill only works for 2D transformations, 3D with perspective do not work (yet). Have to figure out how to do this. Cause of that, also my spec only works for 2D. I've found some spect for the 3D matrix, wich then also could maybe reused: https://drafts.csswg.org/css-transforms-2/#accumulated-3d-transformation-matrix-computation

SebastianZ:
I think it makes more sense to add another option to `getBoxQuads()` to consider the transformed coordinates instead of introducing a separate function `getCompleteTransform()` for that.

I imagine something like `getBoxQuads({considerTransformations: true})`.

Sebastian

jogibear9988:
> One issues atm: My polyfill only works for 2D transformations, 3D with perspective do not work (yet). Have to figure out how to do this. Cause of that, also my spec only works for 2D. I've found some spect for the 3D matrix, wich then also could maybe reused: https://drafts.csswg.org/css-transforms-2/#accumulated-3d-transformation-matrix-computation

should be fixed in current polyfill

jogibear9988:
> I think it makes more sense to add another option to `getBoxQuads()` to consider the transformed coordinates instead of introducing a separate function `getCompleteTransform()` for that.
> 
> I imagine something like `getBoxQuads({considerTransformations: true})`.
> 
> Sebastian

it's in a extra method, cause it is used from the 3 other functions in the spec

cssom-view-1/Overview.bs
```diff
@@ -1781,46 +1781,76 @@ Document includes GeometryUtils;
 typedef (Text or Element or CSSPseudoElement or Document) GeometryNode;
 </pre>
 
-The <dfn method for=GeometryUtils lt="getBoxQuads(options)|getBoxQuads()">getBoxQuads(<var>options</var>)</dfn> method must run the following steps:
-
+The helper method getCompleteTransform(anchestor) must run the following steps:
```
emilio, line 1784: You need to get the element as a parameter, and use `<div algorithm>` for this for example:

```
<div algorithm="get the complete transform">
  In order to <dfn export>get the complete transform</dfn> of an node <var>node</var> relative to a node <var>ancestor</ancestor> ...
</div>

File: cssom-view-1/Overview.bs [hunk 2]
```diff
@@ -1781,46 +1781,76 @@ Document includes GeometryUtils;
...
-
-  scale to 0 means divide by zero, return 0x0
-
-  cross-frames not allowed, throw WrongDocumentError?
-
-  points are flattened (3d transform), z=0. like getClientRect
-
-  test block in inline
+1. let currentElement be the element getCompleteTransform() called on.
+2. let transformationMatrix be the current transformation matrix of currentElement (https://www.w3.org/TR/css-transforms-2/#ctm)
```
emilio, line 1787: Please use bikeshed references for this.

cssom-view-1/Overview.bs
```diff
@@ -1781,46 +1781,76 @@ Document includes GeometryUtils;
...
+ 1. create a DOMQuad from the rect via DOMQuad.fromRect
+ 2. transform each point of the DOMQuad from 1. via convertPointFromNode
 </ol>
 
 The <dfn method for=GeometryUtils lt="convertPointFromNode(point, from, options)|convertPointFromNode(point, from)">convertPointFromNode(<var>point</var>, <var>from</var>, <var>options</var>)</dfn>
 method must run the following steps:
 
 <ol>
- <li><p class=issue>...
+1. get the resulting matrix between the element and the document.defaultView via the algorithm defined in getCompleteTransform
```
emilio, line 1838: All these algorithms need to be in terms of `node`s, as `GeometryUtils` also applies to text nodes and such.

cssom-view-1/Overview.bs
```diff
@@ -1781,46 +1781,76 @@ Document includes GeometryUtils;
...
+2. get the resulting matrix between the from and the document.defaultView via the algorithm defined in getCompleteTransform
+3. inverse the matrix in 2.
+4. Depending on the value of options.fromBox modify the point with the calculated style values of the element.
+   - 'border': no change
+   - 'margin': x - margin.left, y - margin.top
+   - 'padding': x + border.leftWidth, y + border.topWidth
+   - 'content': x + border.leftWidth + padding.left, y + border.topWidth + padding.top
```
emilio, line 1845: This is not actually based on the computed style but on the actual used padding box etc. Those boxes are defined already elsewhere and we should reuse the definitions.

emilio (changes requested):
Thanks, this looks like a good start. Please look at how other specs (specially recent ones) use bikeshed in order to get the formatting and references correct. Also this needs to work in terms of nodes and boxes, not elements.

whimboo:
Hi @jogibear9988, would you maybe have the time to follow-up on the review from Emilio? Thanks

cssom-view-1/Overview.bs
jogibear9988: Is node correct? Or should it be "GeometryNode"?
image

cssom-view-1/Overview.bs
jogibear9988: Where should I use a bikeshed reference? (Is there an editor wich helps in editing bs files)

cssom-view-1/Overview.bs
jogibear9988: I don't really find where I should link to.
maybe: https://drafts.csswg.org/css2/#Computing_widths_and_margins

cssom-view-1/Overview.bs
jogibear9988: I don't get what you mean with this....
getCompleteTransform has a parameter, wich is anchestor

jogibear9988: Added node to helper function

cssom-view-1/Overview.bs
Psychpsyo: > I don't really find where I should link to. maybe: https://drafts.csswg.org/css2/#Computing_widths_and_margins

Pretty sure those boxes are defined here: https://drafts.csswg.org/css-box-4/#box-edges

jogibear9988: Yeah, but I need to link to the Algorythm how they are calculated, or is this enough?

Psychpsyo: > Yeah, but I need to link to the Algorythm how they are calculated, or is this enough?

There is no algorithm like that for these. They are defined in English instead. Browsers have algorithms internally that match those descriptions and can call into those when a spec says to do something with one of these boxes, so linking to their definitions is enough to unambiguously tell an implementer what needs to happen.

More modern specs like to define their concepts in terms of algorithms to derive them, because that makes implementation easier and leaves less room for creative (mis-)interpretation of the definition.

TL;DR: You want to link to a concept's definition when using it. Whether that is an algorithm or not doesn't matter.

jogibear9988:
@emilio 
is it okay like now?

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
-  p1 = top left even in RTL
-
-  scale to 0 means divide by zero, return 0x0
-
-  cross-frames not allowed, throw WrongDocumentError?
-
-  points are flattened (3d transform), z=0. like getClientRect
-
-  test block in inline
+ 1. let currentElement be node.
```
emilio: currentNode, right? I think this should use `<var>current</var>` and other bikeshed markup to display properly.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+ 2. let transformationMatrix be the current transformation matrix of currentElement (https://www.w3.org/TR/css-transforms-2/#ctm)
+ 3. while currentElement is not null do
+ <ol>
```
emilio: Do you really need the `<ol>`?

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
 typedef (Text or Element or CSSPseudoElement or Document) GeometryNode;
 </pre>
 
-The <dfn method for=GeometryUtils lt="getBoxQuads(options)|getBoxQuads()">getBoxQuads(<var>options</var>)</dfn> method must run the following steps:
-
+The helper method getCompleteTransform(node, anchestor) must run the following steps:
```
emilio: I think you should probably add a `<dfn>` so that you can reference this properly elsewhere.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+	 6. if currentElement is the anchestor or the anchestor is null return the transformationMatrix
+ </ol>
+</ol>
 
-  pseudo-elements before/after are children of the element
+The <dfn method for=GeometryUtils lt="getBoxQuads(options)|getBoxQuads()">getBoxQuads(<var>options</var>)</dfn> method must run the following steps:
 
-  viewport boxes are all the same
+<ol>
+ 1. create an array as result container
```
emilio: `Let |result| be an empty [=list=] of [=DOMQuad=] objects.` etc, please look at existing bikeshed files to see how to make this work.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+ 2. for each css-fragment of the element run the following steps
```
emilio: of what element? Probably of `this`? Also fragment should link to a definition.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+ <ol>
+    1. Create a DOMRect of the Element with the x,y value of 0 and the offsetWith and offsetHeight.
```
emilio: This doesn't seem right, `offsetWidth` rounds to integers to begin with which this code shouldn't do, but also is not defined for non-elements, so this should probably do something like:

 * If `this` is a `Document`, then you need to use the layout viewport.
 * If `this` is an `Element` or `Text` node, then you need to iterate over the fragments.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+    2. Depending on the value of options.box modify the Rect from 1. with the calculated style values of the element.
+    - 'border': no change
+    - 'margin': https://drafts.csswg.org/css-box-4/#box-edges - use the margin-box
```
emilio: This should not just drop the urls like this, it should link to the definition. `ResizeObserver` has _very_ similar functionality, maybe crib from there?

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+    - 'padding': https://drafts.csswg.org/css-box-4/#box-edges - use the padding-box
+    - 'content': https://drafts.csswg.org/css-box-4/#box-edges - use the content-box
+    3. create a DomQuad from the rect via DOMQuad.fromRect
```
emilio: Proper capitalization `DOMQuad` and linking please.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+    4. Calculate the resulting transformation matrix between the element and options.realtiveTo (if relativeTo is unset use window.defaultView) via the algorithm defined in getCompleteTransform
```
emilio: `window.defaultView` is not a thing, also `s/realtiveTo/relativeTo`.

This should be untransforming the rect to the layout viewport if there's no `relativeTo`.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+ </ol>
+ 3. return the result container
 </ol>
 
 The <dfn method for=GeometryUtils lt="convertQuadFromNode(quad, from, options)|convertQuadFromNode(quad, from)">convertQuadFromNode(<var>quad</var>, <var>from</var>, <var>options</var>)</dfn> method
 must run the following steps:
 
 <ol>
- <li><p class=issue>...
+ 1. transform each point of the DOMQuad via convertPointFromNode
```
emilio, line 2385: This should link to `convertPointFromNode`, but also probably should have more precise details like:

```
 * Let <var>p1</var> be the result of calling [=convertFromFromNode=] with <var>quad</var>.p1, <var>from</var> and <var>options</var>.
 * ...
 * Return a new [=DOMQuad=] with <var>p1</var>...
```

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+ 1. create a DOMQuad from the rect via DOMQuad.fromRect
+ 2. transform each point of the DOMQuad from 1. via convertPointFromNode
 </ol>
 
 The <dfn method for=GeometryUtils lt="convertPointFromNode(point, from, options)|convertPointFromNode(point, from)">convertPointFromNode(<var>point</var>, <var>from</var>, <var>options</var>)</dfn>
 method must run the following steps:
 
 <ol>
- <li><p class=issue>...
+ 1. get the resulting matrix between the node and the document.defaultView via the algorithm defined in getCompleteTransform
```
emilio: Give these meaningful names like:

```
 * Let <var>thisTransformToViewport</var> be the result of calling [=getCompleteTransform=] with <var>this</var> and <var>this</var>' [=associated document=].
 * Let <var>fromTransformToViewport</var> be the result of ...
 * Invert <var>fromTransformToViewport</var>.
 * ...
```

Then reference those.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
+ 2. get the resulting matrix between the from and the document.defaultView via the algorithm defined in getCompleteTransform
+ 3. inverse the matrix in 2.
+ 4. Depending on the value of options.fromBox modify the point with the calculated style values of the element.
+   - 'border': no change
+   - 'margin': x - margin.left, y - margin.top
+   - 'padding': x + border.leftWidth, y + border.topWidth
+   - 'content': x + border.leftWidth + padding.left, y + border.topWidth + padding.top
```
emilio: Same, refer to standard terms. It seems `ResizeObserver` refers to the CSS2 border box area, etc.

cssom-view-1/Overview.bs
```diff
@@ -1861,46 +1861,74 @@ Document includes GeometryUtils;
...
 
 The <dfn method for=GeometryUtils lt="convertPointFromNode(point, from, options)|convertPointFromNode(point, from)">convertPointFromNode(<var>point</var>, <var>from</var>, <var>options</var>)</dfn>
 method must run the following steps:
 
 <ol>
- <li><p class=issue>...
+ 1. get the resulting matrix between the node and the document.defaultView via the algorithm defined in getCompleteTransform
+ 2. get the resulting matrix between the from and the document.defaultView via the algorithm defined in getCompleteTransform
+ 3. inverse the matrix in 2.
+ 4. Depending on the value of options.fromBox modify the point with the calculated style values of the element.
```
emilio: This should use `<var>point</var>` etc.

emilio (changes requested):
This still needs a bit of work to get the spec into a reasonable shape, but looks in the right direction. Thanks!

whimboo:
Hi @jogibear9988. Would you maybe have the time to update this PR based on the review feedback from Emilio? Thanks a lot!

jogibear9988:
Sorry, I've read them, and hopefully fix in the next days

whimboo:
Hi @jogibear9988. I wanted to check back with you about the status of this PR. Is it still something you can continue working on? Thanks!

jogibear9988:
@whimboo 
sorry, had not much time to work on this.
at the moment I fix my polyfill (https://github.com/jogibear9988/getBoxQuadsPolyfill) to also respect offset-.... css properties. Didn't know they exist. 
Also there is a small issue with webcomponents I try to fix.
Then I look to fix this spec.

jogibear9988:
Also I think the spec needs an upgrade to also include a function to calculate the complete transformation matrix between two elements.
My polyfill does this, and I need this also some time in my designer, and this is not possible only with the boxQuads result.
is this something I also should add here, or should this be an extra issue/pull req?

bramus:
> Also I think the spec needs an upgrade to also include a function to calculate the complete transformation matrix between two elements. My polyfill does this, and I need this also some time in my designer, and this is not possible only with the boxQuads result. is this something I also should add here, or should this be an extra issue/pull req?

This would need a separate PR – one that would need a CSS WG Resolution first to get the group’s consensus to adding it.

whimboo:
Hello @jogibear9988. I want to check back with you if you may have some time that this PR could be moved forward? Or is it blocked based on the last comment? Thanks!

cssom-view-1/Overview.bs
```diff
@@ -2192,26 +2194,107 @@ Document includes GeometryUtils;
...
+    1. Let <var>rect</var> be a {{DOMRect}} describing the <a>border edge</a> of
+        the <a spec=css-break>box fragment</a>,
+        with {{DOMRect/x}} and {{DOMRect/y}} set to 0,
+        and {{DOMRect/width}} and {{DOMRect/height}} set to the width and height
+        of the [=border box=] of the fragment.
+    1. If <var>options</var>.{{BoxQuadOptions/box}} is
+        {{CSSBoxType/"margin"}},
+        adjust <var>rect</var> to describe the <a>margin edge</a> of the fragment
+        by subtracting the used value of 'margin-left' from {{DOMRect/x}},
+        subtracting the used value of 'margin-top' from {{DOMRect/y}},
```
Psychpsyo: These calculations are not how you get a box fragment's margin/padding/border box edges. Whether a box fragment has margins, borders and padding along any given edge depends on many factors beyond just its used values. (Whether that edge adjoins a fragmentainer break and whether or not the box is fragmented along that edge and the value of the `box-decoration-break` CSS property on that box and the value of `margin-trim` on any ancestors whose inner edge that edge touches...) I'd just say which edge the values should correspond to and that the DOMRect should be positioned so that the border edge is at 0,0 instead of trying te re-define the conversion between the different edges from first principles somehow.

jogibear9988: Thanks, that makes sense.

I changed this in the latest update. `getBoxQuads()` now defines the fragment rect in local coordinates with the border edge at 0,0, and for the `"margin"`, `"padding"`, and `"content"` cases it refers to the fragment’s corresponding box area instead of re-deriving those edges from used values.

So this should now follow the standard box definitions rather than trying to reconstruct them from first principles.

jogibear9988:
also added some description to my polyfill: https://github.com/jogibear9988/getBoxQuadsPolyfill/blob/main/csswg-getBoxQuads-calculation.md

jogibear9988:
I also have a new library wich uses getBoxQuads API, to be able to create screenshots or vector exports of HTML pages: https://github.com/node-projects/layout2vector

whimboo:
@jogibear9988 is this PR ready for another review? If yes, mind re-requesting review from Emilio? Thanks.

jogibear9988:
Yeah, should be ready

cssom-view-1/Overview.bs
```diff
@@ -2177,7 +2179,7 @@ The <dfn attribute for=MouseEvent>offsetY</dfn> attribute must follow these step
 enum CSSBoxType { "margin", "border", "padding", "content" };
 dictionary BoxQuadOptions {
   CSSBoxType box = "border";
-  GeometryNode relativeTo; // XXX default document (i.e. viewport)
+    GeometryNode relativeTo; // if omitted, coordinates are relative to the layout viewport
```
emilio: nit: Indentation. Also the "if omitted.." should be in the spec prose

cssom-view-1/Overview.bs
```diff
@@ -2189,7 +2191,7 @@ interface mixin GeometryUtils {
   sequence&lt;DOMQuad> getBoxQuads(optional BoxQuadOptions options = {});
   DOMQuad convertQuadFromNode(DOMQuadInit quad, GeometryNode from, optional ConvertCoordinateOptions options = {});
   DOMQuad convertRectFromNode(DOMRectReadOnly rect, GeometryNode from, optional ConvertCoordinateOptions options = {});
-  DOMPoint convertPointFromNode(DOMPointInit point, GeometryNode from, optional ConvertCoordinateOptions options = {}); // XXX z,w turns into 0
+    DOMPoint convertPointFromNode(DOMPointInit point, GeometryNode from, optional ConvertCoordinateOptions options = {}); // result is flattened to 2D
```
emilio: nit: Ditto

cssom-view-1/Overview.bs
```diff
@@ -2200,26 +2202,158 @@ Document includes GeometryUtils;
...
-  DOM order
-
-  p1 = top left even in RTL
-
-  scale to 0 means divide by zero, return 0x0
-
-  cross-frames not allowed, throw WrongDocumentError?
+The {{GeometryUtils}} methods operate on a node's local box geometry.
+The {{GeometryUtils/getBoxQuads()}} method returns one {{DOMQuad}} per relevant
+box fragment, in content order.
```
emilio: This should be in flattened tree order, and should link to the definition.

cssom-view-1/Overview.bs
emilio: It's already, so just remove the comment.

cssom-view-1/Overview.bs
```diff
@@ -2200,26 +2202,158 @@ Document includes GeometryUtils;
...
+<div algorithm="get the complete transform">
+    To <dfn export>get the complete transform</dfn> of a {{GeometryNode}} <var>node</var>
+    relative to a {{GeometryNode}} <var>ancestor</var>,
+    run the following steps:
+
+    1. Let <var>current</var> be <var>node</var>.
+    1. Let <var>transformationMatrix</var> be a new {{DOMMatrix}}.
+    1. While <var>current</var> is not null:
+        1. If <var>current</var> is <var>ancestor</var>, return <var>transformationMatrix</var>.
+        1. Let <var>parent</var> be null.
```
emilio: Parent is not `parent` right? It's more of `ancestor`

emilio (changes requested):
Thanks, this looks a lot better! Some nits and tweaks but looks good to me with those fixed.

emilio (approved):
I think this is in a pretty good state now. Could you squash please? Also I don't know what needs to happen to make the IPR bot happy... @svgeesus can you help?

whimboo:
> I think this is in a pretty good state now. Could you squash please? Also I don't know what needs to happen to make the IPR bot happy... @svgeesus can you help?

As it looks like @jogibear9988 needs to [join the Cascading Style Sheets (CSS) Working Group](https://www.w3.org/groups/wg/css/join). Once done it should be fine.

jogibear9988:
Don‘t think this is so easy
https://www.w3.org/groups/wg/css/instructions/

Somebody needs to invite me

Psychpsyo:
> Don‘t think this is so easy https://www.w3.org/groups/wg/css/instructions/
> 
> Somebody needs to invite me

You can send an application to become an Invited Expert without explicitly being invited beforehand.

jogibear9988:
I‘d send it

svgeesus:
@jogibear9988 I will send you a request for an individual IPR commitment for this PR.

(There is no need for Invited Expert for this one PR)

jogibear9988:
> @jogibear9988 I will send you a request for an individual IPR commitment for this PR.
> 
> (There is no need for Invited Expert for this one PR)

Okay for me

jogibear9988:
can someone retrigger the IPR check?

jogibear9988:
and if we have this merged, what are the next steps? Is there a way to get this going forward in browsers? (for example move the API in FF away from the flag? get it implemented in chrome? (nearly every browser already has 90% of this already for the dev tools)

Psychpsyo:
> and if we have this merged, what are the next steps? Is there a way to get this going forward in browsers? (for example move the API in FF away from the flag? get it implemented in chrome? (nearly every browser already has 90% of this already for the dev tools)

For one, you can open browser bugs and write web platform tests for the feature (both of those should happen regardless). Implementing a well-tested feature that someone asked for is going to be an easier sell than just "spec says to do this". But the only thing that will definitely get this implemented in browsers is if you do the implementation yourself / pay someone to do it. There isn't much other special magic that can compel a browser vendor into doing something the spec tells them to, and every browser vendor has to prioritize between hundreds of things already.

jogibear9988:
> > and if we have this merged, what are the next steps? Is there a way to get this going forward in browsers? (for example move the API in FF away from the flag? get it implemented in chrome? (nearly every browser already has 90% of this already for the dev tools)
> 
> For one, you can open browser bugs and write web platform tests for the feature (both of those should happen regardless). Implementing a well-tested feature that someone asked for is going to be an easier sell than just "spec says to do this". But the only thing that will definitely get this implemented in browsers is if you do the implementation yourself / pay someone to do it. There isn't much other special magic that can compel a browser vendor into doing something the spec tells them to, and every browser vendor has to prioritize between hundreds of things already.

there already exist some:
https://github.com/web-platform-tests/wpt/blob/master/css/cssom-view/cssom-getBoxQuads-001.html
https://github.com/web-platform-tests/wpt/blob/master/css/cssom-view/cssom-getBoxQuads-002.html

yeah, only wonder, firefox has it since 2014 (12 years), and kept it behind a flag since then.

jogibear9988:
I added a few more tests:

https://github.com/web-platform-tests/wpt/pull/61015

whimboo:
@jogibear9988 I also saw that you created https://github.com/web-platform-tests/wpt/pull/50314 a while ago and which is still open and not reviewed. Is that PR up-to-date so that we could try to see it reviewed as well?

Regarding the remaining steps for Firefox, we seem to still have a [couple of bugs open for `getBoxQuads`](https://bugzilla.mozilla.org/buglist.cgi?query_format=advanced&short_desc_type=allwordssubstr&short_desc=getBoxQuads&bugidtype=include&list_id=18028832), and possibly some related issues as well.

@emilio, could you maybe provide an overview of what would still be needed on our side to move this API out from behind the flag once this PR got landed?

whimboo:
FYI I triggered the IPR revalidation and it seems to be all fine now. Now someone with permissions is needed to get it merged.

emilio:
> @emilio, could you maybe provide an overview of what would still be needed on our side to move this API out from behind the flag once this PR got landed?

More tests would be good, but on the Gecko side the main blocker was the spec and ideally some implementation interest from other browsers, see https://bugzilla.mozilla.org/show_bug.cgi?id=2026051

emilio (approved):
Let's try this, thanks!

jogibear9988:
https://bugzilla.mozilla.org/show_bug.cgi?id=1107559

whimboo:
Thanks a lot @jogibear9988 for getting it finished! It's great to see that we are closer to be able to offer support for these APIs via WebDriver BiDi (https://github.com/w3c/webdriver-bidi/issues/787).

jogibear9988:
> > and if we have this merged, what are the next steps? Is there a way to get this going forward in browsers? (for example move the API in FF away from the flag? get it implemented in chrome? (nearly every browser already has 90% of this already for the dev tools)
> 
> For one, you can open browser bugs and write web platform tests for the feature (both of those should happen regardless). Implementing a well-tested feature that someone asked for is going to be an easier sell than just "spec says to do this". But the only thing that will definitely get this implemented in browsers is if you do the implementation yourself / pay someone to do it. There isn't much other special magic that can compel a browser vendor into doing something the spec tells them to, and every browser vendor has to prioritize between hundreds of things already.

did this:
https://chromium-review.googlesource.com/c/chromium/src/+/8039580

PR #10546, opened 2024-07-09
[css-viewport] Fix bikeshed errors so the spec can be generated.

darktears:
[css-viewport] Fix bikeshed errors so the spec can be generated.


darktears:
@chrishtr please take a look.

chrishtr (approved):

PR #10548, opened 2024-07-09
[css-viewport] Define the viewport property in window

darktears:
In preparation to add the segments property we need to make sure that the viewport property is properly specified. This patch already adds the zoom property to the viewport interface as it was already defined.

### DIFF
--- a/css-viewport/Overview.bs
+++ b/css-viewport/Overview.bs
@@ -362,8 +362,27 @@ only the value previously set to it.
 }
 </pre>
 
-<h2 id='zoom-property'>
-	The 'zoom' property
+<h2 id=extensions-to-the-window-interface>Extensions to the {{Window}} Interface</h2>
+
+<pre class=idl>
+partial interface Window {
+[SameObject, Replaceable] readonly attribute Viewport? viewport;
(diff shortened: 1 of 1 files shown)


darktears:
I wasn't entirely sure about the `zoom` property and where it's supposed to be defined so I can rework the patch if it's not supposed to be on the `window.viewport` object.

bramus:
I don’t think `zoom` is an attribute that needs to be defined on the `Window` object. What the css-viewport spec describes is the CSS `zoom` property.

chrishtr:
> I don’t think `zoom` is an attribute that needs to be defined on the `Window` object. What the css-viewport spec describes is the CSS `zoom` property.

Oops, I didn't read this PR carefully enough. Indeed, it's not a property. I'll revert this PR.

bramus:
Rest of the PR looks good, though.

darktears:
I could have just reverted the IDL.

chrishtr:
I think it'd be better to just redo the PR. I'll review it promptly.

darktears:
@chrishtr see #10564

PR #10564, opened 2024-07-12
[css-viewport] Add the segments property (#9237).

### LINKED ISSUE #9237: [css-viewport-1] Bring the segments property into viewport

darktears:
This property exposes in JavaScript what is already exposed to CSS via env() variables.

TAG review w3ctag/design-reviews#689.

CSSWG Approval: [here](https://github.com/w3c/csswg-drafts/issues/9237#issuecomment-2160730855)

Fixes #9237.

### DIFF
--- a/css-viewport-1/Overview.bs
+++ b/css-viewport-1/Overview.bs
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
 	The {{Window/devicePixelRatio}} of a frame is multiplied by the [=effective zoom=] inherited by its parent frame.
 </h3>
 
+<h2 id=extensions-to-the-window-interface>Extensions to the {{Window}} Interface</h2>
+
+<pre class=idl>
+partial interface Window {
+[SameObject, Replaceable] readonly attribute Viewport viewport;
+};
+</pre>
+
+<h2 id=viewport>Viewport</h2>
+
+<h3 id="the-viewport-interface">The {{Viewport}} Interface</h3>
+
+<pre class=idl>
+[Exposed=Window]
+interface Viewport {
+  readonly attribute FrozenArray&lt;DOMRect>? segments;
+};
+</pre>
+
+<h2 id='segments'>
+	The <dfn attribute for=Viewport>segments</dfn> property
+</h2>
+
+The {{segments}} property is an array of {{DOMRect}} that represent the dimensions of each existing viewport segment.
+Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in CSS ''<length>/px''.
+
+Additonal details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
+
+The {{segments}} attribute must run these steps:
+1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return null.
+2. Returns null if there is only a single viewport segment and abort these steps.
+3. Otherwise, return the {{Viewport}}'s [[css-env-1#viewport-segments|segments]] array.
+
+<div class=non-normative>
+<em>This section is not normative.</em>
+
+The {{segments}} property represents an immutable snapshot of the device {{segments}} when the value was queried.
+If the device state changes (rotation, window resize, <a spec="DEVICE-POSTURE">posture</a> change), the value previously retrieved will be invalid. Developers can listen to {{ScreenOrientation/onchange|Screen Orientation}}, {{Window/resize|Window Resize}}, or {{DevicePosture/onchange|Posture}} changes events to detect device state changes.
+
+<div class="example">
+If a viewport of 400px by 400px is split horizontally into two side-by-side segments, the {{segments}} array will have the following content:
+<code highlight=javascript>[DOMRect(0, 0, 400, 200), DOMRect(0, 200, 400, 200)]</code>
+</div>
+
+</div>
+
 <h2 class="no-num" id="changes">Appendix A. Changes</h2>
 
 This appendix is <em>informative</em>.


bramus (approved):

darktears:
I folded part of the reverted #10548 into this.

css-viewport/Overview.bs
```diff
@@ -526,6 +526,39 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
 	The {{Window/devicePixelRatio}} of a frame is multiplied by the [=effective zoom=] inherited by its parent frame.
 </h3>
 
+<h2 id=extensions-to-the-window-interface>Extensions to the {{Window}} Interface</h2>
+
+<pre class=idl>
+partial interface Window {
+[SameObject, Replaceable] readonly attribute Viewport? viewport;
```
chrishtr, line 533: Is there a resolution to add this?

darktears: This was the suggested path by the resolution to add `segments` here rather than `window.visualViewport` as it was originally proposed. @bramus do you agree?

chrishtr: The resolution speaks of a "new layout viewport tbd". You're naming it viewport instead of layoutViewport - why? There is already a visualViewport on Window.

darktears: I don't have a strong opinion on the name. I know that @bramus had some thoughts on adding more properties beside `segments` so I'm not sure what's the best way to name it. @bramus @fantasai : any thoughts?

bramus: When specs speak of "the viewport" they mean the Layout Viewport, so `viewport` seems fine.

css-viewport/Overview.bs
```diff
@@ -526,6 +526,39 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+};
+</pre>
+
+<h2 id='segments'>
+	The <dfn attribute for=Viewport>segments</dfn> property
+</h2>
+
+The segments property is an array of DOMRects that represent the dimensions of each existing viewport segment. Additonal details about the definition of a viewport segment can be found here : [[css-env-1#viewport-segments]].
+
+The {{segments}} property represent an immutable snapshot of the device {{segments}} when the value was queried.
```
chrishtr, line 554:
```suggestion
The {{segments}} property represents an immutable snapshot of the device {{segments}} when the value was queried.
```

css-viewport/Overview.bs
```diff
@@ -526,6 +526,39 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+If the device state changes (rotation, window resize, <a spec="DEVICE-POSTURE">posture</a> change), the value previously retrieved will be invalid. Developers can listen to {{ScreenOrientation/onchange|Screen Orientation}}, {{Window/resize|Window Resize}}, or {{DevicePosture/onchange|Posture}} changes events to detect device state changes.
```
chrishtr, line 555: Make the second sentence a non-normative note.

css-viewport/Overview.bs
```diff
@@ -526,6 +526,39 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+
+The {{segments}} attribute must run these steps:
+1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return 0.
+2. Returns null if there is only a single viewport segment and abort these steps.
```
chrishtr, line 559: Why not return one segment?

darktears: Well it’s useless information since in a foldable device the state in which we have one segment the value would be the viewport width and height. It provides no additional information that the developer already has. When the device has one segment it should be treated as a normal device.

chrishtr: Yes but it's more consistent to just return the segments available, even if there is only one. Then developers don't have to put more conditionals in their code.

darktears: Also on Android when the device is a state where there is no multiple segments the API basically returns a null object signifying the developers there is nothing special going on.

darktears: Let’s take a step back, in my opinion there is no one segment view, the view is either “segmented” or it’s not. So always two or more.

chrishtr: Why though? Why shouldn't segments be considered a generalization of viewports? If developers want to know if there is no folding going on, they can look to see if there is one segment.

darktears: If they want to know if folding state they should use the Device Posture API not the number of segments.

chrishtr: ok...yet another reason not to return null then?

css-viewport/Overview.bs
```diff
@@ -526,6 +526,39 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+interface Viewport : EventTarget {
+  readonly attribute FrozenArray&lt;DOMRect> segments;
+};
+</pre>
+
+<h2 id='segments'>
+	The <dfn attribute for=Viewport>segments</dfn> property
+</h2>
+
+The segments property is an array of DOMRects that represent the dimensions of each existing viewport segment. Additonal details about the definition of a viewport segment can be found here : [[css-env-1#viewport-segments]].
```
chrishtr, line 552:
```suggestion
The {{segments}} property is an array of DOMRects that represent the dimensions of each existing viewport segment. Additonal details about the definition of a viewport segment can be found here : [[css-env-1#viewport-segments]].
```

css-viewport/Overview.bs
chrishtr: Ok. It's good that it's consistent with the environment variables.

css-viewport/Overview.bs
```diff
@@ -526,6 +526,39 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+
+The {{segments}} property represent an immutable snapshot of the device {{segments}} when the value was queried.
+If the device state changes (rotation, window resize, <a spec="DEVICE-POSTURE">posture</a> change), the value previously retrieved will be invalid. Developers can listen to {{ScreenOrientation/onchange|Screen Orientation}}, {{Window/resize|Window Resize}}, or {{DevicePosture/onchange|Posture}} changes events to detect device state changes.
+
+The {{segments}} attribute must run these steps:
+1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return 0.
+2. Returns null if there is only a single viewport segment and abort these steps.
+3. Otherwise, return the {{Viewport}}'s segments array.
```
chrishtr, line 560: Is there somewhere you can link to that defines the bounds of these segments? If not, spell it out here?

darktears: I link `[[css-env-1#viewport-segments]]` which points [here](https://drafts.csswg.org/css-env-1/#viewport-segments) earlier in the paragraph. It explains the concept of segments but doesn't explicitly say what would represent the DOMRect. Should I spell out something like "each DOMRect represent the geometry (x, y, width, height) of each segment".

chrishtr: Yes, and define in what coordinate space those 4 lie and how they are computed. Also add a non-normative Example or two.

chrishtr (approved):

css-viewport-1/Overview.bs
```diff
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+};
+</pre>
+
+<h2 id=viewport>Viewport</h2>
+
+<h3 id="the-viewport-interface">The {{Viewport}} Interface</h3>
+
+<pre class=idl>
+[Exposed=Window]
+interface Viewport : EventTarget {
```
emilio, line 543: Given we currently have no use for it being an `EventTarget`, we probably shouldn't make it so?

css-viewport-1/Overview.bs
```diff
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
 	The {{Window/devicePixelRatio}} of a frame is multiplied by the [=effective zoom=] inherited by its parent frame.
 </h3>
 
+<h2 id=extensions-to-the-window-interface>Extensions to the {{Window}} Interface</h2>
+
+<pre class=idl>
+partial interface Window {
+[SameObject, Replaceable] readonly attribute Viewport? viewport;
```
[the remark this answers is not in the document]
darktears: I answered in the closed conversations in this issue but null is when the device doesn’t have any segments (which is basically non foldables or foldables not in a folded state). This match the CSS MQ which only match for 2 and above segments. Have a look at the conversion and I can clarify further if needed.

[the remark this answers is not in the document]
emilio: If window.viewport can return null, then it should be defined in the spec when it does so. However making it return null when there are no segments seems wrong? Specially given we want to extend this object further.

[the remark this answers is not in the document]
darktears: Sorry my bad, I thought you asked about the segments property. I was opening this on my mobile phone and got quite confused.

I agree with you, this object will be extended so we should return it always even if the fields are null, in this case for now just the segments field.

darktears:
@emilio added your suggestions.

css-viewport-1/Overview.bs
```diff
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+</h2>
+
+The {{segments}} property is an array of {{DOMRect}} that represent the dimensions of each existing viewport segment.
+Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in ''<length>/px''.
+
+Additonal details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
+
+The {{segments}} attribute must run these steps:
+1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return 0.
+2. Returns null if there is only a single viewport segment and abort these steps.
```
bramus, line 559: Why not return an array with only 1 segment `{ x:0, y:0, width:window.innerWidth, height:window.innerHeight }` here? Seems more ergonomic for authors than checking for `null`/_falsy_.

bramus (changes requested):

bramus:
Left some comments looking at this from an author’s POV. Could be I missed some earlier discussion that explains why things are specced the way it currently is _(I was OOO for 4 weeks)_.

css-viewport-1/Overview.bs
darktears: I answered earlier in the comments. 

>Personally, I can't wrap my head around a single segment. It's either the viewport is segmented or it's not. If it's segmented, then it will be two or more segments.

> Currently in chromium-based browsers the media queries also will not resolve when the viewport is not segmented,@media (vertical-viewport-segments: 1) doesn't work. Also returning a segment that is the exact viewport size doesn't bring any value and that's why the Android native API doesn't give any value when the viewport is not segmented. I feel like developers will start writing code @media (vertical-viewport-segments: 1) or if (window.viewport.segments == 1) which is not what we want, there is no special treatment to be done when the view is not segmented, it's a regular device.

> Overall, we want the developers to think about what to do when the view is segmented, when it's not it is the default behavior (foldable device or not)."

css-viewport-1/Overview.bs
```diff
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+	The <dfn attribute for=Viewport>segments</dfn> property
+</h2>
+
+The {{segments}} property is an array of {{DOMRect}} that represent the dimensions of each existing viewport segment.
+Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in ''<length>/px''.
+
+Additonal details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
+
+The {{segments}} attribute must run these steps:
+1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return 0.
```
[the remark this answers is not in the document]
darktears: We can be consistent and use null.

css-viewport-1/Overview.bs
bramus: Thanks for the clarification.

css-viewport-1/Overview.bs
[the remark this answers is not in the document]
bramus: I think `null` in this case makes more sense.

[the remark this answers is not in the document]
darktears: Done.

bramus (approved):

css-viewport-1/Overview.bs
```diff
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+</h2>
+
+The {{segments}} property is an array of {{DOMRect}} that represent the dimensions of each existing viewport segment.
+Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in ''<length>/px''.
+
+Additonal details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
+
+The {{segments}} attribute must run these steps:
+1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return null.
+2. Returns null if there is only a single viewport segment and abort these steps.
```
emilio, line 559: The IDL is not nullable. You need to fix that too.

css-viewport-1/Overview.bs
```diff
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+3. Otherwise, return the {{Viewport}}'s segments array.
```
emilio, line 560: This should link to css-env-1

darktears: Fixed.

css-viewport-1/Overview.bs
darktears: Fixed.

darktears:
@emilio resolved your comments

css-viewport-1/Overview.bs
```diff
@@ -526,6 +526,52 @@ The <dfn export>unscaled</dfn> value of a CSS length relative to an element is t
...
+  readonly attribute FrozenArray&lt;DOMRect>? segments;
+};
+</pre>
+
+<h2 id='segments'>
+	The <dfn attribute for=Viewport>segments</dfn> property
+</h2>
+
+The {{segments}} property is an array of {{DOMRect}} that represent the dimensions of each existing viewport segment.
+Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in ''<length>/px''.
```
emilio, line 553: Might be worth pointing out that they are CSS pixels, not device pixels.

emilio (approved):
Thanks!

css-viewport-1/Overview.bs
darktears: What do you think of the last change?

emilio: WFM, we can always clarify in follow-ups as needed.

Issue #6867, opened 2021-12-08
[selectors] Pseudo-class to indicate when a slot has content


plinss:
I'm developing several web components and I'm finding a need to tell when a slot has content or not. (e.g. to apply margins or not so empty slots don't impact layout, but layout well when they do, etc.)

The :empty pseudo-class applies when the slot doesn't have direct content, even when it has slotted content, so doesn't work. Using slot:has(::slotted(*)) wouldn't detect text-only content.

Once possibility would be to redefine :empty to detect slotted content, but it may be useful to keep the current behavior to differentiate empty slots from slots with default content (and the change may break exiting content). So a new pseudo-class that only detects slotted content may be useful.

WickyNilliams:
I have also needed this on several occasions. At the moment, you need to (at the minimum) inspect `assignedNodes`/`assignedElements` to determine this, and then add css classes to elements to modify styling. This is far from ideal for such a common use case. 

A typical use case for me is to hide (`display:none`) a `<slot>` if it is has no assigned content. The reason being that i often set a slot's `display` to something other than `contents` e.g. making a slot a flex container so that i can use `gap` to space slotted children. 

So being able to target a slot with assigned content is the most pressing use-case. Though I think it's also worth considering whether there should be separate pseudo-class for each of these states:

* assigned content
* default content
* empty (i.e. no default or assigned content)

MaxArt2501:
I'd like to point out that a `:has-assigned-content` pseudo-class (or better name) would be nice to have since:
* developers would like to avoid using JavaScript for what's basically a declarative effect;
* working around with `:host(:not(:empty)) slot` (for unnamed slots) or `:host(:has([slot='foo'])) slot[name='foo']`:
  * is verbose;
  * has content limitations (e.g. for components with multiple slots);
  * doesn't work for manual node assignment.

I'm quite keen myself to set a `display` value different than `contents` for slots, but this would also allow to style sibling or parent nodes (via `:has()`).

castastrophe:
Perhaps inline with `:has`, `:has-assigned` (content could be a misnomer because it might contain text or nodes). This would enable styles such as:

```
slot[name='foo']:not(:has-assigned) {
   display: none;
}
```

A good partner to this spec would be #6620

castastrophe:
This seems like a related request in the WICG space: https://github.com/WICG/webcomponents/issues/936

AutoSponge:
[...]

Issue #5482, opened 2020-08-28
[css-scoping] Consider aligning ::slotted() for fallback content with implementations


lilles:
The scoping spec states that:

"The ::slotted() pseudo-element represents the elements assigned, after flattening"

According to the flattening algorithm in the HTML spec, that also includes the fallback children of a slot if that slot has no assigned nodes. This behavior was agreed on in [WebComponents issue 631](https://github.com/w3c/webcomponents/issues/631) and issues were reported for Blink and WebKit, but have not been fixed.

The current implementations walk the assignedSlot chain and apply ::slotted rules from each slot's scope, assignedSlot being null for fallback elements.

This is an example of how this works (in all browsers per current spec) with nested slot assignment:

```
<!doctype html>
<div id="host"><span class="a">Orange with blue border</span></div>
<script>
  const root = host.attachShadow({mode:"open"});
  root.innerHTML = `
    <style>
      ::slotted(.a) { border: 2px solid blue }
    </style>
    <div id="host">
      <slot></slot>
    </div>
  `;
  const inner_root = root.querySelector("#host").attachShadow({mode:"open"});
  inner_root.innerHTML = `
    <style>
      ::slotted(.a) { color: orange; }
    </style>
    <slot></slot>
  `;
</script>
```

Per current spec, moving the span into fallback content for the slot should also result in orange text and blue border, but this is unstyled in all browsers):

```
<!doctype html>
<div id="host"></div>
<script>
  const root = host.attachShadow({mode:"open"});
  root.innerHTML = `
    <style>
      ::slotted(.a) { border: 2px solid blue }
    </style>
    <div id="host">
      <slot><span class="a">Orange with blue border</span></slot>
    </div>
  `;
  const inner_root = root.querySelector("#host").attachShadow({mode:"open"});
  inner_root.innerHTML = `
    <style>
      ::slotted(.a) { color: orange; }
    </style>
    <slot></slot>
  `;
</script>
```

If the author wants to style a certain type of element the same whether it is assigned, or fallback, it can be accomplished with the current implementations adding an extra selector like this:

```::slotted(.a), slot > .a```

However, that only works in the scope where the fallback is. For the example above, you can do this for the tree with the border rule, but not when that slot is assigned to the slot of the inner root.

[...]

PR #10586, opened 2024-07-17
[css-scoping-1] add :has-slotted() psuedo class

keithamus:
Based on the discussion in https://github.com/w3c/csswg-drafts/issues/6867, and the conclusions within, I thought I'd have a go at specifying a `:has-slotted( <<compound-selector>> )` pseudo class.

I put this in `css-scoping-1`; the resolution was to add it in `css-selectors-5` but I thought it might be useful living next to `:host`, `:host()`, `:host-context()` and `::slotted()`. I am of course happy to move it if folks disagree with this choice.

My intent for this PR is largely to present something worthy of discussion. I hope I've achieved that, apologies if this PR amounts to just noise.

### DIFF
--- a/css-scoping-1/Overview.bs
+++ b/css-scoping-1/Overview.bs
@@ -424,6 +424,44 @@ Selecting Slot-Assigned Content: the ''::slotted()'' pseudo-element</h4>
 	The only way to style assigned text nodes
 	is by styling the <a>slot</a> and relying on inheritance.
 
+<h4 id='has-slotted-pseudo'>
+Matching Selectors Against Slot-Assigned Content: the '':has-slotted()'' pseudo-class</h4>
+
+	The <dfn selector>:has-slotted()</dfn> pseudo-class, when evaluated 
+	<a>in the context of a shadow tree</a>,
+	matches the <a>slot</a>, if one of its
+	<a lt="find flattened slottables">assigned, after flattening,</a> matches the provided <<compound-selector>>.
+	In any other context, it matches nothing. This pseudo-class only exists on <a>slots</a>.
+
+	The [=specificity=] of '':has-slotted()'' is that of a pseudo-class,
+	plus the [=specificity=] of its argument.
+
+	The grammar of the '':has-slotted()'' pseudo-element is:
+
+	<pre class=prod>:has-slotted( <<compound-selector>> )</pre>
+
+	<div class="example">
+		For example, say you had a component with both children and a shadow tree,
+		like the following:
+
+		<pre>
+			&lt;x-foo>
+				&lt;div id="one" slot="foo" class="foo">...&lt;/div>
+				&lt;"shadow tree">
+					&lt;slot id="theslot" name="foo">&lt;/slot>
+				&lt;/"shadow tree">
+			&lt;/x-foo>
+		</pre>
+
+		For a stylesheet within the <a>shadow tree</a>,
+		selectors like '':has-slotted(*)'', '':has-slotted(div)'', '':has-slotted(.foo)'', and '':has-slotted(#one)''
+		will match ''#theslot'', while '':has-slotted(p)'', '':has-slotted(:not(.foo))'', '':has-slotted(#two)''
+		will not.
+	</div>
+
+	Note: '':has-slotted()'' will only match the ''slot'' element, not the contents. Slotted contents
+	can be matched with the ''::slotted()'' pseudo-element.
+
 <!--
  ██████     ███     ██████   ██████     ███    ████████  ████████
 ██    ██   ██ ██   ██    ██ ██    ██   ██ ██   ██     ██ ██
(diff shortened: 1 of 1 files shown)


css-scoping-1/Overview.bs
```diff
@@ -424,6 +424,44 @@ Selecting Slot-Assigned Content: the ''::slotted()'' pseudo-element</h4>
 	The only way to style assigned text nodes
 	is by styling the <a>slot</a> and relying on inheritance.
 
+<h4 id='has-slotted-pseudo'>
+Matching Selectors Against Slot-Assigned Content: the '':has-slotted()'' pseudo-class</h4>
+
+	The <dfn selector>:has-slotted()</dfn> pseudo-class, when evaluated 
+	<a>in the context of a shadow tree</a>,
+	matches the <a>slot</a>, if one of its
+	<a lt="find flattened slottables">assigned, after flattening,</a> matches the provided <<compound-selector>>.
```
justinfagnani, line 433: "if one of its assigned, after flattening, matches "

Is there a missing work there? "child elements"?

css-scoping-1/Overview.bs
```diff
@@ -424,6 +424,44 @@ Selecting Slot-Assigned Content: the ''::slotted()'' pseudo-element</h4>
...
+	<pre class=prod>:has-slotted( <<compound-selector>> )</pre>
+
+	<div class="example">
+		For example, say you had a component with both children and a shadow tree,
+		like the following:
+
+		<pre>
+			&lt;x-foo>
+				&lt;div id="one" slot="foo" class="foo">...&lt;/div>
+				&lt;"shadow tree">
```
justinfagnani, line 450: Do you want to use the `<::shadow>` markup here?

Westbrook:
Exciting start! Points to the need for `:host:has(...)` support so that we aren't required to ship arbitrary wrapping elements, but that's a argument for another day.

If you haven't already, I think I might take a pass some tentative WPTs based on this outline during the week of the 12th.

css-scoping-1/Overview.bs
```diff
@@ -424,6 +424,44 @@ Selecting Slot-Assigned Content: the ''::slotted()'' pseudo-element</h4>
 	The only way to style assigned text nodes
 	is by styling the <a>slot</a> and relying on inheritance.
 
+<h4 id='has-slotted-pseudo'>
+Matching Selectors Against Slot-Assigned Content: the '':has-slotted()'' pseudo-class</h4>
+
+	The <dfn selector>:has-slotted()</dfn> pseudo-class, when evaluated 
+	<a>in the context of a shadow tree</a>,
+	matches the <a>slot</a>, if one of its
+	<a lt="find flattened slottables">assigned elements, after flattening,</a> matches the provided <<compound-selector>>.
```
Westbrook, line 433: "after flattening" is interesting here because it give "non-slotted" content while being required to gather _all_ slotted content. E.G.:

```
<div>
  <template shadowrootmode="open">
    <style>
      div {
        color: green;
      }
      :has-slotted(div) {
        color: red;
      }
    </style>
    <slot><div>Test</div></slot>
  </template>
</div>
```
In the example above "Test" will display _red_ in the case that flattened elements are leveraged.

However, if you don't use flattened elements:
```
<div>
  <template shadowrootmode="open">
    <div>
      <template shadowrootmode="open">
        <style>
          div {
            color: green;
          }
          :has-slotted(div) {
            color: red;
          }
        </style>
        <slot></slot>
      </template>
      <slot><div>Test</div></slot>
    </div>
  </template>
</div>
```
This "Test" will also be _red_. And, without flattened elements,
```
<div>
  <template shadowrootmode="open">
    <div>
      <template shadowrootmode="open">
        <style>
          div {
            color: green;
          }
          :has-slotted(div) {
            color: red;
          }
        </style>
        <slot></slot>
      </template>
      <div>Test</div>
    </div>
  </template>
</div>
```
This "Test" will also be _red_.

I'm not sure it's a real thing, but is there a place between `::slotted(*)` and `.assignedElements({flatten: true})` that would resolve the _external_ assigned to a slot and not the _default_ ones internal to the shadow host?

sorvell: For consistency, I think it's important that `:has-slotted` matches using the same rules for matchable elements as `::slotted`. 

Concretely, since `::slotted()` never matches either `<slot>` or fallback content, `:has-slotted` also should *not match* when it would by selecting either `<slot>` or fallback content.

See https://github.com/w3c/csswg-drafts/issues/5482. IMO, the spec is still very confusing and possibly indicates that this should match, but the CSS working group resolved that the current behavior is intended in https://github.com/w3c/csswg-drafts/issues/5482.

I think matching against fallback content both for `::slotted` and `:has-slotted` is valuable, but I suggest it should be tackled as a follow-on to this feature.

LeaVerou:
@keithamus could you please rebase to resolve conflicts? Also, I wonder if it may be better to do this in two steps: get the non-functional form in first (`:has-slotted`) which _already_ addresses some of the use cases and is way easier to implement, and _then_ extend it to the functional form which appears to require more design discussion.

Looking at the [resolution](https://github.com/w3c/csswg-drafts/issues/6867#issuecomment-1332344714) it is unclear if it even allows for a functional form, given the original proposal was for a non-functional form. In fact, the non-functional form is not simply an alias for `:has-slotted(*)`, since the latter **would not detect slotted text nodes**.

yisibl:
ping~

Issue #6517, opened 2021-08-13
[mediaqueries-5] Script control of (prefers-*) queries


tabatkins:
In [this Discourse thread](https://discourse.wicg.io/t/allow-browsers-extensions-and-sites-to-set-user-preference-media-queries/5493/4), jimmyfrasche requests the ability for pages to override the (prefers-*) MQs in script, so they can easily integrate a manual toggle in the page without having to do silly duplication of their styles in both an MQ *and* a selector.

My initial response was that custom MQs would solve this, but they're actually somewhat non-trivial to do.

The trivial solution of "just make a script-based custom MQ that initially defers to (prefers-color-scheme), and can be overridden to true/false if you manually toggle" means that if script doesn't load, you have an unknown `(--dark-mode)` MQ in your page that is just always false, so you apply light-mode styles even if the user normally wants dark mode.

Instead you *must* double-layer them:

```
<style>
@custom-media --dark-mode (prefers-color-scheme: dark);
/* use @media (--dark-mode) {...} for stuff */
</style>
<script>
/* Assuming that a JS registration overrides any existing CSS-based ones */
CSS.customMedias.set("--dark-mode", ...);
</script>
```

And this only works for boolean MQs. If they're multi-state, it's trickier; you have to set up distinct custom MQs for each of the states and override all of them in script.

Would it instead make sense to allow authors to override *at least* the (prefers-*) queries with `CSS.customMedias`, supplying a CSS value that's valid for the MQ in question? That way you could instead write:

```
<style>
@media (prefers-color-scheme: dark) {...}
/* always works */
</style>
<script>
/* override when manual toggle is engaged */
CSS.customMedias.set("prefers-color-scheme", ...);
</script>
```

(I don't see a particular reason to limit it to *only* the (prefers-*) queries, but nor do I see a reason to *allow* it for things like (width). I mildly lean towards allowing override of anything, just so we don't have to think about which ones are allowed and which aren't and keep that list up to date, but I could easily be convinced otherwise.)

jimmyfrasche:
Thank you for filing the issue @tabatkins

While out of scope, I do think it's important to consider web extensions here. Namely, a very natural extension is one that lets you toggle preferences per site and if there's only one thing to set then there's a race between the site and the extension: if they both attempt to override the preference the last one to do so wins.

That will be confusing for users and a perennial source of consternation for the poor maintainer fielding bug reports.

[...]

PR #10597, opened 2024-07-19
[mediaqueries-5] Migrate Web Preferences API proposal

lukewarlow:
Per resolution at https://github.com/w3c/csswg-drafts/issues/6517#issuecomment-2220975121 this migrates the [Web Preferences API](https://wicg.github.io/web-preferences-api/) to the Media Queries Level 5 spec.

### DIFF
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -34,6 +34,7 @@ spec:css-values-4;
 		text:<number>
 		text:<resolution>
 	type:dfn; text:relative length
+spec:dom; type:dfn; text:origin
 
 </pre>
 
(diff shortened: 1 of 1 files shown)


mediaqueries-5/Overview.bs
```diff
@@ -3216,6 +3217,306 @@ Automatic handling of User Preferences</h3>
 		allows for unlimited data or is on a metered plan.
 	</div>
 
+<h2 id=auto-pref>
+Script Control of User Preferences</h3>
+
+It is common for website authors to want to respect the user's system preferences while also allowing
```
lukewarlow, line 3223: Probably needs a bit of a better introduction here. Perhaps with more from https://wicg.github.io/web-preferences-api/#sec-intro

lukewarlow:
TODO: It seems like the proposal and the spec use different flavours of bikeshed syntax, so I probably need to migrate to match the spec @tabatkins I don't suppose there's a tool to convert from the shorthand to the more HTML like syntax is there?

TODO: Add to the privacy section, at the very least an inline issue to signpost that it needs filling in.

tabatkins:
> so I probably need to migrate to match the spec

Don't worry about it.

> Add to the privacy section, at the very least an inline issue to signpost that it needs filling in.

Doesn't need much since there isn't *actually* a privacy concern, but it would be good to highlight *why* there isn't, yes, along the lines of what we've discussed in the issues. Maybe something like:

> The {{PreferenceManager}} object allows querying some user-preference [=media features=]. This is not a privacy leak, as that information was always trivially available by using the [=media features=] themselves in the normal fashion. 
>
> The {{PreferenceManager}} object also allows overriding these user-preference [=media features=]; this is also neither a privacy nor accessibility regression, as the [=media features=] were already trivially ignorable by simply not querying them.

lukewarlow:
@tabatkins apologies for the massive delay in getting back to this. I've added in your suggestions to the privacy section. I'd appreciate an overall review when you get time. I'll work on migrating the issues across to this repo, probably once this PR is merged.

tabatkins (approved):

lukewarlow:
@tabatkins is this ready to merge? Just an fyi I don't have merge rights so I can't do it myself :)

lukewarlow:
Thanks!

PR #10598, opened 2024-07-19
[css-align-3] Example 3 changing the order of images to match the text

alkorlos:
In the first paragraph of the example and section heading, the order of the keywords `safe` and `unsafe`, but the figure for illustrates this keywords in the opposite order, which may cause confusion.

Changed the order of the images and the order of the image descriptions.


tabatkins:
Thanks, this does improve the readability. I tweaked the text a little more to make the example look better before merging.

PR #10603, opened 2024-07-20
[css-exclusions-1]  Fix spurious / in <img>

alkorlos:
Fix bikeshed errors.

The command:
```shell
bikeshed spec css-exclusions-1\Overview.bs
```
Throws errors when called locally:
```shell
LINE 206:5: Spurious / in <img>.
LINE 218:7: Spurious / in <img>.
...
```

### Bikeshed Documentation

> The spec command is the most common one you’ll use. It turns a Bikeshed source file into an output HTML file. The rest of this document mostly describes how to format a source file for use with this command.

[Source](https://speced.github.io/bikeshed/#cli-spec)
Could you please tell where the description how specifications are generated for the drafts.csswg.org domain can be found? Apparently, it's not done using the command `bikeshed spec`, otherwise the document can not be generated. Would like to have the ability to fully reproduce the process locally.

### DIFF
--- a/css-exclusions-1/Overview.bs
+++ b/css-exclusions-1/Overview.bs
@@ -203,7 +203,7 @@ The 'wrap-flow' property</h4>
   block's descendants to wrap around its <a>exclusion area</a>.
 
   <figure>
-    <img alt="LTR text wrapping on left edge, RTL text wrapping on right edge, and vertical text wrapping on top edge." src="images/exclusion-writing-modes.png" style="width: 70%" />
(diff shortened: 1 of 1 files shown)


tabatkins:
> Could you please tell where the description how specifications are generated for the drafts.csswg.org domain can be found? Apparently, it's not done using the command bikeshed spec, otherwise the document can not be generated. Would like to have the ability to fully reproduce the process locally.

They're generated with the `-f` flag, as in `bikeshed -f spec`, which prevents fatal errors from stopping generation.

alkorlos:
Thanks for the answer!

PR #10607, opened 2024-07-21
[css-view-transitions-2] Stop extending CSSRule with obsolete pattern

### LINKED ISSUE #10606: [css-view-transitions-2] CSSRule interface should not be extended
https://drafts.csswg.org/css-view-transitions-2/#cssom

If you look at https://drafts.csswg.org/cssom-1/#the-cssrule-interface , you'll see that the numbers are purely historical. In fact CSS `@scope` and `@starting-style` both don't return any number.

@khushalsagar @noamr @vmpstr @bokand  I assume this was just an oversight, but my suggestion would be to delete the `partial interface CSSRule {` bit.

nt1m:
Fixes #10606


noamr (approved):
Thanks!

PR #10614, opened 2024-07-24
[css-conditional-3] Fix Web IDL, `bool` should have been `boolean`

tidoust:
Introduced by recent commit:
https://github.com/w3c/csswg-drafts/commit/3dab354901b4c9b2a716475efb98445317c65727

There's no `bool` type in Web IDL, only `boolean`.


svgeesus (approved):
Thanks

PR #10617, opened 2024-07-24
[css-overflow-5] Create scroll markers from <a> elements.

flackr:
This changes the scroll marker primitive to be based on a regular <a> element within a focusgroup allowing multiple groups of markers to be created and leaning on the focusgroup proposal to establish the interaction behavior.

### DIFF
--- a/css-overflow-5/Overview.bs
+++ b/css-overflow-5/Overview.bs
@@ -207,6 +207,9 @@ Overflow Controls: the 'scroll-marker-group' property and pseudo-elements</h2>
 		::scroll-marker-group { contain: size !important; }
 	</pre>
 
+	The 'scroll-marker-group' implicitly behaves as a single focusable component,
+	establishing a <a href="https://open-ui.org/components/focusgroup.explainer/">focusgroup</a>.
+
 <h3 id="scroll-marker-pseudo">The ''::scroll-marker'' pseudo-element</h3>
 
 	When the computed 'content' value of a <dfn selector>::scroll-marker</dfn> pseudo-element is not 'none'
@@ -215,9 +218,9 @@ Overflow Controls: the 'scroll-marker-group' property and pseudo-elements</h2>
 	on its nearest ancestor [=scroll container=].
 	These boxes are added in the [=tree order=] of their <a>originating element</a>.
 
-	These pseudo-elements behave as a <{button}>
-	with {{HTMLButtonElement/scrollTargetElement}} set to its <a>originating element</a>.
-	They can be focused and invoked.
+	These pseudo-elements have their {{HTMLLinkElement/scrollTargetElement}} set to their <a>originating element</a>.
+	They have a <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex"><code>tabindex</code></a> of '-1',
+	making them focusable within their '::scroll-marker-group' either by arrow key navigation or via the tab key when active.
 
 <h2 id="fragmentation" class=no-num>
 Appendix A: Redirection of Overflow</h2>
@@ -885,9 +888,10 @@ Appendix B: Scroll navigation controls</h2>
 However, there is little feedback to the user regarding the current content being viewed, and
 the interaction model does not match the expectations of modern accessible UI components.
 
-This specification adds a mechanism for creating scroll navigation controls.
-The active marker, reflecting the current position, can be styled to give the user an indication of which section they are in.
-The set of markers are treated as a component following the accessibility guidelines for keyboard navigation within components.
+This specification adds a mechanism for creating scroll navigation controls within groups of <{a}> elements.
+A group is created through the <a href="https://open-ui.org/components/focusgroup.explainer/">focusgroup</a> attribute
+on an ancestor containing one or more <{a}> elements whose <a spec=html>indicated part</a> is on the current page.
+Within each group, the active marker reflects the current scroll position, and can be styled to give the user an indication of which section they are in.
 
 Use cases include a table of contents with links to relevant contents,
 markers for scrolling carousel pages,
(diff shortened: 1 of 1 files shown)


css-overflow-5/Overview.bs
```diff
@@ -885,51 +888,46 @@ Appendix B: Scroll navigation controls</h2>
...
-Issue: Explore whether scrolltarget can be more directly associated with anchor tags.
-
-<h3 id="scroll-target"><{button/scrolltarget}> attribute</h3>
+<h3 id="scroll-target">{{HTMLLinkElement/scrollTargetElement}} attribute</h3>
 
-	The <dfn element-attr for=button>scrolltarget</dfn> attribute turns a <{button}> element into a [=scroll marker control=].
-	This takes the ID of the element to target as its value.
+	An <{a}> element has its {{HTMLLinkElement/scrollTargetElement}}
+	initialized to the <a spec=html>indicated part</a> of its <dfn element-attr for=a>href</dfn> attribute value if it is on the current page,
+	turning it into a [=scroll marker control=].
```
tabatkins, line 906: An "indicated part" is something a document has (based on its url), not a url itself. Rephrase to something like:

> An a element, if it's a local link, has its scrollTargetElement initialized to the part of the [=document=] that would be the [=indicated part=] if the a element's url was the current url.

(Using whatever the correct version of terms like "local link" and "current url" are.)

css-overflow-5/Overview.bs
```diff
@@ -938,47 +936,51 @@ Issue: Explore whether scrolltarget can be more directly associated with anchor
...
+			1. Let <var>targets</var> be the set of {{HTMLLinkElement/scrollTargetElement}} for those controls sorted in [=tree order=].
+			1. Let <var>selected</var> be the first element in <var>targets</var>, or null if <var>targets</var> is empty.
+			1. For each <var>target</var> in <var>targets</var>:
+				1. Set <var>selected</var> to <var>target</var>.
+				1. Let <var>targetPosition</var> be the position that would be scrolled to if we scroll <var>target</var> into view.
+				1.	: If <var>targetPosition</var>'s scroll block offset is less than or equal to <var>position</var>'s scroll block offset, and
+						<var>targetPosition</var>'s scroll inline offset is less than or equal to <var>position</var>'s scroll inline offset.
+					::
+						Update <var>selected</var> to <var>target</var>.
+						Break.
```
tabatkins, line 960: This is super algorithmic, to a degree that I think hampers readability rather then helps clarity. Can boil it down to:

> Let |markers| be ....
> 
> Let |targets| be the {{HTMLLinkElement/scrollTargetElement}}s of |markers|, associated with the item of |markers| they came from, and sorted in tree order.
> 
> For each item in |targets|, associate it with a |target position|, which is ...
> 
> Let |selected target| be the largest-indexed item of |targets| whose associated |target position| is ...
> If there is no such item, return.
> 
> Let |selected marker| be the marker associated with |selected target|.
> If multiple items of |markers| are associated with |selected target|,
> set |selected marker| to be the marker that is earliest in tree order.
> 
> ...

css-overflow-5/Overview.bs
```diff
@@ -938,47 +936,51 @@ Issue: Explore whether scrolltarget can be more directly associated with anchor
...
+						::
+							Focus <var>marker</var>
+					1. Set the <a href="https://open-ui.org/components/focusgroup.explainer/#last-focused-memory">last-focused element</a> of the <var>group</var>
+						to <var>marker</var>.
+				: Otherwise,
+				::
+					Set the 'checked' state of all controls in <var>markers</var> to false.
 	</div>
 
 	Issue: Should we allow for none of the markers to be currently active, e.g. if not yet scrolled past the position of the first marker.
```
tabatkins, line 975: This already happens in this algorithm, right? I think you can remove this issue now.

tabatkins (changes requested):
Two nits, but otherwise looks good.

css-overflow-5/Overview.bs
flackr: Looks like the [select the indicated part](https://html.spec.whatwg.org/multipage/browsing-the-web.html#select-the-indicated-part) algorithm already returns null if the target URL is not local, so I can invoke that.

css-overflow-5/Overview.bs
flackr: So in the previous algorithm we always select at least one marker. In your suggested update we don't. I think using your suggestion for the updated algorithm is fine.

css-overflow-5/Overview.bs
flackr: Thanks for the suggestion. I've adapted the algorithm to follow roughly what you suggested here with the appropriate sections filled in.

css-overflow-5/Overview.bs
```diff
@@ -938,47 +938,45 @@ Issue: Explore whether scrolltarget can be more directly associated with anchor
...
-			: Otherwise,
-			::
-				Set the 'checked' state of all controls in <var>markers</var> to false.
-	</div>
+		1. For each focusgroup <var>group</var> containing one or more [=scroll marker control=] elements whose {{HTMLLinkElement/scrollTargetElement}} is a descendant of [=scroll container=]:
 
-	Issue: Should we allow for none of the markers to be currently active, e.g. if not yet scrolled past the position of the first marker.
+			1. 	Let <var>markers</var> be all of the [=scroll marker control=] elements which are a part of the [=scroll marker group=] for the [=scroll container=].
+			1. 	Let <var>targets</var> be the {{HTMLLinkElement/scrollTargetElement}}s of |markers|, associated with the item of |markers| they came from, and sorted in [=tree order=].
+			1. 	For each |target| in |targets|, <a>determine the scroll-into-view position<a> of |target|, storing this as the associated |target position| of |target|.
```
tabatkins, line 954:
```suggestion
			1. 	For each |target| in |targets|, <a>determine the scroll-into-view position</a> of |target|, storing this as the associated |target position| of |target|.
```

css-overflow-5/Overview.bs
```diff
@@ -938,47 +938,45 @@ Issue: Explore whether scrolltarget can be more directly associated with anchor
...
+			1. 	Let |selected target| by the largest-indexed item of |targets|
```
tabatkins, line 955:
```suggestion
			1. 	Let |selected target| be the largest-indexed item of |targets|
```

css-overflow-5/Overview.bs
```diff
@@ -938,47 +938,45 @@ Issue: Explore whether scrolltarget can be more directly associated with anchor
...
+				whose associated |target position| is equal to or before |position| in both block and inline axis in the current writing mode direction of the [=scroll container=].
```
tabatkins, line 956:
```suggestion
				whose associated |target position| is equal to or before |position| in both the block and inline axises in the current writing mode direction of the [=scroll container=].
```

tabatkins (approved):
Just some spelling/markup nits you can accept edits on, then good to merge 👍

css-overflow-5/Overview.bs
tabatkins: I don't think that's true? You have <[`8d81d76` (#10617)](https://github.com/w3c/csswg-drafts/pull/10617/commits/8d81d767074d6fb4c96624f8b5dbe33518dbb128#diff-fdbff78a41c06b32b51fc29f900df82da5efa11aa9e2e3394047d39c889569f9L961)> which explicitly handles the "selected is non-null" and has an "else" condition for when selected *is* null. The algorithm itself appears to have been able to fail to find a selection, too, if nothing in the list is before the target.

Oh, hm, I think I see the issue; the algo contradicts itself. It *always* sets `selected` to `target` at the start of the loop over `targets` (<[`8d81d76` (#10617)](https://github.com/w3c/csswg-drafts/pull/10617/commits/8d81d767074d6fb4c96624f8b5dbe33518dbb128#diff-fdbff78a41c06b32b51fc29f900df82da5efa11aa9e2e3394047d39c889569f9L954)>), but then it conditionally sets `selected` to `target` again later in the loop (<[`8d81d76` (#10617)](https://github.com/w3c/csswg-drafts/pull/10617/commits/8d81d767074d6fb4c96624f8b5dbe33518dbb128#diff-fdbff78a41c06b32b51fc29f900df82da5efa11aa9e2e3394047d39c889569f9L959)>) before breaking. Ah, and it also set `selected` to the first item in `targets` *before* the loop (<[`8d81d76` (#10617)](https://github.com/w3c/csswg-drafts/pull/10617/commits/8d81d767074d6fb4c96624f8b5dbe33518dbb128#diff-fdbff78a41c06b32b51fc29f900df82da5efa11aa9e2e3394047d39c889569f9L952)>)

So the "if non-null" check didn't do anything unless `targets` was *empty*; otherwise `selected` was guaranteed to be (redundantly) set to something in `targets`. And if I'm reading it correctly now, I think it would actually be guaranteed to only select either the first item in `targets` that is before the eventual position (if any are), or the last item (if nothing is), when it should have been selecting the *last* item that is before the eventual position.

Anyway, it's no problem if you *do* think one of them should be selected by default when they're all below the eventual position, but if so, I suspect the default should be the first rather than the last. But yeah, I'm inclined to go with "none are selected" in that case instead.

PR #10621, opened 2024-07-25
Fix reverse HLG transform

benjamin-otte:
The +b needs to happen outside of the exp() call to be the reverse of the original HLG.

We found this while writing tests for GTK doing roundtrips to HLG and back. See https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7493


chrisn (approved):

svgeesus:
Thanks for noticing!

PR #10629, opened 2024-07-26
[css-view-transitions-2] Initial spec for nested view transitions

### LINKED ISSUE #10334: [css-view-transitions-2] Proposal for a view-transition-tree property (name tbd)
View transitions right now construct a tree of pseudo elements that in some sense "flattened". That is, each `::view-transition-group` is a direct child of `::view-transition`, which is a pseudo child of `:root`. This allows some nice effects where each piece of a transition moves independently.

However, there is also desire to have some relationship between the constructed groups. 
1. Some elements should move relative to their containers while their containers are also participating in a transition. This could be something like a box moving across the screen and the contents within that box also animate in an orthogonal axis. 
2. Some elements want to be clipped by the ancestor that is also participating in a transition. For example a button that stays the same size but its contents expand within the clip and crossfade.

Essentially what we want is for some elements to maintain their hierarchical relationship, where a `::view-transition-group` can have another `::view-transition-group` as a child.

The proposal is to add a new `view-transition-tree` property that takes two values: `flatten` and `preserve`. This property would only have an effect if a `view-transition-name` is also specified.

The `flatten` value would be the default value and act as it acts today, the element on which this appears does not affect how the view transition descendants are constructed.

The `preserve` value would act as a sort of a "containing block" for the view transition descendants: all of the descendants would be nested under their nearest `preserve` ancestor.

All of the names are bikesheddable, of course

/cc @nt1m @khushalsagar @noamr @jakearchibald

noamr:
This specifies the new 'view-transition-group' CSS property and how it used, with some overview and an initial example.

Based on [this resolution](https://github.com/w3c/csswg-drafts/issues/10334#issuecomment-2165761156).

There are still a lot of open issues but we will open them separately to avoid inflating the initial issue.

Closes #10334


css-view-transitions-2/Overview.bs
```diff
@@ -1080,8 +1081,24 @@ When capturing the old or new state for an element, perform the following steps
 
 ## Capturing and applying 'view-transition-group' ## {#vt-group-algorithm}
 
-<div algorithm="additional capture steps (group)">
-When capturing the old or new state for an element, perform the following steps given a [=captured element=] |capturedElement| and an [=element=] |element|:
+<div algorithm="additional old apture steps (group)">
```
khushalsagar, line 1084:
```suggestion
<div algorithm="additional old capture steps (group)">
```

css-view-transitions-2/Overview.bs
```diff
@@ -1080,8 +1081,24 @@ When capturing the old or new state for an element, perform the following steps
...
+When capturing the old or new state for an element, perform the following steps given a [=captured element=] |capturedElement|, an [=element=] |element| and a {{ViewTransition}} |transition|:
```
khushalsagar, line 1085:
```suggestion
When capturing the old state for an element, perform the following steps given a [=captured element=] |capturedElement|, an [=element=] |element| and a {{ViewTransition}} |transition|:
```

css-view-transitions-2/Overview.bs
```diff
@@ -1080,8 +1081,24 @@ When capturing the old or new state for an element, perform the following steps
...
+
+	1. Let |containingGroupName| be the [=used value=] of |element|'s ''view-transition-group''.
+
+	1. If |containingGroupName| is not null, then:
+
+		1. Let |groupContainerElement| be |transition|'s [=ViewTransition/named elements=][containingGroupName].
```
khushalsagar, line 1091: Don't we need a check to ensure that containingGroupName is an ancestor of this element?

css-view-transitions-2/Overview.bs
```diff
@@ -1080,8 +1081,24 @@ When capturing the old or new state for an element, perform the following steps
...
+
+		1. If |groupContainerElement| is not null, then:
+
+			1. Multiply |capturedElement|'s [=captured element/old transform=] by the inverse of |groupContainerElement|'s [=captured element/old transform=].
```
khushalsagar, line 1095: "Set capture’s old transform to a <transform-function> that would map element’s border box from the group container element's border box to its current visual position."

This seems more explicit about the intent than the current statement.

css-view-transitions-2/Overview.bs
```diff
@@ -1097,6 +1114,8 @@ When setting up the transition pseudo-element for a [=captured element=] |captur
 		1. Let |parentGroup| be the ''::view-transition-group()'',
 			with its [=view transition name=] set to |capturedElement|'s [=containing group=].
 
+		1. Multiply |group|'s 'transform' by the inverse of |parentGroup|'s 'transform'.
```
khushalsagar, line 1117: Same here.

noamr: Done

css-view-transitions-2/Overview.bs
noamr: Done

css-view-transitions-2/Overview.bs
```diff
@@ -732,6 +737,95 @@ div.box {
...
 
+# Nested view-transitions # {#nested-view-transitions}
+
+## Overview ## {#nested-overview}
+
+*This section is non-normative.*
+
+By default, setting `view-transition-name` on multiple elements generates a flat [=view transition tree=], where all the ''::view-transition-group()'' pseudo-elements are children of the ''::view-transition'' pseudo-element.
+This is sufficient for many simple use-cases, but there are some styling use-cases that cannot be achieved with a flat tree.
+The most prominent use-case is clipping: with a flat tree, all pseudo-elements are clipped to the root element, so clipping elements in the normal tree would lose their clipping during the view-transition, resulting in a broken visual effect.
```
khushalsagar, line 748:
```suggestion
The most prominent use-case is clipping: with a flat tree, all pseudo-elements are clipped to the [=snapshot containing block=], so clipping elements in the normal tree would lose their clipping during the view-transition, resulting in a broken visual effect.
```

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1080,56 @@ When capturing the old or new state for an element, perform the following steps
...
+
+				1. Set |capturedElement|'s [=captured element/old transform=] to the transform that would map |element|’s [=border box=] from | groupContainerElement|'s [=border box=] to its current visual position.
+
+				1. Set |capturedElement|'s [=captured element/containing group=] to |containingGroupName|.
+</div>
+
+<div algorithm="additional new capture steps (group)">
+When capturing the new state for an element, perform the following steps given a [=captured element=] |capturedElement| and an [=element=] |element|:
+
+	1. Set |capturedElement|'s [=captured element/containing group=] to the [=used value=] of |element|'s 'view-transition-group'.
```
khushalsagar, line 1108: Can you link to the issue for deciding the hierarchy mismatch situation. That's TBD.

Note for the current spec text: it's overriding the parent from the old DOM, if the name existed in the old DOM, without updating the old transform which seems sketchy. On the flip side, if we keep the ancestor from the old DOM but it doesn't exist in the new DOM then how do we compute the new transform..? Reviewing this, I'm more convinced of the use the LCA option.

css-view-transitions-2/Overview.bs
```diff
@@ -732,6 +737,95 @@ div.box {
...
+		:: This is equivalent to ''view-transition-group/normal'', except descendants with a [=computed value=] of ''view-transition-group/normal'' resolve this element's 'view-transition-name'.
+
+		: <dfn><<custom-ident>></dfn>
+		:: The '':view-transition-group()'' generated by this element's 'view-transition-name' should be a child of the ''::view-transition-group()'' pseudo-element
+			that matches the given <<custom-ident>>.
+	</dl>
+
+## Changes to '':view-transition-group()'' ## {#pseudo-element-hierarchy-nested}
+
+When the view [=view transition tree=] is generated, every generated ''::view-transition-group()'' becomes a child of an existing ''::view-transition-group()'' with the same name as its generating element's 'view-transition-group'.
```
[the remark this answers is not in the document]
khushalsagar: Actually you do have a section suggesting changes to this algorithm below. So maybe this is leftover from a previous PR?

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1080,56 @@ When capturing the old or new state for an element, perform the following steps
...
+</div>
+
+<div algorithm="additional new capture steps (group)">
+When capturing the new state for an element, perform the following steps given a [=captured element=] |capturedElement| and an [=element=] |element|:
+
+	1. Set |capturedElement|'s [=captured element/containing group=] to the [=used value=] of |element|'s 'view-transition-group'.
+</div>
+
+<div algorithm="additional pseudo-element setup steps (group)">
+When setting up the transition pseudo-element for a [=captured element=] |capturedElement|, given a |transitionName| and a {{ViewTransition}} |transition|:
```
khushalsagar, line 1112: For this one, could we do 2 sections for each algorithm that needs to be updated?

- https://drafts.csswg.org/css-view-transitions-1/#setup-transition-pseudo-elements-algorithm where we construct the tree once.
- https://drafts.csswg.org/css-view-transitions-1/#style-transition-pseudo-elements-algorithm where we compute the new element's transform each frame.

css-view-transitions-2/Overview.bs
khushalsagar: Sorry, I don't see the edits...

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs
noamr: Done

css-view-transitions-2/Overview.bs
noamr: Oops, pressed resolve instead of commit. Good now

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs
```diff
@@ -732,6 +737,95 @@ div.box {
...
+	Initial: normal
+	Inherited: no
+	Percentages: n/a
+	Computed Value: as specified
+	Animation type: discrete
+	</pre>
+
+	The 'view-transition-group' property can be used in conjuction with 'view-transition-name' to generate a hierarchy of [=named view transition pseudo-element=].
+
+	The [=used value=] for 'view-transition-group' resolves to a 'view-transition-name' in its ancestor chain, or to ''view-transition-name/none''. When generating the [=named view transition pseudo-element=], the ''::view-transition-group()'' with that name
```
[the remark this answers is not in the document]
noamr: Confirmed that this is OK... "used value" is what we want here.

css-view-transitions-2/Overview.bs
```diff
@@ -732,6 +738,95 @@ div.box {
...
+	would be the parent of the ''::view-transition-group()'' generated for this element's 'view-transition-name'.
+
+	When the [=computed value=] of 'view-transition-name' for an element is ''view-transition-name/none'', its 'view-transition-group' [=used value=] is always resolved to ''view-transition-name/none'' as well.
+
+	<dl dfn-type=value dfn-for=view-transition-group>
+		: <dfn>normal</dfn>
+		:: The '':view-transition-group()'' generated by this element's 'view-transition-name' should be a child of the ''::view-transition'' pseudo-element,
+			or of the nearest ancestor with a 'view-transition-group' [=computed value=] of ''view-transition-group/contain''.
```
khushalsagar, line 811:
```suggestion
			or of the nearest named ancestor with a 'view-transition-group' [=computed value=] of ''view-transition-group/contain''.
```

css-view-transitions-2/Overview.bs
```diff
@@ -732,6 +737,95 @@ div.box {
...
+
+		: <dfn>nearest</dfn>
+		:: The '':view-transition-group()'' generated by this element's 'view-transition-name' should be a child of the ''::view-transition-group()'' pseudo-element
+			generated by the nearest ancestor with a 'view-transition-name'.
+
+		: <dfn>contain</dfn>
+		:: This is equivalent to ''view-transition-group/normal'', except descendants with a [=computed value=] of ''view-transition-group/normal'' resolve this element's 'view-transition-name'.
```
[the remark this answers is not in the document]
khushalsagar: Please fix.

css-view-transitions-2/Overview.bs
```diff
@@ -732,6 +737,95 @@ div.box {
...
+
+		: <dfn><<custom-ident>></dfn>
+		:: The '':view-transition-group()'' generated by this element's 'view-transition-name' should be a child of the ''::view-transition-group()'' pseudo-element
+			that matches the given <<custom-ident>>.
```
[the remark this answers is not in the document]
khushalsagar: Please fix. If its TBD then leave a link to the issue.

css-view-transitions-2/Overview.bs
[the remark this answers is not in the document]
khushalsagar: Please fix.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
 	Note: This is written in a monkey-patch manner, and will be merged into the algorithm once the L1 spec graduates.
 </div>
 
+## Capturing and applying 'view-transition-group' ## {#vt-group-algorithm}
+
+<div algorithm="additional old capture steps (group)">
+When capturing the old state for an element, perform the following steps given a [=captured element=] |capturedElement|, an [=element=] |element| and a {{ViewTransition}} |transition|:
```
khushalsagar, line 1087: Can you link to the algorithm where this needs to go eventually. IIUC it's this one: https://drafts.csswg.org/css-view-transitions-1/#capture-old-state-algorithm

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1080,56 @@ When capturing the old or new state for an element, perform the following steps
...
 </div>
 
+## Capturing and applying 'view-transition-group' ## {#vt-group-algorithm}
+
+<div algorithm="additional old apture steps (group)">
+When capturing the old or new state for an element, perform the following steps given a [=captured element=] |capturedElement|, an [=element=] |element| and a {{ViewTransition}} |transition|:
+
+	1. Let |containingGroupName| be the [=used value=] of |element|'s ''view-transition-group''.
+
+	1. If |containingGroupName| is not null, then:
```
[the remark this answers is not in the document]
khushalsagar: Please fix.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1080,56 @@ When capturing the old or new state for an element, perform the following steps
...
+
+		1. Let |groupContainerElement| be |transition|'s [=ViewTransition/named elements=][containingGroupName].
+
+		1. If |groupContainerElement| is not null, then:
```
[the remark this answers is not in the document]
khushalsagar: Please fix.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
...
+
+			1. Let |containerElement| be the old element that is being captured for |containingGroupName|.
+
+			1. If |containerElement| is an ancestor of |element|, then:
```
khushalsagar, line 1099: Unnecessary, used value must be a valid ancestor if not none.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
...
+
+				1. Set |capturedElement|'s [=captured element/old transform=] to the transform that would map |element|’s [=border box=] from |containerElement|'s [=border box=] to its current visual position.
```
khushalsagar, line 1101:
```suggestion
				1. Set |capturedElement|'s [=captured element/old transform=] to the transform that would map |element|’s [=border box=] from |containerElement|'s [=border box=] origin to its current visual position.
```

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
...
+
+				1. Set |capturedElement|'s [=captured element/containing group=] to |containingGroupName|.
+</div>
+
+<div algorithm="additional new capture steps (group)">
+When capturing the new state for an element, perform the following steps given a [=captured element=] |capturedElement| and an [=element=] |element|:
```
khushalsagar, line 1107: Here as well, link to the algorithm where this needs to go: https://drafts.csswg.org/css-view-transitions-1/#capture-new-state-algorithm.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
...
+
+	1. Set |capturedElement|'s [=captured element/containing group=] to the [=used value=] of |element|'s 'view-transition-group'.
+</div>
+
+<div algorithm="additional pseudo-element setup steps (group)">
+When setting up the transition pseudo-element for a [=captured element=] |capturedElement|, given a |transitionName|:
```
khushalsagar, line 1113: Same here, please link to the relevant algorithm where these steps are added: https://drafts.csswg.org/css-view-transitions-1/#setup-transition-pseudo-elements-algorithm.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
...
+		1. Let |parentGroup| be the ''::view-transition-group()'',
+			with its [=view transition name=] set to |capturedElement|'s [=containing group=].
+
+		1. Append |group| to |parentGroup|.
+
+	Note: It is TBD how this is resolved when the old and new groups mismatch. See <a href="https://github.com/w3c/csswg-drafts/issues/10631">Issue 10631</a>.
+</div>
+
+<div algorithm="additional pseudo-element style update steps (group)">
+When update the style of the transition pseudo-element for a [=captured element=] |capturedElement|, given a |transitionName| and a {{ViewTransition}} |transition|:
```
khushalsagar, line 1129: Same here, a link to the relevant algorithm: https://drafts.csswg.org/css-view-transitions-1/#style-transition-pseudo-elements-algorithm.

khushalsagar, line 1129:
```suggestion
When updating the style of the transition pseudo-element for a [=captured element=] |capturedElement|, given a |transitionName| and a {{ViewTransition}} |transition|:
```

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
...
+
+	1. If |capturedElement|'s [=containing group=] is not null, then:
+
+		1. Let |group| be the ''::view-transition-group()'',
+			with its [=view transition =name=] set to |transitionName|.
+
+		1. Let |groupContainerElement| be |transition|'s [=ViewTransition/named elements=][|capturedElement|'s [=containing group=]].
```
khushalsagar, line 1136: Isn't "containing group" here just a string? You probably want to look it up in the map to get a reference to the element associated with the containing group.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1081,75 @@ When capturing the old or new state for an element, perform the following steps
...
+
+		1. Set |group|'s 'transform' to the transform that would map |capturedElement|’s [=new element=] [=border box=] from |groupContainerElement|'s [=border box=] to its current visual position.
+</div>
+
+<div algorithm="additional pseudo-element update style steps (group)">
```
khushalsagar, line 1141: Looks like this was unintentionally added, it's a copy of the algorithm above.

noamr: Removed

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1077,59 @@ When capturing the old or new state for an element, perform the following steps
...
+When updating the style of the transition pseudo-element for a [=captured element=] |capturedElement|, given a |transitionName| and a {{ViewTransition}} |transition|:
+
+	1. If |capturedElement|'s [=containing group=] is not null, then:
+
+		1. Let |group| be the ''::view-transition-group()'',
+			with its [=view transition name=] set to |transitionName|.
+
+		1. Let |containingGroupName| be |transition|'s [=ViewTransition/named elements=][|capturedElement|'s [=containing group=].
+
+		1. Let |groupContainerElement| be |transition|'s [=ViewTransition/named elements=][|containingGroupName|].
```
khushalsagar, line 1128: groupContainerElement is the captured element struct while we need the new element. And that can be null (for the exit transition case). So we need to check if `groupContainerElement`'s new element is not null and then do the transform stuff.

css-view-transitions-2/Overview.bs
```diff
@@ -805,28 +805,24 @@ and by applying ''view-transition-group'' to the internal element referencing th
 
 	When the [=computed value=] of 'view-transition-name' for an element is ''view-transition-name/none'', its 'view-transition-group' [=used value=] is always resolved to ''view-transition-name/none'' as well.
 
+	The <dfn>nearest containing group name</dfn> for an element is the 'view-transition-name' [=computed value=] of its nearest ancestor which has a  'view-transition-group' [=computed value=] of ''view-transition-group/contain'', or ''view-transition-group/none'' if such element does not exist.
```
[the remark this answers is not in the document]

css-view-transitions-2/Overview.bs
noamr: I refactored these algorithms, where the transform adjustment happens when setting the styles.

css-view-transitions-2/Overview.bs
```diff
@@ -1079,55 +1084,63 @@ When capturing the old or new state for an element, perform the following steps
...
-	1. If |capturedElement|'s [=containing group=] is not null, then:
+		1. Let |groupContainerElement| be |transition|'s [=ViewTransition/named elements=][|containingGroupName|].
 
 		1. Let |group| be the ''::view-transition-group()'',
-			with its [=view transition =name=] set to |transitionName|.
+			whose [=view transition name=] is set to |transitionName|.
 
 		1. Let |parentGroup| be the ''::view-transition-group()'',
-			with its [=view transition name=] set to |capturedElement|'s [=containing group=].
+			whost [=view transition name=] is set to |containingGroupName|.
```
khushalsagar, line 1124:
```suggestion
			whose [=view transition name=] is set to |containingGroupName|.
```

css-view-transitions-2/Overview.bs
```diff
@@ -1079,55 +1084,63 @@ When capturing the old or new state for an element, perform the following steps
...
 
 		1. Append |group| to |parentGroup|.
 
 	Note: It is TBD how this is resolved when the old and new groups mismatch. See <a href="https://github.com/w3c/csswg-drafts/issues/10631">Issue 10631</a>.
 </div>
 
+### Adjust the group's 'transform' to be relative to its containing ''::view-transition-group()'' ### {#vt-group-transform-adjust}
 <div algorithm="additional pseudo-element style update steps (group)">
-When updating the style of the transition pseudo-element for a [=captured element=] |capturedElement|, given a |transitionName| and a {{ViewTransition}} |transition|:
+When [[update pseudo-element styles|updating the style of the transition pseudo-element]], perform the following steps before setting the [=captured element/group styles rule=], given a [=captured element=] |capturedElement|, given a |transform| and a {{ViewTransition}} |transition|:
```
khushalsagar, line 1135: This is not linking to the algorithm?

css-view-transitions-2/Overview.bs
```diff
@@ -1079,55 +1084,63 @@ When capturing the old or new state for an element, perform the following steps
...
-			with its [=view transition name=] set to |transitionName|.
-
-		1. Let |containingGroupName| be |transition|'s [=ViewTransition/named elements=][|capturedElement|'s [=containing group=].
+	1. If |capturedElement|'s [=containing group name=] is not null, then:
 
-		1. Let |groupContainerElement| be |transition|'s [=ViewTransition/named elements=][|containingGroupName|].
+		1. Let |groupContainerElement| be |transition|'s [=ViewTransition/named elements=][|capturedElement|'s [=containing group name=]].
 
-		1. Set |group|'s 'transform' to the transform that would map |capturedElement|’s [=new element=] [=border box=] from |groupContainerElement|'s [=border box=] to its current visual position.
+		1. [=Multiply=] |transform| by the inverse matrix of |groupContainerElement|'s [=transform from snapshot containing block=].
```
khushalsagar, line 1143: This looks like a bug for the exit transition case. Since |transform| passed to this algorithm would be the "old transform" here: https://drafts.csswg.org/css-view-transitions-1/#style-transition-pseudo-elements-algorithm:~:text=Set%20transform%20to%20capturedElement%E2%80%99s%20old%20transform.

And we already multiplied the "old transform" with an inverse of its ancestor's transform in "setting up the transition pseudo-element" above?

css-view-transitions-2/Overview.bs
```diff
@@ -1079,55 +1084,63 @@ When capturing the old or new state for an element, perform the following steps
...
 
 		1. Let |group| be the ''::view-transition-group()'',
-			with its [=view transition =name=] set to |transitionName|.
+			whose [=view transition name=] is set to |transitionName|.
 
 		1. Let |parentGroup| be the ''::view-transition-group()'',
-			with its [=view transition name=] set to |capturedElement|'s [=containing group=].
+			whost [=view transition name=] is set to |containingGroupName|.
+
+		1. Set |capturedElement|'s [=captured element/old transform=] to the result of [=multiplying=] |capturedElement|'s [=captured element/old transform=] by the inverse matrix of |groupContainerElement|'s [=captured element/old transform=].
```
khushalsagar, line 1126: IIUC we expect both the captured element's old transform and the groupContainerElement's old transform to be relative to the snapshot containing block here.

Wouldn't there be a bug for the case where we have A nesting B which nests C then? Since we build the pseudo tree top down, when we get to C, B's transform would already be relative to A but C's transform would be relative to the snapshot containing block.

Maybe we want to cache a "old transform from snapshot containing block" which is not mutated after capture?

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs
noamr: Fixed, not multiplying in the previous phase.

css-view-transitions-2/Overview.bs
noamr: Fixed, not changing the old transform in the struct, only multiplying when setting the style.

css-view-transitions-2/Overview.bs
```diff
@@ -983,6 +1083,67 @@ When capturing the old or new state for an element, perform the following steps
...
+
+		1. Let |group| be the ''::view-transition-group()'',
+			whose [=view transition name=] is set to |transitionName|.
+
+		1. Let |parentGroup| be the ''::view-transition-group()'',
+			whose [=view transition name=] is set to |containingGroupName|.
+
+		1. Append |group| to |parentGroup|.
+
+		1. When setting the animation keyframes given |transform|, [=multiply=] |transform| by the inverse matrix of |groupContainerElement|'s [=old transform=].
```
khushalsagar, line 1129: When we get to merge the specs, would be nice to rename this to "transform from snapshot containing block".

khushalsagar (approved):

Issue #10580, opened 2024-07-16
[css-easing-2] Linear easing function requires at least two points?


pyoor:
Is it correct to assume that the linear-easing-function requires at least two points?  If that's the case, shouldn't the syntax be
`[ <linear-stop> ]#{2,}`?
 
https://github.com/w3c/csswg-drafts/blob/065eb9119738d0e99e1bd8cef5f4cc97a22595b2/css-easing-2/Overview.bs#L173

tabatkins:
Nah. Similar to gradients, you can supply two positions for a single stop, so a grammatical requirement like that would prevent authors from writing some otherwise reasonable things. (It would allow `linear(0 0%, 0 100%)`, but disallow the identical-meaning `linear(0 0% 100%)`.)

However, the text as currently written does fail if you provide a single stop, whether you give it two %s or not. ("create a linear easing function" explicitly fails if there is only a single stop, regardless of whether it's doubled or not; "calculate linear easing output" implicitly fails with an OOB access if there's only a single stop (not doubled).

The first failure should get fixed; there's nothing wrong it conceptually. The second should also be fixed, just to avoid grammar oddities, the same as how we just resolved to allow single-stop gradients; there's a trivial behavior for it. I'll propose some edits.

tabatkins:
Agenda+ to request the following changes:

1. Allow "create a linear easing function" to accept a single stop (making it no longer possible to fail). (So `linear(.4 0% 100%)` is valid, identical to `linear(.4 0%, .4 100%)`.
2. Fix "calculate linear easing output" to handle having a single stop (acting as a constant function at that output value). (So `linear(.4)` is valid, identical to `linear(.4 0% 100%)` in behavior, tho it'll serialize as only a single stop.)r

cdoublev:
Accepting a single stop would close #8932 (related/duplicate issue) with no change.

tabatkins:
Fixed: grammar, parsing, serialization, and output all now correctly handle a single stop.

PR #10643, opened 2024-07-30
[css-easing-2] correct the `<linear-stop-list>` syntax

OnkarRuikar:
The [parsing algo](https://drafts.csswg.org/css-easing/#linear-easing-function-parsing) for the `linear()` easing function says, "3. If there are less than two items in stopList, then return failure."

So in the current syntax `<linear-stop-list> = [ <linear-stop> ]#`, the `{2,}` would be more accurate than `#`.

### DIFF
--- a/css-easing-2/Overview.bs
+++ b/css-easing-2/Overview.bs
@@ -170,7 +170,7 @@ A [=linear easing function=] has the following syntax:
 
 <pre class="prod">
   <dfn>&lt;linear-easing-function></dfn> = <dfn function lt="linear()">linear(<<linear-stop-list>>)</dfn>
-  <dfn>&lt;linear-stop-list></dfn> = [ <<linear-stop>> ]#
+  <dfn>&lt;linear-stop-list></dfn> = [ <<linear-stop>> ]{2,}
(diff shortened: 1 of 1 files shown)


tabatkins:
The parsing algo is the broken bit here, not the grammar; it should accept a single stop with two positions. It just wasn't updated when the grammar was changed to match gradients and allow two positions on a single stop. (It should also accept a single stop with a single position, like gradients now do.)

OnkarRuikar:
@tabatkins you mean to say `linear(.3 20% 70%)` is valid but `linear(.3)` and `linear(.3 20%)` are not valid?
The `linear-gradient(red 80%)` doesn't work in FF&Chrome nightlies but `linear-gradient(red 80% 90%)` does.

tabatkins:
`linear(.3 20% 70%)` *should* be valid, yes - it's two stops. The grammar correctly accepts it, but the parsing algo doesn't, and needs fixing. 

`linear(.3)` and `linear(.3 20%)` are not currently valid per the grammar, but should be, to match with gradients. See <https://github.com/w3c/csswg-drafts/issues/10580#issuecomment-2240428017>

`linear-gradient(red 80%)` was resolved to be valid just a few weeks ago; impls haven't caught up yet.

PR #10659, opened 2024-08-01
[css-size-adjust-1] Fix inconsistency aliases

alkorlos:
Fix bikeshed errors.

The command:
```
bikeshed spec css-size-adjust-1\Overview.bs
```

Throws errors:
```
LINE ~69: The biblio refs [[CSS2]] and [[CSS21]] are both aliases of the same base reference [[CSS21]]. Please choose one name and use it consistently.
```

A search through the repositories shows that the alias [[!CSS2]] is used three times more often than [[!CSS21]], so I replaced all aliases with [[!CSS2]].

### DIFF
--- a/css-size-adjust-1/Overview.bs
+++ b/css-size-adjust-1/Overview.bs
@@ -60,8 +60,8 @@ Module interactions {#placement}
 --------------------------------
 
   This module adds additional features that are not defined in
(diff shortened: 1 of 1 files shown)


frivoal (approved):

SebastianZ:
Merged. Thanks, @alkorlos!

Sebastian

PR #10669, opened 2024-08-02
[css-lists-3] Correct the counter-set example

OnkarRuikar:
For following reasons I think [the example](https://drafts.csswg.org/css-lists/#example-838dca3e) has a typo and it should use `counter-set` instead of `counter-reset`:
- the example is part of the `Manipulating Counter Values: the counter-increment and counter-set properties` section
- `counter-reset` creates a new counter so the comment `/* Set section to 0 */` is not accurate. The `counter-set` makes the comment more accurate.


tabatkins:
Thanks, that was indeed a typo on my part. ^_^

Issue #5998, opened 2021-02-16
[css-forms-1] appearance: base to enable interoperable styling of controls/components


gregwhitworth:
The below comment is in reference to an [explainer level proposal](https://github.com/salesforce/standards-explainers/blob/master/indicator-psuedo/explainer.md) regarding a new pseudo element and a way for an author to opt-in to a standardized DOM and styles, currently being proposed is `appearance: base`.

> @gregwhitworth why not make the proposed behavior for `appearance: base` be the behavior for `appearance: none` for radio buttons and checkboxes? Would it not be web compatible? There are likely *some* sites that use `appearance: none` plus a background image to show the checkmark, so this would show a double checkmark, though I don't know how common it is. Maybe it's even an acceptable change? The content would still work, just not rendered as originally intended. When styling form controls, that shouldn't come as a huge surprise I would think.
> 
> CSS UI says that widgets should be *usable* even with `appearance: none`, but checkboxes and radio buttons as implemented today aren't. https://drafts.csswg.org/css-ui/#appearance-semantics
> 
> (Feel free to split this into a new issue if you want to separate the discussion around `appearance` from `::indicator`.)
> 
> cc @frivoal

_Originally posted by @zcorpan in https://github.com/w3c/csswg-drafts/issues/5914#issuecomment-778089292_

gregwhitworth:
@zcorpan I had not considered that given that `appearance: base` takes the step of requiring a defined DOM and styles. You're correct at the definition that the behavior should remain but without that required DOM & styles we don't take a step towards the goal of this which is to enable styling of the built-in control/component without having to replace it.

I don't have a strong opinion regarding the switch that enables that DOM & styles, @frivoal recommended over on the explainer repo a `base` attribute on the control/component that would likewise enable this and would also not conflate `appearance` which some people think will lead to author confusion.

You both have more history investigating the `appearance` property so I'll happily take guidance to ensure that the following goals are met:

1. Requires defined DOM & styles to enable an author to style the control/component
2. Doesn't introduce author confusion
3. Allows user agents & the CSSWG to incrementally enhance more controls & components without compat concerns

@fantasai brought up that very issue with the current proposal of `appearance: base` as authors may apply it to many controls and thus hindering our future additions. @tabatkins recommended a `appearance: base-checkbox` style approach to the problem and as noted @frivoal recommended a similar approach but with a HTML attribute.

cc: @tantek as he had opinions on this aspect as well

[...]

PR #10691, opened 2024-08-05
[css-ui-4] Define appearance:base

josepharhar:
This was discussed here: https://github.com/w3c/csswg-drafts/issues/5998

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -2111,7 +2111,7 @@ Switching appearance: the 'appearance' property</h3>
 
 	<pre class="propdef">
 	Name: appearance
-	Value: ''appearance/none'' | ''auto'' | <<compat-auto>> | <<compat-special>>
+	Value: ''appearance/none'' | ''auto'' | ''base'' | <<compat-auto>> | <<compat-special>>
 	Initial: none
 	Applies To: all elements
 	Inherited: no
@@ -2133,7 +2133,10 @@ Switching appearance: the 'appearance' property</h3>
 	In particular, using ''appearance: none'' allows authors
 	to suppress the <a>native appearance</a> of <a>widgets</a>,
 	giving them a <dfn export>primitive appearance</dfn>
-	where CSS can be used to restyle them.
+  where CSS can be used to restyle them. Using ''appearance: base'' also
+  suppresses the <a>native appearance</a> of <a>widgets</a> and additionally
+  enables additional styling capabilities and interoperability, giving them a
+  <dfn export>base appearance</dfn>.
 
 	<details class=note>
 		<summary>Note on the history of this feature</summary>
@@ -2198,6 +2201,18 @@ Switching appearance: the 'appearance' property</h3>
 			which elements represent which <a>widgets</a>.
 
 			Elements other than <a>widgets</a> must be rendered as for ''appearance/none''.
+
+    <dt><dfn>base</dfn>
+    <dd>
+      Just like ''appearance/none'', the element is rendered following the
+      usual rules of CSS. Replaced elements other than <a>widgets</a> are not
+      affected by this and remain replaced elements. <a>Widgets</a> must not
+      have their <a>native appearance</a>, and instead must have their <a>base
+      appearance</a>.
+
+      <a>Widgets</a> which don't have a <a>base appearance</a> must be rendered
+      as for ''appearance/auto''. The host language is responsible for defining
+      which <a>widgets</a> have a <a>base appearance</a>.
 
 		 <dt><dfn type="">&lt;compat-auto></dfn>
 		 <dd>
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -2138,10 +2138,9 @@ Switching appearance: the 'appearance' property</h3>
 	Using ''appearance: base'' gives <a>widgets</a> a <dfn export>base
 	appearance</dfn>. <a>Widgets</a> in <span>base appearance</span> have their
 	<a>native appearance</a> suppressed where CSS can be used to restyle them,
-	just like <a>primitive appearance</a>, but also have additional styling
-	capabilities and default styles which are usable and interoperable. This is
-	unlike <a>base appearance</a> which can cause some <a>widgets</a> to
-	disappear completely.
+	just like <a>primitive appearance</a>, but also have default styles which are
+	usable and interoperable. This is unlike <a>base appearance</a> which can
+	cause some <a>widgets</a> to disappear completely.
 
 	<details class=note>
 		<summary>Note on the history of this feature</summary>
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -2223,6 +2223,7 @@ Switching appearance: the 'appearance' property</h3>
 				:: <a>Widgets</a> which don't have a <a>base appearance</a> must be
 					rendered as for ''appearance/auto''. The host language is responsible
 					for defining which <a>widgets</a> have a <a>base appearance</a>.
+			</dl>
 
 		 <dt><dfn type="">&lt;compat-auto></dfn>
 		 <dd>
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -2147,9 +2147,9 @@ Switching appearance: the 'appearance' property</h3>
 	<span>Native appearance</span> and <span>primitive appearance</span> should
 	have the same styles.
 
-	ISSUE: Do not to ship ''appearance: base'' until its stylesheet has been
-	properly and thoroughly defined (and implemented per spec) for all form
-	controls.
+	ISSUE: ''appearance:base'' is not ready to ship until its stylesheet has been
+	fully defined in the HTML specification, and implemented accordingly, for all
+	HTML form controls.
 
 	<details class=note>
 		<summary>Note on the history of this feature</summary>
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -2142,6 +2142,10 @@ Switching appearance: the 'appearance' property</h3>
 	usable and interoperable. This is unlike <a>primitive appearance</a> which
 	can cause some <a>widgets</a> to disappear completely.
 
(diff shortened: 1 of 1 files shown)


css-ui-4/Overview.bs
```diff
@@ -2199,6 +2207,24 @@ Switching appearance: the 'appearance' property</h3>
 
 			Elements other than <a>widgets</a> must be rendered as for ''appearance/none''.
 
+		<dt><dfn>base</dfn>
+		<dd>
+			The effect of ''appearance/base'' depends on the element it is applied to:
+			<dl class=switch>
```
josepharhar, line 2213: Running bikeshed with this dl here, even without any further content, gives me the following error:
```
jarhar@jarhar-mac ~/csswg-drafts/css-ui-4 % pipx run bikeshed spec Overview.bs                             
LINE 2344: 'property' definitions don't use a 'for' attribute, but this one claims it's for 'appearance' (perhaps inherited from an ancestor). This is probably a markup error.
<dfn bs-line-number="2344" data-export="" data-dfn-type="property">-webkit-appearance</dfn>
FATAL ERROR: 'property' definitions don't use a 'for' attribute, but this one claims it's for 'appearance' (perhaps inherited from an ancestor). This is probably a markup error.
<dfn data-dfn-type="property">field-sizing</dfn>                                                                              
FATAL ERROR: 'property' definitions don't use a 'for' attribute, but this one claims it's for 'appearance' (perhaps inherited from an ancestor). This is probably a markup error.
<dfn data-dfn-type="property">input-security</dfn>                                                                            
```
It seems totally unrelated but doesn't happen without this dl. Am I missing something?

css-ui-4/Overview.bs
```diff
@@ -2135,6 +2135,14 @@ Switching appearance: the 'appearance' property</h3>
 	giving them a <dfn export>primitive appearance</dfn>
 	where CSS can be used to restyle them.
 
+	Using ''appearance: base'' gives <a>widgets</a> a <dfn export>base
+	appearance</dfn>. <a>Widgets</a> in <span>base appearance</span> have their
+	<a>native appearance</a> suppressed where CSS can be used to restyle them,
+	just like <a>primitive appearance</a>, but also have additional styling
+	capabilities and default styles which are usable and interoperable. This is
```
nt1m, line 2142: I would remove "additional styling capabilities", in general we should try to make as many capabilities available to `none` as well as `base`. 

`auto` is more the odd one out, where the UA can choose to not honor (parts or all) custom styling from authors, because the platform doesn't support it.

nt1m (commented):
Haven't looked into detail, but I do have one comment.

css-ui-4/Overview.bs
josepharhar: Thanks for taking a look! done

css-ui-4/Overview.bs
```diff
@@ -2135,6 +2135,13 @@ Switching appearance: the 'appearance' property</h3>
 	giving them a <dfn export>primitive appearance</dfn>
 	where CSS can be used to restyle them.
 
+	Using ''appearance: base'' gives <a>widgets</a> a <dfn export>base
+	appearance</dfn>. <a>Widgets</a> in <span>base appearance</span> have their
+	<a>native appearance</a> suppressed where CSS can be used to restyle them,
+	just like <a>primitive appearance</a>, but also have default styles which are
+	usable and interoperable. This is unlike <a>base appearance</a> which can
+	cause some <a>widgets</a> to disappear completely.
```
annevk, line 2143: Don't we have to define the style system magic here or elsewhere that makes this possible?

nt1m: I wouldn't necessarily define the specifics of it (seems like an implementation thing), but maybe just at least the end result?

annevk: I think it has to define that a widget can have some style sheet associated with it (that is defined wherever that widget is defined) and how resolving that style sheet with the existing user agent style sheets and any author and user style sheets is supposed to work.

css-ui-4/Overview.bs
```diff
@@ -2135,6 +2135,13 @@ Switching appearance: the 'appearance' property</h3>
 	giving them a <dfn export>primitive appearance</dfn>
 	where CSS can be used to restyle them.
 
+	Using ''appearance: base'' gives <a>widgets</a> a <dfn export>base
+	appearance</dfn>. <a>Widgets</a> in <span>base appearance</span> have their
+	<a>native appearance</a> suppressed where CSS can be used to restyle them,
+	just like <a>primitive appearance</a>, but also have default styles which are
+	usable and interoperable. This is unlike <a>base appearance</a> which can
```
dbaron, line 2142: I think this intends to say it's unlike the `primitive appearance` (rather than `base appearance`)

css-ui-4/Overview.bs
dbaron: I think you're missing `</dl>`.

css-ui-4/Overview.bs
```diff
@@ -2199,6 +2206,24 @@ Switching appearance: the 'appearance' property</h3>
 
 			Elements other than <a>widgets</a> must be rendered as for ''appearance/none''.
 
+		<dt><dfn>base</dfn>
+		<dd>
+			The effect of ''appearance/base'' depends on the element it is applied to:
+			<dl class=switch>
+				: Replaced elements other than <a>widgets</a>
```
dbaron, line 2213: I think you should drop the word "Replaced" here, since with that word dropped the `<dl class=switch>` becomes complete.

css-ui-4/Overview.bs
dbaron: Yeah, I think I agree that we need to define a bit more here, since (unless I'm misunderstanding the current plan) I think this is going to cause some styles to get applied that change the computed values of other properties, both on this element and likely also on observable pseudo-elements inside of it.

(On the flip side, I think the current plan is that `appearance: base` doesn't change whether any pseudo-elements *exist* (if such a thing is even detectable) since it doesn't change the UA shadow DOM at all.  (But I think it does change whether some parts of the UA shadow DOM are `display:none` or not.)  At least, I think that's the current plan, and it certainly does make things cleaner architecturally even if it requires some local ugliness in places.)

annevk: Oh, yes, the internal shadow tree plan would be good to clarify as well.

css-ui-4/Overview.bs
josepharhar: That fixed it, thanks!

css-ui-4/Overview.bs
josepharhar: So should I add a paragraph that looks like this?

> A widget can have some style sheet associated with it and those styles can change based on whether the widget has a native appearance, primitive appearance, or base appearance.

css-ui-4/Overview.bs
josepharhar: Good catch, thanks!

css-ui-4/Overview.bs
josepharhar: done

css-ui-4/Overview.bs
dbaron: It's possible that we only want the style sheet variation to occur between base appearance and the other values.  I think the difference between native appearance and primitive appearance has historically not shown up in the computed values of the properties (e.g., border) that are affected.  (Though I'm not 100% sure of this.)

fantasai:
I think this PR needs some very prominent ISSUE: text warning not to ship this feature until its stylesheet has been properly and thoroughly defined (and implemented per spec) for all form controls. We don't want an implementer reading it, making stuff up that matches the current high-level description, and shipping their own interpretation of what `appearance: base` means.

css-ui-4/Overview.bs
josepharhar: How about this?

> A widget can have some style sheet associated with it and those styles can change based on whether the widget has a base appearance. Native appearance and primitive appearance should have the same styles.

josepharhar:
> I think this PR needs some very prominent ISSUE: text warning not to ship this feature until its stylesheet has been properly and thoroughly defined (and implemented per spec) for all form controls. We don't want an implementer reading it, making stuff up that matches the current high-level description, and shipping their own interpretation of what appearance: base means.

Done

css-ui-4/Overview.bs
josepharhar: Ok, I added the text I suggested in my last comment

css-ui-4/Overview.bs
```diff
@@ -2135,6 +2135,22 @@ Switching appearance: the 'appearance' property</h3>
...
+	can cause some <a>widgets</a> to disappear completely.
+
+	A widget can have some style sheet associated with it and those styles can
+	change based on whether the widget has a <span>base appearance</span>.
+	<span>Native appearance</span> and <span>primitive appearance</span> should
+	have the same styles.
+
+	ISSUE: Do not to ship ''appearance: base'' until its stylesheet has been
+	properly and thoroughly defined (and implemented per spec) for all form
+	controls.
```
dbaron, line 2152: "Do not to ship" needs to be reworded somehow.  But I'd suggest a slightly larger change:

```
  ''appearance:base'' is not ready to ship until its stylesheet has been fully defined
  in the HTML specification, and implemented accordingly, for all HTML form controls.
```

This proposed rewording also makes it clear that (a) this is going to be defined in the HTML spec and (b) the requirement is specific to all HTML form controls (and not any other theoretical form controls).

josepharhar: Thanks, I replaced the issue text with your proposed text

css-ui-4/Overview.bs
```diff
@@ -2199,6 +2215,25 @@ Switching appearance: the 'appearance' property</h3>
 
 			Elements other than <a>widgets</a> must be rendered as for ''appearance/none''.
 
+		<dt><dfn>base</dfn>
+		<dd>
+			The effect of ''appearance/base'' depends on the element it is applied to:
+			<dl class=switch>
+				: elements other than <a>widgets</a>
+				:: Non-<a>widget</a> replaced elements are not affected by this and
+					remain replaced elements.
```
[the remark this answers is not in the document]
josepharhar: done

dbaron (approved):

Issue #10017, opened 2024-02-29
[css-values] specification for calc() should be clearer about when the result has a percentage


dbaron:
There are a number of cases where layout algorithms in CSS (and perhaps other things) care about whether a value has a percentage.  In particular, values that have a percentage are treated differently when there's nothing for the percentage to resolve against:  inside of something with `auto` height a `height: calc(30px)` is a fixed value but a `height: calc(30px + 0%)` is treated as `height: auto`.

With `calc()` as it was specified in css-values-3, this could essentially be implemented as part of the type computation; a `calc()` expression could effectively be either a `<length>` or a `<length-percentage>` as its toplevel type, and those that were `<length-percentage>` act as though they have a percent.

Newer levels of the specification introduce features that make this approach insufficient:
* division by values with units
* [`sign()`](https://drafts.csswg.org/css-values-4/#calc-type-checking)
* [`progress()` and related functions](https://drafts.csswg.org/css-values-5/#progress)
* [`calc-size()`](https://drafts.csswg.org/css-values-5/#calc-size)

(I think this is clearly defined for `calc-size()` because it's very important there, but I don't think it's clearly defined for the other cases.)

It should be clearer whether these other things "erase" the percentage-ness of their arguments when they erase the types of those arguments, or whether they still produce toplevel values that are treated as having a percentage.

tabatkins:
Unit division is actually well-defined, by reference to [Typed OM's concept of "matching"](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-match). Something matches `<length>`/etc only if it has `{length -> 1}` *and* its percent hint is null. I suppose technically it's only by implication that a non-null percent hint means it contains a percentage, tho, so we could probably put in a definition for that.

In-flight edit: actually it looks like I am invoking the types wrong for cases like min(); there I just say the args must have "the same type" and define that the function resolves to that type. This doesn't handle percent hints properly. I'll need to tweak it.

---------

sign() does indeed currently erase the percentage information, tho, as does anything that returns a value of a different type than its input calculations. We should think about whether it *should* transfer that information thru - doing so via the calculation type (copying the percent hint of the input calculation) would mean it doesn't match `<number>` anymore.

----------

The progress() family just need to define their type as the sum of their input calculation's types, same as min()/etc. (Subject to the changes I just discovered I need to make.)

-------

calc-size() makes an end-run around the issue, since it defines that it simply acts like its basis argument in all ways.

[...]

PR #10727, opened 2024-08-12
[css-values-4] Math functions also match mixed types

cdoublev:
> A math function resolves to `<number>`, `<length>`, `<angle>`, `<time>`, `<frequency>`, `<resolution>`, `<flex>`, or `<percentage>` [...] **If it can’t match any of these, the math function is invalid.**

From [CSS Typed OM](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-match):

  > A type matches `<length-percentage>` if its only non-zero entry is either `«[ "length" → 1 ]»` or `«[ "percent" → 1 ]»`.

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -4458,6 +4458,7 @@ Type Checking</h3>
 	according to which of those productions its [=CSSNumericValue/type=] [=CSSNumericValue/matches=].
 	(These categories are mutually exclusive.)
 	If it can't [=CSSNumericValue/match=] any of these,
+	or the <<length-percentage>>/etc mixed types,
 	the [=math function=] is invalid.
 
 	Note: Algebraic simplifications do not affect the validity of a [=math function=] or its resolved type.


tabatkins:
Any `<length>` values are valid in a `<length-percentage>` grammar. We don't need to explicitly indicate that everything that's a length is *also* a length-percentage; that's part of the *definition* of length-percentage.

cdoublev:
My point was that this text almost literaly says that if a math function match `<length-percentage>` but not `<length>`, which happens when it has a percent hint, it is invalid. (In this text, *"matches"* to the type checking algorithm in CSS Typed OM.)

I thought it was an oversight, not something obvious that did not need to be explicited.

tabatkins:
The current text is right. Take a calculation like `10px + 50%`, in a context that resolves %s against a length, like the 'width' property:

* per [determine the type of a calculation](https://drafts.csswg.org/css-values-4/#determine-the-type-of-a-calculation), the type is `{length -> 1}`, with a percent hint of "length".
* per [match a type](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-match), a type matches "length" if it's only non-zero entry is for "length" (true), and since the context it's used in allows percentages, the percent hint isn't consulted.

The "match a type" algo has text for matching `<length-percentage>`/etc that *additionally* allows a `{percent -> 1}` type, tho I think that might now be legacy code and effectively unreachable. I should probably just more explicitly fold it into the plain `<length>` section, as that "If the context in which the value is used does not allow `<percentage>`" text is basically talking about that case (or rather, the inverse).

cdoublev:
Ah yes, this branch is obsolete. You cannot get `[ percent: 1 ]` if the percentage resolves to another type.

cdoublev:
I am not very familiar with all the features defined by CSS Typed OM: is it possible to *successfully* add a percentage and a numeric value of another type, if the percentage value does not resolve to a value of this type? Or is the second branch of step 3 of [add two types](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-add-two-types) also obsolete? (this has also been [discussed](https://github.com/w3c/csswg-drafts/issues/10017#issuecomment-1979143330) a bit).

tabatkins:
That algo's branch is still valid - when you're doing operations on plain CSSNumericValue objects, you don't yet know the context they'll be used in, so you can't tell whether a % will become a length/etc. So that algo handles unifying types correctly when this is the case - so long as your % doesn't end up resolving against *two* different types, it's fine to add it to other dimensions.

Issue #10693, opened 2024-08-05
[selectors][css-scoping] Should `:host:has()` match?


tabatkins:
In <https://github.com/web-platform-tests/wpt/pull/47441> a few tentative tests are suggested that test whether `:host:has(...)` works. The test author correctly notes that this isn't currently specified to work - only the [logical combination pseudo-classes are allowed to match featureless elements](https://drafts.csswg.org/selectors/#featureless) and `:has()`, while defined in the "logical combination" section, isn't on the list of logical combo pseudo-classes.

I agree with the tester, tho, that this *should* work. `:host h1` works just fine, so `:host:has(h1)` should work the same. I suggest that we add `:has()` to the list of pseudo-classes allowed to work on featureless elements.

(Note the first few tests in the WPT PR are testing the different scenario of `:host(:has(...))`, which *is* already well-defined to match, in a different way to what's discussed here. I'm talking about the latter three which are marked "tentative" in their filenames.)

Westbrook:
Thanks for bringing the conversation here @tabatkins! I think the `:host:has(…)` combination will also be a boon to the in progress `slot:has-slotted(…)` API that is pending in https://github.com/w3c/csswg-drafts/pull/10586, and will be quite powerful as an appropriate customization of`:host:has(slot:has-slotted)`. 🥹

Loirooriol:
To clarify, are you saying that both `:host:has(h1)` and `:has(h1)` should be able to match the host, or only the former?

In https://github.com/w3c/csswg-drafts/issues/10179#issuecomment-2093869789 the idea is

> Compound selectors are allowed to match the host only if all the contained simple selectors are allowed to match the host.

tabatkins:
I think it should match in both cases, just like the logical combination pseudos. The reason the host is featureless is because you can't control the markup of the host element, and so it's easy to write a selector you expect to only match the markup you control (in the shadow) and have it accidentally match the host element due to the outer page doing things you can't predict.

This doesn't apply to `:has()` - it's matching based on markup you *do* control (in the shadow), and thus is completely predictable. Even if you use a complex selector in the `:has()` argument, because it's matching inside the shadow and uses that matching context, it can't match the host element unexpectedly.

[...]

PR #10749, opened 2024-08-16
[selectors-4] Add :has() to the list of logical combination pseudo-classes

kbabbitt:
As discussed in #10693.

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -1043,7 +1043,7 @@ Logical Combinations</h2>
 	[=compound selector|compounding=] (logical AND),
 	[=selector lists=] (logical OR),
 	and the <dfn>logical combination pseudo-classes</dfn>
-	'':is()'', '':where()'', and '':not()''.
+	'':is()'', '':where()'', '':not()'', and '':has()''.
 	The [=logical combination pseudo-classes=]
 	are allowed anywhere that any other [=pseudo-classes=] are allowed,
 	but pass any restrictions to their arguments.


Westbrook (approved):

cdoublev:
I think pseudo-elements cannot be followed by `:has()`.

  > This specification allows any pseudo-element to be followed by any combination of the logical combination pseudo-classes and the user action pseudo-classes.

There is probably a good reason for `:has()` to be in the "Logical Combinations" section but I see it as a tree-structural pseudo-class.

kbabbitt:
> I think pseudo-elements cannot be followed by `:has()`.
> 
> > This specification allows any pseudo-element to be followed by any combination of the logical combination pseudo-classes and the user action pseudo-classes.

Good point, this is an unintended consequence of the proposed edit I hadn't realized. `::first-line:has(...)` seems like a reasonable thing to express but could lead to dependency cycles so I don't think we want to allow it. And I don't see how `::before`, `::after`, or `::first-letter` could possibly match `:has()`, so there's no reason to allow those either.

> There is probably a good reason for `:has()` to be in the "Logical Combinations" section but I see it as a tree-structural pseudo-class.

I can see both sides of this. On one hand, because `:has()` takes a `<relative-selector-list>` there's an implied OR between the selectors in that list. On the other hand, unlike `:is()`, `:where()` and `:not()`, `:has()` requires examining the tree beyond the subject of the selector. @tabatkins / @fantasai wdyt?

tabatkins:
It's not really a logical combination pseudo, which is why we didn't include it in that list in the first place. Logical combination pseudos don't do anything at all, theoretically, that you couldn't achieve by just writing out the selectors manually. (Well, `:not()` is a new ability, but it's such a clear mirror to what its argument can do that it's *morally* not new.) So allowing them, with the caveat that you can only use arguments that could already show up there by themselves, doesn't raise any questions.

`:has()` *is* totally new functionality, and can't be treated as cavalierly. Editing for #10693 needs to be a bit more fiddly, unfortunately.

PR #10750, opened 2024-08-17
[selectors-4] Fix bikeshed warnings

kbabbitt:
- Add link-defaults for two definitions
- Fix numerous "Ambiguous for-less link" warnings
- Convert spaces to tabs on line 1980
- Tag var 'argument' with 'ignore' where it's used in a Note

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -37,11 +37,12 @@ spec:css-pseudo-4; type:selector;
 	text: ::after
 	text: ::first-line
 	text: ::first-letter
+spec:dom; type:dfn; text:descendant
+spec:html; type:dfn; text:muted
 spec:html; type:element-attr; for:a; text:href
 spec:html; type:element;
 	text:meta
 	text:audio
-
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -585,16 +586,16 @@ Structure and Terminology</h3>
...
 
 	A <dfn id="complex" export>complex selector</dfn> is
 	a sequence of one or more <a>compound selectors</a>
 	and/or [=pseudo-compound selectors=],
-	with [=compound selectors=] separated by <a>combinators</a>.
+	with [=compound selectors=] separated by <a for=selector>combinators</a>.
 	It represents a set of simultaneous conditions
 	on a set of elements in the particular relationships
-	described by its <a>combinators</a>.
+	described by its <a for=selector>combinators</a>.
```
kbabbitt, line 598: This edit fixes the warning, but for some reason it's still not link-ifying, and I'm not sure why.

kbabbitt:
@tabatkins  / @fantasai friendly ping on this PR. Any concerns with the changes, or any idea why the edit I commented on might not be link-ifying properly?

selectors-4/Overview.bs
tabatkins: Oh, it works just fine, it's just that, in order to reduce visual clutter in heavily linked text, Bikeshed suppresses instances of a given autolink past the first in a given paragraph. (We tried to just visually reduce the contrast, but that was nixed by a11y people, so instead of "this link exists, but it's intentionally a little hard to distinguish from surrounding text" we just have "this link doesn't exist at all" now. :( )

tabatkins (approved):

Issue #116, opened 2016-05-16
[css-backgrounds-4] Add background-repeat-x/y


mstange:
It looks like recent versions of Edge and Chrome support the properties background-repeat-x and background-repeat-y. I haven't checked how far back support goes. Gecko does not support them, and Webkit doesn't appear to either.
If this is something people want, then it should be added to the spec.

yisibl:
Agree, I think it should be consistent with background-position-x/y.

fantasai:
What's the use case for cascading them independently?

SebastianZ:
Backgrounds Level 4 already [defines those two properties plus `background-position-inline` and `background-position-block`](https://drafts.csswg.org/css-backgrounds-4/#background-position-longhands). So, I'm wondering why this issue was opened.

> Gecko does not support them, …

That info is outdated. They were [implemented in Gecko 49](https://bugzilla.mozilla.org/show_bug.cgi?id=550426).

> … and Webkit doesn't appear to either.

I don't have the means to test it, but the related MDN articles suggest that WebKit does support [`background-position-x`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position-x) and [`background-position-y`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position-y). Could someone please verify that and update the articles accordingly (or let me know about it, so I can do that)?

Sebastian

mstange:
Sebastian, this is about background-*repeat*, not background-position. (Also, I'm the person who implemented background-position-x/y in Gecko.)

Fantasai, I don't have a use case. Maybe it's just about consistency with background-position? To me, background-repeat "feels" like a shorthand property because it allows you to specify different values for each axis.

fantasai:
If there's no use case for cascading them independently, then we should not add longhands. And Chrome and Edge should remove them.

tabatkins:
[Our usage numbers](https://www.chromestatus.com/metrics/css/popularity#background-repeat-y) show it at .05%, above our standard removal threshold.

SebastianZ:
> Sebastian, this is about background-_repeat_, not background-position. (Also, I'm the person who implemented background-position-x/y in Gecko.)

Oops! Please forgive me my blindness! I saw you've also already filed [bug 1273244](https://bugzilla.mozilla.org/show_bug.cgi?id=1273244) for it.

Sebastian

dbaron:
[...]

PR #10755, opened 2024-08-18
[css-backgrounds-4] Added repeat-block and `repeat-inline` keywords for `background-repeat`

SebastianZ:
This adds the logical values `repeat-block` and `repeat-inline` to `background-repeat` as [resolved in #116](https://github.com/w3c/csswg-drafts/issues/116#issuecomment-1854430181).

Sebastian

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -143,7 +143,7 @@ Tiling Images Shorthand: the 'background-repeat' property</h3>
 	<p>This shorthand sets the values for the
 	'background-repeat-x' and 'background-repeat-y' longhand properties.
 	Where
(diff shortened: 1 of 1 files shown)


css-backgrounds-4/Overview.bs
```diff
@@ -479,6 +479,14 @@ Tiling Images Shorthand: the 'background-repeat' property</h3>
 	<dd>
 	Computes to ''no-repeat repeat''.
 
+	<dt><dfn>repeat-block</dfn>
+	<dd>
+	Computes to ''repeat'' in [=block flow direction|block direction=] and ''no-repeat'' in [=inline base direction|inline direction=].
```
fantasai: "in the [=block axis=]" and "in the [=inline axis=]" should work here. Otherwise lgtm, feel free to merge after fixing.

PR #10767, opened 2024-08-22
[css-view-transitions] Rename UpdateCallback to something more specific.

emilio:
The WebIDL namespace is global, it's a bit better if we call this something a bit more view-transitions-specific so that other specs don't rely on it, since UpdateCallback is otherwise very generic.

I don't think this would be a normative change since the name of a callback isn't exposed anywhere author visible.

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -171,14 +171,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 	A successful [=view transition=] goes through the following phases:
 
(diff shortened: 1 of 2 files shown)


emilio:
cc @nt1m in case there's any objection :)

nt1m (approved):
I think this is fine:
* WebKit already names this `ViewTransitionUpdateCallback`: https://searchfox.org/wubkat/source/Source/WebCore/dom/ViewTransitionUpdateCallback.idl
* Chrome names this `ViewTransitionCallback` : https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/view_transition/view_transition_callback.idl

noamr (approved):
LGTM

PR #10777, opened 2024-08-24
[css-images-4] Clarified that total width is length of paint line

### LINKED ISSUE #8160: [css-images] 1D "painting area" for stripes() is confusing
`stripes()` keeps referring to a "painting area", e.g.

> a [`<percentage>`](https://drafts.csswg.org/css-values-4/#percentage-value) is relative to the painting area

But this is very confusing since an area is 2D but `stripes()` is a `<1d-image>`. And for example,

```css
width: 0;
height: 0;
border: 50px solid;
border-radius: 100%;
border-color: stripes(blue 50%, orange 50%);
```

I think it would look like the image below, so the actual areas are in a proportion of 1:4, not 1:1

I think "painting area" should be renamed to something like "length of the paint line".

SebastianZ:
Fixes #8160


SebastianZ:
This is lying around for more than a year now.
@fantasai @tabatkins @LeaVerou Could one of you please have a quick look whether the change is fine?

Sebastian

SebastianZ:
@fantasai @tabatkins @LeaVerou Friendly reminder that this is waiting for review. Ignoring the whitespace changes, this is really just a single-sentence change. So it should be a quick one.

Sebastian

PR #10783, opened 2024-08-26
[css-values-5] Require only the name of the media feature

cdoublev:
[`media-progress()`](https://drafts.csswg.org/css-values-5/#media-progress-func) currently wants a full (range) [media feature](https://drafts.csswg.org/mediaqueries-4/#typedef-media-feature). For example: `1px < width`.

But I think `media-progress(width from 0px to 2px)` is expected, ie. the media feature *name*.


cdoublev:
Either `container-progress()` should also take `<mf-name>`, or I am missing how these functions are supposed to work with full media/size feature.

cdoublev:
I guess `<media-feature>` is implicitly restricted to `<mf-boolean>` by *"The specified media query must be a valid “range” type query"*. Sorry.

Issue #6402, opened 2021-06-21
[css-contain-3] Define a syntax for state-based container queries


mirisuzanne:
In the discussion of [what container features can be queried](https://github.com/w3c/csswg-drafts/issues/5989), there was interest in queries based on the "current state" of a container – such as whether a `position: sticky` container is currently "stuck" (displaced from its in-flow position), or whether the box is currently visible on screen.

In the draft of [CSS Containment L3](https://drafts.csswg.org/css-contain-3/#container-type) we currently allow a `state` value for `container-type`, but different states may require different types of containment (or none at all), so I don't think that single type value will work. We also need to define which states are supported initially, and how these states are queried in the `@container` rule.

For example, something along the lines of:

```css
header {
  container: is-stuck is-visible / header;
  position: sticky;
  top: 0;
}

@container header (is-stuck) { /* … */ }
@container header (is-visible) { /* … */ }
```

fantasai:
Agenda+ to defer this to L4

matthew-dean:
@mirisuzanne 

I don't know how to formulate this idea, but it would be nice to query state / behavior from things like Grid / Flexbox, particularly around things like:

- Is this flexbox wrapped?
- Which "row" of a flexbox am I in?
- Which row or column of a Grid am I in?

The reason why I'm thinking about this is that there are tricks in flex-basis that people are using right now that create kind of an equivalent of container queries. The problem with something like flex-basis and flex-wrap is that you have a layout behavior applied at an unknown point / state. So you can't exactly use a container query to "capture" when the state / layout of that box / container changes.

So, like, there's a Switcher layout in the Every Layout library that, when content size is below a certain width, using flex-basis magic, it will change a 3-column layout to a single column layout w/ 3 items. It would be nice to query when this happens, because the Flexbox doesn't "know" when that flex layout happens, and we're now in a single column.

Because it's not based on a single property, and not necessarily a single "width", I really don't know how you would query it? But I just thought I'd put it out there. Here's a Codepen showing a Switcher with neither container nor media queries. https://codepen.io/matthewdean/pen/bGaKwRx

matthew-dean:
I wonder.... if there's something like a syntax missing like a way to "push" state in CSS based on something... happening? I think it should be more like a way to push a state like `:active` or `:focused`, such that the syntax is more like:
```
header:sticky {
  background: red;
}
```
[...]

Issue #10796, opened 2024-08-28
[css-conditional][css-anchor-position][scroll-animations] Snapshotting post-layout State


lilles:
Multiple CSS specifications now rely on post layout state, like scroll state, for doing style and layout:

- Scroll/View progress timelines are animation timelines based on scroll progress of a scroll container
- Anchor positioning does scroll adjustments to anchored element to align with anchor, and scroll progress can affect style and layout through choosing different fallbacks: https://drafts.csswg.org/css-anchor-position-1/#fallback-apply
- Scroll-state container queries apply style directly or indirectly based on scroll progress of a scroll container:
  - Whether a sticky positioned have offsets applied in a given direction
  - Whether an element is scroll snapped in a given direction
  - Whether an element has scrollable overflow in a given direction

The current state on the spec and implementation side:

- Snapshot timing is not specified for anchor positioning (see [issue](https://drafts.csswg.org/css-anchor-position-1/#issue-5446f06a)), but the Chromium implementation uses the same timing as for the scroll animations below.
- Scroll animations snapshotting is [specified](https://drafts.csswg.org/scroll-animations/#event-loop), but the HTML spec references are [out-of-date](https://github.com/w3c/csswg-drafts/issues/10769), and the “stale timelines” concept does not exist in the Chromium implementation. A second update can happen, even for non-stale timelines, in the Chromium implementation:
  - Snapshotting is specified to happen at the beginning of [update animations and send events](https://drafts.csswg.org/web-animations-1/#update-animations-and-send-events)
  - A second update of the timelines happens (for stale timelines per spec) after the first style and layout update in the resizeObserver loop of the HTML spec. There seems to be a discrepancy between the spec and the implementation in that the spec says the update should happen after the resizeObserver loop, but not very clear.
- For scroll-state() queries, there is a prototype implementation where the snapshotting happens at the same time as the former two.

The two-stage snapshotting happens to avoid a first rendering where there is one frame without a snapshot, since there is no previous frame, with a second frame that applies the state from the first frame, leading to flicker for the first rendering.

Instead of having multiple specs describing when this snapshotting happens, and how it interacts with the HTML spec, we could introduce common spec steps for doing post-layout state snapshots.

- Introduce a new “run snapshot post-layout state steps” to CSSOM View spec. Say that this is where the post-layout state used for style and layout is snapshotted for any specs that need to. Leave for each spec to say what they snapshot as part of “run snapshot post-layout state steps”.
[...]

PR #10798, opened 2024-08-28
[css-conditional-5] Draft spec for scroll-state() #6402

lilles:
Per resolution in [1], add scroll-state() as a query function to query sticky, snap, and overflow scroll states with a new container-type for scroll-state queries.

[1] https://github.com/w3c/csswg-drafts/issues/6402#issuecomment-1812973013

### DIFF
--- a/css-conditional-5/Overview.bs
+++ b/css-conditional-5/Overview.bs
@@ -460,7 +460,7 @@ Creating Query Containers: the 'container-type' property</h3>
 
 	<pre class='propdef'>
 		Name: container-type
-		Value: normal | size | inline-size
+		Value: normal | [ [ size | inline-size ] || scroll-state ]
 		Initial: normal
 		Inherited: no
 		Applies to: all elements
@@ -468,20 +468,12 @@ Creating Query Containers: the 'container-type' property</h3>
 		Animation type: not animatable
 	</pre>
 
-	The 'container-type' property establishes the element
-	as a <dfn export>query container</dfn>
-	for the purpose of [=container queries=] that require explicit containment
-	(such as [=container size queries=]),
-	allowing [=style rules=] styling its descendants
-	to query various aspects of its sizing and layout,
-	and respond accordingly.
-
-	Unless otherwise noted,
-	all elements are [=query containers=]
-	for the purpose of [=container queries=]
-	that do not require explicit [=containment=]
-	(such as [=container style queries=]),
-	regardless of the specified 'container-type'.
+	The 'container-type' property establishes the element as a
+	<dfn export>query container</dfn> for certain types of queries. For size
+	[=container queries=], that require certain types of containment, elements
+	are explicitly made [=query containers=] through this property. For other
+	types of [=query containers=] any element can be a query container, such as
+	for [=container style queries=].
 
 	Values have the following meanings:
 
@@ -502,10 +494,13 @@ Creating Query Containers: the 'container-type' property</h3>
 			[=style containment=],
 			and [=inline-size containment=]
 			to the [=principal box=].
+		<dt><dfn>scroll-state</dfn>
+		<dd>
+			Establishes a [=query container=] for [=scroll-state queries=]
 		<dt><dfn>normal</dfn>
 		<dd>
 			The element is not a [=query container=]
-			for any [=container size queries=],
+			for any [=container size queries=] or [=scroll-state queries=],
 			but remains a [=query container=] for [=container style queries=].
 	</dl>
 
@@ -546,6 +541,23 @@ Creating Query Containers: the 'container-type' property</h3>
 		</pre>
 	</div>
 
+	<div class=example>
+		Containers can also expose state that depend on scroll offset. Here is an
+		example of how to style a descendant of a sticky positioned element when it
+		is stuck to the top edge:
+
+		<pre class=lang-css>
+		#sticky {
+			container-type: scroll-state;
+			position: sticky;
+		}
+		@container scroll-state(stuck: top) {
+		  #sticky-child {
+				background-color: lime;
+		  }
+		}
+		</pre>
+	</div>
 
 <h3 id="container-name">
 Naming Query Containers: the 'container-name' property</h3>
@@ -657,19 +669,27 @@ Container Queries: the ''@container'' rule</h3>
 	<pre class="prod def">
 	<dfn><<container-condition>></dfn> = [ <<container-name>> ]? <<container-query>>
 	<dfn><<container-name>></dfn> = <<custom-ident>>
-	<dfn><<container-query>></dfn>     = not <<query-in-parens>>
-	                      | <<query-in-parens>> [ [ and <<query-in-parens>> ]* | [ or <<query-in-parens>> ]* ]
-	<dfn><<query-in-parens>></dfn>     = ( <<container-query>> )
-	                      | ( <<size-feature>> )
-	                      | style( <<style-query>> )
-	                      | <<general-enclosed>>
+	<dfn><<container-query>></dfn> = not <<query-in-parens>>
+	                  | <<query-in-parens>> [ [ and <<query-in-parens>> ]* | [ or <<query-in-parens>> ]* ]
+	<dfn><<query-in-parens>></dfn> = ( <<container-query>> )
+	                  | ( <<size-feature>> )
+	                  | style( <<style-query>> )
+	                  | scroll-state( <<scroll-state-query>> )
+	                  | <<general-enclosed>>
 
-	<dfn><<style-query>></dfn>         = not <<style-in-parens>>
-	                      | <<style-in-parens>> [ [ and <<style-in-parens>> ]* | [ or <<style-in-parens>> ]* ]
-	                      | <<style-feature>>
-	<dfn><<style-in-parens>></dfn>     = ( <<style-query>> )
-	                      | ( <<style-feature>> )
-	                      | <<general-enclosed>>
+	<dfn><<style-query>></dfn>     = not <<style-in-parens>>
+	                  | <<style-in-parens>> [ [ and <<style-in-parens>> ]* | [ or <<style-in-parens>> ]* ]
+	                  | <<style-feature>>
+	<dfn><<style-in-parens>></dfn> = ( <<style-query>> )
+	                  | ( <<style-feature>> )
+	                  | <<general-enclosed>>
+
+	<dfn><<scroll-state-query>></dfn>     = not <<scroll-state-in-parens>>
+	                         | <<scroll-state-in-parens>> [ [ and <<scroll-state-in-parens>> ]* | [ or <<scroll-state-in-parens>> ]* ]
+	                         | <<scroll-state-feature>>
+	<dfn><<scroll-state-in-parens>></dfn> = ( <<scroll-state-query>> )
+	                         | ( <<scroll-state-feature>> )
+	                         | <<general-enclosed>>
 	</pre>
 
 	The keywords ''container-name/none'', ''and'', ''not'', and ''or''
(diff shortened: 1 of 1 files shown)


css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+	which may lead to scroll state changes as a result of layout. The same issue exists for
+	[=scroll progress timelines=], and scroll state is handled in a similar manner.
+
+	To avoid such layout cycles, ''scroll-state'' [=query containers=] update their
+	current state once as the last step of [=run the scroll steps=]. Then, after the
+	resizeObserver loop in the
+	<a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model">HTML event loop processing model</a>,
+	the scroll state of every ''scroll-state'' [=query container=] is updated.
+	If that state has changed since the scroll state update in [=run the scroll steps=],
+	re-run the style and layout update if necessary.
```
dbaron, line 1115: I find this a bit confusing.  Is the "current state" mentioned in the first sentence the same as the "scroll state" in the second sentence?

In the second sentence, avoid the passive voice by saying "update the scroll state" rather than "the scroll state ... is updated".

It's also not clear to me how this prevents cycles.  It sounds like this is saying to re-run an arbitrary number of times.  Or does it mean to re-run just once and then stop?

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+		Value: top | right | bottom | left | block-start | inline-start | block-end | inline-end
+		Type: discrete
+	</pre>
+
+	The '@container/stuck' [=container feature=] queries whether a
+	''position/sticky'' positioned container is visually shifted to stay inside
+	the [=sticky view rectangle=] for the given edge. The logical edges map to
+	physical based on the direction and writing-mode of the [=query container=].
+
+	In the boolean context, the query matches if visual shift is applied in any
```
dbaron, line 1132: I assume that if the query container is not `position: sticky` then none of the values match?  It would be good to state that explicitly.  (It might make sense to have a bit less formal definition for each value in the `<dl>` since they have a lot in common.)

I also think that "visually shifted" should perhaps link to the definition in css-position-3.

It might be worth noting that a container can be stuck to two opposite sides at the same time.  (At least, after reading the text in css-position-3, I think that's the case.)

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+
+	The '@container/snapped' [=container feature=] queries whether a [=snap target=]
+	is snapped to its [=snap container=] in the given axis. It matches in the boolean
+	context if it is snapped in at least one of the directions.
+
+	<dl dfn-type=value dfn-for="@container/snapped">
+		<dt><dfn>x</dfn>
+		<dd>
+			'@container/snapped' [=container feature=] matches ''x''
+			if the container is a horizontal [=snap target=] for its [=scroll container=]
```
dbaron, line 1180: I think "the container" is "the query container", right?

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+			in the block direction of the [=snap container=].
+		<dt><dfn>inline</dfn>
+		<dd>
+			'@container/snapped' [=container feature=] matches ''inline''
+			if the container is a [=snap target=] for its [=scroll container=]
+			in the inline direction of the [=snap container=].
+	</dl>
+
+<h4 id="overflowing">
+Sticky positioning: the '@container/overflowing' feature</h4>
```
dbaron, line 1198: I don't think "Sticky positioning" is the right header here.

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+			The ''position/sticky'' container is shifted to stay inside the right edge.
+		<dt><dfn>bottom</dfn>
+		<dd>
+			The ''position/sticky'' container is shifted to stay inside the bottom edge.
+		<dt><dfn>left</dfn>
+		<dd>
+			The ''position/sticky'' container is shifted to stay inside the left edge.
+		<dt><dfn>block-start</dfn>
+		<dd>
+			The ''position/sticky'' container is shifted to stay inside the block-start edge.
```
dbaron, line 1150: link the logical values to their [definitions](https://drafts.csswg.org/css-writing-modes-4/#logical-directions)?

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+			The [=scroll container=] has [=scrollable overflow=] past the right edge.
+		<dt><dfn>bottom</dfn>
+		<dd>
+			The [=scroll container=] has [=scrollable overflow=] past the bottom edge.
+		<dt><dfn>left</dfn>
+		<dd>
+			The [=scroll container=] has [=scrollable overflow=] past the left edge.
+		<dt><dfn>block-start</dfn>
+		<dd>
+			The [=scroll container=] has [=scrollable overflow=] past the block-start edge.
```
dbaron, line 1230: link the logical values to their [definitions](https://drafts.csswg.org/css-writing-modes-4/#logical-directions)?

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+
+	<pre class="descdef mq">
+		Name: overflowing
+		For: @container
+		Value: top | right | bottom | left | block-start | inline-start | block-end | inline-end
+		Type: discrete
+	</pre>
+
+	The '@container/overflowing' [=container feature=] queries whether a
+	[=scroll container=] has [=scrollable overflow=] in the given direction for the
```
dbaron, line 1208: I think it should be clear that none of these match if the query container isn't a scroll container.

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
...
+	current scroll progress. The logical values map to physical based on the direction
+	and writing-mode of the [=query container=].
+
+	In the boolean context, the query matches if there is scrollable overflow in any
+	direction.
+
+	<dl dfn-type=value dfn-for="@container/overflowing">
+		<dt><dfn>top</dfn>
+		<dd>
```
dbaron, line 1217: I think it should be clear if these match when there is overflow even if the container is not scrollable in that direction.

I also wonder if there's demand for both variants, that is, an `(overflowing:left)` that is true only for RTL or vertical-rl pages where you can actually scroll to the left, and a differently named `(overflowing: left)` that is true if there is overflow to the left whether or not it can be scrolled to.

(I don't think it's worth distinguishing based on `overflow: hidden`... particularly since that's *programatically* scrollable.  But if things do depend on "is it scrollable in that direction" it should be clear whether or not `hidden` makes a difference.)

dbaron (commented):
This looks reasonable.  One big question at the top (about whether it's supposed to be in this file at all), and then a bunch of specific comments on the text.

css-conditional-5/Overview.bs
```diff
@@ -460,28 +460,20 @@ Creating Query Containers: the 'container-type' property</h3>
 
 	<pre class='propdef'>
 		Name: container-type
-		Value: normal | size | inline-size
+		Value: normal | [ [ size | inline-size ] || scroll-state ]
```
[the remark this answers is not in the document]
lilles: > One high-level question here is whether these edits should go in css-conditional-5. The resolution explicitly called for a delta spec. (I realized the base spec has moved from css-contain-3 to css-conditional-5, but I think not doing a delta spec is still surprising.)
> 
> If this gets turned into a delta spec... then some of my later comments probably don't make as much sense.

I was unsure and asked Tab, who proposed adding it to this level (but he's not editing this spec).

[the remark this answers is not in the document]
lilles: > > One high-level question here is whether these edits should go in css-conditional-5. The resolution explicitly called for a delta spec. (I realized the base spec has moved from css-contain-3 to css-conditional-5, but I think not doing a delta spec is still surprising.)
> > If this gets turned into a delta spec... then some of my later comments probably don't make as much sense.
> 
> I was unsure and asked Tab, who proposed adding it to this level (but he's not editing this spec).

Adding @fantasai for another opinion. Should scroll-state() queries be added to a css-conditional-6 delta spec?

css-conditional-5/Overview.bs
lilles: > I assume that if the query container is not `position: sticky` then none of the values match? It would be good to state that explicitly. (It might make sense to have a bit less formal definition for each value in the `<dl>` since they have a lot in common.)

Done.

> I also think that "visually shifted" should perhaps link to the definition in css-position-3.

"visually shifted" does not have a dfn in that spec.

> It might be worth noting that a container can be stuck to two opposite sides at the same time. (At least, after reading the text in css-position-3, I think that's the case.)

Done.

css-conditional-5/Overview.bs
lilles: > I think it should be clear if these match when there is overflow even if the container is not scrollable in that direction.
> 
> I also wonder if there's demand for both variants, that is, an `(overflowing:left)` that is true only for RTL or vertical-rl pages where you can actually scroll to the left, and a differently named `(overflowing: left)` that is true if there is overflow to the left whether or not it can be scrolled to.

The intention is that matching is based on the current scroll position, so `left` and `right` will both match when a horizontal scrollbar is not a 0 nor at the max position. If there is content overflowing to the left of an ltr container, the content is not reachable through user scrolling, and `(overflowing:left)` will not match.

> (I don't think it's worth distinguishing based on `overflow: hidden`... particularly since that's _programatically_ scrollable. But if things do depend on "is it scrollable in that direction" it should be clear whether or not `hidden` makes a difference.)

I used the term [_scrollable overflow_](https://drafts.csswg.org/css-overflow-3/#scrollable-overflow) to exclude `overflow:hidden` and overflow that's not reachable through user scrolling. I've made it more explicit in the text now.

css-conditional-5/Overview.bs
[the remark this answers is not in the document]
svgeesus: I think it is fine to put this in Conditional 5, which is already a delta over Conditional 3 and 4.

svgeesus (approved):
Overall approve, as I prefer to see good PR merged sooner and details worked out later. But I will wait for one of the other editors to do the merge. Not all of the @dbaron comments have been resolved.

css-conditional-5/Overview.bs
lilles: > I find this a bit confusing. Is the "current state" mentioned in the first sentence the same as the "scroll state" in the second sentence?

Changed the latter "scroll state" to "current state" which should be clearer.

> In the second sentence, avoid the passive voice by saying "update the scroll state" rather than "the scroll state ... is updated".

Done.

> It's also not clear to me how this prevents cycles. It sounds like this is saying to re-run an arbitrary number of times. Or does it mean to re-run just once and then stop?

Just a single time. Clarified.

css-conditional-5/Overview.bs
```diff
@@ -460,28 +460,20 @@ Creating Query Containers: the 'container-type' property</h3>
...
 
-	The 'container-type' property establishes the element
-	as a <dfn export>query container</dfn>
-	for the purpose of [=container queries=] that require explicit containment
-	(such as [=container size queries=]),
-	allowing [=style rules=] styling its descendants
-	to query various aspects of its sizing and layout,
-	and respond accordingly.
-
-	Unless otherwise noted,
```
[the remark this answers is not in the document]
lilles: Moved the definition now.

css-conditional-5/Overview.bs
```diff
@@ -1067,6 +1087,159 @@ Style Container Features</h3>
 	are [=computed value|computed=] with respect to the [=query container=],
 	the same as other values.
 
+<h3 id="scroll-state-container">
+Scroll State Container Features</h3>
+
+	A <dfn export>container scroll-state query</dfn> allows to query a container for various
+	state that depends on scroll position. The ''scroll-state'' [=query container=] can either
```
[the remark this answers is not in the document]
lilles: > This feels like it's trying to be a definition of what the query container is in this case, but it doesn't feel like it gives a clear definition. I think it either (a) needs to be a clear definition or (b) needs to sound less like it's trying to define what the query container is. (I think (b) is the right thing because it's defined above, although I'm not sure.)

I changed the text to focus on scroll-state queries instead of the query container.

lilles:
> Overall approve, as I prefer to see good PR merged sooner and details worked out later. But I will wait for one of the other editors to do the merge. Not all of the @dbaron comments have been resolved.

Thanks!

svgeesus:
@dbaron are you now satisfied with the responses to your helpful comments on this PR? If so please r+

css-conditional-5/Overview.bs
dbaron: It's still not clear what happens when the sticky algorithm shifts both sides of the container in the same axis.  It's OK to address that as a followup, though.

css-conditional-5/Overview.bs
```diff
@@ -1206,11 +1227,15 @@ Sticky positioning: the '@container/overflowing' feature</h4>
 
 	The '@container/overflowing' [=container feature=] queries whether a
 	[=scroll container=] has [=scrollable overflow=] in the given direction for the
-	current scroll progress. The logical values map to physical based on the direction
-	and writing-mode of the [=query container=].
+	current scroll progress. Content not reachable through user scrolling, like
```
dbaron, line 1231: Hmmm, now that I'm rereading this text and your replies... what does "current scroll progress" mean here.  Does it mean that an LTR overflow:scroll container is `overflowing(left)` once it's been scrolled away from its initial horizontal position?  (That contradicts the following sentence, which I had suggested you add!)

css-conditional-5/Overview.bs
```diff
@@ -1115,9 +1115,9 @@ Updating Scroll State</h4>
 	current state once as the last step of [=run the scroll steps=]. Then, after the
 	resizeObserver loop in the
 	<a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model">HTML event loop processing model</a>,
-	the scroll state of every ''scroll-state'' [=query container=] is updated.
+	update the current state of every ''scroll-state'' [=query container=].
 	If that state has changed since the scroll state update in [=run the scroll steps=],
-	re-run the style and layout update if necessary.
+	re-run the style and layout update a single time if necessary.
```
dbaron, line 1121: At some point it needs to be clear what order this "single time" update has relative to the other features that do a "single time" update (since when both features are used together it can affect the results).  That doesn't need to be resolved right now, but it might be worth adding an issue.

dbaron: (or maybe the "scroll steps" don't have other things that do single-time update?  There certainly are a bunch of other features that work that way, but maybe none that hook into scroll steps?)

css-conditional-5/Overview.bs
```diff
@@ -657,19 +669,27 @@ Container Queries: the ''@container'' rule</h3>
...
+	<dfn><<style-query>></dfn>     = not <<style-in-parens>>
+	                  | <<style-in-parens>> [ [ and <<style-in-parens>> ]* | [ or <<style-in-parens>> ]* ]
+	                  | <<style-feature>>
+	<dfn><<style-in-parens>></dfn> = ( <<style-query>> )
+	                  | ( <<style-feature>> )
+	                  | <<general-enclosed>>
+
+	<dfn><<scroll-state-query>></dfn>     = not <<scroll-state-in-parens>>
+	                         | <<scroll-state-in-parens>> [ [ and <<scroll-state-in-parens>> ]* | [ or <<scroll-state-in-parens>> ]* ]
+	                         | <<scroll-state-feature>>
```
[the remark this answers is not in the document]
dbaron: I think this still needs a grammar production, not just prose.

dbaron (approved):
I left four new comments (two of which show up directly above, and two of which only show up in the middle of the earlier threads), but I'm fine with this landing either in its current state or with an attempt to address those.

css-conditional-5/Overview.bs
lilles: > At some point it needs to be clear what order this "single time" update has relative to the other features that do a "single time" update (since when both features are used together it can affect the results). That doesn't need to be resolved right now, but it might be worth adding an issue.

That is what issue #10796 mentioned at the top of this section is about.

> (or maybe the "scroll steps" don't have other things that do single-time update? There certainly are a bunch of other features that work that way, but maybe none that hook into scroll steps?)

The "scroll steps" updating of the current state takes effect in the "update style and layout" in the resizeObserver loop in the html spec. The state update with an extra style and layout update _after_ the resizeObserver loop does have the issue of which does that first when taking anchor positioning and scroll timelines into account. That's what issue #10796 tries to address.

css-conditional-5/Overview.bs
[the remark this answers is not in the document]
lilles: It refers to the `media feature` syntax just like the size and style features do, so if that needs further fixing it should be done consistently across the feature types.

css-conditional-5/Overview.bs
lilles: > It's still not clear what happens when the sticky algorithm shifts both sides of the container in the same axis. It's OK to address that as a followup, though.

How is that even possible? It can only be shifted in one direction per axis, right?

css-conditional-5/Overview.bs
lilles: Reworded.

css-conditional-5/Overview.bs
dbaron: Ah, never mind.  I guess I missed a sentence earlier in the [spec's text](https://drafts.csswg.org/css-position-3/#stickypos-insets):

> If this results in a [sticky view rectangle](https://drafts.csswg.org/css-position-3/#sticky-view-rectangle) size in any axis less than the size of the [border box](https://drafts.csswg.org/css-box-4/#border-box) of the sticky box in that axis, then the effective [end](https://drafts.csswg.org/css-writing-modes-4/#css-end)-edge inset in the affected axis is reduced (possibly becoming negative) to bring the sticky view rectangle’s size up to the size of the border box in that axis (where end is interpreted relative to the [writing mode](https://drafts.csswg.org/css-writing-modes-4/#writing-mode) of the [containing block](https://drafts.csswg.org/css-display-4/#containing-block)).

that removes the possibility that this can happen.  (But otherwise, reading just the algorithm in the following paragraph, it seems like it could... although I admit it's not clear what the priority would be!)

Issue #9822, opened 2024-01-19
[css-view-transitions-2] Define what happens when an outbound cross-document view transition starts before rendering


noamr:
Envision the current scenario:
- A document has an view-transition MPA opt-in (`@view-transition { navigation: auto }`)
- That document has some render-blocking resources that take a long time
- The user navigates away before the first frame

The current specified behavior is that navigating away would be stalled until rendering is unblocked.
Suggesting instead that until the page is revealed (the first frame is rendered), outbound cross-document view-transitions would be automatically skipped.

noamr:
Proposed resolution from internal sync: A page that hadn't rendered yet would not trigger a cross-document view-transition.

More precisely, a page that hasn't been revealed yet (probably due to render-blocking) would not count as opting in to an outbound view-transition.

PR #10815, opened 2024-09-02
[css-view-transitions-2] When hiding the document, and inbound cross-document view transition should be skipped.

noamr:
When hiding the document, and inbound cross-document view transition should be skipped.
This is a missing piece of #9822 that wasn't edited in.

Closes web-platform-tests/wpt#47828


nt1m (commented):
Are they added before or after the steps from level 1?

noamr:
> Are they added before or after the steps from level 1?

There would be no observable difference, but I made it clear that it's after to avoid amiguity.

khushalsagar (approved):

Issue #7708, opened 2022-09-07
[css-overflow] Is continue: discard working in the fragment tree useful?


emilio:
The main usefulness of it is clipping overflowing content from e.g. `line-clamp`, but it seems it'd be easier (and more compatible with how `-webkit-line-clamp` works now) if this behaved more like `visibility: hidden` (i.e., stop painting them) rather than `display: none` on the overflowing items?

It'd also be easier to implement, generally, and I'm not sure when it'd be less useful. It might be that I'm missing something obvious?

cc @bfgeek @frivoal @fantasai

tabatkins:
If it just stopped painting them, then the content would still be overflowing, but scrolling would just reveal a big empty area, yeah? I'm not sure why that would be preferable.

(`-webkit-line-clamp` is incredibly hacky and broken, I wouldn't look to it for guidance in any respect.)

emilio:
Well, we could spec that line-clamp affects the scrollable overflow, by not accounting for overflow caused by clamped lines, the same way it affects the auto block size, right? I think that'd still be much easier to implement and maybe more intuitive to authors

tabatkins:
At that point, what's making it different from discarding the fragments after the break, then? You already have the ability to shift the fragments to the next rendering context, so I'm not clear what the problem you're trying to solve here is.

emilio:
Well, for once, it doesn't change the interactions with the OM (e.g., `getBoundingClientRect()` etc keep behaving as if they were in non-fragmented contexts). FWIW in Gecko there's a difference between "pushing stuff to the next page / column" and "making it disappear into the void". We have no precedent for the later I don't think.

tabatkins:
So you're proposing we define that the overflow after the break opportunity doesn't break, but instead becomes invisible ink overflow?

(I'm also curious what the essential difference is between those two cases, btw. I would naively think that "push these to nowhere" is similar in difficulty, if not easier?)

bfgeek:
One specific difference is the OOFs would propagate up from the clamped content.

tabatkins:
Ah, in the "become invisible ink overflow" case? Whereas in the "drop the fragments" case the OOFs after the break would just be gone?

bfgeek:
> Ah, in the "become invisible ink overflow" case? Whereas in the "drop the fragments" case the OOFs after the break would just be gone?

Yup that's correct.

bfgeek:
The other difference/side-effect is the geometry APIs

[...]

PR #10816, opened 2024-09-02
[css-overflow-4] Draft spec for `continue: collapse` (#7708)

andreubotella:


bfgeek:
@emilio - Does this look reasonable to you?

andreubotella:
Before now, this PR allowed clamping between any two blocks, to make it closer to the `continue: discard` behavior, but in hallway conversations in TPAC we agreed that having it clamp to the last line before clamp, even if there are lineless boxes after it that will still fit, was a better behavior. I just updated the spec text to reflect this.

tabatkins:
> but in hallway conversations in TPAC we agreed that having it clamp to the last line before clamp, even if there are lineless boxes after it that will still fit, was a better behavior

In particular, the argument was that `auto` clamping should be equivalent to the largest `<integer>` clamp that avoids overflow; it's weird if `auto` actually lets you include an amount of content *between* two `<integer>` values, and thus can't be reproduced directly.

css-overflow-4/Overview.bs
```diff
@@ -1048,6 +1059,15 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
 			do not cause any content to be discarded.
+
+		<dt><dfn>collapse</dfn>
+		<dd>
+			If the box is a [=block container=], then it must
+			[=establish an independent formatting context=] that also becomes a [=line-clamp container=].
+			This causes all content after the [=clamp point=] to be visually hidden, and to not be
+			taken into account for the box's [=automatic block size=].
+
+			This keyword has no effect on [=multicol containers=].
```
frivoal, line 1070: Couldn't we make it work on multicols too? Is there any particular difficulty in inserting a clamp point after N lines or after the last line of the last non-overflow column, and hiding overflow columns in any case?

css-overflow-4/Overview.bs
```diff
@@ -1115,6 +1135,9 @@ Fragmentation of Overflow: the 'continue' property</h3>
 	would cause intrinsic sizes to depend on layout, which would cause circularities.
 -->
 
+	Note: This differs from the ''visibility: hidden''-like behavior of content after the [=clamp
+	point=] with ''continue: collapse''.
```
frivoal, line 1139: Editorial:
```suggestion
	Note: This differs from the ''visibility: hidden''-like behavior
	of content after the [=clamp point=] with ''continue: collapse''.
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
```diff
@@ -1123,12 +1146,12 @@ Fragmentation of Overflow: the 'continue' property</h3>
 
 	Additionaliy,
 	for compatibility (see [[#webkit-line-clamp]]),
-	when the [=computed value=] of the 'continue' property is ''discard''
+	when the [=computed value=] of the 'continue' property is ''collapse'' or ''-webkit-collapse''
```
frivoal, line 1149:
```suggestion
	when the [=computed value=] of the 'continue' property is '''collapse'', ''-webkit-collapse'', or ''discard''
```

Strictly speaking, I think it is not necessary to list the -webkit- value here, since it is defined "to behaves identically to collapse". But I agree it is clearer to include it anyway.

I suggest including ''discard'' as well. If implementations ever support both, that makes it easier to switch between the two.

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+
+	If a [=block container=] contains a [=clamp point=], within itself or in any of its descendants,
+	its [=automatic block size=] will not take into account any of its children after the clamp point.
+	This also applies for the [=line-clamp container=] itself.
+
+	NOTE: If there are any floats before the [=clamp point=], the [=line-clamp container=]'s
+	automatic size must grow to encompass the clearance, just like it would if it were a regular
+	[=block formatting context=] root that only contained its contents before the [=clamp point=].
+	This is regardless of whether this would cause any content after the [=clamp point=] to be
+	within the container's bounds.
```
frivoal, line 1202: Editorial:
```suggestion
	NOTE: If there are any floats before the [=clamp point=],
	the [=line-clamp container=]'s automatic size must grow to encompass the clearance,
	just like it would if it were a regular [=block formatting context=] root
	that only contained its contents before the [=clamp point=].
	This is regardless of whether this would cause any content after the [=clamp point=]
	to be within the container's bounds.
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+	Any boxes in a [=line-clamp container=] that follow its [=clamp point=] in the box tree, as well
+	as any [=line boxes=] that follow it inside an [=inline formatting context=], will be invisible
+	(behaving like ''visibility: hidden''). This includes [=independent formatting contexts=] as
+	well as [=out-of-flow=] boxes, and all of their descendants. Any overflow such boxes and line
+	boxes might have is always counted as [=ink overflow=] rather than [=scrollable overflow=].
+
+	NOTE: This differs from the ''display: none''-like behavior of not rendered content with
+	''continue: discard''.
+
+	If a [=block container=] contains a [=clamp point=], within itself or in any of its descendants,
```
frivoal, line 1194: Editorial:
```suggestion
	If a [=block container=] contains a [=clamp point=],
	within itself or in any of its descendants,
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+	point in the block formatting context.
+
+	Any boxes in a [=line-clamp container=] that follow its [=clamp point=] in the box tree, as well
+	as any [=line boxes=] that follow it inside an [=inline formatting context=], will be invisible
+	(behaving like ''visibility: hidden''). This includes [=independent formatting contexts=] as
+	well as [=out-of-flow=] boxes, and all of their descendants. Any overflow such boxes and line
+	boxes might have is always counted as [=ink overflow=] rather than [=scrollable overflow=].
+
+	NOTE: This differs from the ''display: none''-like behavior of not rendered content with
+	''continue: discard''.
```
frivoal, line 1192: Editorial:
```suggestion
	NOTE: This differs from the ''display: none''-like behavior
	of not rendered content with ''continue: discard''.
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+<h4 id=line-clamp-containers>Line-clamp containers</h4>
+
+	A <dfn>line-clamp container</dfn> is an [=independent formatting context|independent=] [=block formatting context=] that
+	additionally follows the rules in this section. All line-clamp containers contain a
+	<dfn>clamp point</dfn>, which is one of the following positions inside it:
+
+	- A point immediately after a [=line box=] in the line-clamp container's [=block formatting
+	  context=].
+	- The end of the line-clamp container, if there are any boxes after the last [=inline formatting
+	  context=] in the line-clamp container's [=block formatting context=].
```
frivoal, line 1175: As discussed in an earlier comments, I am not sure I understand what the difference/benefit is of placing a clamp point at the end, instead of saying there's no clamp point. We can define that a clamp point at the end has no effect (which I what I believe the definition bellow does), but would it not be simpler to just say that there isn't one?

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
 	as the box would be a [=flex container=] rather than a [=block container=],
 	and thus the 'continue' property would not apply.
 
+<h4 id=line-clamp-containers>Line-clamp containers</h4>
+
+	A <dfn>line-clamp container</dfn> is an [=independent formatting context|independent=] [=block formatting context=] that
+	additionally follows the rules in this section. All line-clamp containers contain a
+	<dfn>clamp point</dfn>, which is one of the following positions inside it:
```
frivoal, line 1170: Editorial
```suggestion
	A <dfn>line-clamp container</dfn> is an [=independent formatting context|independent=] [=block formatting context=]
	that additionally follows the rules in this section.
	All line-clamp containers contain a <dfn>clamp point</dfn>,
	which is one of the following positions inside it:
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+
+	- A point immediately after a [=line box=] in the line-clamp container's [=block formatting
+	  context=].
```
frivoal, line 1173: Editorial
```suggestion
	- A point immediately after a [=line box=]
		in the line-clamp container's [=block formatting context=].
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
frivoal, line 1175: Editorial
```suggestion
	- The end of the line-clamp container,
		if there are any boxes after the last [=inline formatting context=]
		in the line-clamp container's [=block formatting context=].
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+	all previous possible clamp points, the line-clamp container's [=automatic block size=] (as
+	determined below) is not greater than the [=block size=] the box would have if its automatic
+	block size were infinite; or if that is not the case for any clamp points, to the first clamp
+	point in the block formatting context.
+
+	Any boxes in a [=line-clamp container=] that follow its [=clamp point=] in the box tree, as well
+	as any [=line boxes=] that follow it inside an [=inline formatting context=], will be invisible
+	(behaving like ''visibility: hidden''). This includes [=independent formatting contexts=] as
+	well as [=out-of-flow=] boxes, and all of their descendants. Any overflow such boxes and line
+	boxes might have is always counted as [=ink overflow=] rather than [=scrollable overflow=].
```
frivoal, line 1189: Editorial
```suggestion
	Any boxes in a [=line-clamp container=] that follow its [=clamp point=] in the box tree,
	as well as any [=line boxes=] that follow it inside an [=inline formatting context=],
	will be invisible (behaving like ''visibility: hidden'').
	This includes [=independent formatting contexts=]
	as well as [=out-of-flow=] boxes,
	and all of their descendants.
	Any overflow such boxes and line boxes might have
	is always counted as [=ink overflow=] rather than [=scrollable overflow=].
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1163,44 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+	- The end of the line-clamp container, if there are any boxes after the last [=inline formatting
+	  context=] in the line-clamp container's [=block formatting context=].
+
+	If the [=line-clamp container=]'s [=block formatting context root=] has a [=computed value=] of
+	'max-lines' other than ''max-lines/none'', then that property will determine the [=clamp point=].
+	Otherwise, the clamp point will be set to the last possible clamp point such that, for it and
+	all previous possible clamp points, the line-clamp container's [=automatic block size=] (as
+	determined below) is not greater than the [=block size=] the box would have if its automatic
+	block size were infinite; or if that is not the case for any clamp points, to the first clamp
+	point in the block formatting context.
```
frivoal, line 1183: Editorial:
```suggestion
	If the [=line-clamp container=]'s [=block formatting context root=]
	has a [=computed value=] of 'max-lines' other than ''max-lines/none'',
	then that property will determine the [=clamp point=].
	Otherwise, the clamp point will be set to the last possible clamp point
	such that, for it and all previous possible clamp points,
	the line-clamp container's [=automatic block size=]
	(as determined below)
	is not greater than the [=block size=] the box would have
	if its automatic block size were infinite;
	or if that is not the case for any possible clamp points,
	to the first possible clamp point in the block formatting context.
```
See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

I also added the word "possible" before "clamp point" in the last two lines.

css-overflow-4/Overview.bs
```diff
@@ -707,12 +707,15 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
...
+	to indicate the continuity of truncated / interrupted content.
 	It only affects line boxes contained directly by the <a>block container</a> itself,
 	but as it inherits, will have an effect on descendants’ line boxes unless overridden.
-	If the box contains no line box immediately preceding a <a>region break</a>,
-	then this property has no effect.
+
+	This property only affects a line box if it immediately precedes
+	eiher a [=region break=] or a [=clamp point=] in the [=block formatting context=].
+	If the [=clamp point=] is placed at the end of the [=line-clamp container=], then the line will
+	not be affected.
```
frivoal, line 718: Editorial
```suggestion
	This property only affects a line box
	if it immediately precedes eiher a [=region break=]
	or a [=clamp point=] in the [=block formatting context=].
	If the [=clamp point=] is placed at the end of the [=line-clamp container=],
	then the line will not be affected.
```
Reason: this spec uses semantic line-breaks, and I'd prefer to stay consistent with that. See https://rhodesmill.org/brandon/2012/one-sentence-per-line/

frivoal:
@tabatkins , for your comment in https://github.com/w3c/csswg-drafts/pull/10816#issuecomment-2394275025, I'd suggest following up in https://github.com/w3c/csswg-drafts/issues/10868, because that aspect exists before this PR. The text of the spec may need to change depending on where we land on that issue (and others), but this PR isn't changing it, so I think we should take one problem at a time.

css-overflow-4/Overview.bs
andreubotella: In inserting a clamp point after N lines, no. For adding it in the last line of the last non-overflowing column, at least in my implementation in Chromium it might not be easy, because the machinery to figure out whether a line is the last line that doesn't overflow, including accounting for border/padding/margin, would have to interact with the fragmentation machinery.

css-overflow-4/Overview.bs
```diff
@@ -947,28 +952,33 @@ Forcing a Break After a Set Number of Lines: the 'max-lines' property</h3>
...
+
+	- If the box is a [=fragmentation container=] that captures [=region breaks=],
+	    a <a>region break</a> or
+	    is forced after its <var>N</var>th
+	    descendant <a>in-flow</a> <a>line box</a>.
+	    If fewer than <var>N</var> line boxes exist,
+	    then 'max-lines' introduces no <a>region break</a>.
+	- If the box is a [=line-clamp container=], its [=clamp point=] is set after its <var>N</var>th
+	    descendant <a>in-flow</a> <a>line box</a>.
+	    If fewer than <var>N</var> line boxes exist, then the clamp point is set to the end of the
```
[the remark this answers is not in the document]
andreubotella: I added a clamp point at the end because, for `line-clamp: auto`, that is enough to explain the lack of visible clamping when the max-height is larger than the box's auto size. The last clamp point that fits is the clamp point at the end, which is visually equivalent to no clamping.

I don't feel strongly about this, though, so we could change this to not have a clamp point at the end, as long as we also say that the last line doesn't have a following clamp point (maybe unless it's followed by non-inline content, depending on whether we allow clamping between lineless blocks).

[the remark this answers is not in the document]
andreubotella: With the changes to bring this PR into alignment with the resolution in #10868, I also changed it to not allow the clamp point at the end.

css-overflow-4/Overview.bs
```diff
@@ -947,28 +957,34 @@ Forcing a Break After a Set Number of Lines: the 'max-lines' property</h3>
...
+	    a <a>region break</a>
+	    is forced after its <var>N</var>th
+	    descendant <a>in-flow</a> <a>line box</a>.
+	    If fewer than <var>N</var> line boxes exist,
+	    then 'max-lines' introduces no <a>region break</a>.
+	- If the box is a [=line-clamp container=],
+	its [=clamp point=] is set after its <var>N</var>th
+	    descendant <a>in-flow</a> <a>line box</a>.
+	    If fewer than <var>N</var> line boxes exist,
+	    then the clamp point is set to the end of the line-clamp container.
```
tabatkins, line 980:
```suggestion
		a <a>region break</a>
		is forced after its <var>N</var>th
		descendant <a>in-flow</a> <a>line box</a>.
		If fewer than <var>N</var> line boxes exist,
		then 'max-lines' introduces no <a>region break</a>.
	- If the box is a [=line-clamp container=],
		its [=clamp point=] is set after its <var>N</var>th
		descendant <a>in-flow</a> <a>line box</a>.
		If fewer than <var>N</var> line boxes exist,
		then the clamp point is set to the end of the line-clamp container.
```

css-overflow-4/Overview.bs
```diff
@@ -947,28 +957,34 @@ Forcing a Break After a Set Number of Lines: the 'max-lines' property</h3>
...
+
+
+	Only line boxes in the same <a>Block Formatting Context</a>
```
tabatkins, line 983:
```suggestion
	Only line boxes in the same <a>block formatting context</a>
```

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1170,55 @@ Fragmentation of Overflow: the 'continue' property</h3>
 	as the box would be a [=flex container=] rather than a [=block container=],
 	and thus the 'continue' property would not apply.
 
+<h4 id=line-clamp-containers>Line-clamp containers</h4>
+
+	A <dfn>line-clamp container</dfn> is an [=independent formatting context|independent=] [=block formatting context=]
```
tabatkins, line 1175:
```suggestion
	A <dfn export>line-clamp container</dfn> is an [=independent formatting context|independent=] [=block formatting context=]
```

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1170,55 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+	that additionally follows the rules in this section.
+	All line-clamp containers contain a <dfn>clamp point</dfn>,
+	which is one of the following positions inside it:
```
tabatkins, line 1178:
```suggestion
	which is one of the following block-axis positions inside it:
```

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1170,55 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+	has a [=computed value=] of 'max-lines' other than ''max-lines/none'',
+	then that property will determine the [=clamp point=].
+	Otherwise, the clamp point will be set to the last possible clamp point
+	such that, for it and all previous possible clamp points,
+	the line-clamp container's [=automatic block size=]
+	(as determined below)
+	is not greater than the [=block size=] the box would have
+	if its automatic block size were infinite;
+	or if that is not the case for any possible clamp points,
+	to the first possible clamp point in the block formatting context.
```
tabatkins, line 1196: This could use an example of what the "Otherwise" behavior is trying to guard against.

css-overflow-4/Overview.bs
```diff
@@ -1140,6 +1170,55 @@ Fragmentation of Overflow: the 'continue' property</h3>
...
+
+	Any boxes in a [=line-clamp container=] that follow its [=clamp point=] in the box tree,
+	as well as any [=line boxes=] that follow it inside an [=inline formatting context=],
+	will be invisible (behaving like ''visibility: hidden'').
```
tabatkins, line 1200:
```suggestion
	become [=invisible boxes=].
```

tabatkins (approved):
LGTM after an example or explanation of the one case I called out.

css-overflow-4/Overview.bs
andreubotella: It's not guarding against anything. In this PR, currently, `line-clamp: 3` will set `max-lines: 3`, and that will clamp based on the number of lines; while `line-clamp: auto` will set `max-lines: none`, which will clamp based on the height. This height-based clamping is what the entire "otherwise" clause describes.

I will definitely have to do some rewriting and clarification, though, because this is definitely not obvious from just skimming the spec.

andreubotella:
I just noticed that @syncbot for some reason got rid of three commits I made updating this PR to match the April F2F. I've just cherry-picked them.

@tabatkins, would you mind reviewing those again?

tabatkins (approved):

PR #10817, opened 2024-09-02
[css-view-transition-2] Clarify timeout behavior for inbound view transition

### LINKED ISSUE #10800: [css-view-transitions-2] Clarify inbound transition timeout semantics
The current spec mentions a UA defined timeout as part of `resolve inbound cross-document view-transition`, which then cancels the active view transition (in a queued task).

The `pageswap-long-delay.html` test is using a very-long synchronous `pageswap` event to trigger the timeout, which suggests that the timed period starts before this event (which is fairly removed and before the `resolve inbound cross-document view-transition` algorithm).

It would be good to clarify when the timed period starts.

The test is also expecting the `pagereveal` event's `viewTransition` parameter to be null, which is different from queued cancellation of the view transition. I'm not sure if this is a test bug, or a spec bug.

noamr:
The previous wording of this was somewhat handwavy, made it more rigorous.

- The timeout starts when the transition is created (before pageswap)
- If the timeout expires before the transition is resolved (pagereveal), the inbound transition params are cleared and the behavior is similar to having the transition skipped at pageswap.
- If the timeout expires while the navigation is active, it is skipped with a `TimeoutError`.

Closes #10800

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -995,14 +995,16 @@ The [=captured element=] struct should contain these fields, in addition to the 
 
 			Note: The inbound transition is activated after the dispatch of {{Window/pagereveal}} to ensure mutations made in this event apply to the captured new state.
 
-		1.At any given time, the UA may decide to skip the inbound transition, e.g. after an [=implementation-defined=] timeout. To do so, the UA should [=queue a global task=] on the [=DOM manipulation task source=] given |newDocument|'s [=relevant global object=] to perform the following steps:
+		1. To skip the transition after a timeout, the user agent may perform the following steps [=in parallel=]:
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -875,6 +875,9 @@ It has the following [=struct/items=]:
 
 	: <dfn>initial snapshot containing block size</dfn>
 	:: a [=tuple=] of two numbers (width and height).
+
+	: <dfn>resolved transition</dfn>
```
khushalsagar, line 879: This doesn't look right. This data structure is meant to be serialized across Documents and I don't think we can or want to serialize the script "ViewTransition" object?

noamr: It's null when you serialize it from the old document... Also serialization is somewhat of an implementation concern.
We can make this something that's "associated" with the view transition params but I don't think it matters too much.

khushalsagar (approved):

PR #10826, opened 2024-09-03
[css-view-transitions-1] Swap the order between setting the phase and calling the update callback

### LINKED ISSUE #10822: [css-view-transitions] State management seems broken in the spec
Per spec, this snippet calls the transition callback twice:

```js
document.startViewTransition(async function() {
  console.log("cb");
  document.startViewTransition(() => {});
});
```

This is in fact what happens in Safari TP, and in my experimental Firefox implementation.

This is because [call the update callback](https://drafts.csswg.org/css-view-transitions-1/#call-the-update-callback) says:

  * Otherwise, set callbackPromise to the result of [invoking](https://webidl.spec.whatwg.org/#invoke-a-callback-function) transition’s [update callback](https://drafts.csswg.org/css-view-transitions-1/#viewtransition-update-callback).
  * If transition’s [phase](https://drafts.csswg.org/css-view-transitions-1/#viewtransition-phase) is not "done", then set transition’s phase to "update-callback-called".

But that's already too late, because if something (like `startViewTransition`) skips the transition, the state is not "update-callback-called".

It seems for this specific case, just swapping those two steps should work. But might be worth looking at similar things in the spec.

cc @khushalsagar @vmpstr @mattwoodrow @nt1m

noamr:
This fixes an issue where skipping the transition from within thet update callback would call the update callback twice.

Closes #10822

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1600,11 +1600,11 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |callbackPromise| be null.
 
+		1. If |transition|'s [=ViewTransition/phase=] is not "`done`", then set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
+
 		1. If |transition|'s [=ViewTransition/update callback=] is null,
 			then set |callbackPromise| to [=a promise resolved with=] undefined,
 			in |transition|'s [=relevant Realm=].
-
-		1. If |transition|'s [=ViewTransition/phase=] is not "`done`", then set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
 		1. Otherwise, set |callbackPromise| to the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1604,10 +1604,10 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 			then set |callbackPromise| to [=a promise resolved with=] undefined,
 			in |transition|'s [=relevant Realm=].
 
-		1. Otherwise, set |callbackPromise| to the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
-
 		1. If |transition|'s [=ViewTransition/phase=] is not "`done`", then set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
 
+		1. Otherwise, set |callbackPromise| to the result of [=/invoking=] |transition|'s [=ViewTransition/update callback=].
```
nt1m, line 1609: Is "otherwise" here correct?

noamr: Oops, fixed

css-view-transitions-1/Overview.bs
```diff
@@ -1600,14 +1600,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 		1. Let |callbackPromise| be null.
 
+		1. If |transition|'s [=ViewTransition/phase=] is not "`done`", then set |transition|'s [=ViewTransition/phase=] to "`update-callback-called`".
```
emilio, line 1603: May as well move it further up to keep the code using `callbackPromise` together?

emilio (approved):

css-view-transitions-1/Overview.bs
noamr: Sure, done

Issue #3473, opened 2018-12-30
[css-text] Preventing too-short final lines of blocks (Last Line Minimum Length)


s10wen:
CodePen example:
https://codepen.io/s10wen/pen/GPWWyP?editors=1100#0

Tweet + replies:
https://twitter.com/s10wen/status/1076079575506083840

Wikipedia explanation:
https://en.wikipedia.org/wiki/Widows_and_orphans

CSS Text 3 w3 Spec:
https://www.w3.org/TR/css-text-3/

The above links have led me here, to further pursue this. I'm wondering if anything currently exists, or could be implemented to handle this. My idea is that `orphan: 2` would always leave two strings of text together, please see the CodePen for an example. Or, it could be that `orphan: true`, would mean that orphans always had at least 2 words.

frivoal:
There are already controls for widows and orphan lines and page/column breaks in https://drafts.csswg.org/css-break/#widows-orphans.

A control for widowed words on the last line could be useful, but it doesn't exist yet. However, I suspect it needs to be paired with a better line breaking algorithm than the current greedy one to achieve good results. If all the lines of the paragraph are re-balanced to  push the needed word(s) to the last line, all may be fine, but just pulling from the line before last is probably going to lead to sub-optimal results. Also, we'd need to figure out what this means in languages that do not separate words with spaces, such as Japanese or Chinese.

I suppose there's prior art in other software, and we should have a look at what they do there. InDesign maybe?

astearns:
We have previously talked about the idea of being able to specify a minimum last line length, but in characters or a percentage of width, not in the number of words:

https://drafts.csswg.org/css-text-4/#last-line-limits

jonjohnjohnson:
@s10wen chiming in on current arounds. If you're willing to put presentational matters in html, like your demos use of `<br>`, it's instead optimal to place a `&nbsp;` character between the last two words, making it so they don't have to create a new line if they fit on their current line together.

s10wen:
Hey all, thanks for the conversation around this.

https://drafts.csswg.org/css-break/#widows-orphans seems to be most likely what I'd like to see. Is there anywhere I can see the progress of this being implementing to browsers and test?

AmeliaBR:
I've just changed the title of this issue to make it clearer that we're talking about orphaned words on a line, not orphaned lines on the top/bottom of a page or column (which is what the `widows` and `orphans` properties are about).

[...]

PR #10837, opened 2024-09-05
[css-text] Introduce a text-wrap-style value focused on orphans

frivoal:
See https://github.com/w3c/csswg-drafts/issues/3473#issuecomment-1646267487

### DIFF
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -5501,7 +5501,7 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
 
 	<pre class="propdef">
 	Name: text-wrap-style
-	Value: auto | balance | stable | pretty
+	Value: auto | balance | stable | pretty | reduce-orphans
 	Initial: auto
 	Applies to: [=block containers=] hat establish an [=inline formatting context=]
 	Inherited: yes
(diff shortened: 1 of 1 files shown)


css-text-4/Overview.bs
```diff
@@ -5612,9 +5612,87 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
...
+
+			<div class=advisement>
+				The necessary computations may be expensive,
+				especially when applied to large amounts of text.
+				Authors are encouraged to asses the impact on performance
+				when using ''text-wrap-style: pretty'',
+				and possibly use it selectively where it matters most.
+			</div>
+
+		<dt><dfn>reduce-orphans</dfn>
```
fantasai, line 5638:
```suggestion
		<dt><dfn>avoid-orphans</dfn>
```

css-text-4/Overview.bs
```diff
@@ -5612,9 +5612,87 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
...
+		<dd>
+			Specifies the UA <em>should</em> avoid excessively short last lines,
+			and is expected to consider more than one line
+			when making break decisions.
+			The user agent may decide against improving the last line
+			when it would make some prior line(s) substantially unbalanced.
+			The user agent should not attempt to improve other aspects of the layout
+			beyond what ''text-wrap-style/auto''
+			if that comes at a significant performance cost.
```
fantasai, line 5647:
```suggestion
			if those improvements come at a significant performance cost.
```

css-text-4/Overview.bs
```diff
@@ -5626,10 +5704,12 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
...
 	with this default value,
 	but as optimal results often take more time,
-	''text-wrap-style/pretty'' is offered as an opt-in
+	''text-wrap-style/pretty'' and ''reduce-orphans'' are offered as opt-ins
 	to take more time for better results.
-	The ''text-wrap-style/pretty'' value is intended for body text,
-	where the last line is expected to be a bit shorter than the average line;
+	Both values are intended for body text,
+	where the last line is expected to be a bit shorter than the average line,
+	but ''reduce orphans'' aims to avoid excessively short last lines and to stay performant,
```
fantasai, line 5711:
```suggestion
	but ''reduce-orphans'' aims to avoid excessively short last lines while staying performant,
```

css-text-4/Overview.bs
frivoal: cc: @astearns The key is to find a term that indicates that the value is about getting rid of orphans, but not in all circumstances, and not at any cost.
`avoid` as suggested here by @fantasai is already used in CSS (see https://drafts.csswg.org/indexes/) in a number of places with the meaning "don't, unless [X]", so I think it can work: the precise criteria for when not to do it is left up to the UA, but it is there.
To me, `reduce` could work too, but the meaning is almost too weak: instead of a rule generally applied with exceptions as implied by avoid, it would seem to indicate a slant, a preference. It's not wholly inappropriate here, especially given that everything is largely up to the user agent anyway, but that seems weaker than the author intent we're aiming for.

css-text-4/Overview.bs
```diff
@@ -5626,10 +5704,12 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
 	UAs can experiment with better line breaking algorithms
 	with this default value,
 	but as optimal results often take more time,
-	''text-wrap-style/pretty'' is offered as an opt-in
+	''text-wrap-style/pretty'' and ''reduce-orphans'' are offered as opt-ins
 	to take more time for better results.
-	The ''text-wrap-style/pretty'' value is intended for body text,
-	where the last line is expected to be a bit shorter than the average line;
+	Both values are intended for body text,
+	where the last line is expected to be a bit shorter than the average line,
```
astearns, line 5710: Let’s change “is expected to be” to “might happen to be”

astearns (changes requested):

css-text-4/Overview.bs
```diff
@@ -5612,9 +5612,87 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
...
+			and may include things such as:
+			reducing the variation in length between lines;
+			avoiding <a href="https://en.wikipedia.org/wiki/River_(typography)">typographic rivers</a>;
+			prioritizing different classes of [=soft wrap opportunities=], [=hyphenation opportunities=], or [=justification opportunities=];
+			avoiding hyphenation on too many consecutive lines…
+
+			<div class=advisement>
+				The necessary computations may be expensive,
+				especially when applied to large amounts of text.
+				Authors are encouraged to asses the impact on performance
```
frivoal, line 5633:
```suggestion
				Authors are encouraged to assess the impact on performance
```

css-text-4/Overview.bs
```diff
@@ -5612,9 +5612,87 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
...
+			Specifies the UA <em>should</em> avoid excessively short last lines,
+			and is expected to consider more than one line
+			when making break decisions.
+			The user agent may decide against improving the last line
+			when it would make some prior line(s) substantially unbalanced.
+			The user agent should not attempt to improve other aspects of the layout
+			beyond what ''text-wrap-style/auto''
+			if those improvements come at a significant performance cost.
+
+			Note: In western typography, an “orphan” typically refers to a single word on the last line of a paragraph.
```
frivoal, line 5649:
```suggestion
			Note: In western typography, an “orphan” can refer to a single word on the last line of a paragraph.
```

css-text-4/Overview.bs
frivoal, line 5710: How about:
```suggestion
	where the last line is often a bit shorter than the average line,
```

css-text-4/Overview.bs
```diff
@@ -5612,9 +5612,87 @@ Selecting How to Wrap: the 'text-wrap-style' property</h3>
...
+				Authors are encouraged to assess the impact on performance
+				when using ''text-wrap-style: pretty'',
+				and possibly use it selectively where it matters most.
+			</div>
+
+		<dt><dfn>reduce-orphans</dfn>
+		<dd>
+			Specifies the UA <em>should</em> avoid excessively short last lines,
+			and is expected to consider more than one line
+			when making break decisions.
```
frivoal, line 5642:
```suggestion
			when making break decisions
			(for example, to avoid "fixing" an orphan
			by making the previous line too short).
```

css-text-4/Overview.bs
frivoal: > authors are going to expect that all orphans will be eliminated

If they don't think about too much, probably. But if you show them examples of cases where it would be bad to do so, and it doesn't do so, I suspect they'll be happy with the result, since forcibly doing it would be ugly.

arthurdapaz (approved):

PR #10839, opened 2024-09-05
[css-shadow-parts][css-pseudo] Move various ::part() details to 'part-like pseudo-element'.

### LINKED ISSUE #10083: [css-pseudo][css-shadow-parts] Define "part-like pseudo-element" concept

tabatkins:
Fixes #10083.

@fantasai, could you review? I'm not an editor on the Pseudo spec. ^_^

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -1305,27 +1305,47 @@ Tree-Abiding and Part-Like Pseudo-elements</h2>
     parsing/tree-abiding-pseudo-elements.html
   </wpt>
 
+<h3 id=partlike>
+Part-Like Pseudo-Elements</h3>
+
   A subset of [=tree-abiding pseudo-elements=],
-  the <dfn>part-like pseudo-elements</dfn>,
+  the <dfn export lt="part-like|part-like pseudo-element">part-like pseudo-elements</dfn>,
   have slightly stronger requirements:
   they act as if they have a well-defined location in the document tree.
   This enables them to interact with some other platform features
-  as if they were real elements.
+  as if they were real elements
+  (and, in fact, they often <em>are</em> real elements,
+  such as in the case of ''::part()'',
+  which simply aren't accessible in the current tree).
 
-  <div class=example>
-    For example, a [=part-like pseudo-element=]
-    can be used in the <{html-global/exportparts}> attribute,
-    to masquerade as a ''::part()''
-    for the component it's in:
+  [=Part-like pseudo-elements=] inherit from their [=originating element=] by default,
+  like a standard [=tree-abiding pseudo-element=],
+  but can declare that they inherit from another element instead
+  (possibly not accessible).
+  (For example, ''::part()'' inherits from
+  the parent of the element it represents in the shadow tree.)
 
-    <xmp class=html>
-      <template id=custom-element-template>
-        <p exportparts="::before : preceding-text">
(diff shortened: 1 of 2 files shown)


fantasai:
I'm just gonna merge it and fix it up later, cuz that's easier. ^_^

Issue #10440, opened 2024-06-13
Styling form control pickers


annevk:
As part of the discussion around [`appearance: base`](https://github.com/w3c/csswg-drafts/issues/5998) colleagues and I [presented a vision](https://lists.w3.org/Archives/Public/www-archive/2024May/att-0004/stylable-form-controls.pdf) of how this could work. The idea outlined in that presentation is that stylability of form control pickers (the overlay box for select options, selecting a date, etc.) would have its own independent opt-in. This is because we think that form controls (excluding their pickers) can probably be mostly done all at once with a simple `appearance: base` switch and accompanying pseudo-elements.

Form control pickers however are quite a bit more involved and we expect it to take longer to tackle them and some we expect will never be stylable (e.g., the file picker). The idea is that when a form control uses `appearance: base` you can address its picker using `::picker(form-control-identifier)` and independently set that to `appearance: base` as well. E.g., `select, select::picker(select) { appearance:base }` would give you a stylable select. `::picker(form-contorl-identifier)` would represent the outermost box of the form control picker and additional pseudo-elements will be needed to address the various pieces it consists of.

josepharhar:
Thanks for making an issue for this! I imagine I could replace the `::select-fallback-datalist` or `::select-datalist` pseudo-element that [I've implemented and proposed](https://github.com/w3c/csswg-drafts/issues/10462) with `::picker(select)` and have everything work the same.

I like how this enables incremental opt-in, but would still require us to implement/ship appearance:base for all the non-picker form controls to truly make it incremental, right?

I also like how it enables appearance:base for the button and the popup separately, but I imagine that some authors wouldn't like the additional CSS needed to go from `select { appearance:base-select; }` to `select { appearance:base; } select::picker(select) { appearance:base; }`.

una:
Just catching up on the latest status on this. I want to advocate for simplicity for this feature from a user perspective.

It doesn't seem like there is a benefit to developers (or a difference) between `select { appearance:base; }` and `select::picker(select) { appearance:base; }`. Why force developers to remember both selectors that they will need to target, degrading the developer experience?

The most likely and straightforward developer experience would be to name these selects, i.e. with a class, and access them like `.account-dropdown { appearance: base; /* styles */ }`. Ideally that is all that would be needed to enter into this new behavior, and we wouldn't require an additional pseudo class. With the picker proposal, developers would need to write:

```
.account-dropdown {
  appearance: base;
  /* styles */
}

[...]

PR #10865, opened 2024-09-10
[css-pseudo-4] Define ::picker()

josepharhar:
This was discussed here: https://github.com/w3c/csswg-drafts/issues/10440

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -1337,6 +1337,43 @@ details[open]::details-content {
 }</pre>
   </div>
 
+<h3 id="picker-pseudo">
+Styling form control pickers: the ''::picker()'' pseudo-element</h3>
+
+  The <dfn>::picker()</dfn> pseudo-element targets the popup picker of form
+  control elements which have popup pickers, such as the <{select}> element. It
+  is a [=part-like pseudo-element=].
+
+  <pre class=prod>
+    ::picker() == ::picker( <<ident>>+ )
+  </pre>
+
+  The ''::picker()'' pseudo-element only matches when the <a>originating
+  element</a> supports <a>base appearance</a> and has a popup picker. The
+  specified <<ident>> must also match the unique picker name of the
+  <a>originating element</a>. For example, the unique picker name for the
+  <{select}> element is "select".
+
+  There is no restriction on which properties apply to the ''::picker()''
+  pseudo-element.
+
+  In order for the ''::picker()'' pseudo-element to be rendered, it and its
+  <a>originating element</a> must be rendered with <a>base appearance</a> by
+  applying ''appearance: base'' to both of them.
+
+  <pre class="example">
+&lt;style>
+  select, select::picker(select) {
+    appearance: base;
+  }
+  select::picker(select) {
+    border: 5px solid red;
+    background-color: blue;
+  }
+&lt;/style>
+  </pre>
+
+
 <h2 id="interactions">
 Overlapping Pseudo-element Interactions</h2>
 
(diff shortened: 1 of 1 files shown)


css-pseudo-4/Overview.bs
```diff
@@ -1337,6 +1337,43 @@ details[open]::details-content {
...
+  select, select::picker(select) {
+    appearance: base;
+  }
+  select::picker(select) {
+    border: 5px solid red;
+    background-color: blue;
+  }
+&lt;/style>
+  </pre>
+
```
dbaron, line 1375: I think it's probably worth adding a note (you can just start a paragraph with "Note: " and it will get the right styles) that once all of the controls that might support picker styling have it, then we might add a (no-argument) `::picker` pseudo-element as well.  (I think it's worth adding because this may be a ways in the future, and the people who made these plans today might not be around to do that part!)

css-pseudo-4/Overview.bs
```diff
@@ -1337,6 +1337,43 @@ details[open]::details-content {
...
+  The ''::picker()'' pseudo-element only matches when the <a>originating
+  element</a> supports <a>base appearance</a> and has a popup picker. The
+  specified <<ident>> must also match the unique picker name of the
+  <a>originating element</a>. For example, the unique picker name for the
+  <{select}> element is "select".
+
+  There is no restriction on which properties apply to the ''::picker()''
+  pseudo-element.
```
dbaron, line 1358: This is implied by it being part-like, although maybe it's worth stating anyway.  But maybe start with "Since it is a [=part-like pseudo-element=], ".

css-pseudo-4/Overview.bs
```diff
@@ -1337,6 +1337,43 @@ details[open]::details-content {
...
+  control elements which have popup pickers, such as the <{select}> element. It
+  is a [=part-like pseudo-element=].
+
+  <pre class=prod>
+    ::picker() == ::picker( <<ident>>+ )
+  </pre>
+
+  The ''::picker()'' pseudo-element only matches when the <a>originating
+  element</a> supports <a>base appearance</a> and has a popup picker. The
+  specified <<ident>> must also match the unique picker name of the
```
dbaron, line 1353: I think this should probably be defining the term "unique picker name" (for other specs to reference), unless you think it should be defined somewhere else.

dbaron (commented):
This looks mostly fine.  A few comments below.

css-pseudo-4/Overview.bs
josepharhar: I agree. @fantasai are you ok with this?

css-pseudo-4/Overview.bs
josepharhar: I copied this from a bunch of other part-like pseudo-elements in this spec like `::file-selector-button` and `::details-content`.

> But maybe start with "Since it is a [=part-like pseudo-element=], ".

Done

css-pseudo-4/Overview.bs
dbaron: Yes, I think HTML should define it.

css-pseudo-4/Overview.bs
dbaron: Yeah, a bunch of stuff needs cleaning up, but we may as well set a good example here.

css-pseudo-4/Overview.bs
dbaron: I think you can go ahead without an explicit response from @fantasai here given that it matches the last sentence of the middle bullet of https://github.com/w3c/csswg-drafts/issues/10440#issuecomment-2303554166 .

dbaron (approved):

PR #10922, opened 2024-09-19
[css-view-transitions-2] Allow `auto` as a keyword

### LINKED ISSUE #8320: [css-view-transitions-2] view-transition-name determined by element
Thinking of this example https://codepen.io/jaffathecake/pen/VwBprqL

In this case the developer has to manually give each list item a unique `view-transition-name`. This could be avoided via something like:

```css
.list-item {
  view-transition-name: element-uuid();
}
```

Where `element-uuid()` returns a value unique and constant for that DOM node.

The proposal for CSS `random()` has a `per-element` value that could be generalised as `element-uuid()`. https://github.com/w3c/csswg-drafts/issues/2826#issuecomment-1204305712 (cc @tabatkins)

The downside to this is it would only work for same-document transitions, since `element-uuid()`s will always be different across documents. It also depends on element usage being consistent, which some frameworks don't always get right.

**Edit:** This would depend on https://github.com/w3c/csswg-drafts/issues/8319, otherwise styling the resulting pseudos is impossible.

noamr:
`auto` resolves to either the element's ID, or to a unique identifier that is stable for the duration of the transition and is not web-observable.

Resolution: https://github.com/w3c/csswg-drafts/issues/8320#issuecomment-2344208387
Closes #8320


css-view-transitions-2/Overview.bs
```diff
@@ -755,7 +755,20 @@ div.box {
...
+## Additions to 'view-transition-name' ## {#additions-to-vt-name}
+
+In addition to the existing values, the 'view-transition-name' also accepts an <dfn for=view-transition-name type=value export>auto</dfn> keyword.
+To resolve the [=used value=] of 'view-transition-name' for |element|:
+	1. Let |computed| be the [=computed value=] of 'view-transition-name'.
+	1. If |computed| is <css>none</css>, return null.
+	1. If |computed| is a <<custom-ident>>, return |computed|.
+	1. Assert: |computed| is <css>auto</css>.
+	1. If |element| has an associated [=Element/id=], then return |element|'s [=Element/id=].
+	1. Return a unique string, which is not a <<custom-ident>>. This string should remain consistent at least for the lifetime of |element|'s [=node document=]'s [=active view transition=].
```
khushalsagar, line 769: I'm not sure how exactly to reword this but a couple of things are not obvious right now:

- The string needs to uniquely identify the element. Right now it's ambiguous whether we mean unique for that specific transition.
- For the cross-doc case, this sounds like any `auto` element without an id will become an exit animation since there won't be a match in the new Document. Is that the intention? If so, the string should remain unique even the document changes.

css-view-transitions-2/Overview.bs
```diff
@@ -755,7 +755,20 @@ div.box {
 
 # Nested view-transitions # {#nested-view-transitions}
 
-## Overview ## {#nested-overview}
+## Overview ## {#auto-name-overview}
```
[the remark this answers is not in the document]
noamr: Oops, erased the new title by mistake.

css-view-transitions-2/Overview.bs

css-view-transitions-2/Overview.bs
```diff
@@ -753,6 +753,66 @@ div.box {
...
+In addition to the existing values, the 'view-transition-name' also accepts an <dfn for=view-transition-name type=value export>auto</dfn> keyword.
+To resolve the [=used value=] of 'view-transition-name' for |element|:
+	1. Let |computed| be the [=computed value=] of 'view-transition-name'.
+	1. If |computed| is <css>none</css>, return null.
+	1. If |computed| is a <<custom-ident>>, return |computed|.
+	1. Assert: |computed| is <css>auto</css>.
+	1. If |element| has an associated [=Element/id=], then return |element|'s [=Element/id=].
+	1. Return a unique string, which is not a <<custom-ident>>. This string should remain consistent and unique for this element and {{Document}}, at least for the lifetime of |element|'s [=node document=]'s [=active view transition=].
+
+		Note: this string is not web-observable, and is used for addressing the element in internal algorithms.
```
khushalsagar, line 813: Maybe also leave a note that this means for cross-document view transitions, elements using a unique string will turn into exit animations.

khushalsagar (approved):

css-view-transitions-2/Overview.bs
noamr: Will do. Actually I need to make a change that keeps the `ID` check tree-scoped, so will ask for a re-review.

khushalsagar: Actually do we need any scoping text here? It looks like we do that when querying the computed value for `view-transition-name` here: https://drafts.csswg.org/css-view-transitions-1/#document-scoped-view-transition-name.

So maybe just the change the step 1 here to say, "Let |computed| be the be the element’s document-scoped view transition name." linking back to that algo?

Note to self: A bunch of places like these will need to change for scoped transitions. :)

noamr: Yes, because we read the ID. The ID might have a different tree-scope than the style where the `view-transition-name` property is defined (if it's a `::part`), we need to specifically account for that and only read the ID if it's in the same scope.

khushalsagar: Ah, the ID is a good point. Right now since there's only document scoped transitions, the "document-scoped view transition name" will be not none only if it was set via `::part`. In which case it has outer tree scope.

But the id attribute on the element should always be the inner tree scope, irrespective of how it was set. Correct me if I got it wrong (just ran it past @tabatkins). I'm assuming so because inline styles from the style attribute are also tied to the inner tree scope.

So does that imply `view-transition-name: auto` set via `::part` can never use the id based name..?

noamr: Correct, look at the note above (row 812)

khushalsagar: Got it. In that case I'm having a hard time understanding the wording of the id point. 😅 How about something like this: "If |element| has an associated [=Element/id=], and the tree scope of |computed| is the same as the tree scope of the element, then ...". Because that's what the code does right? We compare the tree scope of the node and the computed value.

And I feel like we still need to mention that we only apply the computed value if it's tree scope matches the Document? It's getting a bit ambiguous on how this interacts with the algorithm in L1 that reads the names.

noamr: Sure, that would have the same meaning. We do have the other tree scope wording in L1 already

noamr: Done

khushalsagar: > Sure, that would have the same meaning. We do have the other tree scope wording in L1 already

Yea but the way it's written right now it's not clear if it's overriding what's defined in L1. For example why are we only looking at the tree scope for ID but not the custom ident? I'm wondering if we should just phrase this as an addition to the "document-scoped view transition name" algorithm. That already deals with how we go from computed value to used value.

noamr: > > Sure, that would have the same meaning. We do have the other tree scope wording in L1 already
> 
> 
> 
> Yea but the way it's written right now it's not clear if it's overriding what's defined in L1. 

I don't understand. This is just for `auto` which is not in L1

khushalsagar: I just meant that the used value for `auto` happens after we do a tree scope check on the computed value (defined in L1) right? So should we say that somewhere here, the algorithm right now just says we're using the computed value directly. And returns it for the custom ident case.

noamr: L1 just says `Each [=view transition name=] is a [=tree-scoped name=]`. We don't change this, I don't think we need to add anything. The only special thing we do is check that scope vs the element's ID's scope if needed

noamr: Oh I see now, let me fix this

noamr: Check now

css-view-transitions-2/Overview.bs
```diff
@@ -753,6 +753,72 @@ div.box {
...
+To resolve the [=used value=] of 'view-transition-name' for |element|:
+	1. Let |computed| be the [=computed value=] of 'view-transition-name'.
+	1. If |computed| is <css>none</css>, return null.
+	1. If |computed| is a <<custom-ident>>, return |computed|.
+	1. Assert: |computed| is <css>auto</css>.
+	1. If |element| has an associated [=Element/id=], and |computed| is associated with the same [=tree/root=] as |element|'s [=tree/root=], then return the value of |element|'s [=Element/id=].
+
+		Note: this means that a ''::part()'' pseudo-element wouldn't resolve to its matching element's [=Element/id=].
+
+	1. Return a unique string. should remain consistent and unique for this element and {{Document}}, at least for the lifetime of |element|'s [=node document=]'s [=active view transition=].
```
khushalsagar, line 814:
```suggestion
	1. Return a unique string. The string should remain consistent and unique for this element and {{Document}}, at least for the lifetime of |element|'s [=node document=]'s [=active view transition=].
```

khushalsagar (approved):

css-view-transitions-2/Overview.bs
```diff
@@ -753,6 +753,72 @@ div.box {
...
+```css
+li {
+	view-transition-name: auto;
+}
+```
+</div>
+
+## Additions to 'view-transition-name' ## {#additions-to-vt-name}
+
+In addition to the existing values, the 'view-transition-name' also accepts an <dfn for=view-transition-name type=value export>auto</dfn> keyword.
```
cdoublev, line 804: It would have been useful for w3c tools, to define it with `New values` in a (partial) property definition table ([exemple](https://github.com/w3c/csswg-drafts/blob/0ab99f38873352f272ec0f430c10084f9621bbe7/css-anchor-position-1/Overview.bs#L805)).

PR #10959, opened 2024-09-26
[css-scoping] First pass at adding `:has-slotted`

LeaVerou:
In #6867 we [resolved](https://github.com/w3c/csswg-drafts/issues/6867#issuecomment-1332344714) to add `:has-slotted`. This PR is a first pass to add this.

The reference to `assignedNodes()` reference doesn’t work, not sure why. 🤷🏽‍♀️

There is [another open PR](https://github.com/w3c/csswg-drafts/pull/10586) by @keithamus but it adds `:has-slotted()` instead, which is both considerably harder to implement, and not what the resolution is about. I did however include a note that this will be included in a future version.

### DIFF
--- a/css-scoping-1/Overview.bs
+++ b/css-scoping-1/Overview.bs
@@ -512,11 +512,11 @@ Selecting Slot-Assigned Content: the ''::slotted()'' pseudo-element</h4>
 Matching on the Presence of Slot-Assigned Nodes: the '':has-slotted'' pseudo-class</h4>
 
 	The <dfn selector>:has-slotted</dfn> pseudo-class
-	matches <{slot}> elements 
+	matches <{slot}> elements
 	which have a non-empty list of <a lt="find slottables">slotted nodes</a>.
 
 	When '':has-slotted'' matches a slot with fallback content,
-	we can conclude that the fallback content is being displayed.
+	we can conclude that the fallback content is <em>not</em> being displayed.
 
 	Note: Even a single whitespace text node is sufficient to make '':has-slotted''' apply.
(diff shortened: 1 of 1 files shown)


css-scoping-1/Overview.bs
```diff
@@ -507,6 +508,25 @@ Selecting Slot-Assigned Content: the ''::slotted()'' pseudo-element</h4>
 	The only way to style assigned text nodes
 	is by styling the <a>slot</a> and relying on inheritance.
 
+<h4 id='the-has-slotted-pseudo'>
+Matching on the Presence of Slot-Assigned Nodes: the '':has-slotted'' pseudo-class</h4>
+
+	The <dfn selector>:has-slotted</dfn> pseudo-class,
+	applies to any slot element with a non-empty list of <a lt="find slottables">slotted nodes</a>.
+	It is false on any other element.
```
tabatkins, line 516:
```suggestion
	The <dfn selector>:has-slotted</dfn> pseudo-class
	matches <{slot}> elements 
	which have a non-empty list of <a lt="find slottables">slotted nodes</a>.
```

css-scoping-1/Overview.bs
```diff
@@ -507,6 +508,25 @@ Selecting Slot-Assigned Content: the ''::slotted()'' pseudo-element</h4>
...
+
+	When '':has-slotted'' matches a slot with fallback content,
+	we can conclude that the fallback content is being displayed.
```
tabatkins, line 519: I think you mean the opposite here.

tabatkins (changes requested):

css-scoping-1/Overview.bs
LeaVerou: Oops, yes!

PR #10964, opened 2024-09-27
[css-view-transitions-2] Clarify a few nesting details

### LINKED ISSUE #10633: [css-view-transitions-2] `view-transition-group` and tree-scoping
Follow up on #10334 

It's unclear how to specify `view-transition-group` in terms of shadow trees.
Sure, we can make the `<<custom-ident>>` a tree-scoped reference, which means it can't reference names from a different scope. However, what about `nearest` keyword etc? Keywords are not usually "tree-scoped".

Since `view-transition-name` is tree-scoped and `view-transition-group` doesn't make much sense in isolation from it, suggesting that both idents and keywords in `view-transition-group` are tree-scoped, and this property always resolves to `none` inside a shadow-tree.

//cc @khushalsagar @vmpstr 

See also #10529

### LINKED ISSUE #10780: [css-view-transition-2] Should non-default `view-transition-group` act like `contain`?
Right now `view-transition-group` is a bit inconsistent, because:
- `nearest` and `<custom-ident>` refer to nesting from the descendant's point of view
- `contain` refers to nesting from the ancestor's point of view.

Perhaps it would make more sense that `nearest` and `<custom-ident>` would also act as `contain`? I think the discussion in the WG [here](https://github.com/w3c/csswg-drafts/issues/10334#issuecomment-2165761156) was sort of saying it but the resolution was ambiguous.

If we do that, what happens when there's an invalid `<custom-ident>`?

cc @vmpstr @khushalsagar @fantasai, also @ydaniv that raised this concern in the last VT breakout.

noamr:
- view-transition-group maintains the tree-scoping relationship
- nearest/custom-ident act like `contain` for their descendants

Closes #10780
Closes #10633

See resolutions in the above issues.

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -45,6 +45,7 @@ spec:css-view-transitions-1;
 	text: view transition name; type: dfn;
 	text: group styles rule; type: dfn;
 	text: update pseudo-element styles rule; type: dfn;
+	text: document-scoped view transition name; type: dfn;
 spec:dom; type:dfn; text:document
 spec:css22; type:dfn; text:element
 spec:selectors-4; type:dfn;
@@ -884,24 +885,6 @@ and by applying ''view-transition-group'' to the internal element referencing th
 	The [=used value=] for 'view-transition-group' resolves to a 'view-transition-name' in its ancestor chain, or to ''view-transition-name/none''. When generating the [=named view transition pseudo-element=], the ''::view-transition-group()'' with that name
 	would be the parent of the ''::view-transition-group()'' generated for this element's 'view-transition-name'.
 
-	When the [=computed value=] of 'view-transition-name' for an element is ''view-transition-name/none'', its 'view-transition-group' [=used value=] is always resolved to ''view-transition-name/none'' as well.
-
-	<dl dfn-type=value dfn-for=view-transition-group>
-		: <dfn>normal</dfn>
-		: <dfn>contain</dfn>
-		:: The [=used value=] is the element's [=nearest containing group name=].
-
-			Note: An element with ''view-transition-group/contain'' becomes the [=nearest containing group name=] for its [=tree/descendants=].
-
-		: <dfn>nearest</dfn>
-		:: The [=used value=] is the 'view-transition-name' [=computed value=] of the nearest [=tree/ancestor=] whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none''.
-
-		: <dfn><<custom-ident>></dfn>
-		:: The [=used value=] is the given <<custom-ident>> if the element has an [=tree/ancestor=]  whose 'view-transition-name' [=computed value=] is that <<custom-ident>>, otherwise the element's [=nearest containing group name=].
-	</dl>
-
-	When the ''view-transition-group'' and ''view-transition-name'' properties are declared in a different [=tree=], ''view-transition-group'' resolves to the [=nearest containing group name=].
-
 # Algorithms # {#algorithms}
 
 ## Data structures ## {#cross-doc-data-structures}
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -820,20 +820,20 @@ and by applying ''view-transition-group'' to the internal element referencing th
 
 	When the [=computed value=] of 'view-transition-name' for an element is ''view-transition-name/none'', its 'view-transition-group' [=used value=] is always resolved to ''view-transition-name/none'' as well.
 
+	The <dfn>relevant tree</dfn> for the 'view-transition-group' property is the [=tree=] that includes the element in which the property itself is declared.
```
khushalsagar, line 823: This is trying to capture the tree scope for the computed value right? Would it be enough to say that the used value for group is "normal" if it doesn't match view transition name's scope. Because the group only applies if the name does and we have checks to only apply the name if the tree scope matches the transition's tree scope.

css-view-transitions-2/Overview.bs
```diff
@@ -1114,9 +1114,9 @@ When capturing the old or new state for an element, perform the following steps
 ### Compute the nearest containing 'view-transition-group' ### {#vt-group-nearest-contain}
 <div algorithm="nearest containing group name">
 
-To get the <dfn>nearest containing group name</dfn> for an {{Element}} |element|, perform the following steps:
+To get the <dfn>nearest containing group name</dfn> for an {{Element}} |element| and a [=tree=] |tree|, perform the following steps:
 
-	1. Let |nearestAncestorWithContain| be |element|'s nearest ancestor whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none'' and whose 'view-transition-group' [=computed value=] is ''view-transition-group/contain''.
+	1. Let |nearestAncestorWithContain| be |element|'s nearest [=tree/ancestor=] in |tree|, whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none'' and whose 'view-transition-group' [=computed value=] is not ''view-transition-group/normal''.
```
khushalsagar, line 1119: I don't think we need to pass tree here? We can ignore the group if its computed value's tree scope doesn't match view-transition-name's tree scope.

css-view-transitions-2/Overview.bs
noamr: Done

css-view-transitions-2/Overview.bs
noamr: Done

khushalsagar:
Eh, it looks like you missed publishing new changes. :)

noamr:
> Eh, it looks like you missed publishing new changes. :)

Oops, look now

css-view-transitions-2/Overview.bs
```diff
@@ -888,20 +888,20 @@ and by applying ''view-transition-group'' to the internal element referencing th
...
-		:: The [=used value=] is the 'view-transition-name' [=computed value=] of the nearest ancestor whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none''.
-
 		: <dfn>contain</dfn>
 		:: The [=used value=] is the element's [=nearest containing group name=].
 
-			Note: Descendant elements are contained within this group by default, as it would be their element's [=nearest containing group name=].
+			Note: An element with ''view-transition-group/contain'' becomes the [=nearest containing group name=] for its [=tree/descendants=].
+
+		: <dfn>nearest</dfn>
+		:: The [=used value=] is the 'view-transition-name' [=computed value=] of the nearest [=tree/ancestor=] whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none''.
```
khushalsagar, line 897: Should we use the computed or used value of the name? With auto based naming seems like it should be used.

css-view-transitions-2/Overview.bs
```diff
@@ -888,20 +888,20 @@ and by applying ''view-transition-group'' to the internal element referencing th
...
 
 		: <dfn><<custom-ident>></dfn>
-		:: The [=used value=] is the given <<custom-ident>> if the element has an ancestor whose 'view-transition-name' [=computed value=] is that <<custom-ident>>, otherwise the element's [=nearest containing group name=]
+		:: The [=used value=] is the given <<custom-ident>> if the element has an [=tree/ancestor=]  whose 'view-transition-name' [=computed value=] is that <<custom-ident>>, otherwise the element's [=nearest containing group name=].
```
khushalsagar, line 900: Same question, not sure if we need the used or computed value of v-t-name.

css-view-transitions-2/Overview.bs
```diff
@@ -888,20 +888,20 @@ and by applying ''view-transition-group'' to the internal element referencing th
 
 	<dl dfn-type=value dfn-for=view-transition-group>
 		: <dfn>normal</dfn>
-		:: The [=used value=] is the element's [=nearest containing group name=].
```
khushalsagar, line 891: We have no spec text for "normal" now. Is that intentional..?

css-view-transitions-2/Overview.bs
```diff
@@ -888,20 +888,20 @@ and by applying ''view-transition-group'' to the internal element referencing th
...
+
+		: <dfn>nearest</dfn>
+		:: The [=used value=] is the 'view-transition-name' [=computed value=] of the nearest [=tree/ancestor=] whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none''.
 
 		: <dfn><<custom-ident>></dfn>
-		:: The [=used value=] is the given <<custom-ident>> if the element has an ancestor whose 'view-transition-name' [=computed value=] is that <<custom-ident>>, otherwise the element's [=nearest containing group name=]
+		:: The [=used value=] is the given <<custom-ident>> if the element has an [=tree/ancestor=]  whose 'view-transition-name' [=computed value=] is that <<custom-ident>>, otherwise the element's [=nearest containing group name=].
 	</dl>
 
+	When the ''view-transition-group'' and ''view-transition-name'' properties are declared in a different [=tree=], ''view-transition-group'' resolves to the [=nearest containing group name=].
```
khushalsagar, line 903: Reading this line out loud, if the tree scope for the group and name on an element is different then group "resolves" to the nearest containing group name.  resolves here means used value? So basically if a named element participates in a transition in the outer scope but group was set in the inner scope, then the used group is the nearest containing group. This is the same as if the group is set to normal.

Worth a note since it's subtle, just to say this: "This means that if the group's computed value has a different tree scope than the name, the group's value is treated the same as normal".

css-view-transitions-2/Overview.bs
```diff
@@ -1182,7 +1182,7 @@ When capturing the old or new state for an element, perform the following steps
 
 To get the <dfn>nearest containing group name</dfn> for an {{Element}} |element|, perform the following steps:
 
-	1. Let |nearestAncestorWithContain| be |element|'s nearest ancestor whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none'' and whose 'view-transition-group' [=computed value=] is ''view-transition-group/contain''.
+	1. Let |nearestAncestorWithContain| be |element|'s nearest [=tree/ancestor=] whose 'view-transition-name' [=computed value=] is not ''view-transition-name/none'' and whose 'view-transition-group' [=computed value=] is not ''view-transition-group/normal''.
```
khushalsagar, line 1185: Just confirming that there's no edge case here. There are cases where we ignore element's with a name here: https://drafts.csswg.org/css-view-transitions-1/#capture-the-old-state. It's in step 8 for elements without a box or if it's c-v: hidden/auto. In all such cases the element's descendants will also be ignored so it doesn't really matter.

If that sounds good, maybe leave a note.

css-view-transitions-2/Overview.bs
noamr: Did an overhaul of this to be more algorithmic, PTAL

css-view-transitions-2/Overview.bs
noamr: See above

css-view-transitions-2/Overview.bs
noamr: See refactor

css-view-transitions-2/Overview.bs
noamr: See refactor

css-view-transitions-2/Overview.bs
```diff
@@ -1177,28 +1160,64 @@ When capturing the old or new state for an element, perform the following steps
...
+
+### Resolving the containing group name ### {#vt-containing-group-name-resolve}
+
+<div algorithm="resolve nearest containing group name">
+
+To resolve the <dfn>nearest containing group name</dfn> of an {{Element}} |element|, perform the following steps given a {{ViewTransition}} |viewTransition|:
+	1. Assert: |element| participates in |viewTransition|.
+	1. Let |ancestorGroup| be |element|'s nearest [=flat tree=] [=tree/ancestor=] who participates in |viewTransition| and whose [=document-scoped view transition group=] is not ''view-transition-group/normal''.
+	1. If |ancestorGroup| exists, return |ancestorGroup|'s [=document-scoped view transition name=].
+	1. Teturn ''view-transition-name/none''.
```
khushalsagar, line 1181:
```suggestion
	1. Return ''view-transition-name/none''.
```

khushalsagar (approved):
This is a lot better, thank you!

PR #10974, opened 2024-09-30
[cssom] Don't serialize empty rules

andruud:
With CSSOM, it's possible to reach into a CSSNestedDeclarations rule and remove its declarations, causing it to serialize as the empty string. Such rules add "whitespace junk" when serializing the outer rule, which looks unattractive.

Fixed by ignoring rules that serialize to the empty string.


emilio:
This is only relevant for the `CSSNestedDeclarations` stuff, and it's an obvious improvement / bugfix. It also only applies in edge cases (script-modified declarations).

PR #10984, opened 2024-10-01
[css-shapes-2] Clarify shape/path interpolation

### LINKED ISSUE #10740: [css-shapes-2] Interpolation between `shape()` and `path()` needs more detail
[CSS Basic Shapes 2](https://drafts.csswg.org/css-shapes-2/#interpolating-shapes)  says that `shape()` and `path()` can interpolate if they have the same set of commands, specifically:

> For this purpose, commands are "the same" if they use the same command keyword, and use the same [<by-to>](https://drafts.csswg.org/css-shapes-2/#typedef-shape-by-to) keyword. For [curve](https://drafts.csswg.org/css-shapes-2/#valdef-shape-curve) and [smooth](https://drafts.csswg.org/css-shapes-2/#valdef-shape-smooth), they also must have the same number of control points.

But there's some ambiguity here, because `shape()` takes an initial "from", and it's not clear if that has to correspond to the first Move command in the path. Specifically, it's not clear if these should interpolate:

```
shape(from 0 0, move to 200px 200px, close)
```
and 
```
path(M 100 100 z)
```

Likewise what if the first command in the path is not an absolute move?

```
shape(from 0 0, line to 200px 200px, close)
```
and 
```
path(m 100 100 z)
```

noamr:
Match the first moveTo of the `path()` with the "from" of the `shape()`

This is guaranteed to work since `path()` always starts with an M or m.

See resolution: https://github.com/w3c/csswg-drafts/issues/10740#issuecomment-2379740446

Closes #10740

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -351,7 +351,7 @@ Interpolating the ''shape()'' Function</h5>
 
 	''shape()'' and ''path()'' functions can be <l spec=css-values-4>[=interpolated=]</l> with each other
 	if their associated list of path data commands is the same length
-	and has the same commands, in order, with the first value of the ''path()'' function interpolating with the
+	and has the same commands, in order, with the first command of the ''path()'' function interpolating with the
 	initial <<coordinate-pair>> in the ''shape()'' function.
 
 		Note: The first command of a ''path()'' function is guaranteed to be a ''move'', see <a href="https://www.w3.org/TR/SVG2/paths.html#PathDataMovetoCommands">moveTo</a> in the SVG spec.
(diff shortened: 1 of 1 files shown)


css-shapes-2/Overview.bs
```diff
@@ -351,7 +351,11 @@ Interpolating the ''shape()'' Function</h5>
 
 	''shape()'' and ''path()'' functions can be <l spec=css-values-4>[=interpolated=]</l> with each other
 	if their associated list of path data commands is the same length
-	and has the same commands, in order.
+	and has the same commands, in order, with the first value of the ''path()'' function interpolating with the
```
astearns, line 354: perhaps s/value/command/, as I could naively interpret the first “value” being `m` or `M`?

noamr: Done

astearns (approved):

Issue #10908, opened 2024-09-17
[css-ui] Pseudo-elements for checkmark and dropdown icon for appearance:base `<select>`


josepharhar:
In [this issue](https://github.com/w3c/csswg-drafts/issues/10857), I proposed putting `option::before` in the UA stylesheet to create checkmarks and `select::after` to create a dropdown icon.

It was suggested that we could create new pseudo-elements for these instead like `::check` and `::select-arrow`.

I used ::before and ::after because it was simpler than creating new pseudo-elements. If we do create new pseudo-elements, then we need to at least make sure that its easy for authors to completely disable their rendering - which can be done with the ::before/::after solution by putting display:none on them.

(not sure if css-ui is the right spec for this, but maybe that depends on what solution we choose)

josepharhar:
Using ::marker was also mentioned, but I imagine it would make things harder since it has logic tied to list item rendernig that I imagine could be hard to work around.

josepharhar:
Some things I'd like to resolve on, in this order:
1. Should we create new pseudo-elements for the dropdown icon and checkmark? Presumably they would be different.
2. What would their behavior be? Can it be exactly the same as ::before and ::after, or at least have an easy opt-out for authors wanting to remove the checkmark next to options?
3. What should their names be?

josepharhar:
From an implementation point of view, I would prefer to make these like ::before and ::after because I want the content property to work on them, and if it has a backing element in the UA shadowroot then I don't think I can easily make the content property work on them easily due to the current state of chromium.

I don't know whether that means they should be "part-like" or not, but I at least want to think of them as being "before-like"

josepharhar:
I created a PR to define them as ::check and ::select-arrow here: https://github.com/w3c/csswg-drafts/pull/10986

I copied the functionality of ::before/::after and made them part-like pseudo-elements.

yisibl:
Maybe we should consider using the same model as `::select-arrow` for the `::marker` in the details element. This would avoid many of the limitations of `::marker`.

josepharhar:
> I copied the functionality of ::before/::after and made them part-like pseudo-elements.

In the PR, I changed them from part-like pseudo-elements to tree-abiding pseudo-elements which accept all properties.

josepharhar:
Proposed resolution: Add tree-abiding pseudo-elements which accept all properties called `::check` and `::select-arrow` for the option checkmark and select drop-down icon

josepharhar:
[...]

PR #10986, opened 2024-10-01
[css-forms-1] Add new pseudo-elements for customizable select

josepharhar:
This PR adds the ::check and ::select-arrow pseudo-elements so that select elements can have a standardized and customizable dropdown icon on the button and checkmarks next to options.

Fixes https://github.com/w3c/csswg-drafts/issues/10908


josepharhar:
I mostly copied the definitions from before/after. If before/after aren't actually part-like, then should I copy the text from ::backdrop that looks like this instead of saying these pseudos are part like?

> The ::backdrop pseudo-element is a tree-abiding pseudo-element. It accepts all properties. It inherits from its originating element.

css-pseudo-4/Overview.bs
```diff
@@ -1634,6 +1634,32 @@ Styling form control pickers: the ''::picker()'' pseudo-element</h3>
...
+Select dropdown icon: the ''::select-arrow'' pseudo-element</h3>
+
+  The ''::select-arrow'' pseudo-element only matches when the <a>originating
+  element</a> is a <{select}> element.
+
+  ''::select-arrow'' is a [=part-like pseudo-element=], so there is no
+  restriction on which properties apply to the ''::select-arrow'' pseudo-element.
+
+  ''::select-arrow'' generates a box as if it was an immediate child of its
+  <a>originating element</a>, preceding any boxes generated by the ''::before''
```
dbaron, line 1647: So I'm wondering if the idea that it precedes the `::before` is incompatible with being part-like.  I'm a little worried that it might be.

css-pseudo-4/Overview.bs
```diff
@@ -1634,6 +1634,32 @@ Styling form control pickers: the ''::picker()'' pseudo-element</h3>
...
+  pseudo-element, with content as specified by 'content'.
```
dbaron, line 1648: I'm also worried about allowing `content` here when it doesn't (in reality) work on elements, which part-like pseudo-elements should fully behave like.

css-pseudo-4/Overview.bs
```diff
@@ -1634,6 +1634,32 @@ Styling form control pickers: the ''::picker()'' pseudo-element</h3>
...
+<h3 id="select-arrow">
+Select dropdown icon: the ''::select-arrow'' pseudo-element</h3>
+
+  The ''::select-arrow'' pseudo-element only matches when the <a>originating
+  element</a> is a <{select}> element.
+
+  ''::select-arrow'' is a [=part-like pseudo-element=], so there is no
+  restriction on which properties apply to the ''::select-arrow'' pseudo-element.
+
+  ''::select-arrow'' generates a box as if it was an immediate child of its
```
dbaron, line 1646: I don't like the term "immediate child"; "child" is fine.  You don't need "immediate" to differentiate "child" from "descendant".

dbaron, line 1646: Are there conditions on when this generates a box?  Does it only do something for `appearance: base-select`?

dbaron (commented):
A few comments (and questions) here -- I think they apply equally to both pseudos, but I've only made them once.

css-pseudo-4/Overview.bs
```diff
@@ -1634,6 +1634,32 @@ Styling form control pickers: the ''::picker()'' pseudo-element</h3>
...
+  <a>originating element</a>, preceding any boxes generated by the ''::before''
+  pseudo-element, with content as specified by 'content'.
+
+<h3 id="check">
+Option checkmark icon: the ''::check'' pseudo-element</h3>
+
+  The '::check'' pseudo-element only matches when the <a>originating
+  element</a> is a <{option}> element.
```
lilles, line 1654: As I commented in the chromium prototype CL, should this instead say that the pseudo-element is only generated for option elements? That's an observable difference for `getComputedStyle(div, '::check')`.

josepharhar:
> I mostly copied the definitions from before/after. If before/after aren't actually part-like, then should I copy the text from ::backdrop that looks like this instead of saying these pseudos are part like?
> 
> > The ::backdrop pseudo-element is a tree-abiding pseudo-element. It accepts all properties. It inherits from its originating element.

It sounds like this is ok, so I copied the text from ::backdrop to make it not part-like anymore

css-pseudo-4/Overview.bs
josepharhar: I made it not part like anymore. Does that resolve this?

css-pseudo-4/Overview.bs
josepharhar: I made it not part like anymore. Does that resolve this?

css-pseudo-4/Overview.bs
josepharhar: I agree, done

css-pseudo-4/Overview.bs
josepharhar: I changed it to say generated instead of matches

css-pseudo-4/Overview.bs
josepharhar: Sure, I added the restriction that there must be a select element with base appearance.

fantasai (changes requested):
Please rebase this over the css-forms-1 draft. :)

josepharhar:
> Please rebase this over the css-forms-1 draft. :)

done

tabatkins:
Per the resolutions from this week's call, these should be renamed to `::checkmark` and `::picker-icon`.

Also, I thought from the discussion in the call that `::checkmark` was meant to reflect the check in a checkbox/radio as well, but this currently defines it to only exist on option elements. I presume this is just old text that doesn't reflect more recent discussion?

josepharhar:
> Also, I thought from the discussion in the call that ::checkmark was meant to reflect the check in a checkbox/radio as well, but this currently defines it to only exist on option elements. I presume this is just old text that doesn't reflect more recent discussion? Similarly, I was under the impression that ::picker-icon would apply to several input types as well, not just select.

Yes people were talking about where else these pseudos could be used in order to get better names so they could be used for more things than select and options.

I'm not exactly sure what all the cases these pseudos will be used in yet because I only have a full definition/implementation of base appearance for select and not for all the other form controls.

I changed the definition for ::picker-icon which elements to match to be more vague/generic to support other future elements.

I changed the definition of ::checkmark to apply to anything which supports :checked and has base appearance.

tabatkins:
Ah, k, if that talk about other usage sites was somewhat theoretical, and they'll be handled in a different PR (if at all), then that's fine. I was just a little confused.

PR #10996, opened 2024-10-03
[css-view-transitions-1] Don't animate transform/size by default when prefers-reduced-motion is on

### LINKED ISSUE #10267: [css-view-transitions] Users need to be able to disable view transitions
This is based on feedback from a related [TAG review](https://github.com/w3ctag/design-reviews/issues/886).

View transitions also need to be considered in #5321 and #7440.

There seems to be an attitude that view transitions are primary content under the complete control of the page author. This violates the priority of constituencies and potentially creates inaccessible content for many users.

It needs to be made clearer that view transitions are an enhancement and the web page should not break if they are disabled (and given the script component, it may be necessary to define behavior when they are disabled). The user also needs a mechanism to disable them.

noamr:
Closes #10267

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1538,6 +1538,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 								backdrop-filter: <var>backdropFilter</var>;
 							}
 						}
+
+						@media (prefers-reduced-motion) {
+							@keyframes -ua-view-transition-group-anim-<var>transitionName</var> {
+								from {
+									backdrop-filter: <var>backdropFilter</var>;
+								}
+							}
+						}
 					</pre>
 
 					Note: The above code example contains variables to be replaced.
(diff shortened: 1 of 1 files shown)


css-view-transitions-1/Overview.bs
```diff
@@ -1538,6 +1538,14 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 								backdrop-filter: <var>backdropFilter</var>;
 							}
 						}
+
+						@media (prefers-reduced-motion) {
```
khushalsagar, line 1542: The thinking here was to reset the `animation` property on the group pseudo instead of modifying the keyframe. So if the author does want to reinstate the size/position animation they'll have access to the keyframe. But I didn't realize that other properties are part of this keyframe too.

And actually more will be added with the new capture mode.

cookiecrook, line 1542: Is there a WPT test for this? Asking b/c I don't know offhand if the user preference media features can be tested in WPT. If not, merging this PR may be dependent on that test-ability requirement.

noamr: It's not currently testable. See https://github.com/web-platform-tests/wpt/issues/14839

PR #11016, opened 2024-10-10
Fix the caret-animation example

schenney-chromium:
[css-ui-4] The "see how it renders in your browser" example for caret-animation had `caret-animation: none` which is neither auto nor `manual` as the text would suggest. Fix it.


frivoal (approved):

frivoal:
Nice catch. This was a leftover from before a rename.

Issue #11023, opened 2024-10-12
[css-anchor-position] Potentially moving the `anchor()` function "Validity" section


jwatt:
The [Anchor-Based Positioning](https://drafts.csswg.org/css-anchor-position-1/#positioning) section has a bunch of subsections addressing various topics. Two of those are [The anchor() function](https://drafts.csswg.org/css-anchor-position-1/#funcdef-anchor) subsection and the [Validity](https://drafts.csswg.org/css-anchor-position-1/#anchor-valid) subsection. The latter is about the validity of the former (the `anchor()` function), but for some reason these two subsections are separated by having two other subsections placed between them (the "Centering on the Anchor" and "Taking Scroll Into Account" subsections).

I'd propose that the "Validity" subsection should be a child of the "The anchor() function" subsection, since they really seem to belong together.

If this change is made, then for consistency it would seem like the "Validity" subsection for "The [anchor-size() Function" subsection should also be indented so that the former is a child of the latter.

fantasai:
It looks like this is done. ^_^

PR #11026, opened 2024-10-12
[css-anchor-position-1] Nest Validity subsections #11023

jwatt:


fantasai:
We landed these changes at some point earlier so closing out the PR.

Issue #10784, opened 2024-08-26
[css-conditional] @container scroll-state(snapped) and snapchanged vs snapchanging


lilles:
There are a few options for when the snap target changes for scroll-state(snapped: ...) queries

1. Matches same target as the `snapchanged` event
2. Matches same target as the `snapchanging` event
3. Introduce two different snapped feature names, one for each of (1) and (2)

If we only have a single feature, I assume (2) would be the preferred behavior?

If we decide to have two features, what should their names be? `snapped` and `snapping`?

lilles:
PR landed. There is also a test: https://wpt.fyi/results/css/css-conditional/container-queries/scroll-state-snapped-snap-changing.html

PR #11031, opened 2024-10-14
[css-conditional-5] snapped matches scrollsnapchanging #10784

lilles:
Edit to clarify the resolution[1] in #10784 that scroll-state(snapped) matches the snap target for scrollsnapchanging, not scrollsnapchanged. That is, the snapped query changes evaluation during a scroll operation.

[1] https://github.com/w3c/csswg-drafts/issues/10784#issuecomment-2379901508


svgeesus (approved):
Thanks!

PR #11037, opened 2024-10-15
[cssom-view-1] Add: run snapshot post-layout state steps #10796

lilles:
Add "run snapshot post-layout state steps" as resolved in issue #10796. Let other specifications specify what they snapshot on a feature-by-feature basis.

Resolution: https://github.com/w3c/csswg-drafts/issues/10796#issuecomment-2379885032

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -2015,6 +2015,18 @@ Issue: In what order are scrollend events dispatched? Ordered based on scroll st
 </table>
 
 
+Post-Layout State Snapshotting {#post-layout-snapshot}
+======================================================
+
+Some CSS features use post-layout state, like scroll position, as input to the next style and layout update.
+
+When asked to <dfn export>run snapshot post-layout state steps</dfn> for a {{Document}} <var>doc</var>, run these steps:
+
+1. For each CSS feature that needs to snapshot post-layout state, take a snapshot of the relevant state.
+
+The state that is snapshot is defined in other specifications. These steps must not invalidate <var>doc</var> or any other {{Document}}s in such a way that other post-layout snapshotting steps can observe that such snapshotting happened. It follows that the order of which such snapshotting takes place should not matter.
+
+
 Security and Privacy Considerations {#priv-sec}
 ===============================================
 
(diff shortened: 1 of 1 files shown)


lilles:
@emilio Does this look like a reasonable approach?

HTML would then invoke this step in the appropriate place, and other specs would refer to this and say what they snapshot.

cssom-view-1/Overview.bs
```diff
@@ -2015,6 +2015,18 @@ Issue: In what order are scrollend events dispatched? Ordered based on scroll st
...
 
 
+Post-Layout State Snapshotting {#post-layout-snapshot}
+======================================================
+
+Some CSS features use post-layout state, like scroll position, as input to the next style and layout update.
+
+When asked to <dfn export>run snapshot post-layout state steps</dfn> for a {{Document}} <var>doc</var>, run these steps:
+
+1. For each CSS feature that needs to snapshot post-layout state, take a snapshot of the relevant state.
```
emilio, line 2025: This probably needs to reference `doc` somehow? Maybe "take a snapshot of the relevant state in `doc`" or so?

cssom-view-1/Overview.bs
```diff
@@ -2015,6 +2015,18 @@ Issue: In what order are scrollend events dispatched? Ordered based on scroll st
...
+
+The state that is snapshot is defined in other specifications. These steps must not invalidate <var>doc</var> or any other {{Document}}s in such a way that other post-layout snapshotting steps can observe that such snapshotting happened. It follows that the order of which such snapshotting takes place should not matter.
```
emilio, line 2027: I don't quite know what the right thing to do is for other specs to reference these kind of things, but be good to have them listed here or something actually... @tabatkins do you know what the best thing to do here is?

emilio (approved):
This looks good conceptually, a nit and few doubts about editorial-ish things... Thanks!

cssom-view-1/Overview.bs
lilles: Done

cssom-view-1/Overview.bs
tabatkins: We just need some defined term here for other specs to hook, like `<dfn export>snapshotted post-layout state</dfn>`.

And fwiw, the "should" in this line needs to be "must".

lilles: I've added the "should" -> "must" to https://github.com/w3c/csswg-drafts/pull/11056

PR #11045, opened 2024-10-17
[css-view-transitions-2] First pass at layered capture

### LINKED ISSUE #10585: [css-view-transitions-2] Optionally capture some properties (e.g. opacity/border) as style instead of snapshot
The current way we capture view-transition participating elements is tuned to a flat tree of crossfade+transform animations: we save the element's relevant properties (transform from root, mix-blend-mode, color-scheme) and everything else is baked into the snapshot.

However, #10334 (nested transition groups) makes it so that this pattern might need to be expanded: to display clipping and tree effects in an expected manner, those would have to be copied into the group rather than baked into the snapshot.

The following (non-exhaustive?) list of CSS features would have a different behavior in a nested transition tree:
- tree effects (opacity, filter, image-mask)
- clipping (clip-path, overflow, border-radius)
- 3D (actual transform, preserve-3d, perspective, transform-origin)

The "odd" one out of these is `border-radius` together with `overflow`: since it has a unique effect of clipping its descendant but not clipping the border.

To render a rounded-corner element with nested clipped descendants correctly in terms of paint-order, the entire set of box decorations (backgrounds & borders) would have to be captured and applied to the group.

Also as per the resolution in #8282, we sometimes capture geometry only (when the old element is out of the viewport).
* TBD whether this needs its own bespoke thing or fits in with the capture modes concept.

So at the very least there could be 4 capture modes (each one includes the ones above it):
- `geometry`
- `flat` (geometry + snapshot + blend)
- `composite` (clipping + 3D + tree effects)
- `layered` (borders + background + shadow)
(names bikesheddable to the bone)

Superficially, it would seem like this can be done automatically:
- Use `geometry` when out of viewport
- Use `flat` when in viewport
- Use `composite` when has nested descendants
- Use `layered` when has nested descendants and clipping border-radius

However, the issue with these capture modes is that by default they're *incompatible with each other* - so if we used one capture mode in the old state and one in the new state, the animation and the captured image would be out of sync, creating a buggy-looking animation like this: https://codepen.io/noamr-the-selector/pen/OJeymbe.

One path forward is to enable the author to define the capture mode (based on the above proposal or some other subdivision we decide on), and encourage authors to use this when using nested transitions, but not choose a capture mode automatically (except for `geometry` when out of screen).

noamr:
Describe how layered capture works, which CSS properties it targets and how it affects rendering. Based on [this resolution](https://github.com/w3c/csswg-drafts/issues/10585#issuecomment-2302433008).

This is a first pass, and still has open issues (will open separately):
1. Should there be a CSS property to decide on this? What should be the default?
2. How do we capture overflow/box-sizing?
3. (More capture questions if they come up)

Closes #10585

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -943,6 +943,7 @@ with disregard to that property:
 	- 'mask'
 	- 'opacity'
 	- 'outline'
+	- 'padding'
 
 # Algorithms # {#algorithms}
 
@@ -1397,9 +1398,7 @@ When [[css-view-transitions-1#style-transition-pseudo-elements-algorithm|updatin
 	1. Let |style| be |element|'s [=layered capture style=].
 	1. Set |capturedElement|'s [=new box properties=] to |element|'s [=layered capture properties=].
 	1. Append the [=string/concatenate|concatentation=] of « `"::view-transition-group("`, |transitionName|, `") {"`, |style| , `"}"` » to the constructed user-agent stylesheet.
-	1. Let |oldPadding| be <css>none</css> if |capturedElement|'s [=old box properties=] is null, otherwise the padding edge computed based on |capturedElement|'s [=old box properties=]'s [=layered box properties/content box=] in [=new box properties=]'s [=layered box properties/padding box=].
 	1. Let (|oldContentWidth|, |oldContentHeight|) be (|capturedElement|'s [=captured element/old width=], |capturedElement|'s [=captured element/old height=]) if |capturedElement|'s [=old box properties=] is null, otherwise  |capturedElement|'s [=old box properties=]'s [=layered box properties/content box=]'s size.
-	1. Let |newPadding| be <css>none</css> if  |capturedElement|'s [=new box properties=] is null, otherwise the padding edge computed based on |capturedElement|'s [=new box properties=]'s [=layered box properties/content box=] in [=new box properties=]'s [=layered box properties/padding box=].
 	1. Let (|newContentWidth|, |newContentHeight|) be (|width|, |height|) if [=new box properties=] is null, otherwise |capturedElement|'s [=new box properties=]'s [=layered box properties/content box=]'s size.
 	1. Append the next string (with replaced variables) to the user agent stylesheet:
 
@@ -1409,10 +1408,6 @@ When [[css-view-transitions-1#style-transition-pseudo-elements-algorithm|updatin
 					width: <var>oldContentWidth</var>;
 					height: <var>oldContentHeight</var>;
 				}
-			}
-
-			:root::view-transition-group(<var>transitionName</var>) {
-				padding: <var>newPadding</var>;
 			}
 
 			:root::view-transition-image-pair(<var>transitionName</var>) {
@@ -1427,15 +1422,6 @@ When [[css-view-transitions-1#style-transition-pseudo-elements-algorithm|updatin
 				animation-duration: inherit;
 			}
 		</pre>
-	1. Change the group keyframes rule:
-			<pre highlight="css">
-				@keyframes -ua-view-transition-group-anim-<var>transitionName</var> {
-					from {
-						/* existing properties.. */
-						padding: <var>oldPadding</var>;
-					}
-				}
-			</pre>
 
 	Note: the ':view-transition-image-pair' pseudo-element is using ''position/relative'' positioning so that the group's 'padding' will take effect.
 
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -902,7 +902,7 @@ In [[css-view-transitions-1]], the old and new states are captured as snapshots,
 creating a crossfade animation by default. This is a simple model and mostly creates the desired outcome.
 
 However, crossfading two flattened snapshots is not always the most expressive animation. CSS allows animating borders, gradient backgrounds, 'filter', 'box-shadow' etc.
-in a way that can feel more expressive and natural than crossfade.
+which can feel more expressive and natural than crossfade depending on the desired UX.
 
 In addition, when using <a href="#nested-view-transitions">nested view transitions</a>, some of the animations could look "wrong" when the CSS properties are flattened to a snapshot.
 This includes tree effects, such as 'opacity', 'mask', 'clip-path' and 'filter', and also clipping using 'overflow'. These effects are designed to apply to the whole tree of elements, not just to one element and its content.
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -1357,20 +1357,21 @@ To compute the <dfn>layered capture geometry</dfn> of an {{Element}} |element|, 
 	and whose [=layered box properties/content box=] is |element|'s [=/content box=].
 </div>
 
+To compute the <dfn>default group size</dfn> given |element|
+	return |element|'s [=/content box=]'s size if |element|'s [=computed value|computed=] 'box-sizing' is ''box-sizing/content-box'', otherwise |element|'s [=/border box=]'s size.
+
 ### Capture the old layered properties ### {#capture-new-layered-props-algorithm}
 <div algorithm="additional old capture steps (layered)">
 When [[css-view-transitions-1#capture-new-state-algorithm|capturing the old state for an element]], perform the following steps given a [=captured element=] |capturedElement| and an [=element=] |element|:
 
 	1. Set |capturedElement|'s [=old layered-capture style=] to |element|'s [=layered capture style=].
 	1. Set |capturedElement|'s [=old box properties=] to |element|'s [=layered capture geometry=].
+	1. Set |capturedElement|'s ([=old width=], [=old height=] to |element|'s [=default group size=].
 </div>
 
 ### Adjustment to image capture size ### {#capture-image-size-algorithm}
 <div algorithm="adjust image capture size (layered)">
-When [=capturing the image=] given |element|:
-	1. Let |geometry| be |element|'s [=layered capture geometry=].
-	1. If |geometry| is not null and |geometry|'s [=layered box properties/box sizing=] is ''box-sizing/content-box'',
-		then the [=natural dimensions=] of the image will be based on |geometry|'s [=layered box properties/content box=]'s size rather than on the [=border box=].
+When [=capturing the image=], the [=natural dimensions=] of the image will be based on the element's [=/content box=]'s size rather than on the [=border box=].
 </div>
 
 ### Render the snapshot with layered capture ### {#layered-capture-rendering}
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -939,15 +992,20 @@ It has the following [=struct/items=]:
...
+		: <dfn>class list</dfn>
 		:: a [=/list=] of strings, initially empty.
 
-		: <dfn for="captured element">containing group name</dfn>
+		: <dfn>containing group name</dfn>
 		:: Null or a string, initially null.
 
-		: <dfn for="captured element">transform from snapshot containing block</dfn>
-		:: A [=matrix=], initially the identity matrix.
+		: <dfn>old layered-capture style</dfn>
```
khushalsagar, line 1002: Maybe drop a link to https://github.com/w3c/csswg-drafts/issues/11078 here.

css-view-transitions-2/Overview.bs
```diff
@@ -939,15 +992,20 @@ It has the following [=struct/items=]:
...
+		:: A string.
+
+		: <dfn>transform from snapshot containing block</dfn>
+		: <dfn>old nested group transform
```
khushalsagar, line 1006: why do we need new transforms here...? we dealt with this for nesting, layered capture shouldn't need to track new transforms.

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1295,86 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+<div algorithm="additional old capture steps (layered)">
+When [[css-view-transitions-1#capture-new-state-algorithm|capturing the old state for an element]], perform the following steps given a [=captured element=] |capturedElement| and an [=element=] |element|:
+
+	1. Set |capturedElement|'s [=old layered-capture style=] to |element|'s [=layered capture style=].
+	1. Set |capturedElement|'s [=old nested group transform=] to the [=nested group transform=] given |element|.
+</div>
+
+### Render the snapshot with layered capture ### {#layered-capture-rendering}
+
+When capturing an element into a snapshot, apply the following style to the element in addition to the [=capture rendering characteristics=]:
```
khushalsagar, line 1338: I don't think this would be the right way to define this since we don't actually want to remove these styles. We could lean on the concept of "contents" defined in the contain spec: https://drafts.csswg.org/css-contain/#skips-its-contents:~:text=the%20flat%20tree%20descendants%20of%20the%20element%2C%20including%20both%20text%20and%20elements%2C%20or%20the%20replaced%20content%20of%20a%20replaced%20element to describe what should be painted when generating this snapshot. @vmpstr might have more input here.

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1295,86 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+
+<div algorithm="compute layered capture style">
+To compute the <dfn>layered capture style</dfn> of an {{Element}} |element|:
+
+	1. Let |styles| be a « ».
+	1. For each |property| in the list of [=layered capture properties=], [=list/append=] the [=string/concatenate|concatentation=] of « |property|, `":"`, |element|'s [=computed value=] of |property|, `";"` » to |styles|.
+	1. Return the [=string/concatenate|concatentation=] of |styles|.
+</div>
 
+### Capture the old layered properties ### {#capture-new-layered-props-algorithm}
```
khushalsagar, line 1328: We need to clearly define a consistent size for this snapshot. L1 defines it as the border box bounds with any painted content outside of it as ink overflow. This helps because irrespective of how big the snapshot is (based on overflow computed later in the pipeline), css properties which apply on top of image bounds (like `object-view-box`) behave consistently.

Also helps to not have to compute the snapshot size for layout.

And for layered capture, the snapshot size should probably always be content box size as far as web observable effects or layout is concerned. Since the goal is to capture the content and similar to flat capture, we can ensure that the content box origin in the image aligns with the content box origin of replaced old/new image pseudo.

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1295,86 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+	outline-color: transparent;
+}
+```
+
+Note: this style should only apply to the element while painting, it should not be web-observable via e.g. {{Window/getComputedStyle()}}.
+
+Note: we apply 'outline-color' and 'border-color' instead of 'outline' and 'border' so that layout is not affected by capturing.
+
+### Compute the nested group transform ### {#vt-compute-nested-group-transform}
+To compute the <dfn>nested group transform</dfn> given an {{Element}} |element|, return a [=matrix=] equivalent to a 2D translation of (|element|'s [=computed value|computed=] ''border-left-width'', |element|'s [=computed value|computed=] ''border-top-width'').
```
khushalsagar, line 1360: Ah, so this is for the border offset. could we cache a 2D offset instead of a transform, and just name it explicitly like `border offset`.

css-view-transitions-2/Overview.bs
noamr: Done

css-view-transitions-2/Overview.bs
```diff
@@ -885,6 +893,49 @@ and by applying ''view-transition-group'' to the internal element referencing th
...
+	- 'clip-path'
+	- 'filter'
+	- 'mask'
+	- 'opacity'
+	- 'outline'
+	- 'overflow'
+	- 'overflow-clip-margin'
+	- 'padding'
+
+	Issue: the behavior of 'overflow' needs to be clarified, especially with ''overflow/scroll'' and ''overflow-auto''. Should the scroll bars be part of the snapshot?
```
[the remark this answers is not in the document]
noamr: Added to the note, also moved the note

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1314,118 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+		then the [=natural dimensions=] of the image will be based on |geometry|'s [=layered box properties/content box=]'s size rather than on the [=border box=].
+</div>
+
+### Render the snapshot with layered capture ### {#layered-capture-rendering}
+
+When capturing an element into a snapshot, only the [=element contents=] are painted, without the element's effects and box decorations.
+Specifically, the element's 'background', 'border', 'border-image', 'box-shadow', and 'outline' are not painted, and its 'border-radius', 'clip-path', 'filter', 'mask', and 'opacity' are not applied.
+
+### Adjust the nested group transform ### {#vt-adjust-nested-group-transform}
+To <dfn>adjust the nested group transform</dfn> given a [=matrix=] |transform|, a [=matrix=] |parentTransform|, |borderBoxWidth|, |borderBoxHeight|, and a [=layered box properties=] |boxProperties|, perform the following steps:
```
khushalsagar, line 1380: nitty nit: It looks like |transform| is being passed by reference to this function which is not obvious when read the algorithm. Maybe have it return the adjusted transform?

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1314,118 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+
+	1. Set |capturedElement|'s [=old layered-capture style=] to |element|'s [=layered capture style=].
+	1. Set |capturedElement|'s [=old box properties=] to |element|'s [=layered capture geometry=].
+</div>
+
+### Adjustment to image capture size ### {#capture-image-size-algorithm}
+<div algorithm="adjust image capture size (layered)">
+When [=capturing the image=] given |element|:
+	1. Let |geometry| be |element|'s [=layered capture geometry=].
+	1. If |geometry| is not null and |geometry|'s [=layered box properties/box sizing=] is ''box-sizing/content-box'',
```
khushalsagar, line 1370: The snapshot capture is independent of box-sizing, `box-sizing` only impacts the CSS of the pseudos.  But the snapshot would always be content box size because there's no reason to capture borders as a part of it. So it wouldn't make sense for the snapshot to be larger than the element's content.

You can fold this into the "layered-capture-rendering" algorithm below.

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1314,118 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+		then the [=natural dimensions=] of the image will be based on |geometry|'s [=layered box properties/content box=]'s size rather than on the [=border box=].
+</div>
+
+### Render the snapshot with layered capture ### {#layered-capture-rendering}
```
khushalsagar, line 1374: The root would be special, doesn't make sense to allow layered capture there. File an issue to resolve on that separately.

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1314,118 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+
+When capturing an element into a snapshot, only the [=element contents=] are painted, without the element's effects and box decorations.
```
khushalsagar, line 1376: Would love to get @vmpstr's input on this phrasing too.

Taking inspiration from the wording in L1: https://drafts.csswg.org/css-view-transitions-1/#capture-the-image-algorithm:~:text=containing%20block.-,Otherwise,-%3A.
- We want to say what is painted, which is covered here.
- We want to say what the image size is internally in the UA which I think would be the union of ink overflow rectangles of all the descendants + size of replaced content..?
- And we want to define the web observable size of the image which is covered above as content box.

css-view-transitions-2/Overview.bs
```diff
@@ -1237,25 +1314,118 @@ When [[css-view-transitions-1#setup-transition-pseudo-elements-algorithm|setting
...
+
+	1. Let |style| be |element|'s [=layered capture style=].
+	1. Set |capturedElement|'s [=new box properties=] to |element|'s [=layered capture properties=].
+	1. Append the [=string/concatenate|concatentation=] of « `"::view-transition-group("`, |transitionName|, `") {"`, |style| , `"}"` » to the constructed user-agent stylesheet.
+	1. Let (|oldContentWidth|, |oldContentHeight|) be (|capturedElement|'s [=captured element/old width=], |capturedElement|'s [=captured element/old height=]) if |capturedElement|'s [=old box properties=] is null, otherwise  |capturedElement|'s [=old box properties=]'s [=layered box properties/content box=]'s size.
+	1. Let (|newContentWidth|, |newContentHeight|) be (|width|, |height|) if [=new box properties=] is null, otherwise |capturedElement|'s [=new box properties=]'s [=layered box properties/content box=]'s size.
+	1. Append the next string (with replaced variables) to the user agent stylesheet:
```
khushalsagar, line 1403: https://chromium-review.googlesource.com/c/chromium/src/+/5953705 is appending the extra styles for image-pair only if layered capture is in use. Did you mean to do the same here as well? No need to override the existing styles in image-pair in flat capture mode.

css-view-transitions-2/Overview.bs
```diff
@@ -885,6 +889,46 @@ and by applying ''view-transition-group'' to the internal element referencing th
 	The [=used value=] for 'view-transition-group' resolves to a 'view-transition-name' in its ancestor chain, or to ''view-transition-name/none''. When generating the [=named view transition pseudo-element=], the ''::view-transition-group()'' with that name
 	would be the parent of the ''::view-transition-group()'' generated for this element's 'view-transition-name'.
 
+# Layered capture # {#layered-capture}
+
+## Overview ## {#layered-capture-overview}
+
+In [[css-view-transitions-1]], the old and new states are captured as snapshots, and the initial and final geometry are captured,
```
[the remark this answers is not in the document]
noamr: Done

css-view-transitions-2/Overview.bs
noamr: Not sure I agree. A lot of operations on transforms work by reference, e.g. the [=multiply=] operation.

css-view-transitions-2/Overview.bs
noamr: Yea, but we haven't resolved yet on "when are we in capture mode". So currently the capture mode spec changes assume that we're in that mode.

css-view-transitions-2/Overview.bs
noamr: But this is the spot in L1 where we define the natural dimension of the image and how it's positioned...

css-view-transitions-2/Overview.bs
noamr: Done https://github.com/w3c/csswg-drafts/issues/11150

css-view-transitions-2/Overview.bs
khushalsagar: Eh ok, my main point was about not making this conditional based on `box-sizing`. Does that make sense?

css-view-transitions-2/Overview.bs
noamr: It's already there in the overview...

noamr (commented):
Revised `box-sizing` thing to only apply to the width/height rather than to the capture.

css-view-transitions-2/Overview.bs

Issue #9908, opened 2024-02-05
[view-transitions-2] Add method to get all animations of a ViewTransition object


bramus:
This is a side-issue to https://github.com/w3c/csswg-drafts/issues/9901 where [this demo](https://twitter.com/bramus/status/1749461824817311947) was mentioned.

To make this work, I basically need to hijack all animations and either pause them (when manually updating their `currentTime`) or give them non-`DocumentTimeline`s (when attaching them to a `ScrollTimeline`).

The code to get all these animations is this:

```js
const transition = document.startViewTransition(…);
await transition.ready;

const animations =
  document
    .getAnimations()
    .filter((anim) => {
      return anim.effect.target === document.documentElement && anim.effect.pseudoElement?.startsWith("::view-transition")
    });
```

While this works, it’s not very ergonomic. Therefore I’m suggesting something like `ViewTransition.getAnimations()` to easily get them:

```js
const transition = document.startViewTransition(…);
await transition.ready;
const animations = transition.getAnimations();
```

This would also benefit [Scoped Transitions](https://github.com/w3c/csswg-drafts/issues/9890), where you can have multiple View Transitions going on at once.

khushalsagar:
`document.documentElement.getAnimations` would make this simpler and also work with scoped transitions. Does that suffice?

I'm ok with adding it to VT as well.

bramus:
> `document.documentElement.getAnimations` would make this simpler and also work with scoped transitions. Does that suffice?

This doesn’t work because the result of this doesn’t include the animations attached to the pseudos. For `HTMLElement.geAnimations()` to return anything usable here you’d need to get the animations from the subtree by calling `document.documentElement.getAnimations({ subtree: true });`.

However, that way you would get you _all_ animations from the _entire_ subtree and you’d to manually check things again.

vmpstr:
Another approach would be something like `document.documentElement.getAnimations({ pseudo: "::view-transition", subtree: true });` which would only get the subtree of the `::view-transition` pseudo. That doesn't currently exist, but it's similar to `getComputedStyle(element, pseudo)`

bramus:
I prefer `ViewTransition#getAnimations()` over `document.documentElement.getAnimations({ pseudo: "::view-transition", subtree: true });` as with the former you can pass the `ViewTransition` object into any method and it can get the animations from it. With the latter you’d also need to pass in the element on which you triggered the VT.

[...]

PR #11050, opened 2024-10-18
[web-animations-1] Add pseudoElement to getAnimations options

vmpstr:
[web-animations-1] Add pseudoElement to getAnimations options #9908

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -5631,27 +5631,33 @@ The {{KeyframeEffect}} interface {#the-keyframeeffect-interface}
 
       On setting,
       sets the [=target pseudo-selector=] of the [=animation effect=]
-      to the provided value
-      after applying the following exceptions:
+      to the result of <dfn>pseudo-element parsing</dfn> on the provided value,
+      defined as the following:
 
-      * If the provided value is not `null`
-          and is an [=invalid selector|invalid=] <<pseudo-element-selector>>,
-          the user agent must [=throw=] a {{DOMException}}
-          with error name "{{SyntaxError}}" and leave the
-          [=target pseudo-selector=] of this [=animation effect=] unchanged.
+        1. Given the value |value|, perform the following steps:
+          1. If |value| is not `null`
+               and is an [=invalid selector|invalid=] <<pseudo-element-selector>>,
 
-          Note: Invalid in this context follows the definition of
-          an [=invalid selector=] defined in [[!SELECTORS-4]]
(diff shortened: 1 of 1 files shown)


web-animations-1/Overview.bs
```diff
@@ -6733,12 +6734,21 @@ The <code>Animatable</code> interface mixin {#the-animatable-interface-mixin}
...
   ::
-      Returns the set of [=relevant animations=] for this object,
-      or, if an {{Animatable/getAnimations(options)/options}} parameter
-      is passed with {{GetAnimationsOptions/subtree}} set to true,
-      returns the set of [=relevant animations for a subtree=] for this object.
+      1. Let |object| be the object on which this method was called.
 
-      The returned list is sorted using
+      1. If {{Animatable/getAnimations(options)/options}} is passed
+          with a non-null {{GetAnimationsOptions/pseudoElement}},
```
flackr, line 6740: We should probably have a similar SyntaxError exception thrown as we do for trying to create an animation effect targeting an invalid pseudo, e.g. see https://www.w3.org/TR/web-animations-1/#dom-keyframeeffect-pseudoelement

You can see this is also referenced by the KeyframeEffect constructor: https://www.w3.org/TR/web-animations-1/#ref-for-dom-keyframeeffect-pseudoelement, though in this case since we're not setting `pseudoElement` it probably can't point to the same reference.

vmpstr: Done. I've tried to define the algorithm in the old spot and reference it, instead of copying the text. But let me know what you prefer. Also, do you think we need a separate resolution for the DOMException, since I don't believe getAnimations can throw prior to this change

flackr: So looking at `getComputedStyle` which is very similar, it looks like MDN [says it can throw on an invalid selector](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle#exceptions) however I'm not sure where this is called out in the [spec](https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle) or how to trigger an exception in chrome, whereas for element.animate it is easy to trigger the exception with `elem.animate({}, {pseudoElement: ':invalid-pseudo'});`.

So we should probably get a consistent story for when we should throw on invalid pseudo selectors, but maybe this should be a separate issue explaining the current behavior of the APIs we have using pseudoElement selectors.

flackr: I'd propose for now we land this with the exception for self consistency within the same spec, and follow up on this.

flackr (approved):

PR #11056, opened 2024-10-21
[css-conditional-5] Link scroll-state snapshotting to cssom-view #10796

lilles:
scroll-state() snapshotting should happen as part of "run snapshot post- layout state steps".

Resolution: https://github.com/w3c/csswg-drafts/issues/10796#issuecomment-2379885032


lilles:
@tabatkins Emilio left a question for you whether the specs snapshotting state should be called out explicitly in the cssom-view spec, which is relevant for this review: https://github.com/w3c/csswg-drafts/pull/11037#discussion_r1801511393

PR #11070, opened 2024-10-22
[css-view-transitions-2] Allow transitions when traversing into a document that was created using cross-origin redirects

### LINKED ISSUE #11063: [css-view-transitions-2] Using 'was created via cross-origin redirects' Document creation state causes unexpected results
If you navigate into a site via a cross-origin redirect, then the 'was created via cross-origin redirects' bit gets set on the Document.

If you then navigate to a different page within the same site, the navigation is eligible for cross document view transition.

If you then navigate back to the original page, and this uses the BFCache, then this second navigation is not eligible, since the Document is still marked as having been created by a cross-origin redirect.

This seems unexpected to me, as the cross-origin redirect during navigation is unrelated to the current navigation.

noamr:
This restriction only makes sense for the initial navigation, not for traversal.

Closes #11063

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -992,9 +992,11 @@ The [=captured element=] struct should contain these fields, in addition to the 
 
 	1. If |oldDocument|'s [=Document/origin=] is not [=same origin=] as |newDocument|'s [=Document/origin=], then return false.
 
-	1. If |navigationType| is {{NavigationType/traverse}}, then return true.
+	1. If |newDocument| [=was created via cross-origin redirects=] and |newDocument|'s [=latest entry=] is null, then return false.
 
-	1. If |newDocument| [=was created via cross-origin redirects=], then return false.
+		Note: A [=Document=]'s [=latest entry=] would be null if this is a new navigation, rather than a restore from BFCache.
+
+	1. If |navigationType| is {{NavigationType/traverse}}, then return true.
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -992,10 +992,10 @@ The [=captured element=] struct should contain these fields, in addition to the
 
 	1. If |oldDocument|'s [=Document/origin=] is not [=same origin=] as |newDocument|'s [=Document/origin=], then return false.
 
-	1. If |newDocument| [=was created via cross-origin redirects=], then return false.
-
 	1. If |navigationType| is {{NavigationType/traverse}}, then return true.
```
khushalsagar, line 995: If it's a non-BFCache traverse navigation, then the cross-origin restriction should still apply. It's no different from a push navigation then.

That said, we might have been overly conservative limiting navigations with cross-origin redirects. I don't quite remember the rationale other than "there could be an information leak and it's a rare use-case". So also happy to re-open that conversation if you'd like.

mattwoodrow: Does the HTML spec's wording for the page swap event work here? https://html.spec.whatwg.org/multipage/browsing-the-web.html#updating-the-traversable:latest-entry

noamr: Yes, I was thinking the same. Amended the PR

khushalsagar (approved):

PR #11073, opened 2024-10-22
[cssom-view-1] Apply scroll snap to scroll-into-view-position

flackr:
When determining the scroll into view position of an element, the UA should account for defined scroll snap positions. These are applied when scrolling due to https://www.w3.org/TR/css-scroll-snap-1/#choosing added in https://github.com/w3c/csswg-drafts/commit/8401d613596842b7ab337e1a3190a771601240aa however applying them here ensures that specs using the scroll into view position are snapping aware.

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1496,6 +1496,10 @@ run the following steps:
             <dt>If <var>element edge D</var> is outside <var>scrolling box edge D</var> and <var>element width</var> is less than <var>scrolling box width</var>
             <dd>Align <var>element edge D</var> with <var>scrolling box edge D</var>.
         </dl>
+    1. If <var>target</var> is an <a for="/">Element</a>, and the target element defines some <a>scroll snap positions</a>,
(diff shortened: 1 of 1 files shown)


flackr:
@fantasai I think this should address the concern you had that the scroll-into-view position was not snap aware. Does this look reasonable?

flackr:
@tabatkins would you be able to take a look and see if this looks reasonable?

cssom-view-1/Overview.bs
```diff
@@ -1496,6 +1496,10 @@ run the following steps:
             <dt>If <var>element edge D</var> is outside <var>scrolling box edge D</var> and <var>element width</var> is less than <var>scrolling box width</var>
             <dd>Align <var>element edge D</var> with <var>scrolling box edge D</var>.
         </dl>
+    1. If <var>target</var> is an <a for="/">Element</a>, and the target element defines some <a>scroll snap positions</a>,
```
tabatkins, line 1499:
```suggestion
    1. If <var>target</var> is an [=/Element=], and the target element defines some <a>scroll snap positions</a>,
```

Issue #10271, opened 2024-04-30
[css-color-hdr] Add interpolation between multiple values of dynamic-range-limit


ccameron-chromium:
This came up when [writing WPT tests](https://chromium-review.googlesource.com/c/chromium/src/+/5501416) for `dynamic-range-limit`.

The `dynamic-range-limit` [specification](https://drafts.csswg.org/css-color-hdr/#the-dynamic-range-limit-property) allows for interpolation between two `dynamic-range-limit` values by the `dynamic-range-limit-mix` function. It was pointed out that interpolation between multiple different values may happen implicitly during animation.

The syntax chosen in the spec lends itself to fixing this issue fairly simply. The current syntax is
```
  dynamic-range-limit-mix() = dynamic-range-limit-mix( [ <ident> && <percentage [0,100]>? ]#{2})
```

The suggestion that I was given was to change this to
```
  dynamic-range-limit-mix() = dynamic-range-limit-mix( [ <ident> && <percentage [0,100]> ]+)
```

**Example**: Halfway between `standard` and `high`.
```
  dynamic-range-limit-mix(standard 50%, high 50%)
```
This is the same as the existing syntax.

**Example**: One-fifth between `constrained-high` and `dynamic-range-limit-mix(standard 40%, high 60%)`:
```
  dynamic-range-limit-mix(constrained-high 20%,  dynamic-range-limit-mix(standard 40%, high 60%) 80%)
```
this simplifies to
```
  dynamic-range-limit-mix(standard 32%, high 48%, constrained-high 20%)
```

**Notes**

In the syntax we didn't allow percentages to be omitted (cause it can get a bit difficult to define more than one unspecified value, unlike `color-mix` which has natural definitions).

If the sum of the arguments' percentages is `0%`, then the function fails.

The arguments' percentages are normalized so that they add to `100%`.

The percent for `standard`, `high`, and `constrained-high` is computed as the of the weighted sum of its percents in each argument.

I'll try to put together a patch that does this.

tabatkins:
Actually, I think it would be preferable to switch the `dynamic-range-limit-mix()` function's grammar to match that of the [`mix-*()` family of functions](https://drafts.csswg.org/css-values-5/#calc-mix) (and ideally switch the name order as well, to `mix-dynamic-range-limit()`).

In other words:

```
mix-dynamic-range-limit( <progress>, <'dynamic-range-limit'>, <'dynamic-range-limit'> )

sboukortt:
[...]

PR #11086, opened 2024-10-25
[css-color-hdr] Update dynamic-range-limit-mix to reflect #10271.

ccameron-chromium:
Update the definition of dynamic-range-limit-mix to facilitate interpolations of interpolations. Add explicit algorithm for the computed value when the specified value is dynamic-range-limit-mix.


svgeesus (approved):
LGTM

PR #11088, opened 2024-10-25
[css-values-5] Allow a single `<syntax-component>`

cdoublev:
The current syntax requires two or more syntax components, which is probably not intentional.


tabatkins:
Since I just touched that line for a different patch, I committed the fix myself. Thanks, that was a silly error. ^_^

PR #11106, opened 2024-10-29
[css-text-4] use "see individual properties" terminology for whitespace property

keithamus:
The `whitespace` property includes `Inherited: individual properties`, but it seems to be the only instance of this value in all of the specs. Lots of other property uses `Inherited: see individual properties`. I presume `whitespace` should use this as well, so I've made this small editorial PR to correct this.


astearns (approved):

songjianet (approved):

kojiishi (approved):

kojiishi:
Thank you for the PR.

Issue #10393, opened 2024-06-05
Proposal: CSS Gap Decorations Level 1


kbabbitt:
[Explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSGapDecorations/explainer.md)

This is a fresh take on a topic that has been brought up previously in discussions such as #2748 and #6748. Authors have long been asking for ways to paint gap decorations in grids and other containers.

## Why a new proposal?

The core of this proposal is similar to ones we've seen previously: We introduce `row-rule-width`, `row-rule-style`, and `row-rule-color` as complements to `column-rule-width`, `column-rule-style`, and `column-rule-color` and apply both sets of properties to grid, flex, and masonry containers. But one reason for the fresh take is to address use cases such as [this one](https://github.com/w3c/csswg-drafts/issues/2748#issuecomment-595889781) that call for varying gap decorations over a single container, or a subregion of a container. To do that, I've drawn inspiration from grid area syntax and the parallel lists + list cycling behavior in CSS Transitions and Animations to allow things like:

```css
.alternate-heavy-light {
  display: grid;
  grid-template: repeat(auto-fill, 30px) / repeat(3, 100px);
  grid-gap: 10px;
  row-rule: 2px solid black / 1px solid lightgray;
}
```
![Example grid with alternating heavy and light row separator](https://github.com/MicrosoftEdge/MSEdgeExplainers/raw/main/CSSGapDecorations/images/example-heavy-light.png)

```css
.grid-multiple-decoration-areas {
  display: grid;
  grid-template-rows: [top] 30px [main-top] repeat(6, 30px) [bottom];
  grid-template-columns: [left] 100px [main-left] repeat(3, 100px) [right];
  grid-gap: 10px;
  grid-row-rule-area: left / top / main-left / bottom,
                      main-left / main-top / right / bottom;
  row-rule: 1px solid lightblue,
            1px solid black;
  grid-column-rule-area: main-left / top / main-left / bottom;
  column-rule: 1px solid lightblue;
}
```
![Example grid with varying gap decorations](https://github.com/MicrosoftEdge/MSEdgeExplainers/raw/main/CSSGapDecorations/images/example-multiple-areas.png)

Applying to the specific use case I cited above:

```css
.container {
  gap-rule-style: solid:
  gap-rule-color: lightgray;
  column-rule-width: 1px repeat(auto, 2px) 1px;
  row-rule-width: 0px repeat(auto, 2px 1px);
[...]

PR #11115, opened 2024-10-30
[css-gaps-1] Initial editor's draft. #10393

kbabbitt:
Major changes from the [unofficial draft](https://kbabbitt.github.io/css-gap-decorations/):

- Defer gap decoration areas out of Level 1
- Drop the direction-relative properties
- Generalize `*-rule-break` to apply to all container types
- Remove the `join` keyword from `*-rule-outset` and fold its behavior into `*-rule-break`
- Define algorithms for:
  - Determining gap decoration endpoints
  - Expanding repeaters in color/style/width properties


alisonmaher:
> Drop the direction-relative properties

I can't remember, was this based on feedback from the working group?

kbabbitt:
Formatted version of draft can be found here: https://kbabbitt.github.io/css-gap-decorations/pr-11115/Overview.html

kbabbitt:
> > Drop the direction-relative properties
> 
> I can't remember, was this based on feedback from the working group?

It was based on feedback from the issue discussion.

https://github.com/w3c/csswg-drafts/issues/10393#issuecomment-2161008801
> * (possibly?) add main-rule and cross-rule for Flexbox (personally I'd prefer to just use row-rule and column-rule)

https://github.com/w3c/csswg-drafts/issues/10393#issuecomment-2161126754
> Me too. There is no main-gap or cross-gap property, so I don't think we need main-rule or cross-rule either.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+Layout and painting {#layout-painting}
+--------------------------------------
+
+	This specification does not generally define the geometry of gaps in container layouts;
+	that is left to other specifications.
+	See [[CSS-ALIGN-3#gaps]] for definitions pertaining to
+	<a>multicol containers</a>, <a>flex containers</a>, and <a>grid containers</a>.
+
+	<a>Gap decorations</a> do not take up space.
+	That is, the presence or thickness of a <a>gap decoration</a> will not alter the placement of anything else.
```
KurtCattiSchmidt, line 89: Is "thickness" the best word here? "dimensions" seems more generic. Or should you use "width" to better align with the properties defined below?

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	If a <a>gap decoration</a> is wider than its gap,
+	any adjacent boxes will overlap the decoration,
+	and in some cases the decoration may possibly extend outside the box of the container.
```
KurtCattiSchmidt, line 92: Nit: "possibly" seems unnecessary

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	<a>Gap decorations</a> are painted just above the border of the container.
+	For scrollable containers,
+	note that while the border and background of the container obviously aren't scrolled,
```
KurtCattiSchmidt, line 95: Nit: "obviously" seems unnecessary

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	the decorations need to scroll along with items in the container.
+
+	A <dfn>gap intersection point</dfn> exists in each of the following locations:
+	<ul>
+		<li>The center of an intersection between a gap and the content edge of the container.
```
KurtCattiSchmidt, line 100: These should have closing `</li>`'s

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	in the center of the corresponding gap and parallel to its edges.
+
+Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break', and 'gap-rule-break' properties {#break}
+---------------------------------------------------------------------------------------------------------------------------
+
+	<pre class='propdef'>
+		Name: column-rule-break, row-rule-break
+		Value: ''intersection'' | ''spanning-item'' | ''none''
+		Initial: ''spanning-item''
+		Applies to: <a>grid containers</a>, <a>flex containers</a>, <a>multicol containers</a>, and <a>masonry containers</a>
```
KurtCattiSchmidt, line 114: Do these need `href`'s?

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+		Inherited: no
+		Percentages: refer to the <a>crossing gap width</a>
+		Animation type: by computed value type
+	</pre>
+
+	These properties can be used to offset the endpoints of <a>gap decorations</a> relative to the
+	<a>gap intersection points</a> which would normally determine their endpoints.
+	The initial value of ''50%'' places the <a>gap decoration</a> endpoint in the center of the intersection.
+	With a value of ''0'', the <a>gap decoration</a> endpoint will coincide with the edge of the intersection.
+	Negative values may be used to recede from the intersection.
```
KurtCattiSchmidt, line 224: Should this indicate direction? "...may be used to recede from the intersection in the negative direction", or is that implied with "recede"?

KurtCattiSchmidt (changes requested):
Looks great overall, mostly just nits

css-gaps-1/Overview.bs
kbabbitt: Nope, Bikeshed will generate them automatically, see https://speced.github.io/bikeshed/#autolinking

alisonmaher (commented):
Overall, looks good to me, as well, with a few suggestions below.

css-gaps-1/Overview.bs
alisonmaher: +1

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	For readability they have not been repeated explicitly.
+
+Gap decorations {#gap-decorations}
+==================================
+
+	Various layouts in CSS such as
+	<a>multicol containers</a>, <a>flex containers</a>, <a>grid containers</a>, and <a>masonry containers</a>
+	position child boxes adjacent to each other with gaps, also known as <a>gutters</a>, between them.
+	Each such gap may contain a <dfn>gap decoration</dfn> which authors
+	may use to provide a visual separator between adjacent boxes.
```
alisonmaher, line 49: Is "visual separator" the right terminology if `gap` is defining the separation. Maybe something that notes it is used to style the gaps instead of as the separator itself?

If so, it might make sense to have a link to the definition of `gap` here (but maybe that wouldn't make sense since I don't think multicol has a `gap` property)

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+<pre class='anchors'>
+urlPrefix: https://drafts.csswg.org/css-grid-3/; type: dfn; spec: CSS-GRID-3; text: masonry container
+</pre>
+
+Introduction {#intro}
+=====================
+
+	<em>This section is not normative.</em>
+
+	[[CSS-MULTICOL-1#column-gaps-and-rules]] allows for rules to be drawn
```
alisonmaher, line 22: Should we note that this will eventually be deprecated in favor of this specification?

css-gaps-1/Overview.bs
kbabbitt: For context, this paragraph was copied verbatim from https://drafts.csswg.org/css-multicol-1/#column-gaps-and-rules. But I agree with this and the next couple of suggested changes to the wording, so I'll go ahead and make them.

css-gaps-1/Overview.bs
kbabbitt: Added throughout.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+		Inherited: no
+		Animation type: discrete
+	</pre>
+
+	Sets the paint order for <a>gap decorations</a> in two-dimensional containers.
+
+Color, style, and width {#color-style-width}
+============================================
+
+Lists of values and the ''repeat()'' notation {#lists-repeat}
```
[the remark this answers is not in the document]
kbabbitt: https://drafts.csswg.org/css-grid-1/#grid-definition places the repeat() definition after the grid-template-* properties but before the shorthands. Changed the ordering in this spec to match.

css-gaps-1/Overview.bs
kbabbitt: Changed the wording a bit to clarify. In short: a negative value moves the endpoint away from the center of the intersection.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+		<dt>
+			At the content edge of the container
+		<dd>
+			The <a>crossing gap width</a> is ''0''.
+
+		<dt>
+			At an intersection with another gap
+		<dd>
+			The <a>crossing gap width</a> is the used value of the 'column-rule-width'
+			or 'row-rule-width' property, whichever applies to the other gap.
```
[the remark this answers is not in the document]
kbabbitt: Changed to "the intersecting gap" - is that clearer?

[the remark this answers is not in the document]
kbabbitt: I could also change the `<dd>` above to "At an intersection with a perpendicular gap" and then change this to "the perpendicular gap." (I was trying to leave room for possible future layouts that have gaps in directions other than vertical or horizontal - for example a hexagonal layout - but could set that aside for now if "perpendicular" would be clearer.)

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	</dl>
+
+	To <dfn>compute the offset</dfn> for a given <a>gap intersection point</a> within a given gap:
```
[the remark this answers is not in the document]
kbabbitt: Yes, a given gap will have multiple gap intersection points - at least 2 at each edge of the container, plus 1 anywhere there's an intersection with another gap.

What I'm trying to communicate on this line is that the specific gap intersection point we're talking about is an input to the algorithm below. I'll go back through the algorithms and assign names to the inputs/given values using `|...|` to make this clearer.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+			Let |subset| be a list that initially contains only the item removed from |endpoints|.
+
+		<li>
+			Remove the first item from |endpoints| and add it to the end of |subset|.
+
+		<li>
+			If |endpoints| is non-empty, and either of the following conditions is true...
+			<ol>
+				<li>
+					|break| is ''column-rule-break/spanning-item'',
```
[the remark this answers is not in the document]
kbabbitt: What I'm trying to do here is formalize how we test for "a child item intersects with / interrupts the gap". I'll see if I can come up with some visuals to illustrate it.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+
+		<li>
+			If |endpoints| contains fewer than 2 items, return |pairs|.
+
+		<li>
+			Remove the first item from |endpoints|.
+			Let |subset| be a list that initially contains only the item removed from |endpoints|.
+
+		<li>
+			Remove the first item from |endpoints| and add it to the end of |subset|.
```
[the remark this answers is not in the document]
kbabbitt: It's intentionally in both, though reading the algorithm again I realized I could make it clearer with some edits.

css-gaps-1/Overview.bs
kbabbitt: > Is "visual separator" the right terminology if gap is defining the separation. Maybe something that notes it is used to style the gaps instead of as the separator itself?

Re-worded this a bit; let me know what you think.

> If so, it might make sense to have a link to the definition of gap here (but maybe that wouldn't make sense since I don't think multicol has a gap property)

The "gutter" link is intended to serve this purpose - this is the term defined in css-align-3 and used in the multicol and grid specs.

css-gaps-1/Overview.bs
kbabbitt: Yes, but probably not here since this section is not normative. Added a sentence to the section with the normative property definitions.

kbabbitt:
@alisonmaher @KurtCattiSchmidt Thanks for the comments! I've pushed some changes with updates to the text; will work on making more visuals.

css-gaps-1/Overview.bs
kbabbitt: Done

css-gaps-1/Overview.bs
kbabbitt: Done

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+		Applies to: Same as 'column-rule-outset' and 'row-rule-outset'
+	</pre>
+
+	Sets the 'column-rule-outset' and 'row-rule-outset' properties to the same value.
+
+Gap decoration paint order: The 'gap-rule-paint-order' property {#paint-order}
+------------------------------------------------------------------------------
+	<pre class='propdef'>
+		Name: gap-rule-paint-order
+		Value: row-over-column | column-over-row
```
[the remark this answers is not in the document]
kbabbitt: Added.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	The initial value of ''50%'' places the <a>gap decoration</a> endpoint in the center of the intersection.
+	With a value of ''0'', the <a>gap decoration</a> endpoint will coincide with the edge of the intersection.
+	Negative values may be used to recede from the intersection.
+	These offsets also apply at the edges of the container,
+	where positive values may extend beyond the content bounds of the container.
+
+	<div class="example">
+		The following examples illustrate adjustment of gap decoration endpoints using the 'column-rule-outset' property.
+		<dl>
+			<dt><pre>column-rule-outset: 0px;</pre>
```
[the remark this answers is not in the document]
kbabbitt: Added.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+
+	<a>Gap decorations</a> are painted relative to pairs of <a>gap intersection points</a>,
+	in the center of the corresponding gap and parallel to its edges.
+
+Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break', and 'gap-rule-break' properties {#break}
+---------------------------------------------------------------------------------------------------------------------------
+
+	<pre class='propdef'>
+		Name: column-rule-break, row-rule-break
+		Value: ''intersection'' | ''spanning-item'' | ''none''
```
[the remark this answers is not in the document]
kbabbitt: Added.

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,601 @@
...
+	That is, the presence or thickness of a <a>gap decoration</a> will not alter the placement of anything else.
+	If a <a>gap decoration</a> is wider than its gap,
+	any adjacent boxes will overlap the decoration,
+	and in some cases the decoration may possibly extend outside the box of the container.
+	<a>Gap decorations</a> are painted just above the border of the container.
+	For scrollable containers,
+	note that while the border and background of the container obviously aren't scrolled,
+	the decorations need to scroll along with items in the container.
+
+	A <dfn>gap intersection point</dfn> exists in each of the following locations:
```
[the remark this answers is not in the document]
kbabbitt: Added a visual and tweaked the wording slightly.

kbabbitt:
Now updated with more visuals, and I've pushed an update to the formatted copy at https://kbabbitt.github.io/css-gap-decorations/pr-11115/Overview.html

Ready for another look, thanks!

css-gaps-1/Overview.bs
```diff
@@ -0,0 +1,785 @@
...
+	</div>
+
+	<div class="example">
+		<pre>
+			.break-spanning-item {
+				gap-rule-break: spanning-item;
+				grap-rule-outset: 0px;
+			}
+		</pre>
+		<img src="images/example-break-spanning-item.png">
```
kbabbitt, line 283: Just realized there's an error in this image: The gap decorations should not be drawn through the intersection neighbored by 4, 6, 7, and 8. Will fix.

kbabbitt: Hmm, actually never mind, as defined by the algorithm, the decorations *should* be drawn through that intersection.

css-gaps-1/Overview.bs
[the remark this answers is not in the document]
kbabbitt: Added an explanatory note and a visual.

alisonmaher (approved):
Changes lgtm, thanks for adding all the visuals. This helped clarify all the questions I had.

The only other thoughts I had were if we should add an open issue for how to handle fragmentation and how to handle subgrid?

Otherwise, this looks good to me, thanks Kevin!

kbabbitt:
@fantasai / @tabatkins would very much appreciate any feedback you have on this draft, could you please take a look? Thanks!

PR #11119, opened 2024-10-30
[CSS-UI-4] Clarify which element caret properties come from

schenney-chromium:
The spec is not clear as to which element's properties should be used to style the caret. Try to clarify that.

I'm not sure this is the best way to do it, but the question came up when discussing Mozilla's position on the caret-animation property. Suggestions for alternate prose are welcome.


schenney-chromium:
Ping!

frivoal:
@schenney-chromium Sorry, I had missed this PR, and created another one that does something similar (and then some): https://github.com/w3c/csswg-drafts/pull/11373/

Are you OK with closing this one in favor in the other, due to effectively being a subset?

schenney-chromium:
Yes. I'm happy to close this. Thanks,.

PR #11135, opened 2024-11-01
[css-nesting-1] Specificity of '&' is zero without a parent rule

### LINKED ISSUE #10196: [css-cascade-6] Specificity of Implicitly-Added `:scope` in Scoped Rules
Current [WPT test](https://github.com/web-platform-tests/wpt/blob/master/css/css-cascade/scope-specificity.html) for scoped styles' specificity seems to assume that an unscoped rule has the same specificity as the identically written out scoped rule.
However, the [spec's example](https://drafts.csswg.org/css-cascade-6/#example-66991251) seems to imply that `:scope <descendant-combinator>` is implicitly added, and that `:scope`'s specificity is equal to that of other pseudo-classes (i.e. class-like +1).

WebKit and Blink are shipping the [behaviour matching the WPT](https://wpt.fyi/results/css/css-cascade/scope-specificity.html?label=experimental&label=master&aligned).

This leads to behaviours where functionally-identical selectors with explicitly-added `:scope` selector wins despite the ordering. In the example below, WebKit and Chrome both show `.foo` being purple:

```
<!doctype html>
<style>
.foo { color: green }
@scope (:root) {
  :scope .foo { color: purple }
  .foo { color: blue }
}
</style>
<div class="foo">Color?</div>
```

cc: @emilio, @mdubet, @andruud

andruud:
Fixes #10196.

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -966,7 +966,8 @@ Nesting Selector: the ''&'' selector {#nest-selector}
 	The <a>specificity</a> of the <a>nesting selector</a>
 	is equal to the largest specificity among the complex selectors
 	in the parent style rule's selector list
-	(identical to the behavior of '':is()'').
+	(identical to the behavior of '':is()''),
+	or zero if no such selector list exists.
 
 	<div class="example">
 		For example, given the following style rules:


argyleink (approved):

andruud:
@argyleink Thanks! Now please click the merge button, I don't have access.

PR #11137, opened 2024-11-01
[css-viewport-1] Add automation support for viewport segments

darktears:
This commits define ways to emulate display features of a device. A display feature is a hardware feature that acts as a divider and creates logically separate regions of the viewport.

Using WebDriver developers and user agents will have the opportunity to test or emulate several types of devices.


bramus:
I don’t think I am the right person to co-review this. I would assume the authors of the https://w3c.github.io/webdriver/ spec are the people you are looking for.

OrKoN:
Thanks for the ping, let me try to review it:

1) Is there a rendered version that includes changes?
2) have you considered specifying it as part of https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport instead? 
3) is anything expected to happen when `[=[[DisplayFeaturesOverride]]=]` is set? currently, it seems the override is set for the future use by the algorithm and the command returns?

darktears:
@OrKoN

1. If by rendered, you mean after bikeshed is run, not yet. I can find a way to throw the generated html somewhere for you to see.
2. No but I'm wondering if it's a good fit. setViewport command only allows to set the width/height and devicePixelRatio but do not seem to include any other viewport properties. If we decide to add a parameter in setViewport then where would all the explanations and diagrams go?
3. I don't understand your question.  Setting [=[[DisplayFeaturesOverride]]=] is the very purpose of this. When the override is set then the `segments` property is recalculated using the array of display features from DisplayFeaturesOverride slot coming from the setdisplayfeatures command.

OrKoN:
> If by rendered, you mean after bikeshed is run, not yet. I can find a way to throw the generated html somewhere for you to see.

Yes, usually https://github.com/w3c/spec-prod is used to publish a preview but looks like it is not set up in this repo. 

> No but I'm wondering if it's a good fit. setViewport command only allows to set the width/height and devicePixelRatio but do not seem to include any other viewport properties. If we decide to add a parameter in setViewport then where would all the explanations and diagrams go?

I assume it works the same way (it is also Bikeshed-based), in a non-normative section, with files being put to a folder. Note that I am not saying it is required.

> I don't understand your question. Setting [=[[DisplayFeaturesOverride]]=] is the very purpose of this. When the override is set then the segments property is recalculated using the array of display features from DisplayFeaturesOverride slot coming from the setdisplayfeatures command.

Probably I made an assumption that is not correct. If this command is only meant to override the segments getter available to the page, the current steps look sufficient. Or does the command imply a way to emulate the viewport with segments visually? Is the command expected to also trigger relevant Screen Orientation, Window Resize or Posture change events mentioned in the spec or any other events? And to confirm the DOM attribute cannot be overriden from JavaScript for testing?

css-viewport-1/Overview.bs
```diff
@@ -119,6 +124,22 @@ property for the first <code class=html>&lt;BODY&gt;</code> element of
 an HTML or XHTML document. For other document types, it is the
 computed 'direction' for the root element.
 
+<h3 id="display-feature"><dfn>Display feature</dfn> definition</h4>
```
OrKoN, line 127:
```suggestion
<h3 id="display-feature"><dfn>Display feature</dfn> definition</h3>
```

css-viewport-1/Overview.bs
```diff
@@ -554,10 +575,12 @@ Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in C
 
 Additonal details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
 
-The {{segments}} attribute must run these steps:
+The {{segments}} attribute getter steps are:
 1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return null.
-2. Returns null if there is only a single viewport segment and abort these steps.
-3. Otherwise, return the {{Viewport}}'s [[css-env-1#viewport-segments|segments]] array.
+2. Let |topLevelTraversable| be |document|'s [=relevant global object=]'s [=/navigable=]'s [=browsing context/top-level traversable=].
```
OrKoN, line 580: |document| is not defined

css-viewport-1/Overview.bs
```diff
@@ -554,10 +575,12 @@ Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in C
...
+3. If |topLevelTraversable|.[=[[DisplayFeaturesOverride]]=] is non-null, return {{Viewport}}'s [[css-env-1#viewport-segments|segments]] array calculated from |topLevelTraversable|.[=[[DisplayFeaturesOverride]]=].
+4. Returns null if there is only a single viewport segment and abort these steps.
```
OrKoN, line 582:
```suggestion
4. Return null if there is only a single viewport segment and abort these steps.
```

OrKoN:
That sounds good.The PR looks good to me % minor comments. 
Do you plan to add WPTs for these WebDriver methods in https://github.com/web-platform-tests/wpt/tree/master/webdriver/tests/classic?

OrKoN (approved):
LGTM

darktears:
> That sounds good.The PR looks good to me % minor comments. Do you plan to add WPTs for these WebDriver methods in https://github.com/web-platform-tests/wpt/tree/master/webdriver/tests/classic?

I intend to do something similar to https://github.com/web-platform-tests/wpt/tree/master/device-posture and add tests to verify the `segments` property, CSS MQs and environment variables.

OrKoN:
It would be also nice to add WebDriver-specific tests to test error conditions, wrong input, some happy path cases. Some browsers might not actually implement testdriver.js using WebDriver so it is also a good indication for external users for the level of support in the corresponding driver.

darktears:
> It would be also nice to add WebDriver-specific tests to test error conditions, wrong input, some happy path cases. Some browsers might not actually implement testdriver.js using WebDriver so it is also a good indication for external users for the level of support in the corresponding driver.

I can investigate that.

emilio:
Is there a precedent to put these kind of spec text _inside_ the CSS spec? I haven't seen it before... cc @fantasai @tabatkins

OrKoN:
I have not seen a precedent with CSS specs but here other specs defining WebDriver extension commands https://dontcallmedom.github.io/webdex/e.html#extension%20commands%40%40webdriver%25%25dfn (not sure if the tool omits some specs)

darktears:
@emilio what you want me to land for editorial improvements? 

The DisplayFeature part with its relevant diagram? The concept is really linked to the automation part.

For sure the tab fix on the IDL of the segments property? Anything else?

darktears:
The CSS WG has resolved that WebDriver extensions should go in their respective specs. I'd love a review and approval. Thanks!

Etilmetilketilvinil:
Lemeztelen=zabi

css-viewport-1/Overview.bs
```diff
@@ -572,6 +595,188 @@ If a viewport of 400px by 400px is split horizontally into two side-by-side segm
 
 </div>
 
+<h3 id='automation'>Automation</h3>
```
emilio, line 598: Should this be an appendix instead?

emilio (approved):

css-viewport-1/Overview.bs
darktears: Done

darktears:
My bikeshed version was updated and was triggering an error to generate the spec so I took the opportunity to fix it in the same commit.

darktears:
Can someone please merge this? Thanks.

emilio:
Sorry for the lag, my GH notifications are... not fun :(

darktears:
> Sorry for the lag, my GH notifications are... not fun :(

No worries. Thanks!

PR #11139, opened 2024-11-01
[css-backgrounds-4] Define background-repeat|position longhands with a logical property group

cdoublev:
Follow-up on https://github.com/w3c/csswg-drafts/issues/116#issuecomment-2295641610.

For example, flow-relative and physical `background-position-*` need to be in the same logical property group in order to maintain the appropriate order of declarations within a CSS declaration block, eg. with `style.cssText = 'background-repeat: repeat; background-repeat-block: no-repeat; background-repeat-y: repeat'`.

Cf. [serialize a CSS declaration block](https://drafts.csswg.org/cssom-1/#serialize-a-css-declaration-block)  and [set a CSS declaration](https://drafts.csswg.org/cssom-1/#set-a-css-declaration).

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -48,6 +48,7 @@ Tiling Images: the 'background-repeat-x', 'background-repeat-y', 'background-rep
 		Percentages: N/A
 		Computed value: as specified
 		Animation type: discrete
+		Logical property group: background-repeat
 	</pre>
 
 	<pre class="prod">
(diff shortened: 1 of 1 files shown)


fantasai (approved):

Loirooriol:
Properties in a logical property group share computed values. So you can't simply say they are in a logical property group if they have different grammars. Is there some mapping from some values to others? Do the properties accept computed values that can't be represented by their grammar?

PR #11147, opened 2024-11-02
[css-scroll-snap-2] [editorial] remove group notation from scroll-start-target

keithamus:
Similar to #11094 & https://github.com/w3c/csswg-drafts/pull/11053, the css-scroll-snap-2 has a redundant group around the `scroll-start-target` property value. It is currently specified as `[ auto | none ]`, but I think it could be better expressed without the `[]`, as `auto | none` instead.


argyleink (approved):

keithamus:
@argyleink I don't have merge rights yet. I'm wondering if you'd be keen on pushing the big shiny green button right...


:arrow_down:  here

PR #11148, opened 2024-11-02
[css-rythm-1] [editorial] remove redundant `px` dimension from `line-height-step` initial value

keithamus:
The `line-height-step` property in css-rythm-1 defines its initial value as `0px`. AIUI - it being a zero value means the `px` is redundant.

It seems to be more common to use `0` rather than `0px`. In fact here are all the places that have an initial of `0`:
 - css-box (`margin*`, `padding*`)
 - css-scroll-snap (`scroll-margin*`)
 - css-inline (`baseline-shift`)
 - css-shapes (`shape-margin`, `shape-padding`)
 - css-text (`text-indent`, `line-padding`, `hyphenate-limit-zone`)
 - css-text-decor (`text-decoration-trim`)
 - css-ui (`outline-offset`)
 - css-borders (`border-*-radius` & `border-radius`)
 - css-logical (okay maybe an unfair comparison as it's mostly logical equivalents of box/borders)
 
And here are the places that have an initial of `0px`:
 - css-rythm (`line-height-step` - that's this one)
 - css-tables (`border-spacing` - this initial value is `0px 0px`)
 - css-animation (`animation-delay` - `0s`)
 - css-animation (`transition-delay`, `transition-duration` - both `0s`)
 - css-overflow (`overflow-clip-margin*` - all `0px`)
 - css-backgrounds (`background-position*` - all `0%`)

If this PR gets merged I'll follow up with one correcting `css-overflow` and perhaps the others 😉

### DIFF
--- a/css-rhythm-1/Overview.bs
+++ b/css-rhythm-1/Overview.bs
@@ -356,7 +356,7 @@ Adjusting Line Box Heights: the 'line-height-step' property {#line-height-step}
 	<pre class='propdef'>
 	Name: line-height-step
 	Value: <<length [0,∞]>>
-	Initial: 0px
+	Initial: 0
 	Applies to: block containers
 	Inherited: yes
 	Percentages: N/A


frivoal (approved):

cdoublev:
If I am not mistaken, only zero [`<length>`](https://drafts.csswg.org/css-values-4/#lengths) outside math functions can be unitless (and zero [`<angle>`](https://drafts.csswg.org/css-values-4/#angles) in some places explicitly allowing it), which seems to be a legacy thing (akin to [`<quirky-length>`](https://drafts.csswg.org/css-values-4/#typedef-quirky-length)) so leaning towards `0px` seems more appropriate imo, as this evidently leads people to make wrong assumptions.

fantasai:
Ok, on reflection, I think @cdoublev is probably right, and it might be a good idea to always include the unit even for zero when stating the initial value. In particular, some of the ones you listed above (`0s` and `0%`) cannot be represented as `0`, since this has different behavior.

I suggest opening a meta-issue about what style we want to go with. :)

tabatkins:
Yes, "unitless zero" is an authoring convenience/legacy compat thing. We should be clear with a unit whenever we write something officially.

PR #11176, opened 2024-11-11
[css-transitions-2][css-animations-2] Define order of transitioncancel and animationcancel events

### LINKED ISSUE #11064: [css-transitions-2][css-animations-2] The event order of the cancelled transitions/animations
Recently, we noticed a Gecko bug about the event order of `transitioncancel` in [Bug 1923208](https://bugzilla.mozilla.org/show_bug.cgi?id=1923208). Per the example this bug, we may have a case when `transitioncancel` and `transitionrun` have the same scheduled time.

The example is something like this, when we are replacing a running transition:

```
<script>
function run() {
  target.style.marginLeft = "100px; // We start to run a transition.
  setTimeout(() => {
    target.style.marginLeft = "200px"; // We replace the running transition
  }, 500);
}
</script>
<body onload="run()">
  <div id="target" style="transition: margin-left 100s;"></div>
<body>
```
So basically, in Blink and Webkit, the event order when replacing the transition is:
```
transitioncancel, transitionrun, .transitionstart
```

However, in Gecko, the order is:
```
transitionrun, transitionstart, transitioncancel
```

These events have the same schedule time, and Gecko puts `transitioncancel` last.

Per spec, we use the following way to decide the order of them (copied from the comments in the bug):

* If the scheduled event times are the same then the algorithm is to sort by [composite order](https://drafts.csswg.org/web-animations-1/#animation-composite-order). That's going to mean we sort as follows:

        If A and B’s associated animations differ by class, sort by any inter-class composite order as defined for the corresponding classes.
        If A and B are still not sorted, sort by any class-specific composite order defined by the common class of A and B’s associated animations.
        If A and B are still not sorted, sort by the position of their associated animations in the global animation list.

* Since the classes are the same ("transition"), we'll sort by the [transition-specific composite order](https://drafts.csswg.org/css-transitions-2/#animation-composite-order), and then we'll hit step 2 there because the cancelled transitions don't have the owning element:

        Otherwise, if only one of A or B has an owning element, let the animation with an owning element sort first.

So it seems we should put the event of the cancelled transition first because it doesn't have the owning element. However, this is not what we expected because we cancel the transition first, and then create the new transition to replace the old one.

We may need to update the spec words to address the case for cancelled transition, e.g. per Brian's suggestion in this Gecko bug:
```
When sorting cancel events, use the owning element and transition generation from when the transition was cancelled
```

Also, we have the similar issue for `animationcanel` (in [css-animations-2](https://drafts.csswg.org/css-animations-2/#animation-composite-order)), and we probably need to figure out how to use `animation-name` for a cancelled animation.

cc @birtles

birtles:
- **[css-transitions-2] Define order for sorting transitioncancel events**
- **[css-animations-2] Define order for sorting animationcancel events**

Fixes #11064.

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -234,6 +234,18 @@ last) as follows:
     computed value of the 'animation-name' property of the (common) <a>owning
     element</a>.
 
+When determining the composite order in order
+to sort [[#events|animation events]]
+where either or both of the events is an {{animationcancel}} event,
+treat the CSS Animation(s) for which the {{animationcancel}} event was generated
+as having an [=owning element=] corresponding to
+the owning element in use at the moment when the CSS Animation was cancelled.
+Furthermore,
+use the position of the animation in the 'animation-name' property
+in effect at the time when the CSS Animation was cancelled
+sorting such that positions of cancelled animations sort
+before positions of animations that have not been cancelled.
+
 The composite order of CSS Animations <em>without</em> an <a>owning element</a>
 is based on their position in the <a>global animation list</a>.
 
--- a/css-transitions-2/Overview.bs
+++ b/css-transitions-2/Overview.bs
@@ -363,6 +363,11 @@ are sorted in composite order (first to last) as follows:
     such that &lsquo;-moz-column-width&rsquo; sorts before
     &lsquo;column-width&rsquo;).
 
+When determining the composite order in order to sort transition events
+where either or both of the events is a {{transitioncancel}} event,
+use the [=owning element=] and [=transition generation=]
+that were set immediately prior to cancelling the transition.
+
 Transitions generated using the markup defined in this specification are
 <em>not</em> added to the <a>global animation list</a> when they are created.
 Instead, these animations are appended to the <a>global animation list</a> at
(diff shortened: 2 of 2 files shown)


graouts:
@birtles are you planning on writing WPT test cases for this change?

birtles:
> @birtles are you planning on writing WPT test cases for this change?

@graouts Yes, @BorisChiou has added some over in [Mozilla bug 1923208](https://bugzilla.mozilla.org/show_bug.cgi?id=1923208): [here](https://phabricator.services.mozilla.com/D226212), [here](https://phabricator.services.mozilla.com/D225117) and [here](https://phabricator.services.mozilla.com/D226213)

graouts:
> > @birtles are you planning on writing WPT test cases for this change?
> 
> @graouts Yes, @BorisChiou has added some over in [Mozilla bug 1923208](https://bugzilla.mozilla.org/show_bug.cgi?id=1923208): [here](https://phabricator.services.mozilla.com/D226212), [here](https://phabricator.services.mozilla.com/D225117) and [here](https://phabricator.services.mozilla.com/D226213)

I'm not familiar with the Mozilla development process, but presumably these tests will be upstreamed upon landing in the Mozilla repository?

birtles:
> I'm not familiar with the Mozilla development process, but presumably these tests will be upstreamed upon landing in the Mozilla repository?

Yes, that's right.

Issue #10711, opened 2024-08-07
[css-ui] Support setting offscreen content inert


flackr:
## Problem

Tab and carousel UI patterns that can be mostly created with pure HTML and CSS. #9745 is aiming to expand upon how rich of an experience can be declaratively built. The primary use cases explored here are scrolling paginated tabs or carousels. E.g.
* Tabs supporting swiping implemented by a horizontal snapping scrolling element, examples:
  * https://gui-challenges.web.app/tabs/dist/
* Carousels implemented by a horizontal snapping scrolling elements, examples:
  * https://github.com/kevin-powell/netflix-carousel-css-only
  * https://gui-challenges.web.app/carousel/dist/

The [ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/) recommends that only the currently visible content is in the focus order, and has examples which achieve this either by:

1. using display: none (example [1](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/), [2](https://www.w3.org/WAI/tutorials/carousels/working-example/)) when they can (i.e. when the switch between slides is immediate), or
2. by setting `aria-hidden` from script (example [3](https://www.w3.org/WAI/tutorials/carousels/working-example/)).

In many cases, display: none can't be used, e.g. in most of the above cases because the previous and next slide may be visible simultaneously. However, since the alternative is an HTML attribute, it requires developers to write script that listens to the appropriate events and set these states, which can be a tricky to correctly detect and is unfortunate to need for cases where the interaction is otherwise declarative (e.g. scrolling or anchor links).

## Proposal

I believe there are a few ways we could approach this problem:

1.  Adding a CSS inertness property, e.g. `interactivity: auto | inert` which was previously deferred in #7021 [awaiting better use cases](https://github.com/w3c/csswg-drafts/issues/7021#issuecomment-1041931621). This has the potential to help with animation related cases, e.g. #8389 where we currently try to infer this [temporarily visible but inert state from the underlying style](https://github.com/w3c/csswg-drafts/issues/8389#issuecomment-1460542364). With this property, developers would still need to have a way to trigger this style change as a result of whether something is currently visible which could be based on a view-timeline, e.g.
    ```css
    @keyframes interactive-when-visible {
      0% { interactivity: auto; }
      100% { interactivity: auto; }
    }
    .page {
      interactivity: inert;
      animation: interactive-when-visible;
      animation-timeline: view();
    }
[...]

PR #11178, opened 2024-11-11
[css-ui-4] Add 'interactivity' property, per #10711

tabatkins:


tabatkins:
@flackr @frivoal Mind reviewing? I wanted to match the flat-tree behavior of [inert] in the HTML spec, so I couldn't *actually* make it an inherited property, but it acts similarly.

css-ui-4/Overview.bs
```diff
@@ -1972,6 +1974,86 @@ Exclusion from Hit-testing: the 'pointer-events' property</h3>
...
+Authors should not specify elements as inert 
+unless the content they represent are also visually obscured in some way. 
+In most cases, 
+authors should not specify the inert attribute on individual form controls. 
+In these instances, the <{input/disabled}> attribute is probably more appropriate.
+
+<div class=issue>
+	The HTML <{html-global/inert}> attribute is meant to be stronger
+	than the 'interactivity' property,
+	per CSSWG resolution.
```
emilio, line 2037: Hmm, do you know where is this resolution? In general:

 * I don't think that UA CSS works for shadow DOM (does `@scope` cross shadow boundaries).
 * It'd be nice to implement inert-escaping using this property (that's how it works in gecko already fwiw), and I don't see much reason for an author not to do the same?

But I guess I see the point of making `inert` take precedence... It just probably needs to be style system magic rather than expressed in terms of `@scope`.

tabatkins: Rereading #10711, I think that "[inert] must win over CSS" might have been a carryover in my memory from when it was being proposed as a `visibility` value, to avoid existing code that sets `visibility` from starting to escape inertness accidentally. It looks like, aside from modal dialogs, we're actually okay with CSS defeating the inert attribute, so long as it's done by a new property that won't trigger problems in legacy code.

But also, you're right, `@scope` doesn't extend into shadow trees, and inertness needs to (particularly for the forced inertness from modal dialogs). So yeah, we'll need to track a bit coming from the host language.

I'm thinking:

1. The host language can indicate that a given element is "forced inert", which causes `interactivity: auto` to (behave as? compute to?) `inert`. This is set on the rest of the page when a modal dialog is active, for example.
2. We add a UA rule for `[inert] { interactivity: inert; }`, just relying on inheritance.
3. We add a UA rule (either using a normal selector, or a new pseudo-class if the qualities aren't exposed to selectors currently) for `dialog, etc { interactivity: auto; }`, so they'll escape inertness by default.

alice: Sorry I don't have the energy today to make a better thought through comment, but please consider tying inert-escaping behaviour to top layer. At least, don't hastily add some inert-escaping behaviour which will have all the potential downsides we were trying to avoid by not allowing it in the first place.

lilles: > 1. The host language can indicate that a given element is "forced inert", which causes `interactivity: auto` to (behave as? compute to?) `inert`. This is set on the rest of the page when a modal dialog is active, for example.

From an implementer point of view, I'd prefer "behave as". Also, I think about it as the interactivity property affects the inertness in the host language, not that the host language affects the computed interactivity in CSS.

> 2. We add a UA rule for `[inert] { interactivity: inert; }`, just relying on inheritance.

:+1: 

> 3. We add a UA rule (either using a normal selector, or a new pseudo-class if the qualities aren't exposed to selectors currently) for `dialog, etc { interactivity: auto; }`, so they'll escape inertness by default.

I don't think we need this.` interactivity:auto` is the initial value, so this won't have an effect (unless you add !important). Also, the host language would just make sure dialog and its descendants don't have forced inertness?

emilio: @lilles how wouldn't we need it? Note that the original comment about this not being an inherited property is no longer relevant.

emilio: I think you still need the `dialog { interactivity: auto }` to make a modal `<dialog>` inside something like `<div inert>` work? I think it works everywhere now fwiw.

emilio: That said... Do we really need the "force inert" concept? In Gecko, we currently implement this with an internal pseudo-class that applies to the top of inert subtrees ([here](https://searchfox.org/mozilla-central/rev/b765e2890b0eb85b24f54bc7ff04491fd0704e30/layout/style/res/ua.css#180-182)).

When a modal dialog is shown, or some element is fullscreen, we apply that pseudo-class to the root element, and another (`:-moz-topmost-modal`) to the dialog or fullscreen element.

I wonder if the solution is a single property authors can only set to `inert` (i.e. you can mark subtrees inert with CSS, but not escape inertness, that is handled by the UA)? I think that also addresses @alice's concerns, and doesn't require keeping two flags around.

lilles: > I think you still need the `dialog { interactivity: auto }` to make a modal `<dialog>` inside something like `<div inert>` work? I think it works everywhere now fwiw.

You mean that the dialog should not be inert in the two cases below? Regardless of whether it's in the top layer or not?

```html
<div inert>
  <dialog></dialog>
</div>
<div style="interactivity:inert">
  <dialog></dialog>
</div>
```

flackr: Being able to escape inertness was a requested ability https://github.com/w3c/csswg-drafts/issues/10711#issuecomment-2379610168 .  Is it possible that for dialog or other top layer blocking UI we add a `!important` inert style to prevent escaping it in the html content but otherwise allow the inert attribute to be escaped?

alice: @emilio's suggestion to only allow inert-escaping by the UA would certainly be the most conservative option and would indeed address my concerns.

My main concern is that allowing inert-escaping behaviour essentially based on specificity means that it would be possible to accidentally end up with elements/subtrees which escape inert without realising it, since there is intrinsically no visual impact of making something inert or un-inert (and the un-inerted element may actually be visually inaccessible). 

My suggestion to tie un-inerting to the top layer is based on the principle that AT behaviour follows from visual behaviour. Something being in the top layer is a reasonably reliable heuristic that it is visually obscuring other elements, particularly if it has a backdrop.

Working backwards through @scottaohara's very helpful example list of cases for un-inerting from the comment linked above:

(3) Elements outside of a modally-showing dialog which nevertheless need to be showing, in particular using popover elements which exist outside of the dialog. Frankly, this seems like a bug in popover to me, and could be addressed via Emilio's suggestion of having the UA manage un-inerting. In general, if something is rendered in the top layer, it probably _shouldn't_ be inert unless explicitly made inert.

(2) Live regions for global announcer systems. I'm incredibly sympathetic to this case, but as Scott notes, the ARIA notification proposal is a better fit for the cases not included in (3), and seems to be making good progress. Also, if we make top layer elements automatically un-inert when the page is inerted because of another top layer element, perhaps this could make use of that to promote the announcer element into the top layer and thus out of inert?

(1) UI tours, where (I think) both the popover explaining the relevant part of the UI, and the relevant part of the UI, should be un-inert and un-obscured. I'll admit, this is potentially the most compelling counter-example to my logic. That said, how is the "page is mostly dimmed" behaviour implemented? Could the same mechanism be used to implement "page is mostly inert" using the new property?

flackr: > In general, if something is rendered in the top layer, it probably shouldn't be inert unless explicitly made inert.

I agree. I think UA rule to uninert may work for this case? E.g.

```css
/* TODO: match other things that may be in the top layer. */
dialog:open {
  interactivity: auto; /* Not !important so an author rule on the dialog can override this. */
}

[inert] {
  interactivity: inert;
}
```

tabatkins: There's been more in this thread, but specifically:

> interactivity:auto is the initial value, so this won't have an effect (unless you add !important)

No, putting a rule into the cascade, even if it matches the initial value, causes it to defeat inheritance, as we only use the inherited value if *nothing* wins the cascade.

mfreed7: > (3) Elements outside of a modally-showing dialog which nevertheless need to be showing, in particular using popover elements which exist outside of the dialog. Frankly, this seems like a bug in popover (or even dialog) to me, and could be addressed via Emilio's suggestion of having the UA manage un-inerting. In general, if something is rendered in the top layer, it probably _shouldn't_ be inert unless explicitly made inert.

This is a good point, which I've heard a few times. I opened this issue https://github.com/whatwg/html/issues/10811 to discuss this problem more generally. It likely should be "fixed" in WHATWG and not in CSSWG, since the top layer stack is there?

nt1m: `dialog:modal` would be more accurate than `dialog:open` in the UA rule above fwiw (in terms of matching the current behavior).

dholbert:
@tabatkins The html version at https://drafts.csswg.org/css-ui-4/ doesn't seem to have this update yet (it shows "22 December 2024" as its datestamp and doesn't mention "interactivity") - does that need to be regenerated?

Maybe/probably that's already on your radar, if so, apologies for the noise.

nt1m:
@tabatkins @flackr How do the UA styles cause modal dialogs to make the rest of the page inert? I don't see that yet in the current UA styles.

mfreed7:
> @tabatkins @flackr How do the UA styles cause modal dialogs to make the rest of the page inert? I don't see that yet in the current UA styles.

That's handled by HTML. See the (recently opened) PR here: https://github.com/whatwg/html/pull/10956

nt1m:
> > @tabatkins @flackr How do the UA styles cause modal dialogs to make the rest of the page inert? I don't see that yet in the current UA styles.
> 
> That's handled by HTML. See the (recently opened) PR here: [whatwg/html#10956](https://github.com/whatwg/html/pull/10956)

I think `blocked by a modal dialog` should be defined in terms of `interactivity` (either set as a presentational hint on the root, or using an UA stylesheet rule), because otherwise `dialog:modal { interactivity: auto }` won't do anything (because `auto` is the default value. I don't think this is what the current HTML spec PR.

lilles:
> > > @tabatkins @flackr How do the UA styles cause modal dialogs to make the rest of the page inert? I don't see that yet in the current UA styles.
> > 
> > 
> > That's handled by HTML. See the (recently opened) PR here: [whatwg/html#10956](https://github.com/whatwg/html/pull/10956)
> 
> I think `blocked by a modal dialog` should be defined in terms of `interactivity` (either set as a presentational hint on the root, or using an UA stylesheet rule), because otherwise `dialog:modal { interactivity: auto }` won't do anything (because `auto` is the default value. I don't think this is what the current HTML spec PR.

No one's been able to come with UA styles that work correctly for shadow DOM.

I believe `dialog:modal { interactivity: auto; }` is there to escape from inherited ancestor inertness style when the dialogs are modal in:

```html
<div style="interactivity:inert">
  <dialog></dialog>
</div>
<div inert>
  <dialog></dialog>
</div>
```

PR #11179, opened 2024-11-11
[css-values-5] Make it clear that `allow-keywords` can't just interpolate with intrinsic sizes.

Psychpsyo:
The current wording is a tiny bit ambiguous; it could also be interpreted to mean that `interpolate-size: allow-keywords` only lets you interpolate between an intrinsic size and a length percentage, but not two length percentages.

I do realize how unlikely this interpretation is, but I had to do a double-take while reading so I thought I'd PR it.


Psychpsyo:
I think this is non-substantive, but if it is ruled to be, I will make a W3C account.

Issue #10715, opened 2024-08-08
[css-overflow-5][css-scroll-snap-2] Snapping and generating scroll-marker pseudo-elements from fragments


flackr:
For common carousel patterns as outlined in #9745, developers will have content being fragmented into different boxes. Often, developers want to create an interface that scrolls through these fragments by pages, e.g. [demo](https://flackr.github.io/carousel/examples/carousel/image/). Similarly, for existing column fragmentation use cases developers may wish to snap to particular columns. Issues #6017 and #5911 both have prior discussion on snapping specifically.

As with snapping, it would be very convenient to be able to generate a [::scroll-marker](https://drafts.csswg.org/css-overflow-5/#scroll-marker-pseudo) pseudo-element per generated fragment to set up automatic markers based on the available space and number of resulting fragments the content ended up needing.

This explores the two issues together as they feel similar in nature, though we don't necessarily have to resolve on the same solution for both.

For the examples below, we have the following common styles:
```css
.scroller {
  overflow: auto;
  height: 80vh;
  scroll-snap-type: x mandatory;
  columns: 1; /* Fragment content to horizontally scrollable full pages */
}
```

Options:
1.  Support [scroll-snap-align](https://drafts.csswg.org/css-scroll-snap/#scroll-snap-align) (as suggested in #6017) and [::scroll-marker](https://drafts.csswg.org/css-overflow-5/#scroll-marker-pseudo) pseudo-elements on a `::fragment` selector[^1]. This would allow a paginated carousel-like snapping experience to be created as follows:
    ```html
    <style>
      .scroller::fragment {
        scroll-snap-align: center;
        &::scroll-marker {
          content: ' ';
        }
      }
    </style>
    <div class="scroller">
      <!-- flowing content or items, e.g. -->
      <div>Item 1</div>
      <div>Item 2</div>
      ...
      <div>Item n</div>
    </div>
    ```
[...]

PR #11186, opened 2024-11-12
[css-multicol-2] Add ::column pseudo, per WG resolution. #10715

tabatkins:
@frivoal @rachelandrew Hi y'all, just drafting the ::column pseudo per #10715. Feel free to review/merge. I can also move the new section elsewhere if desired (or, potentially, to the Pseudo spec instead, but I feel like this is the better spot for it).

### DIFF
--- a/css-multicol-2/Overview.bs
+++ b/css-multicol-2/Overview.bs
@@ -82,6 +82,48 @@ Stacking context</h3>
 
 	Issue: Add final content from previous level
 
+<h3 id=column-pseudo>
+Styling Columns: the ''::column'' pseudo-element</h3>
+
+The <dfn selector>::column</dfn> pseudo-element
+represents the individual columns
+in a [=multi-column container=].
+It only exists on [=multi-column containers=].
+
+A [=multi-column container=] has as many ''::column'' pseudo-elements
+as it has columns.
+They cannot be individually addressed;
+any styles applied to a multicol's ''::column'' pseudo-elements
+apply to all of them on that element.
+Each ''::column'' has the same size and position
+as its corresponding column:
+the same size as the column's [=available space=] in the [=inline axis=]
+and the container's [=content box=] in the [=block axis=].
+(It does not cover the gaps/rules between columns.)
+
+''::column'' pseudo-elements are treated as children of their [=multi-column container=],
+without any contents.
+They <em>do not</em> wrap the contents of the column,
+just fill the same space.
+
+The ''::column'' pseudo-element accepts only a very limited set of properties:
+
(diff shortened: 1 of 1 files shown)


css-multicol-2/Overview.bs
```diff
@@ -82,6 +82,48 @@ Stacking context</h3>
...
+Each ''::column'' has the same size and position
+as its corresponding column:
+the same size as the column's [=available space=] in the [=inline axis=]
+and the container's [=content box=] in the [=block axis=].
+(It does not cover the gaps/rules between columns.)
+
+''::column'' pseudo-elements are treated as children of their [=multi-column container=],
+without any contents.
+They <em>do not</em> wrap the contents of the column,
+just fill the same space.
```
fantasai, line 107: I'm not sure this is true. I think we might reasonably say that they *do* wrap the contents of the column; and in the far future when we allow fragment-based styling, we might want to.

tabatkins: We very specifically resolved that ::column was a leaf node, with no contents. See <https://github.com/w3c/csswg-drafts/issues/10715#issuecomment-2379924982> You even made some comments in support of that. ^_^

tabatkins:
Do we want to add 'scroll-behavior' to the list of supported ::column properties? It's the only thing checked on the element by the [perform a scroll](https://drafts.csswg.org/cssom-view/#perform-a-scroll) algorithm, just to see if the scroll should be smooth or not. (While you can't (yet) scroll to a ::column via JS, the same algo is invoked when you click the ::scroll-marker associated with the ::column.)

rachelandrew:
I think here is as good a place as any for it for now, we could always move it later. Unless @frivoal has concerns I'm happy to merge it here.

AmeliaBR:
Is there an open issue for discussing which other properties might apply to a `::column` pseudo-element? As suggested in:

> Note: This list of properties and pseudo-elements that work on [::column](https://drafts.csswg.org/css-multicol-2/#selectordef-column) will likely be expanded in the future, as we develop more features that could usefully care about the position of a column without caring about the contents.

For example, I think it would be safe to add a background color & maybe full backgrounds, if someone wanted the column area to be coloured but not the gaps. Although then there might be a desire to add padding around the text but inside the coloured area, and... yeah, that's why there should probably be a standing issue (linked from that note) where you can start scoping out what is or isn't possible for this psuedo-element.

rachelandrew:
@AmeliaBR I don't think we specifically have an issue for multicol, there's mention of fragment styling in [overflow-4](https://www.w3.org/TR/css-overflow-4/#fragment-styling) which would be one way to achieve this. It's definitely something people ask for.

PR #11200, opened 2024-11-13
[css-view-transitions-2] Editorial: fix a few nits for WD publishing

noamr:
- Remove `type` from dfn
- Remove typo fixes from changelog

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -555,7 +555,7 @@ An '':active-view-transition'' pseudo-class matches the [=document element=] whe
 
 ### The '':active-view-transition-type()'' pseudo-class ### {#the-active-view-transition-type-pseudo}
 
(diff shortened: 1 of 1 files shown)


fantasai (approved):
I think you can commit directly to main for this kind of fixes. ;) r+ fantasai anyway

css-view-transitions-2/Overview.bs
```diff
@@ -822,7 +822,7 @@ li {
 
 ## Additions to 'view-transition-name' ## {#additions-to-vt-name}
 
-In addition to the existing values, the 'view-transition-name' also accepts an <dfn for=view-transition-name type=value export>auto</dfn> keyword.
+In addition to the existing values, the 'view-transition-name' also accepts an <dfn for=view-transition-name export>auto</dfn> keyword.
```
cdoublev, line 825: Would that be a problem to add the following before this line?

```html
<pre class="propdef partial">
Name: view-transition-name
New values: auto
</pre>
```

cdoublev: Oups, too late. Later then =)

PR #11207, opened 2024-11-13
[css-shapes-2] Specify the grammar for `shape()`

### LINKED ISSUE #10649: [css-shapes-2] `curve to` keyword `using` seems a bit off
We adopted https://github.com/w3c/csswg-drafts/pull/9797 which allowed reordering and switched `via` to `using` but we typically use prepositions, not verbs, for these types of keywords and also `using` is a bit vague in any case. Opening this issue to look for alternatives.

My opening bid is `around`. :)

### LINKED ISSUE #10666: [css-shapes-2] Order of points and control points in `shape()`
In https://github.com/w3c/csswg-drafts/issues/5841 we resolved that the points and the control points could be specified in any order, e.g. these are equivalent:
```
curve using 10px 30px by 20px 20px
curve by 20px 20px using 10px 30px
```

I'm having second thoughts. The `by` or `to` affect the interpretation of the control point coordinate pairs; they are either absolute (relative to the reference box), or relative to (the current point). So it's odd to specify the control points before you've given the `by` or `to`.

I think we should revert this part of the change.

noamr:
grammar changes:
- `<<position>>` can be used instead of an absolute point
- The relevant `<<position>>`'s dimension can be used for `hline`/`vline`
- Since keywords define the different components, order of components is flexible
- `with` is used for curve control points
- Two curve control points are separated by /
- Relative control points can be relative to segment-start/segment-end/reference-box, defaulting to segment start.

Resolution: https://github.com/w3c/csswg-drafts/issues/10649#issuecomment-2474354244

Closes #10649
Closes #10666

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -185,7 +185,7 @@ The ''shape()'' Function</h4>
 						[ to [ <<length-percentage>> | top | center | bottom | y-start | y-end ]
 						| by <<length-percentage>> ]
 		<<curve-command>> = curve
-						[ [ to <<position>> && with <<control-point>> [ / <<control-point>> ]? ]
+						[ [ to <<position>> && [ with <<control-point>> [ / <<control-point>> ]? ] ]
 						| [ by <<coordinate-pair>> && with <<relative-control-point>> [ / <<relative-control-point>> ]? ] ]
 		<<smooth-command>> = smooth
 						[ [ to <<position>> && [with <<control-point>> ]? ]
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -190,9 +190,9 @@ The ''shape()'' Function</h4>
 		<<smooth-command>> = smooth
 						[ [ to <<position>> && [ with <<control-point>> ]? ]
 						| [ by <<coordinate-pair>> && [ with <<relative-control-point>> ]? ] ]
-		<<arc-command>> = arc <<command-end-point>>
-								&& of <<coordinate-pair>>
-								&& <<arc-sweep>>? && <<arc-size>>? && [rotate <<angle>>]?
+		<<arc-command>> = arc [ <<command-end-point>>
+								&& [ of <<coordinate-pair>> ]
+								&& <<arc-sweep>>? && <<arc-size>>? && [rotate <<angle>>]? ]
 
 		<<command-end-point>> = [ to <<position>> | by <<coordinate-pair>> ]
 		<<control-point>> = [ <<position>> | <<relative-control-point>> ]
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -306,8 +306,9 @@ The ''shape()'' Function</h4>
 		<dt><dfn><<relative-control-point>></dfn> = <<coordinate-pair>> [ from [ start | end | origin ] ]?
 		<dd>
 			Provides a control point to a quadratic or cubic Bézier curve.
-			When a <css>from</css> component is provided, with values of <css>start</css>, <css>end</css>, or <css>origin</css>,
-			the given <<coordinate-pair>> is relative to the command's starting point, the command's end point, or the [=reference box=], respectively.
+			When a <css>from</css> keyword is specified followed by <css>start</css>, <css>end</css>, or <css>origin</css>,
+			the given <<coordinate-pair>> is relative to
+			the command's starting point, the command's end point, or the [=reference box=], respectively.
 			If such component is not provided, the <<coordinate-pair>> is relative to the segment's start.
 
 		<dt><dfn><<arc-command>></dfn> = <dfn value>arc</dfn> [[<<by-to>> <<coordinate-pair>>] || [of <<length-percentage>>{1,2}] || <<arc-sweep>>? || <<arc-size>>?|| rotate <<angle>>? ]
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -271,7 +271,8 @@ The ''shape()'' Function</h4>
 						| [ by <<coordinate-pair>> && with <<relative-control-point>> [ / <<relative-control-point>> ]? ] ]
 		<dd>
 			Adds a Bézier curve command to the list of path data commands,
-			ending at the point specified by the <<position>> following the <css>to</css> keyword, or the <<coordinate-pair>> following the <css>by</css> keyword, as specified by <<command-end-point>>.
+			ending at the point specified by the <<position>> following the <css>to</css> keyword,
+			or the <<coordinate-pair>> following the <css>by</css> keyword, as specified by <<command-end-point>>.
 
 			The <css>with</css> component specifies control points for the curve:
 			if a single <<control-point>> or <<relative-control-point>> is provided,
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -28,6 +28,16 @@ spec:css-writing-modes-4; type:concept; text:physical
 spec:css-writing-modes-4; type:concept; text:"writing mode"
 spec:svg2; type:property;
 	text:fill-rule
+spec:css-values-5; type:value;
+	text:top;
+	text:right;
+	text:bottom;
+	text:left;
+	text:center;
+	text:x-start;
+	text:y-start;
+	text:x-end;
+	text:y-end;
 </pre>
 
 <style type="text/css">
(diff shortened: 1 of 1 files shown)


css-shapes-2/Overview.bs
```diff
@@ -185,19 +210,21 @@ The ''shape()'' Function</h4>
...
 		<dd>
 			Every command can be specified in "absolute" or "relative" coordinates,
-			determined by their <<by-to>> component.
-			''to'' indicates that any <<coordinate-pair>>s in the command
+			determined by their <css>by</css> or <css>to</css> component.
+			<css>to</css> indicates that any <<coordinate-pair>>s in the command
 			are relative to the top-left corner of the [=reference box=],
-			while ''by'' indicates that the <<coordinate-pair>>s
+			while <css>by</css> indicates that the <<coordinate-pair>>s
 			are relative to the command's starting point.
```
smfr, line 220: This is no longer true when `from start` / `from end` are used. There probably needs to be a new paragraph that explains how control points are resolved, or a "see below for details".

noamr: Right... There is a paragraph for control points, I'll refer to that.

fantasai: I think you can use `''shape()/by''` for the markup. I think.

css-shapes-2/Overview.bs
```diff
@@ -162,17 +172,32 @@ The ''shape()'' Function</h4>
 	Each command's starting point is the previous command's ending point.
 
 	<pre class=prod>
-		<<shape-command>> = <<move-command>> | <<line-command>> | <<hv-line-command>> |
-		                  <<curve-command>> | <<smooth-command>> | <<arc-command>> | close
+		<<shape-command>> = <<move-command>> | <<line-command>> | <<horizontal-line-command>> |
+						<<vertical-line-command>> | <<curve-command>> | <<smooth-command>> |
+						<<arc-command>> | close
```
fantasai, line 177: Not sure what indentation you're aiming to use but 4 consecutive tabs seems very suspicious.

I might suggest ordering these as
```suggestion
		<<shape-command>> = <<move-command>> | <<line-command>> | close
		                                        <<horizontal-line-command>> | <<vertical-line-command>> | 
		                                        <<curve-command>> | <<smooth-command>> | <<arc-command>>
```

so that related commands are on a single line.

css-shapes-2/Overview.bs
```diff
@@ -209,42 +236,55 @@ The ''shape()'' Function</h4>
...
+			Adds a vertical
 			<a href="https://www.w3.org/TR/SVG/paths.html#PathDataLinetoCommands">lineto</a> command
 			to the list of path data commands.
 
 			This is equivalent to a ''line'' command
-			with the <<length-percentage>> given as the horizontal or vertical component, respectively,
-			of the <<coordinate-pair>>,
-			and the other component specified appropriately to make the line horizontal or vertical.
+			with the <<length-percentage>> given as the vertical component of the <<coordinate-pair>>.
+			Specifying the horizontal component of <<position>> instead of a <<length-percentage>> (''top'', ''center'', ''bottom'', ''y-start'', or ''y-end''),
```
fantasai, line 267:
```suggestion
			Specifying the horizontal component of <<position>>
			(''top'', ''center'', ''bottom'', ''y-start'', or ''y-end'')
			instead of a <<length-percentage>>,
```

css-shapes-2/Overview.bs
```diff
@@ -209,42 +236,55 @@ The ''shape()'' Function</h4>
...
+			would draw a line to that <<position>>, with the <<position>>'s horizontal component remaining the same as the starting point.
 
-		<dt><dfn><<curve-command>></dfn> = <dfn value>curve</dfn> [[<<by-to>> <<coordinate-pair>>] || [using <<coordinate-pair>>{1,2}]]
+		<dt><dfn><<curve-command>></dfn> = <dfn value>curve</dfn> [ [ to <<position>> && with <<control-point>> [ / <<control-point>> ]? ]
+						| [ by <<coordinate-pair>> && with <<relative-control-point>> [ / <<relative-control-point>> ]? ] ]
 		<dd>
 			Adds a Bézier curve command to the list of path data commands,
-			ending at the point specified by the first <<coordinate-pair>>.
+			ending at the point specified by the <<position>> following the <css>to</css> keyword, or the <<coordinate-pair>> following the <css>by</css> keyword, as specified by <<command-end-point>>.
```
fantasai, line 274: Suggested line break

```suggestion
			ending at the point specified by the <<position>> following the <css>to</css> keyword,
			or the <<coordinate-pair>> following the <css>by</css> keyword, as specified by <<command-end-point>>.
```

css-shapes-2/Overview.bs
```diff
@@ -259,11 +299,22 @@ The ''shape()'' Function</h4>
...
 
+		<dt><dfn><<control-point>></dfn> = [ <<position>> | <<relative-control-point>> ]
+		<dd>
+			Provides a control point to a quadratic or cubic Bézier curve.
+
+		<dt><dfn><<relative-control-point>></dfn> = <<coordinate-pair>> [ from [ start | end | origin ] ]?
+		<dd>
+			Provides a control point to a quadratic or cubic Bézier curve.
+			When a <css>from</css> component is provided, with values of <css>start</css>, <css>end</css>, or <css>origin</css>,
+			the given <<coordinate-pair>> is relative to the command's starting point, the command's end point, or the [=reference box=], respectively.
```
fantasai, line 310: Not sure about this edit, but it's slightly shorter if you like it.

```suggestion
			When a <css>from</css> keyword is specified followed by <css>start</css>, <css>end</css>, or <css>origin</css>,
			the given <<coordinate-pair>> is relative to
			the command's starting point, the command's end point, or the [=reference box=], respectively.
```

css-shapes-2/Overview.bs
noamr: Fixed

css-shapes-2/Overview.bs
```diff
@@ -162,17 +172,32 @@ The ''shape()'' Function</h4>
...
+						| [ by <<coordinate-pair>> && with <<relative-control-point>> [ / <<relative-control-point>> ]? ] ]
+		<<smooth-command>> = smooth
+						[ [ to <<position>> && [with <<control-point>> ]? ]
+						| [ by <<coordinate-pair>> && [ with <<relative-control-point>> ]? ] ]
+		<<arc-command>> = arc <<command-end-point>>
+								&& of <<coordinate-pair>>
+								&& <<arc-sweep>>? && <<arc-size>>? && [rotate <<angle>>]?
+
+		<<command-end-point>> = [ to <<position>> | by <<coordinate-pair>> ]
+		<<control-point>> = [ <<position>> | <<relative-control-point>> ]
```
weinig, line 198: When implementing this, I mentioned this to @smfr, and he didn't quite agree, but I thought I would bring it to a slightly larger audience.

While generally I don't really see the benefit in disallowing <position> to be used for `by` (since <position> is really just sugar for [ <length-percentage> <length-percentage> ], but if the distinction is going to be made, I think the grammar can be a bit more friendly:

```
<control-point> = [ <position> | <relative-control-point> ]
```

with one like so: 

```
<absolute-control-point> = <position> [ from [ start | end | origin ] ]?
```

or, if you dislike the duplication, of the `from` section,

```
<control-point-from> = [ from [ start | end | origin ] ]
<absolute-control-point> = <position> <control-point-from>?
<relative-control-point> = <coordinate-pair> <control-point-from>?
```

This allows a user who is already using an absolute control point, for example like so:

```
... curve to center center with top left ...
```

who decides that `with top left` (aka with 0% 0%) should really be `with top left from start` to do so. In the current grammar, they have to translate their top left to percentage form and write out  `with 0% 0% from start`.

(doing this also removes a slightly annoying, though manageable, ambiguity in the grammar when parsing `<<control-point>` since its not clear if "10% 10%" should match `<position>` or `<coordinate-pair>`.

noamr: > When implementing this, I mentioned this to @smfr, and he didn't quite agree, but I thought I would bring it to a slightly larger audience.
> 
> While generally I don't really see the benefit in disallowing to be used for `by` (since is really just sugar for [ ], but if the distinction is going to be made, I think the grammar can be a bit more friendly:
> 
> ```
> <control-point> = [ <position> | <relative-control-point> ]
> ```
> 
> with one like so:
> 
> ```
> <absolute-control-point> = <position> [ from [ start | end | origin ] ]?
> ```
> 
> or, if you dislike the duplication, of the `from` section,
> 
> ```
> <control-point-from> = [ from [ start | end | origin ] ]
> <absolute-control-point> = <position> <control-point-from>?
> <relative-control-point> = <coordinate-pair> <control-point-from>?
> ```

This allows `to bottom left from end`, how would you interpret that?

weinig: As identical to ‘to 100% 0% from end’?

noamr: > As identical to ‘to 100% 0% from end’?

Ah so like transpose the reference rect to the start/end point... not sure it's a useful box for positioning, might create more confusion than usefulness.

weinig: More confusing than the fact that ‘to 100% 0% from end’ is valid and ‘to bottom left’ is valid but not the merged form?

Mostly the point I am trying to make is that treating <position> as something other than syntax sugar for two <length- percentage>s seems undesirable. But changing this later is also non-breaking as far I can tell, so I really don’t want to push this any further than I have.

noamr: > More confusing than the fact that ‘to 100% 0% from end’ is valid and ‘to bottom left’ is valid but not the merged form?

`to 100% 0% from end` is not valid, only `with 100% 0% from end` or `with 10px 220px / 100% 0% from end`, the `from start|end|origin` syntax is only for curve control points. 

I read `with bottom left` as a shorthand for `with 100% 0% from origin`, so expanding it to `with 100% 0% from origin from end` is indeed confusing.

> 
> Mostly the point I am trying to make is that treating as something other than syntax sugar for two s seems undesirable. But changing this later is also non-breaking as far I can tell, so I really don’t want to push this any further than I have.

Sounds good!

noamr: Resolving this for now, feel free to file a new issue about this if you'd like.

tabatkins (approved):
just some syntax fixes, i'll merge them myself

PR #11222, opened 2024-11-15
[css-overflow-5] Editoral - move scroll buttons under scroll navigation controls

flackr:


flackr:
@tabatkins What do you think of grouping markers and buttons under the common scroll navigation controls section? I thought this might be useful for having a place to hang the focus order between them.

If you think it should be separate then I'll probably move scroll markers to be the top level section instead.

tabatkins:
Ah, I made this change myself when I was writing the focus order text. Weird that this isn't recording a conflict...

flackr:
Thanks!

Issue #11198, opened 2024-11-13
[css-overflow-5] Active marker in 2d scroller?


flackr:
When you have a group of scroll targets in a scroller that can scroll in multiple directions, how do you determine which one is active?

Options:
1. Declare a single direction to use in some manner. This could be a property on the scrolling container indicating, or on the marker group / pseudo marker group creation. My preference here would be to use something like the `primary-overflow-direction` from #10060.
2. Automatically infer a single direction to use. We could do this until a mechanism for declaring a direction exists / continue to support it with an automatic value. The automatic direction would likely be either the direction with more scrollable overflow.
3. Use both axes of scroll in the determination of the active marker. We could possibly do this by using the distance from the current target scroll position to the marker's aligned positions.

With options 1 and 2, we could use the single direction as the primary direction but still use the secondary direction when the targeted location in the primary direction is the same. E.g. the first step could be determining the active set of markers on the primary direction, then repeat with the secondary direction.

flackr:
I think I'd propose we go with option 2, choosing the direction with more scrollable overflow, as the primary axis and in the case of a tie choosing the block axis. I could be convinced that for predictability we should go with option 1, however I am somewhat concerned with doing this if we don't have the mechanism for changing the primary axis as many of the use cases we're imagining would be in the non-default horizontal axis.

Issue #11098, opened 2024-10-28
[css-overflow-5] What is the active :checked marker when some markers point to elements within different scrolling containers?


flackr:
With `<a>` elements, it is possible to link to any content anywhere on the page. This means that we can create a scroll marker group where multiple markers might be respectively active. E.g.

Nested scrollers might be used within sections of contents:
```html
<div focusgroup>
  <ol>
    <li><a href="#s1">Section 1</a>
      <ol>
        <li><a href="#s1_1">Subsection 1</a></li>
      </ol>
    </li>
    <li><a href="#s2">Section 2</a>
      <ol>
        <li><a href="#s2_1">Subsection 1</a></li>
      </ol>
    </li>
  </ol>
</div>
<h1 id="s1">Section 1</h1>
<div class="scroller">
  <h2 id="s1_1">Subsection 1</h2>
</div>
```

Sibling / cousin scrollers (I'm less sure of the use case but it would need a defined behavior):
```html
<div focusgroup>
  <ol>
    <li><a href="#i1">Item 1</a></li>
    <li><a href="#i2">Item 2</a></li>
  </ol>
</div>
<div class="scroller">
  <h1 id="i1">Item 1</h2>
</div>
<div class="scroller">
  <h1 id="i2">Item 2</h2>
</div>
```

In both of these cases there are many scroll markers which could be said to be `:checked` as their nearest ancestor scrolling element is currently scrolled to them.

flackr:
I think to handle this in a general way we should find the common ancestor scrolling container (and tree of contained scrolling containers) of all of the scroll markers in the group, and then work from the top-down to first find the active scroller, then repeat within that until you arrive at a particular marker.

flackr:
To make this proposal more concrete. The proposed way to select the active marker among targets in different scrolling containers is as follows:

Define/share an algorithm for determining the active target for a list of targets in a single scroller (which as discussed in #10917 will for now will be a bit flexible),
To determine the active marker, we start from the root (or the nearest ancestor scrolling container) S. Then
1. Let targets be all of the scroll targets whose scrolling container is S, and all of the scroll containers within S which have descendent scroll targets.
2. Find the active target T amongst targets in S.
3. If T is not a scroll target, let S = T and repeat from step 1.
4. Return T as the active target.

[...]

Issue #10917, opened 2024-09-18
[css-overflow-5] Improve active marker calculation


flackr:
Currently the active scroll marker is [determined by](https://drafts.csswg.org/css-overflow-5/#scroll-container-scroll) finding the target which is scrolled to or before the current scroll position. However, this means that even when the next section is 90%+ visible until it reaches the top of the scrollport (or it's scroll aligned position) the active marker does not update to this section. See https://flackr.github.io/carousel/examples/scroll-marker/scrolltarget/ for a demo of this behavior.

Instead, I propose we have the active marker calculated using logic similar to how snap areas are selected where we will select the next section if the previous target is above its desired scroll alignment and the next one is more than half-way to being aligned.

This could be analogous to snap areas where we create a virtual snap area from the start of one scroll target to the start of the next, however unlike snap areas we still want to strongly consider the previous if the next one is not yet sufficiently in view. E.g. if you're scrolled halfway through section 1 even though the target (the header) is scrolled above the viewport you should still be considered to be in section 1 until section 2 is mostly in view.

flackr:
We need to balance for uses where scroll marker targets may be
* Small and thus should still be current even when next item may be less than half a scrollport from being aligned, or
* larger than the scrollport, and shouldn't eagerly switch to next item until their content is more than half offscreen, and/or
* aligned to other edges, where we shouldn't choose a marker that is not yet visible.

Proposal, when deciding whether we should select the next item, we do so when:
1. We are past the ideal alignment of the previous (current algorithm), and
2. The next marker target intersects the viewport, and
3. The distance from the previous item's aligned scroll position to the next item's aligned scroll position (capped to one scrollport to not eagerly choose the next item) is more than twice than the distance to the next item being aligned.

Note that we unfortunately cannot use the exact same logic as snap areas. Since the target may only be the header of the entire section, we should consider the effective target to be everything from the first target being aligned to the next target being aligned. We could consider re-using the snap logic using this virtual area.

kizu:
The proposed algorithm sounds good at the first glance. I recently implemented something similar to it in my blog (https://blog.kizu.dev/captured-custom-properties/ for example, using scroll-driven animations, so works only in Chrome for now, [source CSS behind it](https://github.com/kizu/kizu-blog/blob/main/src/styles/features/scroll-markers.css)). I am using an `inset: 50%` with the “markers” on headers, which seems to work better than any of my other attempts to do so.

flackr:
[...]

PR #11224, opened 2024-11-15
[css-overflow-5] Remove normative active marker algorithm.

flackr:
As resolved in #10917:
- removes the normative algorithm for determining the active marker and leaves it up to the UA,
- additionally defines an example algorithm that takes into account improvements mentioned in the issue for selecting the next marker when it is nearer to being in view.

The example algorithm also
- includes recursive marker selection within nested scroll containers as proposed in #11098, and
- uses the writing mode as the primary axis following the proposed primary axis algorithm in #11198.

### DIFF
--- a/css-overflow-5/Overview.bs
+++ b/css-overflow-5/Overview.bs
@@ -288,33 +288,51 @@ The 'scroll-marker-group' property</h4>
 	This 'eventual scroll position' is used to determine the active marker within each [=scroll marker group=].
 	Since markers themselves may represent just the start of the content (e.g. headers), we consider the active marker to be the first one which we are at or beyond the scroll position of.
 
-	<div algorithm="scroll tracking">
-		Whenever a [=scroll container=] is scrolled, or layout changes the scroll position, the user agent must run these steps to determine the active marker:
+	Whenever a [=scroll container=] participating in one or more [=scroll marker groups=] is scrolled, or layout changes the eventual scroll position,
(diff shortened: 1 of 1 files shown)


css-overflow-5/Overview.bs
```diff
@@ -288,33 +288,51 @@ The 'scroll-marker-group' property</h4>
...
+		1. Let |active| be scroller.
+		1. While |active| is a [=scroll container=] containing [=scroll target=] elements targeted by |group|:
+			1. Let <var>scroller</var> be |active|.
+			1. Let <var>targets</var> be the set of the [=scroll target=] elements whose nearest ancestor [=scroll container=] is |scroller|
+			   and the [=scroll container=] elements which contain [=scroll target=] elements targeted by the [=scroll marker group=] whose nearest ancestor [=scroll container=] is |scroller|.
+			1. Let <var>primary</var> be the primary scrolling axis, assumed to be the block direction of the container's writing-mode.
+			1. Let <var>secondary</var> be the scrolling axis perpendicular to primary.
+			1. Let <var>position</var> be the 'eventual scroll position' considering ongoing scrolling operations.
+			1. For each <var>axis</var> of |primary|, followed by |secondary|:
+				1. 	Let <var>scrollport size</var> be the client size of |scroller| in the dimension |axis|.
```
flackr, line 308: We probably want the size here and the positions used to account for the [optimal viewing region](https://drafts.csswg.org/css-scroll-snap/#optimal-viewing-region), though perhaps this is a followup?

PR #11227, opened 2024-11-18
[selectors-4] Missing link in a single paragraph

OnlyWick:
Impact on readability.

Documentation source link: https://www.w3.org/TR/selectors-4/#attribute-representation

Bikeshed link: https://speced.github.io/bikeshed/#metadata-remove-multiple-links

Before:
xx

After:
xx

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -4,6 +4,7 @@ Group: CSSWG
 Shortname: selectors
 Level: 4
 Status: ED
+Remove Multiple Links: false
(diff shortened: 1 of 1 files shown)


tantek (approved):

OnlyWick:
Ok, I see. But this can be misleading on first reading. What if we split
this sentence into two sentences, so that we can reference it correctly

Tab Atkins Jr. ***@***.***> 于 2024年11月20日周三 01:35写道：

Issue #7348, opened 2022-06-11
[css-cascade-6] Add ability to scope rules from an imported stylesheet


faceless2:
This is following on from a brief conversation with @tabatkins at CSSDay.

While I like the scoping proposal, I think it really needs a way to scope styles from an imported stylesheet. As it is now you can define your scopes - great for modularization - but the rules have to be inline in the same sheet, which seems to defeat the purpose a bit to me.

There are several ways to do this - an `@import` inside a `@scope` is one, but I think for consistency with cascade layers the syntax should be like

```css
@import uri(module.css) scope(.media-object) to (.content);
```

Also useful but slightly more controversial (it doesn't have the desired behaviour when the attribute is not recognised) would be adding a `scope` and (pending a better idea) a `scope-to` attribute to `<link>`, the same way that it now has `layer` (see https://github.com/whatwg/html/pull/7658)

```html
<link rel="stylesheet" href="module.css" scope=".media-object" scope-to=".content">
```

tabatkins:
Yeah, this seems reasonable to me. Might quibble a bit on the syntax, but being able to apply a scope to the rules in a stylesheet seems at least as useful as being able to put a stylesheet on a particular cascade layer.

DarkWiiPlayer:
I think an easier way of handling scope in HTML would simply be:

```html
<link rel="stylesheet" href="module.css" scope=".media-object to .content">
```

mirisuzanne:
My proposal would be that both a `scope()` function in CSS `@import`, and a `scope` attribute on the `<link>` element could use accept the full `@scope` condition syntax:

```css
@import uri(module.css) scope((.media-object) to (.content) inclusive);
```
```html
<link rel="stylesheet" href="module.css" scope="(.media-object) to (.content) exclusive">
```

The only edge case that might warrant a special-case is importing a scope with _only the scope root_. Without any exceptions, that would result in double-parentheses:

```css
@import uri(module.css) scope((.media-object));
```

---

I also want to recognize that adding attributes like this to HTML `<link>` also requires our proposed (but not yet fully specified or approved) update to the `<link media="">` attribute, so that it's possible to test for support of scoping, and only load the linked CSS file _if the scoping will be applied_. This is also needed for a proposed `layer` attribute.

There's a [PR in progress](https://github.com/whatwg/html/pull/7658/), but I could use help on some of the details.

bramus:
[...]

PR #11237, opened 2024-11-18
[css-cascade-6] Support @import scope() syntax

andruud:
#7348


andruud:
Note that this PR has two commits. The first commit copies over the @import section from level 5, verbatim.

css-cascade-6/Overview.bs
```diff
@@ -92,19 +92,23 @@ Importing Style Sheets: the ''@import'' rule</h2>
 
 <pre class='prod'>
 		@import [ <<url>> | <<string>> ]
-		        [ layer | layer(<<layer-name>>) ]?
-		        <<import-conditions>> ;
+		        [[ layer | layer(<<layer-name>>) ]
+		         || [ scope( [ (<<scope-start>>) ]? [ to (<<scope-end>>) ]? | <<scope-start>> ) ]
```
cdoublev, line 96:
```diff
- [ scope( [ (<<scope-start>>) ]? [ to (<<scope-end>>) ]? | <<scope-start>> ) ]
+ [ scope( [ [ (<<scope-start>>) ]? [ to (<<scope-end>>) ]? ]! | <<scope-start>> ) ]
```

(do not allow empty alternative)

andruud: But `@import scope()` is valid, though. It gives you an "implicit scope", i.e. equivalent to `@scope { }`.

cdoublev: Ok, that is intentional then. Sorry! =)

cdoublev:
Is it possible to edit Cascade 4 and 5 to prevent `w3c/reffy` from extracting [`<import-conditions>`](https://drafts.csswg.org/css-cascade-5/#typedef-import-conditions), which would be removed by this PR in Cascade 6?

```diff
- @import [ <url> | <string> ] ... <import-conditions> ;
- <import-conditions> = [ supports( [ <supports-condition> | <declaration> ] ) ]? <media-query-list>?
+ @import [ <url> | <string> ] ... [ supports( [ <supports-condition> | <declaration> ] ) ]? <media-query-list>? ;
```

---

Is it possible to define `<supports()>` and `<media()>` in Conditional 5, and `<scope()>` in Cascade 6, to avoid duplicate/overlapping definitions? 

For example, `supports()` is defined differently in [CSS Values 5](https://drafts.csswg.org/css-values-5/#typedef-if-test), [Conditional 5](https://drafts.csswg.org/css-conditional-5/#funcdef-supports), Cascade 4/5.

`<supports()>`, `<media()>`, `<media-query-list>`, would define the import conditions.

```
@import [ <<url>> | <<string>> ]
        [[ layer | layer(<<layer-name>>) ] || <scope()> || <supports()> || <media()>]?
        <<media-import-condition>>? ;

<scope()> = scope( [ <scope-selector> | <scope-start> ]? )
<supports()> = supports( [ <supports-condition> | <declaration> ] )
<media()> = media( <media-query-list> )
```

`<scope-selector>` would be defined below the definition of `@scope` with `@scope <scope-selector>? { ... }`.

At least, the grouping brackets no longer seem necessary around `scope()`, now that it is implicitly optional.

css-cascade-6/Overview.bs
romainmenke: Shouldn't this then be `@import scope` in parallel with `@import layer`?

I'd rather not have authors needing to remember:
| valid | invalid |
| --- | --- |
| `@import layer` | `@import layer()` |
| `@import scope()` | `@import scope` |

andruud: @romainmenke Ah, yes, that seems better.

andruud:
@cdoublev That does sound quite reasonable, but I'll wait for an editor to have an opinion on what this should look like.

`<media()>` would be spec'ed later / separately, since we don't have a resolution to add it yet.

---

Trying another editor, since @mirisuzanne is unresponsive.

css-cascade-6/Overview.bs
```diff
@@ -51,6 +51,227 @@ Introduction and Missing Sections</h2>
...
+		(such as the requirement that ''@namespace'' rules must not have any other rules other than
+		''@import'' preceding it),
+		it only applies between constructs in the same stylesheet.
+
+	<p class='example'>
+		For example, [=declarations=] in style rules from imported stylesheets interact with the cascade
+		as if they were written literally into the stylesheet at the point of the ''@import''.
+
+	Any ''@import'' rules must precede all other valid at-rules and style rules in a style sheet
+	(ignoring ''@charset'' and <a href="#layer-empty"><css>@layer</css> statement</a> rules)
```
mirisuzanne, line 88: Need `layer-empty` added to the `oldids` on line 46, or this PR doesn't build

css-cascade-6/Overview.bs
```diff
@@ -51,6 +51,227 @@ Introduction and Missing Sections</h2>
...
+		which [=scopes=] the [=style rules=] within the stylesheet,
+		using the [=scoping roots=] and [=scoping limits=]
+		as described by [[#scope-limits]].
+
+		Note: While the [=style rules=] within the imported stylesheet
+		become [=scoped=],
+		they do not become [=nested style rule|nested=].
+		In particular,
+		top-level selectors are not re-interpreted as [=relative selectors=],
+		and the ''&'' pseudo-class maintains its non-nested behavior.
```
mirisuzanne, line 130: I see the reasoning behind this, and think it's probably the best behavior - but it is a departure from `@scope`. I image we would also apply this behavior to any html `scope` attribute, so those would be consistent. Is it something that deserves it's own discussion?

mirisuzanne (commented):
I agree with the proposed change from Romain, and left a couple other notes. Overall it looks good.

css-cascade-6/Overview.bs
romainmenke: Is it correct that wrapping an entire stylesheet's contents with `@scope` would behave differently from importing that same stylesheet with a `scope()` condition?

I am unsure if that is the best option.
I think it would be surprising to authors as they would assume this to be purely a syntactic difference.

If the behavior is different it also implies that bundling becomes impossible, which might make the entire feature unusable in practice as it would force CSS authors to serve unbundled stylesheets and take the associated performance hit.

mirisuzanne: Fair points. It sounds like this does in fact deserve it's own issue. :)

andruud: > I see the reasoning behind this

I'm not sure I do (anymore). I remember concluding that anything else was impossible, but I don't remember why.

We'd have to allow relative selectors top-level, but we could do that and spec that they are relative against `:scope`.

(Re-interpreting `&` is deeply incompatible with Blink's implementation of nesting, but we don't have to let that guide the discussion.)

mirisuzanne:
`<import-conditions>` were given their own construction for the sake of reference from an HTML attribute, although I don't think that proposal is moving forward anymore.

css-cascade-6/Overview.bs
andruud: Done.

css-cascade-6/Overview.bs
andruud: Resolved in https://github.com/w3c/csswg-drafts/issues/11756.

css-cascade-6/Overview.bs
andruud: Done.

andruud:
@mirisuzanne How do we now feel about this? I refreshed this months ago, but forgot to follow up on it ...

andruud:
@cdoublev Let's see if we should do a follow-up here, since I left your comment hanging ...

> Is it possible to edit Cascade 4 and 5 to prevent `w3c/reffy` from extracting [`<import-conditions>`](https://drafts.csswg.org/css-cascade-5/#typedef-import-conditions), which would be removed by this PR in Cascade 6?

What bad situation would arise if we don't do this? Accidentally linkage to an obsolete dfn?

> Is it possible to define `<supports()>` and `<media()>` in Conditional 5, and `<scope()>` in Cascade 6, to avoid duplicate/overlapping definitions?

That amounts to a normative change that goes beyond this PR? Granted, these things _should_ be consistent, but we don't have a resolution right now to change css-conditional-5 (`@when`, specifically).

cdoublev:
> What bad situation would arise if we don't do this? Accidentally linkage to an obsolete dfn?

It would simply have been a convenience to prevent `@webref/css` from including "orphan" grammar entities.

If some level of a spec defines `@rule <foo> ;` and another level `@rule <bar> ;`, `@webref/css` will include a definition for `@rule`, `<foo>`, `<bar>`, with both `<foo>` and `<bar>` possibly defined with `"for": ["@rule"]`. So a human intervention is needed to determine that `<foo>` is obsolete and does not need to be documented on MDN, for example.

`<border-style>`, `<border-width>`, `<margin-width>`, `<padding-width>`, `<shape>`, are the only "orphan" types I know of, and they are all defined in CSS 2. That said, I can survive adding an exclusion of `<import-conditions>`. =)

  > That amounts to a normative change that goes beyond this PR? Granted, these things should be consistent, but we don't have a resolution right now to change css-conditional-5 (`@when`, specifically).

I had not suspected these differences could be intentional when I wrote this comment, indeed. w3c/csswg-drafts#12903 already exists for this.

PR #11239, opened 2024-11-19
link to absolute-position containing block

gitspeaks:
In this context, 'the original containing block' is the 'absolute-position containing block'


tantek (approved):

Loirooriol (changes requested):
Why are you changing CSS2 in a PR that says [css-position-3]?

AFAIK CSS2 is in a state where we are not sure what changes have happened since REC, and further divergence would make things messier.

Better stick to CSS Position.

gitspeaks:
> Why are you changing CSS2 in a PR that says [css-position-3]?
> 
> AFAIK CSS2 is in a state where we are not sure what changes have happened since REC, and further divergence would make things messier.
> 
> Better stick to CSS Position.

This was not intentional. I thought GitHub would prompt me. How can I now exclude the CSS2 changes from the pull request?

Loirooriol:
Reset your branch to the right commit and force-push it.

gitspeaks:
> Reset your branch to the right commit and force-push it.

Done.

gitspeaks:
> > AFAIK CSS2 is in a state where we are not sure what changes have happened since REC, and further divergence would make things messier.
> > Better stick to CSS Position.

The CSS2 section contains relevant information not present in the CSS Position section

Loirooriol (commented):
There are two concepts:
-  https://drafts.csswg.org/css-position/#original-containing-block
-  https://drafts.csswg.org/css-position/#absolute-position-containing-block

It seems like the spec was trying to link to the former one. However, testcase:

```html
<!DOCTYPE html>
<div style="display: inline-grid; grid-template: 50px 50px 50px / 50px 50px 50px; border: solid; position: relative">
  <div style="grid-area: 2 / 2 / 3 / 3; background: cyan; position: absolute; width: 100%; height: 100%"></div>
</div>
```

The percentages resolve to the grid area in Gecko, Blink and WebKit, so it's not the "original containing block".

So the edit seems good but it has behavior implications. Then I think it becomes substantive for IPR, and requires approval by editor discretion or CSSWG.

Loirooriol (approved):
OK, so some archeology:

 - The "[original containing block](https://drafts.csswg.org/css-position/#original-containing-block)" term was added in May 2024 in e0a332989bffe66a13d5bb35cbcf038267d189e8
 - The "[absolute-position containing block](https://drafts.csswg.org/css-position/#absolute-position-containing-block)" term was added in Aug 2024 in 046ce5282cc87a6b0d9a12ab47671bde9315061d
 - The reference to the "original [containing block](https://drafts.csswg.org/css-display-4/#containing-block)" was added in Sep 2021 in 40361aff4ebe0d8486f265560a6d82bd52598ec4

Therefore, despite using the same words, it wasn't referring to the [original containing block](https://drafts.csswg.org/css-position/#original-containing-block). Then I think this PR isn't affecting conformance after all.

But better let the editors confirm, I guess.

gitspeaks:

fantasai:
Thanks @gitspeaks and @Loirooriol for catching this.

PR #11266, opened 2024-11-23
[css-conditional] Fix link to CSS-CONTAIN-3

Psychpsyo:
Just a typo, that's all.

( = I believe this to be non-substantive so probably won't need to make a W3C account)


SebastianZ:
Thank you for the fix!

Sebastian

Issue #11097, opened 2024-10-28
[css-color-adjust] Should forced-colors support `color-mix()`?


cssence:
[CSS Color Adjustment Module Level 1](https://drafts.csswg.org/css-color-adjust-1/)

I realized that Firefox performs a `color-mix()` in Forced Colors Mode when all the colors in the mix are CSS System Colors:
https://cssence.com/2024/color-mix-in-forced-colors-mode/
https://codepen.io/cssence/pen/abeVWYK

Chromium-based browsers do not. After filing an issue ([Bugzilla ticket #1927252](https://bugzilla.mozilla.org/show_bug.cgi?id=1927252)), Mozilla responsed they implemented this on purpose, i.e. not a bug but a feature. We should clarify what the behavior should be, the Chromium-based way or the Firefox way.

kbabbitt:
[From the spec](https://drafts.csswg.org/css-color-adjust-1/#forced):

> Forced colors mode is an accessibility feature intended to increase the readability of text through color contrast. Individuals with limited vision often find it more comfortable to read content when there is a particular type of contrast between foreground and background colors.

Allowing mixing, even of system colors, seems to me like it would be counter to the intent of the feature.

nt1m:
I would prefer allowing mixing of system colors (which WebKit also supports fwiw), but also allow the UA to override it with `currentColor` when forced colors mode is enabled.

nt1m:
It isn't the job of CSS to prevent illegible/non-ideal colors from being created fwiw. But if the colors are non-ideal for forced colors mode, then forced colors mode should be able to override them.

kbabbitt:
> It isn't the job of CSS to prevent illegible/non-ideal colors from being created fwiw.

This is true, but it's orthogonal to the question of allowing mixing in forced colors mode. The goal of forced colors mode is to give the user control over foreground/background color combinations. The canonical use case is high contrast, but it can be any combination - some users might find low contrast more comfortable to read, for example. Mixing these colors, however, would not be applying them as they were set by the user.

nt1m:
Yeah, my point is that forced colors mode should be able to prevent the colors from being mixed, and override the mixed colors with appropriate colors.

svgeesus:
> if its [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value) is a color other than a system color, its [used value](https://drafts.csswg.org/css-cascade-5/#used-value) is instead forced to a system color as follows:
[...]

PR #11277, opened 2024-11-27
[css-color-adjust-1] Add example for `color-mix()` in Forced Colors Mode. #11097

kbabbitt:
The resolution was to close no change, but it was suggested to add an example illustrating the behavior.


svgeesus (approved):
Great, thanks!

PR #11287, opened 2024-11-27
[css-overflow-3] Correct scrollbar space description

gitspeaks:
The terms "inner" and "outer" are not defined for border and padding edges in [The CSS Box Model](
https://www.w3.org/TR/css-box-4/#box-model). These terms are unnecessary, undefined in this context, and their removal makes the description more clear.

### DIFF
--- a/css-overflow-3/Overview.bs
+++ b/css-overflow-3/Overview.bs
@@ -728,11 +728,11 @@ Scrollbars and Layout</h2>
 <h3 id="scrollbar-sizing">
 Scrollbar Contributions to Sizing</h3>
 
-	When reserving space for a scrollbar placed at the edge of an element's box,
-	the reserved space is inserted between the inner border edge
-	and the outer padding edge.
-	For the purpose of the <a spec=css-backgrounds>background positioning area</a> and <a spec=css-backgrounds>background painting area</a> however,
-	this reserved space is considered to be part of the [=padding box=].
+	The space reserved for the scrollbar is located between the 
+	padding area and the border area of the scroll container box.
+	However, for the purpose of the <a spec=css-backgrounds>background positioning area</a> 
+	and <a spec=css-backgrounds>background painting area</a>, this 
+	reserved space is treated as part of the [=padding area=].
 
 	<div class="example">
 		In the following document fragment,
--- a/css-overflow-3/Overview.bs
+++ b/css-overflow-3/Overview.bs
@@ -728,11 +728,11 @@ Scrollbars and Layout</h2>
 <h3 id="scrollbar-sizing">
 Scrollbar Contributions to Sizing</h3>
 
-	The space reserved for the scrollbar is located between the 
-	padding area and the border area of the scroll container box.
+	The space reserved for the scrollbar is located between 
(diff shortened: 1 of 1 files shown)


Loirooriol (commented):
It's true that these terms are not defined, but your proposal seems to imply that the scrollbar is placed on the border area.



CSS Overflow could be clearer but it's basically separating the padding edge from CSS Box into 2 edges: the outer padding edge (on the inside) and the inner border edge (on the outside). The area between them is where the scrollbar is placed, i.e. the scrollbar gutter.

gitspeaks:
> It's true that these terms are not defined, but your proposal seems to imply that the scrollbar is placed on the border area.

I'm not proposing anything. I just edited what is otherwise the only logical interpretation under the current definition of edges. If this results in an incorrect placement of the gutter, feel free to dismiss this pull request once proper definitions are in place.

>CSS Overflow could be clearer 

It’s not about clarity. The issue is that there’s simply no definition for "inner" or "outer" in the context of borders and padding.

>but it's basically separating the padding edge from CSS Box into 2 edges: the outer padding edge (on the inside) and the inner border edge (on the outside). The area between them is where the scrollbar is placed, i.e. the scrollbar gutter.

This doesn’t make sense to me. The border area is defined as the space between the padding edge and the border edge. What you’re calling "inner border" is actually the padding edge. Under the current definition of edges as perimeters, there is no additional "area between edges" beyond what is already specified.

Loirooriol:
> Under the current definition

I'm claiming that CSS Overflow is implicitly patching the CSS Box definition

gitspeaks:
> I'm claiming that CSS Overflow is implicitly patching the CSS Box definition

This patch seems to introduce a conceptual bug.

gitspeaks:

css-overflow-3/Overview.bs
```diff
@@ -728,11 +728,11 @@ Scrollbars and Layout</h2>
...
-	When reserving space for a scrollbar placed at the edge of an element's box,
-	the reserved space is inserted between the inner border edge
-	and the outer padding edge.
-	For the purpose of the <a spec=css-backgrounds>background positioning area</a> and <a spec=css-backgrounds>background painting area</a> however,
-	this reserved space is considered to be part of the [=padding box=].
+	The space reserved for the scrollbar is located between the 
+	padding area and the border area of the scroll container box.
+	However, for the purpose of the <a spec=css-backgrounds>background positioning area</a> 
+	and <a spec=css-backgrounds>background painting area</a>, this 
+	reserved space is treated as part of the [=padding area=].
```
fantasai:
```suggestion
	The space reserved for the scrollbar is located between 
	the [=padding area=] and the [=border area=] of the [=scroll container=] box.
	However, for the purpose of the <a spec=css-backgrounds>background positioning area</a> 
	and <a spec=css-backgrounds>background painting area</a>,
	this reserved space is treated as part of the [=padding area=].
```

fantasai (approved):
For future reference, we use https://rhodesmill.org/brandon/2012/one-sentence-per-line/ ; but other than that this looks good.

PR #11290, opened 2024-11-28
[css-sizing-3] Supplement the 'Applies to' field for width and heigh

OnlyWick:
`Applies to` lacks relevant description. Additionally, the additional information I provided may not be formal enough.

### DIFF
--- a/css-sizing-3/Overview.bs
+++ b/css-sizing-3/Overview.bs
@@ -370,7 +370,7 @@ Preferred Size Properties: the 'width' and 'height' properties</h4>
 	Name: width, height
 	Value: auto | <<length-percentage [0,∞]>> | min-content | max-content | <nobr>fit-content(<<length-percentage [0,∞]>>)</nobr> | <<width/calc-size()>>
 	Initial: auto
-	Applies to: all elements except <a>non-replaced</a> <a>inlines</a>
+	Applies to: all elements except <a>non-replaced</a> inline elements. 'width' cannot be applied to row groups
 	Inherited: no
 	Logical property group: size
 	Percentages: relative to width/height of <a>containing block</a>


OnlyWick:
[See this.](https://www.w3.org/TR/CSS22/visudet.html#the-width-property)

OnlyWick:
It seems that there are some 'Applies to' lacking corresponding descriptions. I will commit it later.

css-sizing-3/Overview.bs
```diff
@@ -370,7 +370,7 @@ Preferred Size Properties: the 'width' and 'height' properties</h4>
 	Name: width, height
 	Value: auto | <<length-percentage [0,∞]>> | min-content | max-content | <nobr>fit-content(<<length-percentage [0,∞]>>)</nobr> | <<width/calc-size()>>
 	Initial: auto
-	Applies to: all elements except <a>non-replaced</a> <a>inlines</a>
+	Applies to: all elements except <a>non-replaced</a> inline elements. 'width' cannot be applied to table rows, and row groups. 'height' cannot be applied to table columns, and column groups
```
Loirooriol, line 373: With the right `writing-mode`, `width` can indeed apply to table rows, and `height` to table columns and column groups.

```html
<!DOCTYPE html>
<table style="writing-mode: vertical-lr">
  <col style="height: 50px">
  <tr style="width: 100px">
    <td style="background: cyan"></td>
  </tr>
</table>
```

Loirooriol (changes requested):

css-sizing-3/Overview.bs
OnlyWick: You're right. But do we really want to put such a long sentence into 'Applies to'? Should we consider adding some examples below to illustrate these cases?

Loirooriol: No, I think it's better to leave "Applies to" as is. It would make sense for `inline-size` and `block-size`, but it would be a bit weird to have properties in a logical property group with different "Applies to". So maybe in prose.

OnlyWick:
I still want to add an Example below this property to indicate the impact of `writing-mode` on it.

Loirooriol (changes requested):
Same problem?

OnlyWick:
My mistake, I tested the row groups, but I neglected it.

OnlyWick:
My mistake, I misunderstood your meaning. I tested the row groups, but I neglected it. :-(

PR #11320, opened 2024-12-03
[css-overflow-5] Treat the last scroll target as active

### LINKED ISSUE #10738: [css-overflow-5] Allowing markers to be active even when not scrollable to aligned position
There are many situations in which it may not be possible to scroll to a position that aligns with the natural scrolling position of every marker. For example, in https://flackr.github.io/carousel/examples/carousel/thumbnails/ is is not possible to scroll to a position that centers the first two or last two items. E.g.

In the following scrolling element, it is not possible to scroll the first two boxes to their center snap aligned position:
scroller

This means that if the active marker is determined solely by the item closest to being aligned https://drafts.csswg.org/css-overflow-5/#scroll-container-scroll it will not be possible to show the first two / last two as active.

Should we consider a model where the browser tracks the actual element that has requested to be scrolled into view? This would allow the edge elements to show as active when clicking on those markers. This could have additional benefits if we treat this element as the anchor node for [anchor node selection](https://drafts.csswg.org/css-scroll-anchoring/#anchor-node-selection) as well. Note that subsequent user scrolling would likely undo the active styling.

Alternately, if we can't solve this in a nice way we could add a note recommending authors avoid these situations by:
* adding padding to the edges of their scrolling element, or
* ensuring their scroll marker elements are scrollport sized, or
* enabling infinite / cyclical scrolling as in #5411

flackr:
When a targeted scroll is performed, use that target to determine the active scroll marker. Fixes #10738.


css-overflow-5/Overview.bs
```diff
@@ -155,6 +155,28 @@ its content will be integrated into this specification,
...
+and is reset to <code>null</code> after any scrolling operations in that [=scroll container=]
+initiated by the user,
+or any script initiated operations that do not have a target,
+or when the target is removed from the document.
+Scrolling operations in the [=scroll container=] with a target Element or PseudoElement,
+set the [=current scroll target=] to that target Element or PseudoElement.
+
+Issue: Setting the current scroll target should be defined in how to <a>scroll a target into view</a>.
+
+Issue: Use the current scroll target as an <a>anchor priority candidate</a> for scroll anchoring.
```
flackr, line 178: I was trying to link this to the definition in css-scroll-anchoring, but it doesn't seem to be working. I'm sure I missed something simple.

css-overflow-5/Overview.bs
```diff
@@ -155,6 +155,28 @@ its content will be integrated into this specification,
...
+Every [=scroll container=] maintains a <dfn export>current scroll target</dfn>.
+It is initially <code>null</code>,
+and is reset to <code>null</code> after any scrolling operations in that [=scroll container=]
+initiated by the user,
+or any script initiated operations that do not have a target,
+or when the target is removed from the document.
+Scrolling operations in the [=scroll container=] with a target Element or PseudoElement,
+set the [=current scroll target=] to that target Element or PseudoElement.
+
+Issue: Setting the current scroll target should be defined in how to <a>scroll a target into view</a>.
```
flackr, line 176: Do you think I should make these edits in the other specs with this PR or leave them as issues here?

tabatkins: Can leave as issues for the moment.

css-overflow-5/Overview.bs
tabatkins: Scroll Anchoring defines the term as just "priority candidates", which isn't great. I'll go fix that spec, it should really be exporting as "anchor priority candidate"; this will start automatically linking then.

flackr: Great, tested and it is now linking.

css-overflow-5/Overview.bs
flackr: Sounds good!

PR #11323, opened 2024-12-04
[css-inline-3] Clarify the condition when inline formatting context is established

gitspeaks:
A block container establishes an inline formatting context only when it directly contains exclusively inline-level content


fantasai:
To the extent that block boxes can contain only inline-level or only block-level content, the "only" is implied. And if the model changes to describe a mix somehow (which IIRC is what browser implementations actually do), then the "only" is no longer true... so altogether I think we shouldn't take this change.

PR #11351, opened 2024-12-11
[css-color-hdr-1] Change dynamic-range-limit-mix syntax to use comma separation #11349

ccameron-chromium:


svgeesus:
Already done in https://github.com/w3c/csswg-drafts/pull/11349

PR #11359, opened 2024-12-12
[css-values-5] Allow explicit null-namespace in attr-name

andruud:
Otherwise we're making a non-backwards-compatible change against content:attr() from css-content-3 [1].

[1] https://drafts.csswg.org/css-content-3/#strings


tabatkins:
Thanks, that was definitely an accident.

Issue #11385, opened 2024-12-17
[css-gaps-1] Duplicate definition of repeat()


tidoust:
CSS Gaps defines a [`repeat()` function](https://drafts.csswg.org/css-gaps-1/#funcdef-repeat) already defined in [CSS Grid Layout](https://drafts.csswg.org/css-grid-2/#funcdef-repeat). Both definitions are global, so that creates a duplicate, see for example the [cross-referencing database](https://respec.org/xref/?term=repeat%28%29&types=css-function).

Is the definition in CSS Gaps meant to supersede the one in CSS Grid Layout?
If not, it needs to be [namespaced](https://speced.github.io/bikeshed/#dfn-for) somehow.

cdoublev:
I think it would make sense to namespace *both* definitions: the one from CSS Grid `for="<track-repeat>,<auto-repeat>,<fixed-repeat>,<name-repeat>"` and the one from CSS Gap `for="<repeat-line-color>,<auto-repeat-line-color>"`.

kbabbitt:
Thanks for pointing this out. Created #11388 to namespace in both specs.

PR #11388, opened 2024-12-17
[css-grid-2][css-gaps-1][editorial] Namespace repeat() definitions. #11385

kbabbitt:


kbabbitt:
@tabatkins could you ptal and let me know if this looks right?

Issue #10995, opened 2024-10-03
[css-view-transitions-2] Allow an auto-generated `view-transition-name` that doesn't default to ID


noamr:
In #8320, we resolved on using `view-transition-name: auto` as a way to generate names, using the element's id if exists.
In the resolution, we said that we'll have another value for generating the name without falling back to ID, and perhaps yet another one for generating it *only* from ID.

Proposing:
- Use `view-transition-name: self` as the keyword for generating the name ignoring ID. Other proposals that were raised were `from-element` and `element-uuid()`.
- No need for "generate from ID", once we have `attr(<ident> id)`.

bramus:
Copying over [my comment from 8320](https://github.com/w3c/csswg-drafts/issues/8320#issuecomment-2389511249) for visibility:

> One thing I don’t like about `from-element` is that it reads like _“read the id from the element”_ which authors will interpret as _“read the id attribute from the element”._ Ideally the value makes it clear that it’s an internal `id`, unrelated to the `id` attribute.
>
> I think `element-uuid` as suggested in the OP comes closest to this.

noamr:
> Copying over [my comment from 8320](https://github.com/w3c/csswg-drafts/issues/8320#issuecomment-2389511249) for visibility:
> 
> > One thing I don’t like about `from-element` is that it reads like _“read the id from the element”_ which authors will interpret as _“read the id attribute from the element”._ Ideally the value makes it clear that it’s an internal `id`, unrelated to the `id` attribute.
> > I think `element-uuid` as suggested in the OP comes closest to this.

Any thoughts on `self`?

bramus:
Not a big fan of it, but also not against it.

---

I recently read up on [`first-valid()`](https://www.w3.org/TR/css-values-5/#first-valid) – which I didn’t know existed up until then.

This give me a new idea _(which would alter the resolution from #8320)_: what if there was a function that allowed authors to list certain options and it returns the first non empty one?

- `auto` = auto generate a name _(element identity)_.
- `attr(id "<ident>")` = use id attribute
- `first-non-empty(attr(id "<ident>"), auto)` = try getting id attribute first. if there is none, fall back to `auto` behavior.

That would prevent a lot of confusion on the author side, as `auto` here now means “auto-generate a v-t-name”.

[...]

PR #11393, opened 2024-12-18
[css-view-transitions-1][css-view-transitions-2] Add view-transition-name: match-element     

noamr:
`match-element` works the same as `auto`, except doesn't try to use the ID as the name.
   It is excluded as a custom-ident in css-view-transitions-1.
  Resolution: https://github.com/w3c/csswg-drafts/issues/10995#issuecomment-2551949161

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -791,13 +791,20 @@ li:nth-child(4) { view-transition-name: item1; }
 ...
 ```
 
-With ''view-transition-name/auto'', this CSS would work:
+With ''view-transition-name/auto'' or ''view-transition-name/match-element'', this CSS would work:
 
 ```css
 li {
 	view-transition-name: auto;
+	/* or */
+	view-transition-name: match-element;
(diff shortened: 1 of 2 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -791,13 +791,20 @@ li:nth-child(4) { view-transition-name: item1; }
...
 li {
 	view-transition-name: auto;
+	/* or */
+	view-transition-name: match-element;
 }
 ```
+
+The ''view-transition-name/auto'' and ''view-transition-name/match-element'' keywords have a minor
+difference, where ''view-transition-name/auto'' would use the element's [=Element/id=] as the 'view-transition-name'
+if exists, making it potentially work across documents, while ''view-transition-name/match-element'' only matches
```
bramus, line 806: I think there’s a word missing?

```suggestion
if it exists, making it potentially work across documents, while ''view-transition-name/match-element'' only matches
```

bramus (approved):

PR #11421, opened 2024-12-30
[scroll-animations-1] Add examples

bramus:
Add some examples to the scroll-animations-1 spec.


bramus:
I believe this PR is ready for review.

One thing I struggled with is there to place the examples for `animation-range`. I chose **not** to put them under [Appendix A](https://drafts.csswg.org/scroll-animations-1/#timeline-ranges) because the entire section is marked as _“This section should move to CSS-ANIMATIONS-2 and WEB-ANIMATIONS-2.”_.

If I were an author, I would like to see the examples that use `animation-range` with scroll-driven animations in the `scroll-animations-1` spec. Because of that, I have put the examples under the https://drafts.csswg.org/scroll-animations-1/#scroll-driven-attachment section which mentions restricting the active interval.

As a result, though, the section itself might need some rewording / extra sentences which I can tackle later.

svgeesus (approved):
LGTM

svgeesus:
> As a result, though, the section itself might need some rewording / extra sentences which I can tackle later.

That should not hold up merging this PR however

fantasai (commented):
Some minor comments. General comment is, the goal of example text is to step away from the spec's precision and generality and help the reader see concretely what it does, using the example as an illustration of what the spec means in more practical and illustrative terms.

scroll-animations-1/Overview.bs
```diff
@@ -238,6 +238,93 @@ spec:selectors-4; type:dfn; text:selector
 	References to the [=root element=] propagate to the document viewport
 	(which functions as its [=scroll container=]).
 
+	<div class=example>
+		Consider the following style sheet and markup:
+
+		```css
+		@keyframes grow {
+			to { scale: 2; }
```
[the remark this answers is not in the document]
bramus: Good call. Updating.

scroll-animations-1/Overview.bs
```diff
@@ -238,6 +238,93 @@ spec:selectors-4; type:dfn; text:selector
...
+
+		The `.subject`‘s animation is set up to be driven by
+		an anonymous [=Scroll progress timeline=] created with ''scroll()''.
+		Because no <<scroller>> or <<axis>> are passed into the function,
+		the default values of ''nearest'' and ''block'' respectively are used.
+
+		Because `.scroller` is the nearest ancestor [=scroll container=],
+		this results in the `.scroller` element driving the animation: 
+		as you scroll `.scroller` up and down, the `grow` animations’s progress
+		moves forwards or backwards in direct response.
```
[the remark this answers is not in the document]
bramus: Added.

scroll-animations-1/Overview.bs
```diff
@@ -770,6 +999,85 @@ spec:selectors-4; type:dfn; text:selector
 	can also be further restricted to a particular timeline range
 	(see [[#named-range-animation-declaration]]).
 
+	<div class=example>
+		In the following example each child of the `.scroller` element
+		reveals itself as it enters the scrollport instead of when entirely crossing it.
```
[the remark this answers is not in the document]
bramus: I added the clarification of what `view()` does to the example that introduced it _(at around line 750)_. .

bramus:
Addressed the comments. Re-requesting a review.

scroll-animations-1/Overview.bs
```diff
@@ -238,6 +238,99 @@ spec:selectors-4; type:dfn; text:selector
...
+		an anonymous [=Scroll progress timeline=] created with ''scroll()''.
+		Because no <<scroller>> or <<axis>> are passed into the function,
+		the default values of ''nearest'' and ''block'' respectively are used.
+
+		Because `.scroller` is the nearest ancestor [=scroll container=],
+		this results in the `.scroller` element driving the animation: 
+		as you scroll `.scroller` up and down, the subject’s animations progress
+		moves forwards or backwards in direct response.
+
+		This results in the `background-color` to be
```
bramus, line 281:
```suggestion
		This results in the `background-color` being
```

scroll-animations-1/Overview.bs
```diff
@@ -560,6 +734,38 @@ spec:selectors-4; type:dfn; text:selector
...
+			animation: reveal linear both;
+			animation-timeline: view();
+		}
+		```
+
+		For every element matched by the selector,
+		a [=view progress timeline=] gets created to drive the animation.
+		Because no arguments are passed into ''view()'' it uses the default values
+		for <<axis>> and <<'view-timeline-inset'>>,
+		resulting in the scroller being tracked in the ''block'' direction.
```
bramus, line 756:
```suggestion
		thus tracking its scroll position in the [=block axis=].
```

scroll-animations-1/Overview.bs
```diff
@@ -670,6 +876,27 @@ spec:selectors-4; type:dfn; text:selector
 	The values of {{ViewTimeline/subject}}, {{ScrollTimeline/source}}, and {{AnimationTimeline/currentTime}}
 	are all computed when any of them is requested or updated.
 
+	<div class=example>
+		In the following example, each child of the `.scroller` element
+		will reveal itself as it crosses the scrollport.
```
bramus, line 881: TODO: Make more explicit that this is the same as the CSS example

scroll-animations-1/Overview.bs
```diff
@@ -1054,6 +1352,87 @@ spec:selectors-4; type:dfn; text:selector
 
 	ISSUE: Define application to time-driven animations.
 
+### Examples
+
+	<div class=example>
+		In the following example, each direct child of the `.scroller` element
+		reveals itself as it enters the scrollport instead of when entirely crossing it.
```
bramus, line 1359: Clarify what this looks like from the user's perspective: it's opaque once fully visible.

scroll-animations-1/Overview.bs
```diff
@@ -1054,6 +1352,87 @@ spec:selectors-4; type:dfn; text:selector
...
+		.scroller > * {
+			animation: reveal linear both,
+			           hide linear forwards;
+			animation-timeline: view();
+			animation-range: entry, exit;
+		}
+		```
+
+		The `reveal` effect is linked to the ''entry'' range
+		while the `hide` effect is linked to the ''exit'' range.
```
bramus, line 1398: TODO: Add extra explanation about the visual output

bramus (commented):
Went over the PR with @fantasai. Added remarks together for me to resolve.

flackr (approved):
Thanks for the examples and details. This will be really helpful.

PR #11433, opened 2025-01-03
Update Open Font Format href in biblio.ref

JohnRiv:
I noticed the current href for `OPEN-FONT-FORMAT` (https://standards.iso.org/ittf/PubliclyAvailableStandards/c052136_ISO_IEC_14496-22_2009%28E%29.zip) returns an HTML page with this message:

> The page you're looking for has either moved or no longer exist.

I was able to find the updated URL in the ISO Standards website here: https://www.iso.org/standard/74461.html

When this PR is merged, it will help to resolve https://github.com/tobie/specref/issues/843

### DIFF
--- a/biblio.ref
+++ b/biblio.ref
@@ -3174,7 +3174,7 @@
 %T Information technology — Coding of audio-visual objects — Part 22: Open Font Format
 %I International Organization for Standardization.
 %R ISO/IEC 14496-22:2009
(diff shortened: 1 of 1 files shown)


tabatkins:
That file (`biblio.ref`) isn't used by anything any longer, and should be removed from the repo. All our spec tooling uses SpecRef data exclusively.

Feel free to make the edit in SpecRef; I can review the PR.

PR #11446, opened 2025-01-06
[css-mixins-1] Describe custom function cycles

andruud:


css-mixins-1/Overview.bs
```diff
@@ -412,7 +413,78 @@ is the [=guaranteed-invalid value=].
...
+may reference other [=custom functions=] or [=custom properties=],
+and may therefore create [[css-variables-1#cycles|cycles]].
+
+For each element, add a node for every specified [=custom function=]
+to the graph described in [[css-variables-1#cycles]];
+add a node for each [=local variable=]
+defined within each of those functions;
+then, for each [=custom function=] <var>func</var>, add edges as follows:
+
+* Between <var>func</var> and any [=custom function=]
```
andruud, line 427: I now realize this a _directed_ dependency graph, so probably we should not use "between A and B", but instead "from A to B". css-variables-1 has the same issue.

andruud: Also, I'm missing some edges ..

andruud: Done.

PR #11453, opened 2025-01-07
[web-animations-1] Fix markup

cdoublev:
Fixes the numbering of the list items in this procedure: https://drafts.csswg.org/web-animations-1/#animation-silently-set-the-current-time


birtles (approved):
Thank you!

PR #11464, opened 2025-01-09
[css-shapes-2] by/to should always come before rest of parameters.

### LINKED ISSUE #10666: [css-shapes-2] Order of points and control points in `shape()`

noamr:
This applies to arc/curve/smooth.

Also fixes a bug where the arc target point was using only coordinate-pair.

Resolution: https://github.com/w3c/csswg-drafts/issues/10666#issuecomment-2578298392

Closes #10666

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -318,7 +318,7 @@ The ''shape()'' Function</h4>
 			the command's starting point, the command's end point, or the [=reference box=], respectively.
 			If such component is not provided, the <<coordinate-pair>> is relative to the segment's start.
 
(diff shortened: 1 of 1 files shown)


noamr:
> The `&&` in the other segment types has to change too.

Which ones? Only curve, smooth & arc have additional parameters.

smfr (approved):
Oh, I confused this with https://github.com/w3c/csswg-drafts/issues/10666. Your change is fine for this issue.

noamr:
> Oh, I confused this with #10666. Your change is fine for this issue.

You actually weren't, I forgot to push one of the commits.

PR #11473, opened 2025-01-10
[css-shapes-2] Fix syntax typos

### LINKED ISSUE #11368: [css-shapes-2] Minor problems with the new `<arc-command>`
Its [production](https://drafts.csswg.org/css-shapes-2/#ref-for-typedef-shape-arc-command%E2%91%A0) requires specifying the size with two `<length-percentage>`:

  > ```
  > <arc-command> = arc [ 
  >      <command-end-point>
  >   && [ of <coordinate-pair> ]
  >   && <arc-sweep>? 
  >   && <arc-size>? 
  >   && [rotate <angle>]? 
  >]
  > <coordinate-pair> = <length-percentage>{2}
  > ```

But then its [contextual definition](https://drafts.csswg.org/css-shapes-2/#typedef-shape-arc-command) has a different value definition and a prose that allows omitting the second component value:

  > `<arc-command> = arc [[<by-to> <coordinate-pair>] || [of <length-percentage>{1,2}] || <arc-sweep>? || <arc-size>?|| rotate <angle>? ]`
  >
  > [...] The `of` component specifies the size of the ellipse that the arc is taken from. The first `<length-percentage>` provides the horizontal radius of the ellipse and the second provides the vertical radius. [...]
  >
  > If only one `<length-percentage>` is provided, both radiuses use the provided value. In that case, `<percentage>` values are resolved against the direction-agnostic size of the reference box (similar to the circle() function).

cdoublev:
Follow-up to 233bbec, 8b76107, and partially fixes #11368.

### DIFF
--- a/css-shapes-2/Overview.bs
+++ b/css-shapes-2/Overview.bs
@@ -173,7 +173,7 @@ The ''shape()'' Function</h4>
 
 	<pre class=prod>
 		<<shape-command>> = <<move-command>> | <<line-command>> | close |
-		                  <<horizontal-line-command>> | <<vertical-line-command>> | 
+		                  <<horizontal-line-command>> | <<vertical-line-command>> |
 		                  <<curve-command>> | <<smooth-command>> | <<arc-command>>
 
 		<<move-command>> = move <<command-end-point>>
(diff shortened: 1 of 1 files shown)


cdoublev:
Superseded by #11488.

PR #11487, opened 2025-01-13
[css-grid-2] Gloss in-flow/out-of-flow jargon

cvrebert:
This hyperlinks the terms "in-flow" & "out-of-flow" to their formal definitions for clarity.
Currently Bikeshed links it to https://drafts.csswg.org/css-display-4/#in-flow etc. If the CSS2 definition is preferred instead, just let me know.

----

Note to self:
```
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:css-display-4; type:dfn; text:in-flow
spec:css21; type:dfn; text:in-flow
```

### DIFF
--- a/css-grid-2/Overview.bs
+++ b/css-grid-2/Overview.bs
@@ -1059,9 +1059,9 @@ Limiting Large Grids</h3>
 Grid Items</h2>
 
 	Loosely speaking, the <dfn export id="grid-item" lt="grid item">grid items</dfn> of a <a>grid container</a>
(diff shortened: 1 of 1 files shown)


cvrebert:
I am very amenable to deeming this a non-substantive change, but IDK whether the W3C would pedantically deem the hyperlinking of a CSS2-era term to be ["resolv[ing an] ambiguit[y]".](https://www.w3.org/policies/process/#class-2)

svgeesus (approved):

PR #11512, opened 2025-01-16
[css-values-5] Drop duplicate definition of ident()

tidoust:
Previous update introduced a definition of the `ident()` function in the prose. The definition in the production block needs to be dropped not to duplicate it.


bramus (approved):

bramus:
_(FYI @tabatkins, as they added that extra dfn at the last minute)_

Issue #11317, opened 2024-12-03
[css-pseudo] Allow custom Properties in ::selection pseudos (revisited)


schenney-chromium:
Chromium continues to try to ship the Highlight Inheritance model for selection. The third attempt has failed because CSS pre-processing and generation tools create `::selection` rules like this:
```html
div::selection {
   --my-opacity: 1;
   background-color: rgb(0 0 0/var(--my-opacity));
}
```
In https://github.com/w3c/csswg-drafts/issues/6264 we allowed custom properties on highlight pseudos.

Then in  https://github.com/w3c/csswg-drafts/issues/6641 we also said that all custom properties should be copied from the root for the `root::selection`, and then start adding those in the `::selection` inheritance path.

This proved insufficient when launched in Chromium because many sites rely on `::selection` inheriting custom properties from their originating elements. So we discussed disallowing custom properties in highlights and inheriting from the originating elements, and resolved to do that in https://github.com/w3c/csswg-drafts/issues/9909.

@tabatkins commented in https://github.com/w3c/csswg-drafts/issues/6641 that the example above must be allowed for their support. @LeaVerou also raised the same concern in one of the issues above. I did not pay enough attention to that. So the change from #9909 broke many many major sites.

To make it possible to ship highlight inheritance for `::selection`, and bring `::selection` implementations into line with other highlight pseudos, I believe we need this ...

*`::selection` pseudo elements inherit custom properties from their `originating-element` and then apply custom properties from the `::selection` rule itself. Custom properties from `::selection` rules apply only within that rule, and are not inherited by descendants.*

My current thinking is that we limit this behavior to `::selection` and not other highlight pseudos, as a back-compatibility requirement. It's not a strongly held position.

Proposing async resolution but feel free to add to the agenda. 4 week time limit given holidays.

@delan @fantasai @frivoal

schenney-chromium:
I should note that implementing the proposed behavior is likely straight-forward and efficient (it's a 2 line change in Chromium). Implementing dual inheritance paths, through both originating and highlight trees, is likely very bad for performance,

astearns:
I am a bit concerned about making `::selection` work differently than other highlight pseudos. As use of the other pseudos rises, won’t people be creating similar rules for them with pre-processors and expect those to work like `::selection`?

schenney-chromium:
I'm totally fine with making it apply to all highlight pseudos.

[...]

PR #11528, opened 2025-01-18
Custom properties highlights

schenney-chromium:
[CSS-Pseudo-4] Correct custom properties in highlights
As was resolved in https://github.com/w3c/csswg-drafts/issues/11317, CSS Highlight Pseudos should allow custom property definitions within the rule itself, but take any missing custom properties from the originating element.

This PR updates the spec to re-list custom properties as supported on highlights. Furthermore, it fixes the description of custom property inheritance that was wrong even for the prior version of the spec.


css-pseudo-4/Overview.bs
```diff
@@ -910,6 +910,8 @@ Styling Highlights</h3>
     <wpt>
     textpath-selection-011.html
     </wpt>
+
+    <li>[=custom properties=] [[!CSS-VARIABLES-1]], but these are not inherited
```
andruud, line 914: This special inheritance behavior should be described in the normative text of `3.5. Cascading and Per-Element Highlight Styles` IMO.

css-pseudo-4/Overview.bs
```diff
@@ -1201,18 +1203,19 @@ Cascading and Per-Element Highlight Styles</h3>
...
-    A div's selection highlight would be a green background to the
-    selected content, with a 1px thick blue underline. The
-    ''--background-color'' custom property is not found on the
-    ''div::selection'' nor ''::selection'' so it is taken from the
-    ''::root'' custom property match. The ''--decoration-thickness''
-    property is inherited from ''::selection'' via the highlight
-    cascade and ''--decoration-color'' is found in
-    ''div::selection'' itself.
+    The universal selection uses the user-agent styling because it only defines
+    custon properties, with no properties that influence the appearance.
```
andruud, line 1207: nit: "custon"

css-pseudo-4/Overview.bs
```diff
@@ -1201,18 +1203,19 @@ Cascading and Per-Element Highlight Styles</h3>
...
+    A div's selection highlight must be a green background to the
+    selected content, with a 10px thick blue underline. The
+    ''--background-color'' and ''--decoration-thickness'' custom properties
+    are not found on the ''div::selection'' so come from the originating div element,
+    which itself inherits the custom properties from the root.
+    The ''--decoration-color'' is found in the ''div::selection'' itself,
+    so the color from the originating elemnent is not used.
```
andruud, line 1214: nit: Are we not doing [sembr](https://sembr.org/) in this spec? 

Can we at least not end lines with "the"?

css-pseudo-4/Overview.bs
schenney-chromium: In hindsight, yes.

css-pseudo-4/Overview.bs
schenney-chromium: I had no idea that existed. Looks like writing as poetry, but in principle I get it. I'll address the "the".

css-pseudo-4/Overview.bs
schenney-chromium: Done.

css-pseudo-4/Overview.bs
schenney-chromium: Done.

css-pseudo-4/Overview.bs
```diff
@@ -1201,18 +1202,19 @@ Cascading and Per-Element Highlight Styles</h3>
...
-    A div's selection highlight would be a green background to the
-    selected content, with a 1px thick blue underline. The
-    ''--background-color'' custom property is not found on the
-    ''div::selection'' nor ''::selection'' so it is taken from the
-    ''::root'' custom property match. The ''--decoration-thickness''
-    property is inherited from ''::selection'' via the highlight
-    cascade and ''--decoration-color'' is found in
-    ''div::selection'' itself.
+    The universal selection uses the user-agent styling because it only defines
+    custom properties, with no properties that influence the appearance.
```
fantasai, line 1206: Clarify that it's the default styling. Rewrap on phrase boundaries.

```suggestion
    The universal ''::selection'' uses the user-agent's default styling
    because it only defines custom properties,
    with no properties that influence the appearance.

```

css-pseudo-4/Overview.bs
```diff
@@ -1201,18 +1202,19 @@ Cascading and Per-Element Highlight Styles</h3>
...
+    A div's selection highlight must be a green background to the  selected content,
+    with a 10px thick blue underline.
```
fantasai, line 1208: Avoid RFC2119 terminology in examples.

```suggestion
    A div's selection highlight would apply a green background to the selected content,
    with a 10px thick blue underline.
```

css-pseudo-4/Overview.bs
```diff
@@ -1201,18 +1202,19 @@ Cascading and Per-Element Highlight Styles</h3>
...
+    The ''--background-color'' and ''--decoration-thickness'' custom properties
+    are not found on the ''div::selection'' so come from the originating div element,
```
fantasai, line 1210:
```suggestion
    Since ''--background-color'' and ''--decoration-thickness'' custom properties
    are not specified on the ''div::selection'' peudo-element,
    they are inherited from its originating <{div}> element,
```

css-pseudo-4/Overview.bs
```diff
@@ -1201,18 +1202,19 @@ Cascading and Per-Element Highlight Styles</h3>
...
+    which itself inherits the custom properties from the root.
+    The ''--decoration-color'' is found in the ''div::selection'' itself,
+    so the color from the originating elemnent is not used.
```
fantasai, line 1213:
```suggestion
    However since ''--decoration-color'' is specified on the ''div::selection'' itself,
    so its value from the [=originating element=] is not used.
```

fantasai (approved):
Looks good overall. Made some suggestions for adjustments. Recommend adding an entry to the changelog (or modifying an existing one on the topic, if there's one already).

Please SQUASH and rebase your changes before merging.

PR #11540, opened 2025-01-20
[css-contain-2] Grammar improvements

Psychpsyo:
It is "then not ...", not "not then ...", and especially not "not them ...".


SebastianZ:
Thanks for the fix, @Psychpsyo!

Sebastian

PR #11554, opened 2025-01-22
[css-flexbox-1] Fix fallback alignment for align-content:stretch

aethanyc:
The commit b6c01a41bb375777763b913006a64af87e5fa73e meant to edit only `space-*` values. Also, CSS Align spec has `align-content:stretch` fallback value as `flex-start` per
https://drafts.csswg.org/css-align-3/#valdef-align-content-stretch


aethanyc:
I've rebased this change to resolve conflicts. Ready for merge.

PR #11566, opened 2025-01-24
[css-conditional-5] Clarify how container query length units are evaluated

Psychpsyo:
This doesn't change how they work, it just makes it more clear.
Saying that "[container query length](https://drafts.csswg.org/css-conditional-5/#container-query-length) units are evaluated as [container size queries](https://drafts.csswg.org/css-conditional-5/#container-size-query) on the relevant axis" didn't make much sense, since a container query length unit needs to evaluate to a pixel amount while a container size query evaluates to a boolean or unknown.

### DIFF
--- a/css-conditional-5/Overview.bs
+++ b/css-conditional-5/Overview.bs
@@ -1581,9 +1581,9 @@ Container Relative Lengths: the ''cqw'', ''cqh'', ''cqi'', ''cqb'', ''cqmin'', '
 	</table>
 
 	For each element,
-	[=container query length=] units are evaluated
-	as [=container size queries=] on the relevant axis (or axes)
-	described by the unit.
+	[=container query length=] units are evaluated using the same
+	rules as [=container size queries=] when querying the size
+	on the relevant axis (or axes) described by the unit.
 	The [=query container=] for each axis
 	is the nearest ancestor container
 	that accepts [=container size queries=] on that axis.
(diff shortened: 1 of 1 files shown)


Psychpsyo:
@dbaron I hope you're the right person to ping about this since you're listed as a spec editor on the spec.

This small wording improvement has been sitting for almost a year at this point.

css-conditional-5/Overview.bs
```diff
@@ -1581,9 +1581,9 @@ Container Relative Lengths: the ''cqw'', ''cqh'', ''cqi'', ''cqb'', ''cqmin'', '
 	</table>
 
 	For each element,
-	[=container query length=] units are evaluated
-	as [=container size queries=] on the relevant axis (or axes)
-	described by the unit.
+	[=container query length=] units are evaluated using the same
+	rules as [=container size queries=] when querying the size
+	on the relevant axis (or axes) described by the unit.
```
fantasai: I think this is a little more direct?
Also rewrapping to match https://rhodesmill.org/brandon/2012/one-sentence-per-line/

```suggestion
	[=container query length=] units are evaluated
	using the same rules as [=container size queries=]
	on the relevant axis (or axes) described by the unit.
```

fantasai:
@Psychpsyo Merged in with some modifications. If you think it needs something else, let us know. (Though maybe in a new issue/PR/email, since this one is now closed...)

PR #11579, opened 2025-01-27
[*][editorial] Drop autolinks to value in `<pre>`

cdoublev:
`<pre>''keyword''</pre>` no longer links to `keyword` definition, which means `''keyword''` now generates `'keyword'`, which is inappropriate.

Related: https://github.com/speced/bikeshed/issues/3011

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -3734,9 +3734,9 @@ Specifying Predefined Colors: the ''color()'' function</h3>
 		<dfn>color()</dfn> = color( <<colorspace-params>> [ / [ <<alpha-value>> | none ] ]? )
 		<dfn>&lt;colorspace-params></dfn> = [ <<predefined-rgb-params>> | <<xyz-params>>]
 		<dfn>&lt;predefined-rgb-params></dfn> = <<predefined-rgb>> [ <<number>> | <<percentage>> | none ]{3}
-		<dfn>&lt;predefined-rgb></dfn> = ''srgb'' | ''srgb-linear'' | ''display-p3'' | ''a98-rgb'' | ''prophoto-rgb'' | ''rec2020''
+		<dfn>&lt;predefined-rgb></dfn> = srgb | srgb-linear | display-p3 | a98-rgb | prophoto-rgb | rec2020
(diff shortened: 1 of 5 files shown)


svgeesus (approved):
I guess, since this no longer works :(

cdoublev:
As noted in the linked issue, `<l>''value''</l>` *should* work and some find these links useful. I have no opinion on this. Ideally, any symbol (keyword, function, type) in a production rule should probably be automatically wrapped in a link pointing to its definition, if any.

cdoublev:
Sorry, I am no longer sure that this PR was appropriate, based on 85e0ca0a90427eb70d3aecb9a0133c67f645fb0e from a few hours before. ¯\_(ツ)_/¯

Please feel free to revert it.

tabatkins:
It only ever really worked accidentally; I didn't process any of the autolink syntaxes inside of "opaque" elements, of which `<pre>` is one (also script/style/xmp). That is, `<pre>[=foo=]</pre>` was not processed, it was left as plain text. It just happened that `<<foo>>` and `''foo''` syntaxes were handled by a completely different parsing layer (just regexes over the source) and thus didn't know what the context was yet, so they worked.

The new parser handles everything consistently. I did still carve out an exception for `<<foo>>`, allowing that to still link, because it's (a) already *extremely* heavily used, (b) obviously useful (non-terminals *often* link to a distant part of the spec or different specs entirely), and (c) impossible for it to accidentally be an innocent usage (because it's super invalid HTML), unlike single quotes.

The opt-out for triggering autolinks within opaques is to use the `<l>` element (a bikeshed-specific element meaning "link"). But I realize that `<l>''foo''</l>`, specifically, doesn't work; that's a bug on my part. (Again, wasn't necessary in the old parser.) Let me go fix that.

tabatkins:
And it's fixed. `<l>''foo''</l>` now works, as it always should have (and a bare `''foo''` in a `<pre>` doesn't, as it always shouldn't have)

cdoublev:
Thanks for the follow up. 

@svgeesus, do you want me to revert this PR and replace `''keyword''` with `<l>''keyword''</li>`?

tabatkins:
Yes, please.

svgeesus:
Yes, please do.

PR #11585, opened 2025-01-28
[css-mixins-1] Fix build error

bramus:
This PR fixes the following two issues with the css-mixins-1 spec:

- Build errors when building the spec, caused by rogue `</pre>` tags.
- ~~Add the opening paren after `@function`. I believe this was erroneously removed in fe94b66b.~~

### DIFF
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -578,7 +578,6 @@ or acting as if nothing exists at that location otherwise.
 	The value of the '@function/result' descriptor
 	is <code>20px</code> if the media query's condition is true,
 	and <code>16px</code> otherwise.
-	</pre>
 </div>
 
 <div class='example'>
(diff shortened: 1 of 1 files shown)


css-mixins-1/Overview.bs
```diff
@@ -104,7 +104,7 @@ The ''@function'' rule defines a [=custom function=],
 and its syntax is:
 
 <pre class="prod def" nohighlight>
-&lt;@function> = @function <<function-name>> <<function-parameter-list>>? )
+&lt;@function> = @function <<function-name>>( <<function-parameter-list>>? )
```
andruud, line 107: `<<function-name>>` is a `<function-token>`. It already includes a `(`. It would be nice if this was readily apparent, but I couldn't figure out something better than this.

andruud (commented):
It would be nice if the commit title explained what the fix _is_.

css-mixins-1/Overview.bs
bramus: Ah sorry, that wasn’t clear to me. Took me some clicks to find it in css-values-3.

Will undo this change.

bramus:
> It would be nice if the commit title explained what the fix _is_.

Revert the incorrect commit + adjusted the title of the PR.

Issue #10647, opened 2024-07-31
[css-shapes] Overload `path()` for CSS-y SVG path syntax instead of taking up `shape()`


LeaVerou:
Apparently [WebKit is implementing](https://github.com/w3c/csswg-drafts/issues/5841#issuecomment-2254583993) `shape()` so this is somewhat urgent.

SVG paths don't have the best ergonomics, and there have been long discussions about how to improve on the ergonomics as well, not just expose the existing model with CSS syntax. See #9889 for one discussion on this

Can we perhaps overload `path()` for the API that is mapped closely to SVG paths so that `shape()` is free to host something better without having to worry about disambiguation?

The current signature of `path()` is very weird anyway, so it would be good to make it sensible.

smfr:
Do you thoughts for what a future `shape()` might be used for?

I think we'll be able to map SVG paths to the new shape syntax losslessly, but will not be able to map in the other direction.

ydaniv:
I think it's good that `path()` and `<path>` are interchangeable. If anything, I'd prefer that in the future we would get a `<shape>` in SVG (setting aside how realistic that is for now).

jyasskin:
I don't know anything about the need to improve the shape language in the future, but I'm concerned about the sentence in the spec that "The [shape()](https://drafts.csswg.org/css-shapes-2/#funcdef-shape) function defines a path". `shape()` should define a _shape_, and `path()` should define a path.

These naming issues are sometimes inevitable, but it seems like this one isn't.

noamr:
> I don't know anything about the need to improve the shape language in the future, but I'm concerned about the sentence in the spec that "The [shape()](https://drafts.csswg.org/css-shapes-2/#funcdef-shape) function defines a path". `shape()` should define a _shape_, and `path()` should define a path.

Sure, we can change the wording.

noamr:
[#9889](https://github.com/w3c/csswg-drafts/issues/9889) is closed and we have no specific proposals being discussed around doing something different with `path()`. We've implemented `shape()` in chromium as well and almost ready to ship. @LeaVerou do you still want to pursue this? If not, can we close this issue?

LeaVerou:
Well, I still think `path()` only supporting a `<string>` argument is inconsistent with all other shape functions, so no, I'm not in favor of closing this with no WG resolution.

noamr:
[...]

PR #11586, opened 2025-01-28
[css-shapes-2] Refine use of the word "path" in the intro

noamr:
A "shape" is a set of commands to generate a path dynamically, rather than them being synonomous terms.


css-shapes-2/Overview.bs
cdoublev, line 1: I have not followed the related issue (#10647, I presume), sorry. But I think the "main" spec authors use [semantic line breaks](https://sembr.org/) and I guess it should be *"when it **is** used for rendering"*.

cdoublev (approved):

css-shapes-2/Overview.bs
noamr: Fixed, thanks!

PR #11594, opened 2025-01-29
[web-animations-1] Array-form keyframes argument can include `null` a…

cdoublev:
This is implied by the procedure to [process a keyframes argument](https://drafts.csswg.org/web-animations-1/#process-a-keyframes-argument) but is not accepted by the non-normative WebIDL type definition of the `keyframes` argument (appearing above the procedure).

While I do not see how this behavior can be useful to authors, they might appreciate having this clarified in the spec.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -6100,7 +6100,7 @@ The {{KeyframeEffect}} interface {#the-keyframeeffect-interface}
       (CompositeOperationOrAuto or sequence<CompositeOperationOrAuto>) composite = [];
     };
 
-    typedef (sequence<Keyframe> or PropertyIndexedKeyframes) KeyframeArgument;
+    typedef (sequence<Keyframe?> or PropertyIndexedKeyframes) KeyframeArgument;
   </xmp>
 
   The meaning and allowed values of each argument is as follows:


birtles:
Yes, looks like this is covered by WPT here:

https://github.com/web-platform-tests/wpt/blob/d46d9abdea3ea1eb0f1eebdc4430f21a13afaae3/web-animations/interfaces/KeyframeEffect/processing-a-keyframes-argument-001.html#L382-L410

web-animations-1/Overview.bs
```diff
@@ -6100,9 +6100,12 @@ The {{KeyframeEffect}} interface {#the-keyframeeffect-interface}
       (CompositeOperationOrAuto or sequence<CompositeOperationOrAuto>) composite = [];
     };
 
-    typedef (sequence<Keyframe> or PropertyIndexedKeyframes) KeyframeArgument;
+    typedef (sequence<Keyframe?> or PropertyIndexedKeyframes) KeyframeArgument;
   </xmp>
 
+  Null and undefined items in an array of keyframes
+  are converted to a keyframe with no animation property.
```
birtles, line 6107: I wonder if this note is necessary? (If it is, I think we mostly try to mark up `null` and `undefined` as `<code class=esvalue>null</code>` etc.)

cdoublev: I removed it.

`null` and `undefined` are always unintentional, I think. Conscientious authors would first check [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Keyframe_Formats), which does not define the specified behavior, and then the spec. They would read the non-normative introduction and find the (corrected) conceptual type. So at this point they know that these values are not errors. They might *suspect* the actual behavior because it seems to be the *least surprising* one. But then they would check the procedure, which is not easy to understand, but that should be fine...

birtles:
Thank you!

PR #11595, opened 2025-01-29
[css-custom-highlight-1] Fix IDL definition of `HighlightRegistry`

tidoust:
A recent update added a method to the `HighlightRegistry` interface through a `partial interface` definition. That's totally fine. However, the initial `interface` definition was also turned into a `partial interface` definition. That is incorrect, a `partial interface` definition always extends an interface that must be defined (without `partial`) somewhere.

This update drops `partial` from the core interface definition accordingly.


ffiori:
LGTM, thanks for noticing!

ffiori (approved):

PR #11597, opened 2025-01-29
[css-fonts-4] Fix seeming typo in Font Feature and Variation Resolution

foolip:
https://drafts.csswg.org/css-fonts-4/#font-feature-variation-resolution looks strange currently, rendering as "either via the use of ‘font-variant’, ‘font-feature-settings’, ‘font’‘font-variation-settings’ in a style rule or within an @font-face rule".

I'm not totally sure of the intention, but I've made the most plausible fix to me.

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -6815,8 +6815,8 @@ Font Feature and Variation Resolution</h3>
 As described in the previous section,
 font features and variations can be enabled in a variety of ways,
 either via the use of 'font-variant!!property',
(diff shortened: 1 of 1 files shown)


svgeesus (approved):

svgeesus:
This odd markup was in [the original commit](https://github.com/w3c/csswg-drafts/blame/ab8c9738327c1ed9665995e1604ce35e35e1d05a/css-fonts-4/Overview.bs) so it is unclear what was intended.

Your change makes sense.

Issue #11598, opened 2025-01-29
CSS Snapshot: incorrect reference to section on text decoration


stiiin:
<p>Since CSS Snapshot 2020, the following two modules are mentioned in the fairly stable modules section (emphasis mine):</p><blockquote><dl><dt><a href="https://www.w3.org/TR/css-text-3/">CSS Text Module Level 3</a> <a href="https://www.w3.org/TR/css-2020/#biblio-css-text-3">[CSS-TEXT-3]</a></dt><dd>Extends and supersedes <mark>CSS2§16 excepting §16.2</mark>, defining properties for text manipulation and specifying their processing model. It covers line breaking, justification and alignment, white space handling, and text transformation.</dd><dt><a href="https://www.w3.org/TR/css-text-decor-3/">CSS Text Decoration Level 3</a> <a href="https://www.w3.org/TR/css-2020/#biblio-css-text-decor-3">[CSS-TEXT-DECOR-3]</a></dt><dd>Extends and supersedes <mark>CSS2§16.3</mark>, providing more control over text decoration lines and adding the ability to specify text emphasis marks and text shadows.</dd></dl></blockquote><p>CSS2§16.2 is titled "Alignment: the 'text-align' property", and CSS Text Module Level 3 in fact does redefine <code>text-align</code> (<a href="https://www.w3.org/TR/css-text-3/#text-align-property">CSS-TEXT-3§6.1: Text Alignment: the text-align shorthand</a>).</p><p>I think "<a href="https://github.com/w3c/csswg-drafts/blob/e3587066d404d558f6c975ee1c8469a50c686eb1/css-2024/Overview.bs#L454">CSS2§16 excepting §16.2</a>" should therefore instead read "CSS2§16 excepting §16.3".</p>

PR #11604, opened 2025-01-29
[css-2024] Fix section number for text-decoration (#11598)

stiiin:
CSS Text Module Level 3 extends and supersedes everything in chapter 16 of CSS2, except text-decoration. text-decoration is defined in CSS2 section 16.3, not 16.2.


stiiin:
Please let me know whether the second commit, retroactively fixing the typo in older snapshots, should be done differently (multiple commits?) or not at all.

svgeesus (approved):

PR #11606, opened 2025-01-30
[css-borders-4] Specify `corner-shape` based on superellipse

### LINKED ISSUE #10993: [css-shapes-2][css-borders-4] corner-shape support for superellipses
I worked a bit on trying to find an option for `corner-shape` on the new API that supported squircles and had inner/outer symmetry.

After a bit of work I'm proposing we add a value `se(x)` (name TBD) that exposes a [superellipse](https://en.wikipedia.org/wiki/Superellipse):

$$ \left|\frac{x}{a}\right|^n + \left|\frac{y}{b}\right|^n = 1 $$

Where $a$ and $b$ are the width and height of the shape. And the default parameter $n$ works in the interval $[0,\infty)$ where:
- $n=0$ is a inwards square
- $n=0.5$ is a inwards circle
- $n=1$ is bevel
- $n=2$ is a circular border (like rounded corner)
- $n=4$ is a squircle
- $n\rightarrow\infty$ is a rectangle

My proposal would be to use $x$ as a parameter where $n=2^x$. This leads to a function that is symmetrical on $0$ where positive values are outside the bevel and negative values are inside, such as:
- $n\rightarrow-\infty$ is a inwards square
- $n=-1$ is a inwards circle
- $n=0$ is bevel
- $n=1$ is a circular border (like rounded corner)
- $n=2$ is a squircle
- $n\rightarrow\infty$ is a rectangle

I've wrote a small demo with this function [here](https://fserb.com/css/squircles).

The superellipse is very easily parameterizable (both angular and $[0,1]$), so I'm guessing it shouldn't be a problem to implement it in an efficient way. There's a small issue where if we are "inside the bevel" the corners can sometimes overlap. For those cases, we will have to do some math and limit the size of the corners to prevent it from happening (the demo currently doesn't handle those cases).

There are other available squircle formulations (like the Fernandez-Guasti squircles) where the parameter is more intuitive ($0$ means circle, $0.5$ is squircle, and $1$ is square). Unfortunately, those formulations don't extend to inwards the bevel, so we would lose some expressiveness. 

We could, together with this, also provide certain easy-to-use values (like `squircle` meaning `se(2)`, or `bevel` meaning `se(0)`, etc...). And they could also contain parameters (i.e. `squircle(y)` mapping $[0,1]$ to $[1,\infty) $).

Highlights:
* squircles!
* the symmetry on the bevel and around $0$ is very nice.
* It's very expressive of a lot of shapes that people care about.

Lowlights:
* *"superellipsis"* is an awful name and very non-descriptive of the intent of the shape. 
* All parameter values (for example, $2$ meaning squircle) are not intuitive or really meaningful.

noamr:
This specifies the `corner-shape` group, including:

- general description and interaction with border-radius
- all the individual corners, side shorthands, and overall shorthand
- multiple keywords, and how they translate to a `superellipse()`
- The superellipse formula, and how it is rendered
- How the exponent of the superellipse interpolates

Open issues (will open separately):
- Add a few examples
- Resolve on "straight" vs "none" for the convex angle.
- Resolve on the exact interpolation formula
- Define restrictions for border rendering

This overwrites the existing `corner-shape` definition.

Closes #10993
Based on resolution https://github.com/w3c/csswg-drafts/issues/10993#issuecomment-2622578937

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -386,7 +386,7 @@ Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: che
 		<dd>Border radii define a convex 90deg angle at the corner.
 			This would have the same visual effect as a 'border-radius' of 0. This is different from having a 'border-radius' of 0 when animating.
 			Equivalent to <css>superellipse(infinity)</css>.
-		<dt><dfn>''squircle''</dfn>
+		<dt><dfn>squircle</dfn>
 		<dd>Border radii define a convex curve between an ellipse and an convex angle, equivalent to <css>superellipse(4)</css>.
 	</dl>
 
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -405,9 +405,19 @@ Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: che
 		|x/a|^k + |y/b|^k = 1
 	</pre>
 
+<h4 id=corner-shape-shorthand>
+Corner Shaping: the 'corner-shape' and 'corner-*-shape' properties</h4>
 
-<h4 id=corner-shape-properties>
-Corner Shaping: the 'corner-*-*-shape' properties</h4>
+	<pre class="propdef">
+		Name: corner-shape
+		Value: <<corner-shape-value>>{1,2} / [ <<corner-shape-value>>{1,2} ]?
+		Initial: round
+		Applies to: all elements where 'border-radius' can apply
+		Inherited: no
+		Animation type: see individual properties
+	</pre>
+
+	Applies the shape to all corners, following the same rules as ''border-radius''.
 
 	<pre class=propdef>
 	Name: corner-top-left-shape, corner-top-right-shape, corner-bottom-right-shape, corner-bottom-left-shape, corner-start-start-shape, corner-start-end-shape, corner-end-start-shape, corner-end-end-shape
@@ -434,13 +444,6 @@ Corner Shaping: the 'corner-*-*-shape' properties</h4>
 	with the first start/end giving the block axis side,
 	and the second the inline-axis side
 	(i.e. patterned as 'corner-<var>block</var>-<var>inline</var>-shape').
-
-<h4 id="corner-shape-side-shorthands">
-Shaping The Corners Of One Side:
-The 'corner-top-shape', 'corner-right-shape',
-'corner-bottom-shape', 'corner-left-shape',
-'corner-block-start-shape', 'corner-block-end-shape',
-'corner-inline-start-shape', 'corner-inline-end-shape' shorthands</h4>
 
 	<pre class=propdef>
 		Name: corner-top-shape, corner-right-shape, corner-bottom-shape, corner-left-shape,
@@ -469,19 +472,6 @@ The 'corner-top-shape', 'corner-right-shape',
 	and start-end, end-end for 'corner-inline-end-shape'.
 	If the second value is omitted it is copied from the first.
 
-<h4 id=corner-shape-shorthand>
-Shaping all the corners at once</h4>
-
-	<pre class="propdef">
-		Name: corner-shape
-		Value: <<corner-shape-value>>{1,2} / [ <<corner-shape-value>>{1,2} ]?
-		Initial: round
-		Applies to: all elements where 'border-radius' can apply
-		Inherited: no
-		Animation type: see individual properties
-	</pre>
-
-	Applies the shape to all corners, following the same rules as ''border-radius''.
 
 <h4 id=corner-shape-interpolation>
 Interpolating corner shapes</h4>
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -357,8 +357,8 @@ Like 'border-radius', 'corner-shape' clips elements according to the [=overflow=
 Since stroking a superellipse accurately may be computationally intensive, user agents may approximate the path using bezier curves,
 as well as account for sharp edges and overlaps.
 
-Note that `border-radius` already has a <a href="https://drafts.csswg.org/css-backgrounds/#corner-overlap">corner overlap restriction</a>,
-and it's unclear whether 'corner-shape' would require additional restrictions.
+Issue: 'border-radius' already handles *adjacent* corners overlapping by shrinking the radiuses proportionally.
+A negative ''superellipse()'' parameter allows for *opposite* corners to sometimes overlap, and needs additional restrictions defined.
 
 Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: check if we need additional rendering restrictions.
 
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -392,7 +392,7 @@ Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: che
 
 	Issue <a href="https://github.com/w3c/csswg-drafts/issues/11607">#11607</a>: resolve on ''straight'' vs <css>none</css>.
 
-	The <dfn>superellipse()</dfn> function describes the curvature of the corner.
+	The <dfn>superellipse( <<number>> | infinity )</dfn> function describes the curvature of the corner.
 	It accept a <<number>> <dfn>superellipse exponent</dfn>, which defines the curvature of the corner, or the exponent of the formula.
 	The [=superellipse exponent=] accepts values between 0 (a straight concave angle) and <css>infinity</css> (a straight convex angle),
 	with the values in between representing the curvatures in between.
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -380,7 +380,7 @@ Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: che
 		<dd>Border radii define a concave elliptical curve at the corner. Equivalent to <css>superellipse(0.5)</css>.
 		<dt><dfn>bevel</dfn>
 		<dd>Border radii define a diagonal slice at the corner. Equivalent to <css>superellipse(1)</css>.
-		<dt><dfn>''notch''</dfn>
+		<dt><dfn>notch</dfn>
 		<dd>Border radii define a concave 90deg angle at the corner. Equivalent to <css>superellipse(0)</css>.
 		<dt><dfn>straight</dfn>
 		<dd>Border radii define a convex 90deg angle at the corner.
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -393,7 +393,7 @@ Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: che
 	Issue <a href="https://github.com/w3c/csswg-drafts/issues/11607">#11607</a>: resolve on ''straight'' vs <css>none</css>.
 
 	The <dfn>superellipse( <<number>> | infinity )</dfn> function describes the curvature of the corner.
-	It accept a <<number>> <dfn>superellipse exponent</dfn>, which defines the curvature of the corner, or the exponent of the formula.
+	It accepts a <dfn>superellipse exponent</dfn>, which defines the curvature of the corner, or the exponent of the formula.
 	The [=superellipse exponent=] accepts values between 0 (a straight concave angle) and <css>infinity</css> (a straight convex angle),
 	with the values in between representing the curvatures in between.
 
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -370,7 +370,7 @@ Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: che
 	</pre>
 
 	<dl dfn-type="value" dfn-for="corner-shape-value">
-		<dt><dfn>''round''</dfn>
+		<dt><dfn>round</dfn>
 		<dd>
 			Border radii define a convex elliptical curve at the corner. Equivalent to <css>superellipse(2)</css>.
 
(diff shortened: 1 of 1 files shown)


noamr:
(Sorry for the multiple reviewers, but it's probably a good idea to have a few eyes on this)

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,171 @@ The 'border-radius' shorthand</h4>
...
+
+<h4 id=corner-shape-value>
+'corner-shape' values</h4>
+
+	<pre class=prod>
+		<dfn><<corner-shape-value>></dfn> = ''round'' | ''scoop'' | ''bevel'' | ''notch'' | ''straight'' | ''squircle'' | superellipse(<<number [0,&infin;]>>)
+	</pre>
+
+	<dl dfn-type="value" dfn-for="corner-shape-value">
+		<dt><dfn>''round''</dfn>
```
fserb, line 370: does it say somewhere that this is the default? 

Also, is it worth mentioning that this is the "current behavior", i.e., rounded corners?

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,171 @@ The 'border-radius' shorthand</h4>
...
+Issue: describe the rendering requirements in more detail.
+
+<h4 id=corner-shape-value>
+'corner-shape' values</h4>
+
+	<pre class=prod>
+		<dfn><<corner-shape-value>></dfn> = ''round'' | ''scoop'' | ''bevel'' | ''notch'' | ''straight'' | ''squircle'' | superellipse(<<number [0,&infin;]>>)
+	</pre>
+
+	<dl dfn-type="value" dfn-for="corner-shape-value">
```
fserb, line 369: I think the proposal was to make the `superellipse()` function symmetrical on 0, right? If so, it would be:

```
notch -> -inf
scoop -> -1
bevel -> 0
round -> 1
squircle -> 2
straight -> inf
```

But we can discuss that on the separate issue after then.

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,171 @@ The 'border-radius' shorthand</h4>
...
+		<dt><dfn>''round''</dfn>
+		<dd>Border radii define a convex elliptical curve at the corner. Equivalent to <css>superellipse(2)</css>.
+		<dt><dfn>''scoop''</dfn>
+		<dd>Border radii define a concave elliptical curve at the corner. Equivalent to <css>superellipse(0.5)</css>.
+		<dt><dfn>''bevel''</dfn>
+		<dd>Border radii define a diagonal slice at the corner. Equivalent to <css>superellipse(1)</css>.
+		<dt><dfn>''notch''</dfn>
+		<dd>Border radii define a concave angle at the corner. Equivalent to <css>superellipse(0)</css>.
+		<dt><dfn>''straight''</dfn>
+		<dd>Border radii define a convex angle at the corner.
```
fserb, line 379: Maybe we can find a better definition than "convex angle"/"concave angle"? Just to clarify that is 90deg? "convex 90deg angle"?

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,171 @@ The 'border-radius' shorthand</h4>
...
-	<dl dfn-type="value" dfn-for="corner-shape">
-		<dt><dfn>''round''</dfn>
-		<dd>Border radii define a convex elliptical curve at the corner.
-		<dt><dfn>''angle''</dfn>
-		<dd>Border radii define a diagonal slice at the corner.
-	</dl>
+<h4 id=corner-shape-interpolation>
+Interpolating corner shapes</h4>
+
+Since a <<corner-shape-value>> can always be expressed a ''superellipse()'' with an [=superellipse exponent=] variable, interpolating between two
```
fserb, line 482: "expressed by a"

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,171 @@ The 'border-radius' shorthand</h4>
...
+	1. If |exponent| is &infin;, return 1.
+	1. Return <code>1/(2^(1/|exponent|))</code>.
+
+To convert a <<number [0,1]>> |interpolationValue| back to a [=superellipse exponent=]:
+	1. If |interpolationValue| is 0, return 0.
+	1. If |interpolationValue| is 1, return &infin;.
+	1. Return <code>ln(0.5)/ln(|interpolationValue|)</code>.
+</div>
+
+Issue: resolve on this or another interpolation formula.
```
fserb, line 499: I'm assuming we need to investigate if this looks good right?

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,171 @@ The 'border-radius' shorthand</h4>
...
+
+<pre>
```
fserb, line 501: do you think it's worth mentioning the radius restriction just as a placeholder?

css-borders-4/Overview.bs
noamr: It does, see the different properties

css-borders-4/Overview.bs
noamr: Created #11609

css-borders-4/Overview.bs
noamr: Yes, opened #11608. It would also be impacted by a change in what `k` means (#11609).

css-borders-4/Overview.bs
noamr: The radius restriction holds here regardless of `corner-shape`. How would you word this?

css-borders-4/Overview.bs
noamr: Done

css-borders-4/Overview.bs
fserb: Maybe link to [the current restriction](https://drafts.csswg.org/css-backgrounds/#corner-overlap), and have a todo to see if this is enough or if we need to update the algorithm for the new curves? I see a lot of "if the radii overlap", on the previous one, not sure if it would apply as is.

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+Rendering 'corner-shape'</h4>
+'corner-shape' works alongside 'border-radius', and does not have any visual effect with a 'border-radius' of 0.
+It acts as an alteration on top of the default round 'border-radius', and thus can be used as progressive enhancement.
+
+Like 'border-radius', 'corner-shape' clips elements according to the [=overflow=] rules, and applies to the rendering of the border.
+Since stroking a superellipse accurately may be computationally intensive, user agents may approximate the path using bezier curves,
+as well as account for sharp edges and overlaps.
+
+Note that `border-radius` already has a <a href="https://drafts.csswg.org/css-backgrounds/#corner-overlap">corner overlap restriction</a>,
+and it's unclear whether 'corner-shape' would require additional restrictions.
```
tabatkins, line 361:
```suggestion
Issue: 'border-radius' already handles *adjacent* corners overlapping by shrinking the radiuses proportionally.
A negative ''superellipse()'' parameter allows for *opposite* corners to sometimes overlap, and needs additional restrictions defined.
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+
+<h4 id=corner-shape-value>
+'corner-shape' values</h4>
+
+	<pre class=prod>
+		<dfn><<corner-shape-value>></dfn> = ''round'' | ''scoop'' | ''bevel'' | ''notch'' | ''straight'' | ''squircle'' | superellipse(<<number [0,&infin;]>>)
+	</pre>
+
+	<dl dfn-type="value" dfn-for="corner-shape-value">
+		<dt><dfn>''round''</dfn>
```
tabatkins, line 373:
```suggestion
		<dt><dfn>round</dfn>
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+		<dd>
+			Border radii define a convex elliptical curve at the corner. Equivalent to <css>superellipse(2)</css>.
+
+			Note: this is the initial value of 'corner-shape' properties, as elements with 'border-radius' would be rounded.
+
+		<dt><dfn>''scoop''</dfn>
```
tabatkins, line 379:
```suggestion
		<dt><dfn>scoop</dfn>
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: check if we need additional rendering restrictions.
+
+<h4 id=corner-shape-value>
+'corner-shape' values</h4>
+
+	<pre class=prod>
+		<dfn><<corner-shape-value>></dfn> = ''round'' | ''scoop'' | ''bevel'' | ''notch'' | ''straight'' | ''squircle'' | superellipse(<<number [0,&infin;]>>)
+	</pre>
+
+	<dl dfn-type="value" dfn-for="corner-shape-value">
```
tabatkins, line 372:
```suggestion
	<dl dfn-type="value" dfn-for="<corner-shape-value>">
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+		<dt><dfn>''round''</dfn>
+		<dd>
+			Border radii define a convex elliptical curve at the corner. Equivalent to <css>superellipse(2)</css>.
+
+			Note: this is the initial value of 'corner-shape' properties, as elements with 'border-radius' would be rounded.
+
+		<dt><dfn>''scoop''</dfn>
+		<dd>Border radii define a concave elliptical curve at the corner. Equivalent to <css>superellipse(0.5)</css>.
+		<dt><dfn>''bevel''</dfn>
```
tabatkins, line 381:
```suggestion
		<dt><dfn>bevel</dfn>
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+		<dd>Border radii define a diagonal slice at the corner. Equivalent to <css>superellipse(1)</css>.
+		<dt><dfn>''notch''</dfn>
```
tabatkins, line 383:
```suggestion
		<dt><dfn>notch</dfn>
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+		<dd>Border radii define a concave 90deg angle at the corner. Equivalent to <css>superellipse(0)</css>.
+		<dt><dfn>''straight''</dfn>
```
tabatkins, line 385:
```suggestion
		<dt><dfn>straight</dfn>
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+		<dd>Border radii define a convex 90deg angle at the corner.
+			This would have the same visual effect as a 'border-radius' of 0. This is different from having a 'border-radius' of 0 when animating.
+			Equivalent to <css>superellipse(infinity)</css>.
+		<dt><dfn>''squircle''</dfn>
```
tabatkins, line 389:
```suggestion
		<dt><dfn>squircle</dfn>
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+		<dd>Border radii define a convex curve between an ellipse and an convex angle, equivalent to <css>superellipse(4)</css>.
+	</dl>
+
+	Issue <a href="https://github.com/w3c/csswg-drafts/issues/11607">#11607</a>: resolve on ''straight'' vs <css>none</css>.
+
+	The <dfn>superellipse()</dfn> function describes the curvature of the corner.
+	It accept a <<number>> <dfn>superellipse exponent</dfn>, which defines the curvature of the corner, or the exponent of the formula.
```
tabatkins, line 396:
```suggestion
	It accepts a <dfn>superellipse exponent</dfn>, which defines the curvature of the corner, or the exponent of the formula.
```

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+		<dd>Border radii define a convex 90deg angle at the corner.
+			This would have the same visual effect as a 'border-radius' of 0. This is different from having a 'border-radius' of 0 when animating.
+			Equivalent to <css>superellipse(infinity)</css>.
+		<dt><dfn>''squircle''</dfn>
+		<dd>Border radii define a convex curve between an ellipse and an convex angle, equivalent to <css>superellipse(4)</css>.
+	</dl>
+
+	Issue <a href="https://github.com/w3c/csswg-drafts/issues/11607">#11607</a>: resolve on ''straight'' vs <css>none</css>.
+
+	The <dfn>superellipse()</dfn> function describes the curvature of the corner.
```
tabatkins, line 395:
```suggestion
	The <dfn>superellipse( <<number>> | infinity )</dfn> function describes the curvature of the corner.
```

tabatkins: (if we rescale the argument, this needs to gain `-infinity` as a keyword)

css-borders-4/Overview.bs
```diff
@@ -336,29 +336,178 @@ The 'border-radius' shorthand</h4>
...
+	where <code>a</code> is the horizontal ''border-radius''
+	<code>b</code> is the vertical ''border-radius'', and <code>k</code> is the [=superellipse exponent=]:
+
+	<pre>
+		|x/a|^k + |y/b|^k = 1
+	</pre>
+
+
+<h4 id=corner-shape-properties>
+Corner Shaping: the 'corner-*-*-shape' properties</h4>
```
tabatkins, line 410: While we *usually* do longhands first, followed by the shorthand, we tend to use the opposite order when the shorthand is going to be used the most often and the longhands are just it exploded to every possible side/corner. Mind swapping the order of these? They can all be grouped under one heading, maybe "Corner Shaping: the 'corner-shape' and 'corner-*-shape' properties".

tabatkins (changes requested):

css-borders-4/Overview.bs
noamr: Sure!

css-borders-4/Overview.bs
noamr: Done

css-borders-4/Overview.bs
noamr: Done

tabatkins (approved):

PR #11617, opened 2025-01-30
[css-env-1] Spec safe-area-max-inset-* envvars

### LINKED ISSUE #11019: [css-env-1] Maximum safe area inset values to allow sliding bottom bar
On some platforms, the safe-area inset can change dynamically while scrolling. For example, when the browser viewport extends to the bottom of an Android or iOS mobile screen, there is usually a system bar drawn on top of the content which also consumes input events which would prevent users from being able to read / interact with that content.

Sites creating bottom bars use the safe area to avoid overlapping this. However, if this requires a relayout it can result in a jittery experience, (e.g. see the linked demo on the [chromium intent](https://groups.google.com/a/chromium.org/g/blink-dev/c/hXkimpjYaMg/m/qiPFAW92AwAJ))

I created a demo site to show an example of two approaches times two positioning methods developers can use to create bottom bars that avoid this area:
https://flackr.github.io/web-demos/css-env/safe-area-inset/index.html

Using the `bottom` property to avoid the area in theory allows the browser to apply this as a translation similar to static sticky / fixed position offsets:
```css
.footer {
  position: fixed;
  background: gray;
  bottom: env(safe-area-inset-bottom, 0px);
}
```

However, this means that when the safe area is greater than 0, you will be able to see behind the footer's background. Using padding would grow the footer, however means that updating the footer is no longer an easily compositable effect. I think it would be useful to give developers the maximum safe area inset so that they could size the footer to the maximum size so that it can simply slide up when needed, e.g.:

```css
.footer {
  position: fixed;
  background: gray;
  padding-bottom: env(safe-area-max-inset-bottom, 0px);
  bottom: calc(env(safe-area-inset-bottom, 0px) - env(safe-area-max-inset-bottom, 0px));
}
```

This would ensure that there is extra space reserved in the footer, positioned offscreen when not needed. The linked [demo](https://flackr.github.io/web-demos/css-env/safe-area-inset/index.html) shows this approach using an assumed oversized maximum inset.

TLDR; can/should we add the maximum dynamic inset so developers can easily create a bottom bar just large enough to slide up when needed?

bramus:
This PR specifies the `safe-area-max-inset-*` envvars.

Fixes #11019.

### DIFF
--- a/css-env-1/Overview.bs
+++ b/css-env-1/Overview.bs
@@ -126,7 +126,7 @@ safe area rectangle.
 
 <index type=value for=env()></index>
 
-Maximum Safe area inset variables {#safe-area-max-insets}
+Safe area maximum inset variables {#safe-area-max-insets}
 ------------------------------------------------------------------
 
 <table dfn-type=value dfn-for="env()">
@@ -152,11 +152,11 @@ Maximum Safe area inset variables {#safe-area-max-insets}
 		<td>0 (scalar)
 </table>
 
-The maximum safe area insets are four [=environment variables=]
+The safe area maximum insets are four [=environment variables=]
 that are tied to the <a href="#safe-area-insets">safe area inset variables</a>.
 Unlike the safe area inset variables which are dynamic values,
-the maximum safe area insets are static values
 that represent the maximum value of their dynamic counterpart.
+the safe area maximum insets are static values
 
 <index type=value for=env()></index>
 
--- a/css-env-1/Overview.bs
+++ b/css-env-1/Overview.bs
@@ -155,8 +155,10 @@ Safe area maximum inset variables {#safe-area-max-insets}
 The safe area maximum insets are four [=environment variables=]
 that are tied to the <a href="#safe-area-insets">safe area inset variables</a>.
 Unlike the safe area inset variables which are dynamic values,
-that represent the maximum value of their dynamic counterpart.
 the safe area maximum insets are static values
+that represent the maximum value of their dynamic counterpart
+when dynamic UA interfaces are retracted,
+making the [=layout viewport=] size the [=large viewport size=].
 
 <index type=value for=env()></index>
 
--- a/css-env-1/Overview.bs
+++ b/css-env-1/Overview.bs
@@ -126,6 +126,40 @@ safe area rectangle.
 
 <index type=value for=env()></index>
 
+Maximum Safe area inset variables {#safe-area-max-insets}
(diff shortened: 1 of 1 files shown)


css-env-1/Overview.bs
```diff
@@ -126,6 +126,40 @@ safe area rectangle.
...
+		<td><dfn>safe-area-max-inset-left</dfn>
+		<td><<length>>
+		<td>0 (scalar)
+</table>
+
+The maximum safe area insets are four [=environment variables=]
+that are tied to the <a href="#safe-area-insets">safe area inset variables</a>.
+Unlike the safe area inset variables which are dynamic values,
+the maximum safe area insets are static values
+that represent the maximum value of their dynamic counterpart.
```
flackr, line 159: I think you should mention and draw analogy to the small and large viewport: https://www.w3.org/TR/css-values-4/#viewport-variants . In particular, these will be the safe-area-insets when using the large viewport.

This helps differentiate what we mean by "maximum" from things like the values could increase by rotating the device or entering fullscreen.

bramus: Would adding _“for the current device orientation”_ at the end be sufficient to cover this?

flackr: Is that sufficient to exclude fullscreen? I didn't think entering fullscreen necessarily counted as a different orientation.

css-env-1/Overview.bs
```diff
@@ -126,6 +126,40 @@ safe area rectangle.
 
 <index type=value for=env()></index>
 
+Maximum Safe area inset variables {#safe-area-max-insets}
```
vmpstr, line 129: nit: Safe Area Max Inset Variables

css-env-1/Overview.bs
flackr: How about:

Unlike the safe area inset variables which are dynamic values,
the maximum safe area insets are static values
that represent the maximum value of their dynamic counterpart
when dynamic UA interfaces that are retracted making the [=viewport size=] the [=large viewport size=].

css-env-1/Overview.bs
bramus: Addressed in e712091

css-env-1/Overview.bs
bramus: Added in 45843b2. Did have to fix some bikeshed-linking, as there is no term `[=viewport size=]` defined in any spec to link to.

flackr (approved):

Issue #11624, opened 2025-01-31
[web-animations-2] Web IDL syntax error


autokagami:
🤖 This is an automatic issue report for Web IDL syntax error. 🤖

[webidl2.js](https://github.com/w3c/webidl2.js) says:

```
Syntax error at line 6 in web-animations-2,15, since `interface AnimationTrigger`:
 CSSKeywordValue or DOMString) rangeStart = "normal";
                                          ^
```

> Error: Unterminated attribute, expected `;`

Please file an issue at https://github.com/saschanaz/webidl-updater/issues/new if you think this is invalid or should be enhanced.

tidoust:
I suspect the intent of the change was to set default values for the attributes. Default values are not allowed in `interface` attributes in Web IDL, only for `dictionary` members.

tidoust:
Also note that, even without the default value, the definition of the attribute remains invalid for the same reason as that noted in
https://github.com/w3c/csswg-drafts/issues/9355#issuecomment-1719439559: "dictionaries aren't allowed as attribute types"

ydaniv:
Thanks! I'll push a fix

ydaniv:
I think this should do as a fix: https://github.com/w3c/csswg-drafts/pull/11630

tidoust:
Oh, failed to notice, but another IDL hiccup will appear once you fix the other ones: a nullable union cannot include a dictionary type, see https://webidl.spec.whatwg.org/#idl-nullable-type

That will happen because dictionary members in `AnimationTriggerOptions` are defined with a `?`.

ydaniv:
Fixed.

PR #11630, opened 2025-01-31
Removed bad syntax adding defaults to interface attributes + any instead of dictionaries as attribute types

ydaniv:
Fix issue #11624


ydaniv:
@flackr @tabatkins 
I think this should fix the issue above.

Do we also need to take care of #9360 ?

ydaniv:
@tidoust should this suffice?

PR #11637, opened 2025-02-03
[css-gaps-1] Fix typo

cdoublev:
There is a typo in the initial value of `column-rule-color`/`row-rule-color`.


kbabbitt (approved):
Thanks for the fix!

PR #11651, opened 2025-02-04
[css-borders-4] Remove the second value from the corner-*-shape variant

### LINKED ISSUE #11650: [css-borders-4] Remove `/` in `corner-*-shape` grammars
The grammar of [`corner-*-shape`](https://drafts.csswg.org/css-borders-4/#propdef-corner-shape) shorthands requires using `/` to separate horizontal/vertical radii.

  > **Name:** `corner-shape`
  > **Value:** `<corner-shape-value>{1,2} [ / <corner-shape-value>{1,2} ]?`
  > 
  > **Name:** `corner-top-shape`, ...
  > **Value:** `<corner-shape-value> [ / <corner-shape-value> ]`
  >
  > The `corner-*-shape` shorthands set the two `corner-*-*-shape` longhand properties of the related side. If values are given before and after the slash, then the values before the slash set the horizontal radius and the values after the slash set the vertical radius.

(nit: the second `<corner-shape-value>` in `corner-<side>-shape` should probably be optional).

But there is no horizontal/vertical definition of a corner shape like with `border-radius`. Is it an oversight? If so, I guess they should follow the same rules than `margin`, `padding`, right?

(edit, clarification) The canonical order of `corner-shape` longhands would be `corner-top-left-shape`, `corner-top-right-shape`, `corner-bottom-right-shape`, `corner-bottom-left-shape`.

noamr:
`corner-shape` has only one value, there is no point in providing two values for each side, it was a copy-paste mistake from `border-radius`.

Closes #11650

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -445,34 +445,6 @@ Corner Shaping: the 'corner-shape' and 'corner-*-shape' properties</h4>
 	and the second the inline-axis side
 	(i.e. patterned as 'corner-<var>block</var>-<var>inline</var>-shape').
 
-	<pre class=propdef>
-		Name: corner-top-shape, corner-right-shape, corner-bottom-shape, corner-left-shape,
-			corner-block-start-shape, corner-block-end-shape, corner-inline-start-shape, corner-inline-end-shape
-		Value: <<corner-shape-value>> [ / <<corner-shape-value>> ]
-		Initial: round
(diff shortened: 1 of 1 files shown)


cdoublev:
I do not follow, sorry. :worried: 

`corner-shape` has one value for each corner. That is 4. And `corner-<side>-shape` makes sense if the author only wants to define the 2 values of `<side>`.

nt1m (approved):

PR #11656, opened 2025-02-05
[css-borders-4] Initial (somewhat scaffoldy) spec for 'border-shape'

### LINKED ISSUE #6997: [css-shapes-2][css-borders-4] Add a way to change an element's shape
@LeaVerou mentioned an `element-shape` property in https://github.com/w3c/csswg-drafts/issues/6980#issuecomment-1020301916 as a way to influence the shape of an element. The idea behind that property is to change the shape of the element so that border, shadows, etc. follow that shape.

So far we have the `shape-outside` property which influences the float area. I suggest to extend this property to be able to specify that it not only affects the float area but also all the other boxes created for the element.

Sebastian

noamr:
Closes #6997

This defines the 'border-shape' property with a narrow set of details that we've resolved on, leaving room for discussing some details as we go along.

Specifically, this adds the two variants of 'border-shape' (single/double <basic-shape>).

Issues to be opened separately:
* interaction with border-{width|color|style}
* Clipping replaced elements
* Lots of examples

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -19,6 +19,7 @@ Warning: Not Ready
 <pre class="link-defaults">
 spec:css-text-4; type:value; text:collapse
 spec:css-shapes-2; type:function; text:path()
+spec:css-shapes-2; type:property; text:shape-inside
 </pre>
 
 <link rel="stylesheet" href="style.css" />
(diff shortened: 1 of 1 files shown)


css-borders-4/Overview.bs
```diff
@@ -1087,6 +1088,58 @@ Layering, Layout, and Other Details</h4>
...
+
+A 'box-shadow' follows both the inner and outer border paths.
+
+'border-shape' does not affect geometry or layout,
+which is still computed using the existing 'border-width' properties.
+
+'border-shape' does not affect the flow of content inside the box.
+Note: An author can use 'border-shape' in tandem with 'shape-inside' to create effects that decorate the box and control its text flow at the same time.
+
+The inner 'border-shape' clips the content of the element, in the same manner as 'border-radius'.
```
smfr, line 1133: Perhaps worth spelling out that "in the same manner" means that only if `overflow:hidden` is set

css-borders-4/Overview.bs
```diff
@@ -1087,6 +1088,58 @@ Layering, Layout, and Other Details</h4>
...
+
+Issue: how should this affect clipping replaced elements?
+
+The fill and stroke color, as well as the stroke width, are taken from the respective 'border-color' and 'border-width' properties.
+
+Issue: define this in detail. Perhaps it should be overridable somehow?
+
+Issue: what do we do about 'border-style'? It can't exactly work for every arbitrary shape.
```
smfr, line 1141: We could support dotted and dashed on the single path variant, but not the two-path variant.

smfr (approved):

css-borders-4/Overview.bs
noamr: > We could support dotted and dashed on the single path variant, but not the two-path variant.

Are we going to stroke the inner and outer paths? Anyway we will have a separate issue for this

css-borders-4/Overview.bs
noamr: Reworded

PR #11664, opened 2025-02-06
[web-animations-2] Use `dfn` type for animation trigger values

tidoust:
The values for "animation trigger state" and "animation trigger type" were defined with a `value` definition type. That type is reserved for CSS values: https://speced.github.io/bikeshed/#dfn-types

This makes tools that crawl specs to extract machine-readable information wrongly assume that the spec introduces new CSS values.

There is no specific definition type available for generic values. This update uses a regular `dfn` and updates references to these values accordingly.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -2379,19 +2379,19 @@ Issue: Should there be any effect of triggers on scroll-driven animations?
 The behavior an [=animation trigger=] applies to its associated [=animation=]’s
 playback is defined by its <dfn export>animation trigger type</dfn>, and
 depending on its [=animation trigger state|state=] can apply one of two
-distinct behavior types: ''primary'' or ''inverse''.
+distinct behavior types: [=animation trigger state/primary=] or [=animation trigger state/inverse=].
 Given an internal |did trigger| flag, the values of [=animation trigger type=]
 can be one of the following:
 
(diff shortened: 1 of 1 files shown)


ydaniv:
Thanks for opening this PR! Regarding `state`, I think correct the dfn to a simple on and not `value` was the thing I was missing, good catch!
Regarding `type`, it's actually a value: https://drafts.csswg.org/css-animations-2/#animation-trigger-type
So these should probably remain as values?

tidoust:
> Regarding `type`, it's actually a value: https://drafts.csswg.org/css-animations-2/#animation-trigger-type
> So these should probably remain as values?

Ah, sorry, I had missed that, indeed! There should be only one normative definition of anything, so if css-animations-2 defines the values, web-animations-2 should reference them and not attempt to re-define them. I pushed a new commit that replaces the `<dfn>` tags with references to the values in css-animations-2.

ydaniv:
@tidoust thanks! LGTM.

> so if css-animations-2 defines the values, web-animations-2 should reference them and not attempt to re-define them.

In css-animations-2 we have a more user-facing definition which describes the behavior of each type for authors.
In web-animations-2 we have a more implementor-facing definition which describes the behavior of each type for implementors.
See [@flackr's comment here](https://github.com/w3c/csswg-drafts/pull/10128#discussion_r1598791377).

Issue #11151, opened 2024-11-04
[selectors] Proposal: Update :visited language to state UAs must preserve user privacy


kyraseevers:
It is well-known that styling visited links differently from unvisited links opens the door to attacks that leak the user’s browsing history. Currently, there are brief mentions of preventing these attacks in Selectors 4. Recently, we have seen cross-browser interest and support for strengthening this language, specifically with respect to the proposal to partition :visited links history: https://github.com/explainers-by-googlers/Partitioning-visited-links-history. This was originally proposed in [WICG issue #100](https://github.com/WICG/proposals/issues/100).

[Selectors Level 4 currently states](https://www.w3.org/TR/selectors-4/#link):
> “Since it is possible for style sheet authors to abuse the :link and :visited pseudo-classes to determine which sites a user has visited without the user’s consent, UAs may treat all links as unvisited links or implement other measures to preserve the user’s privacy while rendering visited and unvisited links differently.”

This might become:
> “Since it is possible for style sheet authors to abuse the :link and :visited pseudo-classes to determine which sites a user has visited without the user’s consent, UAs must either treat all links as unvisited links or attempt to preserve the user’s privacy while rendering visited and unvisited links differently. Once a user visits a link, UAs should not style that link as :visited across all sites. Rather, UAs should only style a link as :visited if the user has previously visited it from this specific context. Browsers can experiment with the scope of the :visited context based on UX and performance outcomes.”

We can then provide more detail in the Appendix regarding our suggested algorithm for achieving this: triple-key partitioning. On the user’s end, this will mean that the browser styles links as :visited if and only if they have been visited from this top-level site and frame origin before*.

*However, we are currently experimenting with an exception to this rule, which would include the concept of “self links.” For a site foo.com, links displayed within this site that also point to foo.com may be styled as :visited even if they have not been visited from this top-level site and frame origin before (this is because the server already knows that this subpage has been visited). Self links may only be styled in top-level frames or same-origin subframes to prevent cross-site leaks.

[...]

PR #11695, opened 2025-02-11
[selectors-4] Terminology edits to the :visited footnote #11151

kyraseevers:
This PR makes a few minor terminology changes to the Partitioned :visited links algorithm footnote.

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -4270,8 +4270,8 @@ while still preserving as much of the <em>usefulness</em> of '':visited'' as pos
 	add an entry to |visited history|
 	recording the navigation's destination URL,
 	the origin of the page containing the link or script,
-	and the (schemeful) site of the top-level site containing that page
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -4259,7 +4259,7 @@ while still preserving as much of the <em>usefulness</em> of '':visited'' as pos
 	containing [=tuples=] of three pieces of information:
```
kyraseevers, line 4259: On line 4254 above, I wanted to note that we recently had an exploit reported which takes advantage of Chrome filtering out 4XX-5XX status codes from History to leak the existence of subpages via :visited history (for more information please see https://github.com/explainers-by-googlers/Partitioning-visited-links-history/issues/8). 

This is an issue we are unable to resolve with partitioning but are working to resolve via a separate solution. However, we won't be able to get rid of the mitigations entirely until this issue is resolved.

PR #11701, opened 2025-02-12
[cssom-view-1] Minor corrections to scrolling algorithms

AtkinsSJ:
3 small corrections to the scrolling-related algorithms that came up while implementing them:

1. The "Return _position_." step in "determine the scroll-into-view position" makes more sense at the top level.
2. The `behavior` parameter to "determine the scroll-into-view position" is not used and seems to be a leftover from before it was split up.
3. Say what value to check `container` against in `scrollIntoView()`. I'm not 100% sure on this but out of the two options, "nearest" makes the most sense.


emilio (approved):

emilio:
It seems this needs a rebase, sorry for not getting to this sooner.

Issue #11672, opened 2025-02-07
[css-color-hdr] Clarification on the grammar / spec text for `dynamic-range-limit-mix()`


weinig:
CSS Color HDR specifies the grammar for `dynamic-range-limit-mix()` is:

```
dynamic-range-limit-mix() = dynamic-range-limit-mix( [ <ident> && <percentage [0,100]> ]#)
```
(https://drafts.csswg.org/css-color-hdr/#dynamic-range-limit-mix)

The examples however make it seems like the actual grammar should be something more like:

```
dynamic-range-limit-mix() = dynamic-range-limit-mix( [ <'dynamic-range-limit'> && <percentage [0,100]> ]#)
```

Perhaps with a more specific [value multiplier](https://drafts.csswg.org/css-values-4/#component-multipliers) than `#`? I don't see any examples in the spec or tests that less than two, so maybe `#{2,}`?

The spec text also seems incorrect. It currently states:

> This function takes two [dynamic-range-limit](https://drafts.csswg.org/css-color-hdr/#propdef-dynamic-range-limit) values, converts them internally to a value in stops above media white, and alters the display to the result of mixing them by the specified amount. The actual calculated result is not exposed.

The examples all show using three, so at least the "two" there should be updated.

cc (@svgeesus, @ccameron-chromium)

PR #11703, opened 2025-02-13
[css-color-hdr] Clarification on the grammar / spec text for dynamic-range-limit-mix() #11672

weinig:
Fixes the grammar to match examples and text.


weinig:
Oops, I see @svgeesus just added the two or more. How about instead I just make the grammar match that.

svgeesus (approved):
LGTM

PR #11720, opened 2025-02-14
[css-gaps-1] Fixes for gap intersection point definition and pairing algorithm.

kbabbitt:
* Revert commit f8b28b2.
* Update illustration for gap intersection points with spanning items in grid.
* Fix "determine pairs of gap intersection points" algorithm to apply `gap-rule-break` as intended.
* Minor markup enhancements.


kbabbitt:
Formatted preview here: https://kbabbitt.github.io/css-gap-decorations/pr-11720/Overview.html

@oSamDavis I can't add you as a "Reviewer" for some reason but could you ptal?

css-gaps-1/Overview.bs
```diff
@@ -190,9 +183,8 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
 		</dd>
```
oSamDavis, line 183: I wonder if it'll be helpful to add _which_ intersection points are considered for ` -rule-break: intersection`. So maybe along the lines of "Gap decorations start and end at every **valid** gap intersection point. A _valid_ gap intersection point intersection is defined as one that forms a visible cross or T intersection." then maybe a "refer to the steps below."

css-gaps-1/Overview.bs
```diff
@@ -206,82 +198,106 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
+					<li>
+						|break| is ''column-rule-break/none''.
+					</li>
+					<li>
+						|break| is ''column-rule-break/spanning-item'',
+						and |start| is not <a>discontiguous</a> with the first item in |endpoints|.
+					</li>
+					<li>
+						|break| is ''column-rule-break/intersection'',
+						and the intersection is not flanked by a spanning item on any side.
```
oSamDavis, line 274: Followed up with this offline; In addition to the discussion, I think instead of using "intersection", we should use |end|.

kbabbitt: Updated per our conversation.

css-gaps-1/Overview.bs
kbabbitt: Agree that the value definitions need to be updated, but I'd prefer not to introduce another new term ("valid") if we can avoid it. I changed the value definitions to be based around visible intersections and added a link to the algorithm steps for precise details. I also reordered the value definitions here to match the order they appear in the algorithm. Let me know what you think.

oSamDavis: Just took a look and it looks great!

oSamDavis (approved):
New edits look good to me. Thanks!

PR #11724, opened 2025-02-16
[css-view-transitions-1] Fix mistake in setup transition pseudo-elements algorithm

Psychpsyo:
The [setup transition pseudo-elements algorithm](https://drafts.csswg.org/css-view-transitions-1/#setup-transition-pseudo-elements) has no [this](https://webidl.spec.whatwg.org/#this) value, because it is not defined on an IDL interface.
It gets its ViewTransition passed in as a parameter.

This fixes the incorrect syntax for accessing the relevant global object.


Psychpsyo:
@emilio @jakearchibald It's been almost a year and I've now ported this fix to the level 2 spec also. Can I get a review on this?

emilio (approved):
Sorry for missing this. I'm not an editor of this spec, but I'm familiar with it and this seems like a straight forward fix. LGTM!

Issue #11011, opened 2024-10-08
[css-pseudo] `color-scheme` should be an allowed property for highlight pseudos


BenjaminAster:
The fact that the `color-scheme` CSS property [is not allowed](https://drafts.csswg.org/css-pseudo-4/#highlight-styling) on highlight pseudo-elements (e.g. `::selection`) has the side effect that the [`light-dark()`](https://drafts.csswg.org/css-color-5/#light-dark) function effectively cannot be used to style highlight pseudos (it will always consider the "light" value). I think it would make sense to just add `color-scheme` to the [list of allowed properties](https://drafts.csswg.org/css-pseudo-4/#highlight-styling), which would make this possible.

fantasai:
This seems like an oversight.

schenney-chromium:
Yes, I agree. I see no problem implementing this in Chromium.

Async resolution is fine by me.

astearns:
The CSSWG will automatically accept this resolution one week from now if no objections are raised here. Anyone can add an emoji to this comment to express support. If you do not support this resolution, please add a new comment.

Proposed Resolution: Add `color-scheme` to the set of properties that work with highlight pseudo-elements

fantasai:
OK, changed my mind. :) We actually have two lists of properties involved here after resolving #7591 so we need to confirm which one we should be adding it to: the list of applicable properties, or the list of properties we steal from the originating element in order to resolve the applicable properties.

We should resolve this the same way we resolve `forced-color-adjust`.

Moving to the regular agenda.

schenney-chromium:
Thinking much more deeply about it ...

* Layout affecting properties are not allowed on highlight pseudos, so such properties must be taken from the originating element when needed to resolve units .
* Custom properties from originating elements support existing use cases and make the change in behavior possible at all.
* Color modifying properties could go either way from a technical perspective. The do not affect layout so could go on the highlight itself, but it's hard to see why you would want the highlight and the originating element to differ in their color adjustment.

The key question then is whether there is a use case in which you want _different_ `forced-color-adjust` and `color-scheme` for the underlying text and it's highlight. I'm having a hard time seeing the benefits and I can definitely see negative outcomes, such as contrast problems. It's also more work for authors if they need to duplicate their `forced-color-adjust` and `color-scheme` properties onto the highlights.

Does anyone have a use case for `forced-color-adjust` and `color-scheme` differing between a highlight pseudo and it's originating element?

PR #11733, opened 2025-02-18
[CSS Pseudo 4] Remove the comment for resolved Issue #11011 

schenney-chromium:
Issue #11011 was resolved to take forced-colors and color-scheme from the originating element. Remove the note for the issue.

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -926,7 +926,6 @@ Styling Highlights</h3>
   applied to the text
   by the <a>originating element</a> and its ancestors and descendants.
 
-
   For any properties not listed above,
   but which are required to resolve the values of applicable properties,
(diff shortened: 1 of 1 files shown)


schenney-chromium:
Note I'll be adding WPT to cover the `light-dark()` usage in highlights, and fixing it in Chromium if it doesn't currently work.

svgeesus:
> Note I'll be adding WPT to cover the `light-dark()` usage in highlights,

@schenney-chromium Would be great if you could drop the filenames into this issue, so I can add them to the CSS Color 5 spec as well. Bikeshed won't automatically notify me, because they won't be under `css/css-color`

schenney-chromium:
Just one WPT test once it lands (probably tomorrow):
css/css-pseudo/highlight-styling-004.html

svgeesus (approved):

PR #11755, opened 2025-02-20
[css-images-4] Add the use of “longer” hue interpolation for the color wheel in the example

### LINKED ISSUE #1222: [css-images-4] Conic Gradient "color wheel" example is backwards
Conic-gradient() color-stops are distributed clockwise around the circle, but in Example 17 the example rendering shows them being laid out counter-clockwise.

yisibl:
Fix #1222


yisibl:
@tabatkins @LeaVerou PTAL

css-images-4/Overview.bs
```diff
@@ -1701,7 +1701,17 @@ Conic Gradient Examples</h4>
...
+
+		<pre class=lang-css>
+			background: radial-gradient(closest-side, gray, transparent),
+			            conic-gradient(in hsl longer hue, red 0 100%);
+			transform: scaleX(-1);
+			border-radius: 50%;
+			width: 200px; height: 200px;
+		</pre>
+
+		<img src="images/conic5.png" alt="color wheel">
```
yisibl, line 1714: There are some differences between the current image (left) and the CSS rendering, should we update the image?

image.

svgeesus: Yes, we should.

The rendering in Firefox (on Win, at least) also shows a rendering bug, but that is a separate issue

svgeesus (changes requested):
If you could add that image to the PR (separate filename, in case it is also used elsewhere) that would be helpful.

css-images-4/Overview.bs
```diff
@@ -1701,7 +1701,17 @@ Conic Gradient Examples</h4>
 			width: 200px; height: 200px;
 		</pre>
 
-		<img src="images/conic5.png" alt="" >
+		You can also achieve the same effect using "longer" hue interpolation. This is more concise and makes it easy to switch to other color spaces.
+
+		<pre class=lang-css>
+			background: radial-gradient(closest-side, gray, transparent),
+			            conic-gradient(in hsl longer hue, red 0 100%);
+			transform: scaleX(-1);
```
tabatkins, line 1709: This is weird, we should just flip the image instead and go with a standard rainbow.

css-images-4/Overview.bs
jfkthame: @svgeesus You wrote:

> The rendering in Firefox (on Win, at least) also shows a rendering bug, but that is a separate issue

It looks OK to me on my Windows machine. Could you clarify what rendering bug you're seeing, please? (Or file a bug at https://bugzilla.mozilla.org with details.) Thanks!

css-images-4/Overview.bs
tabatkins: You can always just reverse your color stops. That won't work for *this particular* rainbow example (you'll have to add some additional stops instead), but that's fine; using `longer hue` specifically to generate rainbows is something that's neat but not intentionally supported as a feature.

css-images-4/Overview.bs
yisibl: > Yes, we should.
@svgeesus Okay, I'll modify it.

svgeesus: > It looks OK to me on my Windows machine. Could you clarify what rendering bug you're seeing, please?

gradient-bug

https://codepen.io/svgeesus/pen/KwKzgQE

Nightly 137.0a1 (2025-02-26) (64-bit) Windows 11

PR #11757, opened 2025-02-20
[css-round-display-1] Update viewport-fit to property of viewport meta element

flackr:
The `@viewport` rule was removed[1] in favor of using viewport meta properties. This updates viewport-fit to follow this change.

[1] https://www.w3.org/TR/css-viewport-1/#since-the-29-march-2016-working-draft


flackr:
@tabatkins can you take a look? This is updating the spec to match the way browsers are implementing viewport-fit - on the meta element.

tabatkins (approved):
LGTM

Issue #11745, opened 2025-02-19
[css-highlight-api] The algorithm for highlightsFromPoint doesn't handle StaticRanges


smaug----:
The algorithm says "The coordinates x,y fall inside at least one of the [DOMRect](https://drafts.fxtf.org/geometry-1/#domrect)s returned by calling [getClientRects()](https://drafts.csswg.org/cssom-view-1/#dom-range-getclientrects) on range."
But StaticRanges don't have getClientRects method

(I thought I had filed this issue already, but can't seem to find it now)

PR #11779, opened 2025-02-25
[css-highlight-api] highlightsFromPoint: add support for StaticRanges for #11745

ffiori:
Add some clarifications to the algorithm for highlightsFromPoint so it includes StaticRanges too.

See issue https://github.com/w3c/csswg-drafts/issues/11745

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -668,15 +668,19 @@ method must return the result of running these steps:
 	* <var>x</var> is greater than the <a>viewport</a> width excluding the size of a rendered scroll bar (if any)
 	* <var>y</var> is greater than the <a>viewport</a> height excluding the size of a rendered scroll bar (if any)
 1. Otherwise, return a [=sequence=] of [=custom highlights=] given by ordering the highlights contained in this {{HighlightRegistry}} in descending order of [=priority=],
-	excluding the highlights without at least one [=range=] <var>range</var> that satisfies the following constraints:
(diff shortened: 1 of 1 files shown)


css-highlight-api-1/Overview.bs
```diff
@@ -668,15 +668,19 @@ method must return the result of running these steps:
 	* <var>x</var> is greater than the <a>viewport</a> width excluding the size of a rendered scroll bar (if any)
 	* <var>y</var> is greater than the <a>viewport</a> height excluding the size of a rendered scroll bar (if any)
 1. Otherwise, return a [=sequence=] of [=custom highlights=] given by ordering the highlights contained in this {{HighlightRegistry}} in descending order of [=priority=],
-	excluding the highlights without at least one [=range=] <var>range</var> that satisfies the following constraints:
-	1. The coordinates <var>x</var>,<var>y</var> fall inside at least one of the {{DOMRect}}s returned by calling {{Range/getClientRects()}} on <var>range</var>.
+	including only those highlights that contain at least one {{AbstractRange}} <var>abstractRange</var> that satisfies the following:
+
+	* Let <var>range</var> be a {{Range}} object whose [=start node=] and [=end node=] are set to <var>abstractRange</var>'s [=start node=] and [=end node=] respectively,
```
sanketj, line 673: This is basically a workaround for the fact that StaticRange doesn't support getClientRects. This is fine for now, but should we also file an issue to support getClientRects (and presumably getBoundingClientRect) on AbstractRange so that it works for all range types?

sanketj (approved):

css-highlight-api-1/Overview.bs
ffiori: Sounds good, I filed an issue here https://github.com/w3c/csswg-drafts/issues/11813

PR #11782, opened 2025-02-26
[css-values-4] A small fix and some tweaks to calc serialization

### LINKED ISSUE #11781: [css-values-4] Calc serialization links to a publicly-inaccessible discussion
At the top of the [Mathematical Expressions/Serialization](https://drafts.csswg.org/css-values-4/#calc-serialize) section, we have this note:

> ISSUE 3
> This section is still [under discussion](https://lists.w3.org/Archives/Member/w3c-css-wg/2016AprJun/0239.html).

(And likewise this is in the [published W3C version](https://www.w3.org/TR/css-values-4/#calc-serialize).)

There are a few issues with this, as an implementer:
- What aspects are undecided?
- Is it even still under discussion? Judging by the URL that discussion is from almost 9 years ago.
- If there are still parts of this that need ironing out, excluding non-csswg implementers (unintentionally or not) means fewer people are able to collaborate the remaining problems.

It's true that css-values-4 is in working draft status, but in practice, `calc()` has long been in common use and websites (and WPT) rely on specific serialization behaviour.

AtkinsSJ:
Hopefully these are self-explanatory. I think they all count as editorial.

The numbering change is because it felt awkward to have non-numbered sequential steps in between an outer numbered list and an inner numbered list for Sum and Product nodes. But while I was at it, it made sense to also do that for Negate and Invert. Steps 2 and 3 feel simple enough that it's not necessary.

Closes #11781.

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -5139,78 +5139,82 @@ Serialization</h3>
 			treating the node's children as the function's comma-separated [=calculation=] arguments,
 			and return the result.
 
-		4. If |root| is a Negate node,
-			let |s| be a [=string=]
-			initially containing "(-1 * ".
+		4. If |root| is a Negate node:
+			
+			1. Let |s| be a [=string=]
+			   initially containing "(-1 * ".
 
-			[=serialize a calculation tree|Serialize=] |root|'s child,
-			and append it to |s|.
+			2. [=serialize a calculation tree|Serialize=] |root|'s child,
+			   and append it to |s|.
 
-			Append ")" to |s|,
-			then return it.
+			3. Append ")" to |s|,
+			   then return it.
 
-		5. If |root| is an Invert node,
-			let |s| be a [=string=]
-			initially containing "(1 / ".
+		5. If |root| is an Invert node:
 
-			[=serialize a calculation tree|Serialize=] |root|'s child,
-			and append it to |s|.
+			1. Let |s| be a [=string=]
+			   initially containing "(1 / ".
 
-			Append ")" to |s|,
-			then return it.
+			2. [=serialize a calculation tree|Serialize=] |root|'s child,
+			   and append it to |s|.
 
-		6. If |root| is a Sum node,
-			let |s| be a [=string=]
-			initially containing "(".
+			3. Append ")" to |s|,
+			   then return it.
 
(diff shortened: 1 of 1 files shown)


css-values-4/Overview.bs
```diff
@@ -5141,78 +5139,82 @@ Serialization</h3>
 			treating the node's children as the function's comma-separated [=calculation=] arguments,
 			and return the result.
 
-		4. If |root| is a Negate node,
-			let |s| be a [=string=]
-			initially containing "(-1 * ".
+		4. If |root| is a Negate node:
+			
+			1. Let |s| be a [=string=]
+			   initially containing "(-1 * ".
```
tabatkins, line 5145:
```suggestion
				initially containing "(-1 * ".
```

css-values-4/Overview.bs
```diff
@@ -5141,78 +5139,82 @@ Serialization</h3>
...
+
+			2. [=serialize a calculation tree|Serialize=] |root|'s child,
+			   and append it to |s|.
```
tabatkins, line 5148:
```suggestion
				and append it to |s|.
```

css-values-4/Overview.bs
```diff
@@ -5141,78 +5139,82 @@ Serialization</h3>
...
+
+			3. Append ")" to |s|,
+			   then return it.
```
tabatkins, line 5151:
```suggestion
				then return it.
```

css-values-4/Overview.bs
```diff
@@ -5141,78 +5139,82 @@ Serialization</h3>
...
-			Append ")" to |s|,
-			then return it.
+			1. Let |s| be a [=string=]
+			   initially containing "(1 / ".
 
-		5. If |root| is an Invert node,
-			let |s| be a [=string=]
-			initially containing "(1 / ".
+			2. [=serialize a calculation tree|Serialize=] |root|'s child,
+			   and append it to |s|.
```
tabatkins, line 5159:
```suggestion
				and append it to |s|.
```

tabatkins (commented):
just adding some syntax nits, still need to review in more detail

tabatkins (commented):
just adding some syntax nits, still need to review in more detail

AtkinsSJ:
Thanks! Sorry about those.

PR #11794, opened 2025-02-27
[css-values-5] Move "exploratory draft" issue from *progress() to *mix()

danielsakhapov:
Since there haven't been any discussions about *progress() for a long time, and both WebKit and Blink have implemented it, I suggest to move this warning to *mix() functions section that's currently under heavy discussion.


danielsakhapov:
@tabatkins @fantasai ptal

PR #11800, opened 2025-02-28
[css-2024] Matching CSS recommendation names to real names

BartolomeSintes:
In sections 2.1 and 2.4 the names of several included recommendations do not match the names that appear in the recommendations. Now they are the same.


svgeesus:
I will merge this, but I doubt an update will be published because css-2025 is the current active work.

Still, in case we do, don't want to loose this.

PR #11812, opened 2025-03-03
[css-animations-2] setting `effect` should not prevent the timeline to set set via `animation-timeline`

graouts:
While most `animation-*` CSS properties should _not_ have an effect once a `CSSAnimation` object has had its `effect` property set via the bindings, the `animation-timeline` property, like the `animation-play-state` and `animation-name` properties, should not follow that pattern since it is relevant to the _animation_ and not the animation _effect_.

The relevant WPT test is updated in https://github.com/web-platform-tests/wpt/pull/51057.

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -142,7 +142,7 @@ the changes from the programming interface take precedence as follows:
     to <code>null</code>
     or some {{AnimationEffect}} other than the original {{KeyframeEffect}},
     all subsequent changes to animation properties other than
-    'animation-name', 'animation-play-state' or 'animation-timeline
+    'animation-name', 'animation-play-state', or 'animation-timeline'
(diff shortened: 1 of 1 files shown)


css-animations-2/Overview.bs
```diff
@@ -142,7 +142,7 @@ the changes from the programming interface take precedence as follows:
     to <code>null</code>
     or some {{AnimationEffect}} other than the original {{KeyframeEffect}},
     all subsequent changes to animation properties other than
-    'animation-name' or 'animation-play-state'
+    'animation-name', 'animation-play-state' or 'animation-timeline
```
annevk, line 145:
```suggestion
    'animation-name', 'animation-play-state', or 'animation-timeline'
```

annevk (approved):
Looks good modulo serial comma and missing closing apostrophe.

PR #11816, opened 2025-03-04
[css-mixins-1] Allow omitting custom function arguments

cdoublev:
[`@function`](https://drafts.csswg.org/css-mixins-1/#at-ruledef-function) parameters are optional. So I think one or more custom function arguments (`#`) is an oversight.


tabatkins:
Yes, in the previous version of the spec it didn't make much sense to have no arguments, since that would just mean you were defining a constant value, but now that the function body sees custom properties on the element it's reasonable.

PR #11817, opened 2025-03-04
[css-shadow-parts-1] replace data-x attribute with lt

foolip:
data-x is used in Wattsi (HTML) but doesn't work in Bikeshed.


tabatkins:
Lol, lt doesn't work either, when it's on a `<span>`. I'll fix it up, one sec.

tabatkins:
All right, fixed in 30765a4db

Gonna go ahead and clean up all the other bikeshed issues too.

PR #11821, opened 2025-03-04
[css-mixins-1] Typo: default return type is `type(*)`

cdoublev:
The default return type of [`@function`](https://drafts.csswg.org/css-mixins-1/#at-ruledef-function) is `*` but is not allowed by the grammar ([`<css-type>`](https://drafts.csswg.org/css-mixins-1/#typedef-css-type)).

This PR adds it as an alternative of `<css-type>`.


tabatkins:
Ah, this was just a typo in the spec, it should say the default is `type(*)`

PR #11823, opened 2025-03-04
Forced Colors Mode WebDriver Emulation

alisonmaher:
Forced Colors Mode is difficult to test via WPT because it is dependent on OS state (Window High Contrast) to trigger the feature to light up.

This change builds on a drafted [webdriver-bidi emulation command](https://www.w3.org/TR/webdriver-bidi/#module-emulation) (see https://github.com/w3c/webdriver-bidi/pull/905) to define a new forced colors emulation (https://alisonmaher.github.io/alisonmaher/css-color-adjust-1/Overview.html#emualte-forced-colors-mode) that keeps track of the current forced colors mode emulation state.

The change also updates the forced colors mode algorithm (https://alisonmaher.github.io/alisonmaher/css-color-adjust-1/Overview.html#forced) to inform the UA what to do when the forced colors mode emulation state is either `"light"` or `"dark"`.

CSSWG resolution for this change: https://github.com/w3c/csswg-drafts/issues/11824#issuecomment-2769343959

This change also fixes a few unrelated tabbing issues in the CSS Color Adjust 1 spec.

### DIFF
--- a/css-color-adjust-1/Overview.bs
+++ b/css-color-adjust-1/Overview.bs
@@ -24,19 +24,14 @@ spec:css-color-4; type:property; text:color
 spec:fill-stroke-3; type:property;
 	text:fill
 	text:stroke
-spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/#; type: dfn;
-	text: error; url: dfn-error
-	text: getting a property; url: dfn-getting-properties
-	text: remote end steps; url: dfn-remote-end-steps
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/; type: dfn;
+	text: navigable; for:Window; url: nav-history-apis.html#window-navigable
+	text: top-level traversable; for:/; url: document-sequences.html#top-level-traversable
 </pre>
 
 <pre class=anchors>
 url: https://html.spec.whatwg.org/multipage/semantics.html#meta-color-scheme;
 	type: attr-value; for: meta/name; spec: html; text: color-scheme;
-</pre>
-<pre class=anchors>
-url: https://262.ecma-international.org/5.1/#sec-4.3.11; type: dfn;
-	spec: ecmascript; text: null;
 </pre>
 
 <link rel="stylesheet" href="style.css" />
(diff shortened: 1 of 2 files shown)


css-color-adjust-1/Overview.bs
```diff
@@ -24,13 +24,27 @@ spec:css-color-4; type:property; text:color
...
+	text: error code; url: dfn-error-code
+	text: extension command; url: dfn-extension-commands
+	text: extension commands; url: dfn-extension-commands
+	text: getting a property; url: dfn-getting-properties
+	text: invalid argument; url: dfn-invalid-argument
+	text: null; url: dfn-null
+	text: remote end steps; url: dfn-remote-end-steps
+	text: session; url: dfn-session
+	text: success; url: dfn-success
+	text: undefined; url: dfn-undefined
```
alisonmaher, line 38: The links for "undefined" and "null" do not work, but I am not sure why. The rest of them are working as expected.

css-color-adjust-1/style.css
```diff
@@ -0,0 +1,24 @@
+.swatch {
```
alisonmaher, line 1: Credit to the CSS Color spec - I reused some of these styles to create the system color swatches.

css-color-adjust-1/Overview.bs
svgeesus: @tabatkins perhaps you can help with this?

tabatkins: Ah, you're using `anchors` values (defining brand new anchors that don't exist in the linking database), but this is a `link-defaults` block (declaring defaults for your autolinks).

If all the others work, then you don't need them at all (or possibly they're ambiguous and these lines, interpreted as link-defaults, are fixing their ambiguity).

These terms in particular, tho, aren't WebDriver definitions at all, they're from the ECMAScript spec. `undefined` is defined properly, use `{{globalThis/undefined}}` to link it. (You can find this by running `bikeshed refs --text=undefined`.) `null` doesn't seem to have a definition, for some reason, I suppose that'll have to get a `<pre class=anchors>` line.

alisonmaher: That did the trick, thanks @tabatkins!

And you were right, some of those weren't required. I removed the ones that were unambiguous, updated `undefined` to use `{{globalThis/undefined}}`, and added `<pre class=anchors>` for `null`.

tabatkins (approved):

alisonmaher:
Opening for official review. Thanks @tabatkins for already taking a look. @fantasai feel free to lmk if you have any feedback. In the meantime, I'll reach out to the WebDriver folks to see if they have any thoughts on this change.

alisonmaher:
Adding @shs96c in comment since I wasn't able to add as a direct reviewer.

He is one of the WebDriver spec editors, and said that the change looks good overall, but offered to take a closer look at the content of the change once back from vacation.

He also noted that there is a new version of the WebDriver spec: 

> "The only comment I would make is that the main focus on the Browser Tools and Testing WG is now WebDriver BiDi. Browser vendors are focusing on implementing that and not the “Classic” spec, so for adoption I’d suggest extending BiDi. 
> 
> The main difference between the two is that while the original spec used URL templates and the verbs of HTTP, BiDi is event based, passing messages using a websocket. In terms of addding to the spec, I think either approach requires about the same amount of text."

So in the meantime, I'll look at the Bidi version of the spec and make the necessary updates to move over to that model

svgeesus (approved):
Looks good to me

Issue #10389, opened 2024-06-04
[css-cascade] Allow declarations directly in @scope?


andruud:
Currently, `@scope` only accepts _rules_ in its body. If you want to apply something to the scoping root itself, you need to use a `:scope` rule:

```
@scope (div) {
  :scope {
    color: green;
  }
}
```

However, ever since [relaxed nesting](https://github.com/w3c/csswg-drafts/issues/7961) was introduced (where we first try to parse something as a declaration, and otherwise try it as a nested rule), we should be unblocked if we wanted to support direct declarations in the rule.

Loirooriol:
But would the rule still be present in CSSOM? Or are you proposing that the `CSSScopeRule` should be able to directly contain declarations?

mirisuzanne:
I think the main source of confusion here would be determining the specificity. Currently these have different specificities:

```css
@scope (#foo) {
  p { /* 0,0,1 */ }
  :scope p { /* 0,1,1 */ }
  & p { /* 1,0,1 */ }
}
```

That's already somewhat confusing, since `:scope p` and `p` have identical behavior. When you remove the nested `p` and consider `:scope` vs bare declarations (with implied-`:scope`), it should likely have the same specificity implications? `0,1,0` vs `0,0,0`?

I don't know if that causes round-trip issues when captured in CSSOM.

andruud:
@mirisuzanne Yeah, true. We already have this problem when the `@scope` is a nested group rule, though:

```
div {
  @scope (#foo) {
    color: green; /* Specificity=? */
  }
}
```

(To be clear, the above is already valid today).

> it should likely have the same specificity implications? 0,1,0 vs 0,0,0?

So you're expecting 0,0,0 here, right?

mirisuzanne:
> So you're expecting 0,0,0 here, right?

That's both my expectation, and a somewhat surprising result.

> color: green; /* Specificity=? */

This would be 0,0,1 - a bit less surprising to me.

andruud:
> This would be 0,0,1 - a bit less surprising to me.

No, keep in mind that the `<scope-start>` selector [acts as the parent rule](https://drafts.csswg.org/css-nesting-1/#nesting-at-scope) for `&`here. So per spec this is 1,0,1. (Warning: not what Blink actually does).

mirisuzanne:
[...]

PR #11832, opened 2025-03-05
[css-mixins-1] Flesh out CSSFunctionRule

andruud:
CSSFunctionRule has so far ignored the existence of a prelude in the `@function` rule. This adds CSSOM support for the function name, return type, and parameters.

### DIFF
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -832,7 +832,7 @@ dictionary FunctionParameter {
 	<dd>
 		The [=parameter type|type=] of the [=function parameter=],
 		represented as a [[css-properties-values-api-1#syntax-strings|syntax string]],
-		or <code>"type(*)"</code> of the [=function parameter|parameter=] has no type.
+		or <code>"type(*)"</code> if the [=function parameter|parameter=] has no type.
 
 	<dt>defaultValue
 	<dd>
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -837,7 +837,7 @@ dictionary FunctionParameter {
 	<dt>defaultValue
 	<dd>
 		The [=default value=] of the [=function parameter=],
-		if any.
+		or `null` if the argument does not have a default.
 </dl>
 
 
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -787,12 +787,59 @@ or acting as if nothing exists at that location otherwise.
 CSSOM {#cssom}
 ==============
 
+The {{CSSFunctionRule}} Interface {#the-function-interface}
+-----------------------------------------------------------
+
 The {{CSSFunctionRule}} interface represents a ''@function'' rule.
 
 <pre class='idl' export>
 [Exposed=Window]
-interface CSSFunctionRule : CSSGroupingRule { };
+interface CSSFunctionRule : CSSGroupingRule {
+	readonly attribute CSSOMString name;
+	sequence&lt;FunctionParameter&gt; getParameters();
+	readonly attribute CSSOMString returnType;
+};
 </pre>
+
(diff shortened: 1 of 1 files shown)


tabatkins:
I presume we'd want to represent the body as well, right? As a .style/.childRules pair, like style rules do?

(The rest looks good on a quick skim.)

andruud:
> I presume we'd want to represent the body as well, right? As a .style/.childRules pair, like style rules do?

It's there, it's just not easily visible: we get `childRules` by inheriting from `CSSGroupingRule`. (Should probably be made more easily visible.)

> .style

There is intentionally no `style` attribute on `CSSFunctionRule`. The "leading" block of declarations (if any) is just wrapped in a `CSSFunctionDeclarations` rule and put in `childRules`. This is consistent with other grouping rules affected by nested declarations, i.e. nested grouping rules, and even `@scope` top-level as of https://github.com/w3c/csswg-drafts/issues/10389. I think it's more consistent this way, vs. treating the leading block as special.

`CSSFunctionDeclarations` has a style attribute, though. :-)

> (The rest looks good on a quick skim.)

Like briefly discussed elsewhere, should we be using a `CSSOMString?` return value for attributes that possibly have no value? (And return `null` instead of `""` when there is no value?)

css-mixins-1/Overview.bs
```diff
@@ -787,13 +787,60 @@ or acting as if nothing exists at that location otherwise.
...
+		If the [=custom function=] has no return type,
+		returns <code>"type(*)"</code>.
+</dl>
+
+
+<pre class='idl' export>
+dictionary FunctionParameter {
+	required CSSOMString name;
+	required CSSOMString type;
+	CSSOMString defaultValue;
```
tabatkins, line 822:
```suggestion
	CSSOMString? defaultValue;
```

css-mixins-1/Overview.bs
```diff
@@ -787,13 +787,60 @@ or acting as if nothing exists at that location otherwise.
...
+<dl dfn-for=FunctionParameter>
+	<dt>name
+	<dd>
+		The name of the [=function parameter=].
+
+	<dt>type
+	<dd>
+		The [=parameter type|type=] of the [=function parameter=],
+		represented as a [[css-properties-values-api-1#syntax-strings|syntax string]],
+		or <code>"type(*)"</code> of the [=function parameter|parameter=] has no type.
```
tabatkins, line 835:
```suggestion
		or <code>"type(*)"</code> if the [=function parameter|parameter=] has no type.
```

css-mixins-1/Overview.bs
```diff
@@ -787,13 +787,60 @@ or acting as if nothing exists at that location otherwise.
...
+
+	<dt>defaultValue
+	<dd>
+		The [=default value=] of the [=function parameter=],
+		if any.
```
tabatkins, line 840:
```suggestion
		or `null` if the argument does not have a default.
```

tabatkins:
> Like briefly discussed elsewhere, should we be using a CSSOMString? return value for attributes that possibly have no value? (And return null instead of "" when there is no value?)

Yup, an *empty* default and *no* default are definitely distinct values, and need to be represented differently in the OM. I've gone ahead and made the default `CSSOMString?` to accommodate.

andruud:
> > Like briefly discussed elsewhere, should we be using a CSSOMString? return value for attributes that possibly have no value? (And return null instead of "" when there is no value?)
> 
> Yup, an _empty_ default and _no_ default are definitely distinct values, and need to be represented differently in the OM. I've gone ahead and made the default `CSSOMString?` to accommodate.

Right, looks like we already did represent it differently: no default = no entry in the dictionary; empty default = empty CSSOMString.

I take it we then prefer an explicit null entry if there is no default?

tabatkins:
> no default = no entry in the dictionary; empty default = empty CSSOMString.

Yeah, much better to have a consistent object shape with a null-valued property.

Issue #1008, opened 2017-02-06
CSS pseudo-selector for h1-h6?


pmaxhogan:
I frequently find myself writing CSS like this:
```CSS
h1, h2, h3, h4, h5, h6{
     ...
}
```
It annoys me that I have to type `h1, h2, h3, h4, h5, h6` so much. But it gets worse:
```CSS
article h1, article h2, article h3, article h4, article h5, article h6{
     ...
}
```
or even
```CSS
article h1.bold:hover, article h2.bold:hover, article h3.bold:hover, article h4.bold:hover, article h5.bold:hover, article h6.bold:hover{
     ...
}
```
Wouldn't it be nice to just type
```CSS
:heading{
     ...
}
```
or
```CSS
article :heading{
     ...
}
```
or 
```CSS
article :heading.bold:hover{
     ...
}
```
`:heading` just matches h1-h6. It would be easy to implement for browser vendors and easy to polyfill. But extremely helpful.

# Having a heading selector like `:heading` would be AWESOME!

notoriousb1t:
Peanut gallery here.  :heading seems ultra specific.  Maybe we would be better served by a tag selector?  Something like 

```css
/h[0-6]/ {
   font-weight: bold;
}
```

SelenIT:
There is the CSS Custom Selectors [proposal](https://drafts.csswg.org/css-extensions/#custom-selectors) (it even has its [PostCSS plugin](https://github.com/postcss/postcss-custom-selectors)) which would solve this exact issue and much more. But doesn't `:matches(h1, h2, h3)` from Selectors level 4 already solve the main problem here (code duplication)?

programmer5000-com:
See #1010

annevk:
https://github.com/whatwg/html/pull/3499 contains a `:heading`/`:heading()` pseudo-class selector proposal. @fantasai wrote that Selectors should maybe define that given that other host languages might have utility for it too. If it's successful for HTML that seems reasonable to me.

tabatkins:
> But doesn't :matches(h1, h2, h3) from Selectors level 4 already solve the main problem here (code duplication)?

No, because HTML has ways of assigning heading levels that go beyond tagnames - `h1` nested inside of sectioning elements automatically becomes a second/third/etc level heading, as appropriate.  (Plus, `:heading` really is pretty dramatically shorter to type than `:matches(h1, h2, h3, h4, h5, h6)`, and is a pretty common thing to want to style.)

[...]

Issue #351, opened 2016-07-28
[css-content][selectors] :level(an+b) Pseudo Class


Crissov:
The [generation of bookmark levels](https://drafts.csswg.org/css-content/#bookmark-generation) has been discussed e.g. in #204 and [in November 2015 on www-style](https://lists.w3.org/Archives/Public/www-style/2015Nov/0310.html). @frivoal had shown that it’s nearly impossible to recreate the HTML5 outline algorithm with existing Selectors, so `bookmark-level: auto` would make some sense. That value hasn’t been added to the drafts, though, and I would like to propose a different solution, a new pseudo-class.

Nesting, stacking or hierarchy levels are of interest for styling in several cases, e.g. headings or sections and lists in document markup languages or layers and sprites in images. So let there be an abstract **level** that depends on the host language and may differ among its element types. Since a pseudo class increases specify, the followin rulesets should describe the usually desired behavior for HTML:

    :level(0) { bookmark-level: none; }
    h1, h2:level(1), h3:level(1), h4:level(1), h5:level(1), h6:level(1) { bookmark-level: 1; }
    h2, h1:level(2), h3:level(2), h4:level(2), h5:level(2), h6:level(2) { bookmark-level: 2; }
    h3, h1:level(3), h2:level(3), h4:level(3), h5:level(3), h6:level(3) { bookmark-level: 3; }
    h4, h1:level(4), h2:level(4), h3:level(4), h5:level(4), h6:level(4) { bookmark-level: 4; }
    h5, h1:level(5), h2:level(5), h3:level(5), h4:level(5), h6:level(5) { bookmark-level: 5; }
    h6, h1:level(6), h2:level(6), h3:level(6), h4:level(6), h5:level(6) { bookmark-level: 6; }

The syntax inside parentheses would probably follow the existing `an+b` syntax, but I‘m not sure we’d need to differentiate `:first-level`, `:last-level` and `:nth-level()`, but it would be consistent.

Likewise, one could easily ensure the same marker style for items in nested lists, no matter if the structure mixes `ol`, `ul` and `dl`.

    ul>li:level(2) { list-style-type: disc; }

tabatkins:
@frivoal Yeah, the semantics are well-defined by the HTML outline algorithm.  Sections are introduced either by the section elements, or implicitly by headings.

frivoal:
[...]

PR #11836, opened 2025-03-05
[css-selectors-5] add `:heading`, `:heading(An+B)` pseudo classes

### LINKED ISSUE #10296: [selectors] Adding a `:heading()` selector for headingoffset?
https://github.com/whatwg/html/issues/5033 proposed to add a new attribute (now called `headingoffset`) which can offset a containers exposing heading level by a given amount, for example `<div headingoffset=1><h1></h1></div>` renders a heading with a level `2` (1+1 = 2) in the accessibility tree.

This begs the question if it's possible to somehow correctly _style_ elements according to their computed offset. Given `headingoffset` is accumulative (it does a flat tree walk) it can be quite difficult to appropriately select for such combinations. 

So, here I am, proposing a pseudo selector to allow for selecting the computed heading level. I propose `:heading(n)` as a selector for heading elements exposes as level `n`. An `<h1>` (with no `headingoffset` parents) would then be selectable as either `:heading(1)` or `h1`, but for example `<div headingoffset=1><div headingoffset=2><h1></h1></div></div>` would be selectable as `:heading(4)` (1+2+1 = 4).

I want to stress that I am of the opinion that `aria-level` or `role=heading` should _not_ impact `:heading()`; they are readily selectable today and I don't think aria should have that dramatic an impact on other selectors.

I think it would be _really helpful_ for us to consider adjusting the UA stylesheets such that the existing for styles like `h1{}` are extended to `h1, :heading(1){}`. I know this looks scarily like the old document outline but as `headingoffset` is both _explicit_ and _opt in_, I'm confident this doesn't introduce the same issus as document outlines.

Thoughts?

keithamus:
This fixes #10296, #351, #1008 by specifying the `:heading` and `:heading(An+B)` pseudo classes, which will match against elements with a defined [heading level](https://html.spec.whatwg.org/multipage/sections.html#heading-level). As it stands, the current elements with a heading level are `h1`, `h2`, `h3`, `h4`, `h5`, and `h6`.

There is a proposal being actively prototyped to allow for more _dynamic_ heading levels that do not strictly adhere to their tagname (see https://github.com/whatwg/html/issues/5033 for more). This proposal (adding the `headingoffset`, `headingreset` attributes) will make it significantly more complex to unilaterally style elements with a heading level of, say, 1, raising the importance of the selectors proposed in this PR.

I've https://github.com/w3c/csswg-drafts/labels/Agenda%2B the issue #10296 in order to discuss _this_, the proposed selectors, that we may resolve and merge this PR, but in the meantime I would appreciate editorial review from those that have time.

### DIFF
--- a/selectors-5/Overview.bs
+++ b/selectors-5/Overview.bs
@@ -16,6 +16,7 @@ Abstract: Selectors Level 5 describes the selectors that already exist in [[!sel
 <pre class="link-defaults">
 spec:selectors-4; type:dfn; text:selector
 spec:html; type:dfn; text:states set
+spec:html; type:dfn; text:heading level
 </pre>
 
 <h2 id="intro">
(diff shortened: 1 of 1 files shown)


tabatkins (approved):

zcorpan (approved):

annevk:
This should really have had a corresponding HTML PR. Will one be created?

Issue #11869, opened 2025-03-09
[css-forms-1] Checkbox stylesheet missing indeterminate handling


lukewarlow:
Currently the checkbox [stylesheet](https://drafts.csswg.org/css-forms-1/#checkboxes--radios) is missing any definition of indeterminate in the UA stylesheet this means it will render identically to a checked/unchecked (depending on the full state) checkbox.

lukewarlow:
Adding something like this might work, uses an em-dash for the horizontal line. (note it doesn't need `:not([switch])` as that never matches indeterminate)

```
input[type=checkbox]:indeterminate::checkmark {
    content: '\2500' / '';
}
```

lukewarlow:
We potentially also want a `user-select: none` on the checkmark parts styles else I think you get odd cursor behaviour.

fantasai:
Do radio buttons have an indeterminate state? What should be the styling for those?

lukewarlow:
> Do radio buttons have an indeterminate state?

No. It only applies to checkboxes (that aren't switches) and progress elements.

fantasai:
Regarding alignment, maybe we should set `line-height: 1` on the checkmark and radio button `<input>`s. I think this is probably a better move than such an aggressive `text-box` trim? :)

fantasai:
[Various horizontal characters.](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cstyle%3Ediv%20%7B%20width%3A%201em%3B%20height%3A%201em%3B%20border%3Asolid%201px%3B%20text-align%3A%20center%3B%20line-height%3A%201%3B%20margin%3A%200.5em%3B%20font-family%3A%20EDITME%3B%20%20%7D%3C%2Fstyle%3E%0A%3Cdiv%3E%26%23x2015%3B%3C%2Fdiv%3E%0A%3Cdiv%3E%26%23x2014%3B%3C%2Fdiv%3E%0A%3Cdiv%3E%26%23x2013%3B%3C%2Fdiv%3E%0A%3Cdiv%3E%26%23x2500%3B%3C%2Fdiv%3E%0A%3Cdiv%3E%26%23x2501%3B%3C%2Fdiv%3E%0A%3Cdiv%3E%26%23x2504%3B%3C%2Fdiv%3E%0A)

Alternatively we could be super obvious about it and put a question mark.

lukewarlow:
[...]

PR #11872, opened 2025-03-09
[css-forms-1] Fix ::checkmark styles to apply when input is checked

lukewarlow:
Fix ::checkmark styles to apply when input is checked

Also apply a border-radius to radio inputs.

With this you now end up with the below rendering:

image

image

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -677,17 +677,21 @@ input:is([type=checkbox]:not([switch]), [type=radio]) {
     display: inline-flex;
     align-items: center;
     justify-content: center;
+    content: '';
 }
 
-input[type=checkbox]:not([switch])::checkmark {
+input[type=radio] {
+    border-radius: 100%;
+}
+
+input[type=checkbox]:not([switch]):checked::checkmark {
     content: '\2713' / '';
 }
 
-input[type=radio]::checkmark {
-    content: '';
+input[type=radio]:checked::checkmark {
     background-color: currentColor;
     display: inline-block;
-    border-radius: 100%;
+    border-radius: inherit;
     height: 100%;
     width: 100%;
 }


css-forms-1/Overview.bs
```diff
@@ -677,14 +677,18 @@ input:is([type=checkbox]:not([switch]), [type=radio]) {
     display: inline-flex;
     align-items: center;
     justify-content: center;
+    content: '';
+}
+
+input[type=radio] {
+		border-radius: 100%;
```
lukewarlow, line 684: Potentially want some sort of padding on this too?

css-forms-1/Overview.bs
```diff
@@ -677,14 +677,18 @@ input:is([type=checkbox]:not([switch]), [type=radio]) {
...
 }
 
-input[type=checkbox]:not([switch])::checkmark {
+input[type=checkbox]:not([switch]):checked::checkmark {
     content: '\2713' / '';
```
lukewarlow, line 688: Needs to handle indeterminate aswell see #11869

css-forms-1/Overview.bs
```diff
@@ -677,14 +677,18 @@ input:is([type=checkbox]:not([switch]), [type=radio]) {
...
 }
 
-input[type=radio]::checkmark {
-    content: '';
+input[type=radio]:checked::checkmark {
     background-color: currentColor;
     display: inline-block;
     border-radius: 100%;
```
nt1m, line 694:
```suggestion
    border-radius: inherit;
```

In the unlikely event where someone changes the border-radius of the radio button, this might be useful.

css-forms-1/Overview.bs
nt1m, line 684:
```suggestion
    border-radius: 100%;
```

nt1m: I intentionally didn't put any so it's not annoying to override, but perhaps it might be a good idea. I haven't thought too much about it

nt1m (approved):

css-forms-1/Overview.bs
lukewarlow: Yeah I'll leave it without for now. We can revisit once the styles are more fleshed out as an overall quality question.

nt1m (approved):

PR #11876, opened 2025-03-09
[css-forms-1] Add `::file-selector-button` to selectors for buttons

lukewarlow:
Previously `::file-selector-button` would be missing many of the styles associated with button-like UI.

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -696,6 +696,7 @@ input[type=radio]::checkmark {
 
 ```css
 button,
+::file-selector-button,
 select,
 input[type="color"] {
   border: 1px solid currentColor;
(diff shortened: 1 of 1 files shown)


lukewarlow:
Alternatively I guess we can get this behaviour by removing all references to this pseudo except for a single rule. That way it will automatically get all of these styles because it's a button?

```css
::file-selector-button {
   appearance: base;
}
```

nt1m (approved):

tabatkins:
> That way it will automatically get all of these styles because it's a button?

For future reference: no, because it's a pseudo-element. The fact that it's element-backed by a button doesn't let it match `button` selectors, you do indeed still need to match it manually by pseudo-element name.

Issue #11838, opened 2025-03-05
[css-forms-1] Should `field-sizing: content` be in base appearance stylesheet?


lukewarlow:
It's unclear to me what the sizing model for base appearance styled elements are but it feels like `field-sizing: content` should be the default behaviour.

nt1m:
I'm not too sure about this, the majority of form controls I've seen are fixed size, including real-life physical form control on paper.

I wouldn't be closed to the idea though, if there is evidence that `field-sizing: content` wouldn't end up being a bogus default that constantly needs to get overridden.

lukewarlow:
I suspect for textarea specifically it would be the better default. Even if the default stylesheet defined a min and max height of say 3lh. That way they can override those properties and it just work. Otherwise they'll have to set field-sizing too?

lukewarlow:
Something like the below styles will give you field-sizing: content behaviour but with a default width and height for textual inputs and textareas. These styles mostly match the existing intrinsic sizes (though they're based on the ch unit for a very specific microsoft font so it's only approximate).

```css
textarea {
    field-sizing: content;
    width: 22ch;
    height: 2lh;
}

input:not([type=checkbox], [type=radio], [type=file], [type=color], [type=range]) {
    field-sizing: content;
    width: 20ch;
}
```

With this authors can get the auto growing behaviour by doing something like:

```css
textarea {
   height: auto;
   /* min-height max-height styles work to constrain size. */
}

input:not([type=checkbox], [type=radio], [type=file], [type=color], [type=range])  {
   width: auto;
   /* min-width max-width styles work to constrain size. */
}
```

lukewarlow:
Also base appearance select effectively forces field-sizing: content to true (though currently the computed style still says fixed in chrome), so perhaps we should also add:

```css
select {
   field-sizing: content !important;
}
```

So that the behaviour is easier to debug?

nt1m:
> Also base appearance select effectively forces field-sizing: content to true (though currently the computed style still says fixed in chrome), so perhaps we should also add:
> 
> select {
>    field-sizing: content !important;
> }
> So that the behaviour is easier to debug?

This makes sense to me!

lukewarlow:
Turns out we might need to consider auto stretching from grid-layout

https://github.com/w3c/csswg-drafts/pull/11897#discussion_r1988112923

nt1m:
I'm started to be convinced that we should do something like this, to make sizing of form controls more similar to other elements.

[...]

PR #11879, opened 2025-03-09
[css-forms-1] Add field-sizing to base select styles

lukewarlow:
Base select is always sized based on its content size rather than using the legacy fixed styling, so we set field-sizing to content !important.

See https://github.com/w3c/csswg-drafts/issues/11838#issuecomment-2708823023

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -695,6 +695,11 @@ input[type=radio]::checkmark {
 ## Selects & buttons
 
 ```css
+select {
+  /* Base appearance select always sizes based on its contents. */
+  field-sizing: content !important;
+}
+
 button,
 ::file-selector-button,
 select,


css-forms-1/Overview.bs
```diff
@@ -695,6 +695,11 @@ input[type=radio]::checkmark {
 ## Selects & buttons
 
 ```css
+select {
+  /* Base appearance select always sizes based on its contents. */
+  field-sizing: content !important;
```
lukewarlow, line 700: !important for now but if we decided to enable the legacy layout (size based on layout of longest option) we could remove that modifier and allow authors to set to fixed to enable it.

lukewarlow, line 700: We maybe only want this for dropdown select and not for listbox select but we can handle that once we have a selector to choose between them.

nt1m (approved):

PR #11881, opened 2025-03-10
[web-animations-1] Make commitStyles use an endpoint-inclusive active interval

### LINKED ISSUE #5394: [web-animations] Could commitStyles always write the end state of the animation?

birtles:
Fixes #5394.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -2801,6 +2801,11 @@ Animation effects {#animation-effects}
       <code>max(min(<a>start delay</a> + <a>active duration</a>,
                     <a for="animation effect">end time</a>), 0)</code>
 
+
+  Furthermore, an <dfn>endpoint-inclusive active interval</dfn> flag
+  may be specified when determining the phase.
+  If not specified, it is assumed to be false.
+
   An [=animation effect=] is in the
   <dfn export for="animation effect">before phase</dfn>
   if the animation effect's [=local time=]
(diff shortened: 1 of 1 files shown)


birtles:
Leaving this as a draft until we've determined that this change is Web-compatible.

birtles:
Putting this up for review since the use counter data we've collected in Firefox [suggests this change is safe to make](https://github.com/w3c/csswg-drafts/issues/5394#issuecomment-2954654330).

There are WPT for this [here](https://github.com/web-platform-tests/wpt/blob/master/web-animations/interfaces/Animation/commitStyles.tentative.html) and once this PR is merged, we'll rename `commitStyles.tentative.html` to `commitStyles.html`.

birtles:
@flackr @graouts Review ping?

flackr (approved):
This looks good, thanks! Sorry I didn't see this earlier. The examples are so much cleaner, I'm glad it looked reasonably safe :-).

birtles:
Thank you!

Issue #5717, opened 2020-11-11
[css-counter-styles] The entire rule being invalid vs. not defining a counter style when name/descriptors are wrong


xiaochengh:
https://github.com/w3c/csswg-drafts/issues/1682#issuecomment-335868951 resolved that:

> RESOLVED: treat invalid counter styles the way we treat invalid font faces

The [spec change log](https://drafts.csswg.org/css-counter-styles-3/#changes-jun-2015) says:

> @counter-style rules that are invalid due to missing descriptors just fail to create a counter style; they’re otherwise still valid rules.

I understand these resolutions as: if a @counter-style rule has wrong descriptors, it should still be parsed into a valid rule and appear in the OM; it only fails to create a counter style and hence do not affect any counter.

However, the current spec text makes the entire @counter-style rule invalid in some cases:

> If the system is cyclic, the symbols descriptor must contain at least one counter symbol, or else the @counter-style rule is invalid.

> If the system is fixed, the symbols descriptor must contain at least one counter symbol, or else the @counter-style rule is invalid. 

> If the system is symbolic, the symbols descriptor must contain at least one counter symbol, or else the @counter-style rule is invalid.

> If the system is alphabetic, the symbols descriptor must contain at least two counter symbols, or else the @counter-style rule is invalid.

> If the system is numeric, the symbols descriptor must contain at least two counter symbols, or else the @counter-style rule is invalid.

> If the system is additive, the additive-symbols descriptor must contain at least one additive tuple, or else the @counter-style rule is invalid.

> If a @counter-style uses the extends system, it must not contain a symbols or additive-symbols descriptor, or else the @counter-style rule is invalid.

Should they be revised into not defining a counter style?

andruud:
I'd interpret "invalid" in "or else the @counter-style rule is invalid" to mean "does not define a counter style", but I suppose it could be interpreted both ways. In general it would be nice to have a clearer distinction between "invalid at parse-time" and "invalid at @-rule interpretation-time". There are other @-rule-defining specs that have the same problem as well.

tabatkins:
Saying that a construct is "invalid" always means at parse-time by default. That said, we resolved that this should instead just prevent the rule from defining a counter style, so I've removed the language saying it's invalid, and adding some more detail to the 'system' section about what the requirements are and what they mean.

Issue #1682, opened 2017-08-03
[css-counter-styles] Invalid @counter-style should just not define a counter style, not be ignored


tabatkins:
In <https://github.com/w3c/csswg-drafts/issues/1133>, we just decided that an `@font-face` rule without its minimum required descriptors should be treated as *valid* (so it shows up in the OM), but just doesn't define a font face for the document.

We should be consistent in applying this decision, which means that an `@counter-style` rule without its minimum required descriptors (`system` and either `symbols` or `additive-symbols`) is valid, but does not create a counter style for the document.

upsuper:
Hmmm... OK... If @dbaron doesn't file a bug, I'll do so tomorrow. (Just to avoid filing duplicate bugs like this time :)

upsuper:
Also, to make it clear, I suppose `@counter-style` rules which use an invalid name (`none`, `decimal`, or `disc`) are still invalid, and still shouldn't show up in the OM. Is that correct?

upsuper:
(FWIW, re-purposed an old [Gecko bug 1299097](https://bugzilla.mozilla.org/show_bug.cgi?id=1299097) for this change.)

tabatkins:
I can go either way on that.  I think I'd prefer to treat it as invalid and drop it.

tabatkins:
@upsuper re: invalid names

Yeah, I intend that to actually invalidate the rule, tho I could go either way if necessary.

xiaochengh:
Hi,

The edit work seems incomplete. The spec still gives a number of scenarios that the @counter-style rule is invalid. There are also some inconsistencies, for example:

> If the system is `cyclic`, the `symbols` descriptor must contain at least one counter symbol, or else the @counter-style rule is invalid.

> The @counter-style rule must have a valid `symbols` descriptor ... otherwise, the @counter-style does not define a counter style (but is still a valid at-rule).

Should I understand the intention as: as long as the at-rule has a correct syntax, then it's valid and appear in OM; it defines a counter style only when the required descriptors have correct values.

In other words, `@counter-style inherit {}` is invalid and shouldn't be in OM; while `@counter-style foo {}` is valid but doesn't not define a counter style.

PR #11883, opened 2025-03-10
[css-counter-styles-3] Extended @counter-style containing *-symbols are valid

### LINKED ISSUE #9258: [css-counter-styles-3] `@counter-style` declared with extended `system` and `*-symbols`
https://drafts.csswg.org/css-counter-styles-3/#valdef-counter-style-system-extends

  > If a `@counter-style` uses the `extends` system, it must not contain a `symbols` or `additive-symbols` descriptor, or else the `@counter-style` rule is invalid.

In Chrome/FF (at least), the rule is *valid* and does not define a counter style.

```html
<style>
  @counter-style counter {
    system: extends cyclic;
    symbols: foo bar;
  }
  ul { list-style: counter }
</style>
<ul>
  <li>one</li> <!-- 1. one-->
  <li>two</li> <!-- 2. two -->
</ul>
<script>
  document.styleSheets[0].cssRules.length // 2
</script>
```

At least, considering their related issues/resolutions, I think *invalid* may be an oversight in https://github.com/w3c/csswg-drafts/commit/4068d1027125febca89ac7cc54f262d5a65c1978 / https://github.com/w3c/csswg-drafts/commit/ba2a8cbd69a0205f409fe8019e746d9a81dad9cc.

cdoublev:
Fixes #9258. This PR removes this emphasized part:

  > If a `@counter-style` uses the extends `system`, it must not contain a `symbols` or `additive-symbols` descriptor, **or else the `@counter-style` rule is invalid**.

It was reported in #5717 and should have been fixed in ba2a8cb.

They all follow the resolution in #1682.

---

ba2a8cb and this PR leave multiple requirements (*"must [not] contain"*) without explicitly saying what should happen when they are not met. In my opinion, *"otherwise, the `@counter-style` does not define a counter style (but is still a valid at-rule)"* should be added.

### DIFF
--- a/css-counter-styles-3/Overview.bs
+++ b/css-counter-styles-3/Overview.bs
@@ -709,8 +709,7 @@ Building from Existing Counter Styles: the ''extends'' system
 	rather than taking their initial values.
 
 	If a ''@counter-style'' uses the ''extends'' system,
-	it must not contain a '@counter-style/symbols' or '@counter-style/additive-symbols' descriptor,
-	or else the ''@counter-style'' rule is invalid.
+	it must not contain a '@counter-style/symbols' or '@counter-style/additive-symbols' descriptor.
 
 	If the specified <<counter-style-name>> is an [=ASCII case-insensitive=] match
 	for ''disc'', ''circle'', ''square'', ''disclosure-open'', or ''disclosure-closed''


tabatkins:
> In my opinion, "otherwise, the @counter-style does not define a counter style (but is still a valid at-rule)" should be added.

Yup, sounds good.

Issue #11850, opened 2025-03-06
[css-forms-1] Define pseudo element for textarea resizer


lukewarlow:
Currently there's no pseudo-element for the resizer in the textarea elements, we should define one (and probably improve it).

schenney-chromium:
What aspects of the appearance would be stylable? There's the obvious `color` and maybe `background-color`, but then what?

Would just supporting an image do the trick?

lukewarlow:
For one consistent rendering would be needed, it's not actually always shown in browsers right now. For example iOS Safari doesn't iirc.

Another thing is how it responds to dark mode. Right now there's issues across browsers with contrast.

The handle isn't big enough to meet wcag touch target requirements for example.

It may even be that the current resize UI should be rethought entirely (possibly the edges should be the drag handles rather than a standalone pseudo?)

nt1m:
I originally held off from adding this, mainly because of this:

> It may even be that the current resize UI should be rethought entirely (possibly the edges should be the drag handles rather than a standalone pseudo?)

Perhaps we could add some intermediate `::corner-resizer` or `::resizer(corner)` pseudo-element though until that happens. `::resizer` would probably cause name clashes if a different type of resizer is added.

SebastianZ:
FWIW, a resizer is also shown if an element is a scroll container and the [`resize` property](https://drafts.csswg.org/css-ui/#resize) is set to something else than `none`.

Sebastian

mfreed7:
+1 to supporting a customizable "resizer". It would also be good to just *include* `<textarea>` in https://drafts.csswg.org/css-forms/, since it's only peripherally mentioned in the Appendix.

PR #11897, opened 2025-03-10
[css-forms-1] Add UA style rules for text fields

lukewarlow:
See #11838 as well as discussion in other issues such as https://github.com/w3c/csswg-drafts/issues/11850


css-forms-1/Overview.bs
```diff
@@ -696,6 +696,56 @@ input[type=radio]:checked::checkmark {
     width: 100%;
 }
 ```
+
+## Text fields
+
+```css
+input:not([type]),
+input:is([type=text], [type=search], [type=email], [type=password], [type=tel], [type=url], [type=number]),
```
lukewarlow, line 704: I think this is every textual input?

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,56 @@ input[type=radio]:checked::checkmark {
...
+  resize: both;
+  height: calc(2lh + 4px + 2px);
+  width: calc(22ch + 4px + 2px);
+}
+
+::placeholder {
+  color: color-mix(in srgb, currentColor 50%, transparent);
+  display: none;
+  line-break: anywhere;
+  grid-area: 1/1;
```
lukewarlow, line 732: Used to stack the elements atop each other?

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,56 @@ input[type=radio]:checked::checkmark {
...
+textarea {
+  overflow-y: auto;
+  padding-block: 2px;
+  resize: both;
+  height: calc(2lh + 4px + 2px);
+  width: calc(22ch + 4px + 2px);
+}
+
+::placeholder {
+  color: color-mix(in srgb, currentColor 50%, transparent);
```
lukewarlow, line 729: Possibly too low contrast (same would apply to disabled button-like text)

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,56 @@ input[type=radio]:checked::checkmark {
...
+}
+
+:placeholder-shown::placeholder {
+  display: block;
+}
+
+::field-text {
+  grid-area: 1/1;
+  /* Should be cursor width but that's not exposed */
+  min-width: 1px;
```
lukewarlow, line 742: Not sure what to do about this?

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,56 @@ input[type=radio]:checked::checkmark {
...
+  height: calc(1lh + 2px + 2px);
+  width: calc(20ch + 4px + 2px);
+}
+
+textarea {
+  overflow-y: auto;
+  padding-block: 2px;
+  resize: both;
+  height: calc(2lh + 4px + 2px);
+  width: calc(22ch + 4px + 2px);
```
lukewarlow, line 725: Weird calcs because of `box-sizing: border-box`.

lukewarlow:
cc @mfreed7 due to your comments around flex layout.

css-forms-1/Overview.bs
bfgeek: note an explicit width/height (even if they match what browsers calculate as their intrinsic size) will break things like auto-stretching within grid layout.

lukewarlow: Interesting... Not really sure on a solution to that.

I do think field-sizing: content as the default makes sense still but this might make it trickier...

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,56 @@ input[type=radio]:checked::checkmark {
...
+
+## Text fields
+
+```css
+input:not([type]),
+input:is([type=text], [type=search], [type=email], [type=password], [type=tel], [type=url], [type=number]),
+textarea {
+  display: inline-grid;
+  padding-inline: 2px;
+  field-sizing: content;
```
lukewarlow, line 708: Do we maybe also want to set cursor to text for these?

css-forms-1/Overview.bs
lukewarlow: Would setting a max-width aswell still break the autostretching? Or does it only break when explicitly set as a width or height?

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,51 @@ input[type=radio]:checked::checkmark {
...
+textarea {
+  display: inline-grid;
+  padding-inline: 2px;
+}
+
+input:not([type]),
+input:is([type=text], [type=search], [type=email], [type=password], [type=tel], [type=url], [type=number]) {
+  overflow-x: auto;
+  scrollbar-width: none;
+  padding-block: 1px;
```
lukewarlow, line 714: Should this be 0.0625em? (1/16)

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,51 @@ input[type=radio]:checked::checkmark {
...
+}
+
+textarea {
+  overflow-y: auto;
+  padding-block: 2px;
```
lukewarlow, line 719: ditto

css-forms-1/Overview.bs
lukewarlow: https://github.com/w3c/csswg-drafts/issues/12019#issuecomment-2770409866 per this it seems like calc-size might be the way to go but needs further thinking.

lukewarlow:
@nt1m do you have any thoughts on this? Currently we have nothing defined for these so it would be good to make some progress on it but we don't have a strong definition for placeholder or field-text yet?

css-forms-1/Overview.bs
nt1m: Ideally padding should be consistent across select, buttons and inputs

nt1m:
> @nt1m do you have any thoughts on this? Currently we have nothing defined for these so it would be good to make some progress on it but we don't have a strong definition for placeholder or field-text yet?

They seem reasonable, but also hard to verify without prototyping

lukewarlow:
https://codepen.io/lukewarlow/pen/wBvqovL - here's a custom element with the applied styles that should demonstrate roughly what it looks like.

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,51 @@ input[type=radio]:checked::checkmark {
...
+  grid-area: 1/1;
+}
+
+:placeholder-shown::placeholder {
+  display: block;
+}
+
+::field-text {
+  grid-area: 1/1;
+  min-block-size: 1lh;
```
lukewarlow, line 943: This probably needs a `font-size: max(1em, 24px)` rule

css-forms-1/Overview.bs
```diff
@@ -696,6 +696,51 @@ input[type=radio]:checked::checkmark {
...
+  /* Should be cursor width but that's not exposed */
+  min-inline-size: 1px;
```
lukewarlow, line 945: Should this actually be 24px? So the textarea/input is never smaller than 24px x 24px?

nt1m: I wonder if this really belongs on field-text, ditto for min-block-size/font-size. Ideally they'd just inherit what they need from the parent.

PR #11920, opened 2025-03-12
[css-values-5] Editorial: fix scope of random() values

tidoust:
According to the definitions data model, `dfn-for` needs to a contain a "comma-separated list" of definitions. A recent commit used `|` to separate values definitions for `random()` and `<random-value-sharing>`. References back to the scoped definitions did not work as a result.


tabatkins:
Argh, sorry about that.

PR #11930, opened 2025-03-12
[css-view-transitions-2] `view-transition-name: auto` with ID should only match other IDs

### LINKED ISSUE #11614: [css-view-transitions-2] view-transition-name: auto when matching id should namespace
We have a `view-transition-name: auto` that can use the `id` attribute from the element as the view transition name. (https://www.w3.org/TR/css-view-transitions-2/#auto-vt-name)

I think it can be confusing if, for example, in the old state we match `id=foo` and on the other side that pairs with css `view-transition-name: foo`. 

The proposal is to "namespace" id matching. Basically `id=foo` would only match with `id=foo` and not with `view-transition-name: foo`. Or, in other words, names generated by the `id` attribute can only match elements that have names that are also generated by the `id` attribute.

@nt1m @noamr fyi

noamr:
Closes #11614

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -816,7 +816,7 @@ To resolve the [=used value=] of 'view-transition-name' for |element|:
 	1. If |computed| is a <<custom-ident>>, return |computed|.
 	1. Assert: |computed| is <css>auto</css> or <css>match-element</css>.
 	1. If |computed| is <css>auto</css>, |element| has an associated [=Element/id=], and |computed| is associated with the same [=tree/root=] as |element|'s [=tree/root=], then return a unique string.
-		Two elements with the same [=Element/id=] should match based on this string, regardless of their [=node document].
+		Two elements with the same [=Element/id=] should return the same string, regardless of their [=node document].
 
 		Note: this means that a ''::part()'' pseudo-element wouldn't resolve to its matching element's [=Element/id=].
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -815,7 +815,8 @@ To resolve the [=used value=] of 'view-transition-name' for |element|:
 	1. If |computed| is <css>none</css>, return null.
 	1. If |computed| is a <<custom-ident>>, return |computed|.
 	1. Assert: |computed| is <css>auto</css> or <css>match-element</css>.
-	1. If |computed| is <css>auto</css>, |element| has an associated [=Element/id=], and |computed| is associated with the same [=tree/root=] as |element|'s [=tree/root=], then return the value of |element|'s [=Element/id=].
+	1. If |computed| is <css>auto</css>, |element| has an associated [=Element/id=], and |computed| is associated with the same [=tree/root=] as |element|'s [=tree/root=], then return a unique string.
+		Two elements with the same [=Element/id=] should match based on this string, regardless of their [=node document].
```
vmpstr, line 819:
```suggestion
		Two elements with the same [=Element/id=] should return the same string, regardless of their [=node document].
```

vmpstr (approved):

css-view-transitions-2/Overview.bs
domenic: Why should instead of must?

Issue #11232, opened 2024-11-18
[css-values-5] What does sibling-index()/sibling-count() return if there's no element?


sesse:
sibling-index() (and its sibling, sibling-count()) does not seem to have defined behavior if there's no element, e.g. in a media query. What should the behavior here be? (Returning 0 would be fine by us, if you just want to pick something arbitrary.)

lilles:
There is a question if it should parse at all. Some relevant cases are @media queries, @page properties, @font-face descriptors.

A similar case is anchor-size(). The spec handles anchor-size() as IACVT if not used in a sizing property, but whether anchor-size() should parse at all in a non-element context like the ones mentioned above is not clear to me.

Testing anchor-size() in Chrome turns out that it is valid at parse time for @page properties, but not for @font-face. Also, anchor-size() causes media queries to become general-enclosed.

Additionally, there is the case for @container queries where I think the obvious behavior is that sibling-index()/count() resolves against the container element like font relative units, which should probably have a clarifying comment in the spec for container queries.

cdoublev:
> The spec handles `anchor-size()` as IACVT if not used in a sizing property, but whether `anchor-size()` should parse at all in a non-element context like the ones mentioned above is not clear to me.

In my opinion, this is an error in the spec. `anchor()` and `anchor-size()` are explicitly defined as `New values` of the set of properties for which they are valid. What would be the reason to accept them in any property value just to make them IACVT in other properties?

Fwiw, #10679 is a duplicate of this issue that is extended to other numeric substitution functions, which could also be extended to arbitrary substitution functions (cf. https://github.com/w3c/csswg-drafts/issues/10679#issuecomment-2266364116).

lilles:
Right. Thanks!

PR #11931, opened 2025-03-13
[css-anchor-position-1][editorial] Clarify validity of anchor-*()

cdoublev:
This PR updates the *Validity* sections for [`anchor()`](https://drafts.csswg.org/css-anchor-position-1/#anchor-valid) and [`anchor-size()`](https://drafts.csswg.org/css-anchor-position-1/#anchor-size-valid), which could [confusingly](https://github.com/w3c/csswg-drafts/issues/11232#issuecomment-2720184752) be interpreted as making these functions invalid *at computed value time* when used in properties other than those for which they are explicitly defined as valid.

These functions are defined as `New values` of these properties. It would not make sense to accept them in all properties (like CSS-wide keywords, arbitrary or whole value substitutions) just to make them invalid at computed value time rather than at parse time.

### DIFF
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -1251,8 +1251,7 @@ An ''anchor()'' function is a
 <dfn lt="valid anchor function|invalid anchor function">valid anchor function</dfn>
 only if all the following conditions are true:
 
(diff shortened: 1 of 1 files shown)


cdoublev:
Fixed in dcd4fc2.

PR #11952, opened 2025-03-18
[css-images] Allow specifying any angle as a bare zero in gradients

### LINKED ISSUE #8346: [css-images] Define `<zero>` as an alternative for `<angle>`
According to [CSS Values 4](https://drafts.csswg.org/css-values-4/#angles), `0` is an invalid `<angle>` if not otherwise defined explicitly:

  > For legacy reasons, some uses of `<angle>` allow a bare `0` to mean `0deg`. This is not true in general, however, and will not occur in future uses of the `<angle>` type.

[`<conic-gradient()>`](https://drafts.csswg.org/css-images-4/#funcdef-conic-gradient) and [`<linear-gradient()>`](https://drafts.csswg.org/css-images-3/#funcdef-linear-gradient) define `0` in prose as a valid `<angle>` representing a gradient paremeter (origin or direction). But it is not defined valid in the [color stop list](https://drafts.csswg.org/css-images-4/#color-stop-syntax) of `<conic-gradient()>`, whereas it is valid in Chrome/FF.

If I am not mistaken, these would be the two remaining productions defining `0` as a valid `<angle>` in prose instead of in basic syntax. For example, [`<hue>`](https://drafts.csswg.org/css-color-4/#typedef-hue) in color functions accepts `<number>`, [transform functions](https://drafts.csswg.org/css-transforms-1/#two-d-transform-functions) accept `<zero>`, etc.

I suggest to define `<zero>` as an alternative of `<angle>` in `<conic-gradient()>` and `<linear-gradient()>`, which would allow to remove the above note in CSS Values.

```diff
  conic-gradient() = conic-gradient(
-   [ [ [ from <angle> ]? [ at <position> ]? ] || <color-interpolation-method> ]? ,
+   [ [ [ from [ <angle> | <zero> ] ]? [ at <position> ]? ] || <color-interpolation-method> ]? ,
    <angular-color-stop-list>
  )
```

```diff
  <linear-gradient()> = linear-gradient(
-   [ <angle> | to <side-or-corner> ]? ,
+   [ <angle> | <zero> | to <side-or-corner> ]? ,
    <color-stop-list>
  )
```

```diff
-  <angular-color-hint> = <angle-percentage>
-  <color-stop-angle> = <angle-percentage>{1,2}
+  <angular-color-hint> = <angle-percentage> | <zero>
+  <color-stop-angle> = [<angle-percentage> | <zero>]{1,2}
```

cdoublev:
Fixes #8346.

It explicitly allows declaring a conic/linear gradient rotation with `<zero>` in the grammar, which is already defined in prose.

It allows declaring an angular color stop/hint with `<zero>`, which matches the output in Chrome and Firefox but is currently undefined, so this change is substantive.

As noted in the issue, I believe that this change allows these lines to be removed from CSS V&U [3](https://github.com/w3c/csswg-drafts/blob/2e37f85c673df18fb79998c96124aab9d87e321f/css-values-3/Overview.bs#L1590-L1593) and [4](https://github.com/w3c/csswg-drafts/blob/2e37f85c673df18fb79998c96124aab9d87e321f/css-values-4/Overview.bs#L2635-L2638), because all uses `<angle>` with a bare `0` would now be explicitly allowed in the grammar:

  > Note: For legacy reasons,
  > some uses of `<<angle>>` allow a bare ''0'' to mean ''0deg''.
  > This is not true in general, however,
  > and will not occur in future uses of the `<<angle>>` type.

### DIFF
--- a/css-images-3/Overview.bs
+++ b/css-images-3/Overview.bs
@@ -334,7 +334,7 @@ Linear Gradients: the ''linear-gradient()'' notation {#linear-gradients}
 	<pre class=prod>
 		<<linear-gradient()>> = linear-gradient( [ <<linear-gradient-syntax>> ] )
 
-		<dfn>&lt;linear-gradient-syntax></dfn> = [ <<angle>> | to <<side-or-corner>> ]? , <<color-stop-list>>
+		<dfn>&lt;linear-gradient-syntax></dfn> = [ <<angle>> | <<zero>> | to <<side-or-corner>> ]? , <<color-stop-list>>
 
 		<dfn>&lt;side-or-corner></dfn> = [left | right] || [top | bottom]
 	</pre>
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -1271,7 +1271,7 @@ Linear Gradients: the ''linear-gradient()'' notation {#linear-gradients}
 
 	<pre class=prod>
 		<dfn>&lt;linear-gradient-syntax></dfn> =
-			[ [ <<angle>> | to <<side-or-corner>> ] || <<color-interpolation-method>> ]? ,
+			[ [ <<angle>> | <<zero>> | to <<side-or-corner>> ] || <<color-interpolation-method>> ]? ,
 			<<color-stop-list>>
 		<dfn>&lt;side-or-corner></dfn> = [left | right] || [top | bottom]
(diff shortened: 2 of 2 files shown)


svgeesus (approved):
Thanks, lgtm

PR #11955, opened 2025-03-18
[css-color-hdr-1] Remove duplicate productions

### LINKED ISSUE #11954: [css-color-hdr-1] Extend `<colorspace-params>` instead of `color()`
The [spec](https://drafts.csswg.org/css-color-hdr-1/#color-function) says:

  > In this specification the `color()` function is extended to allow predefined color spaces for HDR [...]

More precisely, it extends `<colorspace-params>` and `<predefined-rgb>` with new alternatives and redefines the other productions with no change.

Would it be possible to only define `<colorspace-params>` and `<predefined-rgb>`?

This would avoid conflicts from duplicate CSS type value definitions exposed by `w3c/webref`, for CSS tools consuming this data. Resolving them is not always straightforward, even between different levels of the same spec.

---

Another step forward could be a value definition syntax like `<extended> += new alternative` that `w3c/reffy` could interpret as "New values" for `<extended>`, akin to the "New values" field in property definition tables.

This would prevent extended definitions from diverging unexpectedly and eliminate the maintenance cost of adding new values ​​to a production extended by another production.

There are not many extended productions though, so it might not be worth bothering the maintainers of `w3c/reffy` for this.

cdoublev:
Fixes #11954.

### DIFF
--- a/css-color-hdr-1/Overview.bs
+++ b/css-color-hdr-1/Overview.bs
@@ -602,7 +602,8 @@ Mixing Dynamic Range Limits: the ''dynamic-range-limit-mix()'' function {#dynami
 	in a particular, given [=color space=]
 	(rather than the implicit sRGB color space that most of the other color functions operate in).
 
-	In this specification the ''color()'' function is extended
+	In this specification, 
+	the <<colorspace-params>> taken by the ''color()'' function is extended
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
Thanks!

Issue #11945, opened 2025-03-16
[css-sizing-4] Should the contain-intrinsic-* properties allow negative lengths?


weinig:
The current grammar for the [contain-intrinsic-* properties](https://drafts.csswg.org/css-sizing-4/#intrinsic-size-override) is `auto? [ none | <length> ]`, but WPT has tests that indicate a negative length is invalid, https://github.com/web-platform-tests/wpt/blob/7a9307ef9e1e88229ced4a29991270d20911669c/css/css-sizing/contain-intrinsic-size/parsing/contain-intrinsic-size-invalid.html#L22, which all browsers currently pass.

Should the grammar be updated to `auto? [ none | <length [0,∞]> ]`?

cc (@tabatkins, @fantasai, @jensimmons)

Loirooriol:
I guess negative lengths would be like having content with negative margins. But the element won't get a negative size anyways so it seems pointless. I would update to `<length [0,∞]>`, yes.

PR #11959, opened 2025-03-18
[css-sizing-4] contain-intrinsic-* properties should not allow negative lengths. See #11945

weinig:


Loirooriol (approved):
But I think this needs editor approval since it has behavior implications.

tabatkins:
> since it has behavior implications.

Only theoretically; I expect in practice impls were clamping it by 0 anyway. This looks like an easy bugfix to me.

Issue #10674, opened 2024-08-02
[css-fonts] Spec does not define how OS font settings affect the default font-size `medium`


JoshTumath:
## Context

Authors typically don't set an explicit absolute `font-size` on the root (e.g. `font-size: 32px`). They know that the default `font-size` is 16 CSS pixels by default, and so they size fonts relative to that initial `font-size`. For example, they might set a heading to be `font-size: 2rem`, expecting a computed value of 32px.

They do this because they have been taught it's best practice to use relative lengths rather than absolute lengths for accessibility reasons. They are are taught that doing so will respect when the user changes the default font size in their operating system or browser's settings.

## Problem

While `medium` does always seem to compute to `16px` by default, that's not actually explicitly defined in the spec. That seems to be the _de facto_ standard.

On Android devices, changing the font size in the operating system settings does affect pages' default font size in Firefox, but not Chrome.

On Safari on iOS, the operating system font size settings do not affect the default font size and there is no way of increasing the font size in the Safari app settings, AFAIK.

Apple also provide a set of vendor-prefixed values for the `font` property to set the font styles (including font-size), based on a feature they've branded 'dynamic type'. This is explained in an [article on system fonts on the WebKit blog.](https://webkit.org/blog/3709/using-the-system-font-in-web-content/)

```css
body {
  font: -apple-system-body;
  font-family: initial;
  font-weight: initial;
}
```

For more info on this practice, see: https://www.tpgi.com/text-resizing-web-pages-ios-using-dynamic-type/

At the BBC, we were surprised by this when noticing that web pages rendered in a Web View in iOS apps don't scale with the dynamic type feature by default.

It feels like CSS needs to be more explicit about how the user's font size accessibility settings should be respected. Users and authors may be surprised by how the default font size often isn't affected by the operating system settings.

## What the spec currently says

The initial value of the `font-size` property is `medium`. However, according to the specs, the `<absolute-size>` values (small, medium, large, etc) are defined by the user agent. The specs do not in any way prescribe what the size should be.

[CSS2 defines roughly what it should be:](https://drafts.csswg.org/css2/#valdef-font-size-medium)

> The following table provides user agent guidelines for the absolute-size mapping to HTML heading and absolute font-sizes. **The [medium](https://drafts.csswg.org/css2/#valdef-font-size-medium) value is the user’s preferred font size** and is used as the reference middle value.

[...]

Issue #11636, opened 2025-02-03
[css-size-adjust] Clarify interaction between text-size-adjust and em units


progers:
It is not clear how `text-size-adjust` and `em` units should interact ([spec](https://drafts.csswg.org/css-size-adjust/#adjustment-control)). Not adjusting em units leads to clipped text. Adjusting em units leads to double-adjustment of line-height.

Example ([live](https://balsam-juvenile-pufferfish.glitch.me)):
```
<!doctype html>
<meta name="viewport" content="width=device-width">
<style>
  html { -webkit-text-size-adjust: 250%; text-size-adjust: 250%; }
</style>

<p>
  height (em, px):
  <div style="height: 1em; width: 16px; background: lightblue; display: inline-block;">a</div>
  <div style="height: 16px; width: 16px; background: lightgreen; display: inline-block;">a</div>
</p>

<p>
  border-width (em, px):
  <div style="border-top: 1em solid blue; width: 16px; background: lightblue; display: inline-block;">a</div>
  <div style="border-top: 16px solid green; width: 16px; background: lightgreen; display: inline-block;">a</div>
</p>

<p>
  line-height (em, px):
  <div style="line-height: 1em; width: 16px; background: orange; display: inline-block;">a</div>
  <div style="line-height: 16px; width: 16px; background: purple; display: inline-block;">a</div>
</p>
```

Comparison between Webkit (left) and Chromium (right):
<img width="375" alt="Image" src="https://github.com/user-attachments/assets/6e3aa5b1-0e86-41e9-8851-55c245e2572e" />

tabatkins:
This seems more generally to be a question about *what `text-size-adjust` actually does*.

Based on both discussion and testing, it seems the behavior definitely doesn't match the spec, but also differs significantly between iOS and Android. In particular:

* In both Android and iOS, `text-size-adjust` changes the value of *both* `font-size` and `line-height`, independently. (The spec doesn't mention `line-height`, and this isn't just a consequence of `line-height: <number>` responding to the em size; see later points on the details.) 
* Also in both browsers, the adjustment *does not happen at computed-value time* (which the spec says it does). Instead, it happens somewhere around used-value time. This is very important, because applying it at computed-value time would cause issues; you'd get the adjustment compounding across descendents.
[...]

PR #11961, opened 2025-03-18
[css-env-1] Add a text-scale env() (name to be bikeshedded). #10674

tabatkins:


css-env-1/Overview.bs
```diff
@@ -227,6 +227,55 @@ Viewport units should be used instead when there is no hardware feature
...
+represents the scale factor
+of the user agent's operating system font scale.
+(Aka, the scale factor applied by ''text-size-adjust: auto''.)
+
+For example, if ''text-size-adjust:auto''
+would cause text sizes to double,
+then ''env(bikeshed-text-scale)'' would resolve to ''2''.
+
+Note: The ''bsem'' unit represents this same information;
+''1em'' is exactly equivalent to ''calc(1em * env(bikeshed-text-scale))''.
```
JoshTumath, line 254: That doesn't seem right to me. In https://github.com/w3c/csswg-drafts/issues/10674#issuecomment-2695942277, I defined it as _the computed value of `font-size: medium`_ × _`env(bikeshed-text-scale)`_

I don't think we discussed in the issue whether the unit could be relative to `1em` unit - the computed value of the selected element's `font-size`. My understanding was the computed value of `1bsem` would be the same regardless of what element it was used on.

JoshTumath (commented):
Thanks for sharing the initial draft!

css-env-1/Overview.bs
davidsgrogan: I think my understanding matches Josh's -- computed value of `1bsem` would be the same regardless of what element it was used on.

css-env-1/Overview.bs
```diff
@@ -227,6 +227,55 @@ Viewport units should be used instead when there is no hardware feature
...
+		<th>Number of dimensions
+	<tr>
+		<td><dfn>bikeshed-text-scale</dfn>
+		<td><<number>>
+		<td>0 (scalar)
+</table>
+
+The ''bikeshed-text-scale'' [=environment variable=]
+represents the scale factor
+of the user agent's operating system font scale.
```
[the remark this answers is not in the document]
tabatkins: None of this should be precluded; the source of the information is outside of CSS's scope in the first place. I just needed a rough definition of what it was, so it actually made sense. I've adjusted the wording a bit to make this a little looser. (You approved it while sitting next to me, so that's a +1 from you. ^_^)

css-env-1/Overview.bs
tabatkins: I was under the impression that the point of the unit was to allow authors to avoid having to manually type `calc(... * env(preferred-text-scale))` all over the place; it was supposed to be syntax sugar for that expression, essentially. Thus, it should scale *with* the local definition of the em; otherwise, authors can just set a variable at the root and use that instead, and we don't care quite as much about the verbosity in that case.

That said, putting some thought into this revealed that the definition of the unit is going to be very problematically tied into the definition of text-size-adjust itself, which is currently both underspecified (it doesn't actually say enough about its effects) and wrongly specified (it claims to be a computed-time adjustment, which testing trivially shows it isn't).

Luckily, the unit isn't meant to be defined by this spec; it goes over into V&U. So I'll just leave the references here for now, and deal with the definition of `pem` elsewhere when we resolve the `text-size-adjust` issues more generally.

tabatkins: See #11636 for discussion on this.

PR #11963, opened 2025-03-19
[css-forms-1] Include textarea in list of elements.

dbaron:
This fixes three cases where `<textarea>` was omitted from the list of elements that `appearance: base` can be applied to.

I think `textarea` should be included in this list, unless I'm missing something.


lukewarlow (approved):
LGTM

nt1m (approved):

PR #11968, opened 2025-03-19
[css-view-transitions-1] Minor edits

### LINKED ISSUE #11721: [css-view-transitions-1] The activate view transition algorithm does not specify a document
The [activate view transition algorithm](https://drafts.csswg.org/css-view-transitions-1/#activate-view-transition), in step 2, says to "set [rendering suppression for view transitions](https://drafts.csswg.org/css-view-transitions-1/#document-rendering-suppression-for-view-transitions) to false".

[rendering suppression for view transitions](https://drafts.csswg.org/css-view-transitions-1/#document-rendering-suppression-for-view-transitions) is a property of Document but the algorithm was neither given a document nor does it say which document to use.
This seems like an oversight and I assume that there should be a step similar to step 1 in [setup view transition](https://drafts.csswg.org/css-view-transitions-1/#setup-view-transition) that mentions which document this refers to.

Noticed this while implementing the feature into [Ladybird](https://github.com/LadybirdBrowser/ladybird).

### LINKED ISSUE #11723: [css-view-transitions-1] The activate view transition algorithm does not provide a reason to skip the view transition.
The [activate view transition algorithm](https://drafts.csswg.org/css-view-transitions-1/#activate-view-transition), in step 3, says to "then [skip the view transition](https://drafts.csswg.org/css-view-transitions-1/#skip-the-view-transition) for transition, and return".

[Skip the view transition](https://drafts.csswg.org/css-view-transitions-1/#skip-the-view-transition) expects a reason, which generally seems to be various kinds of `DOMException`s.
This should probably pass one of those.

Noticed this while implementing the feature into [Ladybird](https://github.com/LadybirdBrowser/ladybird).

### LINKED ISSUE #11725: [css-view-transitions-1] The update pseudo-element styles algorithm treats captured elements as if they were actual elements
The [update pseudo-element styles algorithm](https://drafts.csswg.org/css-view-transitions-1/#update-pseudo-element-styles) says:
> Let *newRect* be [snapshot containing block](https://drafts.csswg.org/css-view-transitions-1/#snapshot-containing-block) if *capturedElement* is the [document element](https://dom.spec.whatwg.org/#document-element), otherwise, *capturedElement*’s [border box](https://drafts.csswg.org/css-box-4/#border-box).

Missing "the" before "snapshot containing block" aside, *capturedElement* is a [captured element](https://drafts.csswg.org/css-view-transitions-1/#captured-element) from the ViewTransition's [named elements](https://drafts.csswg.org/css-view-transitions-1/#viewtransition-named-elements).
That struct can never be the document element and also does not have a border box.

I think this probably intends to use *capturedElement*'s [new element](https://drafts.csswg.org/css-view-transitions-1/#captured-element-new-element), but it should say so if that is the case.

Noticed this while implementing the feature into [Ladybird](https://github.com/LadybirdBrowser/ladybird).

noamr:
- Specify which documents is used for rendering supression
- Specify reason for skipping transition when capture fails
- Specify the correct document element in new capture

Closes #11721
Closes #11723
Closes #11725

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -1291,14 +1291,15 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 
 			Note: This happens if |transition| was [=skip the view transition|skipped=] before this point.
 
-		1. Set [=document/rendering suppression for view transitions=] to false.
+		1. Set |transition|'s [=relevant global object's=] [=associated document=]'s [=document/rendering suppression for view transitions=] to false.
 
 		1. If |transition|'s [=ViewTransition/initial snapshot containing block size=] is not equal to the [=snapshot containing block size=],
-			then [=skip the view transition=] for |transition|, and return.
+			then [=skip the view transition|skip=] |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
+			and return.
 
 		1. [=Capture the new state=] for |transition|.
 
-			If failure is returned, then [=skip the view transition=] for |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
+			If failure is returned, then [=skip the view transition|skip=] |transition| with an "{{InvalidStateError}}" {{DOMException}} in |transition|'s [=relevant Realm=],
 			and return.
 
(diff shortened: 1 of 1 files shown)


noamr:
Thanks @Psychpsyo for pointing these out!

Psychpsyo:
LGTM

Issue #11509, opened 2025-01-15
Proposal: CSS @sheet


KurtCattiSchmidt:
[Link to explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/AtSheet/explainer.md)

CSS `@sheet` is a mechanism for bundling multiple CSS stylesheets into one .css file. There are numerous practical advantages to bundling multiple CSS sheets into one file, including reduced network requests and potentially increased compression ratios.

`@sheet` is an at-block with an identifier:

style.css:
```
@sheet sheet1 {
 ...
}
```

Sheets defined via `@sheet` can be referenced via fragment identifiers in URL's that match the `@sheet` identifier:

`<link rel="stylesheet" href="style.css#sheet1">`

This was discussed in the [CSSWG call on 2023-04-05](https://lists.w3.org/Archives/Public/www-style/2023Apr/0004.html) as well as [this discussion](https://github.com/w3c/csswg-drafts/issues/5629).

We propose expanding this definition to include `@sheet` definitions at the document level, which can similarly be referenced via a local fragment:

```
<style>
@sheet sheet1 {
 ...
}
</style>
<link rel="stylesheet" href="#sheet1">
```

Much like anchor fragments, Shadow DOM nodes can access `@sheet` definitions from their parent scope:

```
<style>
@sheet foo {
  div {
    color: red;
  }
}
</style>
<template shadowrootmode="open">
  <link rel="stylesheet" href="#foo" />
  <span>I'm in the shadow DOM</span>
</template>
```

More details and examples can be found in the explainer link above. I would welcome any comments/suggestions/additional use cases, either in this thread or as issues in the MSEdgeExplainers repo: [New issue](https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/new?assignees=kschmi&labels=AtSheet&title=CSS+At+Sheet+%3CTITLE+HERE%3E)

jyasskin:
[...]

Issue #10013, opened 2024-02-28
[css-om] Can we lift the restriction on constructed flag for adoptedStylesheets?


keithamus:
Currently `adoptedStyleSheets` has a restriction on `CSSStyleSheets` that have the `constructed` flag set:

> The [set an indexed value](https://webidl.spec.whatwg.org/#observable-array-attribute-set-an-indexed-value) algorithm for adoptedStyleSheets, given value and index, is the following:
>
> 1. If value’s [constructed flag](https://drafts.csswg.org/cssom/#concept-css-style-sheet-constructed-flag) is not set, or its [constructor document](https://drafts.csswg.org/cssom/#concept-css-style-sheet-constructor-document) is not equal to this [DocumentOrShadowRoot](https://dom.spec.whatwg.org/#documentorshadowroot)'s [node document](https://dom.spec.whatwg.org/#concept-node-document), throw a "[NotAllowedError](https://webidl.spec.whatwg.org/#notallowederror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

I'm here to rattle this Chesterton's fence. It would be _really_ useful to be able to pluck a documents stylesheets and adopt it into a shadowroots `adoptedStyleSheets`, co-opting the pages styles while applying the shadow's own styles. Currently this is possible with a rather unfortunate amount of JS by duplicating a `<link>` tag and stuffing it in your shadow-dom, but it feels like we can make this easier?

(Cue discussion about `@import` and media queries...)

mayank99:
Some assorted thoughts:

The primary use case cited is for adopting `document.styleSheets`, but this will also enable adopting stylesheets from a parent shadow-root (which is equally useful!)

```js
const parentRoot = el.shadowRoot.host.getRootNode();
el.shadowRoot.adoptedStyleSheets.push(
  ...parentRoot.styleSheets,
  ...parentRoot.adoptedStyleSheets,
);
```

---

I think there is (or should be) a difference between "constructed" stylesheets and "adoptable" stylesheets.

The `@import` restriction could continue to apply to *constructed* stylesheets, but does not need to apply to other adoptable stylesheets.

---

Should there be an easier way to feature detect?

Currently I can see a `try`/`catch` way of polyfilling.

```js
function requiresConstructing() {
  try {
    const doc = document.implementation.createHTMLDocument();
    doc.body.appendChild(doc.createElement("style");
    doc.adoptedStyleSheets = doc.styleSheets;
    return false;
[...]

PR #11980, opened 2025-03-21
[css-cascade-7] @sheet proposal #11509 

KurtCattiSchmidt:
Initial version, still needs lots of work

### DIFF
--- a/css-cascade-7/Overview.bs
+++ b/css-cascade-7/Overview.bs
@@ -1,0 +1,319 @@
+<pre class='metadata'>
+Title: CSS Cascading and Inheritance Level 7
+Shortname: css-cascade
+Level: 7
+Status: ED
+Prepare for TR: no
+Work Status: Revising
+Group: csswg
+ED: https://drafts.csswg.org/css-cascade-7/
+Previous Version: https://www.w3.org/TR/2021/WD-css-cascade-6-20211221/
+Editor: Elika J. Etemad / fantasai, Apple, http://fantasai.inkedblade.net/contact, w3cid 35400
+Editor: Miriam E. Suzanne, Invited Expert, http://miriamsuzanne.com/contact, w3cid 117151
+Editor: Tab Atkins Jr., Google, http://xanthir.com/contact/, w3cid 42199
+Editor: Kurt Catti-Schmidt, Microsoft, kschmi@microsoft.com, w3cid 146156
+Abstract: This CSS module describes how to collate style rules and assign values to all properties on all elements. By way of cascading and inheritance, values are propagated for all properties on all elements.
+Abstract:
+Abstract: New in this level is [[#at-ruledef-sheet]].
+Ignored Terms: auto, flex items, <supports-condition>, <declaration>, scope proximity, encapsulation context
+Informative Classes: ex
+</pre>
+
+<pre class=link-defaults>
+spec:dom; type:dfn;
+	text:shadow tree
+	for:tree; text:root
+	text:shadow root; for:/
+spec:dom; type:dfn; text:parent element
+spec:css-color-4; type:property; text:color
+spec:css-values-3; type: value; text:ex
+spec:css-conditional-3; type:at-rule; text:@media
+spec:mediaqueries-4; type:type; for:@media; text:all
+spec:mediaqueries-4; type:type; text:<media-query>
+spec:selectors-4; type:dfn; text:subject
+spec:selectors-4; type:dfn; text:selector
+spec:selectors-4; type:dfn; text:combinator
+spec:html; type:element; text:style
+spec:css-scoping-1; type:dfn; text:shadow host
+</pre>
+
+<pre class=ignored-specs>
+spec:mediaqueries-5
+spec:css-values-4
+spec:css-fonts-4
+</pre>
+
+<h2 id="intro" oldids='filtering,fragments,stages-examples,actual,used,computed,cascaded,declared,specified,value-stages,all-shorthand,aliasing,shorthand,content-type,import-processing,conditional-import,at-import,defaulting,initial-values,inheriting,defaulting-keywords,initial,inherit,inherit-initial,default'>
+Introduction and Missing Sections</h2>
+
+	Issue: This is a diff spec over <a href="https://www.w3.org/TR/css-cascade-6/">CSS Cascading and Inheritance Level 6</a>.
+	It is currently an Exploratory Working Draft:
+	if you are implementing anything, please use Level 6 as a reference.
+	We will merge the Level 6 text into this draft once it reaches CR.
+
+<!-- Big Text: Stylesheet 
+
+ ███▌  █████▌ █   ▐▌ █▌    █████▌  ███▌  █▌  █▌ █████▌ █████▌ █████▌
+█▌  █▌   █▌   ▐▌  █  █▌    █▌     █▌  █▌ █▌  █▌ █▌     █▌       █▌  
+█▌       █▌    █ ▐▌  █▌    █▌     █▌     █▌  █▌ █▌     █▌       █▌  
+ ███▌    █▌    ▐▌█   █▌    ████    ███▌  █████▌ ████   ████     █▌  
+    █▌   █▌     █▌   █▌    █▌         █▌ █▌  █▌ █▌     █▌       █▌  
+█▌  █▌   █▌     █▌   █▌    █▌     █▌  █▌ █▌  █▌ █▌     █▌       █▌  
+ ███▌    █▌     █▌   █████ █████▌  ███▌  █▌  █▌ █████▌ █████▌   █▌  
+-->
+
+<h2 id='stylesheet-section'>
+Stylesheet</h2>
+
+	A <dfn export>stylesheet</dfn>
+	is a collection of style rules. 
+	
+	These collections may exist as:
+	<ol>
+		<li>An instance of file with a .css extension and/or MIME type of "text/css" embedded within a document.
+		<li>Collection of CSS rules within a document.
+		<li>CSSOM Stylesheet objects.
+		<li>Any of the above may be subdivided into named stylesheets with the ''@sheet'' keyword.
+	</ol>
+
+	A <dfn>sheet-name</dfn> is a string identifying
+	 inner named sheets within a stylesheet.
+
+	Syntactically, a stylesheet may include an explicit [=sheet-name=], and is represented
+	by the <<sheet-name>> in ''@sheet'' and ''@import'' rules,
+	which is a single <<ident>> token:
+
+	<pre class='prod'>
+		<dfn><<sheet-name>></dfn> = <<ident>>
+	</pre>
+
+<h3 id="at-ruledef-sheet">
+Named Stylesheets: the `@sheet` block at-rule</h3>
+
+	The ''@sheet'' [=block at-rule=]
+	assigns its child style rules to a particular named [=stylesheet=].
+	This sheet-assignment syntax is:
+
+	<pre class='prod'>
+		@sheet <<sheet-name>> {
+			<<rule-list>>
+		}
+	</pre>
+
+	Such ''@sheet'' block rules have the same restrictions and processing
+	as a [=conditional group rule=] [[CSS-CONDITIONAL-3]]
+	with a true condition.
+
+	<div class=example>
+		For example, ''@sheet'' and ''@media'' can be mixed:
+
+		<pre class=lang-css>
+		@sheet framework {
+			h1, h2 { color: maroon; background: white;}
+
+			@media (prefers-color-scheme: dark) {
+				h1, h2 { color: red; background: black; }
+			}
+		}
+		</pre>
+	</div>
+
+	Note: ''@sheet'' [=block at-rules=] cannot be interleaved with ''@import'' rules.
+
+<h3 id="sheet-declaration">
+Declaring Named Stylesheets</h3>
+
+	Named stylesheets can be declared:
+
+	* using an ''@import'' rule with the ''from'' keyword
+		assigning the contents of the imported file into that sheet.
+	* using an [[#at-ruledef-sheet|@sheet block at-rule]],
+		assigning its child style rules into that sheet.
+
+<h3 id="at-ruledef-sheet-block">
+Subdividing Stylesheets: the `@sheet` block at-rule</h3>
+
+<pre class='prod'>
+		@import [ <<string>> from ]?
+				[ <<url>> | <<string>> ]				
+		        [ layer | layer(<<layer-name>>) ]?
+		        <<import-conditions>> ;
+
+		<dfn export>&lt;import-conditions></dfn> = [ supports( [ <<supports-condition>> | <<declaration>> ] ) ]?
+		                     <<media-query-list>>?</pre>
+
+	where:
+	* <<string>> ''from'' imports only the named sheet
+		with name <<string>> from the provided URL.
+
+	* the <<url>> or <<string>>
+		gives the URL of the style sheet to be imported.
+
+	* the optional ''layer'' keyword or ''layer()'' function
+		assigns the contents of the style sheet
+		into its own anonymous [=cascade layer=]
+		or into the named [=cascade layer=].
+
+		The layer is added to the layer order
+		even if the import fails to load the stylesheet,
+		but is subject to any [=import conditions=]
+		(just as if declared by an ''@layer'' rule wrapped
+		in the appropriate [=conditional group rules=]).
+
+	* the optional <<import-conditions>>
+		states the [=import conditions=] under which it applies.
+
+	<div class="example">
+		The following <a href="#conditional-import">conditional <css>@import</css> rule</a>
+		only loads the style sheet with a name of ``foo`` from ``narrow.css`` when the UA
+		<a href="https://www.w3.org/TR/css-conditional-3/#support-definition">supports</a> ''display: flex'',
+		and only applies the style sheet on a <a href="https://www.w3.org/TR/CSS2/media.html#media-types">handheld</a> device
+		with a <a href="https://www.w3.org/TR/mediaqueries-4/#width">maximum viewport width</a> of 400px.
+
+		<pre>@import foo from url("narrow.css") supports(display: flex) handheld and (max-width: 400px);</pre>
+	</div>
+
+	<div class="example">
+		The following layer imports load the style sheets named ``foo`` into
+		the ''framework.component'' layer, and an un-named layer, respectively:
+
+		<pre>
+		@import foo from url("tabs.css") layer(framework.component);
+		@import foo from url("override.css") layer;
+		</pre>
+	</div>
+
+	If a <<string>> is provided,
+	it must be interpreted as a <<url>> with the same value.
+
+	<div class="example">
+		The following lines are equivalent in meaning
+		and illustrate both ''@import'' syntaxes
+		(one with ''url()'' and one with a bare string):
+
+		<pre class='lang-css'>
+		@import "mystyle.css";
+		@import url("mystyle.css");
+		</pre>
+	</div>
+
+<!--
+████████ ████████  ██████
+██          ██    ██    ██
+██          ██    ██
+██████      ██    ██
+██          ██    ██
+██          ██    ██    ██
+████████    ██     ██████
+-->
+
+<h2 id="changes">
+Changes</h2>
+
+This appendix is <em>informative</em>.
+
+
+<h3 id="additions-l6">
+Additions Since Level 6</h3>
+
+	The following features have been added since
+	<a href="https://www.w3.org/TR/css-cascade-6/">Level 6</a>:
+
+	* The definition of a [=stylesheet=],
+		as described by a combination of <<scope-start>> and <<scope-end>> selectors.
+	* The ''@sheet'' rule for subdividing stylesheets
+
+<h3 id="additions-l5">
+Additions Since Level 5</h3>
+
+	The following features have been added since
+	<a href="https://www.w3.org/TR/css-cascade-5/">Level 5</a>:
+
+	* The definition of a [=scope=],
+		as described by a combination of <<scope-start>> and <<scope-end>> selectors.
+	* The in-scope ('':in()'') pseudo-class for selecting with lower-boundaries
+	* The ''@scope'' rule for creating scoped stylesheets
+	* The definition of [=scope proximity=] in the cascade
+
+
+<h3 id="additions-l4">
+Additions Since Level 4</h3>
+
+	The following features have been added since
+	<a href="https://www.w3.org/TR/css-cascade-4/">Level 4</a>:
+
+	* Added [=cascade layers=] to the [=cascade=] sort criteria
+		(and defined style attributes as a distinct step of the [=cascade=] sort criteria
+		so that they interact appropriately).
+	* Introduced the ''@layer'' rule for defining cascade layers.
+	* Added ''layer''/''layer()'' option to ''@import'' definition.
+	* Introduced the ''revert-layer'' keyword for rolling back values to previous layers.
+
+<h3 id="additions-l3">
+Additions Since Level 3</h3>
+
+	The following features have been added since
+	<a href="https://www.w3.org/TR/css-cascade-3/">Level 3</a>:
+
+	* Introduced ''revert'' keyword, for rolling back the cascade.
+	* Introduced ''supports()'' syntax for supports-conditional ''@import'' rules.
+	* Added [=encapsulation context=] to the [=cascade=] sort criteria
+		to accommodate Shadow DOM. [[DOM]]
+	* Defined the property two aliasing mechanisms CSS uses to support legacy syntaxes. See [[css-cascade-4#aliasing]].
+<!--
+	* Added definition of how scoped styles would cascade
+		(deferred from Level 3)
+-->
+
+<h3 id="changes-2">
+Additions Since Level 2</h3>
+
+	The following features have been added since
+	<a href="http://www.w3.org/TR/CSS2/cascade.html">Level 2</a>:
+
+	<ul>
+		<li>The 'all' shorthand
+		<li>The ''initial'' keyword
+		<li>The ''unset'' keyword
+		<li>Incorporation of animations and transitions into the <a>cascade</a>.
+	</ul>
+
+<h2 class="no-num" id="acknowledgments">Acknowledgments</h2>
+
+	David Baron,
+	Tantek Çelik,
+	Keith Grant,
+	Giuseppe Gurgone,
+	Theresa O'Connor,
+	Florian Rivoal,
+	Noam Rosenthal,
+	Simon Sapin,
+	Jen Simmons,
+	Nicole Sullivan,
+	Lea Verou,
+	and Boris Zbarsky
+	contributed to this specification.
+
+<h2 id="privacy">
+Privacy Considerations</h2>
+
+	* User preferences and UA defaults expressed via application of style rules
+		are exposed by the cascade process,
+		and can be inferred from the computed styles they apply to a document.
+
+<h2 id="security">
+Security Considerations</h2>
+
+	* The cascade process does not distinguish between same-origin and cross-origin stylesheets,
+		enabling the content of cross-origin stylesheets to be inferred
+		from the computed styles they apply to a document.
+
+	* The ''@import'' rule does not apply the [=CORS protocol=] to loading cross-origin stylesheets,
+		instead allowing them to be freely imported and applied.
+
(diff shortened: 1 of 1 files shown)


css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+	A <dfn export>stylesheet</dfn>
+	is a collection of style rules. 
+	
+	These collections may exist as:
+	<ol>
+		<li>An instance of file with a .css extension and/or MIME type of "text/css" embedded within a document.
+		<li>Collection of CSS rules within a document.
+		<li>CSSOM Stylesheet objects.
+		<li>Any of the above may be subdivided into named stylesheets with the ''@sheet'' keyword.
```
kbabbitt, line 76:
```suggestion
		<li>Any of the above may be subdivided into named stylesheets with the ''@sheet'' rule.
```

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+	A <dfn>sheet-name</dfn> is a string identifying
+	 inner named sheets within a stylesheet.
+
+	Syntactically, a stylesheet may include an explicit [=sheet-name=], and is represented
+	by the <<sheet-name>> in ''@sheet'' and ''@import'' rules,
+	which is a single <<ident>> token:
+
+	<pre class='prod'>
+		<dfn><<sheet-name>></dfn> = <<ident>>
```
kbabbitt, line 87:
```suggestion
		<dfn><<sheet-name>></dfn> = <<custom-ident>> | <<string>>
```

For consistency with [`@keyframes`](https://drafts.csswg.org/css-animations-1/#keyframes)

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+			h1, h2 { color: maroon; background: white;}
+
+			@media (prefers-color-scheme: dark) {
+				h1, h2 { color: red; background: black; }
+			}
+		}
+		</pre>
+	</div>
+
+	Note: ''@sheet'' [=block at-rules=] cannot be interleaved with ''@import'' rules.
```
kbabbitt, line 121: "Note" blocks are considered informative, rather than normative text. I think this line is something we want as a normative statement. I might also include examples of valid and non-valid combinations of `@sheet` and `@import`.

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+		<dfn export>&lt;import-conditions></dfn> = [ supports( [ <<supports-condition>> | <<declaration>> ] ) ]?
+		                     <<media-query-list>>?</pre>
+
+	where:
+	* <<string>> ''from'' imports only the named sheet
+		with name <<string>> from the provided URL.
+
+	* the <<url>> or <<string>>
+		gives the URL of the style sheet to be imported.
```
kbabbitt, line 150:
```suggestion
		gives the URL of the [=stylesheet=] to be imported.
```

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+	* the optional ''layer'' keyword or ''layer()'' function
+		assigns the contents of the style sheet
```
kbabbitt, line 153:
```suggestion
		assigns the contents of the [=stylesheet=]
```

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+	<pre class='prod'>
+		@sheet <<sheet-name>> {
+			<<rule-list>>
+		}
+	</pre>
+
+	Such ''@sheet'' block rules have the same restrictions and processing
+	as a [=conditional group rule=] [[CSS-CONDITIONAL-3]]
+	with a true condition.
```
kbabbitt, line 105: I'm unclear on why this sentence is in the spec - can you explain what it's intended to do beyond what's implied by the grammar? (Explanatory text of this sort would be a good use case for a "Note".)

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+	<div class=example>
+		For example, ''@sheet'' and ''@media'' can be mixed:
```
kbabbitt, line 108:
```suggestion
		For example, an ''@sheet'' block can contain an ''@media'' rule:
```

I'd phrase it like this because I don't think nesting in the other direction is valid (or is it?)

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+		}
+		</pre>
+	</div>
+
+	Note: ''@sheet'' [=block at-rules=] cannot be interleaved with ''@import'' rules.
+
+<h3 id="sheet-declaration">
+Declaring Named Stylesheets</h3>
+
+	Named stylesheets can be declared:
```
kbabbitt, line 126: Here I might do:

```suggestion
	A <dfn>named stylesheet</dfn> can be declared:
```

So that you can autolink "named stylesheet" elsewhere in the document, for example in the text for `@import <string> from`.

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+		        <<import-conditions>> ;
+
+		<dfn export>&lt;import-conditions></dfn> = [ supports( [ <<supports-condition>> | <<declaration>> ] ) ]?
+		                     <<media-query-list>>?</pre>
+
+	where:
+	* <<string>> ''from'' imports only the named sheet
+		with name <<string>> from the provided URL.
+
+	* the <<url>> or <<string>>
```
kbabbitt, line 149: I might suggest defining an additional term above:

```
<url-string> = [ <url> | <string> ]
```

And then using that in the grammar for `@import` above, and on this line. The reason would be to reduce confusion between the `<string>` in the "from" case, and the `<string>` used for the import source.

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+		the ''framework.component'' layer, and an un-named layer, respectively:
+
+		<pre>
+		@import foo from url("tabs.css") layer(framework.component);
+		@import foo from url("override.css") layer;
+		</pre>
+	</div>
+
+	If a <<string>> is provided,
+	it must be interpreted as a <<url>> with the same value.
```
kbabbitt, line 187: Related to my comment above about `<url-string>` - this text should also be associated with that definition, since it only applies there and not to the `<string>` used with "from".

kbabbitt: I might also bring the example below into the same context for `<url-string>`.

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+
+<h3 id="additions-l6">
+Additions Since Level 6</h3>
+
+	The following features have been added since
+	<a href="https://www.w3.org/TR/css-cascade-6/">Level 6</a>:
+
+	* The definition of a [=stylesheet=],
+		as described by a combination of <<scope-start>> and <<scope-end>> selectors.
```
kbabbitt, line 223: This line doesn't seem right?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+
+<h2 id='stylesheet-section'>
+Stylesheet</h2>
+
+	A <dfn export>stylesheet</dfn>
+	is a collection of style rules. 
+	
+	These collections may exist as:
+	<ol>
+		<li>An instance of file with a .css extension and/or MIME type of "text/css" embedded within a document.
```
alisonmaher, line 73: nit: "and/or a MIME"

Does MIME type have a link? Although I see at https://www.w3.org/TR/css-cascade-6/#security we reverence it just as `text/css` without mentioning the "MIME type".

If we just say it has to be `text/css` will that imply both .css files and `text/css` embedded in a document?

In either case, we'll probably want to style "text/css" similarly to the reference in https://www.w3.org/TR/css-cascade-6/#security

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+		<li>Collection of CSS rules within a document.
```
alisonmaher, line 74: "A collection"

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+		<li>CSSOM Stylesheet objects.
```
alisonmaher, line 75: Does this have any link to reference?

css-cascade-7/Overview.bs
alisonmaher: If we do this, we should probably reference it in earlier sections. Example: https://kurtcattischmidt.github.io/kurtspublishedw3cdrafts/AtSheet.html#stylesheet-section

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+	</div>
+
+	Note: ''@sheet'' [=block at-rules=] cannot be interleaved with ''@import'' rules.
+
+<h3 id="sheet-declaration">
+Declaring Named Stylesheets</h3>
+
+	Named stylesheets can be declared:
+
+	* using an ''@import'' rule with the ''from'' keyword
```
alisonmaher, line 128: "from" isn't picking up any link

alisonmaher: Same in a couple other spots, too

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+				[ <<url>> | <<string>> ]				
+		        [ layer | layer(<<layer-name>>) ]?
+		        <<import-conditions>> ;
+
+		<dfn export>&lt;import-conditions></dfn> = [ supports( [ <<supports-condition>> | <<declaration>> ] ) ]?
+		                     <<media-query-list>>?</pre>
+
+	where:
+	* <<string>> ''from'' imports only the named sheet
+		with name <<string>> from the provided URL.
```
alisonmaher, line 147: For some reason the second <<string>> isn't picking up a link. Maybe that is a bikeshed choice if it's in the same sentence?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+	* the <<url>> or <<string>>
+		gives the URL of the style sheet to be imported.
+
+	* the optional ''layer'' keyword or ''layer()'' function
+		assigns the contents of the style sheet
+		into its own anonymous [=cascade layer=]
+		or into the named [=cascade layer=].
+
+		The layer is added to the layer order
+		even if the import fails to load the stylesheet,
```
alisonmaher, line 158: nit: comma at the end not needed

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,319 @@
...
+		but is subject to any [=import conditions=]
+		(just as if declared by an ''@layer'' rule wrapped
+		in the appropriate [=conditional group rules=]).
+
+	* the optional <<import-conditions>>
+		states the [=import conditions=] under which it applies.
+
+	<div class="example">
+		The following <a href="#conditional-import">conditional <css>@import</css> rule</a>
+		only loads the style sheet with a name of ``foo`` from ``narrow.css`` when the UA
```
alisonmaher, line 168: foo and narrow.css aren't showing up as code in https://kurtcattischmidt.github.io/kurtspublishedw3cdrafts/AtSheet.html#example-00f8cdb8

alisonmaher: Same with the other example

css-cascade-7/Overview.bs
KurtCattiSchmidt: This is interesting - `@layer` doesn't allow the `<<string>>` version, but `@keyframes` does. I filed an issue with the CSSWG for this: https://github.com/w3c/csswg-drafts/issues/12005

css-cascade-7/Overview.bs
KurtCattiSchmidt: Good call out. This essentially means "the rules within this block apply", which is what you'd want for `@layer` (which I copied it from), but crucially *not* what we want for `@sheet`. 

I switched the 'true condition' to 'false condition' and added an example.

css-cascade-7/Overview.bs
KurtCattiSchmidt: I removed this example with a better one

css-cascade-7/Overview.bs
KurtCattiSchmidt: This came from https://www.w3.org/TR/css-cascade-5/#at-import, so I probably don't want to change it

css-cascade-7/Overview.bs
KurtCattiSchmidt: I was able to avoid this by linking directly to https://www.w3.org/TR/css-cascade-5/#content-type, as I see https://www.w3.org/TR/css-cascade-5/#content-type does this. I think this is better, because it has an explicit paragraph description so I don't have to redefine it.

css-cascade-7/Overview.bs
KurtCattiSchmidt: This came from the `@layer` spec https://www.w3.org/TR/css-cascade-5/#layer-block, but perhaps that shouldn't be a 
"Note" either. For now, I've removed this statement entirely.

css-cascade-7/Overview.bs
KurtCattiSchmidt: Great suggestion, done!

css-cascade-7/Overview.bs
KurtCattiSchmidt: This wasn't actually correct, so I removed it

css-cascade-7/Overview.bs
KurtCattiSchmidt: Should be good now

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+
+	A <dfn export>stylesheet</dfn>
+	is a collection of style rules. 
+	
+	These collections may exist as:
+	<ol>
+		<li>An instance of file with the <a href="https://www.w3.org/TR/css-cascade-5/#content-type">CSS Style Sheet Content-Type</a>, embedded within a document.
+		<li>A collection of CSS rules within a document.
+		<li><a href="https://www.w3.org/TR/cssom/#the-stylesheet-interface">CSSOM Stylesheet</a> objects.
+		<li>Any of the above may be subdivided into <dfn>named stylesheet</dfn> with the ''@sheet'' rule.
```
alisonmaher, line 76: "a <dfn>named stylesheet</dfn>" (not sure if that is accurate, feel free to ignore if not

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+	</ol>
+
+	A <dfn>sheet-name</dfn> is a string identifying
+	 inner named sheets within a stylesheet.
+
+	Syntactically, a stylesheet must include an explicit [=sheet-name=], and is represented
```
alisonmaher, line 82: Is this really a "must" for cases outside of @sheet?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+		@import [ <<sheet-name-list>> from ]?
+				    [ <<url-string>> ]				
+		        [ layer | layer(<<layer-name>>) ]?
+		        <<import-conditions>> ;
+
+		<dfn export>&lt;import-conditions></dfn> = [ supports( [ <<supports-condition>> | <<declaration>> ] ) ]?
+		                     <<media-query-list>>?</pre>
+
+	where:
+	* the optional <<sheet-name-list>> imports only the named sheet(s)
```
alisonmaher, line 148: Should "named sheet" link to above?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+		The following layer imports the [=named stylesheet=]s named "foo" and "bar" from "tabs.css" into
+		the ''framework.component'' layer, and an un-named layer, respectively:
+
+		<pre>
+		@import foo, bar from url("tabs.css") layer(framework.component);
+		@import foo, bar from url("override.css") layer;
+		</pre>
+	</div>
+
+	If a <url-string> is provided,
```
alisonmaher, line 207: <url-string> isn't showing up in https://kurtcattischmidt.github.io/kurtspublishedw3cdrafts/AtSheet.html#at-ruledef-import-from. Not sure if that is just because it's an older version?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+
+	<pre class='idl' export>
+	[Exposed=Window]
+	interface CSSSheetBlockRule : CSSGroupingRule {
+	  readonly attribute CSSOMString name;
+	};
+	</pre>
+
+	Its <dfn attribute for=CSSSheetBlockRule>name</dfn> attribute represents
+	the <<sheet-name>> declared by the at-rule itself.
```
alisonmaher, line 247: Can at-rule be linked in any way?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+
+	<pre class='idl' export>
+	partial interface CSSImportRule {
+	  [SameObject] readonly attribute SheetNameList sheetNames;
+	  [SameObject] readonly attribute StyleSheetList namedStyleSheets;
+	};
+	</pre>
+
+	Its <dfn attribute for=CSSImportRule>sheetNames</dfn> attribute represents
+	a list of <<sheet-name>> definitions declared by the at-rule itself in the order
```
alisonmaher, line 275: Can at-rule be linked in any way?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+
+<h3 id="cssimportrule-extensions">
+Extensions to the <code>CSSImportRule</code> interface</h3>
+
+	The {{CSSImportRule}} interface is extended as follows:
+
+	<pre class='idl' export>
+	partial interface CSSImportRule {
+	  [SameObject] readonly attribute SheetNameList sheetNames;
+	  [SameObject] readonly attribute StyleSheetList namedStyleSheets;
```
mhochk, line 270: The value of `namedStyleSheets` should be manually available by iterating the existing `styleSheet` attribute and pulling out each `CSSSheetBlockRules` in order. Given that, is there a reason we should be exposing it as part of the standard? (Possibly an expected common use case?)

css-cascade-7/Overview.bs
mhochk: Looks like this one is missing an extra pair of <>, so it's being rendered as a custom HTML node.
`<<url-string>>`

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+			<<rule-list>>
+		}
+	</pre>
+
+	Such ''@sheet'' block rules have the same restrictions and processing
+	as a [=conditional group rule=] [[CSS-CONDITIONAL-3]]
+	with a false condition.
+
+Note: This means that rules within an ''@sheet'' block are not applied 
+until their [=named stylesheet=] is explicitly embedded into a document.
```
kbabbitt, line 108: I'm not sure if "embedded into a document" is the right condition here. In the example below, the `@sheet` rule *is* part of the document; its contents are accessible via CSSOM. Maybe "imported into another context" instead?

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+	</div>
+
+	''@sheet'' [=block at-rules=] may not be nested within any other CSS rules.
+
+<h3 id="sheet-declaration">
+Declaring Named Stylesheets</h3>
+
+	A [=named stylesheet=] may be declared by using
+	an [[#at-ruledef-sheet|@sheet block at-rule]],
+	assigning its child style rules into that sheet.
```
kbabbitt, line 130: I think this paragraph is a better candidate for the `<dfn>` of "named stylesheet" than where it is currently, and that this paragraph says mostly the same thing as the paragraph on lines 93-95. I'd suggest moving this text up to that location, editing down as needed, and making this instance of "named stylesheet" the `<dfn>` for that term.

css-cascade-7/Overview.bs
KurtCattiSchmidt: You're right, I added `[=named stylesheet=]` to make this explicit

css-cascade-7/Overview.bs
KurtCattiSchmidt: Yep, that was it!

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,402 @@
...
+
+<h3 id="the-csssheetblockrule-interface">
+The <code>CSSSheetBlockRule</code> interface</h3>
+
+	The {{CSSSheetBlockRule}} interface represents
+	the ''@sheet'' [[#at-ruledef-sheet|block rule]]:
+
+	<pre class='idl' export>
+	[Exposed=Window]
+	interface CSSSheetBlockRule : CSSGroupingRule {
```
kbabbitt, line 241: I think naming convention would call for `CSSSheetRule` here; see for example subheadings under https://drafts.csswg.org/cssom-1/#css-rules

css-cascade-7/Overview.bs
kbabbitt: I'd actually suggest linking to ''@sheet'' instead of "at-rule" to give a more specific link in the formatted spec.

kbabbitt (changes requested):

css-cascade-7/Overview.bs
KurtCattiSchmidt: But in this case, I went with Kevin's suggestion to link specifically to the `@sheet` definition

css-cascade-7/Overview.bs
KurtCattiSchmidt: Fixed by linking to the `@sheet` definition here too

css-cascade-7/Overview.bs
KurtCattiSchmidt: Yeah this one is tricky to phrase. I think "imported into another context" is more correct, so I'm going with that for now, but this might change in the future.

css-cascade-7/Overview.bs
KurtCattiSchmidt: You're right - I had based this on `@layer`, but that has to disambiguate between the block and statement, but I am omitting `@sheet` statements for now, because it seems there's no value in `@sheet` statements. This could change depending how this issue is resolved: https://github.com/w3c/csswg-drafts/issues/11998

https://www.w3.org/TR/css-cascade-5/#the-csslayerblockrule-interface

css-cascade-7/Overview.bs
KurtCattiSchmidt: You can manually access the individual rules that way, but not the actual `CSSStyleSheet` object.

The `CSSStyleSheet` has more options than the individual rules - https://drafts.csswg.org/cssom/#the-cssstylesheet-interface

This could potentially be plugged into `adoptedStyleSheets` in Shadow DOM as well - https://drafts.csswg.org/cssom/#dom-documentorshadowroot-adoptedstylesheets

(note that it currently says it must be a Constructable Stylesheet, but that was lifted recently and the specs/implementations haven't caught up - https://github.com/w3c/csswg-drafts/issues/10013#issuecomment-2165396092)

KurtCattiSchmidt: Also, the `CSSImportRule` already has a single-value accessor for `styleSheet` - https://www.w3.org/TR/cssom-1/#cssimportrule

That won't work for named stylesheets, but separating them isn't too bad - the existing API will return the stylesheet containing the global, unnamed sheet and the new `namedStylesheets` list contains the `@sheet` nested sheets.

css-cascade-7/Overview.bs
romainmenke: Maybe drop `into another context`?
Because you can also import it into the same context, right?

css-cascade-7/Overview.bs
romainmenke, line 148: Should it be stated how to handle named sheets that don't exist in the provided url?

`@import one, two from url('foo.css')` where 'foo.css':

```css
@sheet one {}
```

Does it fail entirely or does it still import `one`?

css-cascade-7/Overview.bs
```diff
@@ -117,18 +118,11 @@ until their [=named stylesheet=] is explicitly embedded into a document.
 		&lt;h1>Sample text&lt;/h1>
 		</pre>
 		In this example, the ''h1'' tag does not have the maroon color applied, as the ''@sheet'' 
-		named "framework" was not explicitly embedded into the document.
+		named "framework" was not explicitly imported into another context.
 	</div>
 
 	''@sheet'' [=block at-rules=] may not be nested within any other CSS rules.
```
kbabbitt: We resolved in https://github.com/w3c/csswg-drafts/issues/11999#issuecomment-2775867462 to allow nested `@sheet`.

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,396 @@
...
+		only loads the [=named stylesheet=] with a name of "foo" from "narrow.css" when the UA
+		<a href="https://www.w3.org/TR/css-conditional-3/#support-definition">supports</a> ''display: flex'',
+		and only applies the [=stylesheet=] on a <a href="https://www.w3.org/TR/CSS2/media.html#media-types">handheld</a> device
+		with a <a href="https://www.w3.org/TR/mediaqueries-4/#width">maximum viewport width</a> of 400px.
+
+		<pre>@import foo from url("narrow.css") supports(display: flex) handheld and (max-width: 400px);</pre>
+	</div>
+
+	<div class="example">
+		The following layer imports the [=named stylesheet=]s named "foo" and "bar" from "tabs.css" into
```
kbabbitt:
```suggestion
		The following layer imports the [=named stylesheets=] named "foo" and "bar" from "tabs.css" into
```

Bikeshed should be able to match this up to the definition: https://speced.github.io/bikeshed/#conjugating

css-cascade-7/Overview.bs
KurtCattiSchmidt: Dropped "into another context", as "imported" is sufficient, thanks @romainmenke!

css-cascade-7/Overview.bs
KurtCattiSchmidt: Good question, I will file an issue for this.

css-cascade-7/Overview.bs
KurtCattiSchmidt: Added "(besides nested ''@sheet'' blocks)"

css-cascade-7/Overview.bs
KurtCattiSchmidt: That is so cool! Fixed all instances.

css-cascade-7/Overview.bs
KurtCattiSchmidt: Filed https://github.com/w3c/csswg-drafts/issues/12465

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,396 @@
...
+	Syntactically, a [=named stylesheet=] must include an explicit [=sheet-name=], and is represented
+	by the <<sheet-name>> in ''@sheet'' and ''@import'' rules,
+	which is a single <<ident>> token:
+
+	<pre class='prod'>
+		<dfn><<sheet-name>></dfn> = <<ident>>
+	</pre>
+
+<h3 id="at-ruledef-sheet">
+Named Stylesheets: the `@sheet` block at-rule</h3>
```
kbabbitt:
```suggestion
Named Stylesheets: the ''@sheet'' block at-rule</h3>
```

css-cascade-7/Overview.bs
```diff
@@ -0,0 +1,396 @@
...
+	A [=named stylesheet=] may be declared by using
+	an [[#at-ruledef-sheet|@sheet block at-rule]],
+	assigning its child style rules into that sheet.
```
kbabbitt, line 95: As I was looking at the output HTML, I noticed `@sheet` was not link-ified. It might be good to rewrite this paragraph to work in a `<dfn>@sheet</dfn>` for that. The opening paragraph for `@import` in cascade-3 might be a good example for this.

KurtCattiSchmidt: Latest version adjusted the links to work like `@import` in cascade-3 - LMK if this is what you had in mind.

kbabbitt: lgtm!

KurtCattiSchmidt:
@tabatkins and @fantasai, would at least one of you be able to give some feedback before the F2F in August? I'm expecting some issues to come up that would be nice to discuss there.

tabatkins:
Yup, will review this week.

PR #11981, opened 2025-03-22
[css-forms-1] Add fieldset and legend

lukewarlow:
See https://github.com/w3c/csswg-drafts/issues/11983

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -93,7 +93,7 @@ spec:css-forms-1; type:value; for:/; text:::placeholder
   To apply the [=basic appearance=] on individual controls, the following code is used:
 
   <pre class="lang-css">
-    input, textarea, meter, progress, button, select {
+    input, textarea, meter, progress, button, select, fieldset, legend {
       appearance: base;
     }
   </pre>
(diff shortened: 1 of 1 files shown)


nt1m:
I think we should spend some time thinking about what kind of improvements we want to give to fieldset/legend, and whether it makes sense for them to follow the `appearance: base` pattern.

e.g. for `fieldset`, if the main thing we want to change is the border to also be `currentColor` to match the rest, it might just be possible to change the existing `border: 2px groove threedface` to `border: 1px solid currentColor` directly everywhere with negligible/no web compat breakage. We've changed the border-color before for things like `table` or `hr` successfully, so I don't think this is infeasible.

For `legend`, I've heard valid complaints about the layout being hardcoded and not overridable, and there was some attempt from @zcorpan to previously address it using `appearance: none`. I would consider filing an issue against the css-forms-1 to explore this further. Maybe `appearance: base` would also make sense here, but I wouldn't rush into it without having the discussion first.

lukewarlow:
https://github.com/w3c/csswg-drafts/issues/11983

zcorpan:
See https://github.com/whatwg/html/issues/3912

The conclusion in https://github.com/whatwg/html/pull/4035#issuecomment-1476330098 was that `appearance` is not the right mechanism to control the magic of `fieldset` rendering.

zcorpan:
The rendering rules are defined here https://html.spec.whatwg.org/#the-fieldset-and-legend-elements (this could probably be moved to a CSS spec though).

PR #11984, opened 2025-03-22
[css-forms-1] Fix double borders

### LINKED ISSUE #11841: [css-forms-1] Double border style for progres, meter and input[type=checkbox][switch]
Per https://drafts.csswg.org/css-forms/#layout

Meter, Progress and switch inputs get a double border, once on the "host" element, once on the inner track.

```css
input:not([type=file], [type=range]),
...
::track,
meter,
progress {
  border: 1px solid currentColor;
  background-color: transparent;
}
```

I think the meter, and progress elements should be removed from this second rule (they also already get a transparent background-color from the base rule). [type=checkbox][switch] should also be added to the not selector on the first line.

lukewarlow:
Fixes #11841

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -800,14 +800,12 @@ progress {
 
 ## Layout ## {#stylesheet-layout}
 ```css
-input:not([type=file], [type=range]),
+input:not([type=file], [type=range], [type=checkbox][switch]),
 textarea,
 button,
 ::file-selector-button,
 ::slider-track,
-select,
-meter,
-progress {
+select {
   border: 1px solid currentColor;
   background-color: transparent;
 }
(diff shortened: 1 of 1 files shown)


lukewarlow:
cc @nt1m does this seem reasonable to merge? Or should I try and get a discussion in the group for which approach is best? Seems like a bug fix personally?

nt1m:
It kind of depends on which way to go, if you want the more traditional slider style with the track smaller than the thumb, then maybe it makes more sense to remove the border from the host element rather the track.

Either way we go I'm fine accepting it as a bug fix in the short term, but it would be good to discuss that bigger decision in the WG.

lukewarlow:
I've swapped to the other approach where the track has the border NOT the container. As you say for stuff like range where you might want the track to be smaller (even some switches are like this), it doesn't make sense for the border to be on the parent container.

lukewarlow:
Going to merge as a bug fix but like you say we should have a wider discussion on this.

PR #12070, opened 2025-04-11
[css-values-5] All participants in a cycle become invalid

andruud:
The current spec only remembers that we're in a cycle at the single point we're checking for cycles (step 1 of "substitute arbitrary substitution function"). This is not enough:

 --x: var(--y, 1);
 --y: var(--x, 2);

The current spec would produce computed values of --x:2 and --y:1, which doesn't make sense: if --x and --y indeed have non-invalid computed values, then the fallbacks should not have been taken in the first place.

I believe this is an unintended side effect of
f7a077668ea06a0509203c65d4d3cd5ef25ff458, therefore this PR restores the original behavior of making all properties in a cycle invalid.


tabatkins:
Ah, perfect, I'd just noticed this lack when looking at the removed attr() cycle section and checking that I didn't still need the examples.

PR #12074, opened 2025-04-14
[css-values-4] Use "include" when fetching cross-origin CSS resources

### LINKED ISSUE #12073: [css-values-4][css-images-4] Style resource fetching should probably use credential mode "include" in the default case
I was talking with @annevk, and he mentioned that he thought the "[fetch a style resource](https://www.w3.org/TR/css-values-4/#fetch-a-style-resource)" algorithm in CSS Values 4 should probably follow HTML (and at least what WebKit does) and use [credentials mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) is "include" rather than "same-origin" as the default value.

This would better match HTML's "[create a potential-CORS request](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#create-a-potential-cors-request )" algorithm used for most HTML Subresource fetching.

(@annevk, please correct anything I have wrong here).

cc @tabatkins, @fantasai

noamr:
This matches existing implementation, and was a bug in the existing spec.

Closes #12073

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -1169,7 +1169,7 @@ URL Processing Model</h4>
 			whose [=request/destination=] is |destination|,
 			[=request/mode=] is |corsMode|,
 			[=request/origin=] is |environmentSettings|'s [=environment settings object/origin=],
-			[=request/credentials mode=] is "same-origin",
+			[=request/credentials mode=] is "include",
 			[=request/use-url-credentials flag=] is set,
 			[=request/client=] is |environmentSettings|,
 			and whose [=request/referrer=] is |environmentSettings|'s [=API base URL=].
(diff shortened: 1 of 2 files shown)


weinig:
If you make this change, you'll also need to update CSS Values 5 to specify what do for `crossorigin(anonymous)` as it currently just uses the default (https://drafts.csswg.org/css-values-5/#typedef-request-url-modifier-crossorigin-modifier).

noamr:
> If you make this change, you'll also need to update CSS Values 5 to specify what do for `crossorigin(anonymous)` as it currently just uses the default (https://drafts.csswg.org/css-values-5/#typedef-request-url-modifier-crossorigin-modifier).

You're right! Not just that, but existing callers that use `|corsMode| = "cors"` would receive the wrong credentials mode.
Amended so that:
- The URL modifiers in css-values-5 explicitly set `same-origin` when anonymous
- The default for cors-mode is anonymous, and only the no-cors default changes to use-credentials.

tabatkins (approved):

Issue #11957, opened 2025-03-18
[css-viewport] Behavior of `segments` when there is only 1 segment


bramus:
The spec currently reads the following for [the `segments` property of `window.viewport`](https://drafts.csswg.org/css-viewport/#segments)

> 4. Returns `null` if there is only a single viewport segment and abort these steps.

This very counterintuitive when trying things out and I would have expected for 1 segment to reported, with its values at origin `0,0` and its `width`/`height` equal to the `innerWidth` and `innerHeight` respectively.

Admittedly the fallback authors can do is pretty easy here, yet I still find it weird to return `null` as currently specced. That keyword is typically reserved to indicate that _there is no value_ or _there is no such thing_, whereas there is definitely 1 segment here.

Having it return 1 segment would also help distinguish the 1 segment situations vs “the document is currently not active” situation. The latter also returns `null`, as per spec:

> 1. If the [Viewport](https://drafts.csswg.org/css-viewport/#viewport%E2%91%A0)’s associated [Document](https://dom.spec.whatwg.org/#document) is not [fully active](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active), return `null`

In summary, the proposed change is to return 1 segment instead of `null` when there is only 1 segment.

/cc @darktears

darktears:
This has been discussed in the past unfortunately I can't find the discussion anymore.

Semantically a segment is a part of something that is divided. If the viewport is not divided, then there are no segments. When the device is not folded there are no division therefore no segments.

We don't want author to write code against one segment because they should treat that case as the same as if there wasn't any split (meaning the device is used flat in the case of a foldable) or it's just a regular device. Furthermore, that one segment will return the same information as the viewport size bringing no useful information.

Finally, not returning a single segment match the CSS part of this API where the MQs do not resolve for a value of one segment. In other words, MQs values are always >= 2 for the segments.

bramus:
> We don't want authors to write code against one segment because they should treat that case as the same as if there wasn't any split (meaning the device is used flat in the case of a foldable) or it's just a regular device.

It’s not about writing code against 1 segment, but about consistency when enumerating the segments. We keep saying “1 segment” but that does not rhyme with `null`.

> Finally, not returning a single segment match the CSS part of this API where the MQs do not resolve for a value of one segment.

Which MQ are you talking about? AFAIK there is only the `device-posture` media feature and some env-vars that people can use. 

---

[...]

PR #12088, opened 2025-04-16
[css-viewport-1] Make sure to always return one segment

darktears:
When the device is not folded (flat) or it's a regular device always return one segment (typically the size of the viewport). This is to follow developer feedback and consistency with the rest of the platform.

#11957

### DIFF
--- a/css-viewport-1/Overview.bs
+++ b/css-viewport-1/Overview.bs
@@ -572,13 +572,13 @@ interface Viewport {
 The {{segments}} property is an array of {{DOMRect}} that represent the dimensions of each existing viewport segment.
 Each {{DOMRect}} contains the geometry of the segment (x, y, width, height) in CSS ''<length>/px''.
 
-Additonal details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
+Additional details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
 
 The {{segments}} attribute getter steps are:
 1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return null.
 2. Let |topLevelTraversable| be |document|'s [=relevant global object=]'s [=/navigable=]'s [=browsing context/top-level traversable=].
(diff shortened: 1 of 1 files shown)


zcorpan:
This makes sense to me.

css-viewport-1/Overview.bs
```diff
@@ -572,13 +572,13 @@ interface Viewport {
...
 
-Additonal details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
+Additional details about the definition of a viewport segment can be found here: [[css-env-1#viewport-segments]].
 
 The {{segments}} attribute getter steps are:
 1. If the {{Viewport}}'s associated {{Document}} is not <a>fully active</a>, return null.
 2. Let |topLevelTraversable| be |document|'s [=relevant global object=]'s [=/navigable=]'s [=browsing context/top-level traversable=].
 3. If |topLevelTraversable|.[=[[DisplayFeaturesOverride]]=] is non-null, return {{Viewport}}'s [[css-env-1#viewport-segments|segments]] array calculated from |topLevelTraversable|.[=[[DisplayFeaturesOverride]]=].
-4. Returns null if there is only a single viewport segment and abort these steps.
+4. If there is only a single viewport segment (typically the size of the {{Viewport}}) return it and abort these steps.
```
jakearchibald, line 581:
```suggestion
4. If there is only a single viewport segment (typically the size of the {{Viewport}}) return it.
```

If you return, the algorithm is over https://infra.spec.whatwg.org/#algorithm-control-flow

jakearchibald (approved):

bramus:
Thanks! Before merging the change I think we do need a resolution on the issue for this, which can be done async.

darktears:
@bramus what would be the most efficient way to do that? email to www-style?

tabatkins (changes requested):
Needs to say that it still returns an array, containing that single size.

darktears:
@tabatkins updated the wording, good catch. Thanks.

darktears:
The resolution was approved so @tabatkins or @emilio please merge at your earliest convenience.

PR #12105, opened 2025-04-21
[css-contain-2] WIP: Stop monkey-patching the HTML spec.

emilio:
This exports an "update content relevancy for a document" algorithm that HTML can call at the right time.


css-contain-2/Overview.bs
```diff
@@ -1994,6 +1994,37 @@ Suppressing An Element's Contents Entirely: the 'content-visibility' property {#
...
+
+		1. Let <var>changed</var> be false.
+		1. For each <var>element</var> [=connected=] to doc with ''auto'' used value of ''content-visibility''
+			1. Determine proximity to the viewport for <var>element</var>.
+			1. If <var>element</var> is not [=currently relevant to the user=] and <var>element</var> is [=relevant to the user=], then:
+				2. Set <var>changed</var> to true.
+				2. Set <var>element</var>'s [=currently relevant to the user=] flag to true.
+			1. If <var>isFirstIteration</var> is false, then [=continue=].
+			1. If <var>element</var> is [=relevant to the user=], then:
+				1. If <var>element</var>'s [=currently relevant to the user=] flag is false, set <var>changed</var> to true.
```
vmpstr, line 2012: This condition seems to be the same as the second condition above
> If <var>element</var> is not [=currently relevant to the user=] and <var>element</var> is [=relevant to the user=], then

in fact I don't think it would hit since we update `currently relevant to the user` on line 2009. Or am I missing a subtlety here?

css-contain-2/Overview.bs
```diff
@@ -1994,6 +1994,37 @@ Suppressing An Element's Contents Entirely: the 'content-visibility' property {#
...
+	<div algorithm="update content relevancy for a document">
+		To <dfn export>update content relevancy for a document</dfn> given a document <var>doc</var>, and a boolean <var>isFirstIteration</var>:
+
+		1. Let <var>changed</var> be false.
+		1. For each <var>element</var> [=connected=] to doc with ''auto'' used value of ''content-visibility''
+			1. Determine proximity to the viewport for <var>element</var>.
+			1. If <var>element</var> is not [=currently relevant to the user=] and <var>element</var> is [=relevant to the user=], then:
+				2. Set <var>changed</var> to true.
+				2. Set <var>element</var>'s [=currently relevant to the user=] flag to true.
+			1. If <var>isFirstIteration</var> is false, then [=continue=].
```
vmpstr, line 2010: Because of the comment below, it's unclear to me the role isFirstIteration is playing here. `changed` is set in either case

css-contain-2/Overview.bs
```diff
@@ -2308,14 +2339,7 @@ Restrictions and Clarifications {#cv-notes}
 	</wpt>
 
 3. If an element starts or stops [=skipped contents|skipping its contents=],
-	this change happens
-	after the requestAnimationFrame callbacks
-	of the frame that renders the effects of the change
```
vmpstr, line 2313: So there's a bit of a subtlety here. What this is saying is that if we have frames 1 and 2 (separate update the rendering steps), and we determine that an element is relevant in step 1, then the effect of that relevance (ie it allows rendering, doesn't have strict containment etc) happen on frame 2. The difference is that if you try and observe the size of the element between rAFs in a timeout, then between frame 1 and 2 it should act as if it's still not relevant wrt size containment. See the note in this spec below which tries to explain that.

I don't mind if we change this behavior, but we need a larger discussion for that.

emilio: That's what the intention of the _currently relevant to the user_ flag is (maybe I botched the language).

"currently relevant to the user" is meant to be the thing that has effect on the rendering. But you're right I might have messed up my prose ;)

css-contain-2/Overview.bs
emilio: Yeah I think this branch is dead effectively, but just because of the mistake you pointed out above.

css-contain-2/Overview.bs
emilio: The important bit is that if `isFirstIteration` is false, then non-initial determination is not supposed to happen (this is what implements the semantics you were talking about in your other comments). I think I just botched the condition  in line 2007 (it's supposed to deal only with the initial proximity determination).

I'll fix, thanks for the review.

css-contain-2/Overview.bs
emilio: Ok, I think 4248490afa1fbd774f5d385fb952afd4f4ba00d7 preserves the behavior you were talking about. Though I think I'd rather treat the initial and non-initial viewport determination the same, wdyt?

vmpstr:
The new text looks good to me. One unfortunate consequence is that we always run the style&layout steps even if we know we have pending relevance changes which as far as I can tell pretty much guarantee that we'll have to run style&layout again.

Perhaps this is in the realm of browsers being able to optimize this, but it's unclear if there's interactions with other features that do require the initial style&layout before relevant state changes

emilio:
> The new text looks good to me. One unfortunate consequence is that we always run the style&layout steps even if we know we have pending relevance changes which as far as I can tell pretty much guarantee that we'll have to run style&layout again.

I think that's necessary tho, right? Otherwise whether you see the effect of something like:

```
el.focus();
el.style.contentVisibility = "auto";
```

On this frame or the next depends on whether something has updated style and layout in the interim, right?

Issue #6400, opened 2021-06-21
[css-ui][proposal] Current scrolling direction pseudoclasses


AmeliaBR:
A common UI pattern on the web is to hide/show or partially collapse some content based on whether the user is currently scrolling up or down. I'm not particularly fond of this UI pattern, but it is widespread and is often necessary for long pages on mobile screens.

This currently requires JavaScript scroll listeners to change a class on the body or other scroll container in order to trigger the alternate layout of the collapsible headers/footers (and that's assuming it isn't done entirely in JS-framework state propagation!)

Browser-managed scrolling direction pseudoclasses on the scroll container would eliminate the need for many JS scroll listeners. Something like `body:scrolling-forward` vs `body:scrolling-backward` for primary (block) direction scrolling, and some other logical names for cross/inline scrolling. Maybe also some way to distinguish when scrolling has stopped for a while.

Pseudoclasses could also enable browser-managed heuristics for better detecting when the scroll direction has changed (or stopped), based on the device scroll mechanism (touch gesture, keyboard, mouse wheel) or user accessibility customizations. For example, a user with shaky fingers might want to customize how much they need to scroll in a given direction before the layout shifts on them.

(I don't have the capacity to work on this; just throwing it out there in case someone else wants to pick it up!)

jonjohnjohnson:
Because we do have both physical and logical dimensions to deal with in these scenarios, I wonder if it's worth recommending this as a functional pseudo-class that takes a direction/state of the scrolling element as a parameter?

- `body:scrolling(bottom)` _currently being scrolled toward bottom edge_
- `body:scrolling(inline-start)` _currently being scrolled toward inline-start edge_
- `body:scrolling` _currently being scrolled in any direction_

I've found myself disabling `:hover`,`:active`, and even animation stylings depending on the user currently scrolling. The last example could help in these cases.

In general, I do wonder about *some* circularity when this pseudo class matches and then layout/flow changes so the scrollHeight/scrollWidth is changed in a way where the browser must manipulate the scroll position and now the pseudo class wouldn't match?

jeysal:
This was just mentioned as an alternative solution to #5670 :)
One shortcoming I see with an on-or-off pseudoclass is that you aren't literally scrolling out the top bar with e.g. your finger, i.e. it doesn't move with your finger as you move it up and down. It merely switches to a state once your scrolling reaches a threshold, and the best you can do as far as I can tell is give it a transition, which does not correlate to your finger movements.

[...]

PR #12130, opened 2025-04-28
Add scroll-direction state and updated overflowing state to scrollable

tursunova:
[css-conditional-5] Explainer for the proposal in  https://github.com/w3c/csswg-drafts/issues/6400#issuecomment-2794796926.

### DIFF
--- a/css-conditional-5/scroll_state_explainer.md
+++ b/css-conditional-5/scroll_state_explainer.md
@@ -147,12 +147,35 @@ a carousel.
 
 - [Author request on github](https://github.com/w3c/csswg-drafts/issues/7430)
 
-### Overflowing
+### Scrollable
 
-Query whether a container has scrollable overflow. Can be used to indicate
+Query whether a container has [scrollable overflow](https://drafts.csswg.org/css-overflow-3/#scrollable-overflow-region). Can be used to indicate
 there is content to scroll to in a given direction.
 
 Needs further exploration.
+
+### Scroll-direction
+
+Query based on direction of [scrolling](https://drafts.csswg.org/cssom-view/#scrolling). For instance:
+
+```css
+@container scroll-state(scroll-direction: top) {
+  .scrolling-up {
+    translate: 0 0;
+  }
+}
+```
+
+#### Workaround
+
+The current workaround solution to style elements based on scroll direction is by using `transition-delay` trick, see 
+[Solved by CSS Scroll-Driven Animations: hide a header when scrolling down, show it again when scrolling up](https://www.bram.us/2024/09/29/solved-by-css-scroll-driven-animations-hide-a-header-when-scrolling-up-show-it-again-when-scrolling-down/). However, in this approach, if a user was still pressing on the scrollbar after scrolling to the specific direction the `scroll-direction` would be considered as `none` since the scrollbar wasn't moving anymore. This behaviour is inconsistent with [the completed state of scrolling](https://drafts.csswg.org/cssom-view/#scroll-completed).
+
+#### Use Cases and Author Requests
+
+To hide/show or partially collapse some content based on the direction of scrolling.
+
+- [Author request on github](https://github.com/w3c/csswg-drafts/issues/6400)
 
 ### Anchor position fallback
 
@@ -176,6 +199,8 @@ for each state which can be combined with size container types:
 
 - `stuck`
 - `snapped`
+- `scrollable`
+- `scroll-direction`
 
 Query values for `stuck`:
 
@@ -192,6 +217,43 @@ Query values for `stuck`:
 Query values for `snapped`:
 
 - `none`
+- `x`
+- `y`
+- `block`
+- `inline`
+- `both`
+
+Query values for `scrollable`:
+
+- `none`
+- `top`
+- `right`
+- `bottom`
+- `left`
+- `block-start`
+- `inline-start`
+- `block-end`
+- `inline-end`
+- `x`
+- `y` 
+- `block`
+- `inline`
+
+
+Query values for `scroll-direction`:
+
+- `none`
+- `any`
+- `top`
+- `right`
+- `bottom`
+- `left`
+- `block-start`
+- `inline-start`
+- `block-end`
+- `inline-end`
+- `x`
+- `y` 
 - `block`
 - `inline`
 


css-conditional-5/scroll_state_explainer.md
```diff
@@ -147,13 +147,36 @@ a carousel.
...
+  .scrolling-up {
+    translate: 0 0;
+  }
+}
+```
+
+#### Workaround
+
+The current workaround solution to style elements based on scroll direction is by using `transition-delay` trick, see 
+[Solved by CSS Scroll-Driven Animations: hide a header when scrolling down, show it again when scrolling up](https://www.bram.us/2024/09/29/solved-by-css-scroll-driven-animations-hide-a-header-when-scrolling-up-show-it-again-when-scrolling-down/). However, in this approach, if a user was still pressing on the scrollbar after scrolling to the specific direction the `scroll-direction` would be considered as `none` since the scrollbar wasn't moving anymore. This behaviour is inconsistent with [the completed state of scrolling](https://drafts.csswg.org/cssom-view/#scroll-completed).
```
lilles, line 172: > However, in this approach, if a user was still pressing on the scrollbar after scrolling to the specific direction the `scroll-direction` would be considered as `none` since the scrollbar wasn't moving anymore.

This sentence has an assumption about when the scroll direction is updated that is not described in this doc.

Nit: "scrolling to the specific direction" -> "scrolling in the specific direction"

lilles, line 172: In general, could you wrap the lines at roughly 80 chars like the rest of the doc?

css-conditional-5/scroll_state_explainer.md
```diff
@@ -147,13 +147,36 @@ a carousel.
...
+@container scroll-state(scroll-direction: top) {
+  .scrolling-up {
+    translate: 0 0;
+  }
+}
+```
+
+#### Workaround
+
+The current workaround solution to style elements based on scroll direction is by using `transition-delay` trick, see
```
lilles, line 171: A workaround solution to style ... by using a `transition-delay` ...

css-conditional-5/scroll_state_explainer.md
tursunova: Added open questions with options about when `scroll-direction` should be `none`.

tabatkins:
The design assumption here seems to be that the query only returns the direction of a *currently executing* scroll; once the scroll is done, it returns to "none". But I don't think that actually addresses the use-cases linked - the big one is making a header show fully at first, semi-hide when you start scrolling down so you have more space to read the page content, then reveal itself again when you scroll up.

This suggests that it needs to reflect the direction of *the current or most recent scroll operation*. If you scroll down and then stop scrolling and read for a bit, the author would want it to still reflect `down`, so the header stays semi-hidden. It would probably *never* return to `none`.

We should also make sure it does sufficiently solve this case. Like, the simplest code I could imagine an author using is:

```css
.header {
	translate: 0px;
	transition: translate 1s ease-in-out;
}
@container scroll-state(scroll-direction: down) {
	.header {
		translate: -50%;
	}
}
```

This would make the header reveal itself at first, and when the user starts scrolling up, but slide itself up (presumably into the unscrollable region) if they've been scrolling down.

--------

Another question - do we want to remember the scroll direction of each axis independently, or together? Like, given the code above, I suspect the author wants the header's shown/hidden state to care only about the most recent *block-axis* scroll; if the user scrolls sideways it shouldn't stop matching `down`. Are there use-cases for caring about the most recent scroll regardless of axis?

-------

The JS solutions for this sort of header-autohiding also often track the scroll a little - if the header is currently hidden, and you *start* scrolling up a little, it often doesn't immediately pop into fully showing, but rather just slides out a *little*, tracking with the amount of scroll until it's fully revealed. I don't think there's a clean way to do that with *just* this feature; that seems like a scroll-driven animation instead. But we should look into how that would be done, to make sure authors that want that feature don't still have to just go with a full-JS solution instead.

tursunova:
Thanks @tabatkins , this makes sense. Then few open questions with this approach:
- Since now the value reflects the most recent scroll in that direction, value `none` becomes unreachable expect of the initial state, should we leave it then or should we remove it? If we decide to remove `none`:
1. We can have `inline-start` and `block-start` as initial values. 
2. Values `X`, `Y`, `block` and `inline` will always match, so there would be no point of having them.
3. if we have latin script and only vertical scroller, should we match `inline-start` initially?
- As far as I understood, if we want to match the most recent scroll direction (not `none`) once we stop scrolling, then the value reflects the most recent scroll in that direction, e.g. if we scrolled `down` and then `right`, `scroll-direction: bottom` will still match after scrolling right. But  then is there a use case when we want to hide/show some vertical scrolling specific content once we start scrolling horizontally? Then we might have 2 values, let's say `bottom` and `latest-bottom`, where one will still match after scrolling horizontally and another will only match if latest scrolling (not latest _vertical_ scrolling) was `bottom`.

yisibl:
> Are there use-cases for caring about the most recent scroll regardless of axis? If there are, maybe we can let an author opt into that with another keyword, like `scroll-direction: down most-recent`, so that the simpler version `scroll-direction: down` can just care about the direction of the most recent scroll _in that axis_.

The examples in @bramus' article have a lot of use cases that only care about the current scroll direction.

- https://codepen.io/bramus/pen/oNJeLWg
- https://codepen.io/bramus/pen/wvRqVBm

bramus:
> The examples in @bramus' [article](https://www.bram.us/2023/10/23/css-scroll-detection/) have a lot of use cases that only care about the recent scroll direction.

But there’s also demos that depend the last-non-`none` direction. The “Hidey Bar” is the typical example for this, which Tab also [mentioned](https://github.com/w3c/csswg-drafts/pull/12130#issuecomment-2839534205).

In https://brm.us/hidey-bar I figured out how to _“remember”_ the last-non-`none` direction: by (ab)using an infinite `transition-delay` on the custom prop that stores the direction – as documented in the explainer that is part of this PR.

Would be nice/handy if this “remember“ option were built into the API. Some ideas:

1. Perhaps we distinguish between `scroll-direction` and `active-scroll-direction`?
  - `active-scroll-direction` represents the active scroll direction. If you stop scrolling, it becomes `none` again.
  - `scroll-direction` starts as `none` but once you scroll it remembers the last active scroll direction that is not `none`.
2. More values? E.g. `idle-after-down`

Personal preference for 1 – seems more easy for authors.

---

> Another question - do we want to remember the scroll direction of each axis independently, or together?

Good question. I see one use cases for scroll-based games that allow you to control a character walking in a 8-directional movement (cardinal directions + diagonals), which would rely on remembering each axis independently.

- More values such as `down-right`?
- More `scroll-direction` variants? i.e. `(active-)(vertical|horizontal)-scroll-direction`

But you could ask yourself if that’s a common enough use-case to tackle …

css-conditional-5/scroll_state_explainer.md
```diff
@@ -147,12 +147,51 @@ a carousel.
...
 
-Query whether a container has scrollable overflow. Can be used to indicate
-there is content to scroll to in a given direction.
+Query whether a container has [scrollable overflow](https://drafts.csswg.org/css-overflow-3/#scrollable-overflow-region).
+Can be used to indicate there is content to scroll to in a given direction.
 
-Needs further exploration.
+### Direction
+
+Query based on the most recent direction of [scrolling](https://drafts.csswg.org/cssom-view/#scrolling).
```
lilles: Can remove "based on"

css-conditional-5/scroll_state_explainer.md
```diff
@@ -147,12 +147,51 @@ a carousel.
...
+  }
+}
+```
+
+#### Workaround
+
+A workaround solution to create queries based on scrolling direction is
+described in [Solved by CSS Scroll-Driven Animations: hide a header when
+scrolling down, show it again when scrolling up](https://www.bram.us/2024/09/29/solved-by-css-scroll-driven-animations-hide-a-header-when-scrolling-up-show-it-again-when-scrolling-down/),
+uses `transition-delay` trick to get the active scroll direction.
```
lilles: "uses a"

css-conditional-5/scroll_state_explainer.md
```diff
@@ -192,6 +231,43 @@ Query values for `stuck`:
...
+- `inline-end`
+- `x`
+- `y` 
+- `block`
+- `inline`
+
+
+Query values for `direction`:
+
+- `none`
```
lilles, line 259: Should `none` be removed? Or is it still expected that the start value is `none` even for the state that reflects the most recent scroll?

css-conditional-5/scroll_state_explainer.md
```diff
@@ -192,6 +231,43 @@ Query values for `stuck`:
...
+- `any`
```
lilles: Why do we need `any`? I guess we only need it if we keep `none`?

It's common for media and container queries that authors just rely on the boolean context instead.

lilles (changes requested):

css-conditional-5/scroll_state_explainer.md
tursunova: Removed `any`, thanks.

css-conditional-5/scroll_state_explainer.md
tursunova: I think it would make sense to have `none`, so users can identify whether  there is a scroller in some direction or not. We talked about that with Tab, they also agrees on that.

Issue #9562, opened 2023-11-06
[css-values-5] allow specifying flat vs light tree in `sibling-index()` and `sibling-count()`


argyleink:
In a goal to allow author flexibility with `sibling-index()` and `sibling-count()` and the value resolution from the flat tree vs the light tree, I propose adding a parameter to the functions so developers may specify which tree they need. This would allow developers to work with happy path [default flat tree default behavior](https://github.com/w3c/csswg-drafts/blob/762eacee436fdcb1b11e7e482e6cdb3fee161317/css-values-5/Overview.bs#L1432), but when needed they can specify resolution against the light dom.

```css
ul {
  li {
    animation-duration: calc(sibling-index() * 50ms);
  }
}

:host {
  li {
    animation-duration: calc(sibling-index(light-tree) * 50ms);
  }
}
```

Trade offs: 
- it would disrupt the potential for `of <<selector>>` syntax to be used between the parenthesis
- it would depart from the alignment with `nth-child()`

Use cases:
I don't really have any example use cases for wanting light tree, but the fact there are two trees to resolve against makes me think eventually someone will show up with a use case 😅

lilles:
The current spec saying that `sibling-index()` / `sibling-count()` follows the flat tree means it will not correspond with the `:nth-*` selectors for slotted elements. Is that a behavior we want in the default case? I lean towards having the flat-tree counting as an option.

There is a note in the spec that these function may be extended to accept `of <selector>`, which would not work on the flat tree. It would be confusing to have a single function work both on flat tree and not at the same time?

emilio:
FWIW sibling-index and co working on the flat tree by default seems super confusing for authors (CSS uses dom index for virtually everything else).

Whether using the flat tree should be an option I don't have a strong opinion on...

lilles:
The edits for using DOM tree, not flat tree order are in. The issue should be kept open for adding flat tree alternatives.

lilles:
WPT tests are also update fwiw

PR #12144, opened 2025-05-02
[css-values-5] Don't use flat tree for tree-counting #9562

lilles:
Per resolution[1], tree counting functions should not use the flat tree for indexing. Instead use the term 'inclusive siblings' from the DOM spec, which is the same term that :nth-*() selectors use.

[1] https://github.com/w3c/csswg-drafts/issues/9562#issuecomment-2824834068

### DIFF
--- a/css-values-5/Overview.bs
+++ b/css-values-5/Overview.bs
@@ -2755,7 +2755,7 @@ Tree Counting Functions: the ''sibling-count()'' and ''sibling-index()'' notatio
 	as an <<integer>>,
 	the index of the element
 	on which the notation is used
-	among the children of its parent.
+	among its [=inclusive siblings=].
 	Like '':nth-child()'',
 	''sibling-index()'' is 1-indexed.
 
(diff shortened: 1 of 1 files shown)


css-values-5/Overview.bs
```diff
@@ -2803,9 +2803,8 @@ Tree Counting Functions: the ''sibling-count()'' and ''sibling-index()'' notatio
 
 	</div>
 
-	Note: Like the rest of CSS (other than [=selectors=]),
-	''sibling-count()'' and ''sibling-index()''
-	operate on the [=flat tree=].
+	Note: These functions may, in the future, have variants that support counting
+	[=flat tree=] siblings.
```
tabatkins, line 2807:
```suggestion
	Note: These functions, to match [=selectors=] like '':nth-child()'', operate on the DOM tree,
	rather than the [=flat tree] like most CSS values do.
	They may, in the future, have variants that support counting [=flat tree=] siblings.
```

PR #12155, opened 2025-05-02
Fix bikeshed errors in color adjust spec

alisonmaher:
[css-color-adjust-1] This change fixes some auto-links that weren't working and adds exact locations for some links that were ambiguous. This cleans up all current bikeshed warnings for the color adjust spec, except for complaints about the "system color" definition not existing as an exportable definition, as this will require a change to the color spec.

### DIFF
--- a/css-color-adjust-1/Overview.bs
+++ b/css-color-adjust-1/Overview.bs
@@ -29,11 +29,11 @@ spec:fill-stroke-3; type:property;
 	text:fill
 	text:stroke
 spec:css-gaps-1; type:property; text:column-rule-color
-spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/;
+spec: html;
 	text:body; type:element
 	text:meta; type:element
-	text:navigable; type:dfn; for:Window; url:nav-history-apis.html#window-navigable
-	text:top-level traversable; type:dfn; for:/; url:document-sequences.html#top-level-traversable
+	text:navigable; type:dfn; for:Window;
+	text:top-level traversable; type:dfn; for:/;
 </pre>
 
 <pre class=anchors>
@@ -615,13 +615,13 @@ Forced Color Palettes {#forced}
 				<tr>
 					<td>''Mark''</td>
 					<td>
-						N/A - this <<system-color>> keyword should not be adjusted.
+						N/A - this [=system color=] keyword should not be adjusted.
 					</td>
 				</tr>
 				<tr>
 					<td>''MarkText''</td>
 					<td>
-						N/A - this <<system-color>> keyword should not be adjusted.
+						N/A - this [=system color=] keyword should not be adjusted.
 					</td>
 				</tr>
 				<tr>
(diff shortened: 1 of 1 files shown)


css-color-adjust-1/Overview.bs
```diff
@@ -20,13 +20,20 @@ WPT Display: open
...
 spec:css-color-4; type:property; text:color
+spec:css-color-5; type:function; text:color()
 spec:fill-stroke-3; type:property;
 	text:fill
 	text:stroke
-spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/; type: dfn;
-	text: navigable; for:Window; url: nav-history-apis.html#window-navigable
-	text: top-level traversable; for:/; url: document-sequences.html#top-level-traversable
+spec:css-gaps-1; type:property; text:column-rule-color
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/;
```
tabatkins, line 32:
```suggestion
spec: html;
```

css-color-adjust-1/Overview.bs
```diff
@@ -20,13 +20,20 @@ WPT Display: open
...
+	text:body; type:element
+	text:meta; type:element
+	text:navigable; type:dfn; for:Window; url:nav-history-apis.html#window-navigable
+	text:top-level traversable; type:dfn; for:/; url:document-sequences.html#top-level-traversable
```
tabatkins, line 36:
```suggestion
	text:navigable; type:dfn; for:Window;
	text:top-level traversable; type:dfn; for:/;
```

css-color-adjust-1/Overview.bs
```diff
@@ -606,13 +615,13 @@ Forced Color Palettes {#forced}
 				<tr>
 					<td>''Mark''</td>
 					<td>
-						N/A - this [=system color=] keyword should not be adjusted.
+						N/A - this <<system-color>> keyword should not be adjusted.
```
tabatkins, line 618: The term "system color" exists, it's just not exported by accident. This (and other changes to `<<system-color>>` should be reverted; I'll go fix Colors to export the term "system color" properly.

tabatkins (changes requested):
The link-defaults suggestions are just removing the url/urlprefix keys because those don't do anything in a link-defaults block. Someone copy-pasted from an "anchors" block, where those are meaningful keys.

css-color-adjust-1/Overview.bs
alisonmaher: Done

css-color-adjust-1/Overview.bs
alisonmaher: Done

css-color-adjust-1/Overview.bs
alisonmaher: Ah ok good to know, I've reverted all those back to [=system color=].

alisonmaher: @tabatkins Interestingly, I still am seeing these not linking even though I see you landed a change to export "system color" from the color spec.

And if I run `bikeshed refs --text="system color"` all I get back is `[]`

Issue #11857, opened 2025-03-06
[css-color-adjust-1] Forced Colors Mode support for gap decorations


alisonmaher:
The CSS Gap Decoration spec defines the following color-based CSS properties:
1. [`column-rule-color`](https://drafts.csswg.org/css-gaps-1/#propdef-column-rule-color)
2. [`row-rule-color`](https://drafts.csswg.org/css-gaps-1/#propdef-row-rule-color)
3. [`rule-color`](https://drafts.csswg.org/css-gaps-1/#propdef-rule-color) (which is a shorthand for the first two)

`column-rule-color` is already [noted as being handled by Forced Colors Mode](https://drafts.csswg.org/css-color-adjust-1/#forced-colors-properties), so the proposal would be to add `row-rule-color` and `rule-color` to the list of properties affected by Forced Colors Mode, as well.

CC @kbabbitt

svgeesus:
That seems like an obvious and correct change.

tabatkins:
(Apologies for having had to miss this meeting, and not seeing the resolutions until now.)

I don't understand the third resolution. Forced Colors Mode has obvious, author-visible effects on the properties it affects. This set of properties needs to be interoperable between browsers; thus, it needs to be specified in some way. We do not currently have a definition anywhere else of what "color properties" are that we can lean on. This means that moving the existing list to a note just... removes necessary normative text.

We do that "list in a note" thing when it's helpful to list out a category that is defined elsewhere (especially when it's *diffused*, like each property defines itself as being in the category or not). But in this case the list *is* the canonical definition.

We can add a note about the listed set being something that will grow over time, and thus possibly not being *complete* at any given time, but the goal *should* be to keep it complete. Regardless, the current list is a required bit of normative definition, so I object to the resolution as it stands.

svgeesus:
> We do not currently have a definition anywhere else of what "color properties" are that we can lean on. This means that moving the existing list to a note just... removes necessary normative text.

I agree, and found that resolution to be odd. (The PR puts this in a note, in accordance with the WG resolution)

kbabbitt:
As I recall, part of the discussion was: Instead of the normative part being a list that would need ongoing maintenance, state that all `<color>` components of properties must be adjusted when Forced Colors Mode is enabled unless indicated otherwise. That part was, admittedly, not captured in the resolution.

(If there's a reason we really do need a list, then I'd be fine with that - just trying to complete the picture behind that resolution.)

[...]

PR #12166, opened 2025-05-05
Update to color properties in Forced Colors Mode

alisonmaher:
[css-color-adjust-1] This change makes the updates per resolutions taken in https://github.com/w3c/csswg-drafts/issues/11857.

More specifically, it moves the color properties that forced colors applies to within a note, mentions that this is the list at the time of writing and may not be exhaustive as other color properties are added over time, and notes that forced colors mode applies to any shorthand properties, as well.

It updates normative text referencing color properties to note that forced colors applies to the `<color>` components of all properties.

This change also updates `column-rule-color` to `rule-color` to ensure that full support exits for the new color properties added in the Gaps specification.

It adds `accent-color` to the list of color properties (it is mentioned lower down in the algorithm (along with `scrollbar-color`) but it wasn't mentioned in the color properties list).

I also updated some minor inconsistencies in using "computed" vs "computes" within the forced color mode rules.

Lastly, I updated the order of the color properties to be in alphabetical order.

The updates can be found hosted at https://alisonmaher.github.io/alisonmaher/css-color-adjust-1/Overview.html#forced-colors-properties.

### DIFF
--- a/css-color-adjust-1/Overview.bs
+++ b/css-color-adjust-1/Overview.bs
@@ -803,7 +803,8 @@ Properties Affected by Forced Colors Mode {#forced-colors-properties}
 
 	When <a>forced colors mode</a> is active
 	and 'forced-color-adjust' is ''forced-color-adjust/auto'' (see below) on an element,
-	colors on the element are force-adjusted to the user’s preferred color palette.
+	the <<color>> components of all properties on the element
+	are force-adjusted to the user’s preferred color palette.
 
 	<div class="note">
 		Specifically, <a>forced colors mode</a> applies to the following color properties,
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
LGTM

tabatkins:
Sorry, r-. Nothing wrong with the edits on their face, I just disagree with one of the resolutions this is implementing, and wrote my objection in #11857.

tabatkins (changes requested):
r- for now, until we resolve the issue

alisonmaher:
> Sorry, r-. Nothing wrong with the edits on their face, I just disagree with one of the resolutions this is implementing, and wrote my objection in #11857.

@tabatkins 
Updated the text to more clearly state that this applies to `<color>` components of properties, instead of "color properties".

I also noticed a couple of other minor things when making the update that are now fixed in this PR:
1. We used a mix of "computes" and "computed" in one of the lists in this section - I've updated all remaining instances to "computes"
2. `accent-color` was listed as computing to `auto` but wasn't included in the list of properties that are now in the note, so I added it there now, as well

tabatkins (approved):

Issue #8064, opened 2022-11-11
[css-color-adjust-1] Emojis in forced colors mode


alisonmaher:
Currently in [forced colors mode](https://www.w3.org/TR/css-color-adjust-1/#forced-colors-mode), emojis are not adjusted (similarly to images).

For example, when using the following emojis: 🤞😡👍🎉❓🤷‍♂️👆👈🤔🔮🔍👻

Blink and Gecko will produce the following in a dark forced color scheme:
![emoji-current](https://user-images.githubusercontent.com/43683110/201397543-890824a0-9aaf-44de-94f4-ac807c0cdb8c.png)

However, in EdgeHTML and most Windows apps, emojis will fall back to use a monochrome variant of the emoji colored in the Windows High Contrast text color (i.e. `CanvasText`):
![emoji-monochrome](https://user-images.githubusercontent.com/43683110/201398360-a557c56a-2ef5-483a-ba97-dbb1333542c4.png)

The monochrome variant would provide higher levels of contrast and parity with how Windows generally handles emojis in High Contrast. Is this something we should consider recommending or aligning with in the spec?

myfonj:
This is a neat idea. Just two (sad) remarks about status quo from random passers-by:

1. IIUC, Unicode "dictates" only a [subset of Emoji][emoji-variants]([spec], [txt]) to have both Textual and Emoji variants. (This is sad, I wish _every single Emoji_ had mandatory regular monochrome representation alternative.)
2. Actual implementation is currently even worse; it seems that for example Android does not have or does not let you select Text variant for most of those "mandatory" glyphs ([test](https://myfonj.github.io/tst/emoji-variation-sequences-render-test-compact.html)):

<img alt="Google Chrome" width="200" src="https://github.com/w3c/csswg-drafts/assets/1761395/7a43e26b-1150-42e6-a356-37b3a569d9fa"><img alt="Firefox Nightly" width="200" src="https://github.com/w3c/csswg-drafts/assets/1761395/7f047c25-70cc-4138-a748-6a924c3f7ffd">

[emoji-variants]: https://www.unicode.org/emoji/charts/emoji-variants.html
[spec]: https://www.unicode.org/reports/tr51/
[txt]: https://www.unicode.org/Public/15.0.0/ucd/emoji/emoji-variation-sequences.txt

myfonj:
There are quite promising statements in recent Microsoft announcement about Segoe UI in Windows 11 regarding monochrome Emoji variant:

[...]

PR #12173, opened 2025-05-07
[css-color-adjust-1] Emojis in forced colors mode

alisonmaher:
This change adds text to the Color Adjust spec per resolution in https://github.com/w3c/csswg-drafts/issues/8064. This change notes that UAs should force emojis to their monochrome variant depending on the value of `font-variant-emoji` and if a monochrome variant exists.

Updated spec text can be found hosted at https://alisonmaher.github.io/alisonmaher/css-color-adjust-1/Overview.html#forced-colors-properties


svgeesus (approved):
LGTM

PR #12175, opened 2025-05-08
[css-borders-4] Define border/shadow rendering for `corner-shape`

### LINKED ISSUE #11610: [css-borders-4] Specify how borders are rendered for `corner-shape`
Follow-up on #10993 

`corner-shape` cannot be naturally stroked in the same way as ordinary rounded corners or a stroke path, and some details need to be defined.

For example, how should stroking bevel/notch look like? Do the inner border collapse into each other, or stay confined within the corner?

### LINKED ISSUE #11767: [css-borders-4] specify how corner-shape affects shadows, outline etc

noamr:
Based on [this resolution](https://github.com/w3c/csswg-drafts/issues/11610#issuecomment-2769797194)

- Borders are rendered perpendicular to the outer curve, clamped at corner-shape -1 or 1.
- Shadows are rendered like a scaled version of the original shape, converting the shadow spread to a scale.

Added detailed algorithms, explanation, and a chart.

Closes #11610 
Closes #11767

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -375,6 +375,9 @@ To compute the <dfn>adjusted untrimmed inner top-left corner starting point</dfn
 	1. Let |w| be the [=used value=] of the element's 'border-left-width'.
 	1. If |s| is equal or greater than 1, return <code>|w|, 0</code>.
 	1. If |s| is equal or less than -1, return <code>0, -|w|</code>.
+
+		Note: Because of the above clamping, when the shape is more convex than a ''corner-shape/round'' or more concave than a ''corner-shape-scoop'',
+		the inner curve cannot would appear to have a "belly" towards the middle, as it cannot remain constant while maintaining the same curvature.
 	1. Let |k| be <code>0.5<sup>|s|</sup></code>.
 	1. Let |normalizedHalfCorner| be <code>0.5<sup>|k|</sup></code>.
 	1. Let |offset| be <code>2 - &radic;2</code>.
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -428,7 +428,7 @@ To compute the <dfn>corner path</dfn> given a rectangle |cornerRect|, a rectangl
 		1. Swap between |startThickness| and |endThickness|.
 		1. Set |orientation| to (|orientation| + 2) % 4.
 	1. Let |cornerPath| be a path that begins at <code>(0, 1)</code>.
-	1. Switch on|curvature|:
+	1. Switch on |curvature|:
 		<dl class=switch>
 			: 0
 			:: Extend |cornerPath| by adding a straight line to <code>(1, 0)</code>.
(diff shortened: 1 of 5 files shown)


css-borders-4/Overview.bs
```diff
@@ -361,7 +361,77 @@ as well as account for sharp edges and overlaps.
...
+	1. Let |adjustedY| be <code>max(0, 2 * (1 - |normalizedHalfCorner|) - |offset|)</code>.
+	1. Return <code>|adjustedX|, |adjustedY|</code>.
+</div>
+
+<div algorithm="corner-shape-path">
+To compute the inner path of a corner:
+
+	1. Let |adjustedX|, |adjustedY| be the result of the [=adjusted untrimmed inner top-left corner starting point=] steps,
+		modified to consider the corresponding 'border-width' and 'corner-shape' values at the start and end of a path`.
+	1. Create a path based on the [=canonical superellipse formula=], with the outer corner as the center, starting from |adjustedX| and ending at |adjustedY|.
```
tabatkins, line 394: I don't really understand what this is saying. I'm particularly confused by what the "starting from... ending from..." part means; it doesn't match what I thought those variables were for. They're x and y coordinates of a point; how do you "start" and "end" at them individually?

css-borders-4/Overview.bs
```diff
@@ -361,7 +361,77 @@ as well as account for sharp edges and overlaps.
...
+		User agents MAY create an approximation of this path rather than try to follow the formula precisely, for performance reasons.
+	1. Trim the resulting path so that it starts at a point that has a 'border-width' distance from the outer.
```
tabatkins, line 396: Also not sure what this is actually telling me to do. :(

css-borders-4/Overview.bs
```diff
@@ -361,7 +361,77 @@ as well as account for sharp edges and overlaps.
...
+
+<div class=note>
+The superellipse quadrant is scaled so that:
+	* Its center stays at the outer corner.
+	* Its start and end point are perpendicular to the outer starting point.
+	* Its distance from the outer curve is the corresponding 'border-width'.
+
+This is illustrated in the following chart:
+<figure>
+	<img src="images/corner-shape-border.svg"
```
tabatkins, line 407: This could use some details of where the points/etc calculated by the above algorithms are. 

Is the hull calculated in the above algorithms in some way?

tabatkins (changes requested):
Specific comments scattered in. Generally, tho, I can't actually follow the algorithm and am confused by several details, and it handwaves several operations that might be obvious when you're "in the code", but aren't clear from my perspective as a relative outsider.

noamr:
> Specific comments scattered in. Generally, tho, I can't actually follow the algorithm and am confused by several details, and it handwaves several operations that might be obvious when you're "in the code", but aren't clear from my perspective as a relative outsider.

Fair! I redid the whole thing with more explanations and a rigorous algorithm. PTAL!

tabatkins:
I'm in the middle of an editorial pass over this, but I'll merge what's here; it looks better now! (I won't be able to get back to finish the edits until Thursday.)

PR #12177, opened 2025-05-09
[cssom-view-1] Define HTMLElement.scrollParent

flackr:
Define the scrollParent of an element following the discussion and resolution in #1522.


tabatkins:
Hm, it's slightly unfortunate that a `null` can mean *either* that the scroll parent is the root scroller, *or* that the scroll parent is hidden in a closed shadow. The former can still be responded to, by scrolling the root, the latter can't be, and I'm not sure there's an easy way to tell them apart.

Is that okay? Or do we want to be able to distinguish those cases?

flackr:
> The former can still be responded to, by scrolling the root, the latter can't be, and I'm not sure there's an easy way to tell them apart.

Right, I could see a good argument for skipping the hidden parent. WDYT of just having it keep walking up the tree in that case? I think you're right that scrolling the next scroller up should still provide reasonable behavior.

tabatkins (approved):

PR #12187, opened 2025-05-13
[cssom-1] Correct typos and grammar in "serialize a CSS declaration block"

AtkinsSJ:
I noticed the "that that" mistake while implementing this, then identified a couple of other small corrections. Not all of these are "wrong" exactly, they're just not worded the way the specs usually do.

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2353,20 +2353,20 @@ To <dfn export>serialize a CSS declaration block</dfn> <var>declaration block</v
    <li><i>Shorthand loop</i>: For each <var>shorthand</var> in <var>shorthands</var>, follow these substeps:
     <ol>
      <li>Let <var>longhands</var> be an array consisting of all <a>CSS declarations</a> in <var>declaration block</var>'s
-     <a for="CSSStyleDeclaration">declarations</a> that that are not in <var>already serialized</var> and have a
+     <a for="CSSStyleDeclaration">declarations</a> that are not in <var>already serialized</var> and have a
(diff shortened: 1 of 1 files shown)


cssom-1/Overview.bs
```diff
@@ -2353,20 +2353,20 @@ To <dfn export>serialize a CSS declaration block</dfn> <var>declaration block</v
    <li><i>Shorthand loop</i>: For each <var>shorthand</var> in <var>shorthands</var>, follow these substeps:
     <ol>
      <li>Let <var>longhands</var> be an array consisting of all <a>CSS declarations</a> in <var>declaration block</var>'s
-     <a for="CSSStyleDeclaration">declarations</a> that that are not in <var>already serialized</var> and have a
+     <a for="CSSStyleDeclaration">declarations</a> that are not in <var>already serialized</var> and have a
      <a for="CSS declaration">property name</a> that
      maps to one of the shorthand properties in <var>shorthands</var>.
```
AtkinsSJ, line 2358: Should this be "that maps to shorthand"? The loop is only concerned about that, not the entire list of possible shorthands.

AtkinsSJ: I've pushed a commit to correct this.

AtkinsSJ: Actually I've undone that commit. I don't understand why we have both `longhands` and `current longhands`. `longhands` only gets used to then copy into `current longhands`. Maybe it's supposed to only be created once instead of each time in `shorthand loop`?

svgeesus:
@emilio for review

PR #12189, opened 2025-05-13
[css-lists-3] Correct markup for nested lists example

manuel-za:
Non-substantive contribution.

Before:
- Markup for css-lists-3/Example 19 (nested lists) does not render the expected layout
- The body of the line item "Third second-level item, list start=3" indicates the intention of the authors

After:
- Minor correction to the markup matches likely author intent


SebastianZ:
Thanks for the fix, @manuel-za!

Sebastian

Issue #12031, opened 2025-03-31
[css-highlight-api] Should highlightsFromPoint also return ranges under the hit point?


ffiori:
Many use cases for custom highlights that would benefit from `highlightsFromPoint` also need to determine which ranges are hit from the returned highlights, generally those that have many ranges associated to each highlight.
 
For example,
- a spellchecker might want to know exactly which word is being clicked/hovered to give suggestions,
- a code editor may have multiple instances of a variable or keyword highlighted with the same highlight and possibly want to emphasize only the one that the user is interacting with,
- a custom find-on-page might want to emphasize one specific match that the user is clicking when there could be many of them in the webpage.
 
highlightsFromPoint is currently defined as follows in the [spec](https://drafts.csswg.org/css-highlight-api-1/#dom-highlightregistry-highlightsfrompoint):
```
partial interface HighlightRegistry {
  sequence<Highlight> highlightsFromPoint(float x, float y, optional HighlightsFromPointOptions options = {});
}
```
 
So, implementations that use highlightsFromPoint as above then need some extra computation to identify which range was actually hit, using some workaround with caretPositionFromPoint() or calling getClientRects() on Ranges that are built from the AbstractRanges contained by the highlights hit. This results in more complex code and potentially worse performance than if the ranges were already returned by the API.
 
One possible definition for highlightsFromPoint that would help with this issue could be written as follows:
```
partial interface HighlightRegistry {
  sequence<HighlightHitResult> highlightsFromPoint(float x, float y, optional HighlightsFromPointOptions options = {});
};
interface HighlightHitResult {
  Highlight highlight;
  sequence<AbstractRange> ranges; // all of the ranges that were hit for that particular highlight
}
```
 
Otherwise, if we keep highlightsFromPoint as is, a function like intersects(x,y) could be added to Highlight, returning a list of ranges in the highlight that intersect the given coordinates, but maybe worth to return both highlights and ranges with the same highlightsFromPoint API for the sake of simplicity.
 
Interested to hear more thoughts on this. This issue was also brought up before by @schenney-chromium in the [highlightsFromPoint I2S](https://groups.google.com/a/chromium.org/g/blink-dev/c/8OpfR1m_zjU/m/4nr01-gBAQAJ).

ffiori:
PR https://github.com/w3c/csswg-drafts/pull/12215 resolves this issue.

Issue #12045, opened 2025-04-03
[css-highlight-api] highlightsFromPoint(): Inconsistency between spec and test regarding shadow DOM


jnjaeschke:
There is a test which deals with a [live range that starts and ends in light DOM, but has a shadow DOM part inside](https://github.com/web-platform-tests/wpt/blob/dceefcbf785cc73900a88bda459a0edd2eb98a2d/shadow-dom/HighlightRegistry-highlightsFromPoint.html#L81).

The test [asserts](https://github.com/web-platform-tests/wpt/blob/dceefcbf785cc73900a88bda459a0edd2eb98a2d/shadow-dom/HighlightRegistry-highlightsFromPoint.html#L119-L122) that `highlightsFromPoint()` does not return anything when being called over the shadow DOM part.

However, the spec says that [`highlightsFromPoint()` should return a Highlight if _any_ of the DOMRects returned by `getClientRects()` of _any_ range of the highlight intersects the point](https://drafts.csswg.org/css-highlight-api-1/#dom-highlightregistry-highlightsfrompoint:~:text=The%20coordinates%20x%2Cy%20fall%20inside%20at%20least%20one%20of%20the%20DOMRects%20returned%20by%20calling%20getClientRects%28%29%20on%20range%2E).

Both Gecko and Chrome also return DOMRects for the shadow DOM part, therefore the implementation would include the shadow DOM parts as well.

cc @ffiori @emilio @smaug----

ffiori:
PR https://github.com/w3c/csswg-drafts/pull/12215 resolves this issue, let me know if there's any inconsistencies left.

Issue #12003, opened 2025-03-25
[css-highlight-api] HighlightsFromPoint: Not dealing with invalid static ranges


jnjaeschke:
The spec creates [temporary ranges for `StaticRange`s](https://drafts.csswg.org/css-highlight-api-1/#dom-highlightregistry-highlightsfrompoint:~:text=Let%20range%20be%20a%20Range%20object%20whose%20start%20node%20and%20end%20node%20are%20set%20to%20abstractRange%E2%80%99s%20start%20node%20and%20end%20node%20respectively%2C%20and%20start%20offset%20and%20end%20offset%20are%20set%20to%20abstractRange%E2%80%99s%20start%20offset%20and%20end%20offset%20respectively%2E) because `StaticRange`s currently don't support `getClientRect()`. However, this step assumes that creating a temp range succeeds, and also does not cover the case where an underlying `StaticRange` may be [invalid](https://dom.spec.whatwg.org/#staticrange-valid).

cc @ffiori @stephanieyzhang @smaug----

ffiori:
PR https://github.com/w3c/csswg-drafts/pull/12215 resolves this issue, let me know if there's any inconsistencies left.

Issue #12002, opened 2025-03-25
[css-highlight-api] highlightsFromPoint(): Clarify ordering for same-priority highlights


jnjaeschke:
The [spec](https://drafts.csswg.org/css-highlight-api-1/#dom-highlightregistry-highlightsfrompoint:~:text=descending%20order%20of%20priority%2C) currently mentions that highlights returned from HighlightsFromPoint() are returned by priority in descending order, but don't mention ordering for highlights with the same priority. The [tests](https://github.com/web-platform-tests/wpt/blob/3863a5a4ca3f98ce6ad39a88d929207d21dbf5a1/css/css-highlight-api/HighlightRegistry-highlightsFromPoint.html#L72-L73) imply that in this case the highlights should be ordered by insertion order descending. 

cc @ffiori @stephanieyzhang @smaug----

ffiori:
Hi @jnjaeschke, @schenney-chromium, thanks for your comments. From the spec for custom highlights, in the section that is linked to in highlightsFromPoint spec, [Priority of Overlapping Highlights](https://drafts.csswg.org/css-highlight-api-1/#priorities), it's specified which highlight has a higher effective priority if they're tied on their attribute value:

"When two or more [custom highlights](https://drafts.csswg.org/css-highlight-api-1/#custom-highlight) have the same numerical priority, the one most recently [registered](https://drafts.csswg.org/css-highlight-api-1/#registered) has the higher effective [priority](https://drafts.csswg.org/css-highlight-api-1/#priority)."

PR #12215, opened 2025-05-20
[css-highlight-api] Change highlightsFromPoint return type #12031

ffiori:
[css-highlight-api] Change highlightsFromPoint return type #12031

As discussed and resolved in #12031, this PR changes the return type for highlightsFromPoint to HighlightHitResult to include the ranges that are hit at a given point in addition to the custom highlights.

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -586,9 +586,9 @@ Range Updating and Invalidation</h3>
 Interacting with Custom Highlights</h2>
 
 The {{highlightsFromPoint}}(<var>x</var>, <var>y</var>, <var>options</var>) method allows developers to build scenarios
-involving user interaction with [=custom highlights=]. The method returns a [=sequence=] containing the [=custom highlights=]
-at a given <var>x</var>, <var>y</var> coordinate. The [=custom highlights=] are listed in this [=sequence=] in
-descending [=priority=] order.
+involving user interaction with [=custom highlights=]. The method returns a [=sequence=] containing {{HighlightHitResult}} objects
+which encapsulate the [=custom highlights=] and their [=ranges=] that are hit at a given <var>x</var>, <var>y</var> coordinate.
+This sequence is ordered in [=priority|reverse paint order=] of its {{HighlightHitResult}}'s [=custom highlights|highlights=].
 By default, [=custom highlights=] in a [=shadow tree=] are not returned, but the developer has the possibility to pass in
 an optional <var>options</var> [=dictionary=] with a <var>shadowRoots</var> property containing a [=sequence=] of {{ShadowRoot}}
 objects. [=custom highlights|Highlights=] contained within a [=shadow tree=] provided in this way will be returned.
@@ -605,7 +605,7 @@ objects. [=custom highlights|Highlights=] contained within a [=shadow tree=] pro
 					color:red;
 				}
 			</style>
-			<body>abc
+			<body>abcd
 			<script>
 				document.addEventListener('click', (event) => {
 					const mouseX = event.clientX;
@@ -620,8 +620,11 @@ objects. [=custom highlights|Highlights=] contained within a [=shadow tree=] pro
 				let r2 = new Range();
 				r2.setStart(textNode, 1);
 				r2.setEnd(textNode, 2);
+				let r3 = new Range();
+				r3.setStart(textNode, 3);
+				r3.setEnd(textNode, 4);
 
-				let h1 = new Highlight(r1);
+				let h1 = new Highlight(r1, r3);
 				let h2 = new Highlight(r2);
 				h1.priority = 1;
 				h2.priority = 2;
@@ -632,18 +635,19 @@ objects. [=custom highlights|Highlights=] contained within a [=shadow tree=] pro
 		</xmp>
 
 		The code above will display the following styled text, note that "b" is affected by both [=custom highlight|highlights=]
-		<var>h1</var> and <var>h2</var>, whereas "a" is only affected by <var>h1</var>:
+		<var>h1</var> and <var>h2</var>, whereas "a" and "d" are only affected by <var>h1</var>:
 
 		<div class=sample-out style="color:black">
-			<span style="background:yellow;">a</span><span style="background:yellow;color:red;">b</span><span>c</span>
+			<span style="background:yellow;">a</span><span style="background:yellow;color:red;">b</span><span>c</span><span style="background:yellow;">d</span>
 		</div>
 
 		In this example there's an [=event listener=] set on click [=event|events=] that logs the [=custom highlights=]
-		present at the point where the click was made.
+		and their ranges present at the point where the click was made.
 		The following [=sequence|sequences=] are some examples of what will be printed to console after a click:
-			* <code>[ <var>h1</var> ]</code>, if the user clicks on character "a".
-			* <code>[ <var>h2</var>, <var>h1</var> ]</code>, if the user clicks on character "b",
-				as <var>h2</var> has higher priority than <var>h1</var>.
+			* <code>[ HighlightHitResult {highlight: <var>h1</var>, ranges: [<var>r1</var>]} ]</code>, if the user clicks on character "a".
+				Note that only <var>r1</var> is included in the {{HighlightHitResult}} returned since that's the only range in <var>h1</var> that was hit.
+			* <code>[ HighlightHitResult {highlight: <var>h2</var>, ranges: [<var>r2</var>]}, HighlightHitResult {highlight: <var>h1</var>, ranges: [<var>r1</var>]} ]</code>,
+				if the user clicks on character "b", as <var>h2</var> has higher priority than <var>h1</var>.
 			* <code>[]</code>, if the user clicks on character "c".
 	</div>
 
@@ -651,7 +655,12 @@ The method {{highlightsFromPoint}} is defined as part of the {{HighlightRegistry
 
 <pre class=idl>
 partial interface HighlightRegistry {
-	sequence&lt;Highlight> highlightsFromPoint(float x, float y, optional <span>HighlightsFromPointOptions</span> options = {});
+	sequence&lt;HighlightHitResult> highlightsFromPoint(float x, float y, optional <span>HighlightsFromPointOptions</span> options = {});
+};
+
+interface <dfn interface>HighlightHitResult</dfn> {
+	attribute Highlight highlight;
+	attribute sequence&lt;AbstractRange> ranges;
 };
 
 dictionary <dfn dictionary>HighlightsFromPointOptions</dfn> {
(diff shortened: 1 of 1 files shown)


css-highlight-api-1/Overview.bs
```diff
@@ -665,23 +675,29 @@ method must return the result of running these steps:
...
-	* <var>y</var> is greater than the <a>viewport</a> height excluding the size of a rendered scroll bar (if any)
-1. Otherwise, return a [=sequence=] of [=custom highlights=] given by ordering the highlights contained in this {{HighlightRegistry}} in descending order of [=priority=],
-	including only those highlights that contain at least one {{AbstractRange}} <var>abstractRange</var> that satisfies the following:
+	* <var>x</var> is greater than the [=viewport=] width excluding the size of a rendered scroll bar (if any)
+	* <var>y</var> is greater than the [=viewport=] height excluding the size of a rendered scroll bar (if any)
+	* The topmost [=box=] in the [=viewport=] in paint order that would be a target for hit testing at coordinates <var>x</var>,<var>y</var> when applying
+		the [=transforms=] that apply to the descendants of the viewport, has an element associated to it that's in a [=shadow tree=] whose
+		[=shadow root=] is not [=list/contains|contained by=] <var>options</var>.<var>shadowRoots</var>.
+1. Otherwise, return a [=sequence=] of {{HighlightHitResult}} objects ordered in [=priority|reverse paint order=] of their <var>highlight</var> attributes,
```
dandclark, line 683: I think this semi-declarative way of defining this is getting somewhat unclear with the added layer of including the ranges. For example it's no longer obvious why a HighlightHitResult for a given Highlight would be included or not, because now the question of whether to include something or not is focused on the ranges. The point that a HighlightHitResult is only included if the resulting ranges array is empty could be lost, it's just kind of implied that `ranges` is non-empty.

The bit about allowing a `Range` object to be constructed from a `StaticRange` and then using the `range` in the next bullet point is also hard to follow, it's like a bit of imperative logic mixed in with the declarative.

I think it's worth trying a more fully imperative way of defining this, a bit closer to how it'd actually be implemented. Basically:
1. Let `result` be an empty list.
1. For each Highlight in our registry:
    1. Create a HighlightHitResult with the given highlight.  
        1. For each Range in the Highlight:
            1. Do the hit test, and add range to HighlightHitResult if it passes.
    1. If the HighlightHitResult has non-empty `ranges`, add it to `result`.
1. Sort `result` by priority of the highlights // or alternatively, the outer `for` loop could iterate the highlights in prority order
1. Return result.

dandclark (commented):
Thanks for working on this!

For https://github.com/w3c/csswg-drafts/issues/12045 https://github.com/w3c/csswg-drafts/issues/12003 https://github.com/w3c/csswg-drafts/issues/12002, can you include a sentence or so in your PR description for each to mention the way that this PR resolves the issue?

css-highlight-api-1/Overview.bs
ffiori: That makes sense, I was using declarative because I thought that was the norm but imperative makes more sense in this context. Let me know your thoughts on this last commit.

css-highlight-api-1/Overview.bs
```diff
@@ -586,9 +587,9 @@ Range Updating and Invalidation</h3>
 Interacting with Custom Highlights</h2>
 
 The {{highlightsFromPoint}}(<var>x</var>, <var>y</var>, <var>options</var>) method allows developers to build scenarios
-involving user interaction with [=custom highlights=]. The method returns a [=sequence=] containing the [=custom highlights=]
-at a given <var>x</var>, <var>y</var> coordinate. The [=custom highlights=] are listed in this [=sequence=] in
-descending [=priority=] order.
+involving user interaction with [=custom highlights=]. The method returns a [=sequence=] containing {{HighlightHitResult}} objects
+which encapsulate the [=custom highlights=] and their [=ranges=] that are hit at a given <var>x</var>, <var>y</var> coordinate.
+This sequence is ordered in [=priority|reverse paint order=] of its {{HighlightHitResult}}'s [=custom highlights|highlights=].
```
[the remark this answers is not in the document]
ffiori: I agree with you, I'm not strongly opposed to using paint order either but I think it's a bit clearer to use the original concept for it.

css-highlight-api-1/Overview.bs
```diff
@@ -680,24 +681,24 @@ method must return the result of running these steps:
...
-		* <var>abstractRange</var> is either a live {{Range}}, or a [=StaticRange/valid=] {{StaticRange}} that allows to successfully construct a {{Range}} object <var>range</var> 
-		    whose [=start node=] and [=end node=] are set to <var>abstractRange</var>'s [=start node=] and [=end node=] respectively,
-			and [=start offset=] and [=end offset=] are set to <var>abstractRange</var>'s [=start offset=] and [=end offset=] respectively.
-
-		* The coordinates <var>x</var>,<var>y</var> fall inside at least one of the {{DOMRect}}s returned by calling {{Range/getClientRects()}} on <var>range</var>.
+1. Otherwise, let <var>results</var> be an empty [=sequence=].
+1. For each {{Highlight}} <var>highlight</var> in this {{HighlightRegistry}}:
+	1. Create a {{HighlightHitResult}} <var>result</var> with the given <var>highlight</var> as <var>result</var>.{{HighlightHitResult/highlight}} attribute.
+	1. For each {{AbstractRange}} <var>abstractRange</var> in <var>highlight</var>:
+		1. If <var>abstractRange</var> is a live {{Range}} or a [=StaticRange/valid=] {{StaticRange}}, construct a {{Range}} object <var>range</var>
```
dandclark, line 688: It's not really clear what the algorithm is supposed to do if abstractRang is not a live range or a valid staticrange. I think we need to [`continue`](https://infra.spec.whatwg.org/#iteration-continue) if it's not, or add another layer of nesting after this `if`.

dandclark: It might also be more understandable to check the condition that `abstractRange` is an invalid staticrange.

css-highlight-api-1/Overview.bs
```diff
@@ -680,24 +681,24 @@ method must return the result of running these steps:
...
-
-		* The coordinates <var>x</var>,<var>y</var> fall inside at least one of the {{DOMRect}}s returned by calling {{Range/getClientRects()}} on <var>range</var>.
+1. Otherwise, let <var>results</var> be an empty [=sequence=].
+1. For each {{Highlight}} <var>highlight</var> in this {{HighlightRegistry}}:
+	1. Create a {{HighlightHitResult}} <var>result</var> with the given <var>highlight</var> as <var>result</var>.{{HighlightHitResult/highlight}} attribute.
+	1. For each {{AbstractRange}} <var>abstractRange</var> in <var>highlight</var>:
+		1. If <var>abstractRange</var> is a live {{Range}} or a [=StaticRange/valid=] {{StaticRange}}, construct a {{Range}} object <var>range</var> 
+				whose [=start node=] and [=end node=] are set to <var>abstractRange</var>'s [=start node=] and [=end node=] respectively,
+				and [=start offset=] and [=end offset=] are set to <var>abstractRange</var>'s [=start offset=] and [=end offset=] respectively.
+		1. If <var>range</var> could be successfully constructed and the coordinates <var>x</var>,<var>y</var> fall inside at least one of the {{DOMRect}}s
```
dandclark, line 691: I think checking after the fact whether the range was "successfully constructed" is a little hand-wavy. We should probably be explicitly checking for thrown exceptions. There are a couple examples of that to look at in the DOM spec; in https://dom.spec.whatwg.org/ search for "catching any exceptions".

dandclark: Also, when would we actually expect this to fail?

ffiori: I think this could throw an exception if there's an abstractRange that corresponds to a Range that was created but then start and end nodes weren't set afterwards. If you look at the [constructor of Range](https://dom.spec.whatwg.org/#dom-range-range), they set start and end to ([current global object](https://html.spec.whatwg.org/multipage/webappapis.html#current-global-object)’s [associated Document](https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window), 0), so then in the algorithm when we try to construct a new Range and set start node to the old range's start node, it'll be a doctype and it'll throw (from the [set start algorithm for Range](https://dom.spec.whatwg.org/#concept-range-bp-set)).

So if this step in highlightsFromPoint throws an exception I would just continue.

dandclark: Wouldn't the old start node be the `Document`, not the `doctype`, in that case?
For example https://codepen.io/daniec/pen/yyNJzEr?editors=1000.

dandclark (approved):
Thanks for all the changes, I think this is ready.
Just one more question on the thread about the exception.

css-highlight-api-1/Overview.bs
ffiori: Oh yes, I confused doctype with Document. I don't think this could throw then, we're covering other problematic cases by checking that it's not an invalid StaticRange. I just removed the exception catching bit. Thanks for the catch.

dandclark:
@frivoal as an editor of https://drafts.csswg.org/css-highlight-api-1/ do you want a chance to take a look at this before we land it?

As a sidenote, I think it would make sense for @ffiori and I to be added as editors if that sounds alright with you. Sanket Joshi is no longer at Microsoft, and Fernando and I have both been involved with this spec for a long time.

ffiori:
Hi @frivoal, friendly pinging on this issue and following up on @dandclark's comment. I'd appreciate if you could PTAL, and also wanted to confirm if it makes sense to add me and Daniel as editors.

Thanks in advance!

ffiori:
Adding @megangardner as reviewer as well as editor of https://drafts.csswg.org/css-highlight-api-1/. Could you PTAL?

We're also thinking of adding @dandclark and me as editors since Sanket Joshi left Microsoft, does that sound good to you too?

Thanks in advance.

leotlee:
Hi @sanketj , you think you can help review this and maybe add Dan and Fernando as editors? We haven't had luck getting responses from an editor here. Appreciate it, thanks.

atanassov:
It's great to see the PR moving forward. Adding to the agenda to decide and formalize the editors.

leotlee (approved):

leotlee:
> Looks good to me! Also I am not against adding you both as editors. @frivoal thoughts?

@frivoal , are you OK with us merging this PR? Also I'd love to get your thoughts on updating editors.

frivoal:
Sorry for the slow answer, I had missed that this conversation was happening. I'm arriving after the battle, but at this point it all looks good. Since we have an outstanding resolution in favor of this change, and multiple people being happy with the details, I'm merging without waiting for the PR to make it to the top of the agenda+ list (though we can mention it on the call when we get to it).

As for adding you as editors, I am certainly in favor.

Issue #11908, opened 2025-03-11
[css-mixins-1] Problems with public representations of CSSOM values


cdoublev:
They are minors so I hope it is okay if I do not create separate issues.

---

[`CSSFunctionRule.name`](https://drafts.csswg.org/css-mixins-1/#dom-cssfunctionrule-name):

  > **`name`, of type CSSOMString, readonly**
  >
  > The name of the custom function.

Shouldn'it be [serialized as an identifier](https://drafts.csswg.org/cssom-1/#serialize-an-identifier)?

`--cust\\ om()` [becomes](https://drafts.csswg.org/css-syntax-3/#consume-ident-like-token) `<function-token>` whose value is `cust om`, and `<)-token>`. Then these tokens [become](https://drafts.csswg.org/css-syntax-3/#consume-function) a function whose `name` is `cust om`.

The same question applies to [`FunctionParameter.name`](https://drafts.csswg.org/css-mixins-1/#dom-functionparameter-name).

---

[`CSSFunctionRule.returnType`](https://drafts.csswg.org/css-mixins-1/#dom-cssfunctionrule-returntype):

  > **`returnType`, of type CSSOMString, readonly**
  >
  > The return type of the custom function, represented as a [syntax string](https://drafts.css-houdini.org/css-properties-values-api-1/#syntax-string). If the custom function has no return type, returns `"type(*)"`.

`CSSFunctionRule.returnsType` may be internally represented as a `<type()>` (the result from parsing [`returns <css-type>`](https://drafts.csswg.org/css-mixins-1/#at-ruledef-function)) unless otherwise specified, and thus would not represent a valid syntax string.

So I would first suggest:

  > The return type of the custom function:
  >
  >   - if it is a `<syntax-component>`, represented as a syntax string
  >   - if it is a `<type()>`, represented as its value as a syntax string
  >   - if it is not defined, as `*`.

Now I do not see why it should be represented as a syntax string rather than the serialization of `<css-type>`. Is it intentional? Plugging it in `@property/syntax` (the only value defined with a syntax `<string>`) does not seem useful. I would have thought that representing a syntax as a `<string>` would become legacy.

The same problem/question applies to [`FunctionParameter.type`](https://drafts.csswg.org/css-mixins-1/#dom-functionparameter-type).

---

[Serialize a function parameter](https://drafts.csswg.org/css-mixins-1/#serialize-a-function-parameter):

[...]

PR #12245, opened 2025-05-28
[css-mixins-1] Consistently return a syntax string

andruud:
Perhaps these functions should indeed return a serialized `<css-type>` (#11908), but until then they should always return a valid syntax string.


andruud:
PTAL. https://github.com/w3c/csswg-drafts/issues/11908 has other suggested changes, but I'm sending this separately since it's an obvious inconsistency.

cdoublev (approved):

PR #12261, opened 2025-06-02
[css-values-4] [css-cascade-4] [css-color-5] [css-fonts-4] [css-images-4] [css-shapes-2] Clean up fetching

### LINKED ISSUE #12065: [css-images] [css-values] Fetch algorithms require a style sheet, which may not exist
["fetch a style resource"](https://drafts.csswg.org/css-values-4/#fetch-a-style-resource) and thus ["fetch an external image for a stylesheet"](https://drafts.csswg.org/css-images-4/#fetch-an-external-image-for-a-stylesheet) require a non-null `CSSStyleSheet`. However, there may not *be* such a sheet. Specifically, presentational hints and style attributes.

 For example:

```html
<body background="image.png">
```

Here the `background` attribute sets `background-image`:
> When a [body](https://html.spec.whatwg.org/multipage/sections.html#the-body-element) element has a [background](https://html.spec.whatwg.org/multipage/obsolete.html#attr-background) attribute set to a non-empty value [...] the user agent is expected to treat the attribute as a [presentational hint](https://html.spec.whatwg.org/multipage/rendering.html#presentational-hints) setting the element's ['background-image'](https://drafts.csswg.org/css-backgrounds/#propdef-background-image) property to the return value.
> https://html.spec.whatwg.org/multipage/rendering.html#the-page:attr-background

As another example, inline styles:

```html
<div style="background-image: url('image.png')">
```

As far as I'm aware, neither of these have a `CSSStyleSheet` object associated with them. Seems like we should probably fall back to using the Document's base URL in these cases.

### LINKED ISSUE #12068: [css-values] "fetch a style resource" should fall back to the CSSStyleSheet's location
In step 2 of ["fetch a style resource"]() we have:

> Let base be sheet’s [stylesheet base URL](https://drafts.csswg.org/cssom-1/#concept-css-style-sheet-stylesheet-base-url) if it is not null, otherwise environmentSettings’s [API base URL](https://html.spec.whatwg.org/multipage/webappapis.html#api-base-url). [[CSSOM]](https://drafts.csswg.org/css-values-4/#biblio-cssom)

However, only stylesheets constructed from JS have a base URL set. For those that don't, I believe we should fall back to sheet's [location](https://drafts.csswg.org/cssom-1/#concept-css-style-sheet-location), and only afterwards use the API base URL. As an example:

## index.html
```html
<!DOCTYPE html>
<style>
     @import url("data/sheet.css");
</style>
<div id="checkerboard"></div>
```

## data/sheet.css
```css
#checkerboard {
    background: url("image.png");
}
```

With the current algorithm, sheet has no base URL, so it would fall back to the API base URL, which is the document's location. It would thus try to load `image.png` instead of `data/image.png`. Inserting the sheet's location as a fallback means that it would load relative to `data/sheet.css`, and so get `data/image.png`.

### LINKED ISSUE #12086: [css-values-5][cssom] How should <url> modifiers be used when fetching a stylesheet from an @import rule?
As far as I can find, stylesheet fetching uses the algorithm in cssom, [fetch a style sheet](https://drafts.csswg.org/cssom/#fetch-a-css-style-sheet) and does not include use of any of CSS Values 4, [fetch a style resource](https://www.w3.org/TR/css-values-4/#fetch-a-style-resource).

As such, for an @import rule, which uses the `<url>` production so can have URL modifiers, its not clear how exactly those modifiers should apply to the fetch.

(cc @noamr, @annevk, @tabatkins, @fantasai)

### LINKED ISSUE #12147: [css-cascade] [css-values] CSS fetch algorithms assume the style sheet always has a location
Cross-post of https://github.com/whatwg/html/issues/11281, for visibility here, as it's a bug from interactions between HTML and CSS specs.

noamr:
This fixes a few cases around fetching external URLs for style resources:
- The base URL is the sheet's base URL if exists, otherwise the sheet's location, otherwise the document base URL.
- The algorithm can take either a rule or declaration block, and derive the correct stylesheet and base URL from that. Updated the calling sites to take that onto account.
- Removed redundant URL parsing in "@import".

Closes #12065
Closes #12068
Closes #12086
Closes #12147

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -1168,7 +1168,7 @@ URL Processing Model</h4>
 	<div algorithm="resolve a style resource URL">
 		To <dfn>resolve a style resource URL</dfn> from a [=/url=] or <<url>> |urlValue|,
 		and a a [=CSS rule=] or a [=CSS declaration=] |cssRuleOrDeclaration|:
-			1. Let |base| be the [=style resource base URL=] given |cssRuleOrDeclaration|..
+			1. Let |base| be the [=style resource base URL=] given |cssRuleOrDeclaration|.
 			1. Return the result of the [=URL parser=] steps with |urlValue|'s [=/url=] and |base|.
 	</div>
 


AtkinsSJ (approved):
Thanks, this looks pretty solid to me.

One thing that I've been alerted to this morning, (see https://github.com/LadybirdBrowser/ladybird/issues/4972) that's related to this, is that the fetch algorithms operate on relative URLs (and expect them to not be resolved yet) but the values spec requires that the computed value of an URL is absolute:

> When a [<url>](https://drafts.csswg.org/css-values-4/#url-value) appears in the computed value of a property, it is resolved to an absolute URL, as described in the preceding paragraph. The computed value of a URL that the [UA](https://drafts.csswg.org/css-2023/#user-agent) cannot resolve to an absolute URL is the specified value.
https://drafts.csswg.org/css-values-4/#relative-urls

That suggests that either we're absolutizing URLs in two different places with potentially different rules, or that the absolutization in fetch is redundant as the URL will already be absolute. (Or I'm missing something.) Sorry for not catching this earlier!

noamr:
> Thanks, this looks pretty solid to me.
> 
> One thing that I've been alerted to this morning, (see [LadybirdBrowser/ladybird#4972](https://github.com/LadybirdBrowser/ladybird/issues/4972)) that's related to this, is that the fetch algorithms operate on relative URLs (and expect them to not be resolved yet) but the values spec requires that the computed value of an URL is absolute:
> 
> > When a  appears in the computed value of a property, it is resolved to an absolute URL, as described in the preceding paragraph. The computed value of a URL that the [UA](https://drafts.csswg.org/css-2023/#user-agent) cannot resolve to an absolute URL is the specified value.
> > https://drafts.csswg.org/css-values-4/#relative-urls
> 
> That suggests that either we're absolutizing URLs in two different places with potentially different rules, or that the absolutization in fetch is redundant as the URL will already be absolute. (Or I'm missing something.) Sorry for not catching this earlier!

Good spotting, thanks to both you and to the Ladybird project for keeping the specs honest!

I've extracted out the "resolve to absolute URL" bit, and referred to it from https://drafts.csswg.org/css-values-4/#relative-urls. IO think it should make a lot more sense now.

css-values-4/Overview.bs
```diff
@@ -1147,32 +1146,53 @@ URL Modifiers</h4>
...
+				1. If |sheet|'s <a spec=cssom>location</a> is not null, return |sheet|'s <a spec=cssom>location</a>.
+
+			1. Return |cssRuleOrDeclaration|'s [=relevant settings object=]'s [=environment settings object/API base URL=].
+
+	</div>
+
+	<div algorithm="resolve a style resource URL">
+		To <dfn>resolve a style resource URL</dfn> from a [=/url=] or <<url>> |urlValue|,
+		and a a [=CSS rule=] or a [=CSS declaration=] |cssRuleOrDeclaration|:
+			1. Let |base| be the [=style resource base URL=] given |cssRuleOrDeclaration|..
```
AtkinsSJ: Nit: `..` at the end here.

AtkinsSJ:
Arg, I missed a thing until I now tried to implement this! :sweat_smile: 

I think this:
> given an [CSS rule](https://drafts.csswg.org/cssom-1/#css-rule) or a [css declaration](https://drafts.csswg.org/cssom-1/#css-declaration) cssRuleOrDeclaration

...is referring to the wrong definition of "CSS declaration". ***style resource base URL*** checks for `CSSStyleDeclaration` and that seems to be the one we want. The naming of these is _unhelpful_.

noamr:
> Arg, I missed a thing until I now tried to implement this! 😅
> 
> I think this:
> 
> > given an [CSS rule](https://drafts.csswg.org/cssom-1/#css-rule) or a [css declaration](https://drafts.csswg.org/cssom-1/#css-declaration) cssRuleOrDeclaration
> 
> ...is referring to the wrong definition of "CSS declaration". _**style resource base URL**_ checks for `CSSStyleDeclaration` and that seems to be the one we want. The naming of these is _unhelpful_.

Ah yes it needs to be a `CSS declaration block`.

css-values-4/Overview.bs
```diff
@@ -1147,32 +1146,53 @@ URL Modifiers</h4>
...
+
+				1. If |sheet|'s <a spec=cssom>location</a> is not null, return |sheet|'s <a spec=cssom>location</a>.
+
+			1. Return |cssRuleOrDeclaration|'s [=relevant settings object=]'s [=environment settings object/API base URL=].
+
+	</div>
+
+	<div algorithm="resolve a style resource URL">
+		To <dfn>resolve a style resource URL</dfn> from a [=/url=] or <<url>> |urlValue|,
+		and a a [=CSS rule=] or a [=CSS declaration=] |cssRuleOrDeclaration|:
```
AtkinsSJ, line 1170: "a a"

noamr: Thanks, fixed

PR #12263, opened 2025-06-03
[css-borders-4] Improve definition of superellipse()

cdoublev:
`superellipse()` is currently used in [`<corner-shape-value>`](https://drafts.csswg.org/css-borders-4/#typedef-corner-shape-value) with an inline value definition that does not link to its definition. Besides, this value definition uses `<number [-∞,∞]>`, which is just `<number>`.

This definition also includes its parameters, which are different than in `<corner-shape-value>`. Its syntax is not properly defined in a production rule.

This PR fixes all these problems.

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -367,8 +367,7 @@ Issue <a href="https://github.com/w3c/csswg-drafts/issues/11610">#11610</a>: che
 'corner-shape' values</h4>
 
 	<pre class=prod>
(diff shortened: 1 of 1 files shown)


tabatkins:
I'm in the middle of an editorial rewrite of this section that also addresses these issues, so I'll close this just to avoid causing more clashes than necessary. Sorry about that. ^_^

cdoublev:
Dont' worry. Thanks.

PR #12277, opened 2025-06-04
[css-view-transitions-2] Define the view-transition-group-children pseudo element

### LINKED ISSUE #11926: [css-view-transitions-2]: Nested ::view-transition-group should have a container pseudo.
One of the common use cases for `view-transition-group: contain` that we're considering is a scroller with items within in and a view transition that shuffles the items within the container. 

This is achieved with something like
```css
.container {
  view-transition-name: clipper;
  view-transition-group: contain;
}
.container > .item {
  view-transition-name: auto;
}
```
In the current spec this builds the following pseudo tree:
```
:root
|-::view-transition
| |-::view-transition-group(clipper)
|   |-::view-transition-image-pair(clipper)
|   | |-::view-transition-old(clipper)
|   | |-::view-transition-new(clipper)
|   |-::view-transition-group(auto)
|   | |-::view-transition-image-pair(auto)
|   |   |-::view-transition-old(auto)
|   |   |-::view-transition-new(auto)
|   |-...
```
Essentially, `::view-transition-group`s for the items are siblings of `::view-transition-image-pair(clipper)`

Now, to achieve the desired effect we need to put a clip on `::view-transition-group(clipper)`. This works well in a limited number of cases. The problem happens once you have a border (with maybe a rounded clip) on the original clipper element. Because the `::view-transition-group(clipper)` element is sized to be the border box of the original element, and the fact that border itself is baked into `::view-transition-{old/new}(clipper)`, we end up in a situation where it isn't trivial to properly clip the nested groups and also not clip the border. The solution is possible, but not ergonomic where one has to copy the border to the group pseudo, and do some sizing/offset magic to account for the fact that we're sized to the border box.

@bramus put together a demo here:
https://codepen.io/bramus/full/EaxvMXG/11b9fcbe8ebf42034110d3b486db037f

One improvement we can make is to introduce a new pseudo element `::view-transition-group-container` (name tbd) that's sized to the padding box of the original element (I think that's the right clipping box or is it content box?) and hosts the nested groups:

```
:root
|-::view-transition
| |-::view-transition-group(clipper)
|   |-::view-transition-image-pair(clipper)
|   | |-::view-transition-old(clipper)
|   | |-::view-transition-new(clipper)
|   |-::view-transition-group-container(clipper)
|   | |-::view-transition-group(auto)
|   | | |-::view-transition-image-pair(auto)
|   | |   |-::view-transition-old(auto)
|   | |   |-::view-transition-new(auto)
|   | |-...
```

With that, adding a clip to `::view-transition-group-container(clipper)` does not clip the captured box decorations of the original element, and there's less math required.

Cons:
- The clip won't be trivial with corner shape changes
- This clip won't clip other captured content from the original element that is not itself view transition named
- There are likely other issues, but @bramus's demo has an example of what this may look like and it *seems* a bit more ergonomic.

/cc @noamr @nt1m @bramus

noamr:
This new pseudo is an ancestor to the other nested groups, and is positioned after the group's view-transition-image-pair which contains the group's contents.

Resolution:
https://github.com/w3c/csswg-drafts/issues/11926#issuecomment-2916977161

Closes #11926

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -874,13 +874,13 @@ The pseudo-element tree for this would be as follows:
 ```
 	::view-transition
 	├─ ::view-transition-group(container)
-	│  └─ ::view-transition-image-pair(container)
-	│     ├─ ::view-transition-old(container)
-	│     └─ ::view-transition-new(container)
+	│  ├─ ::view-transition-image-pair(container)
+	│  |  ├─ ::view-transition-old(container)
+	│  |  └─ ::view-transition-new(container)
 	│  └─ ::view-transition-group-children(container)
-	│     └─ ::view-transition-group(article)
-	│        ├─ ::view-transition-old(article)
-	│        └─ ::view-transition-new(article)
+	│     ├─ ::view-transition-group(article)
+	│     |  ├─ ::view-transition-old(article)
+	│     |  └─ ::view-transition-new(article)
 	│     └─ ...other nested groups...
 	└─ …other groups…
 ```
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -860,7 +860,7 @@ This example creates a transition where the [=view transition tree=] is nested i
 }
 
 .container,
-::view-transition-group(container) {
+::view-transition-group-children(container) {
 	clip-path: circle();
 }
 
@@ -870,7 +870,22 @@ article {
 }
 ```
 
-By applying the ''clip-path'' to both the containing element and its generated pseudo-element, we preserve the clip during the transition,
+The pseudo-element tree for this would be as follows:
+```
+	::view-transition
+	├─ ::view-transition-group(container)
+	│  └─ ::view-transition-image-pair(container)
+	│     ├─ ::view-transition-old(container)
+	│     └─ ::view-transition-new(container)
+	│  └─ ::view-transition-group-children(container)
+	│     └─ ::view-transition-group(article)
+	│        ├─ ::view-transition-old(article)
+	│        └─ ::view-transition-new(article)
+	│     └─ ...other nested groups...
+	└─ …other groups…
+```
+
+By applying the ''clip-path'' to both the containing element and its generated ''::view-transition-group-children()'' pseudo-element, we preserve the clip during the transition,
 and by applying ''view-transition-group'' to the internal element referencing the container, we make the tree "nested" in a way that would apply this clipping.
 </main>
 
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -870,7 +870,22 @@ article {
...
+	├─ ::view-transition-group(container)
+	│  └─ ::view-transition-image-pair(container)
+	│     ├─ ::view-transition-old(container)
+	│     └─ ::view-transition-new(container)
+	│  └─ ::view-transition-group-children(container)
+	│     └─ ::view-transition-group(article)
+	│        ├─ ::view-transition-old(article)
+	│        └─ ::view-transition-new(article)
+	│     └─ ...other nested groups...
+	└─ …other groups…
```
vmpstr:
```suggestion
	::view-transition
	├─ ::view-transition-group(container)
	│  ├─ ::view-transition-image-pair(container)
	│  |  ├─ ::view-transition-old(container)
	│  |  └─ ::view-transition-new(container)
	│  └─ ::view-transition-group-children(container)
	│     ├─ ::view-transition-group(article)
	│     |  ├─ ::view-transition-old(article)
	│     |  └─ ::view-transition-new(article)
	│     └─ ...other nested groups...
	└─ …other groups…
```

css-view-transitions-2/Overview.bs
```diff
@@ -889,7 +904,21 @@ and by applying ''view-transition-group'' to the internal element referencing th
 	The 'view-transition-group' property can be used in conjuction with 'view-transition-name' to generate a hierarchy of [=named view transition pseudo-element=].
 
 	The [=used value=] for 'view-transition-group' resolves to a 'view-transition-name' in its ancestor chain, or to ''view-transition-name/none''. When generating the [=named view transition pseudo-element=], the ''::view-transition-group()'' with that name
-	would be the parent of the ''::view-transition-group()'' generated for this element's 'view-transition-name'.
+	would create a ''::view-transition-children()'' pseudo-element, which would be the parent of the ''::view-transition-group()'' generated for this element's 'view-transition-name'.
```
vmpstr:
```suggestion
	would create a ''::view-transition-group-children()'' pseudo-element, which would be the parent of the ''::view-transition-group()'' generated for this element's 'view-transition-name'.
```

css-view-transitions-2/Overview.bs
```diff
@@ -889,7 +904,21 @@ and by applying ''view-transition-group'' to the internal element referencing th
...
+
+### View Transition Tree Root: the ''::view-transition-group-children'' pseudo-element ### {#view-transition-group-children-pseudo}
+
+	The <dfn>::view-transition-group-children()</dfn> [=pseudo-element=]
+	is a [=named view transition pseudo-element=]
+	that represents a matching named [=view transition=] capture.
+	A ''::view-transition-group-children()'' [=pseudo-element=]
+	is generated based on the 'view-transition-group' property,
+	as a [=tree/child=] of the corresponding ''::view-transition-group'' [=pseudo-element=],
```
vmpstr:
```suggestion
	as a [=tree/child=] of the corresponding ''::view-transition-group()'' [=pseudo-element=],
```

css-view-transitions-2/Overview.bs
```diff
@@ -889,7 +904,21 @@ and by applying ''view-transition-group'' to the internal element referencing th
...
+	and contains one or more ''::view-transition-group()'' elements, corresponding to the nested groups.
+	See the [=used group name=] and related algorithms.
+
+	The ''::view-transition-group-children()'' [=pseudo-element=] is positioned after its sibling ''view-transition-image-pair''.
```
vmpstr:
```suggestion
	The ''::view-transition-group-children()'' [=pseudo-element=] is positioned after its sibling ''view-transition-image-pair()''.
```

vmpstr (commented):
We also need to add somewhere the fact that we want `border-style: solid` and `border-color: transparent` in the UA stylesheet for the group-children.

That and I'm not sure if it's understood that these should be abspos inset 0 as well?

noamr:
> We also need to add somewhere the fact that we want `border-style: solid` and `border-color: transparent` in the UA stylesheet for the group-children.
> 
> That and I'm not sure if it's understood that these should be abspos inset 0 as well?

Yea, they do.

noamr:
> > We also need to add somewhere the fact that we want `border-style: solid` and `border-color: transparent` in the UA stylesheet for the group-children.
> > That and I'm not sure if it's understood that these should be abspos inset 0 as well?
> 
> Yea, they do.

Done

vmpstr (approved):

PR #12283, opened 2025-06-05
[pointer-animations-1] Prefix `pointer()` definition types

tidoust:
Type definitions share a common namespace and the `<axis>` type is already defined in Scroll-driven Animations:
https://drafts.csswg.org/scroll-animations-1/#typedef-axis

This update prefixes the `<axis>` type (and `<source>` for consistency) with `pointer-` to disambiguate. Feel free to adjust the name as you see fit!

Note: As in Scroll-driven Animations, the spec defines the types with a `noexport` attribute. That seems fine to avoid accidental references from other modules. If that is also meant to create a module-specific namespace for type definitions, we may have a problem as editing tools and validators that I'm aware of and that leverage CSS syntax definitions do not expect that to be possible ;)

### DIFF
--- a/pointer-animations-1/Overview.bs
+++ b/pointer-animations-1/Overview.bs
@@ -222,9 +222,9 @@ spec:selectors-4; type:dfn; text:selector
 	Its syntax is
 
 	<pre class="prod">
-		<<pointer()>> = pointer( [ <<source>> || <<axis>> ]? )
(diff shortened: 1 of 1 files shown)


tidoust:
> If I recall correctly I meant to reuse axis from Scroll-driven Animations

Up to you :) The `noexport` attribute of the `<axis>` definition would just need to be dropped in Scroll-driven Animations.

PR #12286, opened 2025-06-05
[css-values-4] Fetching: use declaration block instead of declaration

noamr:


noamr:
@AtkinsSJ

AtkinsSJ:
Thank you! I'm loving this quick feedback loop.

PR #12305, opened 2025-06-09
[css-values-5] Typo

cdoublev:
The current [definition of the generic interpolation notation](https://drafts.csswg.org/css-values-5/#interpolation-syntax) requires separating interpolation stops by a whitespace:

`[ [ <easing-function> || <segment-options> , ]? <input-position>{1,2} : <output-value> ]*`

The comma should also be hoisted between the optional stop options and the stop.


cdoublev:
I also committed a change that prevents re-ordering the first argument, which is an obvious oversight. 

But please make appropriate corrections as you see fit.

cdoublev:
Replaced a remnant of `<progress>` with `<progress-source>`, which is also an obvious oversight.

PR #12310, opened 2025-06-10
[css-values-5] Allow re-ordering calc-mix() item values

cdoublev:
All functions in the `*-mix()` family allow declaring `<percentage>` before the mixed item value, except `calc-mix()`.


tabatkins:
Yes, the lack of reordering is intentional, as otherwise `<calc-sum>` and `<percentage>` are ambiguous with each other.

Issue #12119, opened 2025-04-25
[css-animations-2][web-animations-2] How should AnimationTrigger work?


DavMila:
To support scroll triggered animations, [`animation-trigger`](https://drafts.csswg.org/css-animations-2/#animation-trigger) was proposed in [#8942](https://github.com/w3c/csswg-drafts/issues/8942) and [included](https://github.com/w3c/csswg-drafts/pull/10128) in the [css-animations-2](https://drafts.csswg.org/css-animations-2/#animation-trigger) and [web-animations-2](https://drafts.csswg.org/web-animations-2/#animation-trigger) specs. I’m filing this umbrella meta issue so the CSS working group can get familiar with the drafted API and share thoughts that might help answer questions in issues ([11914](https://github.com/w3c/csswg-drafts/issues/11914), [11915](https://github.com/w3c/csswg-drafts/issues/11915), [11910](https://github.com/w3c/csswg-drafts/issues/11910), [11918](https://github.com/w3c/csswg-drafts/issues/11918), [11932](https://github.com/w3c/csswg-drafts/issues/11932), [12064](https://github.com/w3c/csswg-drafts/issues/12064)) related to this feature.

To summarize, `animation-trigger` is a shorthand for some new animation triggering behavior usable both as CSS properties and Web Animation properties:

- `animation-trigger-type`: whose values can be
  - `once`: The default value; plays animations only the first time its trigger condition is met. [[Demo](https://davmila.github.io/sta-demos/once.html)]
  - `repeat`: Plays and resets animations each time its trigger conditions are met. [[Demo](https://davmila.github.io/sta-demos/repeat.html)]
  - `alternate` Plays and reverses animations each time its trigger conditions are met. [[Demo](https://davmila.github.io/sta-demos/alternate.html)]
  - `state` Plays and pauses animations each time its trigger conditions are met. [[Demo](https://davmila.github.io/sta-demos/state.html)]
- `animation-trigger-timeline`: the timeline in which the trigger’s conditions are evaluated
- `animation-trigger-range`: the range of the timeline in which the trigger’s play condition is met,
- `animation-trigger-exit-range`: the range of the timeline in which the trigger’s reset/reverse/pause (depending on the `animation-trigger-type`) condition is met.

[...]

PR #12314, opened 2025-06-10
[css-animations-2] [web-animations-2] Animation trigger event

szager-chromium:
Specification for event-based animation triggers, as discused in issue##12029

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -770,8 +770,6 @@ for an [=animation trigger=] that enters its [=animation trigger/active interval
 
 The behavior of each value is defined in [[web-animations-2#trigger-behaviors]].
 
-Issue: Should 'animation-trigger-behavior' allow multiple values (currently it does)?
-
 ## The 'animation-trigger-timeline' property ## {#animation-trigger-timeline}
 
 The 'animation-trigger-timeline' property specifies the <a>timeline</a>
@@ -964,20 +962,21 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
                         it defaults to 100%.
 </dl>
 
-## The 'animation-trigger-name' property ## {#animation-trigger-name}
+## The 'trigger-name' property ## {#trigger-name}
 
-The 'animation-trigger-name' property associates an element with
-all elements having an 'animation-trigger-event' property with
-a trigger name matching a value in this element's 'animation-trigger-name'.
-When an {{Event}} is dispatched with this element as {{Event/target}},
-and with {{Event/type}} matching the <<trigger-event-type>> of a clause in the
-'animation-trigger-event' property of another element, animation
-playback on the other element is modified according to its
+The 'trigger-name' property causes an [=event-based animation trigger=] to be created
+for any animation whose target element has an 'animation-trigger-event' property
+with matching <<trigger-name>>.
+When an {{Event}} is dispatched with the element having the matching 'trigger-name' property
+as {{Event/target}},
+and with {{Event/type}} matching the <<trigger-event-type>> of the
+'animation-trigger-event' property on the animation target, the animation's
+playback is modified according to its
 'animation-trigger-behavior' property, as described for
 [=event-based animation triggers=].
 
 <pre class='propdef'>
-Name: animation-trigger-name
+Name: trigger-name
 Value: none | [ <<trigger-name>> ]#
 Initial: none
 Applies to: all elements
@@ -995,17 +994,17 @@ The 'animation-trigger-event' property is used to specify the criteria
 for [=qualifying events=] that cause an [=event-based animation trigger=]
 to perform its effect on its associated [=animation=]. It consists of two parts:
 a <<trigger-event-type>> that is matched against the {{Event/type}} of a dispatched {{Event}},
-and an <<trigger-name>> that is matched against the 'animation-trigger-name' property
+and an <<trigger-name>> that is matched against the 'trigger-name' property
 of the event's {{Event/target}}. If both criteria match for a given event
 the trigger will perform its effect on its associated [=animation=].
 
-If the <<trigger-name>> is omitted from <<single-trigger-event>> declaration,
+If the <<trigger-name>> is omitted from <<single-event-trigger>> declaration,
 {{Event/target}} is matched against the element
 having the 'animation-trigger-event' property itself.
 
 <pre class='propdef'>
 Name: animation-trigger-event
-Value: none | [ <<single-trigger-event>> ]#
+Value: none | [ <<single-event-trigger>> ]#
 Initial: none
 Applies to: all elements
 Inherited: no
@@ -1014,7 +1013,7 @@ Computed value: as specified
 Animation type: not animatable
 </pre>
 
-<pre class=prod><dfn>&lt;single-trigger-event></dfn> = event ( <<trigger-event-type>> <<trigger-name>>? )</pre>
+<pre class=prod><dfn>&lt;single-event-trigger></dfn> = event ( <<trigger-event-type>> <<trigger-name>>? )</pre>
 <pre class=prod><dfn>&lt;trigger-event-type></dfn> = <<custom-ident>></pre>
 
 ## The 'animation-trigger' property ## {#animation-trigger}
@@ -1027,7 +1026,7 @@ together in a single declaration, specifying the [=animation trigger=] for an an
 
 <pre class='propdef'>
 Name: animation-trigger
-Value: auto | none | <<single-animation-trigger>>#
+Value: <<single-animation-trigger>>#
 Initial: see individual properties
 Applies to: all elements
 Inherited: no
@@ -1038,9 +1037,9 @@ Animation Type: not animatable
 </pre>
 
 <pre class=prod>
-<dfn>&lt;single-animation-trigger></dfn> = [ <<single-animation-trigger-behavior>> | [ <<single-trigger-event>> | <<single-timeline-trigger>> ] ] <<single-animation-trigger-behavior>>?
+<dfn>&lt;single-animation-trigger></dfn> = <<single-animation-trigger-behavior>> || [ <<single-event-trigger>> | <<single-timeline-trigger>> ]
 
-<dfn>&lt;single-timeline-trigger></dfn> = [ <<dashed-ident>> | <<scroll()>> | <<view()>> ] [ normal | <<length-percentage>> | <<timeline-range-name>> <<length-percentage>>? ]{0,4}
+<dfn>&lt;single-timeline-trigger></dfn> = none | auto | [ <<dashed-ident>> | <<scroll()>> | <<view()>> ] [ normal | <<length-percentage>> | <<timeline-range-name>> <<length-percentage>>? ]{0,4}
 </pre>
 
 # Animation Events # {#events}


szager-chromium:
@flackr PTAL

css-animations-2/Overview.bs
```diff
@@ -770,6 +770,8 @@ for an [=animation trigger=] that enters its [=animation trigger/active interval
 
 The behavior of each value is defined in [[web-animations-2#trigger-behaviors]].
 
+Issue: Should 'animation-trigger-behavior' allow multiple values (currently it does)?
```
DavMila: Hm, I don't think it does?

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,67 @@ Animation Triggers {#triggers}
...
+
+<dl>
+  <dt>''animation-trigger-type/once''</dt>
+  <dd>If [=play state=] is [=play state/finished=], do nothing.
+      Otherwise, perform the actions specified for the {{Animation/play}} function.
+  </dd>
+
+  <dt>''animation-trigger-type/repeat''</dt>
+  <dd>First, if [=play state=] is [=play state/running=] or [=play state/paused=]
+      perform the actions specified for the {{Animation/cancel}} function.
```
DavMila, line 2408: More of an FYI comment perhaps: I don't yet know if this is going to be different for event-based triggers but for scroll-triggered what we want isn't cancel() but more of a "reset"... which is not yet a concept defined in the web animations spec. Essentially, it does
```
animation.currentTime = 0;
animation.pause();
```
Since you follow up by playing the animation, maybe this doesn't matter so much, but for CSS animations, cancel() will be accompanied by an `animationcancel` event which we might not want.

web-animations-2/Overview.bs
```diff
@@ -3622,40 +3680,62 @@ Update the type for the AnimationPlaybackEventInit members.
...
 <div dfn-type=constructor class=constructors dfn-for="AnimationTrigger">
   : <dfn lt="AnimationTrigger(options)">AnimationTrigger(|options|)</dfn>
   ::
       Creates a new {{AnimationTrigger}} object using the following procedure:
 
-      1. Create a new {{AnimationTrigger}} object, |trigger|.
+      1. If |options| contains a {{AnimationTriggerOptions/timeline}} entry
+         and also contains either or both of
+         {{AnimationTriggerOptions/eventTarget}} and {{AnimationTriggerOptions/eventType}}
+         <a>throw</a> a <span class="exceptionname">TypeError</span>.
```
DavMila, line 3707: We have this idea of a default trigger which has the document.timeline as its timeline and gives a nice explanation to how pre-existing CSS animations are automatically played. I think the default trigger, which is what you get with `new AnimationTrigger()` should be a timeline-based trigger with the document.timeline as its timeline.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,67 @@ Animation Triggers {#triggers}
...
+
+  <dt>''animation-trigger-type/repeat''</dt>
+  <dd>First, if [=play state=] is [=play state/running=] or [=play state/paused=]
+      perform the actions specified for the {{Animation/cancel}} function.
+      Then, in all cases, perform the actions specified for the {{Animation/play}} function.
+  </dd>
 
-An [=animation trigger=] |trigger|
+  <dt>''animation-trigger-type/alternate''</dt>
+  <dd>Perform the actions specified for the {{Animation/reverse}} function.</dd>
```
DavMila: Will we reverse the animation on the first event? Is that intended?

css-animations-2/Overview.bs
szager: The '#' in the syntax description from line 734 means it accepts a comma-separated list of behaviors.

web-animations-2/Overview.bs
szager: It's an interesting question about whether it's desirable to issue events for these transitions. They are not otherwise directly observable; is that a good thing? bad thing? Indifferent?

web-animations-2/Overview.bs
szager: That sounds good to me, does this spec language conflict with that?

web-animations-2/Overview.bs
szager: Great point! I need to think about this.

szager-chromium: I updated to specify that if play state is idle then the effect is `play()`, otherwise the effect is `reverse()`.

web-animations-2/Overview.bs
DavMila: Right.. sorry, misread some of the text.

css-animations-2/Overview.bs
DavMila: Oh I think that has to do with the fact that multiple animations can be specified in a comma-separated list. Each animation will then be mapped 1:1 with a comma-separated list of triggers. Each trigger still only gets one behavior. Though idk if this has been incorrectly represented in the spec text.

ydaniv: Yes, what @DavMila said is correct. I think this is called a coordinated list.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
 Issue: Should there be any effect of triggers on scroll-driven animations?
 
-### Animation Trigger Internal State ### {#trigger-state}
+There are two types of [=animation triggers=]:
+<dfn export lt='timeline-based animation trigger'>timeline-based</dfn> and
+<dfn export lt='event-based animation trigger'>event-based</dfn>. The
+type of an animation trigger is deduced from its attributes: a
+timeline-based animation trigger will have a non-empty {{AnimationTrigger/timeline}}
+attribute, while an event-based animation trigger will have non-empty
+{{AnimationTrigger/eventTarget}} and {{AnimationTrigger/eventType}} attributes.
```
ydaniv, line 2383: Although we allow 2 types of animations, time-based and progress-based, using the `animation` property, we don't deduce that based on empty properties. On the contrary, we have `auto` initial values for all properties that are not needed when you use that type or the other.
So, IMHO we can't really add more properties to the API and expect those to be empty, both technically and design-wise.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
+A <dfn for="animation trigger" export>qualifying event</dfn> is
+an {{Event!!interface}} with {{Event/type}} matching the trigger's {{AnimationTrigger/eventType}}
+and {{Event/target}} matching the trigger's {{AnimationTrigger/eventTarget}}.
+
+#### Event-Based Animation Trigger Behavior #### {#event-trigger-behavior}
+
+When a [=animation trigger/qualifying event=] is dispatched, the associated
+animation's playback is modified as follows,
+based on the trigger's {{AnimationTrigger/behavior}}
+and the current [=play state=] of the animation:
```
ydaniv, line 2398: I don't think it's correct to specify behaviors based on the type of the trigger. I think we should be specifying how each time handles its internal state and apply its specified behavior, rather than how the behavior acts.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
+
+<dl>
+  <dt>''animation-trigger-type/once''</dt>
+  <dd>If [=play state=] is [=play state/finished=], do nothing.
```
ydaniv, line 2402: This way it's not really a `once` behavior. It will re-play as long as it's not finished.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
+      Otherwise, perform the actions specified for the {{Animation/play}} function.
+  </dd>
+</dl>
+
+### Timeline-Based Animation Triggers ### {#timeline-based-trigger}
+
+Like animations, [=timeline-based animation triggers=] are associated with a [=timeline=]
+and are attached to an [=animation attachment range|attachment range=].
+Changes to animation playback occur when the associated [=timeline=] transitions
+between inside and outside the [=animation attachment range|attachment range=].
```
ydaniv, line 2430: This is not so accurate in spec technical terms. The timeline doesn't transition, rather its progress goes in and out of its [active interval](https://drafts.csswg.org/web-animations-2/#trigger-active-interval)

css-animations-2/Overview.bs
```diff
@@ -962,18 +964,70 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
...
+'animation-trigger-event', 'animation-trigger-range-start', 'animation-trigger-range-end',
 'animation-trigger-exit-range-start', and 'animation-trigger-exit-range-end'
-together in a single declaration,
-specifying the [=animation trigger=] for an animation.
+together in a single declaration, specifying the [=animation trigger=] for an animation.
 
 <pre class='propdef'>
 Name: animation-trigger
-Value: <<single-animation-trigger>>#
+Value: auto | none | <<single-animation-trigger>>#
```
ydaniv: So far we intentionally didn't have a `none` nor `auto` values for the `animation-trigger` property, though this may change depending on what we resolve in #12119

ydaniv: Also, probably if we do allow `auto`/`none` they should be part of the `<single-animation-trigger>` def and be allowed per trigger defintion.

css-animations-2/Overview.bs
```diff
@@ -984,7 +1038,9 @@ Animation Type: not animatable
 </pre>
 
 <pre class=prod>
-<dfn>&lt;single-animation-trigger></dfn> = <<single-animation-trigger-behavior>> || [ none | auto | [ [ <<dashed-ident>> | <<scroll()>> | <<view()>> ] [ normal | <<length-percentage>> | <<timeline-range-name>> <<length-percentage>>? ]{0,4} ] ]
+<dfn>&lt;single-animation-trigger></dfn> = [ <<single-animation-trigger-behavior>> | [ <<single-trigger-event>> | <<single-timeline-trigger>> ] ] <<single-animation-trigger-behavior>>?
```
ydaniv: I'm not sure I follow the logic here. Perhaps you meant to allow `auto` and `none` as possible alternatives and that's what got the syntax confusing here?

css-animations-2/Overview.bs
```diff
@@ -984,7 +1038,9 @@ Animation Type: not animatable
...
+
+<dfn>&lt;single-timeline-trigger></dfn> = [ <<dashed-ident>> | <<scroll()>> | <<view()>> ] [ normal | <<length-percentage>> | <<timeline-range-name>> <<length-percentage>>? ]{0,4}
```
ydaniv: You changed the syntax of timeline here. `timeline` also accepts `none` and `auto` which you removed here.

css-animations-2/Overview.bs
```diff
@@ -962,18 +964,70 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
                         it defaults to 100%.
 </dl>
 
+## The 'animation-trigger-name' property ## {#animation-trigger-name}
```
ydaniv: If this property is not a longhand sub-property of `animation-trigger` then IMO it should not be named using the scheme of `animation-trigger-*`. IIUC you meant to define this property as the event source declaration, am I right?
In that case, this raises some questions:

1. Is that declaration tree scoped?
2. Is the name reused for all event types that are emitted from this target?
3. Does containment properties affect its scope?
4. Can we lift this scope, same as the` timeline-scope` property does?
5. Do names collide and override eachother? Or can they be used multiple times?
6. Will these names be ever used only for the purpose (and scope) of animations?

css-animations-2/Overview.bs
```diff
@@ -962,18 +964,70 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
...
+
+The 'animation-trigger-name' property associates an element with
+all elements having an 'animation-trigger-event' property with
```
ydaniv: I think it's not accurate to say we "associate elements". The event needs to be associated with a target, and the target here is an animation(s).

css-animations-2/Overview.bs
```diff
@@ -962,18 +964,70 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
...
+Computed value: as specified
+Animation type: not animatable
+</pre>
+
+<pre class=prod><dfn>&lt;trigger-name></dfn> = <<dashed-ident>></pre>
+
+## The 'animation-trigger-event' property ## {#animation-trigger-event}
+
+The 'animation-trigger-event' property is used to specify the criteria
+for [=qualifying events=] that cause an [=event-based animation trigger=]
```
ydaniv, line 994: Shouldn't this be the other way around? As in, the event source should declare the types that emit its name, and the event targets should register to the name alone?
This is normally how you would register an event. You add a handler per type, and then these handlers can publish the name to trigger behavior.

web-animations-2/Overview.bs
szager: I'm not sure if your concern is with the WAAPI interface or the CSS syntax, but the attributes on a WAAPI object can be `undef` or `null`. Is it maybe the word `empty` that you object to?

In the rest of the WAAPI proposal I added safeguards to ensure that an instance of `AnimationTrigger` will conform to the requirement:

`!(trigger.eventType ^ trigger.eventTarget) && (trigger.timeline ^ trigger.eventTarget))`

... so the trigger type *is* detectable based on these attributes.

web-animations-2/Overview.bs
szager: Respectfully, I disagree. Events are fundamentally ephemeral occurences, and I think it's more intuitive to express event-based triggers in terms of taking an action in response to an event without introducing the complexity of additional state, most of which is internal and thus not directly detectable. Conceptually, event-based triggers are like event handlers.

web-animations-2/Overview.bs
szager: If the animation is `idle`, then this will play it.
If it's `running`, then this will have no effect.
If it's `paused` this will un-pause it.
If it's `finished`, we do nothing.

That's the intended behavior, is there something unintuitive?

web-animations-2/Overview.bs
szager: I'm not sure I understand your objection. The spec specifies when the trigger state is updated, and at that time if it's detected that the new state will be different from the previous state, then that can be considered a "transition". It's maybe a synthetic concept, but it's rigorous in terms of the current spec.

szager: On a related note: the existing spec needs some clarification over when the trigger [behaviors](https://drafts.csswg.org/web-animations-2/#trigger-behaviors) are applied -- specifically, they should only be applied when a transition occurs.

For example, consider this description of `alternate':

"If state is primary and did trigger is true, the effect is to reverse the associated animation."

If this logic is applied on every animation update, then the animation would be reversed on every frame. The intention is clearly that this behavior should be applied only when a transition occurs, but the spec doesn't say that.

css-animations-2/Overview.bs
szager: Got it; I will update the PR...

css-animations-2/Overview.bs
szager: After discussion with @flackr and @DavMila, I think the plan is to use a scoping mechanism similar to [anchor positioning](https://drafts.csswg.org/css-anchor-position-1/#anchor-scope) -- specifically, there will be a `trigger-scope` property to apply tree scoping, but in the absense of any applicable `trigger-scope` the scope will be global. As with anchor positioning, containment will not affect scope.

The name is only used to specify the event target; the event type is specified via `animation-trigger-event` or the `animation-trigger` shorthand.

I will change the property to `trigger-name`, to avoid confusion with the properties that can be set via `animation-trigger` shorthand.

css-animations-2/Overview.bs
szager: I don't quite follow you here, but I will point out firstly that the term "target" appears to be overloaded here. Because we're working with events, and "event target" is a very well-understood concept, I'm using the term "target" to refer to the event target and not the element that will animate.

There is definitely an "association" happening here:

```
#event-target {
  trigger-name: --my-trigger;
}
#some-other-event-target {
  trigger-name: --my-trigger;
}
#animatable {
  animation-trigger-name: --my-trigger;
  animation-trigger-event: click;
}
#some-other-animatable {
  animation-trigger-name: --my-trigger;
  animation-trigger-event: keydown;
}
```

If a `click` is dispatched to either #event-target or #some-other-event-target, then #animatable will run its animations. If a `keydown` is disaptched to either #event-target or #some-other-event-target, #some-other-animatable will run its animations.

Is there some other term besides "association" that better describes the relationships between these elements?

css-animations-2/Overview.bs
szager: Can you give an example of the syntax you have in mind? It's not clear to me how to express the following using your suggestion:

```
#event-target {
  trigger-name: --my-trigger;
}
#some-other-event-target {
  trigger-name: --my-trigger;
}
#animatable {
  animation-trigger-name: --my-trigger;
  animation-trigger-event: click;
}
#some-other-animatable {
  animation-trigger-name: --my-trigger;
  animation-trigger-event: keydown;
}
```

css-animations-2/Overview.bs
szager: I misunderstood before how coordinated lists were being used; I will fix this.

css-animations-2/Overview.bs
szager: Also me not  understanding coordinated lists; will fix.

css-animations-2/Overview.bs
szager: Also coordinated lists; will fix.

css-animations-2/Overview.bs
ydaniv: OK, I still think that the type should be part of the event declaration on the target.
Is there a specific reason you want to put the type on the animation target? IMHO, from design perspective this seems to belong on the event target.

css-animations-2/Overview.bs
ydaniv: Because we have 2 elements in play here, I was trying to use "source" for the event target, and "target" for the animated element, but let's stick with event target.
I understand the association, we definitely need that mechanism.
I'm saying we associate the event target with an animation, and not really with another element. The animation is associated with its own target, which can be nulled or changed.

css-animations-2/Overview.bs
ydaniv: First of all, this example is rather confusing to me. Did you intentionally used an example with two identical names? Are they supposed to collide? Or to be aggregated?

So to simplify with 2 different names, I was thinking on something like:

```
#event-target {
  trigger-event: --my-trigger click;
}
#some-other-event-target {
  trigger-event: --my-other-trigger keydown;
}
#animatable {
  animation-trigger: repeat event(--my-other-trigger);
}
#some-other-animatable {
  animation-trigger: --my-other-trigger;
}
```

css-animations-2/Overview.bs
szager-chromium: I see your point; I re-worded this in terms of creating an animation trigger for animations on elements having a matching `animation-trigger-event` property.

css-animations-2/Overview.bs
szager-chromium: I think your proposed syntax is isomorphic to mine, but your syntax occupies more of the --dashed-ident namespace, because it refers to an (event type, event target) pair rather than just an event target; and because it seems to require uniqueness of --dashed-ident labels for triggers -- that's why I deliberately created an example with two identical names. The idea here is that there may be multiple event targets that all affect a single animation target. Here's a comparison of the two syntaxes:

```
#event-target-1 {
  trigger-name: --my-trigger;
}
#event-target-2 {
  trigger-name: --my-trigger;
}
#animatable {
  animation-trigger: event(click --my-trigger);
}
```
-or-
```
#event-target-1 {
  trigger-name: --my-trigger-1 click;
}
#event-target-2 {
  trigger-name: --my-trigger-2 click;
}
#animatable {
  animation-trigger: event(--my-trigger-1), event(--my-trigger-2);
}
```

Now suppose that we want either a click or a keypress to either of the event targets to trigger the animation:

```
#event-target-1 {
  trigger-name: --my-trigger;
}
#event-target-2 {
  trigger-name: --my-trigger;
}
#animatable {
  animation-trigger: event(click --my-trigger), event(keypress --my-trigger);
}
```
-or-
```
#event-target-1 {
  trigger-name: --my-click-trigger-1 click, --my-keypress-trigger-1 keypress;
}
#event-target-2 {
  trigger-name: --my-click-trigger-2 click, --my-keypress-trigger-2 keypress;
}
#animatable {
  animation-trigger: event(--my-click-trigger-1), event(--my-keypress-trigger-1), event(--my-click-trigger-2), event(--my-keypress-trigger-2);
}
```

Again, I think these syntaxes are isomorphic. My own preference to avoid --dashed-ident namespace pollution by allowing non-unique identifiers and pushing the event type specification into the animation target's style, but I don't have my heart set on it.

ydaniv: Well, your idea for allowing multiple event targets to trip the same trigger (not animation) is quite interesting, but I can't think right now on anything that works that way, so it seems quite confusing. In most cases I can think of idents override each other, and are not aggregated, but I could be missing something.
I understand now your logic about reducing the amount of surface pollution in the namespace, but still it feels a bit inverted.
I Would be happy to hear more from the group on both.

Regarding the syntax you used in the examples above, since `animation`, and its sub-properties, are  coordinated lists, delimiting the triggers above by commas makes them act on separate animations, not on the same animation. So this would be invalid.

css-animations-2/Overview.bs
```diff
@@ -962,14 +962,67 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
...
+to perform its effect on its associated [=animation=]. It consists of two parts:
+a <<trigger-event-type>> that is matched against the {{Event/type}} of a dispatched {{Event}},
+and an <<trigger-name>> that is matched against the 'trigger-name' property
+of the event's {{Event/target}}. If both criteria match for a given event
+the trigger will perform its effect on its associated [=animation=].
+
+If the <<trigger-name>> is omitted from <<single-event-trigger>> declaration,
+{{Event/target}} is matched against the element
+having the 'animation-trigger-event' property itself.
```
ydaniv, line 1003: That's interesting! So the form of an anonymous event target would be something like `event(<trigger-event-type>)`?

css-animations-2/Overview.bs
```diff
@@ -962,14 +962,67 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
...
+as {{Event/target}},
+and with {{Event/type}} matching the <<trigger-event-type>> of the
+'animation-trigger-event' property on the animation target, the animation's
+playback is modified according to its
+'animation-trigger-behavior' property, as described for
+[=event-based animation triggers=].
+
+<pre class='propdef'>
+Name: trigger-name
+Value: none | [ <<trigger-name>> ]#
```
ydaniv, line 980: If this property has no other sub-properties, why does  it need to be a list?
IIUC this should simply be `none | <<dashed-ident>>`

css-animations-2/Overview.bs
```diff
@@ -962,14 +962,67 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
...
+Applies to: all elements
+Inherited: no
+Percentages: N/A
+Computed value: as specified
+Animation type: not animatable
+</pre>
+
+<pre class=prod><dfn>&lt;trigger-name></dfn> = <<dashed-ident>></pre>
+
+## The 'animation-trigger-event' property ## {#animation-trigger-event}
```
ydaniv, line 991: Shouldn't this be moved up to sit among the other  `animation-trigger-*` sub-properties?

css-animations-2/Overview.bs
```diff
@@ -962,14 +962,67 @@ The values of 'animation-trigger-exit-range-start' have the following meaning:
                         it defaults to 100%.
 </dl>
 
+## The 'trigger-name' property ## {#trigger-name}
```
ydaniv, line 965: I think  this property should be moved to a separate section since it's not a sub-property of `animation-trigger`.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
+associated [=animation=]'s playback.
 
 Issue: Should there be any effect of triggers on scroll-driven animations?
 
-### Animation Trigger Internal State ### {#trigger-state}
+There are two types of [=animation triggers=]:
+<dfn export lt='timeline-based animation trigger'>timeline-based</dfn> and
+<dfn export lt='event-based animation trigger'>event-based</dfn>. The
+type of an animation trigger is deduced from its attributes: a
+timeline-based animation trigger will have a non-empty {{AnimationTrigger/timeline}}
```
ydaniv, line 2381: I don't think "non-empty" is the correct term. Properties have initial values, so an "empty" property will compute to its initial value, and therefore will have a valid value. We just need to specify what that value is and make sure these play nicely together.
For `animation-trigger-timeline` the initial value is `auto`, though we're probably looking for something to reset that value to `none` when the `animation-trigger-event` is specified and not the initial default, right?

web-animations-2/Overview.bs
ydaniv: Same as the comment on the line above, my concern here is with the CSS syntax.
So we could say, for example, something like: "initial value for `animation-trigger-event` is `none`. And if that value is not `none`, then `animation-trigger-timeline` is reset to `none`.
Since historically we started designing triggers using timelines, we decided that the initial value for `-timeline` will be `auto`, and that would match to the `DocumentTimeline`, same as with `animation-timeline`. Though we could revisit this model and consider matching the initial values and default existing behavior of animations using events.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
+attribute, while an event-based animation trigger will have non-empty
+{{AnimationTrigger/eventTarget}} and {{AnimationTrigger/eventType}} attributes.
+
+### Event-Based Animation Triggers ### {#event-based-trigger}
+
+An [=event-based animation trigger=] executes its {{AnimationTrigger/behavior}}
+when a [=qualifying event=] is dispatched.
```
ydaniv, line 2388: I think this is rather vague considering events, and what we've already established for timelines in the context of trigger. More specifically, we know a trigger's behavior enters its `primary` state when its active range is entered, and its `inverse` state when exited.
I assume you mean for events to toggle between these states on each dispatch, correct?

web-animations-2/Overview.bs
ydaniv: I guess if this is specific to event-based triggers then we could say that since they're stateless by nature, they simply toggle the current state of the animation based on their behavior, as opposed to timeline-based triggers which are stateful.

web-animations-2/Overview.bs
ydaniv: I guess I didn't take the `running` part for granted, but that's fine.

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
+
+When a [=animation trigger/qualifying event=] is dispatched, the associated
+animation's playback is modified as follows,
+based on the trigger's {{AnimationTrigger/behavior}}
+and the current [=play state=] of the animation:
+
+<dl>
+  <dt>''animation-trigger-type/once''</dt>
+  <dd>If [=play state=] is [=play state/finished=], do nothing.
+      Otherwise, perform the actions specified for the {{Animation/play}} function.
```
ydaniv, line 2403: Wouldn't it be better to say we invoke the the procedure of `[=playing an animation=]`? Perhaps with `auto-rewind` flag as `false`? Then this would cover this without requiring the if clause?

web-animations-2/Overview.bs
```diff
@@ -2367,14 +2368,70 @@ Animation Triggers {#triggers}
...
+  </dd>
+
+  <dt>''animation-trigger-type/repeat''</dt>
+  <dd>First, if [=play state=] is [=play state/running=] or [=play state/paused=]
+      perform the actions specified for the {{Animation/cancel}} function.
```
ydaniv, line 2408: Again, I think it's better to reuse the procedure for `[=canceling an animation=]`, rather than reusing the Animation's method.

ydaniv (changes requested):
Sorry for the delay! I've started doing a more thorough reading and added more comments.

flackr:
We've had a few resolutions on the trigger syntax in #12652 and #12336 which it looks like have not been reflected into this PR yet. In particular, we switched to `event-trigger-*` properties for event triggers, and decided that all triggers can have an activated / deactivated state with a list of events to progress into and out of the activated state. If no exit events specified it doesn't stay in the activated state or trigger exit behaviors (behaving much like the current design). Behaviors are also for now not a part of the resolved behavior so we don't need event specific behaviors as we'll have a common set of actions #12611 which can be associated with activation (and optionally deactivation).

PR #12349, opened 2025-06-17
[css-values-5] Fix dfn of `<input-position>` and `<output-value>`

tidoust:
A `<dfn>` of type `type` should only enclose a basic data type (something between `<` and `>`). At least, that's been the case across specs until now. The spec was using the dfn to describe a more complex `value`. More likely, the spec was trying to combine the definition of the two types with the definition of the value pattern that uses them in one single dfn.

This update changes the value pattern definition type to `value` and creates additional `type` dfns for the individual types so that they can be referenced from other constructs.

### DIFF
--- a/css-values-5/Overview.bs
+++ b/css-values-5/Overview.bs
@@ -1185,16 +1185,17 @@ Defining the Interpolation Map</h4>
 	The [=interpolation map=] values are defined as follows:
 
 	<dl dfn-for="calc-interpolate(), transform-interpolate(), color-interpolate(), interpolate()">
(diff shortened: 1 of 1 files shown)


tabatkins:
The current markup *is* only enclosing basic data types. It's doing *two* of them, but that's fine. The text content of the `dfn` is irrelevant, since `lt` is specified.

tidoust:
Ah, I was worried you might say that ;) I agree that this does not matter for referencing purpose. Extraction of grammar constructs is a bit more rigid for now (and, for example, cannot always rely on `<dfn>` and term references being used), and it's the only occurrence of that pattern across specs. Anwyay, that's fine, I'll be more flexible. 🤸

tidoust:
Crawler now handles that scenario. Closing!

Issue #1562, opened 2017-06-25
[cssom-view] Consider making scroll methods return promises


mkay581:
[Smooth scrolling](https://drafts.csswg.org/cssom-view/#smooth-scrolling) has been introduced into the spec where the scroll happens over a period of time, which is great. However, I think it would be even more useful if scroll method returned a promise so the developer could wait until the scroll finishes before moving on to next set of tasks. 

Obviously `scroll()`ing on an element where `scroll-behavior` option set to `auto` makes scroll instant and is therefore not inherently asynchronous. But I think, to keep the API consistent, the scroll method should return a promise every time it is used.

So if you had the following HTML...

```html
<div id="scroll-container" style="scroll-behavior: smooth;">
   <!-- other content here --> 
</div>
```

you can do things like this:

```js
element.scroll(0, 400).then(() => {
   // ...scroll has finished
});
```

mkay581:
FWIW, [here](https://discourse.wicg.io/t/proposal-element-scroll-with-ease-options/1720/6?u=mkay581) is a link to the relevant discussion where others have also +1'd this feature.

dartess:
in 2018 it is still very much needed, only now we will use it this way:

```
let element = document.getElementById('scroll-container');
await element.scroll(0, 400);
// ...scroll has finished
```

use case: set focus after scroll

Slapbox:
Is there an ETA for this being spec?

patchthecode:
still waiting...

webdevelopers-eu:
Unlike TabAtkins' opinion in the IRC chat, I strongly support the idea of implementing Promise rejection if a scroll is interrupted.

As I mentioned in #3744 , the intended effect of `scrollIntoView` is to bring the specified element into view. If this does not happen for any reason, the promise should be rejected, as the action was not fulfilled. There are specific scenarios where this should occur, such as:

* The element is removed or stops being rendered before it reaches the visible part of the viewport.
* The smooth, time-dependent scrolling is interrupted programmatically or by user action.

Simply put, the goal of `scrollIntoView` is to bring an element into view. If this is achieved, the promise should be fulfilled; if not, the promise should be rejected. This aligns with the natural and expected behavior that any programmer would anticipate when triggering an action intended to produce a specific result.

markcellus:
> If this is achieved, the promise should be fulfilled; if not, the promise should be rejected. 

[...]

PR #12355, opened 2025-06-17
[cssom-view-1] Scroll methods in Element and Window return Promises

mustaqahmed:
Switch the return type for the scroll methods from `undefined` to `Promise<undefined>` as per the discussion and resolution in https://github.com/w3c/csswg-drafts/issues/1562#issuecomment-389586317.

Fixed: #1562.


flackr:
In the discussion we talked about the argument to the promise being an object which would tell the developer the result of the scroll, e.g. to know if it was cancelled / interrupted as was brought up a few times as a common thing to need to know.

mustaqahmed:
> In the discussion we talked about the argument to the promise being an object which would tell the developer the result of the scroll, e.g. to know if it was cancelled / interrupted as was brought up a few times as a common thing to need to know.

That's correct.  The problem is that it not yet defined how the promise would be resolved to mark scroll completion vs interruption.  I can update this PR to resolve with a `true` to mark completion and `false` "otherwise" but I am hesitant because of a few reasons:

1. It is perhaps not okay to add normative things w/o discussion, right?  E.g. we don't know if we should use a Boolean here vs an enum.

2. What does that "otherwise" mean to be precise?  Does one programmatic scroll request interrupt all pending promises?  Or only the pending ones for the same scroller?

3. We _may_ need additional text in the current scroll algorithms to "collect" pending scrolling requests.  To me a follow-up PR after this one seems cleaner.

I am proposing to land this PR as "closing" #1562, staring a new spec issue to discuss 1+2 above, and then hash out 3 in the PR for that new issue.

flackr:
> > In the discussion we talked about the argument to the promise being an object which would tell the developer the result of the scroll, e.g. to know if it was cancelled / interrupted as was brought up a few times as a common thing to need to know.
> 
> That's correct. The problem is that it not yet defined how the promise would be resolved to mark scroll completion vs interruption. I can update this PR to resolve with a `true` to mark completion and `false` "otherwise"

I would suggest that it not resolve with a boolean but rather with an object, e.g. define a ScrollResult, that we can add to if we later discover that we want to pass additional information. That said, having a boolean there and opening an issue for whether we should change it is a good thing.

> but I am hesitant because of a few reasons:
> 
> 1. It is perhaps not okay to add normative things w/o discussion, right?  E.g. we don't know if we should use a Boolean here vs an enum.

This is normal, the resolution we have is effectively to pursue specifying and prototyping the feature. It's fully expected that in the process of doing this the implementor may make some opinionated decisions in writing the spec and bring those decisions back to the working group. Having something implemented will make it easier to have examples where we can show how the value would be used. 

> 2. What does that "otherwise" mean to be precise?  Does one programmatic scroll request interrupt all pending promises?  Or only the pending ones for the same scroller?

It would be valuable to define this to mean what we think makes the most sense. IMO this is anything that prevents completing the original scroll request. A programmatic scroll request on an unrelated scroller probably should not interrupt it. A scroll on the same scroller would mean that we are no longer actively scrolling to the originally requested location.

> 3. We _may_ need additional text in the current scroll algorithms to "collect" pending scrolling requests.  To me a follow-up PR after this one seems cleaner.
> 
> I am proposing to land this PR as "closing" #1562, staring a new spec issue to discuss 1+2 above, and then hash out 3 in the PR for that new issue.

It is better to have the boolean there and to add an inline issue in the spec that says we need to agree on the type and when it is successful / unsuccessful. This shows spec readers that the API is likely to change, demonstrates the use case and can even point to the issue for people to add their feedback. Here's an example from css-images-5: https://github.com/w3c/csswg-drafts/blob/b5312ce985daef0e0a2f070130472d58892512cc/css-images-5/Overview.bs#L55

mustaqahmed:
Updated the PR with promise resolution with true/false, and filed #12495 for next steps.  Please take a look.

mustaqahmed:
Here is a pdf showing [the dependency of the scroll algorithms](https://drive.google.com/file/d/1QLYGrCwylU7mhwdQIrXIXnSFFZCDFfaG/view?usp=drive_link) in this spec.  Hopefully this will help the review here.

cssom-view-1/Overview.bs
```diff
@@ -460,8 +465,14 @@ When a user agent is to <dfn for="viewport" export>perform a scroll</dfn> of a <
...
 1. <a for="/">Perform a scroll</a> of the <a>viewport</a>'s <a>scrolling box</a> to its current scroll position + (<var>layout dx</var>, <var>layout dy</var>) with <var>element</var> as the
     associated element, and <var>behavior</var> as the scroll behavior.
+    Let <var>scrollPromise1</var> be the promise returned from this step.
 1. <a for="/">Perform a scroll</a> of <var>vv</var>'s <a>scrolling box</a> to its current scroll position + (<var>visual dx</var>, <var>visual dy</var>) with <var>element</var> as the associated
     element, and <var>behavior</var> as the scroll behavior.
+    Let <var>scrollPromise2</var> be the promise returned from this step.
+1. If both <var>scrollPromise1</var> and <var>scrollPromise2</var> get resolved with true,
+    return a promise resolved with true.
+    Otherwise (either one gets resolved with false or gets rejected)
+    return a promise resolved with a false.
```
flackr: This language implies that the above two scrolls are synchronous and we will return a resolved promise, however, I think in practice at least if behavior is smooth we expect both of them to be in progress. I think roughly you want to return a promise that resolves when both promises are resolved with the value true if and only if both dependent promises resolved true.

cssom-view-1/Overview.bs
```diff
@@ -551,12 +562,12 @@ partial interface Window {
...
-    undefined scrollTo(optional ScrollToOptions options = {});
-    undefined scrollTo(unrestricted double x, unrestricted double y);
-    undefined scrollBy(optional ScrollToOptions options = {});
-    undefined scrollBy(unrestricted double x, unrestricted double y);
+    Promise<undefined> scroll(optional ScrollToOptions options = {});
+    Promise<undefined> scroll(unrestricted double x, unrestricted double y);
+    Promise<undefined> scrollTo(optional ScrollToOptions options = {});
+    Promise<undefined> scrollTo(unrestricted double x, unrestricted double y);
+    Promise<undefined> scrollBy(optional ScrollToOptions options = {});
+    Promise<undefined> scrollBy(unrestricted double x, unrestricted double y);
```
flackr, line 598: `Promise<boolean>`?

cssom-view-1/Overview.bs
```diff
@@ -1185,13 +1198,13 @@ partial interface Element {
...
-  undefined scrollTo(unrestricted double x, unrestricted double y);
-  undefined scrollBy(optional ScrollToOptions options = {});
-  undefined scrollBy(unrestricted double x, unrestricted double y);
+  Promise<undefined> scrollIntoView(optional (boolean or ScrollIntoViewOptions) arg = {});
+  Promise<undefined> scroll(optional ScrollToOptions options = {});
+  Promise<undefined> scroll(unrestricted double x, unrestricted double y);
+  Promise<undefined> scrollTo(optional ScrollToOptions options = {});
+  Promise<undefined> scrollTo(unrestricted double x, unrestricted double y);
+  Promise<undefined> scrollBy(optional ScrollToOptions options = {});
+  Promise<undefined> scrollBy(unrestricted double x, unrestricted double y);
```
flackr, line 1329: `Promise<boolean>`?

cssom-view-1/Overview.bs
```diff
@@ -420,22 +420,27 @@ an associated element or [=pseudo-element=] <var>element</var> and optionally a
...
+   'scroll-behavior' property is ''scroll-behavior/smooth'', or
    <li><var>behavior</var> is <code>smooth</code>
   </ul>
-  ...then perform a <a>smooth scroll</a> of <var>box</var> to <var>position</var>. Once the position has finished updating, emit the <a event>scrollend</a> event.
-  Otherwise, perform an <a>instant scroll</a> of <var>box</var> to <var>position</var>. After an <a>instant scroll</a> emit the <a event>scrollend</a> event.
+  then perform a <a>smooth scroll</a> of <var>box</var> to <var>position</var>;
+  otherwise, perform an <a>instant scroll</a> of <var>box</var> to <var>position</var>.
+1. Wait until the position has finished updating.
+1. If the scroll position changed as a result of this call, emit the <a event>scrollend</a> event.
+1. Resolve <var>scrollPromise</var> with true if it is still in the pending state.
```
flackr: This seems a bit clumsy to "wait until the position has finished updating" and then check whether another call resulted in the promise no longer being pending. I wonder if we could change the wait step to wait for the position to finish updating or another call to this method to this method to be made.

cssom-view-1/Overview.bs
```diff
@@ -420,22 +420,27 @@ an associated element or [=pseudo-element=] <var>element</var> and optionally a
...
 
   Note: <code>behavior: "instant"</code> always performs an <a>instant scroll</a> by this algorithm.
 
   Note: If the scroll position did not change as a result of the user interaction or programmatic invocation, where no translations were applied as a result, then no <a event>scrollend</a> event fires because no scrolling occurred.
-</ol>
 
+Issue: What values to use when fulfilling programmatic scroll promises at scroll completion vs interruption? #12355
```
flackr: It seems like it would make more sense to put this with the interface where we will currently say we return a boolean. Also can you make this a bit more descriptive, e.g.

Issue: Should we return a result object and if so what information should it provide? #12355

cssom-view-1/Overview.bs
mustaqahmed: Done without using "another call to this" (because <var>scrollPromise</var> would be settled in Step 2 above in that case).

cssom-view-1/Overview.bs
```diff
@@ -460,8 +463,14 @@ When a user agent is to <dfn for="viewport" export>perform a scroll</dfn> of a <
...
 1. <a for="/">Perform a scroll</a> of the <a>viewport</a>'s <a>scrolling box</a> to its current scroll position + (<var>layout dx</var>, <var>layout dy</var>) with <var>element</var> as the
     associated element, and <var>behavior</var> as the scroll behavior.
+    Let <var>scrollPromise1</var> be the promise returned from this step.
 1. <a for="/">Perform a scroll</a> of <var>vv</var>'s <a>scrolling box</a> to its current scroll position + (<var>visual dx</var>, <var>visual dy</var>) with <var>element</var> as the associated
     element, and <var>behavior</var> as the scroll behavior.
+    Let <var>scrollPromise2</var> be the promise returned from this step.
+1. Wait until either both <var>scrollPromise1</var> and <var>scrollPromise2</var> get resolved with true, or any of them settles in any other state (i.e. gets resolved with false or gets rejected).
+1. If both <var>scrollPromise1</var> and <var>scrollPromise2</var> were resolved with true,
+    return a promise resolved with true.
+    Otherwise return a promise resolved with false.
```
flackr: The promise should be returned well before we know the result. I.e. we don't wait for the scroll to complete before returning the promise.

cssom-view-1/Overview.bs
```diff
@@ -460,8 +463,14 @@ When a user agent is to <dfn for="viewport" export>perform a scroll</dfn> of a <
 1. Let <var>element</var> be <var>doc</var>'s root element if there is one, null otherwise.
 1. <a for="/">Perform a scroll</a> of the <a>viewport</a>'s <a>scrolling box</a> to its current scroll position + (<var>layout dx</var>, <var>layout dy</var>) with <var>element</var> as the
     associated element, and <var>behavior</var> as the scroll behavior.
+    Let <var>scrollPromise1</var> be the promise returned from this step.
 1. <a for="/">Perform a scroll</a> of <var>vv</var>'s <a>scrolling box</a> to its current scroll position + (<var>visual dx</var>, <var>visual dy</var>) with <var>element</var> as the associated
     element, and <var>behavior</var> as the scroll behavior.
+    Let <var>scrollPromise2</var> be the promise returned from this step.
+1. Wait until either both <var>scrollPromise1</var> and <var>scrollPromise2</var> get resolved with true, or any of them settles in any other state (i.e. gets resolved with false or gets rejected).
```
flackr: I think we return scrollPromise2 immediately, but wait on this before resolving the promise.

cssom-view-1/Overview.bs
mustaqahmed: Aah right.  Fixed it here and everywhere else.

emilio:
The boolean stuff is rather confusing. Do we want / is it worth resolving differently for completion vs interruption? Can't stuff technically scroll after the UA resolves the promise, kinda defeating the point of it?

dlrobertson:
What is the relation of this new scroll promise to the [`scrollend`](https://drafts.csswg.org/cssom-view/#eventdef-document-scrollend) event? And are there reasons to use one over the other?

FWIW `scrollend` happens to be in [interop 2025](https://wpt.fyi/interop-2025). The original issue https://github.com/w3c/csswg-drafts/issues/1562 states a need to do the following:

```js
let element = document.getElementById('scroll-container');
element.scroll(0, 400).then(() => {
   // ...scroll has finished
});
```

After interop 2025 completes, the following or something like it should be relatively reliable.

```js
let element = document.getElementById('scroll-container');
let scrollPromise = new Promise(resolve => element.addEventListener("scrollend", resolve, {once: true}));
element.scroll(0, 400);
scrollPromise.then(() => {
   // ...scroll has finished
});
```

mustaqahmed:
Below are my responses, sorry for the delay (I was on vacation for several weeks).

> The boolean stuff is rather confusing. Do we want / is it worth resolving differently for completion vs interruption? Can't stuff technically scroll after the UA resolves the promise, kinda defeating the point of it?

@emilio I agree the [original resolution](https://github.com/w3c/csswg-drafts/issues/1562) does not include such details, and in our first patch we had the IDL type `Promise<undefined>`.  But we found the spec language "resolve a Promise with true/false" far more expressive than "resolve a Promise (with nothing)".  Moreover, this little addition gives the developers as easy way to test a successful call:
```
let scrolled = await elem.scrollBy(10);
```

I think we correctly prevented the possibility of scrolling after promise resolution through the last few steps in our [#perform-a-scroll](https://drafts.csswg.org/cssom-view/#perform-a-scroll) algorithm.  Note that all other scroll algorithms ultimately [rely on this one](https://drive.google.com/file/d/1QLYGrCwylU7mhwdQIrXIXnSFFZCDFfaG/view), and they never resolve any promise with true unless the innermost call does the same.

Please correct me if I missed any call dependency in the spec!

> What is the relation of this new scroll promise to the [`scrollend`](https://drafts.csswg.org/cssom-view/#eventdef-document-scrollend) event? And are there reasons to use one over the other?

@dlrobertson: When multiple scrollers are involved, writing a `scrollend` event based promise is non-trivial.  For example, it is not easy to determine which elements would be affected by the call `elem.scrollIntoView()`, and then wait for all of them to emit `scrollend` events.

emilio:
Sure, but promise resolution is async and, in your example `scrolled` is actually not really whether something scrolled, right? E.g a smooth scroll interrupted by the user will resolve with false but it's not clear what behavior the developer would want... It's more like `completedScroll`, and it seems rather subtle / hard to use right in presence of user scrolls or what not, right? Or am I missing something?

flackr:
> > The boolean stuff is rather confusing. Do we want / is it worth resolving differently for completion vs interruption? Can't stuff technically scroll after the UA resolves the promise, kinda defeating the point of it?
> 
> @emilio I agree the [original resolution](https://github.com/w3c/csswg-drafts/issues/1562) does not include such details, and in our first patch we had the IDL type `Promise<undefined>`. But we found the spec language "resolve a Promise with true/false" far more expressive than "resolve a Promise (with nothing)".

During the discussion in which we took the resolution, this was raised as an important point, from https://github.com/w3c/csswg-drafts/issues/1562#issuecomment-389586317:

> <dael> Rossen_: Meta point, I don't think anyone objects use case of promise pattern. I think the addition to this is adding some data that describes how this was fulfilled.
> <dael> TabAtkins: I can see fulfill with an object that's an enum of completed or interrupted or something like that.

> Sure, but promise resolution is async and, in your example scrolled is actually not really whether something scrolled, right? E.g a smooth scroll interrupted by the user will resolve with false but it's not clear what behavior the developer would want... It's more like completedScroll, and it seems rather subtle / hard to use right in presence of user scrolls or what not, right?

If a user scrolls before we reach the target then it immediately resolves with false. This is because while we have completed the scrolling that is going to happen for that requested scroll, it was interrupted. In the use cases in the OP it was suggested that developers would use this for things like:

> it would be even more useful if scroll method returned a promise so the developer could wait until the scroll finishes before moving on to next set of tasks for things like triggering analytics data after scrolling, unloading resources when they've been scrolled out of view, parallax scrolling, etc.

Having this completion state allows the developer to do something different for interrupted scrolls without needing to read back the scroll offset and try to figure out whether it was interrupted which can be hard when using scrollIntoView to determine exactly what offset you should expect to scroll to.

Personally, I don't think it should be a boolean but rather a dictionary with an enum so that we could add additional arguments if we think of other things you might want to know about the completed scroll.

mustaqahmed:
> It's more like `completedScroll`, and it seems rather subtle / hard to use right in presence of user scrolls or what not, right? Or am I missing something?

You are correct, `completedScroll` would have been a better name in my hastily written example above.

Our discussion seems to deserve a separate resolution in the WG.  If you agree, I would suggest moving this discussion over to #12495 for a wider audience and settle this PR in one of the following ways:
1. Land the PR as-is, quoting #12495.
1. Change this PR to "always resolve with `false` or `undefined`", quoting that same issue.
1. Change this PR to resolve with an enum value which will be defined later on through that issue.
 
@emilio What do you think?

emilio:
I'd probably resolve with undefined for now. Looks good with that. Thanks!

mustaqahmed:
I have switched the PR to `Promise<undefined>`.

Note that [#perform-a-scroll](https://drafts.csswg.org/cssom-view/#perform-a-scroll) still needs a promise that resolves with a Boolean because both [#viewport-perform-a-scroll](https://drafts.csswg.org/cssom-view/#viewport-perform-a-scroll) and [#scroll-a-target-into-view](https://drafts.csswg.org/cssom-view/#scroll-a-target-into-view) need to know when all affected scrolls have completed.  Again, [the dependency of the scroll algorithms](https://drive.google.com/file/d/1QLYGrCwylU7mhwdQIrXIXnSFFZCDFfaG/view?usp=drive_link) would be helpful I believe.

flackr:
> Note that [#perform-a-scroll](https://drafts.csswg.org/cssom-view/#perform-a-scroll) still needs a promise that resolves with a Boolean because both [#viewport-perform-a-scroll](https://drafts.csswg.org/cssom-view/#viewport-perform-a-scroll) and [#scroll-a-target-into-view](https://drafts.csswg.org/cssom-view/#scroll-a-target-into-view) need to know when all affected scrolls have completed. Again, [the dependency of the scroll algorithms](https://drive.google.com/file/d/1QLYGrCwylU7mhwdQIrXIXnSFFZCDFfaG/view?usp=drive_link) would be helpful I believe.

In this model, we shouldn't need the boolean at all. ScrollIntoView just needs to wait until the promises for each of the scrolled scrollers resolve. If another scroll interrupts one or more of those scrolls then those interrupted scroll promises will be resolved immediately.

mustaqahmed:
> In this model, we shouldn't need the boolean at all. ScrollIntoView just needs to wait until the promises for each of the scrolled scrollers resolve. If another scroll interrupts one or more of those scrolls then those interrupted scroll promises will be resolved immediately.

My question is: do the uninterrupted scrollers still continue their scrolling even after one got interrupted?  This is about the observed behavior and not about the returned promise.  To explain in a different way, suppose a ScrollIntoView call starts smooth scrolling in scrollers `A`, `B` and `C`, and then a `B.scrollBy(0)` call interrupts the smooth scrolling of `B`.  Shouldn't `A` and `C` immediately stop scrolling?

flackr:
> My question is: do the uninterrupted scrollers still continue their scrolling even after one got interrupted? This is about the observed behavior and not about the returned promise. To explain in a different way, suppose a ScrollIntoView call starts smooth scrolling in scrollers `A`, `B` and `C`, and then a `B.scrollBy(0)` call interrupts the smooth scrolling of `B`. Shouldn't `A` and `C` immediately stop scrolling?

This is getting beyond the issue of when the promises should be resolved. Whether or not any of the ongoing scrolls is interrupted, the scrollIntoView call should resolve when all of the dependent scrolls' promises have resolved. If they are interrupted, then the overall promise may resolve sooner. We can separately debate how interrupting a scroll should affect ongoing scrolls but I don't see what that has to do with the promise resolution.

That said, this is a simple thing to test, I put together a test case https://codepen.io/flackr/pen/pvgJQxP . Chrome and Safari only interrupt the scroll of B, Firefox stops the scroll of B and C. I think that it doesn't make sense to stop scrolling A since you would still be able to scroll to the position that brings the B scroller into view. Stopping scrolling C (as firefox does) might make sense since it's no longer necessarily going to come into view. Though I could also see an argument for continuing it. Either way, this seems like a completely separate issue.

mustaqahmed:
Yeah, my comment about the observed behavior should have been in #12495.

Done using the simplest possible promise resolution logic, see my latest commit.

cssom-view-1/Overview.bs
```diff
@@ -1572,10 +1570,7 @@ means to run these steps:
         abort any remaining iteration of this loop.
 1. Let <var>scrollPromise</var> be a new promise.
 1. Return <var>scrollPromise</var>, and run the remaining steps <a>in parallel</a>.
-1. Wait until either all promises in <var>ancestorPromises</var> get resolved with true, or any of them settles in any other state (i.e. gets resolved with false or gets rejected).
-1. Resolve <var>scrollPromise</var>.
-1. Resolve with false any promise in <var>ancestorPromises</var> that is still in the pending state.
-
+1. Resolve <var>scrollPromise</var> when all promises in <var>ancestorPromises</var> have been resolved.
```
emilio: Maybe `s/have been resolved/have settled`? not sure if we want to ever reject one of these promises, but that seems nicer? Then again that would eat the rejection. So it's fine.

mustaqahmed: Done, thanks.  While the "inner" promises don't reject so this is functionally equivalent, "settled" seems better for future-proofing against unsettled "outer" promises in case the inner behavior ever changes.

cssom-view-1/Overview.bs
```diff
@@ -420,22 +420,25 @@ an associated element or [=pseudo-element=] <var>element</var> and optionally a
...
-   'scroll-behavior' property is ''scroll-behavior/smooth''
+   'scroll-behavior' property is ''scroll-behavior/smooth'', or
    <li><var>behavior</var> is <code>smooth</code>
   </ul>
-  ...then perform a <a>smooth scroll</a> of <var>box</var> to <var>position</var>. Once the position has finished updating, emit the <a event>scrollend</a> event.
-  Otherwise, perform an <a>instant scroll</a> of <var>box</var> to <var>position</var>. After an <a>instant scroll</a> emit the <a event>scrollend</a> event.
+  then perform a <a>smooth scroll</a> of <var>box</var> to <var>position</var>;
+  otherwise, perform an <a>instant scroll</a> of <var>box</var> to <var>position</var>.
+1. Wait until either the position has finished updating, or <var>scrollPromise</var> has been resolved.
+1. If the scroll position changed as a result of this call, emit the <a event>scrollend</a> event.
```
flackr: If the promise was resolved outside of this path, it means this scroll was interrupted, which means we should defer the scrollend because there is further ongoing scrolling.

cssom-view-1/Overview.bs
```diff
@@ -420,22 +420,25 @@ an associated element or [=pseudo-element=] <var>element</var> and optionally a
 (which is "<code>auto</code>" if omitted),
 the following steps must be run:
 
-<ol>
- <li><a lt="smooth scroll aborted">Abort</a> any ongoing <a>smooth scroll</a> for <var>box</var>.
- <li>If the user agent honors the 'scroll-behavior' property and one of the following are true:
+1. <a lt="smooth scroll aborted">Abort</a> any ongoing <a>smooth scroll</a> for <var>box</var>.
+1. Resolve any pending promise for <var>box</var>.
```
flackr: This should be a bit more specific about what promises its resolving, e.g.

Resolve any pending scroll promises whose scroll container is <var>box</var>.

I'm not sure if we need to explicitly say that we store these in some set or if it can be implicit.

flackr (approved):
A couple minor changes but otherwise looks good.

cssom-view-1/Overview.bs
mustaqahmed: Switched to your suggested line.

mustaqahmed:
@emilio Please let me know if you would like to see any other changes.

cssom-view-1/Overview.bs
```diff
@@ -1538,16 +1558,20 @@ means to run these steps:
...
                         as the scroll behavior.
+                        Add the promise retured from this step in the set <var>ancestorPromises</var>.
             </dl>
     1. If <var>container</var> is not null and <var>scrolling box</var> is a [=shadow-including inclusive ancestor=] of <var>container</var>,
-        abort the rest of these steps.
-
+        abort any remaining iteration of this loop.
+1. Let <var>scrollPromise</var> be a new promise.
+1. Return <var>scrollPromise</var>, and run the remaining steps <a>in parallel</a>.
+1. Resolve <var>scrollPromise</var> when all promises in <var>ancestorPromises</var> have settled.
```
emilio: Would be good to check if we can also like to the `settle` concept here.

cssom-view-1/Overview.bs
```diff
@@ -1538,16 +1558,20 @@ means to run these steps:
...
                     1. Let <var>root element</var> be <var>document</var>'s [=root element=], if there is one, or null otherwise.
                     1. <a for="viewport">Perform a scroll</a> of the <a>viewport</a> to <var>position</var>, with <var>root element</var> as the associated element and <var>behavior</var>
                         as the scroll behavior.
+                        Add the promise retured from this step in the set <var>ancestorPromises</var>.
             </dl>
     1. If <var>container</var> is not null and <var>scrolling box</var> is a [=shadow-including inclusive ancestor=] of <var>container</var>,
-        abort the rest of these steps.
-
+        abort any remaining iteration of this loop.
+1. Let <var>scrollPromise</var> be a new promise.
```
emilio: Probably try to link to `Promise`, does `{{Promise}}` work? It probably needs to be associated to a realm (though in this case it's generally fine to assume that the realm is the owner global of the Element or Viewport object...

cssom-view-1/Overview.bs
mustaqahmed: I avoided "settle" here to exclude [rejection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#:~:text=A%20promise%20is%20said%20to%20be%20settled%20if%20it%20is%20either%20fulfilled%20or%20rejected%2C%20but%20not%20pending) (as per https://github.com/w3c/csswg-drafts/issues/1562).

cssom-view-1/Overview.bs
mustaqahmed: Done.

PR #12362, opened 2025-06-18
[cssom-view-1] Define caretRangeFromPoint()

### LINKED ISSUE #6459: [cssom-view] Specify `document.caretRangeFromPoint`?
Per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/caretRangeFromPoint) it has been supported in Chrome and Safari for many years. While there is an alternative, I guess it should probably be in spec for web compatibility, although I haven't investigated its impact.

keithamus:
This attempts to define `caretRangeFromPoint()` as discussed in #6459. 

This is largely copy/pasted from `caretPositionFromPoint()`, but simply changing the return type to a Range.

Closes #6459.

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -1024,6 +1024,7 @@ partial interface Document {
   Element? elementFromPoint(double x, double y);
   sequence&lt;Element> elementsFromPoint(double x, double y);
   CaretPosition? caretPositionFromPoint(double x, double y, optional <span>CaretPositionFromPointOptions</span> options = {});
+  Range? caretRangeFromPoint(double x, double y);
   readonly attribute Element? scrollingElement;
 };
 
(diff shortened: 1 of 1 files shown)


cssom-view-1/Overview.bs
```diff
@@ -1097,6 +1098,52 @@ result of running these steps:
...
+
+    <dl>
+        <dt><a>range start node</a>
+        <dd>The node corresponding to the text entry widget.
+
+        <dt><a>range end node</a>
+        <dd>The node corresponding to the text entry widget.
+
+        <dt><a>range start offset</a>
+        <dd>The amount of 16-bit units to the left of where the
```
smfr, line 1126: Where does "16-bit units" come from?

keithamus: This is the same spec prose as defined in `caretPositionFromPoint()` - which AFAICT comes from the definition in 532723f180343305397c6ede736854cb90ea6e8e. I don't have further context than that.

cssom-view-1/Overview.bs
```diff
@@ -1097,6 +1098,52 @@ result of running these steps:
...
+        <dt><a>range end offset</a>
+        <dd>The amount of 16-bit units to the left of where the
+        text insertion point indicator would have inserted.
+    </dl>
+1. Otherwise:
+
+    1. Let <var>caretPosition</var> be a <a for=/>tuple</a> consisting of a <var>caretPositionNode</var> (a <a for=/>node</a>) and a <var>caretPositionOffset</var> (a non-negative integer) for the position where the text insertion point indicator would have been inserted when applying
+        the <a>transforms</a> that apply to the descendants of the <a>viewport</a>.
+    1. Let <var>startNode</var> be the <var>caretPositionNode</var> of the <var>caretPosition</var>, and let <var>startOffset</var> be the <var>caretPositionOffset</var> of the <var>caretPosition</var>.
+    1. While <var>startNode</var> is a [=node=], <var>startNode</var>'s [=tree/root=] is a [=shadow root=], and <var>startNode</var>'s [=tree/root=] is not a [=shadow-including inclusive ancestor=] of any of <var>options</var>["{{CaretPositionFromPointOptions/shadowRoots}}"], repeat these steps:
```
emilio, line 1138: Can we reuse `caretPositionFromPoint`'s definition rather than duplicating it?

PR #12374, opened 2025-06-20
[css-mixins-1] Revert to ED

andruud:
Also, fix a markup problem which made the parser unhappy in section 7.1.2.


svgeesus (approved):
Thanks!

PR #12383, opened 2025-06-22
[css-counter-styles] Define extended CJK longhand counter styles range

### LINKED ISSUE #12300: [css-counter-styles] CJK longhand counter styles should support values beyond 9999
The current CSS Counter Styles Level 3 specification intentionally limits all CJK longhand counter styles to a range of `-9999 to 9999`.  While CJK style range extensions are “allowed", there is no clear specification.

The current WPT expects these styles to work with much larger values (up to 1,000,000,000 in the current test case), and there is a fundamental mismatch between the spec's limitation and WPT expectations or real-world requirements for CJK number representation. Moreover, Firefox already has an extension implementation, so a cross-browser incompatibility exists as well.

e.g, japanese-informal) https://wpt.fyi/results/css/css-counter-styles/japanese-informal/counter-japanese-informal.html

### Current Spec Issues

1. **Intentional Range Limitation**: The spec defines `range: -9999 9999;` for all CJK longhand styles with a note that ["opinions differ on how best to represent numbers 10k or greater."](https://drafts.csswg.org/css-counter-styles/#complex-cjk:~:text=Because%20opinions%20differ,cjk%2Ddecimal.) This limitation is problematic because:
   - Unlike the Western system of 1,000 units, East Asian languages use myriad-based systems with natural breakpoints at 10,000 (万/萬/만), 100,000,000 (億/억), etc.
     - more details can be found here:  [Japanese numerals](https://en.wikipedia.org/wiki/Japanese_numerals)
   - This structural difference makes implementation using the current additive system difficult
   - The limitation prevents proper representation of common large numbers in CJK languages
   - All affected counter styles share similar structural patterns for large numbers

2. **Incomplete additive-symbols**: The current definition only goes up to 9000 on purpose:
   ```css
   additive-symbols: 9000 \4E5D\9621, 8000 \516B\9621, ... /* stops at thousands */
   ```

3. **Test-Spec Mismatch**:

| number | Actual(Spec) | Expected(WPT) |
| ---- | ---- | ---- |
| 10,000 | 10000 | 壱萬 |
| 100,000| 100000 | 壱拾萬|
| 1,000,000 | 1000000 | 壱百萬 |
| 100,000,000 | 100000000 | 壱億 |
| 1,000,000,000  | 1000000000 | 壱拾億 |

### Browser Implementation Status

- ✅ Gecko: Supports extended ranges beyond the spec
- ❌ WebKit:  Follows spec limitation and fails WPT tests for values > 9999
  -  CJK counterstyle support was added in 2021. It uses the `toPredefinedAdditiveSystem()` to implement the additive system algorithm. (see [[Bug 227206](https://bugs.webkit.org/show_bug.cgi?id=227206)](https://bugs.webkit.org/show_bug.cgi?id=227206)). 
- ❌ Chromium:   Follows spec limitation and fails WPT tests for values > 9999 

This creates inconsistent cross-browser behaviour, and specs side should better to be fixed because WPT expectations are natural (, moreover, current outputs are quite not true) in the real world.

### Proposed Solutions

1. **The range restriction relaxation**
   ```css
   range: auto; /* or -999999999 999999999 to pass the current testcases at least*/
   ```
   Allow implementations to extend support while maintaining backward compatibility.

2. **Extend additive-symbols**
   Add symbols for myriad units (万/萬/만) and higher denominations to properly support CJK numbering systems. Example for Japanese:
   ```css
   additive-symbols: 
     100000000 \58F1\5104,        /* 壱億 */
     10000000 \58F1\9621\4E07,    /* 壱阡萬 */
     1000000 \58F1\767E\4E07,     /* 壱百萬 */
     100000 \58F1\62FE\4E07,      /* 壱拾萬 */
     10000 \58F1\4E07,            /* 壱萬 */
     /* ... existing symbols ... */
   ```
   Similar extensions would be needed for Korean and Chinese variants.
   Then, fallback to extended CJK style instead of `cjk-decimal`.

3. **Long-term: Consider a new algorithmic approach for CJK**
   For CJK counter styles that follow predictable patterns, a CJK-specific algorithmic system might be more appropriate than additive for very large numbers. e.g, `system: cjk-algorithmic;`?

### References

- Current spec: https://drafts.csswg.org/css-counter-styles/#complex-predefined-counters
- WPT test expecting large values: [[counter-japanese-formal.html](https://github.com/web-platform-tests/wpt/blob/master/css/css-counter-styles/counter-japanese-formal.html)](https://github.com/web-platform-tests/wpt/blob/master/css/css-counter-styles/counter-japanese-formal.html)
- Related discussion on range limitations: #2479

### Impact

This change would:
- Align the specification with actual test expectations
- Enable proper CJK number formatting for common use cases across East Asian languages
- Improve cross-browser compatibility
- (Better support internationalization requirements for a significant portion of the world's population)

Note: While this issue uses Japanese examples, the same structural issues and solutions would apply to all CJK longhand counter styles defined in the specification. 

I would like to see everyone implement the same range for CJK in the right way and have it specified in the specification.

cc: @nt1m

sakupi01:
This attempts to define extended CJK longhand counter styles with clear implementation separation as discussed in #12300.

Closes #12300.

## Change

This is mainly copy/pasted from an older version of the spec, which defined counter values outside this limited range and clearly separated required/optional implementations.
https://www.w3.org/TR/2013/WD-css-counter-styles-3-20130718/#extended-cjk

However, I made a slight change for the Korean range.
The former version of the spec [defines a range between 0 and 10^16](https://www.w3.org/TR/2013/WD-css-counter-styles-3-20130718/#extended-cjk:~:text=the%20Korean%20styles%20are%20defined%20for%20all%20non%2Dnegative%20numbers%20less%20than%201016.), yet the new version defines it down to a negative range because the current ED spec already defines the Korean negative range. See: 
https://drafts.csswg.org/css-counter-styles/#limited-korean

## The current browser behaviour

The cross-browser implementation for the *limited* one is consistent, but the *extended* one is inconsistent.
The new version of spec makes a clear difference between limited and optional, so that implementation and test separation makes sense.
The current implementation status that I could observe is as follows:

- ✅ Gecko: Partially supports extended ranges beyond the spec, but not fully up to 10^16
- ❌ WebKit: Supports spec limited range and fails tests for values > 9999
- ❌ Chromium: Supports spec limited range and fails tests for values > 9999

The Korean negative range is also supported in all 3 browsers.
FWIW, here's a Chromium implementation, according to the current ED spec (not the former one).
https://chromium.googlesource.com/chromium/src/+/d0be3bc63c34b2a904be717bfa3bf0be4e71eb45

## WPT

WPT test will also be updated with:
- Clear separation between limited(*required*) and extended(*optional*) range.
  -  limited(*required*) : counter-[cjk]-*.html
      - e.g; [counter-japanese-formal.html](https://github.com/web-platform-tests/wpt/pull/53312/files#diff-7d745ba2f9bb8db4ac70496aebed2c743cdace9f3d3425a4edc28c18bf872939)
  - extended(*optional*): counter-[cjk]-extended-*.html
    - e.g; [counter-japanese-formal-extended.html](https://github.com/web-platform-tests/wpt/pull/53312/files#diff-dfce9b315a420779230d69863af53804dbff3e23563d918aa91213be813f52b1)
- Addition of extended CJK longhand range support which covers its algorithm
  - https://www.w3.org/TR/2013/WD-css-counter-styles-3-20130718/#extended-cjk

ref: https://github.com/web-platform-tests/wpt/pull/53312/

cc: @nt1m

### DIFF
--- a/css-counter-styles-3/Overview.bs
+++ b/css-counter-styles-3/Overview.bs
@@ -2367,7 +2367,7 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
 
 	This entire section is normative, but <strong>optional</strong>. User-agents may ignore it and still be conformant. If a user-agent implements some of the extended forms described in this section, they must be implemented as described here.
 
-	The Chinese longhand styles are defined out to 10k with a [=Chinese|specialized algorithm=], while the Japanese and Korean longhand styles are defined similarly as ''additive'' styles. However, these styles are defined out to 10<sup>16</sup> in common usage. The following section describes an alternative algorithm for these styles.
+	The Chinese longhand styles are defined out to 10k with a [[#limited-chinese|specialized algorithm]], while the Japanese and Korean longhand styles are defined similarly as ''additive'' styles. However, these styles are defined out to 10<sup>16</sup> in common usage. The following section describes an alternative algorithm for these styles.
 
 	All of the Chinese, Japanese, and Korean styles are defined for all numbers between -10<sup>16</sup> and 10<sup>16</sup>, exclusive. For numbers outside this range, the ''cjk-decimal'' style is used. All of the styles are defined by almost identical algorithms (specified as a single algorithm here, with the differences called out when relevant), but use different sets of characters. The list following the algorithm gives the name of each counter style using this algorithm, and the individual character sets used by each style.
 
--- a/css-counter-styles-3/Overview.bs
+++ b/css-counter-styles-3/Overview.bs
@@ -2041,7 +2041,7 @@ Longhand East Asian Counter Styles</h3>
 	The Korean and Japanese variants of these counter styles can,
 	if limited to the range of -9999 to 9999,
 	be expressed as ''@counter-style'' rules.
-	The Chinese variants use a [=Chinese|specialized algorithm=],
+	The Chinese variants use a [[#limited-chinese|specialized algorithm]],
 	regardless of range.
 	The required implementation is described in [[#limited-range-required]],
 	which defines the styles over these limited ranges,
--- a/css-counter-styles-3/Overview.bs
+++ b/css-counter-styles-3/Overview.bs
@@ -1914,7 +1914,8 @@ Longhand East Asian Counter Styles</h3>
 	as their characters are more difficult to alter into each other.
 
 	<div class="example">
-		The following table shows examples of these styles, particularly some ways in which they differ.
+		The following table shows examples of these styles,
+		particularly some ways in which they differ.
 
 		<table class="data">
 			<thead>
@@ -2034,15 +2035,19 @@ Longhand East Asian Counter Styles</h3>
 		</table>
 	</div>
 
-	Because opinions differ on how best to represent numbers 10,000 or greater 
-	using the longhand CJK styles, the implementation details are split into two sections. 
-	The Korean and Japanese variants of these counter styles can, if limited to the range 
-	of -9999 to 9999, be expressed as ''@counter-style'' rules. 
-	The Chinese variants use a [=Chinese|specialized algorithm=], regardless of range. 
-	The required implementation is described in [[#limited-range-required]], which defines 
-	the styles over these limited ranges, and the optional implementation is described in 
-	[[#extended-range-optional]], which defines them over a larger range following custom algorithms. 
-	Outside the implementation-supported range, 
+	Because opinions differ on how best to represent numbers 10,000 or greater
+	using the longhand CJK styles,
+	the implementation details are split into two sections.
+	The Korean and Japanese variants of these counter styles can,
+	if limited to the range of -9999 to 9999,
+	be expressed as ''@counter-style'' rules.
+	The Chinese variants use a [=Chinese|specialized algorithm=],
+	regardless of range.
+	The required implementation is described in [[#limited-range-required]],
+	which defines the styles over these limited ranges,
+	and the optional implementation is described in [[#extended-range-optional]],
+	which defines them over a larger range following custom algorithms.
+	Outside the implementation-supported range,
 	the fallback is ''cjk-decimal''.
 
 <!--
@@ -2185,30 +2190,35 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
 	</dl>
 
 	The Chinese longhand styles are defined by almost identical algorithms
-	(specified as a single algorithm here, with the differences called out when relevant),
+	(specified as a single algorithm here,
+	with the differences called out when relevant),
 	but use different sets of characters,
 	as specified by the table following the algorithm.
 
 	<ol>
-		<li>If the counter value is 0, the representation is the character for 0
-		specified for the given counter style.  Skip the rest of this algorithm.
+		<li>If the counter value is 0,
+		the representation is the character for 0 specified for the given counter style.
+		Skip the rest of this algorithm.
 
-		<li>Initially represent the counter value as a decimal number.  For each
-		digit that is not 0, append the appropriate digit marker to the digit.
+		<li>Initially represent the counter value as a decimal number.
+		For each digit that is not 0,
+		append the appropriate digit marker to the digit.
 		The ones digit has no marker.
 
-		<li>For the informal styles, if the counter value is between ten and
-		nineteen, remove the tens digit (leave the digit marker).
+		<li>For the informal styles,
+		if the counter value is between ten and nineteen,
+		remove the tens digit (leave the digit marker).
 
-		<li>Drop any trailing zeros and collapse any remaining zeros into a single
-		zero digit.
+		<li>Drop any trailing zeros
+		and collapse any remaining zeros into a single zero digit.
 
-		<li>Replace the digits 0-9 with the appropriate character for the given
-		counter style.  Return the resultant string as the representation of the
-		counter value.
+		<li>Replace the digits 0-9 with the appropriate character
+		for the given counter style.
+		Return the resultant string as the representation of the counter value.
 	</ol>
 
-	For all of these counter styles, the '@counter-style/suffix' is "、" U+3001,
+	For all of these counter styles,
+	the '@counter-style/suffix' is "、" U+3001,
 	the '@counter-style/fallback' is ''cjk-decimal'',
 	the '@counter-style/range' is ''-9999 9999'',
 	and the '@counter-style/negative' value is given in the table of symbols for each style.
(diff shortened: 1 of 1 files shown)


css-counter-styles-3/Overview.bs
```diff
@@ -2361,6 +2361,359 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
...
+
+	This entire section is normative, but <strong>optional</strong>. User-agents may ignore it and still be conformant. If a user-agent implements some of the extended forms described in this section, they must be implemented as described here.
+
+	The Chinese longhand styles are defined out to 10k with a [=Chinese|specialized algorithm=], while the Japanese and Korean longhand styles are defined similarly as ''additive'' styles. However, these styles are defined out to 10<sup>16</sup> in common usage. The following section describes an alternative algorithm for these styles.
+
+	All of the Chinese, Japanese, and Korean styles are defined for all numbers between -10<sup>16</sup> and 10<sup>16</sup>, exclusive. For numbers outside this range, the ''cjk-decimal'' style is used. All of the styles are defined by almost identical algorithms (specified as a single algorithm here, with the differences called out when relevant), but use different sets of characters. The list following the algorithm gives the name of each counter style using this algorithm, and the individual character sets used by each style.
```
sakupi01: The former version of the spec defines a range between 0 and 10^16 for Korean, yet the new version defines it down to a negative range because the current ED spec already defines the Korean negative range.
https://drafts.csswg.org/css-counter-styles/#limited-korean

FWIW, the Korean negative range has been implemented as well. See: https://issues.chromium.org/issues/40504326#comment75

If there's any clue why the former version was limited to the positive range for Korean, and why it still needs to be regulated in the current extended version, please let me know.

nt1m:
Amazing, thanks! I'll tag @fantasai since she's more familiar with this spec than I am.

svgeesus:
If this is

1. What the spec used to say
2. What (at least one) implementation does
3. What WPT tests for

Then review should be easy and fast.

sakupi01:
@svgeesus 
I updated the OP with clearer descriptions.
Thank you for your feedback! I hope it makes sense.

svgeesus:
@fantasai does this look good?

css-counter-styles-3/Overview.bs
```diff
@@ -2184,7 +2184,7 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
 			(It exists for legacy reasons.)
 	</dl>
 
-	The Chinese longhand styles are defined by almost identical algorithms
+	The <dfn dfn>Chinese</dfn> longhand styles are defined by almost identical algorithms
```
fantasai: Don't do this. If you need a subsection, create one (or more), but turning 'Chinese' into a definition for something that isn't the definition of 'Chinese' as a word is confusing.

css-counter-styles-3/Overview.bs
```diff
@@ -2361,6 +2361,359 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
 		</pre>
 	</div>
 
+<h4 id='extended-range-optional'>Extended Implementation (optional)</h4>
+
+	Some counter styles described in earlier chapters have been limited to an artifically small (though still useful) range to reduce the overall complexity of the spec and the task of implementing those styles. However, some implementations might consider the extra complexity worthwhile for the additional range it offers to authors. To accomodate this, this section describes how to extend the limited counter-styles to a larger range.
+
+	This entire section is normative, but <strong>optional</strong>. User-agents may ignore it and still be conformant. If a user-agent implements some of the extended forms described in this section, they must be implemented as described here.
+
+	The Chinese longhand styles are defined out to 10k with a [=Chinese|specialized algorithm=], while the Japanese and Korean longhand styles are defined similarly as ''additive'' styles. However, these styles are defined out to 10<sup>16</sup> in common usage. The following section describes an alternative algorithm for these styles.
```
fantasai: Linking like this should work.

```suggestion
	The Chinese longhand styles are defined out to 10k with a [[#limited-chinese|specialized algorithm]], while the Japanese and Korean longhand styles are defined similarly as ''additive'' styles. However, these styles are defined out to 10<sup>16</sup> in common usage. The following section describes an alternative algorithm for these styles.
```

css-counter-styles-3/Overview.bs
```diff
@@ -2361,6 +2361,359 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
...
+			</li></ul>
+
+		</li><li>For the Korean styles, insert a space (" " U+0020) between each group.
+
+		</li><li>Replace the digits 0-9 with the appropriate character for the given
+		counter style.  Return the resultant string as the representation of the
+		counter value.
+	</li></ol>
+
+	For all of these counter styles, the '@counter-style/suffix' is "、" U+3001, the '@counter-style/fallback' is ''cjk-decimal'', and the '@counter-style/negative' is given in the tables below, or else is the initial value of the descriptor. The '@counter-style/range' is ''-9999999999999999 9999999999999999'' (-10<sup>16</sup>+1 and 10<sup>16</sup>-1).
```
fantasai: Aside from the range, this is all duplicate of the earlier information, so let's just say all the descriptors are the same as for the limited range variants, aside from the range, which you're defining here.

fantasai (changes requested):
The introductory text could maybe be simplified a bit, but that can be done as a follow-up commit. Otherwise it looks pretty good. A few specific comments, and then two general ones on formatting:

- Remove all optional closing tags (e.g. `</td>`, `</tr>`, `</li>`).
- Use [semantic line breaking](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

fantasai:
(Feel free to [contact me directly](https://fantasai.inkedblade.net/contact) for re-review; GH notifications are a lost cause currently... Someday maybe I'll dig myself out of that.)

fantasai:
Also thank you for drafting this up! <3

sakupi01:
@fantasai 

Thanks for reviewing!
I've added some follow-up commits dealing with your suggestions. Thanks!

svgeesus:
I emailed @fantasai requesting re-review. Thanks both for your patience and for the PR.

css-counter-styles-3/Overview.bs
```diff
@@ -2030,22 +2031,25 @@ Longhand East Asian Counter Styles</h3>
...
+	the implementation details are split into two sections.
+	The Korean and Japanese variants of these counter styles can,
+	if limited to the range of -9999 to 9999,
+	be expressed as ''@counter-style'' rules.
+	The Chinese variants use a [=Chinese|specialized algorithm=],
+	regardless of range.
+	The required implementation is described in [[#limited-range-required]],
+	which defines the styles over these limited ranges,
+	and the optional implementation is described in [[#extended-range-optional]],
+	which defines them over a larger range following custom algorithms.
```
fantasai:
```suggestion
	which defines them over a larger range using custom algorithms.
```

css-counter-styles-3/Overview.bs
```diff
@@ -2361,6 +2371,383 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
...
+
+	This entire section is normative, but <strong>optional</strong>.
+	User-agents may ignore it and still be conformant.
+	If a user-agent implements some of the extended forms described in this section,
+	they must be implemented as described here.
+
+	The Chinese longhand styles are defined out to 10k with a [[#limited-chinese|specialized algorithm]],
+	while the Japanese and Korean longhand styles are defined similarly as ''additive'' styles.
+	However, these styles are defined out to 10<sup>16</sup> in common usage.
+	The following section describes an alternative algorithm for these styles.
```
fantasai: I think we can drop this paragraph. It's already explained in the section above.

css-counter-styles-3/Overview.bs
```diff
@@ -2361,6 +2371,383 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
...
+
+	All of the Chinese, Japanese, and Korean styles are defined
```
fantasai:
```suggestion
	All of the Chinese, Japanese, and Korean styles are defined here
```

css-counter-styles-3/Overview.bs
```diff
@@ -2361,6 +2371,383 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
...
+
+		<li>For the Korean styles,
+		insert a space (" " U+0020) between each group.
+
+		<li>Replace the digits 0-9 with the appropriate character
+		for the given counter style.
+		Return the resultant string as the representation of the counter value.
+	</ol>
+
+	For all of these counter styles, the descriptors are the same as for the [[#limited-range-required|limited range variants]], except for the '@counter-style/range', which is ''-9999999999999999 9999999999999999'' (-10<sup>16</sup>+1 and 10<sup>16</sup>-1).
```
fantasai:
```suggestion
	For all of these counter styles, the descriptors are the same as for the [[#limited-range-required|limited range variants]],
	except for the '@counter-style/range', which is ''-9999999999999999 9999999999999999'' (-10<sup>16</sup>+1 and 10<sup>16</sup>-1).
```

fantasai (approved):
A few nits, but otherwise looks great!

Remember to SQUASH when merging (see drop-down on the side of the merge button).

css-counter-styles-3/Overview.bs
```diff
@@ -2030,22 +2031,25 @@ Longhand East Asian Counter Styles</h3>
...
-	Because opinions differ on how best to represent numbers 10k or greater
+	Because opinions differ on how best to represent numbers 10,000 or greater
 	using the longhand CJK styles,
-	all of the counter styles defined in this section are defined to have a range of -9999 to 9999,
-	but implementations may support a larger range.
+	the implementation details are split into two sections.
+	The Korean and Japanese variants of these counter styles can,
+	if limited to the range of -9999 to 9999,
+	be expressed as ''@counter-style'' rules.
+	The Chinese variants use a [=Chinese|specialized algorithm=],
```
tabatkins:
```suggestion
	The Chinese variants use a [[#limited-chinese|specialized algorithm]],
```

css-counter-styles-3/Overview.bs
tabatkins:
```suggestion
```

css-counter-styles-3/Overview.bs
```diff
@@ -2361,6 +2371,380 @@ Chinese: ''simp-chinese-informal'', ''simp-chinese-formal'', ''trad-chinese-info
...
+		<li>For the Korean styles,
+		insert a space (" " U+0020) between each group.
+
+		<li>Replace the digits 0-9 with the appropriate character
+		for the given counter style.
+		Return the resultant string as the representation of the counter value.
+	</ol>
+
+	For all of these counter styles, the descriptors are the same as for the [[#limited-range-required|limited range variants]],
+	except for the '@counter-style/range', which is ''-9999999999999999 9999999999999999'' (-10<sup>16</sup>+1 and 10<sup>16</sup>-1).
```
tabatkins:
```suggestion
	except for the '@counter-style/range', which is ''calc(-1 * pow(10, 16) + 1) calc(pow(10, 16) - 1)''.
```

PR #12402, opened 2025-06-26
[css-ui-4] define appearance: base-select

josepharhar:
This was resolved here:
https://github.com/w3c/csswg-drafts/issues/10440#issuecomment-2332063450

Base appearance for the select element and its picker is going to be published in the HTML spec very soon.

### DIFF
--- a/css-ui-4/Overview.bs
+++ b/css-ui-4/Overview.bs
@@ -2358,16 +2358,16 @@ Switching appearance: the 'appearance' property</h3>
 					for defining which <a>widgets</a> have a <a>base appearance</a>.
 			</dl>
 
-     <dt><dfn>base-select</dfn>
-     <dd>
-       The effect of ''appearance/base-select'' depends on the element it is applied to:
-       <dl class=switch>
-         : The <{select}> element or ''::picker(select)''
-         :: The element is rendered as for ''appearance/base''.
+		<dt><dfn>base-select</dfn>
+			<dd>
+				The effect of ''appearance/base-select'' depends on the element it is applied to:
+				<dl class=switch>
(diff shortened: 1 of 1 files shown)


css-ui-4/Overview.bs
```diff
@@ -2358,6 +2358,19 @@ Switching appearance: the 'appearance' property</h3>
 					for defining which <a>widgets</a> have a <a>base appearance</a>.
 			</dl>
 
+     <dt><dfn>base-select</dfn>
+     <dd>
+       The effect of ''appearance/base-select'' depends on the element it is applied to:
+       <dl class=switch>
```
dbaron: I'd suggest two changes here:

* first, switch the order of the two items in this list so that the second one can just be "Everything else" (and you don't have to repeat "other than the `<select>` element" in the definition either).
* second, for the `<select>`/`::picker(select)` case, I think it's better to fully defer to the definition of `appearance/base` (as you do for the other half) rather than trying to repeat it.

josepharhar: Done and done, thanks!

dbaron (commented):
Looks good now, though I think it would be good if you could also fix the indentation so it matches the surroundings (which I think likely uses tabs -- if so you should also use tabs).

dbaron (approved):
(I thought I hit the "Approve" radio button last time... but see above comment.)

josepharhar:
> Looks good now, though I think it would be good if you could also fix the indentation so it matches the surroundings (which I think likely uses tabs -- if so you should also use tabs).

Done, thanks!

dbaron (approved):

PR #12403, opened 2025-06-26
[css-lists-3] Fix typo under ::marker Pseudo-Element

Krinkle:
> The marker box is generated by the ::marker pseudo-element of a list item as the list item’s first child, before the ::before pseudo-element (if it exists on the element). It is filled with content as defined in § 3.2 Generating Marker Contents.

I assume the word "as" is mean to be "is". The sentence feels incomplete otherwise.


tabatkins:
No, "as" is correct there. It's generated as the first child.

Krinkle:
@tabatkins I see. PEBCAK. I re-read it a dozen times but still managed to miss the first "is".

I was reading it as:
> The [thing] generated by [element], **as** [being] the item's first child, before [other thing] —."

Changing "as" to "is" here would fix that to read:

> The [thing] generated by [element], **is** the item's first child, before [other thing].

I totally missed there was an "is" in first segment already. Sorry about that.

PR #12404, opened 2025-06-27
[selectors-5]: Update heading specs based on resolutions in #10296

keithamus:
I merged #11836 a little hastily, as it does not reflect some of the resolutions we had in the WG meetings (https://github.com/w3c/csswg-drafts/issues/10296#issuecomment-2775988542), specifically:

- The spec needs to make note of the specificity of these selectors. It now does.
- `:heading()` can have comma separated An+B values, so needs to define a production that accommodates that. It now does.


zcorpan (approved):

tabatkins:
While I don't object to specifying it due to the discussion of the specificity, we usually don't mention specificity of pseudo-classes *unless* it's different from that of a class, as that's the generic default behavior.

Issue #12159, opened 2025-05-05
What should `pseudo(type)` return for elements that can't have pseudo elements?


danielsakhapov:
Full context: https://github.com/danielsakhapov/CSSPseudoElementDoc?tab=readme-ov-file#22-elements-without-pseudos

Description:
[CSSPseudoElement](https://www.w3.org/TR/css-pseudo-4/#CSSPseudoElement-interface) object is returned from [pseudo(type)](https://www.w3.org/TR/css-pseudo-4/#window-interface) if `type` is parsed correctly, but what about elements that can't have pseudo elements - e.g. `<input>` and `::before`?

Possible solutions:
* Return a `CSSPseudoElement` object (whose state/computed style then reflects non-generation)
    * **Pros:**
        * API Consistency: `Element.pseudo(validPseudoTypeString)` always returns a `CSSPseudoElement` object. `null` is reserved only for invalid/unrecognized type strings.
        * Separation of Concerns: `Element.pseudo()` provides the handle; [`.exists`](https://github.com/w3c/csswg-drafts/issues/12158) (potentially) on that handle reveals rendering state as per CSS rules.
        * Robust Introspection: Allows developers to get the object and check its computed style (e.g., `display: "none"`, `content: "none"`) to confirm and understand its non-rendered state.
        * Alignment with Highlight Pseudos: Consistent with `::selection` where an object can exist even if no selection is active.
        * Forward Compatibility: If CSS rules change, the API behavior for `Element.pseudo()` remains stable.
    * **Cons:**
        * Indirect "Existence" Check: Developers need to get the object and then inspect its computed styles(or potentially [`.exists`](https://github.com/w3c/csswg-drafts/issues/12158)) to confirm it's not actually rendered, rather than just checking for `null`.
        * Potentially "Useless" Objects: Creates an object that might not correspond to anything visible or affecting layout.
* Return `null`
    * **Pros:**
        * Direct Feedback: Immediately signals to the developer that this specific combination of element and pseudo element type is not valid or won't render.
        * Simpler "Existence" Check: `null` means it doesn't exist in this context; no further checks needed on a pseudo-object.
    * **Cons:**
        * Inconsistent API Signature: `Element.pseudo()` would return `null` for two different reasons (invalid pseudo selector string or valid selector on an incompatible element), making error handling more complex.
        * Less Introspection: Prevents developers from getting a handle to inspect why it doesn't render (e.g., to check default computed styles for such a conceptual pseudo).

[...]

PR #12406, opened 2025-06-27
[css-pseudo-4] Add a note explaining Element.pseudo() behaviour #12159

danielsakhapov:
As per resolution in #12159, Element.pseudo() returns CSSPseudoElement even when invoked on Element that can't generate the requested type.

### DIFF
--- a/css-pseudo-4/Overview.bs
+++ b/css-pseudo-4/Overview.bs
@@ -1829,6 +1829,12 @@ Additions to the CSS Object Model</h2>
   (The UA may drop or regenerate the object for convenience or performance
   if this is not observable.)
 
+  Note: {{CSSPseudoElement}} is returned even when this method is invoked on elements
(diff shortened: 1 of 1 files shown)


astearns:
The note is good to explain and reference the reasoning why, but notes are non-normative. We’ll also need a normative change to the algorithm

danielsakhapov:
The algorithm will be unchanged, won't it? It still returns a `CSSPseudoElement` object if the selector is valid. 
I started the issue and planned this note as just an additional info to clarify the case like e.g. `<input>.pseudo(::before)` returns `CSSPseudoElement` even though `<input>` can't generate `::before:`.
My point is that per current algorithm description says: the selector `::before` is valid, so it returns `CSSPseudoElement`. I don't see what changes we should add with that resolution, could you, please, explain?

danielsakhapov:
@astearns gentle ping

Issue #12267, opened 2025-06-03
[css-forms-1] Rework default styles for slider-like controls


nt1m:
<img width="677" alt="Image" src="https://github.com/user-attachments/assets/35a3569a-f2ed-47ee-8d11-6c95ccfc2e4b" />

I think we might want to do these changes:
- input type range track should be smaller than the thumb
- thumb should be round
- switch should probably get border-radius by default

I think this will make the base appearance styles for the slider easier to customize by default, since developers won't have to override the track to make it smaller, or set the border-radius of thumb.

cc @lukewarlow

nt1m:
cc @jensimmons @fantasai

jensimmons:
We should make these controls look like real, working controls. 

Design Principle 2:
> The controls are recognizable and usable on their own without additional styles.

Right now, they are very hard to understand. They do not look like typical UI.

lukewarlow:
I definitely agree we probably want border radius on the thumb both for range and for switch (as well as probably switch's track/container). As for whether the thumb is larger than the track or not for range I'm fairly agnostic but if it helps make them more identifiable than I'm in favour.

jensimmons:
We could easily design something simple, elegant, very easy to use, generic enough to blend into any website. To do so, we'll probably use `border-radius` and `box-shadow` and lot of other CSS.

Conversely, another design principle is:
> The styles are easily adapted to the website’s branding, without needing complex reset stylesheets:
> - They use minimal code and are easy to override.
> - They do not have a strong voice & tone of their own, and are visually as simple as possible.

This principle does make folks believe we should not use rounded corners and drop shadows, that it's better to leave things unstyled — even if they look raw.

I'm hoping we can thread the needle, but this does raise a question:

If we do use some CSS to give the default some good & useful styling, but that requires developers to override it to apply their own styles (removing a drop shadow for instance), is that putting too much of a burden on the developer???

Thoughts?

tabatkins:
I agree that "minimal" doesn't have to mean "literally the absolute minimum that makes it not broken"; there's room to do a little bit of styling so it's *usable*. We want to be *careful*, because every property we add is another thing authors have to deal with, but I think a border-radius and possibly a box-shadow is likely fine. It just needs to be things that are obvious to spot, and obvious what element and property to target to revert it.

nt1m:
[...]

PR #12427, opened 2025-07-01
[css-forms-1] Add initial UA styles for switch

lukewarlow:
See #12267


lukewarlow:
I'm sure some of these can be reworked into the more generic selectors such as the border-radius on the thumb but for now I think it makes sense to do just the switch ones and then we can build it up.

css-forms-1/Overview.bs
```diff
@@ -829,6 +841,11 @@ ISSUE: Refine meter, progress, switch and range input styling.
   inline-size: 0;
 }
 
+input[type=checkbox][switch]::slider-fill {
+  background-color: transparent;
+  border-radius: inherit;
```
lukewarlow, line 882: Not really sure if this should be needed, for my range input prototype it isn't but for switch it seems to be else the fill overflows the track.

We can adjust accordingly in future if it's not needed.

css-forms-1/Overview.bs
```diff
@@ -813,13 +813,25 @@ select {
...
 ```css
+input[type=checkbox][switch] {
+  display: inline-flex;
+  position: relative;
+  align-items: center;
+  width: 2em;
+}
+
 ::slider-track {
   height: 1em;
```
lukewarlow: Probably need some thought on how to meet WCAG target sizing here. Same goes for existing checkbox styling I believe.

css-forms-1/Overview.bs
```diff
@@ -837,6 +854,16 @@ ISSUE: Refine meter, progress, switch and range input styling.
...
+
+input[type=checkbox][switch]::slider-thumb {
+  border-radius: 100%;
+  height: 1em;
+  position: absolute;
+}
+
+input[type=checkbox][switch]:checked::slider-thumb {
+  inset-inline-end: 0;
+}
```
nt1m, line 909: I wonder if we could use grid and align properties somehow here.

nt1m (commented):
This looks good by itself, I do think we should do this at the same for other controls too though, it's very likely that range inputs (and potentially other slider based controls) end up looking similar in terms of border-radius.

lukewarlow:
> This looks good by itself, I do think we should do this at the same for other controls too though,

Okay I'll make this draft and merge my other PR into this one so it can all be judged together.

css-forms-1/Overview.bs
lukewarlow: I suspect we'll still end up using absolute positioning for range inputs though. So it depends how much cleaner it actually is.

css-forms-1/Overview.bs
```diff
@@ -829,13 +849,24 @@ ISSUE: Refine meter, progress, switch and range input styling.
   inline-size: 0;
 }
 
+input[type=checkbox][switch]::slider-fill {
+  background-color: transparent;
```
lukewarlow: This should likely be changed. Per fantasais comment on the issue. I also can't comment on the correct line but we need to change the range file style too.

@nt1m do you happen to recall where the grey comes from in the screenshots?

css-forms-1/Overview.bs
```diff
@@ -877,8 +877,15 @@ progress:indeterminate::slider-fill {
 }
 
 input[type=checkbox][switch]::slider-fill {
-  background-color: transparent;
+  background-color: color-mix(in lab, currentColor 25%, transparent);
```
lukewarlow: TODO drop the `in lab` per https://github.com/w3c/csswg-drafts/issues/13496

css-forms-1/Overview.bs
nt1m: Sorry, I forgot to get back to this. I think the grays are usually a mix of currentColor and transparency

css-forms-1/Overview.bs
```diff
@@ -837,14 +837,31 @@ select {
...
-ISSUE: Refine meter, progress, switch and range input styling.
+ISSUE(12267): Refine meter, progress, switch and range input styling.
 
 ```css
+input[type="checkbox][switch],
+input[type="range"] {
+  display: inline-flex;
+  position: relative;
+  align-items: center;
+  width: 100%;
```
lukewarlow: Shoudl this be inline-size instead?

css-forms-1/Overview.bs
```diff
@@ -859,14 +876,32 @@ progress:indeterminate::slider-fill {
   inline-size: 0;
 }
 
+input[type=checkbox][switch]::slider-fill {
+  background-color: color-mix(currentColor 25%, transparent);
```
lukewarlow, line 881: Should this apply to range too?

css-forms-1/Overview.bs
nt1m: Very likely, these controls support vertical writing mode

nt1m: In general, the existing uses of physical sizing properties are probably wrong

lukewarlow:
@nt1m does this look good to merge? We'll obviously need follow-ups for stuff like WCAG touch sizing and actually positioning the thumbs and stuff.

PR #12429, opened 2025-07-01
[css-forms-1] Range styles

lukewarlow:
Follow up to #12427 

(Image includes an implementation of sizing and background color for the fill, and positioning of the thumb which aren't included in this PR)

image

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -816,19 +816,27 @@ select {
 ISSUE(12267): Refine meter, progress, switch and range input styling.
 
 ```css
-input[type=checkbox][switch] {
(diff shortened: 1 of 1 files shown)


css-forms-1/Overview.bs
```diff
@@ -829,13 +849,24 @@ ISSUE: Refine meter, progress, switch and range input styling.
...
+}
+
 ::slider-thumb {
   border-radius: 0;
   border: none;
   background-color: currentColor;
   appearance: none;
   width: 1em;
-  height: 100%;
+  height: 1em;
```
lukewarlow, line 863: Need to consider the WCAG touch size requirements for this. I suspect we'll need something bigger.

PR #12449, opened 2025-07-04
[css-overflow-5] Add an example for 'scroll-target-group' property

danielsakhapov:
As requested in Intent to Ship thread:
https://groups.google.com/a/chromium.org/g/blink-dev/c/R_VD_FkYrF8/m/XDzix9lbAwAJ adding an example to show how to use 'scroll-target-group' property to highlight the currently visible chapter within a table of contents.

### DIFF
--- a/css-overflow-5/Overview.bs
+++ b/css-overflow-5/Overview.bs
@@ -250,6 +250,49 @@ The 'scroll target group' property</h4>
 
 	</dl>
 
+	<div class='example'>
+		The following snippet demonstrates how to automatically highlight the currently visible chapter within a table of contents
+		<pre><code highlight="html">
+			&lt;!DOCTYPE HTML&gt;
+			&lt;title&gt;scroll-target-group example with a table of contents&lt;/title&gt;
+			&lt;style&gt;
+			  #toc {
+					background-color: gray;
+					right: 10px;
+					top: 10px;
+					position: fixed;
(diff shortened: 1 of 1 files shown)


css-overflow-5/Overview.bs
```diff
@@ -250,6 +250,49 @@ The 'scroll target group' property</h4>
...
+					position: fixed;
+					/* Triggers the browser's algorithm for calculating :target-current. */
+					scroll-target-group: auto;
+				}
+
+				a {
+					color: blue;
+					display: block;
+					padding: 10px;
+					text-decoration: none;
```
flackr: not necessary but for the purposes of an example you could probably skip most of this styling, i.e.

1. `display: block` isn't necessary if the links are in a block element, commonly table of contents are a `<ul>` or `<ol>` with `<li>` wrapping each of the links (the structure of the spec table of contents itself).
2. Links are normally blue anyways right?
3. In a structured list, padding wouldn't be necessary on the link tag.
4. Hiding the link text decoration seems like an optional thing.

TLDR I think you should remove this block completely, change `#toc` to an `<ol>` element and wrap each of the `<a>` elements in a `<li>`.

Updated: Sorry my comment got a bit mangled but it looks like you got it :-)

flackr (approved):
Looks good but I left some suggestions to make it a bit more concise.

flackr (approved):

PR #12459, opened 2025-07-07
[css-ui][editorial] Markup <x> and <y> productions with <dfn> element

cdoublev:
I could not find the issues mentioned in the message of 9b8f394 in the archives, so I may be missing the reason not to do this change. But it would allow `w3c/reffy` to extract these CSS types.


tabatkins:
I would much rather just lightly rewrite the spec to use `<<number>>` in the grammars and define that they're the X and Y coordinates in prose. The custom production doesn't add anything here, and differs from how we do similar grammars in many other specs.

tabatkins:
All right, editorial fixup of UI 3 and 4 committed, so I'm closing this.

PR #12460, opened 2025-07-07
[css-borders-4][editorial] Remove range of k parameter from superellipse()

cdoublev:
Replaces `<number [-∞,∞]>` with `<number>`, since the range includes the infinity values.


SebastianZ:
Missed that before. Thanks, @cdoublev!

Sebastian

Issue #12380, opened 2025-06-21
[css-env][css-fonts] Changing the initial font size using <meta text-scale>


JoshTumath:
@davidsgrogan, @progers and I would like to propose a new HTML meta tag called `text-scale` with the following values:

* `<meta name="text-scale" content="legacy">`
* `<meta name="text-scale" content="scale">`

When the author specifies the `scale` option, the UA's initial font size is affected by the OS-level text scale setting.

**➡️ For all of the details of the proposal, read our [Explainer on a meta tag for text scaling behaviour.](https://github.com/w3c/csswg-drafts/blob/main/css-env-1/explainers/meta-text-scale.md)**

We'd love to hear your feedback!

## Background

In #10674, I wanted to solve the problem of how web content doesn't scale based on the OS text scale setting, which was an important accessibility issue for us at the BBC.

The CSSWG resolved to adopt:

* a new environment variable `env(preferred-text-scale)`, which will give authors the amount that the user has scaled their text size. 
* a new unit called `pem` which will be relative to `env(preferred-text-scale)`

Since then, we published an [Explainer for `env(preferred-text-scale)`](https://github.com/w3c/csswg-drafts/blob/main/css-env-1/explainers/env-preferred-text-scale.md).

However, after gathering feedback from authors at conferences recently, I found authors didn't like the sound of the `pem` unit. They would rather continue to use `rem` units as they do now. So we no longer believe the `pem` units are worthwhile, and believe the original idea for a `<meta>` tag to turn on the new/correct/proper text scaling behaviour will be much better (see the Explainer for more info).

## Proposed resolutions

* To adopt the proposal in the Explainer for a `<meta text-scale>` tag
* To retract the resolution for a `pem` unit

JoshTumath:
I categorised this as relating to the css-env-1 and css-fonts-5 specs for now, but actually, if we all have consensus to adopt this, it's not clear to me what spec it would live in. It may even be in WHATWG territory.

tabatkins:
To be a little clearer about the reasoning here (since the explainer just mentions the `meta` at the end as an alternative considered):

[...]

PR #12469, opened 2025-07-09
[css-fonts-5]: define text-scale, legacy/scale keywords, env(preferred-text-scale)

keithamus:
This is an attempt at resolving #12380 by defining the `<meta name="text-scale">` tag, how the UA responds to it (with the `legacy` and `scale` keywords) and defines the `env(preferred-text-scale)` value and its relation to this.

### DIFF
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -59,7 +59,83 @@ rules defined in CSS Fonts Level 4.
 This specification is currently a delta to the CSS Fonts Level 4 specification.
 Do not assume that if something is not here, it has been dropped.
 
+<h2 id="environment-variables">
+Environment Variables</h2>
 
+Text Scale {#text-scale-envs}
+------------------------------------------------------------------
+
+<table dfn-type=value dfn-for="env()">
+	<tr>
+		<th>Name
+		<th>Value
+		<th>Number of dimensions
+	<tr>
+		<td><dfn>preferred-text-scale</dfn>
+		<td><<number>>
+		<td>0 (scalar)
+</table>
+
+The text-scale [=environment variables=] define the devices text scale, which
+must be a non-zero number for which all text can be scaled by.
+
+The preferred-text-scale unit must be a value that computes to the documents
+initial font-size, when multiplied by 16px:
+
+<pre highlight=css>
+	:root {
+		font-size: calc(16px * env(preferred-text-scale));
+	}
+</pre>
+
+<h2 id="text-scale-meta">
+Text-Scale <code class=html>&lt;meta&gt;</code> element</h2>
+
+A document with a <code>&lt;meta></code> tag whose <code>name</code> attribute
+is a <a>ASCII case-insensitive</a> match for
+<dfn lt=text-scale><code>"text-scale"</code></dfn> is recognized as setting the
+initial font size of the document. The value of the <code>content</code>
+attribute must be an <a>ASCII case-insensitive</a> match for one of the
+recognized keywords.
+
+Documents without this <code>&lt;meta></code> tag will have an assumed default
+value of <code>legacy</code>.
+
+<h3 id="text-scale-meta-keywords">
+Keywords</h3>
+
+The recognized keywords in the [=text-scale=]
+<code class=html>&lt;meta&gt;</code> element are:
+
+<ul>
+	<li><code class="index" lt="legacy!!text-scale-meta">legacy</code></li>
+	<li><code class="index" lt="scale!!text-scale-meta">scale</code></li>
+</ul>
+
+<h3 id="legacy-keyword">The 'legacy' keyword</h3>
+
+The <dfn for="text-scale" export><code>legacy</code></dfn> property is
+recognized in the [=text-scale=] content attribute value.
+
+When the value of the [=text-scale=] content attribute is
+<a for="text-scale">legacy</a> the user agent should set the initial font size
+to 16px. The user agent may chose a different initial value, such as one based
+on a user preference. The ''env()/preferred-text-scale'' value must be 1.
+
+<h3 id="scale-keyword">The 'scale' keyword</h3>
+
+The <dfn for="text-scale" export><code>scale</code></dfn> property is
+recognized in the [=text-scale=] content attribute value.
+
+When the value of the [=text-scale=] content attribute is
+<a for=text-scale>scale</a> the user agent may determine the initial font size
+based on the operating systems text scale setting. The
+''env()/preferred-text-scale'' value must be a number that, when multiplied by
+16px, provides a <<length>> that matches that of the initial font size.
+
+Note: It is expected that authors will use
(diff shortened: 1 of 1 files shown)


css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,83 @@ rules defined in CSS Fonts Level 4.
...
+
+<h3 id="legacy-keyword">The 'legacy' keyword</h3>
+
+The <dfn for="text-scale" export><code>legacy</code></dfn> property is
+recognized in the [=text-scale=] content attribute value.
+
+When the value of the [=text-scale=] content attribute is
+<a for="text-scale">legacy</a> the user agent should set the initial font size
+to 16px. The user agent may chose a different initial value, such as one based
+on a user preference. The ''env()/preferred-text-scale'' value must be 1.
```
davidsgrogan, line 123: The value is always 1 on desktop, but not necessarily on mobile.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,83 @@ rules defined in CSS Fonts Level 4.
...
+The <dfn for="text-scale" export><code>scale</code></dfn> property is
+recognized in the [=text-scale=] content attribute value.
+
+When the value of the [=text-scale=] content attribute is
+<a for=text-scale>scale</a> the user agent may determine the initial font size
+based on the operating systems text scale setting. The
+''env()/preferred-text-scale'' value must be a number that, when multiplied by
+16px, provides a <<length>> that matches that of the initial font size.
 
+Note: It is expected that authors will use
```
davidsgrogan, line 136: @JoshTumath I'm interested to hear your thoughts on this paragraph :) As a non-author, I don't have a good sense of what authors are expected to do.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,83 @@ rules defined in CSS Fonts Level 4.
 This specification is currently a delta to the CSS Fonts Level 4 specification.
 Do not assume that if something is not here, it has been dropped.
 
+<h2 id="environment-variables">
+Environment Variables</h2>
+
+Text Scale {#text-scale-envs}
+------------------------------------------------------------------
+
+<table dfn-type=value dfn-for="env()">
```
davidsgrogan, line 68: Did you already crib everything appropriate from https://drafts.csswg.org/css-env-1/#text-zoom ?

davidsgrogan (commented):
Thanks so much for writing this!

Not sure if you already saw https://drafts.csswg.org/css-env-1/#text-zoom , which was the earlier iteration of the env() definition and `pem`.

css-fonts-5/Overview.bs
keithamus: Didn’t even spot this, I’ll make sure everything is moved over and update this tomorrow.

css-fonts-5/Overview.bs
JoshTumath: Yeah, as in the Explainer, the `env(preferred-text-scale)` is the OS text scale on mobile. So if the user doubles their text scale, the value is 2, for example.
https://github.com/w3c/csswg-drafts/blob/main/css-env-1/explainers/env-preferred-text-scale.md#proposal

JoshTumath, line 123: Does 'initial font size' need to be formally defined somewhere? My understanding is that the initial font size _is_ the [`font-size: medium` keyword.](https://drafts.csswg.org/css-fonts-4/#font-size-prop) Maybe the definition of `medium` needs updating to formally define it as the UA setting by default in `legacy` mode and the UA setting * the OS text scale in `scale` mode.

css-fonts-5/Overview.bs
JoshTumath: If the meta tag is set to `scale`, authors would have no need to use `calc(16px * env())`. They can just use `font-size: medium`, since `medium`/the initial font size is now effectively that.

keithamus: `font-size: medium` will work for font sizes but presumably this also impacts ems & rems, such that `1rem` will scale in proportion?

I think here I am simply trying to illustrate that authors might expect `16px` to be the browsers font size and `16px * env()` should (ideally?) resolve to the equivalent px units that `1rem * env()` would.

css-fonts-5/Overview.bs
keithamus: Does this mean in the `legacy` mode that `env(preferred-text-scale)` may not be congruent with the font-size? As in `font-size: medium` will resolve to `16px` even though the rest of the user's OS is at 24px because they have a text-scale 1.5? So an author cannot rely on `calc(16px * env(preferred-text-scale))` resolving to the same CSS pixels as `1rem`?

css-fonts-5/Overview.bs
JoshTumath: Oh I think I see what you mean. Yeah if the meta tag is set to the `scale` keyword, provided that the author doesn't change the root font-size, `1rem` will always be equivalent to `16px * env()`. I think I was confused by the phrasing of 'authors are expected to use it' since I wouldn't expect anyone to practically need to do that.

css-fonts-5/Overview.bs
davidsgrogan: That's correct, and is what happens in today's pre-`<meta name="text-scale">` world.

We expect/hope that `<meta name="text-scale" content="scale">` becomes best practice so that authors don't have to deal with today's inconsistencies, one of which is what you pointed out.

svgeesus:
Waiting for the various discussions here to be marked as "resolved" before merging

keithamus:
Yeah I need to tidy this up, I'll hopefully get some time this week.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,83 @@ rules defined in CSS Fonts Level 4.
...
+
+<h2 id="text-scale-meta">
+Text-Scale <code class=html>&lt;meta&gt;</code> element</h2>
+
+A document with a <code>&lt;meta></code> tag whose <code>name</code> attribute
+is a <a>ASCII case-insensitive</a> match for
+<dfn lt=text-scale><code>"text-scale"</code></dfn> is recognized as setting the
+initial font size of the document. The value of the <code>content</code>
+attribute must be an <a>ASCII case-insensitive</a> match for one of the
+recognized keywords.
```
zcorpan, line 99: Shouldn't this be defined in HTML?

davidsgrogan: `<meta name="viewport" content="...">` seems to be [defined in css](https://www.w3.org/TR/css-viewport-1/), not [html](https://html.spec.whatwg.org/multipage/semantics.html#the-meta-element). Couldn't we follow suit?

davidsgrogan: ... Oh, do you mean just this section should be in HTML?

keithamus:
Forgot to get around to this but https://github.com/w3c/csswg-drafts/pull/13052 supersedes it, so closing.

Issue #12430, opened 2025-07-01
[css-forms-1] Checkbox and radio sizing should meet WCAG minimum sizing


lukewarlow:
Per guideline 3 in https://drafts.csswg.org/css-forms-1/#basic-appearance-principles "The controls pass 100% of [WCAG 2.2 AA standards](https://www.w3.org/TR/WCAG22/)."

We should ensure controls meet https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html

For checkbox, and radio inputs we currently have:

```css
input:is([type=checkbox]:not([switch]), [type=radio]) {
    width: 1em;
    height: 1em;
    ...
}
```

I propose we change this to:

```css
input:is([type=checkbox]:not([switch]), [type=radio]) {
    /* These min-size rules ensure accessibility by following WCAG rules:
    * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
    min-block-size: max(1em, 24px);
    min-inline-size: max(1em, 24px);
    aspect-ratio: 1/1;

    /* This font-size rule ensures the checkmark fills the containing box. */
    font-size: max(1em, 24px);
    ...
}
```

cc @nt1m

nt1m:
I don't think we should force a min size because it's a pain to override.

I'd much rather use a width/height here (which can involve the 24px minimum) given it's the more straightforward one to override.

I generally only want to reserve min sizes when strictly necessary, and checkboxes/radios seem like cases where it's not strictly necessary.

lukewarlow:
So instead it'd be the below?

```css
input:is([type=checkbox]:not([switch]), [type=radio]) {
    /* These size rules ensure accessibility by following WCAG rules:
    * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
    block-size: max(1em, 24px);
    inline-size: max(1em, 24px);

    /* This font-size rule ensures the checkmark fills the containing box. */
    font-size: max(1em, 24px);
    ...
}
```

fantasai:
Current draft spec (per above PR) is:

```
   /* These size rules ensure accessibility by following WCAG rules:
     * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
    block-size: 1em;
    inline-size: 1em;
    font-size: max(1em, 24px);
```

Agenda+ to confirm with the WG and close.

SebastianZ:
[...]

PR #12489, opened 2025-07-16
[css-forms-1] Add 24px sizing rule for checkbox and radio inputs

lukewarlow:
See #12430

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -843,8 +843,14 @@ ISSUE: Refine meter, progress, switch and range input styling.
 
 ```css
 input:is([type=checkbox]:not([switch]), [type=radio]) {
-    width: 1em;
-    height: 1em;
+    /* These size rules ensure accessibility by following WCAG rules:
+    * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
+    block-size: max(1em, 24px);
+    inline-size: max(1em, 24px);
+
+    /* This font-size rule ensures the checkmark fills the containing box. */
+    font-size: max(1em, 24px);
+
     display: inline-flex;
     align-items: center;
     justify-content: center;


css-forms-1/Overview.bs
```diff
@@ -843,8 +843,14 @@ ISSUE: Refine meter, progress, switch and range input styling.
...
-    width: 1em;
-    height: 1em;
+    /* These size rules ensure accessibility by following WCAG rules:
+    * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
+    block-size: max(1em, 24px);
+    inline-size: max(1em, 24px);
+
+    /* This font-size rule ensures the checkmark fills the containing box. */
+    font-size: max(1em, 24px);
+
```
nt1m, line 851: font-size would control both block-size and inline-size technically, so this would have the same effect.

```
    block-size: 1em;
    inline-size: 1em;
    font-size: max(1em, 24px);
```

cc @fantasai to look at this

css-forms-1/Overview.bs
```diff
@@ -843,8 +843,14 @@ ISSUE: Refine meter, progress, switch and range input styling.
 
 ```css
 input:is([type=checkbox]:not([switch]), [type=radio]) {
-    width: 1em;
-    height: 1em;
+    /* These size rules ensure accessibility by following WCAG rules:
+    * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
+    block-size: max(1em, 24px);
```
lukewarlow, line 848: I think this should probably get an aspect-ratio: 1/1; by default too so you don't end up with accidentally rectangular checkboxes.

In my demo implementation the checkmark can end up stretching the checkbox weirdly without it.

lukewarlow:
@nt1m @fantasai  does this look okay?

css-forms-1/Overview.bs
fantasai: Clever. :)

css-forms-1/Overview.bs
```diff
@@ -843,8 +843,14 @@ ISSUE: Refine meter, progress, switch and range input styling.
 
 ```css
 input:is([type=checkbox]:not([switch]), [type=radio]) {
-    width: 1em;
-    height: 1em;
+    /* These size rules ensure accessibility by following WCAG rules:
+    * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
```
fantasai:
```suggestion
     * https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html */
```

PR #12520, opened 2025-07-24
[css-color-adjust-1][editorial] Fix typo in emualte-forced-colors-mode anchor

juliandescottes:
Typo noticed while reviewing https://github.com/w3c/webdriver-bidi/pull/905

### DIFF
--- a/css-color-adjust-1/Overview.bs
+++ b/css-color-adjust-1/Overview.bs
@@ -1114,7 +1114,7 @@ Emulation {#color-adjust-emulation}
 	For the purposes of user agent automation and application testing, this document
 	defines the below emulations.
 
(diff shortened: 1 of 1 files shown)


juliandescottes:
cc @alisonmaher not sure if this link is already cross referenced in other specs, but I thought it might only be relevant to the BiDi spec, so better to fix it now?

SebastianZ:
At least, I didn't find any CSS spec. referencing it. So, merged it.

alisonmaher:
> cc @alisonmaher not sure if this link is already cross referenced in other specs, but I thought it might only be relevant to the BiDi spec, so better to fix it now?

Thanks for the catch! This was added specifically for webdriver support, so no other references to update.

Issue #11977, opened 2025-03-21
[css-multicol-2] `column-height` and nested fragmentation


mstensho:
How should `column-height` be handled if the multicol container is nested inside another fragmentation context (e.g. pagination, or another multicol container)? If an outer break is inserted in the middle of an inner row, should a new inner row be created from scratch in the next outer fragmentainer (option 1 below), or should it resume and use whatever is left (option 2 below)?

```html
<!DOCTYPE html>
<style>
  #outer {
    columns: 2;
    column-fill: auto;
    height: 160px;
    line-height: 20px;
    orphans: 1;
    widows: 1;
    column-rule: solid;
  }
  #inner {
    columns: 1;
    column-fill: auto;
    column-height: 80px;
    column-wrap: wrap;
    row-gap: 40px;
  }
</style>
<div id="outer">
  <div id="inner">
    <div style="background:hotpink;">
      a<br>b<br>c<br>d<br>e<br>f<br>g<br>h<br>i<br>j<br>k<br>
    </div>
  </div>
</div>
```

Option 1 (restart from scratch):

![Image](https://github.com/user-attachments/assets/46d746e7-843d-4888-bddf-b66f757c3f93)

Option 2 (resume what's left):

![Image](https://github.com/user-attachments/assets/34112442-cf06-4403-800f-585f8807c906)

@rachelandrew

rachelandrew:
I think we are intending this wrapping to only take effect in continuous contexts (it's essentially creating pagination for web contexts).

In the nested multicol example, I think probably option 1, but I don't have a strong opinion and I don't think it's something that's likely to happen too much.

mstensho:
Option 1 should be easier to implement (no need to keep track of row progress from earlier fragments), and happens to be what we're currently doing in Blink, because I made no conscious choice when implementing it, and got this for free. But I don't think this behavior makes a lot of sense. Looks like a bug, really. Can we have option 2 instead? I also think it matters if there's a column spanner there in the inner multicol that crosses the outer column boundary.

So that:

```html
<!DOCTYPE html>
<style>
  #outer {
    columns: 2;
    column-fill: auto;
    height: 160px;
    line-height: 20px;
    orphans: 1;
    widows: 1;
    column-rule: solid;
  }
  #inner {
    columns: 2;
    column-fill: auto;
    column-height: 80px;
    column-wrap: wrap;
    row-gap: 40px;
  }
</style>
<div id="outer">
  <div id="inner">
[...]

Issue #11474, opened 2025-01-10
[css-multicol] column-span: <integer>


scottkellum:
In [CSS Multicolumn, column-span](https://www.w3.org/TR/css-multicol-2/#column-span) it would be useful to have the ability to create content that spans a specified number of columns using an <integer>. It is common to have images span two or three text columns in print.

I would expect this to work where the vertical and starting position coincides with where the `block` element falls in the column, then clears subsequent adjacent columns at that position of content to expand into like a float. This would likely be commonly used with `break-before: column; height: 100%;` styles to create a full height block.

Example of a print layout with images spanning two columns:

[![print layout example showing images spanning two text columns](https://i0.wp.com/tirsogamboa.com/wp-content/uploads/2020/12/TIRSO_magazinedesign_2A.jpg?fit=1200%2C858&ssl=1)](https://tirsogamboa.com/how-to-design-a-magazine-layout-like-a-true-editorial-designer/)

It is possible to hack this functionality into some browsers (Chrome) by using `<figure>` and `::before` and `::after` pseudo elements alongside `break-before` and `break-after`, thus clearing columns for the figure to expand into. https://codepen.io/scottkellum/pen/qBQYyGO?editors=1100

claell:
I have a paged-media use case that may help clarify the intended behavior. It is similar to the print example in the issue, but involves a leading partial spanner and automatic flow through columns with different available heights.

Consider a three-column newspaper, magazine, or folded-brochure page:

* A masthead or section introduction spans the first two logical columns at the block start.
* The third column remains available from the block start instead of being pushed below the masthead.
* All following articles remain in one semantic content stream. If their lengths change, content automatically reflows through the shortened first and second columns and then through the full-height third column.

The important part is the single automatic flow. CSS Grid can place the masthead across two tracks, but it generally requires content to be assigned to particular grid areas rather than flowing automatically through these irregular column areas. `column-span: all` also cannot express the layout because it shortens every column.

A more general version is a photograph, pull quote, or advertisement encountered later in a four-column article that spans the current and next column while surrounding text continues in document order. The existing example already points towards this float-like case.

These examples expose some of the questions already noted in the draft:

* Should an integer spanner start in the column where it is encountered and cover the next logical columns, or are leading partial spanners handled differently?
* Should adjacent covered columns be cleared only over the spanner's block range, producing irregular flow areas?
[...]

PR #12544, opened 2025-07-30
[css-multicol-2] Initial pass at changes to the model section.

rachelandrew:
There will be more to do here, and I want to add some images, but here's a first pass at the basics so we at least have lines and rows defined.

I haven't added the stuff about nested multicols here, not sure if it's the right place.

### DIFF
--- a/css-multicol-2/Overview.bs
+++ b/css-multicol-2/Overview.bs
@@ -360,11 +360,14 @@ The multi-column model</h2>
 
 	The column boxes are ordered
 	in the [=inline base direction=] of the multicol container
-	and arranged into <dfn lt="multi-column line | multi-col line | multicol line">multicol lines</dfn>.
+	and arranged into 
+	<dfn lt="multi-column row | multi-col row | multicol row">multicol rows</dfn>.
+	A multicol row contains at least one <dfn lt="multi-column line | multi-col line | multicol line">multicol line</dfn>,
+	but may contain additional multicol lines due to spanning elements.
 	The <dfn export>column width</dfn> is the length of the column box in the inline direction.
 	The <dfn export>column height</dfn> is the length of the column box in the block direction.
-	All column boxes in a line have the same column width,
-	and all column boxes in a line have the same column height.
+	All column boxes in a row have the same column width,
+	and all column boxes in a row have the same column height.
 
 	<wpt title="The following tests relate to baseline alignment of the content of columns, 
 	though this is not defined in this specification.">
@@ -529,15 +532,15 @@ The multi-column model</h2>
 	hit-test-in-vertical-rl.html
 	</wpt>
 
-	Within each [=multicol line=] in the multi-column container,
+	Within each [=multicol row=] in the multi-column container,
 	adjacent column boxes are separated by a <dfn noexport>column gap</dfn>,
 	which may contain a <dfn noexport>column rule</dfn>.
 	All column gaps in the same multi-column container are equal.
 	All column rules in the same multi-column container are also equal, if they appear;
 	column rules only appear between columns that both have content.
 
-	In the simplest case a multicol container will contain only one line
-	of columns, and the height of each column will be equivalent to the
+	In the simplest case a multicol container will contain only one line of columns,
+	inside one multicol row, and the height of each column will be equivalent to the
 	used height of the multi-column container's content box.
 	However, [=fragmentation=] or [=spanners=]
 	can split the content of the [=multi-column container=]
@@ -547,10 +550,17 @@ The multi-column model</h2>
 	constrained by the page and the content continues in a new line of
 	column boxes on the next page; a column box never splits across pages.
 
-	The same effect occurs when a <i>spanning element</i> divides the
-	multi-column container: the columns before the spanning element are
+	If a multi-column container in continuous media is set to wrap,
+	the height of each column is defined by the 'column-height' property.
+	If there is more content than will fit in a single row a new multicol row is created
+	in the block direction, with columns of the size defined by the 'column-height' property.
+	This row contains at least one multicol line.
+
+	When a <i>spanning element</i> divides the
+	multi-column container: a new multicol line is created.
+	The columns before the spanning element are
 	balanced and shortened to fit their content. Content after the
-	spanning element then flows into a new, subsequent line of column boxes.
+	spanning element then flows into a new, subsequent multicol line of column boxes.
 
 	<div class="example">
 		<figure>
@@ -559,10 +569,13 @@ The multi-column model</h2>
 		</figure>
 	</div>
 
+	Issue(11976): This issue deals with the scenario where there is not enough space
+	in the current multicol row for a spanning element. Once resolved, add details here.  
+
 	A [=multi-column container=] therefore is a regular [=block container=]
 	that establishes a new [=independent formatting context=]
-	whose contents consist of a series of
-	[=multicol lines=] and multicol spanners.
+	whose contents consist of one or more
+	[=multicol rows=], containing [=multicol lines=] and multicol spanners.
 	Each [=multi-column line=] acts as a [=block-level box=]
 	that establishes a <dfn>multi-column formatting context</dfn>
 	for its [=column boxes=];
--- a/css-multicol-2/Overview.bs
+++ b/css-multicol-2/Overview.bs
@@ -534,8 +534,8 @@ The multi-column model</h2>
 	In the inline direction, within each [=multicol line=] in the multi-column container,
 	adjacent column boxes are separated by a <dfn noexport>column gap</dfn>,
 	which may contain a <dfn noexport>column rule</dfn>.
-	In the block direction, [=column rows=] are separated by a <dfn noexport>column gap</dfn>,
-	which may contain a <dfn noexport>column rule</dfn>.
+	In the block direction, [=column rows=] are separated by a <dfn noexport>row gap</dfn>,
+	which may contain a <dfn noexport>row rule</dfn>.
 	All column gaps in the same multi-column container are equal.
 	All column rules in the same multi-column container are also equal, if they appear;
 	column rules only appear between columns that both have content.
@@ -555,7 +555,6 @@ The multi-column model</h2>
 	the height of each column is defined by the 'column-height' property.
 	If there is more content than will fit in a single row a new multicol row is created
 	in the block direction, with columns of the size defined by the 'column-height' property.
-	This row contains at least one multicol line or one spanning element.
 
 	When a <i>spanning element</i> divides the
 	multi-column container: a new multicol line is created.
--- a/css-multicol-2/Overview.bs
+++ b/css-multicol-2/Overview.bs
@@ -360,14 +360,13 @@ The multi-column model</h2>
 
 	The column boxes are ordered
 	in the [=inline base direction=] of the multicol container
-	and arranged into 
-	<dfn lt="multi-column row | multi-col row | multicol row">multicol rows</dfn>.
-	A multicol row contains at least one <dfn lt="multi-column line | multi-col line | multicol line">multicol line</dfn>,
-	but may contain additional multicol lines due to spanning elements.
+	and arranged into <dfn lt="multi-column row | multi-col row | multicol row">multicol rows</dfn>.
+	A multicol row contains at least one <dfn lt="multi-column line | multi-col line | multicol line">multicol line</dfn>
+	or spanning element.
 	The <dfn export>column width</dfn> is the length of the column box in the inline direction.
 	The <dfn export>column height</dfn> is the length of the column box in the block direction.
-	All column boxes in a row have the same column width,
-	and all column boxes in a row have the same column height.
+	All column boxes in a line have the same column width,
+	and all column boxes in a line have the same column height.
 
 	<wpt title="The following tests relate to baseline alignment of the content of columns, 
 	though this is not defined in this specification.">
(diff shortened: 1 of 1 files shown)


rachelandrew:
@mstensho I couldn't add you as a reviewer for some reason but PTAL at this initial pass on the model section.

css-multicol-2/Overview.bs
```diff
@@ -360,11 +360,14 @@ The multi-column model</h2>
 
 	The column boxes are ordered
 	in the [=inline base direction=] of the multicol container
-	and arranged into <dfn lt="multi-column line | multi-col line | multicol line">multicol lines</dfn>.
+	and arranged into 
+	<dfn lt="multi-column row | multi-col row | multicol row">multicol rows</dfn>.
+	A multicol row contains at least one <dfn lt="multi-column line | multi-col line | multicol line">multicol line</dfn>,
```
mstensho: A row may also very well just contain a spanner (and no lines), right?

css-multicol-2/Overview.bs
```diff
@@ -360,11 +360,14 @@ The multi-column model</h2>
...
+	but may contain additional multicol lines due to spanning elements.
 	The <dfn export>column width</dfn> is the length of the column box in the inline direction.
 	The <dfn export>column height</dfn> is the length of the column box in the block direction.
-	All column boxes in a line have the same column width,
-	and all column boxes in a line have the same column height.
+	All column boxes in a row have the same column width,
```
mstensho: Column boxes are inside a line (which is inside a row). Why this change? It's also wrong in some cases, actually, since fragments generated from the same multicol container (when nested inside another fragmentation context) may have different widths or heights. Since each multicol fragment establishes a separate *line* for its columns (right?), if it has any columns at all, the width and height of column boxes are the same within a line (but not necessarily within a row).

css-multicol-2/Overview.bs
```diff
@@ -529,15 +532,15 @@ The multi-column model</h2>
 	hit-test-in-vertical-rl.html
 	</wpt>
 
-	Within each [=multicol line=] in the multi-column container,
+	Within each [=multicol row=] in the multi-column container,
```
mstensho: This should remain "line" as well, I think.

css-multicol-2/Overview.bs
```diff
@@ -360,11 +360,14 @@ The multi-column model</h2>
 
 	The column boxes are ordered
 	in the [=inline base direction=] of the multicol container
-	and arranged into <dfn lt="multi-column line | multi-col line | multicol line">multicol lines</dfn>.
+	and arranged into
```
mstensho: The column boxes are arranged into lines that are inside a row, no?

css-multicol-2/Overview.bs
```diff
@@ -547,10 +550,17 @@ The multi-column model</h2>
 	constrained by the page and the content continues in a new line of
 	column boxes on the next page; a column box never splits across pages.
 
-	The same effect occurs when a <i>spanning element</i> divides the
-	multi-column container: the columns before the spanning element are
+	If a multi-column container in continuous media is set to wrap,
```
mstensho, line 554: Why just continuous media?

rachelandrew: I don't see what the use case is for doing this in paged media, paged media already overflows into new pages. If there's a use case, then great, but this has always been discussed as for continuous media.

css-multicol-2/Overview.bs
rachelandrew: I'm not so sure conceptually, because the height of the column (`column-height`) is what defines the size of the row. It could, however, contain two lines and a spanner.

css-multicol-2/Overview.bs
rachelandrew: Again, I think this depends on where we see the height as happening.

- The `column-height` is 100px.
- There's a spanner, it is 20px tall.
- We now have two multicol lines, one above and one below the spanner.
- If the column-height is what defines the row height, those lines are each 40px tall.

Therefore the height the author cares about is the height of the column, because that defines the height of the row. If you consider a block direction carousel where each row is 100vh, the author needs to set the `column-height` to 100vh, which defines the row size. The only reason the row would be a different size would be if we had to stretch due to a giant spanner (if we go that way).

css-multicol-2/Overview.bs
mstensho: Yes, a row contains lines and/or spanners. Lines contain columns. So the column boxes are arranged into lines whose height are confined by the containing row, due to `column-height`.

So I was thinking that this part looked right the way it was.

> The column boxes are ordered in the inline base direction of the multicol container and arranged into multicol lines.

*Then* explain that a line is contained by a row?

css-multicol-2/Overview.bs
mstensho: So `column-wrap` only applies in continuous media? It should probably also *not* apply in nested multicol then (even if media may be continuous in such cases).

Not sure about wording.

> If a multi-column container that doesn't participate in a fragmentation context (such as paged media) is set to wrap [...]

?

If we add this limitation, and also do it for `column-height`, it also means that https://github.com/w3c/csswg-drafts/issues/11977 doesn't need to be resolved.

css-multicol-2/Overview.bs
mstensho: Assuming that we have the same amount of content before and after the spanner, then yes, the columns will all have a height of 40px, both before and after the spanner.

```html
<!DOCTYPE html>
<div style="columns:3; column-fill:auto; column-height:100px; line-height:20px;">
  before
  <div style="column-span:all;">spanner</div>
  after<br>after<br>after<br>after<br>after<br>
</div>
```

There's a multicol container. It has one row that is 100px tall. The row has one line, one spanner, then another line.
The first line is 20px tall, the spanner is 20px tall, the second line is 60px tall. The columns in the first line is 20px tall. The columns in the second line are 60px tall. Therefore all columns in a row do not necessarily have the same column height.

rachelandrew: I see what you mean... maybe we need to more strongly define column boxes vs columns? As it's the column that has the height, depending on how many lines there are multiple column boxes within each column. I think that's where I'm confusing things anyway.

mstensho: I'm confused now. :) Multiple column boxes within each column? In my mental model, there's no distinction between "column" and a "column box". In my example above, there should be one column (or column box) above the spanner (there's just one word there, after all), and two columns (or two column boxes) below the spanner. Are you thinking that the first *column box* above the spanner and the first *column box* below the spanner form *one* *column*? Or something else?

rachelandrew: I'm talking about the thing people are defining a height on.

An author defines a `column-height`, this actually defines the height of the multicol row the columns are contained in.

In the simple scenario we then end up with one or more multicol rows, each containing a multicol line. Both of these will be the height was the author defined by setting a `column-height`.

However, if row 1 contains a spanner, the height of the row should remain as set by the author, but the row is split into two multicol lines with the spanner between them. This does result in two column boxes for each column (in the author view, "I've just set the height of my columns".) as the total height of the row shouldn't change in an ideal world.

I think in terms of implementation, the fact the column-height defines the row probably makes no difference, you end up with multiple lines of boxes that behave as they do now (the complexity is in the other issue where we figure out what to do with the too big spanner). For an author, we're using `column-height` to actually define a row height.

Anyway, I think you are correct with lines vs rows, I was just musing on where that confusion originated and whether I needed to do something to prevent that same confusion.

mstensho: Ah, got it! I just read `column-height` out loud now, and got it. I must have been pronouncing it as "row height" all along. And that's what it is, isn't it? Column flow may be interrupted by a spanner, resulting in two column lines (one before, one after), and the total height available for columns will be `column-height` minus the space taken up by the spanner. Grouping column boxes with the same inline offset (before and after a spanner) together doesn't make sense to me.

Are we sure about the naming, then? Maybe it should be `row-height`, or `column-row-height`?

rachelandrew: yeah... maybe the name is just wrong. It's fine until a spanner gets in there, because the column, row, and line are the same height. Once we have the spanner, the row is now the height that's been set but the other two are different heights.

Let me fix this up so the line and row mentions are correct, we could always take the naming issue back to the WG, but it might be easier with this new model in place as it highlights the issue.

mstensho: Ok. Again I'm tempted to conflate "lines" and "rows", and let column spanners live on the outside of that, then define that row gaps and decorations are only inserted if there's column content on both sides of the gap (i.e. only when columns wrap, and not e.g. next to spanners or at the bottom / top when crossing an outer fragmentainer boundary). What would we lose with such a simplified model? `column-height` would be for columns only, then. Any spanner would be pushed after the row/line, take up its space, and a new row/line would resume after it. What have I forgotten about?

rachelandrew: That won't work for the carousel type use cases. You'd end up with a spanner outside of the "page" which would be weird. 

It would be fine for the simplest use case of just making sure people don't need to scroll up and down to read, but if we want an entire screen to be a row AND you had a spanner, you'd then end up with the spanner off the bottom of the screen.

Also it would definitely stop `column-span: <integer>` ever being a thing.

We need the lines.

rachelandrew: To explain about `column-span: <integer>` the most likely way (in my mind) to implement that would be in conjunction with page floats. So you'd have a three column multicol container, and an element spanning two columns floated top right. That's a super common magazine type layout and the sort of thing people raise (for example [#11474](https://github.com/w3c/csswg-drafts/issues/11474)).

That would not work if spanners were this weird thing that existed outside of rows.

mstensho: Do carousel type use cases even want spanners at all, though?

Spanners are taken "out of flow" anyway, and I see no strict reason for them having to belong inside a row. Yes, I think page float is the use case for `column-span: <integer>`. Anything else (non-floated) would be weird, I think.

rachelandrew: I think it would be such an unexpected behavior for spanners to do that, to immediately lose the nice "set of pages at x height" because there's a spanner seems bad. I do thing that a carousel type layout using page floats would likely be something people want if we ever go there, because there's some nice (and fairly obvious) patterns you could achieve with that.

mstensho: I think that makes sense. So I guess we need lines and rows. Sorry for the detour.

And deal with page floats later. :) You seem to want them to be contained by rows? They don't need to live inside multicol, though (in which case there'll be no rows). E.g. when printing. We may want to top/bottom align a page float. When printing with multicol, we might want them to span a certain number of columns. If there's no column wrapping set, there'll still be just one giant row, but we then want the page floats to be aligned with some page edge, since the row is irrelevant in that case.

css-multicol-2/Overview.bs
rachelandrew: I think adding a limitation for nesting is probably ok, I think the only nested multicols I come across are in test cases.

css-multicol-2/Overview.bs
rachelandrew: Changed this based on another suggestion.

css-multicol-2/Overview.bs
rachelandrew: I've tried to do this this way round PTAL.

css-multicol-2/Overview.bs
rachelandrew: changed.

rachelandrew:
@mstensho thanks for all the discussion, I've updated this some more.

css-multicol-2/Overview.bs
```diff
@@ -529,28 +531,37 @@ The multi-column model</h2>
 	hit-test-in-vertical-rl.html
 	</wpt>
 
-	Within each [=multicol line=] in the multi-column container,
+	In the inline direction, within each [=multicol line=] in the multi-column container,
 	adjacent column boxes are separated by a <dfn noexport>column gap</dfn>,
 	which may contain a <dfn noexport>column rule</dfn>.
+	In the block direction, [=column rows=] are separated by a <dfn noexport>column gap</dfn>,
```
mstensho: *row gap*

css-multicol-2/Overview.bs
```diff
@@ -529,28 +531,37 @@ The multi-column model</h2>
...
+	which may contain a <dfn noexport>column rule</dfn>.
```
mstensho: *row rule*

css-multicol-2/Overview.bs
mstensho: Yeah, if we want to avoid this in paged media, we should avoid it for any kind of nested fragmentation - e.g. nested multicol as well.

css-multicol-2/Overview.bs
```diff
@@ -529,28 +531,37 @@ The multi-column model</h2>
...
 	constrained by the page and the content continues in a new line of
 	column boxes on the next page; a column box never splits across pages.
 
-	The same effect occurs when a <i>spanning element</i> divides the
-	multi-column container: the columns before the spanning element are
+	If a multi-column container in continuous media is set to wrap,
+	the height of each column is defined by the 'column-height' property.
+	If there is more content than will fit in a single row a new multicol row is created
+	in the block direction, with columns of the size defined by the 'column-height' property.
+	This row contains at least one multicol line or one spanning element.
```
mstensho: This is true for all rows, and is also mentioned further up.

mstensho (changes requested):
Also include wording about disallowing wrapping for fragmented multicols (paged media / nested multicol)?

css-multicol-2/Overview.bs
rachelandrew: Raised an issue for this https://github.com/w3c/csswg-drafts/issues/12593

Issue #12431, opened 2025-07-01
[css-gaps-1] Define interpolation behavior


kbabbitt:
Because the width, style, and color properties can now take lists of values, including lists with repeaters, we need to define what the interpolation behavior for these more complex values might look like.

flackr:
FYI the repeat syntax is specified in https://www.w3.org/TR/css-gaps-1/#lists-repeat

I think that non-auto repeaters should be able to combine with non-repeater values and other repeaters intelligently, e.g.

Interpolating `repeat(3, red)` to `repeat(2, red) blue` should interpolate at 50% to `repeat(2, red) #770077`.

In general this probably means intersecting the range of the repeater and pulling out any values which are no longer part of it.

Similarly, we should be able to have intelligent repeater list extension when you don't have auto values. When there are auto values we should create additional values before/after the auto to match the other value's preceding / following non-auto lengths, e.g.

Interpolating `green repeat(auto, red) green` to `repeat(auto, blue) green` should interpolate at 50% to `#007777 repeat(auto, #770077) green`.

kbabbitt:
In cases where there are no auto-repeaters, I think we can indeed be smarter.

I think it gets tricky when there are auto-repeaters that don't line up with each other, because then the gap-by-gap interpolations can depend on the total number of gaps, which we don't know at computed-value time. For example, if we have:

```css
from { column-rule-width: 1px repeat(auto, 2px) 3px 4px 5px; }
to   { column-rule-width: 6px 7px repeat(auto, 8px 9px) 10px; }
```

The pairs of values will line up as follows:

```
4 gaps:
1px 3px 4px  5px
6px 7px 8px 10px

5 gaps:
1px 2px 3px 4px  5px
6px 7px 8px 9px 10px

6 gaps:
1px 2px 2px 3px 4px  5px
6px 7px 8px 9px 8px 10px

7 gaps:
1px 2px 2px 2px 3px 4px  5px
6px 7px 8px 9px 8px 9px 10px
```

I can't figure out how to generalize that expansion without it being a function of the number of gaps. The values that pair with 3px and 4px would need to be pulled out of the end of the region covered by the auto repeater. But without knowing the number of gaps, how do you know where in the auto repeater to start pulling values from?

flackr:
Oh yeah, my idea doesn't work if there are multiple values in the auto repeater, since as you point out we can't know which values to pull out at the right side of the auto repeater. However, for single value auto repeaters I think it works.

[...]

PR #12583, opened 2025-08-07
[css-gaps-1] Define Interpolation behavior for GapDecorations. Issue #12431

jav099:
This PR aims to define the interpolation behavior for the `rule-width` and `rule-color` properties in CSSGapDecorations, based on the resolution in https://github.com/w3c/csswg-drafts/issues/12431.

Changes can be seen live in https://jav099.github.io/javier-intersections-draft.github.io/

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -975,13 +975,6 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
 		</ol>
 	</div>
 
-	<div algorithm>
-		To <dfn>assign gap decoration values in reverse</dfn>
-		to a list of <var ignore=''>gaps</var> using a list of <var ignore=''>values</var>,
-		follow the same steps as to <a>assign gap decoration values</a>,
-		except that in step 2, change all instances of "first" to "last".
-	</div>
-
 <h4 id="interpolation-behavior">Interpolation behavior</h4>
 When interpolating ''repeat()'' values, or lists of values for 'rule-color' or 'rule-width', the interpolation proceeds in two steps:
 
@@ -992,8 +985,8 @@ When interpolating ''repeat()'' values, or lists of values for 'rule-color' or '
 	<div class="example">
 		<pre class="lang-css">
 		@keyframes example {
-			from { column-rule-width: 10px }
-			to   { column-rule-width: 20px 40px }
+			from { column-rule-width: 10px; }
+			to   { column-rule-width: 20px 40px; }
 		}
 		</pre>
 		Interpolation of the above values would result in expansion of the
@@ -1014,7 +1007,7 @@ When interpolating ''repeat()'' values, or lists of values for 'rule-color' or '
 		Interpolation of the above values would result in expansion of the
 		"from" and "to" values to create lists of equal lengths:
 		<pre>
-		From:   5px 10px 5px 10px
+		From:    5px 10px  5px 10px
 		At 50%: 10px 15px 10px 15px
 		To:     15px 20px 15px 20px
 		</pre>
@@ -1055,87 +1048,87 @@ When interpolating ''repeat()'' values, or lists of values for 'rule-color' or '
 <h5 id="auto-repeaters">Lists with Auto Repeaters</h5>
 	<div algorithm>
 
-		To <dfn>interpolate with auto repeaters</dfn>, when either of the lists we are interpolating between (let's call them |from| and |to|) include an <a>auto repeater</a>, we conceptually split the list into three segments, similar
-		to how we [=assign gap decoration values=]:
-		- |leading gaps|
-		- |auto repeat|
-		- |trailing gaps|
+		To <dfn>interpolate with auto repeaters</dfn>, when either of the lists we are interpolating between (|from| and |to|) include an <a>auto repeater</a>:
 
 			<ol>
-				<li>Let |segment lengths| be the length of the |leading gaps| and |trailing gaps| segments (i.e. the number of items in each list, after <a>expansion</a>).</li>
-				<li>We split |to| and/or |from| into their respective segments.</li>
-				<li>If either of |leading gaps| or |trailing gaps| in |to| or |from| is empty and the other has an <a>auto repeater</a>, we do [=discrete=] interpolation.</li>
+				<li>
+				Split each of |from| and |to| into segments, similar to how we [=assign gap decoration values=]:
+				Let |leading values| be the values before the <a>auto repeater</a>.
+				Let |trailing values| be the values after the <a>auto repeater</a>.
+				Let |auto values| be the values inside the <a>auto repeater</a>.
+				</li>
+				<li>If either of |leading values| or |trailing values| in |from| or |to| is empty and the other has an <a>auto repeater</a>, we fall back to [=discrete=] interpolation.</li>
 				<li>Expand any integer repeaters on each segment.</li>
-				<li>If the |segment lengths| don't match, we fall back to [=discrete=] interpolation.</li>
-				<li>When both |from| and |to| contain auto-repeaters and |segment lengths| match, we apply <a>list interpolation</a> to each segment. Applying <a>list interpolation</a> to the |auto repeat| means applying it to the list of values inside the repeater.</li>
-				<li>If only one of |from| or |to| contains an auto-repeater, we do [=discrete=] interpolation.</li>
+				<li>
+				If the length of |leading values| in |from| and |leading values| in |to| don't match, we fall back to [=discrete=] interpolation.
+				If the length of |trailing values| in |from| and |trailing values| in |to| don't match, we fall back to [=discrete=] interpolation.
+				</li>
+				<li>If only one of |from| or |to| contains an auto-repeater, we fall back to [=discrete=] interpolation.</li>
+				<li>When both |from| and |to| contain auto-repeaters, and the length of their segments match as described above, we apply <a>list interpolation</a> to each segment.
+				Applying <a>list interpolation</a> to the |auto values| means applying it to the list of values inside the repeater.
+				</li>
 			</ol>
-	</div>
-	<div class="example">
-		<pre class="lang-css">
-		@keyframes example {
-			from { column-rule-width: 10px repeat(auto, 20px) 30px }
-			to   { column-rule-width: 20px repeat(auto, 40px) 40px }
-		}
-		</pre>
-		Length of the |leading gaps| and |trailing gaps| match, so we can apply the [=repeatable list=] algorithm to each segment.
-		<pre>
-		From:   10px repeat(auto, 20px) 30px
-		At 50%: 15px repeat(auto, 30px) 35px
-		To:     20px repeat(auto, 40px) 40px
-		</pre>
-	</div>
 
-	<div class="example">
-		<pre class="lang-css">
-		@keyframes example {
-			from { column-rule-width: 10px repeat(auto, 20px) 30px }
-			to   { column-rule-width: 20px repeat(auto, 40px 50px) 40px }
-		}
-		</pre>
-		Length of the |leading gaps| and |trailing gaps| match, so we can apply the [=repeatable list=] algorithm to each segment,
-		applying <a>list interpolation</a> to list of values in the |auto repeat| segment.
-		<pre>
-		From:   10px repeat(auto, 20px 20px) 30px
-		At 50%: 15px repeat(auto, 30px 35px) 35px
-		To:     20px repeat(auto, 40px 50px) 40px
-		</pre>
-	</div>
+		<div class="example">
+			<pre class="lang-css">
+			@keyframes example {
+				from { column-rule-width: 10px repeat(auto, 20px) 30px }
+				to   { column-rule-width: 20px repeat(auto, 40px) 40px }
+			}
+			</pre>
+			Length of the |leading values| and |trailing values| across |from| and |to| match, so we can apply the [=repeatable list=] algorithm to each segment.
+			<pre>
+			From:   10px repeat(auto, 20px) 30px
+			At 50%: 15px repeat(auto, 30px) 35px
+			To:     20px repeat(auto, 40px) 40px
+			</pre>
+		</div>
 
-	<div class="example">
-		<pre class="lang-css">
-		@keyframes example {
-			from { column-rule-width: 10px repeat(auto, 20px) }
-			to   { column-rule-width: 20px 30px repeat(auto, 40px 50px) }
-		}
-		</pre>
-		Length of the |leading gaps| and |trailing gaps| match, so we can apply the [=repeatable list=] algorithm to each segment,
-		applying <a>list interpolation</a> to list of values in the |auto repeat| segment.
-		<pre>
-		From:   10px 10px repeat(auto, 20px 20px)
-		At 50%: 15px 20px repeat(auto, 30px 35px)
-		To:     20px 30px repeat(auto, 40px 50px)
-		</pre>
-	</div>
+		<div class="example">
+			<pre class="lang-css">
+			@keyframes example {
+				from { column-rule-width: 10px 20px repeat(auto, 20px) 30px }
+				to   { column-rule-width: 20px 30px repeat(auto, 40px 50px) 40px }
+			}
+			</pre>
+			Length of the |leading values| and |trailing values| across |from| and |to| match, so we can apply the [=repeatable list=] algorithm to each segment,
+			applying <a>list interpolation</a> to list of values in the |auto repeat| segment.
+			<pre>
+			From:   10px 20px repeat(auto, 20px 20px) 30px
+			At 50%: 15px 25px repeat(auto, 30px 35px) 35px
+			To:     20px 30px repeat(auto, 40px 50px) 40px
+			</pre>
+		</div>
 
-	<div class="example">
-		<pre class="lang-css">
-		@keyframes example {
-			from { column-rule-width: 10px repeat(auto, 20px) 30px }
-			to   { column-rule-width: 20px repeat(auto, 40px) 40px 50px }
-		}
-		</pre>
-		Length of the |trailing gaps| and |leading gaps| in |to| do not match, so we fall back to [=discrete=] interpolation.
-	</div>
+		<div class="example">
+			<pre class="lang-css">
+			@keyframes example {
+				from { column-rule-width: 10px repeat(auto, 20px) }
+				to   { column-rule-width: 20px 30px repeat(auto, 40px 50px) }
+			}
+			</pre>
+			Length of the |from| |leading values| and |to| |leading values| don't match, so we fall back to [=discrete=] interpolation.
+		</div>
 
-	<div class="example">
-		<pre class="lang-css">
-		@keyframes example {
-			from { column-rule-width: 10px repeat(auto, 20px) 30px }
-			to   { column-rule-width: 20px }
-		}
-		</pre>
-		Only |from| contains an <a>auto repeater</a>, so we fall back to [=discrete=] interpolation.
+		<div class="example">
+			<pre class="lang-css">
+			@keyframes example {
+				from { column-rule-width: 10px repeat(auto, 20px) 30px }
+				to   { column-rule-width: 20px repeat(auto, 40px) 40px 50px }
+			}
+			</pre>
+			Length of the |from| |trailing values| and |to| |trailing values| don't match, so we fall back to [=discrete=] interpolation.
+		</div>
+
+		<div class="example">
+			<pre class="lang-css">
+			@keyframes example {
+				from { column-rule-width: 10px repeat(auto, 20px) 30px }
+				to   { column-rule-width: 20px }
+			}
+			</pre>
+			Only |from| contains an <a>auto repeater</a>, so we fall back to [=discrete=] interpolation.
+		</div>
 	</div>
 
 <h3 id="gap-decoration-shorthands">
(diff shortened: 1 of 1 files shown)


svgeesus:
@jav099 could you please link your W3C account to your GitHub account, so that the IPR bot knows who you work for? Happy to help if you don't know how to do that.

jav099:
@svgeesus I have already done so I think? The IPR bot was failing yesterday and I followed its links to link the accounts. Now if I go to my W3C account and look at my connected accounts I can see my github profile.

css-gaps-1/Overview.bs
```diff
@@ -712,7 +712,7 @@ Gap decoration color: The 'column-rule-color' and 'row-rule-color' properties</h
 		Initial: currentcolor
 		Applies to: <a>grid containers</a>, <a>flex containers</a>, <a>multicol containers</a>, and <a>masonry containers</a>
 		Inherited: no
-		Animation type: by computed value type
+		Animation type: repeatable list, (see [[#repeat-interpolation]]).
```
```suggestion
		Animation type: repeatable list, see [[#repeat-interpolation]].
```

css-gaps-1/Overview.bs
```diff
@@ -807,7 +807,7 @@ Gap decoration width: The 'column-rule-width' and 'row-rule-width' properties</h
 		Applies to: <a>grid containers</a>, <a>flex containers</a>, <a>multicol containers</a>, and <a>masonry containers</a>
 		Inherited: no
 		Computed value: list of absolute lengths, <a>snapped as a border width</a>, or ''0'' under conditions described below
-		Animation type: by computed value type
+		Animation type: repeatable list, (see [[#repeat-interpolation]]).
```
```suggestion
		Animation type: repeatable list, see [[#repeat-interpolation]].
```

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
 		except that in step 2, change all instances of "first" to "last".
 	</div>
 
+<h4 id="repeat-interpolation">Interpolation of ''repeat()''</h4>
+When combining ''repeat()'' values for 'rule-color', and 'rule-width',
```
kbabbitt: Half sentence here (left over from drafting?)

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+
+When interpolating ''repeat()'' values for 'rule-color' or 'rule-width', the interpolation proceeds in two steps:
+
+<ol>
+<li> <dfn>Expansion</dfn>: Expand each ''repeat(n, …)'' into its equivalent list of values. </li>
```
```suggestion
<li> <dfn>Expansion</dfn>: Expand each [=integer repeater=] into its equivalent list of values. </li>
```
(link to defined terms where available for consistency)

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+<li> <dfn>List Interpolation</dfn>: Apply the [=repeatable list=] interpolation algorithm to the expanded lists, interpolating each item against its counterpart.</li>
+</ol>
+<pre class="example">
+/* Repeater → Repeater */
```
kbabbitt: Suggestions for the interpolation examples:
- Applying the `example` class to `<pre>` makes the formatting inconsistent with other examples in the document; apply it to `<div>` instead and perhaps include some explanatory text
- I think it would be helpful to include the expanded `from` and `to` lists also
- Adjust the spacing so that corresponding values line up vertically in the formatted HTML (unfortunately, as best I know, this can take some trial-and-error)

Here's how I might mark one of these up, but feel free to vary the overall form and explanatory text between examples as it makes sense to do so:

```bs
	<div class="example">
		<pre class="lang-css">
		@keyframes example {
			from { column-rule-width: repeat(2, 10px 20px); }
			to   { column-rule-width: 20px; }
		}
		</pre>
		Interpolation of the above values would result in expansion of the
		"from" and "to" values to create lists of equal lengths:
		<pre>
		From:   10px 20px 10px 20px
		At 50%: 15px 20px 15px 20px
		To:     20px 20px 20px 20px
		</pre>
	</div>
```

(I just now discovered `class="lang-css"` while looking at other specs; I'll probably go back and mark up the other examples in this spec with it at some point)

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+</pre>
+
+<h5 id="auto-repeaters">Lists with Auto Repeaters</h5>
+When either of the lists we are interpolating between (let's call them |from| and |to|) include an <a>auto repeater</a> (e.g. ''repeat(auto, …)''), we conceptually split the list into three segments, similar
+to how we [=assign gap decoration values=]:
+- |leading gaps|
+- |auto repeat|
+- |trailing gaps|
+
+Interpolation rules:
```
kbabbitt: This should be marked up as an algorithm (`<div algorithm>`) and incorporate all of the steps done by an implementation, including splitting up the list and (earlier) expanding integer repeaters. See the definition for "assign gap decoration values" in this spec for an example.

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+
+<pre class="example">
+/* Repeater → Non-Repeater */
+from: repeat(2, 10px 20px)
+to:   20px 30px
+/* At 50%: 15px 25px 15px 25px */
+</pre>
+
+<h5 id="auto-repeaters">Lists with Auto Repeaters</h5>
+When either of the lists we are interpolating between (let's call them |from| and |to|) include an <a>auto repeater</a> (e.g. ''repeat(auto, …)''), we conceptually split the list into three segments, similar
```
```suggestion
When either of the lists we are interpolating between (let's call them |from| and |to|) include an <a>auto repeater</a>, we conceptually split the list into three segments, similar
```

(the "e.g." is not needed since the autolink will take the reader to the definition of the term; omit it for brevity)

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
 		except that in step 2, change all instances of "first" to "last".
 	</div>
 
+<h4 id="repeat-interpolation">Interpolation of ''repeat()''</h4>
```
kbabbitt: I think these rules also apply to cases where the "from" and "to" values don't contain any repeaters but do have unequal list lengths, correct? Assuming that's the case, I'd suggest generalizing the section title and introductory test to reflect that.

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+from: 10px repeat(auto, 20px) 30px
+to:   20px repeat(auto, 40px) 40px
+/* At 50%: 15px repeat(auto, 30px) 35px */
+</pre>
+
+<pre class="example">
+/* Auto-Repeater → Auto-Repeater (Discrete interpolation, segment lengths don't match) */
+from: 10px repeat(auto, 20px) 30px
+to:   20px repeat(auto, 40px) 40px 50px
+/* At 50%: 15px repeat(auto, 30px) 40px 50px */
```
kbabbitt: Are we interpolating part of the value continuously and part discretely? That seems weird to me - I'd expect that if the lists can't be lined up, the entire interpolation would be treated as discrete. Though maybe there's precedent for doing it this way elsewhere?

css-gaps-1/Overview.bs
```diff
@@ -982,6 +982,72 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+</pre>
+
+<pre class="example">
+/* Auto-Repeater → Non-Repeater (Discrete interpolation, segment lengths don't match) */
+from: 10px repeat(auto, 20px) 30px
+to:   20px 40px
+/* At 50%: 20px 40px */
```
kbabbitt: For cases where a discrete interpolation happens, the "At 50%" line doesn't add much value. I would instead have prose stating that the interpolation is discrete and explaining the specific reason why (e.g. "because only one value contains an [=auto repeater=]").

css-gaps-1/Overview.bs
```diff
@@ -1066,4 +1132,6 @@ Changes since the <a href="https://www.w3.org/TR/2025/WD-css-gaps-1-20250417/">1
 		<li>Specified the behavior when gaps are coincident due to track collapsing.
 			(<a href="https://github.com/w3c/csswg-drafts/issues/11814">Issue 11814</a>)
 		<li>Added links to WPT suite.
+		<li>Specified the interpolation behavior for ''repeat()'', 'rule-color', and 'rule-width'.
```
```suggestion
		<li>Specified the interpolation behavior for values which may contain repeaters.
```

css-gaps-1/Overview.bs
jav099: Oh no we are not, that was my bad I didn't update the 50% accordingly.

css-gaps-1/Overview.bs
jav099: Great suggestions, thanks!

css-gaps-1/Overview.bs
jav099: Yes good point, this behavior applies to all interpolations

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
 		</ol>
 	</div>
 
+	<div algorithm>
+		To <dfn>assign gap decoration values in reverse</dfn>
+		to a list of <var ignore=''>gaps</var> using a list of <var ignore=''>values</var>,
+		follow the same steps as to <a>assign gap decoration values</a>,
+		except that in step 2, change all instances of "first" to "last".
+	</div>
```
kbabbitt: This looks like an incorrect merge; the "assign gap decoration values in reverse" algorithm was removed in https://github.com/w3c/csswg-drafts/commit/f52013a7641e2f21d04831617e41254a060984c0

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+When interpolating ''repeat()'' values, or lists of values for 'rule-color' or 'rule-width', the interpolation proceeds in two steps:
+
+	<ol>
+	<li> <dfn>Expansion</dfn>: Expand any [=integer repeater=] into its equivalent list of values. </li>
+	<li> <dfn>List Interpolation</dfn>: Apply the [=repeatable list=] interpolation algorithm to the expanded lists, interpolating each item against its counterpart.</li>
+	</ol>
+	<div class="example">
+		<pre class="lang-css">
+		@keyframes example {
+			from { column-rule-width: 10px }
```
```suggestion
			from { column-rule-width: 10px; }
```

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+			to   { column-rule-width: 20px 40px }
```
```suggestion
			to   { column-rule-width: 20px 40px; }
```

(these semicolons aren't technically required but I'd include them for consistency with the other examples)

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+		<pre class="lang-css">
+		@keyframes example {
+			from { column-rule-width: repeat(2, 5px 10px); }
+			to   { column-rule-width: repeat(2, 15px 20px); }
+		}
+		</pre>
+		Interpolation of the above values would result in expansion of the
+		"from" and "to" values to create lists of equal lengths:
+		<pre>
+		From:   5px 10px 5px 10px
```
```suggestion
		From:    5px 10px  5px 10px
```
(leading spaces in front of the single digits for alignment)

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+		<pre>
+		From:   10px 20px 10px 20px
+		At 50%: 15px 25px 15px 25px
+		To:     20px 30px 20px 30px
+		</pre>
+	</div>
+
+
+<h5 id="auto-repeaters">Lists with Auto Repeaters</h5>
+	<div algorithm>
```
kbabbitt, line 1049: General comment: I believe that algorithms in specs should be described with a high degree of precision, so that it's very clear to an implementer how to transcribe the algorithm from English into their programming language of choice. Often, but not always, you end up with something that reads like a literal description of what C++ code is doing. (I'll call out specific points but also wanted to explain the reasoning behind my comments.)

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+
+		To <dfn>interpolate with auto repeaters</dfn>, when either of the lists we are interpolating between (let's call them |from| and |to|) include an <a>auto repeater</a>, we conceptually split the list into three segments, similar
+		to how we [=assign gap decoration values=]:
+		- |leading gaps|
+		- |auto repeat|
+		- |trailing gaps|
+
+			<ol>
+				<li>Let |segment lengths| be the length of the |leading gaps| and |trailing gaps| segments (i.e. the number of items in each list, after <a>expansion</a>).</li>
```
kbabbitt: Within an `algorithm` block, marking up a word or phrase with vertical bars indicates that it's a variable. (The produced markup also has a neat feature where you can click on a variable name in an algorithm and it will highlight all instances of that variable with the same color. Then you can click on a different variable, and it will highlight all instances of *that* variable in its own color.) See also: https://speced.github.io/bikeshed/#var-and-algorithms

This line reads like there's a single variable |segment lengths| that's somehow computed from |leading gaps| and |trailing gaps|, but that's not the case.

Having said all that... I don't know that we need to be completely pedantic and split everything out into |from-segment-lengths|, |from-leading-gaps| and so on either. It might be clear enough to do something like "|from|'s |leading gaps|" and so on. I do think |segment lengths| is a bit overloaded, and maybe we don't even need it - could we instead say things like "the length of |from|'s |leading gaps|" or "the lengths of |from|'s and |to|'s |leading gaps|", for example?

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+		</pre>
+	</div>
+
+
+<h5 id="auto-repeaters">Lists with Auto Repeaters</h5>
+	<div algorithm>
+
+		To <dfn>interpolate with auto repeaters</dfn>, when either of the lists we are interpolating between (let's call them |from| and |to|) include an <a>auto repeater</a>, we conceptually split the list into three segments, similar
+		to how we [=assign gap decoration values=]:
+		- |leading gaps|
```
kbabbitt: Suggest something like |leading values|, |auto values|, and |trailing values| for these, which reflect the parallel with the partitioning in "assign gap decoration values" but also highlight the subtle distinction that we're partitioning a list of values in this case rather than a list of gaps. (I say "something like" because, in an implementation, I suspect these would actually be two variables each, one based on |from| and one based on |to| - see comment below.)

This might also be a convenient spot to specify that (for each list) |auto values| is assigned the list of values within the auto repeater, rather than the auto repeater itself, if we end up with something of the form "Let |leading values| be the list of X. Let |auto values| be the list of Y. Let |trailing values| be the list of Z."

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+				<li>Let |segment lengths| be the length of the |leading gaps| and |trailing gaps| segments (i.e. the number of items in each list, after <a>expansion</a>).</li>
+				<li>We split |to| and/or |from| into their respective segments.</li>
+				<li>If either of |leading gaps| or |trailing gaps| in |to| or |from| is empty and the other has an <a>auto repeater</a>, we do [=discrete=] interpolation.</li>
+				<li>Expand any integer repeaters on each segment.</li>
+				<li>If the |segment lengths| don't match, we fall back to [=discrete=] interpolation.</li>
+				<li>When both |from| and |to| contain auto-repeaters and |segment lengths| match, we apply <a>list interpolation</a> to each segment. Applying <a>list interpolation</a> to the |auto repeat| means applying it to the list of values inside the repeater.</li>
+				<li>If only one of |from| or |to| contains an auto-repeater, we do [=discrete=] interpolation.</li>
+			</ol>
+	</div>
+	<div class="example">
```
kbabbitt: Referring back to variables from the algorithm in these examples is a nice touch. I wonder - does the spec format correctly if these examples are contained within the `<div algorithm>` block, so that the highlighting feature I mentioned above would work in the examples as well?

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+		To:     20px repeat(auto, 40px 50px) 40px
+		</pre>
+	</div>
+
+	<div class="example">
+		<pre class="lang-css">
+		@keyframes example {
+			from { column-rule-width: 10px repeat(auto, 20px) }
+			to   { column-rule-width: 20px 30px repeat(auto, 40px 50px) }
+		}
```
kbabbitt: This pair of values looks to me like it needs to be discrete. The `<pre>` block below suggests we're applying LCM to the leading value lists, but the problem I see is that `10px repeat(auto, 20px)` is not equivalent to `10px 10px repeat(auto, 20px)`. (Or is there a typo in the `from` keyframe?)

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+		- |auto repeat|
+		- |trailing gaps|
+
+			<ol>
+				<li>Let |segment lengths| be the length of the |leading gaps| and |trailing gaps| segments (i.e. the number of items in each list, after <a>expansion</a>).</li>
+				<li>We split |to| and/or |from| into their respective segments.</li>
+				<li>If either of |leading gaps| or |trailing gaps| in |to| or |from| is empty and the other has an <a>auto repeater</a>, we do [=discrete=] interpolation.</li>
+				<li>Expand any integer repeaters on each segment.</li>
+				<li>If the |segment lengths| don't match, we fall back to [=discrete=] interpolation.</li>
+				<li>When both |from| and |to| contain auto-repeaters and |segment lengths| match, we apply <a>list interpolation</a> to each segment. Applying <a>list interpolation</a> to the |auto repeat| means applying it to the list of values inside the repeater.</li>
```
kbabbitt: On this line and the previous line, it's not clear which |segment lengths| need to match.

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+		At 50%: 15px 25px 15px 25px
+		To:     20px 30px 20px 30px
+		</pre>
+	</div>
+
+
+<h5 id="auto-repeaters">Lists with Auto Repeaters</h5>
+	<div algorithm>
+
+		To <dfn>interpolate with auto repeaters</dfn>, when either of the lists we are interpolating between (let's call them |from| and |to|) include an <a>auto repeater</a>, we conceptually split the list into three segments, similar
```
kbabbitt: I might move the check for only one of |from| and |to| having an auto repeater up to the top of the algorithm. Then this step can be rewritten as "Split each of |from| and |to| into..." which would be clearer.

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+		to how we [=assign gap decoration values=]:
+		- |leading gaps|
+		- |auto repeat|
+		- |trailing gaps|
+
+			<ol>
+				<li>Let |segment lengths| be the length of the |leading gaps| and |trailing gaps| segments (i.e. the number of items in each list, after <a>expansion</a>).</li>
+				<li>We split |to| and/or |from| into their respective segments.</li>
+				<li>If either of |leading gaps| or |trailing gaps| in |to| or |from| is empty and the other has an <a>auto repeater</a>, we do [=discrete=] interpolation.</li>
```
kbabbitt: This line could be interpreted as the discrete interpolation applying to just the leading or trailing segments rather than the whole value.

Also, I think it's easier to follow when |from| and |to| are in that order consistently.

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,169 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+				<li>Expand any integer repeaters on each segment.</li>
```
kbabbitt, line 1061: This step is confusing, because an earlier step in the algorithm talked about counting the list lengths "after expansion" but now this step is telling us to expand integer repeaters.

css-gaps-1/Overview.bs
jav099: It does! moved to inside that div.

css-gaps-1/Overview.bs
```diff
@@ -975,6 +975,162 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
...
+				Let |trailing values| be the values after the <a>auto repeater</a>.
+				Let |auto values| be the values inside the <a>auto repeater</a>.
+				</li>
+				<li>If either of |leading values| or |trailing values| in |from| or |to| is empty and the other has an <a>auto repeater</a>, we fall back to [=discrete=] interpolation.</li>
+				<li>Expand any integer repeaters on each segment.</li>
+				<li>
+				If the length of |leading values| in |from| and |leading values| in |to| don't match, we fall back to [=discrete=] interpolation.
+				If the length of |trailing values| in |from| and |trailing values| in |to| don't match, we fall back to [=discrete=] interpolation.
+				</li>
+				<li>If only one of |from| or |to| contains an auto-repeater, we fall back to [=discrete=] interpolation.</li>
```
kbabbitt: Wondering if it would make sense to check this condition first, before we analyze leading/trailing values in depth? I think that would also eliminate the need to check for an auto repeater in step 2.

kbabbitt (approved):
Looks great overall! Just one minor suggestion

Issue #12084, opened 2025-04-15
[css-gaps-1] Gap intersection point definition might need updating for multi-col


jav099:
Right now we have defined gap intersection points [here](https://drafts.csswg.org/css-gaps-1/#layout-painting). However, we might want to expand this such that in multicol containers, we also have an intersection point where a gap meets a spanner. 

For instance in https://jsbin.com/welojoheko/edit?html,output, we can observe that currently we are not painting behind the spanner (cyan) when painting the column rules, so we might want to add intersections where the gaps meet the spanner, so we can implement this behavior as well as our different `rule-breaks`. I should note too, that for the purpose of `outset` we might also want to treat these intersections as if they were intersections with the edge of the container, rather than intersections with other gaps.

`<style>
    body {
        margin: 0px;
    }

    .container {
        border: 2px solid rgb(96 139 168);
        width: 200px;
        height: 200px;
        column-height: 60px;
        column-gap: 10px;
        row-gap: 10px;
        column-rule-width: 10px;
        column-rule-style: solid;
        column-rule-color: blue;
        row-rule-width: 10px;
        row-rule-style: solid;
        row-rule-color: gold;
        column-wrap: wrap;
        column-width: 60px;
        column-count: 3;
        column-fill: auto;
    }

    p {
        background: rgb(96 139 168 / 0.2);
        height: 60px;
        margin: 0px;
    }

    h2 {
        column-span: all;
        background-color: cyan;
        color: #fff;
        margin: 0px;
        opacity: 0.5;
        height:18px;
    }
</style>

<body>
    <div class="container">
        <p></p>
        <p></p>
        <h2></h2>

        <p></p>
        <p></p>
        <p></p>
        <p></p>
        <p></p>
        <p></p>
    </div>
</body>`

kbabbitt:
@jav099 since you implemented this in Chromium, do you have something in mind in terms of what the revised definition for "gap intersection point" might look like?

jav099:
I've tried to think of a generic way to do it, but since multicol spanners are pretty unique to multicolumn i could only really think of doing the following to the definition:

* The center of an intersection between a gap and the content edge of the container, or the edge of a [multi-column spanner](https://drafts.csswg.org/css-multicol/#multi-column-spanner).

Thoughts?

kbabbitt:
I dug into this a bit more. Excerpts from https://www.w3.org/TR/css-multicol-1/#the-multi-column-model:

[...]

PR #12584, opened 2025-08-07
[css-gaps-1] Update gap intersection definition. Issue #12084

jav099:
This PR aims to update the definition of a gap intersection point to solve the concerns raised in issue https://github.com/w3c/csswg-drafts/issues/12084. This required a minor update to css-align-3 in how we define the gap for flex containers, which now will use language that multi column and grid use.

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -112,7 +112,7 @@ Layout and painting</h3>
 
 	A <dfn>gap intersection point</dfn> is defined to exist in each of the following locations:
 	<ul>
-		<li>The center of an intersection between a gap and the content edge of the container.</li>
+		<li>The center of an intersection between a gap and the end edge of a [=gutter=].</li>
 		<li>The center of an intersection between gaps in different directions.</li>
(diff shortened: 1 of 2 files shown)


css-gaps-1/Overview.bs
```diff
@@ -112,7 +112,7 @@ Layout and painting</h3>
 
 	A <dfn>gap intersection point</dfn> is defined to exist in each of the following locations:
 	<ul>
-		<li>The center of an intersection between a gap and the content edge of the container.</li>
+		<li>The center of an intersection between a gap and the end edge of a [=gutter=].</li>
```
```suggestion
		<li>The center of the start or end edge of a [=gutter=].</li>
```

2 points:
1. "Gutter" and "Gap" mean roughly the same thing, and gutters/gaps don't intersect with themselves
2. The definition should include both start and end edges

jav099: I incorporated this feedback.

kbabbitt (approved):

kbabbitt:
@tabatkins could you please review the changes to css-align-3?

Issue #12596, opened 2025-08-12
[css-color] Add display-p3-linear


annevk:
As per https://developer.apple.com/documentation/realitykit/construct-an-immersive-environment-for-visionos#Generate-efficient-textures Display P3 is the best color space for visionOS. Display P3 Linear would still require some conversion, but it's significantly less than with rec-2100-linear for instance. This is because it's very close to the final color space. And as a result display-p3-linear has better performance.

Therefore colleagues and I think it's a good idea to add display-p3-linear to CSS, 2D Canvas, etc.

svgeesus:
I can certainly see the benefit, and would be happy to add it given implementer interest and a CSSWG resolution.

svgeesus:
@emilio Apple proposed this and Chrome is happy to add, do you have an opinion?

(This is calculated as an intermediate step when converting to or from `display-p3` so implementation is trivial, the proposal is just to expose it in CSS syntax)

svgeesus:
The CSS Working Group just [discussed](https://www.w3.org/2025/08/19-css-color-minutes.html#3461) `[css-color] Add display-p3-linear` and agreed to the following:

 * `RESOLUTION: add display-p3-linear to the predefined color-spaces`
 
<details><summary>The full IRC log of that discussion</summary>
ChrisL: this fit nicely with hardware. in practice people use p3, but linear-p3 is calculated anyway <br>
ccameron: this space is coming from the Canvas world <br>
proposal: add display-p3-linear to the predefined color-spaces <br>
lea: maybe we should have linear version of all RGB color-spaces already existing <br>
ccameron: we could start with srgb, 2020 <br>
<lea> s/lea: maybe we should have linear version of all color-spaces already existing/lea: maybe we should have linear version of all RGB color-spaces already existing/ <br>
 `RESOLUTION: add display-p3-linear to the predefined color-spaces`
</details>

romainmenke:
Should `display-p3-linear` also be added as an interpolation color space?
I think [this list](https://drafts.csswg.org/css-color-5/#interpolation-space) is maintained separately.

svgeesus:
Thanks, good catch

jonsneyers:
Just a nitpick on the name: wouldn't it make more sense to call this simply `p3-linear` instead of `display-p3-linear`?

[...]

PR #12601, opened 2025-08-13
[css-color-4] Add display-p3-linear #12596

svgeesus:
Draft for now until CSSWG resolution

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -933,6 +933,7 @@ Representing Colors: the <<color>> type</h2>
 		[[#predefined-sRGB|sRGB]],
 		[[#predefined-sRGB-linear|Linear-light sRGB]],
 		[[#predefined-display-p3|Display P3]],
+		[[#predefined-display-p3-linear|Linear-light Display P3]],
 		[[#predefined-a98-rgb|A98 RGB]],
 		[[#predefined-prophoto-rgb|ProPhoto RGB]],
 		[[#predefined-rec2020|ITU-R BT.2020-2]],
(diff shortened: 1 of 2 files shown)


svgeesus:
The sample code does not need to be updated as it already converts to and from `display-p3-linear` as part of converting to and from `display-p3`.

css-color-adjust-1/Overview.bs
```diff
@@ -3,6 +3,7 @@ Title: CSS Color Adjustment Module Level 1
 Shortname: css-color-adjust
 Level: 1
 Status: ED
+Prepare for TR: no
```
annevk: This seems like it should be separated out?

css-color-4/Overview.bs
```diff
@@ -4119,6 +4120,56 @@ The Predefined Display P3 Color Space: the ''display-p3'' keyword</h3>
 			2d.color.space.p3.toDataURL.with.putImageData.html
 		</wpt>
 
+<h3 id="predefined-display-p3-linear">
+The Predefined Linear-light Display P3 Color Space: the ''display-p3-linear'' keyword</h3>
```
annevk: I would expect "Linear-Light" and also "Keyword" if you are going for title casing.

css-color-4/Overview.bs
```diff
@@ -4714,7 +4765,7 @@ Color Space for Interpolation</h3>
 	Thus, different color spaces may be more appropriate for each interpolation use case.
 
 	* In some cases, the result of physically mixing two colored lights is desired.
-		In that case, the CIE XYZ  or srgb-linear color space is appropriate, because they are linear in light intensity.
+		In that case, the CIE XYZ, display-p3-linear  or srgb-linear color spaces are appropriate, because they are linear in light intensity.
```
annevk: Should these be linked to their definition?

annevk (commented):
Thanks for putting this up! I'm not a color space expert, but I've done an editorial pass.

css-color-4/Overview.bs
svgeesus: Agreed

css-color-4/Overview.bs
svgeesus: Agree on Linear-Light; the rest of the spec consistently has lower case for "keyword" in section titles, so I left that as-is

css-color-adjust-1/Overview.bs
svgeesus: Yes

PR #12606, opened 2025-08-14
[css-scroll-snap] [cssom-view] Use a single scroll queue for all events

### LINKED ISSUE #11164: [cssom-view] Scroll steps don't match implementations
https://drafts.csswg.org/cssom-view/#document-run-the-scroll-steps has a few issues afaict:

 * It doesn't fire `scrollend` events. Instead `scrollend` events are defined to fire "Whenever scrolling is completed", which seems like it should fire sync, which is clearly wrong as it'd fire before the `scroll` events?
 * Per spec all current `scrollsnapchanging` events are dispatched before all other scroll events.
 * Per spec there's no difference between `VisualViewport` scroll event order and other scroll events.

Implementations disagree with this in various ways:

 * My read of the [relevant Chromium code](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/page/page_animator.cc;l=260-265;drc=0afc9ac9afcaab79fc54299039f4d27abf3a086d) is that all these events are stored in a single queue and dispatched in one go. So `scrollend` / `scroll` / `scrollsnap*` events are just sorted in one queue.
 * The relevant [Firefox code](https://searchfox.org/mozilla-central/rev/113b09cdd5282b39db65b2f62abe843558b9471a/layout/base/nsRefreshDriver.cpp#2733-2736) intends to dispatch regular scroll, then VisualViewport scroll, then scrollend (then VisualViewport scrollend). However [it's bogus](https://bugzilla.mozilla.org/show_bug.cgi?id=1929707) and right now it dispatches the scroll and scrollend on the same queue, and then VisualViewport scroll.
 * The relevant [WebKit code](https://searchfox.org/wubkat/rev/d45649fa32a499702c30474e94bd6843c4c67849/Source/WebCore/dom/Document.cpp#5230-5243) dispatches scroll events, then VisualViewport scroll events. Note that WebKit doesn't support scrollend.

We should pick a model for this that makes sense, specially in presence of `VisualViewport` scrollend events and such.

The Chromium model is simple and makes sense to me: just queue all the scroll-related events in a frame into a single list and dispatch it in order. Any scroll-related event triggered by those gets dispatched on the next frame. But I wonder if people feel strongly about the event ordering.

An interesting thing to think about if we decide to use separate queues per event type, is whether we want to treat them "atomically" or not. E.g., whether the events triggered by earlier lists can affect later lists in the same frame. For example, the spec as written means that `scrollsnapchanging` events are always before `scroll`, and a `scrollsnapchanging` event listener triggering a scroll should fire in the same frame, but if the scroll event triggers snapping then that won't fire snap events until next frame which is a bit odd.

Any strong opinions?

cc @dlrobertson @theres-waldo @lilles @smfr @flackr

emilio:
Fixes #11164

### DIFF
--- a/cssom-view-1/Overview.bs
+++ b/cssom-view-1/Overview.bs
@@ -2002,9 +2002,7 @@ When asked to <dfn export for=Document>run the resize steps</dfn> for a {{Docume
 
 This section integrates with the <a for=/>event loop</a> defined in HTML. [[!HTML]]
 
-Each {{Document}} has an associated list of <dfn for=Document>pending scroll event targets</dfn>, initially empty.
-
-Each {{Document}} has an associated list of <dfn for=Document>pending scrollend event targets</dfn>, initially empty.
+Each {{Document}} has an associated list of <dfn for=Document>pending scroll events</dfn>, which stores pairs of ({{EventTarget}}, {{DOMString}}), initially empty.
 
 Whenever a <a>viewport</a> gets scrolled (whether in response to user interaction or by an API), the user agent must run these steps:
 
@@ -2014,8 +2012,8 @@ Whenever a <a>viewport</a> gets scrolled (whether in response to user interactio
     for <var>doc</var> with <var>doc</var>'s <a spec="css-scroll-snap-2">eventual snap target</a> in the block axis
     as newBlockTarget and <var>doc</var>'s <a spec="css-scroll-snap-2">eventual snap target</a> in the inline axis
     as newInlineTarget.
-1. If <var>doc</var> is already in <var>doc</var>'s <a>pending scroll event targets</a>, abort these steps.
-1. Append <var>doc</var> to <var>doc</var>'s <a>pending scroll event targets</a>.
+1. If (<var>doc</var>, <code>"scroll"</code>) is already in <var>doc</var>'s <a>pending scroll events</a>, abort these steps.
+1. Append (<var>doc</var>, <code>"scroll"</code>) to <var>doc</var>'s <a>pending scroll events</a>.
 
 Whenever an element gets scrolled (whether in response to user interaction or by an API), the user agent must run these steps:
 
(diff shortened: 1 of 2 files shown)


nt1m (approved):
I was hoping someone more seasoned with scrolling would review, but for the sake of moving this forward, I'll approve this

cssom-view-1/Overview.bs
```diff
@@ -2056,13 +2043,40 @@ Issue: In what order are scrollend events dispatched? Ordered based on scroll st
...
-    1. Otherwise, <a>fire an event</a> named <a event>scrollend</a> at <var>target</var>.
-1. Empty <var>doc</var>'s <a>pending scrollend event targets</a>.
+    1. If (<var>target</var>, <code>"scrollend"</code>) is already in <var>doc</var>'s <a>pending scroll events</a>, abort these steps.
+    1. Append (<var>target</var>, <code>"scrollend"</code>) to <var>doc</var>'s <a>pending scroll events</a>.
+
+1. For each item (<var>target</var>, <var>type</var>) in <var>doc</var>'s <a>pending scroll events</a>,
+    in the order they were added to the list, run these substeps:
+
+    1. If <var>target</var> is a {{Document}}, and <var>type</var> is <code>"scroll"</code> or <code>"scrollend"</code>,
+        <a>fire an event</a> named <a event>scroll</a> that bubbles at <var>target</var>.
```
nt1m, line 2053: It should fire an event named scroll or scrollend here

emilio: Ah, nice catch, #12835

nt1m:
@emilio Now that this is published, I noticed that this is missing indentation that makes the if/otherwise steps a bit hard to read (notably in "run the scroll steps"): https://drafts.csswg.org/cssom-view/#scrolling-events

Is it possible to fix that?

AtkinsSJ:
> @emilio Now that this is published, I noticed that this is missing indentation that makes the if/otherwise steps a bit hard to read (notably in "run the scroll steps"): [drafts.csswg.org/cssom-view#scrolling-events](https://drafts.csswg.org/cssom-view/#scrolling-events)
> 
> Is it possible to fix that?

I've submitted a fix for this in #13238

PR #12619, opened 2025-08-19
let `wrap-reverse` define more clearer

zhangenming:
Suggestion to move some text about the effect of wrap-reverse from the main prose to the dl defining the various keywords


svgeesus (approved):
Yes, this is clearer

fantasai:
@zhangenming The changes look correct to me! Definitely an improvement. Would you mind unwinding the unrelated changes so we only have the first commit?

PR #12624, opened 2025-08-19
[css-view-transitions-1] Use `position: absolute` instead of `position: fixed` on `::view-transition`

### LINKED ISSUE #12116: [css-view-transitions-1] What's the point of position: fixed in `::view-transition`?
https://drafts.csswg.org/css-view-transitions-1/#snapshot-containing-block-concept says:

> The [snapshot containing block](https://drafts.csswg.org/css-view-transitions-1/#snapshot-containing-block) is considered to be an [absolute positioning containing block](https://drafts.csswg.org/css-position-3/#absolute-positioning-containing-block) and a [fixed positioning containing block](https://drafts.csswg.org/css-position-3/#fixed-positioning-containing-block) for [::view-transition](https://drafts.csswg.org/css-view-transitions-1/#selectordef-view-transition) and its descendants.

Then https://drafts.csswg.org/css-view-transitions-1/#ua-styles:

```css
:root::view-transition {
  position: fixed;
  inset: 0;
}
```

but afaict that `position: fixed` is just confusing, and behaving exactly the same as `position: absolute`, since the snapshot CB is a `position: fixed` containing block, right?

Should we change it just for clarity? Is it just an historical artifact?

cc @noamr @BorisChiou @vmpstr

nt1m:
Fixes #12116

### DIFF
--- a/css-view-transitions-1/Overview.bs
+++ b/css-view-transitions-1/Overview.bs
@@ -857,7 +857,7 @@ urlPrefix: https://wicg.github.io/navigation-api/; type: interface;
 	}
 
 	:root::view-transition {
-		position: fixed;
+		position: absolute;
 		inset: 0;
 	}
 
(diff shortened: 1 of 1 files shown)


BorisChiou:
I'm updating the ua stylesheet in Gecko. @noamr @vmpstr would you mind taking a look at this spec change?

emilio (approved):

nt1m:
@emilio @BorisChiou If you could create a WPT for this, it would be very appreciated, happy to follow up with a WebKit change too.

BorisChiou:
I will update `pseudo-get-computed-style.html` and `pseudo-get-computed-style-clean-style.html` to make sure the computed value of `position` of `::view-transition` (i.e. `getComputedStyle(document.documentElement, "::view-transition").position`) is `absolute`.

UPDATED: in gecko bug 1983941 (https://bugzilla.mozilla.org/show_bug.cgi?id=1983941)

PR #12634, opened 2025-08-21
[css-selectors-5]: move from :heading(<An+B>#) to :heading(<integer>#)

### LINKED ISSUE #12599: Reconsider <An+B># for :heading
I'm not convinced supporting a comma-separated list for the `:heading` pseudo-class alone is worth it. If it shared `<An+B>#` with another pseudo-class it'd be more persuasive, but for a pseudo-class that can match at most six levels (perhaps nine in the near future) it's really not compelling enough to add the additional parsing and serialization complexity.

The argument @keithamus gave me is that `:heading(1, 2, 3)` is easier than `:heading(-n+3)`, but you can also write `:heading(1), :heading(2), :heading(3)` and that will work just as well.

My suggestion is that we use `<An+B>`, matching `:nth-of-type`.

cc @zcorpan @tabatkins @nt1m

keithamus:
Closes #12599

### DIFF
--- a/selectors-5/Overview.bs
+++ b/selectors-5/Overview.bs
@@ -173,12 +173,15 @@ Heading Structures: the heading pseudo-classes '':heading'', and '':heading()''<
 
 	As a functional pseudo-class,
 	<dfn id='heading-functional-pseudo' lt=':heading()'>:heading()</dfn>
-	notation represents elements that have a <a>heading level</a> among <var>An+B</var>.
+	notation represents elements that have a <a>heading level</a> among matching
+	any of the provided <var>integer</var> values.
 	The syntax is:
 
 	<pre class=prod>
-		:heading() = :heading( <<An+B>># )
+		:heading() = :heading( <<level>># )
 	</pre>
+
+	where <dfn><code>&lt;level></code></dfn> is a <<number-token>> with its type flag set to "integer".
 
 	The [=specificity=] of '':heading()'' is that of a class.
 
@@ -188,10 +191,10 @@ Heading Structures: the heading pseudo-classes '':heading'', and '':heading()''<
 		500, additionally heading levels 1 and 2 will be underlined, while 3 and
 		beyond will have no text-decoration:
 
-		<pre>:heading(-n+3) { font-weight: 900; }</pre>
-		<pre>:heading(n+6) { font-weight: 500; }</pre>
+		<pre>:heading(1, 2, 3) { font-weight: 900; }</pre>
+		<pre>:heading(6, 7, 8, 9) { font-weight: 500; }</pre>
 		<pre>:heading(1, 2) { text-decoration: underline; }</pre>
-		<pre>:heading(n+3) { text-decoration: none; }</pre>
+		<pre>:heading(3, 4, 5, 6, 7, 8, 9) { text-decoration: none; }</pre>
 	</div>
 
 	Note: The <a>heading level</a> might be different from an element's


selectors-5/Overview.bs
```diff
@@ -173,11 +173,12 @@ Heading Structures: the heading pseudo-classes '':heading'', and '':heading()''<
...
 	As a functional pseudo-class,
 	<dfn id='heading-functional-pseudo' lt=':heading()'>:heading()</dfn>
-	notation represents elements that have a <a>heading level</a> among <var>An+B</var>.
+	notation represents elements that have a <a>heading level</a> among matching
+	any of the provided <var>integer</var> values.
 	The syntax is:
 
 	<pre class=prod>
-		:heading() = :heading( <<An+B>># )
+		:heading() = :heading( <integer># )
```
AtkinsSJ: `<integer>` would allow things like `calc()` in here. This probably only wants to permit literal integers? IDK, maybe calculations on env vars could be useful.

Though... I don't think CSS has an "integer literal" type anyway. Just a `<number-token>` one.

keithamus: Really? https://drafts.csswg.org/css-values-4/#integers doesn't seem to suggest that, instead saying it's a subset of `<number-token>`.

SebastianZ:
Nit: The commit message and title misspell "heading" as "headng".

Sebastian

selectors-5/Overview.bs
AtkinsSJ: I think the important phrase there is "When written literally". It's not super clear in that phrasing, but you can definitely use `calc()` for other things that take `<integer>`, for example [z-index](https://drafts.csswg.org/css2/#z-index)

keithamus: I'm sure `calc()` isn't allowed inside of selectors but I cannot find the relevant prose in spec to say where `calc()` is allowed.

Perhaps @tabatkins can point us in the right direction.

AtkinsSJ: I don't think we've previously had a situation where a numeric type is allowed in a selector, though I might have missed that. I'm definitely nit-picking on technicalities here. :sweat_smile: But yeah Tab will know.

keithamus: I guess `calc()` is supported for the `<length>` of a media query (although not a selector, it demonstrates that `calc()` has support outside of declaration values).

Perhaps I can update the spec prose here to explicitly call out that only literal `<integer>` values are allowed and `calc()` would result in a parse error? I can also update the tests to reflect this.

zcorpan: https://drafts.csswg.org/css-values-4/#math says 

> A [math function](https://drafts.csswg.org/css-values-4/#math-function) represents a numeric value, one of:
...
> - [&lt;integer>](https://drafts.csswg.org/css-values-4/#integer-value)
> [...] and can be used wherever such a value would be valid.

Are arguments to functional selectors "values"? It's not super-clear to me if they should apply.

tabatkins: Yeah, selectors are kinda a weird space here.  As we add more "generic" value-handling functions (like `ident()`), we'll have to make a general decision on whether they're allowed in places like a rule's selector. I lean towards disallowing, matching how An+B only allows literal ints.

So I guess, just to be clear, I recommend using `<number-token>`, with prose specifying that the token's type flag has to be "integer", a la [the an+b definition](https://drafts.csswg.org/css-syntax/#typedef-integer).

(Looking at this, I just realized we have two competing definitions of `<integer>`, one from Syntax and one from Values, which have different definitions. That's not great. I should scope the Syntax definitions there to be specifically for an+b.)

keithamus: I opted for `<level>` as a discrete type just for a little more clarity. Hopefully this works, happy to use `<number-token>` instead though.

PR #12640, opened 2025-08-22
[css-conditional-5] Use the generic <boolean-expr>

cdoublev:
As a consequence of 5f19369, [`<boolean-condition>`](https://drafts.csswg.org/css-conditional-5/#typedef-boolean-condition) is extracted by `w3c/reffy` and exposed to `w3c/webref` users.

However, `<boolean-condition>` was intended to be defined generically, as noted under its definition, and this has been done with [`<boolean-expr>`](https://drafts.csswg.org/css-values-5/#typedef-boolean-expr) in 4fd47d8 in CSS Values 5.

Therefore this PR replaces `<boolean-condition>` with `<boolean-expr[ <if-test> ]>` in the prelude of `@when` and `@else`. The leaves of the boolean condition/expresion are consequently different:

```diff
- media( [ <mf-plain> | <mf-boolean> | <mf-range> ] )
+ media( <media-feature> | <media-condition> ) |
- supports( <declaration> )
+ supports( [ <ident> : <declaration-value> ] | <supports-condition> ) |
+ style( <style-query> )
```

I assume these differences should not exist and will be fixed later if required, and that accepting `style()` queries is ok.


css-conditional-5/Overview.bs
```diff
@@ -161,28 +161,11 @@ Generalized Conditional Rules: the ''@when'' rule</h2>
 	It is defined as:
 
 	<pre class=prod>
-		@when <<boolean-condition>> {
+		@when <<boolean-expr[ <if-test> ]>> {
```
AtkinsSJ: I believe you need to use double `<>` around if-test, see the other example here:

https://github.com/w3c/csswg-drafts/blob/c327d940c7b5d623ea87e01ecba383ef8c07f9b9/css-values-5/Overview.bs#L1591

cdoublev: Oups, fixed.

fantasai (approved):

Issue #12236, opened 2025-05-27
[css-conditional-6] Behavior of style(0 = 0px)


andruud:
A literal `0` may be used for any numeric value where a unit would normally be expected, e.g. `width: 0`. Should this unitless behavior also work for `style()` range syntax? In other words, is `style(0 = 0px)` "false" due to a type mismatch, or is it "true"?

This was also discussed in https://github.com/w3c/csswg-drafts/issues/8376, and the consensus seems to be that `0 = 0px` => "true" is the correct behavior. (cc @mirisuzanne ,@kizu)

There was also a question about whether or not we currently allow unitless zero in MQs with range syntax (@emilio), and the answer to that is "yes" (https://jsbin.com/taleroderi).

andruud:
Perhaps we don't need to spend call time on this detail.

Proposed async resolution: `A unitless zero matches any numeric data type in range queries`

Loirooriol:
Any numeric data type? Like `0 = 0deg` even though `<angle>` doesn't accept unitless 0? Seems weird to me.

Crissov:
`<hue>` does, though.

andruud:
OK, I guess we do need to spend call time on it.

tabatkins:
Note that in calculations, a unitless zero is *not* equivalent to a unit'd zero. Unitless zeros for lengths/angles are just an authoring convenience only allowed when used directly to satisfy a `<length>` (and in a few cases that explicitly call it out, an `<angle>`).

I'm supportive of `0 = 0px` being a type mismatch and not being allowed.

(It's allowed in MQs because, given any particular MQ name, we know what the corresponding value type is, and thus can apply the normal "0 is allowed as a `<length>`" rule.)

PR #12656, opened 2025-08-26
[css-conditional-5] Evaluating unitless zero against <length> values …

tursunova:
Resolved in https://github.com/w3c/csswg-drafts/issues/12236#issuecomment-3204826253

### DIFF
--- a/css-conditional-5/Overview.bs
+++ b/css-conditional-5/Overview.bs
@@ -1372,6 +1372,9 @@ Style Container Features</h3>
 			If this cannot be done, evaluate to false.
 		4. If each <<style-range-value>> from the range have the same type,
 			compute each and evaluate the comparison.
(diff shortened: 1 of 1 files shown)


css-conditional-5/Overview.bs
```diff
@@ -1372,6 +1372,9 @@ Style Container Features</h3>
 			If this cannot be done, evaluate to false.
 		4. If each <<style-range-value>> from the range have the same type,
 			compute each and evaluate the comparison.
+			If one value is a unitless zero ''0'' and the other value has a <<length>> type,
+			parse the unitless zero as a <<length>> with a value of ''0''
```
andruud: We parsed in step (3), so we should avoid an instruction to parse again here. Probably we should just say that a unitless zero must be treated as being a zero-&lt;length> when compared against a &lt;length>.

PR #12657, opened 2025-08-26
Add explainer for CSS if() function

tursunova:
https://www.w3.org/TR/css-values-5/#if-notation

### DIFF
--- a/css-values-5/if-explainer.md
+++ b/css-values-5/if-explainer.md
@@ -1,0 +1,141 @@
+# Explainer: CSS `if()` function
+
+This document is an explainer for the `if()` function proposed for CSS Values and Units Module Level 5.
+
+## Introduction
+
+The `if()` function introduces powerful, inline conditional logic to CSS properties. It allows authors to select a value for a property based on a set of ordered conditions, similar to `if/else` constructs in many programming languages. This provides a more dynamic and streamlined way to write CSS without relying on verbose selectors or JavaScript.
+
+## Motivation
+
+Currently, CSS authors often need to repeat selectors to apply different styles under different conditions. For example, to change a property based on a media query, one would write:
+
+```css
+.my-element {
+  color: blue;
+}
+
+@media (min-width: 600px) {
+  .my-element {
+    color: red;
+  }
+}
+```
+
+While functional, this pattern can lead to verbose and fragmented code, especially when multiple conditions are involved. The `if()` function aims to simplify this by allowing conditional logic to be expressed inline within a single declaration:
+
+```css
+.my-element {
+  color: if(media(min-width: 600px): red; else: blue);
+}
+```
+
+This approach improves readability and co-locates related logic, making stylesheets easier to maintain. This is especially powerful when combined with CSS Custom Properties, allowing for the creation of dynamic, themeable components with logic encapsulated directly in the CSS.
+
+## Syntax
+
+The `if()` function's syntax is formally defined as:
+
+```
+<<if()>> = if( [ <<if-branch>> ; ]* <<if-branch>> ;? )
+<if-branch> = <if-condition> : <declaration-value>?
+<if-condition> = <boolean-expr[ <if-test> ]> | else
+<if-test> =
+  supports( [ <ident> : <declaration-value> ] | <supports-condition> ) |
+  media( <media-feature> | <media-condition> ) |
+  style( <style-query> )
+```
+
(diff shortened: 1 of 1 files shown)


css-values-5/if-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+  background-color: if(media(width >= 600px): blue; else: green);
+}
+```
+
+### 2. Feature Support
+
+Conditionally apply a value based on whether the browser supports a particular CSS feature.
+
+```css
+.card {
```
andruud: It's not the best example; you can achieve the same with:

```
.card { display: block; display: grid; }
```

Probably something like what you have in the other example is better, e.g. "if display:grid is supported, then use some other font size".

css-values-5/if-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+Conditions can be combined with boolean keywords like `and`, `or`, and `not`. This allows for more complex logic, such as applying a style only when multiple conditions are met.
+
+```css
+.widget {
+  --base-size: 2rem;
+  --enhanced-size: 3rem;
+
+  /* Use the enhanced size only on wide screens that also support subgrid */
+  font-size: if(
+    media(width >= 1024px) and supports(display: subgrid): var(--enhanced-size);
```
andruud: Does `display:subgrid` actually exist?

css-values-5/if-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+    else: var(--base-size)
+  );
+}
+```
+
+## Fallback Behavior
+
+If no condition in an `if()` function evaluates to true and no `else` branch is provided, the function resolves to an empty token stream. This makes the declaration invalid at computed-value time, causing the property to fall back to its inherited or initial value (behaving like `unset`).
```
andruud: I don't think this is accurate? If no branch hits, then we return `<guaranteed-invalid>`, don't we? (Which is not the same as an empty token stream.)

svgeesus:
Is this an explainer for TAG review? If so you should read [Writing Effective Explainers](https://w3ctag.github.io/explainer-explainer/#introduction) where the TAG state what they are looking for in an explainer.

css-values-5/if-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+  supports( [ <ident> : <declaration-value> ] | <supports-condition> ) |
+  media( <media-feature> | <media-condition> ) |
+  style( <style-query> )
+```
+
+In essence, the function is a list of conditional branches separated by semicolons. Each branch contains a condition followed by a colon (`:`) and a value. The conditions are evaluated in order, and the value from the first branch with a true condition is used.
+
+The `else` keyword can be used as a condition that is always true. This makes it useful for providing a final fallback value, as any branches after an `else` branch will be ignored.
+
+## Core Use Cases
```
andruud: There is a lot of "fluff" here. Perhaps we can just drop it and let the examples speak for themselves.

The "killer feature" for `if()` is in my opinion being able to query _your own_ custom properties ("theming" per this section); perhaps we should highlight that in the motivation instead of inline MQs. Just a thought.

css-values-5/if-explainer.md
tursunova: replaced with coklore example, thanks

css-values-5/if-explainer.md
tursunova: I guess no 😅 Thanks, replaced with `display: flex`

css-values-5/if-explainer.md
tursunova: removed, thanks

css-values-5/if-explainer.md
tursunova: Replaced, thanks.

tursunova:
> Is this an explainer for TAG review? If so you should read [Writing Effective Explainers](https://w3ctag.github.io/explainer-explainer/#introduction) where the TAG state what they are looking for in an explainer.

Yes, it was initially, but now the TAG review is already closed: https://github.com/w3ctag/design-reviews/issues/1045. They left comments to add better use cases and transfer the explainer to the spec repository, which I tried to address here.

I went through the link you shared, I'm not sure if components 1 and 6 are relevant for `if()` now though. Could you please say if there is anything specific that needs to be fixed/changed here? Thank you.

andruud (approved):

Issue #11600, opened 2025-01-29
[css-overflow-5] Making ::scroll-marker existence unconditional?


tabatkins:
Currently, the existence of `::scroll-marker-group` is dependent on the scroll container having a non-`none` value for `scroll-marker-group. <https://drafts.csswg.org/css-overflow-5/#scroll-marker-group-pseudo> Then `::scroll-marker` existence is dependent on its corresponding `::scroll-marker-group` existence (+ markup information defining the element as being a "scroll marker").

This prevents us from making `::scroll-marker` a [":has-allowed pseudo-element"](https://drafts.csswg.org/selectors/#has-allowed-pseudo-element), since its existence is determined by style, and using `:has(::scroll-marker)` would allow cyclic styling loops. This, unfortunately, means it seems hard to define examples like this:

![Image](https://github.com/user-attachments/assets/47ec4c1d-dd51-4e4a-91b5-d3347d72ef01)

Note that the scroll-markers before the :target-current one are solid white; the current and subsequent ones are partially transparent. @argyleink tried to achieve this with:

```css
.carousel li:has(~ li::scroll-marker:target-current)::scroll-marker {
  opacity: 1;
}
```

But it doesn't work because `::scroll-marker` isn't marked as being valid in `:has()`.

-------

So, question: can we make the *existence* of `::scroll-marker` (and perhaps `::scroll-marker-group`) unconditional, based just on the markup information? We'd change the wording so the *box generation* is dependent on styling as currently specified, but the pseudo-element would be treated as *existing* regardless, so we could make `::scroll-marker` a ":has-allowed pseudo-element", and Adam's code would work.

argyleink:
Another use case?  
on hover if you want to show the relevant scroll element that a marker was created for:

```css
.carousel li:has(::scroll-marker:hover) {
  outline: 1px solid LinkText;
}
```

flackr:
The proposed resolution sounds good to me.

flackr:
> So the proposal is that we solve this case by instead specifying :target-past and :target-future. These match any scroll markers before or after the current scroll marker (or, if no scroll marker is current, before/after the break where a current marker could theoretically be.

This sounds good to me.

> However, this doesn't solve the "style the target element when the marker is hover" use-case Adam brought up afterwards.

[...]

PR #12662, opened 2025-08-27
[css-overflow-5] Add :target-before and :target-after pseudo classes #11600

danielsakhapov:
As resolved in https://github.com/w3c/csswg-drafts/issues/11600, `:target-before` and `:target-after` pseudo classes allow to select scroll markers before/after the `:target-current` one.

### DIFF
--- a/css-overflow-5/Overview.bs
+++ b/css-overflow-5/Overview.bs
@@ -366,21 +366,62 @@ The 'scroll-marker-group' property</h4>
 	or via the tab key when currently active or when no other ''::scroll-marker'' is active and this is the first marker in the group,
 	ensuring the group has a <a href="https://open-ui.org/components/focusgroup.explainer/#guaranteed-tab-stop">guaranteed tab stop</a>.
 
-<h4 id="active-scroll-marker">
-Selecting The Active Scroll Marker: the '':target-current'' pseudo-class</h4>
+<h4 id="active-before-after-scroll-markers">
+Selecting Scroll Markers: '':target-current'', '':target-before'' and '':target-after'' pseudo-classes</h4>
 
 	Exactly one [=scroll marker=] within each [=scroll marker group=] is determined to be active at a time.
 	Such "active" [=scroll markers=] match the <dfn selector>:target-current</dfn> pseudo-class.
 
-	<div class='example'>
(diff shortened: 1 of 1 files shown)


css-overflow-5/Overview.bs
```diff
@@ -458,6 +458,59 @@ Selecting The Active Scroll Marker: the '':target-current'' pseudo-class</h4>
 		1. 	Set the active state of all other [=scroll marker=] elements in |group| to false.
 	</div>
 
+<h4 id="target-before-after">
+Selecting Scroll Markers Before and After the Active Marker: the '':target-before'' and '':target-after'' pseudo-classes</h4>
```
flackr: I would expect that this would specified together with `:target-current`, either rolling this in to the above section or keeping the above section about how to calculate the active marker and specifying `:target-current` here as well, whichever you think makes more sense.

css-overflow-5/Overview.bs
```diff
@@ -458,6 +458,59 @@ Selecting The Active Scroll Marker: the '':target-current'' pseudo-class</h4>
...
+
+In addition to the <a>:target-current</a> pseudo-class, this specification introduces the <dfn selector>:target-before</dfn> and <dfn selector>:target-after</dfn> pseudo-classes for use with [=scroll marker=] elements.
+
+These pseudo-classes match [=scroll marker=]s that are, respectively, before or after the [=active marker=] (the one matching <a>:target-current</a>) within the same [=scroll marker group=], as determined by [=flat tree order=].
+
+- <b>:target-before</b> matches all [=scroll marker=]s that precede the [=active marker=] in the flat tree order within the group.
+- <b>:target-after</b> matches all [=scroll marker=]s that follow the [=active marker=] in the flat tree order within the group.
+
+If there is no [=active marker=] in the group, neither pseudo-class matches any [=scroll marker=] in that group.
```
flackr: How can a group have no active marker? The only way I can think of is if there are no markers - in which case no markers can match any of the pseudo classes.

danielsakhapov: Agree, just being cautious

css-overflow-5/Overview.bs
danielsakhapov: Moved them to :target-current

flackr (approved):

PR #12709, opened 2025-09-02
[css-viewport-1] Add defination for `<meta name="viewport">`.

shubhamg13:
Defines the `viewport <meta>` tag 

Fixes: #12710 [Issue 4](https://drafts.csswg.org/css-viewport/#issue-76355744)

### DIFF
--- a/css-viewport-1/Overview.bs
+++ b/css-viewport-1/Overview.bs
@@ -143,7 +143,18 @@ an horizontal display feature" style="width: 70%;height: auto; margin:auto; disp
 <h2 id="viewport-meta">
 Viewport <code class=html>&lt;meta&gt;</code> element</h2>
 
-Issue: Specify me
+A Viewport <a href="https://drafts.csswg.org/css-viewport/#viewport-meta">meta</a>
+element is a HTML element that allows sites to control the behaviour of
+browsers over various <a href="https://drafts.csswg.org/css-viewport/#meta-properties">aspects</a>
+of the viewport. This tag is primarily for mobile devices; most desktop browsers
+ignore it. This gives instruction to browser on how to control the page's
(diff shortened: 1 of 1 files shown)


shubhamg13:
@fantasai @tabatkins  Please take a look.

cc: @xiaochengh

tabatkins:
Apologies, but this doesn't actually meaningfully define the viewport meta; it's a non-normative intro. The point of that issue is that we need to actually define the parsing and parameters of the attribute, and how it maps to the functionality defined in this spec.

(Also, I don't recognize your GitHub username, are you a CSSWG member?)

shubhamg13:
> Apologies, but this doesn't actually meaningfully define the viewport meta; it's a non-normative intro. The point of that issue is that we need to actually define the parsing and parameters of the attribute, and how it maps to the functionality defined in this spec.

Yeah, I will try to properly tune it with all the points you mentioned above.  

> (Also, I don't recognize your GitHub username, are you a CSSWG member?)

Not yet, but I would like to be part of it.

tabatkins:
> Not yet, but I would like to be part of it.

Ah, well an issue then is that if you're not a CSSWG member, you can't submit "normative" text (anything that actually defines terms or behavior, etc). Text like that requires you to release patent rights, which WG members do automatically.

Is your company part of the W3C? If not, that's the usual first step. Alternately, individuals can become Invited Experts if their company isn't appropriate for W3C membership (if it doesn't have anything to do with the web, for instance), but we usually prefer people to interact with the WG first, to show that they understand the sort of work that needs to be done. You can do this by opening an issue to talk about how the definition should look.

I'm going to go ahead and close this PR for now, then.

fantasai:
We actually have the ability to take substantive contributions from non-members. But I have to agree with Tab that this PR, while it represents a reasonable introduction to the concept, isn't really a spec.

Issue #11332, opened 2024-12-06
[css-color] AccentColor in Forced Colors Mode


alisonmaher:
In #5900, we resolved that "AccentColor takes its value from accent-color, and AccentColorText takes its value from the appropriately contrasting foreground color for AccentColor; these resolve at computed value time".

In [Forced Colors Mode](https://drafts.csswg.org/css-color-adjust/#forced-colors-mode), for various color properties, "if its [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value) is a color other than a system color, its [used value](https://drafts.csswg.org/css-cascade-5/#used-value) is instead forced to a system color as follows".

Because `AccentColor` and `AccentColorText` are considered system colors, if these are used by the author, we wouldn't override those colors to another system color in Forced Colors Mode. As such, I am wondering if we should add a clarification to the above resolution that this does not apply in Forced Colors Mode so that we ensure that actual system color is exposed, as opposed to the author `accent-color` in this case?

svgeesus:
The CSS Working Group just [discussed](https://www.w3.org/2025/08/19-css-color-minutes.html#4872) `[css-color] AccentColor in Forced Colors Mode` and agreed to the following:

 * `RESOLUTION: we don't take author specified from accent-color property for forced color mode`

<details><summary>The full IRC log of that discussion</summary>
alisonmaher: we resolved previosuly about accentcolor system color takes it value from the accent-color property<br>
alisonmaher: in forced color mode, various color properties will used ???<br>
proposal is : we dont take author specified from accent-color property for forced color
mode<br>
RESOLUTION: we don't take author specified from accent-color property for forced color mode<br>
florian: are we ignoring it only as far as the keyword are concerned (accent-color / accent-color-text) ?<br>
florian: but the effect on things remain ?<br>
ChrisL: yes<br>
lea: is there any way to combine multiple options of forced color ?<br>
TabAtkins: no forced color force stuff thats the point<br>
florian: ok so in forced color, everything is using the system accent and the property is just use for its own computed value no effect whatsoever<br>
</details>

alisonmaher:
Updates added via https://github.com/w3c/csswg-drafts/pull/12733#event-19513127157

Issue #5900, opened 2021-01-27
[css-color] [css-ui-4] Authors should have access to accent-color value to use in their code


LeaVerou:
Edit: [Resolved](https://github.com/w3c/csswg-drafts/issues/5900#issuecomment-2380321760) to do this

---

https://drafts.csswg.org/css-ui-4/#widget-accent

As discussed in today's telcon, there should be a keyword (akin `currentColor`) or other way to refer to the value of `accent-color` so that authors can use this in their designs as well.
The advantage of this over a custom property, is decoupling: that libraries can hook into existing code using `accent-color` (especially given proposed guidance to set it on the root), without the code needing to be aware of the existence of the library.

bradkemper:
I agree, but I was thinking it should be an environment variable, e.g. `env(accent-color)`. 

Hopefully to do it in a way that doesn't allow using it for fingerprinting.

samuelbradshaw:
I can see the value in both referencing the author-defined accent color on a section of the page, and referencing the user agent's default accent color. Why not both? :)

SebastianZ:
If this should be made available via an environment variable, avoiding fingerprinting is an important point here. Access to `accent-color` probably needs to be restricted to avoid reading the actual value, otherwise the user agent, operating system or device could be determined when they use a unique color.

Though I guess that's something that applies to all environment variables.

Sebastian

LeaVerou:
@SebastianZ This is something set in CSS, so I'm not sure how fingerprinting applies?

SebastianZ:
Well, I assume @bradkemper's suggested environment variable will compute to an actual color value, which will be the color chosen by the user agent by default. So the environment variable needs to compute to the `auto` value or to itself and it needs to be ensured that there's no other way to get the actual color value via CSSOM or other means in order to avoid fingerprinting.

Sebastian

frivoal:
I guess it can't be an env variable: such variables are evaluated independently of where you are in the tree, and thus would always return the UA's default accent color, which we cannot share for fingerprinting reasons (see in https://github.com/w3c/csswg-drafts/issues/6158). If we want the local accent-color, as set by the author, we need something context sensitive, so that wouldn't be an env variable, but something more like the `currentColor` keyword.

That said, I wonder if this is really worth the cost, as there's an easy work-around. If an author needs this, they can just as well set their own `--accent-color` custom property, and use its value to set both the actual `accent-color` property and whatever other place they wanted to use that color in.

[...]

PR #12733, opened 2025-09-04
[css-color-4] Updated `AccentColor` to take its value from `accent-color`, unless Forced Colors Mode is enabled

alisonmaher:
These updates were resolved in both https://github.com/w3c/csswg-drafts/issues/5900 and https://github.com/w3c/csswg-drafts/issues/11332#issuecomment-3200774167

The change to system color text can be found hosted at https://alisonmaher.github.io/alisonmaher/css-color-4/Overview.html#css-system-colors (at the end of the section just before example 15).

For some reason, the second mention of "accent color" does not show as linked, even though it uses the same syntax as the first reference (perhaps a bikeshed issue?)

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -41,6 +41,8 @@ spec: css-backgrounds-3; type: property; text: border-top-color
 spec: css-backgrounds-3; type: property; text: border-left-color
 spec: css-backgrounds-3; type: property; text: border-bottom-color
 spec: css-backgrounds-3; type: property; text: border-right-color
+spec: css-ui-4;	type: dfn; text: accent color
+spec: css-ui-4; type: property; text: accent-color
 </pre>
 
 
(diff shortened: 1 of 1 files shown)


svgeesus:
> For some reason, the second mention of "accent color" does not show as linked, even though it uses the same syntax as the first reference (perhaps a bikeshed issue?)

As intended, Bikeshed only auto-links the first definition in a paragraph, if there are multiple identical ones.

css-color-4/Overview.bs
```diff
@@ -7285,8 +7294,8 @@ Changes</h2>
 <h3 id="changes-from-20250424">Changes since the 
 	<a href="https://www.w3.org/TR/2025/CRD-css-color-4-20250424/">Candidate Recommendation Draft of 24 April 2025</a></h3>
 <ul>
-	<!-- to 23 Apr 2025 -->
-	<li><em>none</em></li>
+	<!-- to 4 Sep 2025 -->
```
svgeesus, line 7297: Probably that isn't all the changes to 4 Sept, but I can fix that up afterwards

PR #12734, opened 2025-09-04
[mediaqueries-4][mediaqueries-5][editorial] Added WPTs

SebastianZ:
I added all the Web Platform Tests that currently cover the features in Media Queries 4 and 5.

In level 5 I've also added all the tests that are in level 4. Is that ok or should the tests only be in in one spec. i.e. 4?

Sebastian

### DIFF
--- a/mediaqueries-4/Overview.bs
+++ b/mediaqueries-4/Overview.bs
@@ -21,6 +21,8 @@ Abstract:
 Abstract: Media Queries Level 4 describes the mechanism and syntax of media queries, media types, and media features. It extends and supersedes the features defined in Media Queries Level 3.
 Ignored Terms: min-resolution, max-resolution, none, view-mode, mediaText, CSSOMString
 Link Defaults: css-break-3 (property) break-inside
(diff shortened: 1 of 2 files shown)


svgeesus:
Tests for other levels need to still be listed (otherwise bikeshed complains) but, for delta specs, put them in &lt;wpt hidden>. You can add a title, like

```html
<wpt hidden title="Level 4 tests">
foo.html
bar.html
</wpt>

SebastianZ:
> Tests for other levels need to still be listed (otherwise bikeshed complains) but, for delta specs, put them in <wpt hidden>. You can add a title, like

Thank you for the hint, @svgeesus! In this case, the spec. already has everything from level 4 in it, i.e. it is _not_ a delta spec. anymore. The question was rather whether they should still go into a `<wpt hidden>` section even when it is _not_ a delta spec. or if they should be in both like in the patch right now.

Sebastian

svgeesus:
Oh I see, sorry I should have checked. In that case, no need to hide them.

svgeesus (approved):

PR #12738, opened 2025-09-05
[css-backgrounds-4] Allowed combining `border-area` and `text` in `background-clip`

### LINKED ISSUE #10696: [css-backgrounds-4] Can you chain `border-area` and `text` in `background-clip`?
#9456 added `background-clip: border-area`.

I think it would be useful to support `background-clip: border-area text` in case you want your background to bleed through both the text and the border-area of an element.

The new syntax would then be:

```
<bg-clip> = <visual-box> | [ border-area || text ]
```

SebastianZ:
This adjusts the syntax of `background-clip` to accept a combination of `border-area` and `text`, as it was resolved on in https://github.com/w3c/csswg-drafts/issues/10696#issuecomment-2461103750.

Closes #10696

Sebastian

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -326,7 +326,7 @@ Painting Area: the 'background-clip' property</h3>
 	The syntax of the property is given with
 
 	<pre class=prod>
-	<dfn>&lt;bg-clip></dfn> = <<visual-box>> | border-area| text
+	<dfn>&lt;bg-clip></dfn> = <<visual-box>> | [ border-area || text ]
 	</pre>
 
 	Issue: Or should this be defining the <css>-webkit-background-clip</css> property,


css-backgrounds-4/Overview.bs
```diff
@@ -326,7 +326,7 @@ Painting Area: the 'background-clip' property</h3>
 	The syntax of the property is given with
 
 	<pre class=prod>
-	<dfn>&lt;bg-clip></dfn> = <<visual-box>> | border-area| text
+	<dfn>&lt;bg-clip></dfn> = <<visual-box>> | [ border-area || text ]
```
Loirooriol, line 329: The spec needs to explain what happens when 2 values are provided. Is the background clipped to the intersection? Union?

SebastianZ: Thanks for the hint! See https://github.com/w3c/csswg-drafts/pull/12739.

PR #12745, opened 2025-09-07
[css-inline-3][editorial] <text-edge> requires both values

cdoublev:
Fixes an oversight in 764a39b: both keywords are mandatory since this commit.

  > `<text-edge> = [ text | ideographic | ideographic-ink ] | [ text | ideographic | ideographic-ink | cap | ex ] [ text | ideographic | ideographic-ink | alphabetic ]`


cdoublev:
I misread the specification and resolution.

PR #12747, opened 2025-09-08
[css-cascade-6] <scope-start> and <scope-end> produce <complex-real-selector-list>

cdoublev:
This PR allows `w3c/webref` to extract syntax rules for both, encoding the exclusion of pseudo-elements. It also clarifies the type they produce.


cdoublev:
`<scope-end>` might need to be defined as a `<relative-real-selector-list>`  or something (#13290), so I convert this PR to a draft.

Issue #8721, opened 2023-04-14
[css-overflow-4] Unexpected behavior of `continue: discard` on multicolumn containers with spanners


andreubotella:
The definition of `continue: discard` in [css-overflow-4, section 5.3](https://w3c.github.io/csswg-drafts/css-overflow-4/#continue) (not the one in appendix A) says that, for a multicolumn container, overflow columns are discarded, which is to say they are not laid out or painted.

Consider the following case, though:

```css
.multicol {
  columns: 3;
  continue: discard;
}
.multicol p {
  break-after: column;
  break-inside: avoid;
}
.spanner {
  column-span: all;
}
```

```html
<div class="multicol">
  <p>Column 1</p>
  <p>Column 2</p>
  <p>Column 3</p>
  <p>Overflow column</p>
  <div class="spanner">Spanner</div>
</div>
```

In the case of block containers, or of region breaks in a multicolumn container, `continue: discard` will discard the remainder of the container, which can be done in implementations by stopping layout at the end of a fragment. However, with overflow columns before a spanner, the layout must resume at the spanner.

My understanding is that the layout at the spanner, and at any further multicol lines after it, is independent from the layout the overflow columns would have, so there is no actual contradiction. I think this could be implemented correctly by having multicol line boxes in the box tree that would end before a spanner, such that only the remainder of the line box would be discarded:

```
MulticolBox
  MulticolLineBox
    BlockBox ("Column 1")
    BlockBox ("Column 2")
    BlockBox ("Column 3")
    BlockBox ("Overflow column")
  MulticolSpanner
  MulticolLineBox
    ...
```

But this would be unexpected behavior all the same, since in terms of the DOM tree there would be a range that is skipped by layout. And Chromium's LayoutNG implementation of multicol, at the very least, doesn't use multicol line boxes – it instead treats spanners as forcing a break in the multicol line.

cc @frivoal @bfgeek

frivoal:
Interesting question.

> the layout must resume at the spanner.

I am not sure. The spanner is taken out of flow, and so you could put it back in the multicol, but it's static position is in the overflow columns, which are discarded, so maybe it should be discarded too.

If I understand you correctly, I agree:
* From a layout point of view, I think it would be doable to not discard it, and just place it below the content of the first multi-col line, whose height is only affected by the content in the non discarded columns. That seems fine.

[...]

PR #12785, opened 2025-09-12
[css-overflow] drop overflow columns and subsequent content with discard

frivoal:
See https://github.com/w3c/csswg-drafts/issues/8721


css-overflow-4/Overview.bs
```diff
@@ -1070,6 +1068,91 @@ Fragmentation of Overflow: the 'continue' property</h3>
 			(such as pagination of this box itself)
 			do not cause any content to be discarded.
 
+			Additionally,
+			if the box is a [=multi-column container=],
+			any content in or after [=overflow columns=] is also not rendered,
```
andreubotella: Is saying that the content is not rendered, rather than the columns / multicol lines, significant? This wording could imply that `column-rule`s between overflow columns, or in a following multicol line, *are* rendered.

PR #12788, opened 2025-09-12
[cssom-1][editorial] Added WPTs

SebastianZ:
I added all the Web Platform Tests that currently cover the features in CSSOM 1.

Note that I am not totally sure about the tests that intersect with other specs (see "Tested in other specs"). They actually cover features defined in other specs, though they are currently placed in css/cssom/. So they _could_ be added to CSSOM as well.
@emilio You are the author of some of those tests. What do you think?

Sebastian


svgeesus:
The 'tested in other specs' tests should be added to those other specs, if they are not already there (and would need a path attribute on &lt;wpt> because they are not on the default path.

I don't see a particular need to hide them here, though. It is fine for a given test to show up in multiple specs.

PR #12798, opened 2025-09-15
[css-text-decor-3] Disallow negative text-shadow blur

Calme1709:
The first two length values for `text-shadow` (x and y offsets) are allowed to be negative but the third value (blur) is not.

This behaviour is respected in all major browsers and is tested in WPT, see the second test [here](http://wpt.live/css/css-text-decor/text-shadow/parsing/text-shadow-invalid.html)

### DIFF
--- a/css-text-decor-3/Overview.bs
+++ b/css-text-decor-3/Overview.bs
@@ -861,7 +861,7 @@ Text Shadows: the 'text-shadow' property</h2>
 
 	<pre class="propdef">
 	Name: text-shadow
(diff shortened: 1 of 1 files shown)


Loirooriol:
`box-shadow` and L4 `text-shadow` are fine, since [`<shadow>`](https://drafts.csswg.org/css-backgrounds-3/#typedef-shadow) already uses `<length [0,∞]>`.
So indeed it's just L3 `text-shadow` that is wrong.

Issue #12801, opened 2025-09-15
[css-break-3] Should broken boxes always be extended to fill the gap?


Psychpsyo:
[CSS Break 3](https://www.w3.org/TR/css-break-3/) says, in [5.3. Splitting Boxes](https://www.w3.org/TR/css-break-3/#box-splitting), that when a box gets split, it's block size is effectively increased. This leads to its backgrounds and borders being painted into the remaining empty space in that fragmentainer.

This usually seems desirable, but it gets ugly when a table cell with the `rowspan` attribute gets broken due to the rows that it spans falling into different fragmentainers. In this case, the empty space that comes to exist between those rows has nothing in it, apart from the `rowspan`-cell, which I'd expect to only render next to the rows that it spans.

I think there should either be an exception for these cells, or a CSS property that can control this behavior, similar to `box-decoration-break`. Alternatively, `box-decoration-break` could be extended to accept something like `[ slice | clone ] || skip-gap` instead of just `slice | clone`. This new behavior should, in my opinion, be applied to rowspan cells by default, so this could be added to the UA stylesheet:
```css
td[rowspan], th[rowspan] {
    box-decoration-break: skip-gap;
}
```
The UA stylesheet addition would be a breaking change, though currently only Safari manages to properly break these cells at all. Both Chrome and Firefox fail to lay them out correctly in different ways. (Chrome doesn't extend the height on the rowspanned cell so it doesn't reach the end of the second row and Firefox draws the background only inside the first fragmentainer and only next to the first row.)

TL;DR:
<img width="1310" height="653" alt="Image" src="https://github.com/user-attachments/assets/789bbcb1-69ab-4781-872d-0378335b3d6f" />

fantasai:
I think what you're saying is that if we're breaking between rows, then we should actually break the (visual) box of a rowspanned cell at that point, too (if it's breakable at that point). This makes sense to me. Probably the same should apply for Grid. I don't think we're at the point where we can require this behavior, but we can suggest it.

Psychpsyo:
> I think what you're saying is that if we're breaking between rows, then we should actually break the (visual) box of a rowspanned cell at that point, too (if it's breakable at that point). This makes sense to me. Probably the same should apply for Grid. I don't think we're at the point where we can require this behavior, but we can suggest it.

Yes, pretty much. Though I think "breaking the (visual) box" of something is not yet a concept that exists in the spec.
[...]

PR #12805, opened 2025-09-15
[css-break-3] Add the concept of grouping boxes

Psychpsyo:
This improves layout in tables and grid containers which contain cells that span across multiple rows which may be broken into separate fragmentainers.

Here is an example of what behavior this aims to improve:
489477628-789bbcb1-69ab-4781-872d-0378335b3d6f

This is not meant to be a requirement for UAs, but rather just a suggestion. I hope the use of "should" instead of "must" has that power. (if not, that can be changed)

More context: https://github.com/w3c/csswg-drafts/issues/12801

I'd be happy to sign the IPR thingy, but I can't seem to do that on my own on the website. (I'm not a member of the working group, nor do I work for a member organization)

### DIFF
--- a/css-break-3/Overview.bs
+++ b/css-break-3/Overview.bs
@@ -1003,6 +1003,30 @@ Splitting Boxes</h3>
     </p>
   </div>
 
+<h4 id="grouping-boxes">
+Grouping Boxes</h4>
+
+  <p>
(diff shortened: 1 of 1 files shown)


fantasai:
I think this is a simpler way of saying the same thing? Or am I missing something? https://github.com/w3c/csswg-drafts/pull/13810

Issue #9326, opened 2023-09-07
[css-grid-3][masonry] variable track sizes + dense packing has poor performance


fantasai:
From @bfgeek in https://github.com/w3c/csswg-drafts/issues/9041#issuecomment-1710806191
> Variable track sizes is problematic with items that span multiple tracks with "dense" packing. To correctly place the item in the correct tracks, you need O(N) layout passes on the item.

tabatkins:
It also interacts badly if you want to allow "auto-spanning", where an element will get a span based on its width. This might be a good mode to allow, possibly even with some author control - if an image is >X% of the track width, make it span 2 instead of 1 (or a larger span, as appropriate).

With variable-width tracks this is *at minimum* problematic for layout, and might be inherently circular (I haven't given it enough thought yet to be certain).

fantasai:
@tabatkins I think auto-spanning is a separate issue. :) And an equally applicable request for Grid.

tabatkins:
Nah, it's part of this issue - *if* we do auto-spanning (and I think Masonry libraries do sometimes have this feature), then variable-size tracks are even more problematic. (And it would be equally problematic for Grid, I think.)

fantasai:
@tabatkins I don't see what that has to do with dense packing. Filed as https://github.com/w3c/csswg-drafts/issues/10089

fantasai:
Back to the original issue...

Agreed that having N different track sizes means dense packing would require O(N) layout passes on items added after a gap, and this would be a problem. We suggest adopting a heuristic that brings this closer to O(2) for most cases:

1. Lay out the item in the widest track, and in the narrowest track, and find the shortest height among the two.
2. Skip trying to fit into any gaps that are shorter than that height.

It's possible, but unlikely, that an item's actual shortest height would be at a middling width; and this allows us to avoid a lot of that extra work in the much more common cases where the item definitely won't fit.

fantasai:
Ok, discussing with @stwrt, new proposal for dense-packing in masonry:
1. Lay out the item in the track where it would be without dense packing.
2. If the item is short enough to fit into an earlier gap _in a track with the same used size as this track_, move it into that gap.

This avoids doing more than one layout, and it works as expected for explicitly-placed masonry in mixed-size tracks as well as for auto-placed masonry in consistently-sized tracks, and does a reasonable job of dense-packing large collections of auto-placed masonry in mixed-size tracks as well.

astearns:
[...]

PR #12807, opened 2025-09-16
[css-grid-3][masonry] Define dense packing for masonry #9326

fantasai:


css-grid-3/Overview.bs
```diff
@@ -873,6 +887,18 @@ Masonry Layout and Placement Algorithm</h3>
 			and then layout the item.
 			Set the [=running position=] of the spanned  [=grid axis=] tracks
 			to <code><var>max_pos</var> + [=outer size=] + 'grid-gap'</code>.
+
+		<li>
+			If the [=masonry container=] uses ''item-pack/dense'' packing,
+			and there exists skipped empty space (e.g. due to spanning items)
```
alisonmaher, line 893: nit: I'd remove this "and" since there is another one later in the list

css-grid-3/Overview.bs
```diff
@@ -873,6 +887,18 @@ Masonry Layout and Placement Algorithm</h3>
...
+			earlier in the layout into which the item, as it is sized now,
+			could have fit without impacting other items’ placement
+			if it were earlier in the placement order,
+			and the tracks it would have spanned at that point
+			have the same total used size as the tracks into which it is currently placed,
+			move the item into that position
+			and rewind the [=auto-placement cursor=] and the [=running position=]
+			to their values before this item’s placement.
```
alisonmaher, line 901: One open question I have is the order in which we break ties with openings of the same start offset. I presume it is using the auto-placement cursor, but do we start from the earlier openings in the current track, or the next track from the one we are currently starting in.

css-grid-3/Overview.bs
fantasai: The items in this list are a bit complex, and themselves include commas, so I felt it was easier to understand the sentence with the extra emphasis from the explicit "and" here.

css-grid-3/Overview.bs
fantasai: It follows the same rules as normal placement... in other words, we move the item into the earliest placement order that would not change the placement of items after it. Maybe I should clarify this text with that concept...

css-grid-3/Overview.bs
alisonmaher: I see, yeah that sounds good to me then. Resolving

css-grid-3/Overview.bs
fantasai: Ok, tweaked the wording a bit. I'll merge this, and if you have suggestions on how to clarify more we can try for improvements. :)

alisonmaher: I can't tell what text changed since I think commits got squashed maybe, but interesting. I'm curious what "earliest in placement order" means - the order in which items get placed, or the earliest in terms of offset? Given that items can be explicitly placed, the former would be tough to track.

I would have guessed we follow the placement cursor to some degree (likely starting from the track the item is originally placed in), and then loop back around to find the opening with the earliest start offset. If there is a tie, you would choose the one closest to the original auto placement cursor.

Or are you saying we should always start looking for track openings at track 0 (i.e. prefer track openings in earlier tracks if there is a tie?)

tabatkins: Yeah, we don't have an in-built ordering such that we can just say "earliest". Individual items are ordered, but their gaps aren't, unlike Grid.

Also, I think this shouldn't change the ordering of things if there *isn't* a hole somewhere; that is, if your only valid choices are the ends of columns, we should continue to pack as normal, following the placement cursor.

So I recommend saying that, *if* there are holes that don't correspond to the existing non-dense placements, we choose the "highest" one, tie-breaking with simple layout direction order. No attention paid to or affect on the cursor when this happens.

fantasai: We do have a placement order, so what this is saying is if there's a gap earlier than the current location, then you place it as if it were in the earliest possible position where it would satisfy these constraints.

If you have a clearer way to express that, sure, but I don't want to repeat the placement rules here.

alisonmaher: > So I recommend saying that, if there are holes that don't correspond to the existing non-dense placements, we choose the "highest" one, tie-breaking with simple layout direction order. No attention paid to or affect on the cursor when this happens.

This makes sense to me and I think is worth updating to clarify since in Chromium we just implemented it based on the placement cursor position for tie breaking, and the current text doesn't really make it clearer to me that the auto placement cursor is ignored when choosing the potential openings

tabatkins: I don't think we have a well-defined "placement order" that includes gaps, no. We have a set of possible placements within the threshold distance of the highest one, and cursor+placement direction breaks ties when that set is more than one. This isn't using that set, tho, or even modifying it - it *first* finds a standard placement as normal, *then* looks for gaps with the same track size and decides if it wants to move to one of those instead.

(@fantasai and I chatted offline) Ah, you're talking about "what hypothetical rearranging of this item in the list of children would have put it in that hole; choose the position with the earliest index". This implies replaying the placement algorithm to some extent, and at minimum requires recording the state of the auto-placement cursor after each placement; this hypothetical rearrangement could, I think, have had knock-on effects as well on the placement of other items.

Instead, I just want to choose the highest valid hole, and if there are multiple valid holes within the tie threshold of the highest, choose the start-most one. In almost all cases this'll be the same as your intended position, but it's stateless and requires no memory of how placement was performed originally. All you need to maintain is the size and position of the holes you form as placement occurs. (I've given you some proposed text in our shared doc.)

PR #12819, opened 2025-09-17
[css-view-transitions-2] Create two-phase-transition-explainer.md

noamr:
SSIA

### DIFF
--- a/css-view-transitions-2/two-phase-transition-explainer.md
+++ b/css-view-transitions-2/two-phase-transition-explainer.md
@@ -1,0 +1,54 @@
+# Two-phase view transition
+Making cross-document view-transitions feel instant and seamless.
+
+See w3c/csswg-drafts#10616
+
+# Current state
+View-transitions don’t start immediately - rendering is paused until the new state is ready. 
+This is especially visible in cross-document view transitions, where the transition only captures the old when the new document’s response arrives, and starts when the new document is ready to render.
+
+This can create a jarring or abrupt experience.
+
+# Current workarounds
+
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/two-phase-transition-explainer.md
```diff
@@ -0,0 +1,54 @@
...
+
+## In a nutshell
+Proposing to prototype a “two-phase” view-transition: 
+Instantly start a transition to a state that can be computed quickly enough or synchronously. Call this a “preview” state.
+When the to-preview transition ends, transition from there to the final state.
+Only interrupt the first transition after a timeout, otherwise stall the navigation commit until finished.
+This should ideally not delay the LCP / loading experience of the new page, as the content keeps loading (and potentially prerendering) in the background.
+
+## Phase 1: script-invoked preview
+We can create this kind of seamless/instant experience without any new CSS, and potentially without needing to fully spec it normatively by changing the behavior as follows:
```
vmpstr, line 34: I'm worried that we might be blocked by some unrelated view transition as a result of this (especially some long running one). I wonder if we can explicitly mark one as a preview transition via types or something

vmpstr (approved):

css-view-transitions-2/two-phase-transition-explainer.md
noamr: > I'm worried that we might be blocked by some unrelated view transition as a result of this (especially some long running one). I wonder if we can explicitly mark one as a preview transition via types or something

Sure or call navigateEvent.startViewTransition or so

PR #12825, opened 2025-09-18
[css-overflow-4] Clarify (independent) formatting contexts and ellipsis insertion

frivoal:
The sentence as it stood was unclear, but this seems to be the only reasonable thing it can mean.

### DIFF
--- a/css-overflow-4/Overview.bs
+++ b/css-overflow-4/Overview.bs
@@ -729,11 +729,12 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
 		block-ellipsis-repaint-002.html
 	</wpt>
 
-	This property only affects a line box
-	if it immediately precedes either a [=region break=]
-	or a [=clamp point=] in the [=block formatting context=],
-	ignoring any intervening [=absolutely positioned=] element
-	or any element closing boundary.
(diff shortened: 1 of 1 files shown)


css-overflow-4/Overview.bs
```diff
@@ -729,11 +729,12 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
...
-	if it immediately precedes either a [=region break=]
-	or a [=clamp point=] in the [=block formatting context=],
-	ignoring any intervening [=absolutely positioned=] element
-	or any element closing boundary.
+	Ignoring any intervening [=absolutely positioned=] element
+	or any element closing boundary,
+	this property only affects a line box
+	if it immediately precedes either a [=region break=] or a [=clamp point=],
+	and the line box is in the same [=independent formatting context=]
+	as the [=clamp point=] or the [=region break=].
```
fantasai, line 737:
```suggestion
	Ignoring any intervening [=absolutely positioned=] elements
	or element-closing boundaries,
	this property affects a [=line box=]
	that immediately precedes a [=region break=] or [=clamp point=]
   when both participate in the same [=formatting context=].
```

andreubotella: I incorporated this change into #13525.

PR #12835, opened 2025-09-19
[cssom-view] Fix scrollend event on document.

emilio:
As per the comments in #11164.


emilio:
Actually it's just a trivial typo, so merging :)

PR #12844, opened 2025-09-22
[css-borders-4][editorial] Fix syntax rewrite mistake

cdoublev:
The syntax of `corner-top-left` and other "single corner" shorthands is:

  > Value: 	`<'border-top-left-radius'>{1,2} || <'corner-top-left-shape'>`

`{1,2}` is an oversight, because the syntax of `border-top-left-radius` is:

  > Value: `<length-percentage [0,∞]>{1,2}`

Before 5b61925, he syntax of `corner-top-left` and other "single corner" shorthands was:

  > Value: `<length-percentage [0,&infin;]>{1,2} || <corner-shape-value>`

My apologies.


noamr (commented):
LGTM

PR #12847, opened 2025-09-23
[css-forms-1] Clarify accessibility principle

### LINKED ISSUE #12480: [css-forms-1] WCAG conformance design principle
This issue is to discuss potential clarifications to the "The controls pass 100% of [WCAG 2.2 AA standards](https://www.w3.org/TR/WCAG22/)." design principle of CSS Forms base appearance mode.

This is strongly worded and while a principle rather than a claim I think it could be good to clarify the scope of what is and isn't included in that. For example, the title attribute tooltip that shows up has a number of WCAG issues (that it avoids by virtue of the fact its browser provided).

Pickers are also largely not included in the spec but the statement might give a false sense that this sovles all those issues.

Potentially we should simply say something more generic that the controls should try to account for accessibility requirements where possible.

Raising as its something that's come up on socials (e.g. https://toot.cafe/@aardrian/114851791346613604).

lukewarlow:
cc @aardrian

Fixes #12480 

Happy to tweak as appropriate, I've incorporated the feedback on the issue but I'm not wedded to any particular wording. Also apologies for the delay on getting to this.

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -62,7 +62,7 @@ spec:css-forms-1; type:value; for:/; text:::placeholder
 
   1. The styles are identical in every user agent.
   1. The controls are recognizable and usable on their own without additional styles.
-  1. The controls are accessible out of the box in default configurations, ideally passing <a href="https://www.w3.org/TR/WCAG22/">WCAG 2.2 AA standards</a>.
+  1. The controls target accessiblity out of the box, ideally passing <a href="https://www.w3.org/TR/WCAG22/">WCAG 2.2 AA standards</a>.
 
   NOTE: Authors may need to do extra adjustments to ensure their controls are accessible depending on any customizations as well as the context the form control is placed into.
   The basic appearance does not prevent adjustments by the author that are inaccessible.


css-forms-1/Overview.bs
```diff
@@ -62,7 +62,11 @@ spec:css-forms-1; type:value; for:/; text:::placeholder
 
   1. The styles are identical in every user agent.
   1. The controls are recognizable and usable on their own without additional styles.
-  1. The controls pass 100% of <a href="https://www.w3.org/TR/WCAG22/">WCAG 2.2 AA standards</a>.
+  1. The controls are accessible out of the box in default configurations, ideally passing <a href="https://www.w3.org/TR/WCAG22/">WCAG 2.2 AA standards</a>.
```
lukewarlow: "are accessible" - This makes me feel slightly uneasy, as it's kind of meaningless.

Perhaps something like this instead?

```suggestion
  1. The controls target accessiblity out of the box, ideally passing <a href="https://www.w3.org/TR/WCAG22/">WCAG 2.2 AA standards</a>.
```

PR #12850, opened 2025-09-24
[css-position-4] Add "afterRemoval" steps to top layer removal algorithms

keithamus:
Refs https://github.com/whatwg/html/issues/11694.

The spec for implicit anchors is slightly wrong, as the timing of when an implicit anchor can be removed is awkward. 

Implicit anchors for elements in the top layer need to be removed _after_ transitions, otherwise the anchor positioning may cause reflow. 

Chrome "cheats" and skips reflow if the implicit anchor is being removed, but this can cause glitches if the anchored element reflows during this transition, as it will appear to move around the page (losing its implicit anchoring) when reflow is triggered.

Rather than doing this, I propose we add a "list of steps |afterRemoval|" to the "request an element to be removed from the top layer (immediately)" steps. This way we can use this as a callback to determine when the implicit anchor association should be removed.


css-position-4/Overview.bs
```diff
@@ -579,7 +579,7 @@ Top Layer Manipulation {#top-manip}
 
 <div algorithm>
 	To <dfn export lt="request an element to be removed from the top layer | request removal from the top layer">request an element to be removed from the top layer</dfn>,
-	given an {{Element}} |el|:
+	given an {{Element}} |el| and a list of steps |afterRemoval|:
```
lukewarlow: Should we make these optional? We don't need them for dialogs for example?

css-position-4/Overview.bs
```diff
@@ -353,10 +355,13 @@ Top Layer Manipulation {#top-manip}
...
 
-	1. For each element |el| in |doc|'s [=pending top layer removals=]:
-		if |el|&apos;s computed value of 'overlay' is ''none'',
-		or |el| is [=not rendered=],
-		[=remove from the top layer immediately=] |el|.
+	1. For each (element |el| and steps |afterRemoval|) in |doc|'s
+		[=pending top layer removals=]: if |el|&apos;s computed value of
+		'overlay' is ''none'', or |el| is [=not rendered=]:
+
+		1. [=remove from the top layer immediately=] |el|.
```
lukewarlow, line 359: This should just pass after removal through as they're called in that algorithm.

css-position-4/Overview.bs
```diff
@@ -353,10 +355,13 @@ Top Layer Manipulation {#top-manip}
...
+
+		2. Run |afterRemoval|.
```
lukewarlow, line 361: Drop this.

css-position-4/Overview.bs
```diff
@@ -316,7 +316,7 @@ Top Layer Manipulation {#top-manip}
 
 <div algorithm>
 	To <dfn export lt="request an element to be removed from the top layer | request removal from the top layer">request an element to be removed from the top layer</dfn>,
-	given an {{Element}} |el|:
+	given an {{Element}} |el| and an optional ist of steps |afterRemoval|:
```
jakearchibald:
```suggestion
	given an {{Element}} |el| and an optional list of steps |afterRemoval|:
```

css-position-4/Overview.bs
```diff
@@ -341,6 +341,8 @@ Top Layer Manipulation {#top-manip}
 	3. Remove the UA !important ''overlay: auto'' rule targeting |el|,
 		if it exists.
 
+	4. Run |afterRemoval|.
+
 	Note: This algorithm is only intended to be used in special cases
```
jakearchibald, line 343: Given this is synchronous, I don't think this needs a callback.

css-position-4/Overview.bs
jakearchibald: It feels like this should stay, and the callback to `remove from the top layer immediately` should be removed. It seems weird to have a callback for something that happens at the end of a sync algorithm.

css-position-4/Overview.bs
keithamus: Good point.

css-position-4/Overview.bs
lukewarlow: Yeah I guess we can drop it and then call afterRemoval inside of the process steps. Or we can drop the call from there and just pass the afterRemoval through to here?

I think its slightly neater (especially on the HTML spec side) to just call it here?

keithamus: It does now as the other algorithms rely on it.

css-position-4/Overview.bs
```diff
@@ -327,12 +327,12 @@ Top Layer Manipulation {#top-manip}
...
 	3. Remove the UA !important ''overlay: auto'' rule targeting |el|.
 
-	4. [=set/Append=] |el| to |doc|'s [=pending top layer removals=].
+	4. [=set/Append=] (|el|, |afterRemoval|) to |doc|'s [=pending top layer removals=].
 </div>
 
 <div algorithm>
 	To <dfn export lt="remove an element from the top layer immediately | remove from the top layer immediately">remove an element from the top layer immediately</dfn>,
-	given an {{Element}} |el|:
+	given an {{Element}} |el| and an optional list of steps |afterRemoval|:
```
jakearchibald: Why would specs use this rather than just call `remove from the top layer immediately` then do whatever they want?

keithamus: Probably because a lot of specs will make a condition of whether or not they remove immediately or not, with a likely list of cleanup steps, so it would be something like:

```
1. If shouldRemoveImmediately
  1. remove an element from the top layer immediately given el
  2. Run doTheThing
1. Otherwise, remove an element from the top layer given el and doTheThing
```

With this provision we allow the following instead:

```
1. If shouldRemoveImmediately, remove an element from the top
   layer immediately given el and doTheThing, otherwise, remove
   an element from the top layer given el and doTheThing
```

Having said that, I could go either way.

**Edit** changed.

jakearchibald (approved):

jakearchibald:
This should get approval from someone else before it lands 😄

lukewarlow:
Non-editorial LGTM. This should wait till the HTML PR is ready and get merged together probably?

Not sure who is the spec editor for this CSS spec but they should check this PR too.

keithamus:
I have requested review from both spec editors.

css-position-4/Overview.bs
josepharhar: Should this be changed to be a list of element with steps instead of just a set of elements?

keithamus:
@tabatkins and/or @fantasai did you want to take a look at this? Should go towards addressing https://github.com/w3c/csswg-drafts/issues/13595 & https://github.com/whatwg/html/issues/11694.

fantasai:
@tabatkins and I looked this over, and we aren't sure this is ready to merge? It looks like some of the review comments haven't been addressed, and it also doesn't look the issue discussion had a real conclusion either.

josepharhar (approved):

PR #12855, opened 2025-09-25
[css-env-1] Correct note to use pem not em

lukewarlow:
[css-env-1] Correct note to use pem not em

### DIFF
--- a/css-env-1/Overview.bs
+++ b/css-env-1/Overview.bs
@@ -253,7 +253,7 @@ would cause text sizes to double,
 then ''env(preferred-text-scale)'' would resolve to ''2''.
 
 Note: The ''pem'' unit represents this same information;
-''1em'' is exactly equivalent to ''calc(1em * env(preferred-text-scale))''.
+''1pem'' is exactly equivalent to ''calc(1em * env(preferred-text-scale))''.
 When directly sizing things, ''bsem'' is just a more convenient length to use.
 
 <div class=example>


css-env-1/Overview.bs
```diff
@@ -253,7 +253,7 @@ would cause text sizes to double,
 then ''env(preferred-text-scale)'' would resolve to ''2''.
 
 Note: The ''pem'' unit represents this same information;
-''1em'' is exactly equivalent to ''calc(1em * env(preferred-text-scale))''.
+''1pem'' is exactly equivalent to ''calc(1em * env(preferred-text-scale))''.
```
lukewarlow, line 256: I think this is a simple typo right? Unless I'm massively misunderstanding.

css-env-1/Overview.bs
```diff
@@ -253,7 +253,7 @@ would cause text sizes to double,
...
 When directly sizing things, ''bsem'' is just a more convenient length to use.
```
lukewarlow, line 257: Is bsem defined anywhere? I can't seem to find it?

css-env-1/Overview.bs
svgeesus: Yeah this is a simple typo or copy-pasta

css-env-1/Overview.bs
svgeesus: I don't see bsem in https://drafts.csswg.org/indexes/#types

svgeesus (approved):

css-env-1/Overview.bs
tabatkins: "bsem" is short for `bikeshed-em`, it's a placeholder while we decided on the name.

PR #12856, opened 2025-09-25
[css-borders-4][editorial] Incorporated rest of CSS Backgrounds 3 text

SebastianZ:
This is an initial import of the full CSS Backgrounds 3 text related to borders and box decorations.

I've left out the note about the spec. _not_ defining how borders of different styles should be joined in the corner.

It also adds the WPTs from CSS Backgrounds 3. Please let me know if I did this right!

I've mostly left the formatting as in the other spec.

Also, some of the syntax definitions probably should be reworked. Though I thought that's out of the scope of this initial import.

Sebastian


css-borders-4/Overview.bs
```diff
@@ -372,23 +929,112 @@ The 'border-top-radius', 'border-right-radius',
...
+		<pre>
+		border-top-left-radius:     2em 0.5em;
+		border-top-right-radius:    1em 3em;
+		border-bottom-right-radius: 4em 0.5em;
+		border-bottom-left-radius:  1em 3em;
+		</pre>
+	</div>
+
+<h3 id="corner-shaping">
 Corner Shaping: the 'corner-*-shape' properties</h3>
```
svgeesus, line 1038: I don't think autolink will work here :)

svgeesus (approved):
Looks good overall

css-borders-4/Overview.bs
```diff
@@ -141,6 +174,19 @@ Line Colors: the 'border-color' properties</h3>
 	If only one value is given,
 	it applies to both the <a>start</a> and <a>end</a> sides.
 
+	<wpt pathprefix="css/css-backgrounds/">
```
svgeesus: I think that should be /css/css-backgrounds otherwise it is relative

SebastianZ: I made it absolute, though there is a wild mix of relative and absolute paths throughout the specs. And note that you also [used relative paths](https://github.com/w3c/csswg-drafts/blame/8ee7a05f87123313c9241f67449aef9b72b707ce/css-color-4/Overview.bs#L4822) for some tests.

css-borders-4/Overview.bs
SebastianZ: Right. I've added it together with similar terms to the ignored terms.

SebastianZ:
Thanks for the approval, @svgeesus! I've merged this now after fixing the two remarks you had.

Sebastian

Issue #12479, opened 2025-07-14
[css-values-5] Clarify how `<attr-unit>` should be parsed


AtkinsSJ:
The spec for this [says](https://drafts.csswg.org/css-values-5/#ref-for-typedef-attr-type%E2%91%A0):

> If given as an [`<attr-unit>`](https://drafts.csswg.org/css-values-5/#typedef-attr-unit) value, the value is first parsed as if `type(<number>)` was specified, then the resulting numeric value is turned into a dimension with the corresponding unit, or a percentage if % was given. Values that fail to parse as a `<number>` trigger fallback.

`<number>` can mean a literal number, but it can also be a math function, or various other functions that produce `<number>`. Chromium added a [crash test](http://wpt.live/css/css-values/crashtests/chrome-405422528-crash.html) which checks that a non-literal appearing doesn't break the browser:

```html
<!DOCTYPE html>
<link rel="help" href="[https://crbug.com/405422528](view-source:https://crbug.com/405422528)">
<div data-crash="sibling-index()" style="width: attr(data-crash px)"></div>
<div data-crash="sign(1em - 1px)" style="width: attr(data-crash px)"></div>
```

But their [fix](https://github.com/chromium/chromium/blob/1b76fc9f583b8fa1be1fc2dc13d61e205bc6f977/third_party/blink/renderer/core/css/css_attr_type.cc#L85-L95) assumes that only `<number>` literals are allowed.

The spec is ambiguous here. Only allowing literals is one interpretation (and certainly the easiest to implement!) but the other interpretation (which I read the spec as meaning) is that this:

```html
<div data-crash="sign(1em - 1px)" style="width: attr(data-crash px)"></div>
```

...should produce a `width` value equivalent to `calc(sign(1em - 1px) * 1px)`.

cc: @lilles

fantasai:
I suspect we might want to allow the author to use either interpretation somehow.

cdoublev:
I think [parsing](https://drafts.csswg.org/css-syntax-3/#css-parse-something-according-to-a-css-grammar) `sibling-index()` against `<number>` with a context associated to an element should succeed, as well as for `sign(1em - 1px)` regardless of the context, but not against `<number-token>`. I do not see why an author would want to disallow functions resolving to `<number>` though.

yisibl:
Some users have already noticed this issue and reported it as a bug in Chrome.
https://issues.chromium.org/issues/441267435

PR #12862, opened 2025-09-26
Add number keyword for attr() and specify how <attr-unit> should be parsed

tursunova:
Resolved in https://github.com/w3c/csswg-drafts/issues/12479#issuecomment-3329561248

### DIFF
--- a/css-values-5/Overview.bs
+++ b/css-values-5/Overview.bs
@@ -1930,8 +1930,9 @@ Ian's proposal:
 		trigger fallback.
 
 		If given as the <css>number</css> keyword,
-		it causes the attribute's literal value
-		to be matched against <<number-token>> production.
+		it causes the attribute's literal value,
+		after [=strip leading and trailing ASCII whitespace|stripping leading and trailing whitespace=],
+		to be [=CSS/parsed=] as a <<number-token>>.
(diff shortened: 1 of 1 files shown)


css-values-5/Overview.bs
```diff
@@ -1929,12 +1929,19 @@ Ian's proposal:
 		Values that fail to parse according to the syntax
 		trigger fallback.
 
+		If given as the <css>number</css> keyword,
+		it causes the attribute's literal value
+		to be matched against <<number-token>> production.
```
tabatkins:
```suggestion
		If given as the <css>number</css> keyword,
		it causes the attribute's literal value,
		after [=strip leading and trailing ASCII whitespace|stripping leading and trailing whitespace=],
		to be [=CSS/parsed=] as a <<number-token>>.
```

css-values-5/Overview.bs
```diff
@@ -1929,12 +1929,19 @@ Ian's proposal:
...
+		Attributes literal values that do not corresponds to the <<number-token>> production
+		trigger fallback.
```
tabatkins:
```suggestion
		Values that fail to parse trigger fallback.
```

tabatkins (approved):

PR #12865, opened 2025-09-27
Correct CSS Cascading and Inheritance Level 5 specification name in 2…

BartolomeSintes:
….2 section

The actual name of the specification is "CSS Cascading and Inheritance Level 5", not "CSS Cascading and Inheritance Module Level 5".

[css-spec-shortname-1] Brief description which should also include the #issuenum-or-URL and/or link to relevant CSSWG minutes.

Copy the above line into the Title and replace with the relevant details. Fill in any additional details here. See https://github.com/w3c/csswg-drafts/blob/master/CONTRIBUTING.md for more info.

PR #12866, opened 2025-09-27
[css-cascade-4][css-cascade-5][css-cascade-6][editorial] Added "Module" to spec title

SebastianZ:
I believe, the titles of the different specs should be consistent regarding their naming. And https://github.com/w3c/csswg-drafts/pull/12865 gave me the incentive to start with the CSS Cascade specs.

CSS Cascade 3 is already a recommendation, so I thought I won't touch that one, even when it's also missing the word "Module" in its title.

Also, I'm not sure whether changing the title of a spec is actually only editorial. Please let me know whether that requires a formal resolution!

Sebastian

### DIFF
--- a/css-cascade-4/Overview.bs
+++ b/css-cascade-4/Overview.bs
@@ -1,5 +1,5 @@
 <pre class='metadata'>
-Title: CSS Cascading and Inheritance Level 4
+Title: CSS Cascading and Inheritance Module Level 4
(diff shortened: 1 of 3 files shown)


svgeesus (approved):

svgeesus:
It is just a consistency fix, it seems editorial to me.

PR #12872, opened 2025-09-29
[css-overflow-4] Refactor the text about block-ellipsis insertion

frivoal:
The text had accreted various rules and constraints, but no longer had clear structure.

This attempt to reorganize the whole thing in a more understandable manner.

No new behavior is (intended to be) introduced by this reformulated text, though some previously implicit consequences are made explicit, for clarity.

The PR is split into two commits, for better readability: [one](802f5cb185632c9006318406a97263eb19d9de03) reorganizing the text, and [one](f3cdfcb354f9e31624f8d7a5238c8c7d3b843787) re-attributing the relevant test annotations to the relevant part (but making no change to the text).

cc: @andreubotella @fantasai

### DIFF
--- a/css-overflow-4/Overview.bs
+++ b/css-overflow-4/Overview.bs
@@ -835,23 +835,32 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
 		that would still allow the entire <a>block overflow ellipsis</a> to fit on the line,
 		applying [[css-text-4#white-space-rules]] including [[css-text-4#white-space-phase-2]]
 		before insertion.
-		For this purpose, <a>soft wrap opportunities</a> added by 'overflow-wrap' are ignored,
-		as are those inhibited by ''text-wrap-mode: nowrap''.
-		If this results in the entire contents of the line box being displaced,
-		the line box is considered to contain a [=strut=], as defined in [[CSS2/visudet#leading]].
 
 		<wpt>
 			block-ellipsis-001.html
+			block-ellipsis-028.html
+			webkit-line-clamp-025.html
+		</wpt>
+
+		For this purpose, <a>soft wrap opportunities</a> added by 'overflow-wrap' are ignored,
+		as are those inhibited by ''text-wrap-mode: nowrap''.
+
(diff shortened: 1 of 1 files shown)


css-overflow-4/Overview.bs
```diff
@@ -800,55 +800,84 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
...
+		<wpt>
+			block-ellipsis-002.html
+			block-ellipsis-003.html
+			block-ellipsis-004.html
+			block-ellipsis-005.html
+			block-ellipsis-006.html
+			block-ellipsis-010.html
+		</wpt>
+
+	* The user agent makes room as necessary of the [=block overflow ellipsis=]
```
andreubotella, line 832: It'd be better to split this paragraph. That would also help with the WPT annotations.

css-overflow-4/Overview.bs
```diff
@@ -800,55 +800,84 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
...
+			block-ellipsis-027.html
+			block-ellipsis-028.html
+			webkit-line-clamp-025.html
+		</wpt>
+
+	* The anonymous inline of [=block overflow ellipsis=] is placed
+		after any remaining content,
+		after [[css-text-4#white-space-phase-2]],
+		as a direct child of the <a>block container</a>’s <a>root inline box</a>,
+		with the appropriate bidi embedding level
```
andreubotella:
```suggestion
		with the appropriate bidi embedding levels
```

With `block-ellipsis: <string>`, you can have multiple embedding levels, and even isolates.

andreubotella (approved):

css-overflow-4/Overview.bs
frivoal: You mean split the second bullet in several paragraphs? Yes, we could do that. Would you suggest 3 paragraphs  (1 per sentence)?

andreubotella: Yeah, that would work.

Issue #10686, opened 2024-08-04
[cssom-2] Start CSSOM 2


SebastianZ:
The CSS Object Model spec. is [in Working Draft status since 2011](https://www.w3.org/TR/2011/WD-cssom-20110712/) and many new features were added and changes were made over the years.

To let this specification finally progress to CR, I suggest to make a cut regarding new features. Anything extending the interfaces defined there should be added to a new Level 2 spec.
Also, some features currently defined there that require more discussion and/or a proper definition should be deferred to level 2. *Which* of them to defer needs to be discussed.

Given, this spec. is based on the [Document Object Model (DOM) Level 2 Style Specification](https://www.w3.org/TR/2000/REC-DOM-Level-2-Style-20001113/), which is already a Recommendation, I even put up for discussion whether the spec. should be turned into a Level 3 spec. instead (and a Level 4 spec. created for everything new).

Sebastian

SebastianZ:
@therealglazou @emilio

SebastianZ:
Adding this to the agenda.

Proposed resolutions:
* Feature-freeze CSSOM 1 and start the next level. Move newer extensions that still need more discussion to the new level.
* Change the level of CSSOM 1 to 3.

Sebastian

SebastianZ:
> <ydaniv> SebastianZ: I could add WPT coverage, to see what's stable and what's not

Just for the record, this was incorrectly minuted. The WPT coverage is already up-to-date.

Sebastian

PR #12879, opened 2025-09-30
[many][editorial] Added "Module" and "Level" to all specs

SebastianZ:
To unify the specification titles, they now all include the word "Module" and also explicitly the mention their level.
See also https://github.com/w3c/csswg-drafts/pull/12866.

I also added "Module Level 1" to CSSOM 1 and CSSOM View 1. See also https://github.com/w3c/csswg-drafts/issues/10686.

Sebastian

### DIFF
--- a/mediaqueries-4/Overview.bs
+++ b/mediaqueries-4/Overview.bs
@@ -1,5 +1,5 @@
 <pre class='metadata'>
-Title: Media Queries Module Level 4
+Title: Media Queries Level 4
 Group: csswg
 Shortname: mediaqueries
 Level: 4
--- a/mediaqueries-5/Overview.bs
+++ b/mediaqueries-5/Overview.bs
@@ -1,5 +1,5 @@
 <pre class='metadata'>
-Title: Media Queries Module Level 5
+Title: Media Queries Level 5
 Group: csswg
 Shortname: mediaqueries
 Level: 5
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -1,5 +1,5 @@
 <pre class='metadata'>
-Title: Selectors Module Level 4
+Title: Selectors Level 4
 Group: CSSWG
 Shortname: selectors
 Level: 4
--- a/selectors-5/Overview.bs
+++ b/selectors-5/Overview.bs
@@ -1,5 +1,5 @@
 <pre class='metadata'>
-Title: Selectors Module Level 5
+Title: Selectors Level 5
 Group: CSSWG
 Shortname: selectors
 Level: 5


frivoal:
I'm ok with the addition of the Level on every spec.

As for the word "Module", it works in most cases, but there are a few where I'm not sure:
* Media Queries & Selectors: while the primary application of these specs is CSS, they are usable in other contexts as well. Is it "you can used Media Queries with CSS as well as with `Window.matchMedia()`, or is it "You can also use the CSS ’Media Queries’ Module with `Window.matchMedia()`"? Same with Selectors. I've mostly tought of it so far in the first way, though we could make it work the second way too I suppose.
* cssom-view : I can't quite put a finger on what's wrong, but adding the word module feels strange. Then again, the name of that spec in general feels pretty strange to me, so ¯\\\_(ツ)\_/¯

svgeesus:
I agree with @frivoal that MQ and Selectors are special; they are intended to be used in multiple places of which CSS is but one, so I would avoid "Module" for those two.

SebastianZ:
@frivoal @svgeesus Ok, I've reverted the titles for the Selectors and Media Queries specs. Though I want to mention that if we applied that reasoning consequently, we'd have to _remove_ "CSS" and "Module" from many other specs as well, like CSSOM, CSSOM View, CSS View Transitions 1, CSS Highlight API 1, all the Houdini specs, and more.

Sebastian

SebastianZ:
And just to be clear, you are saying that the word "Module" shouldn't be added to those specs, because they can be used outside of CSS? Though allowing the features to be used in other contexts as well or even requiring them, isn't that rather a reason to remove _"CSS"_ from the specs? "Module" to me is only an indicator that it's part of something bigger.

Sebastian

frivoal:
"The CSS XYZ Module" means it's a part of CSS. Selectors and Media Queries are not called CSS Selectors or CSS Media Queries, so you can't add module to make them part of a whole, since we haven't specified what whole.

And the lack of CSS in the name is deliberate, as they're conceptualized not as a part of CSS that can be used with other things too, but as a separate thing on top of which CSS is built. A normative dependency of CSS, not a part of it. Being maintained by the CSS WG, the distinction is somewhat blurry, but that's how I've understood it.

SebastianZ:
The counter-argument is that Media Queries 3 and Selectors 3 are part of the [official definition of CSS](https://drafts.csswg.org/css-2025/#css-official) for many years now.

Sebastian

svgeesus:
Yes, CSS uses them.

Not having "CSS" in the titles was a deliberate choice some decades ago, however.

SebastianZ:
@frivoal @svgeesus Ok, so with the changes to the Selectors and Media Queries specs reverted, what do you think?

Sebastian

SebastianZ:
Given @svgeesus's 👍🏻, I went ahead and merged this.

Sebastian

PR #12880, opened 2025-09-30
[css-values-4] Export "style resource base URL" definition.

dbaron:
This exports the "style resource base URL" definition so that it can be used in the definition of a urlpattern() function.


dbaron:
(I got a verbal ok from @tabatkins on this.)

PR #12892, opened 2025-10-01
[css-color-adjust-1] Expand on how forced colors mode handles emojis

alisonmaher:
The team in Chromium making the updates to forced colors mode emoji handling noted that it would be helpful to specify that `font-variant-emoji` gets overridden to `text` to ensure that the monochrome variant is chosen if available. This change adds more text to specify the exact details.

I wasn't sure if a new resolution was needed for this given the resolution in https://github.com/w3c/csswg-drafts/issues/8064#issuecomment-2163389692 but more than happy to file a new issue as needed.

### DIFF
--- a/css-color-adjust-1/Overview.bs
+++ b/css-color-adjust-1/Overview.bs
@@ -890,7 +890,7 @@ Properties Affected by Forced Colors Mode {#forced-colors-properties}
 	* 'accent-color' computes to ''accent-color/auto''
 	* If 'font-variant-emoji' computes to ''font-variant-emoji/normal'' or ''font-variant-emoji/unicode'',
 		UAs should force any emoji on the page to its monochrome variant, if available,
-		by forcing the [=used value=] of 'font-variant-emoji' to ''font-variant-emoji/text''.
+		by forcing the [=computed value=] of 'font-variant-emoji' to ''font-variant-emoji/text''.
 
 	UAs may further tweak these <a>forced colors mode</a> heuristics
 	to provide better user experience.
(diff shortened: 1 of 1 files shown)


fantasai (approved):
Seems like an editorial clarification.

Does bring up a few questions though:
- Should this be a used-value change or a computed-value change?
- What happens if an emoji has a variation selector that indicates color emoji? If I'm reading the spec correctly here https://www.w3.org/TR/css-fonts-4/#propdef-font-variant-emoji `font-variant-emoji` doesn't have an effect in this case...? But it's a bit unclear from the spec. CC @svgeesus

alisonmaher:
> Should this be a used-value change or a computed-value change?

@fantasai this is a good question and one that I was going back and forth on. I chose used value time since we check against the computed value of `font-variant-emoji` before overriding, but that can also be done at computed value time.

And I suppose computed value time is more consistent with all of the other non-color based properties, so perhaps that is a better choice. If others agree, I can make that update

svgeesus:
> * What happens if an emoji has a variation selector that indicates color emoji? If I'm reading the spec correctly here https://www.w3.org/TR/css-fonts-4/#propdef-font-variant-emoji `font-variant-emoji` doesn't have an effect in this case...? But it's a bit unclear from the spec.

I believe that is correct, but may take a closer reading of the spec and perhaps making assumptions about what "appended" means exactly in code point order. My reading is that the "as if appended" puts it last, thus over-riding any actual VS in the text.

Suggestions for simpler but clearer wording welcome, of course (but need not hold up merging this PR).

> And I suppose computed value time is more consistent with all of the other non-color based properties, so perhaps that is a better choice. If others agree, I can make that update

I think it is a more consistent choice, yes.

alisonmaher:
> >And I suppose computed value time is more consistent with all of the other non-color based properties, so perhaps that is a better choice. If others agree, I can make that update
> 
> I think it is a more consistent choice, yes.

Sounds good, updated to computed value time for consistency

PR #12897, opened 2025-10-02
[css-borders-4] [css-shapes-1] Specify rendering of single-path border-shape

### LINKED ISSUE #11662: [css-borders-4] Interaction of  single-path `border-shape` with non-uniform `border-width`
> One proposal for `border-width` interaction. 
> 
> Break it down in two steps.
> 
> Step 1. Assume a `width(t)` function that returns the width of the border parametrized over its length (`t=0` begin of the path,, `t=1` end of a continuous path). For each path, the `width(t)` function can be applied to create an outer shape by moving the path outside in the direction of the normal of the path.
> 
> Step 2. How to build `width(t)`. One option is to allow the user to specify a function or 1d gradient directly for each continuous path. But let's ignore that for now. Can we build a reasonable function given `border-width`? I think so. We can define "control points" (similar to what we have today): `0 - 0.25 - 0.5 - 0.75 - 1.0`. Each interval gets one width. And then we specify a transition range: so, something like: `width(t) = border-top-width [0.05, 0.2], lerp(top, right) [0.2, 0.3], right [0.3, 0.45] ... etc`. We can decide how much control/flexibility to give. This function would be applied to every continuous path on the shape.
> 
> Would something like this work?
>  

 _Originally posted by @fserb in [#6997](https://github.com/w3c/csswg-drafts/issues/6997#issuecomment-2622601997)_

noamr:
border-shape takes its color (and thickness if single-path) from one of the border sides, namely the first one (in logical order) that has a border-style other than none.

Also define half-border-box as per resolution, in css-shapes-1.

Resolution: https://github.com/w3c/csswg-drafts/issues/11662#issuecomment-3357273733
Closes #11662

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -1634,7 +1634,7 @@ as described in <a href="https://drafts.csswg.org/css-backgrounds-3/#corner-clip
 Issue: how should this affect clipping replaced elements?
 
 <div algorithm="choose-relevant-side-for-border-shape">
-To choose the <dfn>relevant side for border shape</dfn> given an [=element=] |element|:
+An element's <dfn>relevant side for border shape</dfn> is the first side (in the order [=block-start=], [=inline-start=], [=block-end=], and [=inline-end]) that has a non-''border-style/none'' [=border style=], or [=block-start=] if they're all ''border-style/none''.
 	1. If |element|'s [=computed value|computed=] 'border-block-start-style' is not ''border-style/none'', then return [=block-start=].
 	1. If |element|'s [=computed value|computed=] 'border-inline-start-style' is not ''border-style/none'', then return [=inline-start=].
 	1. If |element|'s [=computed value|computed=] 'border-block-end-style' is not ''border-style/none'', then return [=block-end=].


css-borders-4/Overview.bs
```diff
@@ -1594,23 +1594,33 @@ The 'border-shape' property</h3>
...
 		Animation type: by computed value
 	</pre>
 
 The 'border-shape' property is provided with either a single <<basic-shape>> or two <<basic-shape>>s,
 resulting in one or two paths, respectively.
-The single-path border shape varint uses the existing ''border'' properties of the element to stroke the
-path that the given <<basic-shape>> resolves to,
-while the double-path border shape variant fills the area between the two paths as if it were the border.
+
+When two <<basic-shape>> values are given, the border is rendered between the paths generated shapes.
```
SebastianZ, line 1605: I think this small change makes the sentence more readable.

```suggestion
When two <<basic-shape>> values are given, the border is rendered as the shape between the two paths.
```

css-borders-4/Overview.bs
```diff
@@ -1623,11 +1633,15 @@ as described in <a href="https://drafts.csswg.org/css-backgrounds-3/#corner-clip
 
 Issue: how should this affect clipping replaced elements?
 
-The fill and stroke color, as well as the stroke width, are taken from the respective 'border-color' and 'border-width' properties.
-
-Issue: define this in detail. Perhaps it should be overridable somehow?
+<div algorithm="choose-relevant-side-for-border-shape">
+To choose the <dfn>relevant side for border shape</dfn> given an [=element=] |element|:
```
tabatkins: Can simplify this significantly to:

```suggestion
An element's <dfn>relevant side for border shape</dfn> is the first side (in the order [=block-start=], [=inline-start=], [=block-end=], and [=inline-end]) that has a non-''border-style/none'' [=border style=], or [=block-start=] if they're all ''border-style/none''.
```

SebastianZ:
Sorry, little late. I just had one remark regarding the wording.

Sebastian

css-borders-4/Overview.bs
noamr: Fixed

Issue #7422, opened 2022-06-27
[selectors] Add pseudo-classes for `<select>` being a drop-down box vs a list box


zcorpan:
See https://github.com/whatwg/html/issues/7050

Chromium [supports](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/resources/html.css;l=909?q=html.css&ss=chromium) a `:-internal-list-box` pseudo-class in the UA stylesheet, which matches [`<select>` elements that render as a list box instead of as a dropdown box](https://html.spec.whatwg.org/multipage/rendering.html#the-select-element-2).

Gecko similarly [supports](https://searchfox.org/mozilla-central/source/layout/style/res/forms.css#249) `:-moz-select-list-box` in the UA stylesheet.

WebKit [seems to use attribute selectors instead](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/css/html.css#L1060), although I think it's not entirely accurate since [the decision per spec](https://html.spec.whatwg.org/multipage/rendering.html#list-box) depends on the [parsing result of the `size` attribute](https://html.spec.whatwg.org/multipage/form-elements.html#concept-select-size). (Demo: `<select size="+02xyz">`)

Web developers could rely on the `multiple` attribute being present to target list boxes, but that might not always work so well, in particular if you're writing CSS that should work for arbitrary HTML.

I think it makes sense to add a pseudo-class that matches `select` elements that are list boxes, e.g. `:select-list-box`. To match drop-downs, you can do `select:not(:select-list-box)`.

Spec: https://drafts.csswg.org/selectors/

zcorpan:
Is there interest in standardizing this pseudo-class (and make it also available to author CSS)?

cc @mfreed7 

Via git blame of WebKit's html.css I found https://bugs.webkit.org/show_bug.cgi?id=220376 -- I think a pseudo-class as suggested here would be less prone to bugs like this (compared to attribute selectors). cc @pxlcoder

Via git blame of Gecko's html.css I found https://bugzilla.mozilla.org/show_bug.cgi?id=1643246 where attribute selectors were replaced with a pseudo-class. cc @emilio @heycam

mfreed7:
It certainly seems tricky to select this based on attribute selectors, so I'm supportive of trying to standardize a pseudo class.

[...]

Issue #12510, opened 2025-07-21
[css-forms-1] UA styles for base appearance listbox select elements


josepharhar:
We already decided on UA styles for base appearance drop-down single-select elements here: https://github.com/w3c/csswg-drafts/issues/10857

I'd like to decide on UA styles for listbox select elements, with or without the multiple attribute. Most of the styles from drop-down single-selects can be reused, but here are some changes I'd like to propose. Until https://github.com/w3c/csswg-drafts/issues/7422 gets resolved, I'm going to write these with a `:has-picker` pseudo-class.

Make these style which currently apply to the select element only apply to drop-down selects:
```css
select:has-picker {
  border-radius: 0.5em;
  padding-block: 0.25em;
  padding-inline: 0.5em;
  min-block-size: calc-size(auto, max(size, 24px, 1lh));
  min-inline-size: calc-size(auto, max(size, 24px));
  display: inline-flex;
  gap: 0.5em;
  field-sizing: content !important;
  box-sizing: border-box;
}
```

Add these styles which only apply to listbox selects:
```css
select:not(:has-picker) {
  display: inline-block;
  overflow-inline: auto;
  overflow-block: auto;
}
```

Change the selectors of the hover/active/disabled rules to only apply when the select has a picker:
```css
select:has-picker:enabled:hover {
  ...
}
select:has-picker:enabled:active {
  ...
}
select:has-picker:disabled {
  ...
}
```

josepharhar:
Here is a PR which is probably easier to read than the style rules I put in the OP: https://github.com/w3c/csswg-drafts/pull/12511

josepharhar:
Instead of comparing the styles to select with a picker, maybe just listing the styles which will apply when its a listbox is easier to understand:
```css
select:not(:has-picker) {
  overflow-inline: auto;
  overflow-block: auto;
  border: 1px solid currentColor;
  display: inline-block;
  user-select: none;
}
```

josepharhar:
comments from meeting:
- no border
- no intrinsic sizing, not sure what the field-sizing value should be to represent this
- box-sizing: border-box
- maybe something else for display value?

josepharhar:
In order to incorporate the last comment's suggestions, here are my new proposed styles for list box selects:
```css
select {
  overflow-inline: auto;
  overflow-block: auto;
  field-sizing: content;
  display: inline-block;
  box-sizing: border-box;
  user-select: none;
}
```
[...]

PR #12908, opened 2025-10-07
[css-forms] Add select list box styles

lukewarlow:
Based on discussion within https://github.com/w3c/csswg-drafts/issues/12510

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -867,14 +867,12 @@ input[type=radio]:checked::checkmark {
     width: 100%;
 }
 ```
-## Selects & buttons ## {#stylesheet-select-button}
+
+## Buttons (and button-like controls) ## {#stylesheet-button}
+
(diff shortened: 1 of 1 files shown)


lukewarlow:
I still think we should resolve https://github.com/w3c/csswg-drafts/issues/7422 - so these can be written much more neatly. But for now spitting it up like Joey has done in the HTML PR feels like the best approach.

nt1m (approved):

css-forms-1/Overview.bs
```diff
@@ -1034,6 +1047,18 @@ select::picker-icon {
...
 ```
 
+### List box selects ### {#stylesheet-listbox-select}
+
+These styles should apply when the select is a [list box](https://html.spec.whatwg.org/multipage/rendering.html#list-box).
+
+```css
+select {
+	overflow: auto;
+  display: inline-block;
```
josepharhar, line 1057: overflow has a tab character but display has two spaces

lukewarlow: Oops. I'll fix that in a follow-up at some point.

PR #12923, opened 2025-10-09
[cssom-view-1] Fix indentation of "run the scroll" steps

tidoust:
The markdown assumed that a 2-space indent was enough to create substeps, but four spaces are needed and the structure of the algorithm was wrong in the generated document as a result.


tidoust:
Superseded by #13238.

PR #12957, opened 2025-10-16
[css-view-transitions-2] Resolve finished promise at own task when animations are done

### LINKED ISSUE #12442: [css-view-transitions-1] finished promise timing
In [css-view-transitions-1's Handle Transition Frame](https://www.w3.org/TR/css-view-transitions-1/#handle-transition-frame-algorithm), step 4 says

> 1. Set transition’s [phase](https://www.w3.org/TR/css-view-transitions-1/#viewtransition-phase) to "done".
> 2. [Clear view transition](https://www.w3.org/TR/css-view-transitions-1/#clear-view-transition) transition.
> 3. [Resolve](https://webidl.spec.whatwg.org/#resolve) transition’s [finished promise](https://www.w3.org/TR/css-view-transitions-1/#viewtransition-finished-promise).
> 4. Return.

Since we resolve the promise in 3, the earliest opportunity for script to run is at the next microtask checkpoint. However, these steps are executed via [perform pending transition operations](https://www.w3.org/TR/css-view-transitions-1/#ref-for-handle-transition-frame) which is invoked in the [Update the Rendering loop](https://html.spec.whatwg.org/#event-loop-processing-model:perform-pending-transition-operations), which doesn't have a microtask checkpoint before producing a visual frame.

This means that we are guaranteed to present one visual frame _after_ view transition is ended but _before_ script had a chance to run.

In some situations, this causes a flicker if we're transferring some style from the group pseudo to the final destination element. Here's an example where the box shadow is applied to the group pseudo and transferred to the final element when the transition finishes: https://vmpstr.github.io/htmldemos/vt/flicker.html (you might have to click a few times, but the flicker is definitely there in both Chrome and Safari). The workaround toggle listens to the first Animation object's finish promise (which resolves and runs outside of the rendering loop) to do the same work, and the flicker disappears.

My proposal to avoid this flicker is to move step 4 of Handle Transition Frame to run after the Update the Rendering step. Something along the lines of marking the transition as "pending done", and scheduling a task to finalize the transition. Finalizing the transition would then confirm that the state is "pending done", and run the four steps of current step 4 of Handle Transition Frame

/cc @nt1m @emilio @noamr

noamr:
Closes #12442

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -1788,6 +1788,7 @@ Issue: Is this updated by a <code>rootEl.startViewTransition()</code> call, too?
 			1. "`pending-capture`".
 			1. "`update-callback-called`".
 			1. "`animating`".
+			1. "`pending-done`".
 			1. "`done`".
 
 			Note: For the most part, a developer using this API does not need to worry about the different phases, since they progress automatically.


noamr:
@vmpstr ping for review

Issue #11213, opened 2024-11-14
[css-conditional-5][css-overflow-5] Valid size query containers and box tree re-ordering


lilles:
In the spec: [https://drafts.csswg.org/css-conditional-5/ref-for-query-container①⑧](https://drafts.csswg.org/css-conditional-5/#ref-for-query-container%E2%91%A0%E2%91%A6), valid query containers are said to be selected from the element's flat tree ancestors.

We have encountered an issue when the box tree order does not follow the flat tree order. The box for `::scroll-marker-group` becomes a sibling of its originating element. While normally pseudo elements can query the size of their originating element, this causes an issue with layout escaping the container causing circularity.

Can we say something generally about box tree re-ordering and size query container candidates?

For the `::scroll-marker-group` case we know that if it generates a box, it will be a sibling of its originating element's box. For future cases, if the box tree order depends on the element's style, we might not know at the time we query the container where the element's box will end up.

If we know about the box re-ordering at query time we could either:

1. Not consider ancestors which will not be box tree ancestors as valid size query containers
2. Always evaluate size queries to unknown for such containers

PR #12960, opened 2025-10-16
[css-conditional-5] Size query containers and box tree re-ordering #11213

lilles:
Per resolution:

https://github.com/w3c/csswg-drafts/issues/11213#issuecomment-3407344774


css-conditional-5/Overview.bs
```diff
@@ -399,6 +399,21 @@ Container Queries</h2>
...
+	sibling	box of its originating element's box. If we allowed querying the size
+	of a sibling box, it would introduce layout cycles.
+
+	<div class=note>
+		The ''::scroll-marker-group'' and ''::scroll-button()'' pseudo-elements
+		generate sibling boxes of their originating element's box.
+		These pseudo-elements will not be able to query their originating scroller
+		for [=container size queries=]. They will, however, be able to query other
+		eligible [=query container=] ancestors.
+	</div>
```
tabatkins:
```suggestion

	<div class=note>
		There are cases where the box for a [=pseudo-element=] is generated as a
		sibling box of its originating element's box. If we allowed querying the size
		of a sibling box, it would introduce layout cycles.
		
		For example, the ''::scroll-marker-group'' and ''::scroll-button()'' pseudo-elements
		generate sibling boxes of their originating element's box.
		These pseudo-elements will not be able to query their originating scroller
		for [=container size queries=]. They will, however, be able to query other
		eligible [=query container=] ancestors.
	</div>
```

PR #12968, opened 2025-10-17
[css-view-transitions-2] Specify view-transition-nane: match-element

vmpstr:
[css-view-transitions-2] Specify view-transition-nane: match-element

Issue #10995

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -537,6 +537,27 @@ spec:css2; type:dfn; text:viewport
 		: <dfn>none</dfn>
 		:: The [=/element=] will not participate independently in a view transition.
 
+		: <dfn>match-element</dfn>
+		:: The [=/element=] will participate in the transition
+			with a user-agent generated name that is unique to the [=/element=].
+
+			<div class="example">
+				```html
+				<style>
+					.card {
+						view-transition-name: match-element;
+					}
+				</style>
+				<div id=e1 class=card>Card 1</div>
+				<div id=e2 class=card>Card 2</div>
+				```
+				In this example, both `#e1` and `#e2` participate in the transition and match against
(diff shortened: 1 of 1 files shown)


noamr:
I am pretty sure I did this already in https://github.com/w3c/csswg-drafts/pull/11393

Was it omitted when merging L1 and L2?

bramus:
> Was it omitted when merging L1 and L2?

Sure looks like it. Because there are partial mentions of match-element in the current drafts (namely that match-element is not accepted as one of the custom-idents)

bramus:
I dug in a bit more and I figured out: there is [a section on auto naming with `auto`/`match-element`](https://drafts.csswg.org/css-view-transitions-2/#auto-vt-name) in the spec, but the problem is that [the definition of the `view-transition-name` prop](https://drafts.csswg.org/css-view-transitions-2/#view-transition-name-prop) does not mention these keywords are possible values.

> ```bs
> 	<pre class=propdef>
> 	Name: view-transition-name
> 	Value: none | <<custom-ident>>
> 	Initial: none
> 	Inherited: no
> 	Percentages: n/a
> 	Computed Value: as specified
> 	Animation type: discrete
> 	</pre>
> ```

So my first suggestion would be to extend list of values for `view-transition-name` as follows:

```diff
- 	Value: none | <<custom-ident>>
+ 	Value: none | match-element | auto | <<custom-ident>>
```

Both terms could then use a (simple) description but one without an example, as the description can link to [the section on auto naming](https://drafts.csswg.org/css-view-transitions-2/#auto-vt-name) for examples.

I’d also recommend to use “element identity” in the descriptions, as that’s the terminology we’ve used in the discussions and one that make it clear.

In https://developer.chrome.com/blog/view-transitions-in-2025#auto-name_elements_with_view-transition-name_match-element I also included the following note, which could make a good note in the spec as well:

> Note: Elements in different documents have a different identity, therefore match-element can only be used in same-document view transitions.

PR #12972, opened 2025-10-17
[css-backgrounds-4][editorial] Incorporated rest of CSS Backgrounds 3 text

SebastianZ:
This is an initial import of the full CSS Backgrounds 3 text related to backgrounds.

It also adds the WPTs from CSS Backgrounds 3 and updates the tests.

I've renamed the "Backgrounds" section to "Defining Backgrounds" and moved the sections "Layering Multiple Background Images" and "Backgrounds of Special Elements" to the same level.
The formatting is mostly the same with a few tweaks.

Some of the links to other specs should probably be updated, though that's something that should be done in a separate patch for both levels.

Sebastian


SebastianZ:
Thanks @svgeesus!

Issue #7762, opened 2022-09-19
[css-animations] Clarify whether `<keyframes-name>` supports the empty string


yisibl:
The spec allows `<string>` here https://github.com/w3c/csswg-drafts/issues/118#issuecomment-227655060, but does not further clarify whether empty strings(`""` or `" "`) are allowed. 

@ewilligers Says: 
> Also, do we want to ban the empty string '""' ?
> https://github.com/w3c/csswg-drafts/issues/2435#issuecomment-445630241

- From the CSS author's point of view, empty strings seem to be meaningless.
- For current implementations across browsers, this has significant compatibility issues.

## `@keyframes "" {}`

> If the browser supports it, the page background color will be green.

```css
@-webkit-keyframes "" {
    to {
        background: green
    }
}

@-moz-keyframes "" {
    to {
        background: green
    }
}

body {
    background: red;
    -webkit-animation: "" 0s ease 1 forwards;
    -moz-animation: "" 0s ease 1 forwards;
}
```

- ` -webkit-` (Chrome) ✅
- `-webkit-`(Safari) ❌
- `-moz-`(Firefox) ❌
- without prefix(both) ❌ [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1554737)

## `@keyframes " " {}`

```css
@-webkit-keyframes " " {
    to {
        background: green
    }
}

@-moz-keyframes " " {
    to {
        background: green
    }
}

body {
    background: red;
    -webkit-animation: " " 0s ease 1 forwards;
    -moz-animation: " " 0s ease 1 forwards;
}
```

- ` -webkit-` (Chrome) ✅
- `-webkit-`(Safari) ✅
- `-moz-`(Firefox) ✅

yisibl:
Firefox has a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1554737

tabatkins:
I'm inclined to say that empty string is not allowed, as that cannot be reproduced as an ident. (All other strings can be, with suitable escapes.)

fantasai:
I agree with @tabatkins

javifernandez:
@fantasai is the PR https://github.com/w3c/csswg-drafts/pull/12985 enough to solve this issue ?

yisibl:
[...]

PR #12985, opened 2025-10-20
[css-animation-1] Handle <string> values in keyframes-name as identitiers

javifernandez:
Per resolution:

https://github.com/w3c/csswg-drafts/issues/2435#issuecomment-428646888 

> 1. Accept `<string>` or `<custom-ident>` as we currently do.
> 2. If `<string>` is one of the CSS-wide keywords or "default" then it is invalid (at parse / variable-resolution time) as per the rules for custom-idents: https://www.w3.org/TR/css-values-3/#custom-idents
> 3. When serializing, if the value doesn't have [whitespace](https://www.w3.org/TR/css-syntax-3/#whitespace)(??) serialize without quotes (i.e. as <custom-ident>), otherwise serialize as `<string>`.

Fixes issues #2435 and #7762

### DIFF
--- a/css-animations-1/Overview.bs
+++ b/css-animations-1/Overview.bs
@@ -229,6 +229,10 @@ Declaring Keyframes</h2>
 	the names are fully <a>case-sensitive</a>;
 	two names are equal only if they are codepoint-by-codepoint equal.
 	The <<custom-ident>> additionally excludes the ''animation-name/none'' keyword.
+	The <<string>> should be treated as <<custom-ident>>, hence parsed as invalid if
+	is one of the <a>CSS-wide keywords</a> or "default"; empty string is not allowed
+	either. It's serialized as identifier when possible and as <<string>> otherwise.
+
 
 	<div class=example>
 		For example, the following two ''@keyframes'' rules have the same name,
@@ -254,8 +258,8 @@ Declaring Keyframes</h2>
 		@keyframes None { /* ... */ }
 		</pre>
 
-		However, those names <em>can</em> be specified with a <<string>>,
-		so the following are both <em>valid</em>:
+		They are also <em>invalid</em> if those names are specified with a <<string>>
+		because they can not be serialized as valid <<custom-ident>>:
 
 		<pre class=lang-css>
 		@keyframes "initial" { /* ... */ }


bkardell:
I'm really not sure this should have been merged, there are definitely some disagreements.  This would make 2 otherwise interoperable implementations incorrect on quoted names, right?  And @nt1m is making a case on one of the linked issues that it doesn't make sense.  

Also, minor nit but 

> hence parsed as invalid if is one of 

doesn't this miss the work "it"?

css-animations-1/Overview.bs
```diff
@@ -229,6 +229,10 @@ Declaring Keyframes</h2>
 	the names are fully <a>case-sensitive</a>;
 	two names are equal only if they are codepoint-by-codepoint equal.
 	The <<custom-ident>> additionally excludes the ''animation-name/none'' keyword.
+	The <<string>> should be treated as <<custom-ident>>, hence parsed as invalid if
+	is one of the <a>CSS-wide keywords</a> or "default"; empty string is not allowed
+	either. It's serialized as identifier when possible and as <<string>> otherwise.
```
yisibl, line 234: > empty string is not allowed either.

I believe we should also clarify how `animation-name: ""` is parsed.

The explanation is as follows:

Since empty string are not permitted in `@keyframes`, `animation-name: ""` will be parsed as `none`.

javifernandez: This PR has been reverted because the changes related to make invalid a ```<string>``` matching a css-wide keyword.  However, the sentence about the empty string would be valid, indeed. 

I'm not sure we need to specify that the value will be none, because an invalid parsing always lead to the property hold its "initial" value.

dholbert: > I'm not sure we need to specify that the value will be none, because an invalid parsing always lead to the property hold its "initial" value.

Seems to me like we should ensure it's specified.  Right now, Chrome Canary/Safari/Firefox all differ on how to serialize `animation-name: ''` (and specifically whether you get the empty string vs. `none` in the specified & computed style.) See https://bugzilla.mozilla.org/show_bug.cgi?id=1998933#c6 for testcases and results.

This is a web-exposed behavior;  the WPT (https://wpt.fyi/results/css/css-animations/parsing/animation-name-invalid.html ) happens to test the specified-style serialization where Chrome/Safari happen to now agree and diverge from Firefox, and the divergence is why Firefox is currently marked as failing a subtest there.

Chrome agrees with Firefox on the **computed**-style-serialization, though; so if there were a WPT for that (maybe there is?) then Chrome/Firefox would get one result and Safari would get another.

javifernandez: Maybe it'd be better to continue the discussion in the issue #7762

Issue #12799, opened 2025-09-15
[selectors-4] Move to CR?


SebastianZ:
Given the [huge list of features that are seem to be ready to release](https://github.com/w3c/csswg-drafts/issues/12797), I wonder whether we should try to get Selectors 4 in shape for moving to Candidate Recommendation status.

To do that, I suggest moving the following features to [level 5 of the spec.](https://drafts.csswg.org/selectors-5/) and feature-freeze level 4:

* `:local-link` (already there, though needs to be removed from level 4)
* `:interest-source` and `:interest-target`
* `:blank`
* All grid-structural selectors, i.e. the `||` combinator, `:nth-col()`, and `:nth-last-col()`
* All time-dimensional pseudo-classes , i.e. `:current`, `:current()`, `:past`, and `:future`

@fantasai @tabatkins What's your opinion?

Sebastian

svgeesus:
Stabilizing and feature-freezing seems like a good plan. Then:

- [x] Update (and trim, for removed features) the changes section
- [x] Add an (empty) Security Considerations
- [x] Add WPT annotations for the  relevant [tests](https://wpt.fyi/results/css/selectors?label=master&label=experimental&aligned)
- [x] Republish a WD
- [x] Get horizontal review
- [ ] Republish WD including changes from horizontal review
- [ ] Triage not-for this-level issues
- [ ] Request CR

fantasai:
I am in favor of this plan. Pushing Selectors 4 to CR is long overdue.

Are we proposing to defer `:current()`, `:current()` and `:current`, or all time-dimensional pseudo-classes?

SebastianZ:
[`:past`](https://developer.mozilla.org/en-US/docs/Web/CSS/:past) and [`:future`](https://developer.mozilla.org/en-US/docs/Web/CSS/:future) already seem to be implemented in Chrome, Edge, Safari while [`:current`](https://developer.mozilla.org/en-US/docs/Web/CSS/:current) (and `:current()`) is not, according to MDN. There are no WPTs to verify that, though.

Anyway, I wouldn't have a problem to defer all time-dimensional pseudo-classes if that helps avoiding controversies.

Sebastian

fantasai:
FWIW, WebKit implements the resource state pseudo classes, so we may want to keep them if they are not riddled with issues.

SebastianZ:
@fantasai Fine, I've removed the resource state pseudo classes from the list.

[...]

PR #12986, opened 2025-10-20
[selectors-4][editorial] Added WPTs

SebastianZ:
I added all the Web Platform Tests that currently cover the features in Selectors 4.

For the reviewers: Please especially check whether the tests under `old-tests/` are placed in the correct section and the tests in "Features from other specs and unrelated tests".

Note that this PR does _not_ include hints about missing WPT coverage. This needs to be done separately.

(This is part of the effort to get Selectors 4 into shape for CR, see https://github.com/w3c/csswg-drafts/issues/12799#issuecomment-3291663905.)

Sebastian


svgeesus (commented):
This looks reasonable to me on a quick diff inspection of test names (meaning I have not grepped for rel="help" in the tests, far less read the source of each test)

PR #12988, opened 2025-10-21
[css-nesting-1] Fix character reference errors

andruud:
For example:
`LINE 330:39: Character reference '&Bar' didn't end in ;`.

Replaced various instances of `&` with `&amp;`.

### DIFF
--- a/css-nesting-1/Overview.bs
+++ b/css-nesting-1/Overview.bs
@@ -327,10 +327,10 @@ Syntax {#syntax}
 		as nesting is not a syntax transformation,
 		but rather matches on the actual elements the parent selector matches.
 
-		It is also true that the selector ''&Bar'' is invalid in CSS in the first place,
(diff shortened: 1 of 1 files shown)


tabatkins:
> I'd argue that the enclosing in `''` seems also wrong to me, because Bikeshed then tries to link those terms, but that can be done in a follow up.

Note that the `''foo''` syntax has traditionally been used *both* for linking CSS values *and* for just marking stuff up in a particular style as CSS content. Bikeshed's cool with that. When it's occasionally problematic, the `<css>...</css>` fake element can be used for styling.

So continuing to use it for selectors as it is here is fine.

(This mixing *is* problematic in Bikeshed's parser impl - it needs to simultaneously parse as HTML *and* parse as raw text, and decide which is correct only after resolving links - but oh well.)

Issue #10443, opened 2024-06-13
[css-cascade][css-syntax] New `!revertable` flag to mark a declaration as "can be reverted when IACVT"


LeaVerou:
This came out of #5319 but I figured I should make a new post for this for visibility, since it’s a distinct proposal.
#5319 discusses a keyword like `revert-declaration`, however I don’t think a keyword will work, since anything that is part of the value cannot be known in advance (it can be part of a variable for example). 

The problem is that UAs throw the other declarations away when they see a new one in the same scope. It's impossible to know whether the new one will ever use that keyword so they can keep the old ones around too. It works with `revert` and `revert-layer` because the other declarations are kept around *anyway* since they’re in a separate origin/layer.

However, I think a new !-flag may actually work. TBB, but some potential names could be `!revertable`, `!revocable`, `!transient`, `!layered`. Since `!`-flags need to be present at parse time and are always at the end, it doesn’t have the issues that a keyword would.
When a declaration with this flag is encountered, the UA keeps old declarations around instead of replacing them with this.

```css
/* Will not reset border-radius to 0 if --pill-radius is not set, it will just cascade normally */
border-radius: var(--pill-radius) !revertable; 
```

There can be more than one:

```css
* {
	/* Will not override animation if --effect is not set to one of these values */
	animation: if(style(--effect: grow), 1s grow) !revertable;
	animation: if(style(--effect: pulse), 4s pulse infinite) !revertable;
}
```

One open question is how would this work in the CSS OM. But we could just expose the last declaration, and authors can use Typed OM to get all of them (which allows append).

andruud:
There's overlap here with the ideas in https://github.com/w3c/csswg-drafts/issues/1594.

> The problem is that UAs throw the other declarations away when they see a new one in the same scope.

Ideally, we'd just not do that, and make it always possible to revert to the previous "winner" of the cascade (using a keyword). We'd need to investigate the performance implications of this, though.

LeaVerou:
That’s the idea for the opt-in keyword: that authors would not be doing this for every single declaration, so the performance implications should be minimal.

Loirooriol:
```css
#foo {
  height: 1px;
  block-size: 2px;
  height: var(--invalid) !revertable;
}
```

Assuming an horizontal writing mode, should this revert to 2px or 1px? I guess 2px like `revert-layer` would do in different layers, but it seems extra annoying for CSSOM.

I wonder if it would make more sense to add an entirely new rule like

```css
[...]

PR #12992, opened 2025-10-21
[css-cascade-5] Add the revert-rule keyword

andruud:
A link-defaults entry is needed for 'if()' to avoid links to css-conditional-values-1.

#10443

### DIFF
--- a/css-cascade-5/Overview.bs
+++ b/css-cascade-5/Overview.bs
@@ -31,6 +31,7 @@ Informative Classes: ex
 spec:dom; type:dfn; text:shadow tree
 spec:css-color-4; type:property; text:color
 spec:css-values-3; type: value; text:ex
+spec:css-values-5; type: function; text:if()
 spec:css-conditional-3; type:at-rule; text:@media
 spec:mediaqueries-4; type:type; for:@media; text:all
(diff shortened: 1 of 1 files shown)


css-cascade-5/Overview.bs
```diff
@@ -1954,6 +1955,42 @@ Rolling Back Cascade Layers: the ''revert-layer'' keyword</h4>
...
+
+		Given an element <code>&lt;div class=apply-sharp></code>,
+		the above style sets ''border-radius'' to <code>0px</code>
+		only when the [=computed value=] of <code>--mode</code>
+		is <code>sharp</code>.
+		</pre>
+	</div>
+
+	The ''revert-rule'' keyword behaves like ''revert-layer''
+	in the [=author origin=].
```
Loirooriol, line 1992: How come? It would be useless to add it if it just behaves as `revert-layer`.

andruud: Yeah, I noticed that myself earlier today. It's just a mistake. https://github.com/w3c/csswg-drafts/commit/14a546238ca4d8fae8db44dc3f1473fef32a347c

Issue #13001, opened 2025-10-21
[css-ui] regression in definition of cursor values with URLs


tyrasd:
#5831 introduced `url-set` definitions of the `cursor` property. However, it also introduced a subtle change in the https://github.com/w3c/csswg-drafts/commit/78fb03bd72db009decb0bdad200cd31f9dd9a235#diff-cc2f666bf7d93dcefcc3fffc4761793a54fba00c03b53b9dbdf81f5056f0b56eR604

before: 
```
[ [<<url>> [ <x> <y> ]?,]* [ auto | … ] ] 
```

after:
```
[ [ <<url>> | <<url-set>> ] [ <x> <y> ]? ]#? [ auto | … ]
```

While looking very similar, the two versions are not equivalent: the former requires a comma after each URL, resulting in a comma between the last URL and the "fallback" predefined cursor (e.g. as in `cursor: url(example.png), auto`). The latter does however use the `#`[^1] syntax which does not allow for a comma to be present after the list of URLs (e.g. hypothetically resulting in `cursor: url(example.png) auto`).

[^1]: > A hash mark (#) indicates that the preceding type, word, or group occurs one or more times, separated by comma tokens

As latter (`#?`) version is not backwards compatible to CSS Level 3, incompatible with how browsers currently handle the definitions of cursors using URLs and does not match the example given on the current draft version of the page as well as other related sources like [wpt](https://github.com/web-platform-tests/wpt/blob/master/css/css-ui/cursor-image-015-manual.html#L12), I assume this change was introduced by mistake.

PS: with 1a57a4025c8778dab45ef8d5fbd482e4d6c4ebce, this mistake also made it into the CSS Level3 definitions on this repository.

emilio:
cc @frivoal

frivoal:
That was indeed accidental. We should restore the `,]*` version of the syntax.

frivoal:
Fixed.

PR #13002, opened 2025-10-21
[css-ui] fix missing comma in value definition for "cursor" property #13001

tyrasd:
A comma is necessary between cursor image url(s) and the predefined "fallback" cursor. Fixes https://github.com/w3c/csswg-drafts/issues/13001

### DIFF
--- a/css-ui-3/Overview.bs
+++ b/css-ui-3/Overview.bs
@@ -921,7 +921,7 @@ While the content is being scrolled, implementations may adjust their rendering 
 
 <pre class="propdef">
 Name: cursor
-Value: <<cursor-image>>#? <<cursor-predefined>>
+Value: <<cursor-image>>#? , <<cursor-predefined>>
 Initial: auto
 Applies to: all elements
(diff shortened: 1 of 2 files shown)


tyrasd:
superseded by #13018

frivoal:
Sorry, I had not seen this PR before making #13018

tyrasd:
All good :relaxed: . PS: One thing I had noticed when creating this small fix was that the `… ,]* …` approach seems to be used more in relatively older parts of the spec, while the equivalent `…#? , …` seems to be preferred in the more recent parts (e.g. [`css-mixins-1`](https://drafts.csswg.org/css-mixins-1/#mixin-rule), [`css-gaps-1`](https://drafts.csswg.org/css-gaps-1/#gap-decoration-shorthands)) and in the prominent [`css-backgrounds`](https://drafts.csswg.org/css-backgrounds-3/#background). Maybe it would be worth to keep those definitions consistent across the different parts of the spec?

PR #13017, opened 2025-10-26
[css-animations-2] Use correct property names for timeline-triger-range

keithamus:
This seems to have been a change from https://github.com/w3c/csswg-drafts/pull/13009.


keithamus:
Fixed in 7daaa2f4629702b620462f0886c832190388ff53

PR #13021, opened 2025-10-27
[css-borders-4] Clarify how `border-shape` renders `box-shadow`.

noamr:
This doesn't require a new resolution, as there is no previous resolution on how border-shape shadow rendering works.
Stroking the inner/outer path by the width of the shadow spread matches how box-shadow works elsewhere.


noamr:
> having an inner shadow sounds a bit weird? but I guess more features for authors is a good way.

That already exists (`box-shadow: *** inset`).

PR #13026, opened 2025-10-27
[css-ui-3][css-ui-4][editorial] Added WPTs

SebastianZ:
I added all the Web Platform Tests that currently cover the features in CSS UI 3 and 4.

Note that this PR does _not_ include hints about missing WPT coverage or notes about tests not being necessary for specific sections. Those should be added separately.

Sebastian

### DIFF
--- a/css-ui-3/Overview.bs
+++ b/css-ui-3/Overview.bs
@@ -1772,8 +1772,15 @@ and
 
 This appendix is <em>informative</em>.
 
-This appendix describes changes from the
-<a href="https://www.w3.org/TR/2017/CR-css-ui-3-20170302/">Candidate Recommendation (CR) of 2 March 2017</a>.
+<h3 id="changes-2018-06-21" class="no-num">
+Changes from the <a href="https://www.w3.org/TR/2018/REC-css-ui-3-20180621/">21 June 2018 Recommendation (REC)</a></h3>
+
+<ul>
+	<li>Added Web Platform Tests coverage
+</ul>
+
+<h3 id="changes-2021-03-16" class="no-num">
+Changes from the <a href="https://www.w3.org/TR/2017/CR-css-ui-3-20170302/">2 March 2017 Candidate Recommendation (CR)</a></h3>
 
 <ul>
 	<li>Updated references to latest versions


css-ui-3/Overview.bs
```diff
@@ -1256,6 +1579,11 @@ but still generate boxes. For example,
 if the element has ''visibility: hidden'' but not ''display: none'',
 boxes are generated for it and its cursor is used for the canvas.
 
+<wpt hidden title="Manual tests">
```
frivoal: Why hidden?

css-ui-3/Overview.bs
```diff
@@ -1634,3 +1991,1017 @@ frame[noresize] {
...
+historical/moz-user-modify-01.html
+historical/moz-user-modify-02.html
+historical/moz-user-modify-03.html
+historical/user-modify-01.html
+historical/user-modify-02.html
+historical/user-modify-03.html
+historical/webkit-user-modify-01.html
+historical/webkit-user-modify-02.html
+historical/webkit-user-modify-03.html
+</wpt>
```
frivoal: How about converting these tests to [tentative tests](https://web-platform-tests.org/writing-tests/file-names.html#test-features) instead? In that case they wouldn't need to be listed at all.

Until that's done, this seems appropriate, but marking them as tentative would seem better.

css-ui-4/Overview.bs
```diff
@@ -3170,3 +4448,101 @@ Appendix D. Default style sheet additions for HTML</h2>
...
+	historical/moz-user-modify-01.html
+	historical/moz-user-modify-02.html
+	historical/moz-user-modify-03.html
+	historical/user-modify-01.html
+	historical/user-modify-02.html
+	historical/user-modify-03.html
+	historical/webkit-user-modify-01.html
+	historical/webkit-user-modify-02.html
+	historical/webkit-user-modify-03.html
+	</wpt>
```
frivoal: Same comment as https://github.com/w3c/csswg-drafts/pull/13026/files#r2467502659 :
> How about converting these tests to tentative tests instead? In that case they wouldn't need to be listed at all.
>
> Until that's done, this seems appropriate, but marking them as tentative would seem better.

frivoal (commented):
Without reviewing every single test in detail, which seems like more work than I can justify at the moment, a review of this PR can only somewhat shallow, but having looked through it at that high level, it seems fine.

I think `<wpt>` annotation should ideally be spread through the spec and be inserted after each paragraph that describes something testable, rather than grouped at the start of each section, but doing so takes a tremendous amount of work, and getting the tests into the right sections, as this PR does, is already useful.

I left a few minor comments, but overall, this seems fine.

css-ui-3/Overview.bs
SebastianZ: I initially added the manual tests as hidden and after realizing that that's not necessary, I simply forgot this spot. Removed it now.

css-ui-3/Overview.bs
SebastianZ: Right. Done that, see https://github.com/web-platform-tests/wpt/pull/55701.

css-ui-4/Overview.bs
SebastianZ: Right. Done that, see https://github.com/web-platform-tests/wpt/pull/55701.

SebastianZ:
> think `<wpt>` annotation should ideally be spread through the spec and be inserted after each paragraph that describes something testable, rather than grouped at the start of each section, but doing so takes a tremendous amount of work, and getting the tests into the right sections, as this PR does, is already useful.

I agree they should ideally be moved to the paragraphs they belong to. Though it already took quite some time to add all the 1312 tests.😅 So that may be done in a follow-up change.

Sebastian

SebastianZ:
@frivoal Thank you for the approval! Please note that I added a last change that restructured the Changes section to be consistent with the one of level 4 and added a note about the WPTs there.

Sebastian

PR #13024, opened 2025-10-27
[css-gaps-1] Update `outset` properties #12603, #12848, #8402

jav099:
https://jav099.github.io/javier-intersections-draft.github.io/

Issue #8402, opened 2023-02-06
[css-text-decor-4] Consider renaming `text-decoration-trim`


SebastianZ:
Besides the general discussion in #4557 to introduce a way to modify the length of the decoration, there was also some discussion around how to call the property. As I [previously summarized](https://github.com/w3c/csswg-drafts/issues/4557#issuecomment-1184944512), name suggestions for it so far were:

* `text-decoration-trim`
* `text-decoration-inset`
* `text-decoration-limit`
* `text-decoration-clip`

As there are also use cases for _extending_ the decorations beyond their normal size. In my opinion, "trim" doesn't express that well enough. Therefore I slightly tend towards "limit".

We have a poll running in comments below:
https://github.com/w3c/csswg-drafts/issues/8402#issuecomment-3408427694
https://github.com/w3c/csswg-drafts/issues/8402#issuecomment-3408451101

Sebastian

jfkthame:
We're in the process of implementing this, so if there's going to be some naming bikeshedding, we'd appreciate it happening sooner rather than later.

(Personally, I'm happy with `text-decoration-trim`; the idea that "negative trimming" extends the line doesn't bother me. Neither `-limit` nor `-clip` seem nearly as clear; `-inset` would be OK but no better than `-trim`.)

SebastianZ:
I'd argue that `*-inset` is definitely better than `*-trim` and would also be more conform with other properties like `view-timeline-inset` or `inset-*`.

All the other properties that have "trim" in their name like `white-space-trim`, `text-box-trim`, `margin-trim`, etc. are keyword-based. And `trim` rather indicates cutting off something rather than potentially expanding it. Therefore I am strongly against using `*-trim` and `*-clip` and prefer either `*-inset` or `*-limit`.

Sebastian

jfkthame:
Thinking about this again, I'm feeling less in favor of `-inset` here. Not sure how well I can articulate this, but fwiw....

To my mind, `inset` properties are about adjusting their target element _in relation to another element_ (a parent or containing box). Given an element that is being sized or placed within an area that has been defined by some external thing, it makes sense to have the ability to "inset" (or expand, by use of negative inset) that area.

But with decoration lines, it's not readily apparent to an author that there is any such "container" to apply insets to. The decoration line seems more like an attribute of a range of text, and its length is determined by the length of that text, not in relation to some containing element, so the idea of "insetting" from such a container is a little odd.

[...]

PR #13039, opened 2025-10-29
[css-text-decor-4] Renamed text-decoration-trim to text-decoration-inset

SebastianZ:
According to the resolution from https://github.com/w3c/csswg-drafts/issues/8402#issuecomment-3462630143.

See also #13024 for the part of the `{row|column}-rule-*` properties.

Sebastian

### DIFF
--- a/css-text-decor-4/Overview.bs
+++ b/css-text-decor-4/Overview.bs
@@ -860,7 +860,7 @@ Text Decoration Line Uniformity</h3>
 
 
 <h4 id="text-decoration-inset-property">
-Adjusting Text Decoration Line Lengths: the 'text-decoration-inset' property</h4>
+Trimming, Expanding, and Shifting Text Decoration Lines: the 'text-decoration-inset' property</h4>
 
 	<pre class="propdef">
 	Name: text-decoration-inset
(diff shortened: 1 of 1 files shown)


jfkthame:
There are also a number of mentions of "trim" in the description of the `auto` value that should be reworded, I think.

(While you're there, perhaps also fix the long-obsolete mention of `text-decoration-skip-inset` in the caption of the Chinese `auto` illustration.)

SebastianZ:
> There are also a number of mentions of "trim" in the description of the `auto` value that should be reworded, I think.

Replaced that by "length adjustments".

> (While you're there, perhaps also fix the long-obsolete mention of `text-decoration-skip-inset` in the caption of the Chinese `auto` illustration.)

I renamed them all to `text-decoration-inset` now.

It seems we also had a `text-decoration-skip-inset` longhand for the `text-decoration-skip` shorthand at some point. As it seems it got replaced by `text-decoration-inset` at some point, I've now removed the remnants of the other property. That also means, `text-decoration-inset` is _not_ a longhand for `text-decoration-skip` anymore.

Sebastian

css-text-decor-4/Overview.bs
```diff
@@ -894,38 +894,38 @@ Text Decoration Line Trimming and Extension: the 'text-decoration-trim' property
...
 				        width="223"
 				        height="77">
-				<p class="caption">''text-decoration-skip-inset: auto'' for <code>&lt;u>石井&lt;/u>&lt;u>艾俐俐&lt;/u></code>
+				<p class="caption">''text-decoration-inset: auto'' for <code>&lt;u>石井&lt;/u>&lt;u>艾俐俐&lt;/u></code>
 </p>
 			</div>
 	</dl>
 
-	Text decoration trimming is subject to 'box-decoration-break':
+	The adjustment of the text decoration's length is subject to 'box-decoration-break':
```
fantasai: We're not adjusting the length, we're adjusting the start and endpoints. It could keep the length constant, for example if it were specified as `-2px 2px`. Open to other ideas, but I think keeping with the "trimming" terminology is clearer here, even if the name of the property has changed. "Length adjustment" is not precise enough.

SebastianZ: Good point! I tried to adust the wording. Let me know if it's better now!

jfkthame (approved):
LGTM, thanks for doing this.

PR #13043, opened 2025-10-30
[css-backgrounds-4] Clarified that decorations and emphasis marks are included in `background-clip: text` and text color has no influence on clipping

### LINKED ISSUE #900: [css-backgrounds] Background-clip and text underline
Chrome and Firefox do not provide a way to render underlined text when using ```background-clip:text```.

https://jsfiddle.net/c3p74zuo/3/

image
image

> The background is painted within (clipped to) the union of the border box and the geometry of the text in the element and its in-flow and floated descendants. 
> https://drafts.csswg.org/css-backgrounds-4/#background-clip

### LINKED ISSUE #10805: [css-backgrounds-4] The `background-clip: text` needs to contain `text-emphasis`
At #900 we've already decided that `text-decoration` is considered part of the text, so it will be clipped as well.

`text-emphasis` should also be treated the same.

There are currently some issues with both Safari and Firefox, see examples: 

https://codepen.io/yisi/pen/zYVJqyR?editors=0100

## Safari TP 202

Looks like it's dropped `U+FE0F` (VS16)

image

## Firefox 
image

cc @smfr @nt1m

SebastianZ:
See the resolution in https://github.com/w3c/csswg-drafts/issues/900#issuecomment-342251900.

Closes #900, closes #10805

Sebastian

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -1035,7 +1035,8 @@ Painting Area: the 'background-clip' property</h3>
 		<dd>
 			The background is painted within (clipped to)
 			the intersection of the border box
-			and the geometry of the text in the element and its [=in-flow=] and floated descendants.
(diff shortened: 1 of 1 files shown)


yisibl:
Awesome!

PR #13047, opened 2025-10-31
[css-mixins-1] Correct @function grammar #13046 

### LINKED ISSUE #13046: [css-mixins-1] @function grammar is missing open-paren
The definition of the `@function` syntax is:

```html
<pre class="prod def" nohighlight>
&lt;@function> = @function <<function-token>> <<function-parameter>>#? )
	[ returns <<css-type>> ]?
{
	<<declaration-rule-list>>
}

<dfn><<function-parameter>></dfn> = <<custom-property-name>> <<css-type>>? [ : <<default-value>> ]?
<dfn><<css-type>></dfn> = <<syntax-component>> | <<@function/type()>>
<dfn><<default-value>></dfn> = <<declaration-value>>
<dfn function lt="type()" for="@function">&lt;type()></dfn> = type( <<syntax>> )
</pre>
```

The literal close-paren (`)`) that follows the `<function-parameter>` list is represented, but the _open_ paren (`(`) that should fall between the `<function-token>` and the (optional) parameter list is not.

The simplest solution is for the grammar to read:

```html
<pre class="prod def" nohighlight>
&lt;@function> = @function <<function-token>> ( <<function-parameter>>#? )
	[ returns <<css-type>> ]?
{
	<<declaration-rule-list>>
}

<dfn><<function-parameter>></dfn> = <<custom-property-name>> <<css-type>>? [ : <<default-value>> ]?
<dfn><<css-type>></dfn> = <<syntax-component>> | <<@function/type()>>
<dfn><<default-value>></dfn> = <<declaration-value>>
<dfn function lt="type()" for="@function">&lt;type()></dfn> = type( <<syntax>> )
</pre>
```

Unless `<function-token>` is meant to include the open-paren, but since it isn't defined that way anywhere (formally or informally), that seems unlikely.

(Also, it would make more sense structurally if `<function-token>` were then defined to include not only the open-paren, but also the argument list and the close-paren. Its definition could then be decomposed further.)

ferdnyc:
Add the missing literal open-paren (`(`) to the definition.

Fixes #13046

### DIFF
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -194,7 +194,7 @@ optionally a <dfn>parameter type</dfn>, described by a [=syntax definition=];
 and optionally a <dfn>default value</dfn>.
 
 <pre class="prod def" nohighlight>
-&lt;@function> = @function <<function-token>> <<function-parameter>>#? )
+&lt;@function> = @function <<function-token>> ( <<function-parameter>>#? )
 	[ returns <<css-type>> ]?
 {
 	<<declaration-rule-list>>


tabatkins:
`<function-token>` actually already includes the opening paren.

PR #13052, opened 2025-11-02
[css-fonts-5] Define behavior for <meta text-scale> from #12380

davidsgrogan:
Initial definition of `<meta name="text-scale" content="...">` resolved on in https://github.com/w3c/csswg-drafts/issues/12380

### DIFF
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -78,20 +78,15 @@ For readability they have not been repeated explicitly.
 <h2 id="text-scale-meta">
 Text-Scale <{meta}> element</h2>
 
-The <dfn>text scale factor</dfn> is a multiplier
-used to calculate the computed value of the ''font-size/medium'' font size.
-The computed value of ''font-size/medium''
-is 16px
-multiplied by the [=text scale factor=].
-This factor also
-affects the computed size of the other <<absolute-size>> keywords,
-but not necessarily by the same factor as ''font-size/medium''.
-The UA should non-uniformly adjust
-keywords other than ''font-size/medium'' for improved readability.
+User agents must calculate the computed value of ''font-size/medium'' by multiplying 16px by
+the <dfn>text scale factor</dfn>.
+User agents must also apply this factor when determining the computed sizes of the other <<absolute-size>> keywords.
+However, user agents should scale these other keywords non-linearly
+to preserve readability.
 
-Note: Scaling all font sizes uniformly would lead to large fonts being too large.
-See <a href="https://github.com/w3c/wcag3/issues/261#issuecomment-2598860630">
-WCAG discussion about minimum and maximum critical font sizes</a>.
+Note: Applying a linear scale to all keywords would render large fonts excessively large. See the
+<a href="https://github.com/w3c/wcag3/issues/261#issuecomment-2598860630">
+WCAG discussion on minimum and maximum critical font sizes</a>.
 
 A document with a <{meta}> tag
 whose <{meta/name}> attribute


davidsgrogan:
Hi @JoshTumath , this is my fork of Keith's initial PR. I left the definition of env(preferred-text-scale) in css-env-1 and just referred to it from here. Even though this PR is not fully complete (I didn't mention media queries, for example), I'm hoping to get it in as a kind of MVP spec. Let me know if you think something major is missing and definitely let me know if something is not correct.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,63 @@ rules defined in CSS Fonts Level 4.
 This specification is currently a delta to the CSS Fonts Level 4 specification.
 Do not assume that if something is not here, it has been dropped.
 
+<h2 id="text-scale-meta">
+Text-Scale <code class=html>&lt;meta&gt;</code> element</h2>
 
+A document with a <code>&lt;meta></code> tag whose <code>name</code> attribute
```
fantasai: This needs to be more specific, should refer to the HTML <{meta}> tag specifically. (Random XML documents with a META tag don't count.)

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,63 @@ rules defined in CSS Fonts Level 4.
...
+is a <a>ASCII case-insensitive</a> match for
+<dfn lt=text-scale><code>"text-scale"</code></dfn> is recognized as setting the
+initial font size of the document. The value of the <code>content</code>
```
fantasai: We're not setting the initial font size. We're setting the computed value of the `medium` font-size, and scaling all the other absolute size keywords accordingly.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,63 @@ rules defined in CSS Fonts Level 4.
...
+attribute must be an <a>ASCII case-insensitive</a> match for one of the
```
fantasai, line 98: Must -> otherwise what?

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,63 @@ rules defined in CSS Fonts Level 4.
...
+<h3 id="text-scale-meta-keywords">
+Keywords</h3>
+
+The recognized keywords in the [=text-scale=]
+<code class=html>&lt;meta&gt;</code> element are:
+
+<ul>
+	<li><code class="index" lt="legacy!!text-scale-meta">legacy</code></li>
+	<li><code class="index" lt="scale!!text-scale-meta">scale</code></li>
+</ul>
```
fantasai: I think it would be better to follow the DL format that we have for property values: list them in the DT with a DFN tag, and give them a definition that's understandable to authors wrt what they do.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,63 @@ rules defined in CSS Fonts Level 4.
...
+
+<h3 id="legacy-keyword">The 'legacy' keyword</h3>
+
+The <dfn for="text-scale" export><code>legacy</code></dfn> property is
+recognized in the [=text-scale=] content attribute value.
+
+When the value of the [=text-scale=] content attribute is
+<a for="text-scale">legacy</a> the user agent should set the initial font size
```
fantasai: should or must? If not must, under what conditions is it reasonable to ignore the should?

css-fonts-5/Overview.bs
```diff
@@ -59,7 +59,63 @@ rules defined in CSS Fonts Level 4.
...
+</ul>
+
+<h3 id="scale-keyword">The 'scale' keyword</h3>
+
+The <dfn for="text-scale" export><code>scale</code></dfn> property is
+recognized in the [=text-scale=] content attribute value.
+
+When the value of the [=text-scale=] content attribute is
+<a for=text-scale>scale</a> the user agent may determine the initial font size
+based on a combination of the operating system's text scale setting and the user agent's text scale setting. The
```
fantasai: "may" or "must"? Shouldn't this be a "must"? Can we define more definitively what this is a "must" about? Like,

The `medium` font size must match the user's preferred paragraph text size as determined from a combination of OS and UA preferences.

fantasai (changes requested):
Also, in general, please follow https://rhodesmill.org/brandon/2012/one-sentence-per-line/ :)

css-fonts-5/Overview.bs
davidsgrogan: Done.

css-fonts-5/Overview.bs
davidsgrogan: Added 'Otherwise the tag is ignored.'

css-fonts-5/Overview.bs
davidsgrogan: Great, taken verbatim.

css-fonts-5/Overview.bs
davidsgrogan: Changed to must.

css-fonts-5/Overview.bs
davidsgrogan: Changed.

css-fonts-5/Overview.bs
davidsgrogan: Gave it a try. Let me know if you meant something different.

davidsgrogan:
> Also, in general, please follow https://rhodesmill.org/brandon/2012/one-sentence-per-line/ :)

Thanks for pointing that out.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+value of <code>legacy</code>.
+
+<h3 id="text-scale-meta-keywords">
+Keywords</h3>
+
+The recognized keywords in the [=text-scale=]
+<{meta}> element are:
+
+<dl>
+	<dt><dfn for="text-scale"><code>legacy</code></dfn></dt>
```
JoshTumath: 'Legacy' is in a code element, but 'scale' isn't. Maybe this should be removed?
```suggestion
	<dt><dfn for="text-scale">legacy</dfn></dt>
```

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
 This specification is currently a delta to the CSS Fonts Level 4 specification.
 Do not assume that if something is not here, it has been dropped.
 
+<h2 id="text-scale-meta">
+Text-Scale <code class=html>&lt;meta&gt;</code> element</h2>
```
JoshTumath: I noticed the HTML spec also mentions when a metadata name should not appear more than once in the document. Maybe you should add this line somewhere:

> There must not be more than one [meta](https://html.spec.whatwg.org/multipage/semantics.html#the-meta-element) element with its [name](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-name) attribute value set to an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for [text-scale](https://html.spec.whatwg.org/multipage/semantics.html#meta-color-scheme) per document.

css-fonts-5/Overview.bs
JoshTumath: I'm wondering specifying the 'otherwise' case is neccessary here, because, to me, the [HTML spec seems to imply it is ignored anyway.](https://html.spec.whatwg.org/multipage/semantics.html#the-meta-element)

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+of the ''font-size/medium'' font size
+and consequently scaling the computed size of the other <<absolute-size>> keywords.
+
+The value of the <code>content</code>
+attribute must be an <a>ASCII case-insensitive</a> match for one of the
+recognized keywords. Otherwise the tag is ignored.
+
+Documents without this <code>&lt;meta></code> tag will have an assumed default
+value of <code>legacy</code>.
+
```
JoshTumath, line 103: It would be good to add an example block here, [similar to the example in the HTML spec for the `color-scheme` meta tag.](https://html.spec.whatwg.org/multipage/semantics.html#meta-color-scheme)

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+Keywords</h3>
+
+The recognized keywords in the [=text-scale=]
+<{meta}> element are:
+
+<dl>
+	<dt><dfn for="text-scale"><code>legacy</code></dfn></dt>
+	<dd>''env()/preferred-text-scale'' returns the user's OS-level font factor on mobile,
+	but 1 on desktop.
+	The document's initial font size only incorporates UA-level font preferences.
```
JoshTumath: There's still a few other places that say 'initial font size'. These will also need to be updated to `medium`.

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+When the value of the [=text-scale=] content attribute is
+[=text-scale/legacy=]
+the user agent must set the computed ''font-size/medium'' font size
+to 16px multiplied by
+the font scale factor
+the user has chosen from any settings <i>provided by the user agent</i>.
+The ''preferred-text-scale'' value must be 1 on desktop platforms.
+On mobile:
+<ul>
+	<li>if the operating system provides a text scale setting AND the UA hasn't already applied that factor to the initial font size, ''env()/preferred-text-scale'' returns the mulitplier that the user has chosen in the operating system's text scale setting.
```
JoshTumath: Typo
```suggestion
	<li>if the operating system provides a text scale setting AND the UA hasn't already applied that factor to the initial font size, ''env()/preferred-text-scale'' returns the multiplier that the user has chosen in the operating system's text scale setting.
```

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+
+<h3 id="legacy-keyword">The 'legacy' keyword</h3>
+
+When the value of the [=text-scale=] content attribute is
+[=text-scale/legacy=]
+the user agent must set the computed ''font-size/medium'' font size
+to 16px multiplied by
+the font scale factor
+the user has chosen from any settings <i>provided by the user agent</i>.
+The ''preferred-text-scale'' value must be 1 on desktop platforms.
```
JoshTumath: Wherever you reference the `preferred-text-scale` env(), maybe it should be prefixed with 'environment variable' in the prose. Since this is not the css-env spec, it's not clear that you're referencing an env().

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+On mobile:
+<ul>
+	<li>if the operating system provides a text scale setting AND the UA hasn't already applied that factor to the initial font size, ''env()/preferred-text-scale'' returns the mulitplier that the user has chosen in the operating system's text scale setting.
+
+	Note: As of this writing, all combinations of Android, iOS, Gecko, WebKit, Blink satisfy this first condition.
```
JoshTumath: Maybe 'At the time of publishing'?

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+computed ''font-size/medium'' font size must match the user's preferred paragraph text size as determined from a combination of OS and UA preferences.
+
+The ''env()/preferred-text-scale'' value must be a number that,
+when multiplied by 16px,
+provides a <<length>> that matches that of the initial font size.
+
+Further, when the value of the [=text-scale=] content attribute is
+<a for=text-scale>scale</a>, the user agent should skip all font-sizing interventions it would otherwise perform in an attempt to automatically honor the user's preferences. E.g. text autosizing on mobile (See [[css-size-adjust#intro]]) and full-application zoom (<a href="https://github.com/w3c/csswg-drafts/blob/main/css-env-1/explainers/env-preferred-text-scale.md#windows-11">popular browsers do this on Windows)</a>.
 
+Note: It is expected that authors will use
```
JoshTumath: Consider using active voice for the first sentence. So:

> Authors are expected to use [...]

css-fonts-5/Overview.bs
```diff
@@ -59,7 +60,78 @@ rules defined in CSS Fonts Level 4.
...
+''&lt;meta name="text-scale" content="scale"&gt;'' in stylesheets so that the initial font size will reflect a combination of the user's font preferences, whether those are specified at the OS level or the UA level. The author will then be able to use ''rem'' throughout the page to honor the user's font preferences.
```
JoshTumath: Maybe the sentence about using `rem` is actually better explained in an example block.

So you could keep the Note block to explain how authors are expected to use `scale` in their documents, with the example block below it to explain how, if you set `scale` and don't alter the default font size, content sized with `rem` units will be relative to the preferred text scale.

Also, I'm not sure if the Note should actually be an orange Assertion box. But maybe not because assertions are normative? Where I've seen Assertion boxes in the past are places like [the css-flexbox spec, where it tells authors to use the `flex` shorthand rather than the longhands.](https://www.w3.org/TR/css-flexbox/#flex-components)

JoshTumath (commented):
> Hi @JoshTumath , this is my fork of Keith's initial PR. I left the definition of env(preferred-text-scale) in css-env-1 and just referred to it from here. Even though this PR is not fully complete (I didn't mention media queries, for example), I'm hoping to get it in as a kind of MVP spec. Let me know if you think something major is missing and definitely let me know if something is not correct.

This looks great! As far as I can tell, it covers everything in the Explainer.

Should the font-size section of the spec also redefine the absolute size keywords in relation to the meta tag/env var?

Once this is merged, we'll need to add the meta extension to this WHATWG wiki page: https://wiki.whatwg.org/wiki/MetaExtensions

css-fonts-5/Overview.bs
davidsgrogan: Done

css-fonts-5/Overview.bs
davidsgrogan: Done

css-fonts-5/Overview.bs
davidsgrogan: Done

css-fonts-5/Overview.bs
davidsgrogan: Done

css-fonts-5/Overview.bs
davidsgrogan: Done

css-fonts-5/Overview.bs
davidsgrogan: Bikeshed linkifies ''preferred-text-scale'' to https://drafts.csswg.org/css-env-1/#valdef-env-preferred-text-scale. Do you think that's sufficient? (I couldn't tell if you already knew the linkification part)

css-fonts-5/Overview.bs
davidsgrogan: Yeah, Advisement (you meant "advisement", not "assertion". I think :) boxes are supposed to be normative. This Note seems merely descriptive as is now, but I suppose we can strengthen the language to say that authors must use it to obey the user's font preferences. Oh, though, we kind of don't care if authors use the more difficult `env(..)` instead to honor the user's preferences. So maybe we leave it non-normative?

davidsgrogan: > So you could keep the Note block to explain how authors are expected to use `scale` in their documents, with the example block below it to explain how, if you set `scale` and don't alter the default font size, content sized with `rem` units will be relative to the preferred text scale.

I'd like to think we included a good example in the explainer we can crib from.
... and we didn't.

Yeah, an example would be good here.

davidsgrogan: Example added.

css-fonts-5/Overview.bs
JoshTumath: Yeah, I'm using the [Bikeshed preview tool](https://api.csswg.org/bikeshed/) and I'm aware of the linkification, but still think it would be useful to say something like:

```
The ''preferred-text-scale'' environment variable value must be 1 on desktop platforms.
```

That matches how it's worded in the css-env-1 spec as well.

css-fonts-5/Overview.bs
JoshTumath: > Yeah, Advisement (you meant "advisement", not "assertion". I think :) boxes are supposed to be normative. This Note seems merely descriptive as is now, but I suppose we can strengthen the language to say that authors must use it to obey the user's font preferences. Oh, though, we kind of don't care if authors use the more difficult `env(..)` instead to honor the user's preferences. So maybe we leave it non-normative?

Sorry yeah, an advisement box. And yes that makes sense to me.

The example is great! 👍🏼

css-fonts-5/Overview.bs
davidsgrogan: Ah, ok, will do.

davidsgrogan: And aside, how do you use that preview tool for this PR? Do you put something into the "Load from URL" box?
Screenshot 2025-11-14 at 1 49 53 PM

davidsgrogan: Done

JoshTumath: Thanks. And yeah I just pasted the URL of the 'raw' version of the BS file on GitHub.

davidsgrogan:
@fantasai I'd love to get another round of feedback on this, or a stamp of approval if it's sufficient as is!

css-fonts-5/Overview.bs
fantasai: Not seeing where you mean, quote?

Anyway I think we're good here now. :)

css-fonts-5/Overview.bs
```diff
@@ -74,6 +73,122 @@ all properties defined in this specification
...
+<h2 id="text-scale-meta">
+Text-Scale <code class=html>&lt;meta&gt;</code> element</h2>
+
+A document with a <{meta}> tag
+whose <{meta/name}> attribute
+is a <a>ASCII case-insensitive</a> match for
+<dfn lt=text-scale><code>"text-scale"</code></dfn>
+is recognized as setting the computed value
+of the ''font-size/medium'' font size
+and consequently scaling the computed size of the other <<absolute-size>> keywords.
```
fantasai: I think instead of this you want to create a concept and say that it controls this concept.

css-fonts-5/Overview.bs
JoshTumath: > Not seeing where you mean, quote?

It's more because of the absence of anything for me to quote. 😅 None of the other standard metadata names define that the tag is ignored if the content is invalid. Like `color-scheme`, for example. https://html.spec.whatwg.org/multipage/semantics.html#meta-color-scheme

It's not a big deal, though. No harm in including the 'otherwise' clause. 🙂

css-fonts-5/Overview.bs
davidsgrogan: I factored the concept into "text scale factor" and refer to that. Is that approximately what you meant?

PR #13069, opened 2025-11-05
[css-values-4] Properly handle simplification of sum nodes containing negated sum nodes #13020

### LINKED ISSUE #13020: [css-values-4] Incomplete simplification of calculation sum node containing inverted sum nodes
Running the [simplify a calculation tree](https://drafts.csswg.org/css-values-4/#simplify-a-calculation-tree) algorithm on a calculation tree containing a sum node which contains an invert node which in turn contains a sum node that can't be simplified down to a single numeric value results in incomplete simplification.

Take for instance the value `calc(100 - (sign(1em - 1px) - 50))` which produces the following calculation tree:
```
SUM:
  NUMERIC(100)
  NEGATE:
    SUM:
      SIGN:
        SUM:
          NUMERIC(1em)
          NUMERIC(-1px)
      NUMERIC(-50)
```

Chrome and Firefox both correctly simplify this to `calc(150 - sign(1em - 1px))` but the existing algorithm results in no simplification. This is because step 6 only simplifies the Negate node if it's child is a Numeric or Negate node, and step 8 only combines the children of the root Sum node which are Numeric or Sum nodes (not negate).

One way to fix this would be that, when simplifying Negate nodes, we convert a negated sum node into a sum of negated nodes i.e. convert
```
NEGATE:
  SUM:
    SIGN:
      SUM:
        NUMERIC(1em)
        NUMERIC(-1px)
    NUMERIC(-50)
```
into
```
SUM:
  NEGATE:
    SIGN:
      SUM:
        NUMERIC(1em)
        NUMERIC(-1px)
  NUMERIC(50)
```
Which would then be further simplified by step 9 producing an output in line with Chrome and Firefox.

I am happy to take a stab at a PR for this if the above seems like a reasonable approach.

Calme1709:
Closes #13020 

A WPT test already exists in the form of css/css-values/calc-nesting-002.html

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -4906,7 +4906,28 @@ Simplification</h4>
 				(0 - value).
 			2. If |root|'s child is a Negate node,
 				return the child's child.
-			3. Return |root|.
+			3. If |root|'s child is a Sum node:
+
+				1. Let |negated grandchildren| be an empty list
+
+				2. For each |grandchild| of the child's children:
+
+					1. If |grandchild| is a numeric value,
+					    create an equivalent numeric value,
+					    but with the value negated
+					    (0 - value),
+					    and append the result to |negated grandchildren|.
+
+					2. If |grandchild| is a Negate node
+					    append |grandchild|'s child to |negated grandchildren|
+
+					3. Otherwise,
+						create a Negate node with |grandchild| as its child,
+						and append the result to |negated grandchildren|
+
+				3. Return a Sum node with |negated grandchildren| as its children
+
+			4. Return |root|.
 
 		7. If |root| is an Invert node:
 
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -1153,7 +1153,7 @@ URL Processing Model</h4>
 			1. If |cssRuleOrDeclaration| is a [=CSS declaration block=] whose [=parent CSS rule=] is not null,
 				set |cssRuleOrDeclaration| to |cssRuleOrDeclaration|'s [=parent CSS rule=].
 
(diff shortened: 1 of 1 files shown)


css-values-4/Overview.bs
```diff
@@ -4906,7 +4906,28 @@ Simplification</h4>
...
+				2. For each |grandchild| of the child's children:
+
+					1. If |grandchild| is a numeric value,
+					    create an equivalent numeric value,
+					    but with the value negated
+					    (0 - value),
+					    and append the result to |negated grandchildren|.
+
+					2. If |grandchild| is a Negate node
+					    append |grandchild|'s child to |negated grandchildren|
```
tabatkins:
```suggestion
						create an equivalent numeric value,
						but with the value negated
						(0 - value),
						and append the result to |negated grandchildren|.

					2. If |grandchild| is a Negate node
						append |grandchild|'s child to |negated grandchildren|
```

tabatkins:
Looks good to me. Didn't think of this, but yes, pushing a negate into a sum makes perfect sense for simplification purposes.

Looks like Firefox is failing this test with an *odd* transformation - it pushes the Negate inward, but then further transforms the `min()` into a `max()` with negated arguments, which isn't supported in the spec.

Afaict, Safari is failing the test because it actually faithfully follows the spec, including for Product simplification: in this case (`1 * (min(...) - 10px)`), none of the simplification cases apply (it's got two children, one of which is a number and the other is a Sum, but that Sum does *not* contain only numeric values). Chrome (and Firefox) are simplifying the case regardless, just in different ways - Chrome is turning it into `Sum(Negate(Min(...)), 10px)`, while Firefox, as I say above, is turning it into `Sum(Max(...), 10px)`.

I'll open an issue about this latter transform.

Issue #1771, opened 2017-08-31
[css-sizing] Auto-resize iframes based on content


tabatkins:
**In <https://lists.w3.org/Archives/Public/www-style/2017Aug/0045.html>, Craig Francis said:**

Just re-raising the suggestion of allowing an iframe or textarea to increase its height based on its content.

Mats Palmgren suggested that I raise this issue again, as Firefox is unlikely to implement it unless the CSSWG agrees to spec it.

For iframes, this feature would allow easy and secure (isolated) embedding of third party content without issues relating to scroll bars (i.e. needing custom JavaScript to ask the parent page, via postMessage, to change the iframe height).

For textrareas, this is just a convenience feature, as trying to get JavaScript to do this is tricky (e.g. the browser can automatically add/remove scroll bars, which can throw off the calculations).

When we discussed this last time (after the @seamless attribute on the <iframe> was removed from the WHATWG spec), it was suggested that we used:

```
iframe,
textrarea {
    height: max-content;
}
```

In the case of the iframe, the child page would need to send a HTTP header (e.g. Expose-Height-Cross-Origin: 1), so it does not leak information about that website (e.g. if the user is logged in).

PS: This has been discussed in a number of times before, the earliest one I've found is 16 years old:

https://bugzilla.mozilla.org/show_bug.cgi?id=80713

And I've written up some notes about this feature, which includes the suggestion of using it for element height animation (e.g. disclosure widgets):

https://github.com/craigfrancis/iframe-height

tomhodgins:
I don't understand why “expand to fit `scrollHeight`” hasn't been the _default_ behaviour for @<textarea>@ all along. In a lot of my demos where I want to add this sort of functionality I achieve this effect by somehow running the following 2 lines of JavaScript from the context of an element:

```javascript
element.style.height = 'inherit'
element.style.height = element.scrollHeight + 'px'
```

I've written a little JavaScript function that can help me apply an auto-height or auto-width to elements that match CSS selectors here that uses this same logic: https://github.com/tomhodgins/reprocss#auto-expand-mixin

Mixin Demo: https://tomhodgins.github.io/reprocss/test/auto-expand-mixin.html

Another Auto-Height demo: https://codepen.io/tomhodgins/pen/KgazaE?q=expand&limit=mine&order=popularity&depth=everything&show_forks=false

[...]

PR #13099, opened 2025-11-12
[css-sizing-4] Add responsively-sized iframes

kojiishi:
Based on the resolutions at:
https://github.com/w3c/csswg-drafts/issues/1771#issuecomment-989034819

### DIFF
--- a/css-sizing-4/Overview.bs
+++ b/css-sizing-4/Overview.bs
@@ -530,7 +530,7 @@ Overriding Contained Intrinsic Sizes: the 'contain-intrinsic-*' properties</h3>
 
 	<pre class=propdef>
 	Name: contain-intrinsic-width, contain-intrinsic-height, contain-intrinsic-block-size, contain-intrinsic-inline-size
-	Value: auto? [ none | <<length [0,∞]>> ] | from-element <<length [0,∞]>>?
+	Value: [ auto | from-element ]? [ none | <<length [0,∞]>> ]
 	Initial: none
 	Inherited: no
 	Logical property group: contain-intrinsic-size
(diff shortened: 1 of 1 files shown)


css-sizing-4/Overview.bs
```diff
@@ -573,6 +573,11 @@ Overriding Contained Intrinsic Sizes: the 'contain-intrinsic-*' properties</h3>
 			<wpt pathprefix="css/css-contain/">
 			content-visibility/content-visibility-058.html
 			</wpt>
+
+		: <dfn>from-element</dfn>
+		:: This value enables the <a>responsively-sized iframes</a>.
```
chrishtr: Add a sentence here saying something like "from-element applies only to `iframe` elements, and applies regardless of the value of the `contain` CSS property". (`contain` has no impact on the `iframe` element regardless)

chrishtr: We also need to add support for the width and height second and third parameters of `from-element`, in order to avoid/reduce CLS when loading a page.

kojiishi: Done.

Also I missed that the `contain-size` shorthand has its own syntax, so it's updated too.

css-sizing-4/Overview.bs
```diff
@@ -573,6 +573,16 @@ Overriding Contained Intrinsic Sizes: the 'contain-intrinsic-*' properties</h3>
 			<wpt pathprefix="css/css-contain/">
 			content-visibility/content-visibility-058.html
 			</wpt>
+
+		: <dfn>from-element
```
tabatkins:
```suggestion
		: <dfn lt="from-element">from-element <<length [0, ∞]>></dfn>
```

tabatkins:
I did some rejiggering of the text, Koji; please let me know if that's okay. It's mostly editorial, the only real change is that I integrated the grammar of the new value into the existing grammar better. Now you specify `auto` *or* `from-element` to get a magic intrinsic size from layout, and either way you have to specify a fallback size.

I also moved the statement about the value applying to iframes regardless of `contain` to an issue, to make it more obvious as a discussion point. If we want to change that behavior, we'll need to slightly rephrase the propdef block's "Applies To" line as well.

Otherwise, the edits are just reworking the text (including rearranging some existing text), and defining a specific term to hook into ("internal layout intrinsic size").

kojiishi:
Thank you @tabatkins, it looks great to me.

PR #13106, opened 2025-11-14
Put back removed defs for animation trigger state values

ydaniv:
Last push these defs were deleted and it's breaking publication.
Putting those back for now so that we can publish a WD, and they can later be revised.


ydaniv:
ping @birtles @flackr - this is needed for republishing WD.

PR #13134, opened 2025-11-21
[css-navigation-1] Add :link-to() pseudo-class (with navigation-param() function

dbaron:
This adds an initial proposal for the :link-to() pseudo-class and a navigation-param() function that goes inside of it.

This also refactors some of the invocations of the URL pattern "match" algorithm to structure the parameters the way the algorithm expects (and be a little bit clearer that a base URL doesn't need to be provided).

### DIFF
--- a/css-navigation-1/Overview.bs
+++ b/css-navigation-1/Overview.bs
@@ -115,9 +115,11 @@ as follows:
 : <<navigation-test>>
 ::	: <<navigation-location>>
 	: at: <<navigation-location>>
-	:: The result is true if
-		the document's [=Document/URL=] [=URL pattern/match|matches=]
-		the [=navigation location URL pattern=] of <<navigation-location>>.
+	:: The result is whether the result of
+		[=URL pattern/match|match a URL pattern=] is non-null
+		given <var>urlPattern</var> as
+		the [=navigation location URL pattern=] of <<navigation-location>>,
+		<var>input</var> as the document's [=Document/URL=].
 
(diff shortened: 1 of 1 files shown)


css-navigation-1/Overview.bs
```diff
@@ -266,7 +280,98 @@ to model it after.
...
+<dfn><<link-condition>></dfn> = <<link-condition-base>> [ with <<navigation-param-expression>> ]?
+<dfn><<link-condition-base>></dfn> = <<url-pattern()>>
+<dfn><<navigation-param-expression>></dfn> = ( [ <<navigation-param-and>> |
+                                    <<navigation-param-or>> |
+                                    <<navigation-param>> ] )
+<dfn><<navigation-param-and>></dfn> = <<navigation-param-expression>>
+                         [ "and" <<navigation-param-expression>> ]*
+<dfn><<navigation-param-or>></dfn> = <<navigation-param-expression>>
+                        [ "or" <<navigation-param-expression>> ]*
+<dfn><<navigation-param>></dfn> = <<navigation-param-key-value>> | <<navigation-param-function>>
```
noamr, line 301: Does this mean that the syntax is `:link-to(route-name with navigation-param(movie-id))` / `:link-to(route-name with movie-id: 123)` or some such?

dbaron: I think it currently requires one more set of `()` for whatever is after the `with`, but otherwise yes.  (This is changeable, though!)

PR #13140, opened 2025-11-22
[css-view-transitions-2] ViewTransition.types should be [SameObject]

emilio:
As it is expected to return the same object all the time.

This matches implementations (and common sense too).


gb-jos:
@emilio @noamr shouldn't it be `readonly` as well ?

emilio:
Yes, good catch, wanna send a patch?

gb-jos:
sure @emilio https://github.com/w3c/csswg-drafts/pull/13466

Issue #12594, opened 2025-08-12
[css-view-transitions-2][css-conditionals] Exposing navigation/route matching


noamr:
Following from a past discussion around navigation-matching for view-transitions in #8925, there is an emerging proposal to expose "routes" (URL patterns) directly in the platform, and allow CSS to respond to them. 

See [route matching explainer](https://github.com/WICG/declarative-partial-updates/blob/main/route-matching-explainer.md), as well as https://github.com/WICG/declarative-partial-updates/issues/46 for discussing the route-map syntax.

The proposal currently is to have a `<script type=routemap>` that gives names to different URL(pattern)s, and then allowing to select them via CSS, for example:
```html
<script type=routemap>
 { "article": "/article/:id" }
</script>

<style>
@route article {
  section[name=article] {
    display: block;
  }
}
</style>
```

Opening this issue for an introductory presentation at the F2F and see if this kind of direction resonates.

bramus:
Looking up some of my old VT syntax explorations, I have this snippet where I define the routes in CSS.

```css
@routes {
  --articles: urlpattern('/articles');
  --article: urlpattern('/articles/:article_id');
}
```

Seems like a nice alternative to the proposed `script type=routemap` for defining the possible routes.

_(The usage of `urlpattern()` in CSS also came up in https://github.com/w3c/csswg-drafts/issues/10975#issuecomment-2578487201)_

bramus:
Oh, I was also reminded of [`@document`](https://developer.mozilla.org/en-US/docs/Web/CSS/@document) which was removed from css-conditional _(and has partial support in Firefox)_.

```css
@document url("https://www.example.com/")
{
  h1 {
    color: green;
  }
}
```

noamr:
> Looking up some of my old VT syntax explorations, I have this snippet where I define the routes in CSS.
> 
> @routes {
>   --articles: urlpattern('/articles');
>   --article: urlpattern('/articles/:article_id');
> }
> Seems like a nice alternative to the proposed `script type=routemap` for defining the possible routes.
> 
> _(The usage of `urlpattern()` in CSS also came up in [#10975 (comment)](https://github.com/w3c/csswg-drafts/issues/10975#issuecomment-2578487201))_

We are likely to use this route definition for things other than styling though... The styling use case is perhaps the first one but not the only one. For example it would allow defining whether a navigation between two routes should create a new navigation entry by default.

noamr:
[...]

PR #13148, opened 2025-11-24
[css-navigation-1] Provide an initial definition of @route rules.

dbaron:
This was discussed briefly in
https://github.com/w3c/csswg-drafts/issues/12594#issuecomment-3530564088 and this PR fills in a proposed definition of such a rule.


dbaron:
Also, for what it's worth, my plan for the next steps is to add a bunch of examples to the spec, now that (once this PR lands) I've filled in all of the pieces needed to write a set of basic examples.

PR #13152, opened 2025-11-25
[css-borders-4][editorial] Fix Bikeshed links to property value range

cdoublev:
Add missing doubled square brackets for Bikeshed links.


SebastianZ:
Thanks for the fix, @cdoublev!

PR #13154, opened 2025-11-25
[css-navigation-1] Add a bunch of examples.

dbaron:
This adds a bunch of examples and hopefully makes the spec a bit more readable.  (I should probably add more in the future, but this is a start.)

(Also fix an edit I missed adding support for named routes declared by @route to :link-to().)

### DIFF
--- a/css-navigation-1/Overview.bs
+++ b/css-navigation-1/Overview.bs
@@ -404,7 +404,7 @@ A simple example of a ''::link-to()'' selector is this one,
 which matches any links that link to the site's homepage:
 
 <pre highlight=css>
-:link-to(/) {
+:link-to(url-pattern("/")) {
   font-weight: bold;
 }
 </pre>


css-navigation-1/Overview.bs
```diff
@@ -350,20 +392,93 @@ ISSUE: This should probably have a more formal definition of the function,
...
 <dfn id="link-to-pseudo" selector>'':link-to()''</dfn> functional pseudo-class
 that matches link elements that link to a certain URL.
 
+<div class="example">
+
+A simple example of a ''::link-to()'' selector is this one,
+which matches any links that link to the site's homepage:
+
+<pre highlight=css>
+:link-to(/) {
```
noamr: Doesn't this need `url-patterm` etc? URL pattern strings shouldn't be hanging loose

css-navigation-1/Overview.bs
```diff
@@ -350,20 +392,93 @@ ISSUE: This should probably have a more formal definition of the function,
...
+
+   The navigation-param(id) further requires that either the
+   from or the to URL of the current navigation also match
+   the URL pattern represented by the "@route
+   --movie-details" rule, and that that the 'id' named group
+   from that match be the same as the 'id' named group from
+   the match with the link's target.
+   */
+.movie-container:has(> .movie-title:link-to(
+  --movie-details with (navigation-param(id)))) {
```
noamr, line 459: Does this work with one less ()?

css-navigation-1/Overview.bs
dbaron: Indeed.  Fixed.

css-navigation-1/Overview.bs
dbaron: Not right now, but that's why I added an "ISSUE" about it.  (It's sort of hard to write the grammar to fix it, but I probably should.)

PR #13160, opened 2025-11-26
[css-navigation-1] Factor out current from/to URL definitions.

dbaron:
This extracts the defintions of current from URL and current to URL from the definition of the `@navigation` rule, which both allows them to be used more formally in the definition of the ::link-to() pseudo-class and also makes the definitions a bit more readable.

This also cleans up some whitespace in the spec.


noamr (approved):
I think we should have an editorial review for this soon, probably use more algorithm steps. Some of these if statements with a lot of conditions are a bit hard to read. 

But lets move quickly and revisit this as we go along.

dbaron:
Yeah, I agree this needs more work -- I had been thinking the next step for this (now that it's better encapsulated in this spec) was to file an HTML issue on how to export the definitions we need so that we're not linking into non-exported terms in HTML.

PR #13161, opened 2025-11-26
[css-tables-3] mark Keith Cirkel (myself) as editor

keithamus:
@gregwhitworth and @FremyCompany are currently listed as editors for this, but are not active in this WG. I'm active in this group and volunteered to be editor for this spec, understanding that the role of the Editor is to regularly triage and resolve issues pertaining to the spec.


FremyCompany:
Thank you @keithamus! I'm indeed way less active now, but feel free to ask questions if you need any help 🙂 Just tag me, I don't pro-actively look for issues anymore (which means it's wonderful to have an active editor again, thank you for volunteering!)

Issue #8685, opened 2023-04-04
[css-view-transitions-2] MPA: 'back' transitions


jakearchibald:
Transitions tend to happen in the opposite direction on 'back' navigations.

With SPA, we recommend looking at state in the navigation API, and setting a class name on the root.

For MPA, it feels like we should have something higher-level:

```css
@media (vt-nav-type: back) {
  /* … */
}
```

We should also look at ways to add that behaviour for SPAs:

```js
document.startViewTransition({
  isBackNavigation: true,
  update: callback,
});
```

Or we could look at a tighter integration with the navigation API.

noamr:
Is this for when the animations are custom, right?
With the default transform/cross-fade-style animations things should "just work" IIUC.
I would like to get a more detailed use case for this issue to see that we're on the same page.

jakearchibald:
Yeah, but cross fade is the default because neutral starting point to build from. It isn't a good transition in of itself.

See the simple shared-axis demo from https://developer.chrome.com/docs/web-platform/view-transitions/#simple-customization - as soon as you introduce a direction into the transition, you want the other direction when going 'back'.

noamr:
Note that this is possible to achieve with a bit of JS even for MPA. Though of course a declarative solution would be much preferable.

For example, this PoC would slide from the top when going forward, from the bottom when going backwards, and fade in when reloading:
```js
	switch (performance.navigation.type) {
		case PerformanceNavigation.TYPE_RELOAD:
			document.documentElement.dataset.navDirection = "reload";
			break;
		case PerformanceNavigation.TYPE_BACK_FORWARD: {
			const prev = sessionStorage.getItem("prev-url");
			console.log(prev, document.referrer)
			document.documentElement.dataset.navDirection = document.referrer === prev ? "forward" : "back";
			break;
		}
		default:
			document.documentElement.dataset.navDirection = "forward";
	}

	sessionStorage.setItem("prev-url", location.href)
```

```css
@keyframes slide-in {
  from { transform: translateY(-100%) }
  to { transform: none}
}

@keyframes slide-out {
  from { transform: translateY(100%) }
  to { transform: none}
}

@keyframes fade-in {
  from { opacity: 0 }
  to { opacity: 1 }
}

::view-transition-group(root) {
  animation-name: slide-in;
  animation-duration: 1s;
}

html[data-nav-direction="back"]::view-transition-group(root) {
  animation-name: slide-out;
  animation-duration: 1s;
}

[...]

PR #13187, opened 2025-12-04
[css-navigation-1] Add definitions of @navigation (history: ...) queries

dbaron:
These are an initial attempt to address the feature request in #8685.

This change also removes the `@navigation (<route-location>)` syntax that was an alternative for `@navigation (at: <route-location>)`, since I think that syntax no longer makes sense with `@navigation` (rather than `@route`) as the name.

### DIFF
--- a/css-navigation-1/Overview.bs
+++ b/css-navigation-1/Overview.bs
@@ -198,10 +198,15 @@ with <<navigation-condition>> defined as:
                      | <<navigation-in-parens>> [ and <<navigation-in-parens>> ]*
                      | <<navigation-in-parens>> [ or <<navigation-in-parens>> ]*
 <dfn><<navigation-in-parens>></dfn> = ( <<navigation-condition>> ) | ( <<navigation-test>> ) | <<general-enclosed>>
-<dfn><<navigation-test>></dfn> = <<navigation-location>> | <<navigation-keyword>> : <<navigation-location>>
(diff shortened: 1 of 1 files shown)


css-navigation-1/Overview.bs
```diff
@@ -198,10 +198,15 @@ with <<navigation-condition>> defined as:
...
-<dfn><<navigation-keyword>></dfn> = at | from | to
+<dfn><<navigation-test>></dfn> = <<navigation-location-test>> | <<navigation-type-test>>
+
+<dfn><<navigation-location-test>></dfn> = <<navigation-location-keyword>> : <<navigation-location>>
+<dfn><<navigation-location-keyword>></dfn> = at | from | to
 <dfn><<navigation-location>></dfn> = <<route-name>> | <<url-pattern()>>
 <dfn><<route-name>></dfn> = <<dashed-ident>>
+
+<dfn><<navigation-type-test>></dfn> = history : <<navigation-type-keyword>>
+<dfn><<navigation-type-keyword>></dfn> = traverse | back | forward | reload
```
noamr, line 209: Should we have a keyword for the normal case? `none` or `navigate` ?

dbaron: What I was proposing here (and in #8685) is that we don't try to solve the push or replace cases for now, i.e., don't try to figure out how to partition them into keywords or name those keywords.  So I think that means "no"... although if you disagree we could try to solve those now.

Issue #11395, opened 2024-12-19
[css-page] Expose unprintable areas via CSS


mstensho:
Most printers have a small region along each edge of the page sheet which is unprintable, due to the printer's paper handling mechanism. See https://drafts.csswg.org/css-page-3/#page-terms

The size of these unprintable areas are available to applications (such as browsers) in most OSes, but currently it's not web-exposed.

This means that authors have no means of confidently setting page margins to prevent content (inside a page margin box, or even inside the page area) from getting clipped.

Maybe we could use env() variables for this? https://drafts.csswg.org/css-env-1/#safe-area-insets seems to be exactly what we need, in a way, although the spec currently says that "[f]or rectangular displays, these must all be zero". Furthermore, some printers have non-uniform unprintable area widths. For instance, my old HP deskjet had an unprintable area of about 1.5cm at the bottom (which is huge, and more than a typical margin), but more reasonable widths on the other three sides (3 millimeters or less). The problem with this non-uniformity is that the values are relative to the direction the sheet of paper travels through the printer, as far as I understand, and the application has no way of telling whether the printer is going to rotate what it receives from the application. E.g. if the page has landscape orientation (e.g. `@page { size: landscape; }`, and the printer is fed with the short edge first (aka portrait), it will have to rotate it. But will it be 90 or 270 degrees? Up to the printer. If the printer can print on both sides, it may also be rotated.

So the only safe thing seems to be: use the larger of all those unprintable area widths - i.e. just one env() variable for this?

Let's pretend it's called `env(unprintable-area-width)` for now:

```css
<!DOCTYPE html>
<style>
  @page {
    margin: calc(env(unprintable-area-width) + 50px);
    margin-bottom: calc(env(unprintable-area-width) + 50px);
    margin-left: calc(env(unprintable-area-width) + 50px);
    margin-right: calc(env(unprintable-area-width) + 50px);

    @top-center {
      content: "";
      border: solid;
      margin-top: env(unprintable-area-width);
    }
    @bottom-center {
      content: "";
      border: solid;
      margin-bottom: env(unprintable-area-width);
    }
    @right-middle {
      content: "";
      border: solid;
      margin-right: env(unprintable-area-width);
    }
    @left-middle {
      content: "";
      border: solid;
      margin-left: env(unprintable-area-width);
    }
    @top-left-corner {
      content: "";
      width: 30px;
      height: 30px;
      border: solid;
      margin-bottom: auto;
      margin-right: auto;
    }
[...]

PR #13190, opened 2025-12-08
[css-page-3] Proposed spec changes for safe printable insets. #11395

mstensho:
See
https://github.com/w3c/csswg-drafts/issues/11395#issuecomment-3546009354

### DIFF
--- a/css-page-3/Overview.bs
+++ b/css-page-3/Overview.bs
@@ -2077,6 +2077,18 @@ Staying within the printable area: the '@page/page-margin-safety' property</h3>
 	Computed Value: as specified
 	</pre>
 
+	<div class=issue>
+		<p>
+			The "For:" field above should also include the at-rules
+			@top-left-corner, @top-left, @top-center, @top-right,
+			@top-right-corner, @right-top, @right-middle,
+			@right-bottom, @bottom-right-corner, @bottom-right,
+			@bottom-center, @bottom-left, @bottom-left-corner,
+			@left-bottom, @left-middle, @left-top.
+		</p>
+		<p>However, multiple at-rules here are currently not supported by bikeshed.</p>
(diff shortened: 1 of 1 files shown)


mstensho:
@fantasai Does this look okay?

faceless2:
```
	Name: page-margin-safety
	For: @page
	Value: none | clamp | add
	Initial: auto
	Computed Value: as specified
```
Based on the example four paragraphs after that clause, wouldn't `For` be "page margin boxes" not `@page`?

mstensho:
> ```
> 	Name: page-margin-safety
> 	For: @page
> 	Value: none | clamp | add
> 	Initial: auto
> 	Computed Value: as specified
> ```
> 
> Based on the example four paragraphs after that clause, wouldn't `For` be "page margin boxes" not `@page`?

It's actually for both `@page` and `@page` margin boxes. How do I express that in this language?

SebastianZ:
> > ```
> > 	Name: page-margin-safety
> > 	For: @page
> > 	Value: none | clamp | add
> > 	Initial: auto
> > 	Computed Value: as specified
> > ```
> >
> > Based on the example four paragraphs after that clause, wouldn't `For` be "page margin boxes" not `@page`?
> 
> It's actually for both `@page` and `@page` margin boxes. How do I express that in this language?

@tabatkins Is it possible to define a descriptor for multiple at-rules at once? Or should they be referenced differently in that case?

Sebastian

mstensho:
I discussed this with @tabatkins , and bikeshed currently doesn't support specifying multiple at-rules here. I pointed this out in an ISSUE in the spec instead, and listed all the additional at-rules where this applies.

mstensho:
Any more feedback? Ready to land?

Issue #12022, opened 2025-03-28
[css-grid-3] Masonry Switch Syntax


fantasai:
In https://github.com/w3c/csswg-drafts/issues/11243 we ended up resolving to:
- Re-use the `grid-*` properties for templating and explicit placement.
- Adopt the `item-flow` proposal (in principle) for flow control. See https://github.com/w3c/csswg-drafts/issues/11480

With that context, I'm opening this issue to focus on one specific sub-point of the syntax debate: how do we trigger masonry layout?

The three proposals so far are:
- Use a new `display` value. **Adopted but needs name bikeshedding, see below.**
- <del>Use `display: grid` + `grid-template-rows/columns: collapse`.</del>
- Use `display: grid` + `item-pack: collapse` (allowing it to show up in the `item-flow` shorthand as .e.g. `item-flow: row collapse`).

Now that `item-flow` has been adopted, nobody is advocating for the `grid-template-rows/columns` option. But people wanted some time to absorb the new `item-flow` situation before deciding. See https://www.w3.org/TR/css-grid-3/#flow-control

New ideas are also welcome.

**UPDATE:** The CSSWG resolved to use the `display` property together with a value that includes the word `grid`. Currently looking for proposals on what that name should be. It could be `grid-something`, `something-grid`, or `grid something`/`something grid` (two keywords). **[SEE THIS NAMING POLL](https://forms.gle/8qPfodfKqkNdWhQ79)**.

Names proposed so far include:

- `collapsed-grid` or `grid collapse`
- `grid-stack` or `stacked-grid`
- `grid-pack` or `packed-grid` or `grid pack`
- `compact-grid` or `compact grid`
- `grid-masonry` or `masonry-grid`
- `grid-flex` or `flex-grid`
- `grid single-axis` (masonry grid) vs `grid dual-axis` (normal grid)
- `grid stack` (masonry grid) vs `grid grid` (normal grid) -- or likewise with another keyword
- `staggered-grid` or `grid staggered`
- `uneven-grid`
- `semi-grid`
- `lane-grid` or `grid-lanes`
- `axial-grid`

alisonmaher:
After analyzing all recent inputs on the syntax switch thus far, my opinion remains unchanged that `display` makes the most sense as the trigger. 
 
**Re: `display` as the trigger** 
 
[...]

Issue #10961, opened 2024-09-27
[css-grid-3] Should there be an `inline-grid-lanes` legacy keyword for `display`?


keithamus:
As per title; I haven't seen examples of `display: inline masonry`. I also think should ensure to include the legacy `inline-masonry` keyword if we're using `display`.

Loirooriol:
We didn't add a legacy `inline-list-item` (dropped in #1495), `block-ruby` nor `block-math` (#5385), you need to provide 2 keywords for these combinations.

So I'm not convinced that we need `inline-masonry`, authors can just use `inline masonry` instead.

bfgeek:
> So I'm not convinced that we need `inline-masonry`, authors can just use `inline masonry` instead.

We should add it as a convenience (there isn't any reason not to). E.g. most developers will use just `inline-grid` and not `inline grid`, and would be unnecessarily surprised if `inline-masonry` isn't supported. We shouldn't make this more difficult unnecessarily.

Ian

keithamus:
`list-item` is special and `block` is assumed; the only legacy variants are `inline-` prefixed. While
I’m indifferent about adding `inline-masonry` I was largely pointing it out for completeness otherwise IIRC it would be the only inside display value without an inline legacy variant.

Loirooriol:
We shouldn't call these "legacy" if we are going to keep adding them whenever we add new inside display types.

Loirooriol:
> it would be the only inside display value without an inline legacy variant

Actually not, Houdini Layout adds [`layout()`](https://drafts.css-houdini.org/css-layout-api/#valdef-display-layout) but not `inline-layout()`.
It was removed by Ian in https://github.com/w3c/css-houdini-drafts/commit/60ff1c223a8eca286212fb9c6fb292344af7716a

Loirooriol:
Renamed since now it's `grid-lanes` instead of `masonry`.

Agenda+ to decide whether to add a "legacy" `inline-grid-lanes` keyword.

SebastianZ:
I've strongly in favor of _not_ introducing a new legacy keyword. That allows to push the awareness of the double-keyword syntax of `display`.

Sebastian

jensimmons:
Making these work:
```
display: inline-flex;
display: inline-grid;
display: inline flex;
display: inline grid;
display: inline grid-lanes;
```
and making this broken:
```
display: inline-grid-lanes
```

is not a good way to help developers understand what's happening. 

[...]

PR #13191, opened 2025-12-08
[css-grid-3] Update CSS-Grid-3 spec to use new grid-lanes syntax

stwrt:
https://github.com/w3c/csswg-drafts/issues/12022 resolved to declaring masonry using display: grid-lanes. We need to update the spec to match.

### DIFF
--- a/css-grid-3/Overview.bs
+++ b/css-grid-3/Overview.bs
@@ -297,7 +297,7 @@ Reordering and Accessibility</h3>
 			</section>
 			<style>
 			.masonry {
-				//FIXME display: something;
+				display: grid-lanes;
 				grid-template-columns: repeat(5, auto);
 			}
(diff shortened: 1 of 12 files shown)


css-grid-3/Overview.bs
```diff
@@ -348,28 +348,25 @@ Reordering and Accessibility</h3>
...
 
-	<div class="informative">
-		<pre class="propdef partial">
-		Name: display
-		New values: something
-		</pre>
-	</div>
+	<pre class="propdef partial">
+	Name: display
+	New values: grid-lanes | inline-grid-lanes
```
Loirooriol, line 353: `inline-grid-lanes` was not part of the resolution. Please don't ship it in WebKit until #10961 is resolved.

Issue #13221, opened 2025-12-15
[fill-stroke-3][editorial] Image of figure 8 not shown in Firefox


SebastianZ:
The [image to figure 8 of Fill & Stroke](https://drafts.csswg.org/fill-stroke/#:~:text=Four%20types%20of%20line%20joins) is not shown in Firefox.

<img width="901" height="706" alt="Image" src="https://github.com/user-attachments/assets/a5b8522b-1701-4dda-8c00-c8c4c49b6bfd" />

In Chrome it _is_ shown, though it doesn't respect dark mode:

<img width="895" height="663" alt="Image" src="https://github.com/user-attachments/assets/98de7a59-b364-448a-9946-931e30361b6c" />

So the image should be fixed that it works everywhere and looks fine in light and dark themes.

Sebastian

myakura:
That figure works in Safari (26.2).

Looks like the culprit is the path values; each segment is joined by a comma, which should not be (should be just a whitespace).
https://github.com/w3c/csswg-drafts/blob/e92ed204434023199e6ea08da43e573b0343e920/fill-stroke/Overview.bs#L830-L845

Browsers behave differently with such error values.

<img width="800" alt="Screenshot: inspecting invalid path in Chrome DevTools" src="https://github.com/user-attachments/assets/8e94b87c-5b90-42be-9ec3-8522e68c2800" />

<img width="800" alt="Screenshot: inspecting invalid path in Firefox Developer Tools" src="https://github.com/user-attachments/assets/c7b1efe3-a8a2-4f31-83a7-33de688ad841" />

> it doesn't respect dark mode

I guess that's because it's an inline SVG but a separate SVG file?

SebastianZ:
Thanks for the hint, @myakura! I previously reported it without looking for the reason.
With your info, I've now fixed both myself.

Sebastian

PR #13222, opened 2025-12-15
[fill-stroke] fix SVG paths

myakura:
The embedded SVG in Figure 8 uses invalid path syntax which separates each segment by a comma. This removes those commas.

#13221


SebastianZ:
@myakura Oops! I didn't see you also already created a PR for this. Sorry!
But thanks a lot for the hint!

Sebastian

PR #13225, opened 2025-12-16
[pointer-animations-1] Link `source` and `target` back to right type

tidoust:
The values are for the `<timeline-range-center-subject>` type but the wrapping `<>` were missing.


ydaniv:
Thanks, @tidoust!

PR #13227, opened 2025-12-16
[css-forms-1] Use border-radius: 50% instead of 100%

fantasai:
100% relies on the fixup rules to scale down to 50%, so maybe we should just say 50% directly since that's what we mean?

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -880,7 +880,7 @@ input:is([type=checkbox]:not([switch]), [type=radio]) {
 }
 
 input[type=radio] {
-    border-radius: 100%;
+    border-radius: 50%;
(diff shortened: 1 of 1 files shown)


josepharhar:
I haven't prototyped base appearance radio inputs yet, but I don't imagine there will be any issues adding this to the UA stylesheet for base appearance when we get around to that

lukewarlow:
> I haven't prototyped base appearance radio inputs yet, but I don't imagine there will be any issues adding this to the UA stylesheet for base appearance when we get around to that

I have a chromium prototype patch for buttons that I'll try and fix up over christmas and then checkbox and radio are my next targets.

nt1m (approved):

Issue #8843, opened 2023-05-15
[scroll-animations-1] Should `none` be part of the list in scroll/view-timeline-name?


andruud:
The `scroll-timeline-name` and `view-timeline-name` properties have this grammar: `none | <custom-ident>#`. That is fine in isolation, but  it's inconsistent with `animation-name`, which has `[ none | <custom-ident> ]#`, and that made me wonder if it's intentional.

If it _is_ intentional, then feel free to close this.

@fantasai

ydaniv:
I think it's intentional since these are timeline declarations that can be matched by definitions of `animation-timeline`'s and are matched by name, and not by position in the list. So it's either a list of names or `none`.

tabatkins:
Yeah, `animation-name`'s use of `none` as any item in the list (which just "turns off" that entry in the list) is pretty odd actually.

andruud:
OK, in that case, how am I supposed to interpret the grammar of `scroll-timeline`:

>`[ <'scroll-timeline-name'> [ <'scroll-timeline-axis'> || <'scroll-timeline-attachment'> ]? ]#`

I'm aware of css-values-4:

> `These types do not include CSS-wide keywords such as inherit, and also do not include any top-level comma-separated-list multiplier (i.e. if a property named pairing is defined as [ <custom-ident> <integer>? ]#, then <'pairing'> is equivalent to [ <custom-ident> <integer>? ], not [ <custom-ident> <integer>? ]#).`

But `scroll-timeline-name` doesn't have top-level comma-separated-list multipliers.

Even if you ignore that, the grammar would suggest that `scroll-timeline: none, none, none` is allowed, even though it's not allowed by the longhand?

> Yeah, animation-name's use of none as any item in the list (which just "turns off" that entry in the list) is pretty odd actually.

Yet if `scroll-timeline` worked like that it seems to answer all my questions, so maybe not that odd. :-)

fantasai:
Hm, yes, it does seem we have a problem. :) I think the two options are:

- Allow `none` as an alternative to `<custom-ident>` in `*-timeline-name`, essentially disables that entry in the coordinated properly list.
- Rework the grammars explicitly so that both `*-timeline-name` and the `*-timeline` shorthand have `none | <whatever>`.

ydaniv:
ah, for coordinating between longhand lists, I see.
[...]

PR #13231, opened 2025-12-17
[scroll-animations-1] Clarify "Computed value" of scroll/view-timeline-name

### LINKED ISSUE #13139: [scroll-animations-1] Clarification on the computed value of `scroll-timeline-name` and `view-timeline-name`
The properties [`scroll-timeline-name`](https://drafts.csswg.org/scroll-animations-1/#propdef-scroll-timeline-name) and [`view-timeline-name`](https://drafts.csswg.org/scroll-animations-1/#propdef-view-timeline-name) both have the grammar:

```
[ none | <dashed-ident> ]#
```

but state their computed value should be "the keyword `none` or a list of CSS identifiers".

What does that mean the computed value should be in cases like:

```css
scroll-timeline-name: none, none;
```

or 

```css
scroll-timeline-name: --foo, none, --bar;
```

The text "the keyword `none` or a list of CSS identifiers" seems to imply either a single keyword `none` OR a list of identifiers, but when there is more than one use `none`, I'm not sure what should be used.

andruud:
The current text would suggest a grammar of `none | <dashed-ident>#`, but the grammar is actually `[none | <dashed-ident>]#`. We probably just forgot to update the text in the edit for #8843.

Resolves #13139.

### DIFF
--- a/scroll-animations-1/Overview.bs
+++ b/scroll-animations-1/Overview.bs
@@ -527,7 +527,7 @@ spec:selectors-4; type:dfn; text:selector
 	Initial: none
 	Applies to: all elements
 	Inherited: no
-	Computed value: the keyword ''scroll-timeline-name/none'' or a list of [=CSS identifiers=]
+	Computed value: list, each item either a [=CSS identifier=] or the keyword ''scroll-timeline-name/none''
 	Animation type: not animatable
 	</pre>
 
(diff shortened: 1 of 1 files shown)


andruud:
PTAL. Inspired by [`animation-name`](https://drafts.csswg.org/css-animations-1/#animation-name).

andruud:
cc @weinig

PR #13237, opened 2025-12-18
[css-values-5] Add explainer for 'revert-rule'

andruud:


lilles (approved):

css-cascade-5/revert-rule-explainer.md
```diff
@@ -0,0 +1,79 @@
...
+based on a custom property:
+
+```css
+div.basic {
+  display: block;
+}
+div.fancy {
+  display: if(style(--enable-fancy): fancy);
+}
+```
```
yisibl, line 33: Could you add an example of `var()`? I think that would make it easier to understand.

Issue #12581, opened 2025-08-07
[css-animations-2] animation-trigger Name Clash Resolution


DavMila:
The current direction of #12336 is that animations are associated with triggers via names (dashed idents). I’m filing this issue so that we can resolve on 2 questions related to name clashes:

1. How should name clashes between elements be resolved?
For example:
```
#source1 {
 event-trigger-name: --trigger;
}
#source2 {
 event-trigger-name: --trigger;
}
#target {
 animation-trigger: --trigger;
}
```
Should `#target` get the trigger on `#source1` or on `#source2`?
In theory we could specify that the above CSS attaches 2 triggers to #target’s animation. However, 1. that's probably not in line with the way dashed idents typically work & 2. we’ve resolved in #12399 to allow only a single trigger per animation to start - as such, `#target` needs to pick one trigger.

Some precedents:
- The anchor position pattern: among elements with the same anchor-name, [the last in tree-order is selected](https://drafts.csswg.org/css-anchor-position/#:~:text=Return%20the%20last%20element%20el%20in%20tree%20order).
- The timeline-scope pattern: where multiple timelines have the same name within a particular scope, that name is considered to [refer to an inactive timeline](https://drafts.csswg.org/scroll-animations-1/#:~:text=declares%20an%20inactive%20timeline).

I propose we go with the anchor position pattern because I think it is more author-friendly - an author could use it to "override" a trigger declared higher up in the DOM. The downside of this choice is that an author could accidentally have their animation linked with the wrong trigger. However, I think this problem will be significantly mitigated when we introduce a (to-be-bikeshed) `trigger-scope` property which limits visibility of trigger names similar to [anchor-scope](https://drafts.csswg.org/css-anchor-position-1/#anchor-scope). I also think that from an author's perspective, it is no worse than if the trigger name was simply ruled invalid (the timeline-scope approach).

2. How should name clashes between timeline-trigger and event-trigger be resolved?
```
#source {
  event-trigger: --trigger;
  timeline-trigger: --trigger;
}
#target {
  animation: ...;
  animation-trigger: --trigger;
}
```
Does #target’s animation get an EventTrigger or a TimelineTrigger?

I think our options are to:
[...]

PR #13241, opened 2025-12-18
[css-animations-2] Define trigger scope

DavMila:
We [resolved](https://github.com/w3c/csswg-drafts/issues/12581#issuecomment-3206707173) to have a name scoping property for animation triggers and have it behave like anchor-scope.

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -813,7 +813,7 @@ of what can be achieved with the Web Animations API in Javascript,
 allowing simple, common interaction patterns
 to be created and managed purely in CSS.
 
-Currently, two types of <dfn export for=CSS>triggers</dfn> are defined:
+Currently, two types of <dfn export for=CSS lt="trigger">triggers</dfn> are defined:
 
 * [=timeline triggers=], managed by the 'timeline-trigger' properties,
     which allow animations to be triggered by entering or leaving certain timeline ranges.
@@ -826,7 +826,7 @@ Currently, two types of <dfn export for=CSS>triggers</dfn> are defined:
     such as clicking an element or pressing certain keys.
 
 A [=trigger=] is <em>defined</em> on some specific triggering element.
-All triggers have a name,
+All triggers have a <dfn export for=CSS>trigger name</dfn>,
 and the specific type of trigger dictates how and when it's activated.
 A trigger can define multiple "types" of activation.
 (For example, [=timeline triggers=] can do different things on entry and exit.)
@@ -844,20 +844,6 @@ is intentionally somewhat generic,
 intended to support using [=triggers=] for <em>other</em> purposes in the future.
 For now, though, [=triggered animations=] are the only user of this feature.
 
-
-<h3 id="trigger-name-scoping">
-Trigger Name Scoping</h3>
-
-A <dfn export>trigger-instantiating property</dfn> specifies the details for
-instantiating an [=Animation Trigger=] along with a name for the instantiated
-trigger, a <dfn export>trigger name</dfn>.
-
-All [=trigger names=] are document-global by default but may be
-scoped by 'trigger-scope'.
-
-ISSUE: Is "document-global" the right term here given that these "global" names are
-still tree scoped?
-
 If a single element attempts to define multiple [=triggers=] of different types
 with the same [=trigger name=],
 it only exposes one of the [=triggers=],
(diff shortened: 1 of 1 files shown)


css-animations-2/Overview.bs
```diff
@@ -845,29 +845,80 @@ intended to support using [=triggers=] for <em>other</em> purposes in the future
...
+<h4 id="trigger-scope">
+Scoping Trigger Names: the 'trigger-scope' property</h4>
+
+<pre class=propdef>
+Name: trigger-scope
+Value: none | all | <<dashed-ident>>#
+Initial: none
+Applies to: all elements
+Inherited: no
+Animation type: discrete
```
flackr: Given that resolving to a different trigger could result in starting an animation, I think this may need to be "not animatable".

css-animations-2/Overview.bs
```diff
@@ -845,29 +845,80 @@ intended to support using [=triggers=] for <em>other</em> purposes in the future
...
+Computed value: as specified
+</pre>
+
+This property scopes [=trigger names=] to the subtree of the matching element.
+
+The values have same meaning as with the [[css-anchor-position-1#anchor-scope|anchor-scope]]
```
vmpstr: I don't think you need this paragraph. The description stands on its own without anchor-scope references

css-animations-2/Overview.bs
```diff
@@ -845,29 +845,80 @@ intended to support using [=triggers=] for <em>other</em> purposes in the future
...
+property. Thus, the following description of their meaning is similar to the anchor-scope text:
```
flackr: I think this line can be removed.

css-animations-2/Overview.bs
```diff
@@ -845,29 +845,80 @@ intended to support using [=triggers=] for <em>other</em> purposes in the future
...
+<dl dfn-for="trigger-scope" dfn-type=value>
+  <dt><dfn>none</dfn>
+  <dd>
+    No changes in [=trigger name=] scope.
+
+  <dt><dfn>all</dfn>
+  <dd>
+    Specifies that all [=trigger names=] defined by this element or its descendants--
+    whose scope is not already limited by a descendant using 'trigger-scope'--
+    to be in scope only for this element's descendants;
```
vmpstr:
```suggestion
    to be in scope only for this element's flat-tree descendants;
```

css-animations-2/Overview.bs
```diff
@@ -845,29 +845,80 @@ intended to support using [=triggers=] for <em>other</em> purposes in the future
...
+    as if it were a [=tree-scoped name/strictly matched=] [=tree-scoped name=].
+    (That is, ''trigger-scope: all'' acts identically
+    to ''trigger-scope: --foo, --bar, ...'',
+    listing all relevant [=trigger names=].)
+
+  <dt><dfn><<dashed-ident>></dfn>
+  <dd>
+    Specifies that a matching [=trigger name=] defined by this element or its descendants--
+    whose scope is not already limited by a descendant using 'trigger-scope'--
+    to be in scope only for this element's descendants;
```
vmpstr:
```suggestion
    to be in scope only for this element's flat-tree descendants;
```

css-animations-2/Overview.bs
```diff
@@ -845,29 +845,80 @@ intended to support using [=triggers=] for <em>other</em> purposes in the future
...
 
-All [=triggers=] are document-global by default,
-similar to <l spec="css-anchor-position-1">[=anchor names=]</l>.
+A <dfn export>trigger-instantiating property</dfn> specifies the details for
+instantiating an [=Animation Trigger=] along with a name for the instantiated
+trigger, a <dfn export>trigger name</dfn>.
+
+All [=trigger names=] are document-global by default,
+similar to <l spec="css-anchor-position-1">[=anchor names=]</l>, but may be
+scoped by [[#trigger-scope| trigger-scope]].
```
flackr: This line is confusing, anchor names can also be scoped. I think simplest would be to just remove it. If there is any cross referencing, I could see the anchor scope spec calling out how unlike timeline-scope and trigger-scope it has additional restrictions.

DavMila: Sounds good. I kept the "document-global" part as I don't think that's called out elsewhere... or is that covered by the "tree-scoped" references?

css-animations-2/Overview.bs
```diff
@@ -845,29 +845,78 @@ intended to support using [=triggers=] for <em>other</em> purposes in the future
...
+Trigger Name Scoping</h3>
 
-All [=triggers=] are document-global by default,
-similar to <l spec="css-anchor-position-1">[=anchor names=]</l>.
+A <dfn export>trigger-instantiating property</dfn> specifies the details for
+instantiating an [=Animation Trigger=] along with a name for the instantiated
+trigger, a <dfn export>trigger name</dfn>.
+
+All [=trigger names=] are document-global by default but may be
+scoped by [[#trigger-scope| trigger-scope]].
```
tabatkins:
```suggestion
scoped by 'trigger-scope'.
```

PR #13243, opened 2025-12-18
[css-animations-2] Add non-normative accessibility considerations for animations

flackr:
These are similar to the accessibility considerations when using transitions: https://www.w3.org/TR/css-transitions-1/#accessibility

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -1496,7 +1496,7 @@ Users who are sensitive to movement,
 or who require additional time to read or understand content,
 may benefit from user agent features that allow
 animations to be disabled or slowed down.
-(But see [[#privacy]] for information on the privacy implications
+(see [[#privacy]] for information on the privacy implications
 of such mitigations.)
 
 User agent implementors should be aware that Web content


css-animations-2/Overview.bs
```diff
@@ -1448,15 +1448,69 @@ console.log(anim.playState); // Should be 'running'.
...
+if such modifications are needed to meet a user's accessibility needs.
+
+Authors are strongly encouraged to use the '@media/prefers-reduced-motion' media query to provide alternative
+reduced motion or instant animations when the user has indicated a preference to ''@media/prefers-reduced-motion/reduce'' motion.
+
+Users who are sensitive to movement,
+or who require additional time to read or understand content,
+may benefit from user agent features that allow
+animations to be disabled or slowed down.
+(But see [[#privacy]] for information on the privacy implications
```
vmpstr:
```suggestion
(see [[#privacy]] for information on the privacy implications
```

css-animations-2/Overview.bs
```diff
@@ -1448,15 +1448,69 @@ console.log(anim.playState); // Should be 'running'.
...
+of such mitigations.)
+
+User agent implementors should be aware that Web content
+may depend on the firing of <a href="#events">animation events</a>,
+so implementations of such mitigations may wish to fire animation events
```
vmpstr:
```suggestion
may depend on the firing of <a href="#events">animation events</a>.
As such, implementations of such mitigations may wish to fire animation events
```

css-animations-2/Overview.bs
```diff
@@ -1448,15 +1448,69 @@ console.log(anim.playState); // Should be 'running'.
...
+<h2 id="accessibility">
+Accessibility Considerations</h2>
+
+<em>This section is not normative.</em>
+
+<h3 id="accessibility-motion">
+Motion</h3>
+
+This specification provides declarative mechanisms for animations
+that previously needed to be done using script.
```
vmpstr, line 1486: This is a bit revealing of the spec history. It could also be that WAAPI came after css animations, but it doesn't change the content of the section. Not sure if there's some more neutral wording. This is also ok as is

flackr: WAAPI did come well after CSS animations, the "script" in this context is raf / setinterval loops. I think it's fine to get this section started.

PR #13252, opened 2025-12-19
[css-view-transitions-2] Reorder sections, add issues, change nested wording a bit; other editorial edits

vmpstr:


vmpstr:
PTAL. FYI there are a couple of bikeshed errors when I build this, but those were there before I did the changes. Not sure how to fix them just yet

PR #13273, opened 2025-12-29
[css-mixins-1][editorial] Added WPTs

SebastianZ:
I added all the Web Platform Tests that currently cover the features in CSS Mixins 1.

Note that this PR does not include hints about missing WPT coverage or notes about tests not being necessary for specific sections. Those should be added separately.

Also note that there are a few tests referring to "locally substituting a var", which got replaced at some point. So I've added them to "Evaluating Custom Functions". Please let me know if that's ok!

Furthermore, with this change I've added a Changes section. That section should be updated to include all the other changes that were made since the FPWD (though is not part of this PR).

Sebastian


SebastianZ:
A previous version of this PR in #13247 already got approved by @svgeesus. Though that PR was corrupted. So I created this one. And since it already got the approval, I am going to merge this one.

Sebastian

PR #13278, opened 2025-12-30
[css-env-1][editorial] Discuss media queries in meta text-scale explainer

davidsgrogan:


css-env-1/explainers/meta-text-scale.md
```diff
@@ -143,6 +143,33 @@ That means:
 
 We expect `scale` to become best practice for authors to use on all new website designs, just as they use the viewport meta tag. It allows authors to continue to use font-relative units like `rem` and `em` like they normally would and mostly avoid using `env(preferred-text-scale)`.
```
davidsgrogan, line 145: Josh, WDYT of adding a before/after section here that shows what `<meta text-scale>` helps with? I think the section would basically comprise slides 5&6 from the pdf you e-mailed on 19 Dec. (Though maybe reversed to show the "FAIL" state of the world as it is currently?)

If so, is that something you're up for doing, given that you already have the graphics for it?

davidsgrogan (commented):
Oops, didn't mean to make this comment "a review" but whatevs.

css-env-1/explainers/meta-text-scale.md
JoshTumath: @davidsgrogan Done. 👍🏼 I've added them in the comparison section rather than line 145, because it felt like they made more sense there, but feel free to position them somewhere else if you feel that would be better.

PR #13293, opened 2026-01-05
[cssom-1] Validate and normalize CSSStyleSheetInit.baseURL

### LINKED ISSUE #12974: [cssom-1] Should `CSSStyleSheetInit.baseURL` be validated?
[`baseURL`](https://drafts.csswg.org/cssom-1/#dom-cssstylesheetinit-baseurl) is a `CSSStyleSheet.constructor()` argument that allows setting the [style sheet base URL](https://drafts.csswg.org/cssom-1/#concept-css-style-sheet-stylesheet-base-url), which is:

  > The base URL to use when resolving relative URLs in the stylesheet. Null by default. Only non-null for stylesheets that have constructed flag set.

Chrome always ignores it ([WPT](https://wpt.fyi/results/css/cssom/CSSStyleSheet-constructable-baseURL.html?label=master&label=experimental&aligned&q=css%2Fcssom%2FCSSStyleSheet-constructable-baseURL.html)). Firefox seems to handle an absolute URL or a relative URL ending with `/`, eg. `/css/`, `./css/`, `css/`, but `/css`, `./css`, `css`, are ignored.

Besides, [WPT](https://github.com/web-platform-tests/wpt/blob/master/css/cssom/CSSStyleSheet-constructable-baseURL.html) expects `CSSStyleSheet.constructor()` to throw `NotAllowedError` for an *invalid base URL*, which seems unspecified, noting that the style sheet base URL is used to [*compute a style resource base URL*](https://drafts.csswg.org/css-values-4/#style-resource-base-url), while this procedure presumably expects an absolute URL. Should it be absolutized?

cdoublev:
Fixes #12974.

### DIFF
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -1167,7 +1167,7 @@ URL Processing Model</h4>
 	</div>
 
 	<div algorithm="resolve a style resource URL">
-		To <dfn>resolve a style resource URL</dfn> from a [=/url=] or <<url>> |urlValue|,
+		To <dfn export>resolve a style resource URL</dfn> from a [=/url=] or <<url>> |urlValue|,
 		and a [=CSS rule=] or a [=CSS declaration block=] |cssRuleOrDeclaration|:
 			1. Let |base| be the [=style resource base URL=] given |cssRuleOrDeclaration|.
 			1. Return the result of the [=URL parser=] steps with |urlValue|'s [=/url=] and |base|.
--- a/css-values-4/Overview.bs
+++ b/css-values-4/Overview.bs
@@ -1167,7 +1167,7 @@ URL Processing Model</h4>
 	</div>
 
 	<div algorithm="resolve a style resource URL">
-		To <dfn export>resolve a style resource URL</dfn> from a [=/url=] or <<url>> |urlValue|,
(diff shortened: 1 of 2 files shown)


Gae24:
Hi, I stumbled upon this while adding support to CSS modules, from reading the specification [resolve a style resource URL](https://drafts.csswg.org/css-values-4/#resolve-a-style-resource-url) takes as input a [CSS rule](https://drafts.csswg.org/cssom-1/#css-rule) or a [CSS declaration block](https://drafts.csswg.org/cssom-1/#css-declaration-block) and neither of them are a  `CSSStyleSheet`.
Instead the `CSSStyleSheet` is retrieved inside [compute the style resource base URL](https://drafts.csswg.org/css-values-4/#style-resource-base-url), and the _base URL_ will be one of this values:
```
- If sheet’s stylesheet base URL is not null, return sheet’s stylesheet base URL. (but at this point is still null)
- If sheet’s location is not null, return sheet’s location.
- Return cssRuleOrDeclaration’s relevant settings object’s API base URL.
```
So I think this should be simplified to:
`Let absolute baseURL be the result of the URL parser steps with baseURL and sheet’s location.` 
(Since location is setted to the base URL of the associated Document for the current global object.)

cdoublev:
You are right. I am not sure exactly what I had in mind by using these algos. Thanks!

PR #13296, opened 2026-01-05
[css-animations-2] Add trigger-scope explainer

DavMila:
This adds an explainer for the trigger-scope property.

### DIFF
--- a/css-animations-2/TRIGGER_SCOPE.md
+++ b/css-animations-2/TRIGGER_SCOPE.md
@@ -1,0 +1,114 @@
+# Explainer for trigger-scope
+
+This explainer is a description of the `trigger-scope` CSS property which provides a
+scoping mechanism for animation trigger-related names.
+
+## Summary
+Animation triggers created in CSS are given names through which those triggers
+can be matched with animations. For example `timeline-trigger` declares a
+timeline-based trigger which can then be matched with an animation via the
+`animation-trigger` property. Here is an example:
+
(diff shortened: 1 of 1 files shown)


css-animations-2/TRIGGER_SCOPE.md
```diff
@@ -0,0 +1,114 @@
...
+`target` in the previous example. Here is an example:
+
+```CSS
+@keyframes slide-in {
+  from { transform: translateX(-50px) }
+  to { transform: translateX(0xp) }
+}
+
+.section {
+  timeline-trigger: --trigger view() contain;
```
flackr, line 47: It would be far more compelling to show an example where the trigger source is a different element than the one animating. In theory this sort of use case would be possible with anonymous triggers. E.g. maybe the animating element is a descendant of the trigger, or a sibling to it.

css-animations-2/TRIGGER_SCOPE.md
```diff
@@ -0,0 +1,114 @@
...
+modify their code to the following:
+
+```CSS
+@keyframes slide-in {
+  from { transform: translateX(-50px) }
+  to { transform: translateX(0px) }
+}
+
+.section {
+  trigger-scope: all;
```
flackr, line 72: It would probably be better for a canonical example to be explicit about the trigger being scoped here, i.e. `trigger-scope: --trigger`

Issue #11814, opened 2025-03-04
[css-gaps] Gap decorations with collapsed gutters


Loirooriol:
https://drafts.csswg.org/css-grid/#collapsed-gutter

> When a [collapsed track](https://drafts.csswg.org/css-grid/#collapsed-grid-track)’s gutters collapse, they coincide exactly—​the two gutters overlap so that their start and end edges coincide.

Does this mean that the gap decorations will overlap? That doesn't sound great.

kbabbitt:
Interesting. Naively, I'd have expected the two gutters adjacent to a collapsed track to consolidate down to a single gutter rather than overlapping each other.

I suppose one use case for the spec'ed behavior might be where you have varying gap decorations over the grid, and if a track changes state from collapsed to un-collapsed or vice-versa, you want gap decorations in subsequent tracks to "stay with" those tracks. But then that means either you have overlapped decorations, or you need to define which one "wins" like we do for collapsing table borders.

Loirooriol:
This is also somewhat related to #10813. My conclusion there was that a collapsed track gets assigned the size of minus one gutter (and that's why the gutters overlap). But that's not great for decorations, so if we do something else, we need to think how abspos "items" will be affected.

fantasai:
CSS2.1 collapsed border conflict resolution rules: https://www.w3.org/TR/CSS2/tables.html#border-conflict-resolution

oSamDavis:
After further consideration, I’d like the CSSWG to revisit the resolution regarding coincident gap decorations caused by collapsed tracks. The current resolution states that all but one decoration should be suppressed when tracks collapse, using border-collapsing rules to determine which decoration “wins.”

While I understand the intent behind this approach, I believe it introduces unnecessary complexity and reduces predictability. Border-collapsing rules were designed for tables, where competing border styles are visually apparent and hierarchical precedence makes sense. Applying those rules to grid gaps feels less intuitive to web authors. Authors also lose direct control over which gap decorations appear in visible gutters.

Additionally, I believe collapsed tracks are somewhat of an implementation detail of `auto-fit`. Most authors use `auto-fit` to fill available space, not to necessarily reason about collapsed tracks. They typically don’t expect these internal mechanics to influence visual outcomes.

### Proposed resolution
Gap decorations should map cleanly to _visible_ gutters in the order they are specified. This approach is simpler, predictable, and gives authors explicit control.

### Example
Consider:
```css
[...]

PR #13298, opened 2026-01-05
[css-gaps-1] Update gap decoration behavior when gaps are coincident due to track collapsing #11814

oSamDavis:
This PR updates the logic of assigning gap decorations to gaps that are coincident due to track collapsing. Previously, border-collapsing rules were used to resolve conflicts. Now, following the new resolution taken in #11814, gap decorations are applied after collapsing, treating all coincident gaps as a single visible gap for decoration purposes.

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -308,7 +308,7 @@ Layout and painting</h3>
 		grid/subgrid/fragmentation/subgrid-gap-decorations-fragmentation-018.html
 	</wpt>
 
-	gap decorations are [=assigned gap decoration values|assigned=] after collapsing has been applied.
+	Gap decorations are [=assigned gap decoration values|assigned=] after [=gutter collapsing|collapsed gutter=] has been applied.
 	A given set of [=collapsed gutters=] consume exactly one [=gap decoration=]; the next [=gap decoration=]
 	is applied to the next gutter (or set of [=collapsed gutters=]). Therefore, [=collapsed gutters=] are treated as a
 	single gutter for decoration purposes.


css-gaps-1/Overview.bs
```diff
@@ -309,9 +309,10 @@ Layout and painting</h3>
 	</wpt>
 
 	If the positions of one or more gaps coincide exactly due to being [=collapsed gutter|collapsed=],
-	at most one decoration will be drawn for that set of gaps.
-	The decoration that is drawn is selected according to the
-	<a href="https://www.w3.org/TR/CSS2/tables.html#border-conflict-resolution">border conflict resolution</a> rules for tables.
+	gap decorations are [=assigned gap decoration values|assigned=] after collapsing has been applied.
+	Collapsed gutters do not consume any gap decoration, instead the next gap decoration in specified order
```
```suggestion
	A given set of [=collapsed gutters=] consume exactly one [=gap decoration=]; the next [=gap decoration=]
```

1. Not necessarily "in specified order" since we could be looping back in a repeater, for example.
2. [Grid 1](https://www.w3.org/TR/css-grid-1/#gutters) describes gutter collapsing as: "the two gutters overlap so that their start and end edges coincide." It's not that one or the other gutter is considered "collapsed" or becomes "non-visible" but rather that they all overlap.

css-gaps-1/Overview.bs
```diff
@@ -309,9 +309,10 @@ Layout and painting</h3>
...
+	is applied to the next visible gutter. Therefore, coincident gaps due to collapsing are treated as a
```
```suggestion
	is applied to the next gutter (or set of [=collapsed gutters=]). Therefore, [=collapsed gutters=] are treated as a
```

css-gaps-1/Overview.bs
```diff
@@ -309,9 +309,10 @@ Layout and painting</h3>
...
+	single visible gutter for decoration purposes.
```
```suggestion
	single gutter for decoration purposes.
```

css-gaps-1/Overview.bs
```diff
@@ -309,9 +309,10 @@ Layout and painting</h3>
 	</wpt>
 
 	If the positions of one or more gaps coincide exactly due to being [=collapsed gutter|collapsed=],
-	at most one decoration will be drawn for that set of gaps.
-	The decoration that is drawn is selected according to the
-	<a href="https://www.w3.org/TR/CSS2/tables.html#border-conflict-resolution">border conflict resolution</a> rules for tables.
+	gap decorations are [=assigned gap decoration values|assigned=] after collapsing has been applied.
```
```suggestion
	Gap decorations are [=assigned gap decoration values|assigned=] after [=gutter collapsing|collapsed gutter=] has been applied.
```

css-gaps-1/Overview.bs
```diff
@@ -309,9 +309,10 @@ Layout and painting</h3>
 	</wpt>
 
 	If the positions of one or more gaps coincide exactly due to being [=collapsed gutter|collapsed=],
```
```suggestion
```

I think we can drop this line of text and instead have the first sentence read as I've suggested below.

kbabbitt (commented):
Thanks! Looks good overall; made a few wording suggestions to tighten up precision.

kbabbitt:
Just thought of something else. Do we have any test cases covering the collapsing behavior yet, and if so could we add them to a `<wpt>` element near this paragraph? (If we don't have any test cases for this yet, I wouldn't block the PR on having them - we can add them later.)

oSamDavis:
> Just thought of something else. Do we have any test cases covering the collapsing behavior yet, and if so could we add them to a `<wpt>` element near this paragraph? (If we don't have any test cases for this yet, I wouldn't block the PR on having them - we can add them later.)

I don't think we have a set of wpts explictly testing these. But I'll create a crbug to add them so we don't lose track of this.

Issue #12784, opened 2025-09-11
[css-gaps-1] Gap intersections for multicol containers


jav099:
**Background**  
We resolved [csswg-drafts#12084](https://github.com/w3c/csswg-drafts/issues/12084) to allow intersections at the border of spanners when gaps are present. This enabled us to _not_ paint decorations behind spanners, which is the current `column-rule` behavior.

**Open question**  
For containers using `column-wrap: wrap`, what default behavior should gap decorations have? The [reference diagram](https://github.com/w3c/csswg-drafts/issues/11976#issuecomment-2753981259) is helpful context.

We see two plausible approaches:

### Option 1 — Intersections at multicol line boundaries
Intersections are placed at the start/end of each multicol line. Decorations do **not** span the gaps created by `column-wrap` unless the author opts in (e.g., via `rule-outset`).

Example:  
<img width="1005" height="537" alt="Image" src="https://github.com/user-attachments/assets/3ab43d64-c429-4cb0-b6a9-d7558114f5dd" />

### Option 2 — Intersections at the center of row gap/column gap crossings
Intersections are placed at the midpoint where a row gap crosses a column gap. Decorations effectively “bridge” row gaps by default—similar to how a grid might behave—so rules continue across those breaks. While visually coherent, this may be less aligned with multicol semantics.

Example:  
<img width="834" height="560" alt="Image" src="https://github.com/user-attachments/assets/88aaa85c-73a1-449c-a28e-b2de6cafd705" />

It's worth to note, that you can achieve `Option 1` visual behavior even if `Option 2` is the default, by using `column-rule-break: intersection` and `column-rule-outset: 0`.

---

**Request for feedback**  
We’d appreciate input, especially from folks steeped in multicol and from web authors, on which of these should be the default behavior:

- Should decorations stop at `column-wrap` breaks unless explicitly extended (Option 1)?  
- Or should they span across by default, meeting at row/column gap centers (Option 2)?  

Thanks in advance!

jav099:
@rachelandrew @mstensho, since both of you are heavily involved in multicol I was wondering if we could get your opinion on this matter. Thanks!

mstensho:
Option 2 is pretty, well, at least as long as all rows have all columns present (maybe even then?). Is this why you're asking? I mean, I think we should have some good reasons for not behaving the same in grid and multicol.

jav099:
[...]

Issue #11491, opened 2025-01-13
[css-gaps-1] Bikeshedding names for ways to break gap decorations


kbabbitt:
Authors may want one of several different behaviors for drawing gap decorations in a Grid that contains spanning items. The spec defines [the `column-rule-break` and `row-rule-break` properties](https://drafts.csswg.org/css-gaps-1/#break) (and shorthand `gap-rule-break`) to select between these behaviors, with one of three values. (In the examples below, I've also set `gap-rule-outset` to 0 to emphasize the difference between them.)

They might want decorations to stop at each intersection:
```css
gap-rule-break: intersection;
gap-rule-outset: 0px;
```
<img src="https://drafts.csswg.org/css-gaps-1/images/example-break-intersection-grid.png" alt="Gap decoration example in CSS Grid with decorations stopping at every intersection">

Or at each T intersection:
```css
gap-rule-break: spanning-item;
gap-rule-outset: 0px;
```

<img src="https://drafts.csswg.org/css-gaps-1/images/example-break-spanning-item-grid.png" alt="Gap decoration example in CSS Grid with decorations stopping at every T intersection">

Or only at the edges of the container:
```css
gap-rule-break: none;
```

<img src="https://drafts.csswg.org/css-gaps-1/images/example-break-none-grid.png" alt="Gap decoration example in CSS Grid with decorations crossing under spanning items">

Questions:
	1. Are there suggestions for better names for these values?
	2. Currently in the spec, the initial value selects the "T intersection" behavior. Is that what most authors would expect as a default?

denmch:
It seems intuitive to me to use `short` and `long` in place of `intersection` and `spanning-item`, understood as something like “the gap rule break here respects the short edge” or vice vera when areas of different sizes meet.

jav099:
Since we are considering moving away from the notion of **`intersections`** (which are largely `grid`-specific) and toward gap-decoration **`segments`** (described further in <https://github.com/w3c/csswg-drafts/issues/12784#issuecomment-3514411109>), the current meaning of **`break-rules`** becomes less clear.

Today we have these values:

- `rule-break: none`
- `rule-break: intersection`
- `rule-break: spanning-item`

In the *intersection* model, decorations are contiguous, so these names made sense.  
However, in the *segment* model, decoration segments are already conceptually “broken up,” so the terminology no longer aligns well.

![Image](https://github.com/user-attachments/assets/73c1f62f-42c2-40e5-973a-3ea2a237803e)

[...]

PR #13299, opened 2026-01-05
[css-gaps-1] Change decorations to be segments based rather than intersections based #12784

jav099:
This PR moves gap decorations to be "segments" based rather than "intersections" based. TODO add more context.

Live preview: https://jav099.github.io/spec-host/

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -575,7 +575,7 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
 
 	Sets the behavior for breaking decorations within a given gap into segments
 	at visible "T" or "cross" intersections formed by items in the container.
-	Precise details for how to apply these values are given in the steps to [=determine candidate segments=].
+	Precise details for how to apply these values are given in the steps to [=determine painted segments=].
 
 	<dl dfn-type="value" dfn-for="column-rule-break, row-rule-break">
 		<dt><dfn>none</dfn>
@@ -643,8 +643,8 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
 		parsing/rule-break-valid.html
 	</wpt>
 
-<h4 id="determining-candidate-gap-segments">
-Determining Candidate Gap Segments From Break Rules</h4>
+<h4 id="determining-painted-gap-segments">
+Determining Gap Segments From Break Rules</h4>
 
 	Recall that a <a>gap decoration segment</a>, |s|, can be represented as:
 	|s| = [|s|.start, |s|.end]
@@ -660,7 +660,7 @@ Determining Candidate Gap Segments From Break Rules</h4>
 	the gap that span across all perpendicular [=gaps=] contributing to that junction.
 
 	<div algorithm>
-		To <dfn>determine candidate segments</dfn> |C| from segments list |S|:
+		To <dfn>determine painted segments</dfn> |C| from segments list |S|:
 
 		<ol>
 			<li>Let |S| be the list of segments <a lt="determine the gap decoration segments">determined</a> for the given gap, ordered from start to end along the gap's axis.


css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+		<dd>
+			[=row gap=]s and [=column gap=]s
+			refer to the [=gutters=] between <a>grid rows</a> and <a>grid columns</a>,
+			respectively.
+			See [[css-grid-1#gutters]] for precise details.
+	</dl>
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/grid-gaps.png">
```
kbabbitt, line 100: There's some imprecision in this image; the intersecting areas look like they extend too far upward, and the second column gap is not orange all the way to the left. Can these be cleaned up?

css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+			<figcaption>Gaps in a grid container (in orange and blue).</figcaption>
```
```suggestion
			<figcaption>
			    Gaps in a grid container; [=column gaps=] in orange and [=row gaps=] in blue).
			    Note the shaded areas where gaps overlap.
			</figcaption>
```

css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+		</figure>
+	</div>
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/flex-gaps.png">
+			<figcaption>Gaps in a flex container (in orange and blue).</figcaption>
```
```suggestion
             <figcaption>
			    Gaps in a flex container; [=column gaps=] in orange and [=row gaps=] in blue).
			    Note that, unlike in grid, gaps in flex do not overlap.
			</figcaption>
```

css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+		</figure>
+	</div>
+
+	Note: It is important to note that [=column gap=]s in Multicol containers break when encountering a spanner element or a row gap.
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/multicol-gaps.png">
+			<figcaption>Gaps in a multicol container with `column-wrap: wrap` (in orange and blue).</figcaption>
```
```suggestion
			<figcaption>
			    Gaps in a multicol container with `column-wrap: wrap`; [=column gaps=] in orange and [=row gaps=] in blue).
			    Note that [=column gaps=] in multicol containers do not overlap [=row gaps=], similar to flex.
			</figcaption>
```

css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+			<img alt="" src="images/flex-gaps.png">
+			<figcaption>Gaps in a flex container (in orange and blue).</figcaption>
+		</figure>
+	</div>
+
+	Note: It is important to note that [=column gap=]s in Multicol containers break when encountering a spanner element or a row gap.
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/multicol-gaps.png">
```
kbabbitt, line 120: It wasn't immediately obvious to me looking at this image that it's a multicol container. I might suggest adding some text that shows clearly which column boxes flow into each other.

This image also has some precision issues with where the rightmost two column boxes are placed.

css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+			<figcaption>Gaps in a multicol container with `column-wrap: wrap` (in orange and blue).</figcaption>
+		</figure>
+	</div>
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/multicol-gaps-2.png">
```
kbabbitt, line 130: Same suggestion as above re: adding text to the image to illustrate flow. I'd also replicate thet heavier border from the above image into this one - where dashed borders do and don't appear helps cue that these are column boxes.

css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+			<figcaption>Column gaps (orange) in a multicol container with a spanner element.</figcaption>
```
```suggestion
			<figcaption>
			    [=Column gaps=] in orange in a multicol container with a spanner element.
			    Note that [=column gaps=] break when they encounter the spanner.
			</figcaption>
```

css-gaps-1/Overview.bs
```diff
@@ -50,13 +50,75 @@ Value Definitions</h3>
...
+	</div>
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/flex-gaps.png">
+			<figcaption>Gaps in a flex container (in orange and blue).</figcaption>
+		</figure>
+	</div>
+
+	Note: It is important to note that [=column gap=]s in Multicol containers break when encountering a spanner element or a row gap.
```
kbabbitt: Having this note in the middle of a string of examples makes it easy to miss. I'd suggest moving it after the paragraph on multi-column containers ("In the column direction, gap (column gap) refers to...")

css-gaps-1/Overview.bs
```diff
@@ -110,80 +167,107 @@ Layout and painting</h3>
...
 	<ul>
-		<li>The center of the start or end edge of a [=gutter=].</li>
-		<li>The center of an intersection between gaps in different directions.</li>
+		<li><dfn>start segment endpoint</dfn>: Point at which a segment begins.
+		<li><dfn>end segment endpoint</dfn>: Point at which a segment ends.
 	</ul>
 
-	The following examples illustrate the locations of <a>gap intersection points</a>,
-	marked with a magenta <span style="color:#f0f">+</span> in each case.
+	Endpoints are positioned on the centerline of the gap they belong to and are created at the following junctions:
```
```suggestion
	[=Segment endpoints=] are created at the following locations:
```

(we already mentioned they're on the centerline)

css-gaps-1/Overview.bs
```diff
@@ -110,80 +167,107 @@ Layout and painting</h3>
...
+			two endpoints are created which align with the boundaries of the cross/perpendicular gap:
+			<ul>
+				<li>The earlier boundary creates an <a>end segment endpoint</a>.
+				<li>The later boundary creates a <a>start segment endpoint</a>.
+			</ul>
+			<p class="note">This ordering (end segment endpoint, start segment endpoint) reflects that the junction corresponds to
+			the end of one track/item/column-box and the beginning of the next.
+	</ol>
+	The following examples illustrate the locations of <a>segment endpoints</a>,
+	with a cyan <span style="color:#0ff">+</span> for <a> start segment endpoints</a> and a
```
kbabbitt: This cyan does not contrast well against the gray backgrounds of some of the example diagrams.

css-gaps-1/Overview.bs
```diff
@@ -110,80 +167,107 @@ Layout and painting</h3>
...
+			  border: 1px dashed black;
+			}	
+		</pre>
+		<figure>
+			<img alt="" src="images/multi-col-segment-endpoints.png">
+			<figcaption>Locations of <a>segment endpoints</a> in a multi-column container.</figcaption>
+		</figure>
+	</div>
+
+	<p class="note">NB: Should I mention different layout types and their gap behaviors here? Or is it implied? From preceeding sections. Feel Like I should for emphasis.
```
kbabbitt: Does "different layout types and their gap behaviors here" mean "why segment endpoints are placed where they are"? If so - yes, I'd vote for explaining this in each example, similar to the suggestions I left for the gap illustration examples above.

css-gaps-1/Overview.bs
```diff
@@ -110,80 +167,107 @@ Layout and painting</h3>
...
+		<li>Bounds of a Gap:
+			<p>For each gap, endpoints are created at its two bounds:
+			<ul>
+				<li>At the start of the gap, a <a>start segment endpoint</a> is created.
+				<li>At the end of the gap, an <a>end segment endpoint</a> is created.
+			</ul>
+			<p class="note">When a gap runs to the content boundary, these endpoints coincide with the corresponding content edge.
+
+		<li>Within a Gap:
+			<p>When a cross/perpendicular gap touches a given gap (either overlapping it, or abutting it without overlapping),
```
kbabbitt: > either overlapping it, or abutting it without overlapping

I think this is only the case where gaps overlap each other, correct?

In flex and multicol, segment endpoint creation where column gaps abut row gaps should be covered by case 1, since the column gap itself ends (or starts). Or am I missing something?

css-gaps-1/Overview.bs
```diff
@@ -110,80 +167,107 @@ Layout and painting</h3>
...
+  			  background: rgb(255 255 128);
 			}
 			.grid > * {
-				border: 1px dashed black;
+  			  border: 1px dashed black;
 			}
 		</pre>
 		<figure>
-			<img alt="" src="images/example-gap-intersection-grid.png">
+			<img alt="" src="images/example-grid-segment-endpoints.png">
```
kbabbitt, line 256: It's a little hard to tell what the segment endpoints at the edges of the container are supposed to line up with. My first thought would be (a) switch the grid items to solid borders, and (b) used dashed lines to show how the endpoints line up with each other, but there might be a better way to make this more apparent.

css-gaps-1/Overview.bs
```diff
@@ -110,80 +167,107 @@ Layout and painting</h3>
...
 			<figcaption>
-				Locations of <a>gap intersection points</a> in a grid with spanning items.
-				Note the presence of <a>gap intersection points</a>
-				even where a spanning item touches the edge of a container,
-				and where an intersection is flanked by spanning items on either side
-				(for example between #2 and #6, or #4 and #7).
+				Locations of <a>segment endpoints</a> in a grid with spanning items.
```
kbabbitt: I'd suggest retaining some of the text from the previous description, so that we call out the presence of segment endpoints even at the edges of the container, and even where the gap is crossed by a spanning item.

css-gaps-1/Overview.bs
```diff
@@ -308,52 +392,178 @@ Layout and painting</h3>
...
 
-	If the positions of one or more gaps coincide exactly due to being [=collapsed gutter|collapsed=],
-	at most one decoration will be drawn for that set of gaps.
-	The decoration that is drawn is selected according to the
-	<a href="https://www.w3.org/TR/CSS2/tables.html#border-conflict-resolution">border conflict resolution</a> rules for tables.
+	<h4 id="segments">
+	Gap decoration segments</h4>
+
+	A <dfn>gap decoration segment</dfn> is formed by pairing a [=start segment endpoint=] with an [=end segment endpoint=].
+	Each [=gap=] is a collection of <i>n</i> [=gap decoration segment=]s, where <i>n > 0</i>.
```
```suggestion
	Each [=gap=] forms a collection of [=gap decoration segments=].
```

I think the number of segments in a given gap can be 0 when the gap is completely occluded by a spanning item.

Note also that you don't need to leave the 's' for a plural out of the linkified text; Bikeshed is smart enough to recognize common variants of defined terms. See https://speced.github.io/bikeshed/#conjugating

css-gaps-1/Overview.bs
```diff
@@ -308,52 +392,178 @@ Layout and painting</h3>
...
+							</li>
+						</ol>
+					</li>
+				</ol>
+			</li>
+		</ol>
+	</div>
+
+	
+	Note: This should result in <strong>no</strong> overlapping [=gap decoration segment=]s in |S|.
```
```suggestion
```

Don't think this note is needed.

css-gaps-1/Overview.bs
```diff
@@ -308,52 +392,178 @@ Layout and painting</h3>
...
+	Each [=gap=] is a collection of <i>n</i> [=gap decoration segment=]s, where <i>n > 0</i>.
+
+	<div algorithm>
+		To <dfn>determine the gap decoration segments</dfn> for a given [=gap=] |G|:
+		<ol>
+			<li>
+				Let |S|, the collection of [=gap decoration segment=]s for |G|, be an empty list.
+			</li>
+			<li>
+				Let |endpoints| be the list of <a>segment endpoint</a>s that fall along the centerline of |G|,
```
```suggestion
				Let |endpoints| be the list of <a>segment endpoint</a>s created for |G|,
```

To keep terminology consistent with the text in 3.1.1. In fact, it might be good to structure the content of that section related to where endpoints are positioned as a "create segment endpoints for a gap" algorithm, and then linking the word "created" here to that algorithm.

css-gaps-1/Overview.bs
```diff
@@ -308,52 +392,178 @@ Layout and painting</h3>
...
+				ordered from start to end along |G|'s axis.
```
```suggestion
				ordered from start to end.
```

We haven't defined "axis" as a part of a gap yet, and I don't think we need to.

css-gaps-1/Overview.bs
```diff
@@ -308,52 +392,178 @@ Layout and painting</h3>
...
+	<div class="example">
+	<figure>
+			<img alt="" src="images/flex-negative-gap-decoration-segments-2.png">
+			<figcaption>
+				Illustration of <a>gap decoration segment</a>s (blue rectangles) in a flex container, where we end up with <a>negative gap decoration segment</a>s.
+				The segments are drawn between pairs of <a>segment endpoint</a>s,
+				with <a>start segment endpoint</a>s marked with cyan crosses and <a>end segment endpoint</a>s marked with pink crosses.
+
+				In this example there is one pair of ignored <a>segment endpoint</a>s (which form a <a>negative gap decoration segment</a>).
+				These are marked with green and black crosses, these used to be a <a>end segment endpoint</a> and <a>start segment endpoint</a> respectively.
```
kbabbitt: I'd suggest leaving the colors of these consistent with the other segment endpoints. I think that would make it easier to follow the consistent pattern of where end and start segment endpoints are generated relative to flex items, and easier to mentally step through the segment algorithm and see why no segment is generated in that location.

css-gaps-1/Overview.bs
```diff
@@ -308,52 +392,178 @@ Layout and painting</h3>
...
+			<figcaption>
+				Illustration of <a>gap decoration segment</a>s (blue rectangles) in a multicol container (with a spanner element).
+				The segments are drawn between pairs of <a>segment endpoint</a>s,
+				with <a>start segment endpoint</a>s marked with cyan crosses and <a>end segment endpoint</a>s marked with pink crosses.
+			</figcaption>
+		</figure>
+	</div>
+
+	Note: In the end, there may remain ignored [=segment endpoint=]s in |endpoints| (or in |starts|). 
+	These result in <dfn>negative gap decoration segment</dfn>s.
```
kbabbitt: The way the segment algorithm is written, I actually don't think we need this defintion. (But if we do for some reason that I'm missing, it should be clear in the algorithm itself how a negative gap decoration segment gets generated.)

css-gaps-1/Overview.bs
```diff
@@ -308,52 +392,178 @@ Layout and painting</h3>
...
+				Note that no decoration will be painted for these segments, so we don't see any blue rectangle between these two points.
+			</figcaption>
+		</figure>
+	</div>
 
-<h3 id="break">
-Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break', and 'rule-break' properties</h3>
 
+<h3 id="join">
+Joining gap decoration segments: The 'column-rule-join', 'row-rule-join', and 'rule-join' properties</h3>
```
kbabbitt: Let's take the switch from "break" to "join" as a separate proposal/PR related to https://github.com/w3c/csswg-drafts/issues/11491. I'd like to keep this PR scoped to a set of changes we can propose for the switch from intersection points to segments.

css-gaps-1/Overview.bs
```diff
@@ -596,8 +759,129 @@ Pairing gap intersection points into segments</h4>
 		</figure>
 	</div>
 
+<h3 id="visibility-rules">
+Painting Gap Decorations with Visibility Rules</h3>
```
kbabbitt: I'd also like to take this as a separate PR.

css-gaps-1/Overview.bs
```diff
@@ -619,32 +903,30 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
-	These properties can be used to offset the endpoints of <a>gap decorations</a> relative to the
-	<a>gap intersection points</a> which would normally determine their endpoints.
+	These properties can be used to offset the start and end points of <a>gap decorations</a> relative to the
+	<a>segment endpoint</a>s which would normally determine where decorations start and end.
 
 	For the interior properties, 
-	a value of ''-50%'' places the <a>gap decoration</a> endpoint in the center of the intersection,
-	and a value of ''0'' places the <a>gap decoration</a> endpoint at the edge of the intersection.
+	a value of ''-50%'' stretches the <a>gap decoration</a> up to the center of the intersection that is formed by the gaps,
+	and a value of ''0'' places the <a>gap decoration</a> start and end points at the [=start segment endpoint=] and [=end segment endpoint=] of the [=gap decoration segment=].
```
kbabbitt: I would rewrite this paragraph as:

```
A value of ''0'' places the [=gap decoration=] start and end points exactly on the [start segment endpoint=] and [=end segment endpoint=], respectively, of the [=gap decoration segment=].
For the interior properties,
because percentages refer to the [=crossing gap width=],
a value of ''-50%'' places the [=gap decoration=] endpoint exactly in the middle of the adjoining intersection.
```

1. I think 0 is probably the easier case to understand in the new segment world, so I'd talk about that first.
2. I'd avoid using the word "stretches" - that could be misinterpreted as some sort of image manipulation in a future where we have gap decoration images.

css-gaps-1/Overview.bs
```diff
@@ -1574,6 +1856,7 @@ Changes since the <a href="https://www.w3.org/TR/2025/WD-css-gaps-1-20250417/">1
 			(<a href="https://github.com/w3c/csswg-drafts/issues/12603">Issue 12603</a>,
 			<a href="https://github.com/w3c/csswg-drafts/issues/12848">Issue 12848</a>,
 			<a href="https://github.com/w3c/csswg-drafts/issues/8402">Issue 8402</a>)
+	</ul>
```
kbabbitt: This looks like a misplaced end tag. Merge artifact?

kbabbitt (commented):
Thanks for working on this! Left a round of comments, though I'm not positive I caught everything.

One other note: I fetched the branch, ran Bikeshed on this file, and noticed a number of new warnings. I'd like to keep this file Bikeshed clean with the exception of warnings for new WPTs that haven't been added yet.

css-gaps-1/Overview.bs
oSamDavis: Done... but haven't pushed updated commit so not resolving yet .

css-gaps-1/Overview.bs
oSamDavis: Done. Leaving open as I haven't uploaded commit with edit

css-gaps-1/Overview.bs
oSamDavis: Done. Leaving open as I haven't uploaded commit with edit

css-gaps-1/Overview.bs
oSamDavis: Yes I think you might be missing the case where within a main direction (say row) gap in flex, we'll want to capture two endpoints within such a gap when the cross gap abuts it.

css-gaps-1/Overview.bs
oSamDavis: Okay changing to the classic purple and green

css-gaps-1/Overview.bs
oSamDavis: Okay, opting for having solid borders for the items and the edge endpoints sort of having a chevron showing the direction they "flow" in. I'd rather this than a line that might hint towards segments at this point... Let me know what you think

css-gaps-1/Overview.bs
oSamDavis: Done... Added some more info to each caption

css-gaps-1/Overview.bs
oSamDavis: ditto

css-gaps-1/Overview.bs

oSamDavis:
Live preview: https://osamdavis.github.io/css-spec-host/

oSamDavis:
Live preview here: https://osamdavis.github.io/css-spec-host/

@kbabbitt, I think this is ready for another pass

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
-	The following examples illustrate the locations of <a>gap intersection points</a>,
-	marked with a magenta <span style="color:#f0f">+</span> in each case.
+	<div algorithm="create segment endpoints for a gap">
+		To <dfn>create segment endpoints for a gap</dfn> |G|:
+		<ol>
+			<li>Bounds of a Gap:
+				<p>Create endpoints at |G|'s two bounds:
+				<ul>
+					<li>At the start of |G|, create a <a>start segment endpoint</a>.
+					<li>At the end of |G|, create an <a>end segment endpoint</a>.
```
```suggestion
					<li>At the centerpoint of |G|'s start edge, create a <a>start segment endpoint</a>.
					<li>At the centerpoint of |G|'s end edge, create an <a>end segment endpoint</a>.
```

css-gaps-1/Overview.bs
```diff
@@ -50,24 +50,89 @@ Value Definitions</h3>
...
-<!-- Big Text: Decorations 
-
-████▌  █████▌  ███▌   ███▌  ████▌   ███▌  █████▌ ████  ███▌  █    █▌  ███▌ 
-█▌  █▌ █▌     █▌  █▌ █▌  █▌ █▌  █▌ ▐█ ▐█    █▌    ▐▌  █▌  █▌ █▌   █▌ █▌  █▌
-█▌  █▌ █▌     █▌     █▌  █▌ █▌  █▌ █▌  █▌   █▌    ▐▌  █▌  █▌ ██▌  █▌ █▌    
-█▌  █▌ ████   █▌     █▌  █▌ ████▌  █▌  █▌   █▌    ▐▌  █▌  █▌ █▌▐█ █▌  ███▌ 
-█▌  █▌ █▌     █▌     █▌  █▌ █▌▐█   █████▌   █▌    ▐▌  █▌  █▌ █▌  ██▌     █▌
-█▌  █▌ █▌     █▌  █▌ █▌  █▌ █▌ ▐█  █▌  █▌   █▌    ▐▌  █▌  █▌ █▌   █▌ █▌  █▌
-████▌  █████▌  ███▌   ███▌  █▌  █▌ █▌  █▌   █▌   ████  ███▌  █▌   ▐▌  ███▌ 
+<!-- Big Text: GAPS
```
kbabbitt, line 53: Please run `bikeshed source` locally to generate the Big Text into the spec source file.

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+					<li>At the start of |G|, create a <a>start segment endpoint</a>.
+					<li>At the end of |G|, create an <a>end segment endpoint</a>.
+				</ul>
+				<p class="note">When a gap runs to the content boundary, these endpoints coincide with the corresponding content edge.
+
+			<li>Within a Gap ([=Junctions=]):
+				<p>A <dfn>junction</dfn> is the region along |G|'s centerline
+				from the point where a cross gap appears on at least one side of |G|
+				to the point where a cross gap ceases to exist on both sides of |G|.
+				Two endpoints are created for each [=junction=]:
```
```suggestion
				Two [=segment endpoints=] are created for each [=gap junction=]:
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+				<ul>
+					<li>At the start of |G|, create a <a>start segment endpoint</a>.
+					<li>At the end of |G|, create an <a>end segment endpoint</a>.
+				</ul>
+				<p class="note">When a gap runs to the content boundary, these endpoints coincide with the corresponding content edge.
+
+			<li>Within a Gap ([=Junctions=]):
+				<p>A <dfn>junction</dfn> is the region along |G|'s centerline
+				from the point where a cross gap appears on at least one side of |G|
+				to the point where a cross gap ceases to exist on both sides of |G|.
```
```suggestion
				<p>A <dfn>gap junction</dfn> is a region of |G|
				where one or more other [=gaps=] abut or overlap |G|.
```

1. Qualify "junction" to avoid overloading it with the same word in other contexts, should we need it elsewhere.
2. For maintainability, avoid repetition with the statements of where segment endpoints are placed below.

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+				Two endpoints are created for each [=junction=]:
+				<ul>
+					<li>Entering a [=junction=]: At the point where a cross gap appears on at least one side of |G|, create an <a>end segment endpoint</a>.
+					<li>Exiting a [=junction=]: At the point where a cross gap ceases to exist on both sides of |G|, create a <a>start segment endpoint</a>.
```
```suggestion
					<li>Entering a [=gap junction=]: At the point where a cross gap appears on at least one side of |G|, create an <a>end segment endpoint</a>.
					<li>Exiting a [=gap junction=]: At the point where no cross gap exists on either side of |G|, create a <a>start segment endpoint</a>.
```

Tightening up the latter definition to remove ambiguity about whether it's the first cross gap or the last cross gap ending that defines the endpoint.

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+				</ul>
+				<p class="note">This ordering (end segment endpoint, start segment endpoint) reflects that the [=junction=] corresponds to
+				the end of one track/item/column and the beginning of another.
+		</ol>
+	</div>
+	The following examples illustrate the locations of <a>segment endpoints</a>,
+	with a purple <span style="color:#72C">+</span> for <a> start segment endpoints</a> and a
+	green <span style="color:#285">+</span> for <a> end segment endpoints</a> in each case.
+	[=Junctions=] are shown with as a dashed box.
```
```suggestion
	[=Junctions=] are shown with a dashed box.
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
 		<figure>
-			<img alt="" src="images/example-gap-intersection-grid.png">
+			<img alt="" src="images/example-grid-segment-endpoints.png">
 			<figcaption>
-				Locations of <a>gap intersection points</a> in a grid with spanning items.
-				Note the presence of <a>gap intersection points</a>
-				even where a spanning item touches the edge of a container,
-				and where an intersection is flanked by spanning items on either side
-				(for example between #2 and #6, or #4 and #7).
+				Locations of <a>segment endpoints</a> in a grid with spanning items.
```
```suggestion
				Locations of <a>segment endpoints</a> and [=gap junctions=] in a grid with spanning items.
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+				These are created at the bounds of each gap (which coincide with the edges of the container in this case),
```
```suggestion
				[=Segment endpoints=] are created at the bounds of each gap (which coincide with the edges of the container),
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+				and within each gap in an end-start manner where the row and column gaps overlap/intersect.
```
```suggestion
				and within each gap in an end-start manner where the row and column gaps overlap.
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
 		</pre>
 		<figure>
-			<img alt="" src="images/example-gap-intersection-flex.png">
-			<figcaption>
-				Locations of <a>gap intersection points</a> in a flexbox.
-				Note the presence of <a>gap intersection points</a> at edges of the container,
-				and the presence of two <a>gap intersection points</a> in close proximity where
-				items in adjacent rows don't quite line up with each other (between #2, #3, #4, and, #5).
+			<img alt="" src="images/flex-and-segment-endpoints.png">
+			<figcaption>Locations of <a>segment endpoints</a> in a flex container.
```
```suggestion
			<figcaption>Locations of <a>segment endpoints</a> and [=gap junctions=] in a flex container.
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+			}
+
+			.multi-column > * {
+			  height: 100px;
+			  border: 1px dashed black;
+			}	
+		</pre>
+		<figure>
+			<img alt="" src="images/multi-col-segment-endpoints.png">
+			<figcaption>Locations of <a>segment endpoints</a> in a multi-column container.
```
```suggestion
			<figcaption>Locations of <a>segment endpoints</a> and [=gap junctions=] in a multi-column container.
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+				Column gaps (between column boxes) create endpoints at their bounds;some bounds coincide with the container edges
+				(like the start of the first 2 column gaps which coincide with the top edge,
```
```suggestion
				(like the starts of the first 2 column gaps which coincide with the top edge,
```

css-gaps-1/Overview.bs
```diff
@@ -132,80 +192,126 @@ Layout and painting</h3>
...
+				and the end of the last 2 column gaps which coincide with the bottom edge).
```
```suggestion
				and the ends of the last 2 column gaps which coincide with the bottom edge).
```

css-gaps-1/Overview.bs
```diff
@@ -335,6 +441,109 @@ Layout and painting</h3>
...
+	Each [=gap=] forms a collection of [=gap decoration segments=].
+
+	<div algorithm>
+		To <dfn>determine the gap decoration segments</dfn> for a given [=gap=] |G|:
+		<ol>
+			<li>
+				Let |S|, the collection of [=gap decoration segments=] for |G|, be an empty list.
+			</li>
+			<li>
+				Let |endpoints| be the list of all <a>segment endpoint</a>s <a lt="create segment endpoints for a gap">created</a> for |G|,
```
```suggestion
				Let |endpoints| be the list of all <a>segment endpoints</a> <a lt="create segment endpoints for a gap">created</a> for |G|,
```

You don't need to exclude the 's' for linkification to work. https://speced.github.io/bikeshed/#conjugating

(Same comment applies in similar cases throughout.)

css-gaps-1/Overview.bs
```diff
@@ -335,6 +441,109 @@ Layout and painting</h3>
...
+				</ul>
+			</li>
+			<li>
+				Return |S|.
+			</li>
+		</ol>
+
+		Note: Based on the algorithm for creating segment endpoints, the list |endpoints| is guaranteed to contain
+		an even number of elements that strictly alternate between [=start segment endpoint=] and [=end segment endpoint=],
+		beginning with a [=start segment endpoint=].
```
kbabbitt: This note begs the question of why we define "start segment endpoint" and "end segment endpoint" as their own terms. With the introduction of "gap junction" and the way we define segment endpoints around them, do we still need separate terms for starts and ends?

css-gaps-1/Overview.bs
```diff
@@ -409,121 +619,69 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
-		slightly offset to the right of the two middle points,
-		which are not marked.
-		These points will form another pair.)
-	</div>
+	Define an operation <dfn>UNION</dfn>(|s1|, |s2|) which updates |s1| such that:
+	|s1| = [|s1|.start, |s2|.end]
+
+	For two adjacent segments |s| and |t| (where |t| begins where |s| ends),
+	<a>UNION</a>(|s|, |t|) is <dfn>flanked by spanning items</dfn>
+	if spanning items exist on opposing sides of the segments.
```
kbabbitt: It's a little weird that this definition applies only to the output of UNION() rather than to segments in general. We also haven't defined what a "spanning item" is, and I don't think "spanning items exist on opposing sides of the segment" tells the whole story - such a spanning item specifically has to span over a gap in the perpendicular direction, correct?

I'll take a crack at editing, but please double check for correctness:

```suggestion
    A [=gap decoration segment=] is said to be <dfn>flanked by spanning items</dfn>
    if, on each side of the section of gap corresponding to the segment's endpoints,
    the gap abuts or intersects at least one gap in the perpendicular direction,
    and all such gaps in the perpendicular direction have items spanning over them.
```

css-gaps-1/Overview.bs
```diff
@@ -409,121 +619,69 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
 
 	<div algorithm>
-		To <dfn>determine pairs of gap decoration endpoints</dfn> within a given |gap|:
+		To <dfn>determine candidate segments</dfn> |C| from segments list |S|:
```
kbabbitt: I'm not a fan of the word "candidate" in this algorithm name. That word makes it sound like we might still drop segments in a future step, but this is the last step before applying insets, isn't it? I might instead call this algorithm "determine painted segments" and the next one "apply offsets to a segment".

EDIT: ... oh wait I see now, the word "append" in this algorithm links to "append a segment for painting." That's hiding some details that I think should be surfaced more explicitly.

css-gaps-1/Overview.bs
```diff
@@ -409,121 +619,69 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
-		the second and third points from the top are <a>discontiguous</a>.
-		Thus, the two upper <a>gap intersection points</a> form a pair,
-		and the two lower <a>gap intersection points</a> form a pair.
-		(There is an additional pair of <a>gap intersection points</a>,
-		slightly offset to the right of the two middle points,
-		which are not marked.
-		These points will form another pair.)
-	</div>
+	Define an operation <dfn>UNION</dfn>(|s1|, |s2|) which updates |s1| such that:
+	|s1| = [|s1|.start, |s2|.end]
```
kbabbitt: I think the subsequent algorithms would be slightly easier to follow if this were a pure function rather than side-effectual:

```suggestion
	Let <dfn>UNION</dfn>(|s1|, |s2|) be equal to [|s1|.start, |s2|.end].
```

css-gaps-1/Overview.bs
```diff
@@ -409,121 +619,69 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
-				If |break| is other than ''column-rule-break/none'',
-				and the first two items in |endpoints| are <a>discontiguous</a>,
-				remove the first item from |endpoints|, then go back to step 2.
-			</li>
+			<li>If |break| is ''column-rule-break/none'':
+				<ol>
+					<li>Let |s| be the first segment in |S|.
+					<li>Let |last| be the last segment in |S|.
+					<li><a>UNION</a>(|s|, |last|)
+					<li>[=Append a segment for painting|Append=] |s| to |C|
```
kbabbitt: Following my suggestions elsewhere, this would be:

```suggestion
					<li>Let |first| be the first segment in |S|.
					<li>Let |last| be the last segment in |S|.
					<li>Let |segment| be [=UNION=](|first|, |last|).
					<li>[=Apply offsets to a segment|Apply offsets to=] |segment|, then append it to |C|.
```

css-gaps-1/Overview.bs
```diff
@@ -409,121 +619,69 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
-						|break| is ''column-rule-break/spanning-item'',
-						and |start| is not <a>discontiguous</a> with the first item in |endpoints|.
-					</li>
-					<li>
-						|break| is ''column-rule-break/intersection'',
-						and |start| is not <a>discontiguous</a> with the first item in |endpoints|,
-						and the intersection is flanked by spanning items on opposing sides.
-					</li>
+					<li>Let |next| be the first segment in |S|.
+					<li>Let |candidate| be a copy of |s|, then <a>UNION</a>(|candidate|, |next|)
```
```suggestion
					<li>Let |candidate| be <a>UNION</a>(|s|, |next|).
```

css-gaps-1/Overview.bs
```diff
@@ -409,121 +619,69 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
+					<li>If |candidate|'s endpoints are <a>discontiguous</a>, go to step 9.
+					<li>If |break| is ''column-rule-break/intersection'' and |candidate| is not <a>flanked by spanning items</a>, go to step 9.
+					<li>Remove the first segment from |S|.
+					<li><a>UNION</a>(|s|, |next|)
```
```suggestion
					<li>Set |s| to <a>UNION</a>(|s|, |next|).
```

css-gaps-1/Overview.bs
```diff
@@ -409,121 +619,69 @@ Breaking gap decorations into segments: The 'column-rule-break', 'row-rule-break
...
-				Offset |start| by the result,
-				in the forward direction along |gap|'s axis.
-			</li>
-
-			<li>
-				<a>Compute the offset</a> for |end| within |gap|.
-				Offset |end| by the result,
-				in the reverse direction along |gap|'s axis.
-			</li>
+			<li>[=Append a segment for painting|Append=] |s| to |C|.
```
```suggestion
			<li>[=Apply offsets to a segment|Apply offsets to=] |s|, then append it to |C|.
```

css-gaps-1/Overview.bs
```diff
@@ -668,26 +825,27 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
 	Positive values recede inwards from the content edge, while negative values extend outwards, beyond the container's content bounds.
 
-	The ''column-rule-interior-inset-start'' and ''row-rule-interior-inset-start'' properties control the behavior for any [=start endpoint=] that is an [=interior gap intersection point=],
-	while the ''column-rule-interior-inset-end'' and ''row-rule-interior-inset-end'' properties control the behavior for any [=end endpoint=] that is an [=interior gap intersection point=].
-	Positive values recede away from the intersection, while negative values extend towards the center of the intersection.
+	The ''column-rule-interior-inset-start'' and ''row-rule-interior-inset-start'' properties control the behavior for any [=start endpoint=] that is an [=interior segment endpoint=],
+	while the ''column-rule-interior-inset-end'' and ''row-rule-interior-inset-end'' properties control the behavior for any [=end endpoint=] that is an [=interior segment endpoint=].
+	Positive values recede away from the [=segment endpoint=], while negative values extend beyond it.
```
kbabbitt: "Recede away from the segment endpoint" is ambiguous. Here's one idea:

```suggestion
    Positive values shorten the segment, while negative values extend it.
```

Though if we go with that, we should probably update the language for the edge insets to be consistent.

css-gaps-1/Overview.bs
SebastianZ: I was always wondering how those big texts are generated. I didn't think of Bikeshed having a built-in feature for that and previously created them manually in lack of a good online tool.😅
Thanks a lot for the hint, @kbabbitt!

oSamDavis: Done! Will resolve once I add the patch

css-gaps-1/Overview.bs
oSamDavis: I'd still say yes, we can't rely just on junctions. Flex cross gaps do not have junctions at all, so in those cases we depend on the bounds of the gap. We define types (start and end) for endpoints because they tell us what kind of endpoint we're dealing with and how it should be paired when drawing segments.

The note is essentially stating a requirement for the correctness of the algorithm. The pairing algo needs endpoints to alternate between start and end, this holds true because endpoints are typed.  So I see the note as a way of stating the contract the algo depends on.

css-gaps-1/Overview.bs
oSamDavis: Hmm, thinking of defining it in terms of junctions + endpoints + gaps. Might be easier to read.

css-gaps-1/Overview.bs
oSamDavis: Okay, so does the EDIT imply "candidate" is fine or we should change to determine painted segments ?

oSamDavis:
Just responded to last round of feedback @kbabbitt

css-gaps-1/Overview.bs
kbabbitt: Yes, I'd still change it. The edit was me correcting my misunderstanding of the relationship between the two algorithms and a precursor to me filing the next few comments to clarify things, thanks for applying these changes. I should have followed up here after filing those; I do still think "determine painted segments" would be clearer than "determine candidate segments."

css-gaps-1/Overview.bs
kbabbitt: I didn't mean to imply that we only place segment endpoints around junctions. What I meant was that junctions make the placement of *interior* endpoints much simpler than before. And since it is now guaranteed that endpoints will alternate between what we now think of as "start" and "end" - why not do away with the distinction between them, and just say that we place a "segment endpoint" at the start and end of each gap, and the start and end of each junction, and have the segmenting algorithm just take each successive pair of points? It would be one less thing a reader would need to keep in their head as they digest the spec.

kbabbitt: Though having just posted that... maybe in the future we'll discover a case where they aren't strictly alternating for some reason, and we'll need to deal with that. And the distinction probably does make some of the diagrams clearer. So let's keep it for now, I suppose.

css-gaps-1/Overview.bs
oSamDavis: Okay, thanks for the clarification. I just updated.

css-gaps-1/Overview.bs
oSamDavis: Okay sounds good. I vote for keeping it for now too.

PR #13312, opened 2026-01-07
[css-align][css-gaps] Move the gap properties to the Gaps spec and rename per CSSWG resolution. #13089

### LINKED ISSUE #13089: [css-align][css-gaps] Move the gap properties to the Gaps spec
For reasons of "nowhere better to put them at the time", the `gap` properties currently live in [the Align spec](https://drafts.csswg.org/css-align/#gaps). (Alignment can create gaps via distribution, so it wasn't unreasonable for *explicit* gaps to live there.)

But now we have a Gaps spec, which currently just defines the `*-rule` properties, but which *depends on* gaps to exist. We should just move the gap properties to Gaps (and possibly slightly retitle the css-gaps spec to "CSS Gaps", rather than its current more-specific title of "CSS Gap Decorations").

kbabbitt:
The only delta is annotating the Big Text heading to be Bikeshed-generated. Normative text is unchanged.


kbabbitt:
@tabatkins I'm getting the following Bikeshed warning on css-align-3 with these edits. I don't know much about how MDN integration works - should I just let this go, or add `Ignore MDN Failure` as suggested, or are additional edits needed?

```
WARNING: Skipped generating some MDN panels, because the following IDs weren't present in the document. Use `Ignore MDN Failure` if this is expected.
  #column-row-gap
  #gap-shorthand
```

kbabbitt:
@tabatkins or @fantasai, friendly ping, could I please get approval for the changes to css-align-3?

fantasai (approved):

fantasai:
Sorry @kbabbitt! I'm not really on top of GH notifications. :(

css-align-3/Overview.bs
```diff
@@ -2554,6 +2360,8 @@ Changes</h2>
 	* Defined that 'justify-self' affects the [=automatic size=] of block-level boxes
 		the same way it does for flex and grid items.
 		(<a href="https://github.com/w3c/csswg-drafts/issues/12102">Issue 12102</a>)
+	* Moved the gap properties to [[!CSS-GAPS-1]].
+		(<a href="https://github.com/w3c/csswg-drafts/issues/13089">Issue 13089</a>)
```
fantasai: Would be good to add a pile of anchors here for the removed section IDs.

fantasai (commented):
(Feel free to fix up this PR, or start a new one, or whatever. I trust you to move things correctly. ;)

css-align-3/Overview.bs
kbabbitt: I decided to put back the section header to keep section numbering the same, so that the HTML was easier to diff. I left the pile of anchors there.

PR #13321, opened 2026-01-09
[css-fonts-4] Create proper types for `<absolute-size>` and `<relative-size>`

tidoust:
Both `<absolute-size>` and `<relative-size>` were defined as values for `font-size`, but that made Bikeshed link the terms to the underlying type definitions in CSS2.

That's confusing for readers following links because the definition of `<absolute-size>` in CSS2 does not have `xxx-large`. That also confuses other spec processing tools, see https://github.com/w3c/webref/issues/1794

This update turns the value definitions into proper type definitions, which will override those defined in CSS2, and make links to `<absolute-size>` and `<relative-size>` target the local definitions.

### DIFF
--- a/css-fonts-4/Overview.bs
+++ b/css-fonts-4/Overview.bs
@@ -1384,8 +1384,8 @@ Font size: the 'font-size' property</h3>
 
 	Values have the following meanings:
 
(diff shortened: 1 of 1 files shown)


SebastianZ:
@tidoust Thanks!

Issue #12201, opened 2025-05-15
[css-gaps-1] Serializing `column-rule` shorthand from separate longhands


oSamDavis:
In CSS Gaps 1, the column-rule shorthand [has been extended to support lists of values.](https://drafts.csswg.org/css-gaps-1/#propdef-column-rule) This expanded grammar introduces complexity when serializing the shorthand, especially if the individual longhand properties—`column-rule-width`, `column-rule-style`, and `column-rule-color`—contain differing lengths or patterns. For example:

```
column-rule-width: repeat(auto, thin medium);
column-rule-style: repeat(3, dotted) solid;
column-rule-color: red blue repeat(auto, green);
```

In such cases, it becomes unclear how the `column-rule` shorthand should be computed and serialized. What should the resulting computed value be when the longhand properties differ in structure?

cdoublev:
Take a look at #7164.

kbabbitt:
7164 led to [Appendix A: Coordinating List-Valued Properties](https://www.w3.org/TR/css-values-4/#linked-properties), but I don't think that section is applicable in this case. Unlike backgrounds, transitions, or animations, there isn't a single "coordinating property" here: width, style, and color are all on equal footing, so to speak. We could define one of those as the coordinating property, but it's not clear to me that it would make things any simpler in this case. The sticking point I see is auto repeaters, because we don't know at computed-value time how many expansions will be generated.

I would look to CSSOM's [Serializing CSS Values](https://www.w3.org/TR/cssom/#serializing-css-values):

> 1. Let *shorthand* be the first shorthand property, in [preferred order](https://www.w3.org/TR/cssom/#concept-shorthands-preferred-order), that exactly maps to all of the longhand properties in *list*.
> 2. If there is no such shorthand or *shorthand* cannot exactly represent the values of all the properties in *list*, return the empty string.
> 3. Otherwise, [serialize a CSS value](https://www.w3.org/TR/cssom/#serialize-a-css-value) from a hypothetical declaration of the property *shorthand* with its value representing the combined values of the declarations in *list*.

If the repeaters happen to line up across the three properties, it's pretty trivial to construct an equivalent shorthand. If the repeaters don't line up exactly, it *may* still be possible to construct an equivalent shorthand by expanding integer repeaters. (The same should hold for specified values, btw.)

oSamDavis:
[...]

PR #13336, opened 2026-01-12
[css-gaps-1]: Add section for serializing column-rule shorthand and update separator of longhands #12201

oSamDavis:
Per #12201, this PR introduces a section explaining how the `column-rule` shorthand is serialized from its longhand properties. It also changes the value separator for the longhand properties from spaces to commas to ensure consistency with the shorthand format.

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -1071,9 +1071,9 @@ Gap decoration color: The 'column-rule-color' and 'row-rule-color' properties</h
 	<pre class='prod'>
 		<dfn>&lt;line-color-list&gt;</dfn>          = <<line-color-or-repeat>>#
 
-		<dfn>&lt;auto-line-color-list&gt;</dfn>     = <<line-color-or-repeat>>#? ,
-											          <<auto-repeat-line-color>> ,
-											          <<line-color-or-repeat>>#?
+		<dfn>&lt;auto-line-color-list&gt;</dfn>     = [<<line-color-or-repeat>># , ]? 
+											          <<auto-repeat-line-color>>
+											          [ , <<line-color-or-repeat>># ]?
 
 		<dfn>&lt;line-color-or-repeat&gt;</dfn>     = [ <<color>> | <<repeat-line-color>> ]
 
@@ -1124,9 +1124,9 @@ Gap decoration style: The 'column-rule-style' and 'row-rule-style' properties</h
 	<pre class='prod'>
 		<dfn>&lt;line-style-list&gt;</dfn>          = <<line-style-or-repeat>>#
 
-		<dfn>&lt;auto-line-style-list&gt;</dfn>     = <<line-style-or-repeat>>#? ,
-											          <<auto-repeat-line-style>> ,
-											          <<line-style-or-repeat>>#?
+		<dfn>&lt;auto-line-style-list&gt;</dfn>     = [<<line-style-or-repeat>># , ]?
+											          <<auto-repeat-line-style>>
+											          [ , <<line-style-or-repeat>># ]?
 
 		<dfn>&lt;line-style-or-repeat&gt;</dfn>     = [ <<line-style>> | <<repeat-line-style>> ]
 
@@ -1175,9 +1175,9 @@ Gap decoration width: The 'column-rule-width' and 'row-rule-width' properties</h
 	<pre class='prod'>
 <dfn>&lt;line-width-list&gt;</dfn>          = <<line-width-or-repeat>>#
 
-		<dfn>&lt;auto-line-width-list&gt;</dfn>     = <<line-width-or-repeat>>#? ,
-											          <<auto-repeat-line-width>> ,
-											          <<line-width-or-repeat>>#?
+		<dfn>&lt;auto-line-width-list&gt;</dfn>     = [<<line-width-or-repeat>># , ]?
+											          <<auto-repeat-line-width>>
+											          [ , <<line-width-or-repeat>># ]?
 
 		<dfn>&lt;line-width-or-repeat&gt;</dfn>     = [ <<line-width>> | <<repeat-line-width>> ]
 


css-gaps-1/Overview.bs
```diff
@@ -1580,6 +1580,114 @@ Gap decoration shorthands: The 'column-rule' and 'row-rule' properties</h3>
...
+	it serializes as an empty string.
+
+<h4 id="longhand-alignment">Longhand Properties Alignment</h4>
+
+	Let |W|, |S| and |C| be the computed values
+	for the width, style and color longhand properties respectively.
+	Recall that each is a list of items,
+	where an item is either a single value or a ''repeat()'' notation.
+
+	The longhands <dfn export lt="line up">line up</dfn>
```
```suggestion
	The longhands are <dfn>repeater aligned</dfn>
```

* "line up" is a bit too generic of a name for this concept, I think.
* I don't think the definition needs to be exported? The only uses are local to this spec. Ignore me if Bikeshed complains though.
* `lt` should only be necessary if the linking text is different the visible text. See https://speced.github.io/bikeshed/#autolink-manual and https://speced.github.io/bikeshed/#conjugating

css-gaps-1/Overview.bs
```diff
@@ -1580,6 +1580,114 @@ Gap decoration shorthands: The 'column-rule' and 'row-rule' properties</h3>
...
+	if and only if all of the following are true:
+
+	<ol>
+		<li>
+			|W|, |S| and |C| have the same number of items.
+		</li>
+		<li>
+			For each item index |i|,
+			|W|[|i|], |S|[|i|] and |C|[|i|]
+			are all either ''repeat()'' notations or non-repeat values.
```
kbabbitt: I'd suggest dropping this list item and incorporating the non-repeat case into the next list item; see next couple of comments.

css-gaps-1/Overview.bs
```diff
@@ -1580,6 +1580,114 @@ Gap decoration shorthands: The 'column-rule' and 'row-rule' properties</h3>
...
+		</li>
+		<li>
+			For each item index |i| where |W|[|i|], |S|[|i|] and |C|[|i|] are ''repeat()'' notations,
+			they all have the same shape:
+			<ol>
+				<li>
+					they are either all <a>integer repeaters</a> or all <a>auto repeaters</a>.
```
```suggestion
					they are either all non-repeat values, all <a>integer repeaters</a>, or all <a>auto repeaters</a>.
```

css-gaps-1/Overview.bs
```diff
@@ -1580,6 +1580,114 @@ Gap decoration shorthands: The 'column-rule' and 'row-rule' properties</h3>
...
+		<li>
+			|W|, |S| and |C| have the same number of items.
+		</li>
+		<li>
+			For each item index |i|,
+			|W|[|i|], |S|[|i|] and |C|[|i|]
+			are all either ''repeat()'' notations or non-repeat values.
+		</li>
+		<li>
+			For each item index |i| where |W|[|i|], |S|[|i|] and |C|[|i|] are ''repeat()'' notations,
```
```suggestion
			For each item index |i|,
			|W|[|i|], |S|[|i|] and |C|[|i|] all have the same shape:
```

css-gaps-1/Overview.bs
```diff
@@ -1580,6 +1580,114 @@ Gap decoration shorthands: The 'column-rule' and 'row-rule' properties</h3>
 	</pre>
 
 	These shorthands set the corresponding width, style, and color properties as a set.
+	When serializing these shorthands from their corresponding longhand properties,
+	if the values for the corresponding longhand properties do not <a>line up</a>,
+	it serializes as an empty string.
+
+<h4 id="longhand-alignment">Longhand Properties Alignment</h4>
+
+	Let |W|, |S| and |C| be the computed values
```
```suggestion
	Let |W|, |S| and |C| be the values
```

These conditions should apply to any set of width, style, and color values, specifically including specified as well as computed values.

css-gaps-1/Overview.bs
```diff
@@ -1580,6 +1580,114 @@ Gap decoration shorthands: The 'column-rule' and 'row-rule' properties</h3>
...
+				<li>
+					If they are <a>integer repeaters</a>, the <<integer>> values are equal.
+				</li>
+				<li>
+					Each ''repeat()'' notation contains the same number of values.
+				</li>
+			</ol>
+		</li>
+	</ol>
+
```
kbabbitt, line 1615: The main reason for these rules is to ensure shorthand declarations can be serialized back without complicating cases where one of the longhands gets overridden with a different shape. I'd suggest modifying the examples to illustrate this, and lead with examples that do serialize and follow with ones that don't.

```
	<div class="example">
		This shorthand declaration:
		<pre>
			column-rule: 1px solid red, 2px dashed green, 3px dotted blue;
		</pre>

		Breaks down to the following longhand values:

		<pre>
			column-rule-width: 1px, 2px, 3px;
			column-rule-style: solid, dashed, dotted;
			column-rule-color: red, green, blue;
		</pre>

        Which serialize back to the same original shorthand value,
        because the longhands are [=repeater aligned=]:
		all three properties have the same number of items (3),
		and each item is a single value (not a repeater).

        The same behavior would occur if the three longhands were specified directly
        rather than coming from a shorthand.
	</div>

    /* ... insert example with repeater that serializes here ... */

	<div class="example">
		These declarations:
		<pre>
			column-rule: repeat(auto, 1px solid red, 2px dashed green);
			column-rule-width: repeat(auto, 3px, 4px, 5px);
		</pre>

		Break down to the following longhand values:

		<pre>
			column-rule-width: repeat(auto, 3px, 4px, 5px);
			column-rule-style: repeat(auto, solid, dashed);
			column-rule-color: repeat(auto, red, green);
		</pre>

        The 'column-rule' shorthand will serialize back as an empty string
        because the longhands are not [=repeater aligned=]:
		the auto repeaters contain different numbers of values.

        The same behavior would occur if the three longhands were specified directly
        rather than coming from a shorthand.
	</div>

    /* ... perhaps insert other examples that don't serialize here ... */
```

css-gaps-1/Overview.bs
```diff
@@ -1245,11 +1245,11 @@ Lists of values and the ''repeat-line-color/repeat()'' notation</h3>
 			<div class="example">
 				An author may write:
 				<pre>
-					column-rule-color: gray red blue red blue red blue gray;
+					column-rule-color: gray, red, blue, red, blue, red, blue, gray;
 				</pre>
 				Or shorten to the following, which produces the same sequence of colors:
 				<pre>
-					column-rule-color: gray repeat(3, red blue) gray;
+					column-rule-color: gray, repeat(3, red blue), gray;
```
```suggestion
					column-rule-color: gray, repeat(3, red, blue), gray;
```

There should be commas inside repeaters as well. (And similar for other examples)

oSamDavis: Ah, I see. I had initially thought that repeaters maintain its space-separated structure for the repeated values. Will update this and others then.

css-gaps-1/Overview.bs
oSamDavis: Thanks for the guidance! Just updated the examples.

Issue #13226, opened 2025-12-16
[css-forms-1] Use visibility for checkmark control?


fantasai:
The current stylesheet creates and destroys the `::checkmark` pseudo-element whenever the `:checked` state changes. It might be more performant (and consistent) to make it exist always and simply toggle the `visibility`, just like we do for select menus. What do we think? CC @bfgeek

nt1m:
Using `visibility` is my weak preference here. cc @lukewarlow

lukewarlow:
Given we also haven't yet accounted for Indeterminate which will need presumably either the ::checkmark or a different pseudo element (I prefer just using ::checkmark personally). I agree it makes sense to always define ::checkmark and then change it's visibility.

annevk:
Why would we need a different pseudo-element for `:indeterminate`?

lukewarlow:
Well we wouldn't. I mean it's technically not a checkmark, but neither is a radio, as I say just using checkmark seems far better.

annevk:
A, I see #11869 discusses `:indeterminate` and already has a reasonable solution proposed.

lukewarlow:
I think this is in the spec and has test coverage so going to go ahead and close this issue.

PR #13357, opened 2026-01-16
[css-forms-1] Add missing content property to radio checkmark

lukewarlow:
[css-forms-1] Add missing content property to radio checkmark

Without this the radio's checkmark is never generated.


lukewarlow:
Pending https://github.com/w3c/csswg-drafts/issues/13226 this might change but either way I think it's good to fix and then we can change it.

PR #13358, opened 2026-01-16
[css-forms-1] Add checkmark content for indeterminate checkbox.

lukewarlow:
image

see #11869

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -894,9 +894,13 @@ input[type=radio]::checkmark {
     content: '';
 }
 
-input[type=checkbox]:not([switch], :checked)::checkmark {
+input[type=checkbox]:not([switch], :checked, :indeterminate)::checkmark {
 input[type=radio]:not(:checked)::checkmark {
     visibility: hidden;
+}
+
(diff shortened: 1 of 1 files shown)


css-forms-1/Overview.bs
```diff
@@ -891,11 +891,15 @@ input[type=radio]::checkmark {
     content: '\2713' / '';
 }
 
-input[type=checkbox]:not([switch], :checked)::checkmark {
+input[type=checkbox]:not([switch], :checked, :indeterminate)::checkmark {
```
nt1m, line 897: I think this won't work as you'd expect. `:not(:checked)` will also set `visibility: hidden` on the indeterminate checkmark.

Maybe this is closer to what you want? I'd need to test:
```
input[type=checkbox]:not([switch]):not(:checked):not(:indeterminate)
```

lukewarlow: Ah yeah you're probably right.

lukewarlow: Hmm it seems to work fine fwiw. I think your suggested selector is equivalent.

I think it's `NOT (a switch OR checked OR indeterminate)`

Issue #12602, opened 2025-08-13
[css-gaps-1] Gap decorations next to empty grid areas


kbabbitt:
We got some author feedback around unwanted gap decorations appearing next to empty grid areas. Examples:

-----

https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/1099 - unwanted decorations in the empty space to the left of the `content` area:

<img width="2156" height="1194" alt="Image" src="https://github.com/user-attachments/assets/9f78ec9d-3940-45a2-b138-913bc15e022d" />

-----

https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/1100 - unwanted decoration in the top row, to the left of "Other things about me":

<img width="1624" height="560" alt="Image" src="https://github.com/user-attachments/assets/20e8f5d8-18c5-4f5b-9c5a-3e3593937d8e" />

-----

https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/1111 - live version https://codepen.io/bigandy/pen/vENxYzV - unwanted decorations after the last article:

<img width="3176" height="1996" alt="Image" src="https://github.com/user-attachments/assets/dd3b209a-8b0e-4161-b590-4ca22f66d471" />

-----

For the first two cases, I suggested two options: [Gap Decoration Areas](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSGapDecorations/explainer.md#placement-of-gap-decorations) to set different decorations for different sub-areas of the grid using grid lines, or a new property to define how gap decorations interact with empty grid areas, similar to how we have `empty-cells` for tables. In both of those two cases, the authors expressed a preference for Gap Decoration Areas.

The third case is more interesting. It's a responsive grid that re-flows the contained articles based on how wide the container is. I don't see how Gap Decoration Areas might solve the unwanted decorations in such a case, at least not without script to support. So I think this is a good use case for a property to control decorations in empty grid areas.

Perhaps that property is `empty-cells`, though I think in this case there would need to be three states: (1) decorations in all gaps; (2) decorations in gaps with an item on at least one side; (3) decorations in gaps with items on both sides. In offline discussion with @alisonmaher @oSamDavis @jav099 last week, it also came up that there may be cases where the author wants different behaviors in each direction.

alisonmaher:
[...]

PR #13374, opened 2026-01-20
[css-gaps-1]: Add section for rule visibility items # 12602

oSamDavis:
This CL adds a section for `rule-visibility-items` properties which dictate how gap decorations segments are drawn relative to empty cells per #12602 .

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -330,7 +330,7 @@ Layout and painting</h3>
 		grid/subgrid/fragmentation/subgrid-gap-decorations-fragmentation-018.html
 	</wpt>
 
-	Gap decorations are [=assigned gap decoration values|assigned=] after [=collapsed gutter=] has been applied.
+	Gap decorations are [=assigned gap decoration values|assigned=] after [=collapsed gutter|gutter collapsing=] has been applied.
 	A given set of [=collapsed gutters=] consume exactly one [=gap decoration=]; the next [=gap decoration=]
 	is applied to the next gutter (or set of [=collapsed gutters=]). Therefore, [=collapsed gutters=] are treated as a
 	single gutter for decoration purposes.


css-gaps-1/Overview.bs
```diff
@@ -330,7 +330,7 @@ Layout and painting</h3>
 		grid/subgrid/fragmentation/subgrid-gap-decorations-fragmentation-018.html
 	</wpt>
 
-	Gap decorations are [=assigned gap decoration values|assigned=] after [=gutter collapsing|collapsed gutter=] has been applied.
+	Gap decorations are [=assigned gap decoration values|assigned=] after [=collapsed gutter=] has been applied.
```
kbabbitt: Why this change?

css-gaps-1/Overview.bs
```diff
@@ -640,6 +640,123 @@ Pairing gap intersection points into segments</h4>
 		</figure>
 	</div>
 
+<!-- Big Text: Rule-Visibility-Items
```
kbabbitt: I'd like to keep the break and inset sections next to each other and move this new section after inset.

css-gaps-1/Overview.bs
```diff
@@ -640,6 +640,123 @@ Pairing gap intersection points into segments</h4>
...
+████▌  █▌  █▌ █▌    █████▌       █▌   █▌ ████  ███▌  ████ ████▌  ████ █▌    ████ █████▌ █   ▐▌       ████ █████▌ █████▌ █     █  ███▌ 
+█▌  █▌ █▌  █▌ █▌    █▌           █▌   █▌  ▐▌  █▌  █▌  ▐▌  █▌  █▌  ▐▌  █▌     ▐▌    █▌   ▐▌  █         ▐▌    █▌   █▌     ██   ██ █▌  █▌
+█▌  █▌ █▌  █▌ █▌    █▌           █▌   █▌  ▐▌  █▌      ▐▌  █▌  █▌  ▐▌  █▌     ▐▌    █▌    █ ▐▌         ▐▌    █▌   █▌     █▌█ █▐█ █▌    
+████▌  █▌  █▌ █▌    ████   ████▌ ▐▌   █   ▐▌   ███▌   ▐▌  █████   ▐▌  █▌     ▐▌    █▌    ▐▌█   ████▌  ▐▌    █▌   ████   █▌ █ ▐█  ███▌ 
+█▌▐█   █▌  █▌ █▌    █▌            █  ▐▌   ▐▌      █▌  ▐▌  █▌  █▌  ▐▌  █▌     ▐▌    █▌     █▌          ▐▌    █▌   █▌     █▌   ▐█     █▌
+█▌ ▐█  █▌  █▌ █▌    █▌            ▐▌ █    ▐▌  █▌  █▌  ▐▌  █▌  █▌  ▐▌  █▌     ▐▌    █▌     █▌          ▐▌    █▌   █▌     █▌   ▐█ █▌  █▌
+█▌  █▌  ███▌  █████ █████▌         ▐█    ████  ███▌  ████ ████▌  ████ █████ ████   █▌     █▌         ████   █▌   █████▌ █▌   ▐█  ███▌ 
+-->
+<h3 id="visibility-rules">
+Painting Gap Decorations with Visibility Rules</h3>
```
```suggestion
Controlling Gap Decoration Visibility: The '*-rule-visibility-items' properties</h3>
```

css-gaps-1/Overview.bs
```diff
@@ -640,6 +640,123 @@ Pairing gap intersection points into segments</h4>
...
+	<pre class='propdef'>
+		Name: column-rule-visibility-items, row-rule-visibility-items
+		Value: all | around | between
+		Initial: all
+		Applies to: <a>grid containers</a>
```
kbabbitt: I think this can be applied to other container types as well? Especially once we change over to the segment model.

css-gaps-1/Overview.bs
```diff
@@ -640,6 +640,123 @@ Pairing gap intersection points into segments</h4>
...
+		Computed value: as specified
+		Animation type: discrete
+	</pre>
+
+	These properties control whether a gap decoration segment is painted
+	in portions of gutters adjacent to empty cells.
+
+	<dl dfn-type="value" dfn-for="column-rule-visibility-items, row-rule-visibility-items">
+		<dt><dfn>all</dfn>
+		<dd>Paint decorations in all gap segments, regardless of whether cells are occupied. This is the initial value.
```
```suggestion
		<dd>Paint decorations in all gap segments, regardless of whether cells are occupied.
```

css-gaps-1/Overview.bs
oSamDavis: There's a link error with bikeshed where "gutter collapsing" is not found.

kbabbitt: Oh I see now... it's because the vertical bar syntax is backwards. It should be `[=definition|visible text=]` not the other way around. https://speced.github.io/bikeshed/#autolink-inside

```suggestion
	Gap decorations are [=assigned gap decoration values|assigned=] after [=collapsed gutter|gutter collapsing=] has been applied.
```

css-gaps-1/Overview.bs

css-gaps-1/Overview.bs
oSamDavis: Okay, added multi-col. And reworded to use items+areas rather than cells.

Issue #13309, opened 2026-01-07
[filter-effects-1] Syntax definition of `filter-functions`


romainmenke:
https://github.com/w3c/csswg-drafts/blob/dd61a8cf2aff822edd584f4b54a8a0065bac802d/filter-effects-1/Overview.bs#L370-L371

https://github.com/w3c/csswg-drafts/commit/f6f0efa067f2b5c63ca764c00ebb8c316b513c44#diff-8be13fb6cc5f77ad353d66447f373758037137170db50c90abc48867e235a9e4R28-R371

This commit changed the definition to:
`blur( ) | brightness( ) | contrast( ) | drop-shadow( ) | grayscale( ) | hue-rotate( ) | invert( ) | opacity( ) | sepia( ) | saturate( )`

Effectively making arguments invalid for all of these.

Likely an unintended change @tabatkins?

SebastianZ:
@tidoust reverted that change in #13380, so they are linked again.
@romainmenke With that fix I think it's fine to close this issue. Let me know if you disagree.

Sebastian

romainmenke:
Thank you for the follow up @SebastianZ
Good to close 🙇

PR #13380, opened 2026-01-22
[filter-effects-1][editorial] Fix syntax of filter-function type

tidoust:
Fix for #13309. The `''foo''` special syntax does not work in production rules.


SebastianZ:
Seems like an obvious bugfix to me. Thanks, @tidoust!

Sebastian

PR #728, opened 2016-11-17
[cssom] Specify serialization of CSSMediaRule and CSSFontFaceRule

canaltinova:
This PR adds serialization steps of CSSMediaRule and CSSFontFaceRule. I reverse-engineered the grammars and came up with these steps.
Serialization of these rules are also implemented with this approach in servo: https://github.com/servo/servo/pull/14238

cc @Manishearth

PR #13386, opened 2026-01-23
[cssom-1] Append only one newline in @media rule block

cdoublev:
The [current algorithm](https://drafts.csswg.org/cssom-1/#serialize-a-css-rule) definition requires adding two newlines:

  > 3. A single SPACE (U+0020), followed by the string "{", i.e., LEFT CURLY BRACKET (U+007B), followed by **a newline**.
  > 4. The result of performing serialize a CSS rule on each rule in the rule’s cssRules list, filtering out empty strings, indenting each item with two spaces, all joined with newline.
  > 5. **A newline**, followed by the string "}", i.e., RIGHT CURLY BRACKET (U+007D)

Based on [WPT tests](https://github.com/web-platform-tests/wpt/blob/master/css/cssom/serialize-media-rule.html) and the [original commit](https://github.com/w3c/csswg-drafts/pull/728) of the current version, I think this is an obvious bugfix.


tabatkins:
While you're right this needs fixing, this particular fix breaks serilaization of non-empty rules, as it will put the `}` on the same line as the last declaration.

What needs to happen instead is the final newline should be moved into the step that serializes the contents and made dependent on that.

cdoublev:
Thanks for catching this mistake. I also applied an obvious bugfix for `@keyframes` serialization.

Issue #13123, opened 2025-11-19
[css-view-transitions-2] [scoped] Are we happy with contain:v-t as the API for tag containment?


skobes-chromium:
A goal of [Scoped View Transitions](https://github.com/WICG/view-transitions/blob/main/scoped-transitions.md) is to enable composition: outer and inner components, possibly from different authors or libraries, can both run view transitions at the same time without coordinating with each other or knowing about each other.

This means we need some way to encapsulate `view-transition-name` tags so that each component can choose its own tag names without worrying about conflicts in a shared namespace.

The solution proposed in the explainer (and currently implemented by Chromium) is to support CSS `contain: view-transition` and encourage developers to set this property on elements that will be used as scopes (i.e. that will have `startViewTransition()` called on them).  An outer transition's tag search will not descend into an inner element that has `contain: view-transition`.

This is similar to @flackr's proposal at https://github.com/w3c/csswg-drafts/issues/8915#issuecomment-1883461265 for containing other kinds of named items.

Alternatively, we could have a different API that establishes a containment boundary by declaring that an element will be used as a scope, such as:
* `view-transition-scope: <ident> | none` (@vmpstr floated this idea a long time ago)
* `will-change: view-transition`
* `this-element-will-be-a-view-transition-scope: true`

Alternatively, we could _not_ have any API for this, but I would be opposed to that as explained in https://github.com/w3c/csswg-drafts/issues/12321#issuecomment-3549226635. Applying containment "behind the scenes" during an active view transition is not a sufficient answer, since tags might be picked up by an outer transition before the inner transition starts.

skobes-chromium:
A related question is whether `contain: strict` implies `contain: view-transition`.

vmpstr:
Proposal: we should add a way to contain names. Following scroll triggers and anchor pos, maybe this should be named `view-transition-scope: auto | none`

vmpstr:
Related: #12321

skobes-chromium:
Note: Chrome 146.0.7655.0 (current dev channel release) implements `view-transition-scope: auto` instead of `contain: view-transition`. The [explainer](https://github.com/WICG/view-transitions/blob/main/scoped-transitions.md) has been updated to reflect this.

skobes-chromium:
@bramus pointed out that other `-scope` properties, such as `timeline-scope`, use the value `all` rather than `auto`, so we should consider `view-transition-scope: all`.

[...]

PR #13390, opened 2026-01-23
[css-view-transitions-2] Spec view-transition-scope

vmpstr:
[css-view-transitions-2] Spec view-transition-scope

These are tentative edits pending #13123 but replaces `contain: view-transitions` which is also not agreed upon.

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -639,7 +639,7 @@ Note: An <code>element.startViewTransition()</code> call does not update
 			[=Skip the view transition|skip=] |viewTransition| with an "{{InvalidStateError}}" {{DOMException}}.
 
 		1. If [=this=] is not the [=/root element|document root element=],
-			force [=this=] to have ''view-transition-scope: auto'' for the duration of the transition.
+			force [=this=] to have both [=layout containment=] and [=view transition scoping=] for the duration of the transition.
 
 		1. If |document|'s [=active view transition=] is not null and its [=outbound post-capture steps=] is not null,
 			then:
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -639,7 +639,7 @@ Note: An <code>element.startViewTransition()</code> call does not update
 			[=Skip the view transition|skip=] |viewTransition| with an "{{InvalidStateError}}" {{DOMException}}.
 
 		1. If [=this=] is not the [=/root element|document root element=],
-			force [=this=] to have [=layout containment=] for the duration of the transition.
+			force [=this=] to have ''view-transition-scope: auto'' for the duration of the transition.
```
skobes-chromium: Shouldn't it force both (layout containment and v-t-s: auto)?

css-view-transitions-2/Overview.bs
```diff
@@ -935,29 +935,88 @@ the active view transition of an element is exposed to script via a property on
...
-	The ''contain: view-transition'' value establishes <dfn export>view transition name containment</dfn> on an element.
-	This causes any [=document-scoped view transition names=] to only be visible
-	for [=view transitions=] scoped to this element or its descendants.
+	Note: Although 'view-transition-scope' is [=discrete|discretely animatable=],
+	it only has an effect during 'view-transition-name' discovery
 
-	Note: Most elements running a [=scoped view transition=]
-	should use [=view transition name containment=],
-	to ensure that their [=document-scoped view transition names=]
-	don't accidentally interfere with other [=scoped view transitions=].
```
skobes-chromium, line 955: Could we preserve this note about avoiding interference and just reword it to recommend view-transition-scope?

css-view-transitions-2/Overview.bs
vmpstr: Done.

css-view-transitions-2/Overview.bs
vmpstr: Done.

PR #13397, opened 2026-01-26
[css-view-transitions-2]: update visibility vs opacity wording

### LINKED ISSUE #12629: [css-view-transition-1] `visibility: hidden` analogy is misleading, elements remain keyboard-focusable and accessible
Hi,
The current css-view-transitions-1 draft at point [4.2. View Transition Painting Order](https://drafts.csswg.org/css-view-transitions-1/#view-transition-stacking-layer), says:
>When a Document’s active view transition’s phase is "animating", the boxes generated by any element in that Document with captured in a view transition and its element contents, except transition root pseudo-element’s inclusive descendants, are not painted (as if they had `visibility: hidden`) and do not respond to hit-testing (as if they had `pointer-events: none`).

>Note: Elements participating in a transition need to skip painting in their DOM location because their image is painted in the corresponding `::view-transition-new()` pseudo-element instead. Similarly, hit-testing is skipped because the element’s DOM location does not correspond to where its contents are rendered. However, there is no change in how these elements are accessed by assistive technologies or the accessibility tree.

The issue is that the use of `visibility: hidden` in the following sentence:
>are not painted (as if they had `visibility: hidden`) and do not respond to hit-testing (as if they had `pointer-events: none`).

is problematic with this sentence:
>there is no change in how these elements are accessed by assistive technologies or the accessibility tree.

because `visibility: hidden` don't just make the element invisible and not-interactive, it also remove it from the accessibility tree.

Also, `visibility: hidden` doesn’t reflect how Chrome and Firefox currently preserve element focusability and interactivity, while `opacity: 0` combined with `pointer-events: none` matches their behavior.

---

Related issue:
- https://github.com/w3c/csswg-drafts/issues/9365
- https://github.com/w3c/csswg-drafts/issues/10930

---

Testing links:
- https://codepen.io/sb3nder/pen/QwjmmZy
- https://mdn.github.io/dom-examples/view-transitions/mpa/index.html
- https://mdn.github.io/dom-examples/view-transitions/spa

vmpstr:
[css-view-transitions-2]: update visibility vs opacity wording

Fixes #12629


vmpstr:
Note level 1 was fixed via #13378

Issue #13102, opened 2025-11-13
New name for `app-region` property?


diekus:
Builds on #7017 

New name for `app-region` CSS property.

diekus:
During TPAC several of us agreed that the name `app-region` is not meaningful. This CSS property is used to define an area from where the user can point and drag a window in a desktop environment. The name `draggable` should be avoided to circumvent confusing with the [global draggable attribute in HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/draggable). 

Starting to bikeshed on a new name for `app-region`:

-  `drag-area`
- `drag-point`
- `drag-behavior`

emilio:
FWIW Gecko's version of this was `-moz-window-dragging`. Since this only affects the top level page, maybe `window-drag` or so?

mhochk:
Building on the earlier https://github.com/w3c/csswg-drafts/issues/7017#issuecomment-1609114145

Though not mentioned in the spec, the existing implementation (at least in Edge) extends not only the draggable nature of the titlebar to this element, but also other native titlebar behaviors like the default contents of the right-click menu. It seems like this is desirable, as the intended use is to replace the titlebar, so the user agent choosing to adopt other titlebar-like behaviors would be a good thing.

Given that, what about a CSS property like `titlebar-area`?


Alternatively (if there is strong pushback on the above direction):
* If keeping this to only dragging the window, and controllable via CSS, @emilio's `window-drag` seems like the right direction. Maybe `window-dragging`?
* If keeping this to only dragging the window, but potentially not as a CSS property, what about expanding the `draggable` HTML property to include a new `with-window` value. Similar to `contenteditable` having `true`/`false`/`plaintext-only` options.

diekus:
I hadn't notice that the context menu still works! I like the `window-drag` proposal @emilio mentioned above. I think there might be reasons where a developer might want to allow dragging from other areas that are not the top of the window. (side panels, other navigation areas, etc).

@mhochk  Nonetheless it'd mean in Edge we'd want to remove this context menu as it's weird that that UX appears on untrusted area if you think about it?

mhochk:
I chatted with @diekus a bit on this and now agree that CSS (not HTML) is the right place for this to live so that this can be cleanly controlled inside a CSS media query. As such, I withdraw my suggestion to extend `draggable`.

[...]

Issue #13101, opened 2025-11-13
Define possibly inheritable behaviour for `app-region` (now `window-drag`)


diekus:
Builds on #7017 

Research 'inheritable' behaviour of `app-region to better understand use cases and if the feature should only work if the title bar is disabled. 

(also update the explainer to reflect this information)

diekus:
Following up on the discussion at TPAC for #7017, The `app-region` property does inherit, both the `drag` and `no-drag` values. 

I've also looked at the behaviour of native apps on Windows and macOS to determine if this is functionality that should work always. It was not clear at TPAC if certain apps could be moved from the left navigation sidebar. I've looked at how apps behave on macOS, and not Finder, not Maps nor the System Settings have this UX. Windows also does not expose this functionality to move windows from the side panel. 

I therefore think the behaviour should remain as-is, with the CSS property only working when the default's platform moving surface disappears. 

I have updated the explainer accordingly to reflect these findings. 

cc @emilio

emilio:
FWIW I didn't mean the application sidebar, but the toolbar. E.g. the KDE system settings on Linux (and Firefox, and probably other QT apps) can be dragged by the toolbar even when using system window decorations.

Also Chromium on Linux, FWIW...

diekus:
Thanks for the extra context. When you say "system window decorations", is it [this](https://develop.kde.org/docs/plasma/aurorae/)?  I don't see why modifications of the appearance of the theme of the platform would alter the behaviour of the CSS property. The default window chrome would continue to behave as expected on the platform, and if this UX surface disappears (because the developer might want to create custom UX) then the property is enabled to facilitate moving the window.

If a developer wanted to create a custom title bar/toolbar combo they could use something like WCO (if supported) to achieve this. Everything else is up to the platform. I've tried Chrome on ChromeOS (closest I have atm to Chromium on Linux :P) and don't get the behaviour, but Firefox on Windows I do get dragging from the _custom_ title/tool bar. Nonetheless, this would also be done via WCO, as it is not the _default's_ OS UX. 

Now if I understand correctly if I wanted an installed web app to try to achieve the same UX that Firefox has, I need WCO, which enables `app-region`. 

<img width="1100" height="131" alt="Image" src="https://github.com/user-attachments/assets/6c721835-4bc8-4c3e-86b3-37380fb336fc" />

_I can do a web app that looks like the previous screenshot, but I need WCO + `app-region` for that._ 

Unless I am missing something?

[...]

PR #13404, opened 2026-01-27
adds the `window-drag` text to the css-ui-4 spec

diekus:
Adds the `window-drag` (formerly `app-region`) CSS property to the css-ui-4 spec. 

This comes out of the resolutions https://github.com/w3c/csswg-drafts/issues/13102 and 
https://github.com/w3c/csswg-drafts/issues/13101 taken on the 21/01/26.


diekus:
Hola @frivoal ! I wanted to make sure the edit contains the necessary text and any recommendations that you as an editor think could be missing based on the resolutions that we've discussed in the call.

thanks

diekus:
hola @frivoal! Just bringing this to your attention! Let me know if there is anything missing or suggested changes!

css-ui-4/Overview.bs
```diff
@@ -4072,6 +4068,48 @@ Single-Line Text Input Controls</h4>
...
+<h2 id="window-interaction">
+Window Interaction</h2>
+
+<h3 id="window-drag">
+Moving the window: the 'window-drag' property</h3>
+
+    <pre class=propdef>
+	Name: window-drag
+	Value: auto | move | none  
+	Initial: ''move/auto''
```
emilio, line 4080: Initial should only be one of the two, right?

css-ui-4/Overview.bs
```diff
@@ -4072,6 +4068,48 @@ Single-Line Text Input Controls</h4>
...
+	Applies to: all elements
+	Inherited: yes
+	Percentages: N/A
+	Computed value: ''move/auto''
```
emilio, line 4084: Probably as specified?

css-ui-4/Overview.bs
```diff
@@ -4072,6 +4068,48 @@ Single-Line Text Input Controls</h4>
...
+    <wpt>
+    </wpt>
+
+The window chrome of an installed web application has a designated area where it can be 
+dragged and moved. This area is generally at the top of the window, usually being the the 
+ default OS's title, toolbar or both. Some APIs or capabilities like 
+ [[WINDOW-CONTROLS-OVERLAY]] might alter and remove the surfaces where users can move a window.
+
+The 'window-drag' CSS property allows the author to specify an area in the viewport where 
+a user can interact to drag and move the window. Values have the following meanings:
```
fantasai, line 4098:
```suggestion
An installed application's window typically has a designated area where it can be dragged to move the window.
This area is generally at the top of the window, usually part of the titlebar, toolbar, or both.
For installed web aps, APIs or capabilities like [[WINDOW-CONTROLS-OVERLAY]]
can alter and remove the surfaces where users can move a window.
The 'window-drag' property allows the author to specify an area within the page
where a user can drag to move the window.

Values have the following meanings:
```

fantasai (commented):
This probably needs a bit more clarification on what part of the element can be used as a drag target (by tying it to some other definitions, I'm not sure which ones).

Issue #12324, opened 2025-06-11
[css-view-transitions-2] [scoped] What is the layout of the ::v-t pseudo in relation to the scope?


skobes-chromium:
This issue tracks an open question about the behavior of [Scoped View Transitions](https://github.com/WICG/view-transitions/blob/main/scoped-transitions.md). Further context appears in the document [Self-Participating Scopes](https://bit.ly/svt-sps).

**Q: What is the layout of the ::v-t pseudo in relation to the scope?**

Because the scope owns the `::view-transition` pseudo-element, it may seem natural for the pseudo to be laid out as an absolute-positioned child of the scope:

```
/* user-agent style */
::view-transition {
  position: absolute;
  inset: 0;
}
```

However, this becomes problematic when the scope is self-participating (#12319), because
* the pseudo is sized to the scope's **padding box**, which doesn't include borders; and
* if the scope is a scroller, the pseudo is placed **inside the scrolling contents**.

You can see this problem in action [here](https://output.jsbin.com/zodacaj/quiet) (use Chrome 139+ with experimental web platform features enabled). The scope is a scroller that is scrolled down. During the transition, the "outside" view of the scroller (i.e. its border box) is captured and render-redirected into the `::view-transition`... which is at the top of the scrolling contents inside the scroll clip.

An alternative might be to layout the pseudo as a sibling of the scope that is effectively [anchor-positioned](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning) to the scope:

```
::view-transition {
  position-anchor: --scope;
  left: anchor(left);
  top: anchor(top);
  width: anchor-size(width);
  height: anchor-size(height);
}
```

There is some further discussion on [crbug.com/417988089](https://crbug.com/417988089).

cc @noamr @vmpstr @flackr @bramus @jakearchibald

bramus:
There have been other requests, unrelated to View Transitions, for the ability to position something against a specific container:

- #7475 (specifically [this comment](https://github.com/w3c/csswg-drafts/issues/7475#issuecomment-1182982687)) to add an extension to `position: fixed` to allow one to indicate _what_ to posfix against.
- #8762 to add `container-fixed` as a keyword for `position`
- #9868 to add a `position-container` property

I believe either one of those would solve this.

[...]

PR #13408, opened 2026-01-28
[css-view-transitions-2] Add scoped layout section and update UA stylesheet.

skobes-chromium:
[css-view-transitions-2] Add scoped layout section and update UA stylesheet. #12324

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -1335,18 +1335,21 @@ is a child of the [=ViewTransition/root element=] in the [=box tree=], but
 with some behaviors that are different from ordinary children:
 
 * The ''::view-transition'' pseudo-element generates a stacking context
-	which paints after all other children of the [=ViewTransition/root element=],
-	regardless of their 'z-index'.
+	which paints separately from, and on top of, the painting of the
+	[=ViewTransition/root element=] and its other children.
+
+	Note: A consequence of painting separately is that properties on the root element
+	that affect the way the root element is painted, such as 'opacity', do not apply
+	to the ''::view-transition'' pseudo-element. However, if the root element is captured
+	in a ''::view-transition-group'' for its own transition, such properties
+	will be reflected in the captured image.
 
 * If the [=ViewTransition/root element=] has a [=scrolling box=], the
 	''::view-transition'' pseudo-element is not translated by the scroll position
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -1328,6 +1328,26 @@ ISSUE: `::view-transition-group-children()` needs to be specified here.
...
 		However, there is no change in how these elements are accessed by assistive technologies or the accessibility tree.
 
+## Scoped View Transition Layout ## {#scoped-view-transition-layout}
+
+For a [=scoped view transition=], the ''::view-transition'' pseudo-element
+is a child of the [=ViewTransition/root element=] in the [=box tree=], but
+with some behaviors that are different from ordinary children:
+
+* The ''::view-transition'' pseudo-element generates a stacking context
+	which paints after all other children of the [=ViewTransition/root element=],
```
vmpstr: I believe this stacking context also paints on top of the actual root element, right? In cases where we don't have a root capture, but have a child participant, that participant would presumably paint on top of the border box of the root element.

css-view-transitions-2/Overview.bs
```diff
@@ -1328,6 +1328,26 @@ ISSUE: `::view-transition-group-children()` needs to be specified here.
...
+	regardless of their 'z-index'.
+
+* If the [=ViewTransition/root element=] has a [=scrolling box=], the
+	''::view-transition'' pseudo-element is not translated by the scroll position
+	of the [=ViewTransition/root element=] and not clipped to the scrolling area
```
vmpstr: This sentence makes it sound that the clip escape is conditional on the fact that the root element has a scrolling box, but that's not true with overflow: clip.

With my previous comment, perhaps we can rephrase this a bit to say that the stacking context paints on top of the root stacking context and then list consequences of that. 

Note that the issue is below is fine if we plan to actually enumerate all properties affected, but in order to try and be future proof, maybe we can try and phrase it in terms of stacking context and list _some_ properties that are affected and then say something like "and other similar properties"

css-view-transitions-2/Overview.bs
skobes-chromium: Done

css-view-transitions-2/Overview.bs
skobes-chromium: Done

vmpstr (approved):

PR #13412, opened 2026-01-29
[css-view-transitions-2] Add nested view transition groups spec

vmpstr:
[css-view-transitions-2] Add nested view transition groups spec


css-view-transitions-2/Overview.bs
```diff
@@ -1024,9 +1024,125 @@ the active view transition of an element is exposed to script via a property on
...
+	Value: normal | contain | nearest | <<custom-ident>>
+	Initial: normal
+	Inherited: no
+	Percentages: n/a
+	Computed Value: as specified
+	Animation type: discrete
+	</pre>
+
+	The 'view-transition-group' property can be used in conjuction with 'view-transition-name' to generate a hierarchy of [=named view transition pseudo-element=].
+	Specifically, this property has no effect if 'view-transition-name' is ''view-transition-name/none''.
```
skobes-chromium: I would remove the word "Specifically" since this sentence is not a more specific expression of the idea in the preceding sentence. (It's more of a "But, ..." or a "However, ...".)

css-view-transitions-2/Overview.bs
```diff
@@ -1024,9 +1024,125 @@ the active view transition of an element is exposed to script via a property on
...
+
+	<dl dfn-type=value dfn-for=view-transition-group>
+		: <dfn>normal</dfn>
+		:: The ''view-transition-group()'' [=pseudo-element=]
+			generated by this element
+			is a child of the ''view-transition-group()'' [=pseudo-element=]
```
skobes-chromium: Shouldn't this say that it is a child of the `view-transition-group-children()` pseudo-element?

vmpstr:
PTAL

Issue #10579, opened 2024-07-16
[css-color-4] Evaluate static gamut mapping algorithms for oklab/oklch spaces


ccameron-chromium:
Separating the action item "ccameron: review gamut mapping algorithms" [from here](https://github.com/w3c/csswg-drafts/issues/9449#issuecomment-2162711081) into a new issue. Tagging in @foolip @LeaVerou @svgeesus for the more limited discussion here (please feel free to include anyone else).

From the linked resolution in the issue, these algorithms are not to be applied to the various RGB spaces. I'd like to scope my comments to just "gamut mapping to apply to oklab/oklch" spaces, since I feel a bit differently about XYZ and Lab spaces (I would want XYZ and Lab to follow RGB, because they are the canonical profile connection spaces used in ICC based color matching, and don't want to mix this discussion with those issues).

The properties that I'd like for the algorithm to apply to just-oklab/oklch are much looser. An iterative algorithm with quadratic convergence is quite fine.

There were lots of algorithms listed in issue 9449, so I'd like to collect a reduced list of candidates here to discuss.

To provide an extremely short list:
* Projecting in Lab to the Rec2020 gamut boundary along constant L
  * This is what's proposed in the spec (modulo termination conditions and iteration method)
  * We can write out the details for an exact solution that quadraticly converges (it's a bit of work, but I can do it if desired)
  * An exact solution for the exact Rec2020 boundary may not be desirable
    * There is some concave nastiness around the blue corner of the gamut that it would be nice to avoid.
    * There is also chrome reduction as discussed in the spec
    * Creating a "nicer-shaped" surface very close to the Rec2020 gamut might be a good solution here
* Projecting in Rec2020 YUV along constant Y
  * This is extremely fast and probably quite good.

svgeesus:
Thanks for starting this.

> * Projecting in Lab to the Rec2020 gamut boundary along constant L

Why Lab and not Oklab, given the hue curvature of Lab (especially in blues, but also in orange to a lesser extent).

I agree that mapping to a smoothed gamut boundary may well be better.

> * Projecting in Rec2020 YUV along constant Y

That is video luma, right? Where Y is computed without bothering to undo gamma-encoding first?

ccameron-chromium:
> Thanks for starting this.
> 
> > * Projecting in Lab to the Rec2020 gamut boundary along constant L
> 
> Why Lab and not Oklab, given the hue curvature of Lab (especially in blues, but also in orange to a lesser extent).

[...]

PR #13416, opened 2026-01-30
[css-color-4] Correction to ray trace to avoid throwing off our intersect calculation and match MINDE flow #10579

facelessuser:
My suggested fix via https://github.com/w3c/csswg-drafts/issues/10579#issuecomment-3815362986 was not the ideal fix and artificially limited `tnear`. This is a numerical instability issue and should be addressed as such by using an epsilon comparison to prevent rays that are too small. This corrects the fix.

I've also adjusted the algorithm to match the flow of MINDE in regardless to checking lightness before returning out of gamut colors.

I also revised the cleanup step. Since the destination is a gamut-bound space, we can clip there as we do in MINDE.

Related #10579

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -6052,9 +6052,7 @@ Sample Pseudocode for the Ray Trace Gamut Mapping</h4>
 
 	<ol>
 		<!-- check if we need gamut mapping at all -->
-		<li>if |origin| is in gamut for |destination|, 
-			convert |origin| to |destination| and return it as the gamut mapped color
-		</li>
+		<li>if |destination| has no gamut limits (XYZ-D65, XYZ-D50, Lab, LCH, Oklab, OkLCh) convert |origin| to |destination| and return it as the gamut mapped color
 		<!-- we do, so convert to OkLCh -->
 		<li>let |origin_OkLCh| be |origin| converted from |origin color space|
 			to the OkLCh color space</li>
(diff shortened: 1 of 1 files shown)


svgeesus:
@facelessuser could you please:

1. [Create a W3C account](https://www.w3.org/account/request/), if you don't already have one
2. Link your GitHub and W3C accounts (the menu item "Connected accounts" on your [W3C account page](https://www.w3.org/users/174425/)).

I can then make a link for an individual royalty free license commitment (for the whole Ray Trace GM, not just this specific pull request). Thanks!

facelessuser:
Excellent, I'll create an account!

facelessuser:
Cool, I've made the request and also updated the algorithm to better match the CSS MINDE flow. This also addresses the concern about clipping and converting to the destination space. I've updated it to follow what CSS MINDE does.

A final note, I settled on 1E-12 for the ray size threshold. It's generally recommended either 1E-9 or 1E-12 for high precision cases in double floating point. I'm probably indifferent to whichever, as I don't think there would be a significant difference.

facelessuser:
I've made the account and linked GitHub.

svgeesus:
Thanks! The request for [non-participant license commitment](https://www.w3.org/2019/12/05-dak-nplc/) has been sent.

facelessuser:
@svgeesus Sorry for not getting back about this, but I currently get the following when I click the link to the non-participant license commitment.

```
The level of access required for this page is inconsistent with [your account](https://www.w3.org/users/myprofile).
There is not yet an access level assigned to this page. W3C staff may be able to change the access level for this resource using the ,access tool.
```

deniak:
@facelessuser you should be able to submit your commitment using this [link](https://www.w3.org/standards/licensing/contributions/9252268/13416/commit/).

facelessuser:
For a bit, I was restricted from commenting for some reason, but I've filled out the non-participant license agreement.

svgeesus:
There is a known problem with the IPR bot currently; I have reported it but meanwhile it is holding up merging any PR with an individual participant commitment (including this one).

facelessuser:
That makes sense. Thanks for the update. Hopefully we'll get it in sooner rather than later.

PR #13419, opened 2026-01-30
[css-values-4] Correct example range in mod function explanation

zzzzBov:
Fix a mistake in an example for the `mod` function where the correct **B** value is `5px`.


SebastianZ:
Thanks for the fix, @zzzzBov!

Sebastian

Issue #12054, opened 2025-04-07
CSS WG drafts server


svgeesus:
> If you want to start a general conversation with the WG about replacing the draft server, feel free to do so. I welcome it. File another issue for that, and please begin the conversation with a practical plan, a source of funding, or replacement servers, not just complaints about the status quo.

 _Originally posted by @plinss in [#11354](https://github.com/w3c/csswg-drafts/issues/11354#issuecomment-2780186234)_

> On behalf of the W3C team, I am conveying that we do not consider this resolved. 
> 
> Please don't close this issue again until we have agreement with the working group and with the W3C team that it's actually resolved. 

 _Originally posted by @sideshowbarker in [#11354](https://github.com/w3c/csswg-drafts/issues/11354#issuecomment-2780157408)_

There was some [discussion at the CSS f2f in New York regrding server issues](https://www.w3.org/2025/04/01-css-irc.html#T20-56-00)

tabatkins:
I think the most important thing to check, first, is just what services are still being provided by Peter's server, and what we can do to lessen that burden.

1. Current versions of drafts are now built and served by GitHub, and reverse-proxied by Peter's server. (There's currently a redirect built in to the local copies that shifts you to the drafts.csswg.org URL if you visit them from the github.io URL, to avoid the github.io urls from being depended on.) Could we move `drafts.csswg.org` to serving off of GitHub first, so we don't even touch Peter's server for the most common case?
2. The *directory* of the drafts is maintained by Peter's server. This can be pretty trivially taken over by just having an `index.html` here on GitHub that we manually update, I think. Could probably even auto-build one on CI.
3. *Historical* versions of drafts are on Peter's server. Are these required? Are they valuable? Can we make this easier to handle? Can we move *serving* these off your server entirely, in favor of just committing bunches of static documents to a GH repo, plus an auto-generated directory file built in CI?
4. The wiki is hosted on Peter's server. Is this problematic due to bots? Could we move this to a GH wiki?
5. Is there anything else, @plinss?

svgeesus:
> 1. Could we move `drafts.csswg.org` to serving off of GitHub first, so we don't even touch Peter's server for the most common case?

That would seem to be a quick and easy way to lessen the load and to improve responsiveness and uptime.

> 2\. The _directory_ of the drafts is maintained by Peter's server. This can be pretty trivially taken over by just having an `index.html` here on GitHub that we manually update, I think. Could probably even auto-build one on CI.

[...]

PR #13432, opened 2026-02-03
Build markdown documents to HTML in the CI workflow

sideshowbarker:
The legacy draft server converts `.md` files in spec directories (explainers, transition documents, proposals, etc.) to HTML. This adds that conversion to the GitHub Actions workflow so the files are served from GitHub Pages.

A new `bin/build-markdown.py` script finds all `.md` files one level deep in spec directories and converts them to HTML using Python-Markdown, with a minimal stylesheet for readability. The workflow installs the `markdown` package and runs the script after building specs.

Relates to https://github.com/w3c/csswg-drafts/issues/12054

### DIFF
--- a/.github/workflows/build-specs.yml
+++ b/.github/workflows/build-specs.yml
@@ -33,7 +33,7 @@ jobs:
           python-version: "3.14"
           cache: 'pip'
 
-      - run: pip install bikeshed
+      - run: pip install bikeshed markdown
       - run: bikeshed update
 
       # The following chunk of code all stolen from andeubotella
@@ -63,6 +63,8 @@ jobs:
             echo "  $file"
             bikeshed issues-list "$file" || true
           done
+      - name: Build markdown
+        run: python ./bin/build-markdown.py
       - name: Build index & symlinks
         run: python ./bin/build-index.py
       - run: rm -rf ./.git{,attributes,ignore}
--- a/bin/build-markdown.py
+++ b/bin/build-markdown.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+"""Convert markdown files in spec directories to HTML."""
+
+import glob
+import os
+import re
+
+import markdown
+
+TEMPLATE = """\
+<!doctype html>
+<meta charset="utf-8">
+<title>{title}</title>
+<style>
+body {{ max-width: 50em; margin: 2em auto; padding: 0 1em; font-family: sans-serif; line-height: 1.6; }}
+code {{ background: #f4f4f4; padding: .1em .3em; border-radius: 3px; }}
+pre {{ background: #f4f4f4; padding: 1em; overflow: auto; border-radius: 3px; }}
+pre code {{ background: none; padding: 0; }}
+img {{ max-width: 100%; }}
+</style>
+{body}
+"""
+
+
+def extract_title(text):
+    m = re.search(r"^#\s+(.+)", text, re.MULTILINE)
+    return m.group(1).strip() if m else "Untitled"
+
+
+def main():
+    md = markdown.Markdown(extensions=["fenced_code", "tables"])
+
+    for md_file in sorted(glob.glob("*/*.md")):
+        if md_file.startswith("."):
+            continue
+
+        html_file = os.path.splitext(md_file)[0] + ".html"
+        if os.path.exists(html_file):
+            continue
+
+        with open(md_file, encoding="utf-8") as f:
+            text = f.read()
+
+        title = extract_title(text)
+        body = md.convert(text)
+        md.reset()
+
+        with open(html_file, "w", encoding="utf-8") as f:
+            f.write(TEMPLATE.format(title=title, body=body))
+
+        print(f"  {html_file}")
+
+
+if __name__ == "__main__":
+    main()


frivoal:
Thanks, this is the kind of thing we need.

I wonder if we have a dependency on any particular flavor of markdown. I don't know what the current server uses, but it'd be good to check if that's the same as what's proposed here or not. Another consideration is that quite a few of our markdown files start their life as documents served by GitHub, and might be using various features of GitHub flavored markdown (which may or may not be preserved by the current server). If so, and if we want to support that, maybe using a tool like https://github.com/joeyespo/grip could help.

sideshowbarker:
@frivoal Short answer: I think what’s set up here now — using Python-Markdown — is the best choice and does everything y’all need, as far as supporting all markdown features in the sources here. So that’s what I’d recommend.


Longer answer:

> I wonder if we have a dependency on any particular flavor of markdown.

I think there are some markdown tables in some of the sources, and we need to fenced-code syntax-highlighting stuff. So for those, we need GitHub-flavored-markdown support. But the good news is: Python-Markdown supports those. See the bit of the change at https://github.com/w3c/csswg-drafts/pull/13432/files#diff-eeeae3642da5cedb2d38612b2e6193054b2171632b614a104842144be1b5f74eR31:

> ```python
>     md = markdown.Markdown(extensions=["fenced_code", "tables"])
> ```


> I don't know what the current server uses, but it'd be good to check if that's the same as what's proposed here or not. Another consideration is that quite a few of our markdown files start their life as documents served by GitHub, and might be using various features of GitHub flavored markdown (which may or may not be preserved by the current server). If so, and if we want to support that, maybe using a tool like [joeyespo/grip](https://github.com/joeyespo/grip?rgh-link-date=2026-02-03T13%3A52%3A40Z) could help.

I think the only other tool to possibly consider is `cmark-gfm` https://github.com/github/cmark-gfm. It’s a C program (though it has Python bindings we could use). I don’t know what the install story for it is — as far as us being able to add a cross-platform install step, to enable contributors to run it locally if they wanted to. I do know that there is a brew package for it. But regardless, if we were to use that, I think we’d need to have the docs explain how to install it.

For Python-Markdown in contrast, we don’t need to do that.

But anyway, as far as I can see, the only additional GFM feature that switching to `cmark-gfm` would buy us is: You could use the special `> [!NOTE]` etc GFM admonitions stuff, and get the pretty HTML rendering for that in the HTML output. Python-Markdown doesn’t (yet) support converting that to the same pretty rendering.

But none of the existing markdown sources here are using those admonitions. Yet. So there’d be no reason in choosing  `cmark-gfm` for that at this point — I mean, just to get support for a GFM feature that none of sources are using.

That said, it would not be a big deal at all at some point later to update the build here to switch over to using `cmark-gfm`. I would be happy to set up, if at some point later y’ll decide you want to use those GFM admonitions.

Otherwise, all that said, as far as I can see, the choice really is just between Python-Markdown and `cmark-gfm`. There are no other alternatives that need to be considered.

tabatkins:
I think we're fine. GH-flavored markdown's only real innovation is the NOTE/etc blocks, and those are done via an existing markup pattern that looks reasonable enough when rendered by a vanilla markdown renderer.

sideshowbarker:
Not to beat this into the ground, but for the record here: After looking into `cmark-gfm` further, I find it also doesn’t handle the GH > [!NOTE] etc admonitions stuff (with apparently are officially called _“alerts”_).

So there’d be zero point in using `cmark-gfm` rather than Python-Markdown.

But there is one thing that _does_ handle those GFM “alerts”: The GH API `https://api.github.com/markdown` endpoint.

So, if at some point later y’all _do_ find a need to use GFM features that Python-Markdown doesn’t support, then you could switch the build to using that GH API markdown endpoint. The main downside would be the obvious one: That it would requiring having  network connection / being online. There’s also the fact that the API is rate limited, and the fact that the API requests are limited to the source markdown files needing to be 400KB or less.

The rate limit and size limit would not be problems here in practice. But I reckon the requirement to have to be online could be a PITA. And it seems like there’d be no point in introducing an extra PITA for no real additional benefit.

So yeah, I’d imagine all y’all and your contributors would agree that relying on Python-Markdown seems fine.

frivoal:
> I think we're fine. GH-flavored markdown's only real innovation […]

I am more concerned about incompatibilities. Conpare the rendering of this file:
GH: https://github.com/webplatformco/project-image-animation/blob/main/image-animation-property/README.md
python: https://drafts.csswg.org/css-image-animation-1/explainer

I don't think I'm using any unique feature, but the way things get escaped, new lines get processed, etc, is different, and running the GH file through python is just broken. I could probably fix by adjusting the file, but if this explainer is broken, it may not be the only one…

plinss:
The legacy server used [CommonMark](https://pypi.org/project/commonmark/)

sideshowbarker:
> > I think we're fine. GH-flavored markdown's only real innovation […]
> 
> I am more concerned about incompatibilities. Conpare the rendering of this file:
GH: [webplatformco/project-image-animation@`main`/image-animation-property/README.md](https://github.com/webplatformco/project-image-animation/blob/main/image-animation-property/README.md?rgh-link-date=2026-02-04T15%3A17%3A32Z)
python: [drafts.csswg.org/css-image-animation-1/explainer](https://drafts.csswg.org/css-image-animation-1/explainer)

Yipes yeah that’s awful

> The legacy server used [CommonMark](https://pypi.org/project/commonmark/)

OK see https://github.com/w3c/csswg-drafts/pull/13448.

https://github.com/readthedocs/commonmark.py appears to be archived and no longer maintained. So, switching the GitHub-supported CommonMark successor would seem to be the best choice.

frivoal:
https://github.com/w3c/csswg-drafts/pull/13448 made things looks good now. Thanks.

PR #13466, opened 2026-02-07
[css-view-transitions-2] ViewTransition.types should be readonly

gb-jos:
types attribute is marked as [SameObject] but not readonly, which is invalid as per webidl spec


gb-jos:
@emilio no idea what to do about the ipr check

emilio:
I'm not technically an editor, so @noamr or @vmpstr should probably merge it :)

Issue #12826, opened 2025-09-18
[css-overflow-4] Confirm edit about IFCs and block ellipsis


frivoal:
This issue is to confirm the edit already applied by https://github.com/w3c/csswg-drafts/pull/12825 with the Working Group.

After discussing with @andreubotella, we found the sentence in question to be vague and and somewhat ambiguous, but that the only _reasonable_ thing it could mean was what we have now rephrased it to. So I have applied the fix without waiting, but we wanted to notify the group and give an opportunity for review.

Leaving it up to the chairs if you want to include this in a sync call, or handle it through async CfC.

andreubotella:
In particular, the question is: if clamping by height results in the clamp point being immediately after an independent formatting context, but the IFC has a line box at the end that does fit the "immediately preceding" conditions, do we place the ellipsis inside the IFC?

In this edit, the answer is no – no line would have an ellipsis in this case.

astearns:
I think I’d prefer to have this edit explained and validated on a sync call.

andreubotella:
The examples and edits requested landed in #13525. Closing.

PR #13525, opened 2026-02-18
[css-overflow-4] Clarify when `block-ellipsis` affects a line box

andreubotella:
As per #12826, this patch clarifies and adds examples for when the `block-ellipsis` property affects a line box.

This incorporates an edit suggested by @fantasai in #12825 after that PR had been closed.

### DIFF
--- a/css-overflow-4/Overview.bs
+++ b/css-overflow-4/Overview.bs
@@ -808,7 +808,7 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
 		so no ellipsis will be shown:
 
 		```html
-		<div style="continue: collapse; max-lines: 2; block-ellipsis: auto">
+		<div style="continue: collapse; max-height: 2lh; block-ellipsis: auto">
 		  Line 1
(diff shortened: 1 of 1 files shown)


css-overflow-4/Overview.bs
```diff
@@ -759,6 +758,65 @@ Indicating Block-Axis Overflow: the 'block-ellipsis' property</h3>
...
+	</div>
+
+	<div class=example>
+		In this example,
+		the last line box is contained
+		in an [=independent formatting context=],
+		so no ellipsis will be shown:
+
+		```html
+		<div style="continue: collapse; max-lines: 2; block-ellipsis: auto">
```
frivoal: `max-lines` doesn't count lines in independent formatting contexts, so the clamp point would be after line 3. Here I think we need to go with a height based clamp instead.
```suggestion
		<div style="continue: collapse; max-height: 2lh; block-ellipsis: auto">
```

frivoal (changes requested):
Looks good, except for what I think is a bug in the 3rd example. @andreubotella, do you confirm my proposed correction?

Issue #6205, opened 2021-04-11
[css-contain] Similar to window.matchMedia(), Container Queries should have a similar method


clairebaire:
CSS Media Queries have a method on the Window interface (`window.matchMedia()`) giving a script the ability to listen for media query changes. I propose that CSS Container Queries has the same functionality. However, I think this introduces some complexity, and therefore I'd like to present two thoughts for consideration:

### Proposal 1

A new method on the `HTMLElement` interface possibly called `matchContainer()` or something similar. Reason it wouldn't be on the `Window` is due to the nature of container queries - not being based upon viewport size, but size of the container. However, this should be influenced by the question still pending, which is what is the default behavior of a container query when there is no `contain` property on any parent element.

### Proposal 2

A new Observer of sorts, similar to `ResizeObserver` or `MutationObserver` where it's a bit more clear that it is listening for any kind of container change, and that it relays the container as a member of a `ContainerQueryList`. This would make it not pollute the `HTMLElement` interface, and instead be a DOM method. This would solely be used to identify when a container query is activated, and which items are affected by that activation.

Admittedly, the second proposal seems very similar to what `ResizeObserver` does already, but I feel it misses the connection that CSS Contain Level 3 is attempting to solve. This could be something that might be added to ResizeObserver, but might also be its own API.

Thoughts?

andruud:
> but I feel it misses the connection that CSS Contain Level 3 is attempting to solve

Why isn't the existing ResizeObserver (as-is) good enough to listen for resize events on containers?

jogibear9988:
Is there anything new about this? We created a Designer with CSS parsing support (https://github.com/node-projects/web-component-designer) where we would need the "matchContainer" API to find the mathcing container queries

seedee:
Wouldn't just checking a container element's offsetWidth when the window resizes be functionally the same as container query?

jogibear9988:
> Wouldn't just checking a container element's offsetWidth when the window resizes be functionally the same as container query?

no? container queries could not only check for a size....

aradalvand:
I expected this to have already been implemented, such a shame.

Loque-:
Just wanted to add that I run into a use case where I wanted to be able to do; 

`window.matchContainer('(aspect-ratio < 16 / 9)')`

Would have _loved_ to be able to do this instead of writing some custom ResizeObserver which could potentially lead to performance issues if not managed mindful of re-paints for calculating/retrieving values.

[...]

PR #13551, opened 2026-02-23
[css-conditional-5] Add `matchContainer()` and `ContainerQueryList` API text

meyerweb:
As per #6205.


css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+    A simple piece of code that detects changes in the overflow in any of a series of elements can be written as follows:
+
+    <pre highlight=javascript>
+    function handleOverflowChange(event) {
+        if(event.matches) // the thing overflowed
+            &hellip;
+        else
+            &hellip;
+    }
+    var cql = matchContainer("(scroll-state(scrollable))");
```
DevSDK: Don't we need to `element.matchContainer()` since the method seems added on `Element`?

meyerweb: Whoops, yes.  Good catch!

meyerweb:
```suggestion
    var cql = Element.matchContainer("(scroll-state(scrollable))");
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+<ol>
+
+ <li>
+
+  For each {{ContainerQueryList}} object <var>target</var> that has <var>doc</var> as its <a for=ContainerQueryList>document</a>, in the order they were created, oldest first, run these substeps:
+
+  <ol>
+
+   <li>If <var>target</var>'s <a>matches state</a> has changed since the last time these steps were run,
```
DevSDK: How about describe or mention what the `target` is? I initially thought it was element itself. But the recent [issues talks](https://github.com/w3c/csswg-drafts/issues/6205#issuecomment-4022106222) and container query behavior is finding eligible ancestor of the element.

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+</ol>
+
+<h3 id=the-ContainerQueryList-interface>The {{ContainerQueryList}} Interface</h3>
+
+This section integrates with the <a for=/>event loop</a> defined in HTML. [[!HTML]]
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>container query list</dfn> 
+and an associated 
+<dfn for=ContainerQueryList>document</dfn> set on creation.
```
DevSDK: I suppose `document` seems global scope like media query. We may need to change it to `element`.



```suggestion
<dfn for=ContainerQueryList>container query list</dfn> 
and an associated 
<dfn for=ContainerQueryList>element</dfn> set(or assigned?) on creation.
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>container</dfn> which is the 
+<a lt='serialize a container query list'>serialized</a>
```
DevSDK: we missed the serialization rule here. 

We may need to create/adopt the existing rule? https://drafts.csswg.org/cssom-1/#serialize-a-media-query-list

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>container</dfn> which is the 
+<a lt='serialize a container query list'>serialized</a> 
+form of the associated [=ContainerQueryList/container query list=].
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>matches state</dfn> 
+which is true if the associated [=ContainerQueryList/container query list=] 
+matches the state of the <a for=ContainerQueryList>document</a>, and false otherwise.
```
DevSDK: I think we need to describe the container here.
I'm not sure it is enough to describe at least we can refer the same target with `@container`:
https://www.w3.org/TR/css-conditional-5/#query-container

```suggestion
matches the state of the <a for=ContainerQueryList>query container</a>, and false otherwise.
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+
+When asked to <dfn export>evaluate container queries and report changes</dfn> 
+for a {{Document}} <var>doc</var>, run these steps:
+
+<ol>
+
+ <li>
+
+  For each {{ContainerQueryList}} object <var>target</var> that has <var>doc</var> as its <a for=ContainerQueryList>document</a>, in the order they were created, oldest first, run these substeps:
```
DevSDK: I think we need to change the `changing target` here:


```suggestion
When asked to <dfn export>evaluate container queries and report changes</dfn> 
for a {{query container}} of the element <var>ele</var>, run these steps:
<ol>

 <li>

  For each {{ContainerQueryList}} object <var>target</var> that has <var>ele</var>, in the order they were created, oldest first, run these substeps:
```

It would be - https://www.w3.org/TR/css-conditional-5/#query-container

And there are other scenario to fire an event here:

**Changing the container**

e.g., 

```js
const target = document.getElementById("target");
const newContainer = document.getElementById("new-container");

const query = target.matchContainer("(width < 400px)");
query.onchange = (event) => console.log("matches:", event.matches);

// Scenario: The target element is moved to a different container in the DOM.
// if old container has 300px and new container has 500px?
newContainer.appendChild(target);
```

WDYT about if the steps cover above example?

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+  <ol>
+
+   <li>If <var>target</var>'s <a>matches state</a> has changed since the last time these steps were run, 
+   <a>fire an event</a> named <a event>change</a> at <var>target</var>
+   using {{ContainerQueryListEvent}},
```
DevSDK: I suppose we missed definition `ContainerQueryListEvent` idl here?

FYR: https://drafts.csswg.org/cssom-view/#mediaquerylistevent

and also We may need to `ContainerQueryListEventInit`  that correspond with `MediaQueryListEventInit`

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+    <pre highlight=javascript>
+    function handleOverflowChange(event) {
+        if(event.matches) // the thing overflowed
+            &hellip;
+        else
+            &hellip;
+    }
+    var cql = Element.matchContainer("(scroll-state(scrollable))");
+    cql.onchange = handleOverflowChange;
+    </pre>
```
DevSDK:
```suggestion
    <pre highlight=javascript>
    function handleOverflowChange(event) {
        if(event.matches) // the thing overflowed
            &hellip;
        else
            &hellip;
    }
    
    const element = document.getElementById("target");
    var cql = element.matchContainer("(scroll-state(scrollable))");
    cql.onchange = handleOverflowChange;
    </pre>
```

Since we need to instance of the Element, we should find the element from the document.

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+<ol>
+ <li>Let <var>parsed container query list</var> be the result of
+ <a lt='parse a container query list'>parsing</a>
+ <var>query</var>.
+ <li>Return a new {{ContainerQueryList}} object,
+ with [=this=]'s
+ <a for="/">associated <code>Document</code></a>
+ as the <a for=ContainerQueryList>document</a>,
+ with <var>parsed container query list</var>
```
DevSDK:
```suggestion
 <li>Return a new {{ContainerQueryList}} object,
 with [=this=] as the <a for=ContainerQueryList>element</a>,
 as the <a for=ContainerQueryList>document</a>,
 with <var>parsed container query list</var> 
```

ditto with the element instead of document comment.

css-conditional-5/Overview.bs
DevSDK: Related: https://github.com/w3c/csswg-drafts/pull/13551/changes#r2950925618

DevSDK (commented):
I focused on referring to the element's target container which is already defined as a "[query container](https://www.w3.org/TR/css-conditional-5/#query-container)" for `@container`.

DevSDK:
cc + @frivoal can you take a look here? :)

DevSDK:
While I'm working on [the CL](https://chromium-review.googlesource.com/c/chromium/src/+/7656126), there are three things need to be consider:

1. -List name: Media queries can be a list with a ',' separator, but container queries don't seem to have a similar syntax.
2. The query itself: Should we accept a container name or not? If we do, what would be a good approach? matchContainer("container name and condition") or matchContainer("condition", containerName)?
3. Serialization rules (maybe adopt containerRule.containerQuery?)

DevSDK:
Additionally, @andruud mentioned that we should specify the event timing in [the CL](https://chromium-review.googlesource.com/c/chromium/src/+/7656126).

Since the following specs are managed by HTML's event [loop-processing-model](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model):

- **[resize-observer Algorithms](https://drafts.csswg.org/resize-observer/#algorithms):**
	>  Note: The following algorithms are invoked as a part of update the rendering in the html spec: gather active resize observations at depth, has active resize observations, has skipped resize observations, broadcast active resize observations and deliver resize loop error notification.
    
- **[css-conditional-5 Updating Scroll State](https://www.w3.org/TR/css-conditional-5/#updating-scroll-state):**
	> To avoid such layout cycles, scroll-state query containers update their current state as part of run snapshot post-layout state steps which is only run at specific points in the HTML event loop processing model.
        When asked to run snapshot post-layout state steps, update the current state of every scroll-state query container. This snapshotted state will be used for any style and layout updates until the next time these steps are run.


We should consider a similar approach here.

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+The <dfn method for=PreferenceObject>matchContainer(query)</dfn> method, when invoked, must run these steps:
+
+<ol>
+ <li>Let <var>parsed container query list</var> be the result of
+ <a lt='parse a container query list'>parsing</a>
+ <var>query</var>.
+ <li>Return a new {{ContainerQueryList}} object,
+ with [=this=] as the <a for=ContainerQueryList>element</a>,
+ as the <a for=ContainerQueryList>document</a>,
```
frivoal: Incomplete sentence fragment. I suspect `as the <a for=ContainerQueryList>document</a>,` should just be deleted, but if not, something is missing.

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+</ol>
+
+<h3 id=the-ContainerQueryList-interface>The {{ContainerQueryList}} Interface</h3>
+
+This section integrates with the <a for=/>event loop</a> defined in HTML. [[!HTML]]
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>container query list</dfn> 
+and an associated 
+<dfn for=ContainerQueryList>element</dfn> set(or assigned?) on creation.
```
frivoal: Which way to go is editorial, but I'd suggest this:
```suggestion
<dfn for=ContainerQueryList>element</dfn> assigned at creation.
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>container</dfn> which is the
```
frivoal: `container` seems to be a poor name for a serialized form of the query. This name suggests it will instead return a pointer to the container against which the query is resolved, which is not what it does.

I'd suggest renaming this to `query`.

The analogous concept in `matchMedia()` is indeed `media`, but that doesn't really work here. Arguably it could have been called `query` there as well, but they certainly didn't call it `document`, which would be the equivalent of calling this `container`.

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>container</dfn> which is the 
+<a lt='serialize a container query list'>serialized</a> 
+form of the associated [=ContainerQueryList/container query list=].
+
+A {{ContainerQueryList}} object has an associated 
+<dfn for=ContainerQueryList>matches state</dfn> 
+which is true if the associated [=ContainerQueryList/container query list=] 
+matches the state of the <a for=ContainerQueryList>query container</a>, and false otherwise.
```
frivoal:
```suggestion
matches the state of  its <a for=ContainerQueryList>element</a>'s [=query container=], and false otherwise.
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+
+When asked to <dfn export>evaluate container queries and report changes</dfn> 
+for a {{query container}} of the element <var>ele</var>, run these steps:
```
frivoal:
```suggestion
for an {{element}} <var>ele</var>, run these steps:
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+<ol>
+
+ <li>
+
+  For each {{ContainerQueryList}} object <var>target</var> that has <var>ele</var>, in the order they were created, oldest first, run these substeps:
```
frivoal:
```suggestion
  For each {{ContainerQueryList}} object <var>target</var> that has <var>ele</var> as its <a for=ContainerQueryList>element</a>, in the order they were created, oldest first, run these substeps:
```

css-conditional-5/Overview.bs
frivoal: Yes, I agree. these are missing, and need to be created similarly to the Media… equivalents.

css-conditional-5/Overview.bs
frivoal: `target` is defined (in line 2041) as a ContainerQueryList. It's `matches state` evaluation does indeed need to look at the element's query container. See https://github.com/w3c/csswg-drafts/pull/13551/changes#r3551143912

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+  <ol>
+
+   <li>If <var>target</var>'s <a>matches state</a> has changed since the last time these steps were run, 
+   <a>fire an event</a> named <a event>change</a> at <var>target</var>
+   using {{ContainerQueryListEvent}},
+   with its {{Event/isTrusted}} attribute initialized to true,
+   its {{ContainerQueryList/container}} attribute initialized to <var>target</var>'s <a>container</a>,
```
frivoal:
```suggestion
   its {{ContainerQueryListEvent/container}} attribute initialized to <var>target</var>'s {{ContainerQueryList/container}},
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+   and its {{ContainerQueryListEvent/matches}} attribute initialized to <var>target</var>'s <a>matches state</a>. <!--fingerprint-->
```
frivoal:
```suggestion
   and its {{ContainerQueryListEvent/matches}} attribute initialized to <var>target</var>'s <a for=ContainerQueryList>matches state</a>.
```

css-conditional-5/Overview.bs
```diff
@@ -1991,11 +1991,115 @@ The <code>CSSContainerRule</code> interface</h3>
...
+partial interface Element {
+    [NewObject] ContainerQueryList matchContainer(CSSOMString query);
+    };
+</pre>
+
+The <dfn method for=PreferenceObject>matchContainer(query)</dfn> method, when invoked, must run these steps:
+
+<ol>
+ <li>Let <var>parsed container query list</var> be the result of
+ <a lt='parse a container query list'>parsing</a>
```
frivoal, line 2116: This seems ambiguous / incomplete. Unlike a `media query list`, and `container query list` isn't a defined concept.

I presume that what we're intending to parse is `<container-condition>#`?

If that's the case, and if we want to keep the parallel with how `matchMedia` is defined, we'd should:
* Adjust the definition at the beginning of [5.4. Container Queries](https://drafts.csswg.org/css-conditional-5/#container-rule) to so that it includes a dfn for `media query list`
* Include, somewhere, an explicit definition of `parse a container query` similar to https://drafts.csswg.org/cssom-1/#parse-a-media-query-list

Alternatively, if what we want to parse is a subset of that (maybe a single `query-in-parens`?) that should be made explicit.

css-conditional-5/Overview.bs
frivoal: Yes we do.

frivoal (changes requested):

css-conditional-5/Overview.bs
DevSDK: Addressed in d83c1b348d7e3184e861fcd0bac81c970e44c4df.

css-conditional-5/Overview.bs
DevSDK: Yeah. I think so. Removed it in d83c1b3.

css-conditional-5/Overview.bs
DevSDK: Fixed d83c1b3.

css-conditional-5/Overview.bs
DevSDK: Yeah I agree with that. Changed it to `query` in d83c1b3

css-conditional-5/Overview.bs
DevSDK: Added d83c1b3

css-conditional-5/Overview.bs
DevSDK: Addressed in d83c1b3

css-conditional-5/Overview.bs
DevSDK: d83c1b3

css-conditional-5/Overview.bs
DevSDK: d83c1b3

css-conditional-5/Overview.bs
DevSDK: Added in d83c1b3

css-conditional-5/Overview.bs
DevSDK: d83c1b3

css-conditional-5/Overview.bs
DevSDK: d83c1b348d7e3184e861fcd0bac81c970e44c4df

DevSDK (commented):
Thanks for the review!

DevSDK:
Regarding the event timing, how about setting the dirty flag during container query evaluation and then triggering it at the same time as ResizeObserver?


Or maybe we can simply mention the **update the rendering** step.

Something like:

```diff
--- a/css-conditional-5/Overview.bs
+++ b/css-conditional-5/Overview.bs
@@ -2141,6 +2141,20 @@
<h3 id=the-ContainerQueryList-interface>The {{ContainerQueryList}} Interface</h3>

This section integrates with the <a for=/>event loop</a> defined in HTML. [[!HTML]]
+
+As part of [=update the rendering=], for each [=fully active=] {{Document}},
+after its style and layout have been updated
+and its post-layout state has been snapshotted
+(see [=run snapshot post-layout state steps=]):
+for each {{element}} <var>ele</var> in the {{Document}}, in [=tree order=],
+that has one or more associated {{ContainerQueryList}} objects,
+run the [=evaluate container queries and report changes=] steps for <var>ele</var>.
+
+Note: Unlike [=media queries=], [=container queries=] depend on the results of
+style and layout, so these steps run after style and layout have been updated,
+rather than alongside [=evaluate media queries and report changes=],
+which is run before them. [[CSSOM-VIEW-1]]

A {{ContainerQueryList}} object has an associated
<dfn for=ContainerQueryList>container query list</dfn>
```

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,153 @@ The <code>CSSContainerRule</code> interface</h3>
...
+ <li>Return a new {{ContainerQueryList}} object,
+ with [=this=] as its <a for=ContainerQueryList>element</a>
+ and <var>parsed container query list</var>
+ as its associated [=ContainerQueryList/container query list=].
+</ol>
+
+To <dfn export>parse a container query list</dfn> for a given string <var>query</var>,
+parse <var>query</var> as a comma-separated list of <<container-condition>>s,
+and return the resulting list.
+
```
frivoal: This is reasonable, but leaning on [how media queries does it](https://drafts.csswg.org/mediaqueries-4/#typedef-media-query-list), we could be a little bit more specific:

```suggestion
To <dfn export>parse a container query list</dfn> for a given string <var>query</var>,
parse <var>query</var> [=parse a comma-separated list of component values|as a comma-separated list of component values=],
then parse each entry in the returned list as a <<container-condition>>,
and return the resulting list.

```

frivoal:
> Or maybe we can simply mention the update the rendering step.
> […]

I think this is reasonable. And after we land this, we should open a PR on HTML to propose upstreaming that change. Once they accept it, it can be removed from here.

frivoal (approved):
The approach suggested in https://github.com/w3c/csswg-drafts/pull/13551#issuecomment-5089370677 should be added to the PR, and maybe you can consider the suggestion I made in https://github.com/w3c/csswg-drafts/pull/13551#discussion_r3767312239, but other than that, as far as I am concerned, this is ready to merge.

We should, however, open a follow up issue to discuss the point raised in https://github.com/w3c/csswg-drafts/pull/13551#issuecomment-4235816324:

> The query itself: Should we accept a container name or not? If we do, what would be a good approach? matchContainer("container name and condition") or matchContainer("condition", containerName)?

css-conditional-5/Overview.bs
fantasai: Should this be a comma-separated list, or should it be a `<<container-query-list>>`?

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,153 @@ The <code>CSSContainerRule</code> interface</h3>
...
+matches the state of its <a for=ContainerQueryList>element</a>'s [=query container=], and false otherwise.
+
+
+When asked to <dfn export>evaluate container queries and report changes</dfn> 
+for an {{element}} <var>ele</var>, run these steps:
+<ol>
+
+ <li>
+
+  For each {{ContainerQueryList}} object <var>target</var> that has <var>ele</var> as its <a for=ContainerQueryList>element</a>, in the order they were created, oldest first, run these substeps:
```
fantasai: Line break...

```suggestion
  For each {{ContainerQueryList}} object <var>target</var> that has <var>ele</var> as its <a for=ContainerQueryList>element</a>,
  in the order they were created, oldest first, run these substeps:
```

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,153 @@ The <code>CSSContainerRule</code> interface</h3>
...
+the associated [=ContainerQueryList/query=].
+
+The <dfn attribute for=ContainerQueryList>matches</dfn> attribute must return
+the associated <a>matches state</a>. <!--fingerprint-->
+
+The following are the <a>event handlers</a> (and their corresponding 
+<a>event handler event types</a>) that must be supported, as 
+<a>event handler IDL attributes</a>, by all objects implementing the 
+{{ContainerQueryList}} interface:
+
```
fantasai: Line breaking and commas...

```suggestion
The following are the <a>event handlers</a>
(and their corresponding <a>event handler event types</a>)
that must be supported as <a>event handler IDL attributes</a>
by all objects implementing the {{ContainerQueryList}} interface:

```

css-conditional-5/Overview.bs
DevSDK: Applied on [a1008d4](https://github.com/w3c/csswg-drafts/pull/13551/commits/a1008d4ea32c1073d24d7a5e179aca49b028a57c) @frivoal 


@fantasai you suggest like this right?
```suggestion
To <dfn export>parse a container query list</dfn> for a given string <var>query</var>,
parse <var>query</var> [=parse a comma-separated list of component values|as a <<container-query-list>>=],
then parse each entry in the returned list as a <<container-condition>>,
and return the resulting list.
```

WDYT @frivoal ?

css-conditional-5/Overview.bs
DevSDK: [Applied](a1008d4ea32c1073d24d7a5e179aca49b028a57c)

css-conditional-5/Overview.bs
DevSDK: [Applied](a1008d4ea32c1073d24d7a5e179aca49b028a57c)

DevSDK:
Event part added [cbd0b20](https://github.com/w3c/csswg-drafts/pull/13551/commits/cbd0b2088ade35e3783fd7927a17517f756498ba)

DevSDK (commented):
Thanks!

css-conditional-5/Overview.bs
frivoal: I think the text should explicitely say "comma-separated list of component values", as that invokes a specific parsing algorithm, including normalization, error handling, etc. The revised proposal links to the right thing, but the text of the link kind of hides it, and its better to be explicit.

frivoal:
Looks good to go to me, and as I said in https://github.com/w3c/csswg-drafts/pull/13551/changes#r3773459486, I don't think we should make the change suggested in https://github.com/w3c/csswg-drafts/pull/13551/changes#r3768743440.

@fantasai, is that OK with you?

fantasai:
I defer to @tabatkins on this, and he says to keep the current wording.

tabatkins:
Specifically, the container-query-list should parse the same way as media-query-list (manually split on commas using the Syntax algo, then parse each segment as a container-query, so we can drop a malformed CQ without dropping the entire list)

frivoal:
@tabatkins agreed, and I think the current text of the PR does that.
```html
To <dfn export>parse a container query list</dfn> for a given string <var>query</var>,
parse <var>query</var> [=parse a comma-separated list of component values|as a comma-separated list of component values=],
then parse each entry in the returned list as a <<container-condition>>,
and return the resulting list.
```
Good to go?

tabatkins:
Yes, I just wanted to make sure my intent was clear; fantasai's comment was somewhat vague about what I was aiming for. ^_^

Do we need to filter out failures from the list, or is that handled elsewhere in the text?

DevSDK:
@tabatkins 
I think this part covers that (diff [here](https://github.com/w3c/csswg-drafts/pull/13551/changes#diff-916e3a3bc30c2b99ad6db096c6785a59d51fdf07e4a55f766d17f974664b3be0R1137-R1145)):

```

	A <<container-query-list>> is a <dfn export>container query list</dfn>.

	If a [=container query list=] includes multiple <<container-condition>>s,
	each condition will select its own [=query container=],
	and evaluate independently.
	A [=container query list=] is ''true'' if <em>any</em> of its component
	<<container-condition>>s are ''true'',
	and ''false'' only if <em>all</em> of its component
	<<container-condition>>s are ''false''.
```

DevSDK:
I just created an issue related to the event dispatch part: [https://github.com/w3c/csswg-drafts/issues/14430](https://github.com/w3c/csswg-drafts/issues/14430) 

How about mentioning that issue in this change and addressing it in a separate PR?

css-conditional-5/Overview.bs
```diff
@@ -1133,10 +1134,12 @@ Container Queries: the ''@container'' rule</h3>
 	If the <<container-query>> is omitted, the [=query container=] is eligible as
 	long as the <<container-name>> matches.
 
-	If a [=container query=] includes multiple <<container-condition>>s,
-	each condition will select it's own [=query container=],
+	A <<container-query-list>> is a <dfn export>container query list</dfn>.
```
fantasai: This `<dfn>` is missing an actual defining sentence. What _is_ a "container query list", essentially (rather than just syntactically) speaking?

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,168 @@ The <code>CSSContainerRule</code> interface</h3>
...
-	but it provides the additional Container Query syntax and features.
+<h3 id=the-matchContainer-method>The matchContainer() method</h3>
+
+<pre class=idl>
+partial interface Element {
+    [NewObject] ContainerQueryList matchContainer(CSSOMString query);
+    };
+</pre>
+
+The <dfn method for=PreferenceObject>matchContainer(query)</dfn> method, when invoked, must run these steps:
```
fantasai: This sentence defines how to implement it. But what is it? It's a method on the element that returns true when the query argument matches as a container query.

We want to make the spec readable for humans, not just spec-to-C++ compilers, so let's actually build up their understanding of what's going on.

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,168 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+<ol>
+ <li>Let <var>parsed container query list</var> be the result of
+ <a lt='parse a container query list'>parsing</a>
+ <var>query</var>.
+ <li>Return a new {{ContainerQueryList}} object,
+ with [=this=] as its <a for=ContainerQueryList>element</a>
+ and <var>parsed container query list</var>
```
fantasai:
```suggestion
 and the <var>parsed container query list</var>
```

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,168 @@ The <code>CSSContainerRule</code> interface</h3>
...
+ as its associated [=ContainerQueryList/container query list=].
+</ol>
+
+To <dfn export>parse a container query list</dfn> for a given string <var>query</var>,
+parse <var>query</var> [=parse a comma-separated list of component values|as a comma-separated list of component values=],
+then parse each entry in the returned list as a <<container-condition>>,
+and return the resulting list.
+
+To <dfn export>serialize a container query list</dfn> run these steps:
```
fantasai: Does this belong under the matchContainer() heading? It doesn't seem like matchContainer() is using it...

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,168 @@ The <code>CSSContainerRule</code> interface</h3>
...
+As part of [=update the rendering=], for each [=fully active=] {{Document}},
+after its style and layout have been updated
+and its post-layout state has been snapshotted
+(see [=run snapshot post-layout state steps=]):
+for each {{element}} <var>ele</var> in the {{Document}}, in [=tree order=],
+that has one or more associated {{ContainerQueryList}} objects,
+run the [=evaluate container queries and report changes=] steps for <var>ele</var>.
+
+Note: Unlike [=media queries=], [=container queries=] depend on the results of
+style and layout, so these steps run after style and layout have been updated,
```
fantasai: Line break...

```suggestion
Note: Unlike [=media queries=], [=container queries=] depend on the results of style and layout,
so these steps run after style and layout have been updated,
```

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2101,168 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+ <li>Serialize each <<container-condition>> in the list, in the order they appear,
+ and return the results as a comma-separated list.
+ Each <<container-condition>> is serialized as specified,
+ with whitespace simplification but no logical simplification,
+ in the same manner as the {{CSSContainerRule/containerName}} and
+ {{CSSContainerRule/containerQuery}} attributes.
+</ol>
+
+<h3 id=the-ContainerQueryList-interface>The {{ContainerQueryList}} Interface</h3>
```
fantasai: Review this section from the perspective of an author. What information do you want to see first? Definitely, it's not the integration with the HTML event loop. It's the interface and the core definition of what this object is, and what it's parts are.

Reorder the paragraphs so that the section is engaging the reader on topics in an order that builds up an understanding of what we're doing and why first, and how specifically later.

css-conditional-5/Overview.bs
DevSDK: Changed into:
```suggestion
	A <dfn export>container query list</dfn> (<<container-query-list>>) is a condition
	composed of one or more <<container-condition>>s.
```

css-conditional-5/Overview.bs
```diff
@@ -1133,10 +1134,12 @@ Container Queries: the ''@container'' rule</h3>
 	If the <<container-query>> is omitted, the [=query container=] is eligible as
 	long as the <<container-name>> matches.
 
-	If a [=container query=] includes multiple <<container-condition>>s,
-	each condition will select it's own [=query container=],
+	A <<container-query-list>> is a <dfn export>container query list</dfn>.
+
+	If a [=container query list=] includes multiple <<container-condition>>s,
```
DevSDK: Just removed if- sentence.

css-conditional-5/Overview.bs
DevSDK: It seems `matchMedia()` in cssom-view has the [same issue](https://github.com/w3c/csswg-drafts/blob/8a3430c05846c3b8bf6e98f6b390d286191943d1/cssom-view-1/Overview.bs#L552). I'll file an issue regarding that.

Addressed in [the change](https://github.com/w3c/csswg-drafts/pull/13551/changes/3c9920aca49cdcb5a67ba7e0ac26b4fce66347ab#diff-916e3a3bc30c2b99ad6db096c6785a59d51fdf07e4a55f766d17f974664b3be0R2113-R2124)

css-conditional-5/Overview.bs
DevSDK: Fixed

css-conditional-5/Overview.bs
DevSDK: It is used by `ContainerQueryList`. [Moved below to that section](https://github.com/w3c/csswg-drafts/pull/13551/changes/3c9920aca49cdcb5a67ba7e0ac26b4fce66347ab#diff-916e3a3bc30c2b99ad6db096c6785a59d51fdf07e4a55f766d17f974664b3be0R2206)

css-conditional-5/Overview.bs
DevSDK: Thanks for catching this. Yes, the event loop integration was in the way before the interface definition. Reordered the section so the IDL and the object's parts come first, and moved the event loop integration into its own subsection at the end.

css-conditional-5/Overview.bs
DevSDK: Fixed.

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2102,180 @@ The <code>CSSContainerRule</code> interface</h3>
...
+A {{ContainerQueryList}} object has an associated
+<dfn for=ContainerQueryList>matches state</dfn>
+which is true if the associated [=ContainerQueryList/container query list=]
+matches the state of its <a for=ContainerQueryList>element</a>'s [=query container=], and false otherwise.
+
+The <dfn attribute for=ContainerQueryList>query</dfn> attribute must return
+the associated [=ContainerQueryList/query=].
+
+The <dfn attribute for=ContainerQueryList>matches</dfn> attribute must return
+the associated <a>matches state</a>. <!--fingerprint-->
```
fantasai: Fold the definitions of the conceptual "match state" and "query" into the definitions of the attributes. This will reduce the amount of cross-lookup required to understand what's happening.

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2102,180 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+The following are the <a>event handlers</a>
```
fantasai: Should this have an H4?

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2102,180 @@ The <code>CSSContainerRule</code> interface</h3>
...
+            &hellip;
+    }
+
+    const element = document.getElementById("target");
+    var cql = element.matchContainer("(scroll-state(scrollable))");
+    cql.onchange = handleOverflowChange;
+    </pre>
+</div>
+
+To <dfn export>serialize a container query list</dfn> run these steps:
```
fantasai: Should this have an H4?

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2102,180 @@ The <code>CSSContainerRule</code> interface</h3>
 			</div>
 	</dl>
 
-	Issue(6205): Container Queries should have a <code>matchContainer</code> method.
-	This will be modeled on {{matchMedia()}} and the {{MediaQueryList}} interface,
-	but applied to Elements rather than the Window.
-	When measuring layout sizes, it behaves Similar to <code>resizeObserver</code>,
-	but it provides the additional Container Query syntax and features.
+<h3 id=the-matchContainer-method>The matchContainer() method</h3>
```
fantasai: Line break here:


```suggestion
<h3 id=the-matchContainer-method>
The matchContainer() method</h3>
```

Same throughout. You also need to indent all the non-header content by one tab character to match the rest of the file.

This is a bit of a weird formatting scheme, but it's very intentional: it puts the text of the heading flush with the left edge, and indents all the content under it, which makes it easy to scan the source file.

(Note that subsections are not further indented, as they would be with `<section>`, so that headings can be inserted, removed, moved, upgraded, and downgraded without affecting any surrounding content.)

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2102,180 @@ The <code>CSSContainerRule</code> interface</h3>
...
+The <dfn method for=Element>matchContainer(query)</dfn> method, when invoked, must run these steps:
+
+<ol>
+ <li>Let <var>parsed container query list</var> be the result of
+ <a lt='parse a container query list'>parsing</a>
+ <var>query</var>.
+ <li>Return a new {{ContainerQueryList}} object,
+ with [=this=] as its <a for=ContainerQueryList>element</a>
+ and the <var>parsed container query list</var>
+ as its associated [=ContainerQueryList/container query list=].
```
fantasai: LI contents should be indented.

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2102,180 @@ The <code>CSSContainerRule</code> interface</h3>
...
+
+ <li>Serialize each <<container-condition>> in the list, in the order they appear,
+ and return the results as a comma-separated list.
+ Each <<container-condition>> is serialized as specified,
+ with whitespace simplification but no logical simplification,
+ in the same manner as the {{CSSContainerRule/containerName}} and
+ {{CSSContainerRule/containerQuery}} attributes.
+</ol>
+
+<h4 id=evaluate-container-queries>Evaluating Container Queries and Reporting Changes</h4>
```
fantasai: This section and the next section probably should be combined. Think about how you'd want to encounter the information as an author, and reorganize it accordingly.

DevSDK: I moved the "Evaluating Container Queries and Reporting Changes" section under "The ContainerQueryListEvent Interface".
 
  I also considered giving the event its own top-level section instead:
  ```
H3 The change Event
    H4 The ContainerQueryListEvent Interface
    H4 Evaluating Container Queries and Reporting Changes
  ```

That reads a bit more naturally from an author's point of view, but I went with the simpler move for consistency with the sibling sections, which are all "The X Interface".

Please let me know if you prefer that way.

css-conditional-5/Overview.bs
DevSDK: Fixed in c9cdfa273fce18ea6d8c177a39e3b8d87fe0c3c6

css-conditional-5/Overview.bs
DevSDK: Also fixed in c9cdfa273fce18ea6d8c177a39e3b8d87fe0c3c6

css-conditional-5/Overview.bs
DevSDK: Edited in 9f46aff62180c1f1db7732118d6195f3c7c098ca

css-conditional-5/Overview.bs
DevSDK: Sure! f1242bddcae9d4233731020f822de956c65b43d1

css-conditional-5/Overview.bs
DevSDK: f1242bddcae9d4233731020f822de956c65b43d1

css-conditional-5/Overview.bs
DevSDK: I moved the "Evaluating Container Queries and Reporting Changes" section under "The ContainerQueryListEvent Interface" in https://github.com/w3c/csswg-drafts/commit/f1242bddcae9d4233731020f822de956c65b43d1
 
  I also considered giving the event its own top-level section instead:
  ```
   H3 The change Event
     H4 The ContainerQueryListEvent Interface
	 H4 Evaluating Container Queries and Reporting Changes
  ```

That reads a bit more naturally from an author's point of view, but I went with the simpler move for consistency with the sibling sections, which are all "The X Interface".

Please let me know if you prefer that way.

tabatkins:
@DevSDK
> I think this part covers that (diff [here](https://github.com/w3c/csswg-drafts/pull/13551/changes#diff-916e3a3bc30c2b99ad6db096c6785a59d51fdf07e4a55f766d17f974664b3be0R1137-R1145)):

Not quite. That's about resolving the combined truth value of multiple queries, but the result of the parsing is that the query list can contain failures (which are *not* queries, they're just... failures). They need to be filtered out, or otherwise handled in some way.

DevSDK:
@tabatkins Ahh got it. I changed the spec text to handle it ([c31154f](https://github.com/w3c/csswg-drafts/pull/13551/commits/c31154f28426f85d480aca7e25dcf73d619b32a1)). Rather than filtering, I'd like to match how the `@container` rule parses here by adopting [the behavior](https://drafts.csswg.org/cssom/#dom-cssstylesheet-insertrule) of `insertRule()` and throwing a `SyntaxError`.
 E.g.,
```js
   sheet.insertRule("@container foo bar, (max-width: 300px) { .x { color: red } }");
  // SyntaxError
```
  
and the same string behaves the same way in `matchContainer()`:
```js
  el.matchContainer("foo bar, (max-width: 300px)");
 // SyntaxError
```
   
whereas an unknown feature inside parens still parses in both and just evaluates to unknown:
   
  ```js
  el.matchContainer("(hiedg: 10px), (max-width: 300px)").query
  // "(hiedg: 10px), (max-width: 300px)"
  ```

matching what `@container`:

```js
  const i = sheet.insertRule("@container (hiedg: 10px), (max-width: 300px) { .x { color: red } }");
  sheet.cssRules[i].containerQuery
  // "(hiedg: 10px), (max-width: 300px)"
```

css-conditional-5/Overview.bs
```diff
@@ -2098,11 +2102,179 @@ The <code>CSSContainerRule</code> interface</h3>
...
+	  CSSOMString query = "";
+	  boolean matches = false;
+	};
+	</pre>
+
+	The <dfn attribute for=ContainerQueryListEvent>query</dfn> attribute must return
+	the value it was initialized to.
+
+	The <dfn attribute for=ContainerQueryListEvent>matches</dfn> attribute must return
+	the value it was initialized to.
```
fantasai, line 2237: Ok, but what is this event? What fires it, and what do we initialize these values to?

Remember, you're reading this document top to bottom. It should make sense in that order. Things should be defined up front, and readers shouldn't need to dig deep into the algorithms to understand at a basic level what something does. You can reference the algorithm for specifics, but you shouldn't require the reader to internalize all of it to make sense of what's going on.

css-conditional-5/Overview.bs
fantasai: The structure you have seems fine. :)

css-conditional-5/Overview.bs
DevSDK: Changed bf2bb0e

Issue #13355, opened 2026-01-15
[css-forms-1] ::field-component and ::field-separator should be children of ::field-text


fantasai:
For internationalization, localization, and platform-dependent reasons, we can't let authors perform arbitrary layout of ::field-component and ::field-separator. Like, reordering the components using flexbox and `order` would be a real problem: if they hacked it up to work in their own OS configuration it's unlikely to work for anyone else. But it seems reasonable that the author might want to style various bits within the control while keeping this stream of text content as a unit, so I making these all children of a more `::field-text` pseudo-element might be a way forward here. We can force the `display` value of `::field-text` to `flow` and that limits the kind of trouble they can cause within the date/time text representation while still allowing the input field itself and other things like the picker-icon to be styled.

It also gives us a consistent way to style the inner text of a text-input style field: `::field-text` is always available.

See also https://github.com/w3c/csswg-drafts/issues/11852

PR #13556, opened 2026-02-24
[css-forms-1] Update structure and styles for temporal input

lukewarlow:
::field-component and ::field-separator are now children of ::field-text.

Also applies default styles.

Per https://github.com/w3c/csswg-drafts/issues/13355

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -568,13 +568,13 @@ spec:css-forms-1; type:value; for:/; text:::placeholder
 
     ```
     <input type="date">
-    ├─ ::field-text
-    │  ├─ ::field-component (08)
-		│  ├─ ::field-separator (/)
-    │  ├─ ::field-component (22)
-    │  ├─ ::field-separator (/)
-    │  └─ ::field-component (2024)
-    └─ ::picker-icon
+      ├─ ::field-text
+      │  ├─ ::field-component (08)
+      │  ├─ ::field-separator (/)
+      │  ├─ ::field-component (22)
+      │  ├─ ::field-separator (/)
+      │  └─ ::field-component (2024)
+      └─ ::picker-icon
     ```
   </div>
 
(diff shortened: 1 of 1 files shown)


css-forms-1/Overview.bs
```diff
@@ -1113,6 +1118,31 @@ select {
...
 ```
 
+### Temporal inputs ### {#stylesheet-temporal-inputs}
+
+```css
+input[type="date"]::field-text,
+input[type="datetime-local"]::field-text,
+input[type="month"]::field-text,
+input[type="time"]::field-text,
+input[type="week"]::field-text {
```
lukewarlow: This should perhaps be just input::field-text per the resolution, I'm not sure if we only want to protect display for these specific inputs though?

css-forms-1/Overview.bs
```diff
@@ -1113,6 +1118,31 @@ select {
...
+	display: flow !important;
+}
+
+input[type="date"]::field-component,
```
lukewarlow: As these are children of field-text now do these styles still work, I'm guessing yes because slider-fill does? I want to be as specific as possible for performance reasons.

css-forms-1/Overview.bs
```diff
@@ -564,11 +568,12 @@ spec:css-forms-1; type:value; for:/; text:::placeholder
...
     ```
     <input type="date">
-    ├─ ::field-component (08)
-    ├─ ::field-separator (/)
-    ├─ ::field-component (22)
-    ├─ ::field-separator (/)
-    ├─ ::field-component (2024)
+    ├─ ::field-text
+    │  ├─ ::field-component (08)
+		│  ├─ ::field-separator (/)
```
nt1m: indentation is off here

css-forms-1/Overview.bs
nt1m: I think we should just write `input::field-text` here, it shouldn't be limited to date time inputs. Let's limit to display for now.

css-forms-1/Overview.bs
nt1m: The attribute selectors don't really help much for performance in the engine. I'd rather optimize the spec for legibility. Can we use `input::field-component, input::field-separator` ?

lukewarlow: I thought the type attributes were bucketed? Either way I'm fine, provided that if engines are more specific it won't be web observable?

PR #13568, opened 2026-02-26
[css-forms-1] Fix select option enabled selector

### LINKED ISSUE #13383: [css-forms-1] option hover/active styles inside of disabled select listbox
Currently per the CSS forms spec (and as implemented in chromium) if a select renders as a listbox and you hover an option within, it gets a hover style. This feels misplaced as it's disabled and so you can't interact.

My proposal would be:

Before:

```css
select option:enabled:hover {
  background-color: color-mix(in lab, currentColor 10%, transparent);
}
select option:enabled:active {
  background-color: color-mix(in lab, currentColor 20%, transparent);
}
```

After:

```css
select:enabled option:enabled:hover {
  background-color: color-mix(in lab, currentColor 10%, transparent);
}
select:enabled option:enabled:active {
  background-color: color-mix(in lab, currentColor 20%, transparent);
}
```

josepharhar:
option elements can match :enabled while inside a disabled select, so the option:enabled selectors also need to include select:enabled.

Fixes https://github.com/w3c/csswg-drafts/issues/13383

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -1010,10 +1010,10 @@ select option {
    * of making options tall. */
   white-space: nowrap;
 }
-select option:enabled:hover {
+select:enabled option:enabled:hover {
   background-color: color-mix(currentColor 10%, transparent);
 }
-select option:enabled:active {
+select:enabled option:enabled:active {
   background-color: color-mix(currentColor 20%, transparent);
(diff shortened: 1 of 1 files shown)


lukewarlow (approved):

lukewarlow:
Will merge once html pr is merged.

PR #13575, opened 2026-03-01
[mediaqueries-5] Lift parens to <media-in-parens>

Calme1709:
This change was made to level 4 in 1af56e0 but was never carried over to level 5. This is relied on by the `<if-test>` grammar in css-values-5


cdoublev:
I already suggested this change [two years ago](https://github.com/w3c/csswg-drafts/pull/10131). (So MQ authors might have reasons for not being interested in this change.)

tabatkins:
Nah, we're just bad at keeping on top of PRs. I've merged @cdoublev 's original PR, as it fixes slightly more than this one. Thanks!

Issue #11424, opened 2024-12-31
[css-values-5] Invalidation of `ident()`


bramus:
This is a follow-up to #9141 which introduces `ident()`. Something that is not defined yet is what should happen when the input for `ident()` is invalid.

E.g. what happens if you do the following:

```css
@property --length {
  syntax: "<length>";
  inherits: false;
  initial-value: 0px;
}

.target {
  --length: 120px;
  anchor-name: ident("--anchor-" var(--length));
}
```

Because `--length` is a `<length>`, `ident()` can’t generate a valid `<custom-ident>`. I assume that in this case the declaration should become Invalid At Computed Value Time (IACVT) and the value should be considered to be `unset`.

However, we might also consider adding an extra argument to `ident()`, to allow a fallback value in case the constructed value is not a valid `<custom-ident>`.

The definition for `ident()` would then become the following:

```
<ident-fn> = ident(<ident-arg>+, <declaration-value>?);
<ident-arg> = <string> | <integer> | <ident>
```

I prefer this last version, as `ident()` then behaves similar to `var()` and `attr()`: when the input cannot be parsed into a `<custom-ident>` _(or is the guaranteed invalid value)_ then the fallback value is used.

LeaVerou:
Yup, a fallback would be great, and the declaration is just invalid if no fallback is provided. I don't think IACVT needs to be explicitly mentioned here: just like any other function, it can be invalid at parse time if the arguments are known at parse time (e.g. `ident(foo 12px)`.

Just let's make sure that we don't accidentally make things invalid that don't need to be. E.g. people should be able to use an integer `<number>` in there without having to explicitly get it through `round()`.

bramus:
> Just let's make sure that we don't accidentally make things invalid that don't need to be. E.g. people should be able to use an integer `<number>` in there without having to explicitly get it through `round()`.

That is/was discussed in https://github.com/w3c/csswg-drafts/issues/11551

tabatkins:
> Something that is not defined yet is what should happen when the input for ident() is invalid.

Nope, that's perfectly defined already - the function is invalid because it doesn't match the specified grammar. (And because that happens due to var() substitution, it's IACVT rather than a parse failure.)

I don't think we really need a fallback here. In general, CSS functions don't guard against the possibility of some argument being invalid from a nonsense var() substitution. I'm not seeing anything special about ident() that would make it more necessary to add special handling.

LeaVerou:
[...]

PR #13578, opened 2026-03-01
[css-values-5] Remove a closed issue denote on ident() fallback

sakupi01:
This removes the open issue marker(whether `ident()` should allow a fallback value #11424) as it seemed to be resolved without any change.
I also added a short Note explaining why there's no fallback, since authors familiar with var()/attr() might wonder. Happy to drop the Note if it feels unnecessary!

### DIFF
--- a/css-values-5/Overview.bs
+++ b/css-values-5/Overview.bs
@@ -2175,10 +2175,16 @@ Constructing <<custom-ident>> values: the ''ident()'' function</h3>
 	<dfn><<ident-arg>></dfn> = <<string>> | <<integer>> | <<ident>>
 	</pre>
 
-	Issue(w3c/csswg-drafts#11424): Should we allow a fallback value?
-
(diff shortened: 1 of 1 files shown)


sakupi01:
Hmm, seems like something went wrong in ipr... I'm a participant of this working group but it seems like ash-nazg can't resolve my GitHub account to my W3C account for some reason. 
Could someone help me figure this out?

astearns:
Can you check https://www.w3.org/users/myprofile/ and make sure your GitHub account is connected there?

sakupi01:
@astearns 

Thank you! I had accidentally connected my GitHub account to wrong W3C account. Now it should work...
https://www.w3.org/groups/wg/css/participants/

tabatkins:
I'm gonna close this PR and just remove the issue; we don't actually need an explanatory note here. `ident()` is the same as any other normal function in this regard.

PR #13580, opened 2026-03-01
Make the index remember the recent/grouped view

andreubotella:


sideshowbarker:
CI job failing: https://github.com/w3c/csswg-drafts/actions/runs/22549167117/job/65316266251#step:3:1

andreubotella:
> CI job failing: https://github.com/w3c/csswg-drafts/actions/runs/22549167117/job/65316266251#step:3:1

This CI job is for autopublishing css-fonts-4, which this PR doesn't change, and it's only failing because a WPT test was added which isn't listed yet in the spec.

@svgeesus With a repo with so many specs, I don't think it's really reasonable to fail CI for a PR that doesn't touch the autopublished specs, especially when the failure can be due to a completely external change.

andreubotella:
> CI job failing: https://github.com/w3c/csswg-drafts/actions/runs/22549167117/job/65316266251#step:3:1

I just rebased, and it's not failing anymore after #13581 and the corresponding updates to the autopublish workflows.

Merging now.

PR #13581, opened 2026-03-02
helper to generate auto publication workflows

deniak:
Based on @tidoust's [suggestion](https://github.com/w3c/csswg-drafts/pull/13553#pullrequestreview-3848787987-permalink), that PR adds a script to help generate the auto publication workflows for a given list of specs.
That PR assumes that the source document is named Overview.bs and each folder matches the shortname of the spec.

### DIFF
--- a/.github/generate-auto-publish-workflows.py
+++ b/.github/generate-auto-publish-workflows.py
@@ -30,7 +30,7 @@ SPECS = [
         "publicationStatus": "WD",
     },
     {
-        "shortname": "css-fonts-4",
+        "shortname": "css-fonts-5",
         "publicationStatus": "WD",
     },
 ]


.github/auto-publish-template.yml
```diff
@@ -0,0 +1,33 @@
...
+# This is the template used to generate auto-publication workflows for
+# the different documents in the repository. Check generate script for
+# details. Edit this file and run the script again to update the
+# workflows.
+######################################################################
+
+name: Publish {{shortname}} on /TR
+
+on:
+  push:
```
tidoust, line 14: The current workflows have an unconditional `pull_request` trigger, which is too unconstrained because it makes them run on each and every pull request, regardless of what the pull requests are for.

It still seems a good idea to validate pull requests against a particular spec to detect problems at the PR level, though? (especially since the other build job does not run on pull requests either).

```suggestion
  pull_request:
    paths:
      - "{{shortname}}/**"
  push:
```

svgeesus:
I see a failing check, but don't understand it. @deniak @tidoust

.github/generate-auto-publish-workflows.py
```diff
@@ -0,0 +1,94 @@
...
+    {
+        "shortname": "css-color-4",
+        "publicationStatus": "CRD",
+    },
+    {
+        "shortname": "css-color-5",
+        "publicationStatus": "WD",
+    },
+    {
+        "shortname": "css-fonts-4",
```
svgeesus:
```suggestion
        "shortname": "css-fonts-5",
```

svgeesus:
So, now that this is merged, what triggers the python script to run? Or do we run it locally and just commit the results?

deniak:
> So, now that this is merged, what triggers the python script to run? Or do we run it locally and just commit the results?

The script needs to be run manually only when there's a new spec to be autopublished or if the template is updated.

deniak:
> I see a failing check, but don't understand it. @deniak @tidoust

I see the recent checks passed. Please provide a link to the error if I missed something.

svgeesus:
> I see the recent checks passed. Please provide a link to the error if I missed something.

It is because Bikeshed warnings prevent publication. Bikeshed warns on variables only used once. I had been ignoring that because bikeshed is wrong (it does not consider `foo` and `foo[2]` to be two uses of `foo` but I changed the spec to prevent the warning and now it auto-publishes correctly.

tabatkins:
> (it does not consider foo and foo[2] to be two uses of foo

Since spec variables are in no particular language, there is no way for me to tell what part of the contents of a `<var>` are the "name" and what parts are "other stuff". So I'm as right as possible here. If you want it to work "correctly", `<code>|foo|[2]</code>` to just mark up the intended variable name as the var.

Issue #10975, opened 2024-09-30
[selectors] :local-link should have a more precise name


jyasskin:
https://drafts.csswg.org/selectors-4/#the-local-link-pseudo

I noticed this because of @keithamus' [Intent to Prototype](https://groups.google.com/a/chromium.org/g/blink-dev/c/z2Y7O5ZCxgc/m/cj0paGiPAQAJ). I'm concerned because "local" has at least 5 plausible meanings for link targets, with a couple more less-plausible ones, and I don't think the current syntax makes it obvious which you're going to get:

* Links within the same site. (Not in any current CSS proposal, and not necessarily desirable. Arguably the intro text should avoid "[within a site](https://drafts.csswg.org/selectors-4/#the-local-link-pseudo:~:text=within%20a%20site)" to exclude this option.)
* Links within the same origin. (Handled via :local-link(0) in https://drafts.csswg.org/selectors-5/#local-pseudo)
* Links within the same path prefix, ignoring queries. (Handled by :local-link(N) in [selectors-5](https://drafts.csswg.org/selectors-5/#local-pseudo).)
* Links within the same page. (Handled via :local-link on a fragment-less link.)
* Links to the current fragment. (Handled via :local-link on a fragment-ful link.)

And the less-plausible ones:
* A "local" network address (127.0.0.1 or ::1): https://wicg.github.io/private-network-access/#ip-address-space-local
* mdns names (foo.local): https://en.wikipedia.org/wiki/.local

I can see an argument that the desired behavior will tend to follow the fragment-ness of the link, so authors will usually get the right behavior by accident, but it still seems like a footgun that adding `#` to the end of your link will significantly change when this selector matches.

I could imagine :same-site-link, :same-origin-link, :same-page-link, and :same-fragment-link handling the first 4. :same-path-link(N) might handle the 5th: it seems more obvious that the integer argument counts path segments if "path" is mentioned in the selector.

If the WG decides that the current name is right, I don't intend to argue further, but it seems worth double-checking.

nt1m:
I agree with @jyasskin here, there's so many meanings you can put to "local link". Would love for this to be revisited.

LeaVerou:
@astearns since this is being [implemented](https://chromium-review.googlesource.com/c/chromium/src/+/5909506) could we prioritize it and add it to the next telcon's agenda?

LeaVerou:
[...]

PR #13585, opened 2026-03-03
[selectors-5][editorial] Add formal definition and production rule of :local-link()

cdoublev:
This PR wraps the first occurrence of `:local-link()` in `<dfn>` and adds a production rule, following [`:heading`/`:heading()`](https://drafts.csswg.org/selectors-5/#headings) as an example.

`:local-link` has an associated `<dfn>` but `:local-link()` does not. So `@webref/css` includes the first, not the second.

I am aware that there are [discussions](https://github.com/w3c/csswg-drafts/issues/10975) about renaming them, but this "in-between" state is confusing.

### DIFF
--- a/selectors-5/Overview.bs
+++ b/selectors-5/Overview.bs
@@ -66,9 +66,14 @@ The Local Link Pseudo-class '':local-link''</h3>
 	</div>
 
 	As a functional pseudo-class,
-	'':local-link()'' can also accept a non-negative integer as its sole argument,
-	which, if the document's URL belongs to a hierarchical scheme,
-	indicates the number of path levels to match:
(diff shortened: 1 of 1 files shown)


fantasai (changes requested):
I think I would keep the sentence as it was previously (because that flows nicely and quickly gives the conceptual definition up front), and then just use `<<integer>>` in the definition.

cdoublev:
I have no problem restoring the initial sentence flow but note that [`:heading()`](https://drafts.csswg.org/selectors-5/#heading-functional-pseudo) takes `<level>`, which is defined as a `<number-token>`, which makes numeric substitutions (math functions and tree-counting functions) invalid.

Do you intentionally want to accept `<integer>` and these functions in `:local-link()`?

PR #13596, opened 2026-03-04
[css-page-3][editorial] Set the initial page-margin-safety value to "none"

cdoublev:
The definition table of [`page-margin-safety`](https://drafts.csswg.org/css-page-3/#descdef-page-page-margin-safety) has `auto` for its initial value, but it is not included in its value definition.


fantasai (changes requested):
But there is no `auto` value defined for `page-margin-safety` so I think the change should actually be to replace the initial value with `none`. :)

https://github.com/w3c/csswg-drafts/issues/11395#issuecomment-3435760484

PR #13598, opened 2026-03-04
[css-forms-1] Add styles for `::color-swatch`

### LINKED ISSUE #12142: [css-forms-1] Styles for color input
As currently specced the color input never renders the actual chosen color. We need to specify styles specific to the color input. This will track that.

Proposed additions

```css
::color-swatch {
    /* Ensures contrast between chosen color and button background (which by default is transparent so the page background) */
    border: 1px solid currentColor;

    /* Enforces a reasonable minimum size */
    min-inline-size: 2em; 
    min-block-size: 1em;

    /* Ensures it fills it's parent container */
    flex: 1; 

    /* Should this be here? It's designed to prevent forced dark mode from messing with the colour? */
    color-scheme: light only;

    /* This uses a linear gradient to handle showing the alpha of the control value. */
    background: linear-gradient(control-value(<color>)),
      linear-gradient(to bottom right, black 50%, white 50%);
}
```

This should result in the below rendering given `<input type="color" value="color(srgb 0 1 0 / 0.75)" alpha>`

Image

lukewarlow:
These styles along with the existing ones will create an output approximately like:

image

Fixes #12142

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -808,8 +808,6 @@ spec:css-forms-1; type:value; for:/; text:::placeholder
 
   ISSUE: This section needs refining with implementation.
 
-  ISSUE(11837): Color input styles need refining.
-
 ## Basics ## {#stylesheet-basics}
 
 ```css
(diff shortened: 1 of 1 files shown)


lukewarlow:
I believe this matches the resolutions we made. I've left border-radius as is for now as that's a wider question.

Issue #12163, opened 2025-05-05
How to handle addEventListener on `CSSPseudoElement`?


danielsakhapov:
Full context: https://github.com/danielsakhapov/CSSPseudoElementDoc?tab=readme-ov-file#41-addeventlistener-on-a-proxyhandle

Description:
Given that the CSSPseudoElement object acts as a persistent handle, attaching an event listener via addEventListener raises further questions. 

* What does it mean to listen for an event on a handle when the actual pseudo element might not be currently rendered (e.g., due to display: none)?
* Should the listener only become active when the pseudo element is actually rendered?
* How does this interact with the retargeting behavior?
* If event.target remains the Element, how does a listener on the CSSPseudoElement get invoked?
* Does it require a special dispatch phase?

This would require some heavy discussions, as it’s already known that authors want to be able to count clicks on ::scroll-marker pseudo elements for analytics purposes.

Proposed variants are included in the table:
| Model                      | `event.target` Behavior (for existing events like `click`) | `addEventListener` on Pseudo? | Pros                                                                         | Cons                                                                                                  | Feasibility Notes                                                                                             |
|----------------------------|-----------------------------------------------------------|------------------------------|------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------|
| A: Status Quo              | `Element` (retargeted)                                    | No                           | Full web compatibility. Simple.                                              | No direct script interaction with pseudo events.                                                      | Current state.                                                                                                |
| B: Full `EventTarget`      | `CSSPseudoElement`                                        | Yes                          | Direct event handling on pseudo. Conceptually clean object model.            | Major breaking change for web compatibility. Complex bubbling definition needed, esp. for non-tree pseudos. | Highly unlikely due to web compatibility constraints.                                                         |
| C: Hybrid (Parent + Selector) | `Element` (retargeted)                                    | No (Interaction via parent)  | Avoids web compat issues. Sidesteps pseudo lifetime/eventing complexity for the use case. | Less direct object model. Requires specific API pattern (like Web Animations `pseudoElement`).           | Proven feasible for Web Animations L1. Potentially applicable to other interactions.                        |
[...]

PR #13603, opened 2026-03-05
[css-animations-1][css-transitions-1] Add .pseudoTarget to AnimationEvent and TransitionEvent

danielsakhapov:
As resolved in w3c/csswg-drafts#12163, add .pseudoTarget to selected event types and describe the initialization process for it.

### DIFF
--- a/css-animations-1/Overview.bs
+++ b/css-animations-1/Overview.bs
@@ -932,6 +932,13 @@ Attributes</h4>
 
 			The <a>un-initialized value</a> of this attribute MUST be
 			<code>null</code>.
+
+			Note: {{AnimationEvent/pseudoTarget}} and {{AnimationEvent/pseudoElement}} are intentionally
+			not equivalent. Because {{AnimationEvent/pseudoTarget}} returns a {{CSSPseudoElement}}
+			object, it is subject to retargeting and encapsulation. Therefore, if the event's target
+			is a pseudo-element inside a shadow tree, {{AnimationEvent/pseudoTarget}} will
+			return <code>null</code> when observed from outside the shadow tree, while
+			{{AnimationEvent/pseudoElement}} will still contain the pseudo-element's name.
 	</dl>
 
 	<dfn dfn-type=constructor for=AnimationEvent>AnimationEvent(type, animationEventInitDict)</dfn> is an <a>event constructor</a>.
--- a/css-transitions-1/Overview.bs
+++ b/css-transitions-1/Overview.bs
@@ -1343,6 +1343,13 @@ associated with transitions.
     The <a>un-initialized value</a> of this attribute MUST be
     <code>null</code>.
 
+    Note: {{TransitionEvent/pseudoTarget}} and {{TransitionEvent/pseudoElement}} are intentionally
+    not equivalent. Because {{TransitionEvent/pseudoTarget}} returns a {{CSSPseudoElement}}
+    object, it is subject to retargeting and encapsulation. Therefore, if the event's target
+    is a pseudo-element inside a shadow tree, {{TransitionEvent/pseudoTarget}} will
+    return <code>null</code> when observed from outside the shadow tree, while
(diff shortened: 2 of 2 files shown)


danielsakhapov:
The spec text is mostly shaped and influenced by https://github.com/w3c/uievents/pull/413

danielsakhapov:
the test is: https://github.com/web-platform-tests/wpt/blob/master/css/css-pseudo/events-on-pseudo-element.tentative.html

css-animations-1/Overview.bs
```diff
@@ -905,10 +908,43 @@ Attributes</h4>
 			runs (in which case the target of the event is that pseudo-element's corresponding
 			element), or the empty string if the animation runs on an element (which means the
 			target of the event is that element).
+		<dt><dfn>pseudoTarget</dfn>
```
dbaron, line 911: It seems like the definitions of `pseudoTarget` and `pseudoElement` should probably be equivalent, in that I would expect that `pseudoElement` and `pseudoTarget.type` should be the same.

However, the definitions seems sufficiently different right now that I wonder if that's actually the case, for example, if the pseudo-element is in a shadow tree and the retargeting algorithm affects things.

Are they intended to be equivalent as I would expect, or are they intended to be different?  (I think it should be clear either way.)

css-animations-1/Overview.bs
```diff
@@ -905,10 +908,43 @@ Attributes</h4>
...
+
+			To get this attribute, run these steps:
+
+			1. If the [=AnimationEvent/pseudoTargetOrigin=] internal slot is <code>null</code>, then return <code>null</code>.
+			1. Let |currentTarget| be {{Event/currentTarget}}.
+			1. If |currentTarget| is <code>null</code>, then return <code>null</code>.
+			1. Let |origin| be the [=AnimationEvent/pseudoTargetOrigin=] internal slot.
+			1. Let |retargeted| be the result of [=retarget=]ing |origin| against |currentTarget|.
+			1. If |retargeted| is not |origin|, then return <code>null</code>.
+			1. Return {{AnimationEvent/pseudoTarget}}.
```
dbaron: I'm not sure if this is formally correct.  This seems to be treating `pseudoTarget` as an internal slot in addition to being a property, but it doesn't really describe it as an internal slot.  (Someone more familiar with these concepts should probably review.)

smaug----: I guess this is similar to [CommandEvent.source](https://html.spec.whatwg.org/#the-commandevent-interface:~:text=was%20initialized%20to%2E-,The%20source%20getter%20steps%20are%20to%20return%20the%20result%20of%20retargeting%20source%20against%20this%27s%20currentTarget%2E).
See the red box there

css-animations-1/Overview.bs
```diff
@@ -905,10 +908,52 @@ Attributes</h4>
...
+			1. Let |retargeted| be the result of [=retarget=]ing |origin| against |currentTarget|.
+			1. If |retargeted| is not |origin|, then return <code>null</code>.
+			1. Return the [=AnimationEvent/pseudoTarget=] internal slot.
+
+			The <a>un-initialized value</a> of this attribute MUST be
+			<code>null</code>.
+
+			Note: {{AnimationEvent/pseudoTarget}} and {{AnimationEvent/pseudoElement}} are intentionally
+			not equivalent. Because {{AnimationEvent/pseudoTarget}} returns a {{CSSPseudoElement}}
+			object, it is subject to retargeting and encapsulation. Therefore, if the event's target
```
dbaron, line 938: Not sure what the "and encapsulation" means here -- I think the underlying purpose of the retargeting could be described as encapsulation but I don't think there are two separate things happening here (unless I'm missing something).

(same below for `TransitionEvent`)

dbaron (approved):
This looks good to me, with one comment below.  Hopefully @smaug---- can review as well.

Issue #13364, opened 2026-01-16
[css-anchor-position-1][css-animations-2][scroll-animations-1] Fix name visibility for descendants


flackr:
In https://github.com/w3c/csswg-drafts/issues/12581#issuecomment-3206707173 we resolved to use the same name scoping mechanism for the various specs that have scoped names: [timeline-scope](https://www.w3.org/TR/scroll-animations-1/#timeline-scope), [anchor-scope](https://www.w3.org/TR/css-anchor-position-1/#anchor-scope), and [trigger-scope](https://drafts.csswg.org/css-animations-2/#trigger-scope). I still think this is a good thing to do so that developers don't have to navigate nuanced differences between name scoping.

Proposal:
Make position-anchor, animation-timeline, animation-trigger refer to the nearest ancestor declaring that name if found before reaching the `*-scope` element.

I think this is one aspect of timeline scopes that we should adopt for anchor scopes and trigger scopes. It is confusing for developers to have to add an explicit scope when referring to an ancestor. E.g.

```html
<style>
.item {
  view-timeline-name: --item;
}
.animation-target {
  animation-name: parallax;
  animation-timeline: --item;
}
</style>
<div class=item>
  <div class=animation-target></div>
</div>
<div class=item>
  <div class=animation-target></div>
</div>
<div class=item>
  <div class=animation-target></div>
</div>
```

This will significantly de-risk the potential compat issues of updating `timeline-scope` as well as make `anchor-scope` and `trigger-scope` easier to use. E.g. here's a demo where the anchor intends to use the nearest position-anchor but right now requires developers explicitly add anchor-scope: https://codepen.io/flackr/pen/azZJvyL

@andruud FYI

andruud:
> Proposal:
> Make all of the `-scope` properties refer to the nearest ancestor declaring that name if found before reaching the scoping element.

The `-scope` properties don't refer to ancestors. But this makes sense if we mean the properties governed by the `-scope` properties (`animation-timeline`, etc). 

Within the "names are global by default" world, I think this amounts to each named definition also quietly implying a named scope? In other words:

 - `scroll/view-timeline-name:--foo` implies `timeline-scope:--foo`.
 - `anchor-name:--foo` implies `anchor-scope:--foo`.
 - `timeline/event-trigger-name:--foo` implies `trigger-scope:--foo`.

andruud:
> I think this amounts to each named definition also quietly implying a named scope?

[...]

PR #13611, opened 2026-03-06
[scroll-animations-1] Make timeline name scoping last-in-tree-order

andruud:
See #12581, #13364.


andruud:
Hmm, actually, I'm not sure this works correctly.

```
<div id=t1 style="scroll-timeline: --t; animation-timeline:--t">
  <div id=t2 style="scroll-timeline: --t">
  </div>
</div>
```

This PR would cause `animation-timeline:--t` to find `--t` on `#t2`. I'll try again next week.

Issue #12888, opened 2025-10-01
[css-fonts-5] Text Fitting: Scaling of things based on font-size


kizu:
My main point of disagreement with the Chrome team's [feedback](https://github.com/w3c/csswg-drafts/issues/2528#issuecomment-3336857019) is the proposed method of scaling the font.

A few quotes from the feedback where they argue about not scaling any dependent properties or elements based on the change of the font-size, and proposing that it should potentially be implementation-dependent and up to browsers.

<details><summary>Several quotes from the feedback</summary>

> - Other `<length>` values should not be scaled even if they are font-relative.  
>     e.g. padding of inline elements
> - Non-text parts such as inline images and inline-blocks should not be scaled.

> The fit-width text feature needs to work after line breaking because it needs original widths of lines. Adding another font shaping step in the fit-width text processing would be possible, but it's hard to add another style computation in it or to restart from style computation.

> - **`Font-size` and `scale` methods**: If we drop the `<fit-method>` option entirely, the scaling method should be implementation-dependent, and probably it would be either `font-size` or `scale`.  
    These methods produce very similar results, and their differences are likely not significant to most users. In the prototype, the `font-size` method didn't scale `letter-spacing` and `word-spacing`, and the `scale` method scaled them. In order to minimize differences between two scaling methods, `latter-spacing` and `word-spacing` should always be scaled because it's difficult to implement non-scaled `letter-spacing` and `word-spacing` with the scale method.

> - **Existing properties with `<length>`**: Scaling font-relative `<length>` values is not practical due to implementation complexity. In Chromium, we only have a computed `<length>` value at the "layout" stage, as a numeric value without unit information. Changing this to retain unit information would increase memory consumption even on pages not using this feature.  
    The `line-height` property is an exception, as we retain its unit information and can resolve font-relative `<length>` at the "layout" stage.

And probably semi-related

> - **Performance Considerations**: Concerns were raised about the performance of the `font-size` method, as repeatedly reshaping fonts to find the best fit is impractical. The prototype addressed this with a combination of one-shot reshaping and a scaling transformation.

</details>

---

If I were to sum things up as I understand it: current Chromium implementation converts font-relative values at the Style computation step, so we lose that information when we need to apply the fitting algorithm after line breaking. We can still do so for `line-height` as it retains the necessary information.

[...]

Issue #12887, opened 2025-10-01
[css-fonts-5] Text Fitting: Shrinking and Growing


kizu:
In the [original explainer](https://github.com/explainers-by-googlers/css-fit-text/blob/main/README.md) and in [their prototype](https://docs.google.com/document/d/1qFtLRuZcQN8p4UmMO1n6nLw71MW-93C7CPwS1M6jas4/edit?usp=sharing), Chrome's team proposed to introduce two properties that will handle the “fit-to-width” behavior: `text-grow` and `text-shrink`.

I am strongly opposed to having two properties instead of one, and I wrote about it [here](https://blog.kizu.dev/fit-to-width-discussions-and-feedback/#problems-of-text-shrink). To sum up:

- Having both properties is redundant, as anything you can do with `text-shrink` you could do with `text-grow`.
- If we want to fix the accessibility issue and introduce the default limit (https://github.com/w3c/csswg-drafts/issues/12886), having two properties with their separate limits won't work, as either the limit will need to be too small, or the possible range will become too wide, not solving anything.

In [their feedback](https://github.com/w3c/csswg-drafts/issues/2528#issuecomment-3336857019), Chrome's team states that

> - **Shrinking + consistent (per-block) behavior is essential**: Some web developers emphasized the need for this combination to address long words overflowing their containers.

Thinking about it further, I agree that some use cases want you to approach things by expressing the default font-size, and then allowing it to shrink. While it is possible to implement it with `text-grow`, the resulting code and thought process are pretty awkward.

**My alternative proposal**: what if we had just one property like (name to be bikeshedded) `text-fit`, with `grow|shrink` being the _required_ keyword to use to enable it with the corresponding method — `grow` — do the thing when there is free space, `shrink` — do the thing when content overflows.

Then, the minimal syntax could be just `text-fit: grow` or `text-fit: shrink`, without a need to provide the target, the method, or the limit, relying on their default values.

The original Chrome team's proposal had the `<fit-target> = consistent | per-line | per-line-all` as required. I would propose to use `consistent` as default for `grow` and `per-line-all` for `shrink`, as these seem to be the most useful for each corresponding case.

Reasons for `consistent` being a default for `grow` (I think it is pretty obvious that `per-line-all` is better for `shrink`):
- It works better for more cases, which is demonstrated by the need to have both `per-line`  and `per-line-all`.
- It will be more performant, as we need to only calculate everything for the longest line.
[...]

Issue #12886, opened 2025-10-01
[css-fonts-5] Text Fitting: Default scaling limit


kizu:
In the [original issue](https://github.com/w3c/csswg-drafts/issues/2528) we received [valuable feedback](https://github.com/w3c/csswg-drafts/issues/2528#issuecomment-2769947094) that a native fit-to-width solution can potentially introduce a simple way for authors to violate [Success Criterion 1.4.4 Resize Text](https://www.w3.org/TR/WCAG21/#resize-text), in the same way many existing fit-to-text and other responsive typography approaches already do.

My [proposed solution](https://blog.kizu.dev/fit-to-width-discussions-and-feedback/#limiting-the-size-range) involves introducing a default limit equal to 200% of the original font-size. This, with all major browsers being able to provide at least 400% full-page zoom, will allow to scale any text that uses fit-to-width without overriding this limit.

This limit is just a default, allowing authors to control it, and increase/decrease based on their design goals: there are cases where the absence of the limit won't cause the SC 1.4.4 violations, and we shouldn't limit authors there. And, realistically, we can't — but we should do what we can and introduce basic safeguards to the spec, so it won't be _easy_ to violate WCAG with the new property.

In Google Chrome team's [feedback](https://github.com/w3c/csswg-drafts/issues/2528#issuecomment-3336857019) they mention this about this proposal:

> - **Accessibility**: If the width of a container with `text-grow` or `text-shrink` depends on the viewport width, changing the page zoom level might not alter the physical text size in that container because its physical width remains unchanged.  
    We currently do not have a solution for this issue, and are not sure if the "200% size limit" solution discussed in this issue will work well.

I would be interested to know reasons why Chrome's team thinks why this limit won't work well.

But without any alternative proposals to safeguard against violating SC 1.4.4, I believe the built-in default limit is the way to go.

So far, I did not read any other proposals that would cover this issue, but if there are any — please, post them in this issue.

tkent-google:
First, I'm wondering what exactly is meant by the "original font size". If a user is constantly browsing any pages at a 200% page zoom level, does the "&lt;CSS font size> * 200%" constitute the "original font size"?

If that is the case, changing the page zoom level while viewing a page would cause the "original font size" to be scaled as well, which means the 200% limit idea would not resolves the accessibility issue.

If it's defined differently, and the "original font size" is the size before the page zoom level is applied, then this particular user would not see any effect from the fit-width text feature.

kizu:
[...]

Issue #12885, opened 2025-10-01
[css-fonts-5] or [css-text-4] Text Fitting: Which spec to put the feature into?


kizu:
In the original issue it was [resolved](https://github.com/w3c/csswg-drafts/issues/2528#issuecomment-2769621512) to be placed in Fonts 5, however Chrome team’s [feedback](https://github.com/w3c/csswg-drafts/issues/2528#issuecomment-3336857019) argues:

> - We think the feature belongs to css-text, rather than css-fonts.  
    It's similar to the `text-align` property.

We should resolve on if we want it to stay in Fonts 5, or move to Text 4.

I don't have an opinion on this. I'll open the remaining issues with the `[css-fonts-5]` in their titles and labels as it was resolved so far, but we could adjust things as we go.

drott:
Seconding tkent@'s opinion in that I find this a better fit in CSS Text. In my understanding, the algorithms for scaling and fitting are based on shaping results, but the logic for tuning spacing and font-sizing is in layout and does otherwise not affect font properties or font rendering. 

CC @foolip

svgeesus:
I weakly agree that `css-text-4` might be a better fit. Happy to whichever way the group decides.

tkent-google:
Resolved by https://github.com/w3c/csswg-drafts/pull/13616

PR #13616, opened 2026-03-09
[css-text-4][css-inline-3][css-text-decor-4] Introduce `text-fit` property

tkent-google:
[css-text-4][css-inline-3][css-text-decor-4] Introduce `text-fit` property

[css-text-4]
* Add a section to define `text-fit` property.
* `word-spacing`, `letter-spacing`, and `text-autospace` properties apply "used" font-size.

[css-inline-3]
`line-height` property applies "used" font-size.

[css-text-decor-4]
`text-decoration-thickness` and `text-underline-offset` properties apply "used" font-size.

https://github.com/w3c/csswg-drafts/issues/2528
https://github.com/w3c/csswg-drafts/issues/12885
https://github.com/w3c/csswg-drafts/issues/12886
https://github.com/w3c/csswg-drafts/issues/12887
https://github.com/w3c/csswg-drafts/issues/12888

### DIFF
--- a/css-inline-3/Overview.bs
+++ b/css-inline-3/Overview.bs
@@ -1019,8 +1019,9 @@ Line Spacing: the 'line-height' property</h3>
 		<dt><dfn><<percentage [0,∞]>></dfn>
 		<dd>
 			The [=preferred line height=]
-			and [=computed value=] of the property
 			is this percentage of the element's used 'font-size'.
+			The [=computed value=] of the property
+			is this percentage of the element's computed 'font-size'.
 			Negative values are illegal.
 	</dl>
 
--- a/css-inline-3/Overview.bs
+++ b/css-inline-3/Overview.bs
@@ -981,7 +981,7 @@ Line Spacing: the 'line-height' property</h3>
 	Initial: normal
 	Applies to: non-replaced inline boxes and SVG <a>text content elements</a>
 	Inherited: yes
-	Percentages: computed relative to ''1em''
+	Percentages: computed relative to the used 'font-size'
 	Computed value: the specified keyword, a number, or a computed <<length>> value
 	Animation type: by computed value type
 	</pre>
@@ -1002,7 +1002,7 @@ Line Spacing: the 'line-height' property</h3>
 		<dt><dfn>normal</dfn>
 		<dd>
 			Determine the [=preferred line height=]
-			automatically based on font metrics.
+			automatically based on the metrics of the used font.
 
 		<dt><dfn><<length [0,∞]>></dfn>
 		<dd>
@@ -1012,7 +1012,7 @@ Line Spacing: the 'line-height' property</h3>
 		<dt><dfn><<number [0,∞]>></dfn>
 		<dd>
 			The [=preferred line height=] is this number
-			multiplied by the element's computed 'font-size'.
+			multiplied by the element's used 'font-size'.
 			Negative values are illegal.
 			The [=computed value=] is the same as the [=specified value=].
 
@@ -1020,7 +1020,7 @@ Line Spacing: the 'line-height' property</h3>
 		<dd>
 			The [=preferred line height=]
 			and [=computed value=] of the property
-			is this percentage of the element's computed 'font-size'.
+			is this percentage of the element's used 'font-size'.
 			Negative values are illegal.
 	</dl>
 
--- a/css-inline-3/Overview.bs
+++ b/css-inline-3/Overview.bs
@@ -981,7 +981,7 @@ Line Spacing: the 'line-height' property</h3>
 	Initial: normal
 	Applies to: non-replaced inline boxes and SVG <a>text content elements</a>
 	Inherited: yes
-	Percentages: computed relative to the used 'font-size'
+	Percentages: see below
 	Computed value: the specified keyword, a number, or a computed <<length>> value
 	Animation type: by computed value type
 	</pre>
--- a/css-text-4/Overview.bs
+++ b/css-text-4/Overview.bs
@@ -9737,6 +9737,173 @@ Aligning a block of text within its container: the 'text-group-align' property</
 			</wpt>
 	</dl>
 
+<h3 id="text-fit-property">
+Text fitting: the 'text-fit' property</h3>
+
+	<wpt title="
+		This section has no coverage yet.
+	"></wpt>
+
+	<pre class="propdef">
+	Name: text-fit
+	Value: [ none | grow | shrink ] [consistent | per-line | per-line-all]? <<percentage>>?
+	Initial: none
+	Applies to: <a>block containers</a>
+	Inherited: no
+	Percentages: N/A
+	Computed value: specified keywords or computed <<percentage>> value
+	Canonical order: per grammar
+	Animation type: discrete
+	</pre>
+
+	<p>The 'text-fit' property describes how [=inline-level=] contents are
+	scaled if they do not fit to the [=line box=].
+
+	<p>Values have the following meanings:
+
+	<dl dfn-for=text-fit dfn-type=value>
+		<dt><dfn>none</dfn>
+		<dd>
+		    The user agent doesn't scale [=inline-level=] contents to fit to
+			the [=line box=].
+		</dd>
+
+		<dt><dfn>grow</dfn>
+		<dd>
+		    The user agent scales up [=inline-level=] contents to fit to the
+			[=line box=].
+		</dd>
+
+		<dt><dfn>shrink</dfn>
+		<dd>
+		    The user agent scales down [=inline-level=] contents to fit to the
+			[=line box=].
+		</dd>
+
+		<dt><dfn>consistent</dfn>
+		<dd>
+		    Specifies that all lines in the container are scaled with a single
+			scaling factor.
+			This keyword has no effect if ''text-fit/none'' is specified.
+			If none of ''text-fit/consistent'', ''text-fit/per-line'', or
+			''text-fit/per-line-all'' are specified, ''text-fit/consistent'' is
+			assumed.
+		</dd>
+
+		<dt><dfn>per-line</dfn>
+		<dd>
+		    Specifies that each line is scaled with its own scaling factor.
+			However, the last line of the block and lines that end in a forced
+			break are not scaled.
+			This keyword has no effect if ''text-fit/none'' is specified.
+		</dd>
+
+		<dt><dfn>per-line-all</dfn>
+		<dd>
+		    Specifies that each line is scaled with its own scaling factor,
+			including the last line of the block and lines that end in a forced
+			break.
+			This keyword has no effect if ''text-fit/none'' is specified.
+		</dd>
+
+		<dt><dfn>&lt;percentage&gt;</dfn>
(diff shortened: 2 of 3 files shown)


css-inline-3/Overview.bs
```diff
@@ -983,15 +1012,15 @@ Line Spacing: the 'line-height' property</h3>
...
+			multiplied by the element's used 'font-size'.
 			Negative values are illegal.
 			The [=computed value=] is the same as the [=specified value=].
 
 		<dt><dfn><<percentage [0,∞]>></dfn>
 		<dd>
 			The [=preferred line height=]
 			and [=computed value=] of the property
-			is this percentage of the element's computed 'font-size'.
+			is this percentage of the element's used 'font-size'.
```
kojiishi, line 1022: Are we sure we want to make `getComputedStyle()` depends on layout for this property? I see some discussions in 12888 but I didn't find resolutions. Did I miss?

If we want this, I guess the resolution timing needs to be clarified, but I'm not familiar with how we handle layout-dependant computed values. Maybe @tabatkins knows this better.

For the `<number>`, the spec says the computed value is the specified value, but [this test](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cdiv%20id%3D%22e1%22%20style%3D%22line-height%3A%201.2%22%3E%3C%2Fdiv%3E%0A%3Cdiv%20id%3D%22e2%22%20style%3D%22line-height%3A%20120%25%22%3E%3C%2Fdiv%3E%0A%3Cscript%3E%0Afor%20(const%20e%20of%20%5Be1%2C%20e2%5D)%20%7B%0A%20%20e.textContent%20%3D%20%60%24%7Be.style.lineHeight%7D%20%3D%20%24%7BgetComputedStyle(e).lineHeight%7D%60%3B%0A%7D%0A%3C%2Fscript%3E%0A) indicates that all browsers return the computed `<length>` value, so I guess the same applies there too.

css-text-4/Overview.bs
```diff
@@ -9737,6 +9737,173 @@ Aligning a block of text within its container: the 'text-group-align' property</
...
+	this by adjusting spacing between characters, this property scales the
+	font size.
+
+	<p>When ''text-fit/grow'' or ''text-fit/shrink'' is specified,
+	a [=line scaling factor=] is computed for each [=line box=].
+	If ''text-fit/consistent'' applies, all [=line boxes=] are scaled by the
+	smallest computed [=line scaling factor=].
+	Otherwise, each [=line box=] is scaled by its own [=line scaling factor=].
+
+	<p>The 'font-size' used value of text is its 'font-size' [=computed value=]
```
kojiishi: Should this be "The used value of 'font-size' is..."? Or maybe just my English skill failed to read this, sorry if so and please disregard this comment.

css-text-4/Overview.bs
```diff
@@ -9737,6 +9737,173 @@ Aligning a block of text within its container: the 'text-group-align' property</
...
+	multiplied by the [=line scaling factor=].
+	This property does not affect the 'font-size' [=computed value=],
+	and thus does not affect font-size-relative <<length>> values.
+	For example, "line-height: 1.5em" and "letter-spacing: 0.1em" are not
+	affected by this scaling.
```
kojiishi, line 9836: nit: How about making the computed value part a separate paragraph? And maybe put before the used value, as the sentence refers to the computed value. I wondered what would happen to computed value when reading it, then found this later.

css-text-decor-4/Overview.bs
```diff
@@ -359,14 +359,14 @@ Text Decoration Line Thickness: the 'text-decoration-thickness' property</h3>
...
-		<dt><dfn><<line-width>></dfn>
 		<dd>
 			A length value specifies the thickness of text decoration lines as a fixed length.
 			
 			Note: A length will inherit as a fixed value,
 			and will not scale with the font.
 
-			A percentage value specifies the thickness of text decoration lines as a percentage of ''1em''.
+			A percentage value specifies the thickness of text decoration lines as a percentage of
+			the used 'font-size'.
```
kojiishi, line 369: Does this affect the computed value?

kojiishi: From [this quick test](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cdiv%3E%3C%2Fdiv%3E%0A%3Cdiv%20style%3D%22line-height%3A%20normal%22%3E%3C%2Fdiv%3E%0A%3Cdiv%20style%3D%22line-height%3A%201.2%22%3E%3C%2Fdiv%3E%0A%3Cdiv%20style%3D%22line-height%3A%20120%25%22%3E%3C%2Fdiv%3E%0A%3Cdiv%20style%3D%22letter-spacing%3A%20120%25%22%3E%3C%2Fdiv%3E%0A%3Cdiv%20style%3D%22text-decoration-thickness%3A%2010%25%22%3E%3C%2Fdiv%3E%0A%3Cscript%3E%0Afor%20(const%20e%20of%20document.getElementsByTagName(%27div%27))%20%7B%0A%20%20let%20lines%20%3D%20%5B%5D%3B%0A%20%20if%20(e.style.lineHeight)%20%7B%0A%20%20%20%20lines.push(%60%24%7Be.style.lineHeight%7D%20%3D%20%24%7BgetComputedStyle(e).lineHeight%7D%60)%3B%0A%20%20%7D%0A%20%20if%20(e.style.letterSpacing)%20%7B%0A%20%20%20%20lines.push(%60%24%7Be.style.letterSpacing%7D%20%3D%20%24%7BgetComputedStyle(e).letterSpacing%7D%60)%3B%0A%20%20%7D%0A%20%20if%20(e.style.textDecorationThickness)%20%7B%0A%20%20%20%20lines.push(%60%24%7Be.style.textDecorationThickness%7D%20%3D%20%24%7BgetComputedStyle(e).textDecorationThickness%7D%60)%3B%0A%20%20%7D%0A%20%20if%20(lines.length)%20%7B%0A%20%20%20%20e.textContent%20%3D%20lines.join(%27%5Cn%27)%3B%0A%20%20%7D%0A%7D%0A%3C%2Fscript%3E%0A), only `line-height` resolves percentages to length?

css-inline-3/Overview.bs
tkent-google: Oh, I made a mistake. __This feature should not affect computed value of any properties__.  This part should be:

> The preferred line height of the property is this percentage of the elmenet's __used__ 'font-size', and computed value of the property is this percentage of the element's __computed__ 'font-size'.

css-text-decor-4/Overview.bs
tkent-google: As noted in another comment, computed values should not be affected.

css-text-4/Overview.bs
tkent-google: ok, I improve the sentence.

css-text-4/Overview.bs
tkent-google: Thank you for the suggestion.  I swap the order.

css-inline-3/Overview.bs
tkent-google: Done.

css-inline-3/Overview.bs
```diff
@@ -952,7 +981,7 @@ Line Spacing: the 'line-height' property</h3>
 	Initial: normal
 	Applies to: non-replaced inline boxes and SVG <a>text content elements</a>
 	Inherited: yes
-	Percentages: computed relative to ''1em''
+	Percentages: computed relative to the used 'font-size'
```
kojiishi: This change looks to me that `getComputedStyle()` returns the "relative to the used `font-size`", as it says "computed".

Whether this field is for the computed value or for the used value is a bit ambiguous. The computed value of percentages is, as the link to the [Percentages label of bikeshed output](https://www.w3.org/TR/css-values/#percentages) (see 5.5.1) defines:
> Unless otherwise specified (...), the computed value of a percentage is the specified percentage.

I think this part is "unless otherwise specified". Especially when it starts with "computed", changing this part will affect the computed value.

When `text-fit` makes the computed value and the used value different, I think clarifying subjects helps.

How about leaving this line unchanged, or something like:
> computed relative to ''1em'', see ''text-fit'' for the used value

or, maybe you want to fix this because `text-fit` makes "1em" ambiguous? If that's the case, how about using the same description as you wrote below, e.g.,:
> computed relative to the computed 'font-size', see ''text-fit'' for the used value

Another possible idea is to replace this with "see below". I think it's better than a short summary that could possibly lead to misunderstanding.

css-text-4/Overview.bs
```diff
@@ -9779,7 +9947,7 @@ Word Spacing: the 'word-spacing' property</h3>
 	Initial: normal
 	Applies to: text
 	Inherited: yes
-	Percentages: relative to computed 'font-size', i.e. ''1em''
+	Percentages: relative to used 'font-size'
```
kojiishi, line 9950: nit: How about adding "used value is" to clarify the subject?

However, it's also ok to ignore this feedback as the issue exists in the original spec. Unlike `line-height`, its computed value is `%` ([test](https://output.jsbin.com/boxagiw)), so the original is already implicitly specifying the used value.

css-text-4/Overview.bs
```diff
@@ -9977,7 +10145,7 @@ Tracking: the 'letter-spacing' property</h3>
 	Initial: normal
 	Applies to: <a>inline boxes</a> and text
 	Inherited: yes
-	Percentages: relative to computed 'font-size', i.e. ''1em''
+	Percentages: relative to used 'font-size'
```
kojiishi, line 10148: ditto

kojiishi (approved):
LGTM w/nits

css-inline-3/Overview.bs
kojiishi: [This commit](https://github.com/w3c/csswg-drafts/commit/f6f1c74b64c91262f849e0f59c0401dbe6b28679) mentions "The fact that percentages compute to lengths", so the word "computed" in the Percentages field is probably intentional.

tkent-google: Thank you for the feedback.  I changed it to "see below" as suggested.

css-text-4/Overview.bs
tkent-google: I'd like to keep it as is.  IMO this is not confusing.

css-text-4/Overview.bs
tkent-google: Ditto.

tkent-google:
@kojiishi Thank you for the review.
Would you merge it if the latest one is ok please?

fantasai:
@tkent-google This was supposed to go into CSS Text Level 5.
https://github.com/w3c/csswg-drafts/issues/12885#issuecomment-3524545760

tkent-google:
> @tkent-google This was supposed to go into CSS Text Level 5. [#12885 (comment)](https://github.com/w3c/csswg-drafts/issues/12885#issuecomment-3524545760)

Oh, my bad.
It seems it was already moved to css-text-5.  Thanks!

Issue #12122, opened 2025-04-25
[css-overflow-5] Discrete versus navigation scroll marker mode


flackr:
## Overview
When using scroll markers, there are two known main use cases:
1. A table of contents or set of quick navigation points.
2. A set of independently exclusively accessible content (e.g. a carousel, tabs).
3. If there are other use cases, please comment on this issue so we can consider whether they fit into these two modes or not.

Currently, the focus management and behavior of scroll markers is largely designed around the first use case, and requires the developer to use `interactivity` to inert the non-current content should they need the second use case.

However, it would be nice if the developer could opt into this behavior and the browser could use this opt-in to automatically apply appropriate inferred roles to the content.

## Proposal
Authors creating carousels or tab designs would apply `scroll-marker-type: discrete`, making the scroll marker pseudo-elements a [sequential focus navigation scope owner](https://html.spec.whatwg.org/multipage/interaction.html#focus-navigation-scope-owner) for their originating elements. This means that the originating element (and its descendents) follow the scroll marker in focus order. Since only the :target-current scroll marker is focusable, the other scroll markers and their content are not focusable until a different scroll marker is activated.

When this type is applied, the originating element will be given the "tabpanel" role, and the ::scroll-marker-group and ::scroll-marker pseudo-elements will be given the tablist and tab role respectively.

Otherwise, (e.g. for table of contents use cases) the ::scroll-marker-group and ::scroll-marker pseudo-elements will be given the listbox and option respectively.

## Open questions
* Should this be part of the scroll-marker-group property? E.g. `scroll-marker-group: before discrete;`? It probably doesn't make sense (and we couldn't implicitly apply the tablist role) if this was per scroll marker - so it needs to be associated with the common container of the scroll markers and this mean that we don't have this extra property for all elements just in case they have a scroll marker group.
* Does this have an effect on element scroll markers? Presumably applied to the same element that has `scroll-marker-group: contain` per #10916.
* Should this change keyboard focus behavior? E.g. without specifying this property we would have a list-of-links style focus where every scroll marker is in tab order. This would be a breaking change but the impact is likely minimal for now.
* How would this work when scroll markers are established on ::column pseudos? Or would we skip the sequential focus navigation scope order?
  * In particular, some elements will span across multiple ::column pseudo-elements. Could we make the marker a sequential focus navigation scope order only for the fully contained elements or only elements whose first fragment is in that particular ::column?
[...]

PR #13627, opened 2026-03-11
[css-overflow-5] Update scroll-marker-group to include modes #12122

danielsakhapov:


css-overflow-5/Overview.bs
```diff
@@ -370,10 +378,27 @@ The 'scroll-marker-group' property</h4>
...
+
+<h4 id="scroll-marker-modes">Scroll Marker Modes</h4>
+
+	A ''::scroll-marker-group'' operates in either <dfn>links</dfn> or <dfn>tabs</dfn> mode based on the 'scroll-marker-group' property value of its [=originating element=]. The chosen mode significantly affects the roles, focus behaviors, and interactions of the associated elements and pseudo-elements.
+
+	If the mode is [=links=]:
+
+	* The roles of ''::scroll-marker-group'' and ''::scroll-marker'' are `navigation` and `link` respectively.
+	* All ''::scroll-marker'' pseudo-elements are tab stops, in order.
+	* Activating a ''::scroll-marker'' pseudo-element <a lt='scroll a target into view'>scrolls its originating element into view</a> and sets the <a href="https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation-starting-point">sequential focus navigation starting point</a> to the [=originating element=], losing focus from the ''::scroll-marker''.
```
flackr: The activation behavior is also specified in the activation section. I think in practice we try to have a single normative specification and if there are additional locations they are non-normative (e.g. notes).

TLDR; can we either:
* remove the activation behavior here and add references to the sections explaining the behavioral changes based on the mode from this section, or
* in the activation behavior section, instead of having a switch, call to this section to invoke mode specific activation behavior

Issue #12258, opened 2025-06-02
[cssom] Always return escaped idents?


andruud:
When some some CSS construct holds a name of some kind, and a CSSOM API makes it possible to read that name back to JS, should we return a serialized ident, or the actual internal value?

Using the name `a\9 b` (U+0009 => TAB) I checked [a number of at-rules](https://jsbin.com/rolumomodu/edit?html,output) that expose their name in some way, and the result varies from rule to rule and from browser to browser. It doesn't appear to be consistently specified in one way either.

Chrome:

```
@layer (statement): a\9 b
@layer (block): a\9 b
@container: a\9 b
@property: --a	b
@position-try: --a	b
@keyframes: a	b
@font-palette-values: --a	b
@font-feature-values: "a\9 b"
```

Firefox:

```
@layer (statement): a\9 b
@layer (block): a\9 b
@container: a\9 b
@property: --a	b
@position-try: (N/A)
@keyframes: a	b
@font-palette-values: --a\9 b
@font-feature-values: a\9 b
```

Safari:

```
@layer (statement): a\9 b
@layer (block): a\9 b
@container: a\9 b
@property: --a	b
@position-try: (N/A)
@keyframes: a	b
@font-palette-values: --a	b
@font-feature-values: a	b
```

We should establish a guideline here, so that we don't continue to just arbitrarily pick something when adding new rules. I'm not sure what's best.

cc @emilio , @cdoublev

Loirooriol:
Including property names in declarations: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/13829
All 3 browsers agree it's the actual value: `--a	b`
I prefer to never escape, just like in HTML if I use

```html
<div id="target">a &lt; b</div>
<script>alert(target.firstChild.nodeValue)</script>
```

I expect the string `"a < b"`, not `"a &lt; b"`.

emilio:
I think where things starts to get blurry is for things that return CSS syntax like `font-face`'s font-family descriptor and so on...

Loirooriol:
Yes, my "never" was misleadingly strong. Property values and descriptor values are given to the CSS parser so they should be escaped. Continuing with the HTML analogy, it would be like `innerHTML` that does HTML parsing/serialization vs specific pieces of data that tend to expose the raw value.

tabatkins:
Yeah, we're unfortunately inconsistent about this. :/

[...]

PR #13635, opened 2026-03-12
[css-conditional-5] Extend CSSContainerRule for multiple conditions.

### LINKED ISSUE #10845: [css-conditional-5] `CSSContainerRule.containerName`/`containerQuery` should return a list
(Follow-up on this [comment](https://github.com/w3c/csswg-drafts/issues/7551#issuecomment-2014543789).)

According to [this resolution](https://github.com/w3c/csswg-drafts/issues/7551#issuecomment-1854385652), the [`@container`](https://drafts.csswg.org/css-conditional-5/#at-ruledef-container) prelude grammar accepts a list of `<container-condition>`s:

  > `@container <container-condition># { <block-contents> }`
  >
  > `<container-condition> = [ <container-name> ]? <container-query>`

But [`CSSContainerRule.containerQuery`](https://drafts.csswg.org/css-conditional-5/#csscontainerrule) and `CSSContainerRule.containerName` still return a single container query/name.

Can you please tell me if there is any reason to withhold the appropriate changes on the CSSOM? Do you plan to rename the attributes? To remove them? To return a list of strings or a string of comma-separated values?

The current version of Chrome/FF (at least) only accept a single `<container-condition>`.

emilio:
Open questions:

 * Do we want to do this "return string or array thing"? Or should we just return the empty string on `containerName` / `containerQuery` when there are multiple conditions? Maybe that's simpler.

 * If we do the "string or array thing", do we still want to expose the "always array" version (`containerNames` / `containerQueries`)? I _think_ so, but I guess I could go either way.

Fixes #10845

### DIFF
--- a/css-conditional-5/Overview.bs
+++ b/css-conditional-5/Overview.bs
@@ -1464,7 +1464,7 @@ Style Container Features</h3>
 	The <dfn><<style-feature-name>></dfn> can be either a [=supported CSS property=]
 	or a valid <<custom-property-name>>.
 	The <dfn><<style-feature-value>></dfn> production matches any valid <<declaration-value>>
-	as long as it doesn't contain <<mf-lt>>, <<mf-gt>> and <<mf-eq>> tokens. 
+	as long as it doesn't contain <<mf-lt>>, <<mf-gt>> and <<mf-eq>> tokens.
 
 	<wpt>
 		container-queries/style-query-registered-custom-rem-change.html
@@ -1970,12 +1970,22 @@ The <code>CSSContainerRule</code> interface</h3>
 	The {{CSSContainerRule}} interface represents an ''@container'' rule.
 
 	<pre class='idl'>
+	dictionary CSSContainerCondition {
+	  required CSSOMString name;
+	  required CSSOMString query;
+	};
+
 	[Exposed=Window]
 	interface CSSContainerRule : CSSConditionRule {
 		readonly attribute CSSOMString containerName;
 		readonly attribute CSSOMString containerQuery;
+		readonly attribute FrozenArray&lt;CSSContainerCondition> conditions;
 	};
 	</pre>
+
+	Issue: We should try to remove {{CSSContainerRule/containerName}} and
+	{{CSSContainerRule/containerQuery}}, since they don't deal with multiple
+	conditions correctly.
 
 	<wpt>
 		container-queries/container-rule-cssom.html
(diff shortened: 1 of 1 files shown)


emilio:
@tabatkins any strong opinions on the "String or array" thing? I think that probably just returning the empty string when there are multiple conditions is better, because otherwise things like `if (rule.containerName.startsWith("my-thing"))` will start crashing if you put multiple conditions.

mirisuzanne:
I don't have any strong feelings on the shape of this, I'd be fine with either approach.

emilio:
https://phabricator.services.mozilla.com/D287344 has some implementation and tests of this proposal.

tabatkins:
I do weakly prefer returning the empty string when it's multiple conditions, yeah.

emilio:
@mirisuzanne could I get your stamp as editor for the spec text?

css-conditional-5/Overview.bs
```diff
@@ -1956,39 +1966,65 @@ The <code>CSSContainerRule</code> interface</h3>
...
+			<div algorithm="CSSContainerRule.conditionText">
+				1. Let |first| be <code>true</code>.
+				1. Let |result| be the empty string.
+				1. [=list/For each=] |condition| in |conditions|:
+					1. If |first| is <code>false</code>, append <code>", "</code> to |result|.
+					1. Set |first| to <code>false</code>.
+					1. If |condition|'s {{CSSContainerCondition/name}} is not empty,
+							append |condition|'s {{CSSContainerCondition/name}} followed by a
+							single space to |result|.
+					1. Append |condition|'s {{CSSContainerCondition/query}} to |result|.
```
AtkinsSJ: The condition isn't guaranteed to have a query. The grammar requires that it has at least a name or a query, so it could just have a name. In that case it would serialize as "name " with a trailing space. (The old algorithm also has this issue.)

css-conditional-5/Overview.bs
```diff
@@ -1956,39 +1966,65 @@ The <code>CSSContainerRule</code> interface</h3>
...
-			<dl class="switch">
-				<dt data-md>The ''@container'' rule has an associated <<container-name>>
-				<dd data-md>The result of serializing that <<container-name>>.
-				<dt data-md>Otherwise
-				<dd data-md>An empty string.
-			</dl>
+			<div algorithm="CSSContainerRule.containerName">
+				1. Let |conditions| be the result of getting the {{CSSContainerRule/conditions}} attribute.
+				1. If the length of |conditions| is 1:
+					1. Return the only |condition|'s {{CSSContainerCondition/name}}.
```
AtkinsSJ, line 2019: This needs to go through ["serialize an identifier"](https://drafts.csswg.org/cssom-1/#serialize-an-identifier) I believe.

css-conditional-5/Overview.bs
```diff
@@ -1956,39 +1966,65 @@ The <code>CSSContainerRule</code> interface</h3>
...
-				<dd data-md>The result of getting the <code>containerQuery</code> attribute.
-			</dl>
+			<div algorithm="CSSContainerRule.conditionText">
+				1. Let |first| be <code>true</code>.
+				1. Let |result| be the empty string.
+				1. [=list/For each=] |condition| in |conditions|:
+					1. If |first| is <code>false</code>, append <code>", "</code> to |result|.
+					1. Set |first| to <code>false</code>.
+					1. If |condition|'s {{CSSContainerCondition/name}} is not empty,
+							append |condition|'s {{CSSContainerCondition/name}} followed by a
```
AtkinsSJ: This needs "serialize an identifier" too.

css-conditional-5/Overview.bs
```diff
@@ -1956,39 +1966,65 @@ The <code>CSSContainerRule</code> interface</h3>
...
-				<dt data-md>The ''@container'' rule has an associated <<container-name>>
-				<dd data-md>The result of getting the <code>containerName</code> and
-					<code>containerQuery</code> attributes, joined by a single whitespace.
-				<dt data-md>Otherwise
-				<dd data-md>The result of getting the <code>containerQuery</code> attribute.
-			</dl>
+			<div algorithm="CSSContainerRule.conditionText">
+				1. Let |first| be <code>true</code>.
+				1. Let |result| be the empty string.
+				1. [=list/For each=] |condition| in |conditions|:
```
AtkinsSJ, line 2003: To be consistent with the other algorithms here, this should have a "1. Let |conditions| be the result of getting the {{CSSContainerRule/conditions}} attribute." step.

css-conditional-5/Overview.bs
emilio: Hmm, I guess we should just escape in the condition, as per #12258. Everything else falls out of that.

emilio:
@AtkinsSJ thanks for the review! I think I addressed your comments, but lmk otherwise :)

css-conditional-5/Overview.bs
```diff
@@ -1986,39 +1996,67 @@ The <code>CSSContainerRule</code> interface</h3>
...
+		<dt><code>conditions</code> of type <code>FrozenArray&lt;CSSContainerCondition?></code>
+		<dd>The <code>conditions</code> attribute, on getting,
+			must return a value as follows:
+			<div algorithm="CSSContainerRule.conditions">
+				1. Let |result| be an empty [=list=].
+				1. [=list/For each=] <<container-condition>> |condition| specified in the rule:
+					1. Let |dict| be a new {{CSSContainerCondition}} with
+						 {{CSSContainerCondition/name}} set to the
+						 <a lt="serialize an identifier">serialized</a> <<container-name>>
+						 of |condition| escaped if specified, or <code>""</code> otherwise,
```
AtkinsSJ: I don't think you need to say "escaped" here - that's already handled by *serialize an identifier*.

Otherwise this looks good to me.

Issue #13429, opened 2026-02-02
Changing `normal` to `running`


svgeesus:
> **running**: the image animation plays, without synchronization with any other instance of the same image on the page.
> 
> If the value is changed from `normal` to `running`, does it:
> - keep running from where it was, but now with no guarantee of future synchronization
> - start running from the start of the animation 

 _Originally posted by @svgeesus in [#1615](https://github.com/w3c/csswg-drafts/issues/1615#issuecomment-3774246821)_

svgeesus:
This issue was raised before the document was adopted as a CSS ED and refers to wording which has now changed. The [current wording](https://drafts.csswg.org/css-image-animation-1/#valdef-image-animation-running) is:

> **running**
>    Like [normal](https://drafts.csswg.org/css-image-animation-1/#valdef-image-animation-normal), the animation of [animated images](https://drafts.csswg.org/css-image-animation-1/#animated-image) is run normally, as determined by the image format and the host language,

>    However, animation timelines are scoped per element: among the [content image](https://drafts.csswg.org/css-image-animation-1/#content-image) and [decorative images](https://drafts.csswg.org/css-image-animation-1/#decorative-image) of a single element, any [animated images](https://drafts.csswg.org/css-image-animation-1/#animated-image) with the same [absolute URL](https://url.spec.whatwg.org/#syntax-url-absolute), the same image data, and with an [image-animation](https://drafts.csswg.org/css-image-animation-1/#propdef-image-animation) computed value of [running](https://drafts.csswg.org/css-image-animation-1/#valdef-image-animation-running) must be rendered synchronized to the same timeline as a group, distinct from the timeline of images in other elements, with the timeline starting at the time of the least recent addition to the group.

The original question still stands, however.

bramus:
Maybe the synchronization part needs a separate control? Something like a `image-animation-clock` prop to indicate whether the clock is synchronized or not.

DevSDK:
The neutral or intuitive way seems `keep running from where it was, but now with no guarantee of future synchronization`. 

On the implementer's perspective, the unsync is only happened after image is paused. So we can leave it keep synced until image has been paused.

frivoal:
[...]

PR #13638, opened 2026-03-12
[css-image-animation] Define what state 'running' starts from

frivoal:
This is in response to https://github.com/w3c/csswg-drafts/issues/13429.

### DIFF
--- a/css-image-animation-1/Overview.bs
+++ b/css-image-animation-1/Overview.bs
@@ -146,6 +146,10 @@ Controlling Image Animations: the 'image-animation' property {#image-animation}
 			If images are added to the element while the computed value is ''running'',
 			the timeline starts at the time of the least recent addition to the group.
 
+			If the element is created or made visible after having previously been set to ''display: none'',
+			with images already added and with 'image-animation' already set to ''running'',
+			the timeline starts when the element is included in the layout.
+
 			If this property is switched to ''running'' from another value,
 			the beginnig of this timeline is set
 			so that the animation continues from the state that was displayed


css-image-animation-1/Overview.bs
```diff
@@ -141,8 +141,15 @@ Controlling Image Animations: the 'image-animation' property {#image-animation}
 			the same image data,
 			and with an 'image-animation' computed value of ''running''
 			must be rendered synchronized to the same timeline as a group,
-			distinct from the timeline of images in other elements,
-			with the timeline starting at the time of the least recent addition to the group.
+			distinct from the timeline of images in other elements.
+
+			If images are added to the element while the computed value is ''running'',
+			the timeline starts at the time of the least recent addition to the group.
```
DevSDK, line 147: does `timeline starts` mean start from the first frame right? (i,.e., element added with `running`)
On the implementation side, we are just considering about [that](https://chromium-review.googlesource.com/c/chromium/src/+/7596743/comment/fae52653_f5d32ce5/)..

frivoal: Per this definition, the intent is yes. So if you set a blank element to `running`, and then some time later you add set its background (for example) to an animated image, the animation starts from the first frame at the moment you insert the image.

DevSDK: Does it affects `display: none` and back too (e.g., `running` element set display none and back) ? Since CSS Animation seems [reset the animation](https://drafts.csswg.org/css-animations/#animations) when it off the layout via `display:none`  Should we mention kind of those?

frivoal: Good point. My expectation would indeed be that if you switch back from `display: none`, the timeline restarts there. How about this:


```suggestion
			the timeline starts at the time of the least recent addition to the group.

			If the element is created or made visible after having previously been set to ''display: none'',
			with images already added and with 'image-animation' already set to ''running'',
			the timeline starts when the element is included in the layout.
```

PR #13653, opened 2026-03-14
[css-mixins-1] chore: added missing CSS values

mfranzke:
Two lines of CSS are missing values for their CSS properties on the [CSS mixins draft](https://drafts.csswg.org/css-mixins/). Missing those doesn't seem to be on purpose, not being related to that `mixin` and `macro` rule functionality, so they might only being be missing unexpectedly.


tabatkins:
Weird, I must have just gotten distracted while writing that example and left them out. thanks!

PR #13658, opened 2026-03-15
[css-forms] Use :is() in UA sheet.

emilio:
This is more compact, and reveals that `<input type=reset>` was forgotten
in a couple places.


lukewarlow:
Looks like this needs a rebase.

PR #13672, opened 2026-03-17
[css-overflow-3] Note that overflow applies to replaced elements

Psychpsyo:
The spec description of overflow and browsers agree that it has effects on replaced elements.

(Well, WebKit doesn't agree, but Chrome and Firefox do.)

Example HTML that showcases this:
```html
<!DOCTYPE html>
<img style="width: 300px; height: 300px; object-fit: cover; overflow: visible;" src="https://upload.wikimedia.org/wikipedia/commons/0/0d/Great_Wave_off_Kanagawa2.jpg">
```


fantasai (changes requested):
This capability was added in Level 4: https://drafts.csswg.org/css-overflow-4/#overflow-control

Maybe just adding a note pointing to that section from Level 3 would be appropriate.

PR #13699, opened 2026-03-21
[css2] Fix wording for magins collapsing

Psychpsyo:
This matches the notes below that section and also matches implementations.
Consider:
```html
<!DOCTYPE html>
Foo
<p><span style="display: block; margin-block: 1em"></span></p>
Bar
```
The current spec suggests that, since this paragraph has an in-flow child, its margins should not collapse, when in reality, all the margins here collapse into a single `1em` gap, because there's nothing in-between.


Psychpsyo:
I'm also wondering whether or not the spec should really say "zero computed min-height" and  "zero or auto computed height".
Since I can set either of these to `fit-content` on the paragraph in my example above and the margins still collapse.
In this case, the computed value is `fit-content`, not zero or `auto`.

PR #13701, opened 2026-03-23
[css-sizing-4] Add the `frame-sizing` property

kojiishi:
[css-sizing-4] Add the `frame-sizing` property

Based on the resolutions at:
https://github.com/w3c/csswg-drafts/issues/1771#issuecomment-3806716273

* Add the `frame-sizing` property.
* Remove the `from-element` keyword added by #13099.
* Clarified the `<meta>` tag is immutable.

### DIFF
--- a/css-sizing-4/Overview.bs
+++ b/css-sizing-4/Overview.bs
@@ -1253,8 +1253,6 @@ Responsively-sized iframes: the 'frame-sizing' property</h3>
 		but will be updated to ''500px'' once the iframe's document loads.
 	</div>
 
-	Issue: Should this apply even when ''contain:size'' isn't specified?
-
 	In addition,
 	the internal document can call {{window/requestResize()|window.requestResize()}};
 	if the document has the required <{meta}> element


kojiishi:
@tabatkins PTAL.

Since this PR contains a move and edits, I split changes to "a move commit" and "edits commit", in case this helps reviewing.

css-sizing-4/Overview.bs
```diff
@@ -1257,6 +1181,90 @@ Interaction With ''overflow: auto''</h4>
...
+
+		The height of the <{iframe}> will initially be the default iframe height (typically ''150px''),
+		but will be updated to ''500px'' once the iframe's document loads.
+	</div>
+
+	Issue: Should this apply even when ''contain:size'' isn't specified?
+
+	In addition,
+	the internal document can call {{window/requestResize()|window.requestResize()}};
+	if the document still has the required <{meta}> element,
```
tabatkins: This seems to go against the philosophy of freezing the meta value, and I guess would require us to track exactly which `meta` element was the one that set the value. Seems confusing - is there a need to keep this behavior?

css-sizing-4/Overview.bs
```diff
@@ -1257,6 +1181,90 @@ Interaction With ''overflow: auto''</h4>
...
+			iframe {
+				frame-sizing: content-height;
+			}
+		</pre>
+
+		The height of the <{iframe}> will initially be the default iframe height (typically ''150px''),
+		but will be updated to ''500px'' once the iframe's document loads.
+	</div>
+
+	Issue: Should this apply even when ''contain:size'' isn't specified?
```
tabatkins: I think this is no longer relevant, now that `frame-sizing` is what's controlling the behavior.

kojiishi: Removed, thanks!

css-sizing-4/Overview.bs
kojiishi: Good catch, thanks. Can you take a look at updated text and see if it's good for readers?

PR #13704, opened 2026-03-23
Only run publish workflows on w3c/csswg-drafts, not forks

kbabbitt:
Based on the example at: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-only-run-job-for-specific-repository

Also included is a fix to generate-auto-publish-workflows.py so that it generates Unix-style line endings when run on Windows.

### DIFF
--- a/.github/auto-publish-template.yml
+++ b/.github/auto-publish-template.yml
@@ -20,6 +20,7 @@ on:
 jobs:
   publish-TR-{{shortname}}:
     name: Publish {{shortname}}
+    if: github.repository == 'w3c/csswg-drafts'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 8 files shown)


kbabbitt:
@svgeesus could you please review? I recently sync'ed my fork of the repo and was surprised and slightly alarmed to see the publish workflows running on the fork. I don't *think* it did any damage, but this seems like a good precaution.

I know almost nothing about GitHub workflows, but from the example I found this appears to be the correct fix. Is there anything I should do to validate?

svgeesus:
Looks good to me but @deniak could you review?

tabatkins:
Unfortunately the best way to validate is typically just to activate it and see what it does :/

I agree this looks correct.

svgeesus (approved):

PR #13713, opened 2026-03-24
Add `light-dark()` and `none` to the image type, #12513

svgeesus:
Adds the image form of `light-dark()`, which is [defined in CSS Color 5](https://drafts.csswg.org/css-color-5/#light-dark), to the image type, as [resolved](https://github.com/w3c/csswg-drafts/issues/12513#issuecomment-4040713109).

Also adds the `none` value to the image type (rather than special-casing it to `light-dark()`), as [suggested by @emilio](https://github.com/w3c/csswg-drafts/issues/12513#issuecomment-4113654690) and [supported by @tabatkins](https://github.com/w3c/csswg-drafts/issues/12513#issuecomment-4092448365).

### DIFF
--- a/css-images-4/Overview.bs
+++ b/css-images-4/Overview.bs
@@ -73,11 +73,13 @@ Value Definitions {#values}
 	The <<image>> value type denotes a 2D image. It can be a
 	<a section href="#url-notation">url reference</a>,
 	<a section href="#image-notation">image notation</a>,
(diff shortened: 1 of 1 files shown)


tabatkins:
Argh, having a naked `none` be an image is gonna be a much more fraught thing that'll require discussion. If you remove that and go back to `none` being a special-case argument for `light-dark()` for now, we can actually merge the feature we resolved on.

PR #13718, opened 2026-03-25
[css-color-5] Add missing groups in <light-dark-image> syntax

cdoublev:
Follow-up on 100146f. 

https://drafts.csswg.org/css-color-5/#typedef-light-dark-image

@yisibl, `<image> | none` must be grouped because `<image> | none , <image> | none` is equivalent to `<image> | [ none , <image> ] | none`.

  > Juxtaposition is stronger than the double ampersand, the double ampersand is stronger than the double bar, and the double bar is stronger than the bar.

https://drafts.csswg.org/css-values-4/#component-combinators


yisibl:
Thanks for spotting that! I saw the `[]` inside the commented-out `<!-- <pre class='prod'>` section and assumed it was already there.

PR #13732, opened 2026-03-26
[css-gaps-1] Clarify application of writing-mode and direction to gap decorations. #13648, #13649

### LINKED ISSUE #13648: [css-gaps] More vertical-writing cases
https://www.w3.org/TR/2026/WD-css-gaps-1-20260227/

I did not find explicit prose covering `writing-mode` interactions, and the spec text does not surface vertical-writing cases. It should add more spec text and WPT coverage for `writing-mode: vertical-rl`, and `writing-mode: vertical-lr`, like for `*-inset-start`/`*-inset-end`.

### LINKED ISSUE #13649: [css-gaps] Logical geometry
https://www.w3.org/TR/2026/WD-css-gaps-1-20260227/

The spec uses `start` and `end` in several property names, which is good, but I did not find explicit prose tying row/column behavior to writing mode and base direction. A short normative statement in sections 2-4 would remove ambiguity. Row/column gaps, segment, and inset directions should be interpreted according to the relevant layout model’s logical axes under the container’s writing mode and direction.

kbabbitt:


css-gaps-1/Overview.bs
```diff
@@ -169,6 +169,17 @@ Gap decorations</h2>
 	note that while the border and background of the container aren't scrolled,
 	the decorations need to scroll along with items in the container.
 
+	[=Gap decorations=] follow the definitions of "column" and "row" laid out by the container to which the properties are applied.
+	They depend on the container type taking 'writing-mode' and 'direction' into account when determining the meanings of these terms.
```
xfq: Is 'direction' a CSS property here? Unlike writing mode, the text direction is usually determined by the HTML `dir` attribute (this may differ in TTML or other languages using CSS).

kbabbitt: Yes, there is a [direction CSS property](https://drafts.csswg.org/css-writing-modes-3/#direction). For HTML documents, the spec text for that property recommends that authors use the `dir` attribute instead. But for other host languages, we would need to look at the `direction` CSS property, or e.g. in SVG perhaps the [`direction` attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/direction).

xfq: Yeah, I think it's better not adding those quotes and just saying things like "taking writing mode and text direction into account" directly, otherwise it might be mistaken for a CSS property.

kbabbitt: Removed the links.

xfq:
Looks good to me. Thank you!

PR #13735, opened 2026-03-27
[css-conditional-5] Fix autolinks to `<supports-env-fn>`

tidoust:
The markup replaced the definition of the `<supports-env-fn>` type with a scoped `env()` function definition. This created broken links within the spec because `<<supports-env-fn>>` no longer worked as a result.

It also created another exported definition for the `env()` function (less of a problem since it was scoped to `@supports`, but it does not seem needed), and confused tools that attempt to extract grammar syntaxes from CSS specs, see:
  https://github.com/w3c/webref/issues/1886

This update gets back to a regular `<dfn>` and adjusts the reference in the changelog.

### DIFF
--- a/css-conditional-5/Overview.bs
+++ b/css-conditional-5/Overview.bs
@@ -124,7 +124,7 @@ Extensions to the ''@supports'' rule</h2>
 		<dfn>&lt;supports-font-format-fn></dfn> = font-format( <<font-format>> )
 		<dfn>&lt;supports-at-rule-fn></dfn> = at-rule( <<at-keyword-token>> )
 		<dfn>&lt;supports-named-feature-fn></dfn> = named-feature( <<ident>> )
-		<dfn function lt="env()" for="@supports">&lt;supports-env-fn></dfn> = env( <<ident>> )
+		<dfn>&lt;supports-env-fn></dfn> = env( <<ident>> )
 	</pre>
 
 	: <<supports-condition-name>>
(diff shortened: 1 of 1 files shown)


bramus:
> It also created another exported definition for the `env()` function (less of a problem since it was scoped to `@supports`, but it does not seem needed)

The goal here was to (eventually) have `css-env-1` point to the `env()` from this spec for feature detection. I am not 100% familiar with all the details of Bikeshed, but won’t this proposed change prevent that from being possible?

(Looping in @tabatkins as the expert, who also approved my original PR that added this)

tidoust:
> The goal here was to (eventually) have `css-env-1` point to the `env()` from this spec for feature detection. I am not 100% familiar with all the details of Bikeshed, but won’t this proposed change prevent that from being possible?

Creating a function definition for `env()` scoped to `@supports` so that it can be pointed to by `css-env-1` is doable. Now, you cannot have a single definition for the type definition *and* the function definition as that confuses tools, including Bikeshed. For example, the link to `<supports-env-fn>` in the syntax of [`<supports-feature>`](https://drafts.csswg.org/css-conditional-5/#typedef-supports-feature) is currently broken, because Bikeshed is looking for a type and can no longer find it. Same thing for the missing link below the syntax that describes `<supports-env-fn>`.

This PR fixes the problem by getting back to a proper type definition (and then you may add a scoped function definition as needed). Alternatively, you could drop the type definition altogether and only have a scoped function definition, through:

```
<dfn>&lt;supports-feature></dfn> = ... | <<@supports/env()>> | ...
<dfn for="@supports">env()</dfn> = env( <<ident>> )
```

Happy to update this PR to do that. Feel free to also close the PR and do things differently, the PR is just meant to be a heads-up that there are broken links along with a possible solution to fix them ;)

tabatkins:
The spec should be fixed now. I've defined the `@supports/env()` function explicitly, in the list below the grammar block, and made all the other functions follow the same markup pattern. The grammar term is reverted to being a normal type definition.

Issue #13646, opened 2026-03-12
[selectors-4] `language-range` is missing its definition


aphillips:
7.2. The Language Pseudo-class: [:lang()](https://www.w3.org/TR/2026/WD-selectors-4-20260122/#lang-pseudo)
https://www.w3.org/TR/2026/WD-selectors-4-20260122/#language-range

> Each **language range** in [:lang()](https://www.w3.org/TR/2026/WD-selectors-4-20260122/#lang-pseudo) must be a valid CSS [<ident>](https://www.w3.org/TR/css-values-4/#typedef-ident) or [<string>](https://www.w3.org/TR/css-values-4/#string-value). (Thus language ranges containing asterisks, for example, must be either correctly escaped or quoted as strings, e.g. :lang(\*-Latn) or :lang("*-Latn").)

The `<dfn>` for "language range" has no definition and really it should provide one. Either it should be linked to the one in I18N Glossary ([here](https://www.w3.org/TR/i18n-glossary/#dfn-language-range)), to the definition of an _extended language range_ in BCP47 (in the RFC4647 part), or it should provide one locally.

svgeesus:
Certainly a definition should be linked to.

The I18n glossary entry seems imprecise ("similar to"):

> A string similar in structure to a language tag that is used for "identifying sets of language tags that share specific attributes".

svgeesus:
The [definition from BCP47](https://www.rfc-editor.org/rfc/rfc4647.html#section-2.2) looks crisper, with a defined grammar:

> Occasionally, users will wish to select a set of language tags based
   on the presence of specific subtags.  An "extended language range"
   describes a user's language preference as an ordered sequence of
   subtags.  For example, a user might wish to select all language tags
   that contain the region subtag 'CH' (Switzerland).  Extended language
   ranges are useful for specifying a particular sequence of subtags
   that appear in the set of matching tags without having to specify all
   of the intervening subtags.

>   An extended language range can be represented by the following ABNF:

>   `extended-language-range = (1*8ALPHA / "*")`
>  `                           *("-" (1*8alphanum / "*"))`

>   The wildcard subtag '*' can occur in any position in the extended
   language range, where it matches any sequence of subtags that might
   occur in that position in a language tag.  However, wildcards outside
   the first position are ignored by Extended Filtering (see [Section](https://www.rfc-editor.org/rfc/rfc4647.html#section-3.2.2)
[...]

PR #13748, opened 2026-03-31
[selectors-4] Defined language range 

svgeesus:
as BCP 47 extended language range, #13646

### DIFF
--- a/selectors-4/Overview.bs
+++ b/selectors-4/Overview.bs
@@ -2604,9 +2604,8 @@ The Language Pseudo-class: '':lang()''</h3>
 	represents an element whose [=content language=]
 	is one of the languages listed in its argument.
 	Each <dfn>language range</dfn> in '':lang()''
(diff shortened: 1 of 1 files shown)


selectors-4/Overview.bs
```diff
@@ -2604,6 +2604,9 @@ The Language Pseudo-class: '':lang()''</h3>
 	represents an element whose [=content language=]
 	is one of the languages listed in its argument.
 	Each <dfn>language range</dfn> in '':lang()''
+	is an <a href="https://www.rfc-editor.org/rfc/rfc4647.html#section-2.2">extended language range,
+	as defined in BCP 47, and
+	</a>
```
aphillips: Should this anchor close go just after the word "range"? I would probably double-bracket [[BCP47]] as well.

svgeesus: Yes, fixed

Issue #13477, opened 2026-02-10
[css-gaps-1] Decorations between empty areas: default behaviors


jav099:
Currently, the default value for `rule-visibility-items` is set to `all`. This means that rules will by default be shown even in segments where there are no items around the rule or when there is an item only on one side of the rule. This works as intended for grid, so no need to change it.

However, in the recent F2F an example was brought up for multicol https://www.software.hixie.ch/utilities/js/live-dom-viewer/?saved=14470, where `all` would create backwards compatibility issues since currently `column-rules` in multicol essentially behave as `rule-visibility-items: between` where they are only painted when surrounded by content on both sides. Details on `rule-visibility-items` can be found at https://drafts.csswg.org/css-gaps-1/#visibility-rules.

As a result, we'd like to propose `auto` for this property (and make it the default), so that we can resolve to `all` for grid containers (which is the current default) and to `between` for multicol containers to address backwards compat issues.

kbabbitt:
I agree that we should be backward compatible with existing multicolumn content, but based on the discussion in https://github.com/w3c/csswg-drafts/issues/13127#issuecomment-3814200432 I wonder if the new keyword should be `normal` rather than `auto`.

kbabbitt:
To illustrate the issue: https://codepen.io/kevinbcmu/pen/WbGrOGe
```css
.col {
  height: 100px;
  border: 1px solid blue;
  column-rule: 1px solid red;
  column-count: 6;
  width: 500px;
}
```

Behavior without Gap Decorations:

<img width="514" height="116" alt="Image" src="https://github.com/user-attachments/assets/014b149b-1bb1-42f2-b5d6-048b6e9b396b" />

Behavior implied by `rule-visibility-items: all`:

<img width="514" height="118" alt="Image" src="https://github.com/user-attachments/assets/5955cf03-cd94-4464-95ce-25b43a846cf8" />

mirisuzanne:
My instinct for grid was that I would most often want `between`, and I like the consistency with multicol. But (as others pointed out)…

- Many grids will be filled, and get the same result from both
- I'm not actually sure when `between` or `around` will be the better approach for grids with empty cells. I don't know if one of them is clearly a better default than the other. 
- It's very surprising to have a default that potentially results in _no decorations_ as you are building the grid, depending on content.

So even if I most often want one of `between`/`around`, I'm coming around to the argument that `all` is a better default.

kbabbitt:
[...]

PR #13755, opened 2026-04-01
[css-gaps-1] Add 'normal' value to 'rule-visibility-items'. #13477

kbabbitt:


kbabbitt:
@jav099 could you ptal to make sure I captured the resolution correctly?

edit: oh whoops, just noticed you already had a PR for this issue

css-gaps-1/Overview.bs
jav099: Should we update this line to reflect that `all` is not technically the default anymore?

kbabbitt: Yes, thanks for catching this.

Issue #10201, opened 2024-04-11
[css-anchor-position] Details on "anchors-valid"


tabatkins:
The "no-overflow" and "anchors-visible" values of [`position-visibility`](https://drafts.csswg.org/css-anchor-position/#position-visibility) were straightforward to define. The "anchors-valid" value, tho, ended up with several important questions to answer that I'm not sure about.

1. What exactly is a "required anchor reference"? 
    * `top: anchor(--foo bottom);` is clearly one. 
    * Is `top: anchor(--foo bottom, anchor(--bar bottom));` one (on --foo) or two (on --foo and --bar)? Or is it an "or" relationship - if *either* --foo or --bar is a valid reference, it works?
    * When exactly is the default anchor element a required reference? We can't just rely on `anchor(bottom)` references, as several other features reference the default anchor. Note that the initial value isn't "none", it's "implicit", so technically we're *always* **attempting** to reference a default anchor. Even if there isn't a single `anchor(bottom)` function on the element, the author could conceiveably be depending on the implicit anchor existing for its other effects.
    * How does fallback affect this? In general, `position-visibility` runs "after" fallback selection, so I presume we'd rely on the set of properties being used by the active fallback, and references in other fallbacks won't matter.

2. Do we want it to make the element invisible if *any* required references are invalid, or if *all* of them are? Is this something that needs to be controllable? (I'm inclined to say "any", since if you don't care about a given reference not resolving, you can always just give it a fallback value. Except for the default anchor...)

tabatkins:
Note that, since this issue was first published, the initial value of `position-anchor` was changed to `normal`, which is equivalent to `none` unless you're using `position-area`. So we're not longer relying on the default anchor *implicitly*, only if you purposely set `position-anchor` or use `position-area` (which very obviously relies on the default anchor). That simplifies things a little bit.

fantasai:
Opened a PR at https://github.com/w3c/csswg-drafts/pull/13768. While making the edits I noticed that the renaming would make `anchors-visible` inconsistent with `anchor-valid`, so wanted to check with the WG whether we want to:
* Rename both.
* Rename only one. :(
* Rename both and add legacy aliases.
* Something else.

CC @tabatkins @bfgeek @tuankiet65

tabatkins:
`anchors-visible` *also* only checks the default anchor, so we should rename both to `anchor-*` (with legacy aliases if needed)

PR #13768, opened 2026-04-03
[css-anchor-position-1] Depluralize anchors-* in 'position-visibility' and only check default anchor #10201

fantasai:
See https://github.com/w3c/csswg-drafts/issues/10201. While making the edits I noticed that the renaming would make `anchors-visible` inconsistent with `anchor-valid`, so opening a PR to check with the WG whether we want to:
* Rename both.
* Rename only one. :(
* Rename both and add legacy aliases.
* Something else.

### DIFF
--- a/css-anchor-position-1/Overview.bs
+++ b/css-anchor-position-1/Overview.bs
@@ -2219,8 +2219,8 @@ Conditional Hiding: the 'position-visibility' property {#position-visibility}
 
 <pre class=propdef>
 Name: position-visibility
-Value: always | [ anchors-valid || anchors-visible || no-overflow ]
-Initial: anchors-visible
+Value: always | [ anchor-valid || anchor-visible || no-overflow ]
+Initial: anchor-visible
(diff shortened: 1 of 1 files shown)


tuankiet65:
I'll assume if renamed, `anchor-visible` would only care about the default anchor like `anchor-valid`, right? Then I support 1) renaming both and 2) change their definition to only care about the default anchor.

I don't support adding aliases because it doesn't make sense to keep the plural form when we only care about one anchor (the default one). Also I don't want to carry historical baggage.

emilio:
Yeah, anchors-visible caring only about the default anchor but being plural is indeed quite weird. I'd support renaming to `anchor-visible`... But a parse time alias might be cheap enough and worth it?

tabatkins:
> But a parse time alias might be cheap enough and worth it?

I've gone ahead and added a MAY allowing a parse-time alias. We discussed it on the call, and we'll need it in Chrome. But it's not required to be supported.

PR #13799, opened 2026-04-13
[css-forms-1] Inherit basic appearance from fieldset to legend

lukewarlow:
Follow-up to #11981

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -55,6 +55,8 @@ spec:css-forms-1; type:value; for:/; text:::placeholder
 
   NOTE: The inheritance prevents authors from mixing native and non-native parts for the same control.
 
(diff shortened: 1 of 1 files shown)


fantasai:
A few questions:
- Should we add `legend { appearance: inherit; }` to the UA stylesheet?
- What happens if an author sets `legend { appearance: base; }` without `fieldset { appearance: base; }`?

lukewarlow:
> A few questions:
>
> - Should we add legend { appearance: inherit; } to the UA stylesheet?

No. Legend doesn't generally support the appearance property currently and I think we should keep it that way.

> - What happens if an author sets legend { appearance: base; } without fieldset { appearance: base; }?

It would do nothing. We should make the code look at the legends parent fieldset to determine whether the legend is in base mode.

Issue #12134, opened 2025-04-29
[scroll-animations-1] [web-animations-2] at progress timeline boundary and animation ranges


graouts:
I've been looking at a [scroll-driven animation demo](https://scroll-driven-animations.style/demos/carousel-with-markers/css/) which behaves differently in Safari Technology Preview (version 217) and Chrome (version 135.0.7049.115). In this demo, small dots are meant to show which "page" is shown within the scrolling container. Each dot has an animation attached to a scroll timeline for the entire scroll container and an `animation-range` to target the subsection of that scroll container matching the given "page".

The difference in behavior between Safari Technology Preview and Chrome shows when we've scrolled all the way to the end. At this stage, all dots appear as "on" in Safari Technology Preview only. This is use to the [at progress timeline boundary](https://drafts.csswg.org/web-animations-2/#at-progress-timeline-boundary) being computed to true for _all_ animations. Indeed, as far as I can tell from the current states of the spec, we don't account for animation ranges here to see whether we are matching the end of the animation range, which is probably what we should do.

I believe Chrome does _not_ implement the notion of _at progress timeline boundary_ and in fact this demo explicitly works around this by adding a pixel to the `animation-range-end` to make sure the 100% timeline time is still yielding an active animation for the final dot.

I think we should modify how the _at progress timeline boundary_  flag is computed to account for animation ranges. That flag was originally discussed in https://github.com/w3c/csswg-drafts/issues/5223.

Cc @flackr, @bramus, @kevers-google, @andruud, @majido and @birtles.

flackr:
@kevers-google I believe the spec is not correct here. We do implement the notion of at progress timeline boundary, however the way we implemented it is such that only the maximum scrolling position is at the progress timeline boundary, not 100% of a viewtimeline.

This particular exception was explicitly added because being at 100% of the scroll position is a common edge case that developers would overlook and since there can be no animation that starts after it there is no risk of having two animations active where one starts at the time the previous one ends.

TLDR, we should update the spec to correctly explain this.

dshin-moz:
Ran into this while [investigating](https://bugzilla.mozilla.org/show_bug.cgi?id=2047776) ongoing WPT failures with `view-timelines/contain-alignment.html` for Gecko, would like a clarification.

BorisChiou:
Yes. This spec issue is related to the interop test, so it'd be great if we could spec this case ASAP.

kevers-google:
Open PR that is related:
[...]

PR #13819, opened 2026-04-16
Update procedure for determining if at a progress timeline boundary

kevers-google:
[web-animations-2][scroll-animations-1] 

Update spec based on discussion in https://github.com/w3c/csswg-drafts/issues/12134.

Keeping the animation in effect at a phase boundary is intended to alleviate the need for a fill mode, when the scroll range for the animation aligns with the maximum or minimum scroll position.  The update aligns with the original intent of the feature.


web-animations-2/Overview.bs
```diff
@@ -1235,52 +1239,18 @@ an <a>animation effect</a>.
...
-<a>animation</a>'s current time.
+    1.  If the playState is *not* paused:
 
-Issue: This procedure can likely be simplified, and instead determine if at a
-scrolling boundary regardless of playback rate or start time. The surprising
-behavior that this is trying to prevent is an animation becoming inactive
-precisely at the scroll limit, alleviating the need for set a fill-mode with
-a ScrollTimeline. Checking if timeline [=timeline/current time=] is
-0 or timeline duration may be sufficient.
+        1.  If the timeline produces the minimum or maximum possible timeline
```
flackr: I think we need to define these as linkable spec terms, and say that for view or scroll timelines they are the time value at min and max scroll offset, and can just be +/- infinity for time based timelines for now.

web-animations-2/Overview.bs
```diff
@@ -1232,55 +1236,37 @@ condition:
...
-    1.  Let <var>effective timeline time</var> be
-        <code>|unlimited current time| / <a>animation</a>'s
-        [=playback rate=] + |effective start time|</code>
+:   <dfn>at progress timeline boundary</dfn>
+::  Determined using the following procedure:
 
-    1.  Let <var>effective timeline progress</var> be
-        <code><var>effective timeline time</var> / [=timeline duration=]</code>
+    1.  If  the associated <a>animation</a>'s timeline is not a
+            [=progress-based timeline=]
```
flackr: This early return is likely irrelevant if minimum and maximum timeline time are defined and not reachable by regular timelines.

web-animations-2/Overview.bs
```diff
@@ -1232,55 +1236,37 @@ condition:
...
+not change from being in effect to not in effect if the scroll position changes
+while in a paused state.
 
-        :   [=animation/start time=] is resolved:
+<dfn>minimum timeline time</dfn>
 
-        ::  <code>(<var>timeline time</var> - [=animation/start time=])
-            &times; [=playback rate=]</code>
+For a scroll-timeline, the minimum timeline time is the timeline time associated
+with the minimum scroll offset. For a time-based timeline, it is -infinity.
```
flackr: Can you define these timeline values under section 2.2 alongside the duration that was added for timelines:
https://drafts.csswg.org/web-animations-2/#timelines

web-animations-2/Overview.bs
```diff
@@ -1232,55 +1236,37 @@ condition:
...
-    1.  If <var>effective timeline progress</var> is 0 or 1, return true,
-        otherwise false.
+        return false
 
-Issue: This procedure is not strictly correct for a paused
-animation if the <a>animation</a>'s current time is explicitly set, as this can
-introduce a lead or lag, between the <a>timeline</a>'s current time and
-<a>animation</a>'s current time.
+    1.  If the playState is *not* paused, or
+        [=held at progress timeline boundary=] is unresolved:
```
flackr: I'm not sure about this having a live value with a paused animation when held at progress timeline boundary is unresolved - however that happens. Do you have an example of what actions cause this? Can we "resolve" the held value?

web-animations-2/Overview.bs
```diff
@@ -1232,55 +1236,37 @@ condition:
...
 
-Issue: This procedure can likely be simplified, and instead determine if at a
-scrolling boundary regardless of playback rate or start time. The surprising
-behavior that this is trying to prevent is an animation becoming inactive
-precisely at the scroll limit, alleviating the need for set a fill-mode with
-a ScrollTimeline. Checking if timeline [=timeline/current time=] is
-0 or timeline duration may be sufficient.
+        1.  If the timeline produces the [=minimum timeline time=] or
```
flackr: When you say "produces" do you mean the current time?

Shouldn't we only return true if the animation's range includes this time?

web-animations-2/Overview.bs

web-animations-2/Overview.bs
kevers-google: Removed the need for considering playstate.

web-animations-2/Overview.bs
kevers-google: The phase calculation asserts that the local time matches the before-active or active-after boundary. This procedure is simplified to simply indicate if the animation's currentTime is the minimum or maximum possible value given the timeline.  So if you are at the before-active boundary and can't produce a lower value for currentTime based on the timeline then inclusive. Similarly if at the active-after boundary and can't produce a higher value based on the timeline, then inclusive.

web-animations-2/Overview.bs

web-animations-2/Overview.bs
```diff
@@ -210,6 +210,18 @@ Add:
...
+>
+> <dfn>minimum timeline time</dfn>
+>
+> For a scroll-timeline, the minimum timeline time is the timeline time
+> associated with the minimum scroll offset. For a time-based timeline, it is
+> -infinity.
+>
+> <dfn>maximum timeline time</dfn>
+>
+> For a scroll-timeline, the minimum timeline time is the timeline time
```
flackr: Typo s/minimum/maximum

web-animations-2/Overview.bs
flackr: s/at progress timeline boundary/at timeline boundary

web-animations-2/Overview.bs
```diff
@@ -1232,54 +1244,24 @@ condition:
...
-            unresolved or zero, or
-        *   the <a>animation</a>'s [=playback rate=] is zero
-
-        return false
-
-    1.  Let <var>effective start time</var> be the <a>animation</a>'s
-        [=animation/start time=] if resolved, or zero otherwise.
+    <dl class="switch">
+        :   If the [=minimum timeline time=] matches the animation's
+            [=current time=], and playbackRate = -1:
```
flackr: I'm not sure I understand this, the animation's current time is a calculation based on the timeline time, https://www.w3.org/TR/web-animations-1/#the-current-time-of-an-animation . As such, couldn't the animation be at time 0 when the timeline time is not at time 0, matching this condition.

Should this be the timeline current time? https://www.w3.org/TR/web-animations-1/#timeline-current-time
I.e. I think it doesn't care about the animation you're asking about because we already only call this for an animation that is at its before/after-active boundary.

web-animations-2/Overview.bs
flackr: Ah right, I mixed up that this procedure is only meant to be about the timeline's state, and the place we call it has already checked the state of the animation.

web-animations-2/Overview.bs
kevers-google: Based on offline discussion, simplified to require matching timeline current time. This breaks paused animations but avoids complications regarding the effect of start time on running animations.

web-animations-2/Overview.bs

flackr (approved):

web-animations-2/Overview.bs
```diff
@@ -210,6 +210,18 @@ Add:
 > (e.g. scroll) timeline, the duration has a fixed upper bound. In this
 > case, the timeline is a <dfn lt="progress-based timeline">progress-based
 > timeline</dfn>, and its [=timeline duration=] is 100%.
+>
+> <dfn>minimum timeline time</dfn>
+>
+> For a scroll-timeline, the minimum timeline time is the timeline time
+> associated with the minimum scroll offset. For a time-based timeline, it is
+> -infinity.
```
graouts, line 218: I think the terms we want to use here are "progress-based" and "monotonic" rather than "scroll-timeline" and "time-based timeline". This would better align with how the rest of the spec is worded.

web-animations-2/Overview.bs
```diff
@@ -210,6 +210,18 @@ Add:
...
+>
+> <dfn>maximum timeline time</dfn>
+>
+> For a scroll-timeline, the maxiumum timeline time is the timeline time
+> associated with the maximum scroll offset, and matches the
+> [=timeline duration=]. For a time-based timeline, it is +infinity.
```
graouts, line 224: Ditto about "progress-based" and "monotonic".

graouts (approved):

web-animations-2/Overview.bs
```diff
@@ -1232,55 +1244,34 @@ condition:
...
-            unresolved or zero, or
-        *   the <a>animation</a>'s [=playback rate=] is zero
-
-        return false
-
-    1.  Let <var>effective start time</var> be the <a>animation</a>'s
-        [=animation/start time=] if resolved, or zero otherwise.
+    <dl class="switch">
+        :   If the [=minimum timeline time=] matches the
+            [=timeline current time=], and playbackRate = -1:
```
dshin-moz: (As discussed in weekly meeting, directionality of playback may matter, but the exact playback rate should not matter)

web-animations-2/Overview.bs
```diff
@@ -215,6 +215,21 @@ Add:
...
 > timeline</dfn>, and its [=timeline duration=] is 100%.
+>
+> <dfn>minimum timeline time</dfn>
+>
+> For a progress-based timeline, the minimum timeline time is the minimum value
+> that can be produced for the timeline. For a monotonic timeline, minimum timeline time is -infinity.
+>
+> <dfn>maximum timeline time</dfn>
+>
+> For a progress-baed timeline, the maximum timeline time is the maximum
```
flackr: Typo, s/baed/based

Also, progress-based is a linkable term, use `[=progress-based timeline=]`

web-animations-2/Overview.bs
```diff
@@ -215,6 +215,21 @@ Add:
...
+> possible value producible for timeline time, and matches the
+> [=timeline duration=]. For a monotonic timeline, it is +infinity.
```
flackr: s/monotonic timeline/[=monotonically increasing timeline=]

web-animations-2/Overview.bs

web-animations-2/Overview.bs

web-animations-2/Overview.bs
kevers-google: Removed playback rate.

web-animations-2/Overview.bs

web-animations-2/Overview.bs

web-animations-2/Overview.bs
```diff
@@ -218,22 +218,25 @@ Add:
 >
 > <dfn>minimum timeline time</dfn>
 >
-> For a progress-based timeline, the minimum timeline time is the minimum value
-> that can be produced for the timeline. For a monotonic timeline, minimum timeline time is -infinity.
+> For a [=progress-based timeline=], the minimum timeline time is the minimum
+> value that can be produced for the timeline. For a
+[=monotonically increasing timeline=], minimum timeline time is -infinity.
```
flackr: Looks like this line is missing the `> ` at the start to denote that it continues the quoted block.

flackr (approved):

PR #13822, opened 2026-04-17
[css-values-5][editorial] Oversights in 5ef76ec

cdoublev:
The `property` and `index` keywords were changed to [`property-scoped` and `property-index-scoped`](https://drafts.csswg.org/css-values-5/#typedef-random-name) in 5ef76ec.


fantasai:
Thanks for catching those!

PR #13823, opened 2026-04-17
[web-animations-1][editorial]: Pass correct value to "set the associated effect of an animation"

simonwuelker:
The existing specification passes `source`, which is not defined anywhere. Presumably the author meant to pass `effect` instead.


simonwuelker:
I assume this change falls under "non-substantitative changes" that don't need require me to join the CSSWG (https://labs.w3.org/repo-manager/pr/id/w3c/csswg-drafts/13823)

PR #13827, opened 2026-04-20
[css-values-5] Fix and rename random grammar

cdoublev:
The [`random-*()`](https://drafts.csswg.org/css-values-5/#random) grammar is slightly broken since 33682af.

They now take a [`<random-key>`](https://drafts.csswg.org/css-values-5/#typedef-random-key) argument renamed from `<random-cache-key>`. It still produces `auto | <random-name> | fixed <number [0,1)>` but `<random-name>` is left undefined, because its definition and production rule now use `<random-cache-key>`, which was probably not intentional.

This PR tries to fix this but since `fixed <number [0,1)>` is not a random (cache) key but a random base value, and since `auto` produces but is not a random name, and resolves to a random base value, I suggest to introduce `<random-base-value>` to replace `<random-key>`.

### DIFF
--- a/css-values-5/Overview.bs
+++ b/css-values-5/Overview.bs
@@ -2286,7 +2286,7 @@ Generating a Random Numeric Value: the ''random()'' function</h3>
 	optionally limiting the possible values to a step between those limits:
 
 	<pre class=prod>
-	&lt;random()> = random( <<random-key>>? , <<calc-sum>>, <<calc-sum>>, <<calc-sum>>? )
+	&lt;random()> = random( <<random-base-value>>? , <<calc-sum>>, <<calc-sum>>, <<calc-sum>>? )
 	</pre>
 
 	See [[#random-evaluation]] and [[#random-caching]]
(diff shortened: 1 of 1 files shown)


fantasai:
I think this problem is fixed in https://github.com/w3c/csswg-drafts/pull/13845/changes ?

Or is something still broken?

cdoublev:
Yes, it is fixed by #13845.

Not exactly something broken, but some things that generally seem to lack precision:

  - `<random-key>` produces `fixed <number>`, which is a *random base value*, not a *random cache name/key*
  - the *random cache name* is built from the `<random-cache-key>` but is not *entirely* specified with it

  > The [`<random-cache-key>`](https://drafts.csswg.org/css-values-5/#typedef-random-cache-key) specifies the [random cache name](https://drafts.csswg.org/css-values-5/#random-cache-name)

https://drafts.csswg.org/css-values-5/#typedef-random-cache-key

But I assume these will be fixed as you refine these terms and definitions.

PR #13830, opened 2026-04-21
[css-text-5] Only accept `none` as a standalone keyword for `text-fit`

cdoublev:
@tkent-google, could you tell me if this is an oversight in #13616?

Each of `consistent | per-line | per-line-all` is explicitly defined as having no effect if `none` is specified. My understanding is that no scaling factor is applied with `none`, so I do not see how `none <percentage>` could be useful.

https://drafts.csswg.org/css-text-4/#propdef-text-fit


tkent-google (approved):
Yes, this change is reasonable.

PR #13833, opened 2026-04-22
[css-navigation-1] Move grammar note out of production block

tidoust:
The Value Definition Syntax does not have provisions for adding comments to grammar production blocks:
https://drafts.csswg.org/css-values-4/#production-blocks

This update moves the comment out of the block to a follow-up note so that tools can continue to extract the grammar from the spec.


dontcallmedom:
overtaken by https://github.com/w3c/csswg-drafts/commit/60501e835289cd8e76462395608e5c2c869a06fb

PR #13842, opened 2026-04-27
[css-gaps-1]: Introduce spec changes for *-inset-cap-* and *-inset-junction-* #13697

### LINKED ISSUE #13697: [css-gaps-1] `overlap-join` with `between` rule visibility
This came up while @oSamDavis @jav099 and I were discussing implementation of `rule-inset: overlap-join` which we resolved to add in #13137.

The intent of that value is to give authors an easy way to join decoration segments neatly with each other. Where two perpendicular segments meet, it's equivalent to `rule-inset: calc( -(50% + crossing_decoration_width/2) )` or in English, "halfway into the intersection, plus enough to cover the remaining width of the crossing decoration":

Image

But there's an edge case that arises when this value is used with `rule-visibility-items: between`. That value causes us to paint decorations only where there are items on both sides of the gap. The `calc()` treatment above would lead to decorations still protruding into intersections even when there's no crossing decoration for it to join:

Image

My initial reaction was that we should not do that and instead leave such decorations flush with the adjacent items, which would be equivalent to an inset of `0` at those specific endpoints. But I'm somewhat worried that behavior might be too magical. What do folks think?

oSamDavis:
Following the resolution in #13697, this PR includes the necessary spec updates, for the cap and junction model of the inset properties.


css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,82 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
 	These properties can be used to offset the start and end points of <a>gap decorations</a> relative to the
 	<a>segment endpoints</a> which would normally determine where decorations start and end.
 
-	<p>An <dfn>edge segment endpoint</dfn> is any [=segment endpoint=] created at the content edges of the container.</p>
-	<p>An <dfn>interior segment endpoint</dfn> is any [=segment endpoint=] created at a [=gap junction=].</p>
+	[=Gap decorations=] are divided into [=gap decoration segments=] at [=gap junctions=].
+	Once [=gap decoration segments=] have been determined, each [=segment endpoint=] can be classified
+	according to whether its segment joins with another [=gap decoration segment=] at that
```
alisonmaher: Using the word "join" here may be slightly confusing because it can still be a junction but not actually be joined (via the `join` keyword).

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,82 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
 	These properties can be used to offset the start and end points of <a>gap decorations</a> relative to the
 	<a>segment endpoints</a> which would normally determine where decorations start and end.
 
-	<p>An <dfn>edge segment endpoint</dfn> is any [=segment endpoint=] created at the content edges of the container.</p>
-	<p>An <dfn>interior segment endpoint</dfn> is any [=segment endpoint=] created at a [=gap junction=].</p>
+	[=Gap decorations=] are divided into [=gap decoration segments=] at [=gap junctions=].
```
alisonmaher: I'm not sure this sentence fully makes sense to me. Usually when saying something is divided, you would say something "divided between x and y". Here we say they are divided into one thing (gap decoration segments).

Perhaps something along the lines of "Gap decorations are split into two categories of gap decoration segments based on the state of the gap junction it intersects." Or maybe we just remove this sentence altogether since that is somewhat covered by the next sentence

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,82 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+	Once [=gap decoration segments=] have been determined, each [=segment endpoint=] can be classified
+	according to whether its segment joins with another [=gap decoration segment=] at that
+	location. The inset properties below are defined based on this classification:
+
+	<p>
+		A <dfn>cap segment endpoint</dfn> is any [=segment endpoint=] whose [=gap decoration segment=]
+		cannot join with any other [=gap decoration segment=] at that location.
```
alisonmaher: "cannot join with" may be a bit of a strong wording (since it's not that it can't, but that it doesn't have any other decoration to join with.

But I also wonder if we should try to avoid using "join" when describing this, similar to what was mentioned above.

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,82 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+		This includes endpoints at the container's content edges,
+		as well as endpoints left dangling at a [=gap junction=] because no corresponding [=gap decoration segment=]
```
alisonmaher: I might reword to avoid danling (since it may not be obvious what that means without defining it) and just say something along the lines an endpoint within the container that meets a gap junction point that is empty on all sides but one

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,82 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+		exists to join with due to 'rule-break' or 'rule-visibility-items' properties.
+	</p>
+
+	<p>
+		A <dfn>junction segment endpoint</dfn> is any [=segment endpoint=] whose [=gap decoration segment=] can join
+		with one or more other [=gap decoration segments=] at the same [=gap junction=].
```
alisonmaher: Similar comment to above, wondering if we can avoid join and maybe reword to focusing on the number of other segments meeting at the same junciton

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,82 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+			.simple-grid-with-empty-cells {
+				rule-break: intersection;
+				rule-visibility-items: between;
+			}
+		</pre>
+		<figure>
+			<img alt="A 3x3 grid with only the top-left five cells occupied. Where a column rule between filled cells terminates at a gap junction with no row rule reaching it, a red C marker is shown instead of a blue J." src="images/example-cap-junct-less-simple-grid.png">
+			<figcaption>
+				When a [=gap decoration segment=] terminates at an interior [=gap junction=]
+				with no other [=gap decoration segment=] there to join with,
```
alisonmaher: Same thing about join here

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,82 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+		<pre>
+			.donut-grid-layout {
+				rule-break: intersection;
+				rule-visibility-items: between;
+			}
+		</pre>
+		<figure>
+			<img alt="A 4x4 donut-shaped grid with an empty middle row and middle-right cells. Red C markers label the dangling ends of column and row rules bordering the empty region, while blue J markers label the few endpoints where rules genuinely meet." src="images/example-cap-junct-donut-grid.png">
+			<figcaption>
+				In a "donut" grid layout, endpoints bordering the empty interior region are [=cap segment endpoints=].
```
alisonmaher: Since "donut" may not be obvious, I think it could be helpful to define what that means (i.e. "In a "donut" grid layout, or one where items surround an empty space in the middle of the container"

css-gaps-1/Overview.bs
```diff
@@ -960,7 +1024,7 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
 			}
 		</pre>
 		<figure>
-			<img alt="" src="images/example-column-start-end-edge-interior-insets.png">
+			<img alt="" src="images/example-column-start-end-inset-cap-junction.png">
```
mhochk: NIT: It's only a file name so feel free to ignore the suggestion, but consider re-ordering so that `inset` comes earlier and `start-end` always comes last, similar to the property names. Also consider dropping `cap` and `junction` completely, as this picture doesn't illustrate controlling those independently.

```suggestion
			<img alt="" src="images/example-column-inset-start-end.png">
```

css-gaps-1/Overview.bs
mhochk: I agree throughout that saying "can join" or "joins" may be confusing. It's somewhat of a circular definition to say the "join segment endpoints" are there one where you can join the segments, if you want to then use "join segment endpoints" as part of your definition of how segments join.

What about focusing on the number of endpoints in the gap junction instead?

```suggestion
	according to the number of [=segment endpoint=]s in the [=gap junction=].
```

...

```
A <dfn>junction segment endpoint</dfn> is any [=segment endpoint=] inside a [=gap junction=] that also contains another [=segment endpoint=].
```

oSamDavis: Okay removed this in favor of the presence of a gap segment

css-gaps-1/Overview.bs

css-gaps-1/Overview.bs

css-gaps-1/Overview.bs

css-gaps-1/Overview.bs

css-gaps-1/Overview.bs

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,77 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+	<p>
+		A <dfn>junction segment endpoint</dfn> is a [=segment endpoint=] at a [=gap junction=]
+		where one or more other [=gap decoration segments=] are also present.
+	</p>
+
+	<p>
+		A <dfn>cap segment endpoint</dfn> is any [=segment endpoint=] that is not a [=junction segment endpoint=].
+		This includes endpoints at the container's content edges,
+		as well as endpoints at a [=gap junction=] where no other [=gap decoration segment=] is present
+		due to 'rule-break' or 'rule-visibility-items' properties.
```
alisonmaher: "due to combinations of 'rule-break' and 'rule-visibility-items' properties where a segment may be dropped."

We could also leave this out in case there are other places in the future that could lead to dropped segments (making this more future proof)

alisonmaher (approved):

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,77 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
 	<dl>
 		<dt>
-			Edge segment endpoints
+			Segment endpoints that coincide with the edge of the container
 		<dd>
 			The [=crossing gap width=] is ''0''.
 
 		<dt>
-			Interior segment endpoints
+			Segment endpoints that are within the container
```
mhochk: "Segment endpoints that are within the container" kindof feels like it applies to all endpoints. I understand the intention is "in" vs "on the edge of", but could be worth considering a wording tweak to make sure it isn't misunderstood by future readers.

What about re-framing along the lines of "Segment endpoints that are at a [=gap junction=]"? (and then probably update the edge definition as well to match)

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,77 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+				endpoints bordering the empty interior region are [=cap segment endpoints=].
+			</figcaption>
+		</figure>
+	</div>
 
 	Percentages are resolved against the <dfn>crossing gap width</dfn>, defined as follows:
 	<dl>
 		<dt>
-			Edge segment endpoints
+			Segment endpoints that coincide with the edge of the container
```
mhochk: Now that caps are different than "edge of the container", it might be good to call out the distinction in behavior. Maybe adding a Note about how this (correctly) means that for the donut case, setting a cap inset of 50% would impact the interior cap endpoints, but the ones on the edge of the container would evaluate to 0.

mhochk (approved):

css-gaps-1/Overview.bs
jav099: What about the negation? ~ endpoints that are NOT bordering the edge of the container or something along those lines?

oSamDavis: Personally, I find it clearer to keep this Percentage resolution section focused on whether endpoints are on the container edge or not. This approach is simpler to follow, compared to using cap or junction as the basis, which would more likely introduce a caveat.

css-gaps-1/Overview.bs
oSamDavis: Same with my comment above

css-gaps-1/Overview.bs
mhochk: I'm not attached to my suggestion rewording. The main point was that "within the container" could be misinterpreted to include those on the edges, as even within this same document the phrase "within a given container" is used to mean *all* the things, not excluding the things that touch the edges.

"Setting multiple values in this way allows for varying <a>gap decorations</a> within a given container."

If you really want to stick with edges vs interior, Javier's suggestion sounds fine to me as well. `Segment endpoints that do NOT coincide with the edge of the container`
`

css-gaps-1/Overview.bs
mhochk: I'm not sure how to interpret your neighboring comment in this context.

This comment thread is not me suggesting changing the definition, simply helping to make it clear that this is intentionally not the same definition as `cap`. I worry it may be assumed by a reader that they either are the same, or are intended to be the same, since they end up being the same for the "normal" cases.

css-gaps-1/Overview.bs
oSamDavis: Okay, I'll take out

css-gaps-1/Overview.bs
oSamDavis: Oh, I see. I misunderstood then. In that case, I'll go with "...do NOT...". Thanks!

css-gaps-1/Overview.bs
mhochk: Note: This is intentionally *not* the same category as a [=cap segment endpoint=].

oSamDavis: I added somewhat of a more verbose note. PTAL and let me know if that's sufficient.

mhochk: LGTM - thank you.

css-gaps-1/Overview.bs
```diff
@@ -872,14 +930,22 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
 			See <a href="#example-unpaired-endpoints-1">this example</a> for one such case.
 	</dl>
 
+	<p class="note">
+      Note: This classification by position is distinct from the
+      [=cap segment endpoint=] / [=junction segment endpoint=] classification.
+	  A [=cap segment endpoint=] may sit at a container edge or at a [=gap junction=]
+	  (see <a href="#example-cap-junction-donut">this example</a>). Hence, ''column-rule-inset-cap-start: 50%''
+      resolves to ''0'' at a container edge, but to half the [=gap junction=] size at an interior cap.
```
kbabbitt: Bikeshed is throwing warnings on these lines; please fix:
```
LINE 934: Line starts with spaces, but document is tab-indented
LINE 935: Line starts with spaces, but document is tab-indented
LINE 938: Line starts with spaces, but document is tab-indented
```

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,76 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
 	These properties can be used to offset the start and end points of <a>gap decorations</a> relative to the
 	<a>segment endpoints</a> which would normally determine where decorations start and end.
 
-	<p>An <dfn>edge segment endpoint</dfn> is any [=segment endpoint=] created at the content edges of the container.</p>
-	<p>An <dfn>interior segment endpoint</dfn> is any [=segment endpoint=] created at a [=gap junction=].</p>
+	<p>
+		A <dfn>junction segment endpoint</dfn> is a [=segment endpoint=] at a [=gap junction=]
+		where one or more other [=gap decoration segments=] are also present.
```
kbabbitt, line 999:
```suggestion
		where one or more other [=segment endpoints=] are also present.
```

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,76 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+	</p>
+
+	<p>
+		A <dfn>cap segment endpoint</dfn> is any [=segment endpoint=] that is not a [=junction segment endpoint=].
+		This includes endpoints at the container's content edges,
+		as well as endpoints at a [=gap junction=] where no other [=gap decoration segment=] is present.
```
kbabbitt, line 1005:
```suggestion
		as well as endpoints at a [=gap junction=] where no other [=segment endpoint=] is present.
```

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,76 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+			.simple-grid-with-empty-cells {
+				rule-break: intersection;
+				rule-visibility-items: between;
+			}
+		</pre>
+		<figure>
+			<img alt="" src="images/example-cap-junct-less-simple-grid.png">
+			<figcaption>
+				When a [=gap decoration segment=] terminates at an interior [=gap junction=]
+				where no other [=gap decoration segment=] is present,
```
kbabbitt, line 1036:
```suggestion
				where no other [=segment endpoint=] is present,
```

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,76 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+				endpoints bordering the empty interior region are [=cap segment endpoints=].
+			</figcaption>
+		</figure>
+	</div>
 
 	Percentages are resolved against the <dfn>crossing gap width</dfn>, defined as follows:
 	<dl>
 		<dt>
-			Edge segment endpoints
+			Segment endpoints that coincide with the edge of the container
```
```suggestion
			Segment endpoints that coincide with an edge of the container
```

css-gaps-1/Overview.bs
```diff
@@ -847,18 +847,76 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+		<figure>
+			<img alt="" src="images/example-cap-junct-donut-grid.png">
+			<figcaption>
+				In a "donut" grid layout, where items surround an empty space in the middle of the container,
+				endpoints bordering the empty interior region are [=cap segment endpoints=].
+			</figcaption>
+		</figure>
+	</div>
 
 	Percentages are resolved against the <dfn>crossing gap width</dfn>, defined as follows:
```
kbabbitt, line 1058: I think the distinction here is actually:

```bs
<dl>
    <dt>
        Segment endpoints at a [=gap junction=]
    <dd>
        The [=crossing gap width=] is the size of the [=gap junction=]
        /* etc. ... */
    <dt>
        Other segment endpoints
    <dd>
        The [=crossing gap width=] is ''0''.
</dl>
```

The reason I think this is that the percentage basis for segment endpoints next to a spanner in multicol is also 0; see https://codepen.io/kevinbcmu/pen/OPbJwWw

Let's also make sure we have a WPT for that case.

css-gaps-1/Overview.bs
```suggestion
      resolves to half the [=gap junction=] size at an interior cap and ''0'' at other caps such as container edges.
```

(Presuming we take [the suggestion above](https://github.com/w3c/csswg-drafts/pull/13842/changes#r3164146337))

kbabbitt (changes requested):

css-gaps-1/Overview.bs
kbabbitt: Chatted with @mhochk offline about this and he caught me up, comment retracted (and same for similar comments)

css-gaps-1/Overview.bs
kbabbitt: Comment retracted

css-gaps-1/Overview.bs
kbabbitt: Comment retracted

css-gaps-1/Overview.bs

css-gaps-1/Overview.bs
oSamDavis: Done, let me know if that's sufficient.

kbabbitt (approved):

PR #13846, opened 2026-04-28
[filter-effects-1] Document Clickjacking Attacks

arichiv:
This formalizes the proposal here: https://chromestatus.com/feature/5117170452398080

### DIFF
--- a/filter-effects-1/Overview.bs
+++ b/filter-effects-1/Overview.bs
@@ -3548,8 +3548,9 @@ A timing attack is a method of obtaining information about content that is other
 
 <h3 id="clickjacking-attack">Clickjacking Attacks</h3>
 
-User agents must not apply SVG reference <a element>filter</a>s to cross-origin/restricted iframes or web plugins.
+User agents must not apply SVG reference <a element>filter</a>s to cross-origin/restricted iframes or web plugins because of their ability to arbitrarily manipulate pixels.
 If this is permitted, it allows the manipulation of otherwise inaccessable frames, potentially causing users to take actions they might not otherwise.
+See the Clickjacking section of <a href="https://html.spec.whatwg.org/multipage/introduction.html#writing-secure-applications-with-html">HTML Spec</a>.
 
 <h2 id=security>Security Considerations</h2>
 
--- a/filter-effects-1/Overview.bs
+++ b/filter-effects-1/Overview.bs
@@ -3546,6 +3546,11 @@ If any of the above rules are not followed, an attacker could infer information 
 
 A timing attack is a method of obtaining information about content that is otherwise protected, based on studying the amount of time it takes for an operation to occur. If, for example, red pixels took longer to draw than green pixels, one might be able to reconstruct a rough image of the element being rendered, without ever having access to the content of the element. Security studies show that timing differences on arithmetic operations can be caused by the hardware architecture or compiler [[ArTD]].
 
+<h3 id="clickjacking-attack">Clickjacking Attacks</h3>
+
+User agents must not apply SVG reference <a element>filter</a>s to cross-origin/restricted iframes or web plugins.
+If this is permitted, it allows the manipulation of otherwise inaccessable frames, potentially causing users to take actions they might not otherwise.
+
 <h2 id=security>Security Considerations</h2>
 
 Besides the security implications of the privacy issues noted above,
--- a/filter-effects-1/Overview.bs
+++ b/filter-effects-1/Overview.bs
@@ -3548,7 +3548,7 @@ A timing attack is a method of obtaining information about content that is other
 
 <h3 id="clickjacking-attack">Clickjacking Attacks</h3>
 
-User agents must not apply SVG reference <a element>filter</a>s to cross-origin/restricted iframes or web plugins because of their ability to arbitrarily manipulate pixels.
+User agents must not apply SVG reference filters (e.g., <a element>filter</a> or <a element>backdrop-filter</a>) to pixel data from cross-origin/restricted iframes or web plugins.
 If this is permitted, it allows the manipulation of otherwise inaccessable frames, potentially causing users to take actions they might not otherwise.
 See the Clickjacking section of <a href="https://html.spec.whatwg.org/multipage/introduction.html#writing-secure-applications-with-html">HTML Spec</a>.
 


filter-effects-1/Overview.bs
```diff
@@ -3546,6 +3546,11 @@ If any of the above rules are not followed, an attacker could infer information
 
 A timing attack is a method of obtaining information about content that is otherwise protected, based on studying the amount of time it takes for an operation to occur. If, for example, red pixels took longer to draw than green pixels, one might be able to reconstruct a rough image of the element being rendered, without ever having access to the content of the element. Security studies show that timing differences on arithmetic operations can be caused by the hardware architecture or compiler [[ArTD]].
 
+<h3 id="clickjacking-attack">Clickjacking Attacks</h3>
+
+User agents must not apply SVG reference <a element>filter</a>s to cross-origin/restricted iframes or web plugins.
```
progers: Because of their ability to arbitrarily manipulate pixels, `SVG filters` should not be applied to cross-origin frames?

filter-effects-1/Overview.bs
```diff
@@ -3546,6 +3546,11 @@ If any of the above rules are not followed, an attacker could infer information
 
 A timing attack is a method of obtaining information about content that is otherwise protected, based on studying the amount of time it takes for an operation to occur. If, for example, red pixels took longer to draw than green pixels, one might be able to reconstruct a rough image of the element being rendered, without ever having access to the content of the element. Security studies show that timing differences on arithmetic operations can be caused by the hardware architecture or compiler [[ArTD]].
 
+<h3 id="clickjacking-attack">Clickjacking Attacks</h3>
+
```
progers, line 3550: WDYT of linking to the html spec definition of clickjacking at https://html.spec.whatwg.org/#writing-secure-applications-with-html ?

filter-effects-1/Overview.bs
tabatkins: The `feDisplacementMap` restriction, above, already handles some of this; a displacement filter applied to an iframe would be using a tainted input and thus would become a no-op. Do we need to expand the restriction like this?

rebane2001: Yes, because the user will still visually see the output of the SVG filter. This makes it possible to trick the user into performing actions, as well as exfiltrating data through user feedback or QR codes.

I wrote more about the attack vectors here: https://lyra.horse/blog/2025/12/svg-clickjacking/

arichiv:
Chrome is currently discussing whether all iframes, or just iframes that are cross-origin with their parent, should have the SVG filter un-applied. Soliciting feedback on that here.

rebane2001:
From a security perspective, same-origin frames allowing filters would be on-par with how the rest of the restrictions with frames work. The concerns would probably be additional complexity in the case of nested frames, and confusing behavior where filters work on one frame but not another. I personally like the idea of still being able to use filters on my own frames :).

tabatkins:
Yeah, there's no good reason to disallow filters on same-origin iframes. A same-origin iframe is not remotely a security boundary.

filter-effects-1/Overview.bs
tabatkins: Makes sense (and *incredible* blog post).

filter-effects-1/Overview.bs
```diff
@@ -3546,6 +3546,12 @@ If any of the above rules are not followed, an attacker could infer information
...
+User agents must not apply SVG reference <a element>filter</a>s to cross-origin/restricted iframes or web plugins because of their ability to arbitrarily manipulate pixels.
```
Psychpsyo, line 3551: What exactly does this mean?
If I set an SVG reference filter somewhere further up on the tree, so that it applies to some area of the page and the cross-origin frame within, does the filter as a whole become invalid? Or is it expected to leave only the iframe alone somehow and still apply to the content around it?

arichiv: I think it's up to the user agent, though @progers may disagree. Assuming a setup like:

```
<div class="svg-filter">
  [...stuff...]
  <iframe src="remote origin"></iframe>
</div>
```

Ideally the 'stuff' still has the filter applied while the iframe does not, but if the design of the painter for the user agent precludes this somehow, it's better to fully invalidate the filter on all content than apply it to the iframe.

rebane2001 (approved):

filter-effects-1/Overview.bs
Psychpsyo: I think the spec should *at least* be explicit with regards to the fact that this should apply to all reference filters that would/could end up using pixel data from the iframe. Otherwise, to a particularily lazy implementer, the interpretation of "only filters applied to the iframe directly" is on the table, which ends up negating all the intended security gains. Similarily, one might naively expect this to mean "all filters up the iframe's hierarchy", which is problematic as well, because of `backdrop-filter`.

I also worry that, if the details of this are left unspecified, it will lead to other browsers ultimately having to reverse-engineer Chrome once they get bug reports along the lines of "This backdrop filter on my modal isn't applying at all, but it works in Chrome."

Then again, I just tried that and Chrome 151 will still happily apply these filters to an iframe if done via `backdrop-filter` on an abspos element in front of the frame. So maybe it wouldn't be the worst idea for the spec to give some guidance here. WebKit and Gecko seem to get around this by just not applying the filter at all in that case. (In the regular `filter` case, they both manage to exempt only the iframe and not the rest of the page. Although in Gecko, the iframe does turn solid white.)

arichiv: Good point, updated

Psychpsyo:
Edited version looks good to me.

arichiv:
@SebastianZ / @flackr willing to merge?

rebane2001 (approved):

SebastianZ:
The change seems reasonable to me, though I am not involved in this topic enough to provide proper input, nor am I one of the spec editors. So I refer to the editors @dirkschulze and @chrishtr.

Sebastian

filter-effects-1/Overview.bs
```diff
@@ -3546,6 +3546,12 @@ If any of the above rules are not followed, an attacker could infer information
 
 A timing attack is a method of obtaining information about content that is otherwise protected, based on studying the amount of time it takes for an operation to occur. If, for example, red pixels took longer to draw than green pixels, one might be able to reconstruct a rough image of the element being rendered, without ever having access to the content of the element. Security studies show that timing differences on arithmetic operations can be caused by the hardware architecture or compiler [[ArTD]].
 
+<h3 id="clickjacking-attack">Clickjacking Attacks</h3>
+
+User agents must not apply SVG reference filters (e.g., <a element>filter</a> or <a element>backdrop-filter</a>) to pixel data from cross-origin/restricted iframes or web plugins.
```
chrishtr:
```suggestion
User agents should not apply SVG reference filters (e.g., <a element>filter</a> or <a element>backdrop-filter</a>) to pixel data from cross-origin/restricted iframes or web plugins.
```

chrishtr (approved):

PR #13847, opened 2026-04-28
[cssom-1] font-family descriptor should not be serialized if not present, fix #13323

### LINKED ISSUE #13323: [cssom] Serialization of `@font-face` assumes `font-family` descriptor is always present
The [`@font-face` section of the CSS fonts spec](https://drafts.csswg.org/css-fonts-4/#font-face-rule) states that
>@&#8203;font-face rules require a font-family and src descriptor; if either of these are missing, the @&#8203;font-face rule must not be considered when performing the font matching algorithm.

This implies that `@font-face` rules without valid `font-family` descriptors are valid through parse time and just don't take part in font matching, this behavior is exhibited by both Chrome and Firefox.

The issue arises when [serializing the associated `CSSFontFaceRule`](https://drafts.csswg.org/cssom/#ref-for-cssfontfacerule) where step 3 assumes that the `font-family` descriptor is always present.

This issue is avoided in Chrome and Firefox by skipping steps 2-4 of the serialization algorithm if the `font-family` descriptor is not present.

svgeesus:
[cssom-1] Fixes #13323

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -1899,9 +1899,13 @@ To <dfn export>serialize a CSS rule</dfn>, perform one of the following in accor
   The result of concatenating the following:
   <ol>
    <li>The string "<code>@font-face {</code>", followed by a single SPACE (U+0020).</li>
-   <li>The string "<code>font-family:</code>", followed by a single SPACE (U+0020).</li>
-   <li>The result of performing <a>serialize a string</a> on the rule’s font family name.</li>
-   <li>The string "<code>;</code>", i.e., SEMICOLON (U+003B).
+   <li>If the '@font-face/font-family' descriptor is present:
+    <ol>
+      <li>The string "<code>font-family:</code>", followed by a single SPACE (U+0020).</li>
+      <li>The result of performing <a>serialize a string</a> on the rule’s font family name.</li>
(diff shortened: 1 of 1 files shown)


emilio (approved):
Looks good but it reads a bit weird to me? Maybe it should be reworded in terms of a temporary string and appending or something. But seems fine either way :)

cssom-1/Overview.bs
```diff
@@ -1899,9 +1899,13 @@ To <dfn export>serialize a CSS rule</dfn>, perform one of the following in accor
   The result of concatenating the following:
   <ol>
    <li>The string "<code>@font-face {</code>", followed by a single SPACE (U+0020).</li>
-   <li>The string "<code>font-family:</code>", followed by a single SPACE (U+0020).</li>
-   <li>The result of performing <a>serialize a string</a> on the rule’s font family name.</li>
-   <li>The string "<code>;</code>", i.e., SEMICOLON (U+003B).
+   <li>If the '@font-face/font-family' descriptor is present:
+    <ol>
+      <li>The string "<code>font-family:</code>", followed by a single SPACE (U+0020).</li>
```
AtkinsSJ, line 1904: I'm late, but I think the space added in step 1 also needs changing, to be added in step 2.1 instead. Otherwise, if there is no `font-family`, then step 3.1 below will mean there are two spaces between the `{` and the first descriptor name.

PR #13850, opened 2026-04-28
[css-gaps-1] Add spec text for assigning gap decorations to suppressed gaps for fragmentation #13754

### LINKED ISSUE #13754: [css-gaps-1] Dropping decorations for suppressed gaps
In https://github.com/w3c/csswg-drafts/issues/13663#issuecomment-4164597208 we resolved that, when varying gap decorations are assigned to a wrapping flex container, decorations are assigned to gaps in "linear" fashion, such that the sequence continues from one flex line to the next. An open question from that discussion is what to do with the decoration that would have appeared next to the last item in a flex line. Does that decoration get dropped from the sequence, or does it get assigned to the first gap in the following line?

One use case @fantasai brought up was a toolbar with varying decorations to separate items into groups. I constructed a straw example here: https://codepen.io/kevinbcmu/pen/ogzdvgE

Image

In my opinion, it's a little clunky to do this with gap decorations, because you have to keep your markup structure in sync with your decoration sequence. But I can't think of a better way to do it with the tools we currently have. I agree it's a case where dropping the decoration is what the author wants, and I haven't been able to come up with a plausible use case for not dropping it.

The other case where I think we'd want to drop a decoration is a gap suppressed due to fragmentation. Consider this example: https://codepen.io/kevinbcmu/pen/YPyYwBQ?editors=1100

Image

There's an auto repeater to set up dashed decorations for the intermediate rows, followed by a single value to set up solid decorations in the last row gap. If this example were fragmented on that last row gap, the author would not want the solid decoration pushed up to the preceding gap.

A third case worth considering is gaps suppressed due to collapsed tracks in grid. I've considered this a bit but haven't yet been able to construct a use case that points in one direction or the other.

oSamDavis:
This PR adds spec text clarifying the behavior of gap decorations when a gap is suppressed at a fragmentation break. Per #13754, the suppressed gap still consumes a decoration from the assignment pattern, even though no decoration is drawn there. This preserves a consistent association between gaps and decorations across fragmented and unfragmented contexts.

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -2351,6 +2351,8 @@ Changes since the <a href="https://www.w3.org/TR/2026/WD-css-gaps-1-20260227/">2
 			(<a href="https://github.com/w3c/csswg-drafts/issues/13648">Issue 13648</a>,
 			<a href="https://github.com/w3c/csswg-drafts/issues/13649">Issue 13649</a>)
 		<li>Updated with links to additional WPTs.
+		<li>Clarified that gaps suppressed due to fragmentation still consume gap decoration values.
+			(<a href="https://github.com/w3c/csswg-drafts/issues/13754">Issue 13754</a>)
 	</ul>
 
 <h3 id="changes-20250417">
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -1399,11 +1399,7 @@ Fragmentation</h3>
 	[[css-align-3#column-row-gap]] specifies that gaps disappear
 	where they coincide with a [=fragmentation break=].
 	Because there is no gap in such locations,
-	no [=gap decoration=] is drawn there.
-	However, the suppressed gap still consumes a decoration from the
-	[=assign gap decoration values|assignment pattern=], so that the association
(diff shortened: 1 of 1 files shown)


alisonmaher (approved):
LGTM - not sure if we want to add something to the edits section to note this as well?

kbabbitt (changes requested):
I would prefer to put this detail in the "Assigning gap decoration values to gaps" section instead, so that all of the information about assigning decorations is in a single place.

oSamDavis:
@kbabbitt just moved, let me know if that's sufficient

css-gaps-1/Overview.bs
```diff
@@ -1811,6 +1811,11 @@ Assigning gap decoration values to gaps</h3>
 	is applied to the next gutter (or set of [=collapsed gutters=]). Therefore, [=collapsed gutters=] are treated as a
 	single [=gap=] for decoration purposes.
 
+	A gap that is <a href="#fragmentation">suppressed</a> because it coincides with a [=fragmentation break=]
```
```suggestion
	A gap that is suppressed due to <a href="#fragmentation">fragmentation</a>
```

kbabbitt (approved):

Issue #13377, opened 2026-01-21
[mediaqueries] Effect of <meta name=color-scheme> on img and video sources


JoshTumath:
> [!NOTE]
> This was initially discussed in https://github.com/w3c/csswg-drafts/issues/10249#issuecomment-3779908771

@johannesodland has raised that, if the author sets `<meta name=color-scheme content=dark>`, it sets the document's _used color scheme_ to `dark`, but there is no way for it to affect which HTML `source` element is selected based on the _used color scheme_.

An author might assume in this example that the HTML parser will select `dark-mode-image.png` to download, but it won't necessarily:

```html
<!-- User has their color-scheme preference set to light mode. -->
<meta name="color-scheme" content="dark">

<picture>
  <source srcset="dark-mode-image.png" media="(prefers-color-scheme: dark)" />
  <img alt="" src="light-mode-image.png">
</picture>
```

There are two questions:

1. How could the UA choose a dark color scheme image to fetch based on the _page's supported color schemes_ set by `<meta name=color-scheme content=dark>`?
2. How could the UA choose a dark color scheme image to fetch based on the _used color scheme_ (which can be changed for all descendents in the DOM tree using the `color-scheme` CSS property)?

Both are difficult to solve because the HTML parser makes an immediate request for an image as soon as it identifies an image URL. When it encounters `source` elements in the `picture` element, it can only select a source using the information it has at parse time. At parse time, we do not know the used color scheme. The only thing we might know is the _page's supported color schemes_ set by the `<meta name="color-scheme">` tag.

Therefore, I think we are more concerned with solving Question 1. Both of these questions should be resolved in the HTML spec, but I think it's worth considering the CSSWG's thoughts, first.

#### Option 1a. propagate to `prefers-color-scheme` media query

Once the _page's supported color scheme_ is parsed, when the HTML parser encounters a `prefers-color-scheme` media query in a source, it evaluates it using the _page's supported color scheme_. So in this example, it loads the dark-mode-image.png.

```html
<meta name="color-scheme" content="dark">

<picture>
  <source srcset="dark-mode-image.png" media="(prefers-color-scheme: dark)" />
  <img alt="" src="light-mode-image.png">
</picture>
```

This solution doesn't introduce any new HTML syntax. Perhaps #7213 sets a precedent for this kind of behaviour?

#### Option 1b. propagate to a new media query `document-color-scheme`

[...]

Issue #7493, opened 2022-07-13
[css-mediaqueries] Should prefers-color-scheme in iframes be context-dependent?


emilio:
This is a follow-up for #7213.

It seems in that issue, we were in general agreement that doing this for images and maybe even same-origin `<iframe>`s would be ok, but @tabatkins and @smfr mentioned that cross-origin frames might not be ok, and I'm curious about the reasoning for that, since other similar alterations like https://github.com/w3c/csswg-drafts/issues/4772 have no same-origin restrictions. When I talked about this with the security folks at Mozilla, there didn't seem to be a particularly interesting attack vector here.

This would be useful both for consistency, but also because it would allow use cases like https://github.com/w3c/csswg-drafts/issues/7213#issuecomment-1144016642 to work. Otherwise, there's no way to have a transparent iframe without coordination with the embedder page, even if the embedded page could support both light and dark color schemes.

cc @smfr, @chrishtr, @lilles, @tabatkins

jtbandes:
@pxlcoder hi, just wondering if there's any update on this in Safari yet? (https://bugs.webkit.org/show_bug.cgi?id=284973)

Issue #7213, opened 2022-04-13
[css-mediaqueries] Should prefers-color-scheme in SVG images be context-dependent?


emilio:
Consider this test-case:

```html
<!doctype html>
<div style="color-scheme: light">
  <img>
</div>
<div style="color-scheme: dark">
  <img>
</div>
<script>
  for (let img of document.querySelectorAll("img")) {
    img.src = "data:image/svg+xml;base64," + btoa(`
      <svg width="32" height="32" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg">
        <style>
          :root { color: blue }
          @media (prefers-color-scheme: dark) {
            :root { color: purple }
          }
        </style>
        <rect fill="currentColor" width="32" height="32"/>
      </svg>
    `);
  }
</script>
```

Right now the rendering of those images match in all browsers, but one could argue that perhaps they should render differently (top blue, bottom purple)?

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1764354, which is what prompted this discussion.

cc @lilles @smfr

emilio:
cc #6988 which is somewhat related.

tabatkins:
Yeah, so basically you're asking if the used color scheme should automatically inherit into nested documents. That's a communication bit, which we try to avoid adding for cross-origin stuff, at least.

(#6988 would allow passing the color-scheme in purposely, if both the outer and inner page successfully handshaked.)

emilio:
Right. My reasoning for doing this for SVG images is that they cannot run script, etc, so it's a much more limited environment. But I guess you're right that they could trigger different image loads to exfiltrate that bit, potentially.

For the record, Gecko does inherit the color scheme into some frames (so that e.g., extension setting popups behave correctly if you have a dark Firefox theme but a light content theme). This is not exposed to regular `<iframe>`s on websites, though.

lilles:
The comment: https://github.com/w3c/csswg-drafts/issues/7213#issuecomment-1099501969 confused me a bit since @emilio's example affects the preferred color-scheme, not the used color-scheme or color-scheme property directly.

This is about changing the preferred color-scheme in the SVG and the SVG root color-scheme would not be affected in another way than through the preferred color-scheme, right?

emilio:
Yeah, that's right.

emilio:
[...]

PR #13857, opened 2026-04-29
[css-color-adjust-1] Rework color scheme stuff to reflect f2f resolutions

tabatkins:
Just posting this as a PR for review purposes for a few days; I'll merge it next week if there are no objections.

This PR applies the resolution from the recent F2F in #13377, making `(prefers-color-scheme)` reflect the effects of `<meta name=color-scheme>` and adding a (name TBD) `(ua-color-scheme)` MQ to actually reflect the user's preference. This makes `(prefers-color-scheme)` now consistent across top-level and embedded documents (previously, embedded documents used the embedding element's used color scheme instead of the user's preferred color scheme), and allows theme switchers a single source of control (tho the [`PreferenceManager` API](https://drafts.csswg.org/mediaqueries-5/#script-control-user-prefs) should generally be used instead).

It also finally applies the previous resolutions from #7493 and #7213, making all embedded documents use the embedding element's color scheme (also, usefully, matching existing browser behavior).

As part of this, I significantly rearranged and rewrote the color scheme portions of Color Adjust to use the new "page color scheme" and "element color scheme" concepts, and generally promote setting the page color scheme with `meta` rather than using the 'color-scheme' property. This is the main thing I'd like review on from anyone interested.


css-color-adjust-1/Overview.bs
```diff
@@ -124,30 +123,131 @@ Preferred Color Schemes {#preferred}
...
-	lets the author indicate appropriate [=color schemes=]
-	for UA-provided UI and colors in the page.
+	to ensure legibility,
+	generally either both foreground and background should be paired [=system colors=]
+	or both be manually specified colors.
+	[[WCAG22]]
+
+	The user's [=preferred color scheme=]
+	is then combined with the [=page's supported color schemes=]
+	to produce a <dfn export>page color scheme</dfn>.
```
JoshTumath, line 143: I think the 'element color scheme' and 'page color scheme' terms will really approve authors' understanding how how all of this works. 👍🏼 

I don't have a strong opinion on this, but should 'page color scheme' actually be 'document color scheme'?

css-color-adjust-1/Overview.bs
```diff
@@ -368,33 +369,125 @@ Opting Into a Preferred Color Scheme: the 'color-scheme' property {#color-scheme
...
+	a [=color scheme=] preference ([=light=] or [=dark=]),
+	or an <dfn for="color scheme" lt="overriding preference">overriding</dfn> [=color scheme=] preference.
+
+	A <dfn export>color scheme support</dfn>
+	is null,
+	or an ordered list of one or more supported [=color schemes=],
+	along with a flag indicating whether or not it <em>solely</em> supports the indicated [=color schemes=].
+
+	<div algorithm>
+		To determine the <dfn noexport>used color scheme</dfn>,
```
JoshTumath, line 405: Now that we have a concept of an 'element color scheme', I'm trying to understand if the 'used color scheme' is still needed. I guess this is now an internal function for determining both the 'element' and 'page' colour schemes, but it's not a term we'd need to refer to any more in documentation for authors?

svgeesus (commented):
I have read over the color-4 and color-5 changes carefully and these seem clear and correct to me; happy to see them merged.

I also looked over the color-adjust and mq-5 changes and they seem good but the document editors should comment so I am not adding approval at this stage.

svgeesus:
I don't understand the spec-prod failures however.

css-color-adjust-1/Overview.bs
tabatkins: Sure, I can use "document" instead. I suppose that is the more common term here.

tabatkins: Ah, the reason I called it "page color scheme" was just because we already had the term "page's supported color schemes", and the two should be consistent. Still, tho, no real preference on what we call these. If we change "page's supported color schemes", tho, we'll need to make an HTML edit too.

css-color-adjust-1/Overview.bs
tabatkins: The "used color scheme" term is just the algorithm itself, which resolves a preference against a support and returns a color scheme.

Correct that it's no longer a term that needs to be used in author-facing materials; it was previously double-used to also mean what's now called "element color scheme".

tabatkins:
> I also looked over the color-adjust and mq-5 changes and they seem good but the document editors should comment so I am not adding approval at this stage.

I'm those editors, so I'll approve it. ^_^

> I don't understand the spec-prod failures however.

As this PR hasn't been merged yet, the new terms don't yet exist in a crawlable form for SpecRef to find, and thus don't exist in Bikeshed's database. We'll need to ping those publications again after this is merged into the EDs and the data has had time to percolate.

css-color-5/Overview.bs
```diff
@@ -2837,35 +2833,35 @@ or any other color or monochrome output device which has been characterized.
...
-			    list-style-image: light-dark(
-					url("icons/deep-maroon-ball.png"),
-					url("icons/pale-yellow-star.png")
-				);
-			}
-		</pre>
-		</div>
+		<pre class="lang-css">
+		ul.fancy {
+		    list-style-image: light-dark(
```
tabatkins:
```suggestion
			list-style-image: light-dark(
```

css-color-adjust-1/Overview.bs
JoshTumath: Ah I see. Well as I say, I have no strong preference either way. I'll leave it to your discretion.

tabatkins: Same. I'll leave it as-is just to minimize my work. ^_^

PR #13861, opened 2026-04-29
[css-overscroll-1] Spec overscroll-behavior: chain

### LINKED ISSUE #13370: [css-overscroll-1] Add an overscroll-behavior value which disables local boundary default actions
Currently there are 3 [overscroll-behavior](https://drafts.csswg.org/css-overscroll-1/#propdef-overscroll-behavior) values:
* `auto` - which allows chaining, local and non-local boundary default actions
* `contain` - which allows local boundary default actions, but disallows chaining and non-local boundary default actions
* `none` - which disallows chaining, local and non-local boundary default actions

It would be nice to also have a value that
1. allows chaining and non-local boundary default actions
2. disallows local boundary default actions.

This would be a nice effect (or lack thereof) when bringing a side menu that shouldn't overscroll and reveal the background, but still allow scroll chaining:
e.g. https://overscroll-menu-vmpstr.yoyo.codes/

from Safari (but similar behavior on FF and Chrome Canary):

Image

I don't have any specific suggestions for the name, but maybe something like `forward` or `continue` as a somewhat contrived opposite of `contain`

@flackr

vmpstr:
Fixes #13370

### DIFF
--- a/css-overscroll-1/Overview.bs
+++ b/css-overscroll-1/Overview.bs
@@ -162,7 +162,7 @@ use <a>preventDefault</a> instead.
 
 <pre class=propdef>
 Name: overscroll-behavior
-Value: [ contain | none | auto ]{1,2}
+Value: [ contain | none | auto | chain ]{1,2}
 Initial: auto auto
 Applies to: <a>scroll container</a> elements
 Inherited: no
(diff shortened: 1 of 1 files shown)


vmpstr:
Errr, got more than I wanted into this, one sec.

vmpstr:
PTAL

flackr (approved):

Issue #13870, opened 2026-04-30
[css-grid-1] Clarify that rows can fragment


fantasai:
See https://github.com/web-platform-tests/interop/issues/1277

fantasai:
PR at https://github.com/w3c/csswg-drafts/pull/13871 CC @tabatkins @alisonmaher @dholbert

PR #13871, opened 2026-04-30
[css-grid-1][css-grid-2] Clarify that Grid can fragment within rows #13870

fantasai:


css-grid-1/Overview.bs
```diff
@@ -4659,15 +4659,32 @@ Fragmenting Grid Layout</h2>
...
 			between rows or columns (whichever is in the appropriate axis),
-			and <a href="https://www.w3.org/TR/css3-break/#end-block">Class C break opportunities</a> occur
+			and <a href="https://www.w3.org/TR/css-break/#end-block">Class C break opportunities</a> occur
 			between the first/last row (column) and the grid container's content edges.
-			[[!CSS3-BREAK]]
+			[[!CSS-BREAK-3]]
+
+		<li>
+			A grid track may be fragmented
+			by fragmenting all items in the track at that point
```
alisonmaher, line 4670: Items may not actually be fragmented as a result. For example, say you have a definite row size of 50px and all items in the row are 25px. If say your fragmentainer has a block size of 40px, the row fragments, but the items remain unfragmented. (Technically in this case, implementation-wise it would be the container breaking (since rows don't create fragments themselves), but we can think of it as the row being fragmented).

In another example, say your row is definitely sized to 25px but your item is 50px, the item will fragment, but your row will not (and the item will overflow the row as usual).

Also, if an item had a forced break before the fragmentainer boundary, it may have broken at an earlier point than the row would have.

So to summarize, item fragmentation is technically independent of the row. If a row is intrinsically sized, though, the breaking of items can impact the breaking of a row.

I'm thinking perhaps we'll need to word this in a way that is separated from items. Maybe something along the lines of "A grid track may fragment, similar to a grid item, if its block size does not fit within the fragmentainer boundary. Fragmentation breaks within a grid item..."

dholbert:
Thanks for writing this up! This looks reasonable to me (modulo @alisonmaher's notes).

(I appreciate that the text calls out that fragmenting an item can cause the item to grow, which can cause its row to grow, which causes neighboring items to grow if their block-end position is at the end of the row [due to stretch or end alignment]; and this can have downstream effects on the rest of the grid layout as well.)

PR #13882, opened 2026-05-03
[css-easing-2][editorial] Fix path syntax in css-easing-2/images/linear-easing-curve.svg

kbhomes:
Removes the commas from the path curve in this example image that were preventing rendering in Firefox. (It was rendering in Chrome fine, but that seems to be against spec (https://w3c.github.io/svgwg/svg2-draft/paths.html#PathDataBNF) which only allows whitespace between draw commands).

Marked this as editorial since it is only modifying an example image.


SebastianZ:
Thanks for the fix, @kbhomes!

Sebastian

PR #13884, opened 2026-05-04
[css-forms-1] Remove select styles

lukewarlow:
These have been upstreamed to HTML so no need to duplicate them here.


lukewarlow:
No need to duplicate them and some changes are happening here but not to the HTML spec version until later on.

annevk (approved):
Thank you!

PR #13885, opened 2026-05-04
Add `overlap-join` keyword to the css-gaps spec #13137

### LINKED ISSUE #13137: [css-gaps-1]: Change the initial value for insets property to be 0
In [issue 12848](https://github.com/w3c/csswg-drafts/issues/12848), we resolved to change the `outset` property to `inset`. This property controls endpoints of segments created as a result of the `rule-break` property. The initial value for the interior points is currently `-50%` allowing the gap decoration to bleed halfway into the gap in the crossing direction. See below:

```css

     column-rule-color: blue;
     column-rule-style: solid;
     column-rule-width: 5px;      

     row-rule-color: red;
     row-rule-style: solid;
     row-rule-width: 5px;

     rule-break: intersection;
```
Image

I am opening this issue because I believe this should be reconsidered and in favor of setting the initial value to `0px`. For one, just changing the `rule-break` property doesn't make it immediately clear what this property controls. Today, one would have to change the `inset` property to `0` to see the effect of changing the rule-break behavior:

Image

In addition, we're also [introducing an auto keyword](https://github.com/w3c/csswg-drafts/issues/13127) to become the default value for `rule break`. This will allow us have different default behaviors for different layout types with the goal of preserving reading order across these layout types. `rule-break: auto` alone is not sufficient to achieve this. In multi-col, to preserve reading order when using `column-wrap we'd want the column gap decoration not to bleed into the cross gap initially. See:

Image

oSamDavis:
Per the resolution taken in #13137, this PR adds spec text for the overlap-join keyword for gap decorations insets.


css-gaps-1/Overview.bs
```diff
@@ -983,14 +983,18 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+		Value: <<inset-value>>
 		Initial: ''0''
 		Applies to: <a>grid containers</a>, <a>flex containers</a>, <a>multicol containers</a>, and <a>grid lanes containers</a>
 		Inherited: no
 		Percentages: refer to the <a>crossing gap width</a>
 		Animation type: by computed value type
 	</pre>
 
+	<pre class='prod'>
+		<dfn>&lt;inset-value&gt;</dfn> = <<length-percentage>> | <a value for="<inset-value>">overlap-join</a>
```
oSamDavis: @kbabbitt  Had trouble linking a keyword in a prod block (see: inset-value definition). Ended up going for a tag; please let me know if there's a better way.

css-gaps-1/Overview.bs
```diff
@@ -1097,6 +1101,66 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+
+			<p class="note">
+				Note: At a [=junction segment endpoint=], this is equivalent to an inset of
+				''calc(-50% - |w| / 2)'',
+				where |w| is the [=used value=] of the crossing decoration's
+				'column-rule-width' or 'row-rule-width', whichever applies.
+			</p>
+		<li>
+			At a [=cap segment endpoint=], it computes to ''0''; the [=gap decoration=] remains flush
+			at the [=segment endpoint=], because there is no other segment to join with.
```
```suggestion
			At a [=cap segment endpoint=], [=overlap-join=] is treated as ''0''.
```

1. We can't determine at computed-value time whether we have a cap or a junction segment endpoint, and in fact the same computed value might apply to both cap and junction segment endpoints, so this has to be later than computed-value time.
2. Keeping the text as concise as possible reduces future maintenance overhead.

css-gaps-1/Overview.bs
```diff
@@ -1097,6 +1101,66 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+		produces neat joins between [=gap decorations=] that meet at the same [=gap junction=]:
+	</p>
+
+	<ul>
+		<li>
+			At a [=junction segment endpoint=], the [=gap decoration=] extends into the [=gap junction=]
+			by half the [=crossing gap width=], plus half the [=used value=] of the crossing decoration's
+			'column-rule-width' or 'row-rule-width' (whichever applies).
+
+			<p class="note">
```
kbabbitt: This note seems redundant with the normative text; I'd remove it.

css-gaps-1/Overview.bs
```diff
@@ -1097,6 +1101,66 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
 
+	<p>
+		The <dfn value for="<inset-value>">overlap-join</dfn> keyword
+		produces neat joins between [=gap decorations=] that meet at the same [=gap junction=]:
+	</p>
+
+	<ul>
+		<li>
+			At a [=junction segment endpoint=], the [=gap decoration=] extends into the [=gap junction=]
+			by half the [=crossing gap width=], plus half the [=used value=] of the crossing decoration's
```
```suggestion
			by half the [=crossing gap width=], plus half the [=used value=] of the crossing gap's
```

css-gaps-1/Overview.bs
```diff
@@ -1097,6 +1101,66 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+				rule-break: intersection;
+				rule-visibility-items: between;
+			}
+		</pre>
+		<figure>
+			<img alt="" src="images/example-1-overlap-join.png">
+		<figcaption>
+			With ''<inset-value>/overlap-join'', [=gap decorations=] at [=junction segment endpoints=]
+			extend into the [=gap junction=], joining adjacent segments.
+			At [=cap segment endpoints=], ''<inset-value>/overlap-join'' computes to ''0''.
```
```suggestion
			At [=cap segment endpoints=], ''<inset-value>/overlap-join'' is treated as ''0''.
```

css-gaps-1/Overview.bs
```diff
@@ -1097,6 +1101,66 @@ Adjusting gap decoration endpoints: The 'rule-inset' properties</h3>
...
+			}
+		</pre>
+		<figure>
+			<img alt="" src="images/example-2-overlap-join.png">
+		<figcaption>
+			In a "donut" grid layout, where items surround an empty space in the middle of the container,
+			''<inset-value>/overlap-join'' joins adjacent segments
+			at the [=junction segment endpoints=],
+			extending each [=gap decoration=] into the [=gap junction=]. At the [=cap segment endpoints=]
+			bordering the edge of the container and the empty interior region, ''<inset-value>/overlap-join'' computes to ''0''.
```
```suggestion
			bordering the edge of the container and the empty interior region, ''<inset-value>/overlap-join'' is treated as ''0''.
```

kbabbitt (approved):

PR #13887, opened 2026-05-05
[css-gaps-1] Clarify decoration pattern continuity across flex lines #13633

### LINKED ISSUE #13663: [css-gaps-1]:  Assigning multiple decoration values to flex containers
[CSS Gaps 1](https://www.w3.org/TR/css-gaps-1/) proposes that rule style, width and color properties accept [multiple values](https://www.w3.org/TR/css-gaps-1/#typedef-line-color-or-repeat), and that these properties also apply to flex containers. For flexbox, this raises the question of how multiple values should be assigned to gaps in the cross direction (i.e. gap between items).

One option is _**linear assignment**_. Here, each gap consumes one value in order. For example, given:
```css
column-rule-color: red, blue, green, yellow;
```
The first gap would use red, the second blue, the third green and so on, cycling as needed.
This model is simple, intuitive, and predictable.

Image

Another option is a _**per line reset**_. With this approach, the list of values will be reapplied for each flex line. It sort of comes in handy for cases where gaps may be aligned across different flex lines, and authors might want the same value applied consistently across lines. So rather than having:
Image

We have:
Image

Essentially, the core question here is: should decoration values apply to global cross‑axis gaps, or to per‑line (local) cross‑axis gaps?

My view is that the **_linear assignment_** model is more intuitive and straightforward for authors. Flexbox is fundamentally a 1D layout model, and assigning values globally across gaps better reflects that mental model. Resetting values per line risks introducing implicit 2D alignment that flexbox does not guarantee.

jav099:
Clarify in the `#assigning` section that gap decoration patterns are assigned sequentially across a flex container and do not restart at the beginning of each flex line.

### DIFF
--- a/css-gaps-1/Overview.bs
+++ b/css-gaps-1/Overview.bs
@@ -2030,6 +2030,9 @@ Assigning gap decoration values to gaps</h3>
 	is drawn there. This preserves a consistent association between gaps and decorations across
 	fragmented and unfragmented contexts.
 
+	Similarly, when a [=flex container=] wraps into multiple flex lines, the decoration pattern is assigned to gaps sequentially across the flexbox
+	and does not restart at the beginning of each flex line.
+
 	<div algorithm>
 		To <dfn>assign gap decoration values</dfn> to a list of |gaps| using a list of |values|:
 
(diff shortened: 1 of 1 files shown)


kbabbitt (approved):
Normative text LGTM. Two things before merging:

1. If I'm not mistaken this is for 13663 rather than 13754, correct?
2. Could you please add an entry to the Changes section for it?

Issue #12289, opened 2025-06-05
[css-overflow-4] Allow scrollable overflow to be clipped in off-axis


flackr:
When developers create scrolling containers that are only intended to scroll in one direction, they'll sometimes have content which slightly overflow in the other axis. Per the [css-overflow-3](https://www.w3.org/TR/css-overflow-3/#valdef-overflow-clip:~:text=The%20visible/clip%20values%20of%20overflow%20compute%20to%20auto/hidden%20(respectively)%20if%20one%20of%20overflow%2Dx%20or%20overflow%2Dy%20is%20neither%20visible%20nor%20clip.) spec:

> The [visible](https://www.w3.org/TR/css-overflow-3/#valdef-overflow-visible)/[clip](https://www.w3.org/TR/css-overflow-3/#valdef-overflow-clip) values of [overflow](https://www.w3.org/TR/css-overflow-3/#propdef-overflow) compute to [auto](https://www.w3.org/TR/css-overflow-3/#valdef-overflow-auto)/[hidden](https://www.w3.org/TR/css-overflow-3/#valdef-overflow-hidden) (respectively) if one of [overflow-x](https://www.w3.org/TR/css-overflow-3/#propdef-overflow-x) or [overflow-y](https://www.w3.org/TR/css-overflow-3/#propdef-overflow-y) is neither visible nor clip.

This seems to mean that even if you set the off axis overflow to `clip` it will still compute to `hidden`. While this is not a problem for user scrolling, `overflow: hidden` is still programmatically scrollable. Calls to scrollIntoView or anchor link scrolls can accidentally vertically scroll to your overflowing content. Worse yet, a user has no way to undo this after it happens.

This originally came up as a scroll-marker bug [crbug.com/422733086](https://issues.chromium.org/422733086) ([Demo](https://codepen.io/flackr/pen/vEOxoGJ)), however this problem can easily occur with existing content. E.g. consider this demo with "accidental" vertical scrolling when clicking on an item to scroll it into view: https://jsbin.com/voxohip/edit?html,css,js,output

Proposal: We should allow overflow `clip` in the off axis to make that axis not scrollable always at origin. E.g. `overflow: auto clip;` should allow creating a scroll container which *cannot* be scrolled vertically. We would still consider it to be a scroll container so long as one axis has a value of `hidden | scroll | auto`.

flackr:
@una FYI

PR #13903, opened 2026-05-08
Update mentions of 'scroll container' to refer to scrollable axises when appropriate. #12289

tabatkins:
WIP

Specs which mention scroll container:

* [x] css-2024: n/a
* [x] css-2025: n/a
* [x] css-2026: n/a
* [x] css-align-3: mixed references; some care about the scroll axis, others don't
* [x] css-anchor-position-1: always cares about the scroll axises
* [x] css-backgrounds-3: only background-attachment:local, doesn't care about axises
* [x] css-backgrounds-4: ditto
* [x] css-conditional-5: scroller CQs only establish CQ containers for their scrollable axises
* [x] css-display-3: doesn't care about axis
* [x] css-display-4: ditto
* [x] css-flexbox-1: minimum content size cares that the main axis is scrollable
* [x] css-flexbox-2: ditto
* [x] css-grid-1: n/a
* [x] css-grid-2: n/a
* [x] css-multicol-2: n/a
* [x] css-nav-1: a few bits care about whether you can scroll in a direction; these are now axis-aware
* [x] css-overflow-3: edits are already committed
* [x] css-overflow-4: scrollbar mentions, implicitly already about axises
* [x] css-overflow-5: TODO: hmm, unsure about scroll markers
* [x] css-overscroll-1: Only chain to scrollers that are scrollable in that axis.
* [x] css-position-3: stickypos cares about axises
* [x] css-position-4: n/a
* [x] css-scroll-anchoring-1: i don't *think* i need to do anything here? the algos look okay as they are, i think
* [x] css-scroll-snap-1: TODO: leaving alone until the WG decides if snapping on different boxes *should* happen or not
* [x] css-scroll-snap-2: same
* [x] css-scrollbars-1: n/a
* [x] css-sizing-4: auto min size, and min-intrinsic-size, care about axises, since they care about scrollability in that axis to reach something
* [x] css-ui-3: not touching, RECish
* [x] css-ui-4: resize doesn't care about axis
* [x] css-writing-modes-4: just a note mention
* [x] cssom-view-1: adjusted "potentially scrollable" to be axis aware, so scrollLeft/Top will care properly about scroll axis on body
* [x] scroll-animations-1: all the timelines care about axis


css-align-3/Overview.bs
```diff
@@ -888,8 +888,8 @@ Automatic Overflow Alignment Safety</h4>
 <h5 id=auto-safety-scroll>
 Content Distribution for Scroll Containers</h5>
 
-	The default [=overflow alignment=] behavior
-	for [=content distribution=] on [=scroll containers=] is ''unsafe''.
+	In a [=scrollable axis=] of a [=scroll container=],
```
flackr, line 891: What does this mean? I see that `overflow-position` is only a single value (i.e. single axis) property and other references have been updated to refer to the block axis but this one seems like it cares about both axes? Is the default `overflow-position` now per axis but can only be specified for both axes? Just trying to confirm what changes we need to make.

cssom-view-1/Overview.bs
```diff
@@ -1638,7 +1639,7 @@ The <dfn attribute for=Element>scrollTop</dfn> attribute, on getting, must retur
 1. If <var>window</var> is null, return zero and terminate these steps.
 1. If the element is the [=root element=] and <var>document</var> is in <a>quirks mode</a>, return zero and terminate these steps.
 1. If the element is the [=root element=] return the value of {{Window/scrollY}} on <var>window</var>.
-1. If the element is <a>the <code>body</code> element</a>, <var>document</var> is in <a>quirks mode</a>, and the element is not <a>potentially scrollable</a>, return the value of {{Window/scrollY}} on <var>window</var>.
+1. If the element is <a>the <code>body</code> element</a>, <var>document</var> is in <a>quirks mode</a>, and the element is not <a>potentially scrollable</a> in the y axis, return the value of {{Window/scrollY}} on <var>window</var>.
```
flackr: This might be fine, but it's very quirky 🥁  that body.scrollTop and body.scrollLeft may refer to different scroll containers in quirks mode. I wonder if this is necessary or if we could just consider whether it is a scroll container in either axis for both.

flackr (commented):
Overall this looks good, just wanted to get clarification on a couple things.

css-overscroll-1/Overview.bs
```diff
@@ -129,8 +129,8 @@ to the chaining rules applied to it.
 
 
 <dfn>Scroll boundary</dfn> refers to when the scroll position of a <a>scroll container</a> reaches
-the edge of the <a>scrollport</a>. If a scroll container has no potential to scroll, because it does
-not <a>overflow</a> in the direction of the scroll, the element is always considered to be at the
+the edge of the <a>scrollport</a>. If a [=scroll container=] has no potential to scroll, because it does
+not <a>overflow</a> in the direction of the scroll or the scroll does not happen in one of its [=scrollable axises=], the element is always considered to be at the
```
flackr: overscroll-behavior only applies to scroll containers. I suspect it should now only apply to the axes for which it is a scroll container. E.g. `overscroll-behavior-x: contain;` does not normally contain scrolling on `overflow: clip`, it should probably not now contain scrolling for `overflow-x: clip; overflow-y: hidden;`

css-align-3/Overview.bs
tabatkins: You can do content-distribution in either axis via `justify-content` and `align-content`, and specify overflow safety in either axis. The existing spec gives different default behavior based on whether the container is scrollable or not (scrollables default to 'unsafe' because you can just scroll to see the overflow, generally). Now, tho, single-axis scroll containers only want to apply that default in their scrollable axis.

(The other updates are block-only because they're talking about things like baselines.)

cssom-view-1/Overview.bs
tabatkins: I'm fine with either. This is indeed just a quirks mode thing, so probably not important to care about. I'll go ahead and revert it.

css-overscroll-1/Overview.bs
tabatkins: :+1:

cssom-view-1/Overview.bs
tabatkins: I suppose a followup question is what to do with the *setters* for scrollTop/Left when the body isn't a scroller in that axis.

flackr: Presumably the same as trying to scroll any non-scrollable element. The setters and methods exist, but don't do any scrolling.

css-align-3/Overview.bs
flackr: Thanks, this makes sense.

css-conditional-5/Overview.bs
```diff
@@ -1526,32 +1530,37 @@ Style Container Features</h3>
...
+	that each query a single feature of the [=query container=].
+	The syntax of a <dfn><<scroll-state-feature>></dfn> is the same as for a [=media feature=]:
+	a feature name, a comparator, and a value.
 
-	[=Scroll-state features=] can either match state of the scroller itself,
-	or an element that is affected by the scroll position of an ancestor
-	[=scroll container's=] [=scrollport=]. An example of the former is the
-	''scrollable'' feature, ''snapped'' the latter.
+	[=Scroll-state features=] can either match state of the scroller itself
+	(such as the ''scollable'' feature),
```
freedebreuil: nit: Should this be ''scrollable'' instead of ''scollable''?

css-conditional-5/Overview.bs
```diff
@@ -612,7 +612,11 @@ Creating Query Containers: the 'container-type' property</h3>
 			and establishes an [=independent formatting context=].
 		<dt><dfn>scroll-state</dfn>
 		<dd>
-			Establishes a [=query container=] for [=container scroll-state queries=]
+			Establishes a [=query container=] for [=container scroll-state queries=].
+			For queries that match the state of the scroller itself
```
freedebreuil, line 631: @tabatkins Just a quick question about the `scroll-state` change.

Is the implication of this that [if we do a query that checks for a vertical value](https://jsbin.com/duhavec/edit?html,output):
```
@container scroll-state(scrollable:bottom) 
```

Would we expect that to "skip" parents which are not scroll containers on the Y axis?
```
.inner {
  container-type: scroll-state;
  overflow: scroll clip;
}
```
Chatting with @flackr we thought this would be the intended behavior.

With the proposed behavior, does that mean applying `container-type: scroll-state` to an element without `overflow` specified would now have no effect? (for `scrollable` or `scrolled`)
```
.inner {
  container-type: scroll-state;
  overflow: scroll; /* this line was not required before to match */
}
```

Also, @flackr mentioned it could be worth opening a csswg issue to confirm if this is the only container query that is dependent on other styles.

Thanks!

css-overscroll-1/Overview.bs
```diff
@@ -116,29 +116,43 @@ This specification does not mandate if and how scroll chaining or overscroll aff
...
+as the document's <a>scrollingElement</a>,
+both regarding placement in the scroll chain
+as well as adhering to the chaining rules applied to it.
+
+<dfn>Scroll boundary</dfn> refers to when the scroll position of a <a>scroll container</a>
+reaches the edge of the <a>scrollport</a>.
+If a [=scroll container=] has no potential to scroll,
+because it does not <a>overflow</a> in the direction of the scroll,
+the element is always considered to be at the scroll boundary.
+If a [=scroll container=] is not scrollable in the given axis,
```
flackr: I'm a bit worried that "scrollable" might be interpreted as having overflow, rather than being determined purely by the overflow style in that axis. Could we make this more clear? It seems like in other specs we've had a definition that clearly explains this. Maybe we should refer to the definition of scroll container in an axis?

css-conditional-5/Overview.bs
tabatkins: Added a note clarifying the behavior a bit.

I'll open a WG issue about verifying that having a CQ be a container dependent on other styles is okay; I think it should be.

css-conditional-5/Overview.bs
tabatkins: Yup, fixed.

css-overscroll-1/Overview.bs
tabatkins: Good catch, rewrote that sentence to be clearer that it's about scrollable axises.

PR #13910, opened 2026-05-11
[css-viewport-1] Make zoom interpolate by computed value type

### LINKED ISSUE #10872: [css-viewport] Is zoom supposed to be non animatable?
The [zoom property](https://drafts.csswg.org/css-viewport-1/#zoom-property) is defined to be non-animatable. According to [Mozilla bug 1917602](https://bugzilla.mozilla.org/show_bug.cgi?id=1917602#c0) Chrome makes this property discretely animatable, Safari makes it non-animatable, and Gecko does smooth animation (i.e. animates as a number/percent).

We'll fix Gecko, but should we be matching Chrome's behavior or is there a good reason for making zoom non-animatable? (For reference, we normally only make properties non-animatable when there's a strong reason like `transition-*` and `animation-*` properties where making them animatable would introduce obvious cyclic dependencies.)

yisibl:
See: https://github.com/w3c/csswg-drafts/issues/10872#issuecomment-2628543182

Fixes https://github.com/w3c/csswg-drafts/issues/10872

### DIFF
--- a/css-viewport-1/Overview.bs
+++ b/css-viewport-1/Overview.bs
@@ -427,7 +427,7 @@ Note: 'zoom' does not affect or prevent 'transform' scaling.
 	Percentages: Converted to <<number>>
 	Media: visual
 	Computed value: as specified, but with <<percentage>> converted to the equivalent <<number>>
-	Animation type: not animatable
+	Animation type: by computed value type
 </pre>
 
 The values of this property have the following meanings:


yisibl:
@lilles @flackr @fantasai PTAL

miketaylr:
@frivoal @emilio mind taking a look (AIUI, you're the current editors of css-viewport-1)?

emilio (approved):
> [yisibl](https://github.com/yisibl) marked as non substantive for IPR from ash-nazg.

Seems fine but not sure if it's really "non-substantive"

miketaylr:
@yisibl can you take care of the IPR issue? thanks.

yisibl:
> @yisibl can you take care of the IPR issue? thanks.

I'm not entirely sure about the specifics here, but if we don't tag it this way, the Robot won't pass.
I've done the same in other merged PRs, so I don't think there should be any issues.

miketaylr:
> I'm not entirely sure about the specifics here, but if we don't tag it this way, the Robot won't pass.

@yisibl - that is fine for non-substantive changes, but this change is likely considered _substantive_  as it changes the conformance requirements for implementations to require this property to be animatable. See https://www.w3.org/policies/process/#class-3 as well as https://www.w3.org/groups/wg/css/ipr/.

@svgeesus do you have any advice on how to unblock this (I believe you're w3c staff contact for CSSWG, forgive me if I got that wrong)? I see "Staff Contacts will provide instructions for recording the non-participant licensing commitment before the contribution can be accepted" in the IPR link.

Maybe it's just easier for one of the editors to land this (very, very simple) change?

PR #13915, opened 2026-05-12
[cssom-view-1] Clarify the relation of visual viewport and layout viewport with scale transform

shubhamg13:
Makes clear that visual viewport and layout viewport are (inversely) proportional.

### DIFF
--- a/css-viewport-1/Overview.bs
+++ b/css-viewport-1/Overview.bs
@@ -106,9 +106,17 @@ difference between the [=initial viewport=] and the [=layout viewport=].
 		This is the viewport you get after processing the viewport
 		<code>&lt;meta&gt;</code> tag.
 	</dd>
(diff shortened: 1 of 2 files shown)


css-viewport-1/Overview.bs
```diff
@@ -106,9 +106,17 @@ difference between the [=initial viewport=] and the [=layout viewport=].
 		This is the viewport you get after processing the viewport
 		<code>&lt;meta&gt;</code> tag.
 	</dd>
+	<dt><dfn>visual viewport</dfn></dt>
+	<dd>
+		The is the portion of the layout viewport that is currently visible.
```
emilio, line 111: nit: "The is" seems mispelled?

But in general, the visual viewport is not only a scale but also a translation right?

PR #13918, opened 2026-05-13
[css-inline-3] Percentage line-height should not be scaled by text-fit

tkent-google:
[css-inline-3] Percentage line-height should not be scaled by text-fit

`line-height: <percentage>` is treated as a fixed value. See https://drafts.csswg.org/css-inline/#issue-af10b7fe .
So it's difficult to scale it by `text-fit` property.

This reverts a part of https://github.com/w3c/csswg-drafts/pull/13616.

https://github.com/w3c/csswg-drafts/issues/12888


tkent-google:
@kojiishi would you review this please?

kojiishi (approved):
LGTM

Issue #2528, opened 2018-04-11
Feature for making text always fit the width of its parent


tobireif:
This thread shows that it's a widely required feature:
https://twitter.com/sindresorhus/status/979363460411609091

Example of a workaround: Open https://tobireif.com/ and (eg using dev tools responsive mode) resize the page down to 250px width while watching the text "Welcome".

litherum:
This requires performing layout in a loop, which we generally have avoided. Requiring a round-trip through JS is valuable because authors are more likely to realize it has a large perf cost

tobireif:
In the JS at https://tobireif.com/ I perform two passes - that's plenty for a good-enough result, and it doesn't impact perf in any noticeable way (the text-fitting is only done once in addition to the first main run). That could be a great option for browser implementers as well, and it shows that supporting such a CSS feature is very feasible.

timothyis:
If this were a feature, I think it'd be best if it was a CSS function. (similar to `calc` or `minmax`)

Something like `font-size: fit(8px, 48px);` where `8px` is the minimum font-size and `48px` is the maximum font-size.

I think using a function, other than being useful for minimum and maximum sizes, relays the gravity of using the feature since surely it'll have _some_ performance issues in extreme cases.

I'd love to see this in CSS!

tobireif:
Great suggestions!

A lower limit and an upper limit both make sense.

Instead of `font-size: fit(8px, 48px)` it might be better to name it eg `font-size: fit-width(8px, 48px)`.

SergeyMalkin:
Changing font-stretch, especially using variable fonts, is another way to fit text into parent. Or compression during justification . So if there is a css property that instructs layout engine to fit, it should allow different methods and so likely be separate from font-size.

And this kind of functionality may not only be on line operation.  It may be useful for more advanced functionality, like optimal paragraph layout, line clamping, or simple ellipsis.

tobireif:
> Changing font-stretch, especially using variable fonts, is another way to fit text into parent. Or compression during justification . So if there is a css property that instructs layout engine to fit, it should allow different methods and so likely be separate from font-size.

True! (also eg letter-spacing)

> And this kind of functionality may not only be on line operation. It may be useful for more advanced functionality, like optimal paragraph layout, line clamping, or simple ellipsis.

[...]

PR #13965, opened 2026-05-26
[css-text-4] `text-fit` should inherit

tkent-google:
It should be inherited like other text-* properties in order to apply it to anonymous blocks.

This addresses the last paragraph of https://github.com/w3c/csswg-drafts/issues/2528#issuecomment-3636441522 .


tkent-google:
@kojiishi would you review this please?

This is consistent with other text-* properties, and is necessary for `<div style="columns:2; text-fit:grow">`.

kojiishi (approved):
lgtm

PR #14001, opened 2026-06-02
[css-navigation-1] [editorial] Fix typo in selector

bramus:
It’s `:active-navigation()`, not `::active-navigation()`


dbaron:
Tab beat me -- I was going to press the button but then the buttons moved around!

PR #14012, opened 2026-06-04
[css-navigation-1] Add examples

bramus:
This PR adds many more examples to the `css-navigation-1` spec. To help readers grasp the examples, it also introduces a description _(with examples)_ for URLPattern’s [pattern string](https://urlpattern.spec.whatwg.org/#pattern-string).

Were needed, existing content was moved around to allow a more continuous read.

### DIFF
--- a/css-navigation-1/Overview.bs
+++ b/css-navigation-1/Overview.bs
@@ -385,69 +385,6 @@ functional pseudo-class
 that matches link elements that link to a certain URL
 that is related to a navigation that is currently active.
 
-<div class="example">
-
-A an example of the '':active-navigation()'' pseudo-class
-is this example which creates a view transition between
-a item in a list that contains a link (in this document)
-and the details page for that link (in a different document).
-This transition works even when the navigation is a back/forward navigation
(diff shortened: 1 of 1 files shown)


css-navigation-1/Overview.bs
```diff
@@ -270,6 +348,28 @@ which matches any links that link to the site's homepage:
 }
 </pre>
 
+Because there is no dynamic part in the homepage URL,
+you can also pass in a <<url>> directly:
```
noamr: Maybe it's good to mention that this won't match `/#scroll-to-here` or `/?utm_id=something`. So it's usually a best practice to have something like:

```css
@route --home {
  pathname: "/";
  base-url: document;
}
```

noamr (approved):
Great, thanks a lot! See one comment

css-navigation-1/Overview.bs
bramus: Fixed in 064e09a.

noamr (approved):

PR #14019, opened 2026-06-05
[css-fonts-5][editorial] Add more meta text scale examples

davidsgrogan:
Follow up to offline June 1 convo.

### DIFF
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -168,10 +168,17 @@ Further, when the value of the [=text-scale=] content attribute is
 
 Note:
 Authors are expected to use
-<code>&lt;meta name="text-scale" content="scale"&gt;</code> in stylesheets so that the ''font-size/medium'' font size will reflect a combination of the user's font preferences, whether those are specified at the OS level or the UA level. The author will then be able to use ''rem'' throughout the page to honor the user's font preferences.
+<code>&lt;meta name="text-scale" content="scale"&gt;</code> in stylesheets so that the ''font-size/medium'' font size will reflect the user's font preferences, whether those are specified at the OS level or the UA level.
+The author will then be able to use ''rem'' and ''em'' throughout the page to honor the user's font preferences.
+
+Note:
+This requires care to do correctly, given the very large font sizes some low-vision users need.
+Consider the extreme case of an iPhone Mini set to iOS's largest font size of 53px.
+Only ~12-15 characters fit on a line.
+
 
 <div class="example" id="ex-scaled-page">
-	If authors don't alter the '':root'' font size, content sized with ''rem'' units will be relative to the preferred text scale
+	Assuming a page hasn't altered the '':root'' font size, content sized with ''rem'' units will be relative to the preferred text scale
 	<pre class="lang-html">
 &lt;!DOCTYPE html>
 &lt;html> &lt;!-- leave this element's font-size as the default -->
@@ -192,6 +199,50 @@ Authors are expected to use
 		&lt;/div>
 	&lt;/body>
 &lt;/html>
+	</pre>
+</div>
+
+<div class="example" id="ex-non-linear">
+	Header text starts larger than body text, so pages should probably grow headers at a slower rate.
+	<pre class="lang-html">
+		&lt;meta name="text-scale" content="scale" />
+		&lt;style>
+		.h1 {
+			--dampen-factor: calc(0.5 + 0.5 / env(preferred-text-scale));
+			font-size: calc(2rem * var(--dampen-factor));
+			/* Resulting font sizes at various text scale factors:
+			      h1      body
+			1x    32px    16px
+			1.5x  40px    24px
+			2x    48px    32px
+			3x    64px    48px */
+		}
+		&lt;/style>
+	</pre>
+</div>
+
+<div class="example" id="ex-inverse">
+	An extreme version of the above, some text content that starts larger than body text can and should become smaller than body text as the user's font size increases.
+	In this floating button example, the button text is larger than the body text at 1x but gets smaller at ~2.5x font scale.
+	<pre class="lang-html">
+		&lt;meta name="text-scale" content="scale" />
+		&lt;style>
+		.floating-action-button {
+			position: fixed;
+			top: 30px;
+			right: 30px;
+			/* Grow slightly slower than h1 example: */
+			--dampen-factor: calc(0.4 + 0.6 / env(preferred-text-scale));
+			font-size: calc(1.5rem * var(--dampen-factor));
+			/* Resulting font sizes at various text scale factors:
+			      button    body
+			1x    24px      16px
+			1.5x  28.8px    24px
+			2x    33.6px    32px
+			3x    43.2px    48px    &lt;-- inversion */
+		}
+		&lt;/style>
+		&lt;button class="floating-action-button">➕Add&lt;/button>
 	</pre>
 </div>
 
--- a/css-fonts-5/Overview.bs
+++ b/css-fonts-5/Overview.bs
@@ -208,7 +208,7 @@ Only ~12-15 characters fit on a line.
 		&lt;meta name="text-scale" content="scale" />
 		&lt;style>
 		.h1 {
-			--dampen-factor: calc(0.5 + 0.5 / env(preferred-text-scale));
+			--dampen-factor: calc(0.5 + (0.5 / env(preferred-text-scale)));
 			font-size: calc(2rem * var(--dampen-factor));
 			/* Resulting font sizes at various text scale factors:
 			      h1      body


css-fonts-5/Overview.bs
```diff
@@ -195,6 +202,50 @@ Authors are expected to use
 	</pre>
 </div>
 
+<div class="example" id="ex-non-linear">
+	Header text starts larger than body text, so pages should probably grow headers at a slower rate.
+	<pre class="lang-html">
+		&lt;meta name="text-scale" content="scale" />
+		&lt;style>
+		.h1 {
+			--dampen-factor: calc(0.5 + 0.5 / env(preferred-text-scale));
```
progers, line 211: What do you think of adding a comment and a little content, then showing an example of the content at 1.x, 2x, and 3.5x? This will encourage others to test large scales.
```
<meta name="viewport" content="width=device-width">
<meta name="text-scale" content="scale">
<style>
h1 {
  /*
    Multiply the 2em baseline by a dampening factor (between 1.0 to 0.5) to
    ensure headers scale up, but without getting too large. Here are the
    resulting font sizes at various text scale factors:
            h1      body
      1x    32px    16px
      1.5x  40px    24px
      2x    48px    32px
      3x    64px    48px
      3.5x  72px    56px
  */
  --dampen-factor: calc(0.5 + 0.5 / env(preferred-text-scale));
  font-size: calc(2em * var(--dampen-factor));
}
</style>
<h1>Header</h1>
This is some body text. The header grows slower than the body text.
```

Image

davidsgrogan: Yeah, that seems good, as long as @fantasai thinks it's appropriate level of detail for the spec. (We could put it in MDN or some other article/documentation.)

Also, note that you can see this PR at https://davidsgrogan.github.io/css-fonts-5/Overview.html#scale-keyword

css-fonts-5/Overview.bs
```diff
@@ -168,10 +168,17 @@ Further, when the value of the [=text-scale=] content attribute is
...
+
+Note:
+This requires care to do correctly, given the very large font sizes some low-vision users need.
+Consider the extreme case of an iPhone Mini set to iOS's largest font size of 53px.
+Only ~12-15 characters fit on a line.
+
 
 <div class="example" id="ex-scaled-page">
-	If authors don't alter the '':root'' font size, content sized with ''rem'' units will be relative to the preferred text scale
+	Assuming a page hasn't altered the '':root'' font size, content sized with ''rem'' units will be relative to the preferred text scale
```
JoshTumath, line 181: Might be better to say, "Assuming a page hasn't altered the `:root` font size with an absolute length unit like `px`"

css-fonts-5/Overview.bs
```diff
@@ -168,10 +168,17 @@ Further, when the value of the [=text-scale=] content attribute is
 
 Note:
 Authors are expected to use
-<code>&lt;meta name="text-scale" content="scale"&gt;</code> in stylesheets so that the ''font-size/medium'' font size will reflect a combination of the user's font preferences, whether those are specified at the OS level or the UA level. The author will then be able to use ''rem'' throughout the page to honor the user's font preferences.
+<code>&lt;meta name="text-scale" content="scale"&gt;</code> in stylesheets so that the ''font-size/medium'' font size will reflect the user's font preferences, whether those are specified at the OS level or the UA level.
+The author will then be able to use ''rem'' and ''em'' throughout the page to honor the user's font preferences.
+
+Note:
```
JoshTumath, line 174: Bikeshed has put this paragraph in a separate note box, but this seems like a continuation of the previous one. Do they need to be in the same note box?

css-fonts-5/Overview.bs
```diff
@@ -195,6 +202,50 @@ Authors are expected to use
 	</pre>
 </div>
 
+<div class="example" id="ex-non-linear">
+	Header text starts larger than body text, so pages should probably grow headers at a slower rate.
```
JoshTumath, line 206: I think this intro needs more context. How about this:

> If an element such as a heading already has large text, it is likely already large enough for a user who needs a high text scale to read. If the large text scales at the same rate as the body text, it could become too large to comfortably fit in the user's viewport.
> 
> Therefore, authors should consider using a dampening factor to reduce the rate that large text scales.

css-fonts-5/Overview.bs
JoshTumath, line 211: Might be good to use parentheses around the division so that people don't get confused by BIDMAS/PEMDAS. (Because I did. 😅 I was wondering for ages why it's not just `1 / env(preferred-text-scale)`.)
```suggestion
			--dampen-factor: calc(0.5 + (0.5 / env(preferred-text-scale)));
```

css-fonts-5/Overview.bs
```diff
@@ -168,10 +168,17 @@ Further, when the value of the [=text-scale=] content attribute is
 
 Note:
 Authors are expected to use
-<code>&lt;meta name="text-scale" content="scale"&gt;</code> in stylesheets so that the ''font-size/medium'' font size will reflect a combination of the user's font preferences, whether those are specified at the OS level or the UA level. The author will then be able to use ''rem'' throughout the page to honor the user's font preferences.
+<code>&lt;meta name="text-scale" content="scale"&gt;</code> in stylesheets so that the ''font-size/medium'' font size will reflect the user's font preferences, whether those are specified at the OS level or the UA level.
```
JoshTumath, line 171: I think this needs stronger wording so that authors know 'you must only use this if you have tested your website with scaled text'.

JoshTumath (commented):
These examples are really helpful.

css-fonts-5/Overview.bs
fantasai: It's appropriate for a spec.

fantasai: Our specs really should have _more_ illustrations and examples of good practices, not less. :)

css-fonts-5/Overview.bs
davidsgrogan: I wasn't sure either way. So if you think they should go together, they should go together.

css-fonts-5/Overview.bs
davidsgrogan: Done. Taken verbatim.

css-fonts-5/Overview.bs
davidsgrogan: Nice. Taken verbatim.

css-fonts-5/Overview.bs
davidsgrogan: Added.

css-fonts-5/Overview.bs
```diff
@@ -195,6 +201,78 @@ Authors are expected to use
 	</pre>
 </div>
 
+<div class="example" id="ex-non-linear">
+	If an element such as a heading already has large text, it is likely already large enough for a user who needs a high text scale to read. If the large text scales at the same rate as the body text, it could become too large to comfortably fit in the user's viewport.
+
+	Therefore, authors should consider using a dampening formula to reduce the rate that large text scales.
+	There are various mathematical approaches to achieve this; one possible formula is demonstrated below.
```
davidsgrogan, line 208: @JoshTumath This formula is different from yours at https://slidr.io/JoshTumath/let-s-fix-the-web-s-text-size#75 -- Does yours have an advantage over this one other than it's in a `@function`? If so, I will switch to yours. If not, I will leave this as is, or put mine in a `@function` also, or put yours after this example so that both are here?

davidsgrogan:
@progers @JoshTumath I finally updated this PR with y'all's suggestions. Please re-review!

https://davidsgrogan.github.io/css-fonts-5/Overview.html#scale-keyword for the updated text (well, at the moment it still serving the previous version, but I hope it updates soon)

css-fonts-5/Overview.bs
JoshTumath: I'm not sure really, both seem good. If I have time later today, I'll test them both out. But yours is nice because it's simpler, but maybe harder to specify a scaling factor because you have to adjust two numbers?

I'd avoid the `@function`. One, you don't want to have a dependency on the mixins and functions spec; and two, I found at the conference where I mentioned it, people seemed to get confused and think they could _only_ use the formula in an `@function`.

JoshTumath (commented):
Nice one!! Thank you.

(This is making me realise again how much we need to look into a way to automate the damping factor somehow. Maybe we could introduce a new property for it. I'll have a think...)

JoshTumath (approved):

davidsgrogan:
> (This is making me realise again how much we need to look into a way to automate the damping factor somehow. Maybe we could introduce a new property for it. I'll have a think...)

Yeah, I considered how lame it is for every author to have to redefine h1->h6 with a dampening factor. For that, I _think_ we can "just" put a dampening factor in the UA style sheet? E.g. change `font-size: 2em;` to include a calc expression at  https://crsrc.org/c/third_party/blink/renderer/core/html/resources/html.css;l=189;drc=b942aaa95ff860964c4d5cd8ec2a6bfaeb00dd74 ?

Are there non h1->h6 scenarios where the author has to manually dampen? We _could_ add an attribute or value to the meta tag that specifies a few parameters that would make the UA do it for any font size above X, where X is part of these new inputs? Just spitballing here.

css-fonts-5/Overview.bs
davidsgrogan: Oh, great intel that people got confused by the `@function`. (It had attraction for me precisely because it would get rid of the two-number-input problem.)

davidsgrogan:
Now that Josh has approved, @svgeesus or @fantasai , care to review? This is all non-normative for engines, FWIW.

PR #14043, opened 2026-06-12
[web-animations-1][animation-triggers-1] Move AnimationTrigger details

DavMila:
This moves the details of AnimationTrigger from web-animations-2 to the animation-triggers-1 spec.

This is mostly a copy-paste of the content from web-animations-2, with
a few adjustments:

- Drops the Animation frame loop update portion
- Drops the Animation mixin interface update portions
- Drops the KeyframeAnimationOptions 'trigger' addition
(these were structured as ammendments to web-animations-2 but are no
longer relevant anyway.)
- Adjusts the level of some headings so bikeshed is happy.


ydaniv:
Thanks @DavMila! I have local WIP on this, some of it conflicts. Would you like to merge these somehow?

ydaniv:
Mostly I'm concerned with trying to keep the Trigger definition generic, and describe how it interacts specifically with animations.

DavMila:
> Mostly I'm concerned with trying to keep the Trigger definition generic, and describe how it interacts specifically with animations.

Yeah keeping the trigger definition generic makes sense. I think I am trying to do the same thing as you but in 2 phases: move the definition over, then modify the definition. This PR just moves it, this other [PR](https://github.com/w3c/csswg-drafts/pull/14044) modifies the definition.

DavMila:
> Would you like to merge these somehow?

Realizing I didn't respond to this sorry. Um, yeah we should probably be in sync somehow. This particular PR just doesn't really change any of the content, though. Would you have any thoughts on the [PR](https://github.com/w3c/csswg-drafts/pull/14044) which is mostly making the `AnimationTrigger` generic?

ydaniv:
> > Would you like to merge these somehow?
> 
> Realizing I didn't respond to this sorry. Um, yeah we should probably be in sync somehow. This particular PR just doesn't really change any of the content, though.

OK, I'll respond on the second one.

> Would you have any thoughts on the https://github.com/w3c/csswg-drafts/pull/14044 which is mostly making the AnimationTrigger generic?

Yes, but I guess I should add it there. Thanks!

PR #14044, opened 2026-06-12
[animation-triggers-1] Update AnimationTrigger specification

DavMila:
This Updates the details of the AnimationTrigger interface and algorithms and removed TimelineTrigger-specific details which will be updated separately.


animation-triggers-1/Overview.bs
```diff
@@ -106,7 +110,126 @@ urlPrefix: https://www.w3.org/TR/web-animations-1/; type: dfn
...
+## Overview ## {#triggers-overview}
+
+An <dfn export>animation trigger</dfn> can be used to control the playback
+of [=animations=] associated with the trigger.
+
+Each [=animation trigger=] maintains the following fields internally:
+
+1. an <dfn lt="animation trigger active" export>active</dfn> flag, which is initially false,
+
+1. a <dfn lt="animation trigger invokes deactivation" export>invokes deactivation</dfn> flag,
```
DavMila, line 125: FYI @flackr:

The idea with the `invokes deactivation` flag is that EventTriggers will, when initialized, set this flag to true only if configured to have deactivation events (and leave false otherwise). TimelineTriggers always set this flag to true as they inherently have a deactivation "event".

animation-triggers-1/Overview.bs
```diff
@@ -111,223 +115,121 @@ urlPrefix: https://www.w3.org/TR/web-animations-1/; type: dfn
...
+Each [=animation trigger=] associates each [=animation=] in its
+[=animation trigger animation set|animations set=] with two <<animation-action>>s. One action is
+associated with [=invoke activation|activation=] and the other is associated with
+[=invoke deactivation|deactivation=].
 
-Issue: Do we need a formal resolution on the spec of the idle state?
+## Invoking Activation & Deactivation ## {#animation-trigger-activation-deactivation}
 
-## Animation Trigger Behavior ## {#trigger-behaviors}
+When asked to <dfn lt="invoke activation">invoke activation</dfn> given a [=trigger=] <var>trigger</var>,
```
DavMila, line 138: FYI @tabatkins: I realize we'd rather not have algorithms in spec but I figured these algorithms are short and not overly complicated... so perhaps they make are still a good way to spec the attributes/behavior common to different trigger types.

ydaniv:
@DavMila in my local changes I'm changing `Animation Trigger` into simply `Trigger`.
My intent is to a generic spec for Triggers, and then apply that to usage with animations.
Should I push what I have now as a separate PR so we can compare/discuss how to converge?

DavMila:
> @DavMila in my local changes I'm changing `Animation Trigger` into simply `Trigger`. My intent is to a generic spec for Triggers, and then apply that to usage with animations. Should I push what I have now as a separate PR so we can compare/discuss how to converge?

Okay, yeah please push what you have and I'll be happy to work off that. I think we should keep using `Animation Trigger` for now since animations are really the only use cases. I think our reasoning was similar when we chose to name the spec animation-triggers-1 instead of just css-triggers. If we do need the more generic `Trigger`, `Animation Trigger` could simply "inherit" from that if necessary I think.

ydaniv:
@DavMila see my PR on top of your previous PR: https://github.com/w3c/csswg-drafts/pull/14199.

Perhaps we can either merge or port your changes to the WAAPI interface to that PR?

ydaniv:
@DavMila I have merged #14199, you can merge from main

PR #14074, opened 2026-06-21
[css-values-4] Fix typo

### LINKED ISSUE #14034: [css-values] Wrong plural: "constants"
Editorial remark. *10.9.2. Infinities, NaN, and Signed Zero* says:
> NaN (short for "not a number") is the result of certain operations that don’t have a well-defined value. It can be written directly using the math constants NaN

It should be "math constant".

Psychpsyo:
Fixes #14034


Psychpsyo:
@tabatkins
Pinging you as editor of that spec. Can we get this merged?

PR #14101, opened 2026-06-29
[css-view-transitions-2] Exclude `-ua-*` and `none` in types

### LINKED ISSUE #14081: [view-transitions-2] Does @view-transition 'types' `<custom-ident> support values that start with "-ua-"?
The current view-transitions-2 spec for @view-transition 'types' (https://drafts.csswg.org/css-view-transitions-2/#descdef-view-transition-types) doesn't seem to place any restrictions on the values of its `<custom-ident>` sub productions, but [WPT](https://github.com/web-platform-tests/wpt/blob/befefbfef4cb4152dac829f2e29a199c76b24ac3/css/css-view-transitions/navigation/with-types/at-rule-with-types-parsing.html#L38) seems to require that values that start with "-ua-" are invalid.

Is this condition just missing from the spec? Or is it specified elsewhere and I just don't have the full context?

noamr:
Based on resolution https://github.com/w3c/csswg-drafts/issues/9534#issuecomment-1802364085
Closes #14081

### DIFF
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -2039,7 +2039,7 @@ The [=@view-transition/types=] descriptor
 
 		: <dfn><<custom-ident>>+</dfn>
 		:: Sets the [=ViewTransition/active types=] for the transition to the specified custom identifiers,
-			 excluding `none` and identifiers starting with the string "ua-".
+			 excluding `none` and identifiers starting with the string "-ua-".
 	</dl>
 
 	Note: the [=@view-transition/types=] descriptor only applies to the {{Document}} in which it is defined.
--- a/css-view-transitions-2/Overview.bs
+++ b/css-view-transitions-2/Overview.bs
@@ -2423,7 +2423,13 @@ It has the following [=struct/items=]:
 
 		1. If |typesDescriptor|'s [=computed value=] is ''@view-transition/type/none'', then return a [=/set=] « ».
 
-		1. Return a [=/set=] of strings corresponding to |typesDescriptor|'s [=computed value=].
+		1. Let |types| be a [=/set=] of strings corresponding to |typesDescriptor|'s [=computed value=].
+
+		1. [=list/Remove=] "`none`" from |types|.
+
+		1. [=list/Remove=] all items from |types| that start with the string "`-ua-`".
+
+		1. Return |types|.
 	</div>
 
 ## [=Setup view transition|Setup single-document view transition=] ## {#setup-view-transition-algorithm}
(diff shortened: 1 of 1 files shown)


css-view-transitions-2/Overview.bs
```diff
@@ -2033,6 +2033,15 @@ The [=@view-transition/types=] descriptor
 	The '<dfn for="@view-transition">types</dfn>' descriptor sets the [=ViewTransition/active types=] for the transition
 	when capturing or performing the transition, equivalent to calling {{Document/startViewTransition(callbackOptions)}} with that {{StartViewTransitionOptions/types}}.
 
+	<dl dfn-type=value dfn-for="@view-transition/types">
+		: <dfn>none</dfn>
+		:: No transition types are set.
+
+		: <dfn><<custom-ident>>+</dfn>
+		:: Sets the [=ViewTransition/active types=] for the transition to the specified custom identifiers,
+			 excluding `none` and identifiers starting with the string "ua-".
```
AtkinsSJ: Typo: Should be "-ua-" according to the commit message

noamr: Oh oops of course. Fixed

css-view-transitions-2/Overview.bs
```diff
@@ -2033,6 +2033,15 @@ The [=@view-transition/types=] descriptor
 	The '<dfn for="@view-transition">types</dfn>' descriptor sets the [=ViewTransition/active types=] for the transition
 	when capturing or performing the transition, equivalent to calling {{Document/startViewTransition(callbackOptions)}} with that {{StartViewTransitionOptions/types}}.
 
+	<dl dfn-type=value dfn-for="@view-transition/types">
+		: <dfn>none</dfn>
+		:: No transition types are set.
+
+		: <dfn><<custom-ident>>+</dfn>
+		:: Sets the [=ViewTransition/active types=] for the transition to the specified custom identifiers,
+			 excluding `none` and identifiers starting with the string "-ua-".
```
vmpstr: Can you expand a bit on this. Specifically, we either reject the full @view-transition/types string if none and "-ua-" is present, or we accept the types string but filter out none and "-ua-".

Ie what should we do with `types: none foo -ua-bar`

noamr: > Can you expand a bit on this. Specifically, we either reject the full @view-transition/types string if none and "-ua-" is present, or we accept the types string but filter out none and "-ua-".
> 
> Ie what should we do with `types: none foo -ua-bar`

I was going by the original resolution. In this case it would be equivalent to 'types: foo'. But we can make a clarifying resolution that's different.

vmpstr: Yeah that's fine, and I guess that's what the sentence says but my brain is refusing to accept that there is no ambiguity here.

vmpstr (approved):

vmpstr:
lgtm

css-view-transitions-2/Overview.bs
noamr: I made the sentence a bit clearer.

PR #14111, opened 2026-06-30
[css-view-transitions-2] Revisit the `-ua-` types fix to make parsing invalid

noamr:


cdoublev (approved):
That works for me. It is no longer ambiguous. Thanks!

The requirement to exclude some identifiers from matching `<custom-ident>` is phrased in various ways across the specs (some examples below), with some inconsistencies like not mentioning ASCII case-insensitivity or excluding identifiers that are already specified as reserved by [CSS Values](https://drafts.csswg.org/css-values-4/#custom-idents):

  > When parsing positionally-ambiguous keywords in a property value, a `<custom-ident>` production can only claim the keyword if no other unfulfilled production can claim it.

---

[`<animateable-feature> = scroll-position | contents | <custom-ident>`](https://drafts.csswg.org/css-will-change-1/#typedef-animateable-feature)

  > The `<custom-ident>` production used here excludes the keywords `will-change`, `none`, `all`, `auto`, `scroll-position`, and `contents`, in addition to the keywords normally excluded from `<custom-ident>`.

[`<container-name> = <custom-ident>`](https://drafts.csswg.org/css-conditional-5/#container-name)

  > The keywords `none`, `and`, `not`, and `or` are excluded from this `<custom-ident>`.

[`<counter-name> = <custom-ident>`](https://drafts.csswg.org/css-lists-3/#typedef-counter-name)

  > A `<counter-name>` name cannot match the keyword `none`; such an identifier is invalid as a `<counter-name>`.

[`<counter-style-name> = <custom-ident>`](https://drafts.csswg.org/css-counter-styles-3/#typedef-counter-style-name)

  > A `<counter-style-name>` is a `<custom-ident>` that is not an ASCII case-insensitive match for `none`.

PR #14122, opened 2026-07-01
[css-navigation-1] Change `:trigger-link` to `:nav-source` and include submit/form

### LINKED ISSUE #11801: [css-navigation-1] Add pseudo-class selector to target the element that initiated the outgoing navigation
## Intro

One of the properties of the  [the Navigation API](https://html.spec.whatwg.org/multipage/nav-history-apis.html#navigation-api)’s `NavigateEvent` is the `sourceElement`. From [the explainer](https://github.com/WICG/navigation-api/blob/main/README.md#navigation-monitoring-and-interception).

> `sourceElement`: An `Element` or null, indicating what element (if any) initiated this navigation. If the navigation was triggered by a link click, the `sourceElement` will be the `<a href="...">`. If the navigation was triggered by a form submission, the `sourceElement` will be the [the element that sent the `submit` event to the form](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter), or if that is null, the `<form>` being submitted. The `sourceElement` will also be null when a targeting a different `window` (e.g., `<a href="..." target="otherWindow">`).

This property is shipping in Chrome 135 and other browser vendors are supportive of it ([WebKit](https://github.com/WebKit/standards-positions/issues/441) / [Mozilla](https://github.com/mozilla/standards-positions/issues/1154)).

## The request

It would be handy if it was possible to target the element using a CSS selector, e.g. something like `:navigation-trigger` or `:navigation-source-element` _(name TBD, of course)_

A use-case that is very common and would be solved by this are View Transitions where authors want to set a `view-transition-name` onto the element that was clicked. Think of an overview page with a list of links to detail pages. Typically they only want the one title or thumbnail to get a `view-transition-name`.

Right now they need script to set that View Transition name, whereas it would be nice to be able to just do that in CSS:

```css
.list a:navigation-trigger img {
  view-transition-name: thumbnail;
}
```

I’ve personally had this need before _(in [this demo](https://view-transitions.chrome.dev/profiles/mpa/))_, and have also received requests from authors regarding this _([example](https://front-end.social/@Jbasoo@mastodon.social/114001484513694453))_.

## The lifetime of this selector

Something that would need to be thought are UA back / forward navigations. When clicking a link – which sets `:navigation-trigger` – and then hitting the UA back button the same transition should run (but in reverse) and then hitting the UA forward button it should also run again.

~~To solve that, I would say this new selector starts matching onto an element when the navigation gets initiated, but that unmatching is not linked to the navigation completing. Instead, existing `:navigation-trigger` matches get unmatched when the navigation gets cancelled. This can occur when a new navigation starts (e.g. you click a link, but then click an other one: only the last one is matched by `:navigation-trigger`) or when people hit the cancel button of their browser.~~

~~Alternatively some logic could be worked out where the last `:navigation-trigger` match of a page gets remembered and dynamically gets re-applied when the back/forward buttons are used.~~

~~_(Admittedly this part is still a bit vague and needs more thought.)_~~

_UPDATE: This selector only matches while a navigation is active. The animation-on-back-navigation part is to be handled by the https://drafts.csswg.org/css-navigation-1/ spec._

noamr:
Resolution:
https://github.com/w3c/csswg-drafts/issues/11801#issuecomment-4857772799

Closes #11801

### DIFF
--- a/css-navigation-1/Overview.bs
+++ b/css-navigation-1/Overview.bs
@@ -599,7 +599,7 @@ NOTE: Some of the design discussion for this feature has been in
 <h3 id="nav-source-pseudo-class">The '':nav-source'' pseudo-class</h3>
 
 This specification defines a <dfn id="nav-source-pseudo" selector>'':nav-source''</dfn> pseudo-class,
-mathcing the element that triggered the current navigation. It is a CSS reflection of the {{NavigateEvent/sourceElement}} property in the Navigation API.
+matching the element that triggered the current navigation. It is a CSS reflection of the {{NavigateEvent/sourceElement}} property in the Navigation API.
 The trigger element can be an [^a^], [^area^], [^form^], a [^input/type/submit^] input, or a [^button^].
 
 The '':nav-source'' pseudo-class matches any element <var>el</var> where
--- a/css-navigation-1/Overview.bs
+++ b/css-navigation-1/Overview.bs
@@ -596,27 +596,32 @@ by matching the relevant parts of the navigation URL to the link URL.
 NOTE: Some of the design discussion for this feature has been in
 <a href="https://github.com/w3c/csswg-drafts/issues/13163">w3c/csswg-drafts#13163</a>.
 
-<h3 id="trigger-link-pseudo-class">The '':trigger-link'' pseudo-class</h3>
+<h3 id="nav-source-pseudo-class">The '':nav-source'' pseudo-class</h3>
 
-This specification defines a new
-<dfn id="trigger-link-pseudo" selector>'':trigger-link''</dfn>
-that matches link elements that trigger the current navigation.
+This specification defines a <dfn id="nav-source-pseudo" selector>'':nav-source''</dfn> pseudo-class,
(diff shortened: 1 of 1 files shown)


css-navigation-1/Overview.bs
```diff
@@ -596,30 +596,35 @@ by matching the relevant parts of the navigation URL to the link URL.
...
 
-This specification defines a new
-<dfn id="trigger-link-pseudo" selector>'':trigger-link''</dfn>
-that matches link elements that trigger the current navigation.
-
-The '':trigger-link'' pseudo-class matches any element where both:
-* the element matches '':any-link''
-* the [=current navigation state=] is not null, and element is its [=navigation state/source element=].
+This specification defines a <dfn id="nav-source-pseudo" selector>'':nav-source''</dfn> pseudo-class,
+mathcing the element that triggered the current navigation. It is a CSS reflection of the {{NavigateEvent/sourceElement}} property in the Navigation API.
```
bramus: Typo.

```suggestion
matching the element that triggered the current navigation. It is a CSS reflection of the {{NavigateEvent/sourceElement}} property in the Navigation API.
```

bramus (changes requested):

bramus (approved):

PR #14124, opened 2026-07-02
[css-speech][editorial] Enforce positive `<generic-voice>` variant using range notation

cdoublev:
The `<integer>` representing a `voice-family` variant is restricted to non-negative values in prose:

  > **`<integer>`**
  >
  > An integer indicating the preferred variant (e.g. "the second male child voice"). **Only positive integers (i.e. excluding zero) are allowed.** The value 1 refers to the first of all matching voices.

But the [`<generic-voice>`](https://drafts.csswg.org/css-speech-1/#typedef-generic-voice) production rule does not enforce it with a range notation.


SebastianZ:
Seems fine to me. Thanks @cdoublev!

Sebastian

PR #14142, opened 2026-07-08
[css-borders-4] Handle rectangle, self-intersecting, and non-uniform outset corners in the contoured path algorithm

### LINKED ISSUE #13318: [css-backgrounds-3] `spread` variable is undefined in "adjusted radius dimension"
[adjusted radius dimension](https://drafts.csswg.org/css-backgrounds/#adjusted-radius-dimension) lists three parameters:

> To compute the adjusted radius dimension given numbers coverage, radius, and outset:

However, step 1 uses `spread`, which isn't defined:
> 1. If radius is greater than spread, or if coverage is greater than 1, then return radius + outset.

As a side note for anyone who fixes this: The related [outset-adjusted border radius](https://drafts.csswg.org/css-backgrounds/#outset-adjusted-border-radius) algorithm misspells "adjusted" in a couple of variable names: `adustedRadiusWidth` and `adustedRadiusHeight`.

Pinging @noamr as these were both introduced in #12896.

### LINKED ISSUE #14157: [css-borders-4]  Stroke/hull direction uses the normalized (signed) half corner, but implementations require the convex half corner for concave shapes
# [css-borders-4] Stroke/hull direction uses the normalized (signed) half corner, but implementations require the convex half corner for concave shapes

**Spec:** CSS Borders and Box Decorations Level 4 — *To get the border-aligned corner clip-out path*

## Problem

When deriving the inset stroke/hull direction, the algorithm does:

> Let `clampedHalfCorner` be the *normalized superellipse half corner* given `clamp(curvature, -1, 1)`.
> Let `vectorFromStartToControlPoint` be `(2 · clampedHalfCorner − 0.5, 1.5 − 2 · clampedHalfCorner)`.
> Let `strokeA, strokeB` be the perpendicular of the normalized `vectorFromStartToControlPoint`.

The *normalized superellipse half corner* applies a `1 − convexHalfCorner` flip for `s < 0` (the concave shapes: scoop, notch). But both Blink and WebKit use the plain **convex** half corner here (no flip). With the spec's normalized (flipped) value, the resulting `strokeA`/`strokeB` point roughly 90° away from the correct inward direction, so **concave corners are offset the wrong way** — their inner tip is not inset.

This looks like a genuine mismatch between the spec text and shipping implementations, not just an editorial nuance.

## The divergence

Blink (`third_party/blink/renderer/platform/geometry/contoured_rect.{h,cc}`):

```cpp
// contoured_rect.h — curvature is k = 2^s
static constexpr float HalfCornerForCurvature(float curvature) {
  return std::pow(0.5, 1 / ClampCurvature(curvature));   // convex; no "1 - x" flip
}
constexpr bool IsConcave() const { return curvature_ < 1; }
constexpr Corner Inverse() const {
  return Corner({Start(), Center(), End(), Outer()}, 1 / Curvature());
}

// contoured_rect.cc — Corner::AlignedToOrigin(), the inset stroke/hull step
const float clamped_half_corner =
    Corner::HalfCornerForCurvature(ClampTo<float>(curvature, 0.5, 2));
const gfx::Vector2dF normalized_hull_vector = gfx::NormalizeVector2d(
    gfx::ScaleVector2d(
        gfx::Vector2dF(clamped_half_corner, 1 - clamped_half_corner), 2) -
    gfx::Vector2dF(.5, .5));
const gfx::Vector2dF adjusted_offset{normalized_hull_vector.x(),
                                     -normalized_hull_vector.y()};
```

Blink represents concavity as *reciprocal* curvature (`k < 1`, `IsConcave()`), handled by `Inverse()` elsewhere, so `HalfCornerForCurvature` never applies the `1 − convex` transform. Its hull vector is `(2·hc − 0.5, 2·(1 − hc) − 0.5)` with the perpendicular taken via `adjusted_offset = (x, −y)`, where `hc` is the convex half corner. WebKit does the same (`Source/WebCore/platform/graphics/CornerShapeUtilities.cpp`, in `adjustCornerForInset`):

```cpp
// The curvature is clamped to [-1, 1] (Blink clamps the equivalent 2^s to [0.5, 2])
// for the hull-direction computation only; the drawn curvature is unchanged.
double clampedCurvature = std::clamp(original.curvature, -1.0, 1.0);
double convexHalfCorner = std::pow(0.5, std::exp2(-clampedCurvature));   // convex; no "1 - x" flip
auto hullDirection = FloatSize(float(convexHalfCorner * 2.0 - 0.5), float(1.5 - convexHalfCorner * 2.0)).normalized();
strokeA = -hullDirection.height();
strokeB = hullDirection.width();
```

## Concrete example (concave clamp, `s = -1`)

The convex half corner and the spec's normalized (flipped) half corner give *perpendicular* control-point vectors:

| half corner used | `vectorFromStartToControlPoint` | perpendicular → `(strokeA, strokeB)` |
|---|---|---|
| convex `0.25` (implementations) | `(0, 1)` | `(-1, 0)` |
| spec normalized `1 − 0.25 = 0.75` | `(1, 0)` | `(0, 1)` |

With the normalized (flipped) value `(strokeA = 0, strokeB = 1)`, `offset1 = offset4 = 0`, so `adjustedCornerCenter = Center` — **the concave corner's tip is not inset at all**, and the offsets instead move toward `Outer`. The convex value `(strokeA = -1, strokeB = 0)` insets the tip correctly (via `strokeA` on both arms).

## Suggested fix

Use the **convex** half corner — `0.5 ^ (1 / 2^|s|)`, i.e. *without* the `1 − convexHalfCorner` concave flip — when deriving the stroke/hull direction, matching Blink and WebKit. Alternatively, state explicitly that the concave flip must not be applied for this step (it is still correct for the interpolation/`corner-shape` value used elsewhere).

If the current normalized half corner is intended here, it would help to explain how implementations are expected to inset concave corners with it, since as written it moves the tip in the wrong direction.

## Background

Noam Rosenthal's implementation write-up ([Implementing corner-shape](https://developer.chrome.com/blog/implementing-corner-shape), Chrome for Developers) documents this as intentional — concave shapes are computed as the *visually-symmetric reflection* of their convex counterpart:

> In terms of the specification and implementation, we are representing something visual here, so we use the symmetrical equivalents when computing concave shapes. **The rest of the math is done on convex shapes** (k>=1, or positive superellipse values).

That is precisely why the unflipped **convex** half corner is the correct input to the stroke/hull-direction step, and why the `1 − convexHalfCorner` flip should be scoped to the interpolated `corner-shape` value rather than the inset geometry — matching both Blink and WebKit above.

### LINKED ISSUE #14183: [css-borders-4] Editorial: doubled "the" in "get the border-aligned corner clip-out path
**Title:** `[css-borders-4] Editorial: doubled "the" in "get the border-aligned corner clip-out path"`

**Spec:** [CSS Borders and Box Decorations Level 4 — §3.9.4.2 Computing a contoured path](https://drafts.csswg.org/css-borders-4/#contour-path)

## Problem

In the *get the border-aligned corner clip-out path* algorithm, the step that defines `vectorFromStartToControlPoint` has a doubled article:

> Let vectorFromStartToControlPoint be **the the** two-dimensional vector (2 ⋅ clampedHalfCorner - 0.5, 1.5 - 2 ⋅ clampedHalfCorner).

## Suggested fix

Remove the duplicate word:

> Let vectorFromStartToControlPoint be **the** two-dimensional vector (2 ⋅ clampedHalfCorner - 0.5, 1.5 - 2 ⋅ clampedHalfCorner).

Purely editorial; no behavior change.

### LINKED ISSUE #14184: [css-borders-4] "get the border-aligned corner clip-out path" references undefined `outerCorner`
**Title:** `[css-borders-4] "get the border-aligned corner clip-out path" references undefined `outerCorner``

**Spec:** [CSS Borders and Box Decorations Level 4 — §3.9.4.2 Computing a contoured path](https://drafts.csswg.org/css-borders-4/#contour-path)

## Problem

In the *get the border-aligned corner clip-out path* algorithm, the `offset1` and `offset2` steps reference a term **`outerCorner`** that is never defined:

> Let offset1 be the vector between originalCornerStart and **outerCorner**, normalized and scaled by startInset ⋅ strokeA.
> Let offset2 be the vector between **outerCorner** and originalCornerEnd, normalized and scaled by startInset ⋅ strokeB.

Every other step in the same algorithm refers to this point as **`originalCornerOuter`** (the algorithm's first parameter). `outerCorner` appears nowhere else and has no definition, so an implementer cannot resolve it from the text.

## Suggested fix

Rename `outerCorner` to `originalCornerOuter` in both steps:

> Let offset1 be the vector between originalCornerStart and **originalCornerOuter**, normalized and scaled by startInset ⋅ strokeA.
> Let offset2 be the vector between **originalCornerOuter** and originalCornerEnd, normalized and scaled by startInset ⋅ strokeB.

Editorial (undefined-term) fix; the intended value is unambiguous.

### LINKED ISSUE #14185: [css-borders-4] "get the border-aligned corner clip-out path" references undefined `unitVectorFromStartToControlPoint
**Title:** `[css-borders-4] "get the border-aligned corner clip-out path" references undefined `unitVectorFromStartToControlPoint``

**Spec:** [CSS Borders and Box Decorations Level 4 — §3.9.4.2 Computing a contoured path](https://drafts.csswg.org/css-borders-4/#contour-path)

## Problem

In the *get the border-aligned corner clip-out path* algorithm, the `controlPoint` step references a term **`unitVectorFromStartToControlPoint`** that is never defined:

> Let controlPoint be the result of calling mapPointToCorner given vectorFromStartToControlPoint's y and 1 - **unitVectorFromStartToControlPoint**'s x.

The normalized (unit) vector defined earlier in the same algorithm is named **`singlePixelVectorFromStartToControlPoint`**:

> Let **singlePixelVectorFromStartToControlPoint** be the vectorFromStartToControlPoint, normalized.

`unitVectorFromStartToControlPoint` appears nowhere else and has no definition, so it looks like an un-unified rename of `singlePixelVectorFromStartToControlPoint`.

## Suggested fix

Rename `unitVectorFromStartToControlPoint` to `singlePixelVectorFromStartToControlPoint`:

> Let controlPoint be the result of calling mapPointToCorner given vectorFromStartToControlPoint's y and 1 - **singlePixelVectorFromStartToControlPoint**'s x.

(If a different vector was intended here, the definition is missing entirely and should be added.)

Editorial (undefined-term) fix.

sb3nder:
- Refactored `border-aligned corner clip-out path` to:
  - Handle non-square corners by scaling the "normal" vectors.
  - Fix when the miter degenerates and self-intersects for concave corners. Fixed by calculating the miter lines as perpendicular to the "normal" vectors (AKA "tangent").
  (note: in square corners this results in the same miter points as the control-point method. When the corner is not square: in concave corners the miter points diverge from the control-point method, in convex corners the miter points are almost the same as the control-point method.)
  - Fix non uniform outset corners and mixed-inset-sign corners, by computing bevel corners as the tangent to 2 circles with center at corner-start and corner-end, and radius the 2 signed insets. All other intermediary corner shapes are adjusted accordingly and are also fixed.
  - Fix "ripple" issue for 0 < k < 1 by drawing a single superellipse curve without miter lines, ensuring C2 continuity.
  - Fix C1 discontinuity for -1 < k < 0 between the curve and miter lines by drawing an approximate superellipse that can be mapped to any case.
  - Fix C0 discontinuity in self-intersections as a result of the C1 continuity fix.
  - Respect "blue area rule" by avoiding the control point method.
  - All mixed-sign/non-uniform inset cases transition are C0 continuous.

- Refactored `Computing a contoured path` to:
  - Respect the `outset-adjusted border radius` for round corner compat.
    - Fixed by substituting the following algorithm:
      > 4. Let adjustedRadiusInOutsetCoordinates be the outset-adjusted border radius given borderRect’s size, radius, and (-insetX, -insetY).
      > 5. Return (adjustedRadiusInOutsetCoordinates’s width - insetX, adjustedRadiusInOutsetCoordinates’s height - insetY).

      with `adjustedInset`, which calculates the inset from the adjusted radius, and by computing the borderRect outer point from the targetRect and the adjusted inset.

- Applied `opposite corner scale factor` to the used value.
- Made various math fixes.
- Updated explanation for non-linear interpolation.
- Fixed some typos.

Fixes https://github.com/w3c/csswg-drafts/issues/13318
Fixes https://github.com/w3c/csswg-drafts/issues/14183
Fixes https://github.com/w3c/csswg-drafts/issues/14184
Fixes https://github.com/w3c/csswg-drafts/issues/14185
Fixes https://github.com/w3c/csswg-drafts/issues/14157

WPT tests:
- https://github.com/web-platform-tests/wpt/pull/61696
- [corner-shape-interpolation.html](https://github.com/web-platform-tests/wpt/blob/master/css/css-borders/corner-shape/corner-shape-interpolation.html) (already ok)

Questions:
- https://github.com/w3c/csswg-drafts/issues/14270

- https://github.com/w3c/csswg-drafts/issues/14158

### Desmos graph https://www.desmos.com/calculator/xaahq21qrk

screencasts:
current algorithm, chrome implementation:

https://github.com/user-attachments/assets/48523c7b-10b4-4d72-850d-fda7045e363b

circa proposed changes (old video, including only some of the fixes):

https://github.com/user-attachments/assets/78960c00-d2e3-4485-9f14-10db51c510b8

chrome bugs:
https://issues.chromium.org/issues/532190484
https://issues.chromium.org/issues/532199267
https://issues.chromium.org/issues/532163949

### DIFF
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -1580,8 +1580,8 @@ To compute an [=/element=] |element|'s <dfn>border contour path</dfn> given numb
 		1. Let |virtualEndRadius| be <code>|outsetAdjustedEndRadius| - |uniformVirtualOutset|</code>.
 		1. Let |virtualStartInset| be <code>-|uniformVirtualOutset|</code>.
 		1. Let |virtualEndInset| be <code>-|uniformVirtualOutset|</code>.
-		1. Let |vectorToVirtualCornerOuterV3| be |normalizedV3|, [=two-dimensional vector/scale|scaled by=] <code>|insetStart| + |uniformVirtualOutset|</code>.
-		1. Let |vectorToVirtualCornerOuterV2| be |normalizedV2|, [=two-dimensional vector/scale|scaled by=] <code>|insetEnd| + |uniformVirtualOutset|</code>.
+		1. Let |vectorToVirtualCornerOuterV3| be |normalizedV3|, [=two-dimensional vector/scale|scaled by=] <code>|insetEnd| + |uniformVirtualOutset|</code>.
+		1. Let |vectorToVirtualCornerOuterV2| be |normalizedV2|, [=two-dimensional vector/scale|scaled by=] <code>|insetStart| + |uniformVirtualOutset|</code>.
 		1. Let |virtualCornerOuter| be |cornerOuter|, [=point extended by vectors|extended by=] « |vectorToVirtualCornerOuterV3|, |vectorToVirtualCornerOuterV2| ».
 
 		1. Return the [=border-aligned corner clip-out path=] given 
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -1789,7 +1789,7 @@ Interpolating corner shapes</h4>
 
 Since a <<corner-shape-value>> can always be expressed by a ''superellipse()'' with an [=superellipse parameter=] variable, interpolating between two
 <<corner-shape-value>>s is done by interpolating the [=superellipse parameter=] itself.
-Since it uses a <code>log2</code>, interpolating it linearly would result in an effect where concave corners interpolate at a much higher velocity than convex corners.
+Since the bevel, round, and straight corner shapes correspond to parameter values of 0, 1, and infinity, respectively, interpolating between them linearly would make the transition from a bevel to a straight corner shape appear to start very quickly and slow down toward the end.
 To balance that, the <dfn>superellipse interpolation</dfn> formula describes how a [=superellipse parameter=] is converted to a value between 0 and 1, and vice versa:
 
 <div algorithm="superellipse-param-to-interpolation-value">
--- a/css-borders-4/Overview.bs
+++ b/css-borders-4/Overview.bs
@@ -675,12 +675,12 @@ Corners</h2>
 
 	To compute the <dfn export>outset-adjusted border radius</dfn> given the 2-dimensional [=size=]'s |edge|, |radius|, and |outset|:
 		1. Let |coverage| be <code>2 * min(|radius|'s [=width=] / |edge|'s [=width=], |radius|'s [=height=] / |edge|'s [=height=])</code>.
-		1. Let |adustedRadiusWidth| be the [=adjusted radius dimension=] given |coverage|, |radius|'s [=width=], and |outset|'s [=width=].
-		1. Let |adustedRadiusHeight| be the [=adjusted radius dimension=] given |coverage|, |radius|'s [=height=], and |outset|'s [=height=].
-		1. Return (|adustedRadiusWidth|, |adustedRadiusHeight|).
+		1. Let |adjustedRadiusWidth| be the [=adjusted radius dimension=] given |coverage|, |radius|'s [=width=], and |outset|'s [=width=].
+		1. Let |adjustedRadiusHeight| be the [=adjusted radius dimension=] given |coverage|, |radius|'s [=height=], and |outset|'s [=height=].
+		1. Return (|adjustedRadiusWidth|, |adjustedRadiusHeight|).
 
 	To compute the <dfn>adjusted radius dimension</dfn> given numbers |coverage|, |radius|, and |outset|:
-		1. If |radius| is greater than |spread|, or if |coverage| is greater than 1, then return <code>|radius| + |outset|</code>.
+		1. If |radius| is greater than |outset|, or if |coverage| is greater than 1, then return <code>|radius| + |outset|</code>.
 		1. Let |ratio| be <code>|radius| / |outset|</code>.
 		1. Return <code>|radius| + |outset| * (1 - (1 - |ratio|)<sup>3</sup> * (1 - |coverage|<sup>3</sup>))</code>.
 
@@ -898,7 +898,7 @@ Corner Sizing: the 'border-*-*-radius' properties</h3>
 
 		<figcaption>
 			The two values of ''border-top-left-radius: 55pt 25pt''
-			define the curvature of the corner.
+			define the size of the corner.
 		</figcaption>
 	</figure>
 
@@ -1191,15 +1191,15 @@ Corner Shaping: the 'corner-*-shape' properties</h3>
 			when |K| is 0 it's a diamond with perfectly flat sides.
 
 		* Negative values define concave curves,
-			roughly opposite of what you get with positive values:
-			a |K| of -1 gives a nearly elliptical "scoop"
-			that's roughly the opposite of a |K| of 1,
+			exact opposite of what you get with positive values:
+			a |K| of -1 gives an elliptical "scoop"
+			that's the opposite of a |K| of 1,
 			a |K| of -2 gives a "squircle" scoop,
 			a |K| of negative infinity gives a square scoop,
 			etc.
 
 		(Note that most literature on superellipses
-		will write the equation with a simpler <math display=inline><msup><mi>x</mi><mi>K</mi></msup></math> exponent.
+		will write the equation with a simpler <math display=inline><msup><mi>x</mi><mi>n</mi></msup></math> exponent.
 		The <math display=inline><msup><mi>x</mi><msup><mn>2</mn><mi>K</mi></msup></msup></math> form
 		was chosen here
 		to make the argument ranges easier to reason about:
@@ -1521,6 +1521,16 @@ The <dfn for="two-dimensional vector">perpendicular</dfn> of a [=two-dimensional
 
 The <dfn>vector between two points</dfn>, given {{DOMPoint}}s |a| and |b|, is <code>(|b|'s {{DOMPointReadOnly/x}} - |a|'s {{DOMPointReadOnly/x}}, |b|'s {{DOMPointReadOnly/y}} - |a|'s {{DOMPointReadOnly/y}})</code>. 
 
+<div algorithm="sum-vectors">
+To <dfn for="two-dimensional vector list">sum</dfn> a [=/list=] of [=two-dimensional vector=]s |vectors|:
+	1. Let |x| be <code>0</code>.
+	1. Let |y| be <code>0</code>.
+	1. [=list/For each=] |v| in |vectors|:
+		1. Increment |x| by |v|[0];
+		1. Increment |y| by |v|[1];
+	1. Return <code>(|x|, |y|)</code>.
+</div>
+
 <h5 id=contour-path>Computing a contoured path</h5>
 
 This algorithm describes how to compute a path with shaped corners (a path that is either inset or outset from the original).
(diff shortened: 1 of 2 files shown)


noamr:
Thanks for working on this! The screen casts look like they fix the issue but it's hard for me to assess the formula without tests. 

Note that the current WPTs for corner-shape implement the spec'ed formula in JS using canvas, see https://github.com/web-platform-tests/wpt/blob/master/css/css-borders/corner-shape/resources/render-corner-shape.js

So supplementing this PR with a WPT PR that does this and some test cases that fix the aforementioned issues would be very helpful.

(Side note: IPR for this issue needs to be filled out)

sb3nder:
Thank you for the pointers. I've added WPTs to the todo list in the PR, and hopefully I'll be able to implement them.

noamr:
> Thank you for the pointers. I've added WPTs to the todo list in the PR, and hopefully I'll be able to implement them.

Let me know if you need any help. I also might try to get to trying this out.

svgeesus:
s it is a separate repo for WPT, adding tests should be a separate PR right?

sb3nder:
> it is a separate repo for WPT, adding tests should be a separate PR right?

Yep, the todo list is mostly there to remind me what I still need to do for this PR.

sb3nder:
> What should non-round, non-uniform outsets or mixed-sign outset/inset corners look like?

Currently, they look like this:

https://github.com/user-attachments/assets/47a7339d-fb73-4308-9b7d-91803436de00

[overflow-clip-margin](https://codepen.io/sb3nder/pen/VYPxGLo)
[clip-path](https://codepen.io/sb3nder/pen/NPdMLxp)

For rounded corners, what I did is basically the same as Firefox in `overflow-clip-margin`, or Chrome when it uses the contoured path code.

I draw corners with non-uniform outsets as the uniform outset of a virtual corner,
I do this to avoid cases where the `clip-out path` breaks with non-uniform outsets.
(You can see this issue slightly in the video above, in the bevel corner case)
However, this does not look quite right to me. Maybe there is a better way to handle this issue directly in the `clip-out path`?

sb3nder:
In the following case (positive non-uniform outsets):
image

I think the corner should always be drawn without overflowing the blue area:
screen_20260720_162428

Let me know if you agree (👍) or disagree (👎).

cc @noamr

sb3nder:
@noamr no progress yet. I got sidetracked trying to handle non-uniform outsets and was considering giving that another try.
Do you think I should leave that as-is for now and proceed with the WPT tests?

noamr:
> @noamr no progress yet. I got sidetracked trying to handle non-uniform outsets and was considering giving that another try. Do you think I should leave that as-is for now and proceed with the WPT tests?

Up to you!

noamr:
This seems like really stellar work. Thanks a lot. We are testing it on the blink implementation/WPT and review it based on this

sb3nder:
Thanks! I really appreciate it.
Happy to help if anything comes up.

noamr:
I want to merge this.
@emilio @smfr @nt1m can you see if this is OK given your ongoing implementation of the feature?
We tested this in blink/WPT and it seems to give better results along all kinds of edge cases.

smfr:
Do you plan to update the WPT JS to show the code? We'd like to review that, and see if it's different from what we've done in WebKit.

noamr:
> Do you plan to update the WPT JS to show the code? We'd like to review that, and see if it's different from what we've done in WebKit.

https://github.com/web-platform-tests/wpt/pull/61658

sb3nder:
I reworked the test algorithm a bit, added some new test cases, and added support for `clip-path: margin-box` to cover non-uniform outsets.
https://github.com/web-platform-tests/wpt/pull/61696

css-borders-4/Overview.bs
```diff
@@ -1650,39 +1709,81 @@ When concave 'corner-shape' values are present (the [=superellipse parameter=] i
...
-		mapped to (0, 0, |element|'s [=computed value|computed=] 'border-top-left-radius').
+
+	1. Let |topRightRadius| be |element|'s [=computed value=] of 'border-top-right-radius'.
+	1. Let |bottomRightRadius| be |element|'s [=computed value=] of 'border-bottom-right-radius'.
+	1. Let |bottomLeftRadius| be |element|'s [=computed value=] of 'border-bottom-left-radius'.
+	1. Let |topLeftRadius| be |element|'s [=computed value=] of 'border-top-left-radius'.
+
+	1. Let |unmappedTopRightHull| be the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-top-right-shape',
+	1. Let |unmappedBottomRightHull| be a the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-bottom-right-shape',
+	1. Let |unmappedBottomLeftHull| be a the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-bottom-right-shape',
```
noamr:
```suggestion
	1. Let |unmappedBottomLeftHull| be a the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-bottom-left-shape'.
```

css-borders-4/Overview.bs
```diff
@@ -1650,39 +1709,81 @@ When concave 'corner-shape' values are present (the [=superellipse parameter=] i
...
-		rotated by 270deg with (0.5, 0.5) as an origin,
-		mapped to (0, 0, |element|'s [=computed value|computed=] 'border-top-left-radius').
+
+	1. Let |topRightRadius| be |element|'s [=computed value=] of 'border-top-right-radius'.
+	1. Let |bottomRightRadius| be |element|'s [=computed value=] of 'border-bottom-right-radius'.
+	1. Let |bottomLeftRadius| be |element|'s [=computed value=] of 'border-bottom-left-radius'.
+	1. Let |topLeftRadius| be |element|'s [=computed value=] of 'border-top-left-radius'.
+
+	1. Let |unmappedTopRightHull| be the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-top-right-shape',
+	1. Let |unmappedBottomRightHull| be a the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-bottom-right-shape',
```
noamr:
```suggestion
	1. Let |unmappedTopRightHull| be the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-top-right-shape'.
	1. Let |unmappedBottomRightHull| be a the [=normalized inner corner hull=] given |element|'s [=computed value|computed=] 'corner-bottom-right-shape'.
```

noamr (approved):
This is great! (% minor comments)
Way better than what we have today.
I want @smfr to sign off as well (the original resolution for the corner-shape algorithm was that @smfr and myself should do what's needed to make it look good, so having us both sign off would count as following the resolultion).

css-borders-4/Overview.bs
sb3nder: Done.

noamr:
Submitted after discussing with @smfr offline.

Issue #11469, opened 2025-01-09
[web-animations-2] [scroll-animations] should calling `play()` reset the start time of a scroll-driven animation?


graouts:
The WPT test [scroll-animations/scroll-timelines/play-animation.html](https://github.com/web-platform-tests/wpt/blob/master/scroll-animations/scroll-timelines/play-animation.html) has a "Playing a running animation resets a sticky start time" subtest that assumes that a running scroll-driven animation with an explicitly-set start time should reset its start time when `play()` is called:

```javascript
promise_test(async t => {
  const animation = createScrollLinkedAnimation(t);
  animation.play();
  await animation.ready;
  animation.currentTime = CSSNumericValue.parse("10%");
  assert_percents_equal(animation.currentTime, 10);
  animation.play();
  await animation.ready;
  assert_percents_equal(animation.currentTime, 0);
}, 'Playing a running animation resets a sticky start time');
```

From my reading of the spec, when `animation.currentTime` is explicitly set, the animation's _start time_ is resolved and the _hold time_ is unresolved. At this stage, both Chrome and WebKit agree that:

- the animation is not pending,
- the play state is "running",
- the start time is `-10%`.

So upon `animation.play()` being called, the [playing an animation](https://drafts.csswg.org/web-animations-2/#playing-an-animation-section) procedure is called with the _auto-rewind_ flag set to true. In step 3, _has finite timeline_ will be set to true since the animation is associated with a scroll timeline. This means that in step 5, _enable seek_ will be set to false. As a result, reading from the rest of the procedure, I don't see how either the animation hold time or start time is modified during this procedure and we abort the procedure in step 10.

What this test is doing seems fine to me, but it seems the spec does not support it at the moment.

graouts:
Cc @kevers-google who seems to have authored this test, as well as @flackr, @andruud and @bramus.

kevers-google:
The Blink implementation for playing-an-animation has deviated significantly from what is currently in the web-animations-2 spec. Not sure how updating the spec got overlooked (not seeing a correction in scroll-animations-1 either).  The specced version does not accommodate view-timelines as we simply don't know what start time to use until layout and we know the start and end of the active range.  To start, I can summarize the revised algorithm in Blink with an explanation of the rationale here, and then we can look into updating the spec:

1. Let aborted pause be a boolean flag that is true if animation has a pending pause task, and false otherwise.
2. Let has pending ready promise be a boolean flag that is initially false.
[...]

PR #14146, opened 2026-07-08
[web-animations-2] calling play() for a finite timeline should reset the start time to the current time

graouts:
Address https://github.com/w3c/csswg-drafts/issues/11469.

### DIFF
--- a/web-animations-2/Overview.bs
+++ b/web-animations-2/Overview.bs
@@ -654,8 +654,10 @@ as CSS Animations [[CSS-ANIMATIONS-1]].
 
     </dl>
 
-1.  If |has finite timeline| and |previous current time| is unresolved:
-    *   Set the flag |auto align start time| to true.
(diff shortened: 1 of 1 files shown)


ydaniv (approved):
Question: Is the ` and |auto-rewind| is true` really required?
Is it simply to limit this change to the WAAPI `play()` call for now?
If so, I think we can even relax it further and remove that part.

graouts:
> Question: Is the ` and |auto-rewind| is true` really required? Is it simply to limit this change to the WAAPI `play()` call for now? If so, I think we can even relax it further and remove that part.

Maybe @kevers-google can answer this better than I do, but this appears to be related to the final step in the procedure to [seamlessly update the playback rate](https://drafts.csswg.org/web-animations-1/#seamlessly-update-the-playback-rate) which will "run the procedure to [play an animation](https://drafts.csswg.org/web-animations-1/#play-an-animation) for animation with the auto-rewind flag set to false."

PR #14166, opened 2026-07-10
[meta] Don't scrub raw HTML from markdown.

### LINKED ISSUE #14162: Generate anchors for section titles when converting markdown
Our build system which serves files under https://drafts.csswg.org/ builds markdown files, like explainers, into nice HTML. That's great. However, section titles don't get an autogenerated id.

This makes linking to any particular section a pain.

```markdown
## The foobar Property
```
should generate
```html
<h2 id="the-foobar-property">The foobar Property</h2>
```
rather than the bare
```html
<h2>The foobar Property</h2>
```
Currently, this cannot even be manually worked around, because neither of the following syntaxes are supported either:
```markdown
## The foobar Property {#the-foobar-property}
```
```markdown
## The foobar Property <a id="the-foobar-property"></a>
```

The first one treats the `{#…}` part as a litteral part of the title and emmits it as is, and the second one emmits:
```html
<h2>The foobar Property  <!-- raw HTML omitted --><!-- raw HTML omitted --></h2>
```

emilio:
Seems it'd be useful, and the markdown in the repo is trusted.

Sorta-fixes #14162 (at least allows working around it)

### DIFF
--- a/bin/build-markdown.py
+++ b/bin/build-markdown.py
@@ -64,7 +64,8 @@ def extract_title(text):
 def render_markdown(text):
     proc = subprocess.run(
         ["cmark-gfm", "--extension", "table", "--extension", "autolink",
-         "--extension", "strikethrough", "--extension", "tasklist"],
+         "--extension", "strikethrough", "--extension", "tasklist",
+         "--unsafe"],
         input=text, capture_output=True, encoding="utf-8",
     )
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):
Makes sense

svgeesus (approved):

PR #14167, opened 2026-07-10
[css-animations-2] Added `animation-delay-start`/`end` and made `animation-delay` a shorthand

### LINKED ISSUE #13863: [css-animations-2] Add the `animation-end-delay` property
Today the Web Animations API allows setting the [`endDelay`](https://www.w3.org/TR/web-animations-1/#dom-effecttiming-enddelay) property of an effect, which sets the [`end-delay`](https://www.w3.org/TR/web-animations-1/#end-delay) of the animation.

The spec says:
> An [end delay](https://drafts.csswg.org/web-animations-1/#end-delay) may also be specified but is primarily only of use when sequencing animations such as by using a [sequence effect](https://www.w3.org/TR/web-animations-2/#sequence-effect).

It's not exposed as a declarative syntax, and this was fine until now, since `SequenceEffect` and `GroupEffect` are still WIP.

However, with the introduction of `animation-trigger` and `timeline-trigger` it's now possible to create a sequence of synchronized animations using only CSS, and not having to resort to WAAPI.

The problem starts when trying to reverse such a sequence, i.e. using the `play-backwards` action. See [example](https://codepen.io/editor/ydaniv/pen/019dd83c-7254-77ba-8217-dd91260a0f48), or in short, consider this case:

```css
ul {
  timeline-trigger: --tltr view() contain / contain;
}

li {
  animation: paint linear 0.5s calc(sibling-index() * 200ms) both;
  animation-trigger: --tltr play-forwards play-backwards;
}
```

Without `endDelay` set to `max(...[delay + activeDuration for each effect]) - <self-activeDuration> - <self-delay>`, when reversing the animations they all start together.
With `endDelay` set to the above, you get the entire sequence reversed with the staggered delays working as expected.

So, if we just add `animation-end-delay` authors can create such sequences using CSS without resorting to JS.

Another option would be to have some other specific mechanism to solve this problem, but I believe that for starters just exposing this property to CSS would take us there.
There is one side-effect to this, that in the above case all animations end at the same time, since the `end delay` is added to `end time`, but this is already the specced model for WAAPI.

FYI @birtles @flackr

ydaniv:
Fixes #13863 :

- Added `animation-delay-start`, basically mapped to the current definition of `animation-delay` from `css-animations-1`
- Added `animation-delay-end`, with 2 issues inlined
- Made `animation-delay` a shorthand
- Fixed errors and warnings

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -600,7 +600,10 @@ as described in [[#animations]].
 
   The 'animation-delay' property is a [=shorthand property|shorthand=]
   that sets 'animation-delay-start' and 'animation-delay-end'
-  together in a single declaration.
+  together in a single declaration. Specifying a single value will
+  set the 'animation-delay-start' to that value, and will not affect
+  the  value of 'animation-delay-end', for backward compatibility with
+  the 'animation-delay' previous longhand behavior.
 
 ## The 'animation-fill-mode' property ## {#animation-fill-mode}
 
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -69,6 +69,7 @@ spec:web-animations-1; type:dfn;
     text:play state
     text:playback direction
     text:start delay
+    text:end delay
     text:target element
     text:unresolved
 spec:web-animations-2; type:dfn;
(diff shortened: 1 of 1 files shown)


css-animations-2/Overview.bs
```diff
@@ -530,10 +531,76 @@ The above requirements do not apply
...
+## The 'animation-delay' shorthand ## {#animation-delay}
+
+  <pre class="propdef shorthand">
+    Name: animation-delay
+    Value: [ <<'animation-delay-start'>> <<'animation-delay-end'>>? ]#
+  </pre>
+
+  The 'animation-delay' property is a [=shorthand property|shorthand=]
+  that sets 'animation-delay-start' and 'animation-delay-end'
+  together in a single declaration.
```
graouts: Perhaps this is sufficiently clear, but I wonder if we should add a note here that explains that specifying a single value only sets `animation-delay-start` and why, for backward compatibility reasons, it doesn't also set `animation-delay-end`.

graouts (approved):

css-animations-2/Overview.bs
ydaniv: Sure, we can emphasize that again here if it helps

ydaniv: added

Issue #14173, opened 2026-07-13
[web-animations-2] Auto-aligning the start time should apply a pending playback rate


graouts:
Throughout the Web Animations spec we typically [apply the pending playback rate](https://drafts.csswg.org/web-animations-1/#apply-any-pending-playback-rate) of an animation when setting its [start time](https://www.w3.org/TR/web-animations-1/#animation-start-time). It seems an oversight not to do so within the procedure to [auto-aligning the start time](https://drafts.csswg.org/web-animations-2/#auto-aligning-start-time) and @kevers-google and I believe we should update that procedure to apply the pending playback rate between the existing steps 7 and 8.

graouts:
I'm not familiar with the type of resolutions that can be reached "asynchronously", but I think this one is pretty straightforward. Note also that it affects Interop 2026 since WPT issues [61146](https://github.com/web-platform-tests/wpt/issues/61146), [61147](https://github.com/web-platform-tests/wpt/issues/61147) and [61152](https://github.com/web-platform-tests/wpt/issues/61152) are caused by this oversight.

Cc @kevers-google to confirm and @flackr, @birtles, @BorisChiou, @dshin-moz and @ydaniv to review this.

BorisChiou:
I also noticed these failures in Firefox recently (but haven't investigated more). Thanks for filing the issue and I agree that we should update the spec (and/or tests) or at least need a resolution for this issue as early as possible. This definitely blocks interop.

graouts:
I don't think the tests require an update, except perhaps removing the `TODO` in the _A pending playback rate is used when determining timeline range alignment_ subtest of [scroll-animations/scroll-timelines/play-animation.html](https://wpt.fyi/results/scroll-animations/scroll-timelines/play-animation.html).

astearns:
Normative changes (even straightforward bug fixes) need a resolution. I’ll start the process and give folks a week to raise concerns

astearns:
The CSSWG will automatically accept this resolution one week from now if no objections are raised here. Anyone can add an emoji to this comment to express support. If you do not support this resolution, please add a new comment.

Proposed Resolution: [apply the pending playback rate](https://drafts.csswg.org/web-animations-1/#apply-any-pending-playback-rate) of an animation when [auto-aligning the start time](https://drafts.csswg.org/web-animations-2/#auto-aligning-start-time).

graouts:
[...]

PR #14175, opened 2026-07-13
[web-animations-2] Auto-aligning the start time should apply a pending playback rate

graouts:
Throughout the Web Animations spec we typically [apply the pending playback rate](https://drafts.csswg.org/web-animations-1/#apply-any-pending-playback-rate) of an animation when setting its [start time](https://www.w3.org/TR/web-animations-1/#animation-start-time). It is an oversight not to do so within the procedure to [auto-aligning the start time](https://drafts.csswg.org/web-animations-2/#auto-aligning-start-time).

Addresses https://github.com/w3c/csswg-drafts/issues/14173.


kevers-google (approved):
LGTM

BorisChiou:
No objection from me. Thanks for updating the spec.

Issue #2259, opened 2018-02-02
[selectors-4] HTML and case-sensitivity


annevk:
HTML has https://github.com/whatwg/html/issues/3372 (in general) and https://github.com/whatwg/html/issues/2738 (`attr()`-specific).

I'm not entirely sure what can be done here given the constraints, but I suppose at least https://drafts.csswg.org/selectors/#case-sensitive and https://drafts.csswg.org/selectors/#attribute-case could be updated to be more accurate about HTML.

(I also think it's a little misleading to talk about HTML and XML, since you can use XML elements inside HTML through scripting, but the HTML rules apply since it's an HTML document (https://dom.spec.whatwg.org/#concept-document-type).)

SelenIT:
Issue #2101 also is probably related.

svgeesus:
> I'm not entirely sure what can be done here given the constraints, but I suppose at least https://drafts.csswg.org/selectors/#case-sensitive and https://drafts.csswg.org/selectors/#attribute-case could be updated to be more accurate about HTML.

Re-animating this zombie issue.

Selectors 4 now [normatively references infra](https://drafts.csswg.org/selectors/#case-sensitive) for the definition of case-insensitive:

> All Selectors syntax is [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) (i.e. [a-z] and [A-Z] are equivalent), except for the parts that are not under the control of Selectors: specifically, the case-sensitivity of document language element names, attribute names, and attribute values depends on the document language.

and also for [attribute case](https://drafts.csswg.org/selectors/#attribute-case):

> By default case-sensitivity of attribute names and values in selectors depends on the document language.

> To match attribute values [ASCII case-insensitively](https://infra.spec.whatwg.org/#ascii-case-insensitive) regardless of document language rules, the attribute selector may include the identifier i before the closing bracket (]). When this flag is present, UAs must match the attribute’s value ASCII case-insensitively (i.e. [a-z] and [A-Z] are considered equivalent).

and

> Alternately, the attribute selector may include the identifier s before the closing bracket (]); in this case the UA must match the value case-sensitively, with “[identical to](https://infra.spec.whatwg.org/#string-is)” semantics [[INFRA]](https://drafts.csswg.org/selectors/#biblio-infra), regardless of document language rules.

[...]

PR #14176, opened 2026-07-13
[selectors-4] Fix explanation of HTML's case-matching rules. Define the default matching of document-supplied strings. #2259

tabatkins:


svgeesus (approved):
This looks better to me, pointing out the complexity with specific references

PR #14197, opened 2026-07-17
[css-font-loading-3] Add support for the new `width` terminology and alias the old `stretch` #14196

### LINKED ISSUE #14196: [css-font-loading-3] Add support for the new `width` terminology and alias the old `stretch`
The CSS Font Loading Module 3 only specifies [support for stretch](https://drafts.csswg.org/css-font-loading/#fontface-interface). However, the equivalent font-face attribute has since been renamed from `font-stretch` to `font-width` with aliasing support for the old property ([1](https://drafts.csswg.org/css-fonts-5/#idl-index)). It feels appropriate to give the same treatment to the loading equivalent. Since stretch and width refer to the same property we would also have to resolve which to prioritize if both are declared. I think the most straight-forward way of doing so is to always take the new syntax if specified and only use the old one if the new one isn't used. For example: an initializer with the options `{width: "condensed", stretch: "expanded"}` would set font-width to `"condensed"` while `{stretch: "expanded"}` sets it to `"expanded"`.

dcrescalante:


svgeesus:
Why is this PR also changing `CSSOMString` to `UTF8String`?

css-font-loading-3/Overview.bs
```diff
@@ -154,6 +157,9 @@ The <code>FontFace</code> Interface</h2>
 			throw a {{SyntaxError}};
 			otherwise, set the attribute to the serialization of the parsed value.
 
+			Note that ''stretch'' is a legacy alias of ''width'' and therefore they set the same attribute.
+			If both are specified, the value of ''stretch'' is ignored, even if ''width'' does not match the grammar.
+
```
fantasai: Is this note backed up by normative prose somewhere?

dcrescalante: After some looking I couldn't find much to back this up. Closest I got was [here](https://www.w3.org/TR/css-cascade-4/#legacy-shorthand) but it doesn't really say much on how legacy aliases should behave when both `old-property` and `new-property` are initialized at the same time via the JS api. This is how we're planning to resolve that case in Firefox, but you're right, there's no precedent for it. Should I remove the second line (or all) of the note?

dcrescalante:
> Why is this PR also changing `CSSOMString` to `UTF8String`?

Sorry, that must've slipped my mind when I wrote this. I'll change it back.

css-font-loading-3/Overview.bs
svgeesus: It isn't a legacy shorthand; it is a [legacy name alias](https://www.w3.org/TR/css-cascade-4/#legacy-name-alias). However, that part of the spec is silent about what happens when both are being set to different values; it only says that setting one sets them both.

css-font-loading-3/Overview.bs
```diff
@@ -85,7 +85,8 @@ The <code>FontFace</code> Interface</h2>
 		dictionary FontFaceDescriptors {
 			CSSOMString style = "normal";
 			CSSOMString weight = "normal";
-			CSSOMString stretch = "normal";
+			CSSOMString width;
+			CSSOMString stretch; // alias for width, sets the same attribute
```
emilio: `stretch` should still equal to normal.

dcrescalante: Okay. Out of curiosity, what would happen if it didn't?

emilio: Well you'd need to define what happens when not specified :)

dcrescalante: Rightt, makes sense. Thanks!

svgeesus (approved):

Issue #12399, opened 2025-06-25
[web-animations-2][css-animations-2] Should an animation be allowed to have multiple triggers associated?


DavMila:
I followed up #12119 where we resolved to treat AnimationTriggers as external controllers of Animations by filing  #12397 where we should resolve on the explicit API by which Animations will be associated with triggers. One additional question that we should answer is what happens when that explicit API is used to associate an animation with multiple triggers., e.g.:

```
const anim = Animation();
const trigger1 = AnimationTrigger();
trigger1.addAnimation(anim);

const trigger2 = AnimationTrigger();
trigger2.addAnimation(anim);
```

In this example, when the animation is added to `trigger2` should its association with `trigger1` be terminated or left intact? (Is there another option?)

I propose that we leave the previous association intact, i.e. allow animations be associated with multiple triggers. This will enable use cases where an author wishes to orchestrate multiple sources which can trigger the playing of an animation whether they be scroll-based source or event/input-based sources.

Since it could happen that the associated triggers wish to perform conflicting operations, e.g. `play()` vs `pause()`, we should decide what the outcome should be in these scenarios. A simple proposal would be to let the outcome be the outcome of running each trigger's operation in the order in which the triggers were created. While it would be possible to construct cases where this doesn't play so well, I think it gives an author a simple conceptual model to work with. I would propose that we go with this simple model to start, and if we discover use cases that can't be done with this simple model, we can introduce ways of describing interactions between triggers on the same animation  similar to how `animation-composition` was introduced for `animation`.

FYI @szager-chromium @flackr @ydaniv @birtles

ydaniv:
Thanks, @DavMila.
We need to try and consider different use-cases here.

Let's start with considering examples with multiple timeline-based triggers:

- 2 `view()` triggers on same animation, one on the `x` axis and the second on the `y` axis. Should each apply its effect separately? What if the author wants a behavior where only if scrolled into a specific rect, the intersect of these two triggers' ranges, then they apply their effect?
- The above example becomes more concrete when used with 2 `pointer()` timelines, if an author would attempt to apply the triggers' effect only when the pointer is over the intersection of x and y.
- 2 overlapping `view()` or `pointer()` triggers on same animation with `once`, `state`, or `alternate` types.  Does entering one range apply the effect? And when entering the overlap area, do we re-apply the effect?

If we consider event-based triggers, though they aren't fully specified yet we can think of use-cases like:

[...]

PR #14211, opened 2026-07-21
[animation-triggers-1] animation-trigger property value syntax #12399

estelle:
Only one trigger per animation.
This updates the syntax of the property to reflect that , in a comma-separated list of animation-triggers, only one trigger is allowed per animation.

related to issue #12399

On March 25, 2026: 

> RESOLVED: we allow one trigger per animation for now, throw in JS if someone tries to add more than one, and have an explicit intent to allow multiple in the future

Note: PR is failing because I am not listed as being in the CSSWG "Estelle did not make IPR commitments for this group."

### DIFF
--- a/animation-triggers-1/Overview.bs
+++ b/animation-triggers-1/Overview.bs
@@ -559,7 +559,7 @@ urlPrefix: https://www.w3.org/TR/web-animations-1/; type: dfn
 
 	<pre class=propdef>
 	Name: animation-trigger
-	Value: [ none | [ <<dashed-ident>> <<animation-action>>+ ]+ ]#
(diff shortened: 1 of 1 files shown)


astearns:
I’ve added you to the relevant group. Once you accept the invite we should be able to revalidate the ipr check

svgeesus:
@estelle you should also link your W3C account to your GitHub account, otherwise the IPR bot gets things wrong.

estelle:
> @estelle you should also link your W3C account to your GitHub account, otherwise the IPR bot gets things wrong.

Thanks. Done!

PR #14214, opened 2026-07-22
[css-values-5] Fix dfn of `<size-keyword>` type

tidoust:
The final `>` was missing. Also, Bikeshed is looking for a type, not a value.


cdoublev:
Fixed in 6200ae8. I hope it is okay if I close this.

PR #14231, opened 2026-07-27
[cssom-view] Remove a dangling scroll completion para

mustaqahmed:
This seems like a leftover line from https://github.com/w3c/csswg-drafts/pull/12606.


emilio (approved):
Thanks!

PR #14233, opened 2026-07-28
[web-animations-1] Specify handling of duration string values other than auto

### LINKED ISSUE #14206: [web-animations-1] What should happen when `duration` of `EffectTiming` is a string that is not `auto`?
The [`AnimationEffect::updateTiming`](https://drafts.csswg.org/web-animations-1/#dom-animationeffect-updatetiming) method whose `duration` member is either a number or a `DOMString`. It is specified like this:

> **duration, of type (unrestricted double or DOMString), defaulting to "auto"**
    The [iteration duration](https://drafts.csswg.org/web-animations-1/#iteration-duration), which is a real number greater than or equal to zero (including positive infinity) representing the time taken to complete a single iteration of the [animation effect](https://drafts.csswg.org/web-animations-1/#animation-effect).
    In this level of this specification, the string value auto is treated as the value zero for the purpose of timing model calculations and for the result of the [duration](https://drafts.csswg.org/web-animations-1/#dom-effecttiming-duration) member returned from [getComputedTiming()](https://drafts.csswg.org/web-animations-1/#dom-animationeffect-getcomputedtiming). If the author specifies the auto value, user agents must, however, return auto for the [duration](https://drafts.csswg.org/web-animations-1/#dom-effecttiming-duration) member returned from [getTiming()](https://drafts.csswg.org/web-animations-1/#dom-animationeffect-gettiming).

Neither this nor [`update the properties of an animation effect`](https://drafts.csswg.org/web-animations-1/#update-the-timing-properties-of-an-animation-effect) specify how DOMString values that are not `"auto"` should be treated.

birtles:
Fixes #14206.

### DIFF
--- a/web-animations-1/Overview.bs
+++ b/web-animations-1/Overview.bs
@@ -5486,7 +5486,8 @@ The {{AnimationEffect}} interface {#the-animationeffect-interface}
       [=throw=] a {{TypeError}} and abort this procedure.
 
   1. If the {{EffectTiming/duration}} member of |input| [=map/exists=],
-      and is less than zero or is the value <code class=esvalue>NaN</code>,
+      and is less than zero, is the value <code class=esvalue>NaN</code>,
+      or is a string other than the lowercase <code>auto</code>,
       [=throw=] a {{TypeError}} and abort this procedure.
 
   1. If the {{EffectTiming/easing}} member of |input| [=map/exists=]


birtles:
This doesn't need a resolution because it is just restoring accidentally dropped spec text.

Issue #14133, opened 2026-07-04
[css-color-4] Use `0` chroma when hue is powerless during conversion


romainmenke:
https://drafts.csswg.org/css-color-4/#powerless-color-component

> When performing color space conversion to a cylindrical polar color space, user agents shall treat a hue component as powerless if the chroma (or other measure of colorfulness, such as saturation in hsl) is less than the epsilon (ε) specified for that color space. For example, a gray color converted into oklch() may, due to numerical errors, have an extremely small chroma rather than precisely 0%; as a result, the hue component is powerless.

https://drafts.csswg.org/css-color-4/#specifying-oklab-oklch

> Powerless hue ε	C <= 0.000004

So `hue` is powerless when chroma is smaller than `0.000004` but chroma itself is not set to `0`.
And any missing component is treated as `0` for the actual conversion math.

So `color(from oklch(0.5 0.000004 180) display-p3 r g b)` which should have a powerless `hue` still ends up as: `color(display-p3 0.38857936 0.38856994 0.38857244)`

Converting that to `hsl` for example results in `color(srgb 0.38858148 0.38856954 0.3885725)`.

-----

Shouldn't `chroma`/`saturation` also be set to `0` when it matches the powerless `hue` condition? Otherwise the outcome is a color with whatever hue `0deg` is for a given color space. This `hue` might itself be powerless, but when converting to polar spaces it will still surface.

@svgeesus

romainmenke:
It isn't obvious what to do about `hwb` where powerlessness results from the combined value of `w` and `b`. 


Currently I am using:

```js
if ((Math.max(0, (Number.isNaN(out[1]) ? 0 : out[1])) + Math.max(0, (Number.isNaN(out[2]) ? 0 : out[2]))) >= 99.999) {
	out[0] = Number.NaN;

	if ((Math.max(0, (Number.isNaN(out[1]) ? 0 : out[1])) + Math.max(0, (Number.isNaN(out[2]) ? 0 : out[2]))) < 100) {
		if (!Number.isNaN(out[1]) && !Number.isNaN(out[2])) {
			out[2] = 100 - out[1];
		} else if (!Number.isNaN(out[1])) {
			out[1] = 100;
		} else if (!Number.isNaN(out[2])) {
			out[2] = 100;
		}
	}
}
```

1. if `w` + `b` is greater than `99.999`
2. if `w` + `b` is less than `100`
3. if `w` and `b` are not missing
[...]

PR #14244, opened 2026-07-31
[css-color-4] set chroma to 0 when converting powerless hue to missing

romainmenke:
See: https://github.com/w3c/csswg-drafts/issues/14133#issuecomment-5142309536

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -1293,6 +1293,11 @@ will have a [=powerless=] hue component.
 	have an <em>extremely small</em> chroma rather than precisely ''0%'';
 	as a result, the hue component is [=powerless=].
 
+	When changing a powerless hue component to a [=missing component=],
+	the chroma (or other measure of colorfulness, such as saturation in ''hsl'') is set to zero
+	to avoid amplifying floating-point noise.
+	In the case of ''hwb'' set ''b'' to ''100 - w'' or if either ''b'' or ''w'' is missing set the other component to ''100''.
+
 <h3 id=parse-color>
 Parsing a <<color>> Value</h3>
 
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -1296,15 +1296,17 @@ will have a [=powerless=] hue component.
 	When changing a powerless hue component to a [=missing component=],
 	the chroma (or other measure of colorfulness, such as saturation in ''hsl''; see [=analogous components=]) is set to zero
 	to avoid amplifying floating-point noise.
+	Negative values are not clamped to zero,
+	this value is only set to zero when it is larger than zero but smaller or equal to ε.
 	For ''hwb'' follow these steps instead:
 		<pre>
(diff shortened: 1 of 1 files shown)


css-color-4/Overview.bs
```diff
@@ -1293,6 +1293,11 @@ will have a [=powerless=] hue component.
 	have an <em>extremely small</em> chroma rather than precisely ''0%'';
 	as a result, the hue component is [=powerless=].
 
+	When changing a powerless hue component to a [=missing component=],
+	the chroma (or other measure of colorfulness, such as saturation in ''hsl'') is set to zero
```
svgeesus:
```suggestion
	the chroma (or other measure of colorfulness, such as saturation in ''hsl''; see [=analogous components=]) is set to zero
```

svgeesus (commented):
Looks good; one suggested link

css-color-4/Overview.bs
```diff
@@ -1293,6 +1293,11 @@ will have a [=powerless=] hue component.
 	have an <em>extremely small</em> chroma rather than precisely ''0%'';
 	as a result, the hue component is [=powerless=].
 
+	When changing a powerless hue component to a [=missing component=],
+	the chroma (or other measure of colorfulness, such as saturation in ''hsl''; see [=analogous components=]) is set to zero
+	to avoid amplifying floating-point noise.
+	In the case of ''hwb'' set ''b'' to ''100 - w'' or if either ''b'' or ''w'' is missing set the other component to ''100''.
```
romainmenke: For `hwb` this must only be done when `W + B >= 99.999` and `W + B < 100` to avoid clamping.

Maybe clearer in pseudo code than in prose?

```md
1. if `w` + `b` is greater than `99.999` and `w` + `b` is less than `100`
2. if `w` and `b` are not missing
  2.1. set `b` to `100 - w`
3. else if `w` is not missing
  3.1. set `w` to `100`
4. else if `b` is not missing
  4.1. set `b` to `100`
```

svgeesus: Yes, that is much clearer

svgeesus: The if condition on `1.` covers `2.` to `4.` if true, right?

romainmenke: Yes

Maybe clearer:

```
1. if `w` + `b` is greater than `99.999` and less than `100` and `w` and `b` are not missing
...
```

Or:

```
1. if `w` + `b` is greater than `99.999` and `w` + `b` is less than `100`
  1.1. if `w` and `b` are not missing
    1.1.1. set `b` to `100 - w`
  1.2. else if `w` is not missing
    1.2.1. set `w` to `100`
  1.3. else if `b` is not missing
    1.3.1. set `b` to `100`
```

romainmenke: I have applied the second option and also replaced `99.999` with `ε`.

I've also added a clarification that this step should only be applied for other color spaces between zero and `ε`.

svgeesus (approved):

Issue #14229, opened 2026-07-27
[css-typed-om] Should parts of the CSS namespace be exposed on workers?


emilio:
TypedOM currently extends the `CSS` namespace like [this](https://drafts.css-houdini.org/css-typed-om-1/#:~:text=clamp%22%2C%20%7D%3B-,partial%20namespace%20CSS):

```webidl
partial namespace CSS {
    CSSUnitValue number(double value);
    // ...
}
```

However, per spec the `CSS` namespace is only exposed on `Window` (via `[Exposed=Window]` in [here](https://drafts.csswg.org/cssom-1/#namespacedef-css)).

~It seems Blink at least exposes `CSS.number` and co in workers.~

When testing more aspects of the TypedOM API in workers, it seems no engine exports the CSS factory functions in workers. I think doing that would be ok since those interfaces are exposed in workers. Maybe `CSS.escape` and a couple others could work in there as well, they're useful for things like `FontFace`.

But my understanding is that it shouldn't work per spec right now. I have some ideas of how to fix this like:

 * Make the CSS namespace definition a plain `[Exposed=(Window, Worker, ...)] namespace CSS {}`
 * Make the existing CSS partials _except the TypedOM one_ `[Exposed=Window]`.

Is there a cleaner alternative @tabatkins? I worry a bit about future specs just doing `partial namespace CSS` and they not realizing by default it's exposing stuff to workers.

emilio:
It seems WebKit does not expose `CSS.number()` and co. in workers. So maybe the alternative is saying the spec is right and fix Blink?

cc @lilles @janvarga @cdumez, thoughts?

emilio:
Rewording because I misunderstood @janvarga when discussing this over Matrix :)

emilio:
It seems that you can manually construct a lot of these on a worker already via `new CSSUnitValue`, so this would mostly be exposing syntax sugar which is available on the main thread.

janvarga:
Yes, this would help with converting more Typed OM WPT to any.js

javifernandez:
I'm working on the [PR 61330](https://github.com/web-platform-tests/wpt/pull/61330) to updated some of these tests to any.js

lilles:
> It seems that you can manually construct a lot of these on a worker already via `new CSSUnitValue`, so this would mostly be exposing syntax sugar which is available on the main thread.

Yes. sgtm to expose this in workers (and layout and paint worklets)

[...]

PR #14245, opened 2026-07-31
[cssom-1] Expose the CSS namespace beyond Window; pin existing partialls #14229

javifernandez:
The CSS namespace was [Exposed=Window], which made css-typed-om's [Exposed=(Window, Worker)] partial invalid: WebIDL requires a partial's own exposure set to be a subset of the original definition's exposure set.

Widen the base namespace to (Window, Worker, PaintWorklet, LayoutWorklet) and pin every pre-existing member to the exposure it has today, so this is a no-op for everything except the css-typed-om numeric factories:

  * cssom-1: escape() pinned to [Exposed=Window]
  * css-conditional-3: supports() partial pinned to [Exposed=Window]
  * css-highlight-api-1: highlights partial pinned to [Exposed=Window]
  * css-images-4: elementSources partial pinned to [Exposed=Window]

This is being discussed in the issue #14229  and there is the corresponding css-typed-om-1 [PR 1177](https://github.com/w3c/css-houdini-drafts/pull/1177).

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -3440,9 +3440,9 @@ The <code>CSS.escape()</code> Method {#the-css.escape()-method}
 The <code>CSS</code> namespace holds useful CSS-related functions that do not belong elsewhere.
 
 <pre class=idl>
-[Exposed=Window]
+[Exposed=(Window, Worker, PaintWorklet, LayoutWorklet)]
 namespace CSS {
-  CSSOMString escape(CSSOMString ident);
+  [Exposed=Window] CSSOMString escape(CSSOMString ident);
 };
 </pre>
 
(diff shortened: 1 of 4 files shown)


cssom-1/Overview.bs
```diff
@@ -3440,9 +3440,9 @@ The <code>CSS.escape()</code> Method {#the-css.escape()-method}
 The <code>CSS</code> namespace holds useful CSS-related functions that do not belong elsewhere.
 
 <pre class=idl>
-[Exposed=Window]
+[Exposed=(Window, Worker, PaintWorklet, LayoutWorklet)]
 namespace CSS {
-  CSSOMString escape(CSSOMString ident);
+  [Exposed=Window] CSSOMString escape(CSSOMString ident);
```
emilio, line 3445: This could probably also be exposed in workers etc tho, right?

emilio (approved):
This needs a resolution, but looks reasonable to me.

cssom-1/Overview.bs
javifernandez: I think so, but I tried to preserve the original behavior for now. Worth discussing it as part of the eventual resolution.

PR #14253, opened 2026-07-31
[css-mixins-1] Mixins with no parameters serialize without parentheses. #13105

### LINKED ISSUE #13015: [css-mixins-1] Mixins can be called (but not defined) without params
It's mildly annoying that you can _invoke_ a mixin without a parameter list (`@apply --m`), but _defining_ a parameter-less mixin requires an explicitly empty list (`@mixin --m()`).

It seem more civilized to either make the parameter list optional on the defining side as well, or require `()` when applying a mixin with no parameters. 

If mixins are indeed going to gain a stronger underlying connection to custom functions (https://github.com/w3c/csswg-drafts/issues/12927), then I'd slightly prefer to just always require `()`, but I'm happy enough as long as we achieve some "symmetry" between `@apply` and `@mixin`.

kbabbitt:


kbabbitt:
@tabatkins could you ptal?

PR #14257, opened 2026-08-02
[css-backgrounds-4][editorial] Sort `background-origin` and `background-clip` values in canonical order

cdoublev:
The [grammar](https://drafts.csswg.org/css-backgrounds-4/#typedef-bg-layer) is `... || <bg-clip> || <visual-box> || ...` but historically, the canonical order, which is defined with *"per grammar"*, has always put `background-origin` before `background-clip` ([WPT](https://wpt.fyi/results/css/css-backgrounds/parsing/background-valid.html)):

```js
element.style.background = 'content-box padding-box';
[...element.style]; // ..., 'background-origin', 'background-clip'
```

And `background-origin` has priority over `background-clip` to match `<visual-box>`.

  > If two `<visual-box>` values are present, then the first sets `background-origin` and the second `background-clip`.

### DIFF
--- a/css-backgrounds-4/Overview.bs
+++ b/css-backgrounds-4/Overview.bs
@@ -1775,8 +1775,8 @@ Backgrounds Shorthand: the 'background' property</h3>
 	of values where
 
 	<pre class=prod>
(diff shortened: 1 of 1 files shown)


SebastianZ (approved):
Looks like an oversight in my change from https://github.com/w3c/csswg-drafts/commit/4c9075b834028a6c6f8c0331da55864086b90723.

So, thank you, @cdoublev!

Issue #8403, opened 2023-02-06
[css-text-decor-5] Allow percentages in `text-decoration-trim`


SebastianZ:
There was already some discussion in #4557 around allowing to use percentage values in `text-decoration-trim`.

One relatively common use case is to animate an underline from no width to its normal size when hovering a link. I saw this on several sites, lately, though I didn't save the URLs, unfortunately. Nontheless, here are some blogs and fiddles showing this effect:

https://speckyboy.com/underline-text-effects-css/
https://dev.to/afif/100-underline-overlay-animation-the-ultimate-css-collection-4p40
https://css-irl.info/animating-underlines/
https://www.rustcodeweb.com/2023/10/create-animated-underline-links-on-hover-using-html-css-only.html
https://www.dannyguo.com/blog/animated-multiline-link-underlines-with-css
https://www.tutorialkart.com/css/how-to-display-animated-underline-on-hover-for-a-link-using-css/
https://codepen.io/jstn/pen/mdoOZJ

Another one is to partially underline a headline based on its width.

The main question regarding this feature was to what the percentage values relate to.
@fantasai initially defined them to be [relative to the width of the containing block for block containers and relative to the total length of the untrimmed line decoration for inline elements](https://github.com/w3c/csswg-drafts/issues/4557#issuecomment-1117735704).
As decorations don't extend to the edges of a block container, I argued that decoration and text lengths would be disconnected with that definition. My suggestion was to [let them refer to the width of each fragment of the decoration box or the total length of the untrimmed line decoration of the in-flow inline-level children](https://github.com/w3c/csswg-drafts/issues/4557#issuecomment-1117950076).
@jfkthame and @litherum then mentioned that referring to the fragment widths might also not be the best approach and mentioned issues regarding [short fragments vs. long fragments](https://github.com/w3c/csswg-drafts/issues/4557#issuecomment-1176014387) and [bidi runs](https://github.com/w3c/csswg-drafts/issues/4557#issuecomment-1176744075).
@dbaron then [noted](https://drafts.csswg.org/css-break/#break-decoration) that this might be controlled via [`box-decoration-break`](https://drafts.csswg.org/css-break/#break-decoration).

Sebastian

SebastianZ:
Oh, of course, one well-known site that has an underline animation is https://2022.stateofcss.com/.

Sebastian

SebastianZ:
[...]

PR #14269, opened 2026-08-04
[css-text-decor-4] Updated the value for `text-decoration-inset`

dletorey:
The value is now: `<length-percentage>{1,2} | auto` As discussed in [Issue #https://github.com/w3c/csswg-drafts/issues/8403](https://github.com/w3c/csswg-drafts/issues/8403) And Firefox [Issue ](https://bugzilla.mozilla.org/show_bug.cgi?id=2044602#c1)

### DIFF
--- a/css-text-decor-4/Overview.bs
+++ b/css-text-decor-4/Overview.bs
@@ -867,7 +867,7 @@ Trimming, Expanding, and Shifting Text Decoration Lines: the 'text-decoration-in
 
 	<pre class="propdef">
 	Name: text-decoration-inset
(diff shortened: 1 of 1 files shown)


SebastianZ:
Thanks @dletorey. But that's actually not what was resolved on.

The resolution was to let percentages refer to either the total width of decoration boxes or to each individual decoration box depending on `box-decoration-break`.

@svgeesus @keithamus

Loirooriol:
Yeah, so I think this should be reverted to avoid confusion: #14279

PR #14279, opened 2026-08-05
Revert "[css-text-decor-4] Updated the value for `text-decoration-inset`"

Loirooriol:
Reverts w3c/csswg-drafts#14269

Reason: it doesn't follow the resolved behavior at all. Revert to avoid confusing implementers and authors.


keithamus:
Closed by https://github.com/w3c/csswg-drafts/pull/14282

Issue #14155, opened 2026-07-09
[css-gaps-1] In grid lanes containers, alignment should shift gap-decoration placement


jav099:
### Context

For other containers, the spec states that content distribution can enlarge gaps:

> Additional spacing may be added by `justify-content` / `align-content`. Such
> additional space increases the size of the corresponding gaps.

Decorations then get placed inside the “enlarged” gap. We propose that grid lanes containers **diverge** from that behavior along the stacking axis.

### Problem

If alignment is allowed to move each track's stacking-axis gutters independently, decorations that should read as a single continuous rail can be broken apart, for example, around a spanner. 

Beneath a spanner, a rule painted in the stacking-axis gutter beneath the spanner should read as one contiguous segment across the spanned width. Respecting per-track alignment would misalign them and break contiguity.

### Proposal

- In grid lanes containers, **gap-decoration placement is determined solely by
  the authored gutters** and does not shift due to alignment.
- We make this explicit as a grid-lanes-specific divergence from the general "alignment increases the gap" rule in the current css-gaps spec.

<img width="686" height="441" alt="Image" src="https://github.com/user-attachments/assets/22db3e7b-0e9e-44b4-82c0-e3606edd9e38" />

_The rule beneath the full-width spanner (1) stays a single contiguous rail even though item (3) is aligned to the bottom of its track._

tabatkins:
Ah, and this isn't a problem in the other layout modes, because you either can't have staggered gaps (Grid) or can't have spanners across that axis (Flexbox), right?

And this only occurs in Lanes when you have an item trapped between two spanners, or between a spanner and the start/end of the container, so deferring to the pre-alignment gap relative to the spanner is well-defined.

This makes sense to me, I think.

jav099:
After some more discussion, and given https://github.com/w3c/csswg-drafts/issues/14153#issuecomment-5133521003, we want to instead propose that alignment does indeed move the gutters/gaps. This way, each gutter is more tightly coupled with each item. For the stacking axis gaps, `rule-break: none` and `intersection` would then only really have visible effects in the gaps above spanners.

Example:

![Image](https://github.com/user-attachments/assets/8491f0aa-0814-4788-9bdc-83497736ce38)

PR #14280, opened 2026-08-05
[css-gaps-1][css-grid-3] Define gutters and gaps for gap decorations in grid lanes #14153, #14154

### LINKED ISSUE #14153: [css-gaps-1] Define gaps for grid lanes
### Context

- The normative definition of a **gap** per container type covers only multicol, flex, and grid containers. **There is no grid
  lanes entry.**
- The **Applies to** line of `row-gap` / `column-gap`
  (css-gaps-1 `Row and Column Gutters` section) likewise omits grid lanes
  containers.

### Proposal

**Grid axis (similar to grid / flex):** The gutter is the space between adjacent
tracks. It runs the full length of the container along the stacking axis at a
fixed grid-axis position, exactly as column/row gaps do between grid tracks
([css-grid-1 §Gutters](https://drafts.csswg.org/css-grid-1/#gutters)).

**Stacking axis (defined relative to items, not cells):** We define the
stacking-axis gutter relative to items:

1. A stacking-axis gap is the gutter immediately preceding an item along the stacking direction, for every item that is not the first in its track.
2. A stacking-axis gutter is positioned at its owning item's placed (running) position along the stacking axis. These gutters are independent per track and are not shared or aligned across tracks, unlike grid.
3. A spanner owns a **single** stacking-axis gutter immediately before it, spanning the grid-axis extent of all tracks it covers.

Image

Image

### LINKED ISSUE #14154: [css-gaps-1] Empty space in the stacking axis should not be part of a gap
### Context

Along the stacking axis, a grid lanes container can contain empty space that is
*not* an authored gutter.

### Problem

Empty space in the stacking axis is very ambiguous, as it can be used for dense-packing 
and alignment.

### Proposal

- A stacking-axis **gap is exactly the authored gutter** , the used
  `row-gap` / `column-gap` value carried by the owning item. Empty
  space beyond that gutter is **not part of any gap** and therefore has **no gap
  decoration**. Otherwise, there would be no way of aligning the decorations above a spanner, which an author might want with `rule-break: none`;

Image

_A stacking-axis gap is only the authored gutter. The full-width spanners (4) and (5) each take one rule; the empty space left under the shorter items (2) and (3) is not a gap and takes no decoration._

jav099:
This PR makes the relevant edits to css-gaps-1 (and css-grid-3) according to the resolution reached in https://github.com/w3c/csswg-drafts/issues/14153.

In the stacking axis, gutters are placed per item and are not affected by item self alignment, with spanners generating one separate gutter per spanned track.

These issues are related:
* https://github.com/w3c/csswg-drafts/issues/14154
* https://github.com/w3c/csswg-drafts/issues/14155


css-gaps-1/Overview.bs
```diff
@@ -109,13 +110,21 @@ Gaps Between Boxes</h2>
...
 			respectively.
 			See [[css-grid-1#gutters]] for precise details.
+
+		<dt id="gap-grid-lanes"><a>grid lanes containers</a>
+		<dd>
+			In a <a>grid lanes container</a> with column tracks,
+			[=row gap=] is the [=gutter=] between items in a track,
+			and [=column gap=] is the [=gutter=] between adjacent tracks.
+			In a container with row tracks, the reverse applies.
+			See [[css-grid-3#gutters]] for precise details.
```
kbabbitt: Could we instead specify this similarly to how we specify for flex, modulo "grid axis" and "stacking axis" instead of "main axis" and "cross axis"? Using a similar description will help reinforce that they work similarly.

css-gaps-1/Overview.bs
```diff
@@ -174,14 +183,46 @@ Gaps Between Boxes</h2>
...
+			</figcaption>
+		</figure>
+	</div>
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/grid-lanes-gaps-span.png">
+			<figcaption>
+				[=Row gaps=] in blue in a four-column grid lanes container with spanning items.
+				Note that the gap before full-width item 5 has one segment in each track.
```
kbabbitt: In the caption, can we refer to the normative text in Grid 3 to explain the segments per track? Something like
```suggestion
				Note that the gap before full-width item 5
				has one segment in each track,
				as described in [[SECTION-LINK-HERE]].
```

css-gaps-1/Overview.bs
```diff
@@ -174,14 +183,46 @@ Gaps Between Boxes</h2>
...
+		</figure>
+	</div>
+
+	<div class="example">
+		<figure>
+			<img alt="" src="images/grid-lanes-gap-space.png">
+			<figcaption>
+				Gaps in a grid lanes container with spanning items;
+				[=column gaps=] in orange and [=row gaps=] in blue.
+				Note that the empty space below items 2 and 3 is not part of a [=row gap=].
```
```suggestion
				Note that the empty spaces due to item placement
				below items 2 and 3
				are not part of the [=row gaps=],
				as described in [[SECTION-LINK-HERE]].
```

css-gaps-1/Overview.bs
```diff
@@ -2674,6 +2715,11 @@ Changes</h2>
 Changes since the <a href="https://www.w3.org/TR/2026/WD-css-gaps-1-20260624/">24 June 2026 Working Draft</a></h3>
 
 	<ul class="non-normative">
+		<li>Defined [[#gaps|gaps]] for <a>grid lanes containers</a>,
+			by reference to the companion [[css-grid-3#gutters|gutter definitions]] in CSS Grid 3,
+			and added them to the applicability of the 'gap', 'row-gap', and 'column-gap' properties.
+			(<a href="https://github.com/w3c/csswg-drafts/issues/14153">Issue 14153</a>,
+			 <a href="https://github.com/w3c/csswg-drafts/issues/14154">Issue 14154</a>)
```
```suggestion
		<li>Defined [[#gaps|gaps]] for <a>grid lanes containers</a>
			and added them to the applicability of the 'gap', 'row-gap', and 'column-gap' properties.
			(<a href="https://github.com/w3c/csswg-drafts/issues/14153">Issue 14153</a>,
			 <a href="https://github.com/w3c/csswg-drafts/issues/14154">Issue 14154</a>)
```

kbabbitt:
I edited the title to also refer to #14154, since the edits in this PR also cover the resolution in that issue.

jav099:
Addressed the latest feedback!

kbabbitt:
Gaps 1 changes lgtm. An editor for Grid 3 should sign off on those changes.

kbabbitt (approved):

fantasai (approved):
I think having a copy of the figure in the css-grid-3 spec would probably be helpful; also having at least one spanner there. Otherwise seems fine.

Issue #13731, opened 2026-03-26
[css-box-4] Scope margin-trim down to block & multi-column containers only?


bfgeek:
@tabatkins , @mstensho and myself have had some discussions about applying margin-trim to flex/grid.

It was resolved in https://github.com/w3c/csswg-drafts/issues/3255 to allow margin-trim to flex/grid.

This issue didn't have a whole lot of discussion, and the single author comment on the issue appears a little mis-guided about what this would achieve[1].

We think it'd make more sense to make margin-trim focused on block/multi-col layout instead. The use-case for margin-trim in flex/grid doesn't really exist. The use-case for consistent spacing is solved by row-gap/column-gap (and solved it well!).

In addition scoping this down to block/multi-col means that the inline-* values for margin-trim would disappear as they don't apply in these modes.

[1] Wanted a progressive enhancement technique to put grid/flex in a "clean state" to use with gaps, but this isn't what margin-trim does. 
> That way margin-trim: all could always be used to get a clean slate when using gap for spacing.

letrastudio:
Hi, it’s me, the single author who commented on [#3255](https://github.com/w3c/csswg-drafts/issues/3255) back in 2019 :)

I wholeheartedly agree with this proposal.

Clearly I misunderstood how margin-trim would apply to flex/grid containers! Now that I’ve [played around](https://codepen.io/letrastudio/pen/RNoVrbM) with the Safari implementation, I don’t see a use case for that at all. I searched Codepen and Github and only found [a single example](https://codepen.io/arpit-codes/pen/WbbJEWK) applying it to a flex container, and it’s a pointless one where `gap` would have done a better job. It looks like no real use cases have materialized in the past seven years, and it’s been available in Safari for almost half of that time.

I don’t think margin-trim solves anything new for grid or flex containers, since those create block formatting contexts on their children, suppressing margin collapse. But in block and multi-column layout it is not possible to handle every instance of collapsed margins from nested elements — that should be the main goal of `margin-trim`.

[...]

PR #14284, opened 2026-08-06
[css-box-4] Remove flex and grid from margin-trim, as resolved in #13731

mstensho:
[css-box-4] Remove flex and grid from margin-trim, as resolved in issue #13731


mstensho:
@fantasai Does this look okay?

Issue #9091, opened 2023-07-19
[css-pseudo-4] Add a pseudo-element to style all highlight pseudo-elements


dbaron:
A proposal I made during the discussion in https://github.com/w3c/csswg-drafts/issues/6641#issuecomment-1642386369 is that, given the model proposed in #6641 for how custom properties work on highlight pseudo-elements, is that we should have a pseudo-element that matches *all* of the highlight pseudo elements.  (This means both the [Higlight Pseudos](https://www.w3.org/TR/css-pseudo-4/#highlight-pseudos) and [Custom Highlight Pseudos](https://www.w3.org/TR/css-highlight-api-1/#styling-highlights).)

This is a slightly new concept because it means that when an implementation resolves the `::selection` pseudo-element (for example), it needs to look at selectors involving `::selection` and also the selectiors involving this new pseudo-element.  Or, to think about it another way, this is sort of like a pseudo-element alias.  That said, I think that still seems like a relatively straightforward feature.

This feature would make it easier for authors who are changing custom properties in a subtree to make sure that those custom properties will work correctly across all the highlight pseudo-elements, without needing to know exactly which highlight pseudo-elements will use the custom properties that they're modifying.

dbaron:
Two possible names for such a pseudo-element might be `::all-highlights` or `::any-highlight`, though I'm sure there are other options.

fantasai:
Naming-wise it should be `::any-highlight` rather than `::all-highlights`, for consistency with `::any-link`.

However, I'm not convinced this is a good idea. It makes sense for custom properties, but as far as non-custom properties go, it's highly unlikely that the author should be styling all the UA-defined highlights (including ones we haven't defined yet) in the same way.

Having `::highlight()` match all custom highlights would make sense, though.

LeaVerou:
As @tabatkins mentioned in https://github.com/w3c/csswg-drafts/issues/6641#issuecomment-1006172908 this is a much broader issue than highlights. The real author need here is to be able to define document-wide custom property values that are available everywhere.

I think instead of introducing more ad-hoc pseudo-elements, we should introduce a pseudo-element or `@rule` that allows this. Or adopt @tabatkins’s proposal and special case the root, paving the cowpaths.

tomayac:
> Having `::highlight()` match all custom highlights would make sense, though.

I just naively and without knowing this wouldn't work tried the above syntax. Then, less confidently, I tried `::highlight(*)`, then I came here. 

[...]

PR #14301, opened 2026-08-10
[css-highlight-1][css-pseudo-4] Add ::highlight(*)

DevSDK:
[css-highlight-1], [css-pseudo-4] add `::highlight(*)` selector as per https://github.com/w3c/csswg-drafts/issues/9091

### DIFF
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -256,10 +256,14 @@ Registering Custom Highlights</h3>
 		is used to identify the highlight during styling (see [[#styling-highlights]]).
 	</div>
 
-	Note: When registering a [=custom highlight=],
-	authors are advised to use a [=custom highlight name=] that is a valid CSS [=identifier=].
-	Using a name that is not a valid identifier can make the highlight hard,
-	and in some cases impossible, to style via CSS.
+	In order for a [=custom highlight=] to be styled via CSS
+	(see [[#styling-highlights]]),
+	the [=custom highlight name=] it is [=registered=] with
+	must be a valid CSS [=identifier=].
+	A [=custom highlight=] [=registered=] with a name
+	that is not a valid [=identifier=]
+	cannot be matched by ''::highlight()'',
+	not even by ''::highlight(*)''.
 
 	Note: It is possible to [=register=] a [=custom highlight=] with more than one [=custom highlight name=].
 	However, using more than one name to style a highlight
@@ -317,6 +321,19 @@ The Custom Highlight Pseudo-element: ''::highlight()''</h3>
 	is being [=contained=] or [=partially contained=]
 	in all the [=ranges=] of the [=registered=] [=custom highlight=]
 	with the [=custom highlight name=] specified as its argument.
+
+	A ''::highlight()'' selector only matches
+	a [=custom highlight pseudo-element=]
+	if its <<custom-highlight-name-selector>> matches
+	that [=custom highlight pseudo-element=]'s [=custom highlight name=],
+	i.e. if it is either ''*'' or a matching <<custom-ident>>.
+
+	The specificity of a ''::highlight()'' selector
+	with a <<custom-ident>> argument
+	is equivalent to a [=type selector=].
+	The specificity of a ''::highlight()'' selector
+	with a ''*'' argument
+	is zero.
 
 <h3 id=processing-model>
 Processing Model</h3>
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -260,9 +260,6 @@ Registering Custom Highlights</h3>
 	authors are advised to use a [=custom highlight name=] that is a valid CSS [=identifier=].
 	Using a name that is not a valid identifier can make the highlight hard,
 	and in some cases impossible, to style via CSS.
-	In particular, a [=custom highlight=] named <code>*</code>
-	can only be selected by escaping its name (e.g. ''::highlight(\*)''),
-	since an unescaped ''*'' argument matches every [=custom highlight name=].
 
 	Note: It is possible to [=register=] a [=custom highlight=] with more than one [=custom highlight name=].
 	However, using more than one name to style a highlight
--- a/css-highlight-api-1/Overview.bs
+++ b/css-highlight-api-1/Overview.bs
@@ -256,14 +256,13 @@ Registering Custom Highlights</h3>
 		is used to identify the highlight during styling (see [[#styling-highlights]]).
 	</div>
 
-	In order for a [=custom highlight=] to be styled via CSS
-	(see [[#styling-highlights]]),
-	the [=custom highlight name=] it is [=registered=] with
-	must be a valid CSS [=identifier=].
-	A [=custom highlight=] [=registered=] with a name
-	that is not a valid [=identifier=]
-	cannot be matched by ''::highlight()'',
-	not even by ''::highlight(*)''.
+	Note: When registering a [=custom highlight=],
+	authors are advised to use a [=custom highlight name=] that is a valid CSS [=identifier=].
+	Using a name that is not a valid identifier can make the highlight hard,
+	and in some cases impossible, to style via CSS.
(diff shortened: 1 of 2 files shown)


css-pseudo-4/Overview.bs
```diff
@@ -842,12 +842,16 @@ Selecting Highlighted Content: the ''::selection'', ''::search-text'', ''::targe
       grammar-error-003-manual.html
       </wpt>
 
-    <dt><dfn>::highlight(<<custom-ident>>)</dfn>
+    <dt><dfn>::highlight(<<custom-highlight-name-selector>>)</dfn>
```
fantasai: I think we can drop the `custom-` part of `<custom-highlight-name-selector>`.

css-pseudo-4/Overview.bs
```diff
@@ -842,12 +842,16 @@ Selecting Highlighted Content: the ''::selection'', ''::search-text'', ''::targe
...
     <dd>
       The ''::highlight()'' [=functional pseudo-element=] represents
       the portion of the document associated with the [=custom highlight=]
       identified by the given [=custom highlight name=].
-      The <<custom-ident>> argument is required.
+
+	<pre class=prod>
+		<dfn>&lt;custom-highlight-name-selector></dfn> = '*' | <<custom-ident>>
+	</pre>
```
fantasai: Fix the indent.

css-pseudo-4/Overview.bs
```diff
@@ -842,12 +842,16 @@ Selecting Highlighted Content: the ''::selection'', ''::search-text'', ''::targe
...
+      The <<custom-highlight-name-selector>> argument is required.
```
fantasai: I think we can drop this sentence. It's implied by the grammar.

css-pseudo-4/Overview.bs
fantasai: (Alternatively, pull this out of the `<dl>` and give it its own definition.)

css-pseudo-4/Overview.bs
```diff
@@ -842,12 +842,16 @@ Selecting Highlighted Content: the ''::selection'', ''::search-text'', ''::targe
...
       See [[CSS-HIGHLIGHT-API-1]] for details.
```
fantasai, line 867: Also this is not explaining how the argument is required. :) So putting them in the same paragraph is confusing.

css-pseudo-4/Overview.bs
```diff
@@ -842,12 +842,16 @@ Selecting Highlighted Content: the ''::selection'', ''::search-text'', ''::targe
...
-    <dt><dfn>::highlight(<<custom-ident>>)</dfn>
+    <dt><dfn>::highlight(<<custom-highlight-name-selector>>)</dfn>
     <dd>
       The ''::highlight()'' [=functional pseudo-element=] represents
       the portion of the document associated with the [=custom highlight=]
       identified by the given [=custom highlight name=].
-      The <<custom-ident>> argument is required.
+
+	<pre class=prod>
+		<dfn>&lt;custom-highlight-name-selector></dfn> = '*' | <<custom-ident>>
```
fantasai: This is giving the grammar for `*`, but it's not defining what it means. It's only defining what `<<custom-ident>>` means. You need to incorporate a definition somehow... imagine that you're new to programming and you don't know that `*` is typically a wildcard, and write the spec for someone who doesn't have that implied assumption. :)

fantasai (changes requested):
Needs a little bit of work, but overall pretty good: straightforward and short.

css-pseudo-4/Overview.bs
DevSDK: Indent fixed.

css-pseudo-4/Overview.bs
DevSDK: Removed!

css-pseudo-4/Overview.bs
DevSDK: Maybe it is treated by removing the above sentence?

css-pseudo-4/Overview.bs
DevSDK: Put the description here:

```
      A <<custom-ident>> argument selects the single [=custom highlight=]
      whose [=custom highlight name=] is that identifier.
      The ''*'' argument selects every [=custom highlight=],
      regardless of its [=custom highlight name=].
```

Does that work?

css-pseudo-4/Overview.bs
DevSDK: Removed. Thanks

css-pseudo-4/Overview.bs
```diff
@@ -842,12 +842,21 @@ Selecting Highlighted Content: the ''::selection'', ''::search-text'', ''::targe
       grammar-error-003-manual.html
       </wpt>
 
-    <dt><dfn>::highlight(<<custom-ident>>)</dfn>
+    <dt><dfn>::highlight(<<highlight-name-selector>>)</dfn>
```
fantasai, line 845: Actually now that I'm thinking about it, we could probably shorten this even more to `<highlight-selector>`. What do you think?

css-pseudo-4/Overview.bs
```diff
@@ -842,12 +842,21 @@ Selecting Highlighted Content: the ''::selection'', ''::search-text'', ''::targe
...
     <dd>
       The ''::highlight()'' [=functional pseudo-element=] represents
-      the portion of the document associated with the [=custom highlight=]
-      identified by the given [=custom highlight name=].
-      The <<custom-ident>> argument is required.
+      the portion of the document associated with
```
fantasai: This is a pre-existing error, but it should say "originating element" rather than "document" here. Maybe we can fix it while we're here?

```suggestion
      the portion of the [=originating element=] associated with
```

css-highlight-api-1/Overview.bs
```diff
@@ -318,6 +318,19 @@ The Custom Highlight Pseudo-element: ''::highlight()''</h3>
 	in all the [=ranges=] of the [=registered=] [=custom highlight=]
 	with the [=custom highlight name=] specified as its argument.
 
+	A ''::highlight()'' selector only matches
+	a [=custom highlight pseudo-element=]
+	if its <<highlight-name-selector>> matches
+	that [=custom highlight pseudo-element=]'s [=custom highlight name=],
+	i.e. if it is either ''*'' or a matching <<custom-ident>>.
```
fantasai: Given the definition in css-pseudo-4 now, I think we can replace this whole block with:


```suggestion
	selected by its argument.
```

css-highlight-api-1/Overview.bs
```diff
@@ -318,6 +318,19 @@ The Custom Highlight Pseudo-element: ''::highlight()''</h3>
...
+
+	The specificity of a ''::highlight()'' selector
+	with a <<custom-ident>> argument
+	is equivalent to a [=type selector=].
+	The specificity of a ''::highlight()'' selector
+	with a ''*'' argument
+	is zero.
```
fantasai: And then I think this paragraph probably fits better in css-pseudo-4 than here.

css-highlight-api-1/Overview.bs
DevSDK: Thanks, fixed

css-highlight-api-1/Overview.bs
DevSDK: Moved

css-pseudo-4/Overview.bs
DevSDK: Not strong opinion, but I'd lean toward keeping "name" since we already have `<pt-name-selector> = '*' | <custom-ident>` in [view-transition-1](https://www.w3.org/TR/css-view-transitions-1/#typedef-pt-name-selector)

css-pseudo-4/Overview.bs
DevSDK: Sure!

frivoal (approved):

fantasai (approved):

Issue #10960, opened 2024-09-26
[scroll-animations-1] Typo in calculation of ViewTimeline progress


nmoucht:
In https://drafts.csswg.org/scroll-animations-1/#view-timeline-progress, it defines the denominator of the progress calculation as 
> range is the [scroll offset](https://drafts.csswg.org/css-overflow-3/#scroll-offset) corresponding to the start of the [cover](https://drafts.csswg.org/scroll-animations-1/#valdef-animation-timeline-range-cover) range minus the scroll offset corresponding to the end of the cover range

but since the scroll offset of the start cover range is always smaller than for the end cover range, this will result in a negative progress. Shouldn't these be flipped (ie cover range end - cover range start)?

nmoucht:
@flackr @kevers-google @andruud

stubbornella:
@flackr can you take a look at this or is there someone else who would be better?

ydaniv:
Yes, looks like a typo, should be:
```
progress = (current - start) / (end - start)
```
With regard to corresponding scroll offsets. 
I think we wrote that in one of the issues but can't find it now.

bramus:
Good catch. Fixed in c4def77.

PR #14305, opened 2026-08-11
[scroll-animations-1] Fix ViewTimeline progress range calculation (#10960)

lancy69:
Reverses the operands in the ViewTimeline `range` definition so that progress is calculated using `end - start`, as confirmed in #10960.


ydaniv:
cc @bramus @fantasai @flackr @graouts, this looks right

PR #14336, opened 2026-08-17
[css-animations-2][editorial] A single `animation-delay` value resets `animation-delay-end`

### LINKED ISSUE #14219: [css-animations-2] Should a single value for `animation-delay` reset `animation-delay-end`?
https://drafts.csswg.org/css-animations-2/#animation-delay

  > Specifying a single value [for `animation-delay`] will set the `animation-delay-start` to that value, **and will not affect the value of `animation-delay-end`**, for backward compatibility with the `animation-delay` previous longhand behavior.

A strict interpretation is that a single value should not reset `animation-delay-end`. But since it did not exist, I do not see how this would not be backward-compatible. Is it an oversight?

cdoublev:
Fixes #14219.

### DIFF
--- a/css-animations-2/Overview.bs
+++ b/css-animations-2/Overview.bs
@@ -601,8 +601,8 @@ as described in [[#animations]].
   The 'animation-delay' property is a [=shorthand property|shorthand=]
   that sets 'animation-delay-start' and 'animation-delay-end'
   together in a single declaration. Specifying a single value will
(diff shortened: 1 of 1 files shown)


ydaniv (approved):

ydaniv:
Thanks, @cdoublev!

graouts (approved):

Issue #14306, opened 2026-08-11
[css-values][css-fonts] Make the set of properties referred to as "font-*" more clear


weinig:
In the definition of [font-relative units](https://drafts.csswg.org/css-values-4/#font-relative-lengths) there is the text:

> When used in the value of any 'font-*' property on the element they refer to...

Rather than leaving it to chance that the set of properties this refers to are the right set and all font effecting properties have the prefix `font-` and there are no non-font effecting properties that have the prefix `font-`, it would be clearer if this linked to either a complete list of properties or some other definition. 

I believe the intent is that it is any property that might influence one of the font's metrics used by font-relative units. One option would be for CSS Fonts could define that concept and this could link to it.

cc @tabatkins @fantasai @svgeesus

weinig:
An example of a property that I think probably belongs in the `font-*` group but doesn't have the `font-` prefix is [`math-depth`](https://w3c.github.io/mathml-core/#the-math-script-level-property), as it can effect the computed value of `font-size`.

tabatkins:
That is a very good point, as I did not realize that property existed or had such an effect!

tabatkins:
Checking my own understanding here.

In Fonts 4:
* The 'font' longhands are all font-affecting
* So are the 'font-synthesis' longhands, I think? Assuming that, say, synthetic bold might adjust the ch width or something.
* Are the [font-feature properties](https://drafts.csswg.org/css-fonts-4/#font-rend-props)? Hmm, I guess they can affect which glyph is used for 0, and thus the ch width too. font-variant-east-asian could affect `ic`, too. And I guess the generic `font-variant` can affect both of them in theory.
* I guess font-optical-sizing is too?
* And font-language-override, as it could affect the glyph used for `ic`.
* font-variant-emoji is *not* font-affecting, the affected characters aren't relied on by font-relative lengths nor are we likely to add font-relative lengths that are in the future reliant on those characters (or we can define them to be equal to the image variant, if we ever want to expose an emoji-width unit)
* Nor is font-palette, as it's just colors

Fonts 5 doesn't introduce any relevant new properties.

MathML will need to define that math-depth is font-affecting.

And that's it, maybe?

tabatkins:
Also, better name suggestions than "font-affecting properties"?

svgeesus:
@weinig does this new subsection [1.2. Interaction With Font-Relative Units](https://www.w3.org/TR/css-fonts-4/#font-relative) resolve this issue?

[...]

PR #14346, opened 2026-08-19
[css-values-4][editorial] Oversight

cdoublev:
Fixes an oversight in 8f01339, which referenced `math-scale` as a font-affecting property, but this property does not exist, if I am not mistaken.

This commit replaces it with [`math-depth`](https://w3c.github.io/mathml-core/#propdef-math-depth), which was referenced in the related issue (#14306).


tabatkins:
Ah sorry, fixed it myself in another commit when Bikeshed complained at me and I actually read what the error was.

Issue #14070, opened 2026-06-19
[css-color-5] Omissable alpha in relative color serialization


svgeesus:
> @kbhomes In general omitting the `/ 1` on relative colors doesn't round-trip, because omitted alpha defaults to the origin color's alpha, not `1`.
> 
> So while it's technically doable to realize that `red` is always opaque, it might not be with something like `currentColor`.
> 
> Might be a bit weird if `alpha(from currentcolor / 1)` had to serialize `alpha(from currentcolor / 1)` but `alpha(from red / 1)` serialized as `alpha(from red)`? Again, not _impossible_ but maybe a bit weird? 

 _Originally posted by @emilio in [#13992](https://github.com/w3c/csswg-drafts/issues/13992#issuecomment-4744858827)_

svgeesus:
> @emilio That's a good point, and I'd forgotten about the previous discussion that pointed out that the [alpha defaults to the origin color's](https://drafts.csswg.org/css-color-5/#relative-colors:~:text=If%20the%20alpha%20value%20of%20the%20relative%20color%20is%20omitted%2C%20it%20defaults%20to%20that%20of%20the%20origin%20color%20%28rather%20than%20defaulting%20to%20100%25%2C%20as%20it%20does%20in%20the%20absolute%20syntax%29%2E).
> 
> But this would then also apply to the other relative color functions which all currently say:
> 
> followed (if the alpha component is non-unity) by " / " and the alpha component as specified (using the same rules as the color channel arguments, but clamped)
> 
> So `oklch(from rgba(255, 0, 0, 0.5) l c h / 1)` is expected to serialize as `oklch(from rgba(255, 0, 0, 0.5) l c h)` which is a different value. Maybe all of these relative color functions should remove the "non-unity" clause from alpha serialization? 

 _Originally posted by @kbhomes in [#13992](https://github.com/w3c/csswg-drafts/issues/13992#issuecomment-4744948042)_

svgeesus:
From CSS Color 4,  all colors were absolute and thus [Serializing alpha values](https://drafts.csswg.org/css-color-4/#serializing-alpha-values) says:

> If, after clamping to the range [0, 1] the alpha is 1, it is omitted from the serialization; an implicit value of 1 (fully opaque) is the default.

> If the alpha is any other value than 1, it is explicitly included in the serialization as described below.

[...]

PR #14353, opened 2026-08-19
[css-text-decor-5] Updated the syntax of `alpha()`

dletorey:
Updated the syntax of`alpha()` function to have the `<alpha>` value required

Based upon [issue #14070](https://github.com/w3c/csswg-drafts/issues/14070)

### DIFF
--- a/css-color-5/Overview.bs
+++ b/css-color-5/Overview.bs
@@ -1811,7 +1811,7 @@ Relative alpha colors refer to an origin color, and only change the alpha channe
 The grammar of the ''alpha()'' function, new in this level, is as follows:
 
 <pre class='prod'>
-<dfn>alpha()</dfn> = alpha([from <<color>>] / [<<alpha-value>>])
+<dfn>alpha()</dfn> = alpha(from <<color>> / [<<alpha-value>> | none])
 </pre>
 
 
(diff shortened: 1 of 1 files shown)


css-color-5/Overview.bs
```diff
@@ -1811,8 +1811,7 @@ Relative alpha colors refer to an origin color, and only change the alpha channe
 The grammar of the ''alpha()'' function, new in this level, is as follows:
 
 <pre class='prod'>
-<dfn>alpha()</dfn> = alpha([from <<color>>]
-					[ / [<<alpha-value>> | none] ]? )
+<dfn>alpha()</dfn> = alpha([from <<color>>] / [<<alpha-value>>])
```
keithamus:
```suggestion
<dfn>alpha()</dfn> = alpha([from <<color>>] / <<alpha-value>>)
```

yisibl: It looks like `none` was left out as well.

keithamus: Should `none` be retained if #14070 concluded that alpha is required? What value is `alpha(from ... / none)`? I guess it empties the alpha channel for other colours so something like `color-mix(rgba(0, 0, 0, 0.5), alpha(from rgba(1, 1, 1, 0.75) / none))` produces a colour with alpha of 0.5?

So the syntax should be:

```suggestion
<dfn>alpha()</dfn> = alpha(from <<color>> / [<<alpha-value>> | none])
```

yisibl: > so something like color-mix(rgba(0, 0, 0, 0.5), alpha(from rgba(1, 1, 1, 0.75) / none)) produces a colour with alpha of 0.5?

Yes

Issue #14018, opened 2026-06-05
[css-scroll-snap] Clarify snapping behavior with multiple single-axis scrollers


vmpstr:
### Context
We resolved to create single-axis scrollers by allowing a scrollable value in one axis and clip in the other: https://github.com/w3c/csswg-drafts/issues/12289
This allowed us to create interesting effects and we resolved that sticky, for example, tracks per axis scrollers: https://github.com/w3c/csswg-drafts/issues/865#issuecomment-4040585511
@tabatkins put together a draft PR to update a bunch of specs to recognize single-axis scrollers: https://github.com/w3c/csswg-drafts/pull/13903

### Problem & Proposal

One of the questions is whether scroll snap also tracks per-axis scrollers. I think that would be reasonable. One hiccup is that [`scroll-snap-type`](https://drafts.csswg.org/css-scroll-snap/#scroll-snap-type) includes an axis in which this scroller snaps.

So, it's pretty clear what happens in a y-only scroller with a y-only scroll-snap-type: it should only snap in y and snapping in x can happen in some ancestor scroller:

```
/* does _not_ capture x snapping, it's not a scroller in that axis and scroll-snap-type doesn't include that axis */
.scroller {
  overflow-y: scroll;
  overflow-x: clip;
  scroll-snap-type: y mandatory;
}
```

Similarly, if the scroller is a scroller in both axis, then that's just the current state: nothing propagates to ancestor scrollers

However, what happens if we have a y-only scroller with a snap that includes the x axis?

```
.scroller {
  overflow-y: scroll;
  overflow-x: clip;
  scroll-snap-type: both mandatory;
}
```
I think it's reasonable to say that because this container is a not a scroller in the x direction, then it cannot prevent some ancestor from snapping in the x direction, so the scroll-snap-type here is essentially ignored when it comes to x direction of some ancestor scroller. That seems to be similar to how we deal with sticky.

For completeness, @flackr also pointed out that if the developer doesn't want some ancestor scroller to snap in the x axis, they can always say `scroll-snap-align: center none` (in horizontal-tb)

The proposal is to clarify that scroll snap also respects single axis scrollers and "propagates" across such scrollers in the axis for which the scroller clips

flackr:
[...]

PR #14354, opened 2026-08-19
[css-scroll-snap-1] Support scroll snap for single-axis scroll containers #14018 

lcdavid94:
[css-scroll-snap-1] Support scroll snap for single-axis scroll containers #14018 

Clarify scroll snapping behavior when an intermediate container is a single-axis scroll container:
- A single-axis scroll container only establishes snap positions in its scrollable axis; specifying `scroll-snap-type` in a non-scrollable axis has no effect.
- Snap position capturing and snap container resolution are evaluated per-axis, allowing unconsumed snap axes to propagate to ancestor scroll containers.

### DIFF
--- a/css-scroll-snap-1/Overview.bs
+++ b/css-scroll-snap-1/Overview.bs
@@ -302,6 +302,9 @@ Scroll Snapping Rules: the 'scroll-snap-type' property {#scroll-snap-type}
     how <a href="#snap-strictness">strictly</a> it <a>snaps</a>,
     and <a href="#snap-axis">which axes</a> are considered.
     If no strictness value is specified, ''proximity'' is assumed.
+    A [=single-axis scroll container=] only establishes snap positions
+    in its [=scrollable axis=];
+    specifying 'scroll-snap-type' for an axis in which the box does not scroll has no effect.
 
     <div class="example">
       In this example, snapping to headings is enabled in the <a>block axis</a>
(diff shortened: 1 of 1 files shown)


lcdavid94:
cc @flackr @vmpstr

css-scroll-snap-1/Overview.bs
```diff
@@ -302,6 +302,9 @@ Scroll Snapping Rules: the 'scroll-snap-type' property {#scroll-snap-type}
     how <a href="#snap-strictness">strictly</a> it <a>snaps</a>,
     and <a href="#snap-axis">which axes</a> are considered.
     If no strictness value is specified, ''proximity'' is assumed.
+    A [=single-axis scroll container=] only establishes snap positions
+    in its [=scrollable axis=];
+    specifying 'scroll-snap-type' for an axis in which the box does not scroll has no effect.
```
vmpstr: probably want to say ... for an axis which is not one of the box's scrollable axis ...

to disambiguate "does not scroll", since that could mean there is no scrollable overflow, or it's overflow hidden or overflow: visible

lcdavid94: Updated. Thanks!

css-scroll-snap-1/Overview.bs
```diff
@@ -402,15 +405,14 @@ Scroll Snap Strictness: the ''scroll-snap-type/none'', ''proximity'', and ''mand
     content in between can become inaccessible
     in cases where it is longer than the screen.
 
-    A box <dfn export>captures snap positions</dfn>
-    if it is a <a>scroll container</a>
-    <em>or</em> has a value other than ''scroll-snap-type/none'' for 'scroll-snap-type'.
-    If a box’s nearest <a lt="captures snap positions">snap-position capturing</a> ancestor
+    A box <dfn export lt="captures snap positions|capture snap positions">captures snap positions</dfn>
+    in a given axis if it is a <a>scroll container</a> [=scrollable axis|scrollable=] in that axis.
```
tabatkins, line 409: This loses the fact that scroll-snap-type captures snap positions even if it's not scrollable. (Intended to let you "trap" positions without them leaking out to a higher container.)

tabatkins (changes requested):

css-scroll-snap-1/Overview.bs
lcdavid94: The existing behavior we have is that the container captures snap positions iff it scrolls. Its `scroll-snap-type` has no impact on the behavior - I tried https://jsfiddle.net/lcdavid94/bdztrLa2/1/ in all 3 major browers and the `#target` still snaps to `#parent` even though `#space` specifies a `scroll-snap-type`. So should we keep this in the spec?

tabatkins: Hm, I'll raise a separate issue about that. We shouldn't *remove* the behavior from the spec in an incidental edit that's just adding a different feature, tho.

tabatkins: Okay, opened <https://github.com/w3c/csswg-drafts/issues/14445>. We discussed this back in 2019 (when we were the only implementor, and didn't have this behavior) and the WG left it in with the assumption that other browsers would pick it up. Nobody did, however, so I'm proposing we remove it.

PR #14355, opened 2026-08-19
[css-content-3]: Allow symbolic-counter-style in content syntax.

### LINKED ISSUE #14003: [css-content-3]: Allow `content: <list-style-type>` as shorthand for `content: counter(dummy, <list-style-type>)`.
See sibling issue https://github.com/whatwg/html/issues/12516. HTML defines base-select as using `content: counter(fake-counter-name, disclosure-open)` seemingly to get around wanting `disclosure-open` but not really wanting to use a counter.

Maybe it's worth us allowing `content: <list-style-type>`, with something akin to a "guaranteed invalid counter name"? This would allow HTML spec to then use `content: disclosure-open`.

This would have the fridge benefit of allowing e.g. `content: circle` as a more descriptive alternative to `content: \25E6` and so on.

/cc @emilio

keithamus:
/cc Content editors @fantasai / @faceless2. Counter-Style editors @tabatkins 

Closes #14003

### DIFF
--- a/css-content-3/Overview.bs
+++ b/css-content-3/Overview.bs
@@ -29,7 +29,8 @@ WPT Display: open
 	spec:css-pseudo-4; type:selector; text:::file-selector-button
 	spec:selectors-4; type:dfn; text:pseudo-elements
 	spec:css-values-5; type:function; for:/; text:attr()
-	
+	spec:css-counter-styles-3; type:value; text:circle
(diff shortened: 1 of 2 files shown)


fantasai (approved):
I think the example is not super great because you can use `list-style` pretty easily to style `::marker`, so it's not a good motivating example. But this is otherwise great. :)

Issue #14347, opened 2026-08-19
[css-color-4] rewrite `13. Color interpolation` to be more explicit about all the steps and effects


romainmenke:
For example:

[`color-mix(in hsl, hsl(30deg 40% 80% / 25%), hsl(90deg 0 80% / 25%))`](https://preset-env.cssdb.org/color/computed/#JTdCJTIyY2klMjIlM0ElMjJjb2xvci1taXgoaW4lMjBoc2wlMkMlMjBoc2woMzBkZWclMjA0MCUyNSUyMDgwJTI1JTIwJTJGJTIwMjUlMjUpJTJDJTIwaHNsKDkwZGVnJTIwMCUyMDgwJTI1JTIwJTJGJTIwMjUlMjUpKSUyMiU3RA==)
[`color-mix(in hsl, hsl(30deg 40% 80% / 25%), hsl(90deg none 80% / 25%))`](https://preset-env.cssdb.org/color/computed/#JTdCJTIyY2klMjIlM0ElMjJjb2xvci1taXgoaW4lMjBoc2wlMkMlMjBoc2woMzBkZWclMjA0MCUyNSUyMDgwJTI1JTIwJTJGJTIwMjUlMjUpJTJDJTIwaHNsKDkwZGVnJTIwbm9uZSUyMDgwJTI1JTIwJTJGJTIwMjUlMjUpKSUyMiU3RA==)

In both cases browsers do a linear interpolation between `30deg` and `90deg` while in my implementation the `90deg` is powerless and becomes missing. This missing hue is then filled in with the `30deg` from the other color.

I suggest rewriting section 13 to make it clearer what all the steps are.

-----

> 2. changing any [powerless](https://drafts.csswg.org/css-color-4/#powerless-color-component) components to [missing](https://drafts.csswg.org/css-color-4/#missing-color-component) values

This step seems redundant as it is already included in https://drafts.csswg.org/css-color-4/#color-conversion

-----

Add a step between what is now 4. and 5. for ["the missing component is treated as having the other color’s component value"](https://drafts.csswg.org/css-color-4/#interpolation:~:text=If%20a%20color%20with%20a%20carried%20forward%20missing%20component%20is%20interpolated%20with%20another%20color%20which%20is%20not%20missing%20that%20component%2C%20the%20missing%20component%20is%20treated%20as%20having%20the%20other%20color%E2%80%99s%20component%20value.)

_or maybe at another position in the sequence?_

-----

[...]

PR #14374, opened 2026-08-22
[css-color-4] clarify interpolation steps

romainmenke:
see: https://github.com/w3c/csswg-drafts/issues/14347#issuecomment-5346256034


svgeesus (commented):
Thanks!

PR #14385, opened 2026-08-24
[css-navigation-1] Rename :nav-source to :navigation-source

### LINKED ISSUE #14303: [css-navigation-1] Rename `:nav-source` to `:navigation-source`
In https://github.com/w3c/csswg-drafts/issues/12594#issuecomment-5177354464 @fantasai mentioned to rename `:nav-source` to `:navigation-source`:

> I originally was fine with either `:nav-source` or `:navigation-source` when we were reviewing that issue on its own, but in the context of `@navigation` and `navigation()` .…maybe we should go with the full name :navigation-source in order to link the concepts together more strongly.

I discussed this with @noamr and we are fine with this (and I personally also wanted `:navigation-source` anyway 😄).

Tagging for async resolution, as this is only about a rename for an unshipped feature, so there is no author-facing impact.

PROPOSED: Rename `:nav-source` to `:navigation-source`

bramus:
This renames `:nav-source` to `:navigation-source`. Fixes #14303.

### DIFF
--- a/css-navigation-1/Overview.bs
+++ b/css-navigation-1/Overview.bs
@@ -425,31 +425,31 @@ by matching the relevant parts of the navigation URL to the link URL.
 NOTE: Some of the design discussion for this feature has been in
 <a href="https://github.com/w3c/csswg-drafts/issues/13163">w3c/csswg-drafts#13163</a>.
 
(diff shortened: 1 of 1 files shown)


noamr:
cc @mstensho

noamr (approved):

Issue #9519, opened 2023-10-24
[css-scroll-snap-1] Why does scroll-snap-align: both snap in each axis independently?


flackr:
In the [css-scroll-snap-1 spec section 4.1.1](https://drafts.csswg.org/css-scroll-snap-1/#snap-axis) it states that for the scroll-snap-axis both:
> The [scroll container](https://drafts.csswg.org/css-overflow-3/#scroll-container) [snaps](https://drafts.csswg.org/css-scroll-snap-1/#scroll-snap) to [snap positions](https://drafts.csswg.org/css-scroll-snap-1/#scroll-snap-position) in both of its axes independently (potentially snapping to different elements in each axis).

What's the use case for this? I would expect the common case is that the author would expects the UA to snap to both axes of that element if it snaps to that element. If an author wants to snap independently on different axes (e.g. for rows and columns of a table) they can specify snap alignment in a single axis.

Use cases where you'd for sure want alignment in both axes are any sort of 2d selection UI, e.g. a 2d panning map with multiple POI that snaps to a particular POI.

This in particular makes using the snapchanged / snapchanging apis in css-scroll-snap-2 more complicated if you do want to know which element is aligned in both axes.

flackr:
In fact, the description of the axis values even says that it affects "whether snap positions are evaluated independently per axis, or together as a 2D point." Is it not expected that both should be treated as a 2D point or was the intention to add another value for evaluating it as a 2d point?

fantasai:
Some prior discussion on this:
- https://drafts.csswg.org/css-scroll-snap/issues-by-issue#issue-67
- https://www.w3.org/TR/2016/WD-css-scroll-snap-1-20160623/#scroll-snap-type

flackr:
> Some prior discussion on this

Forgive me, i'm having a hard time finding the actual discussion of when you want to snap to a different `scroll-snap-align: both` elements in the vertical and horizontal axes. The closest I see is https://lists.w3.org/Archives/Public/www-style/2014Jan/0591.html where it was discussed that you want to be able to make a scroller snap in one axis but not both.

Can we discuss this? I think it would be really helpful for some common use cases.

If we can't remove independent scroll snapping, could we add a keyword to opt into it? Maybe scroll-snap-type: [**element** | x | y | inline | block | both] [proximity | mandatory] where element snaps to a single element at a time (implied that it's in both axes).

[...]

PR #14386, opened 2026-08-24
[css-scroll-snap-1] Add keyword "pair" to scroll-snap-type for snapping to the same element in X and Y axes

lcdavid94:
[css-scroll-snap-1] Add keyword "pair" to scroll-snap-type for snapping to the same element in X and Y axes #9519


lcdavid94:
@vmpstr @flackr for review

tabatkins (approved):

PR #14388, opened 2026-08-25
[css-image-animations] Discuss interaction of pseudo-class with global setting

### LINKED ISSUE #14363: [css-image-animation] Security Horizontal Review
Dear Group, this is the output of the Security Review

**Security review request:** https://github.com/w3c/security-request/issues/140

**Reviewed version:** https://www.w3.org/TR/2026/WD-css-image-animation-1-20260409/

**Security and Privacy Questionnaire:** https://github.com/w3c/csswg-drafts/issues/14164

Thank you for the explanation provided in the Security and Privacy Questionnaire.

I agree with the clarification requested in the Privacy review: https://github.com/w3cping/privacy-request/issues/218

I also considered the following separate, non-blocking point.

## Avoid coupling `:animated-image` matching to a global animation setting

**Description:** The specification says that, when a user-agent setting globally disables image animations, images that could otherwise be animated are considered static and do not match `:animated-image`. In an implementation following this rule, a page could test a known animated content image and infer whether the setting is active.

**Impact:** This would expose one bit of user-agent preference state.

**Suggestions:** Consider defining selector matching independently of the global setting and letting that setting affect rendering only. If the coupling is intentional, document the exposure.

**Affected Sections:**

- https://www.w3.org/TR/2026/WD-css-image-animation-1-20260409/#animated-image-pseudo

frivoal:
This PR would close #14363.

@astearns, are you comfortable leaving this to editor's discretion? Do you want it on the call? Async maybe?

### DIFF
--- a/css-image-animation-1/Overview.bs
+++ b/css-image-animation-1/Overview.bs
@@ -311,8 +311,30 @@ Telling Animated Images Apart: the '':animated-image'' pseudo-class {#animated-i
 	by the current playing state of the image,
 	nor by the value of the ''image-animation'' property.
 	However, if some user agent setting has globally disabled image animations,
-	even images that could otherwise be animated are considered [=static images=],
-	and do not match.
+	when that setting is on,
+	such a user agent may treat as [=static images=]
+	images that would otherwise be animated but whose animation is being suppressed,
+	and have them not match '':animated-image''.
+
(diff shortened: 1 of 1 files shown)


astearns:
Let’s do an async resolution on the MAY bit this adds

PR #14419, opened 2026-08-31
[css-cascade-6] fix `@import` definition

romainmenke:
see: https://github.com/w3c/csswg-drafts/pull/13698

By the current definition:

`media-import-condition = <media-query-list>`

`media-query-list = <media-query>#`

So removing the `?` in the `@import` syntax definition made a media query required as `#` is one or more.

### DIFF
--- a/css-cascade-6/Overview.bs
+++ b/css-cascade-6/Overview.bs
@@ -96,7 +96,7 @@ Importing Style Sheets: the ''@import'' rule</h2>
 		        [[ layer | layer( <<layer-name>> ) ]
 		         || [ scope | scope( <<scope-start>> | <<scope-boundaries>> ) ]
 		         || <<supports-import-condition>>]?
-		        <<media-import-condition>> ;
+		        <<media-import-condition>>? ;
 
 		<dfn export>&lt;supports-import-condition></dfn> = supports( [ <<supports-condition>> | <<declaration>> ] )
 		<dfn export>&lt;media-import-condition></dfn> = <<media-query-list>></pre>


cdoublev:
> To parse a `<media-query-list>` production, [parse a comma-separated list of component values](https://www.w3.org/TR/css-syntax-3/#parse-a-comma-separated-list-of-component-values), then parse each entry in the returned list as a `<media-query>`. Its value is the list of `<media-query>`s so produced.
  >
  > **Note:** This definition of [<media-query-list>](https://drafts.csswg.org/mediaqueries-5/#typedef-media-query-list) parsing intentionally accepts an empty list.

https://drafts.csswg.org/mediaqueries-5/#mq-syntax

romainmenke:
But that isn't reflected in the syntax definition.
Maybe it should be?

A conditional `<media-import-condition>` in `@import` makes it possible to use that definition in linters. Otherwise such linters flag `@import 'foo.css'` as invalid.

cdoublev:
I have no opinion on whether this should be reflected in the syntax. But I prefer that a superfluous `?` not be added *if* an empty input is *also* defined as valid in prose.

`<forgiving-selector-list>` and `<media-query-list>` are the only two productions that accept an empty input, if I am not mistaken. To accomodate linter's needs while considering my opinion, an alternative could therefore be to redefine how they are parsed, and add `?` wherever it is needed. E.g. `@media {}` is valid and `@media` is currently defined with `@media <media-query-list> {...}`.

Note that `:is()` is *generally* defined to match `: <function-token> <any-value> )` before being validated/parsed against `:is(<forgiving-selector-list>)`. `<any-value>` does not accept an empty input while `<forgiving-selector-list>` does, and all browsers currently accept an empty `:is()`.

romainmenke:
I think that in this case author needs go before technical purity :)
I also think that if something can be trivially expressed in a syntax definition that we should always do that.

Other than that I don't have any preference as to how or where we fix this.

cdoublev:
I tend to agree with you because spec authors can make mistake such as `@custom-media <extension-name> [ <media-query-list> | true | false ] ;`: I do not think `@custom-media --name;` is intended to be valid, so it should be `@custom-media <extension-name> [<media-query-list> | true | false]! ;`.

I will work on an alternative PR later, if that works for you.

romainmenke:
That would be great, thank you so much 🙇

cdoublev:
I preferred to open #14431 since it seems more complicated than I thought, for the reasons explained in this issue.

romainmenke:
Thank you @cdoublev 🙇

Issue #3109, opened 2018-09-14
Do the CSSOM part of #866


frivoal:
Asides from the edits already done in https://github.com/w3c/csswg-drafts/issues/866, @tabatkins mentioned that:

> the remaining detail for how legacy shorthands work needs to show up in CSSOM, specifically in https://drafts.csswg.org/cssom/#serializing-css-values (or more precisely, probably in the definition of "preferred order").

This remains to be done.

cdoublev:
I understand that fixing this issue may not be a priority but can I assume that it can be partially fixed in [*serialize a CSS declaration block*](https://drafts.csswg.org/cssom-1/#serialize-a-css-declaration-block) (instead of *serialize a CSS value*, which seems to better seperation of concerns to ignore a legacy shorthand in a *CSS declaration block*):

```diff
- 3. If `property` maps to one or more shorthand properties, let `shorthands` be an array of those shorthand properties, in preferred order.
+ 3. If `property` maps to one or more shorthand properties, let `shorthands` be an array of those shorthand properties, in preferred order, ignoring legacy shorthands.
```

Also, there are some cases discussed in #8014 that I would like to be clarified.

Assuming an hypothetical legacy shorthand `shorthand` with syntax `a | b` and its target `longhand` with syntax `c | d`, a birectionnal mapping from `b` to `c`, I would expect the following serializations:

```js
style.shorthand = 'a'
style.shorthand; // 'a'
style.longhand;  // ''
style.cssText;   // 'shorthand: a;'

style.shorthand = 'b'
style.shorthand; // 'b'
style.longhand;  // 'c'
style.cssText;   // 'longhand: c;'

style.shorthand = 'c'
style.shorthand; // ''
style.longhand;  // ''
style.cssText;   // ''

style['--custom'] = 'a'
style.shorthand = 'var(--custom)'
style.shorthand; // 'var(--custom)'
style.longhand;  // 'var(--custom)'
style.cssText;   // 'shorthand: var(--custom);'
computedStyle.shorthand; // 'a'
computedStyle.longhand; // ''

style.longhand = 'c'
style.shorthand; // 'b'
style.longhand;  // 'c'
style.cssText;   // 'longhand: var(--custom);'

style.longhand = 'var(--custom)'
style.shorthand; // 'var(--custom)'
style.longhand;  // 'var(--custom)'
style.cssText;   // 'longhand: var(--custom);'
```

Loirooriol:
```js
[...]

Issue #12846, opened 2025-09-23
[css-cascade-5] [cssom] Serialization of shorthands with reset-only subproperties is under-defined


AtkinsSJ:
_Apologies if this is in the spec somewhere and I just couldn't find it (though perhaps that's a sign it could be made clearer)._

According to tested behaviour in WPT, (eg, [here](http://wpt.live/css/css-fonts/font-shorthand-serialization-prevention.html)) this causes `font` to not be serializable:

```css
.target {
	font: "something";
	font-variation-settings: "aaaa" a;
}
```

`font-variation-settings` is a reset-only subproperty of `font`: Using the `font` shorthand always sets it to its initial value. My (wrong) mental model was that reset-only subproperties would not affect the shorthand's value. The [reset-only subproperty definition](https://drafts.csswg.org/css-cascade-5/#reset-only-sub-property) and ["serialize a CSS value" algorithm](https://drafts.csswg.org/cssom-1/#serialize-a-css-value) don't address this.

I think in particular, step 1.1 of "serialize a CSS value" is ambiguous (emphasis mine):

> Let shorthand be the first shorthand property, in preferred order, that **_exactly maps to all of the longhand properties_** in list.

I would not expect a reset-only subproperty to be included in the list of longhand properties, but it seems like it actually should do?

In any case I think this could be clearer. Maybe even just specifying that reset-only subproperties are still considered longhands of the shorthand, so the usual rules apply. As someone coming across these, it's been counter-intuitive, even though I can understand now why it would behave like this.

Tangentially related to #878.

tabatkins:
This is also under the somewhat wider umbrella of #14422 now.

Loirooriol:
The `font` shorthand should continue serializing as empty string when a reset-only longhand isn't set to its initial value, regardless of #14422.

cdoublev:
#14423's description says it fixes this issue, but I do not think it actually did.

Reset-only sub-properties are explicitly defined as being *"like other sub-properties"*, with *"sub-properties"* linking to *"longhand sub-properties"*, which can be confusing, I agree.

In my opinion, CSS Cascade should move "longhand" outside of the term "longhand sub-properties", and put more clearly that:

  - a longhand property is any property that is not a shorthand property
  - a sub-property may be a longhand property or a shorthand property
  - a sub-property may be reset-only
  - a shorthand sets all its longhand sub-properties

I do not think CSSOM should be changed.

Loirooriol:
[...]

PR #14423, opened 2026-08-31
Cleanup + Fix issues for decl-block serialization

tabatkins:
DO NOT SQUASH.

Fixes issues #3109 #12846 

Review, @emilio ?

### DIFF
--- a/cssom-1/Overview.bs
+++ b/cssom-1/Overview.bs
@@ -2650,16 +2650,20 @@ To <dfn export>update style attribute for</dfn> <var>declaration block</var> mea
  <li>Unset <var>declaration block</var>'s <a for="CSSStyleDeclaration">updating flag</a>.
 </ol>
 
-The <dfn export id=concept-shorthands-preferred-order>preferred shorthand order</dfn> of a list of shorthand properties <var>shorthands</var> is as follows:
+<div algorithm>
+	The <dfn export id=concept-shorthands-preferred-order>preferred shorthand order</dfn> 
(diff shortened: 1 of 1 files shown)


cssom-1/Overview.bs
```diff
@@ -2651,7 +2657,7 @@ To <dfn export>update style attribute for</dfn> <var>declaration block</var> mea
 </ol>
 
 <div algorithm>
-	The <dfn export id=concept-shorthands-preferred-order>preferred shorthand order</dfn> 
+	The <dfn export id=concept-shorthands-preferred-order>preferred shorthand order</dfn>
```
emilio: This should probably be in the previous commit.

emilio (changes requested):
All commits but the last look good. I'm not convinced we should do #12846.

IIUC, the intention is that for something like this:

```
<div style='font: 1px "something"; font-variation-settings: "wght" 4;'></div>
```

You serialize `div.style.cssText` to `font: 1px "something"; font-variation-settings: "wght" 4;`, but that doesn't match the behavior of any browser, and it's a bit weird that `div.style.font` will _not_ serialize the `font` shorthand (which is correct, tho)?

Implementation wise, this needs a different serialization codepath for each of these shorthand where we skip/ignore the reset-only properties... Which seems rather unfortunate / error-prone. Is it worth it? Also we probably need a resolution for that?

tabatkins:
Oh, huh, in the middle of this we forgot that didn't actually have a resolution. Popped it from the stack, I'm gonna merge the rest then.

PR #14426, opened 2026-09-01
[css-color-4] Add sample pseudocode for the EdgeSeeker gamut mapping algorithm

svgeesus:
## Summary

CSS Color 4 defines three CSS gamut mapping algorithms: Binary Search with Local MINDE, EdgeSeeker, and Ray Trace. The first and third already had sample pseudocode; EdgeSeeker only had a prose description and a placeholder `Issue: add pseudocode for EdgeSeeker GMA`.

This adds pseudocode for EdgeSeeker, matching the structure/style of the existing two algorithms:

- Constructing the LUT (sampling cusp and curve points per hue slice for the destination gamut)
- Calculating the boundary curvature from the sampled points
- The per-color mapping algorithm (lookup + linear/arc chroma reduction + residual clip)
- The two helper algorithms it depends on (max chroma lookup, arc intersection)

Derived from reading the reference implementation in [color-js/apps](https://github.com/color-js/apps/tree/main/gamut-mapping/methods/edge-seeker) (already cited in the spec as `[[colorjs-EdgeSeeker]]`).

## Test plan

- [ ] Built locally with Bikeshed; spec builds cleanly with no new errors or warnings
- [ ] All four new `dfn`s resolve and cross-reference correctly (verified in local build output)
- [ ] Review from other editors/implementers on correctness of the math (esp. curvature/arc-intersection derivation) before merging

### DIFF
--- a/css-color-4/Overview.bs
+++ b/css-color-4/Overview.bs
@@ -141,7 +141,7 @@ spec:css-color-5; type:value; for:oklch();
 			"title": "EdgeSeeker Gamut Mapping algorithm",
 			"date": "2023",
 			"authors": ["Alexey Ardov"],
-			"href": "https://github.com/color-js/apps/tree/main/gamut-mapping/edge-seeker"
+			"href": "https://github.com/color-js/apps/tree/main/gamut-mapping/methods/edge-seeker"
 		},
 		"colorjs-RayTrace": {
 			"title": "Ray Trace Gamut Mapping",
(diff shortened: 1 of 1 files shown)


svgeesus:
@ardov @facelessuser @jamesnw @romainmenke @jparnzen

facelessuser:
I will try to take a look sometime soon.

jparnzen:
Likewise, I’ll try to take a look in the next day or so.

facelessuser:
One thing I'd probably note. I would probably prefer to update the spec to specifically suggest building the LUT in a linear RGB space. I know the reference implementation does so using gamma corrected P3, but using the linear P3 space would distort the initial data less, requiring less correction, I imagine.

I'm not done going through it, but I just thought I'd make that a suggestion.

Issue #14319, opened 2026-08-12
[css-forms-1] Vertically centering form control content


fantasai:
The UA stylesheet currently applies block-axis centering in a few places: options, radio buttons, checkboxes, and [buttons](https://github.com/w3c/csswg-drafts/issues/14190). But actually, when an input or select is sized larger than the text and icons it contains, its contents should also be centered.

We should apply centering systematically to all single-line form controls:

```
input, select, option {
  align-content: center;
  align-items: center;
}
```

I think the main question I have is how much of this needs to be restricted to `appearance: base`.

emilio:
I think the block axis at least is covered (at least the `select` / `input` / `button` stuff) by the resolution in https://github.com/w3c/csswg-drafts/issues/14190 and relevant HTML spec PRs, right?

I'm not convinced by `align-items` at least? It seems it would do nothing on most inputs, and on most `button`s by default since they use block layout?

fantasai:
Well, `select` is a single-line flexbox in `appearance: base`, and likely a number of `<input>` types will be, too. So `align-content: unsafe center` won't be enough.

fantasai:
Also thinking about this and #14190 a bit more -- buttons should probably have `safe center`, since they can accept block-level content. But input and select might be better to use just `center`. They're clipped, you can't scroll to see their contents, and they're a single line of text. The best we can do for showing information is to vertically center: fully showing ascenders is not as important as showing the middle of the text.

flackr:
@fantasai one thing I don't fully understand, but didn't want to run the clock during the discussion, is how we're enforcing that all content is a single line. While showing the top of large fonts is a risk, I'm worried that if developers do end up with multi-line content unsafe alignment is going to be far worse since typically the start contains the most important context:

https://codepen.io/editor/flackr/pen/01a03ee4-0e35-7edf-a92e-5e34d33740db

I'm also not sure how text-ellipsis is meant to work with unsafe alignment, does it elide both sides?

fantasai:
> how we're enforcing that all content is a single line

We're not. But that's the vast majority of how content renders for these elements, it's super unusual to do anything multiline. Button yes, input/select, not so much. And `option` doesn't get clipped by default.

> since typically the start contains the most important context

Not always :)

> I'm also not sure how text-ellipsis is meant to work with unsafe alignment, does it elide both sides?

This is only aligning in the block axis. The inline axis continues to be start-aligned.

PR #14439, opened 2026-09-03
[css-forms-1] Add align-content and align-items to input styles

lukewarlow:
[css-forms-1] Add align-content and align-items to input styles

Per #14319

This edit only applies it in base appearance if it ends up happening in the HTML spec for appearance auto then when the time comes to upstream we can just account for that.

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -850,6 +850,11 @@ fieldset {
   border: 1px solid currentColor;
   background-color: transparent;
 }
+
+input {
+	align-content: unsafe center;
+	align-items: unsafe center;
+}
 ```
 
 ## Sliders ## {#stylesheet-sliders}


fantasai (approved):
I feel like it'd be helpful to have the `select` styles copied into this stylesheet so we can see them all together and get things edited in quickly before upstreaming the changes. (HTML has the styles split up into so many sections, it's hard to make sense of what's happening there.) That would make it easier to see how `select` and `input` correspond also. But this seems fine for `input` on its own.

lukewarlow:
> I feel like it'd be helpful to have the select styles copied into this stylesheet so we can see them all together and get things edited in quickly before upstreaming the changes.

Fwiw it was removed somewhat at Anne's request because changes were being made here but not making it to the HTML spec. I agree it would be good to have a central stylesheet with select styles to adjust with each resolution but perhaps a gist or a github comment in the meta issue that's edited over time would be better.

Issue #14212, opened 2026-07-21
[css-mixins-1] Is `!important` valid in functions and/or mixins?


kbabbitt:
https://drafts.csswg.org/css-mixins-1/#function-rule specifies that:
- the default value of a parameter is `<declaration-value>` which does not accept `!important`
- the function body is `<declaration-rule-list>` which *could* accept `!important` on declarations

`!important` doesn't seem particularly useful to me in the context of a function or mixin local or parameter, but I wanted to raise the issue to clarify.
Blink's implementation of functions currently drops declarations with it: https://codepen.io/kevinbcmu/pen/bNgjLxM

tabatkins:
In parameters (function or mixin), no, invalid. They're not declarations.

In mixin body, yes, it should be valid. The body of a mixin is just a style rule.

In function body, it doesn't seem necessary to support; the locals don't have anything else to cascade against except each other. I don't have an opinion on whether it's valid or invalid to have `!important`, and am happy to match Blink's behavior of making it invalid.

kbabbitt:
> In mixin body, yes, it should be valid. The body of a mixin is just a style rule.

Oh right, that makes sense for mixin results. Not sure it does for mixin locals, or `@locals` block contents if we end up adopting that - they also don't have anything to cascade against except each other, right?

Though now I find myself wondering if we should just accept `!important` and drop it, so that a declaration cut-and-pasted from elsewhere still works.

thejohnjansen:
Taking a look at making edits for this one. Assigning to myself.

cdoublev:
[CSS Syntax](https://drafts.csswg.org/css-syntax-3/#typedef-declaration-value) already makes `!important` invalid in `<declaration-value>`.

  > The `<declaration-value>` production matches any sequence of one or more tokens, so long as the sequence does not contain `<bad-string-token>`, [...], or **`<delim-token>` tokens with a value of "!"**.

Consequently, https://github.com/w3c/csswg-drafts/commit/af8ba4aefb7e06d4d3f52aedca24a6c9a76d536b was unnecessary (and even undesirable, imo).

PR #14469, opened 2026-09-09
!Important is invalid and cause declaration to be invalid in function bodies and mixin arguments and private blocks

thejohnjansen:
[css-mixins-1] the CSSWG discussed whether !important is valid in functions and/or mixins. Issue #14212 was RESOLVED: in function bodies, function and mix in arguments, and all private blocks !important is invalid and causes declaration to be invalid.

### DIFF
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -1787,7 +1787,7 @@ In an ''@private'' rule, a [=custom property=] declaration
 with an ''!important'' annotation is invalid and ignored.
 
 <wpt>
-	mixins/mixin-private-important.html
+	mixins/important.html
 </wpt>
 
 When a [=style rule=] contains any ''@private'' rules
--- a/css-mixins-1/Overview.bs
+++ b/css-mixins-1/Overview.bs
@@ -233,6 +233,13 @@ then the [=default value=] must [=CSS/parse=] successfully
 according to that [=parameter type=]'s syntax.
 Otherwise, the ''@function'' rule is invalid.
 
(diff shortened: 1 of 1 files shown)


css-mixins-1/Overview.bs
```diff
@@ -233,6 +233,13 @@ then the [=default value=] must [=CSS/parse=] successfully
 according to that [=parameter type=]'s syntax.
 Otherwise, the ''@function'' rule is invalid.
 
+A <<declaration-value>> 
+cannot contain a top-level ''!important'' annotation. 
+This restriction applies to default values and to arguments 
+in <<dashed-function>>s, 
+including those used by ''@apply'' rules. 
+Mixin parameters use the same grammar as function parameters.
```
kbabbitt, line 241: This is repeating what's already specified by the grammar, so I don't think it's needed, at least not as normative text.

We might need normative text specifying what happens if a function parameter list contains malformed content, one example of that being `!important`. Or maybe the existing parsing algorithms in Syntax are sufficient to cover that case, not sure. @tabatkins wdyt?

css-mixins-1/Overview.bs
kbabbitt, line 2559: I like to add an entry to the Changes section at the same time I make substantive edits. It helps keep it up to date and makes things easier when it comes time to republish.

css-mixins-1/Overview.bs
thejohnjansen: Removed that text, but leaving this issue "Open" incase @tabatkins wants to respond.

Issue #14308, opened 2026-08-11
[css-forms-1] Apply border-radius to all base appearance controls


fantasai:
A survey of design systems shows that a vast majority round the corners of their input controls to make them not-sharp: usually by 4px, sometimes by 6px. We propose adopting this general consensus by adding `border-radius: 4px` to the base appearance for `input`, `textarea`, `select` and `::picker`.

```
input, textarea, select, ::picker(...) {
  border-radius: 4px;
}
```

There is also the question of what to do for buttons, if they should change to match this radius or continue to have a different, more pronounced radius. But there was more variation on this point, so that can be a separate discussion if we want.

emilio:
I strongly think that `select` should have the same radius as buttons. They're pretty much the same. A single `<select>` is a button that triggers a dropdown on click, sorta-kinda :)

fantasai:
@emilio The complication with buttons is that (unlike `<select>`, which has a ::picker-icon), we need some way of distinguishing buttons from regular input fields. There are ways to do that besides the border-radius, e.g. background-color, but we'd have to pick something, we can't just make them all match and call it done.

emilio:
Yeah, fwiw I would expect the input field's background to be different from select / buttons, but fair point.

stubbornella:
In a design system survey I did recently, 25 of 33 design systems match the button, input, and select `border-radius`. So your instincts are spot on @Emilio. Most systems choose to make inputs, selects, and buttons all match one of these two values: 

- `4px` - Most systems, used by government and enterprise design systems (Fluent, Spectrum, USWDS, most governments)
- `6px` - Most usage, used by popular and growing design system libraries (Bootstrap, Tailwind, Ant, shadcn, Primer)

Either of those values would be a totally valid choice for buttons, inputs, and selects. Weirdly, no design systems chose to use 5px, so we should not take the average of the two. Personally, I'd weigh usage more and go with 6px. It has the added benefit of making the controls look more interactive.

Far fewer design systems choose pill, square corners, or to round the button more than the other controls. I have more data I can share tomorrow if folks are interested.

emilio:
Thanks for the data @stubbornella!

I'd lean a bit more towards `4px`, because trends change (maybe in a few years we don't want everything so rounded) and 4px is less likely to crop something important... But fine with either really.

Issue #11939, opened 2025-03-14
[css-forms-1] Pseudo-class to match textual inputs


lukewarlow:
Currently for styling of inputs you end up having to write fairly complex selectors (that aren't always correct), or you write a blanket input style and then have to override and undo some of it for specific inputs.

E.g.

```css
input:not([type]),
input:is([type=text], [type=search], [type=email], [type=password], [type=tel], [type=url], [type=number]) {
```
This is annoying to write but also not always correct. If a browser doesn't support a specific input type (e.g. input type week is currently limited in browser support) it falls back to a textual input.

It would probably be useful to have a pseudo class that matches for textual inputs?

lukewarlow:
Possibly also useful would be a button like matcher for input type button, reset, submit, and color, along with buttons themselves? But that's an easier selector to do manually and doesn't have the issue that it's not always a static list.

fstrr:
Agreed that some sort of `:accepts-text` pseudo selector would be useful. It would be excellent to be able to do `:accepts-text:disabled`, `:accepts-text:read-write`, etc.

lukewarlow:
Turns out Chromium actually has a `:-internal-text-field` already which matches exactly the selector I provided above.

lukewarlow:
See also https://github.com/w3c/csswg-drafts/issues/2296 for a discussion that seems related (haven't read through it but I don't think the WICG thing ever happened)

PR #14473, opened 2026-09-10
[css-forms-1] Apply 4px border-radius to all base appearance controls

lukewarlow:
[css-forms-1] Apply 4px border-radius to all base appearance controls

Per #14308

### DIFF
--- a/css-forms-1/Overview.bs
+++ b/css-forms-1/Overview.bs
@@ -991,9 +991,7 @@ input:is([type="color"], [type="button"], [type="reset"], [type="submit"]) {
   padding-block: 0.25em;
   padding-inline: 0.5em;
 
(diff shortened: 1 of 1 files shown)


css-forms-1/Overview.bs
```diff
@@ -1066,7 +1059,16 @@ See base appearance user agent styles specified in
...
 
 ```css
-input::field-content {
+textarea,
+input {
+	border-radius: 4px;
+}
+```
+
+The above must apply to textareas and textual inputs only.
```
lukewarlow, line 1073: This would be better if we had https://github.com/w3c/csswg-drafts/issues/11939

nt1m (approved):

PR #14488, opened 2026-09-14
[css-gaps-1] Define rule visibility for grid lanes #14489

jav099:
This PR defines `rule-visibility-items` behavior for grid lanes containers.

The property applies only to grid-axis decorations, and treats the "lanes"/tracks as items, similar to how multicol behaves with this property.

 It has no effect on stacking-axis decorations.

 Addresses #14488


kbabbitt:
I suppose this comes up when there are one or more empty tracks in a grid-lanes layout? Seems fine to me, but I think we should get a resolution on it.

Issue #14461, opened 2026-09-08
[css-forms-1] Lighten form control backgrounds slightly in dark mode


fantasai:
Splitting out from https://github.com/w3c/csswg-drafts/issues/14254 (see also https://github.com/w3c/csswg-drafts/issues/14252) ...

The WebKit customizable select demo styles adopted a slight lightening of the background color in dark modes, to make form controls feel more interactive.

```
select, input, textarea {
	background: light-dark(transparent, alpha(white / 4%));
}
```

We're proposing to adopt this into the default stylesheet.

stubbornella:
This isn't decorative, it is about communicating importance to the user.

Humans have evolved to pay more attention to stuff that is nearby and less to stuff that is farther away. Think speeding car a mile away versus a speeding car entering the sidewalk you are crossing. 

This applies to web design because we need to make the most important things look closer to the viewer. In light mode, you achieve this by casting a drop shadow from the picker and using a thin border around the buttons, inputs, and selects. That won't have the same effect in dark mode, because cast shadows and subtle borders are close to invisible. Instead, to communicate importance in dark mode, you need to lighten the background color. Lighter stuff appears closer to the user, darker panels recede. 

To achieve this, we need to take 2 steps:
1. Slightly lighten all form control backgrounds
2. More significantly lighten backgrounds in the top layer, like pickers, because they are the _most important_ thing on the page **while they are open**.

In this image you can see the effect, the inputs are lighter and the picker is even more so.

<img width="494" height="374" alt="ligher controls and picker" src="https://github.com/user-attachments/assets/dabe7015-6a41-496e-ac6e-b1f5450fc415" />

kbabbitt:
What happens in Forced Colors Mode? Naively, I'd expect this slight lightening to be overridden by the forced background color. If that's the case, would we need to communicate importance in some other way?

lukewarlow:
I think forced colors mode is its own thing, where colour obviously works quite differently. I think it does need discussing though. I would assume in forced colors mode the controls would actually get an entirely different background color, (presumably Field and then color would be FieldText?). That being said I don't know much about the current implementation.

lukewarlow:
> background: light-dark(transparent, color-mix(in srgb, CanvasText 4%, transparent));

Did we get any progress on `currentBackgroundColor` because I would have assumed we'd want to mix using that. Basing the mix off CanvasText feels fragile.

fantasai:
[...]

PR #14501, opened 2026-09-16
[css-forms-1] Lighten background colors in dark mode

lukewarlow:
See https://github.com/w3c/csswg-drafts/issues/14461


css-forms-1/Overview.bs
```diff
@@ -845,10 +845,13 @@ input:not([type=file], [type=range], [type=checkbox][switch]),
 textarea,
 button,
 ::file-selector-button,
-::slider-track,
+::slider-track {
+  border: 1px solid currentColor;
+  background-color: light-dark(transparent, alpha(from white / 4%));
+}
+
 fieldset {
```
lukewarlow, line 853: I think its right to exclude fieldset for now? We need to discuss its styles further.

nt1m: Yeah agreed.

nt1m (approved):

